Issue #11331, opened 2018-06-19
gluon bug: AttributeError: '_thread._local' object has no attribute 'value'


feevos:
## Description
Dear all, 

I am trying to run mxnet in a distributed HPC environment for embarrassingly parallel (distributed) runs.   
The goal is to use this for bayesian hyperparameter optimization, therefore all communication between nodes is nothing mxnet/gpu specific (lists of hyperparams, like learning rate, batch size etc). For my distributed needs I chose [ray](http://ray.readthedocs.io/en/latest/). Each node has 4 gpus and runs a completely independent run from other nodes. However, I cannot even define a simple gluon layer within 
a ```ray.remote``` function. 

When I am using 2 (or more) nodes with this trivial example, everything is working: 

```Python
import os
import sys
import ray
import time

# mxnet gpu examples 
import mxnet as mx
from mxnet import nd
import numpy as np


@ray.remote(num_gpus = 4)
def f():

    gpus = [int(x) for x in os.environ["CUDA_VISIBLE_DEVICES"].split(',')] # In case of multiple GPUs, comment out 2nd option. 
    tctx = [mx.gpu(i) for i in range(len(gpus))]
    a = nd.random.uniform(shape=[3,4,16,16],ctx=tctx[0])


    return a.asnumpy()

if __name__ == '__main__':
    ray.init( redis_address =  sys.argv[1])
    result1 = ray.get(f.remote())
    result2 = ray.get(f.remote())

    print (result1,result2)
```

However, when I try to use any gluon object that derives from HybridBlock, for example: 

```Python
@ray.remote(num_gpus=4)
def f(x):
    loss = gluon.loss.L2Loss()
    return x
```
I get an error. I've also tested ray with a simple pytorch nn (everything is working), so this is most probably a mxnet/gluon  problem.  

**edit**: The same problem and error message appears if I use [dask.distributed](http://distributed.readthedocs.io/en/latest/) for launching/managing the cluster. 

## Environment info (Required)
All nodes are identical, I've run diagnose.py command on an interactive node with 4 gpus allocated 

```Python

----------Python Info----------
Version      : 3.6.4
Compiler     : GCC 7.2.0
Build        : ('default', 'Jan 16 2018 18:10:19')
Arch         : ('64bit', '')
------------Pip Info-----------
Version      : 9.0.1
Directory    : /home/dia021/Software/anaconda3/lib/python3.6/site-packages/pip
----------MXNet Info-----------
[...]

PR #11332, opened 2018-06-19
[MXNET-558] Fix 'AttributeError: '_thread._local' object has no attribute 'value'' on distributed processing applications

ThomasDelteil:
## Description ##
Using distributed processing frameworks, django or ray, users encountered errors when trying to run mxnet in separate threads.

https://github.com/apache/incubator-mxnet/issues/11331 and https://github.com/dmlc/gluon-cv/issues/156

Calling the `_current.value.get` within the context of the respective object solved the issue.

Does anybody have a suggestion as to how to test this without introducing a dependency on ray?

### Essentials ###

### DIFF
--- a/python/mxnet/symbol/symbol.py
+++ b/python/mxnet/symbol/symbol.py
@@ -2451,8 +2451,9 @@ def var(name, attr=None, shape=None, lr_mult=None, wd_mult=None, dtype=None,
     handle = SymbolHandle()
     check_call(_LIB.MXSymbolCreateVariable(c_str(name), ctypes.byref(handle)))
     ret = Symbol(handle)
-    with AttrScope():
-        attr = AttrScope._current.value.get(attr)
+    if not hasattr(AttrScope._current, "value"):
+        AttrScope._current.value = AttrScope()
+    attr = AttrScope._current.value.get(attr)
     attr = {} if attr is None else attr
     if shape is not None:
         attr['__shape__'] = str(shape)
--- a/python/mxnet/symbol/symbol.py
+++ b/python/mxnet/symbol/symbol.py
@@ -2451,7 +2451,8 @@ def var(name, attr=None, shape=None, lr_mult=None, wd_mult=None, dtype=None,
     handle = SymbolHandle()
     check_call(_LIB.MXSymbolCreateVariable(c_str(name), ctypes.byref(handle)))
     ret = Symbol(handle)
-    attr = AttrScope._current.value.get(attr)
+    with AttrScope():
+        attr = AttrScope._current.value.get(attr)
     attr = {} if attr is None else attr
     if shape is not None:
         attr['__shape__'] = str(shape)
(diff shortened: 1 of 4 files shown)


robertnishihara:
I haven't reproduced the original problem, but what about defining the function and executing it in a separate thread or process (just by using the builtin `threading` or `subprocess` modules). And in the multiprocessing case, serializing the function using cloudpickle. Would something like that succeed at triggering the issue?

ThomasDelteil:
```
from threading import Thread
import mxnet as mx
from mxnet import gluon

def threaded_function():
    net = gluon.nn.Dense(2)
    
thread = Thread(target = threaded_function)
thread.start()
thread.join()
```
reproduce the issue, and is fixed with this patch

feevos:
Just a question, with this public release will one need still to provide a prefix as a solution or this is taken care of automatically? Thank you very much for this!!

ThomasDelteil:
@anirudh2290 it looks like you refactored a lot of this scope code last month, can you review this? 
@feevos when the fix is in you'll be able to use it using the `pip install mxnet-cu9x --pre` nightly build

python/mxnet/symbol/symbol.py
```diff
@@ -2451,7 +2451,8 @@ def var(name, attr=None, shape=None, lr_mult=None, wd_mult=None, dtype=None,
     handle = SymbolHandle()
     check_call(_LIB.MXSymbolCreateVariable(c_str(name), ctypes.byref(handle)))
     ret = Symbol(handle)
-    attr = AttrScope._current.value.get(attr)
+    with AttrScope():
```
anirudh2290: we can include above `not hasattr` logic here too.

tests/python/unittest/test_thread_local.py
```diff
@@ -133,6 +133,19 @@ def f():
...
+    def f():
+        net = mx.gluon.nn.Dense(2)
+        net.initialize()
+        net(mx.nd.array([1, 2, 3]))
+        status[0] = True
+
+    thread = threading.Thread(target=f)
+    thread.start()
+    thread.join()
+    assert status[0], "Failed to create a layer within a thread"
```
anirudh2290: Can you also add the test for running the following functions inside a thread:

```
  def g():
      data = mx.sym.Variable('data', attr={'a': 'b'})
```

```
  def f():
      a = mx.sym.var("a")
      b = mx.sym.var("b")
      a_ = mx.nd.ones((2, 2))
      c_ = a_.copy()
      func1 = (a + b).bind(mx.cpu(), args={'a': a_, 'b': c_})
      func1.forward()[0].wait_to_read()
```

python/mxnet/symbol/symbol.py
ThomasDelteil: That logic is already included in the context

tests/python/unittest/test_thread_local.py
ThomasDelteil: Will do :+1:

python/mxnet/symbol/symbol.py
anirudh2290: we can avoid a dict copy this way.

ThomasDelteil: fair point, I just thought it was cleaner that way and had better separation of concerns. I will update.

anirudh2290:
@piiswrong can you also take a look...

Issue #11241, opened 2018-06-12
Conv1D throws CUDNN_STATUS_EXECUTION_FAILED


eric-haibin-lin:
Setup: 
```
$ pip install mxnet-cu90==1.1.0

$ nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2017 NVIDIA Corporation
Built on Fri_Sep__1_21:08:03_CDT_2017
Cuda compilation tools, release 9.0, V9.0.176

$ ls /usr/local/cuda/lib64/libcudnn.so.7.0.3
/usr/local/cuda/lib64/libcudnn.so.7.0.3
```

Run the following script `debug.py`:
```
import mxnet as mx
W_REQ = 'add'
shape = (1, 65536, 1)
ctx = mx.gpu()
kwargs = {'no_bias': True, 'kernel': (1,), 'num_filter': 1}
x = mx.sym.var('x')
w = mx.sym.var('w')
x_grad = mx.nd.zeros(shape, ctx=ctx)
w_grad = mx.nd.zeros(shape, ctx=ctx)
args_grad = {'x': x_grad, 'w': w_grad}
sym = mx.sym.Convolution(x, w, **kwargs)
executor = sym.bind(ctx, grad_req={'x': 'null', 'w': W_REQ},
                    args={'x': mx.nd.ones(shape, ctx=ctx), 'w': mx.nd.ones(shape, ctx=ctx)},
                    args_grad=args_grad)
executor.forward()
executor.backward([mx.nd.ones((1,1,1), ctx=ctx)])
mx.nd.waitall()
```
Gives the following error:
```
[06:31:41] src/operator/nn/./cudnn/./cudnn_algoreg-inl.h:107: Running performance tests to find the best convolution algorithm, this can take a while... (setting env variable MXNET_CUDNN_AUTOTUNE_DEFAULT to 0 to disable)
terminate called after throwing an instance of 'dmlc::Error'
  what():  [06:31:41] src/engine/./threaded_engine.h:359: [06:31:41] src/operator/nn/./cudnn/cudnn_convolution-inl.h:242: Check failed: e == CUDNN_STATUS_SUCCESS (8 vs. 0) cuDNN: CUDNN_STATUS_EXECUTION_FAILED

Stack trace returned 10 entries:
[bt] (0) /usr/local/lib/python2.7/dist-packages/mxnet/libmxnet.so(+0x2a9e78) [0x7f705d9a3e78]
[...]

PR #11338, opened 2018-06-19
[MXNET-11241] Avoid use of troublesome cudnnFind() results when grad_req='add'

DickJC123:
## Description ##
The problem of issue #11241 arises because cudnnFind() measures convolution algorithm runtimes with an assumed "output blending parameter" beta of 0.  However, algorithms may have specialized kernels for the beta==0 case, different and faster than the generalized beta kernels.  Should the generalized kernels have issues with the problem size different than the beta==0 kernels, then the algos returned by cudnnFind() might fail when invoked with beta==1 (as it is when the convolution op grad_req='add' argument is present).

The demonstrated problem area involves a large 'c' value of 64K, where for the backprop-to-filter kernel only algo 1 handles the beta==1 case.  CudnnFind() was shown to occasionally return algos 0 or 3 as fastest, and both of these return error 8 "execution failed" when run.

The fix is based on the observation that cudnnGet() returns algo 1 for the backprop-to-filter kernel for the troublesome problem sizes.  Thus,  the fix is to avoid cudnnFind() when grad_req='add' and force use of cudnnGet() instead.  The fix maintains the effectiveness of the caching of algo lookups and convolution op instances, so neither cudnnFind() nor cudnnGet() is called repeatedly.  Deconvolution was similarly updated with this fix.

### Essentials ###
- [X ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [X ] Changes are complete (i.e. I finished coding on this PR) [1st commit includes test, 2nd the fix]

### Changes ###

## Comments ##

### DIFF
--- a/src/operator/nn/convolution.cu
+++ b/src/operator/nn/convolution.cu
@@ -41,7 +41,8 @@ static CuDNNConvolutionOp<DType>& GetCuDNNConvOp(const ConvolutionParam& param,
                                                  int backward_compute_type,
                                                  const std::vector<TShape>& in_shape,
                                                  const std::vector<TShape>& out_shape,
-                                                 const RunContext& rctx) {
+                                                 const RunContext& rctx,
+                                                 bool add_to_weight) {
 #if DMLC_CXX11_THREAD_LOCAL
   static thread_local std::unordered_map<ConvSignature,
                                          std::shared_ptr<CuDNNConvolutionOp<DType> >,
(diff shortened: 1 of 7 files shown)


DickJC123:
My second commit did not pass because the CI exposed a flaw in the logic of the fix: while cudnnGet() returned the required algo 1 on Volta, it did not for the Maxwell GPUs that were part of the CI cluster.  A third commit narrows the scope of the fix to the observed failure scenarios (grad_req='add' and c>=64K).  The fix is to force the problematic backprop-to-filter kernel to use algo 1 in this case.  The CI failed on this commit as well, but for seemingly unrelated reasons.  Will try an empty commit.

eric-haibin-lin:
do you mind rebase on mxnet master and see if CI works?

Issue #11016, opened 2018-05-21
[Feature request] temperature parameter in Softmax and SoftmaxOutput


slitsey:
MXNet does not appear to have a native temperature parameter in its softmax functions. I would like this to be added, as it has many useful applications when learning a categorical probability distribution, especially in a reinforcement learning setting. Should default to 1 to reproduce current behavior.

https://en.wikipedia.org/wiki/Softmax_function#Reinforcement_learning

@eric-haibin-lin

srochel:
@slitsey Would you be able to provide an example we can use to validate the requested feature?

slitsey:
@srochel @eric-haibin-lin I'm expecting behavior similar to the short program below:
```
import numpy as np

def softmax(x, temperature=1):
    return np.exp(x/temperature)/sum(np.exp(x/temperature))

x = np.array([ 1,  2,  3])

print(softmax(x))
print(softmax(x, temperature=10))
print(softmax(x, temperature=0.1))
```
which returns
```
[ 0.09003057  0.24472847  0.66524096]
[ 0.30060961  0.33222499  0.3671654 ]
[  2.06106005e-09   4.53978686e-05   9.99954600e-01]
```
This allows interpolation between a uniform distribution in the high-temperature limit and a greedy distribution with all probability mass on the most probable index in the zero-temperature limit. Starting with high temperature and decreasing throughout learning allows transitioning from exploration-heavy to exploitation-heavy policies in RL, if the softmax represents the policy's action distribution. Again, there are potential numerical stability issues with this due to the exponentials. But probably something like
```
import mxnet as mx

data = mx.sym.Variable('data')
net = mx.sym.softmax(data=data, temperature=10)

x = mx.nd.array([ 1,  2,  3])

ex = net.bind(mx.cpu(), args={'data': x, 'softmax2_label': 'softmax2'})
ex.forward()
```
should return
```
[
[ 0.30060961  0.33222499  0.3671654 ]
 <NDArray 3 @cpu(0)>]
```
for example. And of course this behavior should generalize as expected across axes, etc., and also be incorporated into mx.sym.SoftmaxOutput (and anywhere else softmax/Boltzmann distributions might arise in MXNet).

apeforest:
Hi @slitsey I will start working on this Feature Request. There are multiple operators related to softmax. I will first implement the operators mx.sym.softmax and mx.sym.SoftmaxOutput as suggested in your example. Please advise otherwise. Thanks!

[...]

PR #11356, opened 2018-06-21
[MXNET-560][WIP] Add temperature parameter in Softmax and SoftmaxOutput operator

apeforest:
## Description ##
This PR is to address the request to have a native temperature parameter in the softmax functions. See [Issue](https://github.com/apache/incubator-mxnet/issues/11016) for more detailed discussion.

I have added the temperature parameter to both softmax and SoftmaxOutput operators. By default the temperature parameter value is 1.0 and both functions should behave the same as not specifying the temperature parameter.

Verified the change using the following code in Python:

```
import mxnet as mx

data = mx.sym.Variable('data')
net = mx.sym.softmax(data=data, temperature=10)

x = mx.nd.array([ 1,  2,  3])

ex = net.bind(mx.cpu(), args={'data': x, 'softmax2_label': 'softmax2'})
ex.forward()
```
should expect return
```
[
[ 0.30060961  0.33222499  0.3671654 ]
 <NDArray 3 @cpu(0)>]
```

### Essentials ###

### Changes ###

## Comments ##
- In the implementation of softmax function, I added a check for the default case where temperature equals 1.0 to remove the unnecessary divide-by-1 operation in the calculation. I have run an offline experiment of softmax function alone compiled with g++ -O3. The single core runtime can be reduced by around 25% with this optimization .


apeforest:
[WIP] Only added temperature parameter to softmax but not SoftmaxOutput yet since the latter depends on operator change in another repo (MShadow). Please review the change for softmax function. Thanks

eric-haibin-lin:
@slitsey

apeforest:
@eric-haibin-lin Will do once I send out the PR for merge. Since this is my first contribution, I would like to get some initial review on my code change to make sure I am following the proper styles and convention in this community.

eric-haibin-lin:
Yeah because my initial review always involves inspecting tests :)

src/operator/nn/softmax-inl.h
```diff
@@ -145,23 +155,36 @@ __global__ void softmax_compute_kernel(DType *in, DType *out, index_t M, int axi
...
+
   __syncthreads();
   cuda::Reduce1D<red::sum, x_bits>(smem);
   __syncthreads();
   DType ssum = smem[0];
   __syncthreads();
 
-  for (index_t i = x; i < M; i += x_size) {
-    out[base + i*sa] = OP::Map(in[base + i*sa] - smax, ssum);
+  if (temperature == 1.0) {
```
rahul003: I'd suggest removing the if condition and merging these two cases, since it's perfectly okay to use the second equation for temperature=1. The if condition within this kernel can cause branching on CPU and affect performance

src/operator/nn/softmax-inl.h
```diff
@@ -145,23 +155,36 @@ __global__ void softmax_compute_kernel(DType *in, DType *out, index_t M, int axi
   __syncthreads();
 
   red::sum::SetInitValue(smem[x]);
-  for (index_t i = x; i < M; i += x_size) {
-    red::sum::Reduce(smem[x], static_cast<DType>(expf(in[base + i*sa] - smax)));
+  if (temperature == 1.0) {
```
rahul003: Same here

rahul003: Please do that everywhere possible

src/operator/nn/softmax-inl.h
```diff
@@ -127,7 +137,7 @@ inline void SoftmaxGrad(Stream<cpu> *s, DType *out, DType *ograd,
 #ifdef __CUDACC__
 template<int x_bits, typename OP, typename DType, int ndim>
 __global__ void softmax_compute_kernel(DType *in, DType *out, index_t M, int axis,
-                                       Shape<ndim> sshape, Shape<ndim> stride) {
+                                       Shape<ndim> sshape, Shape<ndim> stride, float temperature) {
```
rahul003: Please try to use const for variables which don't change

src/operator/nn/softmax-inl.h
apeforest: I added this if condition for performance concern. I'd assume (correct me if I'm wrong) 90% of cases the temperature to softmax is set to 1. Adding a dividep-by-1.0 operation to expf((in[base + i*sa] - smax)/t) will slow down this computation and I am not aware any compiler can optimize away this.

src/operator/nn/softmax-inl.h
apeforest: This is pass-by-value so it does not make any difference.

src/operator/nn/softmax-inl.h
rahul003: The division (even if not optimized away) should be better than causing a branch. Especially on the GPU branch divergence can have a significant impact on performance. It's better to avoid it.

src/operator/nn/softmax-inl.h
rahul003: Just a matter of style https://google.github.io/styleguide/cppguide.html#Use_of_const

apeforest: Thanks for your suggestion. But I do not seem to find the convention of using const for pass-by-value parameter in the Google style guide: 
```
If a function guarantees that it will not modify an argument passed by reference or by pointer, the corresponding function parameter should be a reference-to-const (const T&) or pointer-to-const (const T*), respectively.
```
In fact, adding unnecessary const declaration will put restriction on the caller of the library function.

src/operator/nn/softmax-inl.h
apeforest: I agree with you about the overhead of the branching. However, there is a trade off in performance over complexity here. It really depends on how critical this piece of computation is to the overall performance of the network. In this case, I would assume softmax is a function called very often in the last layer of a neural network. But I am not an expert to oversee the performance impact.

rahul003: If you've tested or benchmarked it, then okay. Just noticed the comment in your post.

haojin2: @apeforest Could you provide data on which version is faster? We have `test_speed` function: https://github.com/apache/incubator-mxnet/blob/a7952f0b3218363a9520aa606f43db94a34c55b8/python/mxnet/test_utils.py#L1133 ready for you to test the speed.

src/operator/nn/softmax-inl.h
haojin2: adding `const` qualifier could be a safe net by explicitly telling the compiler about your assumption that you're not changing the input value at all in the function.
Say if you have a function like this:
```c++
void foo(int a) {
  return a;
}
```
Here your assumption that the original input should be returned without any changes, and that is essential for correct behavior of this function.
Now if someone happens to change the function to:
```c++
void foo(int a) {
  a++;      // <- new code that happens to change the value of a, and will affect correctness
  return a;
}
```
compiler will not complain about it as you do not have a `const` qualifier here, so an extra `const` qualifier is not necessary, but it could be helpful.

apeforest: I have added const qualifier following your suggestion.

apeforest:
@eric-haibin-lin unit test added this time :)

haojin2:
@apeforest please do a rebase properly, you're including a lot of committed changes.

eric-haibin-lin:
Moved to #11466

PR #11365, opened 2018-06-22
Fix flaky test 10380

piiswrong:
Original input scale is too big, which causes deconvolution output to be too big. This in turn causes accuracy problems for low precision data type.

Scaling down input solved the problem


piiswrong:
moved to https://github.com/apache/incubator-mxnet/pull/11529

Issue #10928, opened 2018-05-14
Optimizers memory usage


jeremiedb:
There's been few comments about memory consumption in the R-package. With large models, the memory footprint can rapidly burst the available memory unless a manual gc() is added in the training loop. 

The problematic appears to be the weights update performed by the optimizers: https://github.com/apache/incubator-mxnet/blob/master/R-package/R/model.R#L221

Memory behaves as if the temporary nd.arrays were kept in memory. 

Looked at how the python optimizers were designed and took a first attempt at using the `mx.nd.sgd.update` type of operators. 
It seemed to help, but looking at https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/optimizer.py#L434, I confused at how the state of optimizers such as momentum or adam and cie is updated. I can't see where the staste is updated other than for the initialization. 

Also, tried to do a direct mutation of the weight using the out parameters of the `mx.nd.sgd.update` to point to the executor ref.arg.arrays, but it got the following message after a first update: 

```
 Error in mx.nd.sgd.update(weight = weight, out = out, grad = grad, lr = learning.rate,  : 
  ./ndarray.h:87: RCheck failed: ptr_->writable && !ptr_->moved Passing a read only NDArray to mutate function 
```
I'm still missing some pieces to figure out a proper way of fixing the memory consumtion of the optimizers. Any help would be much welcome!

@thirdwing

vdantu:
@sandeep-krishnamurthy : Could you please tag this as "question", "R"

@jeremiedb : You could also post this on "discuss.mxnet.io". You would get a wider audience.

jeremiedb:
As an additional pointer, the issue seems tied to R's NDArray being set to read only after an initial mutation. https://github.com/apache/incubator-mxnet/blob/master/R-package/src/ndarray.h#L86 @tqchen ? 

Basic comparative example between R and Python: 

R: 
```
> state <- mx.nd.array(c(1,2,3,3))
> grad <- mx.nd.array(1:4)
> weight <- mx.nd.array(1:4)
> mx.nd.sgd.mom.update(weight, grad, state, lr = 0.1, momentum = 0.5, out = weight)
[1] 1.4 2.8 4.2 5.1
> mx.nd.sgd.mom.update(weight, grad, state, lr = 0.1, momentum = 0.5, out = weight)
Error in mx.nd.sgd.mom.update(weight, grad, state, lr = 0.1, momentum = 0.5,  : 
[...]

Issue #10721, opened 2018-04-27
MxNet R, CNN, VRAM consumption explodes dramatically in dependence of number of filters


thomasmooon:
## Description
I have a toy dataset of 360 samples with 4096 data points each, leading to a tensor of shape `(4096,1,360)`. Hence, each observation has a size of ~ 4 kB. The CNN is very simple: `Conv -> flatten -> fully connected -> fully connected -> softmax`:
![image](https://user-images.githubusercontent.com/29228225/39371516-36cdebbc-4a42-11e8-900e-e2554d48a34a.png)

The VRAM consumption explodes in dependence of the number of filters: Please see the table and the related picture below. Regarding the influence of the **kernel size** and the **batch size**: These have very small influence, I've tested several combinations, but I omit these details for now.  The tables measure a setting using 2 GPUs of my environment (described in the environment setting below). As one can see the VRAM demand of each card increases, as expected, linear with the number of convolution filters. But if it exceeds 10, then the GPUs run out of their 8 GB VRAM. What the hell...? 

It is also remarkable, that in a setting with 1 GPU and 8 kernels is not possible: It exhausts the 8 GB RAM of the single Card.  But using 2 GPUs with everything else unchanged, then each GPU consumes only 0.477 GB, so 2x0.477 = 0.95 GB in total. This is far beyond of what is consumed when using only 1 Card. How can this be??

Things else tested without any effect: The argument `workspace`  in the `mx.symbol.Convolution()`-Function. I played with several values: 1, 64, 128, 512 MB. But his had absolutely none effect disregarding to any combination of varying number of filters. Here's the defintion of `workspace` :

> long (non-negative), optional, default=1024 Maximum temporary workspace allowed for convolution (MB)

**_#VRAM consumption in dependence of the number of filters, using 2 GPUs_**
n_filter | VRAM Consumption / Card
-- | --
1 | 313
2 | 339
4 | 385
8 | 477
10 | 523
11 | out of memory

![image](https://user-images.githubusercontent.com/29228225/39371294-8b0e7454-4a41-11e8-8c41-643351ae0a72.png)

[...]

PR #11374, opened 2018-06-23
[MXNET-563] Refactor R optimizers to fix memory leak

jeremiedb:
Fix R memory leakage through refactor of the optimizers. 
Given the mutatable NDArray isn't supported in R, optimizers have been ported as symbolic update, an executor being created for each weight. 

! Only optimizers update symbols have been used for now, so only SGD, rmsprop and Adam are now supported. Will need to reimplement the manual update for Adagrad and Adadelta. 

Memory is now kept at low level even for very large networks and embeddings. 
Tested on CPU and single GPU, not multiple GPUs.

### DIFF
--- a/R-package/R/optimizer.R
+++ b/R-package/R/optimizer.R
@@ -17,7 +17,7 @@ mx.opt.sgd <- function(learning.rate = 0.01,
                        momentum = 0,
                        wd = 0,
                        rescale.grad = 1,
-                       clip_gradient = Inf,
+                       clip_gradient = -1,
                        lr_scheduler = NULL) {
 
   lr <- learning.rate
--- a/R-package/R/optimizer.R
+++ b/R-package/R/optimizer.R
@@ -19,23 +19,23 @@ mx.opt.sgd <- function(learning.rate = 0.01,
                        rescale.grad = 1,
                        clip_gradient = -1,
                        lr_scheduler = NULL) {
-  
+
   lr <- learning.rate
   count <- 0
   num_update <- 0
-  
+
   sgd <- new.env()
   sgd$lr <- lr
   sgd$count <- 0
   sgd$num_update <- 0
-  
+
   create_exec <- function(index, weight_dim, ctx) {
-    
+
     if (momentum == 0) {
-      
+
       weight <- mx.symbol.Variable("weight")
       grad <- mx.symbol.Variable("grad")
-      
+
       sym <- mx.symbol.sgd_update(weight,
                                   grad,
                                   lr = lr,
@@ -44,11 +44,11 @@ mx.opt.sgd <- function(learning.rate = 0.01,
                                   clip_gradient = clip_gradient,
                                   name = "w")
     } else {
-      
+
       weight <- mx.symbol.Variable("weight")
       grad <- mx.symbol.Variable("grad")
       mom <- mx.symbol.Variable("mom")
-      
+
       sym <- mx.symbol.sgd_mom_update(weight,
                                       grad,
                                       mom,
@@ -64,7 +64,7 @@ mx.opt.sgd <- function(learning.rate = 0.01,
   }
 
   update <- function(index, exec_w, weight, grad) {
-    
+
     if (!is.null(lr_scheduler)){
       lr_scheduler(sgd) ## changing lr
       lr <- sgd$lr
@@ -78,7 +78,7 @@ mx.opt.sgd <- function(learning.rate = 0.01,
         sgd$num_update <- max(sgd$num_update, sgd[[indexKey]])
       }
     }
-    
+
     mx.exec.update.arg.arrays(exec_w, arg.arrays = list(weight = weight,grad = grad), match.name = T)
     mx.exec.forward(exec_w, is.train = F)
     return(exec_w$ref.outputs$w_output)
@@ -115,7 +115,7 @@ mx.opt.rmsprop <- function(learning.rate = 0.002,
                            rescale.grad = 1,
                            clip_gradient = -1,
                            lr_scheduler = NULL) {
-  
+
   lr <- learning.rate
   count <- 0
   num_update <- 0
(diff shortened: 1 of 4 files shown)


jeremiedb:
Provides fix for #10721 #10928

thirdwing:
@hetong007 Please take a look at this.

jeremiedb:
@anirudhacharya Sure, I'll add tests. 
Would be great if you could jump in as well. I was expecting to add the missing Adagrad and Adadelta optimizers within a week in order to match existing functionnalities as soon as possible. Would you be disposed looking at the non-mutatble NDArrays which was actually the root cause leading to refactor optimizers into symbolic execution? Thanks!

anirudhacharya:
@jeremiedb sure! also could you please mail me at [anirudhkrec at gmail]. I would like to sync with you about improvements to the R binding.

jeremiedb:
@hetong007 With Adadelta and Adagrad, the same functionnalities as now supported (and non-centered rmsprop has been added within rmsprop). Tests to be added.

R-package/R/optimizer.R
```diff
@@ -1,32 +1,70 @@
...
                        lr_scheduler = NULL) {
-  # use lr as short for learing rate.
+  
   lr <- learning.rate
-  count       <- 0
-  num_update  <- 0
-
+  count <- 0
+  num_update <- 0
+
```
hetong007: please eliminate trailing whitespaces.

R-package/R/optimizer.R
```diff
@@ -40,105 +78,116 @@ mx.opt.sgd <- function(learning.rate,
...
 
 #' Create an RMSProp optimizer with respective parameters.
 #' Reference: Tieleman T, Hinton G. Lecture 6.5-rmsprop: Divide the gradient by a running average of its recent magnitude[J]. COURSERA: Neural Networks for Machine Learning, 2012, 4(2).
 #' The code follows: http://arxiv.org/pdf/1308.0850v5.pdf Eq(38) - Eq(45) by Alex Graves, 2013.
-#' 
-#' @param learning.rate float, default=0.002
-#'      Step size.
-#' @param gamma1 float, default=0.95
+#'
+#' @param learning.rate float, default=1e-3
```
hetong007: Is there a strong reason to change the default values? It may break other people's code.

R-package/R/optimizer.R
```diff
@@ -392,31 +459,22 @@ mx.opt.create <- function(name, ...) {
...
 #' @export
-mx.opt.get.updater <- function(optimizer, weights) {
-  # This is the list to keep track of internal states of optimzer
-  state.list <- lapply(seq_along(weights), function(i) {
-    if (is.null(weights[[i]])) return(NULL)
-    optimizer$create.state(i, weights[[i]])
+mx.opt.get.updater <- function(optimizer, weights, ctx) {
+  
+  exec_list <- lapply(seq_along(weights), function(i) {
+    if (is.null(weights[[i]])) return(NULL) else
```
hetong007: please format here and below as

```
if (condition) {
   xxx
} else {
  yyy
}
```

R-package/R/optimizer.R
jeremiedb: I wanted to align with the Python's package default. I'll revert to existing default if you see if you see more harms from this change.

hetong007: I suggest to keep the default value as in R package. It's not necessary to set default values the same across interfaces, while it may break user's script if we change it, especially for parameters in optimizer.

hetong007:
You may use the help from your editor to eliminate trailing spaces.  Also, please add tests for new optimizers.

jeremiedb:
@hetong007 Tests added and trailing space fixed.

R-package/R/optimizer.R
```diff
@@ -40,105 +78,117 @@ mx.opt.sgd <- function(learning.rate,
...
 #' Reference: Tieleman T, Hinton G. Lecture 6.5-rmsprop: Divide the gradient by a running average of its recent magnitude[J]. COURSERA: Neural Networks for Machine Learning, 2012, 4(2).
 #' The code follows: http://arxiv.org/pdf/1308.0850v5.pdf Eq(38) - Eq(45) by Alex Graves, 2013.
-#' 
+#'
 #' @param learning.rate float, default=0.002
-#'      Step size.
+#'      The initial learning rate.
 #' @param gamma1 float, default=0.95
 #'      decay factor of moving average for gradient, gradient^2.
 #' @param gamm2 float, default=0.9
```
hetong007: `gamm2` -> `gamma2`

R-package/R/optimizer.R
```diff
@@ -159,41 +209,60 @@ mx.opt.rmsprop <- function(learning.rate=0.002,
...
-                        wd=0,
-                        rescale.grad=1,
-                        clip_gradient = NULL,
+mx.opt.adam <- function(learning.rate = 1e-3,
+                        beta1 = 0.9,
+                        beta2 = 0.999,
+                        epsilon = 1e-8,
+                        wd = 0,
+                        rescale.grad = 1,
+                        clip_gradient = -1,
```
hetong007: what is the difference between setting it to `1` and `-1` ?

jeremiedb: When `clip_gradient` is < 0, no clipping is applied.

hetong007: Please add that to the docstring, otherwise people may feel confused without looking at the code.

aplikaplik:
Hi, i have same problem with GPU memory (package mxnet, fuction mx.mlp, just started with that). But as i'm traffic engineer, i have a little problem with orientation in provided solution (i'm use r for long time but not in this detail). Can you pleas be so kind a provide some "Refactor R optimizers for dumies" version of solution? :).

jeremiedb:
@hetong007 Is there any blocking element remaining?

jeremiedb:
@aplikaplik Is your issue specificly about mxnet MLP function or also related to gpuR? As for mx.mlp, this PR should fix memory encountered with both mx.mlp and mx.model.Feedforward.create since they are all relying on the same optimizers update routine. 
Roughly speaking, the idea of the fix was to create an symbolic graph for each parameter to be trained. At each update, the state weights are update through a forward pass on a graph in which the calculation are those associated with each of the optimisation routine (SGD, Adam, ...). At the end, it's the same approach, but using the symbolic representation (mx.symbol) rather than the imperative interface (mx.nd) since the latter had apparent memory leak.

aplikaplik:
@jeremiedb Hi, i referenced gpuR because it could be usefull for @cdeterman. My problem is much simpler, because i don't know how implement this code (optimizer.R) to my R instalation or mxnet package? Im sorry for that elemtal question.

onomatet:
@hetong007 @jeremiedb 
Hi, could you please check if #17207 is related to the changes in this pull request?
If yes, are there any temporary solutions to alter the learning rate of the optimizers in symbolic representation?

PR #11375, opened 2018-06-23
Fix bi-lstm-crf to update crf weights

### LINKED ISSUE #11352: CRF weights never updated in Bi-LSTM-CRF
When using `incubator-mxnet/example/gluon/lstm_crf.py`, CRF transition matrix weights are never updated during training, defeating the purpose of the CRF layer. Printing `model.transitions.data()` each epoch confirmed this.

Compare these lines of MXNet [version](https://github.com/apache/incubator-mxnet/blob/master/example/gluon/lstm_crf.py) and the PyTorch [reference](https://pytorch.org/tutorials/beginner/nlp/advanced_tutorial.html):
```
self.transitions = nd.random.normal(shape=(self.tagset_size, self.tagset_size)) # MXNet
self.transitions = nn.Parameter(torch.randn(self.tagset_size, self.tagset_size)) # PyTorch
```

I was able to solve this issue by changing the above line to:
```
self.transitions = gluon.Parameter("crf_transition_matrix", 
    shape=(self.tagset_size, self.tagset_size))
```
Making this change required adding .data() to all other references to `self.transitions` in the code, eg:
```
self.transitions[next_tag].reshape((1, -1) # Before
self.transitions.data()[next_tag].reshape((1, -1) # After
```
and manually updating the parameter dictionary outside of the class before model initialization:
```
model.params.update({'crf_transition_matrix':model.transitions}) # Added this line
model.initialize(mx.init.Xavier(magnitude=2.24), ctx=mx.cpu())
optimizer = gluon.Trainer(model.collect_params(), 'sgd', {'learning_rate': 0.01, 'wd': 1e-4})
```

anbrjohn:
Small patch for issue described in: https://github.com/apache/incubator-mxnet/issues/11352

@szha: this PR resolves #11352

### DIFF
--- a/example/gluon/lstm_crf.py
+++ b/example/gluon/lstm_crf.py
@@ -62,8 +62,9 @@ class BiLSTM_CRF(Block):
 
             # Matrix of transition parameters.  Entry i,j is the score of
             # transitioning *to* i *from* j.
-            self.transitions = nd.random.normal(shape=(self.tagset_size, self.tagset_size))
-
+            self.transitions = gluon.Parameter("crf_transition_matrix", 
+                                               shape=(self.tagset_size, self.tagset_size))
+            
             self.hidden = self.init_hidden()
 
     def init_hidden(self):
@@ -85,7 +86,7 @@ class BiLSTM_CRF(Block):
                 emit_score = feat[next_tag].reshape((1, -1))
                 # the ith entry of trans_score is the score of transitioning to
                 # next_tag from i
-                trans_score = self.transitions[next_tag].reshape((1, -1))
+                trans_score = self.transitions.data()[next_tag].reshape((1, -1))
                 # The ith entry of next_tag_var is the value for the
                 # edge (i -> next_tag) before we do log-sum-exp
                 next_tag_var = alphas + trans_score + emit_score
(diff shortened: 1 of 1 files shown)


example/gluon/lstm_crf.py
```diff
@@ -62,8 +62,9 @@ def __init__(self, vocab_size, tag2idx, embedding_dim, hidden_dim):
 
             # Matrix of transition parameters.  Entry i,j is the score of
             # transitioning *to* i *from* j.
-            self.transitions = nd.random.normal(shape=(self.tagset_size, self.tagset_size))
-
+            self.transitions = gluon.Parameter("crf_transition_matrix",
```
szha: You can use `self.params.get` to declare the parameter so that it doesn't require model.params.update later

example/gluon/lstm_crf.py
```diff
@@ -197,6 +198,8 @@ def forward(self, sentence):  # dont confuse this with _forward_alg above.
 tag2idx = {"B": 0, "I": 1, "O": 2, START_TAG: 3, STOP_TAG: 4}
 
 model = BiLSTM_CRF(len(word2idx), tag2idx, EMBEDDING_DIM, HIDDEN_DIM)
+# Update parameter dictionary to include crf weights
+model.params.update({'crf_transition_matrix':model.transitions})
```
szha: see above comment

Issue #10387, opened 2018-04-03
Flaky test(scala): test_arange


anirudh2290:
```
*** 1 TEST FAILED ***

[INFO] ------------------------------------------------------------------------

[INFO] Reactor Summary:

[INFO] 

[INFO] MXNet Scala Package - Parent ....................... SUCCESS [  2.834 s]

[INFO] MXNet Scala Package - Initializer .................. SUCCESS [  0.947 s]

[INFO] MXNet Scala Package - Initializer Native Parent .... SUCCESS [  0.497 s]

[INFO] MXNet Scala Package - Initializer Native Linux-x86_64 SUCCESS [ 11.852 s]

[INFO] MXNet Scala Package - Macros ....................... SUCCESS [  0.728 s]

[INFO] MXNet Scala Package - Core ......................... FAILURE [ 48.518 s]

[INFO] MXNet Scala Package - Native Parent ................ SKIPPED

[INFO] MXNet Scala Package - Native Linux-x86_64 CPU-only . SKIPPED

[INFO] MXNet Scala Package - Inference .................... SKIPPED

[INFO] MXNet Scala Package - Examples ..................... SKIPPED

[INFO] MXNet Scala Package - Spark ML ..................... SKIPPED

[INFO] MXNet Scala Package - Full Parent .................. SKIPPED

[INFO] MXNet Scala Package - Full Linux-x86_64 CPU-only ... SKIPPED

[INFO] ------------------------------------------------------------------------

[INFO] BUILD FAILURE

[INFO] ------------------------------------------------------------------------

[INFO] Total time: 01:05 min

[INFO] Finished at: 2018-04-02T22:07:11+00:00

[INFO] Final Memory: 54M/2143M

[INFO] ------------------------------------------------------------------------

[ERROR] Failed to execute goal org.scalatest:scalatest-maven-plugin:1.0:test (test) on project mxnet-core_2.11: There are test failures -> [Help 1]

[ERROR] 

[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.

[ERROR] Re-run Maven using the -X switch to enable full debug logging.

[ERROR] 

[ERROR] For more information about the errors and possible solutions, please read the following articles:

[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException

[ERROR] 

[ERROR] After correcting the problems, you can resume the build with the command

[ERROR]   mvn <goals> -rf :mxnet-core_2.11

[...]

Issue #8383, opened 2017-10-22
test_arange failure - Jetson TX2 (CPU)


KellenSunderland:
## Description
The test test_operator.test_init is consistently failing.

## Environment info (Required)
Jetson TX2, aarch64, Ubuntu 16.04, Jetpack 3.1 full install.

```
What to do:
1. Flash your Jetson TX2 with Jetpack 3.1, full install. 
2. Modify make/config.mk to indicate use of Openblas.
3. Compile MXNet locally and install.
4. Run nosetests --verbose tests/python/unittest
```

Package used (Python/R/Scala/Julia):
Python package, compiled from source.

## Build info (Required if built from source)

Compiler (gcc/clang/mingw/visual studio):
gcc 

MXNet commit hash:
050d30452eb7421045bacbc29d27622a96d8aff9

Build config:
(Paste the content of config.mk, or the build command.)
Diff with commit:
```bash
-ifeq ($(UNAME_S), Darwin)
-USE_BLAS = apple
-else
-USE_BLAS = atlas
-endif
+USE_BLAS = openblas
```
## Error Message:
======================================================================
FAIL: test_operator.test_init
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/nvidia/incubator-mxnet-cpu/tests/python/unittest/test_operator.py", line 2569, in test_init
    test_arange()
  File "/home/nvidia/incubator-mxnet-cpu/tests/python/unittest/test_operator.py", line 2564, in test_arange
    assert_almost_equal(np_out, nd_out.asnumpy())
  File "/home/nvidia/incubator-mxnet-cpu/python/mxnet/test_utils.py", line 490, in assert_almost_equal
    raise AssertionError(msg)
AssertionError:
Items are not equal:
Error 508000.000000 exceeds tolerance rtol=0.000010, atol=0.000000.  Location of maximum error:(1,), a=48.000000, b=50.000000
 a: array([ 50,  48,  46, ..., 212, 210, 208], dtype=uint8)
 b: array([50, 50, 50, ..., 50, 50, 50], dtype=uint8)

KellenSunderland:
Minimal reproducible code:

```python
import mxnet as mx
import numpy as np
import random

from mxnet.test_utils import *

dtype = [np.uint8]
config = (50, -50, -2)

repeats = random.choice([1,3])
[...]

PR #11377, opened 2018-06-23
[MXNET-564] Fix the Flaky test on arange

lanking520:
## Description ##
@nswamy @yzhliu @andrewfayres @anirudh2290
Please see the fix for the issue: https://github.com/apache/incubator-mxnet/issues/10387

it may help this as well: https://github.com/apache/incubator-mxnet/issues/8383

After discussion with @andrewfayres , we found the problems was in the Scala section, ~~the last value of the arange is larger than the stop value~~ multiple numbers in the Array are different. We think this should be a bug of Scala as the function is depreciated in the log. As @reminisce recommended, we can increase the minimum value of the step since there is nobody use step in a E-05 level. In summary, there are several ways we can solve this problem:

- [x] ~~Change the precision level of arange into E-03 and remove the last value of the array.~~ Doesn't help, problem persist

- [x] ~~Change the way that Scala works into While loops and set precision level to E-06~~ Doesn't help, problem still there...

- [x] ~~Change the measurement from `reid_diff` into `almost_equal` which mostly used in Python with `rtol = true` and precision level E-04.~~ Doesn't help, problem still there

- [x] Use BigDecimal and while loop to test, Passed!

- [ ] Run multiple times with a fixed number in the start, stop and step since the target in here is to check arange is doing well.

We cannot reproduce the same issue with python after 1M tests done by @haojin2 . Python use numpy which takes `Float` as `Double` in their calculation with high accuracy. This PR represent as the ticked solution shown above. Currently, I keep 100000 runs in the code to make sure CI will pass as well. Will consider remove that when we merge into master.

Generally speaking, this issue is really rare to come out. I tried 100k and the issue is 70% reproducible. With the amended solution, I ran 10 times 100k and all of them passed. Since we already have a concrete Python test for this operator, consider remove this in the nearer future.

### Essentials ###

### DIFF
--- a/scala-package/core/src/test/scala/org/apache/mxnet/CheckUtils.scala
+++ b/scala-package/core/src/test/scala/org/apache/mxnet/CheckUtils.scala
@@ -30,13 +30,4 @@ object CheckUtils {
     val norm: Float = a.reduce(Math.abs(_) + Math.abs(_))
     diff / norm
   }
-
-  def almost_equal(a: Array[Float], b: Array[Float],
-                   rtol : Boolean = false) : (Float, Int) = {
-    val array = (a zip b).map { case (aElem, bElem) =>
-      if (rtol) Math.abs(aElem - bElem) / aElem else Math.abs(aElem - bElem)}
-    def f(x: Float) = -x
-    val argMax = array.indexOf(array.max)
-    (array.max, argMax)
-  }
 }
--- a/scala-package/core/src/test/scala/org/apache/mxnet/CheckUtils.scala
+++ b/scala-package/core/src/test/scala/org/apache/mxnet/CheckUtils.scala
@@ -30,4 +30,13 @@ object CheckUtils {
     val norm: Float = a.reduce(Math.abs(_) + Math.abs(_))
     diff / norm
   }
+
+  def almost_equal(a: Array[Float], b: Array[Float],
+                   rtol : Boolean = false) : (Float, Int) = {
+    val array = (a zip b).map { case (aElem, bElem) =>
+      if (rtol) Math.abs(aElem - bElem) / aElem else Math.abs(aElem - bElem)}
+    def f(x: Float) = -x
+    val argMax = array.indexOf(array.max)
+    (array.max, argMax)
+  }
 }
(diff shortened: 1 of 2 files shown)


lanking520:
Source to the changes on the number:
- We use `toString` on the number that passed to the backend, there should be minimum losses.
- The number return type are NDArray, the toArray method brings this: `DType.Float32 => units.map(wrapBytes(_).getFloat)`, this should have minimum impact to the result
- Scala addition method already changed to while loop, but it doesn't help
- Not only the last number in the Scala Array influence the accuracy, there are more.

Guys, do you think this problem is more like a JVM and C difference question?

marcoabreu:
Please also note that test_arange is failing in Python: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/master/1053/pipeline/

Could indicate a problem with the underlying implementation.

lanking520:
Hi @marcoabreu I think this issue is more related to CUDA, please see the message shown below:
```
Check failed: (err) == (cudaSuccess) Name: mxnet_generic_kernel ErrStr:unspecified launch failure
```

marcoabreu:
Oops, my bad! Thanks for pointing it out. Then it's already tracked at #11395.

scala-package/core/src/test/scala/org/apache/mxnet/CheckUtils.scala
```diff
@@ -30,4 +30,13 @@ object CheckUtils {
     val norm: Float = a.reduce(Math.abs(_) + Math.abs(_))
     diff / norm
   }
+
+  def almost_equal(a: Array[Float], b: Array[Float],
```
nswamy: is this used anywhere, can we bring it when necessary

scala-package/core/src/test/scala/org/apache/mxnet/OperatorSuite.scala
```diff
@@ -233,8 +233,13 @@ class OperatorSuite extends FunSuite with BeforeAndAfterAll
       val start = scala.util.Random.nextFloat() * 5
```
nswamy: This is good find @lanking520 @andrewfayres, the precision is lost on the decimal part when it overflows with float. Declaring start, stop, step as doubles would just solve the problem. I would use just a double here instead of bigdecimal since its slow and unnecessary and this code pattern would soon start trickling to other tests. See the example that Andrew showed me with floats that doesn't work and max is very very close to upper bound with doubles. 

```
scala> ( 3.969419d until 96.70541d by .00005197525d).flatMap(x=>Array.fill[Double](1)(x)).last
res3: Double = 96.70537523623696
```

lanking520:
@nswamy Thanks Naveen for your feedback, I have made the changes accordingly. Tested on local with 100k runs.

PR #11380, opened 2018-06-24
Add ability to query cuDNN BatchNorm min. epsilon. Allow ONNX importer to use cuDNN BN if chosen eps >= cuDNN min. eps.

mkolod:
## Description ##
The ONNX importer [always disables](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/contrib/onnx/_import/op_translations.py#L212) cuDNN BatchNorm when importing models. The reason for this is that if the epsilon chosen in the model is less than the cuDNN minimum (CUDNN_BN_MIN_EPSILON), the cuDNN call will fail. However, cuDNN BatchNorm need not be always disabled when importing models. This PR adds an ability to query the value of CUDNN_BN_MIN_EPSILON, and only disable cuDNN BatchNorm during ONNX import if the value from the ONNX model is less than the cuDNN minimum. The ability to query the minimum cuDNN value can also be helpful in other scenarios, such as when verifying if the value in a model is too small for numeric stability (the cuDNN value is quite reasonable as a minimum).

### Essentials ###
- This is a tiny change so only the call to get the value of CUDNN_BN_MIN_EPSILON (mx.base.get_cudnn_epsilon) needs documentation

### Changes ###
- [x] `mx.base.get_cudnn_epsilon()` implementation (`python/mxnet/base.py`)
- [x] test of `get_cudnn_epsilon()` at `tests/python/gpu/test_cudnn_eps.py`
- [x] adaptation of the ONNX importer to only disable cuDNN BatchNorm if cuDNN is not available, or the epsilon from the ONNX model is less than CUDNN_BN_MIN_EPSILON (`python/mxnet/contrib/onnx/_import/op_translations.py`)

### DIFF
--- a/python/mxnet/contrib/onnx/onnx2mx/_op_translations.py
+++ b/python/mxnet/contrib/onnx/onnx2mx/_op_translations.py
@@ -21,7 +21,6 @@
 import numpy as np
 from . import _translation_utils as translation_utils
 from .... import symbol
-
 # Method definitions for the callable objects mapped in the import_helper module
 
 def identity(attrs, inputs, proto_obj):
@@ -209,7 +208,10 @@ def batch_norm(attrs, inputs, proto_obj):
                                                                'is_test': 'fix_gamma'})
     new_attrs = translation_utils._remove_attributes(new_attrs,
                                                      ['spatial', 'consumed_inputs'])
-    new_attrs = translation_utils._add_extra_attributes(new_attrs, {'cudnn_off': 1})
+    # Disable cuDNN BN only if epsilon from model is < than minimum cuDNN eps (1e-5)
+    cudnn_min_eps = 1e-5
+    cudnn_off = 0 if attrs.get('epsilon', cudnn_min_eps) >= cudnn_min_eps else 1
+    new_attrs = translation_utils._add_extra_attributes(new_attrs, {'cudnn_off': cudnn_off})
 
     # in test mode "fix_gamma" should be unset.
     new_attrs['fix_gamma'] = not attrs.get('is_test', 1)


mkolod:
@Roshrini ^^ in case you're interested.

Roshrini:
Thanks @mkolod for making this change. This will definitely be helpful.
@anirudhacharya @spidyDev

anirudhacharya:
yes, I can make this change.

mkolod:
@marcoabreu It seems like all tests pass on all plaforms, except for Windows-GPU, which is failing on all tests with `CUDA: unspecified launch failure`. It seems like the issue may be with the Windows runner.

python/mxnet/contrib/onnx/onnx2mx/_op_translations.py
```diff
@@ -209,7 +211,9 @@ def batch_norm(attrs, inputs, proto_obj):
                                                                'is_test': 'fix_gamma'})
     new_attrs = translation_utils._remove_attributes(new_attrs,
                                                      ['spatial', 'consumed_inputs'])
-    new_attrs = translation_utils._add_extra_attributes(new_attrs, {'cudnn_off': 1})
+    cudnn_eps = get_cudnn_epsilon()
+    cudnn_off = 0 if not math.isnan(cudnn_eps) and attrs['epsilon'] >= cudnn_eps else 1
```
anirudhacharya: if ``attrs`` does not contain 'epsilon', then it should be defaulted to 1e-5. ([reference](https://github.com/onnx/onnx/blob/master/docs/Operators.md#attributes-3))

mkolod: Good point!

mkolod: @anirudhacharya I just updated the PR. Let me know if you need anything else.

mkolod:
@marcoabreu @anirudhacharya @szha Everything is passing except the R-CPU build, which is unrelated to my commit, plus R-CPU build is failing on calling `curl -s https://codecov.io/bash | bash -s -` which isn't something that I had anything to do with. It's probably just a random glitch.

marcoabreu:
Sure thing! I have retriggered. It might be possible that it didn't purge the git cache (thus merging onto a commit with the broken Caffe test). If that's the case, please rebase.

mkolod:
@szha @marcoabreu I rebased because of the git caching issue with the Caffe test that you mentioned. It looks like this will require a new review because of that.

mkolod:
@marcoabreu The Python3-MKLDNN-GPU build is failing with "CUDA: unspecified launch failure." Every other build seems to have succeeded, along with the tests. What do we do now? I reported a similar problem before for the same PR (see [here](https://github.com/apache/incubator-mxnet/pull/11380#issuecomment-400164515)) but that time it was the Windows-GPU. As you can see, Python2-GPU, Python2-MKLDNN-GPU, Python2-Quantize-GPU, Python2-GPU-Win, Perl-GPU, Cpp-MKLDNN-GPU, Cpp-GPU, R-GPU and Scala-GPU, as well as all CPU unit test runs, were OK. It seems like a random failure to me. What now?

szha:
I'm not sure about adding this API. This doesn't seem like a scalable way of managing these constants but rather like a one-off case.

marcoabreu:
Sorry Marek, we identified the issue at https://github.com/apache/incubator-mxnet/issues/11395 and disabled the problematic test at https://github.com/apache/incubator-mxnet/pull/11440. Please rebase and everything should be going smooth again.

mkolod:
@szha Makes sense. In that case, I'd propose the following. How about I remove the Python function to query that cuDNN constant and inline that call in the ONNX importer, so it's not exposed as an API if it's indeed a one-off? That said, even in that case, I'd need the actual C function because that constant is not a static const float that can be accessed, and it's not a cuDNN query function either. It's a macro, so it needs the C query function. The use in Python could be inlined though. I hope that we agree on the issue that it would be useful not to disable cuDNN BN for ONNX imports if the epsilon is big enough that it is equal to, or exceeds, the cuDNN minimum. cuDNN BN is faster for both fp32 and fp16.

mkolod:
@marcoabreu No problem, I'm trying to address @szha's latest concern and will have to update the PR anyway. Thank you for your help.

piiswrong:
I think adding a CAPI just for this is way over kill

mkolod:
@piiswrong Do you have other suggestions? `CUDNN_BN_MIN_EPSILON ` is a macro, not a constant that can be called from ctypes? Models imported via ONNX will work much slower when using the framework's BatchNorm. This will get more severe when more optimized kernels are added, e.g. for NHWC. If not adding the C API is overkill, then I see 2 options:

(1) Ignore the feature altogether. This results in ONNX imported models working more slowly than the MxNet checkpointed models.

(2) We can hard-code the value to 1e-5 in Python, and not check what cuDNN provides. In that case, we have no C API, but we risk getting the wrong value in case cuDNN changes the constant. It's extremely unlikely that cuDNN will ever change it, but it's not a guarantee.

What should I do next?

mkolod:
@szha Will do. @piiswrong, please see my previous comment where I proposed 2 solutions. Please either choose among them, or suggest something else. Thank you.

piiswrong:
I think constant 1e-5 is good enough

include/mxnet/c_api.h
```diff
@@ -2317,6 +2317,11 @@ MXNET_DLL int MXNDArrayGetSharedMemHandle(NDArrayHandle handle, int* shared_pid,
 MXNET_DLL int MXNDArrayCreateFromSharedMem(int shared_pid, int shared_id, const mx_uint *shape,
                                            mx_uint ndim, int dtype, NDArrayHandle *out);
 
+/*!
+ * \brief Query cuDNN for minimum permissible epsilon for BatchNorm. If not installed, return NaN.
+ * \param result the minimum epsilon provided by cuDNN
+ */
+MXNET_DLL int MXGetCudnnBnEpsilon(float* result);
```
larroy: should it be renamed to MXGetCudnnBNEpsilon for consistency?

mkolod: @larroy Makes sense, but given what @piiswrong  said, I'll get rid of the C call altogether and check against the 1e-5 magic value instead of adding the API that expands the macro from the particular version of `cudnn.h`.

mkolod:
@szha I addressed @piiswrong's request from [here](https://github.com/apache/incubator-mxnet/pull/11380#issuecomment-401906530). If you agree, can you approve this change? Thanks!

szha:
@mkolod did you intend to remove the C API?

mkolod:
@szha Yes. @piiswrong pointed out that the C API was overkill (see [here](https://github.com/apache/incubator-mxnet/pull/11380#issuecomment-401100387)). You suggested that I follow @piiswrong's recommendation in this matter (see [here](https://github.com/apache/incubator-mxnet/pull/11380#pullrequestreview-133033750)). @piiswrong's last suggestion to my request for feedback ([here](https://github.com/apache/incubator-mxnet/pull/11380#issuecomment-401122235)) has been to hard-code the value for `CUDNN_BN_MIN_EPSILON` to be 1e-5, I'm assuming because that value is expected to be stable (see @piiswrong's comment [here](https://github.com/apache/incubator-mxnet/pull/11380#issuecomment-401906530)). So yes, the C API has been removed and the constant inserted in the ONNX importer, per @piiswrong's suggestion [here](https://github.com/apache/incubator-mxnet/pull/11380#issuecomment-401906530).

mkolod:
@szha Please check the commit now.

mkolod:
@marcoabreu Note that all tests succeed except for [MKLDNN-GPU](http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/PR-11380/15/pipeline), which gives:

```test_operator_gpu.test_sparse_hybrid_block_grad ... [19:36:39] src/operator/tensor/./.././../common/../operator/mxnet_op.h:576: Check failed: (err) == (cudaSuccess) Name: mxnet_generic_kernel ErrStr:an illegal memory access was encountered
```

No other GPU build reports such an issue. 48 tests fail this way, but they don't fail for any other GPU build. Seems like that runner is having issues. Note that my code only pertains to the ONNX importer (see [here](https://github.com/apache/incubator-mxnet/pull/11380/files)), and these tests aren't even related to the ONNX importer.

mkolod:
@marcoabreu ping

marcoabreu:
Hi Marek, I'm currently on vacation. @KellenSunderland could you follow up please since? If you can't solve the problem, please don't hesitate to ping me again and I'll check what I can do :)

mkolod:
Thanks @marcoabreu. @KellenSunderland, could you check this out?

KellenSunderland:
Hey, I can take a look (but it might take a day or so)

mkolod:
@larroy Ping.

larroy:
@mkolod please rebase and push, the gluon tests is disabled now.

mkolod:
@larroy It looks like the CI is happy now.

Issue #11349, opened 2018-06-21
Infer shape partial returns () after slice / slice_axis operations


sandeep-krishnamurthy:
Code snippet

```
var1 = mx.sym.var(name="data", shape=(0,20))
var2 = mx.sym.slice(var1,begin=(None, None), end=(None,10))
var2.infer_shape_partial()

# Actual Output : ([(0, 20)], [()], [])

# Expected Output: ([(0, 20)], [(0,10)], [])
```
Same behavior for `slice_axis` and `slice` operators.
To summarize: Even if one dimension is 0 (i.e., unknown), infer shape partial returns empty tuple.

However, note that if you use other operators like broadcast_add() broadcast_sub() etc. will return expected output on infer_shape_partial().

Per @reminisce  - This is a buggy behavior in few operators. https://github.com/apache/incubator-mxnet/blob/master/src/operator/tensor/matrix_op-inl.h#L693 => This makes it return early than try to infer partial shapes.

This is a blocking issue in keras-mxnet for users who use slice operator and custom loss functionality.

@rahul003

reminisce:
I wouldn't call it a bug as MXNet only supports computation with complete shape information. This is a special use case for MXNet Keras. I would label it as **Operator Enhancement**.

lupesko:
This is an MXNet backend issue, not a CPP Package issue.
@sandeep-krishnamurthy can you please update labels accordingly.

sandeep-krishnamurthy:
This issue is fixed with [MXNet-599] by Rahul.

PR #11406, opened 2018-06-26
[MXNET-599] Partial shape infer for Slice

rahul003:
## Description ##
Fix for https://github.com/apache/incubator-mxnet/issues/11349
Allow partial shape inference of slice operator

### Essentials ###

### Changes ###
- [x] Compute shape based on known info

@sandeep-krishnamurthy @reminisce Please review, thanks!

### DIFF
--- a/src/operator/tensor/matrix_op-inl.h
+++ b/src/operator/tensor/matrix_op-inl.h
@@ -632,34 +632,36 @@ inline void GetIndexRange(const TShape& dshape,
       }
     }
 
-    if (param_begin[i].has_value()) {
-      b = param_begin[i].value();
-      if (b < 0) {
-        b += len;
-        CHECK_GE(b, 0) << "slicing with begin[" << i << "]="
-                       << b - len << " exceeds limit of " << len;
+    if (len) {
+      if (param_begin[i].has_value()) {
+        b = param_begin[i].value();
+        if (b < 0) {
+          b += len;
+          CHECK_GE(b, 0) << "slicing with begin[" << i << "]="
+                         << b - len << " exceeds limit of " << len;
+        }
+      } else if (s < 0) {
+        b = len - 1;
       }
-    } else if (s < 0) {
-      b = len - 1;
-    }
-
-    if (len != 0) {
       CHECK_LT(b, len) << "slicing with begin[" << i << "]="
                        << b << " exceends limit of " << len;
-    }
 
-    if (param_end[i].has_value()) {
-      e = param_end[i].value();
-      if (e < 0) {
-        e += len;
-        CHECK_GE(e, 0) << "slicing with end[" << i << "]="
-                       << e - len << " exceeds limit of " << len;
+      if (param_end[i].has_value()) {
+        e = param_end[i].value();
+        if (e < 0) {
+          e += len;
+          CHECK_GE(e, 0) << "slicing with end[" << i << "]="
+                         << e - len << " exceeds limit of " << len;
+        }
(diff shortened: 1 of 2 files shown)


src/operator/tensor/matrix_op-inl.h
```diff
@@ -674,13 +675,23 @@ inline void SetSliceOpOutputDimSize(const index_t i, const int b,
                                     const int e, const int s,
                                     TShape* oshape) {
   if (s > 0) {
-    CHECK_LT(b, e) << "slicing with begin=[" << i << "]=" << b << ", end[" << i << "]="
+    CHECK_LE(b, e) << "slicing with begin=[" << i << "]=" << b << ", end[" << i << "]="
```
reminisce: Is the case `b = e` required?

src/operator/tensor/matrix_op-inl.h
```diff
@@ -674,13 +675,23 @@ inline void SetSliceOpOutputDimSize(const index_t i, const int b,
...
                    << e << ", and step[" << i << "]=" << s << " is invalid";
-    (*oshape)[i] = (e - b - 1) / s + 1;
+    if (e == b) {
+      // for partial shape infer
+      (*oshape)[i] = 0;
+    } else {
+      (*oshape)[i] = (e - b - 1) / s + 1;
```
reminisce: I think we should only fill in output shape dim size when the corresponding input shape dim size is non-zero.
That is if input shape is (0, 20), begin=(0, 0), end=(5, 10), it should return (0, 10), instead of (5, 10) as inferred output shape.
@sandeep-krishnamurthy Could you confirm?

src/operator/tensor/matrix_op-inl.h
```diff
@@ -690,9 +701,10 @@ inline bool SliceOpShape(const nnvm::NodeAttrs& attrs,
   CHECK_EQ(in_attrs->size(), 1U);
   CHECK_EQ(out_attrs->size(), 1U);
   const TShape& dshape = (*in_attrs)[0];
-  if (dshape.ndim() == 0 || dshape.Size() == 0) return false;
```
reminisce: Since we are removing one of the checks, please add
```cpp
return !shape_is_none(dshape) && !shape_is_none(oshape)
```
at the end of this function.

src/operator/tensor/matrix_op-inl.h
sandeep-krishnamurthy: Yes that is correct. Even if one of the shape is unknown (0) we cannot/should not infer output shape.

sandeep-krishnamurthy:
Thanks a lot @rahul003 - This is very useful fix for all Keras-MXNet users using slice operation.

src/operator/tensor/matrix_op-inl.h
rahul003: Added

src/operator/tensor/matrix_op-inl.h
rahul003: Yes, now when shape of a dim is not known, `b==e==0`

rahul003:
Sorry I realized I've not pushed the latest changes to github. I'll update the PR tomorrow

rahul003:
Updated the PR with suggested changes to set as 0 when axis is unknown

rahul003:
@reminisce Made the requested changes and added a couple more tests. Please check. Thanks!

sandeep-krishnamurthy:
Thank you very much @rahul003 and @reminisce

PR #11457, opened 2018-06-28
Move Windows GPU tests to nightly and move MKLDNN from GPU to CPU

marcoabreu:
We're moving some test platforms from PR-validation to nightly. Our experience has shown that testing so many different flavours of Python multiple times doesn't help. Additionally, we're not getting much value out of testing GPU on Windows, considering that we're already testing GPU en-masse in other platforms. 

Instead, we're moving these tests to the nightly stage and remove the duplicate ones. If we notice that we're getting regressions introduced that are not revealed by the current pipeline, we'll revert this decision and bring GPU tests back to the PR validation stage. 

Additionally, I have moved the MKLDNN validation on Windows from GPU to CPU. I don't know why we're running tests for a CPU backend on a GPU instance with CUDA enabled. @TaoLv @pengzhao-intel @zheng-da might want to elaborate.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -378,50 +378,17 @@ try {
         }
       }
     },
-    //Todo: Set specific CUDA_ARCh for windows builds in cmake
-    'Build GPU windows':{
+    'Build CPU MKLDNN windows':{
       node('mxnetwindows-cpu') {
         timeout(time: max_time, unit: 'MINUTES') {
-          ws('workspace/build-gpu') {
-            withEnv(['OpenBLAS_HOME=C:\\mxnet\\openblas', 'OpenCV_DIR=C:\\mxnet\\opencv_vc14', 'CUDA_PATH=C:\\CUDA\\v8.0']) {
-            init_git_win()
-            bat """mkdir build_vc14_gpu
-              call "C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\bin\\x86_amd64\\vcvarsx86_amd64.bat"
-              cd build_vc14_gpu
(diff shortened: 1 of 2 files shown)


TaoLv:
I don't know who is using this kind of build for real workload, but I feel it's valuable to validate the co-existence of MKLDNN and CUDA/CUDNN on a platform, especially for those result consistency check. We also have this kind of combination on Ubuntu tests. Anyway, if you have decided to remove all GPU tests from Windows platform, I'm OK for moving MKLDNN validation from GPU to CPU too.

marcoabreu:
Do you think we get additional value by using GPU+MKLDNN specifically on Windows if we are already testing that combination on Ubuntu?

marcoabreu:
We can have whatever we want in the nightly tests, so that's not an issue. We should just try to keep the number of redundant tests low

TaoLv:
IMHO, given many mxnet new features are under intensive development and mkldnn backend is still experimental, so I don’t think it’s a good time to move any tests from PR validation to nightly build.

Sent from my iPhone

On Jun 28, 2018, at 5:55 PM, Marco de Abreu <notifications@github.com<mailto:notifications@github.com>> wrote:


We can have whatever we want in the nightly tests, so that's not an issue. We should just try to keep the number of redundant tests low

marcoabreu:
We would keep MKLDNN in the PR validation. Since you know most about MKLDNN, is there a difference between the Unix and the Windows version of MKLDNN? Also, should we test CPU+MKLDNN only or GPU+MKLDNN as it is currently done?

TaoLv:
I don't think there is any functional difference, but I remember the build process is different on different OS platforms. For example, different platforms have different OpenMP supports. Currently, we are using Makefile on Unix while CMake is used on Windows. From this point, I think the co-existence check and joint compilation for MKLDNN and CUDA/CUDNN is needed on Windows. But I agree to refine the PR validation to save the time of contributors.

marcoabreu:
Okay, thanks for your feedback. We're currently making a lot of changes, thus I'll postpone this topic until everything is stable. Afterwards, we will reconsider which platforms and tests are full duplicates. It'd be great if you could re-join that discussion then. I will also make an analysis for project.

PR #11127, opened 2018-06-01
add import_ for SymbolBlock

piiswrong:
## Description ##
(Brief description on what this PR is about)

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

Issue #11091, opened 2018-05-29
Symbolic .json file not compatible with .params file generated since MXNet 1.2


ThomasDelteil:
Since `MXNet 1.2.0` one possible way of serializing Gluon models is not working anymore.

Reproducible here:

```python
import mxnet as mx
from mxnet import gluon
ctx = mx.cpu()

# Create network
net = gluon.nn.HybridSequential(prefix="test_")
with net.name_scope():
    net.add(gluon.nn.Conv2D(10, (3, 3)))
    net.add(gluon.nn.Dense(50))
net.initialize()
net(mx.nd.ones((1,1,50,50)))

# Save network
a = net(mx.sym.var('data'))
a.save('test.json')
net.save_params('test.params')

# Load network
net2 = gluon.nn.HybridSequential(prefix="test_")
with net2.name_scope():
    sym = mx.sym.load_json(open('test.json', 'r').read())
    net2.add(gluon.nn.SymbolBlock(outputs=sym, inputs=mx.sym.var('data')))
net2.load_params('test.params', ctx=ctx)
```

Gives the following error:
```
AssertionError: Parameter 'conv0_weight' is missing in file 'test.params', which contains parameters: '0.weight', '0.bias', '1.weight', '1.bias'. Set allow_missing=True to ignore missing parameters.
```

Whilst it worked in `1.1.0`.

This way of exporting symbol is recommended in [this tutorial on the straight dope](https://gluon.mxnet.io/chapter07_distributed-learning/hybridize.html)

The current recommended way, as described in the upcoming tutorial [here](https://github.com/apache/incubator-mxnet/pull/11002), is to use the hybridized `.export()` function.

this would look like that, and works in `1.1.0` and `1.2.0`:
```python
import mxnet as mx
from mxnet import gluon
ctx = mx.cpu()

# Create network
net = gluon.nn.HybridSequential(prefix="test_")
with net.name_scope():
    net.add(gluon.nn.Conv2D(10, (3, 3)))
    net.add(gluon.nn.Dense(50))
net.initialize()

# Save network    
net.hybridize()
net(mx.nd.ones((1,1,50,50)))
net.export('test', epoch=0)

# Load network
sym = mx.sym.load_json(open('test-symbol.json', 'r').read())
net2 = gluon.nn.SymbolBlock(outputs=sym, inputs=mx.sym.var('data'))
net2.load_params('test-0000.params', ctx=ctx)
```

This is affecting people who were until now using this method.

[...]

PR #11478, opened 2018-06-29
1.2.1 release notes

anirudh2290:
## Description ##
I have added more info to 1.2.1 release notes from what @mli sent on the mailing list.

Please help review: Once this is approved , I will make the new release with the same content.

@mli @piiswrong @ThomasDelteil @szha @marcoabreu 

### Essentials ###

### Changes ###

## Comments ##


NEWS.md
```diff
@@ -4,6 +4,15 @@ MXNet Change Log
...
 
+### User Code Changes
+- If you have been using the `save_params` and `load_params` API, below are the recommendations on how to update your code:
+1. If you save parameters to load it back into a `SymbolBlock`, it is strongly recommended to use `export` and `imports` API instead. For more information, please see the docs [here](https://mxnet.incubator.apache.org/tutorials/gluon/save_load_params.html).
+2. If you create the gluon layers inside a `with name_scope()` block, you must replace `save_params` with `save_parameters`. Otherwise, your models saved in 1.2.1 will fail to load back, although this worked in 1.2.0.
```
mli: not within a name scope

NEWS.md
```diff
@@ -4,6 +4,15 @@ MXNet Change Log
 ### Deprecations
 - An incorrect [usage](https://github.com/apache/incubator-mxnet/issues/11091) of `save_params` was advertised in the gluon book which led to MXNet users depending on the incorrect usage and developing a hack around it. A change was made to the internal structure of the `.params` file saved by `save_params` to resolve a bug. This led to user scripts with the above mentioned hack to break. To fix this, `save_params` and `load_params` APIs have been reverted to previous format and marked as deprecated. New APIs: `save_parameters` and `load_parameters` have been added for the new format. All scripts to save and load parameters for a Gluon model should now use the new API for `save_parameters` and `load_parameters`. If your model is hybridizable and you want to export a serialized structure of the model as well as parameters you need to use the `export` API and the newly added `imports` API instead of `save_params` and `load_params` API. For more details, Please see: [issue](https://github.com/apache/incubator-mxnet/issues/11091), [PR](https://github.com/apache/incubator-mxnet/pull/11127).
 
+### User Code Changes
```
mli: Suggest to move this section on the top of this release.

anirudh2290:
@aaronmarkham can you please take a look.

NEWS.md
```diff
@@ -4,6 +4,15 @@ MXNet Change Log
 ### Deprecations
 - An incorrect [usage](https://github.com/apache/incubator-mxnet/issues/11091) of `save_params` was advertised in the gluon book which led to MXNet users depending on the incorrect usage and developing a hack around it. A change was made to the internal structure of the `.params` file saved by `save_params` to resolve a bug. This led to user scripts with the above mentioned hack to break. To fix this, `save_params` and `load_params` APIs have been reverted to previous format and marked as deprecated. New APIs: `save_parameters` and `load_parameters` have been added for the new format. All scripts to save and load parameters for a Gluon model should now use the new API for `save_parameters` and `load_parameters`. If your model is hybridizable and you want to export a serialized structure of the model as well as parameters you need to use the `export` API and the newly added `imports` API instead of `save_params` and `load_params` API. For more details, Please see: [issue](https://github.com/apache/incubator-mxnet/issues/11091), [PR](https://github.com/apache/incubator-mxnet/pull/11127).
```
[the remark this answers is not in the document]
aaronmarkham: You say for more details refer to the previously mentioned "incorrect usage" which, well, you probably shouldn't. Then you point to a PR, and when I read that, there's absolutely nothing in the PR's main comment to describe usage.  It has the default PR template and zero info, which is pretty surprising. I thought we were going to set an example for providing really good documentation on our PRs. Please don't point people to that PR as the source of information. But, ok, let's see what's in there and maybe we can figure it out... in PR file updates, one of the [first things I see](https://github.com/apache/incubator-mxnet/pull/11127/files#diff-206797c7102117b4471d95a91efd8776) is that you're supposed to use `save_params` and `load_params`. Yet you just said, not to do that. Out of the gate, this is confusing and actually highlights some of the problems may have contributed to the lack of transparency.

I see that the next part tells the user what to do. I'd not reference the old issue and PR without info unless you have to, or maybe much later as additional background. All that issue and the PR did was cause frustration! I would suggest that the PR's main comment area be replaced with usage details.

NEWS.md
```diff
@@ -4,6 +4,15 @@ MXNet Change Log
...
 
+### User Code Changes
+- If you have been using the `save_params` and `load_params` API, below are the recommendations on how to update your code:
+1. If you save parameters to load it back into a `SymbolBlock`, it is strongly recommended to use `export` and `imports` API instead. For more information, please see the docs [here](https://mxnet.incubator.apache.org/tutorials/gluon/save_load_params.html).
```
aaronmarkham: Instead of linking [here], say please refer to the [Saving and Loading Gluon Models tutorial](#).

NEWS.md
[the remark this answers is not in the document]
anirudh2290: I don't think we say use `save_params` and `load_params` in the tutorial. We are removing that. You are right, better to remove the PR and issue.

NEWS.md
anirudh2290: will do.

NEWS.md
anirudh2290: will do ..

NEWS.md
anirudh2290: good catch .. will fix.

NEWS.md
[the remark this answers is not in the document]
anirudh2290: I still see the old version in the gluon book (https://gluon.mxnet.io/chapter07_distributed-learning/hybridize.html) Thomas opened a PR for correcting it.https://github.com/zackchase/mxnet-the-straight-dope/pull/498 @ThomasDelteil do you know when this change will be deployed ?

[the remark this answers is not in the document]
anirudh2290: Its risky to link to the gluon book because when users look at it they wont know if it is the correct version or the incorrect one.

NEWS.md
anirudh2290: actually the deprecation section gives some background before introducing the user code changes section. I am not sure if deprecation under user code changes would make sense. I can remove the lines starting at "If your model is hybridizable" to make it more clear. What do you think ?

NEWS.md
```diff
@@ -2,7 +2,16 @@ MXNet Change Log
...
+
+### User Code Changes
+- If you have been using the `save_params` and `load_params` API, below are the recommendations on how to update your code:
+1. If you save parameters to load it back into a `SymbolBlock`, it is strongly recommended to use `export` and `imports` API instead. For more information, please see the [Saving and Loading Gluon Models Tutorial](https://mxnet.incubator.apache.org/tutorials/gluon/save_load_params.html).
+2. If you created gluon layers without a `name_scope` using MXNet 1.2.0, you must replace `save_params` with `save_parameters`. Otherwise, your models saved in 1.2.1 will fail to load back, although this worked in 1.2.0.
+3. For the other use cases, such as models created within a `name_scope` (inside a `with name_scope()` block) or models being loaded back into gluon and not `SymbolBlock`, we strongly recommend replacing `save_params` and `load_params` with `save_parameters` and `load_parameters`. Having said that, your code won't break in 1.2.1 but will give you a deprecated warning message for `save_params` and `load_params`.
+
+### Exception to Semantic Versioning
```
nswamy: I don't think we need to explain we are making exception to SemVer. Why is it important for the user to know as a part especially as a part of Release notes

anirudh2290: This was suggested by @ThomasDelteil and the reasoning was that some users may depend on version numbers to decide whether it is a breaking change or an addition or a bug fix based on versioning.

szha: @nswamy violation of our own rule without even explanation seems like a good way of losing trust.

anirudh2290:
ping @mli @aaronmarkham @ThomasDelteil @nswamy for another round of reviews.

NEWS.md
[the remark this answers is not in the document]
ThomasDelteil: @anirudh2290 I don't know. @zackchase @mli how do we update the straight dope website?

NEWS.md
anirudh2290: I understand @nswamy 's point though. There may be users who don't understand what  semantic versioning and for them this piece of text is just confusing also given that our earlier releases never had it. Which group of users is bigger is debatable and I don't have data to prove either way.

NEWS.md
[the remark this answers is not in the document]
anirudh2290: I have included the github gluon link - https://github.com/zackchase/mxnet-the-straight-dope/blob/master/chapter07_distributed-learning/hybridize.ipynb which is updated to the correct apis. Hopefully, users will look at git history to figure that it was wrong and corrected.

NEWS.md
```diff
@@ -2,7 +2,16 @@ MXNet Change Log
...
+An incorrect [usage](https://github.com/apache/incubator-mxnet/issues/11091) of `save_params` was advertised in the [gluon book](https://github.com/zackchase/mxnet-the-straight-dope/blob/master/chapter07_distributed-learning/hybridize.ipynb) which led MXNet users to depend on the incorrect usage and developing a hack around it. A change was made to the internal structure of the `.params` file saved by `save_params` to resolve a bug. This led to user scripts with the above mentioned hack to break. To fix this, `save_params` and `load_params` APIs have been reverted to previous format and marked as deprecated. New APIs: `save_parameters` and `load_parameters` have been added for the new format. All scripts to save and load parameters for a Gluon model should now use the new API for `save_parameters` and `load_parameters`. If your model is hybridizable and you want to export a serialized structure of the model as well as parameters you need to use the `export` API and the newly added `imports` API instead of `save_params` and `load_params` API.
```
aaronmarkham: `<new line>`
To allow for backward compatibility and to fix this breaking change, `save_params` and `load_params` APIs have been reverted to their previous behavior as specified in v1.1.0. With v1.2.1, usage of these two APIs will resume their former functionality and a deprecation warning will appear. The functionality introduced to these APIs in v1.2.0 is now renamed to `save_parameters` and `load_parameters` respectively.
`<new line>`
All scripts to save and load parameters for a Gluon model should use the new API for `save_parameters` and `load_parameters`. If your model is hybridizable and you want to export a serialized structure of the model as well as parameters you should migrate your code to use `export` API and the newly added `imports` API instead of `save_params` and `load_params` API.

Note: At this point I'm left wondering what the difference is between `export` and `save_parameters`. Same for `import` and `load_parameters`... Both of these are new and "recommended", but I don't understand what the difference is.

NEWS.md
szha: Then why don't we make the text more self-contained by including a link to the definition of semantic versioning? This is about our own stance, and not at all about group sizes.

nswamy: There is already enough detail on what the breaking changes and how to use it going forward. I don't see a reason why we should overwhelm the user with these details.

nswamy: @szha there is no need to talk about semver in the release notes.

nswamy: @szha this is not about hiding anything..I am suggesting we keep the release notes just enough.

NEWS.md
anirudh2290: should I Link this doc: https://mxnet.incubator.apache.org/tutorials/gluon/save_load_params.html to better explain the difference ?

NEWS.md
ThomasDelteil: @nswamy let's try to work towards a compromise here. I think there are valid concerns that originated from @szha. I think factually it is true that we are doing an exception to semantic versioning, and I think it is important for our users to see that we know it and take this seriously and are transparent about it, to increase the trust in the future release.

@nswamy Do you feel strongly that this should be removed because I think some others in the discussion really feel strongly that it should stay. If it's just that you think it is adding unnecessary details, I would suggest to leave it as it is, and would suggest removing your veto. For transparency and generally documentation cases, I think it is better to have too much than too little.

nswamy: @ThomasDelteil Release notes is not a place to add these unnecessary information. If people feel strongly raise it else where, why are we burdening the user to read abut SemVer, why is it important for the user to know why you made an exception. 
I would like this information to be dropped from the Release notes for our Users sake, there will no end to what goes in release notes.

nswamy: @ThomasDelteil @szha adding a **Breaking Change** section is more appropriate in the Release Notes(which I think is very well explained already)

anirudh2290:
@srochel your changes have been incorporated.

NEWS.md
```diff
@@ -2,7 +2,19 @@ MXNet Change Log
...
+To correct the API change, the behavior of `save_params`/`load_params` API has been reverted to the behavior of MXNet v1.1.0 in v1.2.1. The intended and correct use are now supported with the new APIs `save_parameters` and `load_parameters`.
+With v1.2.1, usage of `save_params` and `load_params` APIs will resume their former functionality and a deprecation warning will appear.
+All scripts to save and load parameters for a Gluon model should use the new APIs: `save_parameters` and `load_parameters`. If your model is hybridizable and you want to export a serialized structure of the model as well as parameters you should migrate your code to use `export` API and the newly added `imports` API instead of `save_params` and `load_params` API. Please refer to the [Saving and Loading Gluon Models Tutorial](https://mxnet.incubator.apache.org/tutorials/gluon/save_load_params.html) for more information.
```
szha: I disagree with the assessment above.

ThomasDelteil: can you please pinpoint which part is not factual in your opinion?

safrooze: The tutorial that NEWS.md points to for "more information" only mentions save_params() and load_params().

NEWS.md
marcoabreu: It might be rephrased, but I agree that we need something with justification to express that this was a one-time-off (other frameworks make a lot of breaking changes with minor releases). Having a stable API gives us a lot trust and is often a major point for users. We had customer using 0.11 although 1.1 was released just because they were afraid of upgrading. Same for 1.0 users although 1.2 is released - we should make it our highest priority that we gain trust in our versioning and explain it in quite detail if we diverge from it. Ideally, we'd be like Apple and basically have everybody upgrade as soon as an update gets released. Otherwise, we'll always have to keep supporting old versions although we officially say that we don't.

NEWS.md
szha: For example, load_params was not reverted to the behavior of 1.1.0 in 1.2.1. I think Anirudh should already be aware of it.

anirudh2290: The statement talks about `save_params`, `load_params` in conjunction. I think the revert statement makes sense for our users.
From the user point of view, if they were using `save_params`/`load_params` to save their model architecture and params, to load it back later into Symbolblock, they know that the their codes which broke in 1.2 will not break in 1.2.1. They are also informed that they should use export and imports API. For users who used it to create layers without namescope in v1.2.0, they know that this feature didn't work with save_params and load_params in v1.1.0 so they will switch to save_parameters and load_parameters.

anirudh2290: @safrooze thanks for pointing! I have changed the link. Also opened a PR to fix the doc for master branch.

szha: @anirudh2290 I'd suggest to opt for correct description of the behavior (like you just did) over incorrect description of what has been done.

anirudh2290: @szha how are you suggesting that the text be changed ?

anirudh2290: ping @szha , @Roshrini has opened PR to cherry pick warning message changes : https://github.com/apache/incubator-mxnet/pull/11548. Can you please call out specific concerns with the release notes.

szha: I already did above. Please update.

anirudh2290: I have already explained that (`save_params`\`load_params`) is in conjunction and why the revert statement makes sense to our users.

szha: @anirudh2290 you asked for my specific concerns, so I explained that my concern is false statement, and I offered alternative that doesn't require false statement.

Seems weird to me that the justification for making false statement is because it "makes sense to our users".

anirudh2290: How is it false that the behavior of `save_params`/`load_params` API has been reverted to the behavior of 1.1.0 ? Does the line say that `load_params` API has been reverted to 1.1.0 ?

szha: The code for `load_params` (and its behavior) wasn't reverted to what's in 1.1.0, for the best compatibility.

Maybe I misunderstood what you were attempting to say with the distinction between "API has been reverted to the behavior of 1.1.0" v.s. "API has been reverted to 1.1.0". I was treating them as equivalent in this case. So please bear with me :)

My suggestion would be to provide explanation as faithful to the code as possible, so that the explanation would always be true regardless of the users' level of understanding. Besides that, you may add whatever you think would benefit the users as you see fit.

NEWS.md
```diff
@@ -2,7 +2,19 @@ MXNet Change Log
...
+2. If you created gluon layers without a `name_scope` using MXNet 1.2.0, you must replace `save_params` with `save_parameters`. Otherwise, your models saved in 1.2.1 will fail to load back, although this worked in 1.2.0.
+3. For the other use cases, such as models created within a `name_scope` (inside a `with name_scope()` block) or models being loaded back into gluon and not `SymbolBlock`, we strongly recommend replacing `save_params` and `load_params` with `save_parameters` and `load_parameters`. Having said that, your code won't break in 1.2.1 but will give you a deprecated warning message for `save_params` and `load_params`.
+
+### Incompatible API Changes
+- We are breaking semantic versioning by making a backwards incompatible change from 1.2.0 in the 1.2.1 patch release. The breaking use case is documented in point 2 above. The reason for doing this is because the 1.2.0 release broke a documented [use case](https://github.com/apache/incubator-mxnet/issues/11091) from the [gluon book](https://github.com/zackchase/mxnet-the-straight-dope/blob/master/chapter07_distributed-learning/hybridize.ipynb) and this release reverts the breakage.
```
szha: Since we're making an inconvenience to the users, it's probably a good idea to state that "We sincerely apologize for the inconvenience, and we as a community will continue to uphold semantic versioning."

srochel: Suggest: 
We did break the promise made with semantic versioning due to the API behavior change in 1.2.0 and the backward incompatible change between 1.2.0 and 1.2.1 patch release. The breaking use case is documented in point 2 above. The reason for doing this is because the 1.2.0 release broke a documented [use case](https://github.com/apache/incubator-mxnet/issues/11091) from the [gluon book](https://github.com/zackchase/mxnet-the-straight-dope/blob/master/chapter07_distributed-learning/hybridize.ipynb) and this release reverts the breakage. As a community we are apologizing for the inconvenience caused and will continue to strive to uphold semantic versioning.

anirudh2290: i have made the change.

PR #11482, opened 2018-06-29
make gluon rnn layers hybrid blocks

szha:
## Description ##
make gluon rnn layers hybrid blocks

### Essentials ###

### Changes ###
- [x] make gluon rnn layers hybrid blocks


python/mxnet/gluon/rnn/rnn_layer.py
```diff
@@ -173,67 +177,63 @@ def begin_state(self, batch_size=0, func=ndarray.zeros, **kwargs):
             states.append(func(name='%sh0_%d'%(self.prefix, i), **info))
         return states
 
-    def forward(self, inputs, states=None):
-        batch_size = inputs.shape[self._layout.find('N')]
+    def hybrid_forward(self, F, inputs, states=None, **kwargs):
+        if F is ndarray:
+            batch_size = inputs.shape[self._layout.find('N')]
+            if self._input_size == 0:
```
piiswrong: implement infershape instead?

python/mxnet/gluon/rnn/rnn_layer.py
```diff
@@ -173,67 +177,63 @@ def begin_state(self, batch_size=0, func=ndarray.zeros, **kwargs):
...
+                    self.i2h_weight[i]._finish_deferred_init()
         skip_states = states is None
         if skip_states:
-            states = self.begin_state(batch_size, ctx=inputs.context)
-        if isinstance(states, ndarray.NDArray):
+            if F is ndarray:
+                states = self.begin_state(batch_size, ctx=inputs.context)
+            else:
+                states = self.begin_state(0, func=symbol.zeros)
+        if isinstance(states, (ndarray.NDArray, symbol.Symbol)):
```
piiswrong: (ndarray.NDArray, symbol.Symbol) -> tensor_types

python/mxnet/gluon/rnn/rnn_layer.py
```diff
@@ -173,67 +177,63 @@ def begin_state(self, batch_size=0, func=ndarray.zeros, **kwargs):
...
-            layout=self._layout, merge_outputs=True)
-        new_states = []
-        for i in range(ns):
-            state = ndarray.concat(*(j.reshape((1,)+j.shape) for j in states[i::ns]), dim=0)
-            new_states.append(state)
-
-        return outputs, new_states
-
-    def _forward_kernel(self, inputs, states):
+    def __call__(self, inputs, *states):
```
piiswrong: implement infer_shape instead?

python/mxnet/gluon/rnn/rnn_layer.py
szha: this is not possible due to the inverse shape inference in concat.

python/mxnet/gluon/rnn/rnn_layer.py
szha: this is not possible due to the inverse shape inference in concat.

python/mxnet/gluon/rnn/rnn_layer.py
szha: or did you mean overriding block's infer shape? I'm taking a look

szha:
@piiswrong turns out I cannot do that only in infer_shape. The reason is that sometimes the block is used as a child block of other blocks, in which case the infer shape is called from parent, thus bypassing the code path in rnn infer shape.

piiswrong:
I see. Then could you do it without overriding `__call__`?

szha:
I can override forward but it would be pretty much equivalent. The reason I cannot do this in hybrid_forward is that when given partial shape, hybrid_forward would only be invoked with symbols as part of the infer_shape pass.

src/operator/tensor/matrix_op-inl.h
```diff
@@ -182,14 +205,16 @@ inline bool ReshapeShape(const nnvm::NodeAttrs& attrs,
       oshape[inf_idx] = dshape.Size() / oshape.Size();
     }
   } else {
-    return (*out_attrs)[0].ndim();
+    if ((*out_attrs)[0].ndim()) {
+      return ReverseReshapeInferShape(&(*in_attrs)[0], oshape);
+    }
```
haojin2: Should this be
```c++
    if ((*out_attrs)[0].ndim()) {
      return ReverseReshapeInferShape(&(*in_attrs)[0], oshape);
    }
    return 0;
```
instead?

anirudh2290:
the test_norm failures seem to be because of rtol and atol not being set in assert_almost_equal calls. seems to be unrelated to the other failures. https://github.com/apache/incubator-mxnet/blob/master/tests/python/unittest/test_ndarray.py#L1324

anirudh2290:
The other failure seems to be due to `test_operator_gpu.test_kernel_error_checking`. I see that this fails on other stages like MKLDNN GPU to and not specific to the NOCUDNN stage. http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/PR-11482/26/pipeline/

tests/python/gpu/test_gluon_gpu.py
```diff
@@ -0,0 +1,203 @@
...
+from mxnet.base import MXNetError
+from mxnet import autograd
+from numpy.testing import assert_allclose
+
+curr_path = os.path.dirname(os.path.abspath(os.path.expanduser(__file__)))
+sys.path.insert(0, os.path.join(curr_path, '../unittest'))
+from common import setup_module, with_seed, teardown, assert_raises_cudnn_disabled
+from test_gluon import *
+from test_loss import *
+from test_gluon_rnn import *
```
anirudh2290: test_gluon_rnn has assert_raises_cudnn_disabled for `test_rnn`. test_rnn doesnt raise an error though. I think it should be removed from test_gluon_rnn.py

Roshrini:
@szha today is the mentioned code freeze date for MXNet 1.3 release. Could you please check the status of this PR? thanks!

haojin2:
@Roshrini We've found the root cause of the bug, the temporary work-around is the one being built and tested now.

szha:
@Roshrini this PR uncovers a weird bug which should probably be addressed before releasing.

tests/python/unittest/test_gluon_rnn.py
```diff
@@ -502,7 +498,6 @@ def test_cell_fill_shape():
 @assert_raises_cudnn_disabled()
 def test_layer_fill_shape():
     layer = gluon.rnn.LSTM(10)
-    layer.hybridize()
```
anirudh2290: Removing this hybridize call should be enough. is there any reason we are removing all hybridize calls ?

szha: The same error on test_norm still occurs if I only remove this one call, so I was trying out all possible cases

Roshrini:
@szha thanks for the update.

src/operator/nn/concat.cc
```diff
@@ -320,5 +375,41 @@ NNVM_REGISTER_OP(_backward_Concat)
 #endif
 .set_attr<FCompute>("FCompute<cpu>", ConcatGradCompute<cpu>);
 
+
+NNVM_REGISTER_OP(_rnn_param_concat)
```
zheng-da: Please write comments for this operator. How is it different from the normal concat.

tests/python/unittest/test_gluon_rnn.py
```diff
@@ -25,7 +25,6 @@
 from common import assert_raises_cudnn_disabled
 
 
-@assert_raises_cudnn_disabled()
```
zheng-da: why do you remove this?

tests/python/unittest/test_gluon_rnn.py
```diff
@@ -490,7 +505,7 @@ def test_layer_fill_shape():
     layer.hybridize()
     check_rnn_layer_forward(layer, mx.nd.ones((3, 2, 7)))
     print(layer)
-    assert layer.i2h_weight[0].shape[1] == 7, layer.i2h_weight[0].shape[1]
+    assert layer.l0_i2h_weight.shape[1] == 7, layer.l0_i2h_weight.shape[1]
```
zheng-da: is this API change?

tests/python/unittest/test_gluon_rnn.py
szha: It’s not calling cudnn kernel

src/operator/nn/concat.cc
szha: This is a custom concat op with specialized infer_shape, which handles the case where the first one or two inputs may have unknown shape that can be inferred from output shape.

tests/python/unittest/test_gluon_rnn.py
szha: This has not been exposed as a documented attribute before and shouldn’t be considered part of API

PR #11492, opened 2018-06-29
[MXNET-612] Optimize RAT setup and move it into PR stage

marcoabreu:
## Description ##
At the moment, Apache RAT is being installed during runtime. This is not time-efficient and will cause future problems if we move the RAT check into the sanity check.

Now, it will be part of a Dockerfile and thus cached. Execution time is about 20 seconds.

Also, we're moving the Apache RAT check to the PR pipeline instead of running it nightly.

### Essentials ###

### Changes ###
- Moved Apache RAT installation into Dockerfile
- Apache RAT check is now part of PR pipeline

### DIFF
--- a/ci/docker/install/ubuntu_rat.sh
+++ b/ci/docker/install/ubuntu_rat.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
 
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
@@ -17,8 +17,11 @@
 # specific language governing permissions and limitations
 # under the License.
 
+set -ex
 
-#mvn and svn are installed in the docker container via the nghtly test script
+echo "Install dependencies"
+apt-get update
+apt-get install -y subversion maven openjdk-8-jdk openjdk-8-jre
 
 echo "download RAT"
 svn co http://svn.apache.org/repos/asf/creadur/rat/trunk/ #>/dev/null
@@ -31,21 +34,3 @@ mvn -Dmaven.test.skip=true install #>/dev/null
 
 echo "build success, cd into target"
 cd apache-rat/target
-
-
-echo "-------Run Apache RAT check on MXNet-------"
-
-#Command has been run twice, once for the logs and once to store in the variable to parse.
-java -jar apache-rat-0.13-SNAPSHOT.jar -E /work/mxnet/tests/nightly/apache_rat_license_check/rat-excludes -d /work/mxnet
-OUTPUT="$(java -jar apache-rat-0.13-SNAPSHOT.jar -E /work/mxnet/tests/nightly/apache_rat_license_check/rat-excludes -d /work/mxnet)"
-SOURCE="^0 Unknown Licenses"
-
-
-echo "-------Process The Output-------"
-
-if [[ "$OUTPUT" =~ $SOURCE ]]; then
-      echo "SUCCESS: There are no files with an Unknown License.";
-else
-      echo "ERROR: RAT Check detected files with unknown licenses. Please fix and run test again!";
-      exit 1
-fi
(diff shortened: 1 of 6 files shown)


ci/docker/install/ubuntu_rat.sh
```diff
@@ -30,22 +33,4 @@ echo "mvn install"
...
-echo "-------Process The Output-------"
-
-if [[ "$OUTPUT" =~ $SOURCE ]]; then
-      echo "SUCCESS: There are no files with an Unknown License.";
-else
-      echo "ERROR: RAT Check detected files with unknown licenses. Please fix and run test again!";
-      exit 1
-fi
\ No newline at end of file
+cd apache-rat/target
```
KellenSunderland: newline

KellenSunderland: And why do we cd at the end of an installation shell script?

ci/docker/runtime_functions.sh
```diff
@@ -784,7 +784,25 @@ build_docs() {
 #Runs Apache RAT Check on MXNet Source for License Headers
 nightly_test_rat_check() {
     set -ex
-    ./tests/nightly/apache_rat_license_check/license_check.sh
+    pushd .
+    
+    cd /work/deps/trunk/apache-rat/target
+
+    #Command has been run twice, once for the logs and once to store in the variable to parse.
```
KellenSunderland: Not sure what this means.  Which logs are we referring to? Which variables?

tests/nightly/apache_rat_license_check/README.md
```diff
@@ -12,6 +12,12 @@ The license check script called by the Jenkinsfile
 ### How to run the RAT check locally
 The following commands can be used to run a Apache RAT check locally - 
 
+Docker based 1-click-method:
```
KellenSunderland: 1-click mechanisms are our specialty!

ci/docker/install/ubuntu_rat.sh
marcoabreu: Sorry, was copy&paste. Removed

ci/docker/runtime_functions.sh
marcoabreu: Yeah right, we could actually just run it and store the output in a variable as well as print it to the log at the same time.

marcoabreu:
@mbaijal

ci/docker/runtime_functions.sh
```diff
@@ -783,8 +783,26 @@ build_docs() {
...
-    set -ex
-    ./tests/nightly/apache_rat_license_check/license_check.sh
+    set -e
+    pushd .
+    
+    cd /work/deps/trunk/apache-rat/target
+
+    # Use shell number 5 to duplicate the log output. It get sprinted and stored in $OUTPUT at the same time https://stackoverflow.com/a/12451419
+    exec 5>&1
+    OUTPUT=$(java -jar apache-rat-0.13-SNAPSHOT.jar -E /work/mxnet/tests/nightly/apache_rat_license_check/rat-excludes -d /work/mxnet|tee >(cat - >&5))
```
anirudh2290: This may be unrelated but what is file descriptor 5 here. I tried on my machine and it gives bad file descriptor. And when I echo $OUTPUT it says : `Ignored 1 lines in your exclusion files as comments or empty lines.` Maybe its just my env, but it is this rat check verified ?

marcoabreu: It's basically just an empty shell number. 1 is stdout, 2 is stderr. I think the other ones are unused. This is basically a hack to mirror the output and store it in a variable https://stackoverflow.com/a/12451419f

Exactly. $OUTPUT will contain the output of the RAT check. If there are no problems, it will output what you have seen. If there are any problems, it will contain details and always print the string which is stored in ERROR_MESSAGE. We determine whether the run was successful by checking if the output contains that error message.

anirudh2290: I have unknown licenses in my repo and it contains the ERROR_MESSAGE. But if it is verified on your end I am okay with this.

Issue #11413, opened 2018-06-27
MXPredReshape


nihilityworld:
I have used the MXPredCreate to create a hanlder out use input shape [1,3,12,12]. And I use the MXPredReshape to change the input shape  to [1,3,430,360] and it return 0. But the input shape of the new hander out2 did not changed. Is it a bug?

frankfliu:
Hi @nicklhy thanks for the question, @sandeep-krishnamurthy requesting this be labeled under Question

frankfliu:
For questions, please submit on MXNet discussion forum (https://discuss.mxnet.io), where it will get a wider audience and allow other to learn as well.
@sandeep-krishnamurthy to close this issue now in favor of the discussion forum issue you will file, please feel free to re-open if closed in error.
Thanks!

Issue #10937, opened 2018-05-14
c++ API, fixed batch size for predictor. 


ashkanaev:
Hello!
Is't possible to make predictor with c++ API for arbitrary batch size? 
From  https://github.com/apache/incubator-mxnet/blob/97570916f844bcb4515d972c75fb0a75da345d97/example/image-classification/predict-cpp/image-classification-predict.cc

` const mx_uint input_shape_data[4] = { 1,
                                        static_cast<mx_uint>(channels),
                                        static_cast<mx_uint>(height),
                                        static_cast<mx_uint>(width) };
`

`MXPredCreate(static_cast<const char*>(json_data.GetBuffer()),
               static_cast<const char*>(param_data.GetBuffer()),
               static_cast<int>(param_data.GetLength()),
               dev_type,
               dev_id,
               num_input_nodes,
               input_keys,
               input_shape_indptr,
               input_shape_data,
               &pred_hnd);`

look's like predictor hard linked to fixed input data shape. how to make it flexible?

vdantu:
@ashkanaev : You could post this question on discuss.mxnet.io to get a wider audience.  Also, are you seeing errors when you modify the predictor or is it not adhering to your modified input shapes? It would be helpful to know what you mean by making it flexible. 

@sandeep-krishnamurthy : Could you tag this with "C++", "Example" and "Question".

chinakook:
MXPredReshape

RedbackThomson:
I am calling MXPredReshape with an updated input_shape_data (only changing the batch size), but when I do, the MXPredSetInput tells me my data size is larger than the expected shape. I am using the new handle, and even freeing the old handle. Can someone check to see if this is a bug in c_predict_api.cc or if I'm somehow messing it up.

lupesko:
@ashkanaev thank you for submitting this issue!
This is an issue with the MXNet C Predict API. Let's please update the label to "C API"

@nswamy can you please update the label?

ThomasDelteil:
@leleamol could you pick this up ? Thanks!

ou525:
@ashkanaev  Is this problem solved? How to achieve it?

Issue #10502, opened 2018-04-11
MXPredReshape


crqcrq001:
When I Use MXPredCreate create a GPU handle, then use MXPredReshape to reshape the input shape,
it failure. 
This may cause by default context is cpu .
in MXPredReshape ()
.....
// bind line 324
  {
    std::map<std::string, Context> ctx_map;
    std::vector<NDArray> grad_store;
    grad_store.reserve(ret->arg_arrays.size());
    std::vector<OpReqType> grad_req(ret->arg_arrays.size(), kNullOp);

    ret->exec.reset(Executor::Bind(ret->sym, ret->ctx, ctx_map,
                                   ret->arg_arrays,
                                   grad_store, grad_req,
                                   ret->aux_arrays,
                                   p->exec.get()));
       // here I think it reset the executor by ret->ctx (but in MXPredCreate ret->ctx is not assignment)
    ret->out_shapes = out_shapes;
    ret->out_arrays = ret->exec->outputs();
  }  
So I change MXPredCreatePartialOut()
....
 // bind line 234
  {
    std::map<std::string, Context> ctx_map;
    std::vector<NDArray> grad_store(arg_arrays.size());
    std::vector<OpReqType> grad_req(arg_arrays.size(), kNullOp);

    ret->exec.reset(Executor::Bind(sym, ctx, ctx_map,
                                   arg_arrays,
                                   grad_store, grad_req,
                                   aux_arrays));
    ret->ctx=ctx;//here
    ret->out_shapes = out_shapes;
    ret->out_arrays = ret->exec->outputs();
  }

Is this correct?

crqcrq001:
and missing ret->sym=sym

spidyDev:
@eric-haibin-lin  Please label it as: Question

harshp8l:
@mxnet-lablel-bot update [question, c api]

PR #9984, opened 2018-03-04
[MXNET-38]add reshape predicator function to c_predict_api

Ldpe2G:
## Description ##

add a function to c_predict_api  which can change the input shape of an already created predicator .
related issue: #5360

### Essentials ###
- [ ] Passed code style checking (`make lint`)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #11493, opened 2018-06-29
Fix MXPredReshape in the c_predict_api

hqucms:
## Description ##

This PR fixes the `MXPredReshape` implementation in the c_predict_api. 
In https://github.com/apache/incubator-mxnet/pull/9984, `aux_arrays`, `sym`, and `ctx` members were introduced in `struct MXAPIPredictor` but not assigned in `MXPredCreatePartialOut`, 
thus the `PredictorHandle` created by `MXPredCreatePartialOut` cannot be reshaped in `MXPredReshape`.
This was pointed out in https://github.com/apache/incubator-mxnet/issues/10502, and likely also related to https://github.com/apache/incubator-mxnet/issues/11413, https://github.com/apache/incubator-mxnet/issues/10937.

Also removed the clean-up of the original handle,
```
  p->arg_arrays.clear();
  ...
  p->aux_arrays.clear();
```
in `MXPredReshape`, as it makes the original handle invalid. I think it would be better to leave the freedom to the user that whether they want to free the old handle after reshaping, or to free the reshaped new handle after use and revive the old one.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/amalgamation/python/mxnet_predict.py
+++ b/amalgamation/python/mxnet_predict.py
@@ -51,17 +51,17 @@ def c_array(ctype, values):
 def _find_lib_path():
     """Find mxnet library."""
     curr_path = os.path.dirname(os.path.abspath(os.path.expanduser(__file__)))
-    api_path = os.path.join(curr_path, '..', '..', 'lib')
-    dll_path = [curr_path, api_path]
-    dll_path = [os.path.join(p, 'libmxnet.so') for p in dll_path] + \
-        [os.path.join(p, 'libmxnet_predict.so') for p in dll_path] + \
-        [os.path.join(p, 'libmxnet.dll') for p in dll_path] + \
-        [os.path.join(p, 'libmxnet_predict.dll') for p in dll_path]
-    lib_path = [p for p in dll_path if os.path.exists(p) and os.path.isfile(p)]
-    if len(lib_path) == 0:
-        raise RuntimeError('Cannot find the files.\n' +
-                           'List of candidates:\n' + str('\n'.join(dll_path)))
-    return lib_path
+    amalgamation_lib_path = os.path.join(curr_path, '../../lib/libmxnet_predict.so')
+    if os.path.exists(amalgamation_lib_path) and os.path.isfile(amalgamation_lib_path):
+        lib_path = [amalgamation_lib_path]
+        return lib_path
+    else:
+        # from python/setup.py
+        libinfo_py = os.path.join(curr_path, '../../python/mxnet/libinfo.py')
+        libinfo = {'__file__': libinfo_py}
+        exec(compile(open(libinfo_py, "rb").read(), libinfo_py, 'exec'), libinfo, libinfo)
+        lib_path = libinfo['find_lib_path']()
+        return lib_path
 
 
 def _load_lib():
--- a/amalgamation/python/mxnet_predict.py
+++ b/amalgamation/python/mxnet_predict.py
@@ -26,6 +26,7 @@ from __future__ import absolute_import
 import os
 import sys
 import ctypes
+import logging
 import numpy as np
 
 __all__ = ["Predictor", "load_ndarray_file"]
@@ -56,6 +57,7 @@ def _find_lib_path():
         lib_path = [amalgamation_lib_path]
         return lib_path
     else:
+        logging.info('Cannot find libmxnet_predict.so. Will search for MXNet library using libinfo.py then.')
         try:
             from mxnet.libinfo import find_lib_path
             lib_path = find_lib_path()
@@ -68,7 +70,7 @@ def _find_lib_path():
                 lib_path = libinfo['find_lib_path']()
                 return lib_path
             else:
-                raise RuntimeError('Cannot find libinfo.py at %s.\n' % libinfo_path)
+                raise RuntimeError('Cannot find libinfo.py at %s.' % libinfo_path)
 
 
 def _load_lib():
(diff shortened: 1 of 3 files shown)


hqucms:
@szha Any suggestions where the tests should be added? I am not very familiar with the organization of the test units.

reminisce:
I think testing this API set has been missing in MXNet code base. The frontend only adopts this in amalgamation. https://github.com/apache/incubator-mxnet/blob/23f973b7523c033dd8cd18eb7a9ae2bda32d43c0/amalgamation/python/mxnet_predict.py

You can add a file: `tests/python/unittest/test_predictor.py` for adding unit tests.

hqucms:
Thanks for your help, @reminisce. I added a unit test following your suggestion.
@szha Please let me know if this looks good to you.

amalgamation/python/mxnet_predict.py
```diff
@@ -51,15 +51,24 @@ def c_array(ctype, values):
...
-    lib_path = [p for p in dll_path if os.path.exists(p) and os.path.isfile(p)]
-    if len(lib_path) == 0:
-        raise RuntimeError('Cannot find the files.\n' +
-                           'List of candidates:\n' + str('\n'.join(dll_path)))
-    return lib_path
+    amalgamation_lib_path = os.path.join(curr_path, '../../lib/libmxnet_predict.so')
+    if os.path.exists(amalgamation_lib_path) and os.path.isfile(amalgamation_lib_path):
+        lib_path = [amalgamation_lib_path]
+        return lib_path
+    else:
```
apeforest: Add a log message here that indicates libmxnet_predict.so file is not found and MXNet will search for libinfo.py

src/c_api/c_predict_api.cc
```diff
@@ -309,7 +312,6 @@ int MXPredReshape(mx_uint num_input_nodes,
         << " shape has been changed, only allow to change the shape of input data.";
     }
   }
-  p->arg_arrays.clear();
```
apeforest: Excellent change. This follows a good design paradigm.

amalgamation/python/mxnet_predict.py
hqucms: @apeforest 
Done: https://github.com/apache/incubator-mxnet/pull/11493/commits/f3d70681421e627d7ad8481fb6254ede2190d945

hqucms:
> We can also get something useful from #10882

@chinakook I think that would be a good idea, but maybe we should get this one merged ASAP since it is supposed to be a hotfix.

tests/python/unittest/test_predictor.py
```diff
@@ -0,0 +1,87 @@
...
+    a = nd.random.uniform(shape=(7, 3))
+    b = nd.random.uniform(shape=(7,))
+    nd_data = {'a':a, 'b':b}
+    nd.save(nd_file, nd_data)
+
+    # test load_ndarray_file
+    nd_load = load_ndarray_file(open(nd_file, "rb").read())
+    assert(set(nd_data.keys()) == set(nd_load.keys()))
+    for k in nd_data.keys():
+        assert_almost_equal(nd_data[k].asnumpy(), nd_load[k])
```
anirudh2290: nit: can we set specific rtol and atol values here to avoid flakiness.

tests/python/unittest/test_predictor.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # create a predictor
+    predictor = Predictor(open(symbol_file, "r").read(),
+                      open(param_file, "rb").read(),
+                      {'data':input1.shape})
+
+    # forward and get output
+    predictor.forward(data=input1)
+    predictor_out1 = predictor.get_output(0)
+    assert_almost_equal(out1.asnumpy(), predictor_out1)
```
anirudh2290: nit: can we set specific rtol and atol values here to avoid flakiness

marcoabreu:
Hi @hqucms, sorry for the late response. Would you mind addressing the review? We're ready to merge then.

safrooze:
@sandeep-krishnamurthy This is an important fix that we need in our project. Given that the code is already using `assert_almost_equal()`, would you be OK with merging this as is? I want to make sure it is included in 1.3.0 release.

marcoabreu:
The default values of almost equals are too low and could result in flaky behavior. We should address it first

safrooze:
@marcoabreu If the default for `assert_almost_equal()` which is part of `mxnet` is too flaky for mxnet tests, then I'd argue that the defaults should be changed, not every place that it's used.

marcoabreu:
The default is atol=e-20 and rtol=e-5. I don't know by how much we can expect a derivation here.

I think the default values are reasonable. The only case when they can cause problems is when we are working with randomized input data. Since we got both cases, I'd prefer to not increase the default tolerances but instead rather have them increased specifically if we use random data.

safrooze:
What tolerance values are recommended for random data in mxnet?

hqucms:
@marcoabreu @anirudh2290 Sorry for the delay in response. I change to `rtol=1e-5, atol=1e-6` which are the ones used in https://github.com/apache/incubator-mxnet/blob/master/tests/python/mkl/test_mkldnn.py#L132. Please let me know if they are OK.

marcoabreu:
No worries, thanks for your contribution and addressing the comments. Seems like you hit a flaky test. Please just make an empty commit to trigger a new ci run.

hqucms:
@marcoabreu The CI test finally succeeded. Please let me know if everything looks OK to you.

marcoabreu:
Looks good, thank you.

Unfortunately, I don't have time to review his pr right now. Could somebody else take it over please?

johnbroughton2017:
Hi @hqucms @maybepossible ,

I think I am still having the problem mentioned here. Did you guys manage to make it work with MXPredReshape? Would you mind having a quick look at my problem? Thanks a lot in advance.
https://discuss.mxnet.io/t/mxpredreshape-always-fails-return-1-mxnet-1-3-0-c/2003/3

-- J

Issue #11448, opened 2018-06-28
fail to fall back when sparse arrays are passed to MKLDNN-enabled operators.


zheng-da:
Currently, the MKLDNN-enabled operators, such as convolution and pooling, can't handle sparse arrays correctly. The reason is that the storage inference of these operators doesn't return the right dispatch mode.
The MKLDNN-enabled operators include:
- [ ] Activation @azai91 
- [x] BatchNorm @zheng-da
- [ ] Convolution @ZhennanQin  
- [ ] Deconvolution @ZhennanQin 
- [x] FullyConnected @haojin2 
- [ ] LRN @azai91 
- [ ] Pooling @luobao-intel
- [ ] Softmax @luobao-intel

We may also need to test the operators below:
- [x] Concat
- [ ] Copy
- [ ] Sum
- [ ] Flatten

@haojin2 @azai91 @pengzhao-intel @TaoLv

pengzhao-intel:
@zheng-da Could you help provide a test case?  
Does cuDNN backend handle these cases?

frankfliu:
Thank you for submitting the issue!  @sandeep-krishnamurthy requesting this be labeled.

zheng-da:
@eric-haibin-lin  @haojin2  could you please provide a test case?

eric-haibin-lin:
There’s an existing test batch norm training test disabled due to flakyness

eric-haibin-lin:
concat is tested and fixed.

haojin2:
I'll take FullyConnected if no one's working on it right now.

zheng-da:
i created a PR for batchnorm.

pengzhao-intel:
what else need to cover? @zouluobao can help

pengzhao-intel:
@luobao-intel will handle other OPs :)

zheng-da:
@pengzhao-intel @luobao-intel we need to fix them and merge the PRs to the v1.3 release, which is probably before the end of next week.

haojin2:
@zheng-da fullyconnected is done in #11498, if you want to add the MKLDNN backward storage type inference for it please go ahead and do so. Thanks!

zheng-da:
@haojin2 I thought you have fixed its backward storage type inference.

haojin2:
@zheng-da checkout https://github.com/apache/incubator-mxnet/pull/11498/files#diff-f2984a387afd7a02b74fe43b9ef50ee1R214

[...]

PR #11301, opened 2018-06-15
MKLDNN Backward op cache

ZhennanQin:
Hi all,

For MKLDNN, creating operator primitive and corresponding memory will spend a lot of time, for many scenarios,  the created primitive can be reused if it meets requirement for next computing. In this PR, we implemented the caching mechanism for most backward operators, just like the optimization have done for forward operator.  Correctness test and accuracy test are all PASS. Performance test shown that most models can get speed up in training. Please review this, Thanks. 

@zheng-da, @marcoabreu, @azai91, @TaoLv, @pengzhao-intel

This PR covers below commits.
Enable primitive allocation cache for _backward_LRN
Enable primitive allocation cache for _backward_Pooling.
Enable primitive allocation cache for _backward_Activation.
Enable primitive allocation cache for _backward_Deconvolution.
Enable primitive allocation cache for _backward_BatchNorm.
Enable primitive allocation cache for _backward_Fully_Connected.
Enable primitive allocation cache for _backward_Convolution.

**Correntness test result:**

unit_test - PASS

Train model with dummy data when MXNET_MKLDNN_DEBUG=1

model | Status
-- | --
alexnet | PASS
googlenet | PASS
vgg-16 | PASS
vgg-19 | PASS
inception-bn | PASS
inception-v3 | PASS
resnet-50 | PASS
resnet-152 | PASS

**Accuracy test result:**

CiFAR10 + ResNet50 ( Convergence ): Validation accuracy of 99 epochs ( top1: 67.88%, top5: 96.26%)

CiFAR10 + VGG16 ( Convergence ): Validation accuracy of   74 epochs ( top1: 82.56%, top5: 98.52% )

**Performance test result:**

**Skylake 8180 2sockets training BS=32 (img/s)**

Model | Baseline | After Caching Backward Op | Diff%
-- | -- | -- | --
alexnet | 366.0699 | 412.3794 | 112.65%
googlenet | 104.3128 | 152.8598 | 146.54%
vgg-16 | 38.40072 | 39.65964 | 103.28%
vgg-19 | 32.35834 | 33.80966 | 104.49%
inception-bn | 85.71482 | 124.4676 | 145.21%
inception-v3 | 43.0482 | 57.82697 | 134.33%
resnet-50 | 53.15957 | 65.68513 | 123.56%
resnet-152 | 21.67982 | 28.26535 | 130.38%
inception-v4 | 23.82474 | 29.33314 | 123.12%

**Total Time consumption breakdown(ms)**

[...]

PR #11498, opened 2018-06-29
Fix InferStorage for sparse fallback in FullyConnected

haojin2:
## Description ##
#11448

### Essentials ###

### Changes ###
- [x] Fix storage fallback in backward

### DIFF
--- a/src/operator/nn/fully_connected.cc
+++ b/src/operator/nn/fully_connected.cc
@@ -210,17 +210,21 @@ inline static bool BackwardFCStorageType(const nnvm::NodeAttrs& attrs,
   CHECK_EQ(in_attrs->size(), 3U);
   CHECK_EQ(out_attrs->size(), out_expected);
 
-  DispatchMode wanted_mode;
-#if 0
+  bool dispatched = false;
   // TODO(zhengda) let's disable MKLDNN for FullyConnected for now.
   // It seems there is a bug.
-  if (dev_mask == mshadow::cpu::kDevMask)
-    *dispatch_mode = DispatchMode::kFComputeEx;
-  else
-#endif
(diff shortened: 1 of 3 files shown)


haojin2:
@eric-haibin-lin

src/operator/nn/fully_connected.cc
```diff
@@ -210,17 +210,17 @@ inline static bool BackwardFCStorageType(const nnvm::NodeAttrs& attrs,
   CHECK_EQ(in_attrs->size(), 3U);
   CHECK_EQ(out_attrs->size(), out_expected);
 
-  DispatchMode wanted_mode;
-#if 0
+  bool dispatched = false;
   // TODO(zhengda) let's disable MKLDNN for FullyConnected for now.
```
eric-haibin-lin: @zheng-da @azai91  is there a plan to enable FC with MKLDNN, or at least test it out? If there's a bug reveal by this test, this will also help stabilize mxnet-MKLDNN.

zheng-da: ideally we should enable it. i forgot the case that it fails. i don't think MKLDNN FC backward is much faster than our native implementation. so it's not very urgent.
@pengzhao-intel @TaoLv did you check if MKLDNN FC backward works in all cases?

pengzhao-intel: We have enabled this in the #11301 and it passed all unit tests and coverage test in our local.
So, I think MKLDNN FC backward works fine.

zheng-da:
@haojin2 could you write a unit test?

haojin2:
@zheng-da unit test added.

haojin2:
@eric-haibin-lin @zheng-da

PR #11503, opened 2018-06-30
updating installation info to have latest packages and more clarity

aaronmarkham:
## Description ##

There hasn't been an update on installation instructions for Ubuntu in a while. Many new packages are out and not referenced, so I'm trying to fix that.

- Fixed a lot of old info in the Ubuntu instructions.
- Added a table of different pip packages and what versions they support.
- Added recommended installations.
- Added scripts for quick installation.
- Reorganized sections for clarity.
- Untangled some R instructions - these were sort of inter-mingled with the standard instructions.

## Comments ##
- I didn't mark mkl as experimental. LMK if I should change that.
- I made an assumption that I don't want to show CUDA 9.1 packages... since NVIDIA isn't even offering it...
- I assume that CUDA 9.2 and cuDNN 7.1.4 is what is desired.
- I assume that `pip install mxnet-cu92mkl` is recommended for inference.
- I assume that `pip install mxnet-cu92` is recommended for training.
- Once any comments settle down on this PR, I can apply these kind of changes on macOS.
- I didn't try to vet the standard install instructions; this section might need updating, and I would like to see it simplified.

### DIFF
--- a/docs/install/ubuntu_setup.md
+++ b/docs/install/ubuntu_setup.md
@@ -76,10 +76,11 @@ Alternatively, you can use the table below to select the package that suits your
 
 
 #### pip Package Availability
+
 The following table presents the pip packages that are recommended for each version of MXNet.
 
-| | 1.2.1 | 1.1.0 | 1.0.0 | 0.12.1 | 0.11.0 |
-|-|-|-|-|-|
+| Package / MXNet Version | 1.2.1 | 1.1.0 | 1.0.0 | 0.12.1 | 0.11.0 |
+|-|-|-|-|-|-|
 | mxnet-cu92mkl | :white_check_mark: | :x: | :x: | :x: | :x: |
 | mxnet-cu92 | :white_check_mark: | :x: | :x: | :x: | :x: |
 | mxnet-cu90mkl | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :x: |
--- a/docs/install/ubuntu_setup.md
+++ b/docs/install/ubuntu_setup.md
@@ -1,38 +1,189 @@
 # Installing MXNet on Ubuntu
 
-**NOTE:** For MXNet with Python installation, please refer to the [new install guide](http://mxnet.io/install/index.html).
+The following installation instructions are for installing MXNet on computers running **Ubuntu 12 to Ubuntu 16.04**. Support for later versions of Ubuntu is [not yet available](#contributions).
+<hr>
 
-MXNet currently supports Python, R, Julia, Scala, and Perl. For users of R on Ubuntu operating systems, MXNet provides a set of Git Bash scripts that installs all of the required MXNet dependencies and the MXNet library.
+## Contents
 
-The simple installation scripts set up MXNet for Python and R on computers running Ubuntu 12 or later. The scripts install MXNet in your home folder ```~/mxnet```.
+* [CUDA Dependencies](#cuda-dependencies)
+* [Quick Installation](#quick-installation)
+    * [Python](#install-mxnet-for-python)
+    * [pip Packages](#pip-package-availability)
+* [Standard Installation](#standard-installation)
+* [Installing Language Packages](#installing-language-packages-for-mxnet)
+    * [R](#install-the-mxnet-package-for-r)
+    * [Julia](#install-the-mxnet-package-for-julia)
+    * [Scala](#install-the-mxnet-package-for-scala)
+    * [Perl](#install-the-mxnet-package-for-perl)
+  * [Contributions](#contributions)
+  * [Next Steps](#next-steps)
 
-## Prepare environment for GPU Installation
+<hr>
 
-If you plan to build with GPU, you need to set up the environment for CUDA and CUDNN.
+## CUDA Dependencies
 
-First, download and install [CUDA 8 toolkit](https://developer.nvidia.com/cuda-toolkit).
+If you plan to build with GPU, you need to set up the environment for CUDA and cuDNN.
 
-Then download [cudnn 6](https://developer.nvidia.com/cudnn).
+First, download and install [CUDA toolkit](https://developer.nvidia.com/cuda-toolkit). CUDA 9.2 is recommended.
 
-Unzip the file and change to the cudnn root directory. Move the header and libraries to your local CUDA Toolkit folder:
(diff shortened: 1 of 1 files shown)


aaronmarkham:
@nswamy @marcoabreu @szha - Can you please review?

aaronmarkham:
The table is rendering incorrectly, so I'll need to fix that first. It is supposed to look like this:

aaronmarkham:
@nswamy - updated Ubuntu instructions as discussed (noting that >16.04 doesn't work).

aaronmarkham:
@marcoabreu any comments on this PR?

docs/install/ubuntu_setup.md
```diff
@@ -1,67 +1,103 @@
 # Installing MXNet on Ubuntu
 
-**NOTE:** For MXNet with Python installation, please refer to the [new install guide](http://mxnet.io/install/index.html).
+The following installation instructions are for installing MXNet on computers running **Ubuntu 12 to Ubuntu 16.04**. Support for later versions of Ubuntu is [not yet available](#contributions).
```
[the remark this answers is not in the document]
aaronmarkham: So should I remove that Ubuntu 12 part?

docs/install/ubuntu_setup.md
```diff
@@ -1,67 +1,103 @@
...
+## Contents
 
-The simple installation scripts set up MXNet for Python and R on computers running Ubuntu 12 or later. The scripts install MXNet in your home folder ```~/mxnet```.
+* [CUDA Dependencies](#cuda-dependencies)
+* [Quick Installation](#quick-installation)
+    * [Python](#install-mxnet-for-python)
+    * [pip Packages](#pip-package-availability)
+* [Standard Installation](#standard-installation)
+* [Installing Language Packages](#installing-language-packages-for-mxnet)
+    * [R](#install-the-mxnet-package-for-r)
```
marcoabreu: Is the CPP package missing?

docs/install/ubuntu_setup.md
```diff
@@ -1,67 +1,103 @@
...
+## CUDA Dependencies
 
-First, download and install [CUDA 8 toolkit](https://developer.nvidia.com/cuda-toolkit).
+If you plan to build with GPU, you need to set up the environment for CUDA and cuDNN.
 
-Then download [cudnn 6](https://developer.nvidia.com/cudnn).
+First, download and install [CUDA toolkit](https://developer.nvidia.com/cuda-toolkit). CUDA 9.2 is recommended.
 
-Unzip the file and change to the cudnn root directory. Move the header and libraries to your local CUDA Toolkit folder:
+Then download [cuDNN 7.1.4](https://developer.nvidia.com/cudnn).
```
marcoabreu: Do we want to specify a minor cuDNN versions? We're always using the latest one in CI, so we shouldn't give the users the impression that only a certain one is supported

docs/install/ubuntu_setup.md
```diff
@@ -1,67 +1,103 @@
...
-```bash
-    sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E084DAB9
-    sudo add-apt-repository ppa:marutter/rdev
+The following scripts will install Ubuntu 16.04 dependencies for MXNet Python development.
 
-    sudo apt-get update
-    sudo apt-get upgrade
-    sudo apt-get install r-base r-base-dev
+```bash
+wget https://raw.githubusercontent.com/apache/incubator-mxnet/master/ci/docker/install/ubuntu_core.sh
```
marcoabreu: Puh, I'd prefer if we keep these scripts internal and not user-facing. We install a bunch of crap in there because it's required for some rare use-case. Also, they might be changed at any time. It would be better if we keep these two concerns separate and let the installation guide test ensure that everything works properly.

docs/install/ubuntu_setup.md
```diff
@@ -1,67 +1,103 @@
...
+Alternatively, you can use the table below to select the package that suits your purpose.
+
+| MXNet Version | Basic | CUDA | MKL | CUDA/MKL |
+|-|-|-|-|-|
+| Latest | mxnet | mxnet-cu92 | mxnet-mkl | mxnet-cu92mkl |
+
+
+#### pip Package Availability
+
+The following table presents the pip packages that are recommended for each version of MXNet.
```
marcoabreu: That's awesome!

marcoabreu:
We should definitely address the >16.04 issues. Ubuntu is our main platform and we want to support it as early as possible to catch early adopters. @nswamy do we have an issue for that so we can look into the problems independent of this PR?

docs/install/ubuntu_setup.md
aaronmarkham: I was able to do this because the info is available, but scattered. I wish we had this for the other things like cuDNN and numpy and OpenCV. I notice installation problems with the latter two all of the time.

docs/install/ubuntu_setup.md
aaronmarkham: I was thinking about maintainability. We essentially test these constantly. But, I can agree with the bloat issue. What if we get a stripped down version just for installation instructions, and add that to CI?
Then I can work towards removing the dependency on the install page being parsed for a CI test - something that can end up being brittle and doesn't really thoroughly test as it's pretty much just doing pip installs.

docs/install/ubuntu_setup.md
aaronmarkham: I debated that, but I think if we're recommending an install path, then it should be the one we're testing all of the time. However, we should supply a table of the supported versions per release. I don't know what minor or major versions work for 1.1.0, 1.0.0, and so on....

docs/install/ubuntu_setup.md
aaronmarkham: You're right... this needs to be added to this page.

aaronmarkham:
@marcoabreu - if there's nothing _wrong_ in this info, then we should get it merged ASAP. Then make another issue/jira that captures the things that should be added (like CPP, different install script, addressing >16.04). The site currently shows incorrect info, and I'd like to see that updated right away.

docs/install/ubuntu_setup.md
marcoabreu: Sure, feel free to create a new nightly test. You could then reference the resources of that test in the installation instructions.

marcoabreu:
Okay, I'll merge it due to urgency. Please address it in a new PR.

PR #11513, opened 2018-07-01
Add path_suffixes for finding cudnn.lib

nehaljwani:
On Windows, the directory structure layout for CUDNN is:

```
/t/incubator-mxnet ❯❯❯ unzip -l /tmp/cudnn-9.0-windows10-x64-v7.1.zip
Archive:  /tmp/cudnn-9.0-windows10-x64-v7.1.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
   122719  04-28-2018 08:01   cuda/include/cudnn.h
    38963  05-01-2018 11:50   cuda/NVIDIA_SLA_cuDNN_Support.txt
    49070  05-05-2018 08:48   cuda/lib/x64/cudnn.lib
331319808  05-05-2018 08:48   cuda/bin/cudnn64_7.dll
---------                     -------
331530560                     4 files
```

When `CUDA_ROOT` is set to `/path/to/extracted/cudnn/cuda`, the `find_library()` call failed, because it couldn't find `cudnn.lib` 

The `find_library()` call tries to find `cudnn.lib` in `%CUDA_ROOT%` and `%CUDA_ROOT%/include` but doesn't look in `%CUDA_ROOT%/lib` or `%CUDA_ROOT%/lib/x64`. This patch fixes the problem by specifying the path suffixes to include `lib` and `lib/x64`.

### DIFF
--- a/cmake/FirstClassLangCuda.cmake
+++ b/cmake/FirstClassLangCuda.cmake
@@ -38,6 +38,7 @@ function(detect_cuDNN)
 
   find_library(CUDNN_LIBRARY NAMES libcudnn.so cudnn.lib # libcudnn_static.a
                              PATHS ${CUDNN_ROOT} $ENV{CUDNN_ROOT} ${CUDNN_INCLUDE}
+                             PATH_SUFFIXES lib lib/x64
                              DOC "Path to cuDNN library.")
 
   if(CUDNN_INCLUDE AND CUDNN_LIBRARY)


nehaljwani:
I am setting CUDNN_ROOT as an environment variable, if it that makes any difference. Also, I traced it being called from: https://github.com/apache/incubator-mxnet/blob/1.2.0/CMakeLists.txt#L414

nehaljwani:
This is still present in the current HEAD: https://github.com/apache/incubator-mxnet/blob/37c8165748457e1b648ec49f68e70027da2af6f2/CMakeLists.txt#L406

KellenSunderland:
This for the quick reply.  I wanted to verify since this file seems quite out of date to me.  I'll try setting CUDNN_ROOT and see if I follow this code path.  The whole FisrtCuda path is a bit of a mystery to me.  I think it is mostly reliant on what cmake version you have (later versions of some support for native cuda projects).

This is all orthogonal to this PR, which LGTM.  I like the PATH_SUFFIXES, seems more concise than what we have in our modules.

nehaljwani:
FWIW, I was using CMake 3.11.1

KellenSunderland:
I wonder if you'd also be willing to contribute the change to this project?  https://github.com/dmlc/mshadow/blob/master/cmake/Cuda.cmake#L225

Issue #11516, opened 2018-07-01
CCache does not work for NVCC or make-based builds


marcoabreu:
At the moment, we don't apply ccache to NVCC compilation or to make-based builds in general. This is due to third parties not supporting our make-wrappers.

marcoabreu:
Problems with NVCC - CCache log:

```
[2018-07-01T23:46:51.232513 6414 ] === CCACHE 3.4.2 STARTED =========================================
[2018-07-01T23:46:51.232513 6414 ] Config: (default) base_dir =
[2018-07-01T23:46:51.232513 6414 ] Config: (environment) cache_dir = /work/ccache
[2018-07-01T23:46:51.232513 6414 ] Config: (default) cache_dir_levels = 2
[2018-07-01T23:46:51.232513 6414 ] Config: (default) compiler =
[2018-07-01T23:46:51.232513 6414 ] Config: (default) compiler_check = mtime
[2018-07-01T23:46:51.232513 6414 ] Config: (default) compression = false
[2018-07-01T23:46:51.232513 6414 ] Config: (default) compression_level = 6
[2018-07-01T23:46:51.232513 6414 ] Config: (default) cpp_extension =
[2018-07-01T23:46:51.232513 6414 ] Config: (default) direct_mode = true
[2018-07-01T23:46:51.232513 6414 ] Config: (default) disable = false
[2018-07-01T23:46:51.232513 6414 ] Config: (default) extra_files_to_hash =
[2018-07-01T23:46:51.232513 6414 ] Config: (default) hard_link = false
[2018-07-01T23:46:51.232513 6414 ] Config: (default) hash_dir = true
[2018-07-01T23:46:51.232513 6414 ] Config: (default) ignore_headers_in_manifest =
[2018-07-01T23:46:51.232513 6414 ] Config: (default) keep_comments_cpp = false
[2018-07-01T23:46:51.232513 6414 ] Config: (default) limit_multiple = 0.8
[2018-07-01T23:46:51.232513 6414 ] Config: (environment) log_file = /work/mxnet/ccache-log
[2018-07-01T23:46:51.232513 6414 ] Config: (default) max_files = 0
[2018-07-01T23:46:51.232513 6414 ] Config: (environment) max_size = 500.0G
[2018-07-01T23:46:51.232513 6414 ] Config: (default) path =
[2018-07-01T23:46:51.232513 6414 ] Config: (default) prefix_command =
[...]

PR #11520, opened 2018-07-02
[WIP] Improve Ccache

marcoabreu:
## Description ##
https://github.com/apache/incubator-mxnet/issues/11516

### Essentials ###

## Comments ##
- So far, GCC and G++ are properly handled by CCache and we have proper cache hits. The problem here is that NVCC still produces cache misses.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -167,10 +167,6 @@ python/.eggs
 tests/Makefile
 tests/mxnet_unit_tests
 
-# generated wrappers for ccache
-cc
-cxx
-
 # Code coverage related
 .coverage
 *.gcov
--- a/ci/docker/install/ubuntu_adduser.sh
+++ b/ci/docker/install/ubuntu_adduser.sh
@@ -40,4 +40,9 @@ then
     mkdir /work/mxnet
     mkdir /work/build
     chown -R jenkins_slave /work/
+
+    # Later on, we have to override the links because underlying build systems ignore our compiler settings. Thus,
+    # we have to give the process the proper permission to these files. This is hacky, but unfortunately 
+    # there's no better way to do this without patching all our submodules.
+    chown -R jenkins_slave /usr/local/bin
 fi
--- a/ci/docker/runtime_functions.sh
+++ b/ci/docker/runtime_functions.sh
@@ -36,35 +36,37 @@ clean_repo() {
     git submodule update --init --recursive
 }
 
-build_ccache_wrappers() {
+install_ccache_wrappers() {
     set -ex
 
-    rm -f cc
-    rm -f cxx
+    # Recommended by CCache: https://ccache.samba.org/manual.html#_run_modes
+    # Add to the beginning of path to ensure this redirection is picked up instead
+    # of the original ones. Especially CUDA/NVCC appends itself to the beginning of the
+    # path and thus this redirect is ignored. This change fixes this problem
+    # This hacky approach with symbolic links is required because underlying build
+    # systems of our submodules ignore our CMake settings. If they use Makefile,
+    # we can't influence them at all in general and NVCC also prefers to hardcode their
+    # compiler instead of respecting the settings. Thus, we take this brutal approach
+    # and just redirect everything of this installer has been called.
+    # In future, we could do these links during image build time of the container.
+    # But in the beginning, we'll make this opt-in. In future, loads of processes like
+    # the scala make step or numpy compilation and other pip package generations
+    # could be heavily sped up by using ccache as well.
+    export PATH=/usr/local/bin:$PATH
+    ln -s ccache /usr/local/bin/gcc
+    ln -s ccache /usr/local/bin/g++
+    ln -s ccache /usr/local/bin/g++-8
+    ln -s ccache /usr/local/bin/gcc-8
+    ln -s ccache /usr/local/bin/nvcc
+    ln -s ccache /usr/local/bin/clang++-3.9
+    ln -s ccache /usr/local/bin/clang-3.9
+    ln -s ccache /usr/local/bin/clang++-5.0
+    ln -s ccache /usr/local/bin/clang-5.0
+    ln -s ccache /usr/local/bin/clang++-6.0
+    ln -s ccache /usr/local/bin/clang-6.0
 
-    touch cc
-    touch cxx
-
-    if [ -z ${CC+x} ]; then
-        echo "No \$CC set, defaulting to gcc";
-        export CC=gcc
-    fi
-
-    if [ -z ${CXX+x} ]; then
-       echo "No \$CXX set, defaulting to g++";
-       export CXX=g++
-    fi
-
-    # this function is nessesary for cuda enabled make based builds, since nvcc needs just an executable for -ccbin
-
-    echo -e "#!/bin/sh\n/usr/local/bin/ccache ${CC} \"\$@\"\n" >> cc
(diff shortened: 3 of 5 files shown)


ci/docker/install/ubuntu_adduser.sh
```diff
@@ -34,4 +34,5 @@ then
     mkdir /work/mxnet    
     mkdir /work/build
     chown -R jenkins_slave /work/
+    chown -R jenkins_slave /usr/local/bin
```
larroy: I think this is hacky.

ci/docker/runtime_functions.sh
```diff
@@ -33,35 +33,23 @@ clean_repo() {
...
-    # this function is nessesary for cuda enabled make based builds, since nvcc needs just an executable for -ccbin
 
-    echo -e "#!/bin/sh\n/usr/local/bin/ccache ${CC} \"\$@\"\n" >> cc
-    echo -e "#!/bin/sh\n/usr/local/bin/ccache ${CXX} \"\$@\"\n" >> cxx
+install_ccache_wrappers() {
+    set -ex
 
-    chmod +x cc
-    chmod +x cxx
+    # Add to the beginning of path to ensure this redirection is picked up instead
```
larroy: Why is this better than the previous solution? It looks less explicit and hacky. Can you provide some rationale?

marcoabreu: It is the recommended way by CCache: https://ccache.samba.org/manual.html#_run_modes
This way ensures that ccache is always used if the symlinks are set. Otherwise, we'd have to make a possibility in all build-paths (including third party) that they use ccache properly. This caused problems in compiling ps-lite, because that project had **another** third party dependency (ZeroMQ) which did not pick up the propagated compiler.

ci/docker/install/ubuntu_adduser.sh
marcoabreu: The problem is that the entire Docker container is owned by root by default. Since we run in low-privilege mode inside our container (opposed to running everything as root), we have to whitelist certain directories (/work and /usr/local/bin) that the process is allowed to write. This is basically a necessary evil of the least-priviledge-policy. 

An alternative would be to set these softlinks directly during the ccache installation. The problem there would then be that CCache would **always** be activated and we would not be able to turn it on/off during runtime. I wouldn't mind about that, but I didn't want to change the behaviour. What do you think?

marcoabreu:
The problem is that the current method with the ccache wrappers does not work for third party dependencies. There are some scripts (libtool which is part of ZeroMQ which is part of 3rdparty/ps-lite) which identify the compiler by its name. If we redirect the compiler using the wrapper method, the script gets confused and falls back into a backup mode, which isn't able to compile the sub-projects properly. Also, this is the recommended way by CCache: https://ccache.samba.org/manual.html#_run_modes

larroy:
I didn't know it's recommended by the authors. Thanks for the clarification.

ci/docker/runtime_functions.sh
```diff
@@ -489,9 +464,8 @@ build_ubuntu_amalgamation_min() {
 build_ubuntu_gpu_cmake_mkldnn() {
     set -ex
     cd /work/build
+    install_ccache_wrappers
```
lebeg: There should not be any need for wrapper for cmake based builds.

.gitignore
```diff
@@ -169,4 +169,4 @@ tests/mxnet_unit_tests
 
 # generated wrappers for ccache
 cc
-cxx
\ No newline at end of file
+cxx
```
lebeg: if you moved out the wrappers then this few lines are not needed

ci/docker/runtime_functions.sh
```diff
@@ -33,35 +33,23 @@ clean_repo() {
...
+install_ccache_wrappers() {
+    set -ex
 
-    chmod +x cc
-    chmod +x cxx
+    # Add to the beginning of path to ensure this redirection is picked up instead
+    # of the original ones. Especially CUDA/NVCC appends itself to the beginning of the
+    # path and thus this redirect is ignored. This change fixes this problem
+    export PATH=/usr/local/bin:$PATH
+    ln -s ccache /usr/local/bin/gcc
```
lebeg: Would this work if already set?

ci/docker/runtime_functions.sh
marcoabreu: The idea here is that we rather intercept the backend calls to the compilers instead of defining them through the frontend (aka the Cmake parameters). This allows us to be flexible with whatever compilers we define without having to think about ccache because it's handled automatically. Does that make sense?

.gitignore
marcoabreu: Will do

ci/docker/runtime_functions.sh
marcoabreu: Yes, you can define and configure the compilers ($CC, $CXX, $NVCC) as you want, we only intercept the backend calls to the actual compilers. This allows us to be independent of the configuration.

ci/docker/runtime_functions.sh
lebeg: As much as I know cmake will not like this and different problems can appear suddenly. Using cmake parameters is the recommended approach.

ci/docker/runtime_functions.sh
lebeg: I meant if ccache soft link is already there.

nswamy:
@marcoabreu are you going address the comments, if its handled in a different PR, could you please close this?

marcoabreu:
This PR is still WIP and paused due to other priorities. It is still valid.

ci/docker/install/ubuntu_adduser.sh
larroy: What requires permissions to write in /usr/local/bin ?

larroy:
Is this problem also happening with CMake builds or only in Make builds?

stu1130:
Thanks for the contriubtion @larroy!
@marcoabreu could you please review it! Thanks

vandanavk:
@sandeep-krishnamurthy Please change the label from [pr-awaiting-response, pr-work-in-progress] to [pr-awaiting-testing, pr-work-in-progress]

Roshrini:
Hi @marcoabreu, Any update on this PR? Are you still working on it?

roywei:
ping @marcoabreu any updates?

marcoabreu:
Yes, I will come back to this as soon as I got a few hours

lupesko:
@marcoabreu this PR has been inactive for 3 months now.
If you are unable to make progress on it it might be better to close it.

ci/docker/install/ubuntu_adduser.sh
marcoabreu: The symbolic link I'm creating. It's super hacky, but some buildsystems seem to directly access /usr/local/bin in order to access the compiler binary. They ignore any custom setting, so I'm using the super brutal way to just rewrite that entry by overriding it with a symlink

ci/docker/runtime_functions.sh
marcoabreu: In that case, the softlink (or whatever was there previously) would just get replaced. But since this script is executed at the install time of the container, it's highly unlikely that the link already exists

ci/docker/runtime_functions.sh
marcoabreu: I know and I only chose this approach after playing around a lot. The problem is that the underlying submodules we have don't respect that choice.

marcoabreu:
Replaced by https://github.com/apache/incubator-mxnet/pull/13456

PR #11546, opened 2018-07-03
Add verify_ssl option to gluon.utils.download

leezu:
## Description ##
Sometimes datasets may be hosted on servers that serve invalid SSL certificates. If so, this allow disabling the SSL certificate verification. The default behavior is not changed.

### Essentials ###

### Changes ###
- [X] Add verify_ssl option to gluon.utils.download

### DIFF
--- a/tests/python/unittest/test_gluon_utils.py
+++ b/tests/python/unittest/test_gluon_utils.py
@@ -20,7 +20,10 @@ import os
 import tempfile
 import warnings
 
-import mock
+try:
+    from unittest import mock
+except ImportError:
+    import mock
 import mxnet as mx
 import requests
 from nose.tools import raises
--- a/tests/python/unittest/test_gluon_utils.py
+++ b/tests/python/unittest/test_gluon_utils.py
@@ -15,28 +15,53 @@
 # specific language governing permissions and limitations
 # under the License.
 
+import io
 import os
 import tempfile
+import warnings
 
+import mock
 import mxnet as mx
-from nose.tools import *
+import requests
+from nose.tools import raises
+
+
+class MockResponse(requests.Response):
+    def __init__(self, status_code, content):
+        super(MockResponse, self).__init__()
+        assert isinstance(status_code, int)
+        self.status_code = status_code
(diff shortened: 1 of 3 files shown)


larroy:
LGTM

tests/python/unittest/test_gluon_utils.py
```diff
@@ -31,4 +31,12 @@ def test_download_successful():
...
     mx.gluon.utils.download("https://raw.githubusercontent.com/apache/incubator-mxnet/master/README.md",
                             path=tmpfile)
-    assert os.path.getsize(tmpfile) > 100
\ No newline at end of file
+    assert os.path.getsize(tmpfile) > 100
+
+def test_download_ssl_verify():
+    with warnings.catch_warnings(record=True) as warnings_:
+        mx.gluon.utils.download(
+            "https://mxnet.incubator.apache.org/", verify_ssl=False)
```
larroy: I would mock the call to requests to avoid doing network IO on a unit test.

larroy:
Is better to mock it, in python is super simple. Using localhost can also fail if the port is used, so best to avoid it. Let's not overcomplicate things. Mocking the request is easy:  

https://stackoverflow.com/questions/15753390/how-can-i-mock-requests-and-the-response

leezu:
Thanks. I have now changed all `gluon.utils.download` tests to make use of Mocking (the two tests already present before this PR caused network access).

leezu:
Where are test dependencies currently tracked? This PR would introduce `mock` as a new dependency. For linux it seems I could add `mock` the list of installed packages in the CI setup shell scripts, but I am not sure about windows.

In case there is no centralized place to declare test/development dependencies, would it make sense to track these in the `setup.py` script and adapt the CI scripts to install via `pip install -e '.[tests]'` to also install these dependencies?

szha:
AFAIK windows CI slave images are still manually built. @marcoabreu might have update

larroy:
We can workaround this by adding pip install mock after the activate call on the Jenkinsfile .  parametrization of windows host is starting now and will take a while.

PR #11558, opened 2018-07-04
Profiler is already enabled by default in 1.2.0

Ishitori:
## Description ##
A user doesn't need to compile mxnet from source to enable profiling. Installing 1.2.0 version (the one I checked) of mxnet is enough to run the code listed in that notebook. I checked it for both CPU and GPU contexts.

### Changes ###
- [ ] Remove the requirement to compile mxnet from source

### DIFF
--- a/docs/tutorials/python/profiler.md
+++ b/docs/tutorials/python/profiler.md
@@ -49,9 +49,7 @@ While it is possible to use `NDArray.waitall()` before and after operations to g
 
 ## The correct way to profile
 
-The correct way to measure running time of MXNet models is to use MXNet profiler. In the rest of this tutorial, we will learn how to use the MXNet profiler to measure the running time and memory consumption of MXNet models.
-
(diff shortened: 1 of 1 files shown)


indhub:
@piiswrong Could you please confirm it is not necessary to build MXNet with USE_PROFILER=1 to use the profiler? If true, we need to fix documentation in few other places too.

indhub:
These pages specifically needs change:
https://mxnet.incubator.apache.org/faq/perf.html#profiler
https://mxnet.incubator.apache.org/faq/env_var.html#control-the-profiler

indhub:
@cjolivier01

szha:
@indhub USE_PROFILER=1 is still necessary and is on by default in both make and cmake, and is enabled in pip. I didn't think @Ishitori meant that USE_PROFILER=1 is no longer necessary.

indhub:
Okay, got it. It might be helpful to mention in those FAQs that pip package already has the flag enabled.

szha:
@rahul003 oh indeed. I didn't realize. Thanks for the correction.

PR #11566, opened 2018-07-05
[MXNET-626] Add while_loop

junrushao1994:
## Description ##

This PR is part of the proposal of adding a set of control flow operators to MXNet. [Link](https://cwiki.apache.org/confluence/display/MXNET/Optimize+dynamic+neural+network+models+with+control+flow+operators) to proposal.

### Essentials ###

### Changes ###
- Add while_loop operator in `src/operator/control_flow.cc`

## TODO ##
- [x] Add a JIRA_ID
- [x] Add more testcases
- [x] Re-examine all `TODO(Junru)` in changes


tests/python/unittest/test_while_loop.py
```diff
@@ -0,0 +1,969 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
zheng-da: please rename this file to test_control_flow.py

tests/python/unittest/test_while_loop.py
```diff
@@ -0,0 +1,969 @@
...
+                assert_almost_equal(x, y, rtol=1e-4, atol=1e-4)
+
+
+if __name__ == '__main__':
+    # import nose
+    # nose.runmodule()
+    test_while_loop_simple_forward()
+    test_while_loop_for_foreach()
+    test_while_loop_nested()
+    test_while_loop_rnn()
```
zheng-da: you don't need to call the test functions explicitly.

zheng-da:
@szha @piiswrong @eric-haibin-lin please help review this PR.

src/operator/control_flow.cc
```diff
@@ -480,6 +480,512 @@ ForeachGradient(const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ogra
...
+    //    outputs[i]            (step != 0 && i not in loop_var_locs)
+    //    ArrayLike(outputs[i]) (step != 0 && i in loop_var_locs)
+    // iter_req =
+    //    kWriteTo              (step != 0           && i in loop_var_locs)
+    //    req[i]                (step == 0           && i in loop_var_locs)
+    //    kAddTo                (step != n_iters - 1 && i not in loop_var_locs)
+    //    req[i]                (step == n_iters - 1 && i not in loop_var_locs)
+    {
+      size_t i = 0;
+      for (size_t loc : var_locs) {
```
zheng-da: what about the variables behind the loop variables? for example, the inputs of the loop are
[loop_var1, loop_var2, other_var1, other_var2]. How are other_var1 and other_var2 processed?

tests/python/unittest/test_while_loop.py
```diff
@@ -0,0 +1,969 @@
...
+
+    for hybridize in [False, True]:
+        # Case 1.1: result should be sum([1, 2, 3 ... 100])
+        model = _TestBlock(
+            cond=lambda i, s: i <= 5,
+            func=lambda i, s: (None, (i + 1, s + i)),
+            max_iterations=10,
+        )
+        if hybridize:
+            model.hybridize(inline_limit=0)
```
zheng-da: is `inline_limit` necessary?

zheng-da: i don't think people use `inline_limit` by default. we should test common cases.

tests/python/unittest/test_while_loop.py
```diff
@@ -0,0 +1,969 @@
...
+        _, result = model(
+            mx.nd.array([1], dtype="int64"), # i
+            mx.nd.array([0], dtype="int64"), # s
+            mx.nd.array([0], dtype="int64"), # false
+        )
+        assert result[0].asscalar() == 1
+        assert result[1].asscalar() == 0
+        assert result[2].asscalar() == 0
+        # Case 2.1: result should be sum([1, 2, 3 ... 100])
```
zheng-da: the cases are labelled with xx.yy. do xx and yy have special meanings?

tests/python/unittest/test_while_loop.py
```diff
@@ -0,0 +1,969 @@
...
+            model.hybridize(inline_limit=0)
+        (outputs, ), (result_i, result_s, _) = model(
+            mx.nd.array([1], dtype="int64"), # i
+            mx.nd.array([0], dtype="int64"), # s
+            mx.nd.array([1], dtype="int64"), # true
+        )
+        assert all(outputs.asnumpy() == np.arange(1, 1001).reshape(1000, 1))
+        assert result_i.asscalar() == 1001
+        assert result_s.asscalar() == 500500
+        # Case 2.3: very corner case
```
zheng-da: what corner case does it test?

tests/python/unittest/test_while_loop.py
```diff
@@ -0,0 +1,969 @@
...
+                    is_train=is_train,
+                    is_for=True,
+                    **params
+                )
+
+    def case_4(length, single_shape, **params):
+        # It is for the case that inputs & outputs are the same
+        # There are 3 outputs
+        # There are 4 states: i, s_0, s_1, s_2
+        # i is used in both differentiable (take) and non-differentiable (+) occasions
```
zheng-da: take is differentiable, + is non-differentiable?

python/mxnet/ndarray/contrib.py
```diff
@@ -191,3 +191,128 @@ def check_input(inputs, in_type, msg):
...
+    The loop ends when `cond` returns false (zero).
+    The `cond` is variadic, and its signature should be
+    `cond(*loop_vars) => NDArray`.
+
+    `func` is a user-defined function as the loop body.
+    It also consumes `loop_vars`, and produces `step_output` and `new_loop_vars` at each step.
+    The number of elements, shape, dtype of each element in `step_output` should be consistent.
+    The `new_loop_vars` should be consistent with `loop_vars` on each step.
+    The `func` is variadic, and its signature should be
+    `cond(*loop_vars) => (List[NDArray] step_output, List[NDArray] new_loop_vars)`.
```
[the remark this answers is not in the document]
junrushao1994: Fixed :-)

tests/python/unittest/test_while_loop.py
junrushao1994: Oops I forgot to uncomment the lines above. Fixed :-)

tests/python/unittest/test_while_loop.py
junrushao1994: Fixed :-)

tests/python/unittest/test_while_loop.py
junrushao1994: Removed

tests/python/unittest/test_while_loop.py
junrushao1994: In xx.yy, I use `xx` as an id of a testing template, for instance, "def case_1(**param)" is a template for various testcases. I specify different shapes and other conditions in the template, to which I refer as `yy`.

tests/python/unittest/test_while_loop.py
junrushao1994: The loop body is never executed. Thanks for the review, I change the comment to "Case 2.3: a corner case, in which loop body is never executed" to make it clear.

tests/python/unittest/test_while_loop.py
junrushao1994: My mistake. Fixed :-)

src/operator/control_flow.cc
junrushao1994: In [line 745](https://github.com/junrushao1994/mxnet/blob/316b0f7a10dcaff6b7792aa2a68c279c934087f8/src/operator/control_flow.cc#L745), a dummy position is appended to `var_locs`

python/mxnet/symbol/contrib.py
```diff
@@ -336,3 +336,205 @@ def check_data(inputs, in_type, msg):
...
+                                # where tuples are results of 2) and 3)
+        input_id_to_loc = {}    # Dict[int, int], given id(sym), input_id_to_loc maps it
+                                # to a `loc`, where inputs[loc] = sym
+        for graph in graphs:
+            # input_syms: all inputs to the `graph`
+            name_to_input_syms = {sym.name: sym for sym in _get_graph_inputs(graph)}
+            # some loop_vars are inputs to `graph`, some are not
+            name_to_loop_vars = {sym.name: sym for sym in loop_vars}
+            # other inputs to `graph` created by cut_graph
+            name_to_cut_g_syms = {sym.list_outputs()[0]: sym for sym in _cut_subgraph(graph)}
```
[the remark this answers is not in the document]
junrushao1994: I feel like that they are equivalent. Just copied from this line in [foreach](https://github.com/junrushao1994/mxnet/blob/316b0f7a10dcaff6b7792aa2a68c279c934087f8/python/mxnet/symbol/contrib.py#L277).

python/mxnet/symbol/contrib.py
```diff
@@ -336,3 +336,205 @@ def check_data(inputs, in_type, msg):
...
+                assert name in name_to_input_syms   # it should obviously hold
+                # name -> sym
+                if name in name_to_loop_vars:
+                    sym = name_to_loop_vars[name]
+                elif name in name_to_cut_g_syms:
+                    sym = name_to_cut_g_syms[name]
+                else:
+                    sym = copy.deepcopy(name_to_input_syms[name])
+                # do 2), and 1) is implicitly done
+                if id(sym) in input_id_to_loc:
```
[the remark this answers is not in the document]
junrushao1994: There are several subgraphs, more specifically, two subgraphs `cond` and `func` in while_loop. They may have common input symbols, so these symbols may have been added to `input_id_to_loc`.

src/operator/control_flow.cc
```diff
@@ -480,6 +480,512 @@ ForeachGradient(const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ogra
...
+  T data;
+  a.SyncCopyToCPU(&data, 1U);
+  return data;
+}
+
+bool as_bool_scalar(const NDArray &a) {
+  MSHADOW_TYPE_SWITCH(a.dtype(), DType, {
+    return static_cast<bool>(_asscalar<DType>(a));
+  });
+  CHECK(false) << "Unknown dtype";
```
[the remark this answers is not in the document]
junrushao1994: Fixed -:)

src/operator/control_flow.cc
```diff
@@ -480,6 +480,512 @@ ForeachGradient(const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ogra
...
+  CHECK_EQ(inputs.size() + 2U, (size_t) params.num_args);
+  CHECK_EQ(outputs.size(), (size_t) params.num_outputs);
+  CHECK_EQ(outputs.size(), req.size());
+  for (size_t i = 0; i < (size_t) params.num_out_data; i++)
+    CHECK_EQ(params.max_iterations, outputs[i].shape()[0]);
+  for (const auto &arr : outputs)
+    CHECK_EQ(arr.storage_type(), kDefaultStorage)
+          << "The while_loop operator doesn't support the sparse format";
+  // construct inputs and outputs for cond
+  std::vector<NDArray> cond_inputs, cond_outputs = {NDArray()};
```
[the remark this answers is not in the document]
junrushao1994: I just wanted to hold the output of `cond`. Is there better way to do this? Thank you!

junrushao1994:
@szha @piiswrong @eric-haibin-lin Hey could you help take a look at the code?

junrushao1994:
@zheng-da Hey I address all of the comments except the one in `WhileLoopComputeExCPU`, in which I am not sure there is better solution. Could you take a look?

zheng-da:
There is a problem in the current implementation. while_loop doesn't inform users the number of iterations it actually performs, so users can't know how much data in the output arrays is valid. One option is to return a mask or a scalar symbol to indicate the number of iterations.

junrushao1994:
@zheng-da For users, an additional variable like `i` in loop variables could perfectly carry information about number of steps it takes, and intuitively matches how a Python while-loop works.
For developers, the number of steps taken is indicated in `WhileLoopState::n_iterations`, I am not sure how to access it from outside. Will this be a problem in future development?
If we consider change the API, i.e. add an extra symbol indicating number of steps taken, it will be some non-trivial discussion like naming, dtype and ctx of the Symbol, and all those stuff will be hard-coded in mxnet python lib. Not sure if there is better or more concise way to do this.

> There is a problem in the current implementation. while_loop doesn't inform users the number of iterations it actually performs, so users can't know how much data in the output arrays is valid. One option is to return a mask or a scalar symbol to indicate the number of iterations.

python/mxnet/ndarray/contrib.py
```diff
@@ -191,3 +191,128 @@ def check_input(inputs, in_type, msg):
...
+    loop_vars = _to_ndarray_tuple(loop_vars, "loop_vars")
+    # It should be work as fine if loop_vars are empty I guess,
+    # but it is semantically unnecessary to include this case.
+    if len(loop_vars) == 0:
+        raise ValueError("loop_vars should contain at least one element")
+
+    steps = 0
+    outputs = []
+    while steps < max_iterations and \
+            _to_python_scalar(cond(*loop_vars), bool, "Return value of cond"): # loop condition
```
piiswrong: So this could end before reaching max_iterations. Isn't this inconsistent with symbol?

python/mxnet/ndarray/contrib.py
```diff
@@ -191,3 +191,128 @@ def check_input(inputs, in_type, msg):
...
+        step_output, loop_vars = _func_wrapper(loop_vars)
+        outputs.append(step_output)
+        steps += 1
+        if len(outputs) != steps or len(step_output) != len(outputs[0]):
+            raise ValueError("step_output are inconsistent on each step")
+    try:
+        outputs = list(ndarray.op.stack(*item) for item in zip(*outputs))
+    except ValueError:
+        raise ValueError("step_outputs are inconsistent on each step")
```
piiswrong: be explicit about which value is inconsistent. Print out the inconsistent shapes if possible

python/mxnet/symbol/contrib.py
```diff
@@ -336,3 +336,205 @@ def check_data(inputs, in_type, msg):
         states = states[0]
 
     return (outs, states)
+
+def while_loop(cond, func, loop_vars, max_iterations, name="while_loop"):
```
piiswrong: interface different from ndarray?

python/mxnet/ndarray/contrib.py
```diff
@@ -191,3 +191,128 @@ def check_input(inputs, in_type, msg):
...
+    `func` is a user-defined function as the loop body.
+    It also consumes `loop_vars`, and produces `step_output` and `new_loop_vars` at each step.
+    The number of elements, shape, dtype of each element in `step_output` should be consistent.
+    The `new_loop_vars` should be consistent with `loop_vars` on each step.
+    The `func` is variadic, and its signature should be
+    `func(*loop_vars) => (List[NDArray] step_output, List[NDArray] new_loop_vars)`.
+
+    `max_iterations` is a scalar that defines the maximum number of iterations allowed.
+
+    This function returns a list of NDArrays of length `|step_output| + |loop_vars|`.
```
piiswrong: The return value has different format from TF. Any specific reasons?

python/mxnet/ndarray/contrib.py
```diff
@@ -191,3 +191,128 @@ def check_input(inputs, in_type, msg):
     if not_data_list and len(outputs) == 1:
         outputs = outputs[0]
     return (outputs, states)
+
+
+def while_loop(loop_vars, cond, func, max_iterations):
```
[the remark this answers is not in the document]
junrushao1994: My bad. Fixed :-)

python/mxnet/symbol/contrib.py
junrushao1994: My bad. Fixed

python/mxnet/symbol/contrib.py
```diff
@@ -336,3 +336,205 @@ def check_data(inputs, in_type, msg):
...
+    It also consumes `loop_vars`, and produces `step_output` and `new_loop_vars` at each step.
+    The number of elements, shape, dtype of each element in `step_output` should be consistent.
+    The `new_loop_vars` should be consistent with `loop_vars` on each step.
+    The `func` is variadic, and its signature should be
+    `func(*loop_vars) => (List[Symbol] step_output, List[Symbol] new_loop_vars)`.
+
+    `max_iterations` is a scalar that defines the maximum number of iterations allowed.
+
+    This function returns a list of Symbols of length `|step_output| + |loop_vars|`.
+    The i-th element in the first `|step_output|` ones of the list represent
```
eric-haibin-lin: I don't understand this sentence.

python/mxnet/symbol/contrib.py
```diff
@@ -336,3 +336,205 @@ def check_data(inputs, in_type, msg):
         states = states[0]
 
     return (outs, states)
+
+def while_loop(cond, func, loop_vars, max_iterations, name="while_loop"):
+    """Run a while loop with user-defined computation and loop condition.
```
eric-haibin-lin: Is max_iterations always required?

python/mxnet/ndarray/contrib.py
```diff
@@ -191,3 +191,128 @@ def check_input(inputs, in_type, msg):
...
+    outputs: a tuple of two lists, which both contains 0, 1 or more NDArrays.
+        The first list contains the stacked output from each step,
+        The second list contains the final state.
+
+    Examples
+    --------
+    >>> cond = lambda i, s: i <= 5
+    >>> func = lambda i, s: ([i + s], [i + 1, s + i])
+    >>> loop_vars = (mx.nd.array([0], dtype="int64"), mx.nd.array([1], dtype="int64"))
+    >>> outputs, states = mx.nd.contrib.while_loop(loop_vars, cond, func, max_iterations=10)
```
eric-haibin-lin: Can you show the output results of this example?

python/mxnet/symbol/contrib.py
zheng-da: yes, without dynamic shape, a user has to provide max_iterations

python/mxnet/ndarray/contrib.py
```diff
@@ -191,3 +191,128 @@ def check_input(inputs, in_type, msg):
...
+    `cond` is a user-defined function as the loop condition.
+    It consumes `loop_vars`, and produces a scalar MXNet NDArray,
+    indicating the termination of the loop.
+    The loop ends when `cond` returns false (zero).
+    The `cond` is variadic, and its signature should be
+    `cond(*loop_vars) => NDArray`.
+
+    `func` is a user-defined function as the loop body.
+    It also consumes `loop_vars`, and produces `step_output` and `new_loop_vars` at each step.
+    The number of elements, shape, dtype of each element in `step_output` should be consistent.
```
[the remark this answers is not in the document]
junrushao1994: I change this to

> In each step, `step_output` should contain the same number elements.
> Through all steps, the i-th element of `step_output` should have the same shape and dtype.
> Also, `new_loop_vars` should contain the same number of elements as `loop_vars`,
> and the corresponding element should have the same shape and dtype.

Does this seem better?

python/mxnet/ndarray/contrib.py
junrushao1994: Yes, they are not consistent, and I put a warning in the docstring. Should I do some padding stuff so that they look the same?

python/mxnet/symbol/contrib.py
junrushao1994: My bad, this is outdated. Does the following seem better?

> This function returns two lists as a tuple. The first list has the length of `|step_output|`, in which the i-th element are all i-th elements of `step_output` from all steps, stacked along axis 0. The second list has the length of `|loop_vars|`, which represents final states of loop variables.

python/mxnet/ndarray/contrib.py
junrushao1994: The results are

```
>>> outputs
[
 [[ 1]
  [ 2]
  [ 4]
  [ 7]
  [11]
  [16]]
 <NDArray 6x1 @cpu(0)>]
>>> states
[
 [6]
 <NDArray 1 @cpu(0)>,
 [16]
 <NDArray 1 @cpu(0)>]
```
Should I put this snippet into docstring?

python/mxnet/ndarray/contrib.py
junrushao1994: I updated the outdated docstring to the following:

> This function returns two lists as a tuple. The first list has the length of `|step_output|`, in which the i-th element are all i-th elements of `step_output` from all steps, stacked along axis 0. The second list has the length of `|loop_vars|`, which represents final states of loop variables.

Currently we don't have dynamic shape inference, so could not support TF-like dynamic-sized, per-time-step `TensorArray`. So we split our requirement into two parts: 1. per-time-step array with `max_iteration`; 2. loop variables which are of the same shape through the loop.

@zheng-da

python/mxnet/ndarray/contrib.py
zheng-da: i think so.

python/mxnet/ndarray/contrib.py
zheng-da: the interface of this while_loop operator is close to the ONNX definition. https://github.com/onnx/onnx/blob/master/docs/Operators.md#Loop
The interface of TF while_loop requires dynamic shape and also doesn't allow efficient implementation. This interface is more flexible. If it returns `([], loop_vars)`,  it's the same as the TF interface.

junrushao1994:
@eric-haibin-lin @zheng-da  I think while loop could support sparse NDArrays as well, so I remove the checks

python/mxnet/ndarray/contrib.py
zheng-da: i think so.

junrushao1994: @zheng-da So should I pad NDArray to make them consistent?

python/mxnet/ndarray/contrib.py
```diff
@@ -28,7 +28,7 @@
 except ImportError:
     pass
 
-__all__ = ["rand_zipfian"]
+__all__ = ["rand_zipfian", "foreach", "while_loop"]
```
eric-haibin-lin: need to update docs/api/python/ndarray&symbol/contrib.md, too

python/mxnet/symbol/contrib.py
```diff
@@ -336,3 +336,219 @@ def check_data(inputs, in_type, msg):
...
+        if new_loop_vars is None:
+            new_loop_vars = []
+        step_output = _to_symbol_tuple(step_output, "step_output")
+        new_loop_vars = _to_symbol_tuple(new_loop_vars, "new_loop_vars")
+        if len(loop_vars) != len(new_loop_vars):
+            raise ValueError("The number of loop_vars should be consistent during the loop")
+        return list(step_output), list(new_loop_vars)
+
+    def _create_subgraph(graph_vars, graph_func, subgraph_name):
+        with AttrScope(__subgraph_name__=subgraph_name):
```
eric-haibin-lin: Does it matter if user doesn't provide name and subgraph_name duplicates?

zheng-da: probably not. the C code that cuts the subgraph looks for nodes with the attribute of "__subgraph_name__"

python/mxnet/ndarray/contrib.py
zheng-da: it's better to do so, in my opinion. what do you think? @piiswrong

python/mxnet/ndarray/contrib.py
junrushao1994: Updated

python/mxnet/ndarray/contrib.py
junrushao1994: Fixed

python/mxnet/ndarray/contrib.py
junrushao1994: Fixed

python/mxnet/ndarray/contrib.py
piiswrong: Yes, ndarray and symbol functions should give the same result for the same input. Otherwise hybridize may break

python/mxnet/ndarray/contrib.py
piiswrong: So does it return two lists separately or concated together?

junrushao1994: @piiswrong Separately

piiswrong: This function returns two lists as a tuple -> This function returns two lists.

"as a tuple" makes it sounds like the two lists are concated into a tuple

junrushao1994: @piiswrong fixed

junrushao1994:
@piiswrong Can we merge this PR?

fhieber:
@junrushao1994 thanks for working on this operator! I can't wait to have this available in MXNet. I took the freedom of checking out this PR and compiling MXNet on a mac to test this operator.
I tried using the example described in the docstring:
```python
>>> cond = lambda i, s: i <= 5
>>> func = lambda i, s: ([i + s], [i + 1, s + i])
>>> loop_vars = (mx.sym.var('i'), mx.sym.var('s'))
>>> outputs, states = mx.sym.contrib.while_loop(cond, func, loop_vars, max_iterations=10)
```
I am evaluating this symbolic graph like this to inspect the outputs from each iteration:
```python
mx.sym.Group(outputs + states).eval(i=mx.nd.zeros((1,)), s=mx.nd.ones((1,)))[0].asnumpy().astype('int32')
```
As expected the output is padded to `max_iterations`, but the padding values are non-deterministic when repeatedly calling the above line:
```python
In [65]: mx.sym.Group(outputs + states).eval(i=mx.nd.zeros((1,)), s=mx.nd.ones((1,)))[0].asnumpy().astype('int32')
Out[65]:
array([[          1],
       [          2],
       [          4],
       [          7],
       [         11],
       [         16],
       [-2147483648],
       [          0],
       [-2147483648],
       [          0]], dtype=int32)

In [69]: mx.sym.Group(outputs + states).eval(i=mx.nd.zeros((1,)), s=mx.nd.ones((1,)))[0].asnumpy().astype('int32')
Out[69]:
array([[ 1],
       [ 2],
       [ 4],
       [ 7],
       [11],
       [16],
       [ 0],
       [ 0],
       [ 0],
       [ 0]], dtype=int32)
```

It seems that the padding values are non-deterministically numerically unstable.

junrushao1994:
@fhieber Thank you so much for pointing out this. Yes, it is expected behavior documented here https://github.com/junrushao1994/mxnet/blob/while_loop_pr/python/mxnet/ndarray/contrib.py#L266

Acutaully I discussed this with my mentor @zheng-da several days ago, and was suggested not to fill a deterministic value in order to ensure performance. Once dynamic shape inference is out, there will be no padding thing. 

Currently if you wants to know the exact number of steps taken, you could add an ‘i’ into loop variables, and increments the ‘i’ on each step.

Thank you so much for the detailed feedback!

fhieber:
Thanks! I actually missed that comment in the documentation, sorry about that. Makes sense to keep the fill value undefined for performance reasons.

I am considering this operator as the core building block for a beam search algorithm capped with max_iterations. I would then do, exactly as you suggested, keep a loop variable that counts the actual number of steps taken to slice the results from the returned, padded output.

benchmark/python/control_flow/while_loop_rnn.py
```diff
@@ -0,0 +1,214 @@
...
+            max_iterations=self.length,
+        )
+        return out + states
+
+def benchmark_rnn(cell, rnn_data, states, length):
+    ctx = rnn_data.context
+    num_batches = 20
+
+    # Imperative
+    cell0 = copy.deepcopy(cell)
```
szha: nit: there's quite a bit of repetition in this function

python/mxnet/ndarray/contrib.py
```diff
@@ -191,3 +191,170 @@ def check_input(inputs, in_type, msg):
...
+    `max_iterations` is a scalar that defines the maximum number of iterations allowed.
+
+    This function returns two lists.
+    The first list has the length of `|step_output|`,
+    in which the i-th element are all i-th elements of
+    `step_output` from all steps, stacked along axis 0.
+    The second list has the length of `|loop_vars|`,
+    which represents final states of loop variables.
+
+    Warning 1: for now, the axis 0 of all NDArrays in the first list are `max_iterations`,
```
szha: Use markup in sphinx for warning?
```
.. warning::
   For now, the axis 0 of all NDArrays in the first list are `max_iterations`, due to lack of dynamic shape inference.
``

File: python/mxnet/ndarray/contrib.py [hunk 26]
```diff
@@ -191,3 +191,170 @@ def check_input(inputs, in_type, msg):
...
+        step_output, loop_vars = _func_wrapper(loop_vars)
+        outputs.append(step_output)
+        steps += 1
+        if len(outputs) != steps or len(step_output) != len(outputs[0]):
+            raise ValueError("Number of elements in step_output should be the same in each step")
+    stacked_outputs = []
+    for i_th, items in enumerate(zip(*outputs), 1):
+        # `mx.ndarray.pad` only support 4-D or 5-D inputs for now
+        # so we could not use it.
+        items = [x.reshape([1] + list(x.shape)) for x in items]
```
szha: use `expand_dims(0)`

python/mxnet/symbol/contrib.py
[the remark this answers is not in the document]
szha: Why the id instead of checking for the symbol directly?

python/mxnet/ndarray/contrib.py
```diff
@@ -191,3 +191,170 @@ def check_input(inputs, in_type, msg):
...
+        if steps != max_iterations and items:
+            pad_shape = [max_iterations - steps] + list(items[0].shape[1: ])
+            pad = ndarray.empty(
```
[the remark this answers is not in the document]
junrushao1994: Autograd does not support inplace operations like in-place writing an NDArray

python/mxnet/symbol/contrib.py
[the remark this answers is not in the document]
junrushao1994: @szha `mx.sym.Symbol.__eq__` has been overridden and returns an NDArray instead of bool. Thus directly using `sym` as keys of a dict won't work.

python/mxnet/ndarray/contrib.py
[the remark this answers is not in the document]
szha: 🤦‍♂️

benchmark/python/control_flow/while_loop_rnn.py
```diff
@@ -0,0 +1,214 @@
...
+            i = states[0]
+            s = states[1: ]
+            data = inputs.take(i).squeeze(axis=0)
+            out, new_s = self.cell(data, s)
+            new_s = [i + 1] + new_s
+            return out, new_s
+        out, states = F.contrib.while_loop(
+            cond=lambda i, *_: i < self.length,
+            func=_func,
+            # lambda i, *s: [i + 1] + list(self.cell(s)),
```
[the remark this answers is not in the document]
junrushao1994: Fixed. Sorry for that T_T

python/mxnet/ndarray/contrib.py
junrushao1994: Cool! Updated

python/mxnet/ndarray/contrib.py
junrushao1994: Fixed.

python/mxnet/symbol/contrib.py
```diff
@@ -336,3 +336,221 @@ def check_data(inputs, in_type, msg):
...
+
+    This function returns two lists.
+    The first list has the length of `|step_output|`,
+    in which the i-th element are all i-th elements of
+    `step_output` from all steps, stacked along axis 0.
+    The second list has the length of `|loop_vars|`,
+    which represents final states of loop variables.
+
+    .. warning::
+    For now, the axis 0 of all Symbols in the first list are `max_iterations`,
```
szha: there should be an empty line before and after the warning block. Also, indentation is needed for it to work. refer to sphinx doc

python/mxnet/symbol/contrib.py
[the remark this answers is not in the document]
szha: OK

benchmark/python/control_flow/foreach_rnn.py
```diff
@@ -157,16 +157,22 @@ def benchmark_rnn(cell, rnn_data, states):
...
              gluon.rnn.LSTMCell(ndim, prefix='rnn_')]
     ctxs = [mx.cpu(0), mx.gpu(0)]
     for cell in cells:
         for ctx in ctxs:
             for batch_size in batch_sizes:
                 if len(get_gpus()) == 0 and ctx == mx.gpu(0):
                     continue
-
-                if isinstance(cell, gluon.rnn.GRUCell):
+                if isinstance(cell, gluon.rnn.RNNCell):
```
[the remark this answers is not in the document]
junrushao1994: @szha Da and I decide that I rewrite these two files. Will push a commit later today.

junrushao1994:
Just now I rewrote the benchmark code for both `foreach` and `while_loop` in a single file.

PR #11582, opened 2018-07-06
Fix typo in clojure project dependencies.

pjstadig:
## Description ##
A minor typo fix for the Clojure contrib project.

I assume the duplicate dependency is a typo. When I checked out and setup the project, it worked when I used the corrected dependency.


gigasquid:
Thanks so much for the correction! 😄 💯

gigasquid:
@szha - This is good to go as soon as CI passes 💚

marcoabreu:
Thanks for the contribution!

@gigasquid can you add it to do the TODO list to validate this execution path in CI?

PR #11584, opened 2018-07-06
CMake will not detect a variable as empty string

fran6co:
## Description ##
Testing it with CMake 3.12 it's not detecting that CUDA_TOOLSET is empty


CMakeLists.txt
```diff
@@ -210,7 +210,7 @@ if(USE_CUDA)
   find_package(CUDA REQUIRED)
   add_definitions(-DMSHADOW_USE_CUDA=1)
   if(FIRST_CUDA AND (NOT USE_OLDCMAKECUDA))
-    if(CUDA_TOOLSET STREQUAL "")
+    if(NOT CUDA_TOOLSET)
```
yajiedesign: may be ()NOT CUDA_TOOLSET) OR ()CUDA_TOOLSET STREQUAL "") is good

fran6co: NOT checks for empty string and when the variable is not set, so there is no need for the OR

yajiedesign: ok

yajiedesign:
LGTM

Issue #10147, opened 2018-03-18
Mxnet R Package Installation Doc Bug


harryprince:
Wrong:

```
# current repo doc:  https://github.com/apache/incubator-mxnet/tree/master/R-package
cran <- getOption("repos")
cran["dmlc"] <- "https://apache-mxnet.s3-accelerate.dualstack.amazonaws.com/R/CRAN/"
options(repos = cran)
install.packages("mxnet",dependencies = T)
```

Correct:
```
cran <- getOption("repos")
cran["dmlc"] <- "https://s3-us-west-2.amazonaws.com/apache-mxnet/R/CRAN/"
options(repos = cran)
install.packages("mxnet",dependencies = T)
```

## Reference

https://stackoverflow.com/questions/43872455/mxnet-package-installation-in-r

ankkhedia:
@sandeep-krishnamurthy  Could you please tag this as Documentation

harryprince:
same issue like:
#10088 
#7985

harryprince:
I summarise an installation example on GPU/linux in detail for Chinese users:
https://segmentfault.com/a/1190000013899075

harryprince:
I recommend someone officially public a stable version mxnet package to CRAN, just like `pip install mxnet-cu80`, which can make installation and development easier.

anirudhacharya:
This issue has been resolved in the latest master. @sandeep-krishnamurthy please close this issue.

@harryprince feel free to reopen if the problem persists.

Issue #8869, opened 2017-11-29
Outdated documentation for installing R packages


mjmg:
## Description
For mxnet 12.1 the documentation for installing R packages seems to be outdated.
Currently these are the commands:
```
make rpkg
R CMD INSTALL mxnet_current_r.tar.gz
```
Upon checking the Makefile, using the command 
```
make rpkg
```
seems to include installation of the R package from the R-package subdirectoy directly and no mxnet_current_r.tar.gz is generated.

Please see https://github.com/apache/incubator-mxnet/commit/0d294dd876e54023ed5e9c5676092a36a1566ab3#diff-b67911656ef5d18c4ae36cb6741b7965 and https://github.com/apache/incubator-mxnet/commit/93cd0178a012c4867550aae9ce887f08d41aee6b#diff-58231b16fdee45a03a4ee3cf94a9f2c3

## Error Message:
```
Warning: invalid package 'mxnet_current_r.tar.gz'
Error: ERROR: no packages specified
```

## Steps to reproduce
```
$ make rpkg
....
$ R CMD INSTALL mxnet_current_r.tar.gz
```

spirosparaskevas:
First of all thanks for confirming this situation. I have just verified this awkwardness. Can u suggest a workaround?

anirudhacharya:
this PR - https://github.com/apache/incubator-mxnet/pull/11590 will fix this issue.

anirudhacharya:
This issue has been resolved in the latest master. @sandeep-krishnamurthy please close this issue.

@mjmg feel free to reopen if the problem persists.

Issue #8240, opened 2017-10-12
Failing to create R package on Linux.


hadjipantelis:
Hello, I try to create the R-package to install from source on CentOS 7 but loading the package during package building fails. I follow the instructions [here](https://mxnet.incubator.apache.org/get_started/install.html).
In the past (around Feb-March 2017) I have installed and ran `mxnet` fine on this particular system but now the `make rpkg` fails so I am slightly puzzled what is going on. I have the same error as https://github.com/apache/incubator-mxnet/issues/1104 but the proposed solutions (`make clean`/ `make clean_all` prior to built) do not work. Building the MXNet core shared library works fine on face value, but building the MXNet R binding/package fails.

The exact commands I use are:
```
rm mxnet/ -rf
git clone --recursive https://github.com/apache/incubator-mxnet.git mxnet --branch 0.11.0
cd mxnet/
make clean_all #(I tried simple make clean too)
cp make/config.mk .
echo "ADD_CFLAGS += -I/usr/include/openblas" >>config.mk
make -j $(nproc) USE_OPENCV=0 USE_BLAS=openblas USE_LAPACK=1
make rpkg
```
The error that I get is:
```
Error: package or namespace load failed for ‘mxnet’:
 .onLoad failed in loadNamespace() for 'mxnet', details:
  call: dyn.load(file, DLLpath = DLLpath, ...)
  error: unable to load shared object '/home/userh/R/x86_64-redhat-linux-gnu-library/3.4/mxnet/libs/mxnet.so':
  /home/userh/R/x86_64-redhat-linux-gnu-library/3.4/mxnet/libs/mxnet.so: undefined symbol: _ZN2cv8fastFreeEPv
```

R environment:
```
R version 3.4.1 (2017-06-30)
Platform: x86_64-redhat-linux-gnu (64-bit)
Running under: CentOS Linux 7 (Core)

Matrix products: default
BLAS/LAPACK: /usr/lib64/R/lib/libRblas.so

locale:
 [1] LC_CTYPE=en_GB.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_GB.UTF-8        LC_COLLATE=en_GB.UTF-8    
 [5] LC_MONETARY=en_GB.UTF-8    LC_MESSAGES=en_GB.UTF-8   
 [7] LC_PAPER=en_GB.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[...]

PR #11590, opened 2018-07-06
[MXNET-627] Fix Installation instructions for R bindings on Linux systems.

anirudhacharya:
Fix for https://github.com/apache/incubator-mxnet/issues/8240 https://github.com/apache/incubator-mxnet/issues/10147 https://github.com/apache/incubator-mxnet/issues/8869

Preview of the website -  http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-11590/$BUILD_NUM/index.html

@hetong007 @aaronmarkham @ankkhedia


docs/install/index.md
```diff
@@ -987,10 +987,23 @@ $ make -j $(nproc) USE_OPENCV=1 USE_BLAS=openblas
 
 **Build and install the MXNet R binding**
 
+You will need to first install R. MXNet requires R version >3.3 and we can fetch that from CRAN package repository.
```
hetong007: Please add similar instructions for windows and os x users.

anirudhacharya: for macOS and windows we do not build from source, this will not be there. There is only the issue of ensuring that the pre-built binary in the S3 works fine with both R 3.4 and 3.5. 
And that is not part of this PR.

aaronmarkham:
Follow these directions to setup a test server and build the docs. Test the selector to make sure content shows up when you click on the different install options.

https://github.com/apache/incubator-mxnet/blob/master/docs/build_version_doc/README.md#setting-up-a-docs-dev-server

anirudhacharya:
@aaronmarkham It does show here correctly - http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/install/index.html?platform=Linux&language=R&processor=CPU

aaronmarkham:
So it looks like we have duplicated information on the install.md page and each of the $OS_setup.md pages. Also, when you use the install page's selector, you get instructions, but no link to the $OS_setup page.

Wouldn't it be better to provide the minimum info needed on the install page, and leave the detail for the OS specific pages?

anirudhacharya:
@aaronmarkham you have been added as a collaborator to my repo. Please make the required changes and push to this branch. Thanks!

aaronmarkham:
I went ahead and resolved an issue with using the CI scripts as part of the installation instructions for Ubuntu and made a new script for core and python deps. I also made an R installation script.
The instructions are generally simplified now by having access to the two new shell scripts.

**install_mxnet_ubuntu_python.sh** 
* It checks for `nvidia-smi` and sets the make flags accordingly, so now we don't have doubled up instructions.
* I moved the pip installs to a central requirements.txt so that's not duplicated anymore.
* It uses sudo in the pip upgrade step - I didn't find a way around this.

**install_mxnet_ubuntu_r.sh**
* The R deps are now in alignment - I updated CI for R to match.

aaronmarkham:
@hetong007 I updated the install page for OSX and Windows to point to the detailed instructions pages. OSX should be good to go now. Windows is being worked on in a separate PR by @ankkhedia. Please revisit your change request since further updates on Windows instructions is out of scope for this PR.

ci/docker/install/ubuntu_r.sh
```diff
@@ -20,17 +20,31 @@
 # build and install are separated so changes to build don't invalidate
 # the whole docker cache for the image
 
+# Important Maintenance Instructions:
```
KellenSunderland: This is probably unintended, but the change to this file causes CI to occasionally fail.  The keyserver for this gpg key fails approx 1/10 requests.  Is there a good reason we reverted @larroy change which manually installs the key?   It seemed like a reasonable fix for the flaky gpg server issue.  Any objection to me restoring the manual key add?

PR #11357, opened 2018-06-22
[MXNET-331] Single machine All Reduce Topology-aware Communication

ctcyang:
## Description ##
Single machine All Reduce Topology-aware Communication

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] Proposed communication method shows speed-up compared to both existing methods (parameter server and NCCL) on small batch sizes for ResNet-50, VGG-16, Inception-v3 and AlexNet.
- [x] Communication method queries the single-machine multi-GPU link topology, and determines a suitable communication pattern to use.
- [ ] In future, will add auto-tuner to automatically choose between single-machine communication protocols (parameter server, NCCL, method proposed here).

## Comments ##
- [Design Proposal](https://cwiki.apache.org/confluence/display/MXNET/Single+machine+All+Reduce+Topology-aware+Communication)

PR #11591, opened 2018-07-06
[MXNET-331] Single machine All Reduce Topology-aware Communication (Updated)

ctcyang:
## Description ##
Single machine All Reduce Topology-aware Communication

### Essentials ###

### Changes ###
- [x] Proposed communication method shows speed-up compared to both existing methods (parameter server and NCCL) on small batch sizes for ResNet-50, VGG-16, Inception-v3 and AlexNet.
- [x] Communication method queries the single-machine multi-GPU link topology, and determines a suitable communication pattern to use.
- [x] Use feature by `MXNET_KVSTORE_USETREE=1`, default is 0.
- [x] Add knobs for tuning this communication method.
- [ ] In future, will add auto-tuner to automatically choose between single-machine communication protocols (parameter server, NCCL, method proposed here).

## Comments ##
- [Design Proposal](https://cwiki.apache.org/confluence/display/MXNET/Single+machine+All+Reduce+Topology-aware+Communication)
- There are comments from several reviewers are here: https://github.com/apache/incubator-mxnet/pull/11357


marcoabreu:
Please note that kvstore tests are currently disabled

src/kvstore/comm_tree.h
```diff
@@ -0,0 +1,500 @@
...
+                             int merged_row, int priority) {
+    std::vector<std::vector<NDArray>> reduce(devs_.size());
+
+    TreeBufferEntry& random_buf = tree_merge_buf_[0][key];
+    const NDArrayStorageType stype = random_buf.merged[0].storage_type();
+    std::vector<size_t>& topology = topology_[root];
+    NDArray buf_slice;
+
+    if (stype == kDefaultStorage) {
+      // Copy everything into buf.merged for each gpu
```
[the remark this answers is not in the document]
ctcyang: We can avoid this copy, but it makes the code harder to read. Since the source and destination belong to the same GPU, the gain is minimal (10 samples/s i.e. 720 samples/s to 730 samples/s).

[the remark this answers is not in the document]
ctcyang: I tested the throughput difference on a similar intra-GPU `CopyFromTo()`, by commenting out Line 306 in `comm_tree.h`. This ruins the correctness of the output, but it gives an idea of how much savings can be gotten. Testing on VGG-16 on an older commit, I got the following. Geomean difference is getting rid of one CopyFromTo makes it 2.2% faster.

```
v6: One more intra-GPU CopyFromTo than v7
fp32:
BS | v6   | v7
4  | 711  | 745
8  | 999  | 1035
16 | 1449 | 1478
32 | 1638 | 1672
64 | 1695 | 1739

fp16:
BS | v6   | v7
8  | 1552 | 1599
16 | 2127 | 2163
32 | 2916 | 2910
64 | 2720 | 2775
128| 2518 | 2532
```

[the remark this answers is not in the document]
xcgoner: Thanks! 
Have you tried to use tools/bandwidth/measure.py to test the throughput?
According to my experiments, tree mode outperforms device mode on p3.x16.
However, on p2.x16, the performance of tree mode seems worse than device mode.
Could you double-check that?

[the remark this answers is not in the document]
ctcyang: That makes to me, because p2.16x topology is fully connected using GPUDirect PCI-E. In fully connected networks, tree will never outperform device mode. 

The advantage on p3.16x is that device mode is forced to use 4 PCI-E (not GPUDirect, so must go through CPU/QPI), but tree is all NVLink, so it avoids CPU/QPI latency.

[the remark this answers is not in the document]
xcgoner: Actually, p2.x16 has a worse communication topology compared to p3.x16.
For p2.x16, the topology is very special. The communication net is composed of 2 fully connected subnets: the first 9 gpus and the last 7 gpus. Unlike p3.x16, there are no connections between the 2 components.
In that case, I expect that tree mode should still outperform device mode. However, surprisingly the device mode is better. Actually, the performance is similar to local mode on p2.x16, which seems weird to me.

ctcyang:
Yeah, I'm blocked by the test. I can't replicate it on local machine, but I can replicate it on Docker image.

ctcyang:
@haojin2 @eric-haibin-lin @rahul003 Trying to get this into 1.3 release

src/kvstore/comm_tree.h
```diff
@@ -0,0 +1,518 @@
...
+
+          if (!is_dest) {
+            if (reduce[dest_id].size() == 0) {
+              reduce[dest_id].push_back(buf_dest.merged[merged_row]);
+            }
+          } else {
+            if (dest_id != topo_id) {
+              CopyFromTo(buf_from.merged[merged_row],
+                  &(buf_dest.copy_buf[merged_row][is_dest-1]),
+                  priority);
```
haojin2: Align the lines like:
```c++
CopyFromTo(buf_...,
           &(buf_...,
           priority);
```

tests/cpp/kvstore/gpu_topology_test.cc
```diff
@@ -0,0 +1,672 @@
...
+
+  bool connected = mxnet::kvstore::IsConnected(W, num_gpus);
+
+  bool correct_connected = true;
+  ASSERT_EQ(connected, correct_connected);
+}
+
+// ComputeTreesTest with backtracking
+// TODO(carlyang): comment out test for now
+/*TEST(GpuTopology, TestComputeTrees1) {
```
rahul003: What's wrong with these tests? Do they not work?

tests/python/gpu/test_nccl.py
```diff
@@ -29,7 +30,51 @@
...
+        try:
+            self._prev_val = os.environ[self._key]
+        except KeyError:
+            self._prev_val = ''
+        os.environ[self._key] = self._next_val
+
+    def __exit__(self, ptype, value, trace):
+        os.environ[self._key] = self._prev_val
+
+def test_device_pushpull():
```
rahul003: Why is this test in the file test_nccl?

src/kvstore/comm_tree.h
```diff
@@ -0,0 +1,518 @@
...
+ * It is faster if the total device-to-device bandwidths is larger than
+ * device-to-cpu, which is often true for 4 or 8 GPUs. But it uses more device
+ * memory.
+ */
+class CommDeviceTree : public CommDevice {
+ public:
+  CommDeviceTree() {
+    inited_ = false;
+    gpuarray_bound_ = dmlc::GetEnv("MXNET_KVSTORE_GPUARRAY_BOUND", 10000000);
+    backtrack_ = dmlc::GetEnv("MXNET_KVSTORE_BACKTRACK", 0);
```
rahul003: Could you document these environment variables in https://github.com/apache/incubator-mxnet/blob/master/docs/faq/env_var.md

src/kvstore/comm_tree.h
```diff
@@ -0,0 +1,518 @@
...
+    }
+#endif
+  }
+
+  void QueryTopology() {
+#if MXNET_USE_CUDA
+    std::vector<float> link_matrix(devs_.size()*devs_.size());
+    GetP2PWeight(devs_, &link_matrix);
+    if (backtrack_)
+      LOG(WARNING) << "Using Backtracking to generate trees";
```
rahul003: This should be LOG(INFO)

src/kvstore/gpu_topology.h
```diff
@@ -0,0 +1,1061 @@
...
+      if (cost > *g) {
+        *g = cost;
+        *a = row;
+        *b = col;
+      }
+    }
+  }
+}
+
+// Performs partition on each existing partition in graph W if partition has
```
rahul003: Thanks for the great comments for the functions, but could you fix comment style to how it's standard in the codebase. Here's an example https://github.com/apache/incubator-mxnet/blob/b4156da26cfe741619227ae726872b1255194900/src/kvstore/kvstore_utils.h#L37

src/kvstore/comm_tree.h
```diff
@@ -0,0 +1,518 @@
...
+  ///        new intent: for every gpu: store key->memory mapping
+  std::vector<std::unordered_map<int, TreeBufferEntry>> tree_merge_buf_;
+
+  /// \brief NVLink-connected topology in full binary tree format
+  std::vector<std::vector<size_t>> topology_;
+  std::vector<std::vector<size_t>> scan_;
+  std::vector<Context> devs_;
+
+  /// \brief Highest numbered device
+  int max_dev_;
```
rahul003: I tried to see where this variable is used to ensure that cases when gpus '1,5,3,7' are given work. But it looks like this variable is not used? Please remove this then

src/kvstore/comm_tree.h
```diff
@@ -0,0 +1,518 @@
...
+      int start = scan_[0][depth_];
+      int end = scan_[0][depth_+1];
+
+      // In order to generalize to any number of GPUs, we use strategy of having
+      // found the mapping from 0, 1, ..., n_gpus to dev_id i.e.
+      //   idx:    0 1 2 3 4 5 6
+      //   dev_id: 4 2 3 1 7 5 0
+      // and generated an n_gpus x n_gpus link topology matrix:
+      //
+      // 1) The reduction trees are saved as indices on 0, 1, ..., n_gpus
```
rahul003: Could you clarify how many are generated

src/kvstore/comm_tree.h
rahul003: Many other places as well where this change ought to be done

ctcyang:
@Roshrini For keeping track of PR

src/kvstore/comm_tree.h
ctcyang: Added documentation.

tests/cpp/kvstore/gpu_topology_test.cc
ctcyang: They used to segfault only on CI, but now they should be fine. I fixed an off-by-1 bug.

docs/faq/env_var.md
```diff
@@ -83,6 +83,31 @@ export MXNET_GPU_WORKER_NTHREADS=3
...
+  - Values: 0(false) or 1(true) ```(default=0)```
+  - If true, MXNet tries to use tree reduction for Push and Pull communication.
+  - Otherwise, MXNet uses the default Push and Pull implementation.
+  - [Tree reduction technology](http://www.sysml.cc/doc/178.pdf) has been shown to be faster than the standard ```--kv-store device``` Push/Pull and ```--kv-store nccl``` Push/Pull for small batch sizes.
+
+* MXNET_KVSTORE_LOGTREE
+  - Values: 0(false) or 1(true) ```(default=0)```
+  - If true and MXNET_KVSTORE_USETREE is set to 1, MXNet will log the reduction trees that have been generated.
+
+* MXNET_KVSTORE_GPUARRAY_BOUND
```
rahul003: I realize it says multiple trees, but could you call out that this is for tree kvstore? especially because we have a similar variable MXNET_KVSTORE_BIGARRAY_BOUND

docs/faq/env_var.md
```diff
@@ -83,6 +83,31 @@ export MXNET_GPU_WORKER_NTHREADS=3
...
+  - Values: Int ```(default=10000000)```
+  - The minimum size of a "big array".
+  - When the array size is bigger than this threshold and MXNET_KVSTORE_USETREE is set to 1, multiple trees are used to load balance the big gradient being communicated in order to better saturate link bandwidth.
+
+* MXNET_KVSTORE_BACKTRACK
+  - Values: 0(false) or 1(true) ```(Default=0)
```
rahul003: Formatting issue

PR #11596, opened 2018-07-07
[MXNET-531] Integration Test for Scala

lanking520:
## Description ##
This is an approach for us to do integration test on Scala. As more of the examples are adding in the Scala CI, we need a separation between tests to make sure it test different things. However, this solution is not perfect due to this fact:

I have placed `verify` stage on the `example` module and change the unit test to integration test. If we run the verify stage, the unit test will still be run though it takes us 1 ~ 2 minutes to finish them.

@nswamy @marcoabreu @yzhliu @andrewfayres  

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -712,18 +712,6 @@ try {
         }
       }
     },
-    'Scala: GPU': {
-      node('mxnetlinux-gpu') {
-        ws('workspace/ut-scala-gpu') {
-          timeout(time: max_time, unit: 'MINUTES') {
-            init_git()
-            unpack_lib('gpu', mx_dist_lib)
-            docker_run('ubuntu_gpu', 'unittest_ubuntu_gpu_scala', true)
-            publish_test_coverage()
-          }
-        }
-      }
-    },
     'Clojure: CPU': {
       node('mxnetlinux-cpu') {
         ws('workspace/ut-clojure-cpu') {
@@ -975,6 +963,17 @@ try {
             unstash 'cpp_test_score'
             unstash 'cpp_test_optimizer'
             docker_run('ubuntu_gpu', 'integrationtest_ubuntu_gpu_cpp_package', true)
+            publish_test_coverage()
+          }
+        }
+      },
+    'Scala: GPU': {
+      node('mxnetlinux-gpu') {
+        ws('workspace/ut-scala-gpu') {
+          timeout(time: max_time, unit: 'MINUTES') {
+            init_git()
+            unpack_lib('gpu', mx_dist_lib)
+            docker_run('ubuntu_gpu', 'integrationtest_ubuntu_gpu_scala', true)
             publish_test_coverage()
           }
         }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -966,7 +966,8 @@ try {
             publish_test_coverage()
           }
         }
-      },
+      }
+    },
     'Scala: GPU': {
       node('mxnetlinux-gpu') {
         ws('workspace/ut-scala-gpu') {
--- a/Makefile
+++ b/Makefile
@@ -593,7 +593,13 @@ scalapkg:
 			-Dcurrent_libdir="$(ROOTDIR)/lib" \
 			-Dlddeps="$(LIB_DEP) $(ROOTDIR)/lib/libmxnet.a")
 
-scalatest:
+scalaunittest:
+	(cd $(ROOTDIR)/scala-package; \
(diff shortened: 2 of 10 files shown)


Makefile
```diff
@@ -593,7 +593,13 @@ scalapkg:
...
 			-Dlddeps="$(LIB_DEP) $(ROOTDIR)/lib/libmxnet.a")
 
-scalatest:
+scalaunittest:
+	(cd $(ROOTDIR)/scala-package; \
+		mvn integration-test -P$(SCALA_PKG_PROFILE),$(SCALA_VERSION_PROFILE) -Dcxx="$(CXX)" \
+			-Dcflags="$(CFLAGS)" -Dldflags="$(LDFLAGS)" \
+			-Dlddeps="$(LIB_DEP) $(ROOTDIR)/lib/libmxnet.a" $(SCALA_TEST_ARGS))
+
+scalaintegrationtest:
```
nswamy: where do you specify to run tests only under examples ? does it also run unit tests?

lanking520: Currently, all unit test are run on `integration-test` stage in Maven and all integration test runs in `verify` stage. 

Yes, it runs unit test as well

nswamy: I am not sure I see the benefit of creating another stage that runs twice. Am I missing anything ? consider adding the test phase in examples under a maven profile which can passed when you call make integrationtests ?

lanking520: It works like this way:
if you put test in the `test` phase, test will be running in `test`, `package` and all after phases.
If you put test in the `integration-test`, test will be running in `verify`, but not in `package`. 
So it is impossible to keep test runs on a individual phase.

lanking520: Some alternative is applicable for tests done in Java. However we are using `scalatest` plugin, the plugins for Java surefire cannot used in here.

lanking520:
I have make the best use of profile and I think it should works fine without duplicated tests. There is also changes on the build pipeline, please let me know your idea. @nswamy @yzhliu

Makefile
nswamy: Can you not run `verify` of examples under a called profile `integration-test`

lanking520: Now there is no verify. Only Integration test.

yzhliu:
good to me

Makefile
```diff
@@ -593,9 +593,15 @@ scalapkg:
 			-Dcurrent_libdir="$(ROOTDIR)/lib" \
 			-Dlddeps="$(LIB_DEP) $(ROOTDIR)/lib/libmxnet.a")
 
-scalatest:
+scalaunittest:
```
yzhliu: remember to update README

marcoabreu:
Yes, the Scala integration tests are now in parallel with the unit tests. At some point we might move them to nightly, but for now, you don't have to worry but just categorize the tests properly.

Makefile
nswamy: I tried to centralize the two profiles in the main pom file instead of changing these properties in every module. The test plugins are spread in all the modules, After cleaning them and adding two profiles in the main pom <tagExcludes> configuration. The tests are now segfaulting, not sure what is causing. I will stop this effort here and we can see later.

nswamy:
@lanking520 thanks for separating the tests into two phases, this definitely helps in speeding up the CI pipeline.
@marcoabreu consider moving Scala Integration tests to the next stage, we don't need to run Unit Tests and Integration tests in parallel

Jenkinsfile
```diff
@@ -712,18 +712,6 @@ try {
         }
       }
     },
-    'Scala: GPU': {
```
marcoabreu: @nswamy @lanking520 was this intentional? I think you just disabled all unit tests for Scala on GPU ._.

Jenkinsfile
```diff
@@ -979,6 +967,18 @@ try {
           }
         }
       }
+    },
+    'Scala: GPU': {
```
marcoabreu: Please name this Scala Integration: GPU

marcoabreu:
Could you elaborate, naveen? When running integration tests, does it automatically cover unit tests as well? I'm afraid merging them will
1. Increase the time of the critical path
2. Make it hard to distinguish between unittests and integration tests later on - right now it looks like we're not running unit tests at all

lanking520:
@marcoabreu You are right, the CPU should be called as Scala Unit test CPU. Currently, we do not have any unit test running on GPU. Only some tests in integration test run GPU. I will amend with another PR to add this fixes.

PR #11604, opened 2018-07-07
Fix scala maven repos

GnaneshKunal:
## Description ##
The Maven Repo of Linux GPU and MacOS CPU were interchanged in the Scala package installation page. So made the necessary changes.


marcoabreu:
Thanks for noticing and submitting a fix!

PR #11606, opened 2018-07-07
ColorNormalizeAug now accepts mxnet NDArrays as well as stated (wrongly) in the documentation.

ifeherva:
According to the doc of mx.image.ColorNormalizeAug mean and std must be mxnet NDArrays though in reality it accepts only numpy arrays. This commit allows ColorNormalizeAug to accept both numpy and mxnet ndarrays.

Extended the image unit tests to cover this functionality.

### DIFF
--- a/tests/python/unittest/test_image.py
+++ b/tests/python/unittest/test_image.py
@@ -155,6 +155,17 @@ class TestImage(unittest.TestCase):
 
 
     def test_augmenters(self):
+        # ColorNormalizeAug
+        mean = np.random.rand(3) * 255
+        std = np.random.rand(3) + 1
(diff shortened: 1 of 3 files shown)


tests/python/unittest/test_image.py
```diff
@@ -155,6 +155,17 @@ def test_imageiter(self):
 
 
     def test_augmenters(self):
```
marcoabreu: Please add the with-seed annotation

ifeherva: Could you please point to an example so I can add it?

ifeherva: Nevermind, found it :)

PR #11435, opened 2018-06-27
Replace PTB dataset to Sherlock Holmes 

liyujiel:
## Description ##
Replace PTB dataset to self-created dataset Sherlock Holmes 

Replace reason: https://github.com/dmlc/web-data/pull/75

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] files related with Penn TreeBank dataset

## Comments ##
- the model parallel LSTM example cannot run with Penn TreeBank dataset originally, created an issue and wait for contributor update

PR #11626, opened 2018-07-10
[MXNET-651] MXNet Model Backwards Compatibility Checker

piyushghai:
## Description ##
This PR aims to check whether models trained on earlier versions of MXNet are loading fine on the latest version or the latest release candidate. It also aims to do a check for consistency in the inference  results on these trained models.

### Essentials ###
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/browse/MXNET-651) created (except PRs with tiny changes)
The PRs purpose itself is to add more tests for backwards compatibility. 
- A README.md has been added to explain the what Model Backwards Compatibility is and how it can be run.

### TODO ###
- Check the tasks using an actual Jenkins Slave. @marcoabreu Can you help me out with that ? 
- Backfill the S3 bucket with models trained on earlier versions of MXNet.


piyushghai:
Pinging @marcoabreu @ThomasDelteil @piiswrong  for review. 
@marcoabreu Please let me know if I did the CI integration part correctly on this PR or if I need to make more changes to it.

lupesko:
Thanks for this important contribution!

A not about the description, which is lacking, and should be updated as needed, rather than be left with un-relevant stuff from the template. A few examples: 
- the issue description says "The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant JIRA issue created (except PRs with tiny changes)"
This issue is not tiny, and yet it has no JIRA. Please follow the guidelines and file a JIRA and update the issue name.
- "All changes have test coverage" is not checked. I think it should be checked, or else you should add the relevant coverage.
- I recommend you go over the issue description and update as needed. No need to leave things that are not relevant for the PR, and on the other hand add more details on bullet points that are relevant.

piyushghai:
@szha Here's what this PR does : 
"Backward compatibility aims to check whether models trained on earlier versions of MXNet are loading fine on the latest version or the latest release candidate. It also aims to do a sanity check for consistency in the inference on these trained models. " 

Here's the relevant JIRA Issue for it : https://issues.apache.org/jira/browse/MXNET-651

szha:
What kind of "loading" are you testing? The statement, as is, is underspecified.

piyushghai:
I am checking the following scenarios to save/load models : 
1. Hybridized models import/export API from [Gluon Package (https://mxnet.incubator.apache.org/api/python/gluon/gluon.html)
2. Gluon Models load_params/save_params API from [Gluon Package] (https://mxnet.incubator.apache.org/api/python/gluon/gluon.html)
3. Gluon Models load_parameters/save_parameters API from [Gluon Package] (https://mxnet.incubator.apache.org/api/python/gluon/gluon.html)
4. Declarative Models load_checkpoint() from [Model API] (https://mxnet.incubator.apache.org/api/python/model.html)

szha:
Any reason not to do it in unittest, using a smaller model?

szha:
2 is deprecated so there's no need to test.

szha:
How do you test load_parameters and save_parameters? What additional case would the new test case for gluon's save_parameters/load_parameters cover that's not already covered by https://github.com/apache/incubator-mxnet/blob/master/tests/python/unittest/test_gluon.py#L1093 ?

piyushghai:
We cannot unittest it as we are not testing the various model save/load APIs on the same version. We are trying to check whether models trained (and saved) on older MXNet versions are still loaded correctly on the latest release candidate of MXNet and thus any new changes that might have crept in remain backwards compatible. 
@marcoabreu can give more context on this change as well.

tests/nightly/model_backwards_compatibility_check/common.py
```diff
@@ -165,3 +177,163 @@ def forward(self, x):
...
+    def __init__(self, path):
+        self.dictionary = Dictionary()
+        self.download_data_from_s3()
+        self.train = self.tokenize(path + 'train.txt')
+        self.valid = self.tokenize(path + 'valid.txt')
+        self.test = self.tokenize(path + 'test.txt')
+
+    def download_data_from_s3(self, ):
+        print ('Downloading files from bucket : ptb-small-dataset' )
+        bucket = s3.Bucket('ptb-small-dataset')
```
ThomasDelteil: where are we downloading this from? since we can't host the ptb dataset ourselves because of the license, I wonder if it is a good idea to depend on it.

tests/nightly/model_backwards_compatibility_check/lm_rnn_gluon_inference.py
```diff
@@ -17,197 +17,31 @@
...
-    bucket = s3client.Bucket(bucket_name)
-    result = bucket.meta.client.list_objects(Bucket=bucket.name,
-                                         Delimiter=backslash)
-    folder_list = list()
-    for obj in result['CommonPrefixes']:
-        folder_list.append(obj['Prefix'].strip(backslash))
+	'''This function returns the top level folders in the S3Bucket. These folders help us to navigate to the trained model files stored for different MXNet versions. '''
+	bucket = s3client.Bucket(bucket_name)
+	result = bucket.meta.client.list_objects(Bucket=bucket.name,
+										 Delimiter=backslash)
```
ThomasDelteil: fix indent

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
```diff
@@ -0,0 +1,57 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+#Author: Piyush Ghai
+
+run_models() {
+	echo '=========================='
+	echo "Running training files and preparing models"
+	echo '=========================='
+	python mnist_mlp_module_api_train.py
```
ThomasDelteil: what are the consequences of having one of these scripts crash with respect to the integrity of the generated data?

tests/nightly/model_backwards_compatibility_check/common.py
piyushghai: Oh. I wasn't aware of the Penn Tree Bank dataset license issue. I had uploaded a trimmed down version of PTB to an S3 bucket. The trimmed down version was 10k training sentences and 3k each for test and validation. 
Can you suggest a different dataset which we can host ?

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
piyushghai: I did not quite fully understand by what you meant by integrity here. 
Are you suggesting that I do a checksum validation on the generated models and other relevant files while downloading them in the inference files ?

ThomasDelteil: if one of the training script randomly crashes, let's say for dataset unavailability, what are the consequences of the fact that the data stored in s3 will be incomplete for that given run?

piyushghai: That's gracefully handled in this commit : https://github.com/apache/incubator-mxnet/pull/11626/commits/7c41488fe3d8679ec12a5d0694a596cc6a72017d 

As of now, I'm exiting the inference script with a message indicating no models found in S3.

tests/nightly/model_backwards_compatibility_check/common.py
szha: See the dataset introduced in #11435

piyushghai: Thanks. Replaced PTB with Sherlock in : c3c91295b7c4e11241b5a32d4885c3d775f052b6

tests/nightly/model_backwards_compatibility_check/lm_rnn_gluon_inference.py
piyushghai: Done. Fixed in : 34853525c99a72c114c2f754403de421b0854409

marcoabreu:
Please note that we are interested in the model architecture itself. Thus,
I'd like to propose that we only use random weights and run inference on
randon data - I don't know if we actually have to train at all. Then save
that model as well as the input and output. Then run forward on the latest
version and ensure the output is identical.

We don't care whether the values actually make sense. We are rather
interested in whether they are identical. This means that we don't need the
overhead of real models and training. I'm happy to get corrected here if my
assumption is wrong.

piyushghai:
@marcoabreu I totally agree with you. I'm currently working adding toy examples which use random weights on simple model architectures and run one epoch through them. 

However, I also wanted to add practical real world models as use cases as well to this. The training time is not an overhead because of small sizes of datasets used and the number of epochs is also small (2 for now). 

For both of these the goal remains the same : Verify the inference values and assert that they are same.

marcoabreu:
The question here is whether we have to run training at all - or if it is sufficient to have the forward and the backward pass separately verified. The scope of this task here is to verify the forward pass.

I don't see much benefit in using a real world dataset if we are only training one or two epochs. The overhead on the other hand is the sheer size of the datasets although they don't really add value to the test. These datasets can get really big and have license problems.

piyushghai:
@ThomasDelteil What are your thoughts on using real world datasets v/s random ones for this ?

szha:
For loading in past versions, there's no need to perform repeated tests as the code is not being updated. We only need to test them if there's a patch.

For new versions, you can simply generate and save random parameters on s3, and use those in unit test to make sure it loads correctly.

Let's avoid copying the example codes.

piyushghai:
@marcoabreu I have removed real world datasets, and instead I am now using randomly generated datasets. I have also removed training part from all the models.

tests/nightly/model_backwards_compatibility_check/JenkinsfileForMBCC
```diff
@@ -0,0 +1,79 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+
+//This is a Jenkinsfile for the model backwards compatibility checker. The format and some functions have been picked up from the top-level Jenkinsfile.
```
marcoabreu: For now, we will have this in a separate Jenkinsfile for testing purposes. In future, I'd like to make this part of the regular Jenkins nightly pipeline.

ci/docker/runtime_functions.sh
```diff
@@ -854,6 +854,27 @@ nightly_test_javascript() {
...
+}
+
+#Backfills S3 bucket with models trained on earlier versions of mxnet
+nightly_model_backwards_compat_train() {
+    set -ex
+    export PYTHONPATH=./python/
+    VENV=mbcc_py2_venv
+    virtualenv -p `which python2` $VENV
+    source $VENV/bin/activate
+    pip install boto3
```
marcoabreu: Please note (and add a comment or move it somewhere else) that this script can not be called within our regular docker environment because the IAM roles do not get propagated into the container as of now. Also, we have security restrictions here that will have to be addressed before we can bring this into production. I'll cover that later.

tests/nightly/model_backwards_compatibility_check/common.py
```diff
@@ -0,0 +1,199 @@
...
+
+# Set fixed random seeds.
+mx.random.seed(7, ctx = mx.cpu())
+np.random.seed(7)
+logging.getLogger().setLevel(logging.DEBUG)
+
+# get the current mxnet version we are running on
+mxnet_version = mx.__version__
+model_bucket_name = 'mxnet-model-backwards-compatibility-models'
+data_bucket_name = 'mxnet-model-backwards-compatibility-data'
```
marcoabreu: Do we really need two separate buckets? I'd prefer if we could only use one to reduce the maintenance overhead.

tests/nightly/model_backwards_compatibility_check/model_backwards_compat_inference.py
```diff
@@ -0,0 +1,168 @@
...
+	        if len(model_files_meta) == 0:
+	            print ('No trained models found under path : %s' %prefix)
+	            continue
+	        model_files = list()
+	        ## For each file found under a model folder : 
+	        for obj in model_files_meta:
+	            file_name = obj.key.split('/')[2]
+	            model_files.append(file_name)
+	            ## Download this file---
+	            bucket.download_file(obj.key, file_name)
```
marcoabreu: where this it download the file to?

tests/nightly/model_backwards_compatibility_check/model_backwards_compat_inference.py
```diff
@@ -0,0 +1,168 @@
...
+
+def test_lenet_gluon_load_params_api():
+	model_name = 'lenet_gluon_save_params_api'
+	print ('Performing inference for model/API %s' %model_name)
+	## Get data from S3
+	data = download_data_from_s3(model_name)
+	
+	test_data = data['data']
+
+	for folder in get_top_level_folders_in_bucket(s3, model_bucket_name):
```
marcoabreu: This seems like a lot of duplicate code. Could we extract it?

tests/nightly/model_backwards_compatibility_check/model_backwards_compat_train.py
```diff
@@ -0,0 +1,159 @@
...
+
+	upload_model_files_to_s3(files, mxnet_folder)
+
+	clean_model_files(files, model_name)
+
+def train_lstm_gluon_save_parameters_api():
+	## If this code is being run on version >= 1.2.0 only then execute it, since it uses save_parameters and load_parameters API
+	if compare_versions(str(mxnet_version), '1.2.1')  < 0:
+		print ('Found MXNet version %s and exiting because this version does not contain save_parameters and load_parameters functions' %str(mxnet_version))
+		sys.exit(1)
```
marcoabreu: are we sure we want to exit instead of just returning?

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
```diff
@@ -0,0 +1,51 @@
...
+	pip install boto3
+}
+
+echo `pwd`
+cd tests/nightly/model_backwards_compatibility_check
+echo `pwd`
+
+install_boto3
+
+install_mxnet 1.1.0
```
marcoabreu: Could we please make this list automatic by traversing the git branches?

marcoabreu:
Thanks a lot, looks very good for a first iteration!

marcoabreu:
I have created a job at http://jenkins.mxnet-ci.amazon-ml.com/job/test-backwards-compatibility-checker/. This job runs your Jenkinsfile and should be retriggered (SCM polling is 1h) if you make changes to your repository.

tests/nightly/model_backwards_compatibility_check/model_backwards_compat_train.py
piyushghai: My bad. It should have been return. Will update it.

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
piyushghai: I'm not sure if I fully understood your comment here @marcoabreu . Git branches generally have a lot of tags such 1.x.x.rc, 1.x.x.nightly etc...  How can we filter out only the actual releases from the git tags ?

tests/nightly/model_backwards_compatibility_check/model_backwards_compat_train.py
piyushghai: Done.

tests/nightly/model_backwards_compatibility_check/model_backwards_compat_inference.py
piyushghai: Done.

tests/nightly/model_backwards_compatibility_check/model_backwards_compat_inference.py
piyushghai: This gets downloaded to the root folder of the mxnet-repo.

tests/nightly/model_backwards_compatibility_check/common.py
piyushghai: The bucket creation is a one-time job which I've already done so. Any other updates to the bucket will be done programatically only. 
I see your point though. I can however, create two folders under the same bucket. One folder for data and another for models. 
That should reduce the overhead.

ci/docker/runtime_functions.sh
piyushghai: Done.

tests/nightly/model_backwards_compatibility_check/common.py
piyushghai: Done.  c80ee31

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
piyushghai: Done. e764d5a2898d42f263688685bf575234a492a650

piyushghai:
@marcoabreu ping.

ci/docker/runtime_functions.sh
marcoabreu: It's still there, what did you change?

tests/nightly/model_backwards_compatibility_check/common.py
```diff
@@ -0,0 +1,210 @@
...
+        if os.path.isfile(file):
+            os.remove(file)
+
+def download_data_from_s3(model_name):
+    print ('Downloading data files for %s from bucket %s'%(model_name, model_bucket_name + backslash + data_folder))
+    bucket = s3.Bucket(model_bucket_name)
+    prefix = data_folder + backslash + model_name + '-data'
+    data_files_meta = list(bucket.objects.filter(Prefix = prefix))
+    if len(data_files_meta) == 0:
+        print ('No data files found for %s' %model_name)
```
marcoabreu: Please use logging instead of print

tests/nightly/model_backwards_compatibility_check/model_backward_compat_checker.sh
```diff
@@ -0,0 +1,28 @@
...
+
+#Author: Piyush Ghai
+
+echo "Invoking model_backwards_compat_checker.sh script"
+echo `pwd`
+cd tests/nightly/model_backwards_compatibility_check
+echo `pwd`
+
+echo '=========================='
+python model_backwards_compat_inference.py
```
marcoabreu: Wouldn't this rather be in runtime_functions.sh?

tests/nightly/model_backwards_compatibility_check/model_backwards_compat_inference.py
```diff
@@ -0,0 +1,149 @@
...
+	        loaded_model = Net()
+	        loaded_model.load_params(model_name+'-params')
+	        output = loaded_model(test_data)
+	        old_inference_results = mx.nd.load(model_name + '-inference')['inference']
+	        assert_almost_equal(old_inference_results.asnumpy(), output.asnumpy())
+	        clean_model_files(model_files, model_name)
+	        print ('=================================')
+	print ('Assertion passed for model : %s' %model_name)
+
+def test_lenet_gluon_hybrid_imports_api():
```
marcoabreu: Do we really need to make a test case for each individual model? Wouldn't it be possible to just loop through the S3 bucket and automatically determine all models that have to be run? That way, we can add things as required without copy&pasting this boilerplate.

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
```diff
@@ -0,0 +1,69 @@
...
+
+## Fetch the latest release tags, filtering out 'rcs' and filtering out some other irrelevant ones
+## This list is sorted in descending order chronologically. Keeping n = 5 for a precautionary check.
+## Sample output for the below git tag command is : 1.2.0 utils 1.1.0 1.0.0 0.12.1
+previous_versions=($(git tag --sort=-creatordate | grep --invert-match rc | head -n 5))
+count=0
+for version in ${previous_versions[*]}
+do
+	# We just need to train the previous two versions. This logic can be changed later on as welll.
+	if [[ "$count" -gt 1 ]]
```
marcoabreu: Could you elaborate? We have to train everything that's missing.

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
```diff
@@ -0,0 +1,69 @@
...
+	then
+		echo "Successfully trained files for the previous two MXNet release versions"
+		exit 1
+	fi
+
+	## If MXNet major version starts with a number >=1. with a wildcard match for the minor version numbers
```
marcoabreu: How does this handle different major versions?

tests/nightly/model_backwards_compatibility_check/model_backwards_compat_inference.py
piyushghai: We would still need to instantiate the models and load them using different loading APIs. Hence the boiler plate code cannot be done away with. I have tried to keep the boiler plate as small as possible to perform the inference.

tests/nightly/model_backwards_compatibility_check/model_backward_compat_checker.sh
piyushghai: runtime_functions calls the bash script -- 'model_backward_compat_checker.sh'

tests/nightly/model_backwards_compatibility_check/common.py
piyushghai: Done. 9d4d0993b56ec8b3148e1a0ee90c81424bb1f0cb

ci/docker/runtime_functions.sh
piyushghai: I added a comment indicating that it should not be called from the regular docker environment.

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
marcoabreu: I see, thank you. Wasn't it part of the initial design that we test everything under the same major version, independent of the number of releases?

ci/docker/runtime_functions.sh
marcoabreu: Sorry, retrospectively, my previous comment was not very clear. Please excuse that.

Everything inside runtime_functions.sh is intended to be run within docker **only**. Please have a look at https://github.com/apache/incubator-mxnet/blob/master/tests/nightly/broken_link_checker_test/JenkinsfileForBLC#L57 to see how to combine the usage of Docker together with the integration of boto to fulfil that constraint. It basically boils down to doing the core logic inside the Docker environment and then running a second script (in this case it's pure shell, but you can also create a second python script dedicated to the publishing process). Feel free to hit me up if you need further advice here.

ci/docker/runtime_functions.sh
```diff
@@ -854,6 +854,29 @@ nightly_test_javascript() {
     make -C /work/mxnet/amalgamation libmxnet_predict.js MIN=1 EMCC=/work/deps/emscripten/emcc
 }
 
+#Tests Model backwards compatibility on MXNet
+nightly_model_backwards_compat_test() {
+    set -ex
+    export PYTHONPATH=./python/
+    pip install boto3
```
marcoabreu: Please install boto here: https://github.com/apache/incubator-mxnet/blob/master/ci/docker/install/ubuntu_core.sh

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
```diff
@@ -0,0 +1,69 @@
...
+	echo '=========================='
+}
+
+install_mxnet() {
+	version=$1
+	echo "Installing MXNet "$version
+	pip install mxnet==$version
+}
+
+install_boto3(){
```
marcoabreu: As stated previously, please remove installing boto3 here

marcoabreu:
Good point @nswamy ! Shall we do this in a follow-up PR or do you want it here?

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
piyushghai: Good point and it's been taken care of in the inference scripts. They pull off all the model files (they don't filter out on which version files were trained on). So in a way, after a couple of releases the training script would have built a pool of all the previous releases for the inference files to work with.

piyushghai: This line -- "previous_versions=($(git tag --sort=-creatordate | grep --invert-match rc | head -n 5))" is version agnostic, which means even if you change the major version, the models will still be trained. 
With multiple minor versions, the training will happen on all of the tagged git versions. 

So essentially, whenever a new release (be it major version/minor version) comes out, it will be pulled in using the git-tag command and the training file will save a bunch of models for it in S3. 

Please let me know if there's still some confusion here.

ci/docker/runtime_functions.sh
piyushghai: I'm a little confused about it. So this is indeed a couple of lines of shell script which can be easily added in the docker file for MBCC, but how do I install boto on the Jenkins machines since the training script would not be running inside a docker container ?

piyushghai: Well, now that I have another look at https://github.com/apache/incubator-mxnet/blob/master/tests/nightly/broken_link_checker_test/JenkinsfileForBLC#L57 I get the interaction with S3.

piyushghai: Done in 1f63941e9ade0eaa2f4361ed5877bc97defcf6fb

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
piyushghai: I have updated the readme and added more comments to this file in : 1f63941e9ade0eaa2f4361ed5877bc97defcf6fb

ci/docker/runtime_functions.sh
piyushghai: Done.

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
piyushghai: Done.

marcoabreu:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/test-backwards-compatibility-checker/detail/test-backwards-compatibility-checker/6/pipeline

tests/nightly/model_backwards_compatibility_check/JenkinsfileForMBCC
```diff
@@ -0,0 +1,86 @@
...
+
+try {
+  stage('MBCC'){
+    parallel 'ModelBackwardsCompat: CPU': {
+      node('mxnetlinux-cpu') {
+        ws('workspace/modelBackwardsCompat') {
+
+                init_git()
+                // Train models on older versions
+                docker_run('ubuntu_nightly_cpu', 'nightly_model_backwards_compat_train', false)
```
marcoabreu: Could you put the training and the testing into two separate (non-parallel) stages to increase clarity?

tests/nightly/model_backwards_compatibility_check/JenkinsfileForMBCC
```diff
@@ -0,0 +1,86 @@
...
+                // upload files to S3 here outside of the docker environment
+                sh /tests/nightly/model_backwards_compatibility_check/upload_models_to_s3.sh
```
marcoabreu: I this this path is wrong because it is absolute. I'd double check here.

tests/nightly/model_backwards_compatibility_check/JenkinsfileForMBCC
```diff
@@ -0,0 +1,86 @@
...
+                // Perform inference on these models
+                docker_run('ubuntu_nightly_cpu', 'nightly_model_backwards_compat_test', false)
+
+		    }
```
marcoabreu: Please check the indentation

ci/docker/runtime_functions.sh
```diff
@@ -895,6 +895,25 @@ nightly_test_javascript() {
...
+    export PYTHONPATH=./python/
+    ./tests/nightly/model_backwards_compatibility_check/model_backward_compat_checker.sh
+}
+
+#Backfills S3 bucket with models trained on earlier versions of mxnet
+nightly_model_backwards_compat_train() {
+    set -ex
+    export PYTHONPATH=./python/
+    VENV=mbcc_py2_venv
+    virtualenv -p `which python2` $VENV
```
marcoabreu: If you run inside a virtualenv, you won't have access to any installed packages. I'd recommend not to use virtualenv here because you're running inside docker anyways. This is also the reason for the CI error that states it could not find boto3.

tests/nightly/model_backwards_compatibility_check/JenkinsfileForMBCC
piyushghai: Done. 87103d43dee19bb42ac4a7296a9be940ca44dffc

tests/nightly/model_backwards_compatibility_check/JenkinsfileForMBCC
piyushghai: Good catch. Added a '.' before /tests. 
Done in 87103d43dee19bb42ac4a7296a9be940ca44dffc.

tests/nightly/model_backwards_compatibility_check/JenkinsfileForMBCC
piyushghai: Done. 87103d43dee19bb42ac4a7296a9be940ca44dffc

ci/docker/runtime_functions.sh
piyushghai: Ah I see. Fixed in edd68164526cf591b090322f8f6e0805e17c17ee

piyushghai:
@marcoabreu The Jenkins CI [1] build is giving an error on : the import statement for MXNet under the Inference Stage of the JenkinsFileForMBCC. I'm not able to figure out why that's happening. 
Can you have a look at this 69843fbe4d6669c135d3ae85aa56df144bc6c076 and give a second eye opinion ? 

[1] : http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/test-backwards-compatibility-checker/detail/test-backwards-compatibility-checker/12/pipeline/4/

marcoabreu:
Hello,
the following error is shown at http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/restricted-test-backwards-compatibility-checker/detail/restricted-test-backwards-compatibility-checker/4/pipeline/4:
```
Installing collected packages: graphviz, numpy, mxnet

  Found existing installation: graphviz 0.8.4

    Uninstalling graphviz-0.8.4:

      Successfully uninstalled graphviz-0.8.4

  Found existing installation: mxnet 1.2.0

    Uninstalling mxnet-1.2.0:

      Successfully uninstalled mxnet-1.2.0

Successfully installed graphviz-0.8.1 mxnet-1.1.0 numpy-1.13.3

==========================

Running training files and preparing models

==========================

  File "model_backwards_compat_train.py", line 20, in <module>

    from common import *

  File "/work/mxnet/tests/nightly/model_backwards_compatibility_check/common.py", line 22, in <module>

    import mxnet as mx

  File "/work/mxnet/python/mxnet/__init__.py", line 24, in <module>

    from .context import Context, current_context, cpu, gpu, cpu_pinned

  File "/work/mxnet/python/mxnet/context.py", line 24, in <module>

    from .base import classproperty, with_metaclass, _MXClassPropertyMetaClass

  File "/work/mxnet/python/mxnet/base.py", line 197, in <module>

    _LIB = _load_lib()

  File "/work/mxnet/python/mxnet/base.py", line 187, in _load_lib

    lib_path = libinfo.find_lib_path()

  File "/work/mxnet/python/mxnet/libinfo.py", line 74, in find_lib_path

    'List of candidates:\n' + str('\n'.join(dll_path)))

RuntimeError: Cannot find the MXNet library.

List of candidates:

libmxnet.so

/usr/local/lib/libmxnet.so

/work/mxnet/python/mxnet/libmxnet.so

/work/mxnet/python/mxnet/../../lib/libmxnet.so

/work/mxnet/python/mxnet/../../build/libmxnet.so

../../../libmxnet.so

==========================

Successfully trained files for the previous two MXNet release versions
```

It's strange that an error occurs but the run sitll continues. Could you please have a look?

tests/nightly/model_backwards_compatibility_check/common.py
```diff
@@ -42,7 +42,7 @@
 
 # get the current mxnet version we are running on
 mxnet_version = mx.__version__
-model_bucket_name = 'mxnet-model-backwards-compatibility-models'
+model_bucket_name = 'mxnet-ci-prod-backwards-compatibility-models'
```
piyushghai: This bucket name change also has to be made in upload_models_to_s3.sh.

Done that in 33096c02f56186464b34ba6b88480290eb89263f

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
```diff
@@ -57,10 +57,10 @@ do
 	fi
 
 	## If MXNet major version starts with a number >=1. with a wildcard match for the minor version numbers
-	if [[ $version = [1-9]* ]]
+	if [[ $version = [1-9].[0-9].[0-9] ]]
```
marcoabreu: I don't think this will handle multi digit versions.

tests/nightly/model_backwards_compatibility_check/model_backward_compat_checker.sh
```diff
@@ -0,0 +1,28 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+#Author: Piyush Ghai
+
```
marcoabreu: Missing set -ex

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
```diff
@@ -0,0 +1,67 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+#Author: Piyush Ghai
+
```
marcoabreu: Missing set -ex

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
piyushghai: Good point. Done 16d320a926d4f23265f730d2fc197ecec5bc4a2e

tests/nightly/model_backwards_compatibility_check/model_backward_compat_checker.sh
piyushghai: Done ceac705fe133cc7100f8c0b5076f943c4315462d

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
piyushghai: Done ceac705fe133cc7100f8c0b5076f943c4315462d

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
```diff
@@ -0,0 +1,71 @@
...
+	if [[ "$count" -gt 1 ]]
+	then
+		echo "Successfully trained files for the previous two MXNet release versions"
+		exit 0
+	fi
+
+	## If MXNet major version starts with a number >=1. with a wildcard match for the minor version numbers
+	## Could have used a [[:digit:]]+. as well but it was not working as a traditional regex in bash.
+	## so had to resort to using [[:digit:]] [[:digit:]]* to indicate multi-digit version regex match
+	if [[ $version = [[:digit:][[:digit:]]*.[[:digit:]].[[:digit:]] ]]
```
marcoabreu: Would  this also cover 1.12.0? It seems like you only added multi digit for the major version.

Also, could you explain how this exactly determines whether it is a major version match?

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
marcoabreu: As stated in-person, we have to train everything under the same major version.

piyushghai: Done. d8fa75d72bd87466112f58865cebb38883b63171.

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
piyushghai: Done. d8fa75d72bd87466112f58865cebb38883b63171.

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
piyushghai: It handles here : d8fa75d72bd87466112f58865cebb38883b63171.

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
```diff
@@ -0,0 +1,82 @@
...
+## Cuts the string and gives only the major version part.
+## eg : 12.3.0 ---> 12
+get_major_version() {
+    major=$(echo $1 | cut -d. -f1)
+    echo $major
+}
+
+## We read the current major version from libinfo.py file. And we extract the major version from it.
+curr_mxnet_version=$(grep -w "__version__" python/mxnet/libinfo.py | grep -o '".*"' | sed 's/"//g')
+curr_major_version=$(get_major_version $curr_mxnet_version)
```
marcoabreu: Please validate the output to ensure we get a notification/error if somebody breaks this by accident.

piyushghai: Done. ca01aa2bd9a183af10a4c3ae7ca5194c395b4ffb.

marcoabreu:
It seems like that this pipeline has revealed some slight changes between 1.0.0 and our master:

```
INFO:root:Performing inference for model/API lenet_gluon_save_params_api
INFO:root:Fetching files for MXNet version : 1.0.0 and model lenet_gluon_save_params_api
/work/mxnet/python/mxnet/gluon/block.py:420: UserWarning: load_params is deprecated. Please use load_parameters.
  warnings.warn("load_params is deprecated. Please use load_parameters.")
  File "model_backwards_compat_inference.py", line 135, in <module>
    test_lenet_gluon_load_params_api()
  File "model_backwards_compat_inference.py", line 72, in test_lenet_gluon_load_params_api
    assert_almost_equal(old_inference_results.asnumpy(), output.asnumpy())
  File "/work/mxnet/python/mxnet/test_utils.py", line 493, in assert_almost_equal
    raise AssertionError(msg)
AssertionError: 
Items are not equal:
Error 1.969787 exceeds tolerance rtol=0.000010, atol=0.000000.  Location of maximum error:(5, 0), a=0.003014, b=0.003014
 a: array([[ 0.01743407, -0.30903903],
       [ 0.08352755, -0.365019  ],
       [ 0.06324662, -0.4323489 ],...
 b: array([[ 0.01743408, -0.3090391 ],
       [ 0.08352771, -0.365019  ],
       [ 0.06324662, -0.4323488 ],...
```

I will still merge this PR because fixing regressions is outside the scope of this PR.

ChaiBapchya:
Q1 I can see this added as part of nightly test. However, the bucket contains models from 1.0 to till 1.6 Is it discontinued for subsequent minor versions of mxnet? 1.7 & master[2.0]?

Q2. The bucket for this test "mxnet-ci-prod-backwards-compatibility-models" was public. Our AppSec team reviewed it and asked to block public access. I've made that change. Why was this bucket made publicly accessible if its run on nightly within CI account? Was it required somewhere else as well? or was it kept public so that non-AWS folks can test it out as well?

@piyushghai @marcoabreu @szha @josephevans @sandeep-krishnamurthy

ChaiBapchya:
Checked the pipeline: https://jenkins.mxnet-ci.amazon-ml.com/job/restricted-backwards-compatibility-checker/
It's been failing since months. We don't have a ticketing mechanism for this so its a feature that's untracked. Pl advise.

marcoabreu:
Q1: No, it was not discontinued. It seems like it was just not caught. I remember that I created a sevice which sends out a daily email to an email list at Amazon with the results. Maybe somebody turned it off or everybody ignored the email..

Q2: It has to be public as otherwise the results are not reproducible for others. Hence it has to be on public read. Don't forget that this is an Apache and not an Amazon project. So yes, your question at the end is the answer :)

marcoabreu:
I think we need to fix it. That'd be my advice.

marcoabreu:
Btw, thanks for catching it!

PR #11470, opened 2018-06-28
Fix build issue with USE_CUDNN=0

anirudh2290:
## Description ##

This was found during the flaky test investigation https://github.com/apache/incubator-mxnet/issues/10973 . I found that the windows code path is different from linux, so it was testing the MXNet conv and deconv op which was not being tested for linux. Building with USE_CUDNN=0 and USE_CUDA=1 was failing and I have added the fix here. I have also added the build and test as a stage to jenkins. Not sure if there was a reason to not check this in CI (configuration not supported by some ops etc.) 

@marcoabreu 

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

Issue #10973, opened 2018-05-16
Flaky test_deconvolution


asitstands:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/PR-10970/1/pipeline/724

```
======================================================================
FAIL: test_operator_gpu.test_deconvolution
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Anaconda3\envs\py3\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\Anaconda3\envs\py3\lib\site-packages\nose\util.py", line 620, in newfunc
    return func(*arg, **kw)
  File "C:\jenkins_slave\workspace\ut-python-gpu\tests\python\gpu\../unittest\common.py", line 157, in test_new
    orig_test(*args, **kwargs)
  File "C:\jenkins_slave\workspace\ut-python-gpu\tests\python\gpu\../unittest\test_operator.py", line 1169, in test_deconvolution
    pad                 = (1,1)
  File "C:\jenkins_slave\workspace\ut-python-gpu\tests\python\gpu\../unittest\test_operator.py", line 1061, in check_deconvolution_forward_backward
    assert_almost_equal(out + args_grad_addto_npy[0], args_grad_addto[0].asnumpy(), rtol=1e-4, atol=1e-3)
  File "C:\jenkins_slave\workspace\ut-python-gpu\pkg_vc14_gpu\python\mxnet\test_utils.py", line 493, in assert_almost_equal
    raise AssertionError(msg)
AssertionError: 
Items are not equal:
Error 1.029078 exceeds tolerance rtol=0.000100, atol=0.001000.  Location of maximum error:(1, 2, 19, 15), a=0.083483, b=0.082445
 a: array([[[[  -1.03032662,   84.66370263,  -15.19734333, ...,   75.15862218,
           -93.83804658,  -63.14129641],
         [  54.54269176,    7.71739136,   23.10548904, ...,   81.15015121,...
 b: array([[[[  -1.03033042,   84.6636734 ,  -15.19732857, ...,   75.15862274,
           -93.83802795,  -63.14128113],
         [  54.54269409,    7.71736002,   23.10546684, ...,   81.15015411,...
-------------------- >> begin captured logging << --------------------
common: INFO: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=1270064081 to reproduce.
--------------------- >> end captured logging << ---------------------
[...]

Issue #11568, opened 2018-07-05
Issues with spatial transformer op when cudnn disabled


anirudh2290:
## Description
as part of PR: #11470, it was found that spatial transformer op without cudnn enabled doesn't pass tests. 
To reproduce try one of the two scripts below:

Script 1:
```
import numpy as np
import mxnet as mx
from mxnet.test_utils import assert_almost_equal, default_context

np.set_printoptions(threshold=np.nan)
num_filter = 2  # conv of loc net
kernel = (3, 3)  # conv of loc net
num_hidden = 6  # fc of loc net
for n in [1, 2, 3, 4]:
    for c in [1, 2, 3, 4]:
        for h in [5, 9, 13, 17]:  # for convenience test, this third and forth input dim should be 4x + 1
            for w in [5, 9, 13, 17]:
                data_shape = (n, c, h, w)
                target_shape = (int((data_shape[2]+1)/2), int((data_shape[3]+1)/2))
                data = mx.sym.Variable(name="data")
                loc = mx.sym.Convolution(data=data, kernel=kernel, pad=(1, 1), num_filter=num_filter, name="loc_conv")
                loc = mx.sym.Flatten(data=loc)
                loc = mx.sym.FullyConnected(data=loc, num_hidden=num_hidden, name="loc_fc")
                stn = mx.sym.SpatialTransformer(data=data, loc=loc, target_shape=target_shape,
                                                transform_type="affine", sampler_type="bilinear")
                arg_names = stn.list_arguments()
                arg_shapes, out_shapes, _ = stn.infer_shape(data=data_shape)
                # check shape
                assert out_shapes[0] == (data_shape[0], data_shape[1], target_shape[0], target_shape[1])
                #dev = default_context()
                dev = mx.gpu(0)
                args = {}
                args['data'] = mx.random.normal(0, 1, data_shape, ctx=mx.cpu()).copyto(dev)
                args['loc_conv_weight'] = mx.nd.zeros((num_filter, data_shape[1], kernel[0], kernel[1]), ctx=dev)
                args['loc_conv_bias'] = mx.nd.zeros((num_filter,), ctx=dev)
                args['loc_fc_weight'] = mx.nd.zeros((6, num_filter*data_shape[2]*data_shape[3]), ctx=dev)
                args['loc_fc_bias'] = mx.nd.array([0.5, 0, 0, 0, 0.5, 0], ctx=dev)
[...]

PR #11630, opened 2018-07-10
Fix flaky test test_deconvolution

anirudh2290:
## Description ##
This is to fix the flaky test: https://github.com/apache/incubator-mxnet/issues/10973. Prereq PR: https://github.com/apache/incubator-mxnet/pull/11470.
`cublasSgemm` and `cublasDgemm` are currently used for linear algebra operations when cudnn is disabled for the Convolution and Deconvolution operators. The test_deconvolution test on gpu fails on the input gradients comparison for conv and deconv operators. When looking at the failures, the difference in the gradients are as big as 0.1. Replacing `cublasSgemm` with `cublasSgemmex` while keeping all other code the same, causes the failure to go away. Also, the failure doesnt happen on CPU with openblas or MKLDNN or on gpu with cudnn enabled. If both APIs work as documented, both cases should have seen the failure. My guess is that `cublasSgemm` is doing the conversion of float32 to float16 before doing the conversion, causing the precision issues. Digging through the cublas documentation, I couldn't find enough information on the computation for `cublasSgemm`.
Please see: http://lutgw1.lunet.edu/cuda/pdf/CUBLAS_Library.pdf  2.8.11 (cublassgemmex) and 2.7.1 (cublassgemm).
 
@DickJC123 @ptrendx any inputs on if there should be any difference in behavior of two apis when all tensors have float32 dtype.

The important changes are in the file: `src/operator/linalg_impl.h`

### DIFF
--- a/src/operator/linalg_impl.h
+++ b/src/operator/linalg_impl.h
@@ -169,23 +169,50 @@ void linalg_gemm<cpu, mshadow::half::half_t>(const Tensor<cpu, 2, mshadow::half:
 
 // cublas col-major processing accounted for by switching first two operands
 
-#define LINALG_GPU_GEMM(fname, DType) \
-template<> inline \
-void linalg_gemm<gpu, DType>(const Tensor<gpu, 2, DType>& A, const Tensor<gpu, 2, DType>& B, \
-                             const Tensor<gpu, 2, DType>& C, DType alpha, DType beta, \
-                             bool tA, bool tB, Stream<gpu> *s) { \
-  using namespace mxnet; \
-  using mshadow::gpu; \
-  CHECK_NOTNULL(s); \
-  check_gemm(A, B, C, alpha, beta, tA, tB); \
-  CUBLAS_CALL(cublas##fname(Stream<gpu>::GetBlasHandle(s), \
-                            (tB ? CUBLAS_OP_T : CUBLAS_OP_N), \
-                            (tA ? CUBLAS_OP_T : CUBLAS_OP_N), \
-                            C.size(1), C.size(0), (tB ? B.size(1) : B.size(0)), \
-                            &alpha, B.dptr_, B.stride_, A.dptr_, A.stride_, \
-                            &beta, C.dptr_, C.stride_)) \
+#define LINALG_GPU_GEMM(fname, DType)                                      \
+  template <>                                                              \
+  inline void linalg_gemm<gpu, DType>(                                     \
+      const Tensor<gpu, 2, DType>& A, const Tensor<gpu, 2, DType>& B,      \
+      const Tensor<gpu, 2, DType>& C, DType alpha, DType beta, bool tA,    \
+      bool tB, Stream<gpu>* s) {                                           \
+    using namespace mxnet;                                                 \
+    using mshadow::gpu;                                                    \
+    CHECK_NOTNULL(s);                                                      \
+    check_gemm(A, B, C, alpha, beta, tA, tB);                              \
+    CUBLAS_CALL(cublas##fname(                                             \
+        Stream<gpu>::GetBlasHandle(s), (tB ? CUBLAS_OP_T : CUBLAS_OP_N),   \
+        (tA ? CUBLAS_OP_T : CUBLAS_OP_N), C.size(1), C.size(0),            \
+        (tB ? B.size(1) : B.size(0)), &alpha, B.dptr_, B.stride_, A.dptr_, \
+        A.stride_, &beta, C.dptr_, C.stride_))                             \
+  }
+
+#if CUDA_VERSION >= 7050
+template <>
+inline void linalg_gemm<gpu, float>(const Tensor<gpu, 2, float>& A,
+                                    const Tensor<gpu, 2, float>& B,
+                                    const Tensor<gpu, 2, float>& C, float alpha,
+                                    float beta, bool tA, bool tB,
+                                    Stream<gpu>* s) {
+  using namespace mxnet;
+  using mshadow::gpu;
+  CHECK_NOTNULL(s);
+  check_gemm(A, B, C, alpha, beta, tA, tB);
+#if CUDA_VERSION >= 8000
+  cudaDataType_t full_datatype = CUDA_R_32F;
+#else
+  cublasDataType_t full_datatype = CUBLAS_DATA_FULL;
+#endif
+  CUBLAS_CALL(cublasSgemmEx(
+      Stream<gpu>::GetBlasHandle(s), (tB ? CUBLAS_OP_T : CUBLAS_OP_N),
+      (tA ? CUBLAS_OP_T : CUBLAS_OP_N), C.size(1), C.size(0),
+      (tB ? B.size(1) : B.size(0)), &alpha, B.dptr_, full_datatype, B.stride_,
+      A.dptr_, full_datatype, A.stride_, &beta, C.dptr_, full_datatype,
+      C.stride_))
 }
+
+#else
 LINALG_GPU_GEMM(Sgemm, float)
+#endif
 LINALG_GPU_GEMM(Dgemm, double)
 
 // Version where matrix rows are given by first axis.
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -667,6 +680,21 @@ try {
         }
       }
     },
+    'Python3: MKLDNN-GPU-NOCUDNN': {
+      node('mxnetlinux-gpu') {
+        ws('workspace/ut-python3-mkldnn-gpu-nocudnn') {
+          withEnv(['CUDNN_DISABLED=ON']) {
```
marcoabreu: Please make a new script to call the unit tests in the runtime functions (you might just nest the calls to prevent copy and paste). Otherwise, the calls will not be reproducible- also, the environment variables will not be available in the docker container.
I'm on my phone, so please excusme if the question is answered below: why do we have to set that variable if we already compile without cudnn?

tests/python/unittest/common.py
```diff
@@ -94,6 +95,23 @@ def random_seed(seed=None):
         random.seed(next_seed)
 
 
+def assert_raises_cudnn_disabled(assertion_error=False):
+    def test_helper(orig_test):
+        @make_decorator(orig_test)
+        def test_new(*args, **kwargs):
+            cudnn_disabled = (os.getenv('CUDNN_DISABLED') == "ON")
+            if not cudnn_disabled or mx.context.current_context().device_type == 'cpu':
```
marcoabreu: Ah that's the environment variable. Could we give it a name that clearly shows that this environment variable is only for test purposes?

tests/python/unittest/common.py
```diff
@@ -94,6 +95,23 @@ def random_seed(seed=None):
...
+                orig_test(*args, **kwargs)
+            else:
+                if assertion_error:
```
marcoabreu: When do we want to differentiate between assertion errors? I think we should always handle them, no matter what.

anirudh2290: this is to differentiate between failures and the ops that dont support cudnn disabled. for example some ops when used with cudnn disabled would just throw an exception saying not supported. others would proceed but would compute wrongly and throw an assertion error.

tests/python/unittest/common.py
anirudh2290: I have renamed it to `CUDNN_OFF_TEST_ONLY` in the PR: #11470. Let us move the discussion for cudnn off tests to that PR.

Jenkinsfile
anirudh2290: Thanks for the info. I have fixed it in PR: #11470. We are removing the C API in that PR so there is no way of finding out whether cudnn is enabled.

tests/python/unittest/common.py
marcoabreu: I see, thanks for elaborating. I'm afraid that this could be a bit error prone in it's usage.

 Also, I don't think this is what should happen. If a user uses an operator although it's not supported, it should not make a random false calculation but rather throw the error properly. This method would mask a problem that a user would silently encounter. We should rather investigate why the error is not being thrown and fix that.

marcoabreu: Also, is there no CPU implementation of these operators we could fall back to instead?

Jenkinsfile
marcoabreu: Couldn't we just always catch the error and skip it if it contains the message about cudnn not being available?

tests/python/unittest/common.py
anirudh2290: yep the issue was uncovered as part of the PR and is documented here: https://github.com/apache/incubator-mxnet/issues/11568 . As I have mentioned in #11470, this is not something we should be blocking the PR for. This issue seems could have been there for a long time. Its possible that there are user codes depending on the buggy behavior and raising an exception would be a breaking change for these users. Its better to fix the issue with cudnn enabled rather than throwing an exception saying cudnn not supported atleast for 1.3.

marcoabreu: Agree, let's not block that PR. But please add a comment here that this is a temporary solution and link it with the issue. 

Would you mind if we start a thread on dev@ to discuss this topic? I think we should have a thorough analysis on this topic. I agree that this might cause a breaking change, but on the other hand, our users might have worked with wrong operators the entire time which might have caused unknown side effects - something that's probably even worse than a detailed error message.

Jenkinsfile
anirudh2290: well that approach would also work. but this is cleaner. for example, someone adding a new op which doesnt support cudnn off may have a different wording for example.

marcoabreu: Well then the error would be made visible in CI and the person would have to adjust the error message, right? Ideally, we'd have standardized error messages (e.g. error codes).

marcoabreu: I don't know if this is cleaner because that way you are catching all errors independently of whether they are actually related to cudnn. Imagine somebody makes a change to the cudnn behaviour but there is is still a bug - the error gets then omitted and we don't notice it. I'd rather have detailed and specific error handling rather than a global try-catch. What do you think?

tests/python/unittest/common.py
anirudh2290: right now the only op that is affected is `mx.sym.SpatialTransformer`. If cudnn is enabled `CuDNNSpatialTransformerOp` is used. If it is not enabled then the mxnet implementation for the op is used. Why this implementation is incorrect needs to be investigated. Do you think we need to start a discussion for only this op ?

marcoabreu: If it is only that operator, I'd rather say that we should deactivate that test and Mark it as flaky. We can then remove this special catch for assertion errors.

In that case we don't need a discussion.

anirudh2290: we need to still test it when cudnn is enabled.

marcoabreu: We currently have a 0 tolerance policy for faulty tests and deactivate them temporarily while taking reduced test coverage as a pill we have to swallow. It is a release requirement that all tests have to be enabled, so don't worry: it won't stay disables for long.

Jenkinsfile
anirudh2290: well this test will only be ignored if the decorator is added, so only when cudnn off and for operators that dont support the configuration. People making changes to cudnn behavior should not be affected by this.  Its a global try catch when cudnn is disabled for ops that dont support the configuration.

tests/python/unittest/common.py
anirudh2290: why disable the test when it can be avoided. If it is for visibility then we already have an open issue. we can also call it a disabled test since it is disabled for cudnn off.

marcoabreu: Sure, as you prefer. Just wanted to avoid having this hack in there, but I don't mind. Could you please add a comment in that test, linking to the issue and stating that the test is disabled when running without cudnn?

anirudh2290: will do

src/operator/linalg_impl.h
```diff
@@ -169,23 +169,50 @@ void linalg_gemm<cpu, mshadow::half::half_t>(const Tensor<cpu, 2, mshadow::half:
...
+    CHECK_NOTNULL(s);                                                      \
+    check_gemm(A, B, C, alpha, beta, tA, tB);                              \
+    CUBLAS_CALL(cublas##fname(                                             \
+        Stream<gpu>::GetBlasHandle(s), (tB ? CUBLAS_OP_T : CUBLAS_OP_N),   \
+        (tA ? CUBLAS_OP_T : CUBLAS_OP_N), C.size(1), C.size(0),            \
+        (tB ? B.size(1) : B.size(0)), &alpha, B.dptr_, B.stride_, A.dptr_, \
+        A.stride_, &beta, C.dptr_, C.stride_))                             \
+  }
+
+#if CUDA_VERSION >= 7050
```
lupesko: A comment here explaining the #if directive would be helpful for other (and yourself) in the future. Can we add it please?

anirudh2290: i have added the comment. also removed the checklist for the current PR.

anirudh2290:
@asmushetzel sorry about the confusion. To clarify, the documentation states that both APIs should behave exactly the same when all tensors have float32 dtype. I am guessing that the difference in behavior could then be because cublassgemm is converting tensors to float16 dtype (a bug?) for computation and could be the reason for the precision issues. I have asked Nvidia folks for inputs but haven't heard back from them yet.

anirudh2290:
I will wait till end of this week to hear back from Nvidia folks before merging this PR.

PR #11637, opened 2018-07-10
[MXNET-652] Allow for multi-context input to load gluon models from onnx

OneRaynyDay:
## Description ##
Currently, Airbnb is developing applications that use mxnet(particularly `gluon`). We use the `mxnet.contrib.onnx` package, which supports importing to gluon from ONNX format. As this is an extremely important part of our pipeline, we want to add support to multi-gpu imports, which is currently not supported.

This is actually a very simple change, because `_load_init()` already supports, directly, a `Context` object or an iterable of them. However, the current implementation for `import_to_gluon` suggests a `str` of `'CPU'` or `'GPU'`. This is not only different from other parts of the mxnet codebase, but also does not allow us to specify which gpu's we want to load the model into. I am aware that this is to comply with ONNX, but the abstraction should be handled at the abstraction layer, not the core utility for importing ONNX to gluon.

### Essentials ###
- [x] The PR title starts with [MXNET-652], where 652 refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)

### Changes ###
- [x] Modified gluon model loading from ONNX to allow multi-context argument.

## Comments ##
- Currently, mxnet uses `ctx` as an arg for a large portion of the core library. There doesn't seem to be any significant reason for why the `contrib.onnx` code uses `str` instead, rather than it being a minor oversight that passed PR's.

### DIFF
--- a/python/mxnet/contrib/onnx/onnx2mx/import_onnx.py
+++ b/python/mxnet/contrib/onnx/onnx2mx/import_onnx.py
@@ -157,15 +157,15 @@ class GraphProto(object): # pylint: disable=too-few-public-methods
                    }
         return metadata
 
-    def graph_to_gluon(self, graph, context):
+    def graph_to_gluon(self, graph, ctx):
         """Construct SymbolBlock from onnx graph.
 
         Parameters
         ----------
         graph : onnx protobuf object
             The loaded onnx graph
-        context : str
-            context for mxnet module object. Should be 'CPU' or 'GPU'
+        ctx : Context or list of Context
+            Loads the model into one or many context(s).
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


python/mxnet/contrib/onnx/onnx2mx/import_onnx.py
```diff
@@ -157,15 +156,15 @@ def get_graph_metadata(self, graph):
                    }
         return metadata
 
-    def graph_to_gluon(self, graph, context):
+    def graph_to_gluon(self, graph, ctx):
```
szha: Thanks for this fix as we're already using ctx everywhere in gluon.

OneRaynyDay: No problem. Thank you for reviewing! I appreciate the merge.

PR #11650, opened 2018-07-11
remove windows note for dataloader, fix deprecated warning in python3.7

zhreshold:
## Description ##
- multi worker data loader is supported in windows, update the doc correspondingly.
- fix a deprecated feature/bug in python3.7

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/tests/python/unittest/test_gluon_data.py
+++ b/tests/python/unittest/test_gluon_data.py
@@ -123,106 +123,99 @@ class Dataset(gluon.data.Dataset):
     def __getitem__(self, key):
         return mx.nd.full((10,), key)
 
-@unittest.skip("Somehow fails with MKL. Cannot reproduce locally")
 def test_multi_worker():
     data = Dataset()
     loader = gluon.data.DataLoader(data, batch_size=1, num_workers=5)
     for i, batch in enumerate(loader):
-        assert (batch.asnumpy() == i).all()
+        assert (batch.asnumpy().astype('int32') == i).all()
+
+class _Dummy(Dataset):
+    """Dummpy dataset for randomized shape arrays."""
+    def __init__(self, random_shape):
+        self.random_shape = random_shape
+
+    def __getitem__(self, idx):
+        key = idx
+        if self.random_shape:
+            out = np.random.uniform(size=(random.randint(1000, 1100), 40))
+            labels = np.random.uniform(size=(random.randint(10, 15)))
+        else:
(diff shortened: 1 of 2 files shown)


marcoabreu:
Great! Is this covered by tests or do we maybe have a skip defined somewhere?

szha:
the fixed deprecation is on python 3.7 which is not supported in CI. I think we should start supporting py37 soon.

marcoabreu:
Definitely! My question is whether we test the data loader on windows. Sometimes, we have some conditional skips if we run on windows. It would be great if we could verify that this test is actually run on windows.

zhreshold:
Good point, let me go through the old tests and figure it out

zhreshold:
@marcoabreu This is behavioral test, the test will crash if not supported.

marcoabreu:
Okay, can we then remove the print to reduce the log output and replace it with some no-op? We're already having a lot of useless log (stuff like a function printing the numbers from 1 to 10000) and I'd like to keep that low in future if the log messages don't provide any information about the test itself.

tests/python/unittest/test_gluon_data.py
```diff
@@ -205,24 +206,20 @@ def batchify(self, data):
...
-        data = Dummy(True)
-        loader = DataLoader(data, batch_size=40, batchify_fn=data.batchify_list, num_workers=2)
-        for epoch in range(1):
-            for i, data in enumerate(loader):
-                if i % 100 == 0:
-                    print(data)
-                    print('{}:{}'.format(epoch, i))
+    # sanity tests
+    # enable test on windows
+    data = Dummy(False)
```
marcoabreu: Did you intentionally set the randomized shape to false? It used to be true before.

Also, could elaborate on the comment in line 210? Is there anything specific we have to do on Windows or is it working out of the box? If it doesn't need any special configuration, we can safely remove the comment as it is otherwise confusing to people who don't know the history

zhreshold: - I am deliberately switch one test case to fixed shape, and the rest one is responsible for the random test.
- Will remove the unnecessary comments

BTW, my PR disappeared in CI yesterday but show up today, do you have a clue?

marcoabreu: I see, thanks for elaborating.

Yeah, we had some disk space problems on our server. Everything should be fine now :)

zhreshold:
close due to duplicate efforts in #11908

PR #11651, opened 2018-07-11
Add logistic regression tutorial

Ishitori:
## Description ##
Add new tutorial on how to do logistic regression with Gluon API (writing minimal amount of custom code) + provided some explanation why it should be done in that way

### Essentials ###

### Changes ###
- [ ] New tutorial on logistic regression

### DIFF
--- a/docs/tutorials/gluon/logistic_regression_explained.md
+++ b/docs/tutorials/gluon/logistic_regression_explained.md
@@ -192,16 +192,25 @@ for e in range(epochs):
     accuracy.reset()
 ```
 
-    Epoch: 0, Training loss: 0.43, Validation loss: 0.36, Validation accuracy: 0.85, F1 score: 0.00
-    Epoch: 1, Training loss: 0.22, Validation loss: 0.14, Validation accuracy: 0.96, F1 score: 0.35
-    Epoch: 2, Training loss: 0.09, Validation loss: 0.11, Validation accuracy: 0.97, F1 score: 0.48
-    Epoch: 3, Training loss: 0.07, Validation loss: 0.09, Validation accuracy: 0.96, F1 score: 0.53
-    Epoch: 4, Training loss: 0.06, Validation loss: 0.09, Validation accuracy: 0.97, F1 score: 0.58
-    Epoch: 5, Training loss: 0.04, Validation loss: 0.12, Validation accuracy: 0.97, F1 score: 0.59
-    Epoch: 6, Training loss: 0.05, Validation loss: 0.09, Validation accuracy: 0.99, F1 score: 0.62
-    Epoch: 7, Training loss: 0.05, Validation loss: 0.10, Validation accuracy: 0.97, F1 score: 0.62
-    Epoch: 8, Training loss: 0.05, Validation loss: 0.12, Validation accuracy: 0.95, F1 score: 0.63
-    Epoch: 9, Training loss: 0.04, Validation loss: 0.09, Validation accuracy: 0.98, F1 score: 0.65
+    Epoch: 0, Training loss: 0.43, Validation loss: 0.36, Validation accuracy: 0.85, F1 score: 0.00 <!--notebook-skip-line-->
+
+    Epoch: 1, Training loss: 0.22, Validation loss: 0.14, Validation accuracy: 0.96, F1 score: 0.35 <!--notebook-skip-line-->
+
+    Epoch: 2, Training loss: 0.09, Validation loss: 0.11, Validation accuracy: 0.97, F1 score: 0.48 <!--notebook-skip-line-->
+
+    Epoch: 3, Training loss: 0.07, Validation loss: 0.09, Validation accuracy: 0.96, F1 score: 0.53 <!--notebook-skip-line-->
+
+    Epoch: 4, Training loss: 0.06, Validation loss: 0.09, Validation accuracy: 0.97, F1 score: 0.58 <!--notebook-skip-line-->
+
+    Epoch: 5, Training loss: 0.04, Validation loss: 0.12, Validation accuracy: 0.97, F1 score: 0.59 <!--notebook-skip-line-->
+    
+    Epoch: 6, Training loss: 0.05, Validation loss: 0.09, Validation accuracy: 0.99, F1 score: 0.62 <!--notebook-skip-line-->
+    
+    Epoch: 7, Training loss: 0.05, Validation loss: 0.10, Validation accuracy: 0.97, F1 score: 0.62 <!--notebook-skip-line-->
+    
+    Epoch: 8, Training loss: 0.05, Validation loss: 0.12, Validation accuracy: 0.95, F1 score: 0.63 <!--notebook-skip-line-->
+    
+    Epoch: 9, Training loss: 0.04, Validation loss: 0.09, Validation accuracy: 0.98, F1 score: 0.65 <!--notebook-skip-line-->
 
 
 In our case we hit the accuracy of 0.98 and F1 score of 0.65.
@@ -225,3 +234,5 @@ In this tutorial I explained some potential pitfalls to be aware of. When doing 
 1. Encode class labels as 0 or 1
 1. Use `SigmoidBinaryCrossEntropyLoss`
 1. Convert probabilities to classes before calculating Accuracy
+
+<!-- INSERT SOURCE DOWNLOAD BUTTONS -->
--- a/docs/tutorials/gluon/logistic_regression_explained.md
+++ b/docs/tutorials/gluon/logistic_regression_explained.md
@@ -55,7 +55,7 @@ val_dataloader = DataLoader(val_dataset, batch_size=batch_size, shuffle=True)
 
 ## Defining and training the model
 
-The only requirement for the logistic regression is that the last layer of the network must be a single neuron. Apache MXNet allows us to do so by using [Dense](https://mxnet.incubator.apache.org/api/python/gluon/nn.html#mxnet.gluon.nn.Dense) layer and specifying the number of units to 1. The rest of the network can be arbitrary complex. 
+The only requirement for the logistic regression is that the last layer of the network must be a single neuron. Apache MXNet allows us to do so by using [Dense](https://mxnet.incubator.apache.org/api/python/gluon/nn.html#mxnet.gluon.nn.Dense) layer and specifying the number of units to 1. The rest of the network can be arbitrarily complex. 
 
 Below, we define a model which has an input layer of 10 neurons, a couple of inner layers of 10 neurons each, and output layer of 1 neuron. We stack the layers using [HybridSequential](https://mxnet.incubator.apache.org/api/python/gluon/gluon.html#mxnet.gluon.nn.HybridSequential) block and initialize parameters of the network using [Xavier](https://mxnet.incubator.apache.org/api/python/optimization/optimization.html#mxnet.initializer.Xavier) initialization. 
 
@@ -69,7 +69,7 @@ with net.name_scope():
     net.add(nn.Dense(units=10, activation='relu'))   # inner layer 2
     net.add(nn.Dense(units=1))   # output layer: notice, it must have only 1 neuron
 
-net.initialize(mx.init.Xavier(magnitude=2.34))
+net.initialize(mx.init.Xavier())
 ```
 
 After defining the model, we need to define a few more things: our loss, our trainer and our metric.
@@ -78,7 +78,7 @@ Loss function is used to calculate how the output of the network differs from th
 
 Trainer object allows to specify the method of training to be used. For our tutorial we use [Stochastic Gradient Descent (SGD)](https://mxnet.incubator.apache.org/api/python/optimization/optimization.html#mxnet.optimizer.SGD). For more information on SGD refer to [the following tutorial](https://gluon.mxnet.io/chapter06_optimization/gd-sgd-scratch.html). We also need to parametrize it with learning rate value, which defines the weight updates, and weight decay, which is used for regularization.
 
-Metric helps us to estimate how good our model is in terms of a problem we are trying to solve. Where loss function has more importance for the training process, a metric is usually the thing we are trying to improve and reach maximum value. In our example, we are using [Accuracy](https://mxnet.incubator.apache.org/api/python/model.html#mxnet.metric.Accuracy) as a measurement of success of our model. 
+Metric helps us to estimate how good our model is in terms of a problem we are trying to solve. Where loss function has more importance for the training process, a metric is usually the thing we are trying to improve and reach maximum value. We also can use more than one metric, to measure various aspects of our model. In our example, we are using [Accuracy](https://mxnet.incubator.apache.org/api/python/model.html#mxnet.metric.Accuracy) and [F1 score](https://mxnet.incubator.apache.org/api/python/model.html#mxnet.metric.F1) as measurements of success of our model. 
 
 Below we define these objects.
 
@@ -86,11 +86,12 @@ Below we define these objects.
 ```python
 loss = gluon.loss.SigmoidBinaryCrossEntropyLoss()
 trainer = Trainer(params=net.collect_params(), optimizer='sgd', 
-                  optimizer_params={'learning_rate': 0.1, "wd": 0.01})
+                  optimizer_params={'learning_rate': 0.1})
 accuracy = mx.metric.Accuracy()
+f1 = mx.metric.F1()
 ```
 
-The next step is to define the training function in which we iterate over all batches of training data, execute the forward pass on each batch and calculate training loss. On line 19, we sum losses of every batch per an epoch into a single variable, because we calculate loss per single batch, but want to display it per epoch.
+The next step is to define the training function in which we iterate over all batches of training data, execute the forward pass on each batch and calculate training loss. On line 19, we sum losses of every batch per epoch into a single variable, because we calculate loss per single batch, but want to display it per epoch.
 
 
 ```python
@@ -133,6 +134,14 @@ Because of the behaviour above, you will get an unexpected result if you just ap
 
 After these transformations we can pass the result to `Accuracy.update()` method and expect it to behave in a proper way.
 
+For `F1` metric to work, instead of one number per class, we must pass probabilities of belonging to both classes. Because of that, on lines 21-22 we:
+
+1. Reshape predictions to a single vector
+
+2. We stack together two vectors: probabilities of belonging to class 0 (1 - `prediction`) and probabilities of belonging to class 1.
+
+Then we pass this stacked matrix to `F1` score.
+
 
 ```python
 def validate_model(threshold):
@@ -145,11 +154,20 @@ def validate_model(threshold):
         # Similar to cumulative training loss, calculate cumulative validation loss
         cumulative_val_loss += nd.sum(loss(output, val_ground_truth_class)).asscalar()
         
+        # getting prediction as a sigmoid
+        prediction = net(val_data).sigmoid()
+        
         # Converting neuron outputs to classes
-        prediction = mx.nd.ceil(net(val_data).sigmoid() - threshold)
+        predicted_classes = mx.nd.ceil(prediction - threshold)
         
         # Update validation accuracy
-        accuracy.update(val_ground_truth_class, prediction.reshape(-1)) 
+        accuracy.update(val_ground_truth_class, predicted_classes.reshape(-1)) 
+        
+        # calculate probabilities of belonging to different classes. F1 metric works only with this notation
+        prediction = prediction.reshape(-1)
+        probabilities = mx.nd.stack(1 - prediction, prediction, axis=1)
+        
+        f1.update(val_ground_truth_class, probabilities)
 
     return cumulative_val_loss
 ```
@@ -164,44 +182,33 @@ epochs = 10
 threshold = 0.5
 
 for e in range(epochs):
-    cumulative_train_loss = train_model()
-    cumulative_val_loss = validate_model(threshold)
+    avg_train_loss = train_model() / train_data_size
+    avg_val_loss = validate_model(threshold) / val_data_size
     
-    print("Epoch: %s, Training loss: %.2f, Validation loss: %.2f, Validation accuracy: %s" % 
-          (e, cumulative_train_loss, cumulative_val_loss, accuracy.get()[1]))
+    print("Epoch: %s, Training loss: %.2f, Validation loss: %.2f, Validation accuracy: %.2f, F1 score: %.2f" % 
+          (e, avg_train_loss, avg_val_loss, accuracy.get()[1], f1.get()[1]))
 
     # we reset accuracy, so the new epoch's accuracy would be calculated from the blank state
     accuracy.reset()
 ```
 
-    Epoch: 0, Training loss: 447.90, Validation loss: 40.13, Validation accuracy: 0.85 <!--notebook-skip-line-->
-
-    Epoch: 1, Training loss: 356.33, Validation loss: 34.38, Validation accuracy: 0.85 <!--notebook-skip-line-->
-
-    Epoch: 2, Training loss: 238.26, Validation loss: 16.34, Validation accuracy: 0.93 <!--notebook-skip-line-->
-
-    Epoch: 3, Training loss: 106.45, Validation loss: 13.55, Validation accuracy: 0.95 <!--notebook-skip-line-->
-
-    Epoch: 4, Training loss: 77.17, Validation loss: 8.87, Validation accuracy: 0.97 <!--notebook-skip-line-->
-
-    Epoch: 5, Training loss: 60.52, Validation loss: 10.60, Validation accuracy: 0.96 <!--notebook-skip-line-->
-
-    Epoch: 6, Training loss: 55.00, Validation loss: 8.23, Validation accuracy: 0.97 <!--notebook-skip-line-->
-
-    Epoch: 7, Training loss: 56.08, Validation loss: 10.59, Validation accuracy: 0.96 <!--notebook-skip-line-->
-
-    Epoch: 8, Training loss: 56.10, Validation loss: 6.74, Validation accuracy: 0.97 <!--notebook-skip-line-->
-
-    Epoch: 9, Training loss: 51.81, Validation loss: 7.48, Validation accuracy: 0.98 <!--notebook-skip-line-->
+    Epoch: 0, Training loss: 0.43, Validation loss: 0.36, Validation accuracy: 0.85, F1 score: 0.00
+    Epoch: 1, Training loss: 0.22, Validation loss: 0.14, Validation accuracy: 0.96, F1 score: 0.35
+    Epoch: 2, Training loss: 0.09, Validation loss: 0.11, Validation accuracy: 0.97, F1 score: 0.48
+    Epoch: 3, Training loss: 0.07, Validation loss: 0.09, Validation accuracy: 0.96, F1 score: 0.53
+    Epoch: 4, Training loss: 0.06, Validation loss: 0.09, Validation accuracy: 0.97, F1 score: 0.58
+    Epoch: 5, Training loss: 0.04, Validation loss: 0.12, Validation accuracy: 0.97, F1 score: 0.59
+    Epoch: 6, Training loss: 0.05, Validation loss: 0.09, Validation accuracy: 0.99, F1 score: 0.62
+    Epoch: 7, Training loss: 0.05, Validation loss: 0.10, Validation accuracy: 0.97, F1 score: 0.62
+    Epoch: 8, Training loss: 0.05, Validation loss: 0.12, Validation accuracy: 0.95, F1 score: 0.63
+    Epoch: 9, Training loss: 0.04, Validation loss: 0.09, Validation accuracy: 0.98, F1 score: 0.65
 
 
-In our case we easily hit the accuracy of 0.98.
+In our case we hit the accuracy of 0.98 and F1 score of 0.65.
 
 ## Tip 1: Use only one neuron in the output layer
 
 Despite that there are 2 classes, there should be only one output neuron, because `SigmoidBinaryCrossEntropyLoss` accepts only one feature as an input. 
-
-In case when there are 3 or more classes, one cannot use a single Logistic regression, but should do multiclass regression. The solution would be to increase the number of output neurons to the number of classes and use `SoftmaxCrossEntropyLoss`. 
 
 ## Tip 2: Encode classes as 0 and 1
 
@@ -215,8 +222,6 @@ NDArray API has two options to calculate logistic regression loss: [SigmoidBinar
 
 In this tutorial I explained some potential pitfalls to be aware of. When doing logistic regression using Gluon API remember to:
 1. Use only one neuron in the output layer
-2. Encode class labels as 0 or 1
-3. Use `SigmoidBinaryCrossEntropyLoss`
-4. Convert probabilities to classes before calculating Accuracy
-
- <!-- INSERT SOURCE DOWNLOAD BUTTONS -->
+1. Encode class labels as 0 or 1
+1. Use `SigmoidBinaryCrossEntropyLoss`
+1. Convert probabilities to classes before calculating Accuracy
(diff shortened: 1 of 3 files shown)


Ishitori:
@thomelane @ThomasDelteil @safrooze or @indhub please take a look

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+    net.add(nn.Dense(units=1))   # output layer: notice, it must have only 1 neuron
+
+net.initialize(mx.init.Xavier(magnitude=2.34))
+```
+
+After defining the model, we need to define a few more thing: our loss, our trainer and our metric.
+
+Loss function is used to calculate how the output of the network different from the ground truth. In case of the logistic regression the ground truth are class labels, which can be either 0 or 1. Because of that, we are using `SigmoidBinaryCrossEntropyLoss`, which suites well for that scenario.
+
+Trainer object allows to specify the method of training to be used. There are various methods available, and for our tutorial we use a widely accepted method Stochastic Gradient Descent. We also need to parametrize it with learning rate value, which defines how fast training happens, and weight decay which is used for regularization.
```
ifeherva: "We also need to parametrize it with learning rate value, which defines how fast training happens" The learning rate of SGD defines the weight updates, not necessarily how fast training happens. I propose to reword it a bit to avoid confusion that large LR will lead to fast training.

safrooze: Suggest to make a brief statement that you're using SGD, and if needed, include a link for further reading.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+    net.add(nn.Dense(units=10, activation='relu'))   # inner layer 1
+    net.add(nn.Dense(units=10, activation='relu'))   # inner layer 2
+    net.add(nn.Dense(units=1))   # output layer: notice, it must have only 1 neuron
+
+net.initialize(mx.init.Xavier(magnitude=2.34))
+```
+
+After defining the model, we need to define a few more thing: our loss, our trainer and our metric.
+
+Loss function is used to calculate how the output of the network different from the ground truth. In case of the logistic regression the ground truth are class labels, which can be either 0 or 1. Because of that, we are using `SigmoidBinaryCrossEntropyLoss`, which suites well for that scenario.
```
safrooze: I think if instead of defining what a loss is, you'd define what `SigmoidBinaryCrossEntropyLoss` does, user may benefit more. Something that explains there is a `from_sigmoid` argument, which means that the output of the neuron doesn't need to go through sigmoid, but at inference you'd have to pass it through sigmoid. Also include link to API.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+
+
+```python
+loss = gluon.loss.SigmoidBinaryCrossEntropyLoss()
+trainer = Trainer(params=net.collect_params(), optimizer='sgd', 
+                  optimizer_params={'learning_rate': 0.1, "wd": 0.01})
+accuracy = mx.metric.Accuracy()
+```
+
+Usually, it is not enough to pass the training data through a network only once to achieve high Accuracy. It helps when the network sees each example multiple times. The number of displaying every example to the network is called `epoch`. How big this number should be is unknown in advance, and usually it is estimated using trial and error approach.
```
safrooze: Suggest to completely skip description about what an epoch is.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+```python
+loss = gluon.loss.SigmoidBinaryCrossEntropyLoss()
+trainer = Trainer(params=net.collect_params(), optimizer='sgd', 
+                  optimizer_params={'learning_rate': 0.1, "wd": 0.01})
+accuracy = mx.metric.Accuracy()
+```
+
+Usually, it is not enough to pass the training data through a network only once to achieve high Accuracy. It helps when the network sees each example multiple times. The number of displaying every example to the network is called `epoch`. How big this number should be is unknown in advance, and usually it is estimated using trial and error approach. 
+
+Below we are defining the main training loop, which go through each example in batches specified number of times (epochs). After each epoch we display training loss, validation loss and calculate accuracy of the model using validation set. For now, let's take a look into the code, and I am going to explain the details later.
```
safrooze: "validataion dateset"

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+            # Do forward pass on a batch of training data
+            output = net(data)
+        
+            # Calculate loss for the training data batch
+            loss_result = loss(output, label)
+
+        # Calculate gradients 
+        loss_result.backward()
+
+        # Change parameters of the network
```
safrooze: use "update" instead of "change"

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+        
+        # Reshape predictions to match dimension of val_ground_truth_class
+        # and update accuracy with the results for that batch.
+        accuracy.update(val_ground_truth_class, prediction.reshape(-1)) 
+    
+    # in the end of epoch, we print out current values for epoch, training and validation losses, and accuracy
+    print("Epoch: %s, Training loss: %.2f, Validation loss: %.2f, Validation accuracy: %s" % 
+          (e, cumulative_train_loss, cumulative_val_loss, accuracy.get()[1]))
+
+    # we reset accuracy, so the new epoch's accuracy would be calculate from the blank state
```
safrooze: would be ***calculated***

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+        loss_result.backward()
+
+        # Change parameters of the network
+        trainer.step(batch_size)
+
+        # Since we calculate loss per single batch, but want to display it per epoch
+        # we sum losses of every batch per an epoch into a single variable
+        cumulative_train_loss += nd.sum(loss_result).asscalar()
+
+    # iterate over all batches of validation data and calculate validation loss
```
safrooze: Split into train and validate function.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+    for i, (val_data, val_ground_truth_class) in enumerate(val_dataloader):
+        # Do forward pass on a batch of validation data
+        output = net(val_data)
+        
+        # Similar to cumulative training loss, calculate cumulative validation loss
+        cumulative_val_loss += nd.sum(loss(output, val_ground_truth_class)).asscalar()
+        
+        # Applying sigmoid function, to get data in range [0, 1] and then
```
safrooze: In a notebook, better to keep long comments as text, rather than comment section of code.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+        # subtracting threshold, to make 0 serve as a class boundary: below 0 - class 0, above 0 - class 1
+        # Apply mx.nd.ceil to get classes: convert negative values to 0 and positive to 1.
+        prediction = mx.nd.ceil(net(val_data).sigmoid() - threshold)
+        
+        # Reshape predictions to match dimension of val_ground_truth_class
+        # and update accuracy with the results for that batch.
+        accuracy.update(val_ground_truth_class, prediction.reshape(-1)) 
+    
+    # in the end of epoch, we print out current values for epoch, training and validation losses, and accuracy
```
safrooze: Stating the obvious.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+        # Since we calculate loss per single batch, but want to display it per epoch
+        # we sum losses of every batch per an epoch into a single variable
+        cumulative_train_loss += nd.sum(loss_result).asscalar()
+
+    # iterate over all batches of validation data and calculate validation loss
+    for i, (val_data, val_ground_truth_class) in enumerate(val_dataloader):
+        # Do forward pass on a batch of validation data
+        output = net(val_data)
+        
+        # Similar to cumulative training loss, calculate cumulative validation loss
```
safrooze: stating the obvious

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+        # Change parameters of the network
+        trainer.step(batch_size)
+
+        # Since we calculate loss per single batch, but want to display it per epoch
+        # we sum losses of every batch per an epoch into a single variable
+        cumulative_train_loss += nd.sum(loss_result).asscalar()
+
+    # iterate over all batches of validation data and calculate validation loss
+    for i, (val_data, val_ground_truth_class) in enumerate(val_dataloader):
+        # Do forward pass on a batch of validation data
```
safrooze: stating the obvious.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+
+# iterate over training data epochs number of times
+for e in range(epochs):
+    cumulative_train_loss = 0
+    cumulative_val_loss = 0
+    
+    # iterate over all batches of training data and calculate training loss
+    for i, (data, label) in enumerate(train_dataloader):
+        with autograd.record():
+            # Do forward pass on a batch of training data
```
safrooze: stating the obvious.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+```python
+epochs = 10
+threshold = 0.5
+
+# iterate over training data epochs number of times
+for e in range(epochs):
+    cumulative_train_loss = 0
+    cumulative_val_loss = 0
+    
+    # iterate over all batches of training data and calculate training loss
```
safrooze: stating the obvious.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+Usually, it is not enough to pass the training data through a network only once to achieve high Accuracy. It helps when the network sees each example multiple times. The number of displaying every example to the network is called `epoch`. How big this number should be is unknown in advance, and usually it is estimated using trial and error approach. 
+
+Below we are defining the main training loop, which go through each example in batches specified number of times (epochs). After each epoch we display training loss, validation loss and calculate accuracy of the model using validation set. For now, let's take a look into the code, and I am going to explain the details later.
+
+
+```python
+epochs = 10
+threshold = 0.5
+
+# iterate over training data epochs number of times
```
safrooze: stating the obvious.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Tip 1: Use only one neuron in the output layer
+
+Despite that there are 2 classes, there should be only one output neuron, because `SigmoidBinaryCrossEntropyLoss` accepts only one feature as an input. 
+
+In case when there are 3 or more classes, one cannot use a single Logistic regression, but should do multiclass regression. The solution would be to increase the number of output neurons to the number of classes and use `SoftmaxCrossEntropyLoss`. 
+
+## Tip 2: Encode classes as 0 and 1
+
+`Sigmoid` function produces values from 0 to 1. `SigmoidBinaryCrossEntropyLoss` uses these values to calculate the   loss by essentially subtracting values and class labels from 1. [Here is the formula](https://mxnet.incubator.apache.org/api/python/gluon/loss.html?highlight=sigmoidbinarycrossentropyloss#mxnet.gluon.loss.SigmoidBinaryCrossEntropyLoss) used for that calculation (we use default version with `from_sigmoid` is False). That's why it is numerically better to have classes encoded in the same range as a `Sigmoid` output with 0 and 1.
```
safrooze: Don't get the last sentence. What is exactly better?

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+
+If your data comes with a label encoded in a different format, such as -1 and 1, then you can either recode it to 0 and 1 by comparing the initial class to 0, or use another function instead of `Sigmoid`, like [`Tanh`](https://mxnet.incubator.apache.org/api/python/ndarray/ndarray.html?highlight=tanh#mxnet.ndarray.tanh), to produce output in range [-1; 1].
```
safrooze: Not sure if mentioning tanh has any merit here. tanh is just not used for logistic regression. What loss do I use with tanh? I find it confusing.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Tip 3: Use SigmoidBinaryCrossEntropyLoss instead of LogisticRegressionOutput
```
safrooze: Include link to API.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+
+NDArray API has two options to calculate logistic regression loss. One is `SigmoidBinaryCrossEntropyLoss`, which I used in the example. This class inherits from the `Loss` class and is intended to be used as a loss function for logistic regression. But there is also a function called `LogisticRegressionOutput`, which can be applied to any `NDArray`. Mathematically speaking, this function does the same thing as `SigmoidBinaryCrossEntropyLoss`. 
+
+My recommendation would be to use `SigmoidBinaryCrossEntropyLoss`, because this class properly inherits from `Loss` class, while `LogisticRegressionOutput` is just a regular function. `LogisticRegressionOutput` is a   function to go when implementing logistic regression using Symbol API, but in case of using Gluon API, there are no benefits using it. The only case when you may want to consider using `LogisticRegressionOutput` as your loss, is when you need to have a support for sparse matrices.
```
safrooze: This paragraph needs rewording. It doesn't really make a case for why not use LogisticRegressionOutput (i.e. the fact that it is designed to be an output layer when using the Module API). Also what's the problem with using sparse matrices and SigmoidBinaryCrossEntropyLoss?! What happens if I do? What fails? You may consider just getting rid of this tip altogether to avoid any confusion.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Tip 4: Convert probabilities to classes before calculating Accuracy
+
+`Accuracy` metric requires 2 arguments: 1) a vector of ground-truth classes and 2) A tensor of predictions. When tensor of predictions is of the same shape as the vector of ground-truth classes, `Accuracy` class assumes that it should contain predicted classes. So, it converts the vector to `Int32` and compare each item of ground-truth classes to prediction vector.
```
safrooze: Why is one called a vector and the other called a tensor?

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Tip 4: Convert probabilities to classes before calculating Accuracy
+
+`Accuracy` metric requires 2 arguments: 1) a vector of ground-truth classes and 2) A tensor of predictions. When tensor of predictions is of the same shape as the vector of ground-truth classes, `Accuracy` class assumes that it should contain predicted classes. So, it converts the vector to `Int32` and compare each item of ground-truth classes to prediction vector. 
+
+Because of the behaviour above, you will get an unexpected result if you just pass the output of `Sigmoid` function as is. `Sigmoid` function produces output in range [0; 1], and all numbers in that range are going to be casted to 0, even if it is as high as 0.99. To avoid this we write a custom bit of code, that:
```
safrooze: Instead of having a tip here, I recommend to write the rounding as a separate function in notebook and include a paragraph on top explaining what it's doing. This paragraph is really the only motivation for writing this tutorial.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+
+2. Use `mx.nd.ceil` function, which converts all negative values to 0 and all positive values to 1
+
+After these transformations we can pass the result to `SigmoidBinaryCrossEntropyLoss` and expect it to behave in the defined way.
+
+The same is not true, if the output shape of your function is different from the shape of ground-truth classes vector. For example, when doing multiclass regression with `Softmax` as an output, the shape of the output is going to be *number_of_examples* x *number_of_classes*. In that case we don't need to do the transformation above, because `Accuracy` metric would understand that shapes are different and will assume that the prediction contains probabilities of an example to belong to these classes - exactly what we want it to be. 
+
+This makes things a little bit easier, and that's why I have seen examples where `Softmax` is used as an output of prediction. If you want to do that, make sure to change the output layer size to 2 neurons, where each neuron will provide a value of an example to belong to class 0 and 1 respectively.
```
safrooze: Wouldn't call it *easier*. Softmax allows the output layer to have more flexibility in learning because it contains twice the number of parameters to learn.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Conclusion
+
+In this tutorial I explained some potential pitfalls to be aware about when doing logistic regression in Apache MXNet Gluon API. There might be some other challenging scenarios, which are not covered in this tutorial, like dealing with imbalanced classes, but I hope this tutorial will serve as a guidance and all other potential pitfalls would be covered in future tutorials.
```
safrooze: "be aware of"

safrooze: Suggest to keep the conclusion factual and not include reference to one of many challenges or "hoping" the tutorial is useful. Also this is not a blog (no posting time associated with it). Future tutorials has no meaning.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
+
+# Logistic regression using Gluon API explained
```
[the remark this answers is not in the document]
Ishitori: There is a lot of similarity with the tutorial you have mentioned. My personal pet-peeve with the tutorial in The Straight Dope is that it doesn't use library functions for sigmoid, loss and accuracy. I believe it creates an impression that you need to write a lot of boiler plate code instead of using what MXNet offers. 

Also, there were confusions around number of neurons in forum and stackoverflow which is directly relates to the usage of `Accuracy`, but needs a bit more context to cover.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+import numpy as np
+import mxnet as mx
+from mxnet import nd, autograd, gluon
+from mxnet.gluon import nn, Trainer
+from mxnet.gluon.data import DataLoader, ArrayDataset
+
+mx.random.seed(12345)  # Added for reproducibility
+```
+
+In this tutorial we will use fake dataset, which contains 10 features drawn from a normal distribution with mean equals to 0 and standard deviation equals to 1, and a class label, which can be either 0 or 1. The length of the dataset is an arbitrary value. The function below helps us to generate a dataset.
```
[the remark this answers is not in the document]
Ishitori: I didn't want to add extra code for data loading and data processing, because this is not the point of this tutorial. The optimal way would be if there is a binary classification dataset in the mxnet itself, so it can be loaded in one line and no pre-processing would be required.

docs/tutorials/gluon/logistic_regression_explained.md
Ishitori: Good suggestion, fixed.

docs/tutorials/gluon/logistic_regression_explained.md
Ishitori: If you try to use `SigmoidBinaryCrossEntropyLoss` with sparse data, it would throw an exception at some point, where it tries to do a reshape:

```NotSupportedForSparseNDArray: Function reshape with arguments (<class 'tuple'>) is not supported for SparseNDArray and only available in NDArray.```

File: docs/tutorials/gluon/logistic_regression_explained.md [hunk 25]
```diff
@@ -0,0 +1,215 @@
...
+```python
+ctx = mx.cpu()
+train_data_size = 1000
+val_data_size = 100
+batch_size = 10
+```
+
+## Working with data
+
+To work with data, Apache MXNet provides Dataset and DataLoader classes. The former is used to provide an indexed access to the data, the latter is used to shuffle and batchify the data.
```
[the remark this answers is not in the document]
Ishitori: Yes, added links and remove my explanation. Added a link to Datasets and Dataloaders tutorial instead.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+batch_size = 10
+```
+
+## Working with data
+
+To work with data, Apache MXNet provides Dataset and DataLoader classes. The former is used to provide an indexed access to the data, the latter is used to shuffle and batchify the data. 
+
+This separation is done because a source of Dataset can vary from a simple array of numbers to complex data structures like text and images. DataLoader doesn't need to be aware of the source of data as long as Dataset provides a way to get the number of records and to load a record by index. As an outcome, Dataset doesn't need to hold in memory all data at once. Needless to say, that one can implement its own versions of Dataset and DataLoader, but we are going to use existing implementation.
+
+Below we define 2 datasets: training dataset and validation dataset. It is a good practice to measure performance of a trained model on a data that the network hasn't seen before. That is why we are going to use training set for training the model and validation set to calculate model's accuracy.
```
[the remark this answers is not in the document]
Ishitori: Agree. Removed this one as well.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+train_dataloader = DataLoader(train_dataset, batch_size=batch_size, shuffle=True)
+
+val_x, val_ground_truth_class = get_random_data(val_data_size, ctx)
+val_dataset = ArrayDataset(val_x, val_ground_truth_class)
+val_dataloader = DataLoader(val_dataset, batch_size=batch_size, shuffle=True)
+```
+
+## Defining and training the model
+
+In real application, model can be arbitrary complex. The only requirement for the logistic regression is that the last layer of the network must be a single neuron. Apache MXNet allows us to do so by using `Dense` layer and specifying the number of units to 1.
```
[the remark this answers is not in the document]
Ishitori: Done

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+
+Below, we define a model which has an input layer of 10 neurons, a couple inner layers of 10 neurons each, and output layer of 1 neuron, as it is required by logistic regression. We stack the layers using `HybridSequential` block and initialize parameters of the network using `Xavier` initialization.
```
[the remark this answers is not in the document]
Ishitori: Fixed

docs/tutorials/gluon/logistic_regression_explained.md
Ishitori: I mention Stochastic Gradient Descent in the text above. So, made the text shorter and added links to the docs and to the Straight Dope tutorial.

docs/tutorials/gluon/logistic_regression_explained.md
Ishitori: I decided to not remove a sentence explaining what loss does, but add links to official documentation and wikipedia. I also pasted your explanation of from_sigmoid - it might be useful to explain it even when we don't set this argument.

docs/tutorials/gluon/logistic_regression_explained.md
Ishitori: Sure, not much value here.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -0,0 +1,215 @@
...
+with net.name_scope():
+    net.add(nn.Dense(units=10, activation='relu'))  # input layer
+    net.add(nn.Dense(units=10, activation='relu'))   # inner layer 1
+    net.add(nn.Dense(units=10, activation='relu'))   # inner layer 2
+    net.add(nn.Dense(units=1))   # output layer: notice, it must have only 1 neuron
+
+net.initialize(mx.init.Xavier(magnitude=2.34))
+```
+
+After defining the model, we need to define a few more thing: our loss, our trainer and our metric.
```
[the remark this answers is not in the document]
Ishitori: Fixed

docs/tutorials/gluon/logistic_regression_explained.md
Ishitori: Cannot find this typo.

docs/tutorials/gluon/logistic_regression_explained.md
Ishitori: Removed the print comment, simplified the `accuracy.update` comment

docs/tutorials/gluon/logistic_regression_explained.md
Ishitori: Removed the loop comment, but kept the forward pass one.

docs/tutorials/gluon/logistic_regression_explained.md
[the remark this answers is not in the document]
safrooze: "training set" -> "training dataset"
"validation set" -> "validation dataset"

docs/tutorials/gluon/logistic_regression_explained.md
Ishitori: Rewrote whole paragraph simplifying.

docs/tutorials/gluon/logistic_regression_explained.md
Ishitori: Removed Tanh part.

docs/tutorials/gluon/logistic_regression_explained.md
Ishitori: This block was deleted.

docs/tutorials/gluon/logistic_regression_explained.md
Ishitori: I rewrote the paragraph, and I actually deleted the sparse part from it as well. @eric-haibin-lin is going to work on this at some point, so to not update the tutorial later, I just omit this part.

docs/tutorials/gluon/logistic_regression_explained.md
Ishitori: Replaced it with "vector of matrix"

docs/tutorials/gluon/logistic_regression_explained.md
Ishitori: Rewrote the conclusion by reiterating over main points of the tutorial.

Ishitori: Fixed

yifeim:
The tutorial looks awesome! A few comments:

* Is it just me or everybody experiences training loss 8x larger than validation loss? I understand it may be possible due to unstable weights from backward update after every batch. But, still it is not my usual expectation. (Not to say that I have not seen similar issues before and I am still slightly puzzled by the root cause.)

* The model class is linear, but the proposed network is a 3-layer overkill (which may be related to the larger training loss). If you want a fun problem, maybe consider a `xor` function class: https://medium.com/@jayeshbahire/the-xor-problem-in-neural-networks-50006411840b

* Is this `mxnet.gluon.loss.LogisticLoss` equivalent to `SigmoidBCELoss`? Since you are using metrics, it may be worthwhile exploring `mx.metric.F1`.

* There are some magic hyperparameters to be explained: a strong wd=0.01, an Xavier=2.34 initialization. As an elementary tutorial, I would try to simplify them unless they are part of the intended purposes.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -120,96 +108,115 @@ for e in range(epochs):
...
+## Tip 1: Use only one neuron in the output layer
 
-Because of the behaviour above, you will get an unexpected result if you just pass the output of `Sigmoid` function as is. `Sigmoid` function produces output in range [0; 1], and all numbers in that range are going to be casted to 0, even if it is as high as 0.99. To avoid this we write a custom bit of code, that:
+Despite that there are 2 classes, there should be only one output neuron, because `SigmoidBinaryCrossEntropyLoss` accepts only one feature as an input. 
 
-1. Subtracts a threshold from the original prediction. Usually, the threshold is equal to 0.5, but it can be higher, if you want to increase certainty of an item to belong to class 1.
+In case when there are 3 or more classes, one cannot use a single Logistic regression, but should do multiclass regression. The solution would be to increase the number of output neurons to the number of classes and use `SoftmaxCrossEntropyLoss`.
```
safrooze: * "In case there are".
* Multi-class regression refers to the case when one item belongs to multiple classes. In that case you cannot use softmax, but rather multiple logistic regressions, one for each class. I suggest to simply remove this sentence to not create any confusion.

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -59,9 +55,9 @@ val_dataloader = DataLoader(val_dataset, batch_size=batch_size, shuffle=True)
 
 ## Defining and training the model
 
-In real application, model can be arbitrary complex. The only requirement for the logistic regression is that the last layer of the network must be a single neuron. Apache MXNet allows us to do so by using `Dense` layer and specifying the number of units to 1.
+The only requirement for the logistic regression is that the last layer of the network must be a single neuron. Apache MXNet allows us to do so by using [Dense](https://mxnet.incubator.apache.org/api/python/gluon/nn.html#mxnet.gluon.nn.Dense) layer and specifying the number of units to 1. The rest of the network can be arbitrary complex.
```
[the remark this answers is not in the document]
Ishitori: Fixed

docs/tutorials/gluon/logistic_regression_explained.md
```diff
@@ -94,21 +90,13 @@ trainer = Trainer(params=net.collect_params(), optimizer='sgd',
 accuracy = mx.metric.Accuracy()
 ```
 
-Usually, it is not enough to pass the training data through a network only once to achieve high Accuracy. It helps when the network sees each example multiple times. The number of displaying every example to the network is called `epoch`. How big this number should be is unknown in advance, and usually it is estimated using trial and error approach. 
-
-Below we are defining the main training loop, which go through each example in batches specified number of times (epochs). After each epoch we display training loss, validation loss and calculate accuracy of the model using validation set. For now, let's take a look into the code, and I am going to explain the details later.
+The next step is to define the training function in which we iterate over all batches of training data, execute the forward pass on each batch and calculate training loss. On line 19, we sum losses of every batch per an epoch into a single variable, because we calculate loss per single batch, but want to display it per epoch.
```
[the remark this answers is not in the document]
Ishitori: Fixed

docs/tutorials/gluon/logistic_regression_explained.md
Ishitori: Removed

Ishitori:
@yifeim, Thanks for comments, I am answering them in order:

1. The loss is so different, because I didn't do the averaging by number of items in training and validation set respectively. You are right that it is better to do so, otherwise numbers look confusing. Did that.

2. While XOR problem is fun and gives a good understanding why you need activation functions, it requires inputs of 0 and 1 as well. I wouldn't want to limit this part of the tutorial to this, as it seems a better use case to showcase arbitrary inputs.

3. `LogisticLoss` can be used instead of `SigmoidBCELoss`. There is a good explanation about both of them: https://stats.stackexchange.com/questions/229645/why-there-are-two-different-logistic-loss-formulation-notations/231994  

I haven't seen `LogisticLoss` used much in practice. I don't want to mention the second option, because it is going to confuse these who are new to Apache MXNet.

I added a showcase how to use F1 metric.

4. You are probably right - I didn't have an intention of going into default parameters as it may require long explanation. Just removed wd and use default of Xavier.

yifeim:
Awesome. Done and done! Thanks for addressing the questions.

PR #11749, opened 2018-07-13
[MXNET-8230] test_operator_gpu.test_rms fails

frankfliu:
## Description ##
Fix flaky test#8230, test_rms fails.

### Essentials ###

### Changes ###
- [X] Fix flaky test#8230, test_rms fails.

This flaky test was due to compare result using mx.test_utils.almost_equal() with high precision.
Reducing tolerance to 1e-1 for float16 and 1e-2 for float32. The test passed 10K iterations after this fix.

### DIFF
--- a/tests/python/unittest/test_optimizer.py
+++ b/tests/python/unittest/test_optimizer.py
@@ -835,7 +835,6 @@ class PyRMSProp(mx.optimizer.Optimizer):
         if self.clip_weights:
              mx.ndarray.clip(weight, -self.clip_weights, self.clip_weights, out=weight)
 
-# @unittest.skip("Test fails intermittently. Temporarily disabled until fixed. Tracked at https://github.com/apache/incubator-mxnet/issues/8230")
 @with_seed()
 def test_rms():
     opt1 = PyRMSProp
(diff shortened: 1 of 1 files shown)


tests/python/unittest/test_optimizer.py
```diff
@@ -835,8 +835,8 @@ def update(self, index, weight, grad, state):
         if self.clip_weights:
              mx.ndarray.clip(weight, -self.clip_weights, self.clip_weights, out=weight)
 
-@unittest.skip("Test fails intermittently. Temporarily disabled until fixed. Tracked at https://github.com/apache/incubator-mxnet/issues/8230")
-@with_seed(0)
+# @unittest.skip("Test fails intermittently. Temporarily disabled until fixed. Tracked at https://github.com/apache/incubator-mxnet/issues/8230")
```
haojin2: If you're getting rid of the line please delete it.

tests/python/unittest/test_optimizer.py
```diff
@@ -848,6 +848,13 @@ def test_rms():
     wd_options = [{}, {'wd': 0.03}, {'wd': 0.05}, {'wd': 0.07}]
     mp_options = [{}, {'multi_precision': False}, {'multi_precision': True}]
     for dtype in [np.float16, np.float32]:
+        if (dtype == np.float16):
+            rtol=1e-1
+            atol=1e-1
+        else:
+            rtol=1e-2
+            atol=1e-2
```
haojin2: I would prefer:
```Python
rtol = 1e-1 if dtype is np.float16 else 1e-2
atol = 1e-1 if dtype is np.float16 else 1e-2
```

tests/python/unittest/test_optimizer.py
lupesko: +1

tests/python/unittest/test_optimizer.py
lupesko: IMO the way @frankfliu wrote it is clearer and easier to parse.
But would be great to have a comment explaining it (it was learned by trial and error, right?)

lupesko:
@frankfliu thanks for contributing the fix! 
Can you please update the issue description:
- Tick all the boxes to show that you addressed the standard PR requirements (you can remove those that are not relevant)
- Under "Changes", explain the change briefly.

tests/python/unittest/test_optimizer.py
frankfliu: done

tests/python/unittest/test_optimizer.py
frankfliu: done

PR #11765, opened 2018-07-15
Fix file name creation for Windows

Ishitori:
## Description ##
By default filename is generated by just taking last part of the url. If it contains query string, which is separated with "?" symbol, file cannot be created on windows machine.

### Essentials ###

### Changes ###
- [X] Small file name generation fix for one particular tutorial

### DIFF
--- a/docs/tutorials/python/predict_image.md
+++ b/docs/tutorials/python/predict_image.md
@@ -57,8 +57,8 @@ from collections import namedtuple
 Batch = namedtuple('Batch', ['data'])
 
 def get_image(url, show=False):
-    # download and show the image
-    fname = mx.test_utils.download(url)
+    # download and show the image. Remove query string from the file name.
+    fname = mx.test_utils.download(url, fname=url.split('/')[-1].split('?')[0])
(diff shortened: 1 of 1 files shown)


docs/tutorials/python/predict_image.md
```diff
@@ -57,8 +57,8 @@ from collections import namedtuple
 Batch = namedtuple('Batch', ['data'])
 
 def get_image(url, show=False):
-    # download and show the image
-    fname = mx.test_utils.download(url)
+    # download and show the image. Remove query string from the file name.
+    fname = mx.test_utils.download(url, fname=url.split('/')[-1].split('?')[0])
```
zhreshold: should you make sure it's not empty in case url is 'https://foo.bar/?raw=1'

zhreshold: even though it's really rare to happen

Ishitori: Yeah, I don't think it will happen here, because the fix is done only for this particular tutorial - all urls are known in advance.

Issue #10988, opened 2018-05-17
Flaky test: test_operator_gpu.test_countsketch


eric-haibin-lin:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/PR-10983/1/pipeline/

```
test_operator_gpu.test_exc_multiple_waits ... ok (0.0126s)

**test_operator_gpu.test_countsketch ... [INFO] Setting test np/mx/python random seeds, use MXNET_TEST_SEED=0 to reproduce.**

**FAIL**

test_operator_gpu.test_cached ... SKIP: test fails intermittently. temporarily disabled till it gets fixed. tracked at https://github.com/apache/incubator-mxnet/issues/8049

test_operator_gpu.test_sparse_nd_setitem ... [INFO] Setting test np/mx/python random seeds, use MXNET_TEST_SEED=584830846 to reproduce.

ERROR

test_operator_gpu.test_residual ... ERROR

test_operator_gpu.test_lstm_dropout ... ok (0.0739s)

test_operator_gpu.test_parameter_sharing ... [INFO] Setting test np/mx/python random seeds, use MXNET_TEST_SEED=1302250019 to reproduce.

ERROR

test_operator_gpu.test_exc_post_fail ... ok (0.0255s)

test_operator_gpu.test_bce_equal_ce2 ... [INFO] Setting test np/mx/python random seeds, use MXNET_TEST_SEED=2007597932 to reproduce.

ERROR

test_operator_gpu.test_exc_mutable_var_fail ... ok (0.0051s)

test_operator_gpu.test_sparse_nd_slice ... [INFO] Setting test np/mx/python random seeds, use MXNET_TEST_SEED=28702949 to reproduce.

ERROR

test_operator_gpu.test_logistic_loss_equal_bce ... ERROR

test_operator_gpu.test_ndarray_elementwise ... [INFO] Setting test np/mx/python random seeds, use MXNET_TEST_SEED=0 to reproduce.

ERROR

test_operator_gpu.test_residual_bidirectional ... ERROR

test_operator_gpu.test_kl_loss ... [INFO] Setting test np/mx/python random seeds, use MXNET_TEST_SEED=1234 to reproduce.

ERROR

test_operator_gpu.test_parameter_str ... ok (0.0005s)

[...]

PR #11780, opened 2018-07-17
fix flaky test test_operator_gpu.test_countsketch

haojin2:
## Description ##
As title.

### Essentials ###

### Changes ###
- [x] Fix flaky count_sketch

## Comments ##
Passed more than 15000 times on my machine

### DIFF
--- a/src/operator/contrib/count_sketch.cu
+++ b/src/operator/contrib/count_sketch.cu
@@ -129,8 +129,8 @@ inline void CountSketchForward(const Tensor<gpu, 2, DType> &out,
                                     nthreads, out_ptr+bstart*out_dim, h_ptr,
                                     s_ptr, in_ptr+bstart*in_dim, batchlen,
                                     in_dim, out_dim);
-    MSHADOW_CUDA_POST_KERNEL_CHECK(sketch_forward_kernel);
-    // cudaThreadSynchronize();
+    cudaError_t err = cudaDeviceSynchronize();
+    CHECK_EQ(err, cudaSuccess) << "Error occured! CUDA: " << cudaGetErrorString(err);
     bstart = (i+1)*batchlen;
   }
 }
@@ -153,7 +153,7 @@ inline void CountSketchBackward(const Tensor<gpu, 2, DType> &in_grad,
     upper_bound = upper_bound-1;
   }
   // guarantee there are at least one iteration
-  upper_bound = upper_bound > 0? upper_bound:0;
+  upper_bound = upper_bound > 0 ? upper_bound : 0;
   int bstart = 0;
   for ( int i = 0; i <= upper_bound; i++ ) {
     const int batchlen = min(processing_batch_size, n_samples - bstart);
@@ -165,7 +165,8 @@ inline void CountSketchBackward(const Tensor<gpu, 2, DType> &in_grad,
                             nthreads, in_grad_ptr+bstart*in_dim, h_ptr,
                             s_ptr, out_grad_ptr+bstart*out_dim, batchlen,
                             in_dim, out_dim);
-    MSHADOW_CUDA_POST_KERNEL_CHECK(sketch_backward_kernel);
+    cudaError_t err = cudaDeviceSynchronize();
+    CHECK_EQ(err, cudaSuccess) << "Error occured! CUDA: " << cudaGetErrorString(err);
     bstart = (i+1)*batchlen;
   }
 }
--- a/tests/python/gpu/test_operator_gpu.py
+++ b/tests/python/gpu/test_operator_gpu.py
@@ -51,7 +51,10 @@ del test_support_vector_machine_l2_svm
 
 
 def check_countsketch(in_dim,out_dim,n):
-    sym = mx.sym.contrib.count_sketch(name='countsketch',out_dim = out_dim)
+    data = mx.sym.Variable("data")
+    h = mx.sym.Variable("h")
+    s = mx.sym.Variable("s")
+    sym = mx.sym.contrib.count_sketch(data=data, h=h, s=s, name='countsketch',out_dim = out_dim)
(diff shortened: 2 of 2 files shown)


haojin2:
@anirudh2290 @reminisce Please give a review when you have time, thanks!

src/operator/contrib/count_sketch.cu
```diff
@@ -129,8 +129,7 @@ inline void CountSketchForward(const Tensor<gpu, 2, DType> &out,
                                     nthreads, out_ptr+bstart*out_dim, h_ptr,
                                     s_ptr, in_ptr+bstart*in_dim, batchlen,
                                     in_dim, out_dim);
-    MSHADOW_CUDA_POST_KERNEL_CHECK(sketch_forward_kernel);
-    // cudaThreadSynchronize();
+    CHECK_EQ(cudaDeviceSynchronize(), cudaSuccess) << "Problem occured!";
```
szha: Consider adding more info to the log (e.g. cudaSuccess value).

haojin2: ok

haojin2: Done.

src/operator/contrib/count_sketch.cu
```diff
@@ -129,8 +129,8 @@ inline void CountSketchForward(const Tensor<gpu, 2, DType> &out,
                                     nthreads, out_ptr+bstart*out_dim, h_ptr,
                                     s_ptr, in_ptr+bstart*in_dim, batchlen,
                                     in_dim, out_dim);
-    MSHADOW_CUDA_POST_KERNEL_CHECK(sketch_forward_kernel);
-    // cudaThreadSynchronize();
+    cudaError_t err = cudaDeviceSynchronize();
+    CHECK_EQ(err, cudaSuccess) << "Error occured! CUDA: " << cudaGetErrorString(err);
```
anirudh2290: why do we need to replace this? is it to improve the error message ?

src/operator/contrib/count_sketch.cu
```diff
@@ -165,7 +165,8 @@ inline void CountSketchBackward(const Tensor<gpu, 2, DType> &in_grad,
                             nthreads, in_grad_ptr+bstart*in_dim, h_ptr,
                             s_ptr, out_grad_ptr+bstart*out_dim, batchlen,
                             in_dim, out_dim);
-    MSHADOW_CUDA_POST_KERNEL_CHECK(sketch_backward_kernel);
+    cudaError_t err = cudaDeviceSynchronize();
```
lupesko: DRY - this check repeats in line 132 (and maybe elsewhere?) - did you consider refactoring into a common function?

src/operator/contrib/count_sketch.cu
haojin2: It's hard to explain within a short comment, plus all contexts are in the code.

src/operator/contrib/count_sketch.cu
haojin2: This is the only place where we want to use cudaDeviceSynchronize, if I define an extra function it's also 4 extra lines in total, and does not make any difference.

haojin2: Oh it's actually 7 extra lines where I need 5 lines to define the function and 2 extra calls to the function. Now it's only 4 extra lines in total.

src/operator/contrib/count_sketch.cu
haojin2: @anirudh2290 the original tracking issue is #10988, and I can confirm that I've seen multiple seeds that can reproduce the issue on my dev machine. With this fix it's passing more than 15000 times of tests even on the ones that failed previously.

anirudh2290: you mentioned that not replacing with `cudaDeviceSynchronize` caused certain values in result ndarray to be zeros causing test to fail. The stream synchronize should happen as part of `Wait` irrespective. This not happening seems to be a more serious problem and good to open an issue on this.

haojin2: You're right about that, we can discuss in person tomorrow about this so that we create a suitable issue for this.

src/operator/contrib/count_sketch.cu
lupesko: DRY is not only about lines of code. Consider a case where there is an issue with the flow in line 168, and the developer decides to change it from cudaDeviceSynchronize to another API. Without DRY, the developer will need to remember to look for other places in the code that call the API. Sometimes, this will be missed. With DRY, different parts of the code are funneled to the one function or whatever that has this logic, and by updating in one location all code paths are updated. Makes sense?

larroy:
Is cudaDeviceSynchronize really needed here? This is a very heavy operation that can affect other kernels if I'm not mistaken. A bit more insight on this would be appreciated. Thanks!

haojin2:
@larroy I can get a failing example without the sync to show you.

haojin2:
@larroy I tagged you in the initial issue, please take a look at it there.

larroy:
@haojin2 thanks, I saw it. My question is if we can use a weaker synchronization such as cudaStreamSynchronize(0); that will not block the whole device.
I see how the gradient was corrupted due to launching the kernels serially operating on the same pointers if I understand the code correctly.

https://developer.download.nvidia.com/CUDA/training/StreamsAndConcurrencyWebinar.pdf

haojin2:
@larroy Yes that's correct, thanks for your understanding and advice! I can surely do a test on the StreamSync and see if that would solve the problem. Are you in charge of testing too? Maybe I can also tag you on test-related PRs in the future.

larroy:
Feel free to tag me if I can be useful, I'm pretty fluent in C++. I'm not in charge of testing per se, but I help maintaining CI and its related infrastructure.

PR #11782, opened 2018-07-17
added new users list info; sync info with channels page

aaronmarkham:
## Description ##
Adding user@ mail list info.


aaronmarkham:
@srochel

Issue #11314, opened 2018-06-15
Embedding Backward (AddTakeGradLargeBatchCaller) non-deterministic nan values


leezu:
## Description

The `AddTakeGradLargeBatchCaller` operator called during backward of Embedding is broken and results in `nan` at random positions in the gradient array. 

## Environment info (Required)
- Cuda 9.0 or Cuda 9.2 with respective mxnet-cu92, mxnet-cu90 prebuilt binaries (both v1.2 and nightly affected)
- EC2 p3.2xlarge or p2.xlarge instance

While it is occurs rarely with Cuda 9.0 or p2.xlarge it almost always occurs on Cuda 9.2 and p3.2xlarge. 

## Minimum reproducible example

``` python
import mxnet as mx
import numpy as np

N = 50000
ctx = mx.gpu()

embedding = mx.gluon.nn.Embedding(N, 300)
embedding.initialize(ctx=ctx)
i = 0
np.random.seed(1)
idx = mx.nd.array(np.random.randint(0, N, size=(1024, 160)), ctx=ctx)

got_nan = False
while True:
    i += 1
    with mx.autograd.record():
        emb_in = embedding(idx)
        loss = emb_in.sum()
    loss.backward()

    if not np.all(np.isfinite(embedding.weight.grad().asnumpy())):
        nan_rows = np.where(~np.isfinite(embedding.weight.grad().asnumpy()))[0]
        print(f'Got nan {i}\tRetrying with same data. '
              f'(Affected rows: {nan_rows.tolist()}).')
        got_nan = True
    else:
        if got_nan:  # We got nan before and it disappeared now
            print(f'nan disappeared in {i}..')
            break

    if i % 100 == 0:
        print(f'{i}')

```

## Steps to reproduce

Run above script with cuda 9.2 and observe very frequent nan values:

``` 
% python debug_embedding_nan.py
Got nan 3       Retrying with same data. (Affected indices: [14721, 14721], [1, 2]).
Got nan 4       Retrying with same data. (Affected indices: [20, 20, 39, 39, 18232, 18232], [257, 258, 1, 2, 1, 2]).
Got nan 5       Retrying with same data. (Affected indices: [20, 20, 71, 33346, 38015], [257, 258, 258, 130, 130]).
Got nan 6       Retrying with same data. (Affected indices: [20, 20], [257, 258]).
nan disappeared in 7..
% python debug_embedding_nan.py 
[...]

PR #11795, opened 2018-07-18
Fix problematic backward of take & embedding

haojin2:
## Description ##
Fix for #11314 

### Essentials ###

### Changes ###
- [x] Modify the backward of take and embedding to use the previous AddTakeGrad kernel

## Comments ##
@leezu I've used the script in #11314 and verified this can pass more than 3000 trials.

### DIFF
--- a/src/operator/tensor/indexing_op.h
+++ b/src/operator/tensor/indexing_op.h
@@ -548,46 +548,6 @@ struct tcast_clip {
   }
 };
 
-template<typename xpu, typename IndexType, typename DType>
-void AddTakeGradLargeBatchCaller(const OpContext& ctx, mshadow::Tensor<xpu, 2, DType> dst,
-                                 const mshadow::Tensor<xpu, 1, IndexType>& index,
-                                 const mshadow::Tensor<xpu, 2, DType> &src) {
-  using namespace mxnet_op;
-  using namespace mshadow::expr;
-
-  Stream<xpu> *s = ctx.get_stream<xpu>();
-
-  // Calculate amount of temporary storage
-  size_t sort_workspace_size = mxnet::op::SortByKeyWorkspaceSize<int, int, xpu>
-    (index.shape_.Size());
-  size_t addtake_workspace_size = mxnet::op::AddTakeGradLargeBatchWorkspaceSize<int, xpu>
-    (index.shape_.Size());
-  size_t temp_storage_size = std::max(sort_workspace_size, addtake_workspace_size);
-  size_t workspace_size = 2*(index.shape_.Size()*sizeof(int)) + temp_storage_size;
-
-  // Request temporary storage
-  Tensor<xpu, 1, char> workspace =
-    ctx.requested[embedding::kTempSpace].get_space_typed<xpu, 1, char>(
-      Shape1(workspace_size), s);
-
-  // Create tensors
-  size_t pos = 0;
-  Tensor<xpu, 1, int> sorted_data(reinterpret_cast<int*>(&workspace[pos]),
-    Shape1(index.shape_.Size()), s);
-  pos += index.shape_.Size()*sizeof(int);
-  Tensor<xpu, 1, int> original_index(reinterpret_cast<int*>(&workspace[pos]),
-    Shape1(index.shape_.Size()), s);
-  pos += index.shape_.Size()*sizeof(int);
-  Tensor<xpu, 1, char> temp_storage(&workspace[pos], Shape1(temp_storage_size), s);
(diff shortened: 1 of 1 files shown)


haojin2:
Benchmark script:
```Python
import mxnet as mx
import numpy as np
import time

N = 50000
ctx = mx.gpu(0)

embedding = mx.gluon.nn.Embedding(N, 300)
embedding.initialize(ctx=ctx)
# print(embedding.weight.data().asnumpy())
i = 0
np.random.seed(1)
idx = mx.nd.array(np.random.randint(0, N, size=(1024, 160)), ctx=ctx)
# idx = mx.nd.zeros((1,1), ctx=ctx)

print(np.max(np.bincount(idx.asnumpy().flatten().astype(np.int64))))

a = time.time()
for i in range(500000):
    with mx.autograd.record():
        emb_in = embedding(idx)
        # print(emb_in)
        loss = emb_in.sum()
        # print(loss)
        loss.backward()
print(time.time() - a)
```
Benchmark results:
Original implementation: 146.33599019050598
New Implementation: 147.44622945785522

Slowdown is negligible after switching to new kernel.

leezu:
Thanks @haojin2! Regarding your test, did you run on Tesla V100 GPUs with Cuda 9.2? Also it seems that the AddTakeGradLargeBatch kernel and related code is unused now and should be removed?

haojin2:
@leezu I'm encountering some unit test problems in this and will debug that a bit. The benchmarks were on a single K80 on my dev machine, I'll also do a benchmark on a p3 tmr. I'll run a check on whether this is still used elsewhere or we may even need to bring it back if we observe a severe performance regression on V100, I'll keep it for now and marking this PR as WIP. Will let you know once I've got the latest results. Thanks for your quick reply!

haojin2:
@leezu I found that my fix is only slightly faster than turning MXNET_FORCE_ADDTAKEGRAD on with the previously shown benchmark scripts, I think we're good with the current implementation now?

leezu:
Ok, thanks for looking into this! In that case we should make `MXNET_FORCE_ADDTAKEGRAD` the default behavior and delete the AddTakeGradLargeBatch kernel.

leezu:
@Roshrini I believe this should be addressed before freezing code for MXNet 1.3. The MXNet 1.3 release is likely to be used a lot on P3 instances with Cuda 9.2, meaning that the **buggy backward pass of the Embedding operator** will hit many users.

haojin2:
@leezu I would be preferring the original AddTakeGrad as that required minimal code change.

eric-haibin-lin:
Is this complete? I see some discussions on dev@ about this bug

KellenSunderland:
First of all, great work diving into this @haojin2.

One concern I have is that the overhead of running the imperative python loop would dominate the performance, and make the two implementations seem equivalent in terms of execution time.  Would it be possible to run the same test prefixed with nvprof and then past the kernel execution time summary?

haojin2:
@KellenSunderland One argument that I would like to make here is that end-to-end performance is what matters most to actual users, since that's what they would experience on their end. On the other hand, as the LargeBatch kernel is exhibiting flaky problematic behavior, I don't think it should be kept (I think one other guy on the dev list thread about this is also agreeing with this point), not to mention here it's not even making any fundamental performance difference.
In conclusion, I think the LargeBatch version should be got rid of anyways, but I shall investigate whether the new general take grad kernel or the AddTakeGrad kernel is giving a better performance using some kind of profiler, does that sound good to you?

haojin2:
I'll also do some researches on my side, thanks for bringing this issue up!

KellenSunderland:
Actually looking at a quick profile it seems it should be pretty representative.  Mostly GPU work going on.  Attaching screenshot.
screen shot 2018-07-24 at 5 01 24 pm

haojin2:
@KellenSunderland So I think there should be not much difference between the two versions of the kernel?

KellenSunderland:
Yes, I'd agree based on these measurements.

haojin2:
@leezu The AddTakeGrad kernel can pass more than 3500 trials on the reproduction script, I think this PR should be a good fix without any fundamental performance loss.

haojin2:
@KellenSunderland Thanks for the help on profiling!

KellenSunderland:
Could be a valid point Xingjian.  Are you seeing differences when measured
with nvprof?  Can you post a summary?

sxjscience:
Would you help profile the change? I think the current change should not have passed the code interview.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

sxjscience:
I’ll do that later if you do not have time. Currently, there are lots of other urgent issues for me so I’m asking for help.

Thanks,
Xingjian


Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #11800, opened 2018-07-18
[MXNET-656] Update documentation for count_tokens_from_str

vandanavk:
## Description ##
count_tokens_from_str() throws an error when user tries to use special characters such as * . ? etc as delimiters. This occurs because the function makes use of re.split() which considers the delimiters to be Regular Expressions.

### Essentials ###
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue]

### Changes ###
Changed the documentation to let the the user know about the use of special characters as delimiters, along with an example of the correct usage.


szha:
@astonzhang

PR #11399, opened 2018-06-26
[MXNET-107] Add Fused Vanilla RNN and dropout for CPU

lihaofd:
## Description ##
In this PR, it creates Fused Vanilla RNN(tanh/relu) operator and dropout of GRU/LSTM/vRNN for CPU.
@pengzhao-intel, @TaoLv 

### New features ###
- Single layer/Multiple layer and unidirectional/bidirectional Vanilla RNN(tanh/relu), including both forward and backward computation.
- Support dropout of GRU/LSTM/vRNN

### Unit-test changes ###
- Create new testcase in tests/python/unittests/test_operator.py.
- update testcase in example/rnn/bucketing/cudnn_rnn_bucketing.py
- Check consistency with original RNNCell implementation.

### Performance ###
We have tested performance of FusedRNN and NonFused RNNCell on our local Skylake-8180 with 2 Sockets and 56 cores. Use MKL as blas lib in this performance test.
Test input size is from DS2 default parameters(seq_length = 300, batch_size = 20, input_size = 800, hidden_size = 800).

Layer=1 bidirectional = False

| API | Inference time(fwd, samples/sec)  |  Training time(fwd + bwd, samples/sec)           |
| --------           | :-----:                              | :----:                                   |
| rnn.RNNCell - NoFusedRNN(Tanh, CPU)     | 492.61                       |  198.02                            |
| this PR - FusedRNN(Tanh, CPU)              | 952.38                      |  318.98                             |
| speedup         | 1.93x                              |  1.61x                                    | 

| API | Inference time(fwd, samples/sec)  |  Training time(fwd + bwd, samples/sec)           |
| --------           | :-----:                              | :----:                                   |
| rnn.RNNCell - NoFusedRNN(Relu, CPU)     | 277.78                       |  104.17                            |
| this PR - FusedRNN(Relu, CPU)              | 740.74                      |  177                             |
| speedup         | 2.67x                              |  1.7x                                    | 

Layer=5 bidirectional = True

| API | Inference time(fwd, samples/sec)  |  Training time(fwd + bwd, samples/sec)           |
| --------           | :-----:                              | :----:                                   |
| rnn.RNNCell - NoFusedRNN(Tanh, CPU) | 38.91                       |  22.73                            |
| rnn.RNNCell (Tanh, cuda)  | 47.85                       |  26.95                            |
| rnn.RNNCell (Tanh, cudnn)  | 208.33                       |  81.63                           |
[...]

PR #10311, opened 2018-03-29
[MXNET-107]Fused GRU implementation for CPU

lihaofd:
## Description ##
In this PR, it aligns with #10104's registration way and creates a fused GRU operator for CPU.
@pengzhao-intel, @TaoLv , @Sherry-zhang 

### New features ###
- Single layer/Multiple layer and unidirectional/bidirectional GRU, including both forward and backward computation.
- Support kAddTo/kNullOp check for both GRU and LSTM

### Unit-test changes ###
- Create new testcase in tests/python/unittests/test_operator.py.
- update testcase in example/rnn/bucketing/cudnn_rnn_bucketing.py
- Check consistency with original GRUCell implementation.

### Performance ###
We have tested performance of sym.RNN and rnn.GRUCell on our local Skylake-6148 with 2 Sockets and 40 cores. Use MKL as blas lib in this performance test.
Test input size is from DS2 default parameters(seq_length = 300, batch_size = 20, input_size = 800, hidden_size = 800).

Layer=1 bidirectional = False

| API | Inference time(fwd, sec)  |  Training time(fwd + bwd, sec)           |
| --------           | :-----:                              | :----:                                   |
| rnn.GRUCell  | 0.12                       |  0.277                            |
| this PR            | 0.048                      |  0.106                              |
| speedup         | 2.5x                              |  2.6x                                    | 

Layer=5 bidirectional = True

| API | Inference time(fwd, sec)  |  Training time(fwd + bwd, sec)           |
| --------           | :-----:                              | :----:                                   |
| rnn.GRUCell  | 1.275                       |  2.94                            |
| rnn.GRUCell (cuda)  | 1.047                       |  1.653                            |
| rnn.GRUCell (cudnn)  | 0.161                       |  0.442                           |
| this PR            | 0.551                      |  1.72                              |
| speedup -this PR/GRUCell         | 231.4%                              |  171%                                    | 
| speedup -this PR/GRUCell (cuda)        | 190%                              |  96%                                    | 
| speedup -this PR/GRUCell (cudnn)        | 29.2%                              |  25.7%                                   | 

### Convergency Curve ###
We have tested Convergency of FusedGRU on our CPU-Skylake-8180 with 2 Sockets and 56 cores and GPU-P100  by using example/rnn/bucketing/cudnn_rnn_bucketing.py
Test input size is layer = 3, batch_size = 32, num-embed = 800, num-hidden = 800, num-epochs 20
fusedgru_cpu_gpu

## Checklist ##
[...]

PR #10104, opened 2018-03-14
[WIP][MXNET-107] Fused LSTM implementation for CPU

Sherry-Zhang:
## Description
In this PR, a fused LSTM operator for CPU is implemented. More supports for other RNN variants are WIP and will be submitted in other PRs.

### New features
* Fused LSTM implemention, including both forward and backward computation.
* Share the same frontend interfaces with current `sym.RNN` operator
* Share the same algorithm and input layout with current `sym.RNN` operator
* ~~Refactor code and register it with NNVM interfaces~~
* Support both FP32 and FP64 inputs, more types will be supported later, such as int8
* Provide more extensible APIs for other RNN variants(vanilla RNN/GRU)

## Unit-test changes
* Create new test `test_lstm` and `test_lstm_bidirectional`in `tests/python/unittests/test_operator.py` 
* Check consistency with original LSTMCell implementation

## Performance
We have tested performance of `sym.RNN` and `rnn.LSTMCell` on our local Skylake-8180 with 2 Sockets and 56 cores. Use MKL as blas lib in this performance test.
Test input size is from DS2 default parameters(seq_length = 300, batch_size = 20, input_size = 800, hidden_size = 800).

**single layer measurement**:

| API                  | Inference time(fwd, sec)  |  Training time(fwd + bwd, sec)           |
| --------           | :-----:                              | :----:                                   |
| rnn.LSTMCell  | 0.106902                       |  0.273108                            |
| #9977             | 0.050126                      |    ---                                   |
| this PR            | 0.050668                      |  0.130266                               |
| speedup         | 2.1x                              |  2.1x                                    |

**multi layer measurement**: num_layers=5

| API                  | Inference time(fwd, sec)  |  Training time(fwd + bwd, sec)           |
|  :--------:         | :-----:                              | :----:                                   |
| rnn.LSTMCell | 0.532034                       | 1.546486                          |
| sym.RNN(#9977) | 0.18641                         |    ---                                   |
| sym.RNN(this PR) | 0.190032                     |   0.619439                               |
| rnn.LSTMCell(cuda) | 0.231355                     | 0.785780                         |
| sym.RNN(cudnn)  | 0.060647                     |   0.161115                               |
| speedup  <br> #10104 /LSTMCell        | 285.41%                             |  249.66%                                  |
| speedup  <br> #10104 /LSTMCell(cuda) | 124.09%                          |  126.85%                                  |
[...]

PR #11807, opened 2018-07-18
enable CPU kernel for all RNN layer forward

szha:
## Description ##
After #10104, #10311, and #11399 are merged, rnn operator's CPU implementation is compatible with GPU implementation.

### Essentials ###
- Unit tests already cover it.

### Changes ###
- [x] always use CPU kernel for RNN layer forward.

## Comments ##
- Changing RNN layer to hybrid block is out of scope for this PR and will happen in #11482 instead.


eric-haibin-lin:
@zheng-da does this help your case?

szha:
Plenty with {cpu, gpu} * {ndarray, symbol} coverage.

PR #11814, opened 2018-07-19
[MXAPPS-581] Nightly Straight Dope tests.

vishaalkapoor:
## Description ##
The Straight Dope notebooks will retrieved from the Github repo, run and
scanned for warnings and errors. Because we are not checking accuracy of
the training, we set the number of epochs to 1 to reduce the integration
test run time.
* Common functionality for running and testing notebooks has been
  factored into a common test util module.
* Support for running UTF-8 notebooks added (Python2 and 3 compatible).
* Notebooks requiring a single GPU and multi GPUs have been split
  into two different test suites so that they can be run on different
  hardware.
* Add test to make sure that all notebooks are tested.
* Comment out broken notebooks while they are being fixed (I will
  uncomment them in a follow up PR).


ci/docker/runtime_functions.sh
```diff
@@ -880,6 +880,41 @@ nightly_test_javascript() {
     make -C /work/mxnet/amalgamation libmxnet_predict.js MIN=1 EMCC=/work/deps/emscripten/emcc
 }
 
+# Nightly 'MXNet: The Straight Dope' Single-GPU Tests
+set_up_nightly_straight_dope_tests() {
+    set -ex
+    cd /work/mxnet/tests/nightly/straight_dope
+    rm -rf ./straight_dope_book
+    git clone https://github.com/zackchase/mxnet-the-straight-dope straight_dope_book
```
marcoabreu: Would it be possible to do this as part of the test class init? That way, the test would be self contained and doesn't need any bootstrapping.

vishaalkapoor:
@marcoabreu Thanks for the suggestion. I've moved the notebook download to the installation steps.

ci/docker/Dockerfile.build.ubuntu_nightly_gpu
```diff
@@ -74,6 +74,10 @@ ARG GROUP_ID=0
 COPY install/ubuntu_adduser.sh /work/
 RUN /work/ubuntu_adduser.sh
 
+# This needs to be run after /work/mxnet directory is created by ubuntu_adduser.sh.
```
marcoabreu: Sorry, I mean that this should be done in https://docs.python.org/2/library/unittest.html#unittest.TestCase.setUpClass

If you do it here, you have two problems:
1. The directory won't be valid since the docker build is independent from the runtime
2. The result will be cached and thus we might not be able to detect problems in the straight dope up front

I know that my suggestion is not ideal, but otherwise we got a test don't won't be runnable if executed without proper knowledge about dependencies.

vishaalkapoor:
Hi @marcoabreu,
Thanks for the clarification.

I've updated the test harness to clone the repo and removed the old code. I rebased so the changes are in:
https://github.com/apache/incubator-mxnet/pull/11814/commits/688767f026d53f4b9aa780ec1433472a1f1f0996

Please take a look. Thank you!

Vishaal

tests/nightly/straight_dope/straight_dope_test_utils.py
```diff
@@ -0,0 +1,106 @@
...
+    cmd = [GIT_PATH,
+           'clone',
+           GIT_REPO,
+           NOTEBOOKS_DIR]
+
+    proc = subprocess.Popen(
+        cmd,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT)
+    (out, _) = proc.communicate()
```
marcoabreu: Nice! Is there a way we could add a timeout here? Git tends to not be very nice to us in terms of reliability..

vishaalkapoor:
Thanks Marco, I added a timeout. Will keep tabs on the nightly test job you linked.

tests/nightly/straight_dope/straight_dope_test_utils.py
vishaalkapoor: Added a commit to add a timeout.

marcoabreu:
Do you think it would make sense to enable the log output? Right now, the output is pretty empty and you can't really tell which chapters have run so far.

marcoabreu:
CI passed at http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/test-straight-dope-nightly/detail/test-straight-dope-nightly/3/pipeline

We're good to ship for a first initial version.

vishaalkapoor:
Thanks for your help @marcoabreu Agreed!

Issue #10709, opened 2018-04-26
[SPARSE] undeterministic result of sparse dot(csr, dense, transpose_a=True)


eric-haibin-lin:
The following test failed because of the limited precision of fp32 and the usage of atomicAdd in sparse dot implementation: 
```
    def check_dot_determinism(lhs_stype, rhs_stype, lhs_density, rhs_density, transpose_a, transpose_b):
        lhs_shape = (2, 200)
        rhs_shape = (2, 200)
        lhs = rand_ndarray(lhs_shape, lhs_stype, density=lhs_density)
        rhs = rand_ndarray(rhs_shape, rhs_stype, density=rhs_density)
        res1 = mx.nd.sparse.dot(lhs, rhs, transpose_a=transpose_a, transpose_b=transpose_b)
        res2 = mx.nd.sparse.dot(lhs, rhs, transpose_a=transpose_a, transpose_b=transpose_b)

        assert_almost_equal(res1.asnumpy(), res2.asnumpy(), rtol=0.0, atol=0.0)
    check_dot_determinism('csr', 'default', 0.5, 1.0, True, False)
```

eric-haibin-lin:
dot(csr.T, dns) = rsp is fixed in #10717.

eric-haibin-lin:
we still have the same problem for dot(csr.T, dns) = dns

haojin2:
Working on this.

haojin2:
Fix merged, should be good for closing?

PR #11825, opened 2018-07-19
Fix non-determinism of dot(csr.T, dns) = dns with tests

haojin2:
## Description ##
Fix for #10709  

### Essentials ###

### Changes ###
- [x] Change dot(csr.T, dns) = dns to a deterministic algorithm
- [x] Test for determinism of dot(csr.T, dns) = dns

## Comments ##
New determinism test passed 10000 times on local machine.
Correctness check, which takes longer time to run (~5s/trial), have already passed more than 10000 times at this moment.

### DIFF
--- a/src/common/utils.h
+++ b/src/common/utils.h
@@ -663,6 +663,18 @@ constexpr size_t MaxIntegerValue<mshadow::half::half_t>() {
   return size_t(2) << 10;
 }
 
+MSHADOW_XINLINE int ilog2ul(size_t a) {
+  int k = 1;
+  while (a >>= 1) ++k;
+  return k;
+}
+
+MSHADOW_XINLINE int ilog2ui(unsigned int a) {
+  int k = 1;
+  while (a >>= 1) ++k;
+  return k;
+}
+
 }  // namespace common
 }  // namespace mxnet
 #endif  // MXNET_COMMON_UTILS_H_
--- a/src/operator/tensor/dot-inl.cuh
+++ b/src/operator/tensor/dot-inl.cuh
@@ -35,6 +35,13 @@
 namespace mxnet {
 namespace op {
 
+// Returns integer log2(a) rounded up
+inline int log2i(size_t a) {
+  int k = 1;
+  while (a >>= 1) k++;
+  return k;
+}
+
 /*!
  * \brief GPU scalar kernel of dot(csr, dns1) = dns2
  * Parallelization by output matrix elements: 1 thread/element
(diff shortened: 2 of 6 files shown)


src/operator/tensor/dot-inl.cuh
```diff
@@ -573,12 +581,113 @@ inline void DotCsrDnsDnsImpl(const OpContext& ctx,
                   data_out.dptr<DType>(), data_l.dptr<DType>(), indptr_l.dptr<IType>(),
```
eric-haibin-lin: Since DotCsrTransDnsDnsWarpKernel, DotCsrTransDnsDnsThreadBlockKernel, and DotCsrTransDnsDnsWarpBlockKernel are nondeterministic and cannot be used, we might as well just remove them.

haojin2: Sure I can do that.

haojin2: Done.

src/operator/tensor/dot-inl.cuh
```diff
@@ -539,86 +434,120 @@ inline void DotCsrDnsDnsImpl(const OpContext& ctx,
...
+          // Fill csc_cols with copy of csr_indices
+          mxnet_op::Kernel<mxnet_op::op_with_req<mshadow_op::identity, kWriteTo>, gpu>::Launch(
+            s, nnz, csc_cols_ptr, csr_indices_ptr);
+          // Allocate the tensors needed for SortByKey
+          Tensor<gpu, 1, IType> original_idx(original_idx_ptr, Shape1(nnz), s);
+          Tensor<gpu, 1, IType> csc_cols(csc_cols_ptr, Shape1(nnz), s);
+          Tensor<gpu, 1, char> temp_storage(temp_storage_ptr, Shape1(temp_storage_bytes), s);
+
+          int num_bits = log2i(num_csr_cols - 1);
+          SortByKey(csc_cols, original_idx, true, &temp_storage, 0, num_bits);
```
apeforest: Q: should the argument be num_bits -1 based on the SortByKey function signature?

haojin2: Pay attention to the `num_csr_cols - 1` above.

src/operator/tensor/dot-inl.cuh
```diff
@@ -35,6 +35,13 @@
 namespace mxnet {
 namespace op {
 
+// Returns integer log2(a) rounded up
+inline int log2i(size_t a) {
```
apeforest: I feel we should use common utility function with more efficient implementation for such type of computation.

haojin2: This is only used in this file, plus, pay attention to the mismatch between types of input and output.

apeforest: Although this is only used in this file for now, this function is too generic and very likely to be used in the future by other developers. Making it a generic utility will allow us to change its implementation in the future without breaking existing code. It's output can be cast based on our need. I don't see it as a reason to have a separate function to do that.

src/operator/tensor/dot-inl.cuh
```diff
@@ -35,6 +35,13 @@
...
+  int k = 1;
+  while (a >>= 1) k++;
```
apeforest: nit: use ++k instead

src/operator/tensor/dot-inl.cuh
haojin2: IMO a generic utility should be a templated one to support all data types instead of something like this.

apeforest: I did a simple search and found log2i (or with slightly different names) function appeared at least three times in this module. See: indexing_op.h, pooled_storage_manner.h and intrinsics.cuh. It would be nice to consolidate them.

haojin2: For normal usages where the inputs and outputs have the same type there're std implementations of this. On the other hand this is already a very optimized code to perform such functionality.

eric-haibin-lin: Taking log2 integer does sound like a common utility. We always need to calculate this before calling `SortByKey `. Can we have `log2i(size_t)` and `log2i(unsigned int)` in `common/utils.h` and replace other occurrences?

haojin2:
@eric-haibin-lin

src/common/utils.h
```diff
@@ -663,6 +663,18 @@ constexpr size_t MaxIntegerValue<mshadow::half::half_t>() {
   return size_t(2) << 10;
 }
 
+MSHADOW_XINLINE int ilog2ul(size_t a) {
```
apeforest: Looks good. Please also refactor other places using the same utility as we discussed.

src/operator/tensor/dot-inl.cuh
```diff
@@ -510,6 +398,7 @@ inline void DotCsrDnsDnsImpl(const OpContext& ctx,
     return;
   }
 
+  using namespace mshadow;
   using mshadow::cuda::kBaseThreadNum;
```
apeforest: You may not need the mshadow directive since you added the line above.

haojin2: not used now, deleted.

src/common/utils.h
haojin2: Done.

apeforest:
LGTM and thanks for the refactoring.

haojin2:
Benchmark script:
```Python
import mxnet as mx
import sys
import os
import scipy
import numpy as np
from mxnet.test_utils import assert_almost_equal
import time

def measure_cost(repeat, f, *args, **kwargs):
    # start bench
    start = time.time()
    results = []
    for i in range(repeat):
        results.append(f(*args, **kwargs))
    for result in results:
        result.wait_to_read()
    end = time.time()
    diff = end - start
    return diff / repeat

def measure_fallback(repeat, a):
    # start bench
    start = time.time()
    results = []
    for i in range(repeat):
        results.append(a.tostype('default'))
    for result in results:
        result.wait_to_read()
    end = time.time()
    diff = end - start
    return diff / repeat

def main():
    shape_lhs = (35000, 256)
    shape_rhs = (35000, 35000)
    dns = np.random.uniform(size=shape_lhs)
    mx_dns = mx.nd.array(dns, ctx=mx.gpu())
    mx_dns_cpy = mx_dns.copy()
    for density in [0.01, 0.005, 0.001, 0.0005, 0.0001]:
        csr = scipy.sparse.random(shape_lhs[0], shape_lhs[1], density=density, format = 'csr', dtype=np.float32)
        mx_csr = mx.nd.sparse.csr_matrix((csr.data, csr.indices, csr.indptr), shape=shape_rhs, ctx=mx.gpu())
        mx_csr_dns = mx_csr.tostype('default')
        sparse_cost = 0.0
        dns_cost = 0.0
        mx.nd.waitall()
        #warmup
        check = mx.nd.dot(mx_csr, mx_dns, transpose_a=True, forward_stype='default')
        check_np = np.dot(mx_csr_dns.asnumpy().T, dns)
        assert_almost_equal(check.asnumpy(), check_np, atol=1e-5, rtol=1e-4)
        print(check.shape)
        mx.nd.waitall()
        for i in range(50):
            sparse_cost += measure_cost(1, mx.nd.dot, mx_csr, mx_dns, transpose_a=True, forward_stype='default')
            dns_cost += measure_cost(1, mx.nd.dot, mx_csr_dns, mx_dns, transpose_a=True)
        print("%.2f %%" % (density*100), dns_cost / sparse_cost)


if __name__ == "__main__":
    main()
```
Benchmark results on single K80 GPU (all speedup values are based on comparison to the same dense dot computation):
(density of CSR matrix - speedup before the change / speedup after the change)
1.00% - 10.641220445769855 / 77.02531563623299
0.50% - 11.99406732741644 / 89.92680215941569
0.10% - 16.309222808496383 / 102.91791288433697
0.05% - 17.67602831964807 / 103.2725946343818
0.01% - 19.016961107687557 / 114.2781844559389
@eric-haibin-lin

haojin2:
Comparison of performance between before and after on smaller workloads (dot( (100k, 128).T , (100k, 512) )):
(density of csr - speedup compared to dense version before the change / speedup compared to dense version after the change)
1.00 % - 0.11684326423581441 / 3.7605456307819147
0.50 % - 0.13864958837387978 / 4.902107295777013
0.10 % - 0.18179564999182662 / 7.328729395201502
0.05 % - 0.19055669842279166 / 7.820138355111453
0.01 % - 0.20276938703209807 / 8.232172552569692

haojin2:
@eric-haibin-lin File conflicts resolved, should be ready for merge once the build passes.

eric-haibin-lin:
Great fix!

Issue #11620, opened 2018-07-09
LBSGD does not have momentum


al-rigazzi:
https://github.com/apache/incubator-mxnet/blob/8fde4b7526acbbe242fd5a5f41fac9e477f927f3/example/image-classification/common/fit.py#L219

Hi, I was wondering why LBSGD is not listed as a method with momentum, since it looks like the class offers the possibility of setting it.

apeforest:
For questions, please submit on MXNet discussion forum (https://discuss.mxnet.io), where it will get a wider audience and allow other to learn as well.
 I'm closing this issue now in favor of the discussion forum issue you will file, please feel free to re-open if closed in error.
 Thanks!

apeforest:
Thanks for posting this question on the forum. @sandeep-krishnamurthy Please close this issue.

thomelane:
@sandeep-krishnamurthy @apeforest can we reopen and tag, looks like this should actually be updated. Signum and LBSGD both have momentum but aren't included here.

apeforest:
@sandeep-krishnamurthy Please label this issue _Example_, _Good for first time_

vishaalkapoor:
@thomelane I've submitted a pull request that adds signum and lbsgd to the list of optimizers with momentum. Please take a look: [link](https://github.com/apache/incubator-mxnet/pull/11826)

PR #11826, opened 2018-07-19
Adding missing optimizers that use momentum.

vishaalkapoor:
## Description ##
This addresses an omission in the image classification examples where signum and lbsgd are not listed as optimizers having momentum. Resolves issue [11620](https://github.com/apache/incubator-mxnet/issues/11620).


vkapoor3141:
Many thanks, @szha  :)

PR #11833, opened 2018-07-20
[MXNET-688] Fix quantization divide by zero errors

OneRaynyDay:
## Description ##
The current quantization strategy for `calib_mode='entropy'` is to calculate the KL divergence for different thresholds and choose the best threshold. This assumes that the random variable is nonzero for all reals and is a continuous random variable. Because we are discretizing the distribution, we smooth the distribution over the range `[-threshold, threshold]`. What we are not considering is that the entire sampled distribution may be not in the range `[-threshold, threshold]` and thus we end up with all zeros in the sampled candidate `p` distribution inside of `_get_optimal_threshold`.

I have added a check that the distribution(possibly unnormalized) is proper before attempting to smooth or else we'll run into a divide by 0 error.

In most cases, activation functions and layers for classification type problems output numbers symmetric around 0. This is not the case for a regressor's last layer, and there are various other examples where the activation distribution is not around 0, and this was a major blockage for airbnb's adoption into mxnet's quantization capabilities.

### Essentials ###

### Changes ###
- Added tests and extra code inside of quantization.py to set all kl divergence to infinity if the probability distribution formed is all zeros. Also there are some typos and one-off-errors in the original implementation that are now fixed.

### DIFF
--- a/python/mxnet/contrib/quantization.py
+++ b/python/mxnet/contrib/quantization.py
@@ -329,7 +329,7 @@ def _get_optimal_threshold(arr, num_bins=8001, num_quantized_bins=255):
             if norm != 0:
                 q[start:stop] = float(quantized_bins[j]) / float(num_quantized_bins)
         p = _smooth_distribution(p)
-        # There is a chance that q is a completely zero'd probability distribution.
+        # There is a chance that q is an invalid probability distribution.
         try:
             q = _smooth_distribution(q)
         except ValueError:
--- a/python/mxnet/contrib/quantization.py
+++ b/python/mxnet/contrib/quantization.py
@@ -231,110 +231,7 @@ def _collect_layer_outputs(mod, data, include_layer=None, max_num_examples=None,
     return collector.nd_dict, num_examples
 
 
-def _smooth_distribution(p, eps=0.0001):
-    """Given a discrete distribution (may have not been normalized to 1),
-    smooth it by replacing zeros with eps multiplied by a scaling factor and taking the
-    corresponding amount off the non-zero values.
-    Ref: http://web.engr.illinois.edu/~hanj/cs412/bk3/KL-divergence.pdf
-    """
-    is_zeros = (p == 0).astype(np.float32)
-    is_nonzeros = (p != 0).astype(np.float32)
-    n_zeros = is_zeros.sum()
-    n_nonzeros = p.size - n_zeros
-    eps1 = eps * float(n_zeros) / float(n_nonzeros)
-    assert eps1 < 1.0, 'n_zeros=%d, n_nonzeros=%d, eps1=%f' % (n_zeros, n_nonzeros, eps1)
-    hist = p.astype(np.float32)
-    hist += eps * is_zeros + (-eps1) * is_nonzeros
-    assert (hist <= 0).sum() == 0
-    return hist
 
-
-# pylint: disable=line-too-long
-def _get_optimal_threshold(arr, num_bins=8001, num_quantized_bins=255):
-    """Given a dataset, find the optimal threshold for quantizing it.
-    Ref: http://on-demand.gputechconf.com/gtc/2017/presentation/s7310-8-bit-inference-with-tensorrt.pdf
-    """
-    if isinstance(arr, NDArray):
-        arr = arr.asnumpy()
-    elif isinstance(arr, list):
-        assert len(arr) != 0
-        for i, nd in enumerate(arr):
-            if isinstance(nd, NDArray):
-                arr[i] = nd.asnumpy()
-            elif not isinstance(nd, np.ndarray):
-                raise TypeError('get_optimal_threshold only supports input type of NDArray,'
-                                ' list of np.ndarrays or NDArrays, and np.ndarray,'
-                                ' while received type=%s' % (str(type(nd))))
-        arr = np.concatenate(arr)
-    elif not isinstance(arr, np.ndarray):
-        raise TypeError('get_optimal_threshold only supports input type of NDArray,'
-                        ' list of NDArrays and np.ndarray,'
-                        ' while received type=%s' % (str(type(arr))))
-    min_val = np.min(arr)
-    max_val = np.max(arr)
-    th = max(abs(min_val), abs(max_val))
-
-    hist, hist_edeges = np.histogram(arr, bins=num_bins, range=(-th, th))
-    zero_bin_idx = num_bins // 2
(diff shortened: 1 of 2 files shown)


python/mxnet/contrib/quantization.py
```diff
@@ -303,32 +305,34 @@ def _get_optimal_threshold(arr, num_bins=8001, num_quantized_bins=255):
...
             quantized_bins[j] = sliced_nd_hist[start:stop].sum()
         quantized_bins[-1] += sliced_nd_hist[num_quantized_bins * num_merged_bins:].sum()
         # expand quantized_bins into p.size bins
-        q = np.zeros(p.size, dtype=np.float32)
+        q = np.zeros(sliced_nd_hist.size, dtype=np.float32)
         for j in range(num_quantized_bins):
             start = j * num_merged_bins
             if j == num_quantized_bins - 1:
-                stop = -1
+                stop = len(is_nonzeros)
```
OneRaynyDay: This is an off-by-1 error that can be caught via the quantization tests that I added. Indexing a numpy array with `x[a:-1]` excludes the last element.

python/mxnet/contrib/quantization.py
```diff
@@ -303,32 +305,34 @@ def _get_optimal_threshold(arr, num_bins=8001, num_quantized_bins=255):
...
             else:
                 stop = start + num_merged_bins
             norm = is_nonzeros[start:stop].sum()
             if norm != 0:
-                q[start:stop] = float(quantized_bins[j]) / float(norm)
-        q[sliced_nd_hist == 0] = 0
```
OneRaynyDay: This is not representative of the quantized distribution, as setting values to 0 artificially will not correctly represent the quantized activation output.

python/mxnet/contrib/quantization.py
```diff
@@ -303,32 +305,34 @@ def _get_optimal_threshold(arr, num_bins=8001, num_quantized_bins=255):
...
+                q[start:stop] = float(quantized_bins[j]) / float(num_quantized_bins)
```
OneRaynyDay: Originally this was `float(norm)`, and that is not appropriate. Suppose you have the distribution:

```
[0, 0, ... , 0, 1]
```

If the `num_quantized_bins` is `3`, then you theoretically should get:

```
[0, 0, ... , 1/3, 1/3, 1/3]
```

instead of:

```
[0, 0, ..., 1, 1, 1]
```

To make this more clear, suppose your original dist is:

```
[0, 0, 0, ... , 1/3, 1/3, 1/3]
```

This should be equivalent after quantization as the first distribution, but it isn't. Under the rules, you would get the same array back, where you have `[..., 1/3, 1/3 ,1/3]`, and the first distribution would give you `[..., 1, 1, 1]`, off by the multiplier.

python/mxnet/contrib/quantization.py
```diff
@@ -274,22 +279,21 @@ def _get_optimal_threshold(arr, num_bins=8001, num_quantized_bins=255):
     max_val = np.max(arr)
     th = max(abs(min_val), abs(max_val))
 
-    hist, hist_edeges = np.histogram(arr, bins=num_bins, range=(-th, th))
+    hist, hist_edges = np.histogram(arr, bins=num_bins, range=(-th, th))
```
OneRaynyDay: `edges` is mispelled as `edeges` throughout the code

tests/python/quantization/test_quantization.py
```diff
@@ -485,6 +486,27 @@ def test_quantize_sym_with_calib():
...
+    assert_almost_equal(mx.contrib.quant._smooth_distribution(dirac_delta, eps=1e-3), smooth_dirac_delta)
+
+
+@with_seed()
+def test_optimal_threshold_adversarial_case():
+    # The worst case for the optimal_threshold function is when the values are concentrated
+    # at one edge: [0, 0, ..., 1000]. (histogram)
+    # We want to make sure that the optimal threshold in this case is the max.
+    arr = np.array([2]*1000)
+    res = mx.contrib.quant._get_optimal_threshold(arr, num_quantized_bins=5)
```
OneRaynyDay: Using the incorrectly implemented code for `_get_optimal_threshold`, we would result in a divide by 0 error here.

python/mxnet/contrib/quantization.py
```diff
@@ -303,32 +307,34 @@ def _get_optimal_threshold(arr, num_bins=8001, num_quantized_bins=255):
...
-                q[start:stop] = float(quantized_bins[j]) / float(norm)
-        q[sliced_nd_hist == 0] = 0
+                q[start:stop] = float(quantized_bins[j]) / float(num_quantized_bins)
         p = _smooth_distribution(p)
-        q = _smooth_distribution(q)
+        # There is a chance that q is a completely zero'd probability distribution.
+        try:
+            q = _smooth_distribution(q)
+        except ValueError:
+            divergence[i - num_half_quantized_bins] = float("inf")
```
OneRaynyDay: If the distribution is improper, we set the KL divergence to infinity, as it could theoretically model a uniform distribution of parameters `[a,b]` with either variables unbounded, which means KL divergence is infinity.

OneRaynyDay:
For reference:

With new PR, on imagenet, resnet 152, 5 batch, entropy method:

```
INFO:logger:Finished inference with 16000 images
INFO:logger:Finished with 178.458244 images per second
INFO:logger:('accuracy', 0.753)
INFO:logger:('top_k_accuracy_5', 0.922625)
```

Previously:

```
# TODO
```

python/mxnet/contrib/quantization.py
```diff
@@ -303,32 +307,35 @@ def _get_optimal_threshold(arr, num_bins=8001, num_quantized_bins=255):
...
             if j == num_quantized_bins - 1:
-                stop = -1
+                stop = len(is_nonzeros)
             else:
                 stop = start + num_merged_bins
             norm = is_nonzeros[start:stop].sum()
             if norm != 0:
                 q[start:stop] = float(quantized_bins[j]) / float(norm)
-        q[sliced_nd_hist == 0] = 0
+        q[p == 0] = 0
```
OneRaynyDay: According to the slides on page 38: http://on-demand.gputechconf.com/gtc/2017/presentation/s7310-8-bit-inference-with-tensorrt.pdf, the zero'd out bins are meant to be w.r.t reference distribution `p`, rather than the `sliced_nd_hist`.

PR #11834, opened 2018-07-20
Fix mxnet ctc_loss bug

HawkAaron:
1. fix ctc_loss GPU bug
~2. add blank label for gluon CTCLoss~
edit @szha: crossed out second item

### DIFF
--- a/python/mxnet/gluon/loss.py
+++ b/python/mxnet/gluon/loss.py
@@ -409,6 +409,8 @@ class CTCLoss(Loss):
         length respectively.
     weight : float or None
         Global scalar weight for loss.
+    blank_label : {'first', 'last'}, default 'last'
+        Set the label that is reserved for blank label.
 
 
     Inputs:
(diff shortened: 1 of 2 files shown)


python/mxnet/gluon/loss.py
```diff
@@ -409,6 +409,8 @@ class CTCLoss(Loss):
         length respectively.
     weight : float or None
         Global scalar weight for loss.
+    blank_label : {'first', 'last'}, default 'last'
```
szha: it was intentional not to expose this option in gluon.

chinakook:
Is the speed of big vocab size solved of this op?

python/mxnet/gluon/loss.py
HawkAaron: That means I need to revert this commit and  resend a pull request ?

HawkAaron:
@chinakook its looks good to me

python/mxnet/gluon/loss.py
szha: you can simply add a commit that removes the change in blank_label

szha:
@Jerryzcn could you give this a try?

Jerryzcn:
will do

Jerryzcn:
@HawkAaron it seems removing the max negatively affect convergence. Do you observe similar result?

src/operator/contrib/ctc_include/detail/gpu_ctc.h
```diff
@@ -411,12 +411,7 @@ GpuCTC<ProbT>::compute_log_probs(const ProbT* const activations) {
         denoms_, out_dim_, num_elements);
 
     // compute denominators for softmax
-    denoms_handle = reduce_with_axis<red::sum, false>(
-        F<mxnet::op::mshadow_op::exp>(
-            log_probs_handle -
-            broadcast<0>(reduce_with_axis<red::maximum, false>(log_probs_handle, 1),
```
Jerryzcn: max is necessary here.

HawkAaron: the max is reduced in here: https://github.com/apache/incubator-mxnet/blob/master/src/operator/contrib/ctc_include/detail/gpu_ctc.h#L398

and `log_probs -= denoms` is in here: https://github.com/apache/incubator-mxnet/blob/master/src/operator/contrib/ctc_include/detail/gpu_ctc.h#L409

that means the broadcast line (https://github.com/apache/incubator-mxnet/blob/master/src/operator/contrib/ctc_include/detail/gpu_ctc.h#L417) should be zero

Jerryzcn: I see. Did not look at other parts of the code thanks! Let me check why my model is not converging

szha:
@HawkAaron thanks for the fix, and @Jerryzcn thanks for the review

Issue #11775, opened 2018-07-17
NPE in BucketingModule when providedLabel of DataBatch is not set


milandesai:
Package: Scala
Mxnet Version: 1.1.0
Java Version: openjdk 1.8.0_131
Maven Version: [unused]
Scala Runtime: 2.12.4

We create a DataBatch without setting the `providedLabel` field. We then call `BucketingModule.forward(...)` with the data batch and get a NullPointerException:
```
java.lang.NullPointerException was thrown.
java.lang.NullPointerException
	at ml.dmlc.mxnet.DataDesc$.ListMap2Descs(IO.scala:252)
	at ml.dmlc.mxnet.module.BucketingModule.forward(BucketingModule.scala:313)
```

This happens because the implicit method `ListMap2Descs` tries to convert `ListMap[String, Shape]` into `IndexedSeq[DataDesc]` but fails because the former is null by default.

lupesko:
Thanks for reporting the issue @milandesai 

To help the community look into it, please provide more information as asked below (usually the new issue template asks to fill all of that in) - providing complete information in the most concise form is the best way to get help. 
- Package used (Python/R/Scala/Julia):
- For Scala user, please provide:
1. Java version: (`java -version`)
2. Maven version: (`mvn -version`)
3. Scala runtime if applicable: (`scala -version`)
- Build info (Required if built from source)
1. Compiler (gcc/clang/mingw/visual studio):
2. MXNet commit hash (Paste the output of `git rev-parse HEAD` here.)
3. Build config: (Paste the content of config.mk, or the build command.)
- Minimum reproducible example
(If you are using your own code, please provide a short script that reproduces the error. Otherwise, please provide link to the existing example.)

Thanks!

lupesko:
@nswamy please label as "Scala" "Bug"
@lanking520 @andrewfayres @ddavydenko who are working on the Scala API

lanking520:
Adding DataDesc in this PR: https://github.com/apache/incubator-mxnet/pull/11844

Issue #11571, opened 2018-07-05
 RNN Example not supported for MXNet Scala


lanking520:
This model has not been updated since 2017/07, the RNN example for LstmBucketing is no longer runnable. 
Please contribute to RNN support on Scala for the current version. If you would like to run this example for now, please revert your MXNet version to `v0.12`. 

In summary, I suspect that some operator changes caused this issue.

- [ ] LstmBucketing
- [ ] TestCharRnn
- [x] TrainCharRNN

Please contact @lanking520 or @nswamy if you would like to dig in and fix this problem. The following is a way to reproduce the issues I found:

### Setup
Download the required ptb file from here:
```Bash
wget https://raw.githubusercontent.com/dmlc/web-data/master/mxnet/ptb/ptb.train.txt
wget https://raw.githubusercontent.com/dmlc/web-data/master/mxnet/ptb/ptb.valid.txt
```
You can use this [script](https://github.com/apache/incubator-mxnet/blob/master/scala-package/examples/scripts/rnn/run_lstm_bucketing.sh) to run the model 
or manually identify the file path by following [this](https://github.com/apache/incubator-mxnet/blob/master/scala-package/examples/src/main/scala/org/apache/mxnetexamples/rnn/LstmBucketing.scala#L36-L48)

#### Additional step to get it work
Please add the follows to [IO.scala](https://github.com/apache/incubator-mxnet/blob/master/scala-package/core/src/main/scala/org/apache/mxnet/IO.scala#L358)
```Scala
if (shapes.toIndexedSeq(0)._2.length == 2) {
   shapes.map { case (k, s) => new DataDesc(k, s, layout = "NT") }.toIndexedSeq
}
else shapes.map { case (k, s) => new DataDesc(k, s) }.toIndexedSeq
```

#### Segmentation fault caused by C++ backend
```Bash
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x000000012d026fd8, pid=56081, tid=0x0000000000000e03
#
# JRE version: Java(TM) SE Runtime Environment (8.0_171-b11) (build 1.8.0_171-b11)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.171-b11 mixed mode bsd-amd64 compressed oops)
# Problematic frame:
# C  [mxnet-scala+0x120afd8]
[error occurred during error reporting (printing problematic frame), id 0xe0000000]

[...]

Issue #10753, opened 2018-04-30
Scala Module MNIST MLP example not working


gigasquid:
## Description
Running the Scala example for Module MNIST is not working due to the new requirement in DataDesc from this PR https://github.com/apache/incubator-mxnet/commit/b95ae7c9c8c3cf40a12cfc729d5dadaa322ee0dd#diff-196b8fe8b507164f1c13b556f178b7b2R233

## Environment info (Required)
Mac OSX
Java(TM) SE Runtime Environment (build 1.8.0_152-b16)

MXNet commit hash: 4044eb46f7e2a29c7236656887baf1c44a882d5b

Build config:
USE_BLAS = openblas
ADD_CFLAGS += -I/usr/local/opt/openblas/include
ADD_LDFLAGS += -L/usr/local/opt/openblas/lib
ADD_LDFLAGS += -L/usr/local/lib/graphviz/

## Error Message:
```
15:05 $ ./mnist_mlp.sh 
[main] INFO MXNetJVM - Try loading mxnet-scala from native path.
[main] INFO MXNetJVM - Try loading mxnet-scala-osx-x86_64-gpu from native path.
[main] INFO MXNetJVM - Try loading mxnet-scala-osx-x86_64-cpu from native path.
[main] WARN MXNetJVM - MXNet Scala native library not found in path. Copying native library from the archive. Consider installing the library somewhere in the path (for Windows: PATH, for Linux: LD_LIBRARY_PATH), or specifying by Java cmd option -Djava.library.path=[lib path].
[main] INFO org.apache.mxnet.util.NativeLibraryLoader - Replaced .dylib with .jnilib
[main] INFO org.apache.mxnet.util.NativeLibraryLoader - Loading libmxnet-scala.jnilib from /lib/native/ copying to mxnet-scala
[15:05:57] src/io/iter_mnist.cc:110: MNISTIter: load 60000 images, shuffle=1, shape=(10,784)
[15:05:58] src/io/iter_mnist.cc:110: MNISTIter: load 10000 images, shuffle=1, shape=(10,784)
[main] INFO org.apache.mxnetexamples.module.MnistMlp - Run intermediate level api from beginning.
[main] ERROR org.apache.mxnetexamples.module.MnistMlp - requirement failed: number of dimensions in shape :2 with shape: (10,784) should match the length of the layout: 4 with layout: NCHW
[...]

PR #11844, opened 2018-07-20
[MXNET-689] add DataDesc type for the Scala Package

lanking520:
## Description ##
This PR includes massive changes on all Iterators that extends `DataIter`, two new fields are added (`dtype` and `layout`).

@nswamy @yzhliu @andrewfayres 

It intends to solve some issues about the layout information missing in DataDesc. For example: RNN based networks require different layouts(“TNC”/”NTC”). Current implementation of DataBatch, DataIter does not support different layouts and assumes a fixed layout based on the shape of the NDArray returned in the next() call to the iterator.

However I found it hard to have a general solution that both address backward compatibility as well as adding this information. Currently, there are several ways to address this issue:

1. Assume Shape == 2 follows `NT` format, Shape == 3 follows `TNC`, Shape == 4 follows `NCHW` and only change the implicit function in [here](https://github.com/apache/incubator-mxnet/blob/master/scala-package/core/src/main/scala/org/apache/mxnet/IO.scala#L358-L363).

2. Add `Layout` and `DType` defs for all iterator, then users can pass these information to create `DataDesc` instead of `ListMap[Name, Shape]`. However, it breaks the input args for `NDArrayIter` as well as other iterators. Users are required to add more inputs. However, if the iterator are passed in with dynamic shapes (shapes are not determined), it will break the DataDesc checking scheme. 

3. Do what 2 contains and create `this` constructor to make sure BC.

## Test
- [x] IOSuite
- [ ] ModuleSuite
- [x] DataIterSuite

## Historical issues
- https://github.com/apache/incubator-mxnet/issues/11775
- [MLP problem](https://github.com/apache/incubator-mxnet/issues/10753)
- [RNN problem](https://github.com/apache/incubator-mxnet/issues/11571)

### Essentials ###


lanking520:
When I am in the middle of changes, some parts looks confusing to me. Are we defining these field in the `dataiter` class or we fetch these information through `databatch`. There are different solution in different DataIters' child classes. The currently issue that failing the test is actually here:
```Scala
  /**
   * DataIter creator
   * @param handle native memory ptr for the iterator
   * @param params parameter passed to the iterator
   * @return created DataIter
   */
  private def creator(handle: DataIterCreator)(
    params: Map[String, String]): DataIter = {
    val out = new DataIterHandleRef
    val keys = params.keys.toArray
    val vals = params.values.toArray
    checkCall(_LIB.mxDataIterCreateIter(handle, keys, vals, out))
    val dataName = params.getOrElse("data_name", "data")
    val labelName = params.getOrElse("label_name", "label")
    new MXDataIter(out.value, dataName, labelName)
  }
```
This creator did not pass in the layout and dtype information and MXDataIter extends DataIter. By default, the layout are set a "NCHW" that causing the problem.

scala-package/core/src/main/scala/org/apache/mxnet/io/MXDataIter.scala
```diff
@@ -151,12 +175,34 @@ private[mxnet] class MXDataIter(private[mxnet] val handle: DataIterHandle,
     out.value
   }
 
+  /**
+    * Get the DType
+    * @return DType
+    */
+  def getDType(): DType = {
```
nswamy: suggest to add two new parameters to the constructor `dataDescriptor` and `labelDescriptor`.
Make sure not break API backwards compatibility

lanking520:
After some days spent on the issues, I finally figure out the fix on the issues we have. @nswamy and @yzhliu thanks for your comments on the changes and things I am about to apply is:

- Depreciate `provideData` and `provideLabel` methods
- disable implicit transformation in IO to use `DataDesc` everywhere
@andrewfayres please join the review if you have time :-)
Please provide a secondary review before I apply these changes. There will be a final review session once changes are applied and everything I felt is settled.

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
```diff
@@ -59,10 +61,12 @@ class NDArrayIter(data: IndexedSeq[(String, NDArray)],
   def this(data: IndexedSeq[NDArray], label: IndexedSeq[NDArray] = IndexedSeq.empty,
            dataBatchSize: Int = 1, shuffle: Boolean = false,
            lastBatchHandle: String = "pad",
-           dataName: String = "data", labelName: String = "label") {
+           dataName: String = "data", labelName: String = "label",
+           dType: DType = MX_REAL_TYPE, dataLayout: String = "NCHW",
```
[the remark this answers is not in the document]
nswamy: create another this constructor with the old set of constructor args to not break backwards compatibility.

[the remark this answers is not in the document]
nswamy: its ok to keep here. let all the defaults remain in one place.

scala-package/core/src/main/scala/org/apache/mxnet/IO.scala
```diff
@@ -170,6 +180,10 @@ object DataBatch {
     private var label: IndexedSeq[NDArray] = null
     private var index: IndexedSeq[Long] = null
     private var pad: Int = 0
+    private var dataLayout: String = "NCHW"
```
nswamy: For layouts I suggest to use extensible Enums similar to this https://blogs.oracle.com/darcy/mixing-in-an-enum.  
free floating strings are generally not a good practice because it removes compile time type-checks. 
WDYT? does it make sense?

lanking520:
Hi @benkamphaus, could you please take a look at the recent CI failure on Clojure. I believe these lines cause that failure: https://github.com/apache/incubator-mxnet/blob/master/contrib/clojure-package/src/org/apache/clojure_mxnet/io.clj#L194-L225. I have modified the input args for `NDArrayIter` and there will be more optional fields

nswamy:
@benkamphaus, from Clojure perspective are we breaking backwards compatibility. I am curious to know how it is handled, since we added new parameters with defaults.

benkamphaus:
It appears to be the case that the new constructor arglist for `NDArrayIter` breaks backwards compatibility, which is the cause of the current Clojure CI broken message:

```No matching ctor found for class org.apache.mxnet.io.NDArrayIter, compiling:(org/apache/clojure_mxnet/io.clj:216:4)```

The cause is straight forward -- the constructor call [here](https://github.com/lanking520/incubator-mxnet/blob/dataiter/contrib/clojure-package/src/org/apache/clojure_mxnet/io.clj#L216-L223) is no longer valid due to the change to the args required for `NDArrayIter` as [declared here](https://github.com/apache/incubator-mxnet/pull/11844/files#diff-b63b7056575260dd87ef51b1bf01fc9cR44).

This constructor call on the clojure side backs `ndarray-iter`, and there are tests, specs, and examples and the changes in required args to `ndarray-iter` would need those to change as well. (i.e. this impacts the API/contract level, not just an implementation concern below those).

Note that the keyword args in the Clojure API means that `ndarray-iter` could back a call to both the previous NDArrayIter constructor and the form with the new args (using additional, optional args to the same function in the Clojure API). Otherwise, I'm not sure what the expectation/desire is here without more context.

I'll discuss with @gigasquid as well.

**Comment by lanking520** [comment]
@benkamphaus thanks for your detailed analysis. I can tell how hard it is if we change the args in here since I also changed a lot of examples to match this change in Scala. Let's be cautious on this change and discuss if it is necessary here to introduce more parameters.

**Comment by lanking520** [comment]
I have reverted the changes back to Strings. Next step is to add a default value to Layout section so we can bypass the require check

**Comment by lanking520** [comment]
In the recent two PR, add Backward Compatibility fix on DataIter as well as DataBatch. User can still go with the old defs...

File: scala-package/core/src/main/scala/org/apache/mxnet/IO.scala [hunk 4]
```diff
@@ -280,7 +333,10 @@ abstract class DataIter extends Iterator[DataBatch] {
    */
   @throws(classOf[NoSuchElementException])
   def next(): DataBatch = {
-    new DataBatch(getData(), getLabel(), getIndex(), getPad())
+    new DataBatch(getData(), getLabel(), getIndex(), getPad(),
+      null, null, null,
```
nswamy: please make these parameters as Options and don't pass nulls.

lanking520: This will break bcc I think, not sure about the influence

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
```diff
@@ -42,27 +43,40 @@ import scala.collection.immutable.ListMap
...
                   label: IndexedSeq[(String, NDArray)],
                   private val dataBatchSize: Int, shuffle: Boolean,
-                  lastBatchHandle: String) extends DataIter {
-
+                  lastBatchHandle: String,
+                  dataDType: DType, labelDType: DType,
+                  dataLayout: String, labelLayout: String) extends DataIter {
+
+  // scalastyle:off
+  def this(data: IndexedSeq[NDArray], label: IndexedSeq[NDArray],
```
nswamy: you need to create constructors(with original parameters) to maintain backward compatibility

scala-package/core/src/test/scala/org/apache/mxnet/IOSuite.scala
```diff
@@ -99,7 +101,9 @@ class IOSuite extends FunSuite with BeforeAndAfterAll {
       "data_shape" -> "(3,28,28)",
       "batch_size" -> "100",
       "preprocess_threads" -> "4",
-      "prefetch_buffer" -> "1"
+      "prefetch_buffer" -> "1",
```
nswamy: create new tests that take new parameters, so we know that the existing APIs don't break.

scala-package/core/src/main/scala/org/apache/mxnet/DType.scala
```diff
@@ -35,4 +35,13 @@ object DType extends Enumeration {
       case DType.Unknown => 0
     }
   }
+  private[mxnet] def getType(dtypeStr: String): DType = {
+    dtypeStr match {
+      case "UInt8" => DType.UInt8
+      case "Int32" => DType.Int32
+      case "Float16" => DType.Float16
+      case "Float32" => DType.Float32
+      case "Float64" => DType.Float64
```
yzhliu: otherwise throw IllegalArgumentException

scala-package/core/src/main/scala/org/apache/mxnet/IO.scala
```diff
@@ -216,6 +244,30 @@ object DataBatch {
       this
     }
 
+    /**
+      * Set the dtype.
+      * @param dataDType The dtype of the data, default is Float32
+      * @param labelDType The dtype of the label, default is Int32
+      * @return this
+      */
+    def setDType(dataDType: DType, labelDType: DType): Builder = {
```
yzhliu: suggest to have `setDataDType` and `setLabelDType` separately, which will be helpful for Java users who want to set dtype for only one of them.

scala-package/core/src/main/scala/org/apache/mxnet/IO.scala
```diff
@@ -216,6 +244,30 @@ object DataBatch {
...
+      this
+    }
+
+    /**
+      * Set the layout.
+      * @param dataLayout The layout of the data, default is NCHW
+      * @param labelLayout The layout of the label, default is N
+      * @return this
+      */
+    def setLayout(dataLayout: String, labelLayout: String): Builder = {
```
yzhliu: same as setdtype

scala-package/core/src/main/scala/org/apache/mxnet/IO.scala
```diff
@@ -352,7 +431,16 @@ object DataDesc {
...
   def getBatchAxis(layout: Option[String]): Int = {
-    layout.map(_.indexOf('N')).getOrElse(0)
+    if (layout.isEmpty|| layout.get == Layout.UNDEFINED) {
+      logger.info("Found Undefined Layout, will use default index 0")
+      0
+    } else {
+      if (layout.get.contains('N')) {
+        layout.get.indexOf("N")
+      } else {
+       throw new IllegalArgumentException("No N found in Batch Axis!")
```
yzhliu: incorrect indent ?

scala-package/core/src/main/scala/org/apache/mxnet/io/MXDataIter.scala
```diff
@@ -30,32 +31,55 @@ import scala.collection.mutable.ListBuffer
...
   // fix me if any better way found)
   private var currentBatch: DataBatch = null
 
-  private val (_provideData: ListMap[String, Shape],
-               _provideLabel: ListMap[String, Shape],
-               _batchSize: Int) =
+  private val (_provideDataDesc: IndexedSeq[DataDesc],
+               _provideLabelDesc: IndexedSeq[DataDesc],
+                _provideData: ListMap[String, Shape],
+                _provideLabel: ListMap[String, Shape],
```
yzhliu: incorrect indent?

scala-package/core/src/main/scala/org/apache/mxnet/IO.scala
```diff
@@ -136,11 +142,29 @@ class DataBatch(val data: IndexedSeq[NDArray],
...
                 // use ListMap to indicate the order of data/label loading
                 // (must match the order of input data/label)
-                private val providedData: ListMap[String, Shape] = null,
-                private val providedLabel: ListMap[String, Shape] = null) {
+                private val providedData: ListMap[String, Shape],
+                private val providedLabel: ListMap[String, Shape],
+                val dataDType: DType,
+                val labelDType: DType,
+                val dataLayout: String,
+                val labelLayout: String) {
```
yzhliu: I'm wondering whether we can pack them all into dataDesc and labelDesc?

gigasquid:
I ran the Clojure examples and they all look good. They do have a warning
`INFO  org.apache.mxnet.DataDesc: Found Undefined Layout, will use default index 0` but everything is fine.

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
```diff
@@ -42,27 +43,40 @@ import scala.collection.immutable.ListMap
 class NDArrayIter(data: IndexedSeq[(String, NDArray)],
                   label: IndexedSeq[(String, NDArray)],
                   private val dataBatchSize: Int, shuffle: Boolean,
-                  lastBatchHandle: String) extends DataIter {
-
+                  lastBatchHandle: String,
+                  dataDType: DType, labelDType: DType,
+                  dataLayout: String, labelLayout: String) extends DataIter {
+
+  // scalastyle:off
```
nswamy: why?

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
```diff
@@ -101,10 +115,16 @@ class NDArrayIter(data: IndexedSeq[(String, NDArray)],
   private var cursor = -dataBatchSize
 
   private val (_provideData: ListMap[String, Shape],
-               _provideLabel: ListMap[String, Shape]) = {
+               _provideLabel: ListMap[String, Shape],
```
nswamy: can you separate the _provideDataDesc and _provideLabelDesc into separate statement

scala-package/core/src/test/scala/org/apache/mxnet/IOSuite.scala
nswamy: ?

scala-package/examples/src/main/scala/org/apache/mxnetexamples/rnn/BucketIo.scala
```diff
@@ -94,8 +96,22 @@ object BucketIo {
...
-      seperateChar: String = " <eos> ", text2Id: Text2Id = defaultText2Id,
-      readContent: ReadContent = defaultReadContent) extends DataIter {
+      seperateChar: String, text2Id: Text2Id,
+      readContent: ReadContent,
+      dataLayout: String,
+      labelLayout: String,
+      dataDType : DType,
+      labelDType: DType) extends DataIter {
+
+    // scalastyle:off
```
nswamy: why are turning off scalastyle, i saw in a few other places.

scala-package/pom.xml
```diff
@@ -231,6 +231,7 @@
           <skipTests>${skipTests}</skipTests>
           <reportsDirectory>${project.build.directory}/surefire-reports</reportsDirectory>
           <junitxml>.</junitxml>
+          <stdout>F</stdout>
```
nswamy: is this needed?

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
lanking520: I finally figure out the scala style issue. Scala does not allow user to define a function with more than a certain numbers of args...

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
lanking520: @yzhliu recommend do this several days ago in previous CR. Which one works better?

scala-package/pom.xml
lanking520: It brings the whole stacktraces for Scalatest. So yes, needed and good for all developers to debug.

scala-package/examples/src/main/scala/org/apache/mxnetexamples/rnn/BucketIo.scala
lanking520: Same reason explained above

lanking520:
Thanks @nswamy @yzhliu @gigasquid for CR and testing. Next Commit will cover:
- spacing problem fixes
- Introduce DataDesc somewhere to reduce the amount of args
- `setDType` and `setLayout`. I think these fields should be available to be defined in databatches.

nswamy:
Why do think `setDType` and `setLayout` is need when you get it through the constructor as a DataDesc ?

lanking520:
@nswamy How to bypass the `Shape` field in DataDesc? That's the block I have to make this change.

scala-package/core/src/main/scala/org/apache/mxnet/IO.scala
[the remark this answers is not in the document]
nswamy: Look at the constructor you are getting all 4 fields of DataDesc. 
```
                private val providedData: ListMap[String, Shape],
                private val providedLabel: ListMap[String, Shape],
                val dataDType: DType,
                val labelDType: DType,
                val dataLayout: String,
                val labelLayout: String)
```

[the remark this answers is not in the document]
lanking520: But since user always define these field as null, it will introduce crashes when we fetch the information from here, I didn't even know why we need provideData and provideLabel here.

scala-package/core/src/main/scala/org/apache/mxnet/IO.scala
yzhliu: how about replace `providedData` by `ListMap[String, DataDesc]`, and providedLabel respectively. (can create a new constructor to keep backward-compatibility.
Actually there's no guarantee that all inputs share the same layout and dtype.

yzhliu: @nswamy maybe need to keep the `Map`, I think some usage queries by name.

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
yzhliu: My major point was no need to fetch the data twice

scala-package/core/src/main/scala/org/apache/mxnet/IO.scala
nswamy: can you elaborate a bit? DataDesc already has `name`

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
lanking520: Done in the recent Commit

scala-package/core/src/main/scala/org/apache/mxnet/DType.scala
lanking520: Done in the recent commit

lanking520:
@nswamy @yzhliu thanks for your second phase of CR, I have applied changes:

- DataBatch: Use `IndexSeq[DataDesc]` to replace ListMap
- Remove getLayout and getDType in all of the code portion (seemed not being used)

nswamy:
This PR adds quite a lot of new APIs, please add corresponding tests.

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
```diff
@@ -42,27 +43,40 @@ import scala.collection.immutable.ListMap
 class NDArrayIter(data: IndexedSeq[(String, NDArray)],
                   label: IndexedSeq[(String, NDArray)],
                   private val dataBatchSize: Int, shuffle: Boolean,
-                  lastBatchHandle: String) extends DataIter {
-
+                  lastBatchHandle: String,
+                  dataDType: DType, labelDType: DType,
```
nswamy: With this change all inputs/labels will have the same DType and Label, however data and labels can take IndexedSeq[(String, NDArray)] -- this does not seem correct. 

Can the NDArrayIter constructor be
```
class NDArrayIter(data: IndexedSeq[NDArray], 
label: IndexedSeq[NDArray], 
val dataDesc: IndexedSeq[DataDesc],
val labelDesc: IndexedSeq[DataDesc],
val dataBatchSize: Int,
lastBatchHandle: String)
```

scala-package/core/src/main/scala/org/apache/mxnet/IO.scala
```diff
@@ -352,7 +402,16 @@ object DataDesc {
...
   def getBatchAxis(layout: Option[String]): Int = {
-    layout.map(_.indexOf('N')).getOrElse(0)
+    if (layout.isEmpty|| layout.get == Layout.UNDEFINED) {
+      logger.info("Found Undefined Layout, will use default index 0")
+      0
+    } else {
+      if (layout.get.contains('N')) {
+        layout.get.indexOf("N")
+      } else {
+        throw new IllegalArgumentException("No N found in Batch Axis!")
```
nswamy: no Batch Axis('N') found in Layout

scala-package/core/src/main/scala/org/apache/mxnet/Layout.scala
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.mxnet
+
+object Layout {
```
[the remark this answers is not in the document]
nswamy: ?

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
```diff
@@ -42,27 +43,40 @@ import scala.collection.immutable.ListMap
...
+                  dataLayout: String, labelLayout: String) extends DataIter {
+
+  // scalastyle:off
+  def this(data: IndexedSeq[NDArray], label: IndexedSeq[NDArray],
+           dataBatchSize: Int, shuffle: Boolean,
+           lastBatchHandle: String,
+           dataName: String, labelName: String,
+           dataDType: DType, labelDType: DType,
+           dataLayout: String, labelLayout: String) {
+    this(IO.initData(data, allowEmpty = false, dataName),
```
nswamy: how about changing/adding API for `initData` that will give you `IndexedSeq[DataDesc]` instead of `IndexedSeq[(String, NDArray)]
`initData` seems to be giving the default names for Inputs and labels.

```
private def initDataDesc(data: Indexed[ND]): Indexed[DataDesc]
```

scala-package/core/src/main/scala/org/apache/mxnet/io/PrefetchingIter.scala
```diff
@@ -68,6 +70,42 @@ class PrefetchingIter(
...
+    }
+  }
+
+  private val _provideLabelDesc: IndexedSeq[DataDesc] = {
+    if (dataNames == null) {
+      iters.map(_.provideLabelDesc).foldLeft(IndexedSeq[DataDesc]()) { (acc, elem) =>
+        acc ++ elem
+      }
+    } else {
+      iters.zipWithIndex.map(tu => (tu._1.provideDataDesc, tu._2))
```
andrewfayres: Should this be 
```
iters.zipWithIndex.map(tu => (tu._1.provideLabelDesc, tu._2))
```

lanking520: Nice!

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
andrewfayres: Can we put the DType and Label in the same IndexedSeq as the NDArray? It would help make this a little less prone to bugs because they two would be coupled together.

Something like:
```scala
class NDArrayIter(data: IndexedSeq[(DataDesc, NDArray)],
label: IndexSeq[(DataDesc, NDArray)],
val dataBatchSize: Int,
lastBatchHandle: String)
```

nswamy: like this better.

scala-package/core/src/main/scala/org/apache/mxnet/IO.scala
```diff
@@ -332,8 +379,9 @@ abstract class DataPack() extends Iterable[DataBatch] {
 
 // Named data desc description contains name, shape, type and other extended attributes.
 case class DataDesc(name: String, shape: Shape,
-                    dtype: DType = Base.MX_REAL_TYPE, layout: String = "NCHW") {
-  require(shape.length == layout.length, ("number of dimensions in shape :%d with" +
+                    dtype: DType = Base.MX_REAL_TYPE, layout: String = Layout.UNDEFINED) {
```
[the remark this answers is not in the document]
lanking520: It break the Backwards compatibility, user need to define `DataDesc` as `new DataDesc` instead of `DataDesc` directly

scala-package/core/src/main/scala/org/apache/mxnet/IO.scala
```diff
@@ -352,9 +399,19 @@ object DataDesc {
    *         for each data-parallelism device.
    */
   def getBatchAxis(layout: Option[String]): Int = {
-    layout.map(_.indexOf('N')).getOrElse(0)
+    if (layout.isEmpty|| layout.get == Layout.UNDEFINED) {
+      logger.warn("Found Undefined Layout, will use default index 0")
```
yzhliu: will use default index 0 for batch axis

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
```diff
@@ -39,35 +40,35 @@ import scala.collection.immutable.ListMap
...
+    * @param label Same as data, but is not fed to the model during testing.
+    *              Label names will be label_0, label_1, ..., etc.
+    * @param dataBatchSize Batch Size
+    * @param shuffle Whether to shuffle the data
+    * @param lastBatchHandle "pad", "discard" or "roll_over". How to handle the last batch
+    *
+    * This iterator will pad, discard or roll over the last batch if
+    * the size of data does not match batch_size. Roll over is intended
+    * for training and can cause problems if used for prediction.
+    */
```
yzhliu: avoid changes like these.

scala-package/core/src/main/scala/org/apache/mxnet/IO.scala
```diff
@@ -228,15 +264,14 @@ object DataBatch {
...
      */
-    def provideDataShape(name: String, shape: Shape): Builder = {
-      if (datatShapes == null) {
-        datatShapes = ListMap((name, shape))
+    def provideDataShape(dataDesc: DataDesc): Builder = {
+      if (dataShapes == null) {
+        dataShapes = IndexedSeq(dataDesc)
       } else {
-        datatShapes = datatShapes.updated(name, shape)
+        dataShapes = dataShapes ++ IndexedSeq(dataDesc)
```
nswamy: provideDataShape->`provideDataDesc`
this does not look right. which Index of the Input do you want to add to?.  You should maintain the same semantic as in the constructor ie., take `IndexedSeq[DataDesc]` as arguments for Builder methods. 
If this is to keep it compatible with Java, I suggest you create another method that takes `IndexedSeq[DataDesc]`..although i think Scala might cause problems due to erasure

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
```diff
@@ -101,20 +102,30 @@ class NDArrayIter(data: IndexedSeq[(String, NDArray)],
...
 
   private val (_provideData: ListMap[String, Shape],
-               _provideLabel: ListMap[String, Shape]) = {
+               _provideLabel: ListMap[String, Shape],
+               _provideDataDesc: IndexedSeq[DataDesc],
+               _provideLabelDesc: IndexedSeq[DataDesc]) = {
     val pData = ListMap.empty[String, Shape] ++ initData.map(getShape)
     val pLabel = ListMap.empty[String, Shape] ++ initLabel.map(getShape)
-    (pData, pLabel)
+    val pDData = IndexedSeq.empty[DataDesc] ++ initData.map(ele => {
```
nswamy: do you need derive the DataDesc/LabelDesc when they are passed in the constructor.  Am I missing something?

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
```diff
@@ -261,7 +283,24 @@ object NDArrayIter {
...
-      this.label = this.label ++ IndexedSeq((name, label))
+      this.label = this.label ++ IndexedSeq((new DataDesc(name,
+        label.shape, DType.Float32, Layout.UNDEFINED), label))
+      this
+    }
+
+    /**
+      * Add one data input with its DataDesc
+     */
+    def addDataDesc(dataDesc: DataDesc, data: NDArray): Builder = {
```
nswamy: `addDataDesc`->`addDataWithDesc`?

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
```diff
@@ -261,7 +283,24 @@ object NDArrayIter {
...
+      this.data = this.data ++ IndexedSeq((dataDesc, data))
+      this
+    }
+
+    /**
+      * Add one label input with its DataDesc
+      */
+    def addLabelDesc(labelDesc: DataDesc, label: NDArray): Builder = {
```
nswamy: addLabelDesc->addLabelWithDesc?

scala-package/core/src/main/scala/org/apache/mxnet/IO.scala
nswamy: Do you want to the same as you what did for NDArrayIter.builder

scala-package/core/src/main/scala/org/apache/mxnet/io/NDArrayIter.scala
lanking520: yes, the shape information got update

scala-package/core/src/main/scala/org/apache/mxnet/IO.scala
lanking520: @yzhliu the previous builder only allows single dataDesc (shape, name) input for the DataBatch. However, the data passed in is a IndexedSeq, in the new design, I changed them into indexedSeq and remove the old method. 

@nswamy About NDArrayIter, they are single input for both data and DataDesc, I am not planning to do the same thing there.

PR #11862, opened 2018-07-23
[MXNET-710] Change POM files to be able to regularly publish to Apache Release & Maven Central Repo

nswamy:
## Description ##
This brings the ability to publish Maven packages using Apache Maven Release plugin and is integrated with the existing Make file System. The steps to perform the Scala Release is documented here: 
https://cwiki.apache.org/confluence/display/MXNET/MXNet-Scala+Release+Process.

This process was used to publish both 1.2.0 and 1.2.1 releases.  This could be further automated by integrating to the CI system to publish SNAPSHOTS to the Apache SnapShot repo which can be picked up to perform the Release.

@lanking520 @andrewfayres

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -589,6 +589,7 @@ scalaclean:
 scalapkg:
 	(cd $(ROOTDIR)/scala-package; \
 		mvn package -P$(SCALA_PKG_PROFILE),$(SCALA_VERSION_PROFILE) -Dcxx="$(CXX)" \
+		    -Dbuild.platform="$(SCALA_PKG_PROFILE)" \
 			-Dcflags="$(CFLAGS)" -Dldflags="$(LDFLAGS)" \
 			-Dcurrent_libdir="$(ROOTDIR)/lib" \
 			-Dlddeps="$(LIB_DEP) $(ROOTDIR)/lib/libmxnet.a")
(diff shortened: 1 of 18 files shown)


Makefile
```diff
@@ -589,6 +589,7 @@ scalaclean:
 scalapkg:
 	(cd $(ROOTDIR)/scala-package; \
 		mvn package -P$(SCALA_PKG_PROFILE),$(SCALA_VERSION_PROFILE) -Dcxx="$(CXX)" \
+		    -Dbuild.platform="$(SCALA_PKG_PROFILE)" \
```
lanking520: Can't we used this param in the `profile` directly?

Makefile
```diff
@@ -608,12 +609,32 @@ scalaintegrationtest:
 scalainstall:
 	(cd $(ROOTDIR)/scala-package; \
 		mvn install -P$(SCALA_PKG_PROFILE),$(SCALA_VERSION_PROFILE) -DskipTests -Dcxx="$(CXX)" \
+		    -Dbuild.platform="$(SCALA_PKG_PROFILE)" \
 			-Dcflags="$(CFLAGS)" -Dldflags="$(LDFLAGS)" \
 			-Dlddeps="$(LIB_DEP) $(ROOTDIR)/lib/libmxnet.a")
 
+scalarelease-dryrun:
```
lanking520: Could you please also include explanations on these make target in the CWiki and possibly provide the links to explains the reason to run this and the following two make target below?

scala-package/assembly/osx-x86_64-cpu/pom.xml
```diff
@@ -26,10 +26,51 @@
...
+      <version>1.3.0-SNAPSHOT</version>
+    </dependency>
   </dependencies>
 
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>flatten-maven-plugin</artifactId>
+        <!--<version>1.1.0-SNAPSHOT</version>-->
```
lanking520: Remove this comments?

scala-package/dev/compile-mxnet-backend.sh
```diff
@@ -0,0 +1,107 @@
...
+  exit 1
+fi
+PLATFORM=$1
+MXNETDIR=$2
+
+
+# below routine shamelessly copied from
+# https://github.com/apache/incubator-mxnet/blob/master/setup-utils/install-mxnet-osx-python.sh
+
+chkret() {
```
lanking520: What is chkret?

scala-package/dev/compile-mxnet-backend.sh
```diff
@@ -0,0 +1,107 @@
...
+    echo "USE_OPENMP=1" >> ./config.mk
+    echo "USE_BLAS=openblas" >> ./config.mk
+    echo "USE_LAPACK=1" >> ./config.mk
+    echo "USE_DIST_KVSTORE=1" >> ./config.mk
+    echo "USE_S3=1" >> ./config.mk
+    make -j$(cat /proc/cpuinfo | awk '/^processor/{print $3}' | tail -1)
+elif [[ $PLATFORM == "linux-x86_64-gpu" ]]
+then
+    echo "Building MXNet Backend on Linux GPU"
+    echo "Building MXNet Backend on Linux CPU"
```
lanking520: remove this line?

Makefile
nswamy: I think I have tried this, I couldn't use read this from the profile in release/release-perform that is why I had to add an additional property.
```
-Darguments=""-Dbuild\.platform=\""$(SCALA_PKG_PROFILE)\""
```

nswamy:
@lanking520 regarding
```
      <plugin>
        <artifactId>maven-deploy-plugin</artifactId>
       <configuration>
          <skip>true</skip>
        </configuration>
     </plugin>
```
I am not sure if I got your suggestion right, but that is exactly how it is done right now.

lanking520:
@nswamy The normal practice is to add the configuration file in the parent pom and all child modules will extend that. So if you add the `maven-deploy-plugin` in the parent pom, you will see it applies to every where in all of the modules. In that case, you only need to specify the `<skip>false</skip>` to the place you want to be instead of adding `<skip>true</skip>` in every pom file

lanking520:
Agreed. We can create another PR specifically on cleaning the pom files and build cycle.

scala-package/pom.xml
```diff
@@ -213,15 +209,10 @@
...
       </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-deploy-plugin</artifactId>
-        <version>2.8.2</version>
-      </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
-        <version>2.19</version>
```
lanking520: Why do we degrade the version here?

nswamy: good catch. will fix it

PR #11885, opened 2018-07-25
Fix JNI custom op code from deregistering the operator fixes #10438

### LINKED ISSUE #10438: Loading an older model with a custom operator from Java/Scala more than twice causes MXNet to crash
## Version
MXNet 1.1.0 with Scala

## Description
Our Scala application uses MxNet and includes several tests. Many of these tests load the same model files, so as a result the same process ends up loading a model multiple times. This normally works fine, but when upgrading from 0.11.0 to 1.1.0 we discovered an odd issue causing our tests to fail. We discovered that if you have a model that uses a custom operator and was built from an mxnet version older than the current one (thus prompting the "Attempting to upgrade..." message), and you load it more than twice in the same process, then on the third reload it crashes.

## Analysis
Based on my investigation, the relevant lines appear to be:

- `legacy_json_util.cc#UpgradeJSON_FixParsing (line 30)`: This method is run whenever a model older than the current version is loaded. It invokes the attribute parser for each op.
- `legacy_json_util.cc#UpgradeJSON_Parse (line 92)`: This method is always run. It also invokes the attribute parser for each op
- `custom.cc#AttrParser (line 80)`: This line resets the shared pointer to the custom operator's callback list. Note that it also sets up a deleter function that unregisters the op by invoking the function in the next bullet.
- `ml_dmlc_mxnet_native_c_api.cc#opPropDel (line 2426)`: This function unregisters the custom operator. However, there is logic here that tracks how many times this function has been invoked. It only deregisters the custom op on the _third_ invocation.

Suppose we are running MXNet 1.1.0 and load a model built on 0.11.0. Because of the version mismatch, both of the UpgradeJSON stages are run, each one running the attribute parser for each op. When the custom op's attribute parser is run the second time, the shared pointer to the operator's callback list is reset, triggering the unregister function to be invoked. This function, for now, simply updates a counter. The counter now has value 1.

Now suppose during the same process, we load the same model again. The same scenario as above happens again, except this time the counter is incremented to 2.

The third time the model is loaded, again the same scenario reoccurs, except this time the counter is already at value 2, so the unregister function actually unregisters the custom op rather than simply updating the counter. Our custom operator is no longer registered, so shortly thereafter we get an "Operator [op-name] is not registered" fatal error.

andrewfayres:
## Description ##
This PR fixes a bug where custom operators get deregistered while upgrading a model which was trained in a previous version of mxnet. During the upgrade process the engine calls the custom operator method opPropDel every time it encounters a customer operator which it's already seen. The code had a counter that did nothing until the 3rd time called then deregistered the operator. This caused an whenever the same custom operator was encountered 3 times (either by loading in the model multiple times or loading a model which used the operator multiple times).

The fix is to basically do nothing during the delete callback. The purpose of the callback seems to be to clean up resources that were allocated during the creator function and our creator function isn't allocating any new resources but is instead reinitializing existing resources which were created during the Operator.register call.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/scala-package/native/src/main/native/org_apache_mxnet_native_c_api.cc
+++ b/scala-package/native/src/main/native/org_apache_mxnet_native_c_api.cc
@@ -1898,9 +1898,6 @@ JNIEXPORT jint JNICALL Java_org_apache_mxnet_LibInfo_mxRtcFree
 
 // store the user defined CustomOpProp object reference with its name
 std::unordered_map<std::string, jobject> globalOpPropMap;
-// store how many time of the delete function was called
-// for a specific CustomOpProp object
-std::unordered_map<std::string, int> globalOpPropCountMap;
 // store the user defined CustomOp object reference with its name
 std::unordered_map<std::string, jobject> globalOpMap;
 // used for thread safty when insert  elements into
@@ -1908,6 +1905,7 @@ std::unordered_map<std::string, jobject> globalOpMap;
 std::mutex mutex_opprop;
 std::mutex mutex_op;
 
+// Registers a custom operator when called
 JNIEXPORT jint JNICALL Java_org_apache_mxnet_LibInfo_mxCustomOpRegister
   (JNIEnv *env, jobject obj, jstring jregName, jobject jopProp) {
   const char *regName = env->GetStringUTFChars(jregName, 0);
@@ -1915,14 +1913,13 @@ JNIEXPORT jint JNICALL Java_org_apache_mxnet_LibInfo_mxCustomOpRegister
 
   std::unique_lock<std::mutex> lock(mutex_opprop);
   globalOpPropMap.insert({ key, env->NewGlobalRef(jopProp) });
-  globalOpPropCountMap.insert({ key, 0 });
   lock.unlock();
 
+  // lambda function to initialize the operator and create all callbacks
   auto creatorLambda = [](const char *opType, const int numKwargs,
     const char  **keys, const char **values, MXCallbackList *ret) {
     int success = true;
 
-    // set CustomOpProp.kwargs
     std::string opPropKey(opType);
     if (globalOpPropMap.find(opPropKey) == globalOpPropMap.end()) {
       LOG(WARNING) << "CustomOpProp: " << opPropKey << " not found";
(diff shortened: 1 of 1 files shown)


andrewfayres:
@lanking520

nswamy:
@andrewfayres is it possible to add some testing to this?

andrewfayres:
It's more a question of exactly what you want to test. We've got tests for custom operators already and there's already work going on to verify model backward compatibility.

andrewfayres:
Adding a bit of detail based on our conversations this morning

@lanking520 I've seen no memory leaks resulting from the counter being removed. The reason is because the code that would eventually run after getting passed the counter would clean up very little. It'd remove the entry from the globalOperator map and delete the reference in the jvm. This effectively deregistered the operator in the JNI side while it could still be in scope in the scala side. The memory that leaks here is gets allocated during the Operator.register scala call. This is the same before and after this fix (we should add a dispose method and/or a Operator.deregister method but that's a bit beyond the scope of fixing this bug).

@nswamy I do not believe that there are currently any existing unit tests for our custom operator jni code. The whole JNI file should be tested via unit tests and I will create an issue for us to follow through on that. We do have a custom operator example that gets run during the CI and provides a minimal amount of testing. My manual testing mostly involved running a modified version of our example multiple times. Verifying that I could reproduce the bug and that the crash stopped happening after digging into the code and make the change. Then I verified that both my modified version and the original still worked after the change. The code that was being run after the count check was guaranteed to make an operator unusable in the native side and would crash whenever the jvm tried to use that operator.

lanking520:
It's been better to solve customer problems, though it brings a memory leak issues. @andrewfayres please add the problem here as an issue in Scala.

andrewfayres:
#10438 thought I tagged it in the title but that seems to of not worked.

lupesko:
@nswamy bouncing this up for review

andrewfayres:
[Issue](https://issues.apache.org/jira/browse/MXNET-852) to add testing to JNI code.

I'll expand on what I said earlier for testing:
Reproducing the issue was fairly straight forward. Using the existing custom operator as a template, I trained and saved a model with a custom operator. Then I attempted to load the model multiple times. On the 3rd attempt a fatal "Operator not found error" was thrown. I then trained another model with 3 of the same custom operators, saved it, then attempted to load it and received the same error.

In both cases, the error was being thrown whenever the upgrade model process called the opPropDel for the 3rd time. The best that I can tell the opPropDel method is only called by the upgradeJSON method when loading a model trained on an older version of MXNet. In this case, there was a counter in the JNI code that on the 3rd time opPropDel was called the custom operator would be deregistered in the native code making all instances of the custom operator unusable and producing a fatal error on the next attempt.

My change was to remove the code that did the deregister and make opPropDel method simply return success (which is what it was already doing up until it hit the counter threshold).

After making this change I loaded the same models that I had trained previously and this time they loaded without issue.

nswamy:
Thanks @andrewfayres for the detailed explanation. Next time please add the results also in the PR. What makes such PR hard is that we are basing off of manual tests which are not reproducible unless another person does the exact same manual test to verify. The concern I have in such cases is the possibility of introducing regressions somewhere else.

Issue #11176, opened 2018-06-06
Failed to find any forward convolution algorithm.


ThomasDelteil:
See this test failing: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/master/915/pipeline/

with 
```
 src/operator/nn/./cudnn/cudnn_convolution-inl.h:744: Failed to find any forward convolution algorithm.
```

I have encountered this in the wild very rarely too.

marcoabreu:
@DickJC123

marcoabreu:
This is not CI related

eric-haibin-lin:
Is it possible that memory is exhausted on CI?

aluo-x:
Also encountered this error on Windows, CUDA 9.2, cudnn 7.1.4.

Ran the following command `python train_imagenet.py --benchmark 1 --gpus 0 --network inception-v3 --batch-size 64 --image-shape 3,299,299 --num-epochs 1 --kv-store device`

Reducing the batch size to 16 resolved the issue.

codewithsk:
Currently facing this issue. Reducing batch size does not seem to fix the issue. 

Trying to train [fast neural style transfer](https://github.com/SineYuan/mxnet-fast-neural-style/tree/605da92ba3ece0107eff64f3b7e1c13f28c6f0dd )

Issues seems to arise when trying to do mod.save_params() throws the following error:

```
mxnet.base.MXNetError: [23:59:34] src/operator/nn/./cudnn/cudnn_convolution-inl.h:744: Failed to find any forward convolution algorithm.
```

codewithsk:
Update: I've managed to find a rather bizarre workaround to this issue.

I was facing this issue when I was trying to do a `model.save_checkpoint()`. However, if I caught the exception and saved it in the except block, it seemed to work flawlessly

```python
    try:
        mod.save_checkpoint(model_save_path, epoch)
    except Exception as excep:
        print("Exception caught: ", excep)
        mod.save_checkpoint(model_save_path, epoch)
```

codewithsk:
Update: Sleeping for 0.5 seconds before sleeping also seems to help.

```python
    time.sleep(0.5)
    mod.save_checkpoint(model_save_path, epoch)
```

haojin2:
@ThomasDelteil Is this still occurring on CI now? If it's not appearing again would you mind closing this?
[...]

PR #11886, opened 2018-07-25
Improve error message of cudnn operators

haojin2:
## Description ##
From the comments under #11176 seems like there's a wide confusion on the error messages in cudnn convolution operators, enhancing the error messages so that we give hints to users to reduce confusion.

### Essentials ###

### Changes ###
- [x] Enhancement of error messages in cudnn_conv
- [x] Enhancement of error messages in cudnn_deconv

## Comments ##
@ThomasDelteil

### DIFF
--- a/src/operator/nn/cudnn/cudnn_convolution-inl.h
+++ b/src/operator/nn/cudnn/cudnn_convolution-inl.h
@@ -685,6 +685,7 @@ class CuDNNConvolutionOp {
       const int kMaxAlgos = 10;
       int nalgo = kMaxAlgos;
       int i = 0;
+      size_t min_memory_needs = 0;
       // Forward Algorithm Find/Get, v6 and earlier
       if (CUDNN_MAJOR == 6 && param_.layout.value() == mshadow::kNHWC) {
         // In cuDNNv6, for kNHWC, only CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM is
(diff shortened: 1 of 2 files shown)


haojin2:
@anirudh2290 @reminisce @eric-haibin-lin @rahul003

ptrendx:
I'm not sure I like this error message actually - at the time of raising this error you have full info on why there was no algorithm chosen and so you should just state the exact reason. Basically something like:
```
Failed to find suitable forward convolution algorithm. N algorithms we're supported with minimum workspace size of M bytes, whereas the limit on workspace size is W. Please try increasing the allowed workspace size or decreasing batch size.
```
I think it would be much more informative (it would also help if we could know which convolution actually triggered this error) while still not confusing.

haojin2:
@ptrendx Sure I can change it to that way! But I'm afraid I'll have to do it tomorrow, as it's pretty late now in the Bay Area. Will let you know once that version's ready. Thanks for your feedback!

haojin2:
@ptrendx How does the following message look to you?
```
N algorithms with minimum memory requirement M bytes have been tried. There are only X bytes available workspace on your GPU, please consider reduce the batch size or model size.
```

haojin2:
@ptrendx Okay, so how does:
```
N algorithms with minimum memory requirement M bytes have been tried. Workspace size is set to X bytes, please consider reducing the batch/model size or increasing workspace size.
```
look to you?

ptrendx:
Sounds good.

haojin2:
@reminisce @eric-haibin-lin @rahul003 @anirudh2290 Please give a review again as the code has changed a bit since last update.

haojin2:
@eric-haibin-lin Should be good for merge?

PR #11908, opened 2018-07-27
Fix shared memory with gluon dataloader, add option pin_memory

zhreshold:
## Description ##
- This PR fixes gluon data loader which use too aggressive pre-fetching strategy which will cause shared_memory usage being super high (storing too much in data queue). 
- Add option `pin_memory` to allow data loader to pin data into `cpu_pinned` context. Which can potentially speed up the following copy to GPU process.
- Removed the doc description: `num_worker > 0 is not available on windows` since they are runnable. 
- Enabled unittests on windows for multi worker data loader

### Essentials ###

### DIFF
--- a/ci/windows/test_py2_cpu.ps1
+++ b/ci/windows/test_py2_cpu.ps1
@@ -16,7 +16,8 @@
 # under the License.
 
 7z x -y windows_package.7z
-$env:PYTHONPATH=join-path $pwd.Path windows_package\python
+for /r %%i in (*.dll) do xcopy /Y /C "%%i" .\\python\\mxnet\\
+$env:PYTHONPATH=join-path $pwd.Path python
 $env:MXNET_STORAGE_FALLBACK_LOG_VERBOSE=0
 c:\Anaconda3\envs\py2\Scripts\pip install -r tests\requirements.txt
 c:\Anaconda3\envs\py2\python.exe -m nose -v --with-xunit --xunit-file nosetests_unittest.xml tests\python\unittest
--- a/ci/windows/test_py2_cpu.ps1
+++ b/ci/windows/test_py2_cpu.ps1
@@ -16,8 +16,8 @@
 # under the License.
 
 7z x -y windows_package.7z
-for /r %%i in (*.dll) do xcopy /Y /C "%%i" .\\python\\mxnet\\
-$env:PYTHONPATH=join-path $pwd.Path python
+$env:MXNET_LIBRARY_PATH=join-path $pwd.Path windows_package\python\mxnet\libmxnet.dll
+$env:PYTHONPATH=join-path $pwd.Path windows_package\python
 $env:MXNET_STORAGE_FALLBACK_LOG_VERBOSE=0
 c:\Anaconda3\envs\py2\Scripts\pip install -r tests\requirements.txt
 c:\Anaconda3\envs\py2\python.exe -m nose -v --with-xunit --xunit-file nosetests_unittest.xml tests\python\unittest
--- a/ci/windows/test_py2_cpu.ps1
+++ b/ci/windows/test_py2_cpu.ps1
@@ -16,7 +16,7 @@
 # under the License.
 
 7z x -y windows_package.7z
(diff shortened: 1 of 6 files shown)


Jenkinsfile
```diff
@@ -850,8 +850,9 @@ try {
                 xcopy C:\\mxnet\\model model /E /I /Y
                 call activate py2
                 pip install mock
-                set PYTHONPATH=${env.WORKSPACE}\\pkg_vc14_cpu\\python
-                del /S /Q ${env.WORKSPACE}\\pkg_vc14_cpu\\python\\*.pyc
+                set PYTHONPATH=${env.WORKSPACE}\\python
+                for /r %%i in (*.dll) do xcopy /Y /C "%%i" .\\python\\mxnet\\
```
marcoabreu: Would this maybe be a better fit for our makefile?

zhreshold: what do you suggest?

zhreshold:
@piiswrong Can you have a look?

python/mxnet/gluon/data/dataloader.py
```diff
@@ -277,12 +298,16 @@ def default_batchify_fn(data):
 
     num_workers : int, default 0
         The number of multiprocessing workers to use for data preprocessing.
-        `num_workers > 0` is not supported on Windows yet.
+    pin_memory : boolean, default False
```
rahul003: Why default to False?

zhreshold: not necessary for non-gpu instances, and using non-pagable memory for everything can screw up the stability

zhreshold:
@marcoabreu  @larroy For windows to correctly pickle test functions, I copied dll to where unittests are launched, PYTHONPATH is set accordingly, please have a look.

ci/windows/test_py2_cpu.ps1
```diff
@@ -16,7 +16,8 @@
 # under the License.
 
 7z x -y windows_package.7z
-$env:PYTHONPATH=join-path $pwd.Path windows_package\python
+for /r %%i in (*.dll) do xcopy /Y /C "%%i" .\\python\\mxnet\\
```
larroy: Can we explain this? I don't get it, aren't the packed files in the windows_package folder?

tests/python/unittest/test_gluon_data.py
```diff
@@ -130,99 +130,89 @@ def test_multi_worker():
     for i, batch in enumerate(loader):
         assert (batch.asnumpy() == i).all()
 
-@with_seed()
-def test_multi_worker_forked_data_loader():
+class _Dummy(Dataset):
+    """Dummpy dataset for randomized shape arrays."""
```
larroy: typo?

ci/windows/test_py2_cpu.ps1
zhreshold: I am trying to make PYTHONPATH the git cloned one instead of stashed build, because otherwise when unittest is launched in git cloned folder, the pickled functions(from multiprocess) cannot find the correct dlls.

tests/python/unittest/test_gluon_data.py
zhreshold: will fix that

Jenkinsfile
larroy: I also don't get this change. We have MXNET_LIBRARY_PATH to locate the dll.

larroy:
We have a copy of python inside the windows package as well. It should be the same as the checked out repo. But there should be no need to copy the dlls around. What was the problem that made you copy them? I would like to understand this, as we are working on this windows builds actively.

ci/windows/test_py2_cpu.ps1
larroy: Can you instead set the MXNET_LIBRARY_PATH  environment variable? this should be the correct way.

zhreshold:
@larroy Do you know where I can point MXNET_LIBRARY_PATH to? 
Error log: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/PR-11908/12/pipeline/896

Issue #11855, opened 2018-07-23
Distributed learning with Async update does not work.


absalama:
Hi All, 

I am running currently distributed learning using the **launch.py** tool. My plan is to run the distributed learning by using Sync and Async distributed modes. For the Sync PS, the training works fine and I got the number. However, when I switch the kvstore type to "dist_async", the following error shows up: 

`terminate called after throwing an instance of 'dmlc::Error'
  what():  [11:58:12] src/kvstore/././kvstore_dist_server.h:294: Check failed: sync_mode_ Updater needs to be set for async mode
Stack trace returned 10 entries:
[bt] (0) /work/projects/Project00755/envs/mxnet_gpu/lib/python3.5/site-packages/mxnet/libmxnet.so(+0x30cbe2) [0x2aedf8ebdbe2]
[bt] (1) /work/projects/Project00755/envs/mxnet_gpu/lib/python3.5/site-packages/mxnet/libmxnet.so(+0x30d1b8) [0x2aedf8ebe1b8]
[bt] (2) /work/projects/Project00755/envs/mxnet_gpu/lib/python3.5/site-packages/mxnet/libmxnet.so(+0x2ab1f11) [0x2aedfb662f11]
[bt] (3) /work/projects/Project00755/envs/mxnet_gpu/lib/python3.5/site-packages/mxnet/libmxnet.so(+0x2ab6f10) [0x2aedfb667f10]
[bt] (4) /work/projects/Project00755/envs/mxnet_gpu/lib/python3.5/site-packages/mxnet/libmxnet.so(+0x2ab74d4) [0x2aedfb6684d4]
[bt] (5) /work/projects/Project00755/envs/mxnet_gpu/lib/python3.5/site-packages/mxnet/libmxnet.so(+0x2a97b06) [0x2aedfb648b06]
[bt] (6) /work/projects/Project00755/envs/mxnet_gpu/lib/python3.5/site-packages/mxnet/libmxnet.so(+0x2f640d5) [0x2aedfbb150d5]
[bt] (7) /shared/apps/gcc/4.9.4/lib64/libstdc++.so.6(+0xc9090) [0x2aee34139090]
[bt] (8) /lib64/libpthread.so.0(+0x7e25) [0x2aedc9953e25]
[bt] (9) /lib64/libc.so.6(clone+0x6d) [0x2aedca36934d]
`

Any idea why this error happens? 

Thanks :)

rahul003:
[...]

PR #11910, opened 2018-07-27
Improving documentation and error messages for Async distributed training with Gluon

rahul003:
## Description ##
When using asynchronous mode of distributed training, updater has to be provided to the server. Since Symbolic mode of training always does updates on server, it generally has no issue. But Gluon provides the flexibility to determine whether to update weights on server or worker using `update_on_kvstore`. Also the default value for this is false, i.e. to perform weight updates on worker. This results in crash when using async mode with Gluon as this person realized https://github.com/apache/incubator-mxnet/issues/11855. We need to prevent users from running into this.

### Essentials ###

### Changes ###
- [x] add error message if user sets the variable to false explicitly
- [x] document the behavior of kvstore
- [x] added nightly test for async_dist kvstore

### DIFF
--- a/docs/faq/distributed_training.md
+++ b/docs/faq/distributed_training.md
@@ -76,7 +76,17 @@ to see an example usage.
 ### Updating weights
 KVStore server supports two modes, one which aggregates the gradients and updates the weights using those gradients, and second where the server only aggregates gradients. In the latter case, when a worker process pulls from kvstore, it gets the aggregated gradients. The worker then uses these gradients and applies the weights locally. 
 
-When using Gluon there is an option to choose between these modes by passing `update_on_kvstore` variable when you create the [Trainer](https://mxnet.incubator.apache.org/versions/master/api/python/gluon/gluon.html#mxnet.gluon.Trainer) object. 
+When using Gluon there is an option to choose between these modes by passing `update_on_kvstore` variable when you create the [Trainer](https://mxnet.incubator.apache.org/versions/master/api/python/gluon/gluon.html#mxnet.gluon.Trainer) object like this:
+
+```
+trainer = gluon.Trainer(net.collect_params(), optimizer='sgd',
+                        optimizer_params={'learning_rate': opt.lr,
+                                          'wd': opt.wd,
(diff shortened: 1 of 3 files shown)


rahul003:
@eric-haibin-lin

docs/faq/distributed_training.md
```diff
@@ -73,6 +73,13 @@ These can be passed as arguments to the iterator.
 You can look at [example/gluon/image_classification.py](https://github.com/apache/incubator-mxnet/blob/master/example/gluon/image_classification.py)
 to see an example usage.
 
+### Updating weights
+KVStore server supports two modes, one which aggregates the gradients and updates the weights using those gradients, and second where the server only aggregates gradients. In the latter case, when a worker process pulls from kvstore, it gets the aggregated gradients. The worker then uses these gradients and applies the weights locally. 
+
+When using Gluon there is an option to choose between these modes by passing `update_on_kvstore` variable when you create the [Trainer](https://mxnet.incubator.apache.org/versions/master/api/python/gluon/gluon.html#mxnet.gluon.Trainer) object.
```
sandeep-krishnamurthy: Example code snippet will be very easy for reader  here.

python/mxnet/gluon/trainer.py
```diff
@@ -187,6 +187,11 @@ def _init_kvstore(self):
             arg_arrays = {param.name: param.data(self._contexts[0]) for param in self._params}
             kvstore, update_on_kvstore = _create_kvstore(config['kvstore'], len(self._contexts),
                                                          arg_arrays)
+            if kvstore and 'async' in kvstore.type and config['update_on_kvstore'] is not None\
```
sandeep-krishnamurthy: If we are forcing the user to set this param, why don't we set it inside the function itself as default value?

rahul003: If the user does not set that variable explicitly (default way), then I set it to the right value. If the user explicitly sets it to false, then raised the error.

docs/faq/distributed_training.md
rahul003: ok

PR #11923, opened 2018-07-28
[MXAPPS-581] Fixes for broken Straight Dope tests.

vishaalkapoor:
This change fixes a few broken notebooks in the Straight Dope tests. It's a follow up to https://github.com/apache/incubator-mxnet/pull/11814.

* Update relative paths pointing to the data directory to point to the
  correct place in the testing temporary folder.

* Enable the notebooks that were previously broken because of relative
  file paths not pointing to the correct place.

* Move some notebooks we do not plan to test to the whitelist. These
  notebooks are not published in the Straight Dope book.

* Clean-up: Convert print statements to info/warn/error logging
  statements. Add some logging statements for better status.


hasanmua:
Hasan089

PR #11928, opened 2018-07-30
Generalized reshape_like operator

sbodenstein:
This PR implements [this proposal](https://discuss.mxnet.io/t/reshaping-broadcasting-without-hardcoding-target-dimensions/851/5) for generalizing the `reshape_like` operator, so that it has more flexible handling of which dimensions of the input and target tensors are used in the reshaping process.

There are a few differences from the proposal in the previous discussion. For example, to be consistent with the inputs names of `reshape_like` (`lhs` and `rhs` instead of `src` and `target`), the parameters were called `lhs_begin` rather than `src_begin` etc. 

I've also added tests for `reshape_like`, which didn't exist before, along with better documentation.

@ThomasDelteil, @taliesinb


src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -350,10 +350,108 @@ NNVM_REGISTER_OP(_identity_with_attr_like_rhs)
...
 .add_argument("rhs", "NDArray-or-Symbol", "Second input.");
 
+void ReshapeLikeRangeCanonicalize(int ndims, const char *side, int begin,
+                                  const dmlc::optional<int> &end, int *cbegin,
+                                  int *cend) {
+  *cbegin = begin;
+  if (*cbegin < 0)
+    *cbegin += ndims;
+
+  if (!static_cast<bool>(end)) {
```
szha: you can use has_value for better readability.

src/operator/tensor/elemwise_unary_op.h
```diff
@@ -476,6 +476,31 @@ void HardSigmoidBackward(const nnvm::NodeAttrs& attrs,
...
+  dmlc::optional<int> lhs_end, rhs_end;
+  DMLC_DECLARE_PARAMETER(ReshapeLikeParam) {
+    DMLC_DECLARE_FIELD(lhs_begin).set_default(0).describe(
+        "Defaults to 0. "
+        "The beginning index along which the lhs dimensions are to be "
+        "reshaped. Supports negative indices.");
+    DMLC_DECLARE_FIELD(lhs_end)
+        .set_default(dmlc::optional<int>())
+        .describe("Defaults to None. The ending index to be used, "
+                  "The ending index along which the lhs dimensions are to be "
```
taliesinb: This describe comment seems to have a spurious sentence fragment in it.

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -350,10 +350,108 @@ NNVM_REGISTER_OP(_identity_with_attr_like_rhs)
...
+      << "shape " << oshape << " because they have different "
+      << "size.";
+  SHAPE_ASSIGN_CHECK(*out_attrs, 0, oshape);
+  return true;
+}
+
+DMLC_REGISTER_PARAMETER(ReshapeLikeParam);
 NNVM_REGISTER_OP(reshape_like)
-.describe("Reshape lhs to have the same shape as rhs.")
+.describe(R"code(Reshape `lhs` to have the same shape as `rhs`.
```
ThomasDelteil: I would alter this line to say something more along the line:
"Reshape some or all dimensions of `lhs` to have the same shape as some or all dimensions of `rhs`"

taliesinb:
Please see this 'sister' issue that designs what is effectively the dual of this feature: https://discuss.mxnet.io/t/proposal-new-merge-dims-reshaping-op/1524

src/operator/tensor/elemwise_unary_op.h
sbodenstein: Fixed.

src/operator/tensor/elemwise_unary_op_basic.cc
sbodenstein: Agreed, much better. Changed.

src/operator/tensor/elemwise_unary_op_basic.cc
sbodenstein: Ok, that is cleaner. Fixed.

Btw: the reason I used this casting method because it was done [here](https://github.com/apache/incubator-mxnet/blob/51f650e0bf3b905fec4aebfc873c1c56eac61536/src/operator/tensor/matrix_op-inl.h#L1125). Should this be changed as well?

src/operator/tensor/elemwise_unary_op.h
```diff
@@ -476,6 +476,31 @@ void HardSigmoidBackward(const nnvm::NodeAttrs& attrs,
   });
 }
 
+struct ReshapeLikeParam : public dmlc::Parameter<ReshapeLikeParam> {
+  int lhs_begin, rhs_begin;
```
szha: I wonder if these two parameters should be made optional too, given that reshape_like op has been around for 10 months.

sbodenstein: They are optional in the sense that they have default values `DMLC_DECLARE_FIELD(lhs_begin).set_default(0)` that match the old behaviour if not explicitly specified. I thought `dmlc::optional<int>` was simply for the case where you had to handle the `None`-value case, which is not necessary to support for `lhs_begin` and `rhs_begin`. Note that the same is done for `slice_axis`, `begin` is an `int` and `end` is `dmlc::optional<int>`.

Or am I misunderstanding something?

szha: Considering the case where the parameters may be in some serialized format, it may be necessary to support null values to ensure compatibility there too.

taliesinb: @szha Can I clarify something? Maybe I am misunderstanding. Are you saying that it is required to make new parameters on existing layers optional<...> for backward compatibility reasons, *even* if they have defaults?

szha: @taliesinb I think it doesn't hurt in this case and was suggesting that we lean on the safer side. There might be cases where the default value filling on the frontend fails, such as deserializing a graph.

taliesinb: @szha ok. so it sounds like there isn't a particular policy at the moment about how to add new parameters to existing layers. do you mind if i ask on the mailing list about what such a policy should be, just for the next time this happens?

szha: of course not. good idea :)

sbodenstein: @szha: I've changed `lhs_begin` and `rhs_begin` to use optional (and added tests for this case). Can we merge?

PR #11942, opened 2018-07-31
[MXNET-749] Bug fixes in control flow operators

junrushao1994:
## Description ##
This PR fixes some bugs for control flow operators `mx.sym.contrib.while_loop` and `mx.sym.contrib.cond`.

Takeaways:
- if now I understand correctly, we should call `MXSymbolCutSubgraph` to capture free variables of a closure before calling `MXSymbolGetInputSymbols`. I guess we should document this so that others won't be confused.
- `symbol._internal.something` always return a single Symbol, regardless of how many outputs it actually has.

### Essentials ###

### Changes ###
- [x] Bug fixes

## Comments ##
- See takeaways

### DIFF
--- a/python/mxnet/symbol/contrib.py
+++ b/python/mxnet/symbol/contrib.py
@@ -486,12 +486,12 @@ def while_loop(cond, func, loop_vars, max_iterations=None, name="while_loop"):
         input_id_to_loc = {}    # Dict[int, int], given id(sym), input_id_to_loc maps it
                                 # to a `loc`, where inputs[loc] = sym
         for graph in graphs:
-            # input_syms: all inputs to the `graph`
-            name_to_input_syms = {sym.name: sym for sym in _get_graph_inputs(graph)}
             # some loop_vars are inputs to `graph`, some are not
             name_to_loop_vars = {sym.name: sym for sym in loop_vars}
             # other inputs to `graph` created by cut_graph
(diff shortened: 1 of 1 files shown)


junrushao1994:
@zheng-da @piiswrong Could you help review this? Thank you guys!

junrushao1994:
@zheng-da And I am not sure if my understanding is correct, but I guess another thing we should discuss is: is the `SubgraphSanityCheck` we did in [NNVM](https://github.com/dmlc/tvm/blob/142c7da392a232e5610b6d7bcfd5b40f9c4dd027/nnvm/src/core/graph.cc#L24), actually too restrictive?

For example, there is a huge graph, containing two subgraphs `A` and `B`, and `A` and `B` has there own subgraphs, say `A.0` and `B.0`. Some intermediate results are shared across several subgraphs. If `CutSubgraph` fails to copy-and-split them into different subgraphs, then `SubgraphSanityCheck` will raise an error. However, in theory, `CutSubgraph` should do the copy-and-split stuff, but it is not working on my case. I suspect there is more bugs, either in control flow operators, or in subgraph cut. Could we discuss sometime?

zheng-da:
To make sure cutting subgraphs works correctly, I think we should give each subgraph a unique name. Maybe we aren't doing it correctly.

junrushao1994:
@zheng-da I tried to comment out the sanity check, and everything seems to work

junrushao1994:
This fix is pretty simple, I think we should merge this so that it doesn't block subsequent work. Thanks!

Issue #11871, opened 2018-07-24
Proposal: Generalize broadcast_like


sbodenstein:
Now that `broadcast_like` has been merged (#11820), I would like to discuss the extension discussed [here](https://github.com/apache/incubator-mxnet/pull/11820#issuecomment-406445064) by @taliesinb. Quoting it:

> There is a generalization that would be extremely useful for this operator to have. The generalization is very similar to one that was discussed at https://discuss.mxnet.io/t/reshaping-broadcasting-without-hardcoding-target-dimensions/851/6 (you can skip to the last 4 comments, the thread contains an irrelevant proposal although the motivation is relevant).
> 
> In short, the generalization would allow only specific dimensions to be copied from the 'other' tensor. For example:
> 
> input.shape = (1, 2, 1, 3)
> other.shape = (5, 6, 7, 8)
> output = broadcast_like(input, other, input_axes:(0,2), other_axes:(1,3))
> output.shape = (6, 2, 8, 3)
> In other words, what's happening here is that the you can pick exactly which axes of the other tensor you want to use to "fill in" axes of the input tensor. This is how broadcast_axes works, except instead of providing the values via a size parameter, you are providing them from specific axes in the other tensor.
> 
> The reason this is so valuable is that it is common to have another tensor that contains the dimension you want to broadcast amongst a set of irrelevant dimensions. There is simply no other way of "extracting" the relevant dimension from elsewhere in the net, so currently you have to hardcode that dimension into a parameter list, which forces expensive workarounds like bucketing where otherwise cheap reshaping would work to make a net that is compatible with multiple sequence lengths, for example.
> 
> The current behavior of broadcast_like in the PR would be consistent with this generalization if the default value of input_axis is the empty tuple, which means "all axes".

I would like to open this design up for discussion.

@ifeherva

ifeherva:
Sounds reasonable to me.

rondogency:
@sbodenstein I was working on broadcast_like briefly a few weeks ago (#11443 ). This specs looks very reasonable. I am just wondering if you would be comfortable that I work on top of @ifeherva new broadcast_like, and generalize it as discussed above?

sbodenstein:
@rondogency: @ifeherva had also expressed interest to implement it, but whoever has more time this week should probably do it (we need it for next Monday, so I would add it neither of you has time this week).

rondogency:
[...]

PR #11984, opened 2018-08-02
Generalized broadcast_like operator

ifeherva:
## Description ##
This is the generalized implementation of the broadcast_like operator, discussed here:
https://github.com/apache/incubator-mxnet/issues/11871

### DIFF
--- a/src/operator/tensor/broadcast_reduce_op.h
+++ b/src/operator/tensor/broadcast_reduce_op.h
@@ -362,8 +362,10 @@ inline bool BroadcastLikeShape(const nnvm::NodeAttrs& attrs,
   TShape& lhs_shape = (*in_attrs)[0];
   TShape& rhs_shape = (*in_attrs)[1];
 
-  if ((lhs_shape.ndim() == 0) || (lhs_shape.ndim() == 0)) return false;
-    CHECK_EQ(lhs_shape.ndim(), rhs_shape.ndim())
+  if ((lhs_shape.ndim() == 0) || (lhs_shape.ndim() == 0)) {
+    return false;
+  }
+  CHECK_EQ(lhs_shape.ndim(), rhs_shape.ndim())
     << "Operand of shape " << lhs_shape << " cannot be broadcasted to " << rhs_shape;
 
   const BroadcastLikeParam& param = nnvm::get<BroadcastLikeParam>(attrs.parsed);
--- a/src/operator/tensor/broadcast_reduce_op.h
+++ b/src/operator/tensor/broadcast_reduce_op.h
@@ -148,12 +148,12 @@ struct BroadcastToParam : public dmlc::Parameter<BroadcastToParam> {
 };
 
 struct BroadcastLikeParam : public dmlc::Parameter<BroadcastLikeParam> {
-  TShape lhs_axes;
-  TShape rhs_axes;
+  dmlc::optional<TShape> lhs_axes;
+  dmlc::optional<TShape> rhs_axes;
   DMLC_DECLARE_PARAMETER(BroadcastLikeParam) {
-    DMLC_DECLARE_FIELD(lhs_axes).set_default(TShape())
+    DMLC_DECLARE_FIELD(lhs_axes)
       .describe("Axes to perform broadcast on in the first input array");
-    DMLC_DECLARE_FIELD(rhs_axes).set_default(TShape())
+    DMLC_DECLARE_FIELD(rhs_axes)
       .describe("Axes to copy from the second input array");
   }
 };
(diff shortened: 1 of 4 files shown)


src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -137,6 +137,17 @@ struct BroadcastToParam : public dmlc::Parameter<BroadcastToParam> {
   }
 };
 
+struct BroadcastLikeParam : public dmlc::Parameter<BroadcastLikeParam> {
+  TShape input_axes;
+  TShape other_axes;
+  DMLC_DECLARE_PARAMETER(BroadcastLikeParam) {
+    DMLC_DECLARE_FIELD(input_axes).set_default(TShape())
```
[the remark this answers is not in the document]
ifeherva: Well, that's what you proposed :) Can change of course.

sbodenstein:
I find this design counterintuitive:
```
s=mx.sym.broadcast_like(mx.sym.Variable("lhs"), mx.sym.Variable("rhs"), other_axes=(),input_axes=())
s.infer_shape(lhs=(1,7,1,1), rhs=(3,7,9,4))[1]
[(3, 7, 9, 4)]
```
So `input_axes=()` is equivalent to `input_axes=(0, 2, 3)`. Surely an empty tuple should be interpreted as the identity? You could make `other_axes` and `input_axes` optional with default `None`.

ifeherva:
Do you get an error when not passing anything to other_axes and input_axes?

ifeherva:
@sbodenstein There are tests for the empty case.

However, I think passing nothing to the axes params for me means "take the whole" not "take nothing". I find that counterintuitive.

sbodenstein:
Ok, I think we are talking about different cases. Obviously, this PR shouldn't change this case: `mx.sym.broadcast_like()`. It should indeed "take the whole". This case is also being tested for.

But the example I gave was not this, but the case where axes are specified with empty tuples as values: `mx.sym.broadcast_like(lhs_axes=(), rhs_axes=())`. I don't see any tests for this case. And the behaviour seems strange: `lhs_axes=(5,3)` means take two axes, `lhs_axes=(5,)` means take one axis, and `lhs_axes=()` mens take all axes. I'm also happy if it just fails for the empty tuple case.

sbodenstein:
@ifeherva: just checking whether you agree with my comment above.

sbodenstein:
Can't you use `dmlc::optional`?

ifeherva:
Good point, see last commit.

sbodenstein:
Great! Can you just add a test for the empty tuple case?

ifeherva:
Already did:
https://github.com/apache/incubator-mxnet/pull/11984/files#diff-69757562d07268150de8b369ff5b6b61R569

sbodenstein:
Ah, apologies didn't see that. Is it ready to be merged?

ifeherva:
I think so.

sbodenstein:
@szha: could you review this PR?

taliesinb:
@anirudh2290 it would be fantastic if we could get some idea of when a review will be possible.

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -340,20 +351,60 @@ inline bool BroadcastLikeShape(const nnvm::NodeAttrs& attrs,
...
+    return false;
+  }
+
+  const BroadcastLikeParam& param = nnvm::get<BroadcastLikeParam>(attrs.parsed);
+  TShape oshape;
+
+  // lhs or rhs or both params were not specified
+  if (!param.lhs_axes.has_value() || !param.rhs_axes.has_value()) {
+    CHECK_EQ(lhs_shape.ndim(), rhs_shape.ndim())
     << "Operand of shape " << lhs_shape << " cannot be broadcasted to " << rhs_shape;
```
szha: indentation

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -340,20 +351,60 @@ inline bool BroadcastLikeShape(const nnvm::NodeAttrs& attrs,
...
+    }
+  } else {
+    auto lhs_axes = param.lhs_axes.value();
+    auto rhs_axes = param.rhs_axes.value();
+
+    CHECK(rhs_axes.ndim() == lhs_axes.ndim())
+    << "Input_axis and other_axis size does not match";
+
+    CHECK(lhs_axes.ndim() > 0)
+    << "Empty axes tuple is not allowed";
```
szha: indentation

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -340,20 +351,60 @@ inline bool BroadcastLikeShape(const nnvm::NodeAttrs& attrs,
...
+
+    oshape = TShape(lhs_shape);
+    for (index_t i = 0; i < lhs_axes.ndim(); ++i) {
+      auto copyfrom = lhs_axes[i];
+      if (copyfrom < 0) {
+        copyfrom =  lhs_shape.ndim() + copyfrom;
+      }
+      CHECK(copyfrom >= 0 && copyfrom < oshape.ndim())
+      << "Invalid dimension specified in lhs_axes: " << lhs_axes[i];
```
szha: and here

szha: also, either print the whole lhs_axes, or print `i` to make the error message informative.

sbodenstein:
@szha: is it ready to be merged? (I think those indentation and other issues were resolved)

PR #12019, opened 2018-08-03
[MXNET-753] Fallback when using non-MKLDNN supported operators

azai91:
## Description ##
-convert all mkldnn special format to default in using non-mkldnn operator

### Essentials ###

### Changes ###
- [x] add attribute to all mkldnn operators to identify mkldnn
- [x] convert inputs to default array if using fcomputeex operator that is not mkldnn

## Comments ##

### DIFF
--- a/src/executor/attach_op_execs_pass.cc
+++ b/src/executor/attach_op_execs_pass.cc
@@ -159,7 +159,7 @@ class StatefulComputeExExecutor : public OpExecutor {
     op_ctx.run_ctx = rctx;
 #if MXNET_USE_MKLDNN == 1
     InvalidateOutputs(out_array, req);
-    in_array = CreateInputsInputs(in_array);
+    in_array = CreateDefaultInputs(in_array);
     fcompute_(state_, op_ctx, in_array, req, out_array);
     return;
 #endif
@@ -231,7 +231,7 @@ class FComputeExExecutor : public OpExecutor {
     InvalidateOutputs(out_array, req);
     const auto is_mkldnn = Op::GetAttr<bool>("TIsMKLDNN");
     if (!is_mkldnn.get(attrs_.op, false)) {
-      in_array = CreateInputsInputs(in_array);
+      in_array = CreateDefaultInputs(in_array);
       fcompute_(attrs_, op_ctx, in_array, req, out_array);
       return;
     }
--- a/src/executor/attach_op_execs_pass.cc
+++ b/src/executor/attach_op_execs_pass.cc
@@ -159,7 +159,8 @@ class StatefulComputeExExecutor : public OpExecutor {
     op_ctx.run_ctx = rctx;
 #if MXNET_USE_MKLDNN == 1
     InvalidateOutputs(out_array, req);
-    fcompute_(state_, op_ctx, InvalidateInputs(in_array), req, out_array);
+    in_array = CreateInputsInputs(in_array);
+    fcompute_(state_, op_ctx, in_array, req, out_array);
     return;
 #endif
     fcompute_(state_, op_ctx, in_array, req, out_array);
@@ -230,7 +231,8 @@ class FComputeExExecutor : public OpExecutor {
     InvalidateOutputs(out_array, req);
     const auto is_mkldnn = Op::GetAttr<bool>("TIsMKLDNN");
     if (!is_mkldnn.get(attrs_.op, false)) {
-      fcompute_(attrs_, op_ctx, InvalidateInputs(in_array), req, out_array);
+      in_array = CreateInputsInputs(in_array);
+      fcompute_(attrs_, op_ctx, in_array, req, out_array);
       return;
     }
 #endif
--- a/src/executor/attach_op_execs_pass.cc
+++ b/src/executor/attach_op_execs_pass.cc
@@ -159,7 +159,7 @@ class StatefulComputeExExecutor : public OpExecutor {
     op_ctx.run_ctx = rctx;
 #if MXNET_USE_MKLDNN == 1
     InvalidateOutputs(out_array, req);
-    CreateDefaultInputs(in_array, in_array_fallback);
+    CreateDefaultInputs(in_array, &in_array_fallback);
     fcompute_(state_, op_ctx, in_array_fallback, req, out_array);
     return;
 #endif
@@ -232,7 +232,7 @@ class FComputeExExecutor : public OpExecutor {
     // TODO(alex): (MXNET-847) Remove this fallback feature after subgraph implemented
     const auto is_mkldnn = Op::GetAttr<bool>("TIsMKLDNN");
     if (!is_mkldnn.get(attrs_.op, false)) {
-      CreateDefaultInputs(in_array, in_array_fallback);
+      CreateDefaultInputs(in_array, &in_array_fallback);
       fcompute_(attrs_, op_ctx, in_array_fallback, req, out_array);
       return;
     }
--- a/src/executor/attach_op_execs_pass.cc
+++ b/src/executor/attach_op_execs_pass.cc
@@ -159,7 +159,7 @@ class StatefulComputeExExecutor : public OpExecutor {
     op_ctx.run_ctx = rctx;
 #if MXNET_USE_MKLDNN == 1
     InvalidateOutputs(out_array, req);
-    in_array_fallback = CreateDefaultInputs(in_array);
+    CreateDefaultInputs(in_array, in_array_fallback);
     fcompute_(state_, op_ctx, in_array_fallback, req, out_array);
     return;
 #endif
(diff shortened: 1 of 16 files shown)


pengzhao-intel:
@azai91 Very nice improvements!

ZhennanQin:
@azai91  Good improvement before subgraph ready. I'm not familiar with engine part, can this change cover all kinds of executing scenarios, like any combination of NaiveEngine, ThreadedEngine, symbolic, gluon, hybridized gluon?

src/operator/nn/mkldnn/mkldnn_base-inl.h
```diff
@@ -351,6 +351,13 @@ static inline void InvalidateOutputs(const std::vector<NDArray> &arrs,
   }
 }
 
+static inline std::vector<NDArray> InvalidateInputs(const std::vector<NDArray> &arrs) {
```
eric-haibin-lin: The name is confusing and suggests mutation on the inputs.

src/executor/attach_op_execs_pass.cc
```diff
@@ -226,6 +228,11 @@ class FComputeExExecutor : public OpExecutor {
     op_ctx.run_ctx = rctx;
 #if MXNET_USE_MKLDNN == 1
     InvalidateOutputs(out_array, req);
+    const auto is_mkldnn = Op::GetAttr<bool>("TIsMKLDNN");
+    if (!is_mkldnn.get(attrs_.op, false)) {
+      fcompute_(attrs_, op_ctx, InvalidateInputs(in_array), req, out_array);
```
eric-haibin-lin: Is it thread safe to modify inputs?

eric-haibin-lin: nvm

tests/python/mkl/test_mkldnn.py
```diff
@@ -259,6 +259,50 @@ def check_fullyconnected_training(stype):
...
+        def forward(self, is_train, req, in_data, out_data, aux):
+            print(in_data[0])
+            self.assign(out_data[0], req[0], in_data[0])
+
+        def backward(self, req, out_grad, in_data, out_data, in_grad, aux):
+            self.assign(in_grad[0], req[0], out_grad)
+
+    data = mx.symbol.Variable('data')
+    conv = mx.sym.Convolution(data=data, kernel=(5, 5), pad=(1, 1), stride=(1,1), num_filter=8, name="conv", no_bias=True)
+    mlp = mx.symbol.Custom(name='custom', data=conv, op_type='custom')
```
apeforest: nit: renaming mlp to custom or custom_op

src/operator/tensor/elemwise_unary_op.h
```diff
@@ -299,7 +299,11 @@ class UnaryOp : public OpBase {
         }
         break;
       case kWriteInplace:
+// cannot check if ptrs are the same for MKLDNN because we may have
+// created copies of input when reordering. WriteInPlace will still write to original array
+#if MXNET_USE_MKLDNN != 1
```
apeforest: change to #if MXNET_USE_MKLDNN == 0

src/executor/attach_op_execs_pass.cc
```diff
@@ -40,6 +40,11 @@ const OperatorProperty* OpPropGetOpProperty(const NodeAttrs& attrs);
 
 namespace exec {
 
+class MKLDNNOpExecutor : public OpExecutor {
+ protected:
+  std::vector<NDArray> in_array_fallback;
```
apeforest: Why do we need an extra copy rather than just editing the in_array directly?

azai91: there is a race condition if we attempt to reorder the `read_var` in place. other operators may be trying to read from it the same time (since they are expected to be read only consts)

src/operator/nn/activation.cc
```diff
@@ -182,6 +182,7 @@ The following activation functions are supported:
 })
 .set_attr<FCompute>("FCompute<cpu>", ActivationCompute<cpu>)
 #if MXNET_USE_MKLDNN == 1
+.set_attr<bool>("TIsMKLDNN", true)
```
apeforest: Is there a more generic way to add this instead of doing this for all operators? What if we later we add new operators, should we document this somewhere?

azai91: if only needs to be added to MKLDNN operators. this fix was a temporary solution while we get the subgraph feature implemented. we weighed the pros / cons of waiting to release a stable MKLDNN stable with this hack or waiting another month for subgraph to be introduced (possibly with it's own bugs) and decided we would use this short term solution.

apeforest: Please mark this TODO and create a JIRA ticket to remove this later after MKLDNN support is released.

azai91: added TODO to opexecuter

src/executor/attach_op_execs_pass.cc
apeforest: If that's the case, why do we even bother to create a subclass with this extra data member?

src/executor/attach_op_execs_pass.cc
```diff
@@ -153,12 +158,15 @@ class StatefulComputeExecutor : public StorageFallbackOpExecutor {
...
 
 // stateful compute_ex executor
-class StatefulComputeExExecutor : public OpExecutor {
+class StatefulComputeExExecutor : public MKLDNNOpExecutor {
  public:
   void Run(RunContext rctx, bool is_gpu) override {
     op_ctx.run_ctx = rctx;
 #if MXNET_USE_MKLDNN == 1
     InvalidateOutputs(out_array, req);
+    in_array_fallback = CreateDefaultInputs(in_array);
```
apeforest: Can we just define the array here instead?

src/executor/attach_op_execs_pass.cc
```diff
@@ -153,12 +158,15 @@ class StatefulComputeExecutor : public StorageFallbackOpExecutor {
 
 
 // stateful compute_ex executor
-class StatefulComputeExExecutor : public OpExecutor {
+class StatefulComputeExExecutor : public MKLDNNOpExecutor {
```
apeforest: Making this class derived from MKLDNNOpExecutor while it's only valid when MXNET_USE_MKLDNN == 1. This violates the inheritance class is-a relationship

src/executor/exec_pass.h
```diff
@@ -86,6 +86,9 @@ class OpExecutor {
   virtual OpStatePtr state() const {
     return OpStatePtr();
   }
+  
+ protected:
+  std::vector<NDArray> in_array_fallback;
```
apeforest: Do we really need this as a class data member? Or can we just declare it as a local variable when it is created because  I don't see it used anywhere else in the flow.

src/executor/attach_op_execs_pass.cc
```diff
@@ -159,6 +159,9 @@ class StatefulComputeExExecutor : public OpExecutor {
     op_ctx.run_ctx = rctx;
 #if MXNET_USE_MKLDNN == 1
     InvalidateOutputs(out_array, req);
+    in_array_fallback = CreateDefaultInputs(in_array);
```
apeforest: Can we just declare std::vector<NDArray> in_array_fallback here?

azai91: we need the fallback arrays to stay in memory or else we segfault.

apeforest: Cannot see why that is the case. But creating a data member is not a good solution to this. I am not very comfortable of intruding a new data member to this class without much justification. If this PR is only for a temporary workaround I am okay to approve. But please create a JIRA ticket to remove this hack later.

src/executor/exec_pass.h
```diff
@@ -86,6 +86,9 @@ class OpExecutor {
   virtual OpStatePtr state() const {
     return OpStatePtr();
   }
+
+ protected:
+  std::vector<NDArray> in_array_fallback;
```
apeforest: Please make sure to add TODO with JIRA number to remove this hack.

src/operator/nn/mkldnn/mkldnn_base-inl.h
```diff
@@ -356,6 +356,17 @@ static inline void InvalidateOutputs(const std::vector<NDArray> &arrs,
   }
 }
 
+static inline std::vector<NDArray> CreateDefaultInputs(const std::vector<NDArray> &arrs) {
```
apeforest: Also add TODO to remove this unnecessary function when final solution is implemented.

src/operator/nn/mkldnn/mkldnn_base-inl.h
```diff
@@ -356,6 +356,18 @@ static inline void InvalidateOutputs(const std::vector<NDArray> &arrs,
   }
 }
 
+// TODO(alexzai): (MXNET-856) Remove helper function after subgraph feature added
+static inline std::vector<NDArray> CreateDefaultInputs(const std::vector<NDArray> &arrs) {
```
anirudh2290: can we pass a pointer to a vector as an argument instead of returning a vector.

pengzhao-intel:
@azai91 @zheng-da this PR caused the performance regression, such as resnet 50 v1.
Please help fix the issue ASAP.

PR #12025, opened 2018-08-03
disable opencv threading for forked process

zhreshold:
## Description ##
After 3.4, opencv seems to be enabling threading automatically. 
This PR disable opencv threading for forked process, to alleviate thread contention problem.

This PR improved image loading performance from 160 sample/s to 450 sample/s with gluon data loader when `num_workers=4`

example code with real data:

```python

import mxnet as mx
import gluoncv as gcv
from tqdm import tqdm

class NaiveTransform(object):
    def __init__(self, width, height):
        self._width = width
        self._height = height

    def __call__(self, src, label):
        src = mx.image.imresize(src, self._width, self._height, interp=1)
        return src, label

dataset = gcv.data.COCODetection(splits=('instances_train2017',), skip_empty=True, use_crowd=False)
dataloader = mx.gluon.data.DataLoader(dataset.transform(NaiveTransform(512, 512)),
                                      batch_size=1, shuffle=True, num_workers=4)
with tqdm(total=len(dataset)) as pbar:
    for batch in dataloader:
        mx.nd.waitall()
        pbar.update(1)
```

### Essentials ###

### DIFF
--- a/src/initialize.cc
+++ b/src/initialize.cc
@@ -26,6 +26,9 @@
 #include <dmlc/logging.h>
 #include <mxnet/engine.h>
 #include "./engine/openmp.h"
+#if MXNET_USE_OPENCV
+#include <opencv2/opencv.hpp>
+#endif  // MXNET_USE_OPENCV
 
 namespace mxnet {
 #if MXNET_USE_SIGNAL_HANDLER && DMLC_LOG_STACK_TRACE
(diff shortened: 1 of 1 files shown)


rahul003:
I'm concerned if this could affect the speed of the record IO pipeline. Could you verify the speeds for that using this gluoncv script at the largest batch size of 256 per GPU? https://github.com/dmlc/gluon-cv/blob/master/scripts/classification/imagenet/train_imagenet.py

zhreshold:
@rahul003 Using record IO won't trigger the multiprocess worker.

zhreshold:
@rahul003 Right, FYI, I have tested the recordIO speed, with and w/o this PR, there's no sign of degradation.

PR #12031, opened 2018-08-04
[MXNET-1107] Fix CPUPinned unexpected behaviour

ctcyang:
## Description ##
There is unexpected behavior with MXNet Context cpu_pinned for nonzero values of device.
This PR fixes this unexpected behavior.

### Essentials ###

### Changes ###
- [x] Fix unexpected behavior with cpu_pinned Context
- [x] A class mxnet::common::cuda::SetDevice that collects what device was active before cudaSetDevice was called, and restores it to this device once the MXNet has done what it needs to do. This is done when the SetDevice object goes out of scope

## Comments ##
On master branch, if you do:
```
>>> import mxnet as mx
>>> data = mx.nd.zeros((2,3), ctx=mx.Context('cpu_pinned', 7))
>>> print(data)


[[0. 0. 0.]
 [0. 0. 0.]]
<NDArray 2x3 @cpu_pinned(7)>
```

This looks right. However, `nvidia-smi` tells me that this data is actually located on GPU 0:

```
$ nvidia-smi
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|    0    109140      C   python3                                      524MiB |
+-----------------------------------------------------------------------------+
```

To fix this unexpected behaviour, I changed four things:
1) call cudaSetDevice and set current device to GPU X we want to allocate on even for kCPUPinned case (src/storage/storage.cc)
2) before, Context::real_dev_id returns the GPU id for kGPU, and 0 for both kCPU and kCPUPinned even though it stores the dev_id that the user passed into the constructor as a private member (include/mxnet/base.h)
3) add guard to every place cudaSetDevice is called in MXNet C++ code. This guard is implemented using the class mxnet::common::cuda::SetDevice. The purpose of this guard is to prevent race conditions on which device was set first. Other GPU libraries have something similar e.g. 
https://github.com/uber/horovod/blob/master/horovod/torch/cuda_util.cc


include/mxnet/base.h
```diff
@@ -156,7 +156,7 @@ struct Context {
    * \brief Returns dev_id for kGPU, 0 otherwise
```
haojin2: Should also fix the documentation here?

ctcyang: Thanks!

src/common/cuda_utils.h
```diff
@@ -179,113 +283,38 @@ inline DType __device__ CudaMin(DType a, DType b) {
...
- * \param func Expression to call.
- *
- * It checks for cuRAND errors after invocation of the expression.
- */
-#define CURAND_CALL(func)                                       \
-  {                                                             \
-    curandStatus_t e = (func);                                  \
-    CHECK_EQ(e, CURAND_STATUS_SUCCESS)                          \
-        << "cuRAND: " << mxnet::common::cuda::CurandGetErrorString(e); \
+class SetDevice {
```
eric-haibin-lin: Rename to DeviceScope, add a member function SetDevice, and remove all occurrences of `cudaGetDevice`?

ctcyang:
Added [WIP] tag due to bug with `--kv-store dist_sync_device`.

stu1130:

pengzhao-intel:
@eric-haibin-lin Any update on this PR? 

It opens for a long time and we prefer to leverage the feature in this PR :)

Makefile
```diff
@@ -678,3 +678,4 @@ clean_all: clean
 ifneq ($(EXTRA_OPERATORS),)
 	-include $(patsubst %, %/*.d, $(EXTRA_OPERATORS)) $(patsubst %, %/*/*.d, $(EXTRA_OPERATORS))
 endif
+
```
apeforest: nit: remove blank line

src/kvstore/comm.h
```diff
@@ -723,6 +723,9 @@ class CommDevice : public Comm {
     int n = static_cast<int>(gpus.size());
     int enabled = 0;
     std::vector<int> p2p(n*n);
+
+    // Restores active device to what it was before EnableP2P
+    mxnet::common::cuda::SetDevice set_device;
     for (int i = 0; i < n; ++i) {
       cudaSetDevice(gpus[i]);
```
apeforest: As @eric-haibin-lin suggested, replace this with a class method by your RAII object SetDevice (before renaming)

src/kvstore/comm_tree.h
```diff
@@ -339,6 +339,7 @@ class CommDeviceTree : public CommDevice {
     int n = static_cast<int>(gpus.size());
     int enabled = 0;
     std::vector<int> p2p(n*n);
+    mxnet::common::cuda::SetDevice set_device;
     for (int i = 0; i < n; ++i) {
       cudaSetDevice(gpus[i]);
```
apeforest: same here

src/kvstore/kvstore_nccl.h
```diff
@@ -428,6 +428,7 @@ class KVStoreNCCL : public KVStoreLocal {
         mutate_vars.push_back(ptr(dst[i])->var());
     }
     Engine::Get()->PushSync([this](RunContext rctx) {
+        mxnet::common::cuda::SetDevice set_device;
         for (auto cur : nccl_data_) {
           CUDA_CALL(cudaSetDevice(cur.second.dev_id));
```
apeforest: same here

src/storage/storage.cc
```diff
@@ -51,7 +51,13 @@ class StorageImpl : public Storage {
   static void ActivateDevice(Context ctx) {
```
apeforest: Seems this method does not need to be static. It is only used in this class. Besides, making it static without a RAII class around is very dangerous.

apeforest: This method should not be static. It is not protected by any RAII object. Besides, it is not used by any other classes anyway.

src/kvstore/comm.h
apeforest: replace this with a class method by the SetDevice (renaming) object as @eric-haibin-lin suggested.

src/kvstore/comm_tree.h
apeforest: same here

src/kvstore/kvstore_nccl.h
apeforest: same here

src/storage/storage.cc
ctcyang: You raise a good point, but I think that should be addressed in a separate PR.

Makefile
ctcyang: Fixed

Issue #11967, opened 2018-08-01
pad layer only works for float32


sbodenstein:
Padding is a core structural op that should work for all array types, but fails for any type that is not `float32`:
```
import numpy as np
import mxnet as mx

s=mx.sym.pad(mx.sym.Variable("data", dtype="float64"), mode="edge", pad_width=(0,0,0,0,1,1,1,1))
d=mx.nd.ones((1,2,6,6), dtype="float64")
s.bind(mx.cpu(), {'data': d})

MXNetError: Error in operator pad13: [15:30:33] include/mxnet/operator.h:228: Check failed: in_type->at(i) == mshadow::default_type_flag || in_type->at(i) == -1 Unsupported data type 1
```

marcoabreu:
Hello Sebastian,

thanks a lot for noting. Do you think you're able to provide a patch for this?

Roshrini:
@sandeep-krishnamurthy Can you please add label: Operator, FeatureRequest

sbodenstein:
@marcoabreu: yes, I will by next week.

marcoabreu:
Great! Looking forward to it, thank you.

apeforest:
@sandeep-krishnamurthy please close this issue.

PR #12035, opened 2018-08-05
Pad Operator Type Support

sbodenstein:
This PR allows the `pad` operator to operate on any type. Fixes Issue #11967.

One potential issue: I've enabled this layer for integer types, for which gradients don't get computed. Is there a standard way of handling this?

### DIFF
--- a/tests/python/unittest/test_operator.py
+++ b/tests/python/unittest/test_operator.py
@@ -2907,13 +2907,9 @@ def test_roipooling():
 
 def check_pad_with_shape(shape, xpu, pad_width, mode, dtype="float64"):
     # bind with label
-    real_types = ["float16", "float32", "float64"]
     X = mx.symbol.Variable('X', dtype=dtype)
     Y = mx.symbol.Pad(data=X, mode=mode, pad_width=pad_width)
-    if dtype in real_types:
-        x = mx.random.uniform(-1, 1, shape, ctx=mx.cpu(), dtype=dtype).copyto(xpu)
-    else:
-        x = mx.nd.ones(shape, ctx=mx.cpu(), dtype=dtype).copyto(xpu)
+    x = mx.random.uniform(-1, 1, shape, ctx=mx.cpu(), dtype=dtype).copyto(xpu)
     # numpy result
     pad_grouped = list(zip(*[iter(list(pad_width))] * 2))
     np_out = np.pad(x.asnumpy(), pad_grouped, mode)
@@ -2925,26 +2921,25 @@ def check_pad_with_shape(shape, xpu, pad_width, mode, dtype="float64"):
     # compare numpy + mxnet
     assert_almost_equal(out, np_out)
     # grad check
-    if dtype in real_types:
-        check_numeric_gradient(Y, [x.asnumpy()], numeric_eps=1e-2, rtol=1e-2)
+    check_numeric_gradient(Y, [x.asnumpy()], numeric_eps=1e-2, rtol=1e-2)
 
 
 @with_seed()
 def test_pad():
-    ct = default_context()
+    ctx = default_context()
     shape1 = (2, 3, 3, 5)
     pad1 = (0, 0, 0, 0, 1, 2, 3, 4)
     shape2 = (2, 3, 3, 5, 4)
     pad2 = (0, 0, 0, 0, 1, 2, 3, 4, 3, 1)
     # note: this op doesn't support ints yet. Add tests when supported
-    test_types = ["float32", "float64", "float16"]
-    for d in test_types:
-        check_pad_with_shape(shape1, ct, pad1, 'constant', d)
-        check_pad_with_shape(shape1, ct, pad1, 'edge', d)
-        check_pad_with_shape(shape2, ct, pad2, 'constant', d)
-        check_pad_with_shape(shape2, ct, pad2, 'edge', d)
-        check_pad_with_shape(shape1, ct, pad1, 'reflect', d)
-        check_pad_with_shape(shape2, ct, pad2, 'reflect', d)
+    test_types = ["float16", "float32", "float64"]
+    for dtype in dtypes:
+        check_pad_with_shape(shape1, ctx, pad1, 'constant', dtype)
+        check_pad_with_shape(shape1, ctx, pad1, 'edge', dtype)
+        check_pad_with_shape(shape2, ctx, pad2, 'constant', dtype)
+        check_pad_with_shape(shape2, ctx, pad2, 'edge', dtype)
+        check_pad_with_shape(shape1, ctx, pad1, 'reflect', dtype)
+        check_pad_with_shape(shape2, ctx, pad2, 'reflect', dtype)
 
 
 def np_instance_norm(data, weight, bias, eps):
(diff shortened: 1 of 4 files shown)


haojin2:
I think you can get rid of the change from `MSHADOW_REAL_TYPE_SWITCH` to `MSHADOW_TYPE_SWITCH`, that way you still have support for all float types while getting rid of the integer issue.

sbodenstein:
@haojin2: it gets rid of the integer issue by getting rid of integer support (which seems unnecessarily restrictive). But adding integer support could also be a separate PR. What do you think I should do?

haojin2:
@sbodenstein I would say that floating inputs should be the most common use case, so I would suggest that we start with floating types only and if we see needs for integer types later we can address accordingly.

sbodenstein:
@haojin2: ok, done. Can we merge?

haojin2:
@sbodenstein I cannot decide that as I'm not a committer, I'll ping some committers for you.
@anirudh2290 @eric-haibin-lin @reminisce Can you guys take a look?

tests/python/unittest/test_operator.py
```diff
@@ -2905,37 +2905,46 @@ def test_roipooling():
...
-def check_pad_with_shape(shape, xpu, pad_width, mode):
+def check_pad_with_shape(shape, xpu, pad_width, mode, dtype="float64"):
     # bind with label
-    X = mx.symbol.Variable('X')
+    real_types = ["float16", "float32", "float64"]
+    X = mx.symbol.Variable('X', dtype=dtype)
     Y = mx.symbol.Pad(data=X, mode=mode, pad_width=pad_width)
-    x = mx.random.uniform(-1, 1, shape, ctx=mx.cpu()).copyto(xpu)
+    if dtype in real_types:
+        x = mx.random.uniform(-1, 1, shape, ctx=mx.cpu(), dtype=dtype).copyto(xpu)
```
eric-haibin-lin: what's the purpose of looping over real_types and generate a new x to override the previous one?

sbodenstein: Its not looping over `real_types` and its not overriding anything (`X` and `x` are different). This is simply written to be compatible with the integer case one day (happy to remove for now if you want). Unless I'm misunderstanding something?

sbodenstein:
It would be great if I could get some idea of when this can be merged, as we want to make new builds with this feature enabled.

haojin2:
@eric-haibin-lin @anirudh2290 Please take a look when you have time, thanks!

tests/python/unittest/test_operator.py
```diff
@@ -2905,37 +2905,46 @@ def test_roipooling():
...
 
-def check_pad_with_shape(shape, xpu, pad_width, mode):
+def check_pad_with_shape(shape, xpu, pad_width, mode, dtype="float64"):
     # bind with label
-    X = mx.symbol.Variable('X')
+    real_types = ["float16", "float32", "float64"]
+    X = mx.symbol.Variable('X', dtype=dtype)
     Y = mx.symbol.Pad(data=X, mode=mode, pad_width=pad_width)
-    x = mx.random.uniform(-1, 1, shape, ctx=mx.cpu()).copyto(xpu)
+    if dtype in real_types:
```
haojin2: I think you do not really need this if-else branch at this moment as the caller of this function only passes real types.

tests/python/unittest/test_operator.py
```diff
@@ -2905,37 +2905,46 @@ def test_roipooling():
...
     assert_almost_equal(out, np_out)
     # grad check
-    check_numeric_gradient(Y, [x.asnumpy()], numeric_eps=1e-2, rtol=1e-2)
+    if dtype in real_types:
+        check_numeric_gradient(Y, [x.asnumpy()], numeric_eps=1e-2, rtol=1e-2)
 
 
 @with_seed()
 def test_pad():
+    ct = default_context()
```
apeforest: nit: ctx is probably a more commonly used name for context?

tests/python/unittest/test_operator.py
```diff
@@ -2905,37 +2905,46 @@ def test_roipooling():
...
     pad2 = (0, 0, 0, 0, 1, 2, 3, 4, 3, 1)
-    check_pad_with_shape(shape1, default_context(), pad1, 'constant')
-    check_pad_with_shape(shape1, default_context(), pad1, 'edge')
-    check_pad_with_shape(shape2, default_context(), pad2, 'constant')
-    check_pad_with_shape(shape2, default_context(), pad2, 'edge')
-    check_pad_with_shape(shape1, default_context(), pad1, 'reflect')
-    check_pad_with_shape(shape2, default_context(), pad2, 'reflect')
+    # note: this op doesn't support ints yet. Add tests when supported
+    test_types = ["float32", "float64", "float16"]
+    for d in test_types:
```
haojin2: nit: to be consistent with most other tests within this file I would prefer:
```Python
dtypes = [np.float16, np.float32, np.float64]
for dtype in dtypes:
```

tests/python/unittest/test_operator.py
sbodenstein: Removed. I had kept it there as I thought that it will be useful when integer support is added. But removed for now.

tests/python/unittest/test_operator.py
sbodenstein: Changed to `ctx`.

tests/python/unittest/test_operator.py
sbodenstein: Changed.

nswamy:
LGTM, I triggered the CI again, we can merge once CI passes.

nswamy:
@sbodenstein changes to the `test_pad` unit test is failing, could you please fix it
```

ERROR: test_operator.test_pad

  File "C:\Anaconda3\envs\py2\lib\site-packages\nose\case.py", line 197, in runTest

    self.test(*self.arg)

  File "C:\jenkins_slave\workspace\ut-python-cpu\tests\python\unittest\common.py", line 172, in test_new

    orig_test(*args, **kwargs)

  File "C:\jenkins_slave\workspace\ut-python-cpu\tests\python\unittest\test_operator.py", line 3039, in test_pad

    for dtype in dtypes:

NameError: global name 'dtypes' is not defined
```

tests/python/unittest/test_operator.py
```diff
@@ -2926,16 +2926,20 @@ def check_pad_with_shape(shape, xpu, pad_width, mode):
...
     shape2 = (2, 3, 3, 5, 4)
     pad2 = (0, 0, 0, 0, 1, 2, 3, 4, 3, 1)
-    check_pad_with_shape(shape1, default_context(), pad1, 'constant')
-    check_pad_with_shape(shape1, default_context(), pad1, 'edge')
-    check_pad_with_shape(shape2, default_context(), pad2, 'constant')
-    check_pad_with_shape(shape2, default_context(), pad2, 'edge')
-    check_pad_with_shape(shape1, default_context(), pad1, 'reflect')
-    check_pad_with_shape(shape2, default_context(), pad2, 'reflect')
+    # note: this op doesn't support ints yet. Add tests when supported
+    test_types = ["float16", "float32", "float64"]
```
haojin2: should be
```Python
dtypes = [np.float16, np.float32, np.float64]
```
?

PR #12038, opened 2018-08-05
MXNET-776 [Perl] Better documentation/bug fixes.

sergeykolychev:
## Description ##
1) Several new metric classes
2) Improved documenation
3) Bugfixes.

### Essentials ###

### DIFF
--- a/perl-package/AI-MXNetCAPI/mxnet_typemaps.i
+++ b/perl-package/AI-MXNetCAPI/mxnet_typemaps.i
@@ -1215,5 +1215,5 @@
 
 %typemap(in) (void* callback_handle)
 {
-    $1 = (void*)$input;
+    $1 = (void*)newSVsv($input);
 }
--- a/docs/tutorials/index.md
+++ b/docs/tutorials/index.md
@@ -154,6 +154,17 @@ Select API:&nbsp;
 
 <hr>
 
+## Perl Tutorials
+
+* Getting Started
+    * [Machine learning in Perl](http://blogs.perl.org/users/sergey_kolychev/2017/02/machine-learning-in-perl.html)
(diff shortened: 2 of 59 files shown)


perl-package/AI-MXNetCAPI/mxnet_typemaps.i
```diff
@@ -1215,5 +1215,5 @@
 
 %typemap(in) (void* callback_handle)
 {
-    $1 = (void*)$input;
+    $1 = (void*)newSVsv($input);
```
tlby: I suspect you had to do this because  you found`$input` was getting freed?  newSVsv() will return a copy with a refcount of 1 preventing that, but `SvREFCNT_inc($input)` would likely have worked here as well with less effort.  But in both cases, we probably need an `SvREFCNT_dec(...)` somewhere to avoid a (small) memory leak.

sergeykolychev: This was interesting case. It's related to the call backs for the kvstore and the montitor.
The things worked with XS Mouse but when I tried pure perl Mouse the subs that were expected to persists by the swig side were somehow freed on the Perl level.
The reason of the freeing was unclear to me, as I still kept the references to these subs in specifically created for this case mouse accessors.

In any case I have changed the code and now not keeping these subs in perl space at all (the accessors had been removed), so essentially swig copies these subs and keeps them.

```SvREFCNT_inc``` would have worked here as well, probably, but I went with the way that's recommended in the perlapi.

tlby: okay great.  You know me, I think of how everything at the low level will behave in in a tight loop.  Sounds like that's not the case here so no worries.

PR #12042, opened 2018-08-06
Make check_isfinite, check_scale optional in clip_global_norm

leezu:
## Description ##
Make check_isfinite, check_scale optional in clip_global_norm. If both are set to false, clip_global_norm does not force any synchronization and throughput can be increased. Note if check_scale=False, this requires multiplying all arrays with 1 in cases where the multiplication could be skipped by doing a blocking check.

While this PR preserves the old default behavior of using blocking calls, we may want to change the default behavior to improve throughput.

### Essentials ###

### Changes ###
- [X] Make check_isfinite, check_scale optional in clip_global_norm

## Comments ##
- This improves throughput in https://github.com/dmlc/gluon-nlp/pull/253 by around 10%.
@eric-haibin-lin

### DIFF
--- a/python/mxnet/gluon/utils.py
+++ b/python/mxnet/gluon/utils.py
@@ -115,7 +115,7 @@ def split_and_load(data, ctx_list, batch_axis=0, even_split=True):
     return [i.as_in_context(ctx) for i, ctx in zip(slices, ctx_list)]
 
 
-def clip_global_norm(arrays, max_norm, check_isfinite=True, check_scale=True):
+def clip_global_norm(arrays, max_norm, check_isfinite=True):
     """Rescales NDArrays so that the sum of their 2-norm is smaller than `max_norm`.
 
     Parameters
@@ -125,11 +125,6 @@ def clip_global_norm(arrays, max_norm, check_isfinite=True, check_scale=True):
     check_isfinite : bool, default True
          If True, check that the total_norm is finite (not nan or inf). This
          requires a blocking .asscalar() call.
-    check_scale : bool, default True
-         If True, skip array rescaling if max_norm / total_norm >= 1. This
-         requires a blocking call. If False, rescale arrays with min(1,
-         max_norm / total_norm).
-
     """
     def _norm(array):
         if array.stype == 'default':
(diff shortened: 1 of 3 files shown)


lupesko:
@leezu can you please follow the community's convention to file a JIRA and update the issue title?
@eric-haibin-lin bouncing for review.

leezu:
@lupesko I considered this to fall into "PRs with tiny changes" which don't require JIRA. Currently tests are failing, I'll need to add some fixes before review can proceed. Sorry for the delay

python/mxnet/gluon/utils.py
```diff
@@ -115,8 +115,21 @@ def split_and_load(data, ctx_list, batch_axis=0, even_split=True):
...
+    Parameters
+    ----------
+    arrays : list of NDArray
+    max_norm : float
+    check_isfinite : bool, default True
+         If True, check that the total_norm is finite (not nan or inf). This
+         requires a blocking .asscalar() call.
+    check_scale : bool, default True
+         If True, skip array rescaling if max_norm / total_norm >= 1. This
+         requires a blocking call. If False, rescale arrays with min(1,
```
szha: rescale is not blocking

leezu: Yes. But previously there is a blocking call to check if re-scale is necessary.
If `check_scale is False`, we always rescale, possibly by `1`.

szha: I see. Maybe we can use contrib.cond for this

leezu: `contrib.cond` would only help when working with the symbolic interface. For ndarray, `contrib.cond` also uses a blocking `.asscalar()` call. `clip_global_norm` always works with the ndarray API.

leezu: For now the PR preserves the old default behavior. I'm happy to remove the `check_scale` argument and to always rescale, trading off computation against avoiding blocking calls. That would assume that it is always cheaper/better do the rescaling than to wait for a blocking asscalar() and potentially avoid rescaling.

eric-haibin-lin: +1 for removing check_scale. It doesn't seem necessary. We can always perform a scaling.

Issue #11964, opened 2018-08-01
[Feature request] There is no converter API from mx.nd.NDArray to DLPack tensor


wkcn:
Hi, all.

[DLPack](https://github.com/dmlc/dlpack) is important for sharing tensors and operators among Deep Learning Frameworks.

PyTorch provides the API `torch.utils.dlpack.to_dlpack` and `torch.utils.dlpack.from_dlpack`.

MXNet also uses DLPack, however it seems that there is no converter API from mx.nd.NDArray to DLPack tensor ([DLManagedTensor](https://github.com/dmlc/dlpack/blob/master/include/dlpack/dlpack.h#L125)).

I think it's better to add the API.

Thank you!

---------------------------------
Note:

In [TVM](https://github.com/dmlc/tvm),
[DLManagedTensor* NDArray::ToDLPack() const](https://github.com/dmlc/tvm/blob/master/src/runtime/ndarray.cc#L117)
[NDArray NDArray::FromDLPack(DLManagedTensor* tensor)](https://github.com/dmlc/tvm/blob/master/src/runtime/ndarray.cc#L140)

However, it seems that `NDArray` of MXNet and TVM are different.

Roshrini:
@sandeep-krishnamurthy Can you please add label: FeatureRequest

arogozhnikov:
`cupy` (version > 5) also supports dlpack.

tqchen:
This is a good proposal, there is one thing that we need to be cautious about: MXNet uses asynchronous execution and things can run out of order if it is not tracked by mxnet engine.

Here is an example that things can go wrong.
```python
x = mxnet.nd.zeros((3,4))
x = x + 1
# pass to pytorch or tvm
y = other_framework.nd.fromdlpack(x.asdlpack())
# Y may or maynot be 1
print(y)
```
This is because when we directly call ```x.asdlpack()``` and directly convert the data structure, the execution may not yet finished.

### A Fix
We need to also add flags to asdlpack

```python
x = mxnet.nd.zeros((3,4))
x = x + 1
# pass to pytorch or tvm
x.wait_for_read()
y = other_framework.nd.fromdlpack(x.asdlpack)
# Y may or maynot be 1
print(y)
```
This works for read case, similarly, if we plan to modify the content, we need to do
```python
x = mxnet.nd.zeros((3,4))
x = x + 1
# pass to pytorch or tvm
x.wait_for_write()
[...]

PR #12047, opened 2018-08-06
[MXNET-779]Add DLPack Transformation API

wkcn:
## Description ##
Add DLPack Transformation API

[DLPack](https://github.com/dmlc/dlpack) is important for sharing tensors and operators among Deep Learning Frameworks.
Issue: #11964 

### Essentials ###

### Changes ###
- [x] Add `MXNDArrayToDLPack`, `MXNDArrayFromDLPack`, `MXNDArrayCallDLPackDeleter` and `MXNDArrayCallDLPackCapsuleDeleter` in `c_api` 
- [x] Add `ToDLPack()` and `FromDLPack` for the class `NDArray`
- [x] Add the constructor `TBlob(const DLTensor &dltensor)`
- [x] Add the function `DLDataTypeTransform` in `tensor_blob.h` for the transformation from `DLDataType` to `MShadow Type Flag`
- [x] Add `NDArray(const TBlob &data, int dev_id, const std::function<void()>& deleter) function`
- [x] Add `mx.nd.NDArray.to_dlpack_for_read`, `mx.nd.NDArray.to_dlpack_for_write`, `mx.nd.to_dlpack_to_read`, `mx.nd.to_dlpack_for_write` and `mx.nd.from_dlpack`
- [x] Add unit-test for DLPack Transformation API
## Comments ##
[Extra Test with PyTorch](https://gist.github.com/wkcn/1d0151c898582541e6eb74f0162eaa87)


python/mxnet/_ctypes/ndarray.py
```diff
@@ -31,10 +31,10 @@
 
 class NDArrayBase(object):
     """Base data structure for ndarray"""
-    __slots__ = ["handle", "writable"]
+    __slots__ = ["handle", "writable", "dlpack_handle"]
```
tqchen: do not put dlpack_handle inside it, instead directly return a pycapsule that contains the handle

python/mxnet/cython/ndarray.pyx
```diff
@@ -52,12 +53,15 @@ cdef class NDArrayBase:
...
-    def __init__(self, handle, writable=True):
+    def __init__(self, handle, writable=True, dlpack_handle=None):
         self._set_handle(handle)
         self.cwritable = writable
+        self.cdlpack_handle = dlpack_handle
 
     def __dealloc__(self):
         CALL(MXNDArrayFree(self.chandle))
+        if self.cdlpack_handle:
+            CALL(MXNDArrayCallDLPackDeleter(self.cdlpack_handle))
```
tqchen: do not put dlpack_handle inside the object

src/ndarray/ndarray.cc
```diff
@@ -312,6 +312,46 @@ NDArray NDArray::data_ndarray() const {
   return ret;
 }
 
+struct NDArrayDLManager {
+    NDArray handle;  // ref NDArray
+    DLManagedTensor tensor;
+    TShape strides;  // store variable strides
```
tqchen: DLPack allow strides field to be nullptr,  for mxnet do do not need to store strides, as we only support compact tensor for now.

src/ndarray/ndarray.cc
```diff
@@ -312,6 +312,46 @@ NDArray NDArray::data_ndarray() const {
...
+};
+
+DLManagedTensor* NDArray::ToDLPack() const {
+  NDArrayDLManager* dlmanager(new NDArrayDLManager);
+  dlmanager->handle = *this;
+  if (!is_none()) {
+    dlmanager->tensor.dl_tensor = data().dltensor();
+    // assign value for dl_tensor.strides
+    if (!dlmanager->tensor.dl_tensor.strides) {
```
tqchen: strides == nullptr

src/ndarray/ndarray.cc
tqchen: During conversion to mxnet tensor, check the strides field and see if it is compact

tqchen:
see my additional comments in https://github.com/apache/incubator-mxnet/issues/11964

wkcn:
@tqchen Thank you! I will change it.

src/ndarray/ndarray.cc
wkcn: I will fix it.
It seems that PyTorch.utils.from_dlpack doesn't allow strides nullptr.

include/mxnet/c_api.h
```diff
@@ -737,6 +741,57 @@ MXNET_DLL int MXNDArrayGetShape(NDArrayHandle handle,
...
+ * \return 0 when success, -1 when failure happens
+ */
+MXNET_DLL int MXNDArrayCallDLPackDeleter(DLManagedTensorHandle dlpack);
+
+/*!
+ * \brief Delete a dlpack tensor
+ * \param dlpack_capsule the pointer of a PyCapsule storing DLManagedTensor
+ * \return 0 when success, -1 when failure happens
+ */
+MXNET_DLL void MXNDArrayCallDLPackCapsuleDeleter(PyObjectHandle dlpack_capsule);
```
tqchen: I am less certain why we need the deleter function here, can they be directly handled in the python/cython side?

include/mxnet/tensor_blob.h
```diff
@@ -104,6 +104,14 @@ class TBlob {
       : dptr_(dptr), shape_(shape), type_flag_(type_flag) {
     SetDLTensor(dev_mask, dev_id);
   }
+  /*!
+   * \brief constructor that construct TBlob from DLTensor
+   * \param DLTensor Object
+   */
+  explicit TBlob(const DLTensor &dltensor) : dptr_(dltensor.data),
```
tqchen: Need to add compactness check

tqchen: Specifically, TBlob only support compact tensors, need to check strides == null or the strides reflect a compact setting

python/mxnet/_ctypes/ndarray.py
```diff
@@ -31,21 +31,24 @@
 
 class NDArrayBase(object):
     """Base data structure for ndarray"""
-    __slots__ = ["handle", "writable"]
+    __slots__ = ["handle", "writable", "dlpack"]
     # pylint: disable= no-member
 
-    def __init__(self, handle, writable=True):
+    def __init__(self, handle, writable=True, dlpack=None):
```
tqchen: dlpack should not be part of the member, the PyCapsule manages itself

tqchen:
Since it is a C API change @piiswrong can you also take a look?

include/mxnet/c_api.h
```diff
@@ -737,6 +741,57 @@ MXNET_DLL int MXNDArrayGetShape(NDArrayHandle handle,
...
                                void **out_pdata);
+/*!
+* \brief Create a reference view of NDArray that
+*  represents as DLManagedTensor until
+*  all the pending writes with respect NDArray are finished.
+* \param handle the handle to the ndarray
+* \param out_dlpack pointer holder to get pointer of DLManagedTensor
+* \return 0 when success, -1 when failure happens
+*/
+MXNET_DLL int MXNDArrayToDLPackForRead(NDArrayHandle handle,
```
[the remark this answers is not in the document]
wkcn: I will modify it.

include/mxnet/c_api.h
wkcn: I tried to implement a deleter function in python, however the deleter function may be released by Python GC before calling the deleter function.
The Python Frontend of MXNet both uses [Python(ctypes)](https://github.com/apache/incubator-mxnet/tree/master/python/mxnet/_ctypes) and [Cython](https://github.com/apache/incubator-mxnet/tree/master/python/mxnet/cython). It may be impossible to implement the deleter function in ctypes. 
So the deleter function should be implemented in C++.

include/mxnet/tensor_blob.h
wkcn: Thanks. I will move the strides check from ndarray.cpp to tensor_blob.h.

python/mxnet/_ctypes/ndarray.py
wkcn: When calling `FromDLPack` function, NDArray need to manage the release of `NDArrayDLManager::handle` in `ndarray.cc`.
e.g.
```
a = mx.nd.array([1,2,3]) # Denote TBlob x to store the data
pack = a.to_dlpack_for_write()
b = mx.nd.from_dlpack(pack)
del a, pack
# a and PyCapsule pack has been released.
# b need to manage the release TBlob x.
```
NDArray doesn't have the deleter function, so I made dlpack as a member of NDArray.
When the reference count of dlpack in NDArray is zero, the TBlob will be released.
Is there any other way to keep the reference?

tqchen: A better way is to keep NDArray's shared_ptr inside the manager_ctx itself, you can take a look at TVM's NDArray to DLManagedTesnor impl

wkcn: NDArray in MXNet and TVM are different. NDArray in TVM has the function `IncRef` and `DecRef` to change the reference count, however that in MXNet uses `NDArray::ptr_` (std::shared_ptr<Chunk>) to manage the reference count. `NDArray::ptr_` is a private member of `NDArray`.
The PR is similar to the PyTorch DLPack implementation [Code](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/DLConvertor.cpp#L128)
I add a `NDArrayDLManager` to manage the reference count of NDArray. [Code](https://github.com/apache/incubator-mxnet/pull/12047/files#diff-67d2910557c85e38570cd7c4f940cdbdR315) line 315 in src/ndarray/ndarray.cc

Setting `dlpack` as the NDArray(Python Class) member is to avoid the release of NDArrayDLManager, e.g. When the original NDArray and the PyCapsule (DLPack) are release, the new NDArray (generated by from_dlpack) still exists.

tqchen: you can create a new NDArray() that copies the original NDArray(which increases refcount) and put that as a context

wkcn: I use NDArrayDLManager whose member is a copy of NDArray, which increases refcount.
The deleter of DLPack is to delete NDArrayDLManager, which decreases refcount.

If a new NDArray() created by DLPack, the deleter of DLPack will be called by the new NDArray. Similarly, If a PyTorch Tensor createed by DLPack (MXNet), the deleter will be called by the new PyTorch Tensor.

NDArray (CPP and Python class) doesn't have deleter API, so I set the DLPack as the member of NDArray (Python). When NDArray generated by DLPack releases, the DLPack object in the NDArray will release, then the deleter function in DLPack will be called.

tqchen: You just need to create new NDArray that copies the original NDArray, and create a Deleter function that deletes the NDArray in the context. There is no need for python NDArray to keep reference of the dlpack object. As the DLManagedTensor manages itself

tqchen: Actually, your implementation of ToDLPack is fine and the Manager class is OK. Since there is already the Manager class and the deleter will get called, you do not need to keep reference to the dlpack in the NDArray

include/mxnet/c_api.h
tqchen: Take a look at destructor at https://github.com/dmlc/tvm/pull/1573

tqchen: There is some subtlty here but they can never-the-less be implemented

python/mxnet/ndarray/ndarray.py
```diff
@@ -3851,3 +3898,117 @@ def histogram(a, bins=10, range=None):
...
+    >>> u += 1
+    >>> x
+    [[2. 2. 2.]
+     [2. 2. 2.]]
+    <NDArray 2x3 @cpu(0)>
+    """
+    check_call(_LIB.MXNDArrayWaitToWrite(data.handle))
+    dlpack = DLPackHandle()
+    check_call(_LIB.MXNDArrayToDLPack(data.handle, ctypes.byref(dlpack)))
+    return ctypes.pythonapi.PyCapsule_New(dlpack, b'dltensor', pycapsule_dlpack_deleter)
```
tqchen: cannot directly pass in string, instead use a global string object, PyCapsule requires the string to outlive the capsule

src/c_api/c_api.cc
```diff
@@ -494,6 +494,57 @@ int MXNDArrayGetData(NDArrayHandle handle,
...
+  if (dlpack) {
+    DLManagedTensor *p_dlpack = static_cast<DLManagedTensor*>(dlpack);
+    p_dlpack->deleter(p_dlpack);
+  }
+  API_END();
+}
+
+
+typedef struct {
+    char py_object[16];
```
tqchen: This seems to be dangerous unless pycapsule is already standardized and won't change

include/mxnet/c_api.h
wkcn: I ever tried to write a python function as the destructor, but it can't pass CI.
Please see [the MXNet CI result](http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/PR-12047/8/pipeline)
All windows test_dlpack failed because the destructor written in Python is released before calling it.

PyTorch implemented the destructor using Python API in C++, and CuPy implemented it by cython, namely the code will be built by C++.
However, MXNet uses ctypes and cython. I couldn't find a better way to implement the destructor except writing it in MXNet C++ API.

python/mxnet/_ctypes/ndarray.py
wkcn: Sorry. I don't know which object manages the actual memory after from_dlpack is called.
In the following code, 
```
a = mx.nd.array([1,2,3]) # Denote TBlob x to store the data
pack = a.to_dlpack_for_write()
b = mx.nd.from_dlpack(pack)
del a, pack
```
The original NDArray `a` and dlpack are released. However, the Manager `NDArrayDLManager` should be kept.
The new NDArray `b` get the data pointer from DLPack.
Which object will call the deleter?

In PyTorch, the Tensor class has the deleter structrual parameter. [Code1](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/DLConvertor.cpp#L170) [Code2](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/templates/Type.h#L88)
When the refcount of Tensor is zero, the Tensor Object will call the deleter to release `ATenDLMTensor`.
MXNet doesn't have the explicit deleter structual parameter. 
So I set `dlpack` as the member of NDArray (Python) class.
When the refcount of NDArray is zero, the refcount of dlpack in NDArray is zero too.
The dlpack will be released then call the deleter to delete `NDArrayDLManager`.

include/mxnet/c_api.h
tqchen: There is a special trick that need to be used to implement destructor in python. https://github.com/dmlc/tvm/pull/1573 implements the destructor in python and it works

tqchen: Specificialy, the signature of the destructor need to use c_void_p instead of py_object to avoid repeatitive destruction

python/mxnet/_ctypes/ndarray.py
tqchen: In your case, when a get deleted, b still holds a NDArrayDLManager, which is allocated by new, and that object still hold NDArray(which holds a shared_ptr), so the original resource won't be released

tqchen: You need to be careful to use shape from the same NDArray in your NDArrayDLManager

include/mxnet/c_api.h
wkcn: Yes. I knew the trick and tried it in my previous PR. But it failed in Windows Test.
[Related CI](http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/PR-12047/8/pipeline)

It seems that the CI of TVM doesn't have Windows Test so the CI is passed.

src/c_api/c_api.cc
wkcn: Yes, it's dangerous.
I want to call the function `PyCapsule_GetPointer` in c_api.cc,
however MXNet doesn't include Python.h header file.

include/mxnet/c_api.h
tqchen: This is strange as destructor itself sits in the global scope and should be destructed after the dltensors(which have a local scope)

wkcn: Yes.
In [the test code](https://gist.github.com/wkcn/501cd118b8ceb1608bd6c8fa968e7912), it works in Linux but failed in Windows.

python/mxnet/_ctypes/ndarray.py
wkcn: Thanks.

In the other case,
```python
from torch.utils import dlpack
a = torch.array([1,2,3])
pack = dlpack.to_dlpack(a)
b = mx.nd.from_dlpack(pack)
del a, pack
```
When `dlpack.to_dlpack` is called, PyTorch will allocate `ATenDLMTensor` which increases the refcount of Torch Tensor [code](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/DLConvertor.cpp#L129).
After the variables `a` and `pack` are released, `ATenDLMTensor` still exists.
I think the deleter should be called by the new NDArray `b` when the NDArray `b` releases. Refer to [PyTorch FromDLPack](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/DLConvertor.cpp#L170). However, NDArray doesn't have explicit deleter parameter.

In my PR, `from_dlpack` will copy the dlpack object.
When the old dlpack releases, it doesn't call the deleter.
The new dlpack will be a member of the new NDArray as `NDArray(handle=handle, dlpack=dlpack_copy)`.
When the new NDArray releases, the new dlpack will be released, then call the deleter by the new dlpack. And the deleter will release `NDArrayDLManager` or `ATenDLMTensor`. The refcount of the old NDArray will decrease 1.

include/mxnet/c_api.h
tqchen: I see two problems in your particular gist you paste. 

- The destructor need to be declared in the global scope(instead of constructing when passing to the argument)
```python

cfunc = ctypes.CFUNCTYPE(None, ctypes.c_void_p)

def dfunc(dltensor):
    pycaps = ctypes.cast(dltensor, ctypes.py_object)
    pass

c_destructor = cfunc(dfunc)
c_str_dltensor = cstr("dltensor")

def test():
    a = ctypes.pythonapi.PyCapsule_New(1, c_str_dltensor, c_destructor)
test()

```

wkcn: Thank you!
I will test it on Windows.
If it works, I will update the PR.

python/mxnet/ndarray/ndarray.py
wkcn: Solved it. Thank you!

src/c_api/c_api.cc
wkcn: I have removed it in the latest PR.

src/c_api/c_api.cc
```diff
@@ -500,6 +500,33 @@ int MXNDArrayGetData(NDArrayHandle handle,
...
+  NDArray *arr = static_cast<NDArray*>(handle);
+  *out_dlpack = arr->ToDLPack();
+  API_END();
+}
+
+int MXNDArrayFromDLPack(DLManagedTensorHandle dlpack,
+                        NDArrayHandle *out_handle) {
+  API_BEGIN();
+  NDArray *pdata = new NDArray();
+  *pdata = NDArray::FromDLPack(
```
wkcn: Does it need Rvalue Referene？

python/mxnet/_ctypes/ndarray.py
wkcn: I made a copy of NDArray as the member of NDArrayDLManager, and the copy increase the refcount.
I'm confused how to modify the PR.
After creating a new NDArray(Python) from DLPack, then delete the old NDArray(Python) and PyCapsule(DLPack).

Which object will call the deleter function?

In my case, when `a` gets deleted, how does `b` hold the NDArrayDLManager? It seems that `b` only get the pure data pointer from `DLManagedTensor::dl_tensor::data`, the type of the pointer is not a shared pointer.
And how does `b` store the pointer to `NDArrayDLManager` in MXNet NDArray?

tqchen:
My point is that the DLManagerd Tensor should manage its own deletion and you do not have to add dlpack field for mx.nd.NDArray. Even when the frontend NDArray release, the internal data will be kept alive until the pycapsule get released and the DLManagedTensor's deleter get called

wkcn:
@tqchen Hi! I have modified the PR, the DLManaged Tensor will manage itself.
However, there is a problem.
In this case,
```python
import mxnet as mx
pack = mx.nd.array([1,2,3]).to_dlpack_for_write()
b = mx.nd.from_dlpack(pack)
del pack # DLManagedTensor's deleter get called when `pack` get released.
print (b) # b != [1,2,3]
```

Does it need to manage the release of dlpack in NDArray Chunk (C++) ?

tqchen:
pack is only supposed to be consumed once, when from_dlpack is called, the pack itself should be marked as used and its deleter will no longer be called when the capsule get destroyed

wkcn:
@tqchen Yes. The latest PR has implemented it in ndarray.py.
And the new NDArray created by from_dlpack will call the deleter.

wkcn:
@tqchen @piiswrong 
Hi!
Is there any problem? Could you please merge the PR?
Thank you!

include/mxnet/ndarray.h
```diff
@@ -831,7 +855,7 @@ class NDArray {
 
     Chunk(const NDArrayStorageType storage_type_, const TBlob &data,
           const std::vector<TBlob> &aux_data, int dev_id)
-        : static_data(true), delay_alloc(false), storage_type(storage_type_) {
+        : static_data(true), delay_alloc(false), storage_type(storage_type_), deleter_(nullptr) {
```
tqchen: We cannot plug deleter to the TBlob and Chunk, this seems will cause more problems than what it can solve.

std::shared_ptr<Chunk> do support customized deleter already, and maybe that can be used to help solve this issue of having to hide a customized deleter for NDArray

tqchen:
@wkcn sorry I was away for a vacation recently, please see my comments

wkcn:
@tqchen Thank you! I will try to change it.

wkcn:
@tqchen Hi! I have updated the code as your comments. 
I add a new constructor `NDArray(const TBlob &data, int dev_id, const std::function<void()>& deleter) ` for 'NDArray class`, because the performance between `std::make_shared` and 'std::shared_ptr::shared_ptr' is different.

tqchen:
I think it is good to go from my end. thanks @wkcn . @piiswrong please take another look, if there is no issues I will merge this in in two days

wkcn:
@tqchen Thank you so much!

tqchen:
This is now merged

PR #12065, opened 2018-08-07
[MXAPPS-805] Notebook execution failures in CI.

vishaalkapoor:
* Add a retry policy when starting a notebook executor to handle the failure to
 start a notebook executor (due to a port collision, kernel taking too
 long to start, etc.).

* Change logging level for tests to INFO so that we have more
 informative test output.


vishaalkapoor:
Closing this pull request while I investigate whether or not we can reduce the generality of the RuntimeError to something more specific.

PR #12068, opened 2018-08-07
[MXAPPS-805] Notebook execution failures in CI.

vishaalkapoor:
* Add a retry policy when starting a notebook executor to handle the failure to
 start a notebook executor (due to a port collision, kernel taking too
 long to start, etc.).

* Change logging level for tests to INFO so that we have more
 informative test output.

### DIFF
--- a/tests/utils/notebook_test/__init__.py
+++ b/tests/utils/notebook_test/__init__.py
@@ -32,6 +32,13 @@ import nbformat
 
 IPYTHON_VERSION = 4  # Pin to ipython version 4.
 TIME_OUT = 10*60  # Maximum 10 mins/test. Reaching timeout causes test failure.
+RETRIES = 10
+
+try:
+    TimeoutError
+except NameError:
+    # py2
+    TimeoutError = RuntimeError
 
 def run_notebook(notebook, notebook_dir, kernel=None, no_cache=False, temp_dir='tmp_notebook'):
     """Run tutorial Jupyter notebook to catch any execution error.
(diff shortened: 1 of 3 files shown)


tests/utils/notebook_test/__init__.py
```diff
@@ -72,15 +79,21 @@ def run_notebook(notebook, notebook_dir, kernel=None, no_cache=False, temp_dir='
...
+
+        # There is a low (< 1%) chance that starting a notebook executor will fail due to the kernel
+        # taking to long to start, or a port collision, etc.
+        for i in range(RETRIES):
+            try:
+                nb, _ = eprocessor.preprocess(notebook, {'metadata': {'path': working_dir}})
+            except (RuntimeError, TimeoutError) as rte:
+                logging.info("Error starting preprocessor: {}. Attempt {}/{}".format(str(rte), i+1, RETRIES))
+                time.sleep(1)
+                continue
```
marcoabreu: Rethrow exception?

Also, what happens if there is a valid failure? Wouldn't this result in a lot of failures because all notebooks get called 10 times?

marcoabreu: It would be good if we could be more specific than just runtime error

vishaalkapoor: I've added a commit which scans for the error message we would like to retry on that encompasses the issues we're seeing. We won't need to scan for TimeoutErrors as before, just RuntimeErrors with one specific error message.

vishaalkapoor:
Thanks everyone for reviews! :-)

PR #11921, opened 2018-07-28
[MXNET-711] Added updated logos to the powered by page and iot select fix

kpmurali:
## Description ##
Updating logos to the powered by page. Added Clojure to the navigation list for APIs.

### Changes ###
- [x] Added updated logos to the powered by page
- [x] Added Clojure to the navigation list for APIs.
- [x] IOT selector fix in the install page.

PR #12075, opened 2018-08-07
Add Clojure to website nav

aaronmarkham:
## Description ##
This adds Clojure to the navigation. I also alphabetized the list since it is getting pretty long and there didn't seem to be any logical order.

This also is taking over the Clojure aspect from #11921 - since I think that PR is going to be blocked forever and will need to be parsed out and closed.


gigasquid:
Very excited to see this happen 😸 💯 🎆

sergeykolychev:
I think alphabetized list is better than what we have now, but I'd suggest to keep Python at the top as an exception. 90% of users will be looking for Python docs anyway.

aaronmarkham:
No. I will have to rebased my other PRs though.

Issue #12073, opened 2018-08-07
get name from symbol cpp


timprepscius:
This is most trivial, and of course a billion ways to work around, however:

Given a symbol:
`auto a = Symbol::Variable("my_super_name") // could be any symbol
`

How do I retrieve the name of the symbol:
`std::cout << a.GetName() << std::endl;
`

Somehow I cannot find the methods to allow this, nor any examples.

haojin2:
There's indeed no such function in Symbol class in cpp-package, a fix in #12076.

lanking520:
@mxnet-label-bot could you please add [feature, c++]

haojin2:
@timprepscius Fix is merged, can you try that out and let me know if you found and problems? Thanks!

leleamol:
@mxnet-lable-bot  [pending-requestor-info]

leleamol:
The fix is merged and there is no update from the requester.
Can this issue be closed?

@mxnet-jenkins @marcoabreu

PR #12076, opened 2018-08-08
Add GetName function in Symbol class for cpp package

haojin2:
## Description ##
As title.

### Essentials ###

### Changes ###
- [x] Add GetName function for Symbol class

## Comments ##
Address #12073

### DIFF
--- a/cpp-package/include/mxnet-cpp/symbol.hpp
+++ b/cpp-package/include/mxnet-cpp/symbol.hpp
@@ -172,6 +172,14 @@ inline std::vector<std::string> Symbol::ListAuxiliaryStates() const {
   return ret;
 }
 
+inline std::string Symbol::GetName() const {
+  int success;
+  const char* out_name;
+  CHECK_EQ(MXSymbolGetName(GetHandle(), &out_name, &success), 0);
+  CHECK_EQ(success, 1);
+  return std::string(out_name);
+}
+
 inline void Symbol::InferShape(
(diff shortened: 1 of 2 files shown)


cpp-package/include/mxnet-cpp/symbol.hpp
```diff
@@ -172,6 +172,13 @@ inline std::vector<std::string> Symbol::ListAuxiliaryStates() const {
   return ret;
 }
 
+inline std::string Symbol::GetName() const {
+  int success;
+  const char* out_name;
+  CHECK_EQ(MXSymbolGetName(GetHandle(), &out_name, &success), 0);
```
apeforest: What if success == 0? Based on the function description, it seems that out_name could be NULL even if the returned value is 0:

```
/*!
 * \brief Get string attribute from symbol
 * \param symbol the source symbol
 * \param key The key of the symbol.
 * \param out The result attribute, can be NULL if the attribute do not exist.
 * \param success Whether the result is contained in out.
 * \return 0 when success, -1 when failure happens
 */
int NNSymbolGetAttr(SymbolHandle in,
                             const char* key,
                             const char** out,
                             int *out);
```

haojin2: added.

haojin2:
@nswamy Should be good for merge? BTW this is a change to cpp-package, I don't think it's related to backend.

nswamy:
Thanks @haojin2 . I changed the label and merging.

PR #12078, opened 2018-08-08
[MXNET-749] Correct usages of `CutSubgraph` in 3 control flow operators

junrushao1994:
## Description ##
This PR corrects the usage of `CutSubgraph`.

`CutSubgraph` consumes a Symbol `s`, and
1) Assumes the attribute `__subgraph_name__` of `s->outputs[0].node` is the name of the subgraph
2) Assumes the boundary of this subgraph be nodes who has at least one immediate predecessor whose attribute `__subgraph_name__` is not the name of the subgraph
3) Duplicates and puts these immediate predecessors into this subgraph

Calling `CutSubgraph` without very special care will cause disastrous consequences, often when `s->outputa[0].node` has an improper name. For example, in control flow operators, if we write a function `lambda: x` which captures an outside symbol `x` (so the attribute of `x` is not desirable), and an operator passes `x` into `CutGraph` as the first output of subgraph, the entire procedure breaks. Without copying and manually correcting such symbols, this will happen anywhere and be very hard to debug. An example that causes this bug could be https://github.com/dmlc/gluon-nlp/pull/233 in which the `pred` I pass to `cond` is calculated outside the subgraph.

This bug affects all three control flow operators, `foreach`, `while_loop` and `cond`.

CC: @zheng-da 

### Essentials ###

### Changes ###
- [x] 3 control flow operators.
- [x] BTW, remove some unnecessary copies of std::string in the implementation of CutSubgraph.

## Comments ##
- Just looked into the code, seems that another thing we haven't ensure is that each subgraph should have its distinct name, otherwise the CutSubgraph may not work properly. Fixed in #12106.

### DIFF
--- a/python/mxnet/symbol/contrib.py
+++ b/python/mxnet/symbol/contrib.py
@@ -133,25 +133,11 @@ def _construct_subgraph(sym_out, sym_states):
     all_outputs = []
     all_outputs.extend(sym_out)
     all_outputs.extend(sym_states)
-    g = symbol.Group(all_outputs)
-
     flat_out = []
-    all_input_names = g.list_inputs()
-    output_names = [o.name for o in sym_out]
     for o in sym_out:
-        if o.name in all_input_names:
-            flat_out.append(symbol.op.identity(o))
-        else:
-            flat_out.append(o)
-
+        flat_out.append(symbol.op.identity(o))
     for s in sym_states:
-        if s.name in all_input_names or s.name in output_names:
-            # There is a problem if the outputs are the same as the inputs
-            # or the first output. By calling identity, we can make sure that
-            # all symbols will refer to different NDArrays.
-            flat_out.append(symbol.op.identity(s))
-        else:
-            flat_out.append(s)
+        flat_out.append(symbol.op.identity(s))
     return symbol.Group(flat_out)
 
 def foreach(body, data, init_states, name="foreach"):
@@ -469,10 +455,8 @@ def while_loop(cond, func, loop_vars, max_iterations=None, name="while_loop"):
             num_outputs = len(outputs) + len(final_state)
             # nnvm cut-graph does not allow inputs and outputs overlap
             # so we calculate the name of inputs, and copy outputs once it overlaps with inputs
-            all_input_names = symbol.Group(outputs + final_state).list_inputs()
-            make_identity = lambda x: symbol.op.identity(x) if x.name in all_input_names else x
             # group all outputs of graph_func
-            graph = symbol.Group(list(map(make_identity, outputs + final_state)))
+            graph = symbol.Group(list(map(symbol.op.identity, outputs + final_state)))
         return graph, num_out_data, num_outputs
 
     def _union_inputs(*graphs):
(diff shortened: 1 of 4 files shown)


zheng-da:
we shouldn't add identity on all outputs. it's too expensive as a fix for a corner case.

But the current solution may not work either. I think a better way is to check if `__subgraph_name__` is in `symbol.list_attr()` in the output symbol from the function.

junrushao1994:
@zheng-da I agree it is expensive. We need to ensure all output nodes of the symbol fed to CutGraph to have the same sub graph name, so not only should we check existence of  __subgraph_name__, but also should check if they are equal. But currently we have not ensured all subgraphs have different names though...

junrushao1994:
I submitted a PR ensuring subgraphs names to be unique in #12106.

junrushao1994:
@zheng-da I updated this PR just now, and now each symbol is copied only when their `__subgraph_name__` != current subgraph name.

Another fix ensuring control flow operators to have unique names is also submitted #12106.

python/mxnet/symbol/contrib.py
```diff
@@ -137,18 +137,16 @@ def _construct_subgraph(sym_out, sym_states):
 
     flat_out = []
     all_input_names = g.list_inputs()
-    output_names = [o.name for o in sym_out]
+    output_names = {o.name for o in sym_out}
     for o in sym_out:
-        if o.name in all_input_names:
+        if o.name in all_input_names or o.list_attr().get("__subgraph_name__", "") != name:
```
zheng-da: what happens if `o` doesn't have `__subgraph_name__`?

src/operator/control_flow.cc
```diff
@@ -1225,6 +1225,9 @@ static bool BackwardCondStorageType(const nnvm::NodeAttrs& attrs,
     CHECK(sync_in_in(input_locs, out_attrs, &subg_out_attrs, is_udf));
     return ret;
   };
+  for (const dim_t &cond_in : params.cond_input_locs) {
+    (*out_attrs)[cond_in] = 0;
+  }
```
zheng-da: what is this for?

python/mxnet/symbol/contrib.py
junrushao1994: @zheng-da `dict.get(key, defaultvalue)` will return you `defaultvalue` if `key` doesn't exist

src/operator/control_flow.cc
junrushao1994: `cond` don't need to be back-propagated

junrushao1994:
@zheng-da Hi, I add unittests for `foreach`, `cond` and `while_loop` according to the code you provided offline. Thank you so much for providing the testcases!

src/operator/control_flow.cc
eric-haibin-lin: Is 0 a magic number?

junrushao1994: @eric-haibin-lin In this case, anything other than `kBadStorageID` will work because we don't bp through `cond`. Do you suggest that I use `kDefaultStorage` to make it clearer?

junrushao1994:
Thank you guys for have this merged!

PR #12083, opened 2018-08-08
GPU Memory Query to C API

sbodenstein:
This PR adds a function `MXNET_DLL int MXGetGPUMemoryInformation(int dev, int* free_mem, int* total_mem);` to the C API to allow any language binding to easily query for the total and available memory available on a particular GPU. 

Its useful as `nvidia-smi` is not available on certain platforms (like OSX). Having a platform-independent way of querying this (eg to automatically figure out appropriate batch sizes) is very useful for higher-level frameworks built on top of MXNet.

This is similar to the cuTorch function [getMemoryUsage](https://github.com/torch/cutorch/blob/7126f4bc02757838dd597d94e5d8ee67b45e0388/init.c#L710), with the same implementation.

### DIFF
--- a/include/mxnet/base.h
+++ b/include/mxnet/base.h
@@ -334,10 +334,10 @@ inline int32_t Context::GetGPUCount() {
 #endif
 }
 
+inline void Context::GetGPUMemoryInformation(int dev, int *free_mem,
+                                             int *total_mem) {
+#if MXNET_USE_CUDA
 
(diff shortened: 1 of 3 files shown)


sandeep-krishnamurthy:
@sbodenstein - Thank you for the contribution. It would be great if you follow up with a Python API for querying GPU memory?

Issue #12009, opened 2018-08-02
F.pick operator don't support negative indices and silently clip out of range indices


zhreshold:
## Description
F.pick operator don't support negative indices

## Minimum reproducible example
```python
from mxnet import nd

x = nd.array([[1, 2, 3, 4], [5, 6, 7, 8]])
y1 = nd.array([[0], [0]])
print(x.pick(y1, axis=-1)
# [1. 5.]

# negative index
y2 = nd.array([[-1], [1000]])
print(x.pick(y2, axis=-1))
# [1. 8.]
```
Clearly that out of range indices are clipped without any warning.

## What have you tried to solve it?

Like other operators, nd/sym.pick should support negative indices rather than clip the indices, out of range indices should raise error.

zhreshold:
closed by #12090

PR #11326, opened 2018-06-18
[MXNET-381] Enhancement of take operator

haojin2:
## Description ##
Previously our `take` operator only supports axis=0 and mode = 'clip' case, this PR adds support for axis in range [-r, r-1] and an additional mode 'wrap'.

### Essentials ###
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] Support for take on any dimension
- [x] New mode 'wrap' for indices
- [x] Unit tests for enhanced take operator

## Comments ##
The legacy implementation for axis=0 and mode='clip' is still preserved to ensure there's no performance or accuracy regression after the enhancement.

PR #12090, opened 2018-08-08
[MXNET-791] Pick with negative indices

perdasilva:
## Description ##
Updates pick operator to handle negative indices. 
Implements #12009 

### Essentials ###

### Changes ###
Updates pick and pick_grad in broadcast_reduce_op to handle negative indices
Updates pick to include a mode parameter ('pick' or 'wrap') to specify whether out of bounds indices should either be clipped or wrapped.

## Comments ##
Making change as it was tagged with `Call for Contribution`

### DIFF
--- a/src/operator/tensor/broadcast_reduce_op.h
+++ b/src/operator/tensor/broadcast_reduce_op.h
@@ -99,6 +99,8 @@ struct ReduceAxisParam : public dmlc::Parameter<ReduceAxisParam> {
   }
 };
 
+enum PickOpMode {kWrap, kClip};
+
 struct PickParam : public dmlc::Parameter<PickParam> {
   dmlc::optional<int> axis;
   int mode;
@@ -112,6 +114,14 @@ struct PickParam : public dmlc::Parameter<PickParam> {
     DMLC_DECLARE_FIELD(keepdims).set_default(false)
       .describe("If true, the axis where we pick the elements is left "
                 "in the result as dimension with size one.");
+    DMLC_DECLARE_FIELD(mode)
+    .add_enum("wrap", kWrap)
+    .add_enum("clip", kClip)
+    .set_default(kClip)
+    .describe("Specify how out-of-bound indices behave. Default is \"clip\"."
+              " \"clip\" means clip to the range. So, if all indices mentioned are too large,"
+              " they are replaced by the index that addresses the last element along an axis. "
+              " \"wrap\" means to wrap around.");
   }
 };
 
@@ -1108,7 +1118,7 @@ void L2NormComputeEx(const nnvm::NodeAttrs& attrs,
                      const std::vector<NDArray>& outputs);
 
 /*! \brief index element from array along axes */
-template<int ndim>
+template<int ndim, bool clip = true>
 struct pick {
   template<typename DType, typename IType>
   MSHADOW_XINLINE static void Map(int i, DType* out, const DType* a,
@@ -1117,15 +1127,20 @@ struct pick {
                                   mshadow::Shape<ndim> sshape) {
     using namespace broadcast;
     int j = static_cast<int>(idx[i]);
-    if (j < 0) j = 0;
-    else if (j >= M) j = M-1;
+    if (clip) {
+      if (j <= 0) j = 0;
+      else if (j >= M) j = M - 1;
+    } else {
+      j = j % M;
+      j += (j < 0) ? M : 0;
+    }
     j = ravel(unravel(i, sshape), bshape) + j*stride;
     out[i] = a[j];
   }
 };
 
 /*! \brief index element from array along axes */
-template<int ndim>
+template<int ndim, bool clip = true>
 struct pick_grad {
   template<typename DType, typename IType>
   MSHADOW_XINLINE static void Map(int i, DType* igrad, const DType* ograd,
@@ -1134,8 +1149,13 @@ struct pick_grad {
                                   mshadow::Shape<ndim> sshape) {
     using namespace broadcast;
     int j = static_cast<int>(idx[i]);
-    if (j < 0) j = 0;
-    else if (j >= M) j = M-1;
+    if (clip) {
+      if (j <= 0) j = 0;
+      else if (j >= M) j = M - 1;
+    } else {
+      j = j % M;
+      j += (j < 0) ? M : 0;
+    }
     j = ravel(unravel(i, sshape), bshape) + j*stride;
     igrad[j] += ograd[i];
   }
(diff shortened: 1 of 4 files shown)


perdasilva:
Hi,

This is my first PR to the project and I'm still learning the ropes. Any and all suggestions would be much welcome. I'm not sure how to handle the case. For simplicity, I handled it with mod arithmetic. But in numpy, you can index in the range [-len, len). Which behaviour should be correct for this issue?
Should the cropping happen for indices < -1*len and >= len ?

Cheers,

Per

tests/python/unittest/test_operator.py
```diff
@@ -30,6 +30,41 @@
 from common import setup_module, with_seed, teardown, assert_raises_cudnn_disabled, assertRaises
 import unittest
 
+@with_seed()
```
perdasilva: I will move this back to its proper place - didn't get a chance as my family arrived home =P

haojin2:
Welcome to MXNet family!
I think you can use what I did for `take` operator as a reference: https://github.com/apache/incubator-mxnet/pull/11326/files#diff-ed06b8d9798aca630313f2a9dd3fcd68R316 for handling negative indices.

src/operator/tensor/broadcast_reduce_op_index.cc
```diff
@@ -131,7 +130,7 @@ Examples::
   pick(x, y=[0,1], 0) = [ 1.,  4.]
 
   // picks elements with specified indices along axis 1
-  pick(x, y=[0,1,0], 1) = [ 1.,  4.,  5.]
+  pick(x, y=[0,-1,0], 1) = [ 1.,  4.,  5.]
```
haojin2: Please consider adding extra example for the newly supported case instead of modifying the existing cases.
For this case maybe you can do
```
pick(x, y=[0,1,0,-1], 1) = [ 1.,  4.,  5.,  4.]
```

tests/python/unittest/test_operator.py
```diff
@@ -4417,7 +4417,7 @@ def test_pick_helper(index_type=np.int32):
             sshape = bshape.copy()
             sshape[axis] = 1
             data = np.random.uniform(-1, 1, size=bshape)
-            index = np.random.randint(0, bshape[axis], size=sshape)
+            index = np.random.randint(-1*bshape[axis], bshape[axis], size=sshape)
```
haojin2: I would prefer `-bshape[axis]`.

perdasilva:
Thank you for the welcome and guidance ^^ 
I hope I have done everything correctly. I've incorporated the mode into the currently existing pick test. I wonder if they should be split? Or maybe the pick_helper should be extended with the mode?

Cheers!!

Per

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -99,6 +99,10 @@ struct ReduceAxisParam : public dmlc::Parameter<ReduceAxisParam> {
   }
 };
 
+namespace pick_ {  // to avoid name conflict
+enum TakeOpMode {kRaise, kWrap, kClip};
```
haojin2: I would prefer this to be like
```c++
enum PickOpMode {kWrap, kClip};
```
as we do not have `raise` mode at all.

haojin2: This way you do not need the extra namespace either.

tests/python/unittest/test_operator.py
```diff
@@ -4417,7 +4417,12 @@ def test_pick_helper(index_type=np.int32):
             sshape = bshape.copy()
             sshape[axis] = 1
             data = np.random.uniform(-1, 1, size=bshape)
-            index = np.random.randint(0, bshape[axis], size=sshape)
+            mode = np.random.choice(a=['clip', 'wrap'])
```
haojin2: Please test both choices for `mode` using a for-loop instead of randomized testing for `mode`

src/operator/tensor/broadcast_reduce_op_index.cc
```diff
@@ -117,7 +117,7 @@ an output array of shape ``(i0,)`` with::
   output[i] = input[i, indices[i]]
 
 By default, if any index mentioned is too large, it is replaced by the index that addresses
-the last element along an axis (the `clip` mode).
+the last element along an axis (the `clip` mode).
```
haojin2: Please get rid of the unnecessary extra whitespace on this line.

perdasilva:
Will do. I kept the raise there to be consistent with the take implementation. Would you like me to remove raise from take?

perdasilva:
Do you think it would make sense to test 'wrap' with the range -2*-bshape[axis]..2*bshape[axis]? To cover cases where one is wrapping more than the length of the axis in both directions? In my understanding, in numpy indices should generally be [-len, len). So, the current implementation diverges from that.

haojin2:
@perdasilva Sorry for the late reply, I think adding more coverage is totally okay. Regarding the inconsistency with numpy, does this operator has some equivalent op in numpy? There is one for `take`, but this one I'm not sure.

perdasilva:
@haojin2 No problem at all ^^
The only thing I could find that is similar to pick in numpy is [numpy.choose](https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.choose.html). But there's not axis input there. It also seems that 'raise' is the default mode.  Upon closer inspection, 'wrap' works the same here as with numpy. Interestingly though, np.array indexing is in the range [-len, len) ^^

I'll also increase the coverage for the test.

haojin2:
@perdasilva So if there's no direct equivalence of this one in numpy then we may not need to match the behavior with numpy, please simply ensure we have enough documentation to document the behavior.

perdasilva:
I believe that's all done now. Thanks for everything ^^

perdasilva:
@haojin2 I believe everything is done as specified. Please merge when you get a chance ^^

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -112,6 +114,14 @@ struct PickParam : public dmlc::Parameter<PickParam> {
     DMLC_DECLARE_FIELD(keepdims).set_default(false)
       .describe("If true, the axis where we pick the elements is left "
                 "in the result as dimension with size one.");
+    DMLC_DECLARE_FIELD(mode)
+    .add_enum("wrap", kWrap)
+    .add_enum("clip", kClip)
+    .set_default(kClip)
+    .describe("Specify how out-of-bound indices bahave. Default is \"clip\"."
```
sandeep-krishnamurthy: nit: bahave-> behave

haojin2:
Please rebase with the latest master, then do a `git submodule update --init --recursive`, then do a `git push --force origin <your-branch-name>`

src/operator/tensor/broadcast_reduce_op.h
perdasilva: Good catch! Thank you!

perdasilva:
Hmm those dmlc changes seem to have persisted...strange

perdasilva:
Ok - Just did a soft reset, updated the submodule, then re-applied my commits. Seems to have fixed the issue =)

sandeep-krishnamurthy:
@haojin2 - This PR is ready for merge after your approval.

haojin2:
@sandeep-krishnamurthy Ready for merge.

Issue #9816, opened 2018-02-17
Dropout may mask values even when ratio=0.0


DickJC123:
One reasonable expectation is that a dropout layout would pass no values when the dropout ratio=1.0 and would pass all values with the dropout ratio=0.0.  However, the current dropout test fails a ratio=0.0 test under some seeds because some values are masked.  Adapting from the current nn/dropout-inl.h, we have essentially:
```
prob_keep = 1.0 - ratio;
rand_pick = random_uniform_pick(0.0,1.0);
mask_out = (rand_pick < prob_keep ? 1.0 : 0.0) * (1.0/prob_keep);
dropout_out = input_data * mask_out;
```
It must be that rand_pick above can include the 1.0 endpoint, so mask_out becomes 0.0.  A quick fix might change the '<' to '<=', but then some values might be passed when the dropout ratio = 1.0 if the rand_pick can be 0.0.  The possible differences between rng properties between cpu and gpu's should be considered: CUDNN rngs return values in (0.0,1.0] while most others are in the range [0.0,1.0).  This can be reproduced in any of the commits of the ci_test_randomness3 PR #9791 if you edit the line above 'def test_dropout()' in test_operator.py to be only "@with_seed()", then execute: (output shown)
```
MXNET_TEST_SEED=990952066 nosetests --verbose tests/python/unittest/test_operator.py:test_dropout
[INFO] Setting module np/mx/python random seeds, use MXNET_MODULE_SEED=881809903 to reproduce.
[WARNING] *** test-level seed set: all "@with_seed()" tests run deterministically ***
test_operator.test_dropout ... [INFO] Setting test np/mx/python random seeds, use MXNET_TEST_SEED=990952066 to reproduce.
FAIL
======================================================================
FAIL: test_operator.test_dropout
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/dcarter/mxnet_dev/dgx/mxnet/tests/python/unittest/common.py", line 152, in test_new
    orig_test(*args, **kwargs)
  File "/home/dcarter/mxnet_dev/dgx/mxnet/tests/python/unittest/test_operator.py", line 4596, in test_dropout
    check_dropout_ratio(0.0, shape)
  File "/home/dcarter/mxnet_dev/dgx/mxnet/tests/python/unittest/test_operator.py", line 4561, in check_dropout_ratio
[...]

PR #12091, opened 2018-08-08
[MXNET-792] Fix for issue #9816

samskalicky:
## Description ##
Fix for issue #9816 Dropout may mask values even when ratio=0.0

as mentioned by @DickJC123, the problem is that on the CPU and GPU the random number generator works slightly differently. On the CPU the numbers are generated in the range [0,1) and on the GPU they are (0,1]. Therefor on the GPU a value of 1 can be generated and using the existing thresholding mshadow operator this results in an incorrect value since its only performing < operation. 

This fix adds a <= comparison for dropout to fix this issue. 

Another possible solution is to match the random number generator behavior between CPU and GPU, for which there is already a PR in the mshadow repo: https://github.com/dmlc/mshadow/pull/338

When this PR gets merged the behavior of dropout with this fix will still be the same. So this fix works now and in the future.

### Essentials ###

### Changes ###
added mshadow op for threshold_eq (theshold currently does <, this will do <=)

modified dropout operator to use threshold_eq instead of theshold this will ensure equivalent behavior for the random numbers generated on CPU [0, 1) and GPU (0, 1]

removed fixed seed for test_dropout

### DIFF
--- a/tests/python/unittest/test_operator.py
+++ b/tests/python/unittest/test_operator.py
@@ -5514,7 +5514,7 @@ def test_stack():
 
 
 # test fails with seed 990952066: 0 output seen with dropout ratio=0. See issue #9816
-@with_seed(1234)
+@with_seed()
 def test_dropout():
     def zero_count(array, ratio):
         zeros = 0
@@ -5566,6 +5566,7 @@ def test_dropout():
 
         exe.arg_arrays[0][:] = 1
         exe.forward(is_train=True)
+
         if not math.isnan(max_value):
             assert exe.outputs[0].asnumpy().max() > 0
         else:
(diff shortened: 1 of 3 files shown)


eric-haibin-lin:
@DickJC123 could you also help review?

samskalicky:
Thanks @apeforest, thats a good suggestions. I updated the title to be a little more descriptive.

eric-haibin-lin:
This will result in incorrect values when ratio=1 on cpu. let's fix that

tests/python/unittest/test_operator.py
```diff
@@ -5514,7 +5514,7 @@ def test_stack():
 
 
 # test fails with seed 990952066: 0 output seen with dropout ratio=0. See issue #9816
```
haojin2: If no longer flaky please get rid of this line too.

samskalicky:
Thanks @haojin2, i removed that unnecessary comment

PR #12092, opened 2018-08-08
Decrease success rate to make test more stable

Ishitori:
I have added this test back to unit test coverage and decreased success rate even more, to make sure that fails would happen even more rare

## Description ##
Trying to fix another flaky test

### Essentials ###

### Changes ###
- [X] Test is returned back


haojin2:
Have you tried with more than 10000 trials?

Ishitori:
Yes, I confirm it works for 10k runs as well.

PR #12104, opened 2018-08-09
[DO NOT REVIEW] Subgraph API

reminisce:
## Description ##
Incremental change to the dev branch. Not ready for review.

### Essentials ###

### Changes ###

## Comments ##


src/executor/graph_executor.cc
```diff
@@ -1699,6 +1701,146 @@ GraphExecutor::CachedSegOpr GraphExecutor::CreateCachedSegOpr(size_t topo_start,
...
+                                   const std::map<std::string, Context>& ctx_map,
+                                   const std::vector<Context>& in_arg_ctxes,
+                                   const std::vector<Context>& aux_state_ctxes) {
+  auto subgraph_prop = op::SubgraphPropertyRegistry::Get()->CreateSubgraphProperty(prop_name);
+  nnvm::Symbol ret = src.Copy();
+  nnvm::Graph g;
+  g.outputs = ret.outputs;
+  g = InferForwardAttrs(g, arg_shapes, arg_dtypes, arg_stypes, default_ctx,
+                        ctx_map, in_arg_ctxes, aux_state_ctxes);
+  subgraph_prop->SetAttr("graph", g);
```
zheng-da: This interface doesn't allow customizing `g`. I think it's better to provide an interface such as `set_graph` and allow the child class to override this virtual method.

src/executor/graph_executor.cc
```diff
@@ -1718,6 +1860,11 @@ Executor *Executor::SimpleBind(nnvm::Symbol symbol,
                                std::unordered_map<std::string, NDArray>* shared_buffer,
                                Executor* shared_exec) {
   auto exec = new exec::GraphExecutor();
+  if (!exec->subgraph_property().empty()) {
+    symbol = exec::PartitionGraph(symbol, exec->subgraph_property(), arg_shape_map, arg_dtype_map,
```
ZhennanQin: Many subgraph pass can only be used on inference, eg. default_subgraph_op doesn't have FGradient attribute and can't support backward computation. Maybe default_subgraph_op will support backward in future, but some mkldnn fusion pass is designed to be used in inference only. Shall we allow subgraph_property to describe if it's inference only and apply it as it requires?

ZhennanQin:
Can this change cover hybridize gluon model? I don't see the change in cached_op.

ZhennanQin:
Another question: what's the sequence should we execute below 2 kinds of subgraph pass:
1. the pass to cover all operators that backends support.(just like default_subgraph_op)
2. Fusion pass to combine several ops into a new ops.

I'm asking this is because, if we want to apply 2 after 1, then we need to have the ability to traverse into subgraph node to do fusion. Do we support this?

src/executor/graph_executor.cc
reminisce: Is it necessary to indicate whether a subgraph property is for inference only or not?

reminisce:
@ZhennanQin CachedOp will be submitted through a another PR. We want to make this PR as small as possible.

I think what you described about fusion can be done in graph partitioning as well. In `CreateSubgraphNode()`, you can replace the subgraph symbol with a new subgraph symbol consisting of fused operators. Does this make sense?

src/executor/graph_executor.cc
reminisce: I think this can be done in `CreateSubgraphSelector()`?

src/executor/graph_executor.cc
ZhennanQin: Yes, that's also my question. We indeed have some fusion passes only for inference, we need a way to disable it on training. If we can't describe this in subgraph property, then we need to find another way to indicate that.

ZhennanQin:
@reminisce Then we need to call PartitionGraph() inside CreateSubgraphNode() to traverse nodes, otherwise we have to implement a separate traverse method. If that meet your expectation, then I'm OK with this solution.

src/executor/graph_executor.cc
zheng-da: Subgraph selector is created by each node. I feel it's more intuitive to customize `g` somewhere subgraph property is created.

reminisce:
@ZhennanQin Yes, I fine with that.

PR #12110, opened 2018-08-09
[MXNET-730][WIP] Scala test in nightly

lanking520:
## Description ##
This is the PR for migrating Scala test to nightly build. The action I applied here is to provide a new function in `Util` that can check which test to run and how many epochs needed to run. All Integration test will run 1 epoch in default and nightly will run more for performance and accuracy checking.

@nswamy @yzhliu @andrewfayres @marcoabreu 

### Essentials ###

### DIFF
--- a/scala-package/examples/src/main/scala/org/apache/mxnetexamples/Util.scala
+++ b/scala-package/examples/src/main/scala/org/apache/mxnetexamples/Util.scala
@@ -21,9 +21,15 @@ import java.io.File
 import java.net.URL
 
 import org.apache.commons.io.FileUtils
+import org.slf4j.LoggerFactory
 
 object Util {
-
+  /**
+    * a Download wrapper with retry scheme on Scala
+    * @param url the URL for the file
+    * @param filePath the path to store the file
+    * @param maxRetry maximum retries will take
+    */
   def downloadUrl(url: String, filePath: String, maxRetry: Option[Int] = None) : Unit = {
     val tmpFile = new File(filePath)
     var retry = maxRetry.getOrElse(3)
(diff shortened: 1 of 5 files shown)


scala-package/examples/src/main/scala/org/apache/mxnetexamples/Util.scala
```diff
@@ -42,4 +48,30 @@ object Util {
     }
    if (!success) throw new Exception(s"$url Download failed!")
   }
+
+  /**
+    * This Util is designed to manage the tests in CI
+    * @param name the name of the test
+    * @return runTest and number of epoch
+    */
+  def testManager(name: String) : (Boolean, Int) = {
```
nswamy: why is this needed ? can this not be controlled through make/maven targets and configurations?

lanking520: Then there will be 10+ lines maven configuration and fragments of checking in the tests to achieve the same functionalities. I believe placing them all together into one place is better to manage

nswamy: you'll need to change everytime a new test it added. Ideally your test should be configurable and you choose what settings you want to run in the target/script.

I don't think it should be controlled through the code here

lanking520: Let's say this way. If someday you want to add a new flag to split the test further for some other use cases, you need to open each test and add that flag there. Assume you have 10 tests in the plate, then 10 duplicated code change will be applied in the tests. This solution, one change would be necessary. Apart from that, this holds the same scenario of downloading files, we create a standard input and output into a single place will definite helps us for future maintenance.

andrewfayres: Agreeing with Naveen here. Convention is that we control this type of behavior through something like environment variables, build parameters, or config files.

One other thing is that we should try to avoid having code which is unique to our CI testing inside the src folder.

larroy: isn't that why we have src/test/scala in the standard layout?

andrewfayres: Yes, you're right. I meant to say we should avoid having it in src/main folder.

lanking520:
Close this for now as it is not yet settled to an agreement. Will think about a better way and open again.

Issue #11883, opened 2018-07-25
ImageIter last batch / batch padding behavior


kjchalup:
## Description
mxnet.image.ImageIter lacks a way to choose behavior on last batch (when batch size doesn't divide the size of the dataset). In contrast, [mxnet.io.NDArrayIter](https://mxnet.incubator.apache.org/api/python/io/io.html#mxnet.io.NDArrayIter) has a `last_batch_handle` keyword arg which allows the user to choose what to do.

This can lead to unexpected behavior:

```
# Create a diter for an image dataset containing a total of 8 images.
diter = mxnet.image.ImageIter(batch_size=5, ...)
batch1 = diter.next()
batch2 = diter.next()
print(b2.label)
```

output:
```
[
 [3.0000000e+00 1.0000000e+00 4.0000000e+00 2.5249697e-29 2.8025969e-45]
 <NDArray 5 @cpu(0)>]
```

In this case, the last batch has only 3 legit labels, and the last two labels are garbage. I think the user is expected to check the padding manually:

```
print('batch1.pad = {}, batch2.pad = {}'.format(batch1.pad, batch2.pad))
```

output:
```
batch1.pad = 5, batch2.pad=2
```

But this is not documented in the [mxnet.image API reference](https://mxnet.incubator.apache.org/api/python/image/image.html). In addition, b2.data is *not* filled with garbage -- it contains 5 legit images, adding to the confusion.

sandeep-krishnamurthy:
Thanks for reporting. Added to the backlog items we are working on - https://issues.apache.org/jira/browse/MXNET-737

knwng:
The data_batch and label_batch is initialized with a mx.nd.empty() object. Under my configuration, when the data generator is exhausted during filling a data batch, the data_batch will not be padded or filled with new data but remain garbage data even nan, which will raise an index out of bound error when calculate CrossEntropy metric. 
I'm currently change nd.empty into nd.zeros to initialize data_batch in the source code, but I think a automatically refilling or padding mechanism should be added to ImageIter.

PR #12131, opened 2018-08-10
[MXNET-737][WIP] Add last batch handle for imageiter

stu1130:
## Description ##

Add `last_batch_handle` parameter to ImageIter based on #11883
* `last_batch_handle` support `pad`(default), `discard`, `roll_over`

Note that reading record files(.rec) without shuffle(i.e. sequential read) didn't support 'discard'
 
### Essentials ###

### Changes ###
 * Add last_batch_handle feature to ImageIter and adjust test_imageiter to test last_batch_handle parameter
 * Change the shuffle behavior to the same as NDArrayIter where shuffling the data only happen during the iterator initialization

## Comments ##
N/A
@zhreshold

### DIFF
--- a/python/mxnet/image/image.py
+++ b/python/mxnet/image/image.py
@@ -1047,7 +1047,7 @@ class ImageIter(io.DataIter):
     path_imgidx : str
         Path to image index file. Needed for partition and shuffling when using .rec source.
     shuffle : bool
-        Whether to shuffle all images at the start of each iteration or not.
+        Whether to shuffle all images or not.
         Can be slow for HDD.
     part_index : int
         Partition index.
@@ -1059,6 +1059,9 @@ class ImageIter(io.DataIter):
         Label name for provided symbols.
     dtype : str
         Label data type. Default: float32. Other options: int32, int64, float64
+    last_batch_hanle :  str, optional
+        How to handle the last batch. This parameter can be ‘pad’, ‘discard’ or ‘roll_over’.
+        'discard' is not support when reading from record file(.rec) withouting shuffle(=False)
     kwargs : ...
         More arguments for creating augmenter. See mx.image.CreateAugmenter.
     """
@@ -1066,9 +1069,11 @@ class ImageIter(io.DataIter):
     def __init__(self, batch_size, data_shape, label_width=1,
                  path_imgrec=None, path_imglist=None, path_root=None, path_imgidx=None,
                  shuffle=False, part_index=0, num_parts=1, aug_list=None, imglist=None,
-                 data_name='data', label_name='softmax_label', dtype='float32', **kwargs):
+                 data_name='data', label_name='softmax_label', dtype='float32', last_batch_handle='pad', **kwargs):
         super(ImageIter, self).__init__()
         assert path_imgrec or path_imglist or (isinstance(imglist, list))
+        # Reading from record file without shuffle didn't support 'discard' for last_batch_handle
+        assert not path_imgrec or shuffle or last_batch_handle != 'discard'
         assert dtype in ['int32', 'float32', 'int64', 'float64'], dtype + ' label not supported'
         num_threads = os.environ.get('MXNET_CPU_WORKER_NTHREADS', 1)
         logging.info('Using %s threads for decoding...', str(num_threads))
@@ -1084,6 +1089,7 @@ class ImageIter(io.DataIter):
             else:
                 self.imgrec = recordio.MXRecordIO(path_imgrec, 'r')  # pylint: disable=redefined-variable-type
                 self.imgidx = None
+            self.path_imgrec = path_imgrec
         else:
             self.imgrec = None
 
@@ -1130,9 +1136,11 @@ class ImageIter(io.DataIter):
         self.data_shape = data_shape
         self.label_width = label_width
 
-        self.shuffle = shuffle
         if self.imgrec is None:
             self.seq = imgkeys
+            # shuffle 
+            if shuffle:
+                random.shuffle(self.seq)
         elif shuffle or num_parts > 1:
             assert self.imgidx is not None
             self.seq = self.imgidx
@@ -1149,22 +1157,53 @@ class ImageIter(io.DataIter):
         else:
             self.auglist = aug_list
         self.cur = 0
+        self.is_iterated_over = False
+        self._imgrec = None
+        # handle the last batch
+        if self.seq and last_batch_handle == 'discard':
+            new_seq_n = len(self.seq) - len(self.seq) % batch_size
+            self.seq = self.seq[:new_seq_n]
+
+        self.last_batch_handle = last_batch_handle
+        self.num_image = len(self.seq) if self.seq is not None else None
         self.reset()
 
     def reset(self):
         """Resets the iterator to the beginning of the data."""
-        if self.shuffle:
-            random.shuffle(self.seq)
+        if self.last_batch_handle == 'roll_over' and \
+            self.seq and \
+            self.cur > self.num_image:
+            assert self.num_image is not None, 'imgrec without shuffle is not supported'
+            self.cur = (self.cur % self.num_image) % self.batch_size
+        elif self.last_batch_handle == 'roll_over' and \
+                self.is_iterated_over:
+                
+            if self._imgrec is not None:
+                self.imgrec = self._imgrec
+            self.is_iterated_over = False
+        else:
+            self.cur = 0
+            self.is_iterated_over = False
+            if self.imgrec is not None:
+                self.imgrec.reset()
+
+    def hard_reset(self):
+        """Resets the iterator and ignore roll over data"""
         if self.imgrec is not None:
             self.imgrec.reset()
         self.cur = 0
+        self.is_iterated_over = False
 
-    def next_sample(self):
+    def next_sample(self, imgrec=None):
         """Helper function for reading in next sample."""
         if self.seq is not None:
-            if self.cur >= len(self.seq):
+            if self.cur < self.num_image:
+                idx = self.seq[self.cur]
+            elif self.num_image % self.batch_size != 0 and \
+                self.cur < self.batch_size * ((self.num_image // self.batch_size) + 1):
+                idx = self.seq[self.cur % self.num_image]
+            else:
                 raise StopIteration
-            idx = self.seq[self.cur]
             self.cur += 1
             if self.imgrec is not None:
                 s = self.imgrec.read_idx(idx)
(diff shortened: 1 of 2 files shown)


python/mxnet/image/image.py
```diff
@@ -1207,8 +1244,31 @@ def next(self):
...
+
+        # handle padding for sequential read
+        if self.seq is None:
+            pad = batch_size - i
+            #  pad the last batch by creating a new MXRecordIO
+            self._imgrec = recordio.MXRecordIO(self.path_imgrec, 'r')
+            if pad != 0:
+                _ = self.iterate(batch_data, batch_label, i, self._imgrec)
+            
+            if self.last_batch_handle != 'roll_over':
```
zhreshold: is it possible to cache last batch in memory if roll_over? I think open/closing recordio is problematic

python/mxnet/image/image.py
```diff
@@ -1059,16 +1059,21 @@ class ImageIter(io.DataIter):
         Label name for provided symbols.
     dtype : str
         Label data type. Default: float32. Other options: int32, int64, float64
+    last_batch_hanle :  str, optional
```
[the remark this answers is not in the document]
stu1130: The original [NDArrayIter](https://mxnet.apache.org/api/python/io/io.html#mxnet.io.NDArrayIter) call it `last_batch_handle` but I think the name `last_batch` already give much information of what it does. So I would go with your suggestion.

python/mxnet/image/image.py
```diff
@@ -1059,16 +1059,21 @@ class ImageIter(io.DataIter):
...
+        How to handle the last batch. This parameter can be ‘pad’, ‘discard’ or ‘roll_over’.
+        'discard' is not support when reading from record file(.rec) withouting shuffle(=False)
```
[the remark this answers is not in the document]
stu1130: The reason why discard is not supported is that when we read the rec file sequentially we don't know how many images in the file. Therefore, there is no way we can precalculate the number of images we need to discard. The only two solutions that I came up with is 
* iterate the file during the initialization of data iterator 
* allow users to input the number of the images
The first solution would take lots of time if the file is large during the initialization. The second one is not user-friendly. So I decided to give up this option.

python/mxnet/image/image.py
```diff
@@ -1149,22 +1158,44 @@ def __init__(self, batch_size, data_shape, label_width=1,
...
     def reset(self):
         """Resets the iterator to the beginning of the data."""
-        if self.shuffle:
-            random.shuffle(self.seq)
+        if self.last_batch != 'roll_over' or \
+            self._is_allowed_reading is True:
+            if self.imgrec is not None:
+                self.imgrec.reset()
+            self.cur = 0
+        self._is_allowed_reading = True
```
sandeep-krishnamurthy: why do we need to set this again?

python/mxnet/image/image.py
```diff
@@ -1149,22 +1158,44 @@ def __init__(self, batch_size, data_shape, label_width=1,
...
+        """Resets the iterator and ignore roll over data"""
         if self.imgrec is not None:
             self.imgrec.reset()
         self.cur = 0
+        self._is_allowed_reading = True
 
     def next_sample(self):
         """Helper function for reading in next sample."""
+        if self._is_allowed_reading is False:
+            raise StopIteration
```
sandeep-krishnamurthy: Please add user comprehensible message when raising errors.

python/mxnet/image/image.py
```diff
@@ -1149,22 +1158,44 @@ def __init__(self, batch_size, data_shape, label_width=1,
...
         if self.seq is not None:
-            if self.cur >= len(self.seq):
+            if self.cur < self.num_image:
+                idx = self.seq[self.cur]
+            else:
+                if self.last_batch != 'discard':
```
sandeep-krishnamurthy: should this be != discard and == rollover?

python/mxnet/image/image.py
```diff
@@ -1207,8 +1237,28 @@ def next(self):
         except StopIteration:
             if not i:
                 raise StopIteration
+        return i
 
-        return io.DataBatch([batch_data], [batch_label], batch_size - i)
+    def next(self):
+        """Returns the next batch of data."""
+        batch_size = self.batch_size
+        c, h, w = self.data_shape
```
sandeep-krishnamurthy: Are we assuming always channels_first?

tests/python/unittest/test_image.py
```diff
@@ -130,29 +129,59 @@ def test_color_normalize(self):
...
+
+            for test in test_list:
+                imglist = im_list if test == 'imglist' else None
+                path_imglist = fname if test == 'path_imglist' else None
+                
+                test_iter = mx.image.ImageIter(2, (3, 224, 224), label_width=1, imglist=imglist, 
+                    path_imglist=path_imglist, path_root='', dtype=dtype)
+                for _ in range(3):
+                    for batch in test_iter:
+                        pass
```
sandeep-krishnamurthy: can we assert the returned batch size and shape here?

python/mxnet/image/image.py
stu1130: No matter it is `pad` or `roll_over`, we need to "reset the cursor"(assign 0 to the self.cur) to do the padding

python/mxnet/image/image.py
```diff
@@ -1149,22 +1158,44 @@ def __init__(self, batch_size, data_shape, label_width=1,
         else:
             self.auglist = aug_list
         self.cur = 0
+        self._is_allowed_reading = True
+        self._cached_data = None
+        # handle the last batch
+        if self.seq and last_batch == 'discard':
+            new_seq_n = len(self.seq) - len(self.seq) % batch_size
+            self.seq = self.seq[:new_seq_n]
```
zhreshold: this is not correct, you now change the behavior from initial shuffle to per epoch shuffle, so self.seq must be shuffled in reset, and the slice can not be fixed.

python/mxnet/image/image.py
```diff
@@ -1179,17 +1210,16 @@ def next_sample(self):
...
             if s is None:
+                if self.last_batch != 'discard':
+                    self.imgrec.reset()
                 raise StopIteration
             header, img = recordio.unpack(s)
             return header.label, img
 
-    def next(self):
-        """Returns the next batch of data."""
+    def iterate(self, batch_data, batch_label, start=0):
```
zhreshold: use `_collect_batch` as name instead

python/mxnet/image/image.py
```diff
@@ -1207,8 +1237,28 @@ def next(self):
...
+        batch_size = self.batch_size
+        c, h, w = self.data_shape
+        batch_data = nd.empty((batch_size, c, h, w))
+        batch_label = nd.empty(self.provide_label[0][1])
+        i = self.iterate(batch_data, batch_label)
+        # calculate the padding
+        pad = batch_size - i
+        # handle padding for sequential read
+        if pad != 0:
+            if self.seq is not None:
```
zhreshold: the logic of last_batch here is kind weird

python/mxnet/image/image.py
zhreshold: actually `_batchify` could be better IMO

python/mxnet/image/image.py
stu1130: Yes we are

python/mxnet/image/image.py
stu1130: According to (NDArrayIter)[https://mxnet.incubator.apache.org/_modules/mxnet/io.html#NDArrayIter], it shuffled only during the initialization. But you are right. We should shuffle for each epoch.

python/mxnet/image/image.py
```diff
@@ -1207,8 +1230,25 @@ def next(self):
...
+        batch_size = self.batch_size
+        c, h, w = self.data_shape
+        batch_data = nd.empty((batch_size, c, h, w))
+        batch_label = nd.empty(self.provide_label[0][1])
+        i = self._batchify(batch_data, batch_label)
+        # calculate the padding
+        pad = batch_size - i
+        # handle padding for 'pad' and 'roll_over' for the last batch
+        if pad != 0:
+            if self.last_batch == 'discard':
```
stu1130: change the padding logic here

python/mxnet/image/image.py
```diff
@@ -1207,8 +1230,25 @@ def next(self):
...
+        """Returns the next batch of data."""
+        batch_size = self.batch_size
+        c, h, w = self.data_shape
+        batch_data = nd.empty((batch_size, c, h, w))
+        batch_label = nd.empty(self.provide_label[0][1])
+        i = self._batchify(batch_data, batch_label)
+        # calculate the padding
+        pad = batch_size - i
+        # handle padding for 'pad' and 'roll_over' for the last batch
+        if pad != 0:
```
stu1130: refactor the padding logic here

python/mxnet/image/image.py
```diff
@@ -1149,22 +1155,49 @@ def __init__(self, batch_size, data_shape, label_width=1,
...
+                self._is_allowed_reading = True
+
+    def hard_reset(self):
+        """Resets the iterator and ignore roll over data"""
+        if self.seq is not None and self.shuffle:
             random.shuffle(self.seq)
         if self.imgrec is not None:
             self.imgrec.reset()
         self.cur = 0
+        self._is_allowed_reading = True
```
zhreshold: `self._allow_read` for shorter but same meaning

tests/python/unittest/test_image.py
```diff
@@ -130,29 +129,70 @@ def test_color_normalize(self):
...
+                    first_batch_roll_over.data[0][1].asnumpy(), first_image.asnumpy())
+                assert first_batch_roll_over.pad == 2
+                # test iteratopr work properly after calling reset several times when last_batch_handle is roll_over
+                for _ in test_iter:
+                    pass
+                test_iter.reset()
+                first_batch_roll_over_twice = test_iter.next()
+                assert np.array_equal(
+                    first_batch_roll_over_twice.data[0][2].asnumpy(), first_image.asnumpy())
+                assert first_batch_roll_over_twice.pad == 1
```
zhreshold: assert second epoch with size 6?
Also can you add test for `shuffle=True`, just for sanity test, no value assert is required for shuffle mode.

tests/python/unittest/test_image.py
```diff
@@ -130,29 +129,81 @@ def test_color_normalize(self):
...
+                # we've called next once
+                i = 1
+                for _ in test_iter:
+                    i += 1
+                # test the third epoch with size 6
+                assert i == 6
+                # test shuffle option for sanity test
+                test_iter = mx.image.ImageIter(3, (3, 224, 224), label_width=1, imglist=imglist, shuffle=True,
```
stu1130: add shuffle test case

tests/python/unittest/test_image.py
```diff
@@ -130,29 +129,81 @@ def test_color_normalize(self):
...
+                first_batch_roll_over_twice = test_iter.next()
+                assert np.array_equal(
+                    first_batch_roll_over_twice.data[0][2].asnumpy(), first_image.asnumpy())
+                assert first_batch_roll_over_twice.pad == 1
+                # we've called next once
+                i = 1
+                for _ in test_iter:
+                    i += 1
+                # test the third epoch with size 6
+                assert i == 6
```
stu1130: test epoch size

python/mxnet/image/image.py
```diff
@@ -1149,22 +1155,49 @@ def __init__(self, batch_size, data_shape, label_width=1,
...
         """Resets the iterator to the beginning of the data."""
-        if self.shuffle:
+        if self.seq is not None and self.shuffle:
+            random.shuffle(self.seq)
+        if self.last_batch_handle != 'roll_over' or \
+            self._cache_data is None:
+            if self.imgrec is not None:
+                self.imgrec.reset()
+            self.cur = 0
+            if self._allow_read is False:
```
sandeep-krishnamurthy: nit: just self._allow_read = True would work here?

stu1130: yes, it works as well. Only iter with 'pad' would change this flag though.

PR #12157, opened 2018-08-14
Subgraph API for integrating accelerators with MXNet

reminisce:
## Description ##
This is a joint work of @reminisce and @zheng-da on implementing subgraph API for integrating backend accelerators (TVM, MKLDNN, TensorRT, nGraph, etc.) with MXNet. See [design proposal](https://cwiki.apache.org/confluence/display/MXNET/Unified+integration+with+external+backend+libraries).

## Highlighted deliverables
- A graph partitioning algorithm accepting custom partitioning rules (through subclassing `SubgraphProperty`).
- A naive graph partitioning rule called `DefaultSubgraphProperty` partitioning graphs based upon provided operator names. For example, in the left graph, operators B and C are designated to be grouped into a subgraph operator node. After applying the partitioning algorithm, nodes B and C are replaced by a subgraph operator node in the graph on the right-hand-side.
image

- `DefaultSubgraphOperator`. A stateful operator storing a subgraph for execution by `CachedOp`.
- `NDArray` version number for tracking mutations of `NDArray`s. This is useful, for example, in a stateful MKLDNN subgraph operator for checking whether input data has changed since the last forward call to decide if to trigger data format conversion.
- Graph partitioning enabled in `GraphExecutor` for the forward pass. Triggered by an environment variable `MXNET_SUBGRAPH_BACKEND`.

## Coming soon after this PR
- Backward pass support in `DefaultSubgraphOperator`.
- Support graph partitioning for Gluon/CacheOp.
- Tutorial for developers on integrating backend accelerators using the Subgraph API.
- Real backend accelerator integration with MXNet.

### Essentials ###


pengzhao-intel:
Thanks @reminisce it's really helpful for us :)

@zhennanqin @lvtao, please take a look the compatibility of current code with our interface :)

src/operator/subgraph/default_subgraph_property.h
```diff
@@ -0,0 +1,81 @@
...
+ */
+class DefaultSubgraphProperty: public SubgraphProperty {
+ public:
+  static SubgraphPropertyPtr Create() { return std::make_shared<DefaultSubgraphProperty>(); }
+  virtual nnvm::NodePtr CreateSubgraphNode(const nnvm::Symbol &sym,
+                                           const int subgraph_id = 0) const {
+    nnvm::NodePtr n = nnvm::Node::Create();
+    n->attrs.op = Op::Get("_default_subgraph_op");
+    n->attrs.name = "_default_subgraph_op" + std::to_string(subgraph_id);
+    n->attrs.subgraphs.push_back(std::make_shared<nnvm::Symbol>(sym));
```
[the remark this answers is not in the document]
reminisce: `n` is a newly created node and its `attrs.subgraphs` is empty before `push_back` is called. It's actually up to you to decide which index is for the `sym` in your specific subgraph property. Just make sure you get the correct subgraph in your subgraph operator implementation.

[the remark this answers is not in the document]
ZhennanQin: I heard that control dependence will use attrs.subgraphs as well. I'm not sure how this attribute will be used for dependence purpose.  Do we have any instruction to define the sequence for attrs.subgraphs?Eg. always put subgraph symbol in index 0, and following dependence purpose node.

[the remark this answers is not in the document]
reminisce: As far as I know, there is no fixed convention of doing this. @zheng-da 
I don't think you need to worry about this if you can ensure the consistency between your specific subgraph property and subgraph operator.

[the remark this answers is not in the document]
ZhennanQin: We can ensure the consistency between subgraph property and subgraph operator, and we also need to ensure nobody changes it after subgraph partition. It's better to add some check for this consistency in case it breaks.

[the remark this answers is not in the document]
reminisce: One thing you can do to enforce the consistency is defining some enums as indices of forward graph, backward graph, fused graph, etc in `attrs.subgraphs`, and always use those enums to get the desired subgraphs. It also increases the code readability.

src/operator/subgraph/subgraph_property.h
```diff
@@ -0,0 +1,132 @@
...
+// This op name set is for setting the names of operators that should be grouped into
+// subgraphs. In practice, every backend accelerator should have a predefined name set.
+// This set is only used for the testing purpose.
+// key: property name, value: op name set
+typedef dmlc::ThreadLocalStore<std::unordered_map<std::string, std::unordered_set<std::string>>>
+  SubgraphPropertyOpNameSet;
+
+#define MXNET_REGISTER_SUBGRAPH_PROPERTY(Name, SubgraphPropertyType) \
+  static DMLC_ATTRIBUTE_UNUSED auto __make_ ## SubgraphPropertyType ## _ ## Name ## __ = \
+    SubgraphPropertyRegistry::Get()->__REGISTER__(#Name, &SubgraphPropertyType::Create)
```
anirudh2290: Can we implement and call \_\_REGISTER_OR_GET\_\_ here (which checks if it is part of map else calls \_\_REGISTER\_\_). This may be needed because it is possible that \_\_REGISTER\_\_ may be called twice on the same name leading to exception.

reminisce: This is intended. The registration can only happen once.

anirudh2290: With \_\_REGISTER_OR_GET\_\_ also register will happen only once. To explain an use case where register wont work: lets say I create a custom subgraph property and call MXNET_REGISTER_SUBGRAPH_PROPERTY from inside the namespace. Now i include this custom subgraph property header file in multiple source files. If they are part of different translation units \_\_REGISTER\_\_ called multiple times and fails since the static variable scope is limited to the a single translation unit.

reminisce: That's a good catch. I will fix it. Thanks.

src/operator/subgraph/default_subgraph_op.cc
```diff
@@ -0,0 +1,113 @@
...
+    output_ptrs.push_back(&nd);
+  }
+#if DEBUG_SUBGRAPH
+  for (size_t i = 0; i < inputs.size(); ++i) {
+    LOG(INFO) << "inputs[" << i << "].version = " << inputs[i].version();
+  }
+  for (size_t i = 0; i < outputs.size(); ++i) {
+    LOG(INFO) << "outputs[" << i << "].version = " << outputs[i].version();
+  }
+#endif
```
zheng-da: should we clean this?

src/operator/subgraph/partition_graph.cc
```diff
@@ -0,0 +1,775 @@
...
+    + ", index=" + std::to_string(entry.index) + ", version=" + std::to_string(entry.version);
+  LOG(INFO) << ret;
+}
+
+void PrintNodeEntries(const std::vector<nnvm::NodeEntry*>& entries) {
+  for (size_t i = 0; i < entries.size(); ++i) {
+    PrintNodeEntry(*entries[i]);
+  }
+}
+#endif
```
zheng-da: delete the debug code?

src/ndarray/ndarray.cc
```diff
@@ -39,6 +39,7 @@
 #include "../operator/tensor/matrix_op-inl.h"
 #include "../operator/tensor/init_op.h"
 #include "../operator/nn/mkldnn/mkldnn_base-inl.h"
+#include "../engine/engine_impl.h"
```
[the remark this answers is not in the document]
reminisce: Removed.

src/imperative/imperative_utils.h
```diff
@@ -436,7 +436,8 @@ inline void PushFComputeEx(const FComputeEx& fn,
       }
     };
 
-  if (exec_type == ExecType::kCrossDeviceCopy) {
+  if (exec_type == ExecType::kCrossDeviceCopy
+      || exec_type == ExecType::kSubgraphExec) {
```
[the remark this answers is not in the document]
zheng-da: Yes, I added this line. It's actually not necessary because this is PushFComputeEx, which only pushes FComputeEx. could you help me remove the line?

src/operator/subgraph/default_subgraph_op.cc
reminisce: Ah, I see. Removed now.

src/engine/naive_engine.cc
```diff
@@ -165,14 +178,22 @@ class NaiveEngine final : public Engine {
     }
     CHECK(this->req_completed_)
         << "NaiveEngine only support synchronize Push so far";
+    // increment var version
+    for (auto var : mutable_vars) {
```
anirudh2290: In case of deletevariable, will this increment the var version after it is deleted ?

src/executor/graph_executor.cc
```diff
@@ -1428,6 +1430,146 @@ GraphExecutor::CachedSegOpr GraphExecutor::CreateCachedSegOpr(size_t topo_start,
...
+      CHECK(i1 < arg_names.size());
+      CHECK_EQ(arg_names[i1], input_names[i]);
+      arg_shapes.push_back(in_args[i1].shape());
+      arg_dtypes.push_back(in_args[i1].dtype());
+      arg_stypes.push_back(in_args[i1].storage_type());
+      in_arg_ctxes[i1] = in_args[i1].ctx();
+      ++i1;
+    }
+  }
+  return PartitionGraph(src, prop_name, arg_shapes, arg_dtypes, arg_stypes,
```
anirudh2290: should we check for prop_name here. Currently only default is supported correct ?

src/operator/subgraph/default_subgraph_property.cc
```diff
@@ -0,0 +1,76 @@
...
+                                           const int subgraph_id = 0) const {
+    nnvm::NodePtr n = nnvm::Node::Create();
+    n->attrs.op = Op::Get("_default_subgraph_op");
+    n->attrs.name = "_default_subgraph_op" + std::to_string(subgraph_id);
+    n->attrs.subgraphs.push_back(std::make_shared<nnvm::Symbol>(sym));
+    return n;
+  }
+  virtual SubgraphSelectorPtr CreateSubgraphSelector() const {
+    return std::make_shared<ContainOpSelector>(
+        this->GetAttr<std::unordered_set<std::string>>("op_names"));
```
anirudh2290: Can we override GetAttr for DefaultSubgraphProperty, so that it informs the user that the set has to be provided by the user using C API to test default subgraph property? This will also be a good example for developers writing custom subgraph property.

src/executor/graph_executor.cc
```diff
@@ -1428,6 +1430,146 @@ GraphExecutor::CachedSegOpr GraphExecutor::CreateCachedSegOpr(size_t topo_start,
...
+static nnvm::Symbol PartitionGraph(const nnvm::Symbol& src,
+                                   const std::string& prop_name,
+                                   const nnvm::ShapeVector& arg_shapes,
+                                   const nnvm::DTypeVector& arg_dtypes,
+                                   const StorageTypeVector& arg_stypes,
+                                   const Context& default_ctx,
+                                   const std::map<std::string, Context>& ctx_map,
+                                   const std::vector<Context>& in_arg_ctxes,
+                                   const std::vector<Context>& aux_state_ctxes) {
+  auto subgraph_prop = op::SubgraphPropertyRegistry::Get()->CreateSubgraphProperty(prop_name);
```
zheng-da: If `prop_name` doesn't exist, we need to handle the failure of creating a subgraph property.

src/executor/graph_executor.cc
zheng-da: nice catch. we should check in here https://github.com/apache/incubator-mxnet/pull/12157/files/8435acff465725e406bbd467902c3e6bb5c83139#diff-d8b5a5b027d00584737fb6486cba38b9R1467

src/operator/subgraph/default_subgraph_property.cc
zheng-da: this is a templated method. I don't think we can make it a virtual method. but I don't really know what you mean here.

src/executor/graph_executor.cc
reminisce: There is a check inside `CreateSubgraphProperty`.

src/engine/naive_engine.cc
reminisce: Good catch. Should move this before `exec_fun` runs.

src/executor/graph_executor.cc
reminisce: There is a check inside CreateSubgraphProperty.

piiswrong:
Let's not put this in example since the CAPIs are not meant to be called by users

src/operator/subgraph/default_subgraph_property.cc
anirudh2290: To explain my use case: Lets say a user forgot to call setattr for op_names when using the default_subgraph_property. It would fial with the error that op_names cannot be found in subgraph property. A better error message would have been useful. There may be such cases where classes subclassing SubgraphProperty may want to also implement logic for SetAttr and GetAttr in the subclass.

reminisce: It's not meant for users to use and is there only for the flexible testing purpose. Furthermore, template method cannot be virtual. But I still cannot understand your use case. How different could SetAttr and GetAttr be in a subclass than the base?

anirudh2290: for example you only want certain attributes to be set for the subgraph property and you want to do some sanity check on whether the attrs has can be supported or not. For example supported op_names.

zheng-da: for your use case, i think it's fine without customization. Users of subgraph API can write their own code to do sanity check. The main problem is a templated method can't be virtual.

src/operator/subgraph/partition_graph.cc
```diff
@@ -0,0 +1,774 @@
...
+                            SubgraphSelectorPtr subgraph_selector,
+                            const int label,
+                            const size_t snid,
+                            const std::vector<SimpleNodePtr>& simple_nodes,
+                            std::vector<nnvm::Node*>* subgraph_nodes) {
+  std::unordered_set<const nnvm::Node*> excluded_nodes;
+  const size_t max_num_retry = simple_nodes.size() * simple_nodes.size();
+  size_t count = 0;
+  bool success = false;
+  while (!success && count < max_num_retry) {
```
anirudh2290: Can you explain the logic behind ssetting max_num_retry to square of simple_nodes.size(). Shouldnt it just be simple_nodes.size()

src/operator/subgraph/partition_graph.cc
```diff
@@ -0,0 +1,774 @@
...
+                                                    " a loop in the graph";
+      ++count;
+      const nnvm::Node* top = s.top();
+      s.pop();
+      if (top == ancestor) {
+        return true;
+      }
+      for (const auto& entry : top->inputs) {
+        // when searching for the ancestor, the path cannot cross any subgraph node
+        auto it = std::find(snodes.begin(), snodes.end(), entry.node.get());
```
anirudh2290: nit: maybe add check for it == snodes.end() since it wont happen that it will find the input node in the subgraph nodes ?

src/executor/graph_executor.cc
```diff
@@ -42,6 +43,7 @@ using namespace mxnet::common;
 GraphExecutor::GraphExecutor() {
   log_verbose_ = dmlc::GetEnv("MXNET_EXEC_VERBOSE_LOGGING", false);
   need_grad_ = false;
+  subgraph_property_ = dmlc::GetEnv("MXNET_SUBGRAPH_BACKEND", std::string());
```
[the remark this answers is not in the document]
reminisce: Right now, there is no real accelerator integrated with MXNet. Intel team is going to submit their work on MKLDNN using this API set. We can add this evn var along with their PR.

eric-haibin-lin:
LGTM

eric-haibin-lin:
Looks like all comments are addressed. Merging it now

Issue #10856, opened 2018-05-09
Failed OpenMP assertion when loading MXNet compiled with DEBUG=1


anirudh2290:
Reproduction steps:
```
cmake -DUSE_CUDA=ON -DUSE_CUDNN=ON -DUSE_MKLDNN=ON -DBLAS=Open -DUSE_DEBUG=1 -DCMAKE_BUILD_TYPE=Debug -GNinja .. && ninja && cd ..

[INFO] Setting module np/mx/python random seeds, use MXNET_MODULE_SEED=1107463865 to reproduce.
Test should successfully run its course of multi-process/forked data loader without errors ... Assertion failure at kmp_runtime.cpp(6481): __kmp_team_pool == __null.
OMP: Error #13: Assertion failure at kmp_runtime.cpp(6481).
OMP: Hint: Please submit a bug report with this message, compile and run commands used, and machine configuration info including native compiler and operating system versions. Faster response will be obtained by including all program sources. For information on submitting this issue, please see https://bugs.llvm.org/.
Assertion failure at kmp_runtime.cpp(6481): __kmp_team_pool == __null.
OMP: Error #13: Assertion failure at kmp_runtime.cpp(6481).
OMP: Hint: Please submit a bug report with this message, compile and run commands used, and machine configuration info including native compiler and operating system versions. Faster response will be obtained by including all program sources. For information on submitting this issue, please see https://bugs.llvm.org/.
^C[INFO] Setting test np/mx/python random seeds, use MXNET_TEST_SEED=39684479 to reproduce.

----------------------------------------------------------------------
Ran 1 test in 2.106s

OK
```

```
----------System Info----------
('Platform     :', 'Linux-4.4.0-1055-aws-x86_64-with-Ubuntu-16.04-xenial')
('system       :', 'Linux')
('node         :', 'ip-172-31-21-146')
('release      :', '4.4.0-1055-aws')
('version      :', '#64-Ubuntu SMP Thu Apr 5 17:06:36 UTC 2018')
----------Hardware Info----------
('machine      :', 'x86_64')
('processor    :', 'x86_64')
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                32
On-line CPU(s) list:   0-31
Thread(s) per core:    2
Core(s) per socket:    16
Socket(s):             1
NUMA node(s):          1
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 79
[...]

Issue #11417, opened 2018-06-27
libomp.so dependency (need REAL fix)


eLvErDe:
Hello,

I've read nearly most related ticket which usually ends up disabling openmp. Sounds like the worst idea ever.

Is it possible to have an official statement regarding this ? mxnet 1.0.0 was building just fine with -DUSE_OPENMP=1 without adding this unresolved library dependency.

Do you know what has been changed ? Is that really solved ? By which commit, which release ?

Thanks in advance,

Best regards, Adam.

eLvErDe:
Here's a list of related tickets/commits ending with "disabling openmp":

https://github.com/apache/incubator-mxnet/commit/85215b6176ef3612d198f590268a6595b86565fb
https://github.com/apache/incubator-mxnet/issues/10011
https://github.com/apache/incubator-mxnet/issues/10230

marcoabreu:
Hello Adam,
just to clarify: Are you talking about ARM? If yes, please specify which version exactly and how you are trying to compile. There have been some changes recently and we'd like to make sure we are on the same page.
Best regards,
Marco

eLvErDe:
Absolutely not, regular X64. Multiple variants being built with/without MKL-DNN and/or CUDA.

From what I saw, the difference between 1.0.0 and 1.1.1 is in root tree `CMakeLists.txt`:

In 1.0.0:
```
  if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/openmp/CMakeLists.txt)
    # Intel/llvm OpenMP: https://github.com/llvm-mirror/openmp
```
In 1.1.0:
```
  if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/openmp/CMakeLists.txt AND SYSTEM_ARCHITECTURE STREQUAL "x86_64" AND NOT MSVC)
    # Intel/llvm OpenMP: https://github.com/llvm-mirror/openmp
```

The main difference is that this openmp folder DID NOT exist in 1.0.0 sources tree, so it falls back building using system openmp:
```
  else()
    if(OPENMP_FOUND)
      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
[...]

PR #10951, opened 2018-05-15
[MXNET-545] Fix broken cython build

asitstands:
## Description ##

Fix broken cython module for ndarray and also ensure the compatibility with the recent version (0.28) of cython (#10295, #10738). See this [comment](10951#issuecomment-397852430) for the details of the changes.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

Issue #9744, opened 2018-02-08
Performance regression when OMP_NUM_THREADS environment variable is not set


cbalioglu:
## Affected Versions
- 1.0.0
- 1.1.0.rc0
- 1.1.0.rc1

## Build Options
- USE_OPENMP=ON

## Environments Tested
OS: 
- Amazon Linux 2012.03 (Kernel 4.4.48)
- Amazon Linux 2017.03 (Kernel 4.4.48)

Hardware:
- AWS m4.2xlarge
- AWS c4.2xlarge
- AWS p3.2xlarge

## Impact
The benchmark numbers below are produced by training a symbolic NTM algorithm on a quad core (8 logical cores) Intel Xeon E5-2676 CPU. However the regression is reproducible by running any type of symbolic graph on a CPU context.

OMP_NUM_THREADS | 1.0.0 + GNU OpenMP | 1.0.0 + Intel OpenMP | 1.1.0.rc0 + GNU OpenMP
----------------------|----------------------|----------------------|--------------------------
\- | 75s | 116s | 66s
4 | 65s | 77s | 48s

## Root Cause
The `mxnet::engine::OpenMP` class has already an [implementation](https://github.com/apache/incubator-mxnet/blob/3761f2f36e7e335f17e5c1dcbfe215de27f73fd5/src/engine/openmp.cc#L53) for automatically setting up the optimal number of OpenMP threads in case the `OMP_NUM_THREADS` environment variable is not explicitly set. However this implementation does not completely solve the regression problem as the `omp_set_num_threads` function is thread-local.

The setup logic referred above is only [called](https://github.com/apache/incubator-mxnet/blob/3761f2f36e7e335f17e5c1dcbfe215de27f73fd5/src/engine/openmp.cc#L102) during the static initialization of the `OpenMP` class in the main thread. However the `mxnet::engine::ThreadedEngine` class utilizes OpenMP not only in the main thread but also in worker threads. As `omp_set_num_threads` is never called in these threads, they attempt to use all logical cores for parallelized regions, which causes the regression.

## Solution
The `omp_set_num_threads` function has to be explicitly called in all threads in case the `OMP_NUM_THREADS` environment variable is not set by the user.

## Actions Required
As part of this investigation I already implemented an ad-hoc fix and plan to polish and make it production ready in the upcoming days. At this stage no action is requested from the main contributors.

cjolivier01:
Thank you for the excellent detailed explanation.

[...]

PR #12160, opened 2018-08-14
Remove conflicting llvm OpenMP from cmake builds

lebeg:
## Description ##

Using a different OpenMP library from the one supplied by the compiler can lead to undefined behaviour. Multiple problems were reported by users of MXNet.

If there is need to use OpenMP library provided by llvm, one should compile MXNet with latest Clang, which is now supported as a test stage by the CI.

### Essentials ###
- [x] This is a minor change
- [x] Changes are complete

### Changes ###
- [x] Removed OpenMP overwrite from CMakeLists.txt
- [x] Removed otherwise not used OpenMP submodule

### Comments ###
Fixes:
* https://github.com/apache/incubator-mxnet/issues/11417 libomp.so dependency (need REAL fix)
* https://github.com/apache/incubator-mxnet/issues/10856 test_multi_worker_forked_data_loader fails with DEBUG=1

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -431,34 +431,12 @@ endif()
 # ---[ OpenMP
 if(USE_OPENMP)
   find_package(OpenMP REQUIRED)
-  # This should build on Windows, but there's some problem and I don't have a Windows box, so
-  # could a Windows user please fix?
-  if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/openmp/CMakeLists.txt
-     AND SYSTEM_ARCHITECTURE STREQUAL "x86_64"
-     AND NOT MSVC
-     AND NOT CMAKE_CROSSCOMPILING)
 
-    # Intel/llvm OpenMP: https://github.com/llvm-mirror/openmp
-    set(OPENMP_STANDALONE_BUILD TRUE)
-    set(LIBOMP_ENABLE_SHARED TRUE)
-    add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/openmp)
-    list(REMOVE_ITEM mxnet_LINKER_LIBS iomp5)
-    list(APPEND mxnet_LINKER_LIBS omp)
-    if(UNIX)
-      list(APPEND mxnet_LINKER_LIBS pthread)
-    endif()
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
-    add_definitions(-DMXNET_USE_OPENMP=1)
-  else()
-    if(OPENMP_FOUND)
-      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
-      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
-      set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
-      set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
-      add_definitions(-DMXNET_USE_OPENMP=1)
-    endif()
-  endif()
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
+
 elseif(UNIX AND NOT ANDROID)
   list(APPEND mxnet_LINKER_LIBS pthread)
 endif()
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,9 +10,6 @@
 [submodule "3rdparty/dlpack"]
 	path = 3rdparty/dlpack
 	url = https://github.com/dmlc/dlpack
-[submodule "3rdparty/openmp"]
-	path = 3rdparty/openmp
-	url = https://github.com/llvm-mirror/openmp
 [submodule "3rdparty/googletest"]
 	path = 3rdparty/googletest
 	url = https://github.com/google/googletest.git
--- a/ci/jenkins/Jenkins_steps.groovy
+++ b/ci/jenkins/Jenkins_steps.groovy
@@ -30,11 +30,11 @@ mx_lib_cython = 'lib/libmxnet.so, lib/libmxnet.a, 3rdparty/dmlc-core/libdmlc.a, 
 mx_pip = 'build/*.whl'
 
 // mxnet cmake libraries, in cmake builds we do not produce a libnvvm static library by default.
-mx_cmake_lib = 'build/libmxnet.so, build/libmxnet.a, build/3rdparty/dmlc-core/libdmlc.a, build/tests/mxnet_unit_tests, build/3rdparty/openmp/runtime/src/libomp.so'
-mx_cmake_lib_cython = 'build/libmxnet.so, build/libmxnet.a, build/3rdparty/dmlc-core/libdmlc.a, build/tests/mxnet_unit_tests, build/3rdparty/openmp/runtime/src/libomp.so, python/mxnet/_cy2/*.so, python/mxnet/_cy3/*.so'
+mx_cmake_lib = 'build/libmxnet.so, build/libmxnet.a, build/3rdparty/dmlc-core/libdmlc.a, build/tests/mxnet_unit_tests'
+mx_cmake_lib_cython = 'build/libmxnet.so, build/libmxnet.a, build/3rdparty/dmlc-core/libdmlc.a, build/tests/mxnet_unit_tests, python/mxnet/_cy2/*.so, python/mxnet/_cy3/*.so'
 // mxnet cmake libraries, in cmake builds we do not produce a libnvvm static library by default.
 mx_cmake_lib_debug = 'build/libmxnet.so, build/libmxnet.a, build/3rdparty/dmlc-core/libdmlc.a, build/tests/mxnet_unit_tests'
-mx_cmake_mkldnn_lib = 'build/libmxnet.so, build/libmxnet.a, build/3rdparty/dmlc-core/libdmlc.a, build/tests/mxnet_unit_tests, build/3rdparty/openmp/runtime/src/libomp.so, build/3rdparty/mkldnn/src/libmkldnn.so.0'
+mx_cmake_mkldnn_lib = 'build/libmxnet.so, build/libmxnet.a, build/3rdparty/dmlc-core/libdmlc.a, build/tests/mxnet_unit_tests, build/3rdparty/mkldnn/src/libmkldnn.so.0'
 mx_mkldnn_lib = 'lib/libmxnet.so, lib/libmxnet.a, lib/libiomp5.so, lib/libmkldnn.so.0, lib/libmklml_intel.so, 3rdparty/dmlc-core/libdmlc.a, 3rdparty/tvm/nnvm/lib/libnnvm.a'
 mx_tensorrt_lib = 'build/libmxnet.so, lib/libnvonnxparser_runtime.so.0, lib/libnvonnxparser.so.0, lib/libonnx_proto.so, lib/libonnx.so'
 mx_lib_cpp_examples = 'lib/libmxnet.so, lib/libmxnet.a, 3rdparty/dmlc-core/libdmlc.a, 3rdparty/tvm/nnvm/lib/libnnvm.a, 3rdparty/ps-lite/build/libps.a, deps/lib/libprotobuf-lite.a, deps/lib/libzmq.a, build/cpp-package/example/*, python/mxnet/_cy2/*.so, python/mxnet/_cy3/*.so'
--- a/tests/nightly/JenkinsfileForBinaries
+++ b/tests/nightly/JenkinsfileForBinaries
@@ -19,7 +19,7 @@
 //This is a Jenkinsfile for nightly tests. The format and some functions have been picked up from the top-level Jenkinsfile
 
 mx_lib = 'lib/libmxnet.so, lib/libmxnet.a, 3rdparty/dmlc-core/libdmlc.a, 3rdparty/tvm/nnvm/lib/libnnvm.a'
-mx_cmake_lib = 'build/libmxnet.so, build/libmxnet.a, build/3rdparty/dmlc-core/libdmlc.a, build/tests/mxnet_unit_tests, build/3rdparty/openmp/runtime/src/libomp.so'
+mx_cmake_lib = 'build/libmxnet.so, build/libmxnet.a, build/3rdparty/dmlc-core/libdmlc.a, build/tests/mxnet_unit_tests'
 
 node('utility') {
   // Loading the utilities requires a node context unfortunately


szha:
you may want to check this on mac, where the "special" clang doesn't provide -fopenmp option

lebeg:
@szha yes, but for AppleClang we are failing on this [line](https://github.com/apache/incubator-mxnet/pull/12160/files#diff-af3b638bc2a3e6c650974192a53c7291R372) already, so no difference made in this case.

vandanavk:
Is this PR good to go? https://github.com/apache/incubator-mxnet/pull/10951 is waiting on this

KellenSunderland:
@szha @lebeg Wouldn't we expect the CMake behaviour would detect if you have OpenMP (or don't have it) and take appropriate actions without an error?  Why do we even need to specify OPEN_MP=0/1?  If we were setting this automatically we would not run into different versions of MacOS building/failing right?  What do you think?

lebeg:
@KellenSunderland well the option USE_OPENMP i think is good to control explicit need for OpenMP (`REQUIRED` in cmake's find_package) or disable it if needed for testing for example. We could potentially have a string option with values like `Auto`, `Require`, `Disable`, but that would be a separate slightly bigger change.

KellenSunderland:
@lebeg Yes different PR for sure.  In the future it'd be great if we had 'auto', 'on', and 'off, and the default was auto.

lebeg:
@larroy Actually, if you install clang via brew (`brew install llvm`) then OpenMP would work.

lebeg:
And it would use the OpenMP lib that we had in the submodule 😁

KellenSunderland:
FYI here's what the deps look like before / after this patch is applied.  Clearly moving from linking solely to the 3rdparty llvm openmp to the gomp version.  https://gist.github.com/KellenSunderland/3e2852d929c444f6ba36c3c9b46194b8

lebeg:
Yes, this certainly fixes https://github.com/apache/incubator-mxnet/issues/10856 since it's kmp API (thread affinity) failure available only in llvm omp.

CMakeLists.txt
```diff
@@ -370,32 +370,12 @@ endif()
 # ---[ OpenMP
 if(USE_OPENMP)
   find_package(OpenMP REQUIRED)
-  # This should build on Windows, but there's some problem and I don't have a Windows box, so
-  # could a Windows user please fix?
-  if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/openmp/CMakeLists.txt
```
cjolivier01: Are you high?  Why would you remove this?

cjolivier01:
"Using a different OpenMP library from the one supplied by the compiler has side effects."  <-- that doesn't really give any useful information

Using the llvm OpenMP gives a significant performance improvement over gcc's libgomp.  It's loaded anyway for many builds such as mkl.  It should not be removed 7unless there is a VERY good reason to, and no other reasonable solution around the problem exists.

lebeg:
Thanks for your review @cjolivier01 :)

Actually for MKL builds another OpenMP library is provided (from Intel), specifically for MKLML. And it is stated in [MKLDNN README](https://github.com/intel/mkl-dnn/blob/master/README.md) that:

> Using GNU compiler with -fopenmp and -liomp5 options will link the application with both Intel and GNU OpenMP runtime libraries. This will lead to undefined behavior of the application.

In MXNet exactly this happens but with llvm openmp.

lebeg:
I consider improving stability and lowering complexity of the build an enough good reason to do this change.

cjolivier01:
btw, I did a lot of work with mxnet and cython ( https://github.com/apache/incubator-mxnet/tree/cython ), and never had issues with this openmp library (always build with cmake).  More likely the problems you see are symptoms of something else that can be adjusted.  This OpenMP library has 66% less overhead than libgomp when running parallel OMP regions, which adds up to a lot of time quickly.

larroy:
@cjolivier01 are these numbers available or is there some script / instructions on how to reproduce the benchmarks / how are they conducted?

lebeg:
I am about to get some performance numbers on how MXNet performs without the included as submodule OpenMP. I do not think there is any change I can make based on the reviewing comments.

lupesko:
+1 to @cjolivier01 ask to update the issue description to be more informative, and would be great to see perf data.
Thanks for the contribution!

lebeg:
@cjolivier01 

> Using the llvm OpenMP gives a significant performance improvement over gcc's libgomp.

You can see at [this line](https://github.com/apache/incubator-mxnet/pull/12160/files#diff-af3b638bc2a3e6c650974192a53c7291L384) that not GNU OpenMP was replaced, but Intel OpenMP. I have reasons to believe that llvm OpenMP is not significantly faster (if at all) than Intel OpenMP.

Can you provide measurements you did on performance? What build configurations you were using?

azai91:
if the performance gains are neglible (like <5%) with LLVM then I think it is worth removing.

access2rohit:
@lebeg LGTM. Would definitely like to see perf numbers though

vandanavk:
@cjolivier01 Please provide any inputs/references about how to test performance with and without this change. This will help in testing and proceeding with the review process.

vrakesh:
@lebeg Requesting an update on this, has the feedback been addressed?

Roshrini:
@cjolivier01 Could you take a look at this PR again? This PR is pending for a long time now.

ankkhedia:
@lebeg ping!
Could you please follow up on the comments above and move this PR?

asitstands:
How about to introduce a compile time flag to control the use of intel openmp? If entire removing is not a good option currently, a convenient way to turn it off would be enough.

kalyc:
@cjolivier01 could you please provide feedback on the PR as mentioned above?

cjolivier01:
I am cloising this PR because this is roughly equivalent to commenting out an assert as a bugfix. There is something else going badly that is causing the assert as I stated before  in several places and that should be addressed.  Since there is no investigation tot he actual root cause, i am closing this PR.

larroy:
@cjolivier01 you are closing this PR, but what solution is there for this problematic issue? https://github.com/apache/incubator-mxnet/issues/10856

lebeg:
@cjolivier01 as I have stated above, the root cause was identified: it is the undefined behaviour which we are exploiting by mixing OpenMP libraries from different compilers.

larroy:
@lebeg what is the solution you would propose? Is what is presented in this PR?

lebeg:
This PR is exactly reflecting my proposal.

larroy:
@lebeg can we provide evidence that this fixes any of the issues that we have observed? I think this would be valuable to move the conversation forward.

larroy:
@cjolivier01 we are linking against two different versions of omp, is this going to cause undefined behaviour in your opinion?


```
$ ldd build/tests/mxnet_unit_tests  | grep omp
	libomp.so => /home/piotr/devel/mxnet/mxnet_other/build/3rdparty/openmp/runtime/src/libomp.so (0x00007f697bc55000)
	libXcomposite.so.1 => /usr/lib/x86_64-linux-gnu/libXcomposite.so.1 (0x00007f6970a60000)
	libgomp.so.1 => /usr/lib/x86_64-linux-gnu/libgomp.so.1 (0x00007f69660cd000)
```

lebeg:
@pengzhao-intel can we have your input on this?

lebeg:
I cross post the things I already sent to the @dev list:

The issue was closed, but I am strong in my opinion that it's the right thing to do.

**Background**

If you want to use OpenMP pragmas in your code for parallelization you would supply a special flag to the compiler:

- Clang / -fopenmp
https://openmp.llvm.org/

- GCC / -fopenmp
https://gcc.gnu.org/onlinedocs/libgomp/Enabling-OpenMP.html

- Intel / [Q]openmp
https://software.intel.com/en-us/node/522689#6E24682E-F411-4AE3-A04D-ECD81C7008D1

- Visual Studio: /openmp (Enable OpenMP 2.0 Support)
https://msdn.microsoft.com/en-us/library/tt15eb9t.aspx

Each of the compilers would enable the `#pragma omp` directive during C/C++ compilation and arrange for automatic linking of the OpenMP runtime library supplied by each compiler separately.

Thus, to use the advantages of an OpenMP implementation one has to compile the code with the corresponding compiler.

Currently, in MXNet CMake build scripts a bundled version of llvm OpenMP is used ([here](https://github.com/apache/incubator-mxnet/blob/master/CMakeLists.txt#L400-L405) and [here](https://github.com/apache/incubator-mxnet/tree/master/3rdparty)) to replace the OpenMP library supplied by the compiler.

I will quote here the README from the MKL-DNN (Intel(R) Math Kernel Library for Deep Neural Networks):

"Intel MKL-DNN uses OpenMP* for parallelism and requires an OpenMP runtime library to work. As different OpenMP runtimes may not be binary compatible it's important to ensure that only one OpenMP runtime is used throughout the application. Having more than one OpenMP runtime initialized may lead to undefined behavior resulting in incorrect results or crashes." [link](https://github.com/intel/mkl-dnn/blame/master/README.md#L261-L265)

And:

"Using GNU compiler with -fopenmp and -liomp5 options will link the application with both Intel and GNU OpenMP runtime libraries. This will lead to undefined behavior of the application." [link](https://github.com/intel/mkl-dnn/blame/master/README.md#L278-L280)

As can be seen from ldd for MXNet:

```
$ ldd build/tests/mxnet_unit_tests | grep omp
    libomp.so => /.../mxnet/build/3rdparty/openmp/runtime/src/libomp.so (0x00007f697bc55000)
    libgomp.so.1 => /usr/lib/x86_64-linux-gnu/libgomp.so.1 (0x00007f69660cd000)
```

**Performance**

The only performance data related to OpenMP in MXNet I was able to find is here:
https://github.com/apache/incubator-mxnet/issues/9744#issuecomment-367711172

Which in my understanding is testing impact of different environment variables for the same setup (using same bundled OpenMP library).

The libraries may differ in implementation and the [Thread Affinity Interface](https://software.intel.com/en-us/node/522691) may have significant impact on performance.

All compilers support it:

- Clang / KMP_AFFINITY
https://github.com/clang-ykt/openmp/blob/master/runtime/src/kmp_affinity.cpp

- GCC / GOMP_CPU_AFFINITY
https://gcc.gnu.org/onlinedocs/gcc-4.7.1/libgomp/GOMP_005fCPU_005fAFFINITY.html

- Intel / KMP_AFFINITY
https://software.intel.com/en-us/node/522689#6E24682E-F411-4AE3-A04D-ECD81C7008D1

- Visual Studio / SetThreadAffinityMask
https://docs.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-setthreadaffinitymask

I see no reason why this submodule should be kept.

pengzhao-intel:
Thanks @lebeg it's a long thread and I need to go through the change and issue first.
Will back to you soon.

pengzhao-intel:
Also @xinyu-intel  @ZhennanQin they have the experience of different platform, MacOS, Windows and compiler-level knowledge.

lebeg:
[Link](https://lists.apache.org/thread.html/4827f0f742b6e7e070da350ea81226d059401527f3072ce8b33c1fdf@%3Cdev.mxnet.apache.org%3E) to the dev list discussion.

larroy:
Please reopen this PR given the data collected in the wiki:

https://cwiki.apache.org/confluence/display/MXNET/Benchmarking+MXNet+with+different+OpenMP+implementations

ZhennanQin:
@larroy Thanks for continuously watching this issue. My previous compiler developer experience tells me, build project with certain version of certain compiler component is not a good idea. Because compiler may have assumption about the version of omp be used.  We should let toolchain to make the decision, this can avoid many unexpected behavior.
So, reopen this PR +1.

stsukrov:
Yup.
Let's do it properly.

larroy:
@stsukrov can you ping @lebeg ?

lebeg:
Guys, I can not reopen this PR - it was closed by a committer and I assume it takes one to open it again. The branch is still there, I can make any necessary changes if needed after it will be reopened.

cjolivier01:
Is all this still related to an assertion?  Why does it assert?

You should also turn off operators tuning before running performance tests and include mnist.

larroy:
This is not (only) about the assertion. Unit tests hang when compiling with OpenMP in Linux. I always compile without openmp for this reason. 

There are several related issues that we hope that this PR will make go away:

https://github.com/apache/incubator-mxnet/issues/14999

@cjolivier01 is suggesting to run the tests with MXNET_USE_OPERATOR_TUNING=0 so OMP is always used for operators, but is this a useful scenario? If with the provided openmp and tuning we get similar perfomance why does it matter if some operators are tuned or not?

larroy:
I spent some time debugging this assertion, it was long ago, but if I remember correctly is related to the problematic use of pthread_atfork in https://github.com/apache/incubator-mxnet/blob/master/src/initialize.cc#L76
Which causes openmp to be reinitialized and the assertion to be triggered.
I'm not sure what you want to do with this information or what you suggest to do with that problem and that problematic piece of code which caused hard bugs in the past, but I would say it's out of the scope of this PR.

piyushghai:
@lebeg Can you rebase with the latest master branch ?

larroy:
Mnist results:

This PR:

```
(py3_venv) piotr@ec2 cpu:0: ~/m/e/module [omp]> /usr/bin/time python mnist_mlp.py
[23:07:54] ../src/io/iter_mnist.cc:110: MNISTIter: load 60000 images, shuffle=1, shape=(100,784)
[23:07:54] ../src/io/iter_mnist.cc:110: MNISTIter: load 10000 images, shuffle=1, shape=(100,784)
epoch 000: accuracy=0.123883
epoch 001: accuracy=0.679983
INFO:root:Epoch[0] Train-accuracy=0.133283
INFO:root:Epoch[0] Time cost=1.047
INFO:root:Epoch[0] Validation-accuracy=0.252500
INFO:root:Epoch[1] Train-accuracy=0.725100
INFO:root:Epoch[1] Time cost=1.017
INFO:root:Epoch[1] Validation-accuracy=0.866900
batch 000 acc: 0.910
batch 020 acc: 0.890
batch 040 acc: 0.870
batch 060 acc: 0.910
batch 080 acc: 0.840
validation Accuracy: 0.867
accuracy=0.866900
174.10user 8.20system 0:07.17elapsed 2541%CPU (0avgtext+0avgdata 501636maxresident)

####
100 epoch:


4733.37user 13.77system 2:14.40elapsed 3531%CPU (0avgtext+0avgdata 501476maxresident)k
0inputs+0outputs (0major+649505minor)pagefaults 0swaps



```


Master 3b663ef6cc9bc6992d769b8ae7fa8e72e4f3201b

```
(py3_venv) piotr@ec2 cpu:0: ~/m/e/module [master]> /usr/bin/time python mnist_mlp.py
[23:08:09] ../src/io/iter_mnist.cc:110: MNISTIter: load 60000 images, shuffle=1, shape=(100,784)
[23:08:09] ../src/io/iter_mnist.cc:110: MNISTIter: load 10000 images, shuffle=1, shape=(100,784)
epoch 000: accuracy=0.123883
epoch 001: accuracy=0.679983
INFO:root:Epoch[0] Train-accuracy=0.133283
INFO:root:Epoch[0] Time cost=0.781
INFO:root:Epoch[0] Validation-accuracy=0.252500
INFO:root:Epoch[1] Train-accuracy=0.725100
INFO:root:Epoch[1] Time cost=0.935
INFO:root:Epoch[1] Validation-accuracy=0.866900
batch 000 acc: 0.910        
batch 020 acc: 0.890        
batch 040 acc: 0.870        
batch 060 acc: 0.910        
batch 080 acc: 0.840        
validation Accuracy: 0.867  
accuracy=0.866900           
105.58user 83.53system 0:06.92elapsed 2733%CPU (0avgtext+0avgdata 502812maxresident)k
0inputs+0outputs (0major+266146minor)pagefaults 0swaps

#####
With 100 epochs:

2654.15user 2779.46system 2:33.42elapsed 3541%CPU (0avgtext+0avgdata 503584maxresident)k
0inputs+0outputs (0major+697918minor)pagefaults 0swaps



```

larroy:
Gluon unit tests seem to run faster with the system omp:

```
test_gluon.test_np_shape_parameters ... (2, 2016)
(2, 16)
ok
test_gluon.test_gluon_param_load ... ok

Ran 107 tests in 70.958s

OK (SKIP=18)
206.25user 490.32system 1:12.04elapsed 966%CPU (0avgtext+0avgdata 13262532maxresident)k
0inputs+976152outputs (0major+33768421minor)pagefaults 0swaps
```


Master:

```
(2, 16)
ok
test_gluon.test_gluon_param_load ... ok

Ran 107 tests in 259.936s

OK (SKIP=18)
949.25user 1507.26system 4:21.16elapsed 940%CPU (0avgtext+0avgdata 13279492maxresident)k
0inputs+976152outputs (0major+34278780minor)pagefaults 0swaps
```

larroy:
https://github.com/apache/incubator-mxnet/issues/14999

cjolivier01:
As you can see, standard MKL builds also use libomp.

ldd /usr/local/lib/python3.6/dist-packages/mxnet/libmxnet.so
        linux-vdso.so.1 (0x00007ffc989cf000)
        libmklml_intel.so => /usr/local/lib/python3.6/dist-packages/mxnet/libmklml_intel.so (0x00007f0afb7c1000)
        libiomp5.so => /usr/local/lib/python3.6/dist-packages/mxnet/libiomp5.so (0x00007f0afb3e5000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f0afb1dd000)
        libmkldnn.so.0 => /usr/local/lib/python3.6/dist-packages/mxnet/libmkldnn.so.0 (0x00007f0afa7ba000)
        libgfortran.so.3 => /usr/local/lib/python3.6/dist-packages/mxnet/libgfortran.so.3 (0x00007f0afa493000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f0afa28f000)
        libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f0af9f06000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f0af9b68000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f0af9950000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f0af9731000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f0af9340000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f0b073f4000)
        libquadmath.so.0 => /usr/local/lib/python3.6/dist-packages/mxnet/libquadmath.so.0 (0x00007f0af9100000)

larroy:
Why do you mean by your latest comment and what additional info are you requesting?

What is your comment regarding the speed improvements and solving the issues with the bundled OpenMP please elaborate on your veto and how to move forward.

PR #12161, opened 2018-08-14
[WIP] A solution to prevent zombie containers locally and in CI

larroy:
## Description ##
This PR adds mechanisms on the build scripts to cleanup docker containers when there is a cancellation, either from the user by sending SIGINT / SIGTERM or when the process inside the container ends with an error.

It also propagates the environment variables from Jenkins which are used by the process tree killer to identify runaway processes so processes inside the container are killed when the Jenkins job is stopped.

With this patch we catch SIGINT and SIGTERM and install a handler on atexit which cleans (stops) docker containers which were created during execution of the build.

We also switch to python docker API for better management of running containers.

Since Jenkins is suppossed to send SIGINT / SIGTERM on job cancellation this should also provent zombie containers in CI.

### Essentials ###


ci/build.py
```diff
@@ -23,7 +23,11 @@
 """
 
 __author__ = 'Marco de Abreu, Kellen Sunderland, Anton Chernov, Pedro Larroy'
-__version__ = '0.1'
+__version__ = '0.8'
+
+def pip_install(pkg):
```
lebeg: We shouldn't install anything during a build.py run, all requirements should be there already

ci/build.py
```diff
@@ -37,11 +41,54 @@
...
 from typing import *
 from util import *
+try:
+    pip_install('docker')
+    import docker
+    import docker.models
+    import docker.errors
+    import signal
+    import atexit
+except ImportError as e:
```
lebeg: A usual import error I think should be enough

ci/build.py
```diff
@@ -37,11 +41,54 @@
...
+    def add_container(self, container: docker.models.containers.Container):
+        assert isinstance(container, docker.models.containers.Container)
+        self.containers.add(container)
+
+    def remove_container(self, container: docker.models.containers.Container):
+        assert isinstance(container, docker.models.containers.Container)
+        self.containers.remove(container)
+
+    def _cleanup_containers(self):
+        if self.containers:
```
lebeg: Shouldn't we just return if not?

ci/build.py
```diff
@@ -37,11 +41,54 @@
...
+            logging.warning("Cleaning up containers")
+        docker_client = docker.from_env()
+        try:
+            stop_timeout = int(os.environ.get("DOCKER_STOP_TIMEOUT", 3))
```
lebeg: Could we have "3" in a named variable?

ci/build.py
```diff
@@ -37,11 +41,54 @@
...
+        except Exception as e:
+            stop_timeout = 3
+        for container in self.containers:
+            try:
+                container.stop(timeout=stop_timeout)
+                logging.warning("☠: %s", trim_container_id(container.id))
```
lebeg: Should we have a more meaningful warning text than ☠?

ci/build.py
```diff
@@ -37,11 +41,54 @@
...
+            except Exception as e:
+                logging.exception(e)
```
lebeg: That might be a good place for a warning

ci/build.py
lebeg: Do you think we need a warning level here?

ci/build.py
```diff
@@ -121,7 +185,7 @@ def build_docker(platform: str, docker_binary: str, registry: str, num_retries:
     # container match permissions of the local user. Same for the group.
     #
     # These variables are used in the docker files to create user and group with these ids.
-    # see: docker/install/ubuntu_adduser.sh
+    # see: dockerfiles/install/ubuntu_adduser.sh
```
lebeg: Let's not move the files in this change for now. We probably would like this change to be merged to release branches as well and we'd need to keep it as crisp as possible.

ci/build.py
```diff
@@ -180,63 +245,133 @@ def default_ccache_dir() -> str:
...
+                  local_ccache_dir: str,
+                  cleanup: Cleanup,
                   dry_run: bool = False,
-                  interactive: bool = False) -> str:
+                  interactive: bool = False) -> int:
+    #
+    # Environment setup
+    #
+    environment = {
+        'CCACHE_MAXSIZE': '20G',
```
lebeg: You've changed the default value of this which was 500G previously. This would delete almost the whole cache on the ci. It makes sense to check whether the script is run on the ci and if not, set a lower value. But I would leave it for a later change and just fix the value by using the variable `CCACHE_MAXSIZE`  which was previously there.

ci/build.py
```diff
@@ -180,63 +245,133 @@ def default_ccache_dir() -> str:
...
+    # Equivalent command
+    docker_cmd_list = [docker_binary, 'run',
+               '--rm',
                '--shm-size={}'.format(shared_memory_size),
                '-v', "{}:/work/mxnet".format(mx_root),  # mount mxnet root
                '-v', "{}:/work/build".format(local_build_folder),  # mount mxnet/build for storing build artifacts
                '-v', "{}:/work/ccache".format(local_ccache_dir),
                '-u', '{}:{}'.format(os.getuid(), os.getgid()),
-               '-e', 'CCACHE_MAXSIZE={}'.format(CCACHE_MAXSIZE),
+               '-e', 'CCACHE_MAXSIZE=15G',
```
lebeg: See comment above

ci/build.py
```diff
@@ -180,63 +245,133 @@ def default_ccache_dir() -> str:
...
+        #
+        signal.pthread_sigmask(signal.SIG_BLOCK, {signal.SIGINT, signal.SIGTERM})
+        container = docker_client.containers.run(
+            tag,
+            detach=True,
+            command=command,
+            #auto_remove=True,
+            shm_size=shared_memory_size,
+            user='{}:{}'.format(os.getuid(), os.getgid()),
+            volumes={
```
lebeg: This looks so much better!

ci/build.py
```diff
@@ -180,63 +245,133 @@ def default_ccache_dir() -> str:
...
+        # Race condition:
+        # If the previous call is interrupted then it's possible that the container is not cleaned up
+        # We avoid by masking the signals temporarily
+        cleanup.add_container(container)
+        signal.pthread_sigmask(signal.SIG_UNBLOCK, {signal.SIGINT, signal.SIGTERM})
+        #
+        #############################
+
+        stream = container.logs(stream=True)
+        for chunk in stream:
```
lebeg: Why is this necessary?

ci/build.py
```diff
@@ -180,63 +245,133 @@ def default_ccache_dir() -> str:
...
+            sys.stdout.buffer.write(chunk)
+            sys.stdout.buffer.flush()
+        stream.close()
+        try:
+            logging.info("Waiting for status of container %s", trim_container_id(container.id))
+            wait_result = container.wait(timeout=10)
```
lebeg: We should have a variable for the timeout value. In what units is it?

ci/build.py
```diff
@@ -248,7 +383,11 @@ def load_docker_cache(tag, docker_registry) -> None:
     else:
         logging.info('Distributed docker cache disabled')
 
+
 def main() -> int:
+    import pprint as pp
```
lebeg: Should we have it where all other imports are?

ci/dockerfiles/runtime_functions.sh
```diff
@@ -305,6 +305,14 @@ build_ubuntu_cpu() {
     build_ubuntu_cpu_openblas
 }
 
+build_ubuntu_base_cpu() {
+    echo "skip"
```
lebeg: Do we need this?

ci/build.py
larroy: agree, just trying to see if I could get Jenkins to run. Will remove it before final merge.

ci/build.py
larroy: you suggest info?

ci/build.py
larroy: To get the output of the container.

ci/build.py
larroy: Agree, seconds. TODO.

ci/build.py
larroy: I plan to remove this, is just for diagnostics. TODO.

ci/dockerfiles/runtime_functions.sh
larroy: Yes, otherwise build.py -a doesn't work.

ci/build.py
larroy: TODO

ci/build.py
larroy: TODO

marcoabreu:
Could we temporarily close this PR to avoid any accidental executions on prod CI?

larroy:
@marcoabreu I don't think there would be any problem if it's executed by CI, what are you worried about specifically?

larroy:
@lebeg addressed your comments.

ci/build.py
larroy: ⚡

ci/build.py
larroy: ⚡

ci/build.py
larroy: ⚡

ci/build.py
larroy: ⚡

ci/build.py
lebeg: Yes, might be more appropriate

ci/dockerfiles/runtime_functions.sh
lebeg: Ok, I see

marcoabreu:
See Antons comment about your change deleting the entire shared CCache, for example. I'm closing now since there are being new pushes to this PR.

ci/build.py
larroy: done

PR #12181, opened 2018-08-15
Fixed backwards compatibility checker

lebeg:
## Description ##

The script was reporting missing mxnet installation.

### Essentials ###
- [x] Changes are complete

### Changes ###
- [x] Fixed python and pip versions to 3

### DIFF
--- a/tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
+++ b/tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
@@ -25,14 +25,14 @@ run_models() {
 	echo '=========================='
 	echo "Running training files and preparing models"
 	echo '=========================='
-	python model_backwards_compat_train.py
+	python3 model_backwards_compat_train.py
 	echo '=========================='
 }
 
 install_mxnet() {
 	version=$1
 	echo "Installing MXNet "$version
-	pip install mxnet==$version --user
+	pip3 install mxnet==$version --user
 }
 
 ## Cuts the string and gives only the major version part.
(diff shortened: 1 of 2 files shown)


ci/docker/runtime_functions.sh
```diff
@@ -925,13 +925,23 @@ nightly_test_javascript() {
 nightly_model_backwards_compat_test() {
     set -ex
     export PYTHONPATH=/work/mxnet/python/
+
+    cd /work/venv
+    virtualenv -p python3
+    source /work/venv/bin/activate
```
larroy: If we do a virtualenv why not do pip install instead of exporting PYTHONPATH?

Also, why not relative paths?

lebeg: python, python2, python3 are different now, as their corresponding pip, pip2, pip3
Maybe it would be better to make 2 different virtual env's for python2 and python3. Will try that.

piyushghai: Why do we want to go for a virtual env here ?

lebeg: Nevermind. It was just an attempt to insure python and pip are in sync.

tests/nightly/model_backwards_compatibility_check/train_mxnet_legacy_models.sh
```diff
@@ -25,14 +25,14 @@ run_models() {
...
-	python model_backwards_compat_train.py
+	python3 model_backwards_compat_train.py
 	echo '=========================='
 }
 
 install_mxnet() {
 	version=$1
 	echo "Installing MXNet "$version
-	pip install mxnet==$version --user
+	pip3 install mxnet==$version --user
```
piyushghai: Does this resolve the Python module not found issues we are seeing in the MBCC Train Stage on Jenkins ?

lebeg:
Closing due to https://github.com/apache/incubator-mxnet/pull/12223

PR #12195, opened 2018-08-16
[MXNet-744] Fix website build pipeline Python 3 issues

aaronmarkham:
## Description ##
Some Python 3 incompatibilities existed in the website build pipeline. When CI was updated to run docs build in Python 3 mode these surfaced. This PR addresses it and locks in the Python dependency versions to help prevent future breakage.

### Essentials ###

### Changes ###
* Fixed Sphinx plugin mxdoc.py Python 3 issues by using `six`.
* Added `six` to dependencies
* Consolidated Python requirements for docs to one file in `ci/docker/install`
* Pinned dependencies to specific version numbers
* `mock` was not pinned upstream from docs, so we were getting a later version than what docs had been pinned to, so this is upgraded.
* "Upgraded" a variety of packages that were not pinned to the latest in PyPI. Tests indicate things are good, and I'd rather pin things now to prevent future breakage.
* Added a separate `Jenkinsfile-dev` so I don't have to keep hacking the Jenkinsfile every time I need to test something from my fork. This way I can just point to it from the test Jenkins jobs directly, and no more code updates. 
* Updated CI's docker images to use the new Python requirements file

## Comments ##
* We can break out a separate requirements file if some split in deps appears between Python 2/3. For now the packages are the same.


aaronmarkham:
@marcoabreu @larroy Please take a look.

docs/Jenkinsfile-dev
```diff
@@ -0,0 +1,54 @@
+// -*- mode: groovy -*-
```
nswamy: where is this file used? why is it called Jenkinsfile-dev?

nswamy:
LGTM, but I am not sure where is the new Jenkinsfile-dev used. I will merge this to unblock CI and can continue to follow up.

aaronmarkham:
Login to Jenkins and look at test-website-build. That's what points to the Jenkinsfile-dev version. Email notifications for failures are off and it doesn't run code on restricted nodes. Might be a better way to do it, but seemed best for the situation.

PR #12199, opened 2018-08-16
bump timeout on Jenkins for docs/website to 120 min

aaronmarkham:
## Description ##
Patch the Jenkins timeout to run for up to 120 minutes. Also add settings for branches to turn on/off doc generation granularly. 

We're getting [timeouts](http://jenkins.mxnet-ci.amazon-ml.com/job/restricted-website-build/) at 60 minutes as the docs jobs are taking longer that. (Odd, since I had them down to 23 minutes for a full site build, but that was on my own instance, and last week prior to a variety of updates.) Maybe once the cache can kick in we'll see better times.

This should be investigated tomorrow, but in the meantime, let's unblock CI. This is a followup on #12195.

### DIFF
--- a/docs/Jenkinsfile
+++ b/docs/Jenkinsfile
@@ -21,7 +21,7 @@
 // See documents at https://jenkins.io/doc/book/pipeline/jenkinsfile/
 
 // timeout in minutes
-max_time = 60
+max_time = 120
 
 node('restricted-mxnetlinux-cpu') {
   // Loading the utilities requires a node context unfortunately
--- a/docs/Jenkinsfile-dev
+++ b/docs/Jenkinsfile-dev
@@ -21,7 +21,7 @@
 // See documents at https://jenkins.io/doc/book/pipeline/jenkinsfile/
 
 // timeout in minutes
-max_time = 60
+max_time = 120
 
 node('mxnetlinux-cpu') {
   // Loading the utilities requires a node context unfortunately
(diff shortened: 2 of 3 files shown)


aaronmarkham:
This finished successfully pointing at my branch:
http://jenkins.mxnet-ci.amazon-ml.com/job/test-website-build/
Build 24: http://jenkins.mxnet-ci.amazon-ml.com/job/test-website-build/24/

lebeg:
I see that the build took 38 minutes. Is there a need to increase the timeout?

marcoabreu:
The reason for the timeout is the scala doc generation. It's currently quite inefficient and sometimes could spike. I'd say that we can apply this timeout for now and then bring the times back down again. Wdyt?

aaronmarkham:
The Scala docs part is really slow. And when we release a new version it will run there too. I've talked to Qing and Andrew about a way to just use the prebuilt Maven package in the pipeline instead of building from source which will save time and skip some tests that seem to have recently been introduced, adding to time.

I also have clojure docs running as a new feature for master and any new branch. 

Then there's R docs. I'm supposed to something about that soon as they have been disabled. All these add up. 

But, BTW the new pipeline should retain builds, so old versions should get cached thereby lowering time... Eventually.

marcoabreu:
Thanks a lot, Aaron. Could you please turn it back on?

aaronmarkham:
The pipeline is humming along now. 4 builds and 0.11.0 is back in as well.
Also, I got rid of the time traveling home page, so when you view older versions you still see the latest news, but can access the old API docs. I think this is pretty cool.

PR #12191, opened 2018-08-15
[MXNET-696] Define cmp() in Python 3 for line 222

cclauss:
[__cmp()__ was removed in Python 3](https://docs.python.org/3/whatsnew/3.0.html#ordering-comparisons) so this PR recreates it if required leveraging the formula at http://python-future.org/compatible_idioms.html#cmp

## Description ##
Define cmp() in Python 3

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] Define cmp() in Python 3

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #12229, opened 2018-08-17
Revert "[MXNET-696] Define cmp() in Python 3 for line 222"

marcoabreu:
Reverts apache/incubator-mxnet#12191

We reverted the underlying change and thus this PR is not necessary (and is no longer compatible)

### DIFF
--- a/tests/nightly/model_backwards_compatibility_check/common.py
+++ b/tests/nightly/model_backwards_compatibility_check/common.py
@@ -29,13 +29,6 @@ from mxnet.gluon import nn
 import re
 from mxnet.test_utils import assert_almost_equal
 
-try:
-    cmp             # Python 2
-except NameError:
-    # See: https://docs.python.org/3.0/whatsnew/3.0.html#ordering-comparisons
-    def cmp(x, y):  # Python 3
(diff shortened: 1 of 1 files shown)


marcoabreu:
@piyushghai @cclaus fyi

haojin2:
Just wonder why are you also reverting a bunch of other un-related changes?

marcoabreu:
I just pressed the revert but, I don't know what GitHub is doing there ._.

Have a look at the diff, it looks good to me

marcoabreu:
I'm only on my phone right now, so I can't make a revert with the command line. In the end, the commits will be squashed so it should not make much of a difference as long as the diff is good

haojin2:
@marcoabreu Could you do a double-check please? Cause one of the commits on the list shows that it's reverting an important fix for the next release.

piyushghai:
The file diff looks good as it was the only change in that PR. 
Not sure, why the other commits are showing up here.

piyushghai:
Check PR #12231 . 
That includes just one commit required. 
@marcoabreu Feel free to close this one.

marcoabreu:
Thank you!

PR #12270, opened 2018-08-21
[MXNET-848] Pin dockcross base images in CI

KellenSunderland:
## Description ##
This PR will pin our dockcross base images to known working versions.  Pinning these images is necessary as we've had cases where following the 'latest' tag in dockerhub leads us to picking up broken, or incompatible images.  Though the dockcross images are generally high-quality we'd like to avoid these issues from blocking our CI in the future.

The disadvantage here is that we will lose visibility in valid failures for things like new cmake versions that potentially break on ARM builds.  We feel that these test should be done independently of PR validation, and would be best suited for the nightly test runs.

### Essentials ###

### DIFF
--- a/ci/docker/Dockerfile.build.android_armv7
+++ b/ci/docker/Dockerfile.build.android_armv7
@@ -18,7 +18,7 @@
 #
 # Dockerfile to build MXNet for Android ARMv7
 
-FROM dockcross/base:latest
+FROM mxnetci/dockcross-linux-base:08212018
 MAINTAINER Pedro Larroy "pllarroy@amazon.com"
 
 # The cross-compiling emulator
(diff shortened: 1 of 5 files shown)


KellenSunderland:
@marcoabreu

lebeg:
We need to enable OpenMP for other arm platforms, I've build the right dockcross images, once they are uploaded to dockerhub we can update the versions in this change and add armv7 to it.

KellenSunderland:
@lebeg: nice!  The images in these PR are the exact same images we were using prior to pining.  I'd recommend we initially stick with them, then give your fixed images new tags and test them thoroughly on the dev CI.  We can then slowly roll them out to prod (and should be able to easily roll them back).  What do you think?

lebeg:
@KellenSunderland good idea, let's do that

lebeg:
https://github.com/apache/incubator-mxnet/pull/12272 Enable armv7 with pinned docker images

Issue #9067, opened 2017-12-14
how to  convert parameters  dtype from float32 to float64 in gluon?


zhaoningning:
I need to do this because model produce very small values, and float32 is not suitable in 
this case, Does anyone who has any ideas about convert float32 parameter to float64 in gluon?

zhreshold:
Cast data to float64, and the corresponding weights in model will be initialized to float64 accordingly once you call `net(x)`
Or there's per parameter function cast: https://mxnet.incubator.apache.org/api/python/gluon/gluon.html?highlight=parameterd#mxnet.gluon.Parameter.cast

zhaoningning:
@piiswrong @ry @pluskid

reminisce:
Similar to https://github.com/apache/incubator-mxnet/issues/8799

haojin2:
@zhaoningning Hello, do you still experience this issue? If so, would you please provide some example scripts to reproduce your issue? Thanks!

apeforest:
I could reproduce this error using the following example from #8799. I will work on this issue.

```
import mxnet as mx
import numpy as np
from mxnet import autograd

def crit_func(a):
    q, l = mx.nd.linalg.gelqf(a)
    # Could as well write _, l = mx.nd.linalg.gelqf(a)
    # This circumvents the issue, but is highly obscure:
    #bogus = mx.nd.BlockGrad(q) * 0.0
    #return mx.nd.sum(l) + bogus
    return mx.nd.sum(l)

ctx = mx.cpu()

dtype = np.float32
a32 = mx.nd.random.normal(shape=(2, 3), ctx=ctx, dtype=dtype)
a32.attach_grad()
with autograd.record():
    crit32 = crit_func(a32)
    print ('crit32 = %f' % crit32.asscalar())
    crit32.backward()
    print ('grad32\n', a32.grad.asnumpy())

dtype = np.float64
a64 = mx.nd.Cast(a32, dtype=dtype)
a64.attach_grad()
with autograd.record():
    crit64 = crit_func(a64)
    print ('crit64 = %f' % crit64.asscalar())
    crit64.backward()
    print ('grad64\n', a64.grad.asnumpy())
```

apeforest:
@zhaoningning Could you provide a small script of your issue. I would like to verify my fix since your description of the issue is not exactly the same as #8799. Thanks!

lupesko:
@zhaoningning are you able to provide a script reproducing the issue?

[...]

Issue #8799, opened 2017-11-23
Dangling outputs and dtype != float32: Gradient computation fails


mseeger:
Hello,

please see the complete example below for full details.
The problem is that if in the course of a computation, certain outputs of operators are not used downstream (I call this "dangling outputs"), then gradient computation fails IFF dtype != float32.

I suppose this is because somewhere deep inside, output gradients constant 0 are created for these outputs, but this is done in float32 instead of the type of the output. I think this should be really easy to fix for somebody who knows where this happens (I don't!).

You could argue that one should not ever produce dangling outputs. But at least for some operators, this is bound to happen. And in any case, things should not work for float32 and then fail for other dtype, and the user is left with a totally obscure error message (see below).

The way around this problem is to make sure there are no dangling outputs, see example below. I'd argue that this is highly obscure, and users should not have to do this.

Here the complete example. Note that this is *not* specific to linalg.gelqf. I am just using it as an example of an operator with >1 outputs. You can use any other op with >1 output and will get the same error (you can also use an op with 1 output, and then not use it, even though this is arguably pretty dumb).

--- Code ---

import mxnet as mx
import numpy as np
from mxnet import autograd

def crit_func(a):
    q, l = mx.nd.linalg.gelqf(a)
    # Could as well write _, l = mx.nd.linalg.gelqf(a)
    # This circumvents the issue, but is highly obscure:
    #bogus = mx.nd.BlockGrad(q) * 0.0
    #return mx.nd.sum(l) + bogus
    return mx.nd.sum(l)

ctx = mx.cpu()

dtype = np.float32
a32 = mx.nd.random.normal(shape=(2, 3), ctx=ctx, dtype=dtype)
a32.attach_grad()
with autograd.record():
    crit32 = crit_func(a32)
print 'crit32 = %f' % crit32.asscalar()
crit32.backward()
print 'grad32\n', a32.grad.asnumpy()

dtype = np.float64
a64 = mx.nd.Cast(a32, dtype=dtype)
a64.attach_grad()
with autograd.record():
    crit64 = crit_func(a64)
print 'crit64 = %f' % crit64.asscalar()
crit64.backward()
print 'grad64\n', a64.grad.asnumpy()

--- Output ---

crit32 = -0.633861
grad32
[[-0.55487823  0.57657677  0.63104177]
 [-1.29575324  0.44739273  0.3476553 ]]
crit64 = -0.633862

---------------------------------------------------------------------------
[...]

Issue #9686, opened 2018-02-02
[Discussion] MXNet 2.0 Roadmap (was: APIs that might be a good idea to break in 2.0)


szha:
Let's start a discussion here about the roadmap towards MXNet 2.0. We are looking for:

- New features that are useful to your research and development.
- Improvements and patches to existing features.
- APIs that should be fixed.

If you have any item that you'd like to propose to have in the roadmap, please do:

- Create (or locate existing) issue for the item, note the issue number.
- Comment in this issue: 1) the above issue number, 2) one sentence of what the item is about and why it's useful to you.
- Indicate whether you'd be willing to help out on the item.

Given that this would be a major release, we'd have the opportunity to make backward incompatible changes. This would allow us to visit some topics that require large changes such as dropping support for python2, transitioning fully to cmake, making the tensor library numpy-compatible, or even new programming models.

-------------
Now that we decided to follow semantic versioning for releases, it would be a good idea to coordinate features and API changes to make the best use of the next major release. Thus, I propose that we use this issue to track the APIs we'd like to change in the next major version.

The candidates I've collected so far:
1. remove legacy ops such as batch-norm v1
2. reorganizing namespace for utility functions such as `download` in #9671 
3. [transform argument](https://mxnet.incubator.apache.org/versions/master/api/python/gluon/data.html?highlight=data.vision#mxnet.gluon.data.vision.datasets.MNIST) in the constructor of existing vision dataset API.

Once there are more of such requests, I will try to organize these API-breaking requests better.

bhavinthaker:
Do we have sufficient automated testing to catch accidental lapses? 

If not, can we have a volunteer to work on writing these automated test-cases? How do we track this task?

larroy:
Refactors of the cpp-package and other C++ APIs. I would like that.

fhieber:
consistent use of "axis"/"dim(s)" keyword arguments in all operators, for example:
- swapaxes uses dim1, dim2
- expand_dims uses axis

anirudhacharya:
@sandeep-krishnamurthy Please tag this - API Change, Call for Contribution, Roadmap.

szha:
#9881

eric-haibin-lin:
kvstore should not be public API

szha:
we should merge element wise ops with broadcast ops and dispatch the different implementation only based on the shape, so that symbol and ndarray +-*/ are consistent

szha:
#11031

[...]

PR #12290, opened 2018-08-22
[MXNET-798] Fix the dtype cast from non float32 in Gradient computation

apeforest:
## Description ##
This PR fixes the issues #9067 and #8799 where gradient computation for operators with multiple output fails in ndarray if the dtype is not float32. 

The root cause of the issue is that a _zeros operator was added for the other don't care output. The _zeros operator uses float32 dtype by default and it will cause conflict if the dtype in ndarray is not float32. My solution is to create a new _zeros_without_dtype operator that does not take any default dtype and use it to replace the _zeros operator in the computation graph. This change solves the dtype conflict problem and should be backward compatible.  

A unit test is added to test this fix.

### Essentials ###

### Changes ###
- [X] Change the way to infer type for auto-derived zero operator in nnvm::Graph
- [X] Added a unittest for operators with multioutput.

## Comments ##
- This seems to be a general problem for all multioutput operators when computing gradient in imperative mode. A simple example is copied from the original issue below:
- Although the change is small, the impact could be large. Thus thorough review is solicited.
```
import mxnet as mx
from mxnet import autograd


data = mx.nd.arange(16, dtype='float64').reshape((4, 4))
data.attach_grad()

with autograd.record():
    y = mx.nd.split(data, axis=0, num_outputs=2)
y[0].backward()
print(data.grad)
```


apeforest:
@eric-haibin-lin Please review

tests/python/unittest/test_infer_type.py
```diff
@@ -0,0 +1,36 @@
...
+    data.attach_grad()
+
+    with autograd.record():
+        y = mx.nd.split(data, axis=0, num_outputs=2)
+    y[0].backward()
+    assert data.grad.dtype == np.float64
+
+
+if __name__ == "__main__":
+    test_infer_multiout_op()
```
haojin2: I think this should go to something like test_operator.py instead of creating a separate file for it? And, please see https://github.com/apache/incubator-mxnet/blob/master/tests/python/unittest/test_operator.py#L7017-L7018 for how to use nosetests.

tests/python/unittest/test_infer_type.py
```diff
@@ -0,0 +1,55 @@
...
+        test32.backward()
+
+    data64 = mx.nd.Cast(data32, dtype='float64')
+    data64.attach_grad()
+    with autograd.record():
+        test64 = test_func(data64)
+        test64.backward()
+    assert_almost_equal(data64.grad.asnumpy().all(), data32.grad.asnumpy().all())
+
+
```
haojin2: I think this should go to something like test_operator.py instead of creating a separate file for it? And, please see https://github.com/apache/incubator-mxnet/blob/master/tests/python/unittest/test_operator.py#L7017-L7018 for how to use nosetests.

apeforest:
Change to [WIP] to fix some platform dependent unit test failure.

src/executor/infer_graph_attr_pass.cc
```diff
@@ -254,7 +254,8 @@ nnvm::Graph InferAttr(nnvm::Graph &&ret,
         dispatch_mode = &dispatch_modes[nid];
         if (dispatch_modes[nid] == DispatchMode::kUndefined) forward_known = false;
       }
-      auto finfer = finfer_shape.get(inode.source->op(), fdefault);
+      auto finfer = (inode.source->op() == Op::Get("_zeros")) ? fdefault :
+        finfer_shape.get(inode.source->op(), fdefault);
```
eric-haibin-lin: Are you sure about this? This affects all _zero ops, not just for the case you mentioned.

apeforest: You are right, this is breaking some unit test (however, due to unittest of master branch is broken in MacOS, I wan't able to verify before checkin). I have changed the PR to WIP.

apeforest:
@eric-haibin-lin Please review this new implementation. Thanks for your suggestion!

eric-haibin-lin:
What's up with the build?

apeforest:
@eric-haibin-lin Not sure exactly. An earlier build passed https://github.com/apache/incubator-mxnet/pull/12290/commits/dcc5f7877ec1fad7ad4c1f3d2ab3a3c5f52909ca). After I renamed some variables the build on ARM7 failed. I can submit an empty change to trigger the build again.

anirudh2290:
Also,maybe we should add zeros to APIs that may be good to break for 2.0 https://github.com/apache/incubator-mxnet/issues/9686

tests/python/unittest/test_infer_type.py
```diff
@@ -0,0 +1,55 @@
...
+    with autograd.record():
+        test32 = test_func(data32)
+        test32.backward()
+
+    data64 = mx.nd.Cast(data32, dtype=np.float64)
+    data64.attach_grad()
+    with autograd.record():
+        test64 = test_func(data64)
+        test64.backward()
+    assert_almost_equal(data64.grad.asnumpy().all(), data32.grad.asnumpy().all())
```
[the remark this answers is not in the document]
apeforest: Why increase the rtol and atol if the unit test can pass with the default one?

apeforest:
@anirudh2290 The _zeros_without_dtype operator is a private operator used only in building nnvm graph. It is not meant to be exposed to users.

tests/python/unittest/test_infer_type.py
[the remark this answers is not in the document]
anirudh2290: This can be flaky. you are comparing a float32 numpy to a float64 numpy and the atol and rtol  defaults are small.

anirudh2290:
@apeforest what i meant is we can change the dtype default to -1 for zeros operator for 2.0.

apeforest:
@anirudh2290  Thanks for the clarification. I have increased atol and rtol values in unit test. As to changing the dtype default to -1 for zeros, I think it is not related to this PR and may cause a backward compatibility issue with old models. Therefore, I would prefer doing that in a separate PR. Please let me know what you think. Thanks.

anirudh2290:
Not suggesting to do it in this PR. Just wanted to document it in the APIs to break for 2.0 and we can do it before 2.0 release.

PR #12276, opened 2018-08-21
[Don't merge][Review] A solution to prevent zombie containers locally and in CI

larroy:
## Description ##
This PR adds mechanisms on the build scripts to cleanup docker containers when there is a cancellation, either from the user by sending SIGINT / SIGTERM or when the process inside the container ends with an error.

It also propagates the environment variables from Jenkins which are used by the process tree killer to identify runaway processes so processes inside the container are killed when the Jenkins job is stopped.

With this patch we catch SIGINT and SIGTERM and install a handler on atexit which cleans (stops) docker containers which were created during execution of the build.

We also switch to python docker API for better management of running containers.

@aaronmarkham @marcoabreu @lebeg @KellenSunderland

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

PR #12296, opened 2018-08-22
Separate refactoring from #12276 in a prior PR

larroy:
## Description ##
This is a prior refactor separated from #12276 at request from the reviewer.

Fixes --no-dockerhub-cache options and docker's --cache-from so local cache can be used on request by providing --no-dockerhub-cache. 

Checked with mypy and pycharm code analyze.

@KellenSunderland 

### Essentials ###

## Comments ##

### DIFF
--- a/ci/build.py
+++ b/ci/build.py
@@ -23,31 +23,88 @@
 """
 
 __author__ = 'Marco de Abreu, Kellen Sunderland, Anton Chernov, Pedro Larroy'
-__version__ = '0.1'
+__version__ = '0.2'
 
 import argparse
 import glob
 import logging
-import os
 import re
 import shutil
 import subprocess
 import sys
 import tempfile
-import platform
 from copy import deepcopy
 from itertools import chain
-from subprocess import call, check_call
+from subprocess import call, check_call, check_output
 from typing import *
 from util import *
+import pprint
+import requests
+
+
+def retry(target_exception, tries=4, delay_s=1, backoff=2):
+    """Retry calling the decorated function using an exponential backoff.
+
+    http://www.saltycrane.com/blog/2009/11/trying-out-retry-decorator-python/
+    original from: http://wiki.python.org/moin/PythonDecoratorLibrary#Retry
+
+    :param target_exception: the exception to check. may be a tuple of
+        exceptions to check
+    :type target_exception: Exception or tuple
+    :param tries: number of times to try (not retry) before giving up
+    :type tries: int
+    :param delay_s: initial delay between retries in seconds
+    :type delay_s: int
+    :param backoff: backoff multiplier e.g. value of 2 will double the delay
+        each retry
+    :type backoff: int
+    """
+    import time
+    from functools import wraps
+
+    def decorated_retry(f):
+        @wraps(f)
+        def f_retry(*args, **kwargs):
+            mtries, mdelay = tries, delay_s
+            while mtries > 1:
+                try:
+                    return f(*args, **kwargs)
+                except target_exception as e:
+                    logging.warning("Exception: %s, Retrying in %d seconds...", str(e), mdelay)
+                    time.sleep(mdelay)
+                    mtries -= 1
+                    mdelay *= backoff
+            return f(*args, **kwargs)
+
+        return f_retry  # true decorator
+
+    return decorated_retry
+
 
 CCACHE_MAXSIZE = '500G'
 
+
+# noinspection SyntaxError
 def under_ci() -> bool:
     """:return: True if we run in Jenkins."""
     return 'JOB_NAME' in os.environ
 
-def get_platforms(path: Optional[str] = "docker"):
+
+def git_cleanup() -> None:
+    """Clean repo and subrepos, update subrepos"""
+    logging.info("cleaning up repository")
+    with remember_cwd():
+        os.chdir(get_mxnet_root())
+        check_call(['git', 'clean', '-ffdx'])
+        check_call(['git', 'submodule', 'foreach', '--recursive', 'git', 'clean', '-ffdx'])
+        check_call(['git', 'submodule', 'update', '--recursive', '--init'])
+
+
+def get_dockerfiles_path():
+    return "docker"
+
+
+def get_platforms(path: str = get_dockerfiles_path()) -> List[str]:
     """Get a list of architectures given our dockerfiles"""
     dockerfiles = glob.glob(os.path.join(path, "Dockerfile.build.*"))
     dockerfiles = list(filter(lambda x: x[-1] != '~', dockerfiles))
@@ -57,10 +114,11 @@ def get_platforms(path: Optional[str] = "docker"):
 
 
 def get_docker_tag(platform: str, registry: str) -> str:
+    """:return: docker tag to be used for the container"""
     return "{0}/build.{1}".format(registry, platform)
 
 
-def get_dockerfile(platform: str, path="docker") -> str:
+def get_dockerfile(platform: str, path=get_dockerfiles_path()) -> str:
     return os.path.join(path, "Dockerfile.build.{0}".format(platform))
 
 
@@ -68,18 +126,18 @@ def get_docker_binary(use_nvidia_docker: bool) -> str:
     return "nvidia-docker" if use_nvidia_docker else "docker"
 
 
-def build_docker(platform: str, docker_binary: str, registry: str, num_retries: int) -> None:
+def build_docker(platform: str, docker_binary: str, registry: str, num_retries: int, use_cache: bool) -> str:
     """
     Build a container for the given platform
     :param platform: Platform
     :param docker_binary: docker binary to use (docker/nvidia-docker)
     :param registry: Dockerhub registry name
     :param num_retries: Number of retries to build the docker image
+    :param use_cache: will pass cache_from to docker to use the previously pulled tag
     :return: Id of the top level image
     """
-
     tag = get_docker_tag(platform=platform, registry=registry)
-    logging.info("Building container tagged '%s' with %s", tag, docker_binary)
+    logging.info("Building docker container tagged '%s' with %s", tag, docker_binary)
     #
     # We add a user with the same group as the executing non-root user so files created in the
     # container match permissions of the local user. Same for the group.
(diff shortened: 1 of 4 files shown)


ci/build.py
```diff
@@ -23,31 +23,88 @@
...
+    :type target_exception: Exception or tuple
+    :param tries: number of times to try (not retry) before giving up
+    :type tries: int
+    :param delay_s: initial delay between retries in seconds
+    :type delay_s: int
+    :param backoff: backoff multiplier e.g. value of 2 will double the delay
+        each retry
+    :type backoff: int
+    """
+    import time
```
lebeg: I think it's better to do the imports at the top of the file in this case

ci/build.py
```diff
@@ -346,15 +429,21 @@ def use_cache():
...
             if use_cache():
                 load_docker_cache(tag=tag, docker_registry=args.docker_registry)
-            build_docker(platform, docker_binary, args.docker_registry, num_retries=num_docker_build_retires)
+            build_docker(platform, docker_binary, args.docker_registry, num_retries=args.docker_build_retries,
+                         use_cache=use_cache())
             if args.build_only:
                 continue
-            build_platform = "build_{}".format(platform)
-            cmd = ["/work/mxnet/ci/docker/runtime_functions.sh", build_platform]
+            git_cleanup()
```
lebeg: This is good, but maybe we should issue a warning here with confirmation.

ci/build.py
```diff
@@ -346,15 +429,21 @@ def use_cache():
...
             shutil.rmtree(buildir(), ignore_errors=True)
+            build_platform = "build_{}".format(platform)
+            plat_buildir = os.path.abspath(os.path.join(get_mxnet_root(), '..',
+                                                        "mxnet_{}".format(build_platform)))
+            if os.path.exists(plat_buildir):
```
lebeg: How would the path exist if git_cleanup was called before?

ci/build.py
lebeg: Actually, how this would work when all platforms are build one after another? Only the latest artifacts survive?

ci/build.py
```diff
@@ -383,7 +472,8 @@ def use_cache():
 
 ./build.py -a
 
-    Builds for all platforms and leaves artifacts in build_<platform>
+    Builds for all platforms and leaves artifacts in build_<platform>. **WARNING** it performs git
```
lebeg: Maybe a -y for quiet confirmation

ci/build.py
KellenSunderland: How would this work if for example I want to test a new file locally that hasn't yet been added to git?

ci/build.py
```diff
@@ -297,22 +377,24 @@ def script_name() -> str:
...
+        return not args.no_cache or under_ci()
 
     command = list(chain(*args.command))
     docker_binary = get_docker_binary(args.nvidiadocker)
     shared_memory_size = args.shared_memory_size
     num_docker_build_retires = args.docker_build_retries
 
     if args.list:
-        list_platforms()
+        print(list_platforms())
```
lebeg: You used logger before, maybe use it here as well instead of print?

ci/build.py
```diff
@@ -284,7 +364,7 @@ def script_name() -> str:
                         default=1,
                         type=int)
 
-    parser.add_argument("-c", "--cache", action="store_true",
+    parser.add_argument("--no-cache", action="store_true",
```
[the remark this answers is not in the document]
lebeg: Yes, I agree that this is changing behavior of the script and does not fall under "minor refactorings". Even if it's a oneliner - it needs a separate PR.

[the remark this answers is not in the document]
larroy: Is this going to ever end? Some prs are huge and this is a ci script we are talking about.

ci/build.py
larroy: Is different concerns, this prints unconditionally to stdout which is what you want. Logging can be supressed. This is intentional

ci/build.py
larroy: Why is it better? it's perfectly legal to do this, and some consider best practice to scope imports to the minimum scope necessary, like for example function scope.

ci/build.py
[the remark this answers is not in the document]
larroy: It's too much trouble to split, as agreed in 1:1 I improved documentation about this flag.

ci/build.py
larroy: This is addressed, I removed git cleanup.

ci/build.py
larroy: removed that git cleanup.

ci/build.py
lebeg: Ok, I see. Thanks!

ci/build.py
```diff
@@ -284,8 +288,9 @@ def script_name() -> str:
                         default=1,
                         type=int)
 
-    parser.add_argument("-c", "--cache", action="store_true",
-                        help="Enable docker registry cache")
+    parser.add_argument("--no-cache", action="store_true",
```
lebeg: I find the name a bit confusing, maybe `--dockerhub--cache`?

larroy: ⚡️

ci/build.py
```diff
@@ -193,52 +188,61 @@ def container_run(platform: str,
...
-    base = os.path.split(os.path.realpath(__file__))[0]
-    os.chdir(base)
 
-    logging.getLogger().setLevel(logging.INFO)
+def log_environment():
+    instance_id = ec2_instance_id_hostname()
+    if instance_id:
+        logging.info("EC2 Instance id: %s", instance_id)
+    pp = pprint.PrettyPrinter(indent=4)
+    logging.debug("Build environment: %s", pp.pformat(dict(os.environ)))
```
lebeg: Should this be info? The global level is at that level. Or should we have separate levels for ci and local?

ci/build.py
```diff
@@ -193,52 +188,61 @@ def container_run(platform: str,
...
-def main() -> int:
-    # We need to be in the same directory than the script so the commands in the dockerfiles work as
-    # expected. But the script can be invoked from a different path
-    base = os.path.split(os.path.realpath(__file__))[0]
-    os.chdir(base)
 
-    logging.getLogger().setLevel(logging.INFO)
+def log_environment():
+    instance_id = ec2_instance_id_hostname()
+    if instance_id:
```
lebeg: Is this always true? The function returns '?' if something goes wrong

ci/util.py
```diff
@@ -41,3 +42,73 @@ def remember_cwd():
...
+
+    def decorated_retry(f):
+        @wraps(f)
+        def f_retry(*args, **kwargs):
+            mtries, mdelay = tries, delay_s
+            while mtries > 1:
+                try:
+                    return f(*args, **kwargs)
+                except target_exception as e:
+                    logging.warning("Exception: %s, Retrying in %d seconds...", str(e), mdelay)
```
lebeg: Should we propagate the exception if we out of retries?

ci/build.py
```diff
@@ -284,8 +288,10 @@ def script_name() -> str:
                         default=1,
                         type=int)
 
-    parser.add_argument("-c", "--cache", action="store_true",
-                        help="Enable docker registry cache")
+    parser.add_argument("-c", "--no-dockerhub-cache", action="store_true",
```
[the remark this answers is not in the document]
larroy: I would rather have it explicit. but we would have to code it if we want it the way you say.

ci/build.py
larroy: I think it's too verbose. In general I don't think we need to view this environment. I had it to diagnose. I wouldn't bother right now.

ci/build.py
larroy: Empty string is false. In that case is not printed. For example outside of ci or locally.

ci/util.py
larroy: Last call to f is propagating the exception as you say.

lebeg: Yes, you're right

ci/build.py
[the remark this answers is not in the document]
lebeg: The problem is that for "default" behaviour, which I assume is use local cache (as for normal docker builds) we need to provide --no-dockerhub-cache explicitly. I think it should be the other way around.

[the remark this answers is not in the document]
lebeg: `-c` is not a good shortening anymore for the option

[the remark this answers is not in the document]
larroy: The default behavior remains unchanged if I'm not mistaken. Which is to use the dockerhub cache.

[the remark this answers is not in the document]
KellenSunderland: I'm also confused what the difference between this and just not specifying '--docker-registry' is.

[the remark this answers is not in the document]
lebeg: You can see in the diff that at least what the option did changed:
```
parser.add_argument("-c", "--cache"
```
```
parser.add_argument("-c", "--no-dockerhub-cache"
```
Previously it was enable and now it is disable. Am I missing something?

ci/build.py
```diff
@@ -90,41 +93,29 @@ def build_docker(platform: str, docker_binary: str, registry: str, num_retries:
...
-               "--build-arg", "GROUP_ID={}".format(os.getgid()),
-               "--cache-from", tag,
-               "-t", tag,
-               "docker"]
+    #
+    cmd = [docker_binary, "build",
+           "-f", get_dockerfile(platform),
+           "--build-arg", "USER_ID={}".format(os.getuid()),
+           "--build-arg", "GROUP_ID={}".format(os.getgid())]
+    if use_cache:
```
KellenSunderland: Nit: I find this a bit of a misnomer.  It sounds like we're disabling docker-cache, which is possible, but it's not what's done here.  Reading the code I would expect it to pass --no-cache to the docker cmd being run.  Maybe a more specific name could help?

ci/build.py
```diff
@@ -90,41 +93,29 @@ def build_docker(platform: str, docker_binary: str, registry: str, num_retries:
...
+        cmd.extend(["--cache-from", tag])
+    cmd.extend(["-t", tag, get_dockerfiles_path()])
+
+    @retry(subprocess.CalledProcessError, tries=num_retries)
```
KellenSunderland: Really like this.

ci/build.py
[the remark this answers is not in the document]
larroy: Previously It was not functioning properly. 
Also check line 308. I think you are missing something as the effect is now negated. The option disables the cache if present. Default is cache enabled. I tested all this.


What single letter abbreviation do you suggest instead?

ci/build.py
```diff
@@ -193,52 +188,61 @@ def container_run(platform: str,
...
         ret = call(runlist)
 
-    docker_run_cmd = ' '.join(runlist)
     if not dry_run and interactive:
         into_cmd = deepcopy(runlist)
         # -ti can't be after the tag, as is interpreted as a command so hook it up after the -u argument
         idx = into_cmd.index('-u') + 2
         into_cmd[idx:idx] = ['-ti']
-        cmd = '\\\n\t'.join(into_cmd)
+        cmd = ' \\\n\t'.join(into_cmd)
```
KellenSunderland: Thanks for fixing this.

ci/build.py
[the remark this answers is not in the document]
larroy: Kellen. One doesn't need to specify the registry. Hence this option to disable cache which is enabled by default.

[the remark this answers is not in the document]
KellenSunderland: This seems like a work around for broken behaviour to me.  Would you mind if we leave it out of this PR, and try to think about how to fix the root cause of the problem?  To me it makes sense to _not_ use remote caching if a user doesn't specify '--docker-registry'.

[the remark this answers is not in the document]
larroy: That’s bad because we want to use cache by default. This was discussed at length with Marco before and we came to agree on this behavior. I wouldn’t want to change that.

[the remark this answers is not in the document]
larroy: That’s bad because we want to use cache by default. This was discussed at length with Marco before and we came to agree on this behavior. I wouldn’t want to change that.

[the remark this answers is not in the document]
KellenSunderland: Ok, that's the information I was missing here.  In this case it makes sense to me.

sandeep-krishnamurthy:
@lebeg - This PR is ready to go, if your concerns are addressed.

ci/build.py
larroy: I'm open. Any suggestions?

sandeep-krishnamurthy:
@marcoabreu - can you take a look at this?

marcoabreu:
LGTM, thanks for iterating on this one!

PR #10568, opened 2018-04-16
[WIP] [MXNET-325] Model parallelism tutorial.

indhub:
## Description ##

I'm working on a tutorial to show how model parallelism can be done in Gluon. The tutorial builds a stacked LSTM with different LSTM layers on different GPUs (like described in [this](https://mxnet.incubator.apache.org/faq/model_parallel_lstm.html) tutorial). The code borrows from [this](http://gluon.mxnet.io/chapter05_recurrent-neural-networks/rnns-gluon.html) gluon tutorial and makes it model parallel.

This is a Work In Progress. Before I write a tutorial around this code, I wanted to get the approach reviewed to make sure this is the correct way to do model parallelism in Gluon.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)

PR #12298, opened 2018-08-22
Remove model parallelism references from FAQ

ThomasDelteil:
## Description ##
Remove from the FAQ outdated references to model parallelism.
Model parallelism should be avoided as it is less performant than data parallelism, better accumulate gradients through multiple mini batches of small batch size rather than splitting the model to move bigger batches across GPUs.


aaronmarkham:
I just fixed the broken links in that FAQ. :expressionless:
I think content should removed if it is "bad" information, however wouldn't it be better to provide a disclaimer at the top saying **This is considered less performant that x y and z and we recommend you look at [this other amazing tutorial](#) before you try the following model parallel approach.** Even talk a lot more about the reasoning and provide references... but then leave the content. What if someone has some super massive model and they really, really want to do it this way anyway?

ThomasDelteil:
@aaronmarkham I agree, after a previous PR[1] where @indhub had written a Model parallelism tutorial, we had an offline discussion and decided that the penalty incurred by doing model parallelism when you should actually be doing data parallelism outweighed the benefit for the few users that would actually need model parallelism.

[1] https://github.com/apache/incubator-mxnet/pull/10568/files

sandeep-krishnamurthy:
@ThomasDelteil - I think the content is not bad like @aaronmarkham  said. It is only communicating the pros/cons and setting the right context of when is this applicable or not.
Is there any strong  need to remove this?

lupesko:
Agree with @sandeep-krishnamurthy comment.
@ThomasDelteil maybe you want to refine the doc a bit?

ThomasDelteil:
@sandeep-krishnamurthy @aaronmarkham @lupesko closing for now, will follow up with other stakeholders and see if we actually need to remove. For the records, I don't think either strongly that we should remove this FAQ

PR #12323, opened 2018-08-23
Sphinx error reduction

aaronmarkham:
## Description ##
The is a resubmission of #11916.
I had to remove updating the config files that were already updated in the website build pipeline and to keep this one focused on the `toctree` and other content issues.

## Preview
http://34.201.8.176/versions/sphinx_error_reduction/


aaronmarkham:
@szha Can you or someone review this?
Thanks.

docs/architecture/release_note_0_9.md
```diff
@@ -1,49 +0,0 @@
-# MXNet 0.9 (NNVM) Release Note
```
lebeg: We don't need those?

aaronmarkham: All of the release notes are on github in the releases section. We don't maintain any others here. The presence of the page throws an error because it isn't in a ToC. I'd just delete it since the release notes are elsewhere.

PR #12350, opened 2018-08-25
fix bug in 'device' type kvstore

solin319:
When we init a key after another key pushed. This key has no merged_buf_ in file 'comm.h', but the inited_ is true. So it can't pull this new key. The code below will make error.
```
import mxnet as mx
a=mx.nd.array([1,2,3], ctx=mx.gpu(0))
b=mx.nd.array([0,0,0], ctx=mx.gpu(0))
kv=mx.kv.create('device')
kv.init('1', a)
kv.push('1', [a,a,a,a])
kv.pull('1', b)
kv.init('2', a)
kv.pull('2', b)
```

### DIFF
--- a/tests/python/unittest/test_kvstore.py
+++ b/tests/python/unittest/test_kvstore.py
@@ -107,6 +107,23 @@ def test_init():
     check_init(mx.kv.create(), 'a')
 
 @with_seed()
+def test_pull():
+    """test pull"""
+    def check_pull(kv):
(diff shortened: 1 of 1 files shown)


marcoabreu:
Thanks a lot! Would you mind adding a test?

solin319:
We add a test in 'test_kvstore.py'.

solin319:
@marcoabreu  @anirudh2290 
Anything else to do?

PR #12362, opened 2018-08-26
Removing the re-size for validation data, which breaking the validation accuracy of CIFAR training

juliusshufan:
## Description ##
The dataloader for example/image-classification resizes the imagesize of validation dataset to 256 by default, and which breaks the validation accuracy of CIFAR10, this PR trying to fix it.
Below firgure is the validation accuracy trends without this fix (model is ResNet-50).
image

### Changes ###
example/image-classification/common/data.py

## Comments ##

### DIFF
--- a/example/image-classification/common/data.py
+++ b/example/image-classification/common/data.py
@@ -194,7 +194,6 @@ def get_rec_iter(args, kv=None):
         std_r               = rgb_std[0],
         std_g               = rgb_std[1],
         std_b               = rgb_std[2],
-        resize              = 256,
         data_name           = 'data',
         label_name          = 'softmax_label',
         batch_size          = args.batch_size,


juliusshufan:
@szha may I have your comments, I use this example for the regression tracking of mxnet with mkldnn backend, I'll be helpful if this issue can be addressed.
Thanks.

zhreshold:
Does it affect the imagenet validation perf?

juliusshufan:
@zhreshold Thanks for the feedback, I use the pre-trained ResNet-50 model and validation-set of ImageNet-1k, and the accuracy with/w.o my changes are same: see test result:

**Without** my changes:
INFO:root:Finished with 126.286497 images per second
INFO:root:('accuracy', 0.753156969309463)
INFO:root:('top_k_accuracy_5', 0.9257512787723785)

**With** my changes:
INFO:root:Finished with 126.029153 images per second
INFO:root:('accuracy', 0.753156969309463)
INFO:root:('top_k_accuracy_5', 0.9257512787723785)

The imagenet validation perf and accuracy are **same**.

But without my changes, the validation accuracy trends of CIFAR10+ResNet50/VGG16 are as below, obviously it is not expected.

image
image

TaoLv:
Seems this regression also goes into 1.3.0rc0 realese.

juliusshufan:
@zhreshold sorry for pushing, may I have you comments or approval?

Thanks.

zhreshold:
@juliusshufan Thanks for the result, it's fair enough to go through. Again, thanks for the contribution!

PR #12365, opened 2018-08-27
Adagrad optimizer with row-wise learning rate

leezu:
## Description ##
This implements a special version of AdaGrad with a learning rate per row instead of for every parameter. This is useful for learning large embedding matrices, as it significantly reduces the memory requirements for the state array as well as reduces over-fitting due to having less learning rates. Consequently only 2D weight arrays are supported.

~~Compared to the AdaGrad optimizer already included in MXNet, this Optimizer also supports the group lasso proximal operator for inducing group sparsity, where each row of the weight is considered a group. As there is a single learning rate per group, we can use a [closed form solution for the proximal operator](https://en.wikipedia.org/wiki/Proximal_gradient_methods_for_learning#Group_lasso) which is not possible if different learning rates are used for parameters within a single group.~~
To speed up merging this PR, the proximal implementation will be merged later, after the API for eager update is integrated with gluon trainer.

This Optimizer is useful for the GluonNLP project.

### Essentials ###

### Changes ###
- [X] Add contrib.optimizer.GroupAdaGrad

## Comments ##
- If you have any suggestions for the API naming or other suggestions pleas let me know.
- ~~Forcing an eager update of the lazily accumulated group lasso regularization for a set of rows requires access to some private API. I am not sure how this could be solved without access to private API:~~ This will be addressed in a separate PR

        fake_grad = mx.nd.sparse.row_sparse_array(
            (mx.nd.zeros((len(indices_to_force_update), weight.shape[1])), indices_to_force_update))
        weight.grad()[:] = fake_grad
        weight.data()._fresh_grad = True
        trainer._optimizer._index_update_count[0] -= 1
        trainer._optimizer.num_update -= 1
        trainer.step(batch_size=1)

@szha @szhengac @eric-haibin-lin

### DIFF
--- a/src/operator/contrib/optimizer_op.cc
+++ b/src/operator/contrib/optimizer_op.cc
@@ -64,9 +64,7 @@ Updates are applied by::
     div = grad / sqrt(history + float_stable_eps)
     weight -= div * lr
 
-Weights are updated lazily if the gradient is sparse. In particular, before
-using a set of weights for a forward pass, you may want to ensure that the
-lazily accumulated group lasso regularization is applied.
+Weights are updated lazily if the gradient is sparse.
 
 Note that non-zero values for the weight decay option are not supported.
 
--- a/src/operator/contrib/optimizer_op.cc
+++ b/src/operator/contrib/optimizer_op.cc
@@ -23,21 +23,21 @@
  * \brief Optimizer operators
  * \author Leonard Lausen
  */
-#include "./optimizer_op-inl.h"
 #include "../elemwise_op_common.h"
+#include "./optimizer_op-inl.h"
 
 namespace mxnet {
 namespace op {
 
-DMLC_REGISTER_PARAMETER(ProximalGroupAdagradParam);
+DMLC_REGISTER_PARAMETER(GroupAdagradParam);
 
 /*!
- * \brief Shape inference function for Proximal Group AdaGrad.
+ * \brief Shape inference function for Group AdaGrad.
  */
-inline bool ProximalGroupAdagradShape(const nnvm::NodeAttrs &attrs,
-                                      std::vector<TShape> *in_attrs,
-                                      std::vector<TShape> *out_attrs) {
-  CHECK_EQ(in_attrs->size(), 4U);
+inline bool GroupAdagradShape(const nnvm::NodeAttrs &attrs,
+                              std::vector<TShape> *in_attrs,
+                              std::vector<TShape> *out_attrs) {
+  CHECK_EQ(in_attrs->size(), 3U);
   CHECK_EQ(out_attrs->size(), 1U);
 
   SHAPE_ASSIGN_CHECK(*out_attrs, 0, in_attrs->at(0));
@@ -50,8 +50,8 @@ inline bool ProximalGroupAdagradShape(const nnvm::NodeAttrs &attrs,
          (in_attrs->at(0)[0] == in_attrs->at(2)[0]);
 }
 
-NNVM_REGISTER_OP(_contrib_proximal_group_adagrad_update)
-.describe(R"code(Update function for Proximal Group AdaGrad optimizer.
+NNVM_REGISTER_OP(_contrib_group_adagrad_update)
+.describe(R"code(Update function for Group AdaGrad optimizer.
 
 Referenced from *Adaptive Subgradient Methods for Online Learning and Stochastic Optimization*,
 and available at http://www.jmlr.org/papers/volume12/duchi11a/duchi11a.pdf but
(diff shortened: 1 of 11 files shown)


python/mxnet/contrib/optimizer.py
```diff
@@ -0,0 +1,140 @@
...
+        weight += (div + weight * wd) * -lr
+
+    If `l2_regularization_strength > 0` a proximal operator is used to optimize
+    with group lasso objective. Weights are updated lazily if the gradient is
+    sparse. In particular, before using a set of weights for a forward pass,
+    you may want to ensure that the lazily accumulated group lasso
+    regularization is applied. This can be achieved by creating a sparse
+    gradient array that contains explicit 0 data for the indices to be updated:
+
+        fake_grad = mx.nd.sparse.row_sparse_array(
```
eric-haibin-lin: Could you check if the doc for this part renders correctly?

python/mxnet/contrib/optimizer.py
```diff
@@ -0,0 +1,140 @@
...
+            (mx.nd.zeros((len(indices), dim)), indices))
+        weight.grad()[:] = fake_grad
+        weight.data()._fresh_grad = True
+        trainer._optimizer._index_update_count[0] -= 1
+        trainer._optimizer.num_update -= 1
+        trainer.step(batch_size=1)
```
eric-haibin-lin: This seems not very easy to use..

python/mxnet/contrib/optimizer.py
```diff
@@ -0,0 +1,140 @@
...
+
+    For details of the update algorithm see
+    :class:`~mxnet.ndarray.proximal_group_adagrad_update`.
+
+    This optimizer accepts the following parameters in addition to those
+    accepted by :class:`.Optimizer`. Weight decay is not supported.
```
eric-haibin-lin: Maybe remove wd terms from line 43?

python/mxnet/contrib/optimizer.py
```diff
@@ -0,0 +1,140 @@
...
+            last_update_buffer = zeros(1, ctx=weight.context)
+        return (history, last_update_buffer)
+
+    def update(self, index, weight, grad, state):
+        assert (isinstance(weight, NDArray))
+        assert (isinstance(grad, NDArray))
+        self._update_count(index)
+        lr = self._get_lr(index)
+        wd = self._get_wd(index)
+        assert wd == 0
```
eric-haibin-lin: Add err message

python/mxnet/contrib/optimizer.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def create_state(self, index, weight):
+        assert len(weight.shape) == 2
+        history = zeros(
+            (weight.shape[0], 1), weight.context, stype=weight.stype)
+        last_update_buffer = None
+        if self.l2_regularization_strength > 0:
+            last_update_buffer = full(
+                shape=(weight.shape[0], ),
+                val=self.num_update,
```
eric-haibin-lin: is `self.num_update` always 0 when create_state is called?

tests/python/unittest/test_contrib_optimizer.py
```diff
@@ -0,0 +1,125 @@
...
+    eps_options = [{}, {'eps': 1e-8}]
+    cg_options = [{}, {'clip_gradient': 0.4}, {'clip_gradient': 0.5}]
+    rg_options = [{}, {'rescale_grad': 0.14}, {'rescale_grad': 0.8}]
+    l2_options = [{
+        'l2_regularization_strength': 0.0
+    }, {
+        'l2_regularization_strength': 0.05
+    }]
+    for dtype in [np.float32]:
+        for eps_option in eps_options:
```
eric-haibin-lin: This is getting long. Could `itertools.product ` help reduce verbosity?

src/operator/contrib/optimizer_op.cc
```diff
@@ -0,0 +1,95 @@
...
+  SHAPE_ASSIGN_CHECK(*out_attrs, 0, in_attrs->at(1));
+  SHAPE_ASSIGN_CHECK(*in_attrs, 0, out_attrs->at(0));
+  SHAPE_ASSIGN_CHECK(*in_attrs, 1, out_attrs->at(0));
+
+  return out_attrs->at(0).ndim() != 0U && out_attrs->at(0).Size() != 0U &&
+         (in_attrs->at(0)[0] == in_attrs->at(1)[0]) &&
+         (in_attrs->at(0)[0] == in_attrs->at(2)[0]);
+}
+
+NNVM_REGISTER_OP(proximal_group_adagrad_update)
```
eric-haibin-lin: register as _contrib_proximal_group_adagrad_update and invoke it using `nd.contrib.proximal_group_adagrad_update`.
For contrib ops we can skip the sparse alias.

src/operator/contrib/optimizer_op.cc
```diff
@@ -0,0 +1,95 @@
...
+.set_attr<FInferStorageType>("FInferStorageType", ProximalGroupAdagradStorageType)
+.set_attr<nnvm::FMutateInputs>("FMutateInputs",
+  [](const nnvm::NodeAttrs& attrs) {
+    return std::vector<uint32_t>{2, 3};
+  })
+.set_attr<FComputeEx>("FComputeEx<cpu>", ProximalGroupAdagradUpdateEx<cpu>)
+.add_argument("weight", "NDArray-or-Symbol", "Weight")
+.add_argument("grad", "NDArray-or-Symbol", "Gradient")
+.add_argument("history", "NDArray-or-Symbol", "History")
+.add_argument("last_update_buffer", "NDArray-or-Symbol", "Last update buffer")
```
eric-haibin-lin: Rename it as `last_update` instead of `last_update_buffer`

python/mxnet/contrib/optimizer.py
eric-haibin-lin: I'm afraid it's still not very clear how to use this in an end2end training script for new users

leezu: Yes. 

Unfortunately we can't trigger a weight update during forward computation automatically in MXNet as far as I am aware, so the user needs to trigger it manually if he wants to have it.

Alternatively to manually triggering it as introduced above, we can extend the `Trainer` and `Optimizer` APIs to support some kind of `apply_lazy_regularization` method that hides above logic and only requires the user to specify the set of rows to update.

leezu: Currently the PR does not change any such APIs. If we agree on any API changes I would be very happy to remove this workaround.

python/mxnet/contrib/optimizer.py
leezu: Even if it is non-zero, we only care about the difference between `self.num_update` and the `last_update` array. So the actual value of `self.num_update` does not matter.

tests/python/unittest/test_contrib_optimizer.py
leezu: This follows the coding style used in other test files. It seemed nested for loops are preferred for the tests. I'm happy to change it.

leezu:
Test was failing due to `test_mkldnn.test_activation`.

python/mxnet/contrib/optimizer.py
```diff
@@ -0,0 +1,140 @@
...
+
+    For details of the update algorithm see
+    :class:`~mxnet.ndarray.contrib.proximal_group_adagrad_update`.
+
+    This optimizer accepts the following parameters in addition to those
+    accepted by :class:`.Optimizer`. Weight decay is not supported.
+
+    Parameters
+    ----------
+    l2_regularization_strength : float
```
szha: maybe rename to lambda?

leezu:
Perl tests failed due to

```
Downloading model//squeezenet1.1-33ba0f93.zip from http://apache-mxnet.s3-accelerate.dualstack.amazonaws.com/gluon/models/squeezenet1.1-33ba0f93.zip ...

download of url failed! (599 Internal Exception)
```

Restarting

leezu:
Rebased and set up Sphinx to build documentation for `optimizer.contrib` module

szha:
http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-12365/10/api/python/optimization/contrib.html there's some problem in doc rendering

stu1130:
@leezu Could you fix the doc rendering problem? Let us know if you need help!

vandanavk:
@mxnet-label-bot [pr-awaiting-response]

leezu:
- Doc rendering should be fixed now
- The proximal implementation was removed from the PR to facilitate merging the row-wise learning rate feature before a decision is taken on how to best design the eager update API that needs to be integrated with gluon Trainer to improve usability of the lazy proximal update

This should be ready to review / merge now

leezu:
CI was failing for a few days / tries due to seemingly unrelated flaky tests. It finally passed now.

piyushghai:

leezu:
@eric-haibin-lin @anirudh2290 do you have any comments? Can this be merged?

PR #12374, opened 2018-08-27
Fix/public internal header

azai91:
## Description ##
the public header `resource.h` current has internal dependencies. public headers should only define interfaces and should not include internal dependencies as it allows users to access non-exposed interfaces. move random_generator.h into public header domain. move private implementations into respective .cc files.

### Essentials ###

### Changes ###
- [ ]move random_generator.h into public header domain. move private implementations into respective .cc files.

## Comments ##


src/common/random_generator.cu
```diff
@@ -59,6 +59,16 @@ void RandGenerator<gpu, float>::Seed(mshadow::Stream<gpu> *s, uint32_t seed) {
   s->Wait();
 }
 
+
+static void RandGenerator<gpu, float>::AllocState(RandGenerator<gpu, DType> *inst) {
```
lebeg: Isn't the only defined specialization double based?

src/common/random_generator.h
```diff
@@ -1,218 +1,218 @@
...
-  curandStatePhilox4_32_10_t *states_;
-};  // class RandGenerator<gpu, double>
-
-#endif  // MXNET_USE_CUDA
-
-}  // namespace random
-}  // namespace common
-}  // namespace mxnet
-#endif  // MXNET_COMMON_RANDOM_GENERATOR_H_
+///*
```
lebeg: Is this on purpose?

azai91:
@mseth10

include/mxnet/random_generator.h
```diff
@@ -80,7 +79,7 @@ class RandGenerator<cpu, DType> {
     std::mt19937 *engine_;
   };
 
-  static void AllocState(RandGenerator<cpu, DType> *inst) {
+  static void  AllocState(RandGenerator<cpu, DType> *inst) {
```
lebeg: some redundant whitespace

include/mxnet/random_generator.h
```diff
@@ -150,14 +149,9 @@ class RandGenerator<gpu, DType> {
     curandStatePhilox4_32_10_t state_;
   };  // class RandGenerator<gpu, DType>::Impl
 
-  static void AllocState(RandGenerator<gpu, DType> *inst) {
-    CUDA_CALL(cudaMalloc(&inst->states_,
-                         kNumRandomStates * sizeof(curandStatePhilox4_32_10_t)));
-  }
+  static void AllocState(RandGenerator<gpu, DType> *inst);
```
apeforest: Why do we do this differently with CPU implementation?

azai91: what does you mean? we are calling cudaMalloc in this case instead of just "new"

apeforest: See line: https://github.com/apache/incubator-mxnet/pull/12374/files#diff-ba5bcd7d0b76b85a2df1f793dc4d3302R82

Also this is inside the inner class Impl which is supposed to handle all the implementation. I do not see a reason to move these two lines to the cu file. Not to mention the performance advantage of inline function.

azai91: I see what you mean. these cuda calls depends on cuda_utils.h which is in the commons folder. I am not sure if we want to expose those. thoughts?

anirudh2290: @apeforest are you indicating about perf advantages during compile time ? I think its okay to put the definition in random_generator.cu since we don't want to expose cuda_utils.

src/common/random_generator.cu
```diff
@@ -59,6 +59,17 @@ void RandGenerator<gpu, float>::Seed(mshadow::Stream<gpu> *s, uint32_t seed) {
   s->Wait();
 }
 
+template<>
+void RandGenerator<gpu, float>::AllocState(RandGenerator<gpu> *inst) {
```
anirudh2290: doesnt this remove support for other DTypes like half_t ?

samskalicky: No, the random generator always generates floats between 0 and 1, we scale them as needed in wherever the random generator is used

include/mxnet/random_generator.h
azai91: @apeforest any thoughts on what to revise? if we want to be consistent we could create a random_generator.cc file and put the non-cuda implementations in there, but I personally think that is unnecessary for such small functions.

apeforest: @azai91 Thanks for your efforts in refactoring this. Since the main purpose of this PR is refactoring, I hope we can do it in the most elegant way. If we are making this file a public header, I would extract the implementation class Impl to another internal header file so that we do not expose the internal  implementation details. By doing that, we can have put the implementation details for both CPU and GPU there. Please let me know your thoughts.

lupesko:
@anirudh2290 @apeforest can you guys please review status again, and discuss with @azai91 whether you have additional feedback? Thanks!

include/mxnet/random_generator.h
azai91: it seems that they way we are using the random_generator, we expect the developer to be able to access the internal Impl class (https://github.com/azai91/incubator-mxnet/blob/6f7254c91709904a9fb6290f1998fcf2da818d0e/src/operator/random/unique_sample_op.h#L118). 

the class is is public as well. I may be interpretting the developers intentions incorrectly though.

apeforest: I suppose the developers did not intend to directly access internal Impl class but was rather lacking proper interface API. In fact, throughout the repository, I only found one place that uses `RandGenerator<cpu,` GType>::Impl` directly, and therefore I think it's not a big overhead to refactor that one line of code. Regarding how to separating Impl class from the interface, you might find this reference helpful: ehttps://cpppatterns.com/patterns/pimpl.html

aaronmarkham:
I can't build mxnet on master from source anymore… getting this error:
```
make: *** No rule to make target 'include/mxnet/././../../src/common/random_generator.h', needed by 'build/src/operator/random/sample_multinomial_op.o'.  Stop.
```
I updated my submodules... I just pulled latest:
```
ubuntu@ip-172-31-66-78:~/incubator-mxnet$ git log
commit e213286db5f885701ee10d0a312c15564959edd9
Author: Sandeep Krishnamurthy <sandeep.krishna98@gmail.com>
Date:   Fri Sep 14 10:18:49 2018 -0700
```
Any ideas?

PR #12380, opened 2018-08-28
Manually track num_max_thread

azai91:
## Description ##
Currently we rely on the omp lib (omp_set_max_threads/omp_get_max_threads) to set/fetch the max number threads. However, there is known issue with openmp where the num_thread values do not get updated in a multithreaded environment (https://software.intel.com/en-us/forums/intel-c-compiler/topic/494242). We will keep track of the num_threads manually.

This has caused an issue when the library is called in multi-threaded environment. we will instead only call omp_get_max_threads once and cached the result.

### Essentials ###

### Changes ###
- [x] reserve method modifies num_max_thread directly
- [x] GetRecommendedOMPThreadCount uses cached version  

## Comments ##

### DIFF
--- a/src/engine/openmp.cc
+++ b/src/engine/openmp.cc
@@ -79,10 +79,24 @@ void OpenMP::set_reserve_cores(int cores) {
 
 int OpenMP::GetRecommendedOMPThreadCount(bool exclude_reserved) const {
 #ifdef _OPENMP
-  if (omp_num_threads_set_in_environment_ || enabled_) {
+  if (omp_num_threads_set_in_environment_) {
     return omp_thread_max_;
   }
-  return 1;
+  if (enabled_) {
+    if (exclude_reserved) {
+      if (reserve_cores_ >= thread_count) {
+        thread_count = 1;
+      } else {
+        thread_count -= reserve_cores_;
+      }
+    }
+    // Check that OMP doesn't suggest more than our 'omp_thread_max_' value
+    if (!omp_thread_max_ || thread_count < omp_thread_max_) {
+      return thread_count;
+    }
+    return omp_thread_max_;
+  }
+
 #else
   return 1;
 #endif
--- a/src/engine/openmp.cc
+++ b/src/engine/openmp.cc
@@ -79,8 +79,9 @@ void OpenMP::set_reserve_cores(int cores) {
 
 int OpenMP::GetRecommendedOMPThreadCount(bool exclude_reserved) const {
 #ifdef _OPENMP
-  if (omp_num_threads_set_in_environment_ || enabled_)
+  if (omp_num_threads_set_in_environment_ || enabled_) {
     return omp_thread_max_;
+  }
   return 1;
 #else
   return 1;
--- a/src/engine/openmp.cc
+++ b/src/engine/openmp.cc
@@ -97,10 +97,8 @@ int OpenMP::GetRecommendedOMPThreadCount(bool exclude_reserved) const {
     }
     return omp_thread_max_;
   }
-
-#else
-  return 1;
 #endif
+  return 1;
 }
 
 OpenMP *__init_omp__ = OpenMP::Get();
--- a/src/engine/openmp.cc
+++ b/src/engine/openmp.cc
@@ -73,11 +73,7 @@ void OpenMP::set_reserve_cores(int cores) {
   CHECK_GE(cores, 0);
   reserve_cores_ = cores;
 #ifdef _OPENMP
-  if (reserve_cores_ >= omp_thread_max_) {
-    omp_thread_max_ = 1;
-  } else {
-    omp_thread_max_ -= reserve_cores_;
-  }
+  omp_thread_max_ = std::max(omp_thread_max_ - reserve_cores_, 1);
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


azai91:
@cjolivier01 please review

src/engine/openmp.cc
```diff
@@ -74,17 +74,17 @@ void OpenMP::set_reserve_cores(int cores) {
...
 }
 
 int OpenMP::GetRecommendedOMPThreadCount(bool exclude_reserved) const {
 #ifdef _OPENMP
   if (omp_num_threads_set_in_environment_) {
-    return omp_get_max_threads();
+    return omp_thread_max_;
   }
   if (enabled_) {
     int thread_count = omp_get_max_threads();
```
lupesko: You mention you want to call `omp_get_max_threads()` once and cache the results. This is another place where the API is called (in addition to line 55). Is that intentional? will it not create the same issue?

src/engine/openmp.cc
```diff
@@ -74,17 +74,17 @@ void OpenMP::set_reserve_cores(int cores) {
   reserve_cores_ = cores;
 #ifdef _OPENMP
   if (reserve_cores_ >= omp_thread_max_) {
-    omp_set_num_threads(1);
+    omp_thread_max_ = 1;
   } else {
-    omp_set_num_threads(omp_thread_max_ - reserve_cores_);
+    omp_thread_max_ -= reserve_cores_;
```
apeforest: I think this does not solve multithreading problem but rather causing race condition.

src/engine/openmp.cc
azai91: these are the lines I changed. I removed the call and replaced it with the cached result.

the only call is in the constructor (line 55).

src/engine/openmp.cc
azai91: isn't this atomic?

kalyc:
Thanks for your contribution @azai91 

vandanavk:
@apeforest for review

src/engine/openmp.cc
apeforest: Say at one time reserve_cores_ = 4 and omp_thread_max_ = 8

Thread 1: checks `reserve_cores_ >= omp_thread_max` is false so goto the else branch; at this time, thread2 is executing in the core and reserve_cores = 12 and therefore will update the omp_thread_max to 1. And when it comes to Thread 1 to run, it will execute `omp_thread_max_ -= reserve_cores_`. Now `omp_thread_max_` will be negative.

src/engine/openmp.cc
apeforest: @azai91 Can you address @lupesko comments? I also found this change suspicious. It seems once omp_thread_max_ is set to 1 in line 77, all the subsequent call to this function will return 1 if the env variable is set. Which means, this stops multithreading altogether in the operator. Could you provide a test case to verify this change actually works? Thanks

apeforest:
I cannot see how this change will solve the issue. Please provide a test case.

Roshrini:
Hi @azai91, This PR is open since long time now. Can you please address review comments?
@sandeep-krishnamurthy Can you please change lable to "pr-awaiting-response"?

apeforest:
@cjolivier01 Could you please review this change?

src/engine/openmp.cc
```diff
@@ -73,34 +73,14 @@ void OpenMP::set_reserve_cores(int cores) {
...
-        thread_count = 1;
-      } else {
-        thread_count -= reserve_cores_;
-      }
-    }
-    // Check that OMP doesn't suggest more than our 'omp_thread_max_' value
-    if (!omp_thread_max_ || thread_count < omp_thread_max_) {
-      return thread_count;
-    }
+  if (omp_num_threads_set_in_environment_ || enabled_)
```
apeforest: nit: enclose with {} even the body is just one line

azai91: kk

src/engine/openmp.cc
```diff
@@ -73,32 +73,13 @@ void OpenMP::set_reserve_cores(int cores) {
...
-        thread_count = 1;
-      } else {
-        thread_count -= reserve_cores_;
-      }
-    }
-    // Check that OMP doesn't suggest more than our 'omp_thread_max_' value
-    if (!omp_thread_max_ || thread_count < omp_thread_max_) {
-      return thread_count;
-    }
+  if (omp_num_threads_set_in_environment_ || enabled_) {
```
cjolivier01: You're ignoring exclude_reserved in this context

cjolivier01:
What problems are being caused?  It is my understanding that startup omp_set_num_threads() will propagage to child threads.  Beyond that, setting max threads has thread scope, which is what the issue link you pasted shows (which is the expected behavior).

ankkhedia:
@azai91 Could you please follow up on the comments given above?

kalyc:
@azai91 ping! could you please follow up on the comments above?
Thanks!

azai91:
@cjolivier01 addressed above issue. please let me know if anything else needs addressing.

zheng-da:
This PR is problematic. This causes significant performance regression in spmv in the multi-GPU training. It seems the PR affects the number of threads used for spmv. Please fix the problem ASAP.

Issue #10020, opened 2018-03-07
How can I build C++ package?


Mabinogiysk:
I can not find anything about installing mxnet for C++. 
The tutorials about [C++](https://mxnet.incubator.apache.org/tutorials/c++/basics.html) mentions ["building the C++ package"](https://mxnet.incubator.apache.org/get_started/build_from_source.html#build-the-c++-package), but the page can not open.
I wonder is there anyway to build mxnet C++ package?

edmBernard:
You can use this parameter when you run your make command :
`USE_CPP_PACKAGE=1`

marcoabreu:
@aaronmarkham see comment about website

aaronmarkham:
Hi @Mabinogiysk did #12388 fix this for you? If so, can you close this?
There's this page now: https://mxnet.incubator.apache.org/install/c_plus_plus.html

PR #11148, opened 2018-06-05
[MXNET-679] Refactor handling BLAS libraries with cmake

lebeg:
## Description ##

This is fixing a number of issues with handling BLAS libraries with cmake:

[[MXNET-115]](https://issues.apache.org/jira/projects/MXNET/issues/MXNET-115) USE_LAPACK is forced on all platforms
https://github.com/apache/incubator-mxnet/issues/8702 [DISCUSSION] Should we deprecate Makefile and only use CMake?
https://github.com/apache/incubator-mxnet/issues/8532 mxnet-mkl (v0.12.0) crash when using (conda-installed) numpy with MKL // (general MKL handling in comments)
https://github.com/apache/incubator-mxnet/issues/10175 MXNet MKLDNN build dependency/flow discussion
https://github.com/apache/incubator-mxnet/issues/9868 MKL and CMake
https://github.com/apache/incubator-mxnet/issues/9105 libmxnet.so load path error
https://github.com/apache/incubator-mxnet/issues/8729 Build amalgamation using a docker // (LAPACK with cross compilation in the comments)
https://github.com/apache/incubator-mxnet/issues/7852 Trouble installing MXNet on Raspberry Pi 3

### Essentials ###
- [x] The PR title start starts with [[MXNET-679]](https://issues.apache.org/jira/projects/MXNET/issues/MXNET-679)
- [x] Changes are complete
- [x] Code is well-documented
- [x] To the my best knowledge, examples are not affected by this change

### Changes ###
- [x] Moved all BLAS library handling to a separate module
- [x] LAPACK is handled properly and separately for different BLAS libs
- [x] Added LAPACK functionality when using Accelerate
- [x] Moved MKLML handling to a separate module
- [x] MKLML will be downloaded only if needed (previously always)
- [x] Enabled MKLML for mac
- [x] Added documentation for BLAS option in the cmake file
- [x] Made MKLML / MKLDNN separation that finally makes sense (see https://github.com/apache/incubator-mxnet/issues/9868)
- [x] Defined and documented the recommended order for checking availability of BLAS libraries (from most recommended to usual)
- [x] Disabled MKL, MKLML, MKLDNN for not desktop platforms and cross compilation
- [x] Switched on LAPACK for arm and android builds
- [x] Standardised android arm builds
- [x] Upgraded OpenBLAS version for cross-compilation to v0.3.2
- [x] Used standard random generators instead of rand_r for openmp threads used in rnn
[...]

PR #12388, opened 2018-08-29
Installation instructions consolidation

aaronmarkham:
## Description ##
Different installation pages were out of date or out of sync. The install page was over-complicated. The build from source instructions weren't really working.

This PR consolidates install info, directs users to where the more recent updates happen to be, and simplifies the basic user install. I started off just trying to fix the C++ setup info, but found so many other issues, I worked on those too.

## Features
* Fixes a bug that made it so you couldn't preview the dev version's install page without it redirecting you. (options.js)
* Updates the [build_from_source](http://34.201.8.176/versions/cplusplus_install/install/build_from_source.html) page 
    - Point to the OS-focused pages where this information is most up-to-date.
    - Added a [table that maps links to each binding](http://34.201.8.176/versions/cplusplus_install/install/build_from_source.html#installing-mxnet-language-bindings).
    - Leaves the centos/other linux sections there since this info isn't covered elsewhere
    - Leaves the NCCL section alone (is this even current info?)
    - Links this page in the [Download Source section](http://34.201.8.176/versions/cplusplus_install/install/index.html) found at the bottom of install
* Creates a new [C++ setup page](http://34.201.8.176/versions/cplusplus_install/install/c_plus_plus.html) that brings together C++ info found around the project.
* Refactors the install page. 
    - Moves the [validation section to a new page](http://34.201.8.176/versions/cplusplus_install/install/validate_mxnet.html)
    - Removes virtualenv (this was repeated multiple times in the file w/o much real value
    - Super simplifies the pip install sections
    - Links back to each OS's detailed setup guides
    - Links to specific install guide sections for Perl and Julia
    - and more... the file went from 2531 lines to 1062. So much easier to deal with... probably could take more pruning...
* Fixes some issues on the OS guide pages - the Python binding steps were missing

## Preview
http://34.201.8.176/versions/cplusplus_install/install/index.html

### DIFF
--- a/docs/install/build_from_source.md
+++ b/docs/install/build_from_source.md
@@ -1,18 +1,24 @@
 # Build MXNet from Source
 
-This document explains how to build MXNet from sources. Building MXNet from sources is a 2 step process.
+This document explains how to build MXNet from source code. Building MXNet from source is a two step process.
 
 1. Build the MXNet shared library, `libmxnet.so`, from [C++ source files](#build-the-shared-library)
-2. Install the language binding for MXNet. MXNet supports -
-   [C++](#build-the-c-package),
-   [Scala](#build-the-scala-package), [R](#build-the-r-package), and
-   [Julia](#build-the-julia-package).
+2. Install the [language bindings](#installing-mxnet-language-bindings) for MXNet. MXNet supports the following languages:
+    - Python
+    - C++
+    - Clojure
+    - Julia
+    - Perl
+    - R
+    - Scala
+
 
 ## Build the shared library
 
 ### Prerequisites
 
-You need C++ build tools and a BLAS library to build the MXNet shared library. If you want to run MXNet with GPUs, you will need to install CUDA and CuDNN first.
+You need C++ build tools and a BLAS library to build the MXNet shared library. If you want to run MXNet with GPUs, you will need to install [NVDIA CUDA and cuDNN](https://developer.nvidia.com/cuda-downloads) first.
+
 
 #### C++ build tools
 
@@ -26,99 +32,6 @@ You need C++ build tools and a BLAS library to build the MXNet shared library. I
    for Windows) to build the library.
 
 
-Select your preferences and follow the instructions to install MXNet from sources.
-<div class="btn-group opt-group" role="group">
-<button type="button" class="btn btn-default opt active">Linux</button>
-<button type="button" class="btn btn-default opt">macOS</button>
-<button type="button" class="btn btn-default opt">Windows</button>
-</div>
-<script type="text/javascript" src='../../_static/js/options.js'></script>
-
-<div class="linux">
-
-Then select the Linux distribution:
-<div class="btn-group opt-group" role="group">
-<button type="button" class="btn btn-default opt active">Ubuntu</button>
-<button type="button" class="btn btn-default opt">CentOS</button>
-<button type="button" class="btn btn-default opt">Others</button>
-</div>
-
-- **Ubuntu** for systems supporting the `apt-get`
-  package management program
-- **CentOS** for systems supporting the `yum` package
-  management program
-- **Others** for general Linux-like systems building dependencies from scratch.
-
-<div class="ubuntu">
-
-Refer to the <a href="ubuntu_setup.html">MXNet Ubuntu installation guide</a> for build from source instructions as well as installation of language bindings.
-
-</div>
-
-<div class="centos">
-
-Install build tools and git on `CentOS >= 7` and `Fedora >= 19`.
-
-```bash
-sudo yum groupinstall -y "Development Tools" && sudo yum install -y git
-```
-
-</div>
-
-<div class="others">
-
-Installing both `git` and `make` by following instructions on the websites is
-straightforward. Here we provide the instructions to build `gcc-4.8` from source codes.
-
-1. Install the 32-bit `libc` with one of the following system-specific commands:
-
-   ```bash
-   sudo apt-get install libc6-dev-i386 # In Ubuntu
-   sudo yum install glibc-devel.i686   # In RHEL (Red Hat Linux)
-   sudo yum install glibc-devel.i386   # In CentOS 5.8
-   sudo yum install glibc-devel.i686   # In CentOS 6/7
-   ```
-
-2. Download and extract the `gcc` source code with the prerequisites:
-
-   ```bash
-   wget http://mirrors.concertpass.com/gcc/releases/gcc-4.8.5/gcc-4.8.5.tar.gz
-   tar -zxf gcc-4.8.5.tar.gz
-   cd gcc-4.8.5
-   ./contrib/download_prerequisites
-   ```
-
-3. Build `gcc` by using 10 threads and then install to `/usr/local`
-
-   ```bash
-   mkdir release && cd release
-   ../configure --prefix=/usr/local --enable-languages=c,c++
-   make -j10
-   sudo make install
-   ```
-
-4. Add the lib path to your configure file such as `~/.bashrc`:
-
-   ```bash
-   export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/local/lib64
-   ```
-
-</div>
-</div> <!-- linux -->
-
-<div class="windows">
-
-1. If [Microsoft Visual Studio 2015](https://www.visualstudio.com/downloads/) is not already installed, download and install it. You can download and install the free community edition.
-2. Download and Install [CMake](https://cmake.org/) if it is not already installed.
-
-</div>
-
-<div class="macos">
-
-Install [Xcode](https://developer.apple.com/xcode/).
-
-</div>
-
 #### BLAS library
 
 MXNet relies on the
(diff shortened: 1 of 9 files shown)


docs/install/build_from_source.md
```diff
@@ -28,184 +32,102 @@ You need C++ build tools and BLAS library to build MXNet shared library. If you
    for Windows) to build the library.
```
lebeg: A C++ compiler that supports C++ 11.
[G++ (4.8 or later)](https://gcc.gnu.org/gcc-4.8/), [Clang](http://clang.llvm.org/) or [Visual Studio 2015](https://visualstudio.microsoft.com/downloads/) (on Windows) is required.

docs/install/build_from_source.md
```diff
@@ -28,184 +32,102 @@ You need C++ build tools and BLAS library to build MXNet shared library. If you
...
 
-   ```bash
-   wget http://mirrors.concertpass.com/gcc/releases/gcc-4.8.5/gcc-4.8.5.tar.gz
-   tar -zxf gcc-4.8.5.tar.gz
-   cd gcc-4.8.5
-   ./contrib/download_prerequisites
-   ```
+MXNet relies on the
+[BLAS](https://en.wikipedia.org/wiki/Basic_Linear_Algebra_Subprograms) (Basic
+Linear Algebra Subprograms) library for numerical computations. You can install
```
lebeg: ... library for numerical computations. Those can be extended with the LAPACK (Linear Algebra Package) - an additional set of mathematical functions.

docs/install/build_from_source.md
```diff
@@ -28,184 +32,102 @@ You need C++ build tools and BLAS library to build MXNet shared library. If you
...
+any one among [ATLAS](http://math-atlas.sourceforge.net/),
+[OpenBLAS](http://www.openblas.net/) and
+[MKL](https://software.intel.com/en-us/intel-mkl).
```
lebeg: `[Accelerate](https://developer.apple.com/documentation/accelerate) on MAC OSX.`

docs/install/build_from_source.md
```diff
@@ -28,184 +32,102 @@ You need C++ build tools and BLAS library to build MXNet shared library. If you
...
-</div>
-
-<div class="macos">
-
-macOS users can skip this step as `xcode` ships with a BLAS library.
-
-</div>
+### Other Linux
+Installing both `git` and `make` by following instructions on the websites is
+straightforward. Here we provide the instructions to build `gcc-4.8` from source codes.
```
lebeg: Are we sure that we need instructions how to build gcc? Is there any use case where it can not be installed with a package manager?

Say: `sudo apt install gcc-4.8` or `sudo apt install gcc-4.9`

docs/install/build_from_source.md
lebeg: Can we also make CMake the default (instead of GNU Make and cmake only for Windows)?

docs/install/build_from_source.md
```diff
@@ -226,109 +148,62 @@ To build OpenCV from source code, you need the [cmake](https://cmake.org) librar
...
+* Build with both OpenBLAS, GPU, and OpenCV support:
 
-Use [CMake](https://cmake.org/) to create a Visual Studio solution in ```./build```.
+```bash
+make -j USE_BLAS=openblas USE_CUDA=1 USE_CUDA_PATH=/usr/local/cuda USE_CUDNN=1
+```
 
-In Visual Studio, open the solution file,```.sln```, and compile it.
-These commands produce a library called ```mxnet.dll``` in the ```./build/Release/``` or ```./build/Debug``` folder.
+* Build on macOS with the default BLAS library and clang installed with `xcode` (OPENMP is disabled because it is not supported in default by clang):
```
lebeg: To use OpenMP on MacOS you need to install the Clang compiler `brew install llvm` (the one provided by Apple does not support OpenMP)

docs/install/build_from_source.md
```diff
@@ -226,109 +148,62 @@ To build OpenCV from source code, you need the [cmake](https://cmake.org) librar
...
+```
 
-  ```bash
-  make -j USE_BLAS=apple USE_OPENCV=0 USE_OPENMP=0
-  ```
+* Build without using OpenCV:
 
-</div>
+```bash
+make USE_OPENCV=0
```
lebeg: Maybe this would be a good place to put the BLAS libraries explanation from https://github.com/apache/incubator-mxnet/pull/11148. What do you think?

docs/install/index.md
```diff
@@ -63,314 +63,196 @@ Indicate your preferred configuration. Then, follow the customized commands to i
...
-sudo apt-get install graphviz
-pip install graphviz
 ```
-
-**Step 4**  Validate the installation by running simple MXNet code described [here](#validate-mxnet-installation).
-
-**Experimental Choice** If You would like to install mxnet with Intel MKL, try the experimental pip package with MKL:
-```bash
-$ pip install mxnet-mkl
+$ pip install mxnet
```
lebeg: Not mxnet-mkl anymore? There are a lot of other options as well: https://pypi.org/search/?q=mxnet Maybe add some recommendation or at least list some options here?

docs/install/index.md
```diff
@@ -385,2146 +267,796 @@ Follow the four steps in this [docker documentation](https://docs.docker.com/eng
...
-```
+- libblas (for linear algebraic operations)
 
-**Step 4**  Validate the installation by running simple MXNet code described [here](#validate-mxnet-installation).
+- libopencv (for computer vision operations. This is optional if you want to save RAM and Disk Space)
 
-</div> <!-- END of pip -->
+- A C++ compiler that supports C++ 11. The C++ compiler compiles and builds MXNet source code. Supported compilers include the following:
 
+- [G++ (4.8 or later)](https://gcc.gnu.org/gcc-4.8/). Make sure to use gcc 4 and not 5 or 6 as there
```
lebeg: Clang 3.9 - 6 as well

docs/install/index.md
```diff
@@ -385,2146 +267,796 @@ Follow the four steps in this [docker documentation](https://docs.docker.com/eng
...
+    cmake \
+        -DUSE_SSE=OFF \
+        -DUSE_CUDA=OFF \
+        -DUSE_OPENCV=ON \
+        -DUSE_OPENMP=ON \
+        -DUSE_MKL_IF_AVAILABLE=OFF \
+        -DUSE_SIGNAL_HANDLER=ON \
+        -DCMAKE_BUILD_TYPE=Release \
+        -GNinja ..
+    ninja -j1
```
lebeg: -j $(nproc)

docs/install/osx_setup.md
```diff
@@ -102,11 +102,20 @@ If building with ```GPU``` support, add the following configuration to config.mk
...
 - [Julia](#install-the-mxnet-package-for-julia)
 - [Scala](#install-the-mxnet-package-for-scala)
 - [Perl](#install-the-mxnet-package-for-perl)
 
+## Install MXNet for Python
+To install the MXNet Python binding navigate to the root of the MXNet folder then run the following:
+
+```bash
+$ cd python
+$ pip install -e .
```
lebeg:
```
Note that the `-e` flag is optional. It is equivalent to `--editable` and means that if you edit the source files, these changes will be reflected in the package installed.
```

docs/install/build_from_source.md
aaronmarkham: I'd be happy to do that, but so many other downstream instructions use `make`. Are we ready to update all of those too?

docs/install/build_from_source.md
aaronmarkham: seems like this would be a good spot to mention mkl-dnn? Also could link to those instructions within the project... wdyt?

docs/install/build_from_source.md
aaronmarkham: So, with Caffe2 on CentOS, I had to build gcc from scratch. I haven't tried MXNet on CentOS yet, but I wonder if whoever wrote this ran into a similar issue.

Since this is for "other linux" the `apt` instructions won't cut it. 
My preference would be to split out a separate page for CentOS and have thorough, tested instructions. (maybe as a separate PR?)

docs/install/build_from_source.md
aaronmarkham: This caused me to rearrange the page for a better flow. This is all good info. It should be double-checked in case it is only for cmake and won't work for make.

docs/install/index.md
aaronmarkham: I added the link to PyPI as a "footer" for every pip install instruction. That way users can look for them. I could also put in a footer note that adding *-mkl to a package would get them the experimental mkl support.... that would be good?

aaronmarkham:
@lebeg thanks for the review! I've address most of you comments. Please take a look at the build from source page again since I made a lot of changes after incorporating your cmake documentation on BLAS. That was all really good stuff to add here!

A big update would be to switch using cmake as the default and change all of the examples, but I think we should hold off until your cmake PR is merged, right?

aaronmarkham:
I could also add this table to the pip footer...

pip

docs/install/build_from_source.md
lebeg: Well, maybe it can wait then and be done in a separate PR

docs/install/build_from_source.md
lebeg: Definitely

docs/install/build_from_source.md
lebeg: Ah, yes you're right

docs/install/index.md
lebeg: Awesome

lebeg:
> A big update would be to switch using cmake as the default and change all of the examples, but I think we should hold off until your cmake PR is merged, right?

Yes, absolutely.

docs/install/build_from_source.md
```diff
@@ -350,143 +178,56 @@ make -j"$(nproc)"
...
-make rpkg
+make -j USE_BLAS=openblas USE_CUDA=1 USE_CUDA_PATH=/usr/local/cuda USE_CUDNN=1
 ```
 
-## Build the Scala package
-
-Both JDK and Maven are required to build the Scala package.
-
-<div class="ubuntu">
+* Build on **macOS** with the default BLAS library and clang installed with `xcode` (OPENMP is disabled because it is not supported in default by clang):
```
lebeg: Build on **macOS** with the default BLAS library and Clang installed with `xcode` (OpenMP is disabled because it is not supported by default by Apple version of Clang):

aaronmarkham:
This PR also fixes #10020

marcoabreu:
This PR broke the installation guide nightly test: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/NightlyTests/detail/master/79/pipeline

aaronmarkham:
I'll take a look. I removed the virtualenv section so it should be an easy fix.

PR #12397, opened 2018-08-29
update C++ example so it is easier to run

aaronmarkham:
## Description ##
This PR updates a C++ image classification example, fixing grammar, and adding instructions so it can be run through the first time.
Currently, you will segfault and fail a few times until you read through all of the tips and troubleshoot.

## Comments
Maybe this example should be moved into the cpp-package folder? Or, should the examples in the cpp-package folder be moved out to the general /mxnet/example folder?

Maybe this should use something from the model zoo instead of linking to old models on data.mxnet.io?

We can always do either of these later and just get this fixed up for now...

### DIFF
--- a/example/image-classification/predict-cpp/README.md
+++ b/example/image-classification/predict-cpp/README.md
@@ -1,13 +1,24 @@
 # Image Classification Example of C++
-This is a simple predictor which shows how to use c api for image classfication.
+This is a simple predictor which shows how to use the MXNet C++ API for image classification with a pre-trained ImageNet model.
 
-It uses opencv for image reading
+## Prerequisites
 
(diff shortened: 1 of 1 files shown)


example/image-classification/predict-cpp/README.md
```diff
@@ -1,13 +1,24 @@
 # Image Classification Example of C++
-This is a simple predictor which shows how to use c api for image classfication.
+This is a simple predictor which shows how to use the MXNet C++ API for image classification with a pre-trained ImageNet model.
```
lupesko: This example actually uses the C Predict API, and not the C++ API. They are not the same.

Issue #12203, opened 2018-08-16
flaky test: test_operator_gpu.test_depthwise_convolution


lebeg:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/PR-12181/4/pipeline

```
======================================================================

FAIL: test_operator_gpu.test_depthwise_convolution

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/usr/local/lib/python3.5/dist-packages/nose/case.py", line 198, in runTest

    self.test(*self.arg)

  File "/usr/local/lib/python3.5/dist-packages/nose/util.py", line 620, in newfunc

    return func(*arg, **kw)

  File "/work/mxnet/tests/python/gpu/../unittest/common.py", line 172, in test_new

    orig_test(*args, **kwargs)

  File "/work/mxnet/tests/python/gpu/../unittest/test_operator.py", line 1663, in test_depthwise_convolution

    np.testing.assert_allclose(arr1.asnumpy(), arr2.asnumpy(), rtol=1e-3, atol=1e-3)

  File "/usr/local/lib/python3.5/dist-packages/numpy/testing/nose_tools/utils.py", line 1396, in assert_allclose

    verbose=verbose, header=header, equal_nan=equal_nan)

  File "/usr/local/lib/python3.5/dist-packages/numpy/testing/nose_tools/utils.py", line 779, in assert_array_compare

    raise AssertionError(msg)

AssertionError: 

Not equal to tolerance rtol=0.001, atol=0.001



(mismatch 2.040816326530617%)

 x: array([[[[  7.347052,  -1.722254,   7.837829,   4.21605 ,  -1.359475,

            1.55463 ,   6.701931],

         [ 11.283103,  12.302897,  -9.111632,  -3.390831,  -4.708895,...

 y: array([[[[  7.348634,  -1.720118,   7.836634,   4.217753,  -1.361165,

            1.552893,   6.695152],

         [ 11.284711,  12.303747,  -9.112375,  -3.390379,  -4.709879,...

-------------------- >> begin captured logging << --------------------

common: INFO: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=640580015 to reproduce.

--------------------- >> end captured logging << ---------------------

```

ankkhedia:
@lebeg Thanks for filing the issue. We will look into this issue.

mseth10:
Fix in #12402

haojin2:
[...]

PR #12402, opened 2018-08-30
fixed flaky test issue for test_operator_gpu.test_depthwise_convolution

mseth10:
## Description ##
Issue not reproducible. Tolerance parameter (rtol) relaxed. This should fix the flakiness issue #12203 since mismatch percentage is small.

### Essentials ###

### Changes ###
- [x]  Tolerance parameter rtol modified to 1e-2

## Comments ##
- Passed more than 10,000 times on GPU
- @haojin2


lupesko:
@mseth If it is not reproducible, why are we relaxing tolerance?

tests/python/unittest/test_operator.py
```diff
@@ -1659,7 +1658,7 @@ def test_depthwise_convolution():
                             exe2.backward(exe2.outputs[0])
```
anirudh2290: can we check whether a downcast is happening when we do `arr2[:] = arr1` from float64 to float32. If yes lets fix that check if the test is still flaky.

mseth10: arr1 and arr2 are both float32. Downcast is happening during populating arr1 by random normal samples. I have made the casting explicit just to avoid errors.

lebeg:
Failing again: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/master/1549/pipeline/

mseth10:
@lupesko tolerance parameter is relaxed even though the issue is not reproducible as it might be the case that the exact environment is not replicated at our end, also the relaxation we are making is small.

Issue #11849, opened 2018-07-21
gluon.SymbolBlock cannot imports resnet trained with dtype="float16"


wgchang:
## Description
Cannot load fine-tuned resnet101 (incubator-mxnet/example/image-classification/symbols/resnet.py) with dtype="float16"  with "gluon.SymbolBlock.imports" method.

## Error Message:

AssertionError: Failed loading Parameter 'stage3_unit2_conv2_weight' from saved params: dtype incompatible expected <type 'numpy.float32'> vs saved <type 'numpy.float16'>
## Minimum reproducible example

```python
net = gluon.SymbolBlock.imports('resnet-101-symbol.json',['data','softmax_label'],'resnet-101-0007.params')  # This line gives the error message.
net = gluon.SymbolBlock.imports('resnet-101-symbol.json',['data','softmax_label']) 
print(net.collect_params())
```
## My Questions
In incubator-mxnet/example/image-classification/symbols/resnet.py, 
there is mx.sym.Cast for type conversion.
I fine-tuned resnet101 with dtype="float16", and I need to load this model as HybridBlock, However, the method gluon.SymbolBlock.imports makes every params' type in the network as float32. Therefore, the trained model cannot be updated.

Here, resnet-101-0007.params are trained with argument dtype='float16'
In resnet-101-symbol.json file, there is the Cast op.
  {
      "op": "Cast", 
      "name": "cast0", 
      "attrs": {"dtype": "float16"}, 
      "inputs": [[7, 0, 0]]
    }, 
It seems that gluon.SymbolBlock.imports does not consider the type conversion operator.

For now, I think I need to load all parameter manualy, and change types then save.
Is there any other solution to solve this problem?

nicklhy:
Got the same problem here. Any updates here?

ThomasDelteil:
@rahul003 same problem here, can't load back saved float16 models

sandeep-krishnamurthy:
Using below snippet:
```
import mxnet as mx

ctx = mx.cpu(0)
data = mx.nd.zeros((1,3,224,224), ctx=ctx, dtype='float64')
net_fp32 = mx.gluon.model_zoo.vision.resnet34_v2(pretrained=True, ctx=ctx)
net_fp32.cast('float64')
net_fp32.hybridize()
pred = net_fp32.forward(data)
net_fp32.export('resnet34_fp16', 0)
print('export fp16 model')

sm = mx.sym.load('resnet34_fp16-symbol.json')
[...]

PR #12412, opened 2018-08-30
Infer dtype in SymbolBlock import from input symbol

sandeep-krishnamurthy:
## Description ##
* Fix for the issue - https://github.com/apache/incubator-mxnet/issues/11849 
* Currently, Gluon symbol block cannot import any symbol with type other than fp32. All the parameters are created as FP32 leading to failure in importing the params when it is of type fp16, fp64 etc,
* In this PR, we infer the type of the symbol being imported and create the Symbol Block Parameters with that inferred type.

~~Created this PR for getting the early feedback. I am working on adding test cases for this and will update the PR soon~~ 
Added the tests

@szha @hetong007 @apeforest - Can you please take a look at this?
@Roshrini - I think this is important fix to be picked for 1.3. @szha ?

### Essentials ###

### Changes ###
* In this PR, we infer the type of the symbol being imported and create the Symbol Block Parameters with that inferred type.
* If we cannot infer the type, we go by usual route using default dtype (fp32) for creating the parameters.

### Comments ###
Below is the issue:

When you create mx.gluon.SymbolBlock(sm, input). It creates the parameters in the Block. Type is not passed for creating the parameter.
See here - https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/gluon/block.py#L1058 and the behavior is "If there is not parameter to get, it creates one and uses default type (fp32) https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/gluon/parameter.py#L688


python/mxnet/gluon/block.py
```diff
@@ -1053,13 +1053,34 @@ def __init__(self, outputs, inputs, params=None):
...
+            for i, arg in enumerate(arg_params):
+                if arg not in input_names:
+                    self.params.get(arg, allow_deferred_init=True, dtype=arg_types[i])
+
+            for i, aux in enumerate(aux_params):
+                if aux not in input_names:
+                    self.params.get(aux, grad_req='null', allow_deferred_init=True, dtype=aux_types[i])
+        else:
+            # Use default types for params
+            for i, arg in enumerate(arg_params):
```
apeforest: Maybe no enumerate is needed since you don't use index i in this loop?

python/mxnet/gluon/block.py
```diff
@@ -1053,13 +1053,34 @@ def __init__(self, outputs, inputs, params=None):
...
+                if arg not in input_names:
+                    self.params.get(arg, allow_deferred_init=True)
 
-        for i in out.list_auxiliary_states():
-            if i not in input_names:
-                self.params.get(i, grad_req='null', allow_deferred_init=True)
+            for i, aux in enumerate(aux_params):
```
apeforest: Same here, no need to get index i

python/mxnet/gluon/block.py
```diff
@@ -1086,3 +1107,49 @@ def _clear_cached_op(self):
...
+
+    # Try to infer types of other parameters.
+    if input_sym_arg_type and len(input_sym_arg_type) > 0:
+        params = {input_sym_name:input_sym_arg_type[0]}
+        arg_types, _, aux_types = out_params.infer_type(**params)
+        if arg_types is not None and len(arg_types) == len(arg_params) and \
+           aux_types is not None and len(aux_types) == len(aux_params):
+            infer_type_success = True
+
+    return (infer_type_success, arg_types, aux_types)
```
szha: if arg_types or aux_types are None, just fill it with default type. feel free to add a default_type argument to this function to make it flexible.

tests/python/unittest/test_gluon.py
```diff
@@ -336,6 +339,28 @@ def hybrid_forward(self, F, x):
...
 
+    # Test case to verify if initializing the SymbolBlock from a model with params 
+    # other than fp32 param dtype.
+    # Load a resnet model, cast it to fp64 and export
+    tmp = tempfile.mkdtemp()
+    tmpfile = os.path.join(tmp, 'resnet34_fp64')
+    ctx = mx.cpu(0)
+
+    net_fp32 = mx.gluon.model_zoo.vision.resnet34_v2(pretrained=True, ctx=ctx)
+    net_fp32.cast('float64')
```
lupesko: Shall we add another test that casts to float16?
The original issue reported import from float16 failing, and it might be appropriate to cover it as well.

python/mxnet/gluon/block.py
sandeep-krishnamurthy: Done.

tests/python/unittest/test_gluon.py
sandeep-krishnamurthy: Added a test for fp16.

python/mxnet/gluon/block.py
sandeep-krishnamurthy: N/A with latest refactoring.

sandeep-krishnamurthy:
@apeforest @szha @lupesko - Thanks for your time. Addressed your comments. Please have a look.

tests/python/unittest/test_gluon.py
```diff
@@ -336,6 +339,29 @@ def hybrid_forward(self, F, x):
...
 
+    # Test case to verify if initializing the SymbolBlock from a model with params 
+    # other than fp32 param dtype.
+
+    # 1. Load a resnet model, cast it to fp64 and export
+    tmp = tempfile.mkdtemp()
+    tmpfile = os.path.join(tmp, 'resnet34_fp64')
+    ctx = mx.cpu(0)
+
+    net_fp32 = mx.gluon.model_zoo.vision.resnet34_v2(pretrained=True, ctx=ctx, root=tmp)
```
apeforest: Is the model name in model zoo going to be maintained? If there is any change in the name, it would break this unit test. Not sure if we want to keep this dependency

sandeep-krishnamurthy: resnet34_v2 is a public function exposed through model_zoo.vision module. I think it is ok because we are not using string based selection of the model.

sandeep-krishnamurthy:
@szha @apeforest - Is this good to go?

tests/python/unittest/test_gluon.py
```diff
@@ -336,6 +339,34 @@ def hybrid_forward(self, F, x):
     net.hybridize()
     assert isinstance(net(mx.nd.zeros((16, 10))), mx.nd.NDArray)
 
+    # Test case to verify if initializing the SymbolBlock from a model with params 
+    # other than fp32 param dtype.
+
+    # 1. Load a resnet model, cast it to fp64 and export
+    tmp = tempfile.mkdtemp()
```
stu1130: should we delete the temporary directory when done with it?

sandeep-krishnamurthy: Temp gets automatically cleaned up.

stu1130: According to python docs, it seems to be deleted by the user. [The user of mkdtemp() is responsible for deleting the temporary directory and its contents when done with it.](https://docs.python.org/2/library/tempfile.html#tempfile.mkdtemp)

sandeep-krishnamurthy: Right. Thanks. I meant, temp gets automatically cleaned up after all the tests (test session).

sandeep-krishnamurthy: @marcoabreu - Can you please confirm if my understanding is correct? If not, I will add code to delete the temp directory created in the tests. Also, I see similar behavior in all other tests, where it creates temp dir, but, assumes, it will be cleaned up the system.

sandeep-krishnamurthy:
@zhreshold - Can you help look at this once, and merge if looks good?

python/mxnet/gluon/block.py
```diff
@@ -1086,3 +1094,54 @@ def _clear_cached_op(self):
...
+    aux_types = None
+
+    # Get Input symbol details. This will be used to infer types of
+    # other parameters.
+    input_sym_name = in_params.name
+    input_sym_arg_type = in_params.infer_type()[0]
+
+    # Try to infer types of other parameters.
+    if input_sym_arg_type and len(input_sym_arg_type) > 0:
+        params = {input_sym_name:input_sym_arg_type[0]}
```
zhreshold: this does not handle Grouped Symbol because you are slicing [0] from the symbol.

python/mxnet/gluon/block.py
```diff
@@ -1053,13 +1054,20 @@ def __init__(self, outputs, inputs, params=None):
...
-                self.params.get(i, allow_deferred_init=True)
+        # Infer type of parameters. Without this, every parameter will be created with
+        # default type i.e., fp32
+        arg_params = out.list_arguments()
+        aux_params = out.list_auxiliary_states()
 
-        for i in out.list_auxiliary_states():
-            if i not in input_names:
-                self.params.get(i, grad_req='null', allow_deferred_init=True)
+        arg_types, aux_types = _infer_param_types(inputs[0], out, arg_params, aux_params)
```
zhreshold: this does not handle Grouped Symbol because you are only slicing [0] from the symbol.

zhreshold: also, I think the type inference should occur in `cast` as well, otherwise it's buggy when user is trying to cast dtype of a cascaded network with symbolBlock inside.

tests/python/unittest/test_gluon.py
```diff
@@ -336,6 +339,41 @@ def hybrid_forward(self, F, x):
...
+    net_fp64 = mx.gluon.SymbolBlock(sm, inputs)
+    net_fp64.collect_params().load(tmpfile + '-0000.params', ctx=ctx)
+    # 3. Get a conv layer's weight parameter name. Conv layer's weight param is
+    # expected to be of dtype casted, fp64.
+    for param_name in net_fp64.params.keys():
+        if 'conv' in param_name and 'weight' in param_name:
+            break
+    assert np.dtype(net_fp64.params[param_name].dtype) == np.dtype(np.float64)
+
+    # Cast the symbol block to FP32 and try to forward a FP32 data.
```
sandeep-krishnamurthy: @zhreshold - Added a test here to verify the SymbolBlock.cast functionality.

sandeep-krishnamurthy:
@zhreshold - Can you please take a look at this? Thanks.

python/mxnet/gluon/block.py
```diff
@@ -1084,5 +1092,71 @@ def _clear_cached_op(self):
...
+        Defaults to 'float32', if unable to infer type.
+    """
+    arg_types = None
+    aux_types = None
+
+    # Get Input symbol details. This will be used to infer types of
+    # other parameters.
+    input_sym_names = [in_param.name for in_param in in_params]
+
+    # Try to infer input types. If not successful, we will set default dtype.
```
zhreshold: could you elaborate when infer type is not successful, and should we raise if that happens?

sandeep-krishnamurthy: In one of the unit test(pasting below) infer shape will be false and it need not be an issue, users are just assuming default type.

```python
    model = nn.HybridSequential()
    model.add(nn.Dense(128, activation='tanh'))
    model.add(nn.Dropout(0.5))
    model.add(nn.Dense(64, activation='tanh'),
              nn.Dense(32, in_units=64))
    model.add(nn.Activation('relu'))

    model.initialize()

    inputs = mx.sym.var('data')
    outputs = model(inputs).get_internals()

    smodel = gluon.SymbolBlock(outputs, inputs, params=model.collect_params())
```

zhreshold: I don't get why this would fail to infer_type?

sandeep-krishnamurthy: Because, input symbol is not created with dtype.
```python
>>> data = mx.sym.var('data')
>>> data.infer_type()
(None, None, None)
```
```python
>>> data = mx.sym.var('data', dtype='float32')
>>> data.infer_type()
([<class 'numpy.float32'>], [<class 'numpy.float32'>], [])
```

PR #12413, opened 2018-08-30
fixed docs/website build checkout bug

aaronmarkham:
## Description ##
Fixes a bug in the docs build pipeline where a failover case didn't checkout a different branch as expected. This caused master API docs to appear in older versions in CI. Local builds would run fine.

Added a hard error stop on the checkout step to prevent any issue like this popping up silently in CI.
Also fixed the echo statement to print exactly what should be run next.
Also removed Clojure from the v1.2.0 settings for docs generation. These should only generate in master or in 1.3.0+. 

For example, in the CI logs:
```
+ git checkout v1.0.0
error: pathspec 'v1.0.0' did not match any file(s) known to git.
+ git branch v1.0.0 upstream/v1.0.0
Branch v1.0.0 set up to track remote branch v1.0.0 from upstream.
```
It checks if the branch is there already, if not it grabs it from upstream. What is missing here is that it also needed to checkout the newly created branch. If all fails then the process aborts on this step, rather than continue with master.

## Comments
The consistent theme is patched in a separate PR: #12426.

### DIFF
--- a/docs/settings.ini
+++ b/docs/settings.ini
@@ -14,7 +14,7 @@ r_docs = 0
 scala_docs = 1
 
 [document_sets_v1.2.0]
-clojure_docs = 1
+clojure_docs = 0
 doxygen_docs = 1
 r_docs = 0
 scala_docs = 1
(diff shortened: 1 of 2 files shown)


docs/settings.ini
```diff
@@ -14,7 +14,7 @@ r_docs = 0
 scala_docs = 1
 
 [document_sets_v1.2.0]
```
szha: I see two copies of document_sets_1.2.0.

szha: Aaron explained offline that for each version one entry is for release branch while the other is for release tag. Currently, the pipeline is set up so that docs are built from branches.

Issue #12377, opened 2018-08-28
Flaky test: test_mkldnn.test_activation


lebeg:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/master/1529/pipeline

```
======================================================================
FAIL: test_mkldnn.test_activation
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/work/mxnet/tests/python/mkl/../unittest/common.py", line 172, in test_new
    orig_test(*args, **kwargs)
  File "/work/mxnet/tests/python/mkl/test_mkldnn.py", line 297, in test_activation
    check_activation_training(stype)
  File "/work/mxnet/tests/python/mkl/test_mkldnn.py", line 293, in check_activation_training
    check_numeric_gradient(test, in_location, numeric_eps=1e-2, rtol=0.16, atol=1e-4)
  File "/work/mxnet/python/mxnet/test_utils.py", line 912, in check_numeric_gradient
    ("NUMERICAL_%s"%name, "BACKWARD_%s"%name))
  File "/work/mxnet/python/mxnet/test_utils.py", line 491, in assert_almost_equal
    raise AssertionError(msg)
AssertionError: 
Items are not equal:
Error 2.232502 exceeds tolerance rtol=0.160000, atol=0.000100.  Location of maximum error:(0, 0, 0), a=0.445562, b=0.693506
 NUMERICAL_data: array([[[0.44556212, 0.2619341 , 0.77837706],
        [0.        , 0.8214429 , 0.5259812 ],
        [0.        , 0.        , 0.        ]],...
 BACKWARD_data: array([[[0.693506  , 0.26193386, 0.77837765],
        [0.        , 0.8214439 , 0.52598166],
        [0.        , 0.        , 0.        ]],...
-------------------- >> begin captured logging << --------------------
common: INFO: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=1284728931 to reproduce.
--------------------- >> end captured logging << ---------------------
```

pengzhao-intel:
@luobao-intel please take a look for the reason.

luobao-intel:
[...]

PR #12418, opened 2018-08-31
Fix flaky test: test_mkldnn.test_activation #12377

luobao-intel:
## Description ##
Fix the flaky test failure : test_mkldnn.test_activation. #12377
The problem locates in the finite difference method for gradient comparison. In this case, the large eps caused the wrong calculation and this pull request reduces the eps for more accurate calculation.
@pengzhao-intel

### DIFF
--- a/tests/python/mkl/test_mkldnn.py
+++ b/tests/python/mkl/test_mkldnn.py
@@ -281,7 +281,6 @@ def test_pooling():
         check_pooling_training(stype)
 
 
-@unittest.skip("Flaky test: https://github.com/apache/incubator-mxnet/issues/12377")
 @with_seed()
 def test_activation():
     def check_activation_training(stype):
--- a/tests/python/mkl/test_mkldnn.py
+++ b/tests/python/mkl/test_mkldnn.py
@@ -292,7 +292,7 @@ def test_activation():
             in_location = [mx.nd.array(data_tmp).tostype(stype)]
 
             test = mx.symbol.Activation(data, act_type="relu")
-            check_numeric_gradient(test, in_location, numeric_eps=1e-2, rtol=0.16, atol=1e-4)
+            check_numeric_gradient(test, in_location, numeric_eps=1e-6, rtol=0.16, atol=1e-4)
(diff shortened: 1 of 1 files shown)


pengzhao-intel:
Please enable the case which has been skipped in another PR.

pengzhao-intel:
@marcoabreu @lebeg please help take a reveiw

tests/python/mkl/test_mkldnn.py
```diff
@@ -292,7 +291,7 @@ def check_activation_training(stype):
             in_location = [mx.nd.array(data_tmp).tostype(stype)]
 
             test = mx.symbol.Activation(data, act_type="relu")
-            check_numeric_gradient(test, in_location, numeric_eps=1e-2, rtol=0.16, atol=1e-4)
+            check_numeric_gradient(test, in_location, numeric_eps=1e-6, rtol=0.16, atol=1e-4)
```
lebeg: Hm, this change is lowering the epsilon by 4 magnitudes. If this test was flaky before due to computational precision wouldn't this change make it worse?

pengzhao-intel: Actually, it will improve the precision of the reference results because we use the smaller eps for the finite difference method. 
Previously, the flaky test is caused by the large eps which can't simulate the small difference.
The numerical calculation sometimes is a little tricky :)

luobao-intel: This epsilon would have an impact on the baseline calculation instead of mkldnn calculation. And the smaller the epsilon is, the more accurate the baseline (gradient referring to theano) is. So this change won't make it worse.

lebeg: @pengzhao-intel @luobao-intel thanks for your explanations!

marcoabreu:
@szha @eric-haibin-lin

lupesko:
Thanks for the fix @luobao-intel !
Copying in more folks for review: @anirudh2290 @azai91 @mseth10

PR #12426, opened 2018-08-31
Consistent website theme and custom 404

aaronmarkham:
## Description ##
The PR applies the website theme to each version. The navigation will be the same, so the option for Clojure needs to be handled properly for old versions. For this I use the .htaccess file to redirect users to an API error page. For good measure, I also added a custom 404 error page.

This PR stacks on #12413 (has the same changes in build_all_version.sh), plus a change to copy the theme, and fixes my concerns there with the theme.

## Preview
http://34.201.8.176/
You can test the redirect if you switch to an old version like 1.0.0 and go to API --> Clojure
You can look at the 404:
http://34.201.8.176/error/404.html 
If you trigger a 404 by going to an [invalid link](http://34.201.8.176/123) it goes to the non-existent production URL. I didn't use a relative link because that won't update the URL and this causes formatting issues (described in the convo below) when `sidebar.js` does all of its wacky `document.ready` injections. When this PR goes live then it'll all come together.

## Comment
I'm sure there's probably some fancy regex that would collapse the clojure rules to one line, but I'll let someone else get fancy.


docs/error/404.md
```diff
@@ -0,0 +1,3 @@
+# Page Removed or Relocated or Never Existed
```
lupesko: Suggest to update this to something simpler, like: `Page does not exist`

docs/error/404.md
```diff
@@ -0,0 +1,3 @@
...
+
+If you're here that means you requested a page that doesn't exist. Sorry about that! Maybe try the search box to find what you're looking for.
```
lupesko: Suggest to also give a link to the master website: "Try the search box to find what you're looking for, or navigate to [mxnet home page](mxnet.io)"

aaronmarkham:
For anything in /api like http://34.201.8.176/api/python/index2.html you get

...which is unexpected to say the least. Every other directory behaves properly.

Please don't merge this until I figure out what's going on and patch it.

aaronmarkham:
Using the following would leave the URL intact which would then get special theme js applied to the error page when it had `/api` in the URL. Rather than try to fix the js, I updated the redirect to use a static URL which forces the URL to change thereby bypassing the theme.

aaronmarkham:
@lupesko - I made the suggested changes. Can you clear your change request?

Issue #12248, opened 2018-08-20
flaky test: test_operator_gpu.test_bilinear_sampler


lebeg:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/master/1456/pipeline

```
======================================================================
FAIL: test_operator_gpu.test_bilinear_sampler
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/usr/local/lib/python3.5/dist-packages/nose/util.py", line 620, in newfunc
    return func(*arg, **kw)
  File "/work/mxnet/tests/python/gpu/../unittest/test_operator.py", line 4086, in test_bilinear_sampler
    assert_almost_equal(exe.grad_dict['grid'].asnumpy(), grid_grad, rtol=1e-3, atol=1e-5)
  File "/work/mxnet/python/mxnet/test_utils.py", line 491, in assert_almost_equal
    raise AssertionError(msg)
AssertionError: 
Items are not equal:
Error 633.610779 exceeds tolerance rtol=0.001000, atol=0.000010.  Location of maximum error:(0, 1, 32, 24), a=0.007807, b=0.038601
 a: array([[[[ 0.        , -0.0205371 ,  0.        , ...,  0.        ,
           0.        ,  0.        ],
         [ 0.        ,  0.        ,  0.        , ...,  0.        ,...
 b: array([[[[ 0.        , -0.0205371 ,  0.        , ...,  0.        ,
           0.        ,  0.        ],
         [ 0.        ,  0.        ,  0.        , ...,  0.        ,...
----------------------------------------------------------------------
```

ankkhedia:
Hi @lebeg Thanks for filing the issue. We will look into the test.

haojin2:
Fix in #12427

PR #12427, opened 2018-08-31
Improved test for BilinearSampler

haojin2:
## Description ##
Fix for #12248.

### Essentials ###

### Changes ###
- [x] Add cudnn_off option to BilinearSampler
- [x] Remove old flaky unit test for BilinearSampler
- [x] Add new unit test of consistency for BilinearSampler

## Comments ##
CPU, GPU and CuDNN versions all failed on the same seed with the same "wrong" number, which means our 2 versions are consistent with the CuDNN version, so the reference answer calculated by the original helper function within the unit test should be problem here. Changing the test from comparing to the reference answer to a consistency check where the 3 versions are compared against each other to check for correctness.


haojin2:
New test passed more than 10000 trials on CPU, GPU and CuDNN:
```
MXNET_TEST_COUNT=10000 nosetests -s --verbose tests/python/gpu/test_operator_gpu.py:test_bilinear_sampler_versions
[INFO] Setting module np/mx/python random seeds, use MXNET_MODULE_SEED=206885670 to reproduce.
test_operator_gpu.test_bilinear_sampler_versions ... ok

Ran 1 test in 200.524s

OK
```

tests/python/gpu/test_operator_gpu.py
```diff
@@ -1918,6 +1918,65 @@ def test_softmax_activation():
         assert_almost_equal(cpu_a.grad.asnumpy(), gpu_a.grad.asnumpy(),
                 atol = 1e-3, rtol = 1e-3)
 
+
+@with_seed()
+def test_bilinear_sampler_versions():
```
samskalicky: why is this test function being added to test_operator_gpu.py when the old one was in test_operator.py? Do we have a similar test function just for CPU too, or do we now need to run GPU tests to test the CPU implementation for this case?

haojin2: yeah since we're checking the consistency of multiple versions, you need to run it where every version of it is available.

haojin2:
@marcoabreu @larroy @lebeg Can you guys take a look when you have time? Should be good for merge.

tests/python/gpu/test_operator_gpu.py
```diff
@@ -1918,6 +1918,65 @@ def test_softmax_activation():
...
+    data = mx.sym.Variable('data')
+    grid = mx.sym.Variable('grid')
+    sym1 = mx.sym.BilinearSampler(data=data, grid=grid)
+    sym2 = mx.sym.BilinearSampler(data=data, grid=grid, cudnn_off=True)
+    sym3 = mx.sym.BilinearSampler(data=data, grid=grid)
+
+    test_cases = [[(1,3,15,16),(1,2,10,10)],
```
lebeg: You could use [parameterized](https://pypi.org/project/parameterized/)

tests/python/gpu/test_operator_gpu.py
```diff
@@ -1918,6 +1918,65 @@ def test_softmax_activation():
...
+        exe_cudnn = sym3.simple_bind(data=data_shape, grid=grid_shape, ctx=default_context(), grad_req='write')
+        exe_list = [exe_cpu, exe_gpu, exe_cudnn]
+        ref_idx = 0
+        test_data = np.random.uniform(low=-0.1, high=0.1,size=data_shape).astype(np.float32)
+        test_grid = np.random.uniform(low=-2, high=2, size=grid_shape).astype(np.float32)
+        for exe in exe_list:
+            exe.arg_dict['data'][:] = test_data
+            exe.arg_dict['grid'][:] = test_grid
+            exe.forward(is_train=True)
+            assert_almost_equal(exe_list[0].outputs[0].asnumpy(), exe.outputs[0].asnumpy(), rtol=1e-3, atol=1e-5)
```
lebeg: The NumPy documentation suggests [assert_allclose](https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose) as better alternative to [assert_almost_equal](https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_almost_equal.html).

haojin2: I'm simply doing something similar to what we have for `check_consistency` function used in all other consistency-checking unit tests: https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/test_utils.py#L1204

tests/python/gpu/test_operator_gpu.py
haojin2: Seems like this introduces a new dependency, and many other tests would also need refactoring to migrate to this, so you want me to do this for this test only?

lebeg: Well is was just a suggestion, maybe we could consider it for a future improvement

tests/python/gpu/test_operator_gpu.py
lebeg: Same here, might be a future improvement

haojin2:
@anirudh2290 @sandeep-krishnamurthy @nswamy Should be ready for merge.

tests/python/gpu/test_operator_gpu.py
haojin2: Sure, thanks for the suggestion, I'm supportive of making those changes all together in a separate PR.

tests/python/gpu/test_operator_gpu.py
```diff
@@ -1918,6 +1918,65 @@ def test_softmax_activation():
...
+            exe.arg_dict['data'][:] = test_data
+            exe.arg_dict['grid'][:] = test_grid
+            exe.grad_dict['data'][:] = data_initial_grad
+            exe.grad_dict['grid'][:] = grid_initial_grad
+            exe.forward(is_train=True)
+            exe.backward(mx.nd.array(out_grad))
+            assert_almost_equal(exe.grad_dict['data'].asnumpy(), exe_list[ref_idx].grad_dict['data'].asnumpy(), rtol=1e-3, atol=1e-5)
+            assert_almost_equal(exe.grad_dict['grid'].asnumpy(), exe_list[ref_idx].grad_dict['grid'].asnumpy(), rtol=1e-3, atol=1e-5)
+        assert_almost_equal(exe_list[ref_idx].grad_dict['data'].asnumpy(), data_grad + data_initial_grad, rtol=1e-3, atol=1e-5)
+        assert_almost_equal(exe_list[ref_idx].grad_dict['grid'].asnumpy(), grid_grad + grid_initial_grad, rtol=1e-3, atol=1e-5)
```
anirudh2290: Is my understanding correct that we are doing only consistency checks between cpu , gpu and cudnn runs for the bilinear sampler and removing the correctness tests that we had earlier (bilinear sampler forward and backward computation using numpy ops) ? This should be fine assuming the correctness of cudnn op.

haojin2: Yes you're right, and the previous "reference" version actually delivers different results with all 3 versions on certain seeds.

Issue #12327, opened 2018-08-24
[Feature Request] support of diag for N-d arrays


jasonyu1996:
Hi! 

I just found that the `diag` operator does not support N-d arrays where `N > 2`. According to my own experience, it could be made more useful if the `N > 2` cases are properly designed. For example, I find it troublesome to take the diagonals of several matrices of the same shape at the same time. I know this task could be accomplished with a combination of `arange`, `tile` and `pick`, but it would be very complicated, confusing and error-prone. To support this, the behaviour when `N > 2` could be designed as taking the diagonal of the last two axes, i.e., when fed with an array of shape `[d1, d2, d3, ..., dn-2, dn-1, dn]`, where the diagonal of `[dn-1, dn]` is of length `k`, `diag` would return an array of shape `[d1, d2, d3, ..., dn-2, k]`. Of course, this could be designed to be more flexible (allowing specifying the axes to reduce, for example).

PyTorch provides a `diag` operator that behaves in the same way. Tensorflow actually splits it into two operators, `diag` and `diag_part`, the former of which constructs diagonal matrices and the latter takes diagonals from matrices. They are designed to support `N > 2` but not in a way I find useful or flexible.

On the MXNet forum: https://discuss.mxnet.io/t/diag-for-n-d-arrays/1707

ankkhedia:
@mxnet-label-bot [Feature Request, NDArray]

samskalicky:
Hi @jasonyu1996. Thanks for reporting this. Ive been looking to implement a trace operator in mxnet per this request #10500. In preparing for implementing trace (which is really just summing the diagonal of the matix) I also noticed the limited implementation of the diag operator. 

Given that many MXNet users have data in the form of WxHxC (width, height, channel), and then add a 4th dimension for number/batch, what are your thoughts on general N-dimensionality approach for this operator? Is it necessary to support the general case?

And how about implementation of the diag operator, as you mentioned there are already existing and high performance implementation for each sub-computation required. Do you think there is opportunity for further performance improvement (memory, time, etc.) by fusing these together? Or do you think it would be best to just implement diag calling these sub-computations separately (inside the diag operator)?

Let me know if you have thoughts on this. I would be interested in working with you to implement this as well.

jasonyu1996:
[...]

PR #12430, opened 2018-09-01
[MXNET-882] Support for N-d arrays added to diag op.

jasonyu1996:
## Description ##
Github Issue #12327
### Essentials ###
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-12430/3/index.html

### Changes ###
- [x] The `diag` operator now supports N-d arrays where `N > 2`. In this case it behaves in the same way as the `diagonal` operator provided by numpy (https://www.numpy.org/devdocs/reference/generated/numpy.diagonal.html). Axes of the sub-arrays can be specified with `axis1` and `axis2`.

## Comments ##
- This change is an extension to the existing interface, and therefore completely back compatible.


src/operator/tensor/diag_op-inl.h
```diff
@@ -30,33 +30,46 @@
...
+  dmlc::optional<int> axis2;
   DMLC_DECLARE_PARAMETER(DiagParam) {
     DMLC_DECLARE_FIELD(k)
     .set_default(dmlc::optional<int>(0))
     .describe("Diagonal in question. The default is 0. "
               "Use k>0 for diagonals above the main diagonal, "
               "and k<0 for diagonals below the main diagonal. "
               "If input has shape (S0 S1) k must be between -S0 and S1");
+    DMLC_DECLARE_FIELD(axis1)
+    .set_default(dmlc::optional<int>(0))
```
samskalicky: Would it be better to use unsigned here instead of int since the axes cannot be negative?

src/operator/tensor/diag_op-inl.h
```diff
@@ -30,33 +30,46 @@
...
+inline TShape DiagShapeImpl(const TShape& ishape, const nnvm::dim_t k,
+                            const nnvm::dim_t axis1, const nnvm::dim_t axis2) {
   if (ishape.ndim() == 1) {
     auto s = ishape[0] + std::abs(k);
     return TShape({s, s});
   }
 
-  auto h = ishape[0];
-  auto w = ishape[1];
+  auto h = ishape[axis1];
```
samskalicky: related to the previous comment, shouldnt we check that the provided axis1 and axis2 values are within the acceptable range [0:ndim-1] so that users dont get nasty segfaults on incorrect inputs?

samskalicky: I see there is a check later on line 85/86, so maybe consider re-organizing the code so that this is looked up after the check

src/operator/tensor/diag_op-inl.h
```diff
@@ -69,7 +82,28 @@ inline TShape DiagShapeImpl(const TShape& ishape, const nnvm::dim_t k) {
...
+  TShape oshape(n_dim);
+
+  // remove axis1 and axis2 and append the new axis to the end
+  for (int i = 0; i < x1; i ++)
+    oshape[i] = ishape[i];
+  for (int i = x1 + 1; i < x2; i ++)
+    oshape[i - 1] = ishape[i];
+  for (int i = x2 + 1; i <= n_dim; i ++)
+    oshape[i - 2] = ishape[i];
+  oshape[n_dim - 1] = s;
```
samskalicky: Could this be simplified to something like:

int idx = 0;
for(int i=0; i<ndim; i++) {
   if(i != axis1 && i != axis2)
      oshape[idx++] = ishape[i]
oshape[ndim-2] = s;

Also, the final dimension assignment on line 104 looks wrong. Shouldnt it be oshape[n_dim - 2] = s;? Since the output tensor will be 1 dimension smaller than the input tensor (and off-by-1 for 0-based versus ndim being the count).

src/operator/tensor/diag_op-inl.h
jasonyu1996: I think you are right. These checks should be done first before using axis1 and axis2. Also I find it necessary to add some tests testing cases with negative axes.

src/operator/tensor/diag_op-inl.h
jasonyu1996: Yes, the code could be simplified that way, but that does not necessarily make improvements in my opinion. The original implementation is equally easy to read and fairly efficient. As for the final assignment, note that `n_dim` is already the dimension of the input tensor minus 1, so I think it is fine.

jasonyu1996:
@samskalicky Thanks for your review! I have accordingly made modifications to the code.

samskalicky:
@jasonyu1996 regarding this error, what have you tried to debug it? The message doesnt look very meaningful. If you run the same test in the master branch does it fail the same way? You might have to try re-applying some of your changes and see when it starts breaking to debug. 

There are some instructions here to setup debugging (if you're using a Mac): https://cwiki.apache.org/confluence/display/MXNET/MXNet+Developer+Setup+on+Mac but im also working on improving this and putting together a debug guide so please post any steps or processes that you do (whether they worked or not) and i'll be sure to include those.

src/operator/tensor/diag_op-inl.h
samskalicky: If the values are negative, where are negative values acceptible? When they are used later on to index directly into the ishape array on line 76 wouldnt that cause a problem?

src/operator/tensor/diag_op-inl.h
```diff
@@ -80,11 +112,13 @@ inline bool DiagOpShape(const nnvm::NodeAttrs& attrs,
 
     const TShape& ishape = (*in_attrs)[0];
     if (ishape.ndim() == 0) return false;
-    if (ishape.ndim() > 2) LOG(FATAL) << "Input must be 1- or 2-d.";
```
samskalicky: What is the diagonal for a 1D input tensor? Shouldnt that be an error?

src/operator/tensor/diag_op-inl.h
```diff
@@ -30,33 +30,51 @@
...
   }
 };
 
-inline TShape DiagShapeImpl(const TShape& ishape, const nnvm::dim_t k) {
+inline TShape DiagShapeImpl(const TShape& ishape, const nnvm::dim_t k,
+                            const nnvm::dim_t axis1, const nnvm::dim_t axis2) {
   if (ishape.ndim() == 1) {
     auto s = ishape[0] + std::abs(k);
     return TShape({s, s});
   }
```
samskalicky: Can you please explain what this is doing (i understand its not a change you are making) but what is this doing and why is that the way to calculate the diagonal for 1D tensors?

src/operator/tensor/diag_op-inl.h
```diff
@@ -159,21 +274,7 @@ void DiagOpForward(const nnvm::NodeAttrs& attrs,
...
-    });
-  } else {
-    MSHADOW_TYPE_SWITCH(out_data.type_flag_, DType, {
-      MXNET_ASSIGN_REQ_SWITCH(req[0], req_type, {
-        Kernel<diag_gen<req_type>, xpu>::Launch(s, out_data.Size(), out_data.dptr<DType>(),
-                            in_data.dptr<DType>(), Shape2(oshape[0], oshape[1]), param.k.value());
-      });
-    });
-  }
+  DiagOpProcess<xpu, false>(in_data, out_data, ishape, oshape, out_data.Size(), param, s, req);
```
samskalicky: I like that you put all the diag setup work in the DiagOpProcess function, but now that leaves DiagOpForward as a function that doesnt actually do anything. It seems like an unnecessary additional function call to call DiagOpProcess. What do you think about inlining DiagOpProcess inside of DiagOpForward? Or is there some benefit to doing it this way that im missing?

src/operator/tensor/diag_op-inl.h
```diff
@@ -104,26 +138,26 @@ inline bool DiagOpType(const nnvm::NodeAttrs& attrs,
...
+                                  int stride, int offset,
+                                  int base) {
     using namespace mxnet_op;
-    int j = 0;
-    if (k > 0) {
-      j = ravel(mshadow::Shape2(i, i + k), ishape);
-    } else if (k < 0) {
-      j = ravel(mshadow::Shape2(i - k, i), ishape);
+    int idx = i / base;
+    int j = ravel(unravel(idx, oshape), ishape) + offset + stride * (i - idx * base);
```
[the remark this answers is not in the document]
jasonyu1996: @samskalicky The base is equal to the size of the diagonal. The output is reshaped to `oshape + (base,)`, and the input is reshaped to `ishape`. `idx` is therefore the part corresponding to `oshape` and `ishape`. And it corresponds to the mth element in the diagonal, where `m = i % base = I - idx * base`. We need to multiply it by `stride`, which is basically the sum of the strides of `axis1` and `axis2` in order compute the index of the element we want. However, because the diagonal we want is not necessarily the main diagonal (only when `k = 0`), we need to add to the result an offset.

samskalicky:
@jasonyu1996 There is a PR to fix this issue https://github.com/dmlc/nnvm/pull/525 . I know that this issue is blocking everyone else so hopefully this gets through soon and we can get your changes merged too.

src/operator/tensor/diag_op-inl.h
samskalicky: thanks!

src/operator/tensor/diag_op-inl.h
samskalicky: makes sense, thanks!

src/operator/tensor/diag_op-inl.h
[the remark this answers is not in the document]
samskalicky: thanks!

src/operator/tensor/diag_op-inl.h
```diff
@@ -133,13 +167,94 @@ struct diag_gen {
     auto j = unravel(i, oshape);
     if (j[1] == (j[0] + k)) {
```
samskalicky: if the input shape is (3,3) then the output diagonal will be (3), will this cause a problem with indexing j[1] since there wont be a 2nd dimension in the oshape?

jasonyu1996:
@samskalicky Thanks! As what the diag operator does is basically copying some elements from the input directly to output, the backward process is quite straightforward. What it does is simply copying the gradients from the output back to the corresponding elements of the input. This is achieved with a template parameter indicating whether it is forward or backward to control the direction of the assignment in the copy process.

src/operator/tensor/diag_op-inl.h
jasonyu1996: `diag_gen` is what is reserved for generating 2D matrices from 1D inputs (a functionality from the original `diag` operator). It would not be invoked when the input has more than 1 dimension.

samskalicky:
@anirudh2290 Please review/merge

jasonyu1996:
@anirudh2290 @samskalicky Thanks!

src/operator/tensor/diag_op-inl.h
```diff
@@ -69,7 +87,21 @@ inline TShape DiagShapeImpl(const TShape& ishape, const nnvm::dim_t k) {
     s = 0;
   }
 
-  return TShape({s});
+  if (x1 > x2) {
+    std::swap(x1, x2);
+  }
+
+  int n_dim = static_cast<int>(ishape.ndim()) - 1;
```
apeforest: Can n_dim be 0?

samskalicky: No. What is a 0-dimensional array?

src/operator/tensor/diag_op-inl.h
```diff
@@ -133,13 +167,94 @@ struct diag_gen {
...
+
+    int idim = ishape.ndim(), odim = oshape.ndim();
+
+    int minx = x1, maxx = x2;
+    if (minx > maxx)
+      std::swap(minx, maxx);
+
+    int oleading = 1, obody = 1, otrailing = 1;
+    for (int i = 0; i < minx; i ++)
+      oleading *= ishape[i];
```
apeforest: Is it possible to have int overflow here?

src/operator/tensor/diag_op-inl.h
```diff
@@ -133,13 +167,94 @@ struct diag_gen {
...
+    for (int i = minx + 1; i < maxx; i ++)
+      obody *= ishape[i];
```
apeforest: same: int overflow?

src/operator/tensor/diag_op-inl.h
```diff
@@ -133,13 +167,94 @@ struct diag_gen {
...
+    for (int i = maxx + 1; i < idim; i ++)
+      otrailing *= ishape[i];
```
apeforest: same: int overflow?

src/operator/tensor/diag_op-inl.h
apeforest: What if ishape.ndim()==1

src/operator/tensor/diag_op-inl.h
samskalicky: This code is counting the number of elements, so its possible the total number of elements in an array could be greater than 2,147,483,647 (2^31-1). This would be an 8GB array though. Seems like a special case. If you have a data set so large so that the data operated on by a single operator is 8GB you're probably going to batch it up into something smaller.

This would be a different feature request. And we would probably want to review this separately from the feature in this PR (extending 2D to N-D)

src/operator/tensor/diag_op-inl.h
apeforest: Ok, I see. You've already checked that on line 66.

src/operator/tensor/diag_op-inl.h
apeforest: Given this PR is to support N-D tensor, I wouldn't call an array of size greater than 2^31-1 a rare case though. Using int here simply leaves door for an overflow bug in the future. A common way in this module to index array size is to use the variable index_t. Please see https://github.com/apache/incubator-mxnet/blob/ba8a9d13e1b549d061f1933c463cfad5e7bdd7aa/src/operator/nn/softmax-inl.h#L57

src/operator/tensor/diag_op-inl.h
```diff
@@ -104,42 +143,132 @@ inline bool DiagOpType(const nnvm::NodeAttrs& attrs,
...
   }
 };
 
-template<int req>
+template<int req, bool back>
 struct diag_gen {
   template<typename DType>
-  MSHADOW_XINLINE static void Map(int i, DType* out, const DType* a,
+  MSHADOW_XINLINE static void Map(index_t i, DType* out, const DType* a,
                                   mshadow::Shape<2> oshape, int k) {
```
samskalicky: Should "int k" here be "index_t k" also?

src/operator/tensor/diag_op-inl.h
```diff
@@ -104,42 +143,132 @@ inline bool DiagOpType(const nnvm::NodeAttrs& attrs,
...
+                   index_t dsize,
+                   const DiagParam& param,
+                   mxnet_op::Stream<xpu> *s,
+                   const std::vector<OpReqType>& req) {
+  using namespace mxnet_op;
+  using namespace mshadow;
+  if (ishape.ndim() > 1) {
+    // input : (leading + i, body + i, trailing)
+    int x1 = CheckAxis(param.axis1.value(), ishape.ndim());
+    int x2 = CheckAxis(param.axis2.value(), ishape.ndim());
```
samskalicky: While we're at it, shouldnt we change x1 and x2 to index_t type?

src/operator/tensor/diag_op-inl.h
```diff
@@ -104,42 +143,132 @@ inline bool DiagOpType(const nnvm::NodeAttrs& attrs,
...
+
+    int idim = ishape.ndim(), odim = oshape.ndim();
```
samskalicky: index_t idim

src/operator/tensor/diag_op-inl.h
```diff
@@ -104,42 +143,132 @@ inline bool DiagOpType(const nnvm::NodeAttrs& attrs,
...
+
+    int minx = x1, maxx = x2;
```
samskalicky: index_t minx, maxx

src/operator/tensor/diag_op-inl.h
```diff
@@ -104,42 +143,132 @@ inline bool DiagOpType(const nnvm::NodeAttrs& attrs,
...
+    if (minx > maxx) {
+      std::swap(minx, maxx);
+    }
+
+    index_t oleading = 1,
+           obody = 1,
+           otrailing = 1;
+
+    for (int i = 0; i < minx; ++i) {
```
samskalicky: index_t i here too, and for the other for loops below

src/operator/tensor/diag_op-inl.h
```diff
@@ -104,42 +143,132 @@ inline bool DiagOpType(const nnvm::NodeAttrs& attrs,
...
+        itrailing = otrailing * ishape[maxx];
+
+    index_t stride1 = itrailing * obody,
+        stride2 = otrailing;
+
+    if (x1 == maxx) {
+      std::swap(stride1, stride2);
+    }
+    index_t offset;
+    int k = param.k.value();
```
samskalicky: index_t k

src/operator/tensor/diag_op-inl.h
```diff
@@ -104,42 +143,132 @@ inline bool DiagOpType(const nnvm::NodeAttrs& attrs,
   return (*out_attrs)[0] != -1;
 }
 
-template<int req>
+template<int ndim, int req, bool back>
```
samskalicky: should ndim here be index_t too? or maybe im going to far with the index_t.... ;-)

src/operator/tensor/diag_op-inl.h
samskalicky: oh and odim too

src/operator/tensor/diag_op-inl.h
```diff
@@ -104,42 +143,132 @@ inline bool DiagOpType(const nnvm::NodeAttrs& attrs,
...
+        if (back && req[0] != kAddTo && req[0] != kNullOp) {
+          out_data.FlatTo1D<xpu, DType>(s) = 0;
+        }
+        if (ileading == 1) {
+          Kernel<diag<2, req_type, back>, xpu>::Launch(s, dsize, out_data.dptr<DType>(),
+                              in_data.dptr<DType>(), Shape2(obody, otrailing),
+                              Shape2(ibody, itrailing),
+                              stride1 + stride2, offset, oshape[odim - 1]);
+        } else {
+          Kernel<diag<3, req_type, back>, xpu>::Launch(s, dsize, out_data.dptr<DType>(),
```
samskalicky: Why is the ndim template argument now fixed to 2 or 3 here? shouldnt it be generic using the odim variable you define in line 202?

src/operator/tensor/diag_op-inl.h
apeforest: Isn't that way too large for the number of dimensions? :)

samskalicky: what about for 1x1x1x1x1x1x1xx1x1x1x.......x1x1x1x1x1x1 tensors?

src/operator/tensor/diag_op-inl.h
jasonyu1996: That is because for efficiency here I have merged some contiguous axes together, and the resulting shape cannot have more than 3 axes. In the original shape, there is no need to distinguish two axes if they are not separated by `axis1` or `axis2`.

src/operator/tensor/diag_op-inl.h
jasonyu1996: `ndim` here can only be 2 or 3 because I do some merge first.

src/operator/tensor/diag_op-inl.h
jasonyu1996: After some search work, I found that there is a type specially designed for holding the size of an axis: `dim_t`. I will try changing it to this type. `index_t` is not quite suitable since as far as I know it is unsigned, but `k` can be negative in our case.

jasonyu1996:
@samskalicky @apeforest Thanks again!

jasonyu1996:
@samskalicky @apeforest It seems that `dim_t` is poisonous in some cases. I have changed to following examples. `k` changed to `int` following the old `diag` implementation, `axis1` and `axis2` changed to `int32_t` following `swapaxes` and `concat`. I notice this is actually not very consistent in different operator implementations, and believe that if one day it is necessary to have a tensor of that many dimensions, these operators might have to go through some refactoring work.

src/operator/tensor/diag_op-inl.h
samskalicky: More efficiency is always great. But im having trouble following the abstraction to 2 or 3 axes. Can you elaborate some more on the approach/implementation?

jasonyu1996: Okay. Let's assume that the input shape is `(s1, s2, ..., sa - 1, sa, sa+1, ... , sb - 1, sb, sb + 1, ... sn)` where `a = axis1, b = axis2`. What would be the output shape? It would be `(s1, s2, ..., sa - 1, sa + 1, ..., sb - 1, sb + 1, ... sn, sd)` where `sd` is the size of the diagonal of a `sa x sb` matrix. Note that `s1, s2, ..., sa - 1`, `sa + 1, sa + 2, ..., sb - 1` and `sb + 1, sb + 2, ... sn` remain untouched (directly mapped), and for this reason we can merge them respectively. To remove `sa` and `sb` we can also merge them into segments `sa + 1, sa + 2, ..., sb - 1` and `sb + 1, sb + 2 , ... , sn` respectively, thus reshaping the original input into `(s1 * s2 * ... * sa - 1, sa * sa + 1 * ... * sb - 1,  sb * sb + 1 * ... * sn)`, and the corresponding output into `(s1 * s2 * ... * sa - 1, sa + 1 * sa + 2 * ... * sb - 1, sb + 1 * sb + 2 * ... * sn, sd)`. In this way, to get `output(x1, x2, x3, x4)`, we can simply find the index of `input(x1, x2, x3)` plus an offset `x4 * (stride1 + stride2)` where `stride1` and `stride2` are strides of `axis1` and `axis2` in the input respectively.

jasonyu1996:
@apeforest @samskalicky Am I expected to do something more to get this PR merged?

samskalicky:
Sorry @jasonyu1996, i'll take a look at this today and give you a status update. Thanks for your patience!

src/operator/tensor/diag_op-inl.h
samskalicky: Ok cool that makes sense. Thats a nice simplification to compute the index mapping between input/output coordinates. It might be worth while to capture some of this optimization in comments in the code somewhere.

lupesko:
@sandeep-krishnamurthy @nswamy @anirudh2290 - looks like this one is ready to go, can you please review/merge? Thanks!

src/operator/tensor/diag_op-inl.h
jasonyu1996: @samskalicky I will take your advice! Thanks!

PR #12447, opened 2018-09-03
MXNET-884 - Clojure with-resources macro for Improved Memory Handling

gigasquid:
## Description ##

This adds a `with-resources` macro to the Clojure package to provide an idiomatic way, (similar to `with-open`), to explicitly dispose of resources with a `.dispose` call.

It works with any Scala object that has a `.dispose` method -  NDArray, Symbol, Executor, Iterators, etc..

Usage is:

```clojure 
(util/with-resources [a (ndarray/ones [3])
                      b (ndarray/ones [3])]
    (ndarray/+ a b))
```
Any resources in the binding vector will be disposed. Result resources will not.

### Essentials ###

## Comments ##
- The Scala package uses scopes with an NDArrayCollector. However, scopes do not mesh well with the Clojure language and the macro pattern for `with-open` for resources is well established and idiomatic in Clojure.
- This will allow users to better control the resources and memory management, however it is not a full solution. There is still investigation into other avenues ongoing with the Scala package.
- Applications of with-resources to the examples will happen in a separate PR

### DIFF
--- a/contrib/clojure-package/src/org/apache/clojure_mxnet/util.clj
+++ b/contrib/clojure-package/src/org/apache/clojure_mxnet/util.clj
@@ -204,3 +204,18 @@
     (throw (ex-info error-msg
                     (s/explain-data spec value)))))
 
+(defmacro with-resources
+  "bindings => [name init ...]
+  Similar to with-open. Evaluates body in a try expression with names bound to the values
+  of the inits, and a finally clause that calls (.dispose name) on each
+  name in reverse order. Use for NDArrays, Symbols, Iterators, Executors and others that need to be disposed"
+  [bindings & body]
+  (cond
+    (= (count bindings) 0) `(do ~@body)
+    (symbol? (bindings 0)) `(let ~(subvec bindings 0 2)
+                              (try
+                                (with-resources ~(subvec bindings 2) ~@body)
+                                (finally
+                                  (. ~(bindings 0) dispose))))
+    :else (throw (IllegalArgumentException.
+                  "with-open only allows Symbols in bindings"))))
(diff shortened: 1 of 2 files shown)


iperdomo:
This is an improvement that looks more _idiomatic_. Our current approach is to remember to call `.dispose` on those `NDArray` objects we create, something like:
```clojure
(let [a (ndarray/ones [3])
      b (ndarray/ones [3])
      result (ndarray/+ a b)]
  (doseq [nd a]
    (.dispose ^NDArray nd))
  (doseq [nd b]
    (.dispose ^NDArray nd))
  result)
```
With some discipline you can remember to do it, however our current main issues are `NDArray` objects created at the Scala API that we can't _reach_ and call `.dispose` on them.

gigasquid:
@iperdomo Thanks for the feedback. This PR is a step in the direction of improved memory handling but I agree it does not solve your larger issue. I'm continuing to look into solutions with the Scala package.

gigasquid:
@nswamy thoughts on merging this PR or waiting until the largest memory management effort?

The pro for merging would be that it might be a minor help right now and doesn't harm anything
The con would be if we are changing direction and won't be using `close` and we would have to deprecate later

nswamy:
The work I am doing is here
https://github.com/nswamy/incubator-mxnet/blob/using-handle/scala-package/core/src/main/scala/org/apache/mxnet/NativeResource.scala
https://github.com/nswamy/incubator-mxnet/blob/using-handle/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
https://github.com/nswamy/incubator-mxnet/blob/using-handle/scala-package/core/src/main/scala/org/apache/mxnet/NDArray.scala#L566-L570

Do you think it makes sense to wait for these changes to get merged ? but i'll be fine if you want give this interim solution to users, however suggest to mark it experimental so you have the flexibility to modify.

I am adding the NativeResource class so all the MXNet Objects that have native Resources can extend be automatically taken taken care of. NativeResource also implements AutoCloseable - is that something that Clojure can see?

gigasquid:
@nswamy Thanks for the explanation. I'm not sure about the visibility of traits themselves to Clojure, but it can work with Scopes. I think I'm going to go ahead and close this PR since Clojure uses can use `dispose` or java interop with the NDArrayCollector if they need to solve immediate problems and I'll wait for your work to land to integrate the better solution.

Issue #11412, opened 2018-06-27
cannot find documentation for some python api


zchrissirhcz:
## Description
I read [MXNet's module API tutorial](https://mxnet.incubator.apache.org/tutorials/basic/module.html) and confused to some Python API's usage. On [the official python api url](https://mxnet.incubator.apache.org/api/python/) I can find many documentations but some are not find.

## Environment info (Required)

the environment is not required. I search the python API on your official website.

## Build info (Required if built from source)

The build info is not required.  I search the python API on your official website.

## Error Message:
There is not error message since no code run.

## Steps to reproduce

1. Go to https://mxnet.incubator.apache.org/tutorials/basic/module.html
2. Find some functions that don't know their concrete argument meaning
3. For example, `mx.sym.Convolution` can be found by typing `mxnet.symbol.Convolution`; However, `mx.test_utils` can never be found.

Hoping the documentation can be complete, thanks.

frankfliu:
Hi @zchrissirhcz thanks for the question, @sandeep-krishnamurthy requesting this be labeled under Question

frankfliu:
For questions, please submit on MXNet discussion forum (https://discuss.mxnet.io), where it will get a wider audience and allow other to learn as well.
@sandeep-krishnamurthy to close this issue now in favor of the discussion forum issue you will file, please feel free to re-open if closed in error.
Thanks!

sandeep-krishnamurthy:
@aaronmarkham - Can you please help?

ThomasDelteil:
@zchrissirhcz a few module are indeed not in the API documentation as top-level packages as they should, for example the `mxnet.viz` and `mxnet.test_utils`. This is something we need to be fixed IMO, @aaronmarkham ?
 @indhub can you remove the question tag and add a [Bug] tag?
We should also try to handle better the shorthand versions of the modules: e.g `sym`->`symbol` `viz->vizualisation` `nd`->`ndarray`

aaronmarkham:
I'm not sure why this is the case. My guess is that it is probably doxygen config. I can tackle it first thing next week. Can you assign the issue to me?

sad-:
hi @aaronmarkham, what's the status of this?  any updates?

Issue #12318, opened 2018-08-23
Sphinx is unable to access some MXNet ONNX module functions


aaronmarkham:
This causes Sphinx to fail processing the API docs when shorthand references are used. I used a workaround in #12317 to just reference the functions the long way. 

This means the reference is:
`mxnet.contrib.onnx.onnx2mx.import_model.import_model`
When it could be:
`mxnet.contrib.onnx.import_model`

![2018-08-23_15-08-55](https://user-images.githubusercontent.com/5974205/44555096-657b4100-a6e8-11e8-8ea5-49cf3db31063.png)

But the shorthand route doesn't work.

Example:
```
ubuntu@ip-172-31-66-78:~/incubator-mxnet/docs$ python -c "import mxnet.contrib.onnx.onnx2mx"
ubuntu@ip-172-31-66-78:~/incubator-mxnet/docs$ python -c "import mxnet.contrib.onnx.onnx2mx.import_model"
ubuntu@ip-172-31-66-78:~/incubator-mxnet/docs$ python -c "import mxnet.contrib.onnx.import_model"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: No module named import_model
```
@Roshrini

ankkhedia:
@mxnet-label-bot [Doc, Website]

aaronmarkham:
Hi @marcoabreu - this is more of a Python issue on the engineering side than something that can be fixed from within the website or docs systems. I might even label it a bug.

ankkhedia:
@sandeep-krishnamurthy Could you please label above as [Python, Build, Bug] and remove [Doc, Website]

vdantu:
@aaronmarkham : What needs to be done here for the docs?

PR #12455, opened 2018-09-04
Add Python API docs for test_utils and visualization

aaronmarkham:
## Description ##
Fixes https://github.com/apache/incubator-mxnet/issues/11412 - by adding missing modules to the docs coverage.

There appear to be several modules that are not included in the Python docs. This PR specifically addresses those mentioned in #11412 - `test_utils` and `visualization`.

I put these in a "tools" folder. It seemed to make sense.

## Preview
http://34.201.8.176/versions/python_tools/api/python/index.html#tools

## Comments
LMK what other missing APIs should go in the tools folder. There seem to be a lot of things in https://github.com/apache/incubator-mxnet/tree/master/python/mxnet that aren't covered. If someone wants to make an issue and tag me in it, I can handle those in another PR. 

Something is wrong with the shorthand invocations. The way the docs pipeline is setup and the way Sphinx is reading things makes it so `mxnet.viz` doesn't work. Therefore only the long version `mxnet.visualiztion` is mentioned in these docs. This is similar to what I ran into with ONNX: #12318.


aaronmarkham:
@sandeep-krishnamurthy Could you provide some feedback on the other missing modules and whether `docs/api/python/tools` makes sense for the ones in this PR?

sandeep-krishnamurthy:
Thanks @aaronmarkham . This looks good to me and a very useful addition.
Apart from these utilities, im2rec and rec2idx are missing. But, it has bunch of other clean up to be done before making it part of API docs. Created the issue to track - https://github.com/apache/incubator-mxnet/issues/12501

@aaronmarkham - Can you please rebase this PR and we are good to go here.

PR #12456, opened 2018-09-04
[MXNET-910] Multithreading inference.

zheng-da:
## Description ##
MXNet Executor isn't thread-safe and thus the predictor is also not thread-safe. However, there is a use case that we want to parallel inference of the same model with multiple threads. In this case, we need to create multiple executors that share the same weight arrays. The current C predict API doesn't support this. As such, we add a new API to create multiple predictors for parallel inference and show the use of this new API in the example of image classification.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/src/c_api/c_predict_api.cc
+++ b/src/c_api/c_predict_api.cc
@@ -68,7 +68,7 @@ struct MXAPINDList {
 };
 
 void MXPredCreateNewExecutor(PredictorHandle pred_hnd) {
-	MXAPIPredictor *pred = (MXAPIPredictor *) pred_hnd;
+  MXAPIPredictor *pred = (MXAPIPredictor *) pred_hnd;
   auto sym = pred->sym;
   auto ctx = pred->ctx;
   auto key2arg = pred->key2arg;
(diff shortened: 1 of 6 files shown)


marcoabreu:
Shouldn't we hide this abstraction from our users? The engine should be smart enough to determine when to multithread - the APIs just have to support concurrent calls.

zheng-da:
This PR is mainly for demo.
For your comments, I think it's necessary to expose the number of threads for parallel inference. The computation in an executor is parallelized itself. This gives another option for parallelization along with parallelism in an executor. It's hard for a system to figure. Users should try and decide.

src/c_api/c_predict_api.cc
```diff
@@ -232,24 +223,117 @@ int MXPredCreatePartialOut(const char* symbol_json_str,
...
+    std::unique_ptr<MXAPIPredictor> ret(new MXAPIPredictor());
+    ret->sym = sym;
+    ret->ctx = ctx;
+    ret->key2arg = key2arg;
+    ret->arg_arrays = arg_arrays;
+    ret->aux_arrays = aux_arrays;
     ret->out_shapes = out_shapes;
-    ret->out_arrays = ret->exec->outputs();
+
+    if (!lazy) {
```
eric-haibin-lin: Why is this made lazy?

zheng-da: The fundamental problem here is that if we create multiple executors in the same thread (e.g., in the main thread), these executors will share the same temporary resources, which leads to race condition when these executors are used in different threads. To fix this problem, here we avoid creating executors when we create predictors in the main thread. The executors are actually created when the predictor is used in the worker thread for the first time. As long as the executor is always used in this worker thread, there won't be race condition.

JohnLee168: > The fundamental problem here is that if we create multiple executors in the same thread (e.g., in the main thread), these executors will share the same temporary resources, which leads to race condition when these executors are used in different threads. To fix this problem, here we avoid creating executors when we create predictors in the main thread. The executors are actually created when the predictor is used in the worker thread for the first time. As long as the executor is always used in this worker thread, there won't be race condition.

If I use 10 different PredictorHandles created by 
MXPredCreate() in the main thread and for each PredictorHandle calling MXPredSetInput(), MXPredForward() and MXPredGetOutput() functions to inference in 10 threads, Is it safe?
Thread can get one current available PredictorHandle to inference, therefore, one certain thread may get different PredictorHandle to inference. Is this safe?

zheng-da: you can give it a try. i'm not sure.

JohnLee168: > you can give it a try. i'm not sure.

I got dead lock in MXPredSetInput(), MXPredForward() or MXPredGetOutput(), It seems it doesn't support.

Issue #11368, opened 2018-06-22
Enable NCCL Flag in MXNet CI


mbaijal:
## Description
There are currently no tests running in CI with the NCCL flag enabled. 
At a minimum we should, add a docker with NCCL libs, add a build with the NCCL flag and run this test `tests/python/gpu test_nccl.py`

PR #12477, opened 2018-09-07
Added build with nccl

lebeg:
## Description ##

Resolves https://github.com/apache/incubator-mxnet/issues/11368

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/ci/docker/runtime_functions.sh
+++ b/ci/docker/runtime_functions.sh
@@ -305,6 +305,7 @@ build_centos7_gpu() {
         USE_CUDNN=1                               \
         USE_DIST_KVSTORE=1                        \
         CUDA_ARCH="$CI_CUDA_COMPUTE_CAPABILITIES" \
+        USE_NCCL=1                                \
         -j$(nproc)
 }
 
--- a/ci/docker/runtime_functions.sh
+++ b/ci/docker/runtime_functions.sh
@@ -536,6 +536,7 @@ build_ubuntu_gpu_tensorrt() {
         USE_GPERFTOOLS=0                                     \
         ONNX_NAMESPACE=onnx                                  \
         CUDA_ARCH="-gencode arch=compute_70,code=compute_70" \
+        USE_NCCL=1                                           \
         -j$(nproc)
 }
 
@@ -554,6 +555,7 @@ build_ubuntu_gpu_mkldnn() {
         USE_CUDA_PATH=/usr/local/cuda             \
         USE_CUDNN=1                               \
         CUDA_ARCH="$CI_CUDA_COMPUTE_CAPABILITIES" \
+        USE_NCCL=1                                \
         -j$(nproc)
 }
 
@@ -571,6 +573,7 @@ build_ubuntu_gpu_mkldnn_nocudnn() {
         USE_CUDA_PATH=/usr/local/cuda             \
         USE_CUDNN=0                               \
         CUDA_ARCH="$CI_CUDA_COMPUTE_CAPABILITIES" \
+        USE_NCCL=1                                \
         -j$(nproc)
 }
 
@@ -588,6 +591,7 @@ build_ubuntu_gpu_cuda91_cudnn7() {
         USE_CPP_PACKAGE=1                         \
         USE_DIST_KVSTORE=1                        \
         CUDA_ARCH="$CI_CUDA_COMPUTE_CAPABILITIES" \
+        USE_NCCL=1                                \
         -j$(nproc)
 }
 
--- a/src/operator/tensor/elemwise_unary_op_basic.cc
+++ b/src/operator/tensor/elemwise_unary_op_basic.cc
@@ -29,17 +29,19 @@
 namespace mxnet {
 namespace op {
 
+namespace {
+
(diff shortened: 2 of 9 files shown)


Jenkinsfile
```diff
@@ -500,6 +518,20 @@ core_logic: {
...
+    'Python3: GPU NCCL': {
+      node(NODE_LINUX_GPU) {
+        ws('workspace/ut-python3-gpu') {
+          try {
+            utils.init_git()
+            utils.unpack_lib('gpu', mx_lib)
+            python3_gpu_nccl_ut('ubuntu_gpu')
+            utils.publish_test_coverage()
+          } finally {
+            utils.collect_test_results_unix('nosetests_gpu.xml', 'nosetests_python3_gpu.xml')
```
marcoabreu: Please adjust the second filename

ci/docker/runtime_functions.sh
```diff
@@ -693,6 +709,17 @@ unittest_ubuntu_tensorrt_gpu() {
...
 }
 
+unittest_ubuntu_gpu_nccl() {
+    set -ex
+
+    export PYTHONPATH=./python/
+    export MXNET_STORAGE_FALLBACK_LOG_VERBOSE=0
+    export LD_LIBRARY_PATH=/work/mxnet/lib:$LD_LIBRARY_PATH
+
+    nosetests-3.4 $NOSE_COVERAGE_ARGUMENTS --with-xunit --xunit-file nosetests_trt_gpu.xml --verbose \
```
marcoabreu: I think the Tensor RT reference is a copy and paste artifact. See above.

kalyc:
Thanks for your contribution @lebeg 
@mxnet-label-bot[pr-awaiting-testing, pr-work-in-progress]

lebeg:
@marcoabreu that could be an option, but for example `USE_DIST_KVSTORE` is not enabled as well, but as I understood has similar functionality. Enabling it for all builds would mean it's rather a requirement for the library. Maybe an option `USE_NCCL_IF_AVAILABLE` can be introduced, but information needs to be propagated to the API whether MXNet was compiled with NCCL to skip the test if needed. For now it's a separate build and a separate test - with all optimizations that we have on the CI these are almost trivial stages.

lebeg:
Another problem is the version of NCCL - the 1<sup>st</sup> version is [available on github](https://github.com/NVIDIA/nccl/releases) and we can integrate it easily to the CI. But the latest release links to CUDA 8.0 wich requires a separate docker build, since the usual builds are against 9.1 currently.
The 2<sup>nd</sup> version is available only behind a registration: https://developer.nvidia.com/nccl

lebeg:
@yajiedesign yes, I think so. Actually, we could assert that even in the test. The node label `NODE_LINUX_GPU` would be a g3 instance with 2 cards.

lebeg:
Actually, I just found out that libnccl2 is part of the nvidia cuda docker images, so no big deal of enabling it for all gpu builds.

vandanavk:
@lebeg Is this PR ready for review again once the CI build is fixed?

lebeg:
Would need to merge https://github.com/dmlc/ps-lite/pull/140 first to get the make builds running.

lebeg:
PR not in WIP anymore and can be reviewed

ankkhedia:
@lebeg It seems that CI is still failing on the newer commits. Could you please take a look?
Also could you please resolve branch conflicts?

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -30,11 +30,11 @@ namespace mxnet {
 namespace op {
 
 // infer storage function for _identity_with_attr_like_rhs op
-static bool IdentityAttrLikeRhsStorageType(const nnvm::NodeAttrs& attrs,
-                                           const int dev_mask,
-                                           DispatchMode* dispatch_mode,
-                                           std::vector<int> *in_attrs,
-                                           std::vector<int> *out_attrs) {
+static bool InferStorage(const nnvm::NodeAttrs &attrs,
```
eric-haibin-lin: I'm afraid mxnet::op::InferStorage is too generic while its implementation is only for `_identity_with_attr_like_rhs` op. This will cause confusion

lebeg: It's used only in 1 place and the problem with it is that it's debug symbol is too long which creates a warning which is treated as error. This function is not exposed outside of the translation unit as well which lowers the risk of confusion. I would be happy to rename it to anything more useful if it's short enough to silence the warning

lebeg: I changed it to `IdentLikeRhsStorageType` and moved it to anonymous namespace instead of static per rule [SF.22 in CppCoreGuideLines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rs-unnamed2)

vandanavk:
@lebeg is this PR ready for review again?

lebeg:
I don't have time for this right now, I'll close it and reopen once it will be ready.

PR #12482, opened 2018-09-07
fix website render issue on &lt; and &gt;

### LINKED ISSUE #11438: docs using xml in code blocks will improperly render
As this has bitten me several times now, so I'm writing this up as an issue. It is likely a bug in `recommonmark`.

If you use entities that can be interpreted as tags, like

```
#Three opening backticks
'''
<dependency>
...

This will render improperly.

If you use

```xml
<dependency>
...

This will render improperly.

If you use

```html
<dependency>
...

This will render properly.
```

There may be other workarounds, but for now I'm using ```html
This has been coming up in documentation for scala.

### LINKED ISSUE #12168: Error in Operator implementation guide
## Description
Error in the operator implementation guide:
https://mxnet.incubator.apache.org/faq/add_op_in_backend.html

It's correct in the markdown file: https://github.com/apache/incubator-mxnet/blob/master/docs/faq/add_op_in_backend.md

## Error Message:
"Note that forward and backward functions are registered with attribute key FCompute, rather than FCompute."

aaronmarkham:
## Description ##
Fixes #12168 
Fixes #11438

## Comments
I didn't [hear back yet](https://github.com/apache/incubator-mxnet/issues/12168#issuecomment-418555387) from @piiswrong  about the purpose of this particular line. However, removing it fixes two documented render issues, and I don't see any negative effects after removing the line.

## Preview
http://34.201.8.176/versions/tag_render/

### DIFF
--- a/docs/build_version_doc/AddVersion.py
+++ b/docs/build_version_doc/AddVersion.py
@@ -74,7 +74,8 @@ if __name__ == '__main__':
                     version_tag_mobile.extract()
                 navbar.append(version_str)
                 navbar_mobile.append(version_str_mobile)
-                outstr = str(content).replace('&lt;', '<').replace('&gt;', '>')
+                # The following causes rendering errors in code blocks; refer to #12168
+                #outstr = str(content).replace('&lt;', '<').replace('&gt;', '>')
             # Fix link
             if args.current_version == tag_list[0]:
                 print("Fixing " + os.path.join(path, name))


sandeep-krishnamurthy:
@piiswrong - You authored it earlier, can you please suggest if this will not break other expected behavior?

sandeep-krishnamurthy:
Merging this PR, as it fixes couple of issues and I couldn't hear back from Eric. Verified on the rendered site, did not find any issues.

Issue #11359, opened 2018-06-22
Flaky test test_io:test_ImageRecordIter_seed_augmentation


marcoabreu:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/PR-11340/8/pipeline
```
test_io.test_ImageRecordIter_seed_augmentation ... [21:50:02] src/io/iter_image_recordio_2.cc:170: ImageRecordIOParser2: data/cifar/train.rec, use 4 threads for decoding..
[21:50:02] src/io/iter_image_recordio_2.cc:228: Load mean image from data/cifar/cifar10_mean.bin
[21:50:02] src/io/iter_image_recordio_2.cc:246: Load mean image from data/cifar/cifar10_mean.bin completed
[21:50:02] src/io/iter_image_recordio_2.cc:170: ImageRecordIOParser2: data/cifar/train.rec, use 4 threads for decoding..
[21:50:02] src/io/iter_image_recordio_2.cc:228: Load mean image from data/cifar/cifar10_mean.bin
[21:50:02] src/io/iter_image_recordio_2.cc:246: Load mean image from data/cifar/cifar10_mean.bin completed
FAIL


======================================================================
FAIL: test_io.test_ImageRecordIter_seed_augmentation
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/work/mxnet/tests/python/unittest/test_io.py", line 381, in test_ImageRecordIter_seed_augmentation
    assert(np.array_equal(data,data2))
AssertionError
```
Probably due to missing tolerances

marcoabreu:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/master/1042/pipeline

marcoabreu:
14 out of 225 runs failed -> 6% failure rate

wenyangchu:
Hi @marcoabreu, can it be that the multithreading of this iterator does not guarantee the order of the patches?

marcoabreu:
I would have to check, but that could be a good start for further investigation.

larroy:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/PR-11382/6/pipeline

szha:
[...]

PR #12485, opened 2018-09-08
test_ImageRecordIter_seed_augmentation flaky test fix

perdasilva:
## Description ##
Fixes flakiness with ImageRecordIter augmentation seed tests (#11359) by moving the seed_aug parameter and RNG seeding out of the default augmenter. We now seed the RNG before each image augmentation (if seed_aug parameter is set). This guarantees reproducibility while still enabling multithreaded augmentation (thus fixing the flakiness).
Additionally, it makes the test more robust by testing the whole iterator and not just the first image.

### Essentials ###

## Comments ##

### DIFF
--- a/src/io/image_iter_common.h
+++ b/src/io/image_iter_common.h
@@ -131,6 +131,8 @@ struct ImageRecParserParam : public dmlc::Parameter<ImageRecParserParam> {
   size_t shuffle_chunk_size;
   /*! \brief the seed for chunk shuffling*/
   int shuffle_chunk_seed;
+  /*! \brief random seed for augmentations */
+  dmlc::optional<int> seed_aug;
 
   // declare parameters
   DMLC_DECLARE_PARAMETER(ImageRecParserParam) {
@@ -165,6 +167,8 @@ struct ImageRecParserParam : public dmlc::Parameter<ImageRecParserParam> {
         .describe("The data shuffle buffer size in MB. Only valid if shuffle is true.");
     DMLC_DECLARE_FIELD(shuffle_chunk_seed).set_default(0)
         .describe("The random seed for shuffling");
+    DMLC_DECLARE_FIELD(seed_aug).set_default(dmlc::optional<int>())
+        .describe("Random seed for augmentations.");
   }
 };
 
--- a/src/io/iter_image_recordio_2.cc
+++ b/src/io/iter_image_recordio_2.cc
@@ -519,6 +519,13 @@ inline size_t ImageRecordIOParser2<DType>::ParseChunk(DType* data_dptr, real_t* 
       cv::Mat res;
       rec.Load(blob.dptr, blob.size);
       cv::Mat buf(1, rec.content_size, CV_8U, rec.content);
+
+      // If augmentation seed is supplied
+      // Re-seed RNG to guarantee reproducible results
+      if (param_.seed_aug.has_value()) {
+        prnds_[tid]->seed(idx + param_.seed_aug.value() + kRandMagic);
+      }
+
       switch (param_.data_shape[0]) {
        case 1:
 #if MXNET_USE_LIBJPEG_TURBO
--- a/tests/python/unittest/test_io.py
+++ b/tests/python/unittest/test_io.py
@@ -32,6 +32,10 @@ except ImportError:
 import sys
 from common import assertRaises
 import unittest
+try:
+    from itertools import izip_longest as zip_longest
+except:
+    from itertools import zip_longest
 
 
 def test_MNISTIter():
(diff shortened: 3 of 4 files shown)


kalyc:
Thanks for your contribution @perdasilva 

sandeep-krishnamurthy:
@perdasilva - Did you run the tests multiple times to verify it is not flaky anymore?

perdasilva:
@sandeep-krishnamurthy sorry for the delay in the response. I've been absent lately because life got in the way. The issue was the multi-threading, I've just done a rebase from master and it seems the number of threads has been reduced to 1 (which fixes the flakyness). I need to look at the history of the file to see what's changed since I was last looking at this problem.

perdasilva:
I have sketched out a fix to the flakiness here, although its a bit of a hack at the moment. 
The reason for the flakiness has to do with the parallel processing and the random number generators used for the augmenters.

One random number generator is created per preprocessing thread (https://github.com/apache/incubator-mxnet/blob/master/src/io/iter_image_recordio_2.cc#L162). While the image records are retrieved in the same order (https://github.com/apache/incubator-mxnet/blob/master/src/io/iter_image_recordio_2.cc#L508) and always stored in the same index in the output array (https://github.com/apache/incubator-mxnet/blob/master/src/io/iter_image_recordio_2.cc#L567) - they are not guaranteed to be processed by the same thread every time. This matter because the default augmenter seeds the thread's random number generator (https://github.com/apache/incubator-mxnet/blob/master/src/io/image_aug_default.cc#L255) the first time it gets called (https://github.com/apache/incubator-mxnet/blob/master/src/io/iter_image_recordio_2.cc#L563). 

This means that for different runs, the same record can be processed in a different iteration of the usage of the random number generator by the augmenters, thus leading to different random numbers being generated, which lead to different changes to the image -> flakyness.

To remedy the issue, I've changed to code such that the augmenter seed is a parameter to the image parser. I then use this seed and the image record index to seed the random number generator (outside of the default augmenter) before any changes are made to the image. Since the records are always retrieved in the same order, the same record will always have the same generator seed, independent of the number of threads used -> same random numbers being generated -> reproducible. 

It's a bit of a hack, so I would appreciate some input from the code owner (@anirudh2290 )  and the community.

sandeep-krishnamurthy:
@haojin2 - Can you provide some inputs here? Thanks.

Roshrini:
@haojin2 @apeforest Can you please review this PR? Thanks

ankkhedia:
@haojin2 @apeforest ping for reviewing the PR!

src/io/iter_image_recordio_2.cc
```diff
@@ -518,6 +518,17 @@ inline unsigned ImageRecordIOParser2<DType>::ParseChunk(DType* data_dptr, real_t
...
       cv::Mat buf(1, rec.content_size, CV_8U, rec.content);
+
+      if (idx % 1000 == 0) {
+        if (param_.seed_aug.has_value()) {
+          LOG(INFO) << "aug seed: " << param_.seed_aug.value();
+        }
+        LOG(INFO) << "tid: " << tid << " idx: " << idx << " index: " << rec.image_index();
+      }
+      if (param_.seed_aug.has_value()) {
+        prnds_[tid]->seed(idx + param_.seed_aug.value());
```
lebeg: I do not understand why seeds need to be reset all the time. The logic is a bit obscure - the random generators are in the augmentor and get passed to the iterator. 
Anyway, what you could try to do (if not refactoring) is to seed all generators in `DefaultImageAugmenter::Init` [here](https://github.com/apache/incubator-mxnet/blob/master/src/io/image_aug_default.cc#L215) or in a separate method (since the random generator is not passed in to `Init`). I guess the interface `ImageAugmenter` would require a separate `seed` method then that could be called sequentially (instead of openmp parallelized).

perdasilva: There are two requirements here: that setting the seed will yield reproducible results and that parallelization should be used to augment the images. We need to reset the seed for each image because there is no guarantee that the same image will be processed by the same thread. Or that even that it will be the ith image processed by that thread across every run (or even different hardware - the code figures out the number of threads to use for processing). Therefore, resetting the random number generator at the start of processing an image it the only way (at least that I could think of) to guarantee that, in the case of setting a fixed seed, the same random distortions will be applied to the same image, in a multi-threaded environment, independent of the hardware being used. I hope this is clear. It's not the easiest topic to discuss in written form lol.

src/io/image_iter_common.h
```diff
@@ -165,6 +167,8 @@ struct ImageRecParserParam : public dmlc::Parameter<ImageRecParserParam> {
         .describe("The data shuffle buffer size in MB. Only valid if shuffle is true.");
     DMLC_DECLARE_FIELD(shuffle_chunk_seed).set_default(0)
         .describe("The random seed for shuffling");
+    DMLC_DECLARE_FIELD(seed_aug).set_default(dmlc::optional<int>())
```
anirudhacharya: maybe I am missing something here, but the PR says it fixes a flaky test. 

Are we adding this new field just to fix the flakiness of a particular test? Can you please share the root cause of the flakiness based on your investigations?

tests/python/unittest/test_io.py
```diff
@@ -427,7 +427,7 @@ def check_CSVIter_synthetic(dtype='float32'):
     for dtype in ['int32', 'int64', 'float32']:
         check_CSVIter_synthetic(dtype=dtype)
 
-@unittest.skip("Flaky test: https://github.com/apache/incubator-mxnet/issues/11359")
+# @unittest.skip("Flaky test: https://github.com/apache/incubator-mxnet/issues/11359")
```
anirudhacharya: please remove rather than comment.

kalyc:
@perdasilva thanks for your contributions. We are looking forward to merging your PR. Could you please address comments above to proceed further?

src/io/image_iter_common.h
perdasilva: Fair call. However, the problem, I think, is architectural. 
The seed_aug parameter is passed in to the default image augmenter (https://github.com/apache/incubator-mxnet/blob/master/src/io/image_aug_default.cc#L101). But, the actual random number generator is external, but seeded internally (https://github.com/apache/incubator-mxnet/blob/master/src/io/image_aug_default.cc#L254).

As I've explained above, I can't see how to meet the requirements of processing the images in parallel and ensuring that the output is reproducible (by setting the augmentation seed). 

My fix for the flakiness (without violating the aforementioned requirements), is to ensure that the RNG used to process the image is seeded before processing the image within that thread it is processed in.

To be clear, none of the changes made were to facilitate running the tests, but to actually correct the underlying problem. I agree that the PR is a little rough around the edges atm, but what I'm trying to do is reach out to the community to understand what would be the best way to  actually fix the underlying problem.

tests/python/unittest/test_io.py
perdasilva: Done. Although again, I need a review of my current approach to see if its the right one. If I get validation from that, I can then prepare the PR for proper review...
Many of nits I'll fix before removing [WIP]

stu1130:
@perdasilva is it ready for review?

perdasilva:
@stu1130 I will polish it tomorrow and submit it for review. I'm sorry for the delay.

stu1130:

vandanavk:
@anirudhacharya @lebeg could you review this again? the feedback has been addressed

tests/python/unittest/test_io.py
```diff
@@ -427,13 +434,43 @@ def check_CSVIter_synthetic(dtype='float32'):
...
 
-@unittest.skip("Flaky test: https://github.com/apache/incubator-mxnet/issues/11359")
 def test_ImageRecordIter_seed_augmentation():
     get_cifar10()
     seed_aug = 3
 
+    def assert_dataiter_equals(dataiter1, dataiter2):
+        for batch1, batch2 in zip_longest(dataiter1, dataiter2):
+            # ensure iterators are of same length
+            assert(batch1 and batch2)
```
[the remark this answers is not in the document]
perdasilva: This is a function for asserting that two data iterators are producing the same output. The data iterator iterates over batches (lists of arrays). So, basically it runs through the iterators to get the batches, then iterates through the batches and ensures that they are equal.

tests/python/unittest/test_io.py
```diff
@@ -427,13 +434,43 @@ def check_CSVIter_synthetic(dtype='float32'):
...
+            for i in range(0, len(batch1.data)):
+                data1 = batch1.data[i].asnumpy().astype(np.uint8)
+                data2 = batch2.data[i].asnumpy().astype(np.uint8)
+                assert(np.array_equal(data1, data2))
+
+    def assert_dataiter_not_equals(dataiter1, dataiter2):
+        for batch1, batch2 in zip_longest(dataiter1, dataiter2):
+
+            # try to ensure iterators are of same length
+            assert(batch1 and batch2)
```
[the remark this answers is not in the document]
perdasilva: The problem is that (as far as I could tell), there was no easy way of getting the number of batches in an iterator. It doesn't have a size or length method. So, I'm using the zip_longest (https://docs.python.org/3.0/library/itertools.html#itertools.zip_longest) to iterate through the iterators together. If they are don't have the same length, a fill value (None by default) will be returned in place of an item (or batch in this case). Therefore, if one of batch1 or batch2 is none, the assertion will be triggered.

tests/python/unittest/test_io.py
```diff
@@ -32,6 +33,12 @@
 import sys
 from common import assertRaises
 import unittest
+import sys
+
+if sys.version_info >= (3,0):
+    from itertools import zip_longest
+else:
+    from itertools import izip_longest as zip_longest
```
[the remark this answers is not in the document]
perdasilva: I've updated it with your suggestion. Much nicer.

tests/python/unittest/test_io.py
```diff
@@ -489,31 +524,27 @@ def test_ImageRecordIter_seed_augmentation():
...
         seed_aug=seed_aug+1)
-    batch = dataiter.next()
-    data2 = batch.data[0].asnumpy().astype(np.uint8)
-    assert(not np.array_equal(data,data2))
+
+    assert_dataiter_not_equals(dataiter1, dataiter2)
 
     # check whether seed_aug changes the iterator behavior
-    dataiter = mx.io.ImageRecordIter(
+    dataiter1 = mx.io.ImageRecordIter(
```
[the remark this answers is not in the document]
perdasilva: This test was already here when I got here =)  I think it was meant as a sanity check to ensure that the seed aug parameter is not influencing the iterator (other than in the case where the iterator has augmentations applied to its images).

perdasilva:
@anirudhacharya  all good now?

anirudhacharya:
LGTM. 
@anirudh2290 @apeforest for review/merge

apeforest:
@stu1130 Could you please review this change?

perdasilva:
@stu1130 - not 10000x. I will do that tomorrow and confirm ^^

perdasilva:
@stu1130 - ok each test takes around 10 seconds - might take a tad bit longer than expected...

stu1130:
@anirudh2290 could you take a look at this and merge it if it looks good

perdasilva:
@anirudh2290 any chance you could merge this? =)

Roshrini:
@marcoabreu @sandeep-krishnamurthy Can someone merge this PR?
@mxnet-label-bot Update [pr-awaiting-merge]

PR #12489, opened 2018-09-08
[WIP][MXNET-918] Random api

mdespriee:
## Description ##
Introduce Symbol.random and NDArray.random modules in scala API

### Essentials ###

### Changes ###
- Symbol.random api generated by macros
- NDArray.random api generated by macros

## Comments ##
- ~~Is there any JIRA mentioning this feature ?~~ 
- ~~Only implemented for Symbol. Adding the same for NDArray seems quite straightforward. Tell me if I should to it in the same PR or in a separate~~
- ~~I *moved* functions from Symbol.api to Symbol.random. Is it OK ?~~
- The produced API is not strictly similar to the one in python: 
   * in python, each function, eg random.normal, has 2 overrides, one scalar, one symbolic.
   * the generated scala api has different namings: Symbol.random.sample_normal for symbol inputs, and Symbol.random.random_normal for scalars. The arguments naming is also not fully consistent (mu & sigma, vs loc & scale). 
   => What are your recommendations on this ?
- The unit-test is poor. What should be tested for generated code like this ?

### DIFF
--- a/scala-package/macros/src/main/scala/org/apache/mxnet/NDArrayMacro.scala
+++ b/scala-package/macros/src/main/scala/org/apache/mxnet/NDArrayMacro.scala
@@ -25,9 +25,11 @@ import scala.reflect.macros.blackbox
 private[mxnet] class AddNDArrayFunctions(isContrib: Boolean) extends StaticAnnotation {
   private[mxnet] def macroTransform(annottees: Any*) = macro NDArrayMacro.addDefs
 }
+
 private[mxnet] class AddNDArrayAPIs(isContrib: Boolean) extends StaticAnnotation {
   private[mxnet] def macroTransform(annottees: Any*) = macro NDArrayMacro.typeSafeAPIDefs
 }
+
 private[mxnet] class AddNDArrayRandomAPIs(isContrib: Boolean) extends StaticAnnotation {
   private[mxnet] def macroTransform(annottees: Any*) = macro NDArrayMacro.typeSafeRandomAPIDefs
 }
@@ -41,15 +43,20 @@ private[mxnet] object NDArrayMacro extends GeneratorBase {
   def addDefs(c: blackbox.Context)(annottees: c.Expr[Any]*) = {
     impl(c)(annottees: _*)
   }
+
   def typeSafeAPIDefs(c: blackbox.Context)(annottees: c.Expr[Any]*) = {
-    typeSafeAPIImpl(c)(annottees: _*)
+    typedAPIImpl(c)(annottees: _*)
   }
+
   def typeSafeRandomAPIDefs(c: blackbox.Context)(annottees: c.Expr[Any]*) = {
-    typeSafeRandomAPIImpl(c)(annottees: _*)
+    typedRandomAPIImpl(c)(annottees: _*)
   }
+
   // scalastyle:off havetype
 
-  private val ndarrayFunctions: List[NDArrayFunction] = buildFunctionList(false)
+  private val ndarrayFunctions = buildFunctionList(false)
+
+  private val rndFunctions = buildRandomFunctionList(false)
 
   /**
     * Implementation for fixed input API structure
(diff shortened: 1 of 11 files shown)


lanking520:
@mdespriee Firstly for unit test, there are tests covers in Python on these operators. In that case we decide not to test them on Scala end.

```
the generated scala api has different namings: Symbol.random.sample_normal for symbol inputs, and Symbol.random.random_normal for scalars. The arguments naming is also not fully consistent (mu & sigma, vs loc & scale).
```
For this question, let me consult with people working on Python. 

Please also consider adding NDArray with these operators (in NDArrayAPI)

lanking520:
Hi @mdespriee , I have created a JIRA ticket for you: https://issues.apache.org/jira/browse/MXNET-918. You can put it on the title so people know what issue it is related.

scala-package/macros/src/main/scala/org/apache/mxnet/NDArrayMacro.scala
```diff
@@ -91,69 +114,74 @@ private[mxnet] object NDArrayMacro {
     val newNDArrayFunctions = {
       if (isContrib) ndarrayFunctions.filter(
         func => func.name.startsWith("_contrib_") || !func.name.startsWith("_"))
-      else ndarrayFunctions.filterNot(_.name.startsWith("_"))
+      else ndarrayFunctions.filterNot(f => f.name.startsWith("_") &&
+        !f.name.startsWith("sample_") && !f.name.startsWith("random_"))
```
lanking520: You should not try to avoid `sample_` and `random_` from generation:
[This](http://mxnet.incubator.apache.org/api/python/ndarray/ndarray.html#mxnet.ndarray.random_uniform) called by operator `random_uniform`.
and [This](http://mxnet.incubator.apache.org/api/python/ndarray/random.html#mxnet.ndarray.random.uniform) called by operator `_random_uniform`

kalyc:
Thanks for your contribution @mdespriee 

mdespriee:
@lanking520. Underscore removed. We now have this kind of signatures:
```
def random_normal (loc : Option[org.apache.mxnet.Base.MXFloat] = None, scale : Option[org.apache.mxnet.Base.MXFloat] = None, shape : Option[org.apache.mxnet.Shape] = None, ctx : Option[String] = None, dtype : Option[String] = None, name : String = null, attr : Map[String, String] = null) : org.apache.mxnet.Symbol
```
and
```
def sample_normal (mu : Option[org.apache.mxnet.Symbol] = None, shape : Option[org.apache.mxnet.Shape] = None, dtype : Option[String] = None, sigma : Option[org.apache.mxnet.Symbol] = None, name : String = null, attr : Map[String, String] = null) : org.apache.mxnet.Symbol
```

I don't think we can go further. If remove both "_random_" and "_sample_" prefixes, we'll have collisions that the scala compiler can't resolve since all parameters have default values.
(`multiple overloaded alternatives of method uniform define default arguments`)

lanking520:
@mdespriee Firstly I think we have a collide operators. `_sample` and `_random` looks like the same with `random` and `sample`. So it looks like we shouldn't need to generate a new set of API in order to get this work. Instead, we should try to use Macros to do something similar to [python](https://github.com/apache/incubator-mxnet/blob/225f71f744ac5e7bd29868b6d3ba0e4fe2527c43/python/mxnet/symbol/random.py). Create a helper function that can choose between `sample` and `random`. Change the implementation a little bit more for Random Module:

1. Set the args type as `Any` for Random API in order to take in `NDArray/Float` or `Symbol/Float`
2. Do a instance check such as `asInstanceOf[NDArray]` to see if it is an object or numeric value
3. Do `random_` or `sample_` in the function

vandanavk:
Thanks for your contribution @mdespriee. Please have a look at @lanking520's feedback.

vrakesh:
@mdespriee Have you had a chance to look at @lanking520 's feedback? Requesting an update from you on this.

mdespriee:
Hi,
I've been busy the 2 past weeks (spark summit...) I noticed @lanking520
about that.
I'll probably be able to work on it this week. This PR is not dead  :-)
Mathieu


Le mar. 9 oct. 2018 à 09:01, Rakesh Vasudevan <notifications@github.com> a
écrit :

lanking520:
@mdespriee Good to see you back!

mdespriee:
> @mdespriee Firstly I think we have a collide operators. `_sample` and `_random` looks like the same with `random` and `sample`. So it looks like we shouldn't need to generate a new set of API in order to get this work. Instead, we should try to use Macros to do something similar to [python](https://github.com/apache/incubator-mxnet/blob/225f71f744ac5e7bd29868b6d3ba0e4fe2527c43/python/mxnet/symbol/random.py). Create a helper function that can choose between `sample` and `random`. Change the implementation a little bit more for Random Module:
> 
> 1. Set the args type as `Any` for Random API in order to take in `NDArray/Float` or `Symbol/Float`
> 2. Do a instance check such as `asInstanceOf[NDArray]` to see if it is an object or numeric value
> 3. Do `random_` or `sample_` in the function

I've noticed the following inconsistencies between `sample` and `random`:
- multinomial only exists in `sample`-> This is catchable at runtime in case of mismatch
- normal has different params in both worlds : loc&scale vs mu&sigma. -> How do we manage this ?

lanking520:
@mdespriee Firstly thanks for coming back to maintain this PR! 

I see your point, is there a mapping can be applied to these operators? If yes as they are just the difference like that, we can apply something in Macros end to make it change. 

Apart from that, I think you can extend https://github.com/apache/incubator-mxnet/blob/master/scala-package/core/src/main/scala/org/apache/mxnet/Random.scala this class to support Random with your Macros design

andrewfayres:
Can we add a gist of the generated code?

Also, is there a reason that we're adding a new NDArray.random as an api instead of extending NDArray.api (same for Symbol)?

lanking520:
@andrewfayres here is what we are trying to do: https://mxnet.incubator.apache.org/api/python/ndarray/random.html. In NDArray API we already have some random generation function, but a Random namespace would be great since it is compatible with both NDArray/Float.

lanking520:
@mdespriee thanks for your effort to refactor the code. The issue shown above is the mismatch on the parameter. It can be caused by `function name` do not have this parameter, or you put a param in the wrong type. I would recommend you to split the refactoring portion and Random generation portion into two PRs. The good thing: 

1. you can focus on each of them and make the refactoring and Random part even better. 

2. Reduce the change on your code, it's easier for you to find where the problem is and easy for us to help you find where you can make a mistake.

I know this may introduce more work to you but I think this could be the best solution.

mdespriee:
Following @lanking520 comment, I split the work in 2 PRs:
- https://github.com/apache/incubator-mxnet/pull/13038 (refactoring of code gen)
- https://github.com/apache/incubator-mxnet/pull/13039 (Random module itself)

PR #12517, opened 2018-09-11
Updated nnvm with fixed tvm submodule

lebeg:
## Description ##

Follow up on missing tvm commit in nnvm repository: https://github.com/dmlc/nnvm/pull/525


szha:
This is not a proper fix as you're changing the nnvm package content.

lebeg:
@szha you're right, thank you for the fix: https://github.com/apache/incubator-mxnet/commit/ea0aec432f30f0a067ee695b946173d4e3940e4a

PR #12530, opened 2018-09-12
Implement mkldnn convolution fusion and quantization.

ZhennanQin:
Implement mkldnn convolution fusion and quantization.
## Description ##
This PR is the implementation followed this [proposal](https://cwiki.apache.org/confluence/display/MXNET/MXNet+Graph+Optimization+and+Quantization+based+on+subgraph+and+MKL-DNN)
@zheng-da,  @azai91, @TaoLv, @pengzhao-intel @reminisce 

### Essentials ###

### Changes ###
- [x] Feature2, tests, (and when applicable, API doc)

## Comments ##


zheng-da:
Could you please split operator fusion and quantization into two PRs? It doesn't seem that these two works have to be in the same PR.

ZhennanQin:
@zheng-da There's strong dependence between quantization and fusion because they all changed mkldnn_conv.cc, which is the key part of this PR. If I split them into 2 PRs, then this file needs to be reviewed twice, because quantization will change this file a lot(almost rewrite), which eventually increase the burden of code review.

pengzhao-intel:
@zheng-da  @reminisce 
Could you help take a review?  Thanks.

src/operator/nn/mkldnn/mkldnn_convolution.cc
```diff
@@ -257,32 +300,59 @@ void MKLDNNConvolutionForward(const nnvm::NodeAttrs& attrs, const OpContext &ctx
...
     }
   }
-  auto out_mem = CreateMKLDNNMem(out_data[conv::kOut], fwd.fwd_pd.dst_primitive_desc(),
-                                 req[conv::kOut]);
+  mkldnn_output_t out_mem;
+  if (param.mkldnn_param.with_sum) {
+    out_mem = mkldnn_output_t(
+        OutDataOp::Noop,
+        const_cast<mkldnn::memory *>(out_data[conv::kOut].GetMKLDNNDataReorder(
+            fwd->fwd_pd.dst_primitive_desc())));
```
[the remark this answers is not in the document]
ZhennanQin: Actually, out_data always has same memory description with fwd->fwd_pd, so we should use GetMKLDNNData instead of GetMKLDNNDataReorder. Then out_mem will always be the memory in out_data.

src/operator/subgraph/mkldnn/mkldnn_conv.cc
```diff
@@ -0,0 +1,670 @@
...
+      mkldnn_param.quantized ? outputs[kMin].data().dptr<float>() : nullptr;
+  float *out_max_ptr =
+      mkldnn_param.quantized ? outputs[kMax].data().dptr<float>() : nullptr;
+  CHECK_EQ(input_size, idx);
+  bool has_bias = mkldnn_param.with_bn || !conv_param.no_bias;
+  cached_data_ = inputs[in_data];
+  if (mkldnn_param.with_sum)
+    cached_output_ = inputs[in_sum];
+  else
+    cached_output_ = outputs[kOut];
```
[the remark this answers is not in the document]
ZhennanQin: Yes, they are changed every time, so we will reassign cached_input_  and cached_output_  on each forward. You can treat them as a normal variable, as it points to different NDArray according to mkldnn_param.

src/operator/subgraph/mkldnn/mkldnn_conv.cc
```diff
@@ -0,0 +1,678 @@
...
+                                          const OpContext &ctx,
+                                          MKLDNNConvForward *fwd,
+                                          const std::vector<NDArray> &inputs,
+                                          const std::vector<OpReqType> &req,
+                                          const std::vector<NDArray> &outputs) {
+  if (SupportMKLDNNConv(full_param.conv_param, inputs[0])) {
+    // MKLDNN_OPCHECK_INIT(false, outputs.size(), inputs, outputs);
+    MKLDNNConvolutionForwardFullFeature(full_param, ctx, fwd, inputs, req, outputs);
+    // MKLDNN_OPCHECK_RUN(ConvolutionCompute<cpu>, attrs, ctx, inputs, req,
+    // outputs);
```
zheng-da: why comment these two lines?

src/operator/subgraph/mkldnn/mkldnn_conv.cc
```diff
@@ -0,0 +1,678 @@
...
+
+ private:
+  bool initalized;
+  nnvm::Symbol subgraph_sym_;
+  MKLDNNConvFusionParam param;
+  std::shared_ptr<MKLDNNConvForward> fwd;
+  NDArray cached_weight_;
+  NDArray cached_bias_;
+  NDArray cached_data_;
+  NDArray cached_output_;
```
zheng-da: please remove `cached_data_` and `cached_output_` from the class as we discussed last time.

marcoabreu:
How is it possible that GetBoolHash was removed? I thought it was there to support the caching. Is it not actually required?

src/operator/nn/mkldnn/mkldnn_convolution-inl.h
```diff
@@ -35,19 +36,79 @@
...
+              "quantized convolution op to calculate primitive scale");
+    DMLC_DECLARE_FIELD(max_calib_range)
+    .set_default(dmlc::optional<float>())
+    .describe("The maximum scalar value in the form of float32 obtained "
+              "through calibration. If present, it will be used to by "
+              "quantized convolution op to calculate primitive scale");
+  }
+  const int GetBoolHash() const {
+    int hash = 0;
+    hash = hash * 2 + this->with_bn ? 1 : 0;
```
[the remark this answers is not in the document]
ZhennanQin: This function is not used anymore, I simply remove it. Sorry about that.

include/mxnet/c_api.h
```diff
@@ -1535,6 +1534,15 @@ MXNET_DLL int MXSetCalibTableToQuantizedSymbol(SymbolHandle qsym_handle,
...
 
+/*!
+ * \brief Run subgraph pass based on the backend provided
+ * \param sym_handle symbol to be converted
+ * \param backend backend names for subgraph pass
+ * \param ret_sym_handle returned symbol
+ */
+MXNET_DLL int MXGenBackendSubgraph(SymbolHandle sym_handle, const char *backend,
+                                   SymbolHandle *ret_sym_handle);
+
```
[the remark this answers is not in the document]
ZhennanQin: It's not for testing, but for quantization script. For mkldnn quantization, we agreed to do fusion first, and then do quantization. So on python side, we need an api to generate fused graph, and then pass it to quantization pass. Otherwise, we have to allow simple_bind returning the graph after subgraph pass.

src/executor/attach_op_execs_pass.cc
```diff
@@ -159,9 +159,13 @@ class StatefulComputeExExecutor : public OpExecutor {
...
-    CreateDefaultInputs(in_array, &in_array_fallback);
-    fcompute_(state_, op_ctx, in_array_fallback, req, out_array);
-    return;
+    // TODO(alex): (MXNET-847) Remove this fallback feature after subgraph implemented
+    const auto is_mkldnn = Op::GetAttr<bool>("TIsMKLDNN");
+    if (!is_mkldnn.get(attrs_.op, false)) {
+      CreateDefaultInputs(in_array, &in_array_fallback);
+      fcompute_(state_, op_ctx, in_array_fallback, req, out_array);
+      return;
+    }
```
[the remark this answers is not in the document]
ZhennanQin: OK. I will move this part out.

src/executor/attach_op_execs_pass.cc
```diff
@@ -180,12 +184,14 @@ class StatefulComputeExExecutor : public OpExecutor {
...
-  explicit StatefulComputeExExecutor(const OpStatePtr& state,
+  explicit StatefulComputeExExecutor(const NodeAttrs& attrs,
+                                     const OpStatePtr& state,
                                      const FStatefulComputeEx& fcompute,
                                      ExecType exec_type)
-      : state_(state), fcompute_(fcompute), exec_type_(exec_type) {}
+      : attrs_(attrs), state_(state), fcompute_(fcompute), exec_type_(exec_type) {}
 
  private:
+  NodeAttrs attrs_;
```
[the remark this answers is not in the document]
ZhennanQin: To query TIsMKLDNN for StatefulComputeExExecutor.

src/operator/quantization/mkldnn/mkldnn_quantize-inl.h
```diff
@@ -75,6 +75,11 @@ static void MKLDNNQuantizeComputeKer(const std::vector<NDArray>& inputs,
   auto i_mpd = i_mem->get_primitive_desc();
   auto i_desc = i_mpd.desc();
   mkldnn::memory::format i_fmt = static_cast<mkldnn::memory::format>(i_desc.data.format);
+  if (i_fmt == mkldnn::memory::format::nchw ||
+      i_fmt == mkldnn::memory::format::nChw8c ||
+      i_fmt == mkldnn_nChw16c) {
+    i_fmt = mkldnn::memory::format::nhwc;
+  }
```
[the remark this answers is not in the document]
ZhennanQin: For mkldnn, yes. nhwc should be the default int8 layout, just like nchw for fp32.

src/operator/subgraph/mkldnn/mkldnn_conv-inl.h
```diff
@@ -0,0 +1,48 @@
...
+#include "../../nn/batch_norm-inl.h"
+#include "../../nn/mkldnn/mkldnn_convolution-inl.h"
+
+namespace mxnet {
+namespace op {
+
+struct MKLDNNConvFusionParam {
+  MKLDNNConvFullParam full_conv_param;
+  std::shared_ptr<BatchNormParam> bn_param;
+};
```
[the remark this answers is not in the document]
ZhennanQin: It's an abstraction to isolate mkldnn convolution params with fusion params: 
MKLDNNConvFullParam defines in mkldnn_convolution-inl.h, which only contains the options convolution needed, and will pass to MKLDNNConvolutionForwardFullFeature.
MKLDNNConvFusionParam defines in mkldnn_conv-inl.h, which is used for SgMKLDNNConvParamParser, and to support fusion related function in SgMKLDNNConvOperator::Forward.

src/operator/subgraph/mkldnn/mkldnn_conv.cc
ZhennanQin: Will remove it.

src/operator/subgraph/mkldnn/mkldnn_conv.cc
ZhennanQin: Sure.

ZhennanQin:
@marcoabreu Because for original convolution op, the newly created params doesn't be used. And for _sg_mkldnn_conv, it's an stateful op, so it doesn't rely on caching(not use caching mechanism).

example/quantization/imagenet_gen_qsym_mkldnn.py
```diff
@@ -0,0 +1,213 @@
...
+def save_params(fname, arg_params, aux_params, logger=None):
+    if logger is not None:
+        logger.info('Saving params into file at %s' % fname)
+    save_dict = {('arg:%s' % k): v.as_in_context(cpu()) for k, v in arg_params.items()}
+    save_dict.update({('aux:%s' % k): v.as_in_context(cpu()) for k, v in aux_params.items()})
+    mx.nd.save(fname, save_dict)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Generate a calibrated quantized model from a FP32 model with MKL-DNN support')
```
eric-haibin-lin: Could update readme.md with an example to run this script?

example/quantization/imagenet_gen_qsym_mkldnn.py
```diff
@@ -0,0 +1,213 @@
...
+    if calib_mode != 'none':
+        download_calib_dataset('http://data.mxnet.io/data/val_256_q90.rec', args.calib_dataset)
+
+        # download model
+        prefix, epoch = download_model(model_name=args.model, logger=logger)
+        sym, arg_params, aux_params = mx.model.load_checkpoint(prefix, epoch)
+
+    out = SymbolHandle()
+    backend = "MKLDNN"
+    check_call(_LIB.MXGenBackendSubgraph(sym.handle, c_str(backend), ctypes.byref(out)))
```
eric-haibin-lin: Calling C_API in the example seems not user friendly. Do we want to have sth like this in the symbol.py? @zheng-da

python/mxnet/contrib/quantization.py
```diff
@@ -40,7 +40,7 @@
 from ..module import Module
 
 
-def _quantize_params(qsym, params):
+def _quantize_params(qsym, params, th_dict):
```
eric-haibin-lin: Does th_dict mean threashold_dict? I don't understand what th stands for

python/mxnet/contrib/quantization.py
```diff
@@ -408,12 +420,23 @@ def _load_params(params, logger=logging):
         raise ValueError('Unsupported params provided. Must be either a path to the param file or'
                          ' a pair of dictionaries representing arg_params and aux_params')
 
+def save_params(fname, arg_params, aux_params, logger=None):
```
eric-haibin-lin: This is considered as a public API under the contrib.quantization namespace. What is the necessity of adding such an API?

src/operator/quantization/quantize_graph_pass.cc
```diff
@@ -89,17 +89,43 @@ std::vector<NodeEntry> OfflineParams(std::vector<NodeEntry>&& outputs,
   return outputs;
 }
 
-inline bool NeedQuantize(NodePtr node, const std::unordered_set<NodePtr> excluded_nodes) {
-  static auto& quantized_op_map = Op::GetAttr<mxnet::FQuantizedOp>("FQuantizedOp");
-  return quantized_op_map.count(node->op()) && !excluded_nodes.count(node);
+inline bool NeedQuantize(NodePtr node,
+                         const std::unordered_set<std::string> excluded_nodes) {
```
eric-haibin-lin: Use const reference to save a copy

src/operator/subgraph/mkldnn/mkldnn_conv.cc
```diff
@@ -0,0 +1,672 @@
...
+    }
+  }
+
+  if (weight_channelwise_scale) {
+    weight_scales->resize(channel);
+#pragma omp parallel for num_threads(engine::OpenMP::Get()->GetRecommendedOMPThreadCount())
+    for (int c = 0; c < static_cast<int>(channel); ++c) {
+      DType weight_range = MaxAbs(weight_c_min[c], weight_c_max[c]);
+      weight_scales->at(c) = int8_range / weight_range;
+      DType *fp_ptr = weight_ptr + c * offset;
```
eric-haibin-lin: nit: use const Dtype* when it applies

src/operator/subgraph/mkldnn/mkldnn_conv.cc
```diff
@@ -0,0 +1,672 @@
...
+  node->attrs.dict["quantized"] = "true";
+  node->attrs.subgraphs.reserve(attrs.subgraphs.size());
+  for (auto sub : attrs.subgraphs) {
+    node->attrs.subgraphs.push_back(sub);
+  }
+  node->op()->attr_parser(&(node->attrs));
+  return node;
+}
+
+static inline bool StringEndsWith(std::string const &str,
```
eric-haibin-lin: This can probably be moved to common/utils.cc

src/operator/subgraph/mkldnn/mkldnn_conv_post_quantize_property.cc
```diff
@@ -0,0 +1,163 @@
...
+#include "../../quantization/requantize-inl.h"
+
+namespace mxnet {
+namespace op {
+
+class SgMKLDNNConvPostQuantizeSelector : public SubgraphSelector {
+ public:
+  /*! \brief pattern match status */
+  enum SelectStatus {
+    sFail = 0,
```
eric-haibin-lin: Usually we use prefix with `k` for enum (kFail)

src/operator/subgraph/mkldnn/mkldnn_conv_property.cc
```diff
@@ -0,0 +1,239 @@
...
+#include "../subgraph_property.h"
+#include "../../nn/activation-inl.h"
+
+namespace mxnet {
+namespace op {
+class SgMKLDNNConvSelector : public SubgraphSelector {
+ public:
+  /*! \brief pattern match status */
+  enum SelectStatus {
+    sFail = 0,
```
eric-haibin-lin: sFail -> kFail, sStart -> kStart, etc

example/quantization/imagenet_gen_qsym_mkldnn.py
ZhennanQin: We do want to provide resnet50v1 as example, but we don't know where's to put the pre-trained model and its parameter file. Do you have any suggestion where's to  upload them?

python/mxnet/contrib/quantization.py
ZhennanQin: I guess it means threashold_dict, @reminisce can you explain it?

python/mxnet/contrib/quantization.py
ZhennanQin: This is already inside imagenet_gen_qsym.py. Will remove it.

python/mxnet/contrib/quantization.py
reminisce: Yes, it means threshold.

include/mxnet/c_api.h
```diff
@@ -1502,22 +1502,21 @@ MXNET_DLL int MXSymbolInferType(SymbolHandle sym,
...
- * \param ret_sym_handle quantized symbol result
- * \param num_excluded_symbols number of layers excluded from being quantized in the input symbol
- * \param excluded_symbols array of symbols to be excluded from being quantized
- * \param num_offline number of parameters that are quantized offline
- * \param offline_params array of c strings representing the names of params quantized offline
- * \param quantized_dtype the quantized destination type for input data.
- */
-MXNET_DLL int MXQuantizeSymbol(SymbolHandle sym_handle,
-                               SymbolHandle *ret_sym_handle,
+* \brief Convert a symbol into a quantized symbol where FP32 operators are replaced with INT8
```
reminisce: The space is needed.

src/operator/quantization/quantize_graph_pass.cc
```diff
@@ -89,17 +89,41 @@ std::vector<NodeEntry> OfflineParams(std::vector<NodeEntry>&& outputs,
...
-  return quantized_op_map.count(node->op()) && !excluded_nodes.count(node);
+  if (quantized_op_map.count(node->op())) {
+    bool excluded = false;
+    if (node->attrs.subgraphs.size()) {
+      // This is a subgraph node, try to match subgraph name first,
+      // and then try to match inner node.
+      if (excluded_nodes.count(node->attrs.name)) {
+        excluded = true;
+      } else {
+        auto subgraph_sym = node->attrs.subgraphs[0];
```
reminisce: Please add a comment explaining what index 0 correspond to. If possible, add a CHEKC_EQ as well for `node->attrs.subgraphs`.

src/operator/quantization/quantize_graph_pass.cc
```diff
@@ -199,33 +245,45 @@ Graph QuantizeGraph(Graph &&src) {
...
-          size_t num_outputs = mirror_node->num_outputs() - 2;
-          uint32_t min_index = num_outputs + 2 * e.index;
-          uint32_t max_index = num_outputs + 2 * e.index + 1;
-          NodePtr dequantize_node = CreateNode("_contrib_dequantize",
-            e.node->attrs.name + "_dequantize");
-          dequantize_node->inputs.emplace_back(mirror_entry);
-          dequantize_node->inputs.emplace_back(NodeEntry{mirror_node, min_index, 0});
-          dequantize_node->inputs.emplace_back(NodeEntry{mirror_node, max_index, 0});
-          dequantize_node->op()->attr_parser(&(dequantize_node->attrs));
+      if (node->is_variable() && node->attrs.name == "data") {
```
reminisce: What is changed here?

src/operator/subgraph/mkldnn/mkldnn_conv.cc
```diff
@@ -0,0 +1,667 @@
...
+  }
+
+ private:
+  bool initalized;
+  nnvm::Symbol subgraph_sym_;
+  MKLDNNConvFusionParam param;
+  std::shared_ptr<MKLDNNConvForward> fwd;
+  NDArray cached_weight_;
+  NDArray cached_bias_;
+  float cached_data_min;
```
reminisce: Add `_` suffix for data members.

src/operator/subgraph/mkldnn/mkldnn_conv_post_quantize_property.cc
```diff
@@ -0,0 +1,163 @@
...
+        std::make_shared<SgMKLDNNConvPostQuantizeSelector>(disable_all);
+    return selector;
+  }
+
+  void ConnectSubgraphOutput(
+      const nnvm::NodePtr n,
+      std::vector<nnvm::NodeEntry *> *output_entries) const override {
+    for (size_t i = 0; i < output_entries->size(); ++i) {
+      auto entry_ptr = output_entries->at(i);
+      *entry_ptr = nnvm::NodeEntry{n, entry_ptr->index, 0};
```
reminisce: What's the relationship between `n` and `entry_ptr->node`? What justifies that `entry_ptr->index` is the correct index of `n`'s output?

src/operator/subgraph/subgraph_property.h
```diff
@@ -92,6 +92,22 @@ class SubgraphProperty {
   // execute the operators in the subgraph.
   virtual nnvm::NodePtr CreateSubgraphNode(const nnvm::Symbol &s,
                                            const int subgraph_id = 0) const = 0;
+  // Connect subgraph internal output with external output entries. By default,
+  // each output entry will connect to an unique internal output.
+  virtual void ConnectSubgraphOutput(
```
reminisce: `ConnectSubgraphOutputs`.

src/operator/subgraph/subgraph_property.h
```diff
@@ -92,6 +92,22 @@ class SubgraphProperty {
...
+      const nnvm::NodePtr n,
+      std::vector<nnvm::NodeEntry *> *output_entries) const {
+    for (size_t i = 0; i < output_entries->size(); ++i) {
+      *output_entries->at(i) = nnvm::NodeEntry{n, static_cast<uint32_t>(i), 0};
+    }
+  }
+  // Connect subgraph internal input with external input entries. By default,
+  // each input entry will connect in top sorted order.
+  virtual void ConnectSubgraphInput(
```
reminisce: `ConnectSubgraphInputs`.

src/operator/quantization/quantize_graph_pass.cc
ZhennanQin: OK. But this is not a good design that we didn't have a rule that which index should be the subgraph symbol node as I comment in your subgraph PR. We have to assume that the first node(index 0) is the subgraph node, and all developers won't break this.

src/operator/quantization/quantize_graph_pass.cc
ZhennanQin: This is to solve first layer calibration problem. If we try to quantize first layer(eg. conv0), then we will insert an quantize op between data and conv0. After enabling  calib_quantize_op, we will collect calibration info on data, which can't be done by set_monitor_callback as it only monitors op's output, but data is not an output of any op. To solve this problem, we insert an identity op for data, to convert it to identity's output, then we can use set_monitor_callback to do calibration for it.

src/operator/subgraph/mkldnn/mkldnn_conv_post_quantize_property.cc
ZhennanQin: For this case, n is _sg_mkldnn_conv, and entry_ptr->node is the followed requantize op. This is a customized version for this certain fusion pass, as _sg_mkldnn_conv has same output sequence(output, output_min, output_max) as requantize.

ZhennanQin:
@marcoabreu Can you review again to see if the change resolve your concern? Thanks.

src/operator/subgraph/mkldnn/mkldnn_conv-inl.h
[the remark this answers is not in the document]
zheng-da: `MKLDNNConvFullParam` contains the following data structure. It's used for fused convolution, right?
```
struct MKLDNNConvParam : public dmlc::Parameter<MKLDNNConvParam> {
  // When adding more members into this class, please double check GetHash()
  // won't overflow.
  bool with_bn;
  bool with_relu;
  bool with_sum;
  bool with_postsum_relu;
  bool quantized;
  bool weight_channelwise_scale;

```

src/operator/nn/mkldnn/mkldnn_convolution.cc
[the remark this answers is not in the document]
zheng-da: if so, please change it to `GetMKLDNNData`.

example/quantization/imagenet_gen_qsym_mkldnn.py
zheng-da: agree. it's better to provide a Python API for this.

src/operator/quantization/quantize_graph_pass.cc
zheng-da: why are there more than 1 subgraph on a subgraph node?

reminisce: You have the full control of creating subgraph(s) in your subgraph property implementation. As we discussed before, if you plan to have more subgraphs, it's better to use enum as indices which are more descriptive.

src/operator/subgraph/mkldnn/mkldnn_conv-inl.h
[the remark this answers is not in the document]
ZhennanQin: @zheng-da Yes. You're right. MKLDNNConvFullParam contains all the parameters for mkldnn convolution primitive.

src/operator/quantization/quantize_graph_pass.cc
zheng-da: the `subgraphs` field isn't only used by the subgraph API. control flow operators also use this field to store subgraphs.

pengzhao-intel:
As discussion from dev@ (see below link), the current failure of travis-ci/pr can be ignored since the issue from ccache mechanism.

https://lists.apache.org/thread.html/ae5f0151f0398492aed066cdce41f2de2fa675c68612704409458a01@%3Cdev.mxnet.apache.org%3E

src/operator/quantization/quantize_graph_pass.cc
ZhennanQin: CHECK is added.

src/operator/subgraph/mkldnn/mkldnn_conv_property.cc
```diff
@@ -0,0 +1,239 @@
...
+      return true;
+    }
+    return false;
+  }
+
+  bool SelectInput(const nnvm::Node &n, const nnvm::Node &new_node) override {
+    return false;
+  }
+
+  bool SelectOutput(const nnvm::Node &n, const nnvm::Node &new_node) override {
```
zheng-da: how do you make sure the output of convolution is only used by batch_norm or activation?

ZhennanQin: See line 71 to 79

zheng-da:
@pengzhao-intel @KellenSunderland one thing I don't understand is that why my PR can pass all tests?
https://github.com/apache/incubator-mxnet/pull/12698
https://github.com/apache/incubator-mxnet/pull/12693

My concern is that if the CI problem isn't fixed and this PR is merged, all following PRs won't be able to pass CI any more?

ZhennanQin:
@zheng-da Based on my knowledge, merging this PR won't cause following CI failed. The reason that travis timeout is ccache can't find a cached build that similar enough to this PR, causing it almost build from scratch. After this PR merged, ccache will create new cache based on this PR, and following PR can still be built very quickly based on new cache.
@KellenSunderland Is my understanding right?

ZhennanQin:
@marcoabreu Ping again... As you requested changes to this PR, can you review again to list if any concern isn't resolved?

KellenSunderland:
Yes that is correct.

src/operator/quantization/quantize_graph_pass.cc
```diff
@@ -89,17 +89,42 @@ std::vector<NodeEntry> OfflineParams(std::vector<NodeEntry>&& outputs,
...
+inline bool NeedQuantize(NodePtr node, const std::unordered_set<std::string>& excluded_nodes) {
   static auto& quantized_op_map = Op::GetAttr<mxnet::FQuantizedOp>("FQuantizedOp");
-  return quantized_op_map.count(node->op()) && !excluded_nodes.count(node);
+  static auto& fexec_type = nnvm::Op::GetAttr<FExecType>("FExecType");
+  const auto& op = node->op();
+  if (op && quantized_op_map.count(op)) {
+    bool need = true;
+    if (excluded_nodes.count(node->attrs.name)) {
+      need = false;
+    } else if (node->attrs.subgraphs.size()) {
```
KellenSunderland: Nit: consider
```
else if (!node->attrs.subgraphs.empty())
```
A little more explicit / readable.

src/operator/quantization/quantize_graph_pass.cc
```diff
@@ -199,33 +248,45 @@ Graph QuantizeGraph(Graph &&src) {
...
-          uint32_t min_index = num_outputs + 2 * e.index;
-          uint32_t max_index = num_outputs + 2 * e.index + 1;
-          NodePtr dequantize_node = CreateNode("_contrib_dequantize",
-            e.node->attrs.name + "_dequantize");
-          dequantize_node->inputs.emplace_back(mirror_entry);
-          dequantize_node->inputs.emplace_back(NodeEntry{mirror_node, min_index, 0});
-          dequantize_node->inputs.emplace_back(NodeEntry{mirror_node, max_index, 0});
-          dequantize_node->op()->attr_parser(&(dequantize_node->attrs));
+      if (node->is_variable() && node->attrs.name == "data") {
+        // Instert identity for data to collect calib for it.
```
KellenSunderland: Instert -> Insert

src/operator/subgraph/subgraph_property.h
```diff
@@ -92,6 +92,22 @@ class SubgraphProperty {
   // execute the operators in the subgraph.
   virtual nnvm::NodePtr CreateSubgraphNode(const nnvm::Symbol &s,
                                            const int subgraph_id = 0) const = 0;
+  // Connect subgraph internal output with external output entries. By default,
+  // each output entry will connect to an unique internal output.
+  virtual void ConnectSubgraphOutputs(
+      const nnvm::NodePtr n,
+      std::vector<nnvm::NodeEntry *> *output_entries) const {
```
zheng-da: i think it's better check if `n` has the same number of outputs as `output_entries`.

tests/python/mkl/test_subgraph.py
```diff
@@ -0,0 +1,472 @@
...
+      output.wait_to_read()
+
+  sym_sg = sym.get_backend_symbol("MKLDNN")
+  mod_sg = Module(symbol=sym)
+  mod_sg.bind(data_shapes=[('data', data_shape)], label_shapes=[('softmax_label', label_shape)])
+  mod_sg.set_params(arg_params, aux_params)
+
+  mod_sg.forward(batch, is_train=False)
+  for output_sg in mod_sg.get_outputs():
+      output_sg.wait_to_read()
```
zheng-da: shouldn't you compare the outputs of fused version and unfused version?

src/c_api/c_api_symbolic.cc
```diff
@@ -696,3 +696,21 @@ int MXSetCalibTableToQuantizedSymbol(SymbolHandle qsym_handle,
   *ret_qsym_handle = s;
   API_END_HANDLE_ERROR(delete s);
 }
+
+int MXGenBackendSubgraph(SymbolHandle sym_handle, const char *backend,
```
KellenSunderland: By backend in this context do you mean the symbol that's a placeholder for the fused mkldnn call?

src/ndarray/ndarray.cc
```diff
@@ -1592,8 +1601,12 @@ void NDArray::Save(dmlc::Stream *strm) const {
     save_data = nd_cpu.data();
   } else {
     this->WaitToRead();
-    save_data = this->data();
     nd_cpu = *this;
+#if MXNET_USE_MKLDNN == 1
+    if (nd_cpu.IsMKLDNNData())
```
KellenSunderland: NIt: I believe the project encourages braces even on single line conditionals.

src/operator/quantization/mkldnn/mkldnn_quantize-inl.h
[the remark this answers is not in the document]
KellenSunderland: Question for both of you: If we quantize to int8 and store the params, will they be stored in nhwc by default (and thus not need to be converted each time the model is loaded)?

tests/python/mkl/test_subgraph.py
```diff
@@ -0,0 +1,472 @@
...
+@with_seed()
+def test_neg_conv_bn_add_relu():
+  for data_shape in DATA_SHAPE:
+    syms, attrs, excluded_attrs = neg_conv_bn_add_relu()
+    check_neg_fusion(syms, attrs, excluded_attrs, data_shape)
+
+
+if __name__ == "__main__":
+  import nose
+  nose.runmodule()
```
KellenSunderland: nit: newline

src/operator/subgraph/subgraph_property.h
ZhennanQin: This can't be guaranteed as output_entries may have duplicated entries from n.

python/mxnet/contrib/quantization.py
```diff
@@ -53,6 +53,7 @@ def _quantize_params(qsym, params):
     qsym : Symbol
         Quantized symbol from FP32 symbol.
     params : dict of str->NDArray
+    th_dict: dict of min/max pairs of layers' output
```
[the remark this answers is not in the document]
ZhennanQin: It applies to the output as well.

tests/python/mkl/test_subgraph.py
ZhennanQin: Yes, it's done by line 119.

src/c_api/c_api_symbolic.cc
ZhennanQin: What do you mean by a placeholder for the fused mkldnn call?
This API is intend for converting a symbol into a backend specific symbol. On quantization flow, we need do fusion at first, and then do quantization. So on python level, we need to get the backend specific symbol that all backend specific fusion applied, and pass it to quantization pass. That's why we need this API.

src/ndarray/ndarray.cc
ZhennanQin: I don't see such rule from the code itself. Single line conditionals without braces can be found everywhere, even in this file. See line 1591, 1654, 1669 ...

tests/python/mkl/test_subgraph.py
ZhennanQin: Why make lint can't report this?

src/c_api/c_api_symbolic.cc
KellenSunderland: I guess my point is that 'backend' is an overloaded term here, so to me it's confusing when you say you're converting a symbol to a backend specific symbol.

Am I understanding correctly that when you're fusing ops and calling this function the symbol that you begin with (sym_handle) is a symbol representing a graph of NNVM ops which use our default MXNet backend, and the symbol you're converting to represents a fused operator targeting an MKLDNN backend?

tests/python/mkl/test_subgraph.py
KellenSunderland: No idea, for source files it should.

KellenSunderland:
If you guys could rebase and push this it would really help use to verify that Travis is now working correctly.

KellenSunderland:
Thanks for rebasing gentleman!

xinyu-intel:
@KellenSunderland The CI looks good now:)

src/operator/subgraph/subgraph_property.h
```diff
@@ -92,6 +92,22 @@ class SubgraphProperty {
...
+    for (size_t i = 0; i < output_entries->size(); ++i) {
+      *output_entries->at(i) = nnvm::NodeEntry{n, static_cast<uint32_t>(i), 0};
+    }
+  }
+  // Connect subgraph internal input with external input entries. By default,
```
reminisce: Same here.

src/operator/subgraph/mkldnn/mkldnn_conv_property.cc
```diff
@@ -0,0 +1,239 @@
...
+        // Make sure n is the left operand of sum, if not,
+        // switch sum operands sequence to ensure that
+        // the extra sum operand stays in the last of inputs.
+        if (node_sets.count(node->inputs[1].node.get())) {
+          auto tmp = node->inputs[1];
+          node->inputs[1] = node->inputs[0];
+          node->inputs[0] = tmp;
+          std::rotate(input_entries->begin(), input_entries->begin() + 1,
+                      input_entries->end());
+          std::rotate(orig_input_entries->begin(),
```
reminisce: This will change the topo-sorted order of `orig_input_entries`. Have you tested the effect of this change?

src/c_api/c_api_symbolic.cc
ZhennanQin: Yes, your understanding is correct.

src/operator/subgraph/subgraph_property.h
```diff
@@ -92,6 +92,22 @@ class SubgraphProperty {
   // execute the operators in the subgraph.
   virtual nnvm::NodePtr CreateSubgraphNode(const nnvm::Symbol &s,
                                            const int subgraph_id = 0) const = 0;
+  // Connect subgraph internal output with external output entries. By default,
+  // each output entry will connect to an unique internal output.
+  virtual void ConnectSubgraphOutputs(
+      const nnvm::NodePtr n,
```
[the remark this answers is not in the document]
ZhennanQin: @reminisce  Code style is not consistent in this file. SubgraphSelector use different coding style with SubgraphProperty, which making me confusing which coding style should to follow. For example, SubgraphSelector uses /*! */ style comment with param description, and long parameter name(eg nnvm::Node &input_node), while SubgraphProperty uses \\ style comment without param description, and short parameter name(eg. const nnvm::Symbol &s).
Changing these 2 SubgraphProperty functions into SubgraphSelector style doesn't make sense to me. If you insist, I'd like to adjust the code style for whole file.

src/operator/subgraph/mkldnn/mkldnn_conv_property.cc
ZhennanQin: Add code in bind and simple_bind to handle the case that input order change. Add test to verify this.(test_pos_conv_add2 is the case that input order change, we add a test that use bind with input list, and compare the result with non-fusion).

eric-haibin-lin:
Looks like some tests are failing

ZhennanQin:
> Looks like some tests are failing

@eric-haibin-lin Fixed.

src/operator/quantization/mkldnn/mkldnn_quantize-inl.h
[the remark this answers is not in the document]
ZhennanQin: @KellenSunderland Sorry for responding late. For mkldnn quantization flow, we won't offline quantize any params, but using fp32 params(eg. weight and bias) for quantized convolution. We will online quantize convolution params in first forwarding. For the code here, it's for the default output format of 'quantize' op when using in mkldnn quantization flow, this won't effect non-mkldnn quantization flow.

eric-haibin-lin:
Great contribution. Thanks

src/executor/graph_executor.cc
```diff
@@ -1561,15 +1561,32 @@ static nnvm::Symbol PartitionGraph(const nnvm::Symbol& src,
...
       ++i1;
     }
   }
-  return PartitionGraph(src, prop_name, arg_shapes, arg_dtypes, arg_stypes,
-                        default_ctx, ctx_map, in_arg_ctxes, aux_state_ctxes);
+
+  // setup in_args_map
+  std::unordered_map<std::string, NDArray> in_args_map;
+  for (size_t i = 0; i < in_args->size(); ++i) {
+    in_args_map[arg_names[i]] = in_args->at(i);
```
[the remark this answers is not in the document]
ZhennanQin: @mbrookhart I think MXNet requires inputs with unique names, otherwise, how to distinguish the inputs with same names? According to symbol.bind API:
**args (list of NDArray or dict of str to NDArray) –
Input arguments to the symbol.

If the input type is a list of NDArray, the order should be same as the order of list_arguments().
If the input type is a dict of str to NDArray, then it maps the name of arguments to the corresponding NDArray.
In either case, all the arguments must be provided.** 

So basically, name is the only id for each input.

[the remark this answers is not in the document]
mbrookhart: https://github.com/apache/incubator-mxnet/blob/c98b19e2d108a3861d89b475927e8a21a913e540/tests/python/unittest/test_operator.py#L1171

If it's a list, the names don't have to match.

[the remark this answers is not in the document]
mbrookhart: In general, MXNet seems to determine node ID's by shared_ptr memory address. Unfortunately, that trick doesn't work with Partition Graph because the pass copies the nodes. :(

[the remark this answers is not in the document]
ZhennanQin: Yeah, I understand that you can find the order from the code, but it's not guaranteed, and you shouldn't make any assumption based on that, as it may change in future because the order isn't a part of list_arguments() spec, at least for now.

I guess we should answer this question first, which is, should we support inputs with same name? If the answer is shouldn't, then we need to rename those inputs with same name in unit test and add corresponding check and document  to disallow user using like that.

If the answer is we should, then we need to define a clear way for distinguishing inputs apart from name on API level, instead of the undocumented DFS order.

According to the current API design, I guess inputs with same name shouldn't be supported, as the order of list_arguments() is unique only if inputs has unique name. Adding DFS order to list_arguments() spec isn't user friendly, as it's hard for end user to find out the final DFS order from a complex topology, as single op(eg. rand_zipfian) from end-user level may be consist of many small ops in final computing graph.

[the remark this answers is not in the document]
zheng-da: I thought graph partitioning preserves the order of inputs. It now doesn't?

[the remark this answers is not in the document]
ZhennanQin: @zheng-da Now graph partitioning may change the order of inputs. This is a basic requirement from graph optimization purpose. For example, if we found there's a redundant op that doesn't contribute to graph output, shouldn't we remove it? If the answer is yes, then the input list will be changed.
Besides, current API design provides bad support for inputs with same name(if you treat the undocumented behavior is a kind of support). We should fix this anyway.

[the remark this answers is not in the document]
mbrookhart: :) A bug someone is using isn't a bug, it's a feature. I don't know who wrote test_maximum_minimum, but it feels like they were trying to test the duplicate name case?

PR #12540, opened 2018-09-12
redirecting navigation items to latest info

aaronmarkham:
## Description ##
The website's different versions show people old and sometimes inaccurate info. This PR redirects traffic to old version content to the latest info.
API docs traffic is unaffected. Although I redirect ONNX to the latest.

It should combine nicely with the other htaccess redirects for Clojure and error documents that I have in #12426.

### DIFF
--- a/docs/build_version_doc/artifacts/.htaccess
+++ b/docs/build_version_doc/artifacts/.htaccess
@@ -4,3 +4,12 @@ RewriteRule ^get_started.*$ /install/ [R=301,L]
 RewriteRule ^how_to.*$ /faq/ [R=301,L]
 RewriteRule ^api/python/symbol.html$ /api/python/symbol/symbol.html [R=301,L]
 RewriteRule ^community/index.html$ /community/contribute.html [R=301,L]
+
+# Navigation bar redirects to latest info
+RewriteRule ^versions/[^\/]+/architecture/.*$ /architecture/ [R=301,L]
+RewriteRule ^versions/[^\/]+/community/.*$ /community/ [R=301,L]
+RewriteRule ^versions/[^\/]+/faq/.*$ /faq/ [R=301,L]
+RewriteRule ^versions/[^\/]+/gluon/.*$ /gluon/ [R=301,L]
+RewriteRule ^versions/[^\/]+/install/.*$ /install/ [R=301,L]
+RewriteRule ^versions/[^\/]+/tutorials/.*$ /tutorials/ [R=301,L]
+RewriteRule ^versions/[^\/]+/api/python/contrib/onnx.html /api/python/contrib/onnx.html [R=301,L]


aaronmarkham:
@lupesko - please review.

sandeep-krishnamurthy:
Thanks @aaronmarkham 
1. Change is confusing for readers trying to find docs in other version: With this change, I am unable to see any other doc page for other than master version. Everything gets redirected to master. It is confusing for users, say I click v1.0.0 and Tutorials => gets redirected to landing page of v1.0.0 
2. Versions are now applicable only for APIs. So why don't we just show versions under API? Currently, it is not clear versioning is only for API docs.
3. I go to a API page (say scala) then change version, it goes to landing page. Should we go to that current page in older version?

aaronmarkham:
@sandeep-krishnamurthy thanks for the feedback. I add clarifications to the description above in a Stats section and a Details on Each Change section. 

> Change is confusing for readers trying to find docs in other version: With this change, I am unable to see any other doc page for other than master version. Everything gets redirected to master. It is confusing for users, say I click v1.0.0 and Tutorials => gets redirected to landing page of v1.0.0

Basically users are only allowed to browse versions on the API pages. For tutorials I think each one should be tested back through versions and then labelled to what their minimum version is. But since we don't test the old versions now, I think it is better that we don't display them. For your UX flow here, which I think would be quite rare, but we could send them to a page where they get the option to continue to the old broken info, or go to the new stuff.

> Versions are now applicable only for APIs. So why don't we just show versions under API? Currently, it is not clear versioning is only for API docs.

Yes! But apparently it wasn't that easy last time when Krishnan tried. I'd like to tackle that next. But at least with this, people aren't going to bad/incorrect/non-working info or tutorials!

> I go to a API page (say scala) then change version, it goes to landing page. Should we go to that current page in older version?

It's always done that. Really annoying. That should be fixed too though (or moot if we get the versions just working right within the API section. Next PR?

docs/build_version_doc/artifacts/.htaccess
```diff
@@ -4,6 +4,16 @@ RewriteRule ^get_started.*$ /install/ [R=301,L]
...
 RewriteRule ^community/index.html$ /community/contribute.html [R=301,L]
+
+# Navigation bar redirects to latest info
+RewriteRule ^versions/[^\/]+/architecture/.*$ /architecture/ [R=301,L]
+RewriteRule ^versions/[^\/]+/community/.*$ /community/ [R=301,L]
+RewriteRule ^versions/[^\/]+/faq/.*$ /faq/ [R=301,L]
+RewriteRule ^versions/[^\/]+/gluon/.*$ /gluon/ [R=301,L]
+RewriteRule ^versions/[^\/]+/install/.*$ /install/ [R=301,L]
+RewriteRule ^versions/[^\/]+/tutorials/.*$ /tutorials/ [R=301,L]
+RewriteRule ^versions/[^\/]+/api/python/contrib/onnx.html /api/python/contrib/onnx.html [R=301,L]
```
lupesko: ONNX is an API. Why don't we want to allow someone with an older version of MXNet to see ONNX API for that version?

aaronmarkham:
I updated this to be like the Clojure redirects, so that any navbar selection going to ONNX on a version that didn't exist will go to the API error page.
I updated the [preview](http://34.201.8.176/) as well so you can test this out. Go to v1.0.0, then try to hit the Docs-->ONNX link in the nav.
Once this passes CI, please merge.

PR #12550, opened 2018-09-13
[MXNET-908] Enable python tests in Travis

KellenSunderland:
## Description ##
This turns on python tests in Travis in order to increase or MacOS coverage.

### Essentials ###

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,3 +29,4 @@ script:
   - export MXNET_STORAGE_FALLBACK_LOG_VERBOSE=0
   - mv make/osx.mk config.mk
   - make -j 2
+  - python -m nose --verbose tests/python/unittest/
--- a/ci/travis/install.sh
+++ b/ci/travis/install.sh
@@ -25,6 +25,6 @@ if [ ${TRAVIS_OS_NAME} == "osx" ]; then
     brew install libpng
     brew install ImageMagick
     brew install swig
-	python -m pip install --user nose numpy cython scipy requests
-	python3 -m pip install --user nose numpy cython scipy requests
+    python -m pip install --user nose numpy cython scipy requests mock
+    python3 -m pip install --user nose numpy cython scipy requests mock
 fi


kalyc:
Thanks for your contribution @KellenSunderland 
I noticed the build is failing
@mxnet-label-bot[pr-awaiting-response]

KellenSunderland:
The build is failing with the same error users are receiving.  It's a valid
failure which we were trying to reproduce in CI.

marcoabreu:
@KellenSunderland it seems like everything is passing. Is this good to merge?

KellenSunderland:
It's ok to merge.  My preference would be to wait a day and then I'll split
the fix into it's own PR (to make it easy to revert/bisect/cherry pick into
release branches).

.travis.yml
```diff
@@ -29,3 +29,4 @@ script:
   - export MXNET_STORAGE_FALLBACK_LOG_VERBOSE=0
   - mv make/osx.mk config.mk
   - make -j 2
```
lanking520: @KellenSunderland instead of placing python here as a job that travis would always execute, could you please put python build section into a separate target so we can also add Scala-maven test in here in the future.

lanking520: Possibly this will help: https://docs.travis-ci.com/user/build-stages

KellenSunderland: I think it would be great if we could add Scala tests as well, but I do have a few concerns.  How long would they take to execute?  Let's keep in mind that we have a limited number of executors that can be run before we begin interfering with other Apache projects.  There's also a global timeout of ~45 minutes to build and execute on a virtualized MacOS host which takes a few minutes just to start.  I'd ideally like to keep our total test time below 30 minutes to avoid hitting this in the future.

In principle I'm not against build stages, but unless we have a lot of jobs running in parallel that have dependencies on each other I don't see a big benefit in using them.

lanking520: Thanks for your reply! If you mean `scalaunittest`it will be around 10 mins or less. Is (travis wait)[https://stackoverflow.com/questions/43918874/how-to-increase-no-activity-wait-time-in-travis-ci]  possible to apply in here?

KellenSunderland:
Note: this will fail in Travis until this PR is merged: https://github.com/apache/incubator-mxnet/pull/12590

.travis.yml
KellenSunderland: Unfortunately Travis wait won't help in this case.  The tests are active and sending output to Travis, they just take a long time and hit the global timeout for Travis jobs.

10 minutes actually is reasonable.  Let's do a test run with Scala enabled after this is merged and see if we can get it included.  My guess is if we enable folder caching, and cache models we're currently downloading we'll be able to fit it in and not hit the global timeouts.

KellenSunderland:
Rebased, should build properly now.

lanking520:
Looks like we are still suffering the timeout issue: https://travis-ci.org/apache/incubator-mxnet/builds/433814087?utm_source=github_status&utm_medium=notification

PR #12558, opened 2018-09-14
[MXNET-952] Check for correlation kernel size along with unittest

### LINKED ISSUE #9034: Understanding the correlation layer
Hey everyone,

I do not understand the impact of the stride parameter for the mxnet.sym.correlation layer. I have this simple minimal example:

```
import numpy as np
import mxnet as mx

# Network
data1 = mx.symbol.Variable('data1')
#data2 = mx.symbol.Variable('data2')
cor = mx.sym.Correlation(data1=data1, data2=data1, kernel_size=28, stride1=1, stride2=1, pad_size=0, max_displacement=0)
loss = mx.sym.MakeLoss(cor, normalization='batch')
group = mx.symbol.Group([mx.sym.BlockGrad(cor), loss])

# Data
datashape = (1, 1, 28, 28)  # like mnist
data = np.random.rand(1, 1, 28, 28)

#Bind, execute, get computed correlation
executor = group.simple_bind(ctx=mx.cpu(), data1=datashape)
outs = executor.forward(is_train=True, data1=data)
cor = executor.outputs[0]
grad1 = executor.outputs[1]
print(cor)
```

It computes the correlation between 2 identical feature maps with a kernel of the same size as the feature maps. **I do not understand why the output of the correlation layer has shape 1x1x2x2**. If I set the strides to anything bigger than 1, the output is a 1x1x1x1 matrix, that makes more sense to me. As the Kernel has the size of the feature maps, the output should always be a 1x1x1x1 matrix, independent from the value of the strides parameters! 

The documentation notes

> 
> Given a maximum displacement :math:`d`, for each location :math:`x_{1}` it computes correlations :math:`c(x_{1}, x_{2})` only in a neighborhood of size :math:`D:=2d+1`,
>     by limiting the range of :math:`x_{2}`. We use strides :math:`s_{1}, s_{2}`, to quantize :math:`x_{1}` globally and to quantize :math:`x_{2}` within the neighborhood
>     centered around :math:`x_{1}`.
> 
>     The final output is defined by the following expression:
> 
>     .. math::
>       out[n, q, i, j] = c(x_{i, j}, x_{q})
> 
>     where :math:`i` and :math:`j` enumerate spatial locations in :math:`f_{1}`, and :math:`q` denotes the :math:`q^{th}` neighborhood of :math:`x_{i,j}`.

Unfortunately I do not really understand from this the impact of the strides parameter, as s_1 and s_2 do not appear in the equation.

Furthermore, if you just run the code a couple of times you will notice that the correlation matrix sometimes contains "Infinity" values. I do not understand mathematically or conceptually how the correlation can be infinity. 

Any help is much appreciated.

ChaiBapchya:
## Description ##
Added a check statement for kernel size whether it's an odd number or not with respect to Correlation operator.
Also added a test case in test_operator python file
This fixes #9034 

### Essentials ###

### DIFF
--- a/src/operator/correlation-inl.h
+++ b/src/operator/correlation-inl.h
@@ -78,6 +78,7 @@ class CorrelationOp : public Operator {
     using namespace mshadow;
     CHECK_EQ(in_data.size(), 2U);
     CHECK_EQ(out_data.size(), 3U);
+    CHECK_NE(param_.kernel_size%2,0) << "kernel size should be odd number";
     Stream<xpu> *s = ctx.get_stream<xpu>();
     Tensor<xpu, 4, DType> data1 = in_data[Correlation::kData1].get<xpu, 4, DType>(s);
     Tensor<xpu, 4, DType> data2 = in_data[Correlation::kData2].get<xpu, 4, DType>(s);
--- a/src/operator/correlation-inl.h
+++ b/src/operator/correlation-inl.h
@@ -78,7 +78,7 @@ class CorrelationOp : public Operator {
     using namespace mshadow;
     CHECK_EQ(in_data.size(), 2U);
     CHECK_EQ(out_data.size(), 3U);
-    CHECK_NE(param_.kernel_size%2, 0) << "kernel size should be odd number";
+    CHECK_NE(param_.kernel_size % 2, 0) << "kernel size should be odd number";
     Stream<xpu> *s = ctx.get_stream<xpu>();
     Tensor<xpu, 4, DType> data1 = in_data[Correlation::kData1].get<xpu, 4, DType>(s);
     Tensor<xpu, 4, DType> data2 = in_data[Correlation::kData2].get<xpu, 4, DType>(s);
--- a/src/operator/correlation-inl.h
+++ b/src/operator/correlation-inl.h
@@ -78,7 +78,7 @@ class CorrelationOp : public Operator {
     using namespace mshadow;
     CHECK_EQ(in_data.size(), 2U);
     CHECK_EQ(out_data.size(), 3U);
-    CHECK_NE(param_.kernel_size%2,0) << "kernel size should be odd number";
+    CHECK_NE(param_.kernel_size%2, 0) << "kernel size should be odd number";
     Stream<xpu> *s = ctx.get_stream<xpu>();
     Tensor<xpu, 4, DType> data1 = in_data[Correlation::kData1].get<xpu, 4, DType>(s);
     Tensor<xpu, 4, DType> data2 = in_data[Correlation::kData2].get<xpu, 4, DType>(s);
--- a/tests/python/unittest/test_operator.py
+++ b/tests/python/unittest/test_operator.py
@@ -6914,26 +6914,14 @@ def test_spacetodepth():
     test_invalid_depth_dim()
 
 @with_seed()
-def test_correlation_kernel_size():
-    import numpy as np
-    import mxnet as mx
+def test_invalid_kernel_size():
+    invalid_kernel_size = 28
+    assert_exception(mx.nd.Correlation,MXNetError,mx.nd.array(np.random.rand(1, 1, 28, 28)),mx.nd.array(np.random.rand(1, 1, 28, 28)),kernel_size=invalid_kernel_size)
 
-    # Network
-    data1 = mx.symbol.Variable('data1')
-    cor = mx.sym.Correlation(data1=data1, data2=data1, kernel_size=28, stride1=1, stride2=1, pad_size=0, max_displacement=0)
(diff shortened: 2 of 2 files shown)


src/operator/correlation-inl.h
```diff
@@ -78,6 +78,7 @@ class CorrelationOp : public Operator {
     using namespace mshadow;
     CHECK_EQ(in_data.size(), 2U);
     CHECK_EQ(out_data.size(), 3U);
+    CHECK_NE(param_.kernel_size%2,0) << "kernel size should be odd number";
```
Vikas89: spaces 
kernel_size % 2, 0

tests/python/unittest/test_operator.py
```diff
@@ -6913,6 +6913,28 @@ def test_invalid_depth_dim():
...
+    # Data
+    datashape = (1, 1, 28, 28)  # like mnist
+    data = np.random.rand(1, 1, 28, 28)
+
+    #Bind, execute, get computed correlation
+    executor = group.simple_bind(ctx=mx.cpu(), data1=datashape)
+    outs = executor.forward(is_train=True, data1=data)
+    cor = executor.outputs[0]
+    grad1 = executor.outputs[1]
+    print(cor)
```
Vikas89: print? instead do you want to assert?
Also, I would recommend to add a test case with odd version number

ChaiBapchya: Cool incorporating that

src/operator/correlation-inl.h
ChaiBapchya: Yup thanks

lupesko:
Thank you for the contribution @ChaiBapchya !

Note that CI tests have failed, due to cpplint ([link](http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/PR-12558/1/pipeline)) - please address and verify tests pass.

@mxnet-label-bot [pr-awaiting-testing]

tests/python/unittest/test_operator.py
```diff
@@ -6913,6 +6913,28 @@ def test_invalid_depth_dim():
     test_invalid_block_size()
     test_invalid_depth_dim()
 
+@with_seed()
+def test_correlation_kernel_size():
+    import numpy as np
```
apeforest: I think these are already imported in this module

tests/python/unittest/test_operator.py
apeforest: Agree with @Vikas89. I think your PR is mainly to test the error message when kernel size is even. Therefore, you do not need to test the regular case. Instead, you may want to test the case where the exception will be raised.

tests/python/unittest/test_operator.py
ChaiBapchya: True. My bad. Removing it.

kalyc:
Thanks for your contribution @ChaiBapchya 

tests/python/unittest/test_operator.py
ChaiBapchya: Yup. Added both. Thanks for the clarification.

src/operator/correlation-inl.h
```diff
@@ -78,6 +78,7 @@ class CorrelationOp : public Operator {
     using namespace mshadow;
     CHECK_EQ(in_data.size(), 2U);
     CHECK_EQ(out_data.size(), 3U);
+    CHECK_NE(param_.kernel_size%2, 0) << "kernel size should be odd number";
```
apeforest: nit: add space round '%'

tests/python/unittest/test_operator.py
```diff
@@ -6913,6 +6913,16 @@ def test_invalid_depth_dim():
     test_invalid_block_size()
     test_invalid_depth_dim()
 
+@with_seed()
+def test_invalid_kernel_size():
+    invalid_kernel_size = 28
+    assert_exception(mx.nd.Correlation,MXNetError,mx.nd.array(np.random.rand(1, 1, 28, 28)),mx.nd.array(np.random.rand(1, 1, 28, 28)),kernel_size=invalid_kernel_size)
```
apeforest: nit: break the long line.

tests/python/unittest/test_operator.py
```diff
@@ -6913,6 +6913,16 @@ def test_invalid_depth_dim():
...
+
+@with_seed()
+def test_valid_kernel_size():
+    valid_kernel_size = 9
+    mx.nd.Correlation(mx.nd.array(np.random.rand(1, 1, 28, 28)),mx.nd.array(np.random.rand(1, 1, 28, 28)),kernel_size=valid_kernel_size)
```
apeforest: same here

PR #12559, opened 2018-09-14
Add license headers to R-package

anirudhacharya:
## Description ##
Add license headers to all files in R-package. 

Current [MXNet release instructions](https://cwiki.apache.org/confluence/display/MXNET/Release+Process#ReleaseProcess-Step1.9.Createartifactsforthereleaseandpushtothedistfolder) requires all files in the R-package to be removed until licensing issues are resolved, before we build the tar file. This PR is to resolve those licensing issues.

## Comments ##
- @hetong007 @ankkhedia @lupesko @sandeep-krishnamurthy

### DIFF
--- a/R-package/LICENSE
+++ b/R-package/LICENSE
@@ -1,28 +1,202 @@
-Copyright (c) 2015 by Contributors
-All rights reserved.
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions are met:
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
-* Redistributions of source code must retain the above copyright notice, this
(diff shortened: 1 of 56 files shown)


sandeep-krishnamurthy:
Thanks @anirudhacharya 
Please add Apache 2.0 license header to all examples under demo/
We need to update the R-Package LICENSE file as well

ankkhedia:
> @anirudhacharya The R-Package folder also contains a LICENSE file, which is not an Apache2.0 license. How are we addressing it?

Should we just remove that license and make after adding Apache license to every folder.

ankkhedia:
> @ankkhedia why should we remove the file? scala, clojure and cpp package also have LICENSE files.
But python don't have. I don't mind it having over there just that if adding headers  to all the files suffices do we need an extra file?
Maybe @lupesko and @sandeep-krishnamurthy can comment on it.

sandeep-krishnamurthy:
@ankkhedia - I do not have a strong opinion but it is good to have in the base package folder.

PR #12571, opened 2018-09-14
Disable installation nightly test

aaronmarkham:
## Description ##
The nightly installation tests have been mostly disabled. The virtualenv one is breaking because that section was removed from the installation page.

This is a quick fix (like the others) that disables the tests.

## Comments
I recommend that this test concept be reimagined and reimplemented so it has actual coverage. It looks like only one out of seven would be running now.


kalyc:

marcoabreu:
FYI: We don't disable tests on nightly unless we entirely delete them.

aaronmarkham:
@marcoabreu - I was following the pattern that was set with the other several that are disabled. Really that whole lot of tests needs to be deleted and reimagined. I can propose something for it. I have something in mind already.

PR #12572, opened 2018-09-15
Make Gluon download function to be atomic

### LINKED ISSUE #11646: download function should avoid race-condition
Currently the download function would directly write to target file path, which causes file integrity problem when multiple processes use the same logic to write to the same file.
https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/gluon/utils.py#L223-L226

Alternative is to write to a temporary file first, and only move to target path upon download completion after verifying file integrity. Temporary file should be cleaned up at the end. This should help solve issues like #11616

stu1130:
## Description ##
Originally download function within gluon/utils.py would cause race-condition(#11646)(Fixes #11646). 
 * The PR tackle the problem by giving the downloaded file a tempory name which contains UUID and renaming back to target filename atomically. 
 * The atomic_replace should work with pytho2, pytho3 across Linux, OS X and Windows.

### Essentials ###

### Changes ###
* Assign a uuid for the downloaded file and rename back to the target filename 
* Implement a atomic_replace to handle trivial code which take care of pytho2, pytho3 across all the OS.
* Add corresponding unittests to check the correctness within multiprocesses.

## Comments ##


stu1130:

python/mxnet/gluon/utils.py
```diff
@@ -242,23 +277,34 @@ def download(url, path=None, overwrite=False, sha1_hash=None, retries=5, verify_
...
+                random_uuid = str(uuid.uuid4())
+                with open('{}.{}'.format(fname, random_uuid), 'wb') as f:
+                    # create uuid for temporary files
                     for chunk in r.iter_content(chunk_size=1024):
                         if chunk: # filter out keep-alive new chunks
                             f.write(chunk)
+                # if the target file exists(created by other processes),
+                # delete the temporary file
+                if os.path.exists(fname):
+                    os.remove('{}.{}'.format(fname, random_uuid))
```
anirudhacharya: here if `fname` local path exists, then the newly downloaded file is silently deleted. We should probably raise a user warning saying that a local file by the same name exists and the downloaded file has not been saved.

It might also be better to first check for the presence of local file and then make the HTTP call in line 285. This way  we will prevent making unnecessary calls, instead of actually downloading the file and then deleting it.

python/mxnet/gluon/utils.py
```diff
@@ -195,6 +197,39 @@ def check_sha1(filename, sha1_hash):
     return sha1.hexdigest() == sha1_hash
 
 
+if sys.platform != 'win32':
```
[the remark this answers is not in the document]
stu1130: Thanks for raising the question. cygwin meets POSIX requirement and rename is guaranteed to be atomic in POSIX

python/mxnet/gluon/utils.py
stu1130: will add UserWarning and I checked the target file existence before starting downloading at line 280

anirudhacharya: so if the check exists there in line 280. why are we repeating it here. What purpose does this check serve?

stu1130: Under multiprocessing environment, two processes might pass the check and one of them finish the download, rename. while the other one is still downloading or right before rename. In such case, I choose to remove instead of rename because there is a rare case where rename might be not atomic in windows based on some research on MSDN. it uses remove to mitigate the risk

python/mxnet/gluon/utils.py
```diff
@@ -242,23 +280,35 @@ def download(url, path=None, overwrite=False, sha1_hash=None, retries=5, verify_
...
+                    raise RuntimeError('Failed downloading url {}'.format(url))
+                # create uuid for temporary files
+                random_uuid = str(uuid.uuid4())
+                with open('{}.{}'.format(fname, random_uuid), 'wb') as f:
                     for chunk in r.iter_content(chunk_size=1024):
                         if chunk: # filter out keep-alive new chunks
                             f.write(chunk)
+                # if the target file exists(created by other processes),
+                # delete the temporary file
+                if os.path.exists(fname):
```
szha: check for file hash too, as the existing file might be an aborted incomplete download.

python/mxnet/gluon/utils.py
```diff
@@ -195,6 +197,42 @@ def check_sha1(filename, sha1_hash):
...
 
 
+if not sys.platform.startswith('win32'):
+    # refer to https://github.com/untitaker/python-atomicwrites
+    def _replace_atomic(src, dst):
+        """Implement atomic os.replace with linux and OSX. Internal use only"""
+        try:
+            os.rename(src, dst)
+        except OSError:
+            raise OSError('os.rename failed. Might be wrong source path.')
```
sandeep-krishnamurthy: Error message should be more user friendly. Something like - "Moving downloaded temp file - %s, to %s failed. Please retry the download.".format(src, dst) .
Or something similar.

sandeep-krishnamurthy:
@anirudhacharya / @szha - This is good to go, if your review is addressed.

python/mxnet/gluon/utils.py
```diff
@@ -195,6 +197,43 @@ def check_sha1(filename, sha1_hash):
...
+    def _path_to_unicode(x):
+        """Handle text decoding. Internal use only"""
+        if not isinstance(x, text_type):
+            return x.decode(sys.getfilesystemencoding())
+        return x
+
+    def _handle_errors(rv):
+        """Handle WinError. Internal use only"""
+        if not rv:
+            raise WinError()
```
szha: seems not informative. what could be the causes? what should users do when this happens?

stu1130: Good catch! will add more comments in docstring

python/mxnet/gluon/utils.py
```diff
@@ -195,6 +197,54 @@ def check_sha1(filename, sha1_hash):
...
+    text_type = unicode if sys.version_info[0] == 2 else str  # noqa
+
+    def _path_to_unicode(x):
+        """Handle text decoding. Internal use only"""
+        if not isinstance(x, text_type):
+            return x.decode(sys.getfilesystemencoding())
+        return x
+
+    def _format_error(err):
+        return ctypes.FormatError(_path_to_unicode(err))
```
szha: maybe rename to _path_to_unicode to _str_to_unicode, since err is obviously not a path.

anirudhacharya:
lgtm

vandanavk:
@stu1130 could you have a look at the CI so we can proceed with the review/approval

stu1130:
@vandanavk sure hopefully get it fixed by the end of this week

stu1130:
@szha @sandeep-krishnamurthy @anirudhacharya 
I add os.remove when MoveFileExW fail (e.g. fail to acquire file lock) so no temp file (XXX.UUID) will be left

python/mxnet/gluon/utils.py
```diff
@@ -242,31 +300,44 @@ def download(url, path=None, overwrite=False, sha1_hash=None, retries=5, verify_
...
                         if chunk: # filter out keep-alive new chunks
                             f.write(chunk)
+                # if the target file exists(created by other processes)
+                # and have the same hash with target file
+                # delete the temporary file
+                if not os.path.exists(fname) or (sha1_hash and not check_sha1(fname, sha1_hash)):
+                    # atmoic operation in the same file system
+                    _replace_atomic('{}.{}'.format(fname, random_uuid), fname)
+                else:
+                    os.remove('{}.{}'.format(fname, random_uuid))
```
sandeep-krishnamurthy: should this be protected within try..except?

sandeep-krishnamurthy:
@szha - Your review comments are addressed. Can you please take a look? Thanks.

python/mxnet/gluon/utils.py
```diff
@@ -231,7 +289,7 @@ def download(url, path=None, overwrite=False, sha1_hash=None, retries=5, verify_
             fname = os.path.join(path, url.split('/')[-1])
         else:
             fname = path
-    assert retries >= 0, "Number of retries should be at least 0"
+    assert retries >= 0, 'Number of retries should be at least 0'
```
szha: include the value of retries in the error message.

szha:
otherwise looks fine

PR #12594, opened 2018-09-19
[MXNET-867] Pooling1D with "same" padding

ChaiBapchya:
## Description ##
MaxPooling 1D added another pooling convention (aka padding type) called "same"

### Essentials ###

### DIFF
--- a/src/operator/nn/pooling.cc
+++ b/src/operator/nn/pooling.cc
@@ -99,7 +99,7 @@ static bool PoolingShape(const nnvm::NodeAttrs &attrs,
   if (param.pooling_convention == pool_enum::kSame) {
     CHECK_EQ(dshape.ndim(), 3U)
       << "Pooling: Input data should be 3D in (batch, channel, x)"
-      << ", Currently 'same' supports Max Pooling 1-D";
+      << ". Currently 'same' supports Max Pooling 1-D";
   }
   CHECK_GE(dshape.ndim(), 3U)
       << "Pooling: Input data should be  3D in (batch, channel, x)"
--- a/src/operator/nn/pooling.cc
+++ b/src/operator/nn/pooling.cc
@@ -100,6 +100,8 @@ static bool PoolingShape(const nnvm::NodeAttrs &attrs,
     CHECK_EQ(dshape.ndim(), 3U)
       << "Pooling: Input data should be 3D in (batch, channel, x)"
       << ". Currently 'same' supports Max Pooling 1-D";
+    CHECK(param.pad[0] == 0 and param.pad[1] == 0 and param.pad[2] == 0)
+      << "Same pooling convention disables the use of pad parameter.";
   }
   CHECK_GE(dshape.ndim(), 3U)
       << "Pooling: Input data should be  3D in (batch, channel, x)"
@@ -132,12 +134,12 @@ static bool PoolingShape(const nnvm::NodeAttrs &attrs,
                   (dshape[2] + 2 * param.pad[0] - param.kernel[0]) /
                       param.stride[0];
     } else if (param.pooling_convention == pool_enum::kFull) {
-      oshape[2] = 1 + static_cast<int>(ceil(
+      oshape[2] = 1 + static_cast<int>(std::ceil(
                           static_cast<float>(dshape[2] + 2 * param.pad[0] -
                                              param.kernel[0]) /
                           param.stride[0]));
     } else {
-      oshape[2] = static_cast<int>(ceil(
+      oshape[2] = static_cast<int>(std::ceil(
                           static_cast<float>(dshape[2] + 2 * param.pad[0]) /
                           param.stride[0]));
     }
--- a/src/operator/nn/pooling.cc
+++ b/src/operator/nn/pooling.cc
@@ -100,7 +100,7 @@ static bool PoolingShape(const nnvm::NodeAttrs &attrs,
     CHECK_EQ(dshape.ndim(), 3U)
       << "Pooling: Input data should be 3D in (batch, channel, x)"
       << ". Currently 'same' supports Max Pooling 1-D";
-    CHECK(param.pad[0] == 0 and param.pad[1] == 0 and param.pad[2] == 0)
+    CHECK(param.pad[0] == 0 && param.pad[1] == 0 && param.pad[2] == 0)
       << "Same pooling convention disables the use of pad parameter.";
   }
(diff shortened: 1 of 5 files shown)


stu1130:
Thanks for your contribution @ChaiBapchya 

stu1130:
Thanks for your contribution @ChaiBapchya 

tests/python/unittest/test_operator.py
```diff
@@ -6957,6 +6957,17 @@ def test_invalid_depth_dim():
     test_invalid_block_size()
     test_invalid_depth_dim()
 
+@with_seed()
+def test_max_pooling_pad_type_same():
+    input_data=mx.nd.array(np.random.rand(1,1,10))
+    output_data=mx.nd.Pooling(
+        input_data,
+        kernel=(2),
```
anirudh2290: what about other kernel.ndim() >= 2

ChaiBapchya: Specifically address the JIRA ticket that asks for maxpool1D

anirudh2290: so we should check and raise an exception if it is used with other ndims

src/operator/nn/pooling.cc
```diff
@@ -126,8 +126,8 @@ static bool PoolingShape(const nnvm::NodeAttrs &attrs,
       oshape[2] = 1 +
                   (dshape[2] + 2 * param.pad[0] - param.kernel[0]) /
                       param.stride[0];
-    } else {
-      oshape[2] = 1 + static_cast<int>(std::ceil(
+    } else if (param.pooling_convention == pool_enum::kFull) {
```
access2rohit: You introduced "else if" and removed "else". So there is no default behavior here. Are you sure all the cases will fall under these 2 conditionals only ?

tests/python/unittest/test_operator.py
haojin2: @ChaiBapchya You should add checks for illegal inputs that are not supported by your implementation.

ChaiBapchya: Yup, I'm working on it.

src/operator/nn/pooling.cc
```diff
@@ -96,6 +96,11 @@ static bool PoolingShape(const nnvm::NodeAttrs &attrs,
     CHECK(param.p_value.has_value());
   }
   const TShape &dshape = (*in_shape)[0];
+  if (param.pooling_convention == pool_enum::kSame) {
+    CHECK_EQ(dshape.ndim(), 3U)
+      << "Pooling: Input data should be 3D in (batch, channel, x)"
+      << ", Currently 'same' supports Max Pooling 1-D";
```
apeforest: nit: change coma to period.

src/operator/nn/pooling.cc
```diff
@@ -126,11 +131,15 @@ static bool PoolingShape(const nnvm::NodeAttrs &attrs,
...
-    } else {
-      oshape[2] = 1 + static_cast<int>(std::ceil(
+    } else if (param.pooling_convention == pool_enum::kFull) {
+      oshape[2] = 1 + static_cast<int>(ceil(
                           static_cast<float>(dshape[2] + 2 * param.pad[0] -
                                              param.kernel[0]) /
                           param.stride[0]));
+    } else {
+      oshape[2] = static_cast<int>(ceil(
+                          static_cast<float>(dshape[2] + 2 * param.pad[0]) /
```
apeforest: nit: indentation

src/operator/nn/pooling.cc
ChaiBapchya: My bad. Done

tests/python/unittest/test_operator.py
```diff
@@ -6957,6 +6957,21 @@ def test_invalid_depth_dim():
...
     test_invalid_depth_dim()
 
+@with_seed()
+def test_max_pooling_pad_type_same():
+    import math
+    input_data = mx.nd.array(np.random.rand(1,1,10))
+    stride = 2
+    kernel = 2
+    output_data=mx.nd.Pooling(
+        input_data,
```
apeforest: Do you also want to test the case where users selects "same" pad type but also supplies pad values?

apeforest:
Thanks for making this contribution. LGTM overall. You may also want to update the operator description in operator/nn/pooling.cc since you added a new type "same"

src/operator/nn/pooling.cc
ChaiBapchya: Unable to find the indentation issue. Also is there a similar linting tool for python to ensure I don't miss out on those. Do we use pylint?

apeforest: If pylint does not complain, I guess it's fine. But it seems not following the Google python style guide: https://github.com/google/styleguide/blob/gh-pages/pyguide.md

ChaiBapchya:
As discussed with Sina Afrooze and Lin, this is just the first step towards complete solution/fix.

Immediate demand was for Max Pool 1 Dimension to support "same" padding (the way Tensorflow does). Above pull request addresses that.

However, for a complete solution, following things need to be understood

### Asymmetric Padding
Feature request is about "Asymmetric padding"
Currently, MXNet supports Symmetric padding
e.g. Input data -> `1,2,3,4` and `pad`=`2` 
Output data -> `0,0,1,2,3,4,0,0`
 
#### Why isn't Asymmetric padding supported by CUDnn, MXNet?

Due to misalignment of input-output if padding happens only on one side (left or right alone).
It is called as *"Causal Convolution"* (only one side pad)

### Current work-around
Level of abstraction introduced.
By default, one can use `pad` parameter while performing `Pooling` or `Convolution` which performs symmetric padding. But, if there is a specific use-case for performing asymmetric padding, one has to perform following steps,
Perform independent padding (using `pad` operator) and then perform `Pooling` or `Convolution` as desired.

### Future steps
If one has to fully implement asymmetric padding, following changes need to be made
1. Pooling
- Add another parameter such as pad_type (because currently, `pooling_conv` which allows - 'full', 'valid', 'same' has been misused. Original meaning of `pooling_conv` was not supposed to take care of "type of padding"

2. Convolution
- Incorporate asymmetric padding for the same

anirudhacharya:
@ChaiBapchya can you create and issue for the future steps, giving the full description as in the above comment

ChaiBapchya:
Sure. Thanks for suggesting :)

PR #11742, opened 2018-07-13
[MXNET-623] Fixing an integer overflow bug in large NDArray

apeforest:
## Description ##
User filed a bug when converting NDArray to numpy. The rootcause is due to the large tensor size that exceeds the current limit (32-bit unsigned integer). This fix is to replace the unsigned 32-bit integer by int64_t data type to index tensor size. 

I verified the fix using the following code from its original [issue](https://github.com/apache/incubator-mxnet/issues/11495):
```
import mxnet as mx
from mxnet import gluon, nd
m = gluon.nn.Embedding(14000, 128)
m.initialize()
ind = nd.zeros((700000, 128))
x = m(ind)
x.shape
test=x.asnumpy()
```

I have run python memory_profiler and did not see noticeable memory footprint increase. I am still 
evaluating the performance in more cases.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [X] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [X] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Nightly tests are added for complicated/long-running ones: test_large_array.py
- [X ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [X ] Changed the shape index from uint32_t to int64_t.

## Comments ##
- This PR is to fix a bug filed on Github. It is not verified to support large array in all operators or scenarios in MXNet. A project is needed to address large array support in a large scope.
- This PR has a dependency on mshadow submodule. The related PR is https://github.com/dmlc/mshadow/pull/348

PR #12631, opened 2018-09-21
[MXNET-953] Fix oob memory read

KellenSunderland:
## Description ##
Simple bugfix addressing an out-of-bounds memory read.  

### Essentials ###

## Comments ##
- This fixes the problem but I'm open to suggestions on what the correct way to properly assign the memory to avoid similar errors.  Should we be using memcpy in cases like this?

## Testing Done ##
Ran nosetests-3.4 --verbose tests/python/unittest/test_operator.py:test_size_array with an ASAN build (described in more detail here: https://cwiki.apache.org/confluence/display/MXNET/Detecting+Memory+Leaks+and+Buffer+Overflows+in+MXNet ).

ASAN initially reported:
```
SUMMARY: AddressSanitizer: stack-buffer-overflow /work/mxnet/src/operator/tensor/elemwise_unary_op_basic.cc:546 in mxnet::op::SizeComputeCPU(nnvm::NodeAttrs const&, mxnet::OpContext const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&, std::vector<mxnet::OpReqType, std::allocator<mxnet::OpReqType> > const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&)
Shadow bytes around the buggy address:
  0x1000162b7420: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x1000162b7430: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x1000162b7440: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x1000162b7450: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x1000162b7460: f1 f1 f1 f1 f8 f2 f2 f2 f2 f2 f2 f2 f8 f2 f2 f2
=>0x1000162b7470: f2 f2 f2 f2 f8 f2 f2 f2 f2 f2 f2 f2[04]f2 f2 f2
  0x1000162b7480: f2 f2 f2 f2 f8 f2 f2 f2 f2 f2 f2 f2 f8 f2 f2 f2
  0x1000162b7490: f2 f2 f2 f2 f8 f2 f2 f2 f2 f2 f2 f2 f8 f2 f2 f2
  0x1000162b74a0: f2 f2 f2 f2 f8 f2 f2 f2 f2 f2 f2 f2 f8 f2 f2 f2
  0x1000162b74b0: f2 f2 f2 f2 00 00 00 00 00 00 00 00 00 00 00 00
  0x1000162b74c0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
```
After patch test passes as expected.
```
root@bf8e737c8b91:/work/mxnet#  nosetests-3.4 --verbose tests/python/unittest/test_operator.py:test_size_array
[INFO] Setting module np/mx/python random seeds, use MXNET_MODULE_SEED=1029956617 to reproduce.
test_operator.test_size_array ... ok

Ran 1 test in 0.049s

OK
```

### DIFF
--- a/src/operator/tensor/elemwise_unary_op_basic.cc
+++ b/src/operator/tensor/elemwise_unary_op_basic.cc
@@ -492,7 +492,8 @@ void ShapeComputeCPU(const nnvm::NodeAttrs& attrs,
   CHECK_EQ(req.size(), 1U);
   const TBlob& in_data = inputs[0];
   const TBlob& out_data = outputs[0];
-  memcpy(out_data.dptr_, in_data.shape_.data(), in_data.ndim() * sizeof(int64_t));
+  size_t type_size = mshadow::mshadow_sizeof(out_data.type_flag_);
+  memcpy(out_data.dptr_, in_data.shape_.data(), in_data.ndim() * type_size);
 }
 
 NNVM_REGISTER_OP(shape_array)
@@ -542,8 +543,9 @@ void SizeComputeCPU(const nnvm::NodeAttrs& attrs,
   CHECK_EQ(req.size(), 1U);
   const TBlob& in_data = inputs[0];
   const TBlob& out_data = outputs[0];
+  size_t type_size = mshadow::mshadow_sizeof(out_data.type_flag_);
   const index_t size_var = in_data.Size();
-  memcpy(out_data.dptr_, &size_var, 1U * sizeof(int64_t));
+  memcpy(out_data.dptr_, &size_var, type_size);
 }
 
 NNVM_REGISTER_OP(size_array)


KellenSunderland:
@anirudh2290 Does this fix look ok to you?

KellenSunderland:
@mxnet-label-bot can you please add label [pr-awaiting-review]

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -542,7 +542,7 @@ void SizeComputeCPU(const nnvm::NodeAttrs& attrs,
   CHECK_EQ(req.size(), 1U);
   const TBlob& in_data = inputs[0];
   const TBlob& out_data = outputs[0];
-  const index_t size_var = in_data.Size();
+  const int64_t size_var = in_data.Size();
   memcpy(out_data.dptr_, &size_var, 1U * sizeof(int64_t));
```
lebeg: No need to do `1U *` either.

lebeg: If I would want to assign a int64 variable a value, I would rather go for: 
```*static_cast<int64_t*>(out_data.dptr_) = in_data.Size();```

Comment (KellenSunderland) [reply src/operator/tensor/elemwise_unary_op_basic.cc]: Yeah I agree, better to cast the pointer and then assign.

File: src/operator/tensor/elemwise_unary_op_basic.cc [hunk 2]
```diff
@@ -542,8 +542,7 @@ void SizeComputeCPU(const nnvm::NodeAttrs& attrs,
   CHECK_EQ(req.size(), 1U);
   const TBlob& in_data = inputs[0];
   const TBlob& out_data = outputs[0];
-  const index_t size_var = in_data.Size();
-  memcpy(out_data.dptr_, &size_var, 1U * sizeof(int64_t));
+  *static_cast<int64_t*>(out_data.dptr_) = in_data.Size();
```
apeforest: Hmm...why casting to int64_t in the first place? I think maybe the fix should be just `sizeof(index_t)` instead of `sizeof(int64_t)`? Thanks.

KellenSunderland: If we did that what would happen when the value was read elsewhere in code from dtpr?

KellenSunderland: Let me add some logging code to see what types are used here.  My assumption is that uint64 is used for all size operations.  It's large enough to handle large indexes and it's integer type, which makes sense for floats.  Let me do a few test runs here in CI and validate those assumptions.

apeforest: uint32_t is currently used for index_t in mshadow. I have another PR to change index_t type from uint32_t to int64_t. Please see https://github.com/dmlc/mshadow/pull/348 and https://github.com/apache/incubator-mxnet/pull/11742 for more details

KellenSunderland: Thanks for the background @apeforest.  If it may be changing in the future perhaps it's best to make the size dependent on the TBlob type.  That way we don't have to change this again if index_t changes.

KellenSunderland:
Given this type may change in the future, I think the best solution is to cast based on out_data.type_flag_.  This way we don't have to update for the new size, and we correct the OOB.  What do you think @apeforest @lebeg?

Roshrini:
bounce for review @apeforest @anirudh2290

roywei:
ping @apeforest @anirudh2290 could you take a look again? Thanks!

PR #12632, opened 2018-09-21
Fix version dropdown behavior

aaronmarkham:
## Description ##
This PR fixes the bad behavior of the versions dropdown on the website. It will remember what API document you were viewing when you switch versions. It even remembers which anchor/bookmark you're on.

## Preview
You can play with it here: http://34.201.8.176/

This is how it works on the site now. You're in deep on some doc page, and then think, "I wonder if the params are different in the last release...". You click the versions dropdown, select 1.3.0, and then _bam_, you're taken to the 1.3.0 home page. 

This is so annoying. Now you have to go and search for that page again.

With this PR, changing versions does what you'd expect it to do!

If you pick a doc that didn't exist in a previous version, you get a 404 error.

### DIFF
--- a/docs/_static/js/docversion.js
+++ b/docs/_static/js/docversion.js
@@ -1,0 +1,25 @@
+$(document).ready(function () {
+	function setVersion(){
+  		console.log('Applying current file to links: ', doc[1]);
+  		if (document.getElementById('dropdown-menu-position-anchor-version')) {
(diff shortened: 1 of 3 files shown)


stu1130:
It's so sweet that you even made videos.

sandeep-krishnamurthy:
This is super important ux improvement, thanks for following it up and fixing it.

PR #12642, opened 2018-09-22
Enable C++ coverage

marcoabreu:
This PR adds coverage metrics for the C++ backend. 

Something to verify at a later stage: Compare coverage result of optimized builds with non-optimized: https://github.com/pyarmak/cmake-gtest-coverage-example/blob/master/cmake/modules/CodeCoverage.cmake

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,6 +94,7 @@ LDFLAGS = -pthread $(MSHADOW_LDFLAGS) $(DMLC_LDFLAGS)
 
 ifeq ($(ENABLE_TESTCOVERAGE), 1)
         CFLAGS += --coverage
+        LDFLAGS += --coverage
 endif
 
 ifeq ($(USE_TENSORRT), 1)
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -92,6 +92,10 @@ endif
 CFLAGS += -I$(TPARTYDIR)/mshadow/ -I$(TPARTYDIR)/dmlc-core/include -fPIC -I$(NNVM_PATH)/include -I$(DLPACK_PATH)/include -I$(TPARTYDIR)/tvm/include -Iinclude $(MSHADOW_CFLAGS)
 LDFLAGS = -pthread $(MSHADOW_LDFLAGS) $(DMLC_LDFLAGS)
 
+ifeq ($(ENABLE_TESTCOVERAGE), 1)
+        CFLAGS += --coverage
+        LDFLAGS += --coverage
+endif
```
szha: need to make sure this flag is portable to other compilers/linkers. I checked for gcc and clang, both of which supports gcov.
https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#Instrumentation-Options
https://clang.llvm.org/docs/SourceBasedCodeCoverage.html

marcoabreu: It's only compatible with GCC and clang, right. Want me to add a check?

marcoabreu:
Here's an example report: https://codecov.io/gh/apache/incubator-mxnet/tree/26763c7b686f6059beabf7e56dd98bf1e526ab52

PR #12647, opened 2018-09-24
NativeResource Management in Scala

nswamy:
## Description ##
This PR adds automatic Native Resource Management in Scala for resources such as NDArray, Symbol, Executor, Iterator, etc.,. 
User can use ResourceScope.using to automatically collect all the native resources and dispose them at the end of the code block or choose to either rely on Garbage Collector where the native resources are tracked using Phantom References and disposed when GC finds them not reachable. 

### Essentials ###

### Changes ###
- [X] NativeResource, tests, (and when applicable, API doc)
- [X] ResourceScope, tests, (and when applicable, API doc)


scala-package/init/src/main/scala/org/apache/mxnet/init/Base.scala
```diff
@@ -48,6 +48,7 @@ object Base {
     if (os.startsWith("Linux")) {
       System.load(s"$baseDir/linux-x86_64/target/libmxnet-init-scala-linux-x86_64.so")
     } else if (os.startsWith("Mac")) {
+      baseDir = "/Users/wamy/nswamy/deepengine/workspace/mxnet_scala/scala-package/init-native"
```
gigasquid: This just for testing?

stu1130:
Thanks for your contribution @nswamy 

scala-package/core/pom.xml
```diff
@@ -109,5 +109,12 @@
       <artifactId>commons-io</artifactId>
       <version>2.1</version>
     </dependency>
+    <!-- https://mvnrepository.com/artifact/org.mockito/mockito-all -->
```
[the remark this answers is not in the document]
nswamy: Yes

scala-package/init/src/main/scala/org/apache/mxnet/init/Base.scala
nswamy: Yes, I will remove in my next PR

scala-package/core/src/main/scala/org/apache/mxnet/NativeResource.scala
```diff
@@ -0,0 +1,193 @@
...
+    * the finalizer runs much later depending on the GC.
+    * @param resource resource to verify if it has been deAllocated
+    * @return true if already deAllocated
+    */
+  def isDeAllocated(ref: NativeResourceRef): Boolean = {
+    !refMap.containsKey(ref)
+  }
+
+  def cleanup: Unit = {
+    print("NativeResourceRef: cleanup\n")
```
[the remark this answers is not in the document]
nswamy: please ignore all print..its for my debugging they will be removed.

scala-package/core/src/main/scala/org/apache/mxnet/NDArray.scala
```diff
@@ -562,16 +562,20 @@ object NDArray extends NDArrayBase {
  */
 class NDArray private[mxnet](private[mxnet] val handle: NDArrayHandle,
                              val writable: Boolean = true,
-                             addToCollector: Boolean = true) extends WarnIfNotDisposed {
+                             addToCollector: Boolean = true) extends NativeResource {
```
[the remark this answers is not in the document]
nswamy: this is the right patten. NDArray `is a` NativeResource and all the functionality of NativeResource is provided through inheritance provided by NDArray.

scala-package/core/src/main/scala/org/apache/mxnet/NativeResource.scala
```diff
@@ -0,0 +1,193 @@
...
+    // phantomRef will be removed from the map when NativeResource.close is called.
+    val resource = refMap.get(ref)
+    if (resource != 0L)  { // since CPtrAddress is Scala a Long, it cannot be null
+      print("NativeResourceRef: got a reference for resource\n")
+      ref.resourceDeAllocator(resource)
+      refMap.remove(ref)
+    }
+  }
+
+  protected class ResourceCleanupThread extends Thread {
```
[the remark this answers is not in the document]
nswamy: read the comments. remove is a blocking call.

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
```diff
@@ -0,0 +1,84 @@
...
+       case _ => // do nothing
+     }
+     ret
+   } finally {
+     // TODO(nswamy@): handle exceptions
+     resource.close
+   }
+ }
+
+  private[mxnet] val resourceScope = new ThreadLocal[ArrayBuffer[ResourceScope]] {
```
larroy: shall it be some plural name? like resourceScopeRegistry as it's an array....?

scala-package/examples/scripts/run_train_mnist.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
larroy: #!/usr/bin/env bash -exuo pipefail

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
```diff
@@ -0,0 +1,195 @@
...
+    @inline def safeAddSuppressed(t: Throwable, suppressed: Throwable): Unit = {
+      if (!t.isInstanceOf[ControlThrowable]) t.addSuppressed(suppressed)
+    }
+
+    var retThrowable: Throwable = null
+
+    try {
+      val ret = body
+       ret match {
+          // don't de-allocate if returning any collection that contains NativeResource.
```
lanking520: A risk here is we cannot check if an object contains NativeResources

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
```diff
@@ -0,0 +1,195 @@
...
+
+  /**
+    * Get the Last but one Scope from threadLocal Scopes.
+    * @return n-1th scope or None when not found
+    */
+  private[mxnet] def getPrevScope(): Option[ResourceScope] = {
+    val scopes = threadLocalScopes.get()
+    Try(Some(scopes(scopes.size - 2))).getOrElse(None)
+  }
+}
```
lanking520: add skipline

scala-package/core/src/test/scala/org/apache/mxnet/ResourceScopeSuite.scala
```diff
@@ -0,0 +1,151 @@
...
+        ndList
+      }
+      ndListRet.foreach(nd => assert(nd.isDisposed == false,
+        "NativeResources within a returned collection should not be disposed"))
+    }
+    ndListRet.foreach(nd => assert(nd.isDisposed == true,
+    "NativeResources returned from inner scope should be disposed in outer scope"))
+  }
+
+  test("test native resource inside a map") {
```
lanking520: 👍🏿

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
nswamy: The only way I can think of is to do this check this by reflection, which might incur runtime costs, I will think of the best way and address this in the next PR. 
discussed with @frankfliu one option is to create an Annotation that the user will use on such objects if they do not want auto management using ResourceScope. Will address in one of the next PRs.

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
```diff
@@ -0,0 +1,195 @@
...
+    @inline def resourceInGeneric(g: scala.collection.Iterable[_]) = {
+      g.foreach( n =>
+        n match {
+          case nRes: NativeResource => {
+            removeAndAddToPrevScope(nRes)
+          }
+          case kv: scala.Tuple2[_, _] => {
+            if (kv._1.isInstanceOf[NativeResource]) removeAndAddToPrevScope(
+              kv._1.asInstanceOf[NativeResource])
+            if (kv._2.isInstanceOf[NativeResource]) removeAndAddToPrevScope(
```
piyushghai: Do these if's have to be mutually exclusive ? Or can both the if's be executed ?

scala-package/examples/scripts/run_train_mnist.sh
```diff
@@ -0,0 +1,33 @@
...
+set -e
+
+MXNET_ROOT=$(cd "$(dirname $0)/../../.."; pwd)
+echo $MXNET_ROOT
+CLASS_PATH=$MXNET_ROOT/scala-package/assembly/linux-x86_64-cpu/target/*:$MXNET_ROOT/scala-package/examples/target/*:$MXNET_ROOT/scala-package/examples/target/classes/lib/*:$MXNET_ROOT/scala-package/infer/target/*
+
+# model dir
+DATA_PATH=$2
+
+java -XX:+PrintGC -Xms256M -Xmx512M -Dmxnet.traceLeakedObjects=false -cp $CLASS_PATH \
```
piyushghai: Should this traceLeakedObjects flag be false ?

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
```diff
@@ -0,0 +1,195 @@
...
+        case nRes: NativeResource => removeAndAddToPrevScope(nRes)
+        case ndRet: NDArrayFuncReturn => ndRet.arr.foreach( nd => removeAndAddToPrevScope(nd) )
+        case _ => // do nothing
+      }
+      ret
+    } catch {
+      case t: Throwable =>
+        retThrowable = t
+        null.asInstanceOf[A] // we'll throw in finally
+    } finally {
```
andrewfayres: Why are we doing all this odd throwable stuff?

scala-package/examples/scripts/run_train_mnist.sh
nswamy: This is for testing, i'll fix this in the next commit

scala-package/core/src/test/scala/org/apache/mxnet/ResourceScopeSuite.scala
nswamy: thanks for helping with the syntax, need to do the same for Java Iterables.

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
nswamy: So, if the code block throws exception we want to still dispose all the NativeResources

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
nswamy: Not mutually exclusive, NativeResource can be in both Key and Value of the Map

scala-package/core/src/main/scala/org/apache/mxnet/NativeResource.scala
```diff
@@ -0,0 +1,193 @@
...
+    setPriority(Thread.MAX_PRIORITY)
+    setName("NativeResourceDeAllocatorThread")
+    setDaemon(true)
+
+    override def run(): Unit = {
+      while (true) {
```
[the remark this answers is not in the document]
nswamy: ReferenceQueue's remove is a blocking call, it is not spinning, the thread is blocked.

scala-package/core/src/main/scala/org/apache/mxnet/NDArray.scala
```diff
@@ -584,11 +588,10 @@ class NDArray private[mxnet](private[mxnet] val handle: NDArrayHandle,
    * The NDArrays it depends on will NOT be disposed. <br />
    * The object shall never be used after it is disposed.
    */
-  def dispose(): Unit = {
+  override def dispose(): Unit = {
     if (!disposed) {
```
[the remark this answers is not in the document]
nswamy: good point!, i'll change for all and make this variable private.

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  ResourceScope.addToLocalScope(this)
+
+  /**
+    * Releases all the [[NativeResource]] by calling
+    * the associated [[NativeResource.close()]] method
+    */
+  override def close(): Unit = {
+    resourceQ.foreach(resource => if (resource != null) resource.dispose(false) )
```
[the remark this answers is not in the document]
nswamy: because if you remove(from NativeResource) while you are iterating on the collection, the behavior of the iterator's.next is undefined, let me just refactor this part of the code since itsn't obvious

scala-package/core/src/main/scala/org/apache/mxnet/NativeResource.scala
```diff
@@ -0,0 +1,186 @@
...
+    */
+  def isDeAllocated(ref: NativeResourceRef): Boolean = {
+    !refMap.containsKey(ref)
+  }
+
+  def cleanup: Unit = {
+    // remove is a blocking call
+    val ref: NativeResourceRef = refQ.remove().asInstanceOf[NativeResourceRef]
+    // phantomRef will be removed from the map when NativeResource.close is called.
+    val resource = refMap.get(ref)
```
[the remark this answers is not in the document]
nswamy: there is a check see the next line 
```
    // phantomRef will be removed from the map when NativeResource.close is called.
    val resource = refMap.get(ref)
    if (resource != 0L)  { // since CPtrAddress is Scala a Long, it cannot be null
```

scala-package/core/src/main/scala/org/apache/mxnet/NativeResource.scala
```diff
@@ -0,0 +1,186 @@
...
+private[mxnet] class NativeResourceRef(resource: NativeResource,
+                                       val resourceDeAllocator: CPtrAddress => Int)
+        extends PhantomReference[NativeResource](resource, NativeResourceRef.refQ) {}
+
+private[mxnet] object NativeResourceRef {
+
+  private[mxnet] val refQ: ReferenceQueue[NativeResource]
+                = new ReferenceQueue[NativeResource]
+
+  private[mxnet] val refMap = new ConcurrentHashMap[NativeResourceRef, CPtrAddress]()
```
yzhliu: do we really need this Map? can we store resource.nativeAddress (not whole NativeResource.resource) as a member of NativeResourceRef?
My major concern is, given two thread-safe containers cooperate with each other, a small carelessness can result in concurrency disaster. `isDeAllocated` is not really necessary - if you're worry about `WarnIfNotDisposed`, can we simply remove it? since now everything is guaranteed to be disposed finally (we invented that because we removed `dispose()` in `finalize` at that time).

scala-package/core/src/main/scala/org/apache/mxnet/Model.scala
```diff
@@ -290,66 +290,69 @@ object Model {
...
-            updateParams(executorManager.paramArrays,
-              executorManager.gradArrays,
-              updaterLocal, ctx.length,
-              executorManager.paramNames,
-              kvStore)
-          }
-          monitor.foreach(_.tocPrint())
-          // evaluate at end, so out_cpu_array can lazy copy
-          executorManager.updateMetric(evalMetric, dataBatch.label)
+//      ResourceScope.using() {
```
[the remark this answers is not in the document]
nswamy: Yes, that is because executorManager.gradArrays/ executorManager.paramArrays all of these are initialized on the fly during initialization of trainExecs, they will deAllocated at the end of scope so  it crashes in the next epoch.

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
```diff
@@ -0,0 +1,196 @@
...
+    resourceQ.clear()
+    ResourceScope.removeFromLocalScope(this)
+  }
+
+  /**
+    * Add a NativeResource to the scope
+    * @param resource
+    */
+  def add(resource: NativeResource): Unit = {
+    resourceQ.+=(resource)
```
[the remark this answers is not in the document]
nswamy: Yeah I was calling it directly before, as I kept changing the data structure i had to go change lines of code and this abstraction helps with that.

scala-package/core/src/main/scala/org/apache/mxnet/NativeResource.scala
nswamy: This map is used to keep a reference to `NativeResourceRef` without which the GC will just finalize it

scala-package/core/src/main/scala/org/apache/mxnet/Model.scala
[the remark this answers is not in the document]
nswamy: I figure out the problem, its in the Optimizer createState. I'll find a way workaround it.

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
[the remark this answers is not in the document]
nswamy: I meant I need this abstraction

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
andrewfayres: It was more a question of why aren't we doing something like this instead.
```
} finally {
  try {
    curScope.close
  } catch {
    case closeThrowable: Throwable =>
      if (NonFatal(retThrowable) && !NonFatal(closeThrowable))
        retThrowable = closeThrowable
      else safeAddSuppressed(retThrowable, closeThrowable)
   } finally {
     throw retThrowable
   }
}
```
We call curScope.close regardless of the value of retThrowable so I don't think we need the if. NonFatal(null) returns true so the check to see which to throw should still work. I don't think you need the var toThrow because retThrowable can just be overwritten in the if statement if you decide to throw closeThrowable instead.

I'm on the fence here about whether we should just throw the first exception and not the most fatal exception. My intuition says that the first exception will almost certainly be causing the 2nd regardless of which is fatal.

nswamy: Lets say you get OOM(Native) from code block and then you get a error trying to close all of them, I would still want to throw the OOM exception from the block

scala-package/core/src/main/scala/org/apache/mxnet/NativeResource.scala
```diff
@@ -0,0 +1,189 @@
...
+    * @param resource resource to verify if it has been deAllocated
+    * @return true if already deAllocated
+    */
+  def isDeAllocated(ref: NativeResourceRef): Boolean = {
+    !refMap.containsKey(ref)
+  }
+
+  def cleanup: Unit = {
+    // remove is a blocking call
+    val ref: NativeResourceRef = refQ.remove().asInstanceOf[NativeResourceRef]
```
yzhliu: why is it `NativeResourceRef` while `refQ` is `ReferenceQueue[NativeResource]`, which should return `NativeResource`. did I miss something?

nswamy: The way PhantomReference works is, it is inserted to the Queue when the object(NativeResource) is not reachable after that point you cannot access the NativeResource using ref.get(), hence the NativeResourceRef(with NativeAddress and DeAllocator fn) is held in a HashMap
see this article 
https://medium.com/@ramtop/weak-soft-and-phantom-references-in-java-and-why-they-matter-c04bfc9dc792
--- from the article
since Phantom don’t have a link to the actual object, a typical pattern is to derive your own Reference type from Phantom and adding some info useful for the final free, for example filename.
---

nswamy:
Thanks for the reviews, merging this PR.. I'll cherry pick this into the java-api branch

PR #12666, opened 2018-09-25
[WIP] Horovod support for MXNet

ctcyang:
## Description ##
This PR contains the minimal set of changes required to support Horovod distributed training using MXNet.

### Essentials ###

### Changes ###
- [x] Horovod support

## Comments ##
- Design doc: https://cwiki.apache.org/confluence/display/MXNET/Horovod-MXNet+Integration
- Install readme: https://github.com/ctcyang/horovod/blob/mxnet_feature/examples/mxnet/README.md


vandanavk:
@mxnet-label-bot [pr-work-in-progress]

python/mxnet/module/executor_group.py
```diff
@@ -421,17 +421,31 @@ def get_params(self, arg_params, aux_params):
...
+        if copy_to_cpu:
+            for name, block in zip(self.param_names, self.param_arrays):
+                weight = sum(w.copyto(ctx.cpu()) for w in block) / len(block)
+                weight.astype(arg_params[name].dtype).copyto(arg_params[name])
+            for name, block in zip(self.aux_names, self.aux_arrays):
+                weight = sum(w.copyto(ctx.cpu()) for w in block) / len(block)
+                weight.astype(aux_params[name].dtype).copyto(aux_params[name])
+        else:
+            for name, block in zip(self.param_names, self.param_arrays):
+                context = block[0].context
```
apeforest: It seems this `else` block is identical to the `if` block except for the context. Can you add the if check to the context and change context between `block[0].context` and `ctx.cpu()`

python/mxnet/module/bucketing_module.py
```diff
@@ -161,7 +161,7 @@ def output_shapes(self):
         assert self.binded
         return self._curr_module.output_shapes
 
-    def get_params(self):
+    def get_params(self, copy_to_cpu=True):
```
apeforest: Why is this copy_to_cpu argument needed here since it is not used. Nor is it declared the BaseModule

python/mxnet/module/python_module.py
```diff
@@ -93,7 +93,7 @@ def output_shapes(self):
     ################################################################################
     # Parameters of a module
     ################################################################################
-    def get_params(self):
+    def get_params(self, copy_to_cpu=True):
```
apeforest: Why is this argument needed here?

python/mxnet/module/sequential_module.py
```diff
@@ -149,7 +149,7 @@ def output_shapes(self):
         assert self.binded
         return self._modules[-1].output_shapes
 
-    def get_params(self):
+    def get_params(self, copy_to_cpu=True):
```
apeforest: Why is this argument needed here?

src/common/cuda_utils.h
```diff
@@ -68,6 +68,110 @@ inline __device__ bool __is_supported_cuda_architecture() {
 }
 #endif  // __CUDACC__
 
+/*!
```
apeforest: I would prefer this trunk to be reviewed and merged separately in PR: https://github.com/apache/incubator-mxnet/pull/12031

apeforest:
Please add [MXNET-1111] to the PR title. Thanks

python/mxnet/module/bucketing_module.py
ctcyang: It's there, because it's needed to be called by the user here during MXNet training: https://github.com/ctcyang/horovod/blob/0deb0d5d5c6befda6bad49ff713a0e0061fc08da/examples/mxnet_imagenet_resnet50.py#L266

Horovod only supports (input on GPU, output on GPU) type of communication. However, we found that `get_params` returns it as an NDArray on the CPU by default. That is why we need to add this argument here to make Horovod work. Since the default is set to `True`, it maintains backward compatibility with the rest of MXNet code.

python/mxnet/module/executor_group.py
ctcyang: Fixed

src/io/iter_image_recordio_2.cc
```diff
@@ -285,9 +285,9 @@ inline bool ImageRecordIOParser2<DType>::ParseNext(DataBatch *out) {
     shape_vec.push_back(param_.label_width);
     TShape label_shape(shape_vec.begin(), shape_vec.end());
 
-    out->data.at(0) = NDArray(data_shape, Context::CPUPinned(0), false,
+    out->data.at(0) = NDArray(data_shape, Context::CPU(0), false,
```
yuxihu: For my own understanding, why do we change to CPU from CPUPinned in this PR? Is it going to cause issue if we stick to CPUPinned when conduct distributed training with Horovod?

ctcyang: In my benchmark, it doesn't affect performance. You may recall we can't use `horovod.local_rank()` to determine the GPU id because `mxnet` cannot be compile-time dependent on `horovod`. But then if you set all GPUs from 0-7 to use `CPUPinned(0)` you will get 8 processes starting the CUDA driver on GPU 0, which wastes memory and you won't be able to get the largest batch size.

python/mxnet/module/base_module.py
```diff
@@ -625,7 +625,7 @@ def output_shapes(self):
     ################################################################################
     # Parameters of a module
     ################################################################################
-    def get_params(self):
+    def get_params(self, copy_to_cpu=True):
```
yuxihu: For all the places you added this new param, could you please update the comments for the function to explain what this copy_to_cpu is for, like what you did in executor_group.py? May also be good to add some explanation why it is not used here and where it is going to be used. The readers might have the same confusion why the param is in function signature but not used in the implementation.

python/mxnet/module/sequential_module.py
yuxihu: Please check Line 167 in this file: arg, aux = module.get_params(). Do we need to pass in the copy_to_cpu value when calling get_params? arg, aux = module.get_params(copy_to_cpu)?

python/mxnet/module/bucketing_module.py
yuxihu: Please check Line 174 in this file: params = self._curr_module.get_params(). Do we need to change to params = self._curr_module.get_params(copy_to_cpu)?

python/mxnet/module/base_module.py
yuxihu: As we discussed offline, based on our tests of distributed training on 1 node and 2 node p3.16xlarge instances, the training accuracy looks normal. We also saved and compared the module parameters from a rank from each node in the 2 node scenario and they look identical. I think we can remove this copy_to_cpu argument change

ctcyang:
Tag @eric-haibin-lin as reviewer

src/io/iter_image_recordio_2.cc
eric-haibin-lin: @ctcyang what did you benchmark? Did you also run kvstore('local') and kvstore('nccl') to verify the perf impact?

src/kvstore/kvstore_nccl.h
```diff
@@ -64,7 +64,7 @@ class KVStoreNCCL : public KVStoreLocal {
   KVStoreNCCL() : KVStoreLocal() {
     // Due to aggregation, we do not use the Comm interface
     comm_ = nullptr;
-    pinned_ctx_ = Context::CPUPinned(0);
+    pinned_ctx_ = Context::CPU(0);
```
eric-haibin-lin: cc @ptrendx @DickJC123 is this ok to nccl?

ankkhedia:
@eric-haibin-lin Could you please take a look in the above PR?

kevinVegBird:
have the horovod merged into the mxnet1.4.0?why havent I saw any horovod in the mxnet 1.4.0，but the performance of mxnet1.4.0 is really improved compare to mxnet 1.3.0.

eric-haibin-lin:
Yes it was in 1.4 @kevinVegBird

Issue #12676, opened 2018-09-26
Mismatched Shapes in Value Error


ChaiBapchya:
## Description
Shapes given in the Value error seem to be mismatched between expected and got

Python

Compiler: visual studio

MXNet commit hash: 769165b1794f9067e42c65477b3c2fe91a606f89

## Error Message:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/bapac/workspace/incubator-mxnet/python/mxnet/ndarray/utils.py", line 146, in array
    return _array(source_array, ctx=ctx, dtype=dtype)
  File "/Users/bapac/workspace/incubator-mxnet/python/mxnet/ndarray/ndarray.py", line 2489, in array
    arr[:] = source_array
  File "/Users/bapac/workspace/incubator-mxnet/python/mxnet/ndarray/ndarray.py", line 444, in __setitem__
    self._set_nd_basic_indexing(key, value)
  File "/Users/bapac/workspace/incubator-mxnet/python/mxnet/ndarray/ndarray.py", line 710, in _set_nd_basic_indexing
    self._sync_copyfrom(value)
  File "/Users/bapac/workspace/incubator-mxnet/python/mxnet/ndarray/ndarray.py", line 872, in _sync_copyfrom
    str(self.shape), str(source_array.shape)))
ValueError: Shape inconsistent: expected () vs got (1,)
```

## Minimum reproducible example
`mx.nd.array(1)`

Correct way of initialization right now is to feed a list,
```
>>> a=mx.nd.array([1])
>>> a

[1.]
<NDArray 1 @cpu(0)>
```
Shape of such an NDarray is 
```
>>> a.shape
(1,)
```

Meaning, it expects (1,)
However, when we feed ()
It should give error 

`ValueError: Shape inconsistent: expected (1,) vs got ()`

vandanavk:
@mxnet-label-bot [NDArray]

ChaiBapchya:
@mxnet-label-bot [Bug]

vandanavk:
Fix in PR https://github.com/apache/incubator-mxnet/pull/12678 (Ref: https://github.com/apache/incubator-mxnet/pull/12677#issuecomment-424960050)

vrakesh:
@ChaiBapchya  can this be closed as well?

ChaiBapchya:
[...]

PR #12677, opened 2018-09-26
[MXNET-995] Constant Initializer for ND Array

### LINKED ISSUE #12672: Ndarray construction can't convert zero-dimensional numpy arrays
## Description

See example below

Ubuntu 18, python 3.6, pip-installed CPU-only version
MXNet 1.3.0

```python
mxnet.ndarray.array( numpy.arange(1).reshape([1, 1, 1]) ) # ok
mxnet.ndarray.array( numpy.arange(1).reshape([1, 1]) ) # ok
mxnet.ndarray.array( numpy.arange(1).reshape([1]) ) # ok
mxnet.ndarray.array( numpy.arange(1).reshape([]) )
# last one fails: ValueError: Shape inconsistent: expected () vs got (1,)
```

MXNet does support zero-dimensional arrays, so this is the problem of importing.

ChaiBapchya:
## Description ##
Allows 1D (constant) to be used to initialize an NDArray or Symbol
Fixes #12672 and #12676 
### Essentials ###

### Changes ###
- [ ] Checks if the input is a scalar of the type int/float and converts it into a list for NDArray
- [ ] Unit test for checking initialization with NDArray as well as Symbol API (using bind and eval functions)
- [ ] Mismatch of the Value error shapes changed

### DIFF
--- a/python/mxnet/ndarray/ndarray.py
+++ b/python/mxnet/ndarray/ndarray.py
@@ -869,7 +869,7 @@ fixed-size items.
         source_array = np.ascontiguousarray(source_array, dtype=self.dtype)
         if source_array.shape != self.shape:
             raise ValueError('Shape inconsistent: expected %s vs got %s'%(
-                str(self.shape), str(source_array.shape)))
+                str(source_array.shape), str(self.shape)))
         check_call(_LIB.MXNDArraySyncCopyFromCPU(
             self.handle,
             source_array.ctypes.data_as(ctypes.c_void_p),
@@ -2479,6 +2479,8 @@ def array(source_array, ctx=None, dtype=None):
     if isinstance(source_array, NDArray):
         dtype = source_array.dtype if dtype is None else dtype
     else:
+        if isinstance(source_array,(float,int)):
+            source_array=[float(source_array)]
         dtype = mx_real_t if dtype is None else dtype
         if not isinstance(source_array, np.ndarray):
             try:
--- a/python/mxnet/ndarray/ndarray.py
+++ b/python/mxnet/ndarray/ndarray.py
@@ -2479,8 +2479,8 @@ def array(source_array, ctx=None, dtype=None):
     if isinstance(source_array, NDArray):
         dtype = source_array.dtype if dtype is None else dtype
     else:
-        if isinstance(source_array,(float,int)):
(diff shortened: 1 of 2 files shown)


vandanavk:

python/mxnet/ndarray/ndarray.py
```diff
@@ -2479,6 +2479,8 @@ def array(source_array, ctx=None, dtype=None):
     if isinstance(source_array, NDArray):
         dtype = source_array.dtype if dtype is None else dtype
     else:
+        if isinstance(source_array, (float, int)):
```
zhreshold: mx.base.numeric_types is better

python/mxnet/ndarray/ndarray.py
```diff
@@ -869,7 +869,7 @@ def _sync_copyfrom(self, source_array):
         source_array = np.ascontiguousarray(source_array, dtype=self.dtype)
         if source_array.shape != self.shape:
             raise ValueError('Shape inconsistent: expected %s vs got %s'%(
-                str(self.shape), str(source_array.shape)))
+                str(source_array.shape), str(self.shape)))
```
zhreshold: why switch the order of expected vs got?

python/mxnet/ndarray/ndarray.py
zhreshold: my concern here is that you implicitly converted 0-d array(constant) to 1-d array here, which may not be very appropriate

arogozhnikov:
Please check that this case was resolved by PR #12678 too.

python/mxnet/ndarray/ndarray.py
ChaiBapchya: Ah, my logic was since MXNet NDArray needs a 1d array minimum, I should convert constant. Else, is there a way I can do constant initialization support? Would be great to get some help in that direction.

ChaiBapchya:
Ah, alright. So I'll take mine down then. Thanks for your contribution.

python/mxnet/ndarray/ndarray.py
ChaiBapchya: @zhreshold could you confirm if the above issue is right or not? coz the value error it gave sounded confusing due to mismatched shapes.

PR #12637, opened 2018-09-21
[MXNET-912] Refactoring ctc loss operator

apeforest:
## Description ##
The ctc_loss operator is widely used and has been stable in the past a few releases. This PR is to move this operator from contrib to MXNet regular and refactored the implementation using the new operator interface. 

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [X] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [X] Changes are complete (i.e. I finished coding on this PR)
- [X] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [X] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [X] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

## Comments ##
- Refactoring
- Updated unit test from mx.contrib.ctc_loss to mx.ctc_loss

PR #12687, opened 2018-09-27
ignore a 3rd party libray in cpplint

apeforest:
## Description ##
Need this to be merged so my other PR #12637 can pass CI lint.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -568,7 +568,7 @@ lint: cpplint rcpplint jnilint pylint
 
 cpplint:
 	3rdparty/dmlc-core/scripts/lint.py mxnet cpp include src plugin cpp-package tests \
-	--exclude_path src/operator/contrib/ctc_include
+	--exclude_path src/operator/contrib/ctc_include src/operator/nn/ctc_include
 
 pylint:
 	pylint --rcfile=$(ROOTDIR)/ci/other/pylintrc --ignore-patterns=".*\.so$$,.*\.dll$$,.*\.dylib$$" python/mxnet tools/caffe_converter/*.py


vandanavk:

Makefile
```diff
@@ -568,7 +568,7 @@ lint: cpplint rcpplint jnilint pylint
 
 cpplint:
 	3rdparty/dmlc-core/scripts/lint.py mxnet cpp include src plugin cpp-package tests \
-	--exclude_path src/operator/contrib/ctc_include
+	--exclude_path src/operator/contrib/ctc_include src/operator/nn/ctc_include
```
anirudh2290: Hi @apeforest, I think if we are adding it to src/operator/nn we should enforce the same lint checks enforced on rest of the code of MXNet. We can create exception for contrib but not very comfortable adding exception for src/operator/nn dir. I would like to hear other opinions.

apeforest: @anirudh2290 ctc_include is a 3rd party library and I plan to move it from src/operator/contrib/ctc_include to src/operator/nn/ctc_include in my PR #12637. I need this cpp lint to be excluded from Makefile before CI can pass. Thanks

anirudh2290: i understand that but i was concerned about adding 3rdparty library in src/operator dir when it doesnt conform to mxnet style guidelines. Also, what if more files were to be added to this dir in the future, which don't adhere to lint and not caught in review.

apeforest: Hi @szha, what's the guideline of moving an operator from contrib to regular and its 3rd party dependency? Thanks!

apeforest:
Moving the ctc_include to 3rdparty as @szha suggested. Closing this PR.

PR #12689, opened 2018-09-27
[WIP] Search update 

aaronmarkham:
## Description ##
This PR improves version switching on the website. The new feature helps with searching for APIs mostly, but can also work for pages you're searching for that exist in different versions.

The current version selector will kick you to the home page from the search page. I fixed this behavior for the API section of the docs last week, but thought fixing search would be a big UX improvement too.

It's generalized now too. 
* It should work for any section of the site. 
* It should retain both bookmarks `#` and query strings `?`.
* It is overridden by the .htaccess file, so sections like tutorials will still always go to the latest.

**UPDATE**: To deal with requests to the root of the website, I updated the redirects to point to the default version. Right now this means any request to / will go to /versions/master. Or requests to /tutorials/xyz go to /versions/master/tutorials/xyz.

## Usage
As this is a javascript update, please be sure to clear your cache.
Let's say you're looking for `allreduce`, but for whatever reason you're on the v1.0.0 of the site (like coming in from Google):
http://34.201.8.176/versions/1.0.0/search.html?q=allreduce&check_keywords=yes&area=default#
You get no results, but you're prompted to use the versions dropdown.
So you choose 1.3.0:
http://34.201.8.176/versions/1.3.0/search.html?q=allreduce&check_keywords=yes&area=default
You get a result.

## Comment
It would be awesome if the search fails we have it return other versions as suggestions, but I'm not that familiar yet with how Sphinx does its search magic... so that can be another day!

## Test Plan
@safrooze found a hole in the regex coverage, so I think it is worth identifying the various angles of entry and usage.
* Home page redirects to default version (master): http://34.201.8.176/
* Old deep links to resources are redirected to default version: http://34.201.8.176/tutorials
* Deep links to resources on specific versions are accepted: http://34.201.8.176/versions/1.2.1/tutorials/index.html (this is broken unfortunately, so back to the regexes)

### DIFF
--- a/docs/_static/js/docversion.js
+++ b/docs/_static/js/docversion.js
@@ -1,5 +1,5 @@
 function setVersion(){
-        let doc = window.location.pathname.match(/^\/(api\/.*)$/) || window.location.pathname.match(/^\/versions\/[^*]+\/(api\/.*)$/);
+        let doc = window.location.pathname.match(/^\/versions\/[0-9.master]+\/([^*]+.*)$/);
         if (doc) {
             if (document.getElementById('dropdown-menu-position-anchor-version')) {
                     versionNav = $('#dropdown-menu-position-anchor-version a.main-nav-link');
@@ -7,11 +7,11 @@ function setVersion(){
                             currLink = $( el ).attr('href');
                             version = currLink.match(/\/versions\/([0-9.master]+)\//);
                             if (version) {
-                                    versionedDoc = '/versions/' + version[1] + '/' + doc[1] + (window.location.hash || '');
+                                    versionedDoc = '/versions/' + version[1] + '/' + doc[1] + (window.location.hash || '') + (window.location.search || '');
                                     $( el ).attr('href', versionedDoc);
                             }
                     });
-            }        
+            }
         }
 }
 
--- a/docs/_static/searchtools_custom.js
+++ b/docs/_static/searchtools_custom.js
@@ -570,7 +570,7 @@ var Search = {
         }
         Search.title.text(_('Search Results'));
         if (!resultCount)
-          Search.status.text(_('Your search did not match any documents. Please make sure that all words are spelled correctly and that you\'ve selected enough categories.'));
+          Search.status.text(_('Your search did not match any documents in this version of the documentation. You can use the dropdown selector in the navigation bar to try another version. Please make sure that all words are spelled correctly and that you\'ve selected enough categories.'));
         else
             Search.status.text(_('Search finished, found %s page(s) matching the search query.').replace('%s', resultCount));
         Search.status.fadeIn(500);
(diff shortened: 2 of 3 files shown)


vandanavk:

docs/_static/js/docversion.js
```diff
@@ -1,17 +1,17 @@
 function setVersion(){
-        let doc = window.location.pathname.match(/^\/(api\/.*)$/) || window.location.pathname.match(/^\/versions\/[^*]+\/(api\/.*)$/);
+        let doc = window.location.pathname.match(/^\/versions\/[0-9.master]+\/([^*]+.*)$/);
```
safrooze: Wouldn't this fail if the version user is looking at is master, where `/versions/` is missing from the URL?

aaronmarkham: Yep... good catch... I was testing the other side of the use case. 

I've been thinking about a related point. I could just force all the default traffic to /versions/master. This would solve a variety of thorny issues. Also, help with analytics. WDYT about that?

safrooze: Yeah I think I'd personally like to see `/versions/xyz` in the URL all the time.  In the future if we decide to change default to latest release rather than master, we can easily change the redirect and everything else should work as expected.

sandeep-krishnamurthy:
@safrooze @aaronmarkham - ping for next steps here

aaronmarkham:
@sandeep-krishnamurthy @safrooze I've re-enabled the preview. You should clear your cache to make sure you're getting the latest .htaccess file.
This was pretty challenging to resolve and I hope I have the different angles covered with:
https://github.com/apache/incubator-mxnet/pull/12689/files#diff-8752dbf83f90b837af0035f3c1291ecdR31
That should redirect traffic to the root to master. Requests to other versions go through. I added an env var to htaccess to hold the default version, which is currently 'master'.
I kind of wish I wasn't combining the search fix with this kind of structure fix, but anyway, here we are. Please test it out.

ankkhedia:
@safrooze Could you please take a look into this PR?

kalyc:
@aaronmarkham could you please elaborate more on the issue by providing an example? You can consider addressing the regex issue with another PR and have this PR move ahead.

aaronmarkham:
There are two competing routes to make this move forward. 
The first is to add a second match to the regex so that anything /not-in-versions/ folder works, [as mentioned by Sina](https://github.com/apache/incubator-mxnet/pull/12689#discussion_r221328455).
The second, and the route I started to take here, is to make sure everything is in /versions/. That requires a redirect on anything not in versions. Smartly done, it'll take your request to /api and move it the /versions/master/api, where master is what is set to be the default in htaccess and any root requests are forwarded... but it's not working...:
```
RewriteRule .* - [E=default_version:/versions/master]
# other rules
...
# finally, check and forward
# Redirect any root requests to the default version
RewriteRule ^(?!versions\/[0-9.master]+\/)(.*)$ %{ENV:default_version}/$1 [R=301,NC,L]
```

stu1130:
@aaronmarkham are you able to fix the issue you mentioned above?
If it will take you a long time, you can create a separate PR for this so that we can merge this one

roywei:
@aaronmarkham ping for update, thanks!

aaronmarkham:
Closing this in favor of tackling the problems separately. The site should be updated to have traffic directed to the default version. Then we can improve search.

PR #12690, opened 2018-09-27
[MXNET-1000] get Ndarray real value and form it from a NDArray

lanking520:
## Description ##
This is an implementation of visualizing the internal structure of NDArray as well as forming an NDArray from a multi-dimensional array.

**`visualize`**: Shown the structure of the current NDArray, it is also able to visualize some large NDArray

**`toNDArray`**: Convert a Array[Array[....Array[MXNET_PRIMITIVE]...]] matrix into NDArray. Flatten and fill in a long array

@nswamy @andrewfayres @yzhliu @gigasquid  @ustcfd 

### Essentials ###


vandanavk:

scala-package/core/src/main/scala/org/apache/mxnet/NDArray.scala
```diff
@@ -197,6 +197,15 @@ object NDArray extends NDArrayBase {
       "_onehot_encode", Seq(indices, out), Map("out" -> out))(0)
   }
 
+  /**
+    * Get the String representation of NDArray
+    * @param nd input NDArray
+    * @return String
+    */
+  def toString(nd : NDArray) : String = {
+    nd.visualize
```
gigasquid: I want to test this out through the Clojure api. We do a lot of work in the REPL and with a default visualize in the .toString method I want to make sure the usability is not surprising, (especially with large arrays).

I'll take a look and let you know

nswamy:
I am not comfortable adding this API, an analogy to NDArray is BufferedImage, you don't print all the pixel values of a BufferedImage in toString. I understand it helps with debugging, however i would expect the developer/user to use a debugger or override toString then.

lanking520:
> I am not comfortable adding this API, an analogy to NDArray is BufferedImage, you don't print all the pixel values of a BufferedImage in toString. I understand it helps with debugging, however i would expect the developer/user to use a debugger or override toString then.

@nswamy two things I would like to mention:
- the visualize method does not override `toString`.
- the visualize method can smartly handle the large NDArray case. Please see my unittest examples. It would ignore most of the content and only print a part of it. It implementation is very similar to ndarray print function in Python

nswamy:
IMO, It still doesn't warrant to add this , whether you call it visualize or override toString. Another thing its confusing user experience that you are putting arbitrary THRESHOLD values?. what makes these values to be correct, why not more or less. 
```	    val THRESHOLD = 100000      // longest NDArray to show in full
    val ARRAYTHRESHOLD = 1000   // longest array to show in full 
```

lanking520:
> IMO, It still doesn't warrant to add this , whether you call it visualize or override toString. Another thing its confusing user experience that you are putting arbitrary THRESHOLD values?. what makes these values to be correct, why not more or less.
> 
> ```
>     val ARRAYTHRESHOLD = 1000   // longest array to show in full 
> ```
@nswamy since this is just a helper tool for users to debug and see what content inside. I think it's fair to have this tool. This Threshold value is just a limitation on numbers of elements to be shown in the array. The reason I set it as 1000 because 10000 sometimes would crash the JVM because it used too much memory and it's making no sense for a user to visualize that much. If the user seriously wants something to show up, he/she should at least try to slice and get it visualized. We can get this threshold configurable and ask the user to pass in. I am also curious how numpy set up this number in Python. If I can find a number in there I will let you know.

lanking520:
Link to the discussion we discussed before: https://github.com/apache/incubator-mxnet/pull/12536

yzhliu:
Any reason why we should not name the function `toString`? @nswamy @gigasquid Does clojure pkg so far has similar thing that may conflict with this?

nswamy:
quoting from Oracle's toString https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#toString()
```
Returns a string representation of the object. In general, the toString method returns a string that
"textually represents" this object. The result should be a **concise but informative representation**
that is easy for a person to read. It is recommended that all subclasses override this method.
```
It has to be concise: It might be helpful to print the metadata of the NDArray(Shape, DType,  ...).

yzhliu:
I agree. Can we modify the current impl so that it can reach the toString bar?

lanking520:
@nswamy @yzhliu @gigasquid WDYT?

yzhliu:
@nswamy Any thoughts? If it is for pretty-printing ndarray, I think it is suitable to put into toString. As for now we do not have a good toString for NDArray.

lanking520:
As offline discussion with @nswamy, the visualize method will go to a Util class. The reason behind it does not quite explain the idea behind NDArray. We can take it back to `toString` if anytime in the future when this method is more stable.

yzhliu:
would you mind describe at what extent, it describes the idea behind NDArray? I think such a small feature can be easily made correctly at the very first time.

lanking520:
@yzhliu here is the points:

We should not place it at the very core place if we are not totally confident about its functionality.  We should place it elsewhere until we are confident enough user likes it and it is not buggy.
This method is not fully tested so we are using users for the experiment. Apart from that, we are more responsible for all public methods we introduce because the future maintenance for it is big.

yzhliu:
My point is, if we think the method is buggy, then we should not place it anywhere - why create a new Class which for sure will be deprecated later?
I don't see the point why we cannot make it correct now.

So, do we agree, NDArray needs an implement of `toString`? If so, can we spend some efforts make the functionality correct for `toString`?

yzhliu:
So far what need to be improved:
- @nswamy 's suggestion: print also Shape and DType.
- Mine: too many lines printed, suggest to follow numpy's limitation.
Anything else?

nswamy:
@yzhliu my concerns are: 
1) that we are exposing helper utility functions as public API. I am not convinced how printing binary float values would make sense except when you probably are dealing with a 2 or 3 dimensional tensor when axes have small values, i don't think I can find a difference if there is a pixel value different in [3x224x224] Image represented in NDArray.

2) Typically Java object's toString() do not print the binary content of an object - if you print binary content IMO could be very surprising to users. I gave an analogy above how BufferedImage does not print its pixel values in toString. 
Hence I suggested to only print the metadata of NDArray in toString() and move this helper to a utility class.

Lastly look at what other frameworks are doing in their Tensor's toString()
https://www.tensorflow.org/api_docs/java/reference/org/tensorflow/Tensor.html#toString()

lanking520:
I think Tensor is an ideal translation to MXNet Symbol since Tensorflow itself is a symbolic language. It did not contain real-time data when you compute at its stage so showing the detailed information seemed impossible.

I agree with you on showing the entire image buffered data is not necessary unless the user really want to do so.

yzhliu:
Some examples do print values:
ND4J:
https://github.com/deeplearning4j/deeplearning4j/blob/master/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseNDArray.java#L5751

numsca (basically wrapper of ND4J):
https://github.com/botkop/numsca/blob/master/src/main/scala/botkop/numsca/Tensor.scala#L214

jblas:
https://github.com/botkop/numsca/blob/master/src/main/scala/botkop/numsca/Tensor.scala#L214

I agree no one can tell whether two big tensors are the same by eyeball, but it is useful for small tensors, and helpful to things like checking scale of values, layout, etc - This is what we usually do with numpy's pretty-print arrays, it's not precise, but valuable.
If mxnet python print ndarray values, how would users be surprised to see scala does the same thing?

How about we do this, make Visualize.toString have params rowLimit and colLimit, and have NDArray.toString invoke with a small number.

ankkhedia:
@lanking520 ping! Were you able to figure out a way forward here?

lanking520:
@gigasquid Could you please provide some opinions?

Other two Scala committers from the `Contributor.md`: @yanqingmen @terrytangyuan

terrytangyuan:
I agree that we should not print the entire data. Rather we should let users to specify how much to print. I am into Spark's way of doing this: `def toString(maxLines: Int, maxLineWidth: Int)`

gigasquid:
Another reference is from https://docs.scipy.org/doc/numpy/reference/generated/numpy.set_printoptions.html
The default is 1000 - but you can override that with options

I do like the way mxnet python also includes the dtype 
https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/ndarray/ndarray.py#L412

yzhliu:
@terrytangyuan @gigasquid Thanks for comment. I think we all agree that we should not print the entire data. What we haven't reached consensus is whether to have `NDArray.toString()` print a subset of the data values. basically what I proposed is 
```scala
>>> nd.toString()
<NDArray [xxx, xxx, ... xxx] Shape Context Dtype>
```

@nswamy 's proposal is
```scala
>>> nd.toString()
<NDArray Shape Context Dtype>

>>> Visualize.toString(nd)
<NDArray [xxx, xxx, ... xxx] Shape Context Dtype>
```

terrytangyuan:
@yzhliu Thanks for the clarifications. I am voting for the first option since I think it’s more useful and intuitive to users.

gigasquid:
I also think the first option is more intuitive/useful and like mxnet python

lanking520:
Thank you all for your inputs @yzhliu @nswamy @gigasquid @terrytangyuan . Since the majority vote appears in the `nd.toString`, I would transfer my design back to there any apply the limitation on the data to print. I would follow what @gigasquid proposed to set a limitation on the total data and also @terrytangyuan with an option to override the `toString` option.

kalyc:
Hi @lanking520 thanks for your contributions. Could you update the PR as per the discussions above?

stu1130:
@lanking520 ping again. feel free to change the label if I am wrong

lanking520:
> @mxnet-label-bot update [scala, pr-work-in-progress]
> @lanking520 ping again. feel free to change the label if I am wrong

Thanks, you are right about it

roywei:
@lanking520 ping for update, thanks!

sandeep-krishnamurthy:
@lanking520 - ping :-) Looks like a useful utility for debugging purposes.

lanking520:
Indeed, will update when I back from Vacation

lanking520:
Make this back to life now. Updated with the changes in the comments and it should print you something good for now. Please take a look.

lanking520:
@zachgk @piyushghai @andrewfayres Please take a look at this if you have time

scala-package/core/src/main/scala/org/apache/mxnet/NDArray.scala
```diff
@@ -509,6 +510,57 @@ object NDArray extends NDArrayBase {
...
+    */
+  def toNDArray(sourceArr: Array[_], ctx : Context = null) : NDArray = {
+    val shape = ArrayBuffer[Int]()
+    shapeGetter(sourceArr, shape, 0)
+    val finalArr = new Array[Float](shape.product)
+    arrayCombiner(sourceArr, finalArr, 0, finalArr.length - 1)
+    array(finalArr, Shape(shape), ctx)
+  }
+
+  private def shapeGetter(sourceArr : Any,
```
zachgk: I get using an ArrayBuffer for the array data since it is large. Don't do it for the shape because it makes everything complicated.

scala-package/core/src/main/scala/org/apache/mxnet/NDArray.scala
```diff
@@ -694,6 +745,12 @@ class NDArray private[mxnet](private[mxnet] val handle: NDArrayHandle,
   // we use weak reference to prevent gc blocking
   private[mxnet] val dependencies = mutable.HashMap.empty[Long, WeakReference[NDArray]]
 
+  private val traceProperty = "mxnet.setNDArrayPrintLength"
+  private lazy val printLength = {
+    val value = Try(System.getProperty(traceProperty).toInt).getOrElse(1000)
```
zachgk: Do you need the intermediate val value or can this be just a single line?

scala-package/core/src/test/scala/org/apache/mxnet/NDArraySuite.scala
```diff
@@ -85,6 +89,31 @@ class NDArraySuite extends FunSuite with BeforeAndAfterAll with Matchers {
...
+    var nd = NDArray.toNDArray(arr)
+    require(nd.shape == Shape(2, 1, 100, 4))
+    val arr2 = Array(1.0f, 1.0f, 1.0f, 1.0f)
+    nd = NDArray.toNDArray(arr2)
+    require(nd.shape == Shape(4))
+  }
+
+  test("test Visualize") {
+    var nd = NDArray.ones(Shape(1, 2, 1000, 1))
+    logger.info(s"Test print large ndarray:\n$nd")
```
zachgk: Can we make this test more automatic? Maybe test whether it contains the full output or not

scala-package/core/src/main/scala/org/apache/mxnet/MX_PRIMITIVES.scala
```diff
@@ -82,4 +82,10 @@ object MX_PRIMITIVES {
 
   implicit def MX_DoubleToDouble(d: MX_Double) : Double = d.data
 
+  def isValidType(num : Any) : Boolean = {
```
piyushghai: Can this be renamed to ```isValidMxPrimitiveType```

scala-package/core/src/main/scala/org/apache/mxnet/NDArray.scala
```diff
@@ -509,6 +510,63 @@ object NDArray extends NDArrayBase {
...
+    */
+  def toNDArray(sourceArr: Array[_], ctx : Context = null) : NDArray = {
+    val shape = ArrayBuffer[Int]()
+    shapeGetter(sourceArr, shape, 0)
+    val container = new Array[Any](shape.product)
+    arrayCombiner(sourceArr, container, 0, container.length - 1)
+    val finalArr = container(0) match {
+      case f: Float => array(container.map(_.asInstanceOf[Float]), Shape(shape), ctx)
+      case d: Double => array(container.map(_.asInstanceOf[Double]), Shape(shape), ctx)
+      case _ => throw new IllegalArgumentException(s"Unsupported type ${container(0).getClass}")
```
piyushghai: Should we add to the method docs as well that it will throw in an IllegalArgumentException ?

piyushghai: Should the exception message indicate that the supported types are Float/Double ?

scala-package/core/src/main/scala/org/apache/mxnet/NDArray.scala
```diff
@@ -763,6 +823,56 @@ class NDArray private[mxnet](private[mxnet] val handle: NDArrayHandle,
...
+  /**
+    * Helper function to create formatted NDArray output
+    * The NDArray will be represented in a reduced version if too large
+    * @param nd NDArray as the input
+    * @param totalSpace totalSpace of the lowest dimension
+    * @return String format of NDArray
+    */
+  private def buildStringHelper(nd : NDArray, totalSpace : Int) : String = {
+    var result = ""
+    val THRESHOLD = 10      // longest NDArray[NDArray[...]] to show in full
```
piyushghai: Can this be MAX_SIZE_THRESHOLD ?

scala-package/core/src/test/scala/org/apache/mxnet/NDArraySuite.scala
```diff
@@ -85,6 +89,47 @@ class NDArraySuite extends FunSuite with BeforeAndAfterAll with Matchers {
...
+    require(nd.shape == Shape(4))
+    val doubleData = 1.0d
+    nd = NDArray.toNDArray(arrayGen(doubleData))
+    require(nd.shape == Shape(2, 1, 100, 4))
+    require(nd.dtype == DType.Float64)
+  }
+
+  test("test Visualize") {
+    var nd = NDArray.ones(Shape(1, 2, 1000, 1))
+    require(nd.toString.split("\n").length == 33)
```
piyushghai: Should we compare the string representations as well ?

scala-package/core/src/main/scala/org/apache/mxnet/NDArray.scala
```diff
@@ -509,6 +510,63 @@ object NDArray extends NDArrayBase {
...
+        }
+        require(shape(shapeIdx) == arrLength,
+          s"Each Array should have equal length, expected ${shape(shapeIdx)}, get $arrLength")
+        arr.foreach(ele => shapeGetter(ele, shape, shapeIdx + 1))
+      }
+      case _ => throw new IllegalArgumentException(s"Wrong type passed: ${sourceArr.getClass}")
+    }
+  }
+
+  private def arrayCombiner(sourceArr : Any, arr : Array[Any],
```
yzhliu: better named flattenArray

scala-package/core/src/main/scala/org/apache/mxnet/NDArray.scala
yzhliu: how about System.getProperty so that it is configurable

lanking520:
@zachgk Please review again

PR #12733, opened 2018-10-03
Throw exception if MXSymbolInferShape fails.

csm:
* scala-package/native/src/main/native/org_apache_mxnet_native_c_api.cc:
  (Java_org_apache_mxnet_LibInfo_mxSymbolInferShape): throw
  IllegalArgumentException with the content of MXGetError if call to
  MXSymbolInferShape fails.

## Description ##
Properly handle any errors returned from `MXSymbolInferShape`, by throwing an exception to the caller. Otherwise, an error in that function call would mean subsequent calls (e.g. FillSymbolInferShape) would be operating on invalid, uninitialized data.

### DIFF
--- a/scala-package/native/src/main/native/org_apache_mxnet_native_c_api.cc
+++ b/scala-package/native/src/main/native/org_apache_mxnet_native_c_api.cc
@@ -1581,6 +1581,18 @@ JNIEXPORT jint JNICALL Java_org_apache_mxnet_LibInfo_mxSymbolInferShape
   env->ReleaseIntArrayElements(jargShapeData, argShapeData, 0);
   env->ReleaseIntArrayElements(jargIndPtr, argIndPtr, 0);
 
+  if (ret != 0) {
+    if (jkeys != NULL) {
+      for (int i = 0; i < jnumArgs; i++ ) {
+        jstring jkey = reinterpret_cast<jstring>(env->GetObjectArrayElement(jkeys, i));
+        env->ReleaseStringUTFChars(jkey, keys[i]);
+        env->DeleteLocalRef(jkey);
+      }
+    }
+    jclass illArgClass = env->FindClass("java/lang/IllegalArgumentException");
+    return env->ThrowNew(illArgClass, MXGetLastError());
+  }
+
   jclass listClass = env->FindClass("scala/collection/mutable/ListBuffer");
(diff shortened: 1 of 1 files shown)


vrakesh:

gigasquid:
Thanks @csm for submitting the fix!  💯  (The original issue was found using the Clojure api)

scala-package/native/src/main/native/org_apache_mxnet_native_c_api.cc
```diff
@@ -1581,26 +1581,29 @@ JNIEXPORT jint JNICALL Java_org_apache_mxnet_LibInfo_mxSymbolInferShape
...
-        env, listAppend, joutShapeData, outShapeSize, outShapeNdim, outShapeData)) {
-    // TODO(Yizhi): out of memory error thrown, return a specific error code ?
-    return -1;
-  }
-  if (FillSymbolInferShape(
-        env, listAppend, jauxShapeData, auxShapeSize, auxShapeNdim, auxShapeData)) {
-    // TODO(Yizhi): out of memory error thrown, return a specific error code ?
-    return -1;
-  }
+    if (FillSymbolInferShape(
```
lanking520: please don't change the code ident if you are not changing the content inside

nswamy:
@csm thanks for the contribution and welcome to MXNet. this is cool and LGTM. We need to find a way to add tests to these JNI code.

scala-package/native/src/main/native/org_apache_mxnet_native_c_api.cc
nswamy: What @lanking520 is saying is that we follow the [style-guide](https://github.com/apache/incubator-mxnet/blob/master/scala-package/scalastyle-config.xml) and request not make whitespace changes.

csm: Hi! This change placed these three `if (FillSymbolInferShape...` calls inside a new block, the `if (ret == 0) {` now on line 1584. Are you saying indentation shouldn't be changed even if a new scope is introduced around the code?

As for inserting the line break on the first of those calls, without that, your own style checker failed because that line was > 100 characters. I followed the convention with the rest of this file to make that line shorter.

Am I missing something, or do any of these changes actually violate your style guide?

PR #12746, opened 2018-10-05
Add env flag to disable MKLDNN cache (MXNET_MKLDNN_CACHE_ENABLED)

azai91:
## Description ##
MKLDNN by default caches mkldnn memory and primitives to performance reasons. The cache looks uses input shape as a key. However, when models accept variable size inputs, the cache constantly grows and is not cleared. A flag is added MXNET_MKLDNN_CACHE_ENABLED that will disable this.

In a future PR, we can clear the cache automatically if it growths a certain length.

### Essentials ###

### Changes ###
- [ ] Add flag (MXNET_MKLDNN_CACHE_ENABLED) to disable cache.

## Comments ##


vrakesh:

ZhennanQin:
If cache is disabled, shouldn't we skip hashing key as well? As hashing key is useless and wasting time.

docs/faq/env_var.md
```diff
@@ -181,6 +181,10 @@ When USE_PROFILER is enabled in Makefile or CMake, the following environments ca
   - Values: 0, 1 ```(default=1)```
   - Flag to enable or disable MKLDNN accelerator. On by default.
   - Only applies to mxnet that has been compiled with MKLDNN (```pip install mxnet-mkl``` or built from source with ```USE_MKLDNN=1```)
+  
+* MXNET_MKLDNN_CACHE_ENABLED
+  - Values: 0, 1 ```(default=1)```
+  - Flag to enable or disable MKLDNN cache. On by default. This should be disabled your dataset has variable size inputs.
```
lupesko: I suggest we document something a bit more verbose: "Flag to enable or disable MKLDNN cache. It is on by default to optimize for performance. Should be disabled if your model has variable input shapes, as cache size may grow unbounded."

lupesko:
@azai91 it will be difficult for most users to know they have to turn this env var off in case their models have variable input shapes. I think we need a better solution, once that does not require setting an env var, such as the follow up PR you mentioned (cache pruning) or maybe identifying a different cache key (not input shape).

azai91:
@mseth10

docs/faq/env_var.md
```diff
@@ -191,6 +191,10 @@ When USE_PROFILER is enabled in Makefile or CMake, the following environments ca
   - Values: 0, 1 ```(default=1)```
   - Flag to enable or disable MKLDNN accelerator. On by default.
   - Only applies to mxnet that has been compiled with MKLDNN (```pip install mxnet-mkl``` or built from source with ```USE_MKLDNN=1```)
+  
+* MXNET_MKLDNN_CACHE_SIZE
+  - Values: Int ```(default=-1)```
```
lupesko: Why not set to some reasonable default, so that users with variable input shape will not suffer from OOM?

src/operator/nn/mkldnn/mkldnn_base-inl.h
```diff
@@ -147,6 +147,18 @@ static inline bool MKLDNNEnvSet() {
   return is_mkldnn_enabled;
 }
 
+static inline int GetMKLDNNSize() {
+  static int mkldnn_cache_size = dmlc::GetEnv("MXNET_MKLDNN_CACHE_SIZE", -1);
+  return mkldnn_cache_size;
+}
+
+// TODO(alex): (MXNET-1075) Will remove env variable and calculate cache size during runtime
```
lupesko: 👍🏼

lupesko:
@anirudh2290 @eric-haibin-lin @mseth10

docs/faq/env_var.md
ZhennanQin: It's hard to find a reasonable default for common usage, because it depends on the topology depth. Currently, convolution is defined as stateless op, making all convolution layers share same mkldnn primitive cache pool for reusing purpose. So extremely, for enabling caching for the topology which has 1000 convolution layers with different parameters and data shapes, MXNET_MKLDNN_CACHE_SIZE should be larger than 1000. As time moves, we can imagine that the depth of topology will continue growing.

src/operator/nn/mkldnn/mkldnn_act.cc
```diff
@@ -147,6 +147,8 @@ static MKLDNNActForward &GetActForward(const ActivationParam& param,
   auto it = fwds.find(key);
   if (it == fwds.end()) {
     MKLDNNActForward fwd(param, ctx.is_train, in_data, in_mem);
+    if (MKLDNNCacheExceeded(fwds.size()))
+      fwds.clear();
```
ZhennanQin: Shall we just randomly remove one from fwds and insert the new one? It's more reasonable to remove the first inserting one that not be reused, but it's hard to achieve. Randomly removing is acceptable. Our target is balancing cache size with performance. Flushing cache isn't the best choice.

ZhennanQin: Shall we just randomly remove one from fwds and insert the new one? It's more reasonable to remove the first inserting one that not be reused, but it's hard to achieve. Randomly removing is acceptable. Our target is balancing cache size with performance. Flushing cache isn't the best choice.

ZhennanQin:
LGTM. Thanks for enabling this feature.

src/operator/nn/mkldnn/mkldnn_base-inl.h
```diff
@@ -147,6 +147,23 @@ static inline bool MKLDNNEnvSet() {
...
+template<typename S, typename I, typename H>
+static typename std::unordered_map<S, I, H>::iterator AddToCache(
+    std::unordered_map<S, I, H>* cache, const S &key, const I &item) {
+  int mkldnn_cache_size = GetMKLDNNCacheSize();
+  if (mkldnn_cache_size != -1 && static_cast<int>(cache.size()) > mkldnn_cache_size)
+    cache->erase(cache.begin());
```
TaoLv: This is a little crude to me. Can we have a LRU algorithm here? https://en.wikipedia.org/wiki/Cache_replacement_policies#LRU

ZhennanQin: There're 2 approach reasonable to me.
1. Permanently disable cache and remove the overhead of key hashing. By this way, we can get best performance and minimal memory usage for the cache always miss scenarios.
2. Maintain cache size as user expected, and kick off an old one when reaching the memory limitation. Then we can expect cache may hit in future. Just like what cpu cache behaves. 

I don't know why you said 

> if the future inputs are consistently not in the cache then we need to constant remove one element every time which is expensive.

Removing an item from hash map has constant time complexity, which is much less than building key hashing. If you think it's expensive, then we should move to first approach(not hashing key at all).

src/operator/nn/mkldnn/mkldnn_base-inl.h
```diff
@@ -147,6 +148,61 @@ static inline bool MKLDNNEnvSet() {
...
+//  int mkldnn_cache_size = GetMKLDNNCacheSize();
+//  if (mkldnn_cache_size != -1 && static_cast<int>(cache.size()) > mkldnn_cache_size)
+//    cache.erase(cache.begin());
+//  auto ins_return = cache.insert(std::pair<S, I>(key, item));
+//  CHECK(ins_return.second);
+//  return ins_return.first;
+//}
+
+template<typename K, typename V, typename H>
+class MKLDNNCache {
```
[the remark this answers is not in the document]
ZhennanQin: Measure latency change as well.

azai91:
Master branch throughput on Resnet 50 batch size 32 on c5
ubuntu@ip-172-31-3-217:~/incubator-mxnet-2$ python example/image-classification/benchmark_score.py
/home/ubuntu/anaconda3/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
Assertion failure at kmp_runtime.cpp(2582): __kmp_init_serial.
INFO:root:network: resnet-50
INFO:root:device: cpu(0)
/home/ubuntu/incubator-mxnet-2/python/mxnet/module/base_module.py:68: UserWarning: Data provided by label_shapes don't match names specified by label_names ([] vs. ['softmax_label'])
  warnings.warn(msg)
[20:19:14] ../src/operator/nn/mkldnn/mkldnn_base.cc:74: Allocate 147456 bytes with malloc directly
[20:19:14] ../src/operator/nn/mkldnn/mkldnn_base.cc:74: Allocate 589824 bytes with malloc directly
[20:19:15] ../src/operator/nn/mkldnn/mkldnn_base.cc:74: Allocate 2359296 bytes with malloc directly
[20:19:15] ../src/operator/nn/mkldnn/mkldnn_base.cc:74: Allocate 9437184 bytes with malloc directly
INFO:root:batch size 32, dtype float32, images/sec: 107.744738

wikier:
Confirmed the new env variable helps to mitigate the memory issue we've seen.

Roughly, because it's be very network-dependent, for us going from `MXNET_MKLDNN_CACHE_SIZE=512` to `MXNET_MKLDNN_CACHE_SIZE=128` gives a 55% memory saving for a 7% performance loss.

azai91:
Resnet 50 with cache disabled.
ubuntu@ip-172-31-3-217:~/incubator-mxnet$ MXNET_MKLDNN_CACHE_SIZE=-1 python example/image-classification/benchmark_score.py
/home/ubuntu/anaconda3/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
Assertion failure at kmp_runtime.cpp(2582): __kmp_init_serial.
Assertion failure at kmp_runtime.cpp(2582): __kmp_init_serial.
Assertion failure at kmp_runtime.cpp(2582): __kmp_init_serial.
Assertion failure at kmp_runtime.cpp(2582): __kmp_init_serial.
Assertion failure at kmp_runtime.cpp(2582): __kmp_init_serial.
Assertion failure at kmp_runtime.cpp(2582): __kmp_init_serial.
Assertion failure at kmp_runtime.cpp(2582): __kmp_init_serial.
INFO:root:network: resnet-50
INFO:root:device: cpu(0)
/home/ubuntu/incubator-mxnet/python/mxnet/module/base_module.py:68: UserWarning: Data provided by label_shapes don't match names specified by label_names ([] vs. ['softmax_label'])
  warnings.warn(msg)
[23:20:09] ../src/operator/nn/mkldnn/mkldnn_base.cc:74: Allocate 147456 bytes with malloc directly
[23:20:09] ../src/operator/nn/mkldnn/mkldnn_base.cc:74: Allocate 589824 bytes with malloc directly
[23:20:09] ../src/operator/nn/mkldnn/mkldnn_base.cc:74: Allocate 2359296 bytes with malloc directly
[23:20:09] ../src/operator/nn/mkldnn/mkldnn_base.cc:74: Allocate 9437184 bytes with malloc directly

azai91:
Resnet with cache limit.
ubuntu@ip-172-31-3-217:~/incubator-mxnet$ MXNET_MKLDNN_CACHE_SIZE=-1 python example/image-classification/benchmark_score.py
/home/ubuntu/anaconda3/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
Assertion failure at kmp_runtime.cpp(2582): __kmp_init_serial.
Assertion failure at kmp_runtime.cpp(3767): __kmp_gtid_get_specific() == gtid.
INFO:root:network: resnet-50
INFO:root:device: cpu(0)
/home/ubuntu/incubator-mxnet/python/mxnet/module/base_module.py:68: UserWarning: Data provided by label_shapes don't match names specified by label_names ([] vs. ['softmax_label'])
  warnings.warn(msg)
[23:29:25] ../src/operator/nn/mkldnn/mkldnn_base.cc:74: Allocate 147456 bytes with malloc directly
[23:29:25] ../src/operator/nn/mkldnn/mkldnn_base.cc:74: Allocate 589824 bytes with malloc directly
[23:29:26] ../src/operator/nn/mkldnn/mkldnn_base.cc:74: Allocate 2359296 bytes with malloc directly
[23:29:26] ../src/operator/nn/mkldnn/mkldnn_base.cc:74: Allocate 9437184 bytes with malloc directly
INFO:root:batch size 32, dtype float32, images/sec: 105.089997

ZhennanQin:
So for throughput, there's a drop about 2%(from 107.744738 to 105.089997), isn't it? How about latency change? I guess it will be impacted a lot from your current implementation.

azai91:
you're right. I will just remove the cache as suggested above.

ankkhedia:
@azai91 ping!
Any updates on this PR

src/operator/nn/mkldnn/mkldnn_act.cc
```diff
@@ -135,24 +135,21 @@ static MKLDNNActForward &GetActForward(const ActivationParam& param,
                                        const OpContext &ctx, const NDArray &in_data,
                                        const mkldnn::memory &in_mem) {
 #if DMLC_CXX11_THREAD_LOCAL
-  static thread_local std::unordered_map<MKLDNNActSignature, MKLDNNActForward, OpHash> fwds;
+  static thread_local MKLDNNCache<MKLDNNActSignature, MKLDNNActForward, OpHash> fwds;
```
nswamy: one of the reason why the memory usage is high could be because it maintains a cache for each thread? 
I am wondering if the original authors measured contention and performance impact if we maintained one cache for the process? 
I understand it *might* need some kind of locking, has someone explored non-blocking/lockfree data structures?

azai91:
I have done the following experiments and their results are below, we have to pick one of the solutions unless you have better one @ZhennanQin @TaoLv ? 
1) LRU cache with performance hit of 2%.
2) Use the unordered_set that drops elements randomly. I have tested this and it does not cause regression.

ZhennanQin:
@azai91 I prefer the 2nd choice as LRU doesn't bring any help to performance. Thanks for contribution.

docs/faq/env_var.md
anirudh2290: nit: can we be more specific about what the units for cache_size are. Example , number of elements in the cache.

azai91:
closing this PR because it is a dup of https://github.com/apache/incubator-mxnet/pull/13324.

PR #12764, opened 2018-10-09
Updated tvm submodule head

ZhennanQin:
## Description ##
Updated tvm submodule head.

### Essentials ###

### Changes ###

## Comments ##


sandeep-krishnamurthy:
@marcoabreu @larroy  @KellenSunderland - Can you please take a look at this PR?

ZhennanQin:
Explain a bit for removing 'FInplaceIdentity' attr for cast and _backward_cast.
After upgrading tvm, a minor change on memory planning pass is included that allow doing memory inplace optimization on different data type with same size. Eg, for 'cast' op that casting data from fp32 to int32, previously, even  this op has 'FInplaceOption' and 'FInplaceIdentity' attrs, memory inplace won't happen because input and output have different data type(fp32 vs int32). After upgrading tvm, as fp32 and int32 are both 32-bit data size, memory inplace optimization will take effect.
Memory inplace optimization for 'cast' is a good thing, but 'FInplaceIdentity' will cause trouble when inputs has more than 1 dependence, which means the input will be used by other ops, and memory inplace will override the input, leading result incorrect. Removing 'FInplaceIdentity' can fix this issue.
Overall, after upgrading tvm with this fix, 'cast' can correctly get benefit from memory inplace optimization now.
@tqchen What do you think?

KellenSunderland:
Non Blocking / Just Comments not a Review:
It would be great if we could get rid of some of the mem copies associated with casts.  I've seen quite a few cases (in Sockeye) where casts completely dominate performance when executing mixed precision graphs.  Typically what I saw was a bunch of FCs running with TCs in fp16, then a cast before a softmax that took longer to run than the actual softmax, then a cast back.  Granted in that case it was fp16<->fp32 conversions, but if we can at least improve the speed of int32<->fp32 it'd already be a great first step.  I'd imaging people trying to run graphs mixing between int32 and fp32 would be hitting the same issues.

Would this change have any impact on threadsaftey?  Would it be making a tensor that was previously immutable mutable?  Do we have protections in cases where the same input is casted from different threads (even to the same data type)?

ZhennanQin:
@KellenSunderland I guess this change won't have impact on threadsaftey, because the 'cast' mentioned here is a symbolic op, which used in computing graph.

eric-haibin-lin:
There're quite a few ops registered with FInplaceIdentity. Why is it only removed from cast?

szha:
Cast changes the type and others shouldn't

Issue #12767, opened 2018-10-09
Flaky test: test_gluon_gpu.test_slice_batchnorm_reshape_batchnorm


lebeg:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/master/1728/pipeline

```
======================================================================
ERROR: test_gluon_gpu.test_slice_batchnorm_reshape_batchnorm
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/usr/local/lib/python3.5/dist-packages/nose/util.py", line 620, in newfunc
    return func(*arg, **kw)
  File "/work/mxnet/tests/python/gpu/../unittest/common.py", line 172, in test_new
    orig_test(*args, **kwargs)
  File "/work/mxnet/tests/python/gpu/../unittest/test_gluon.py", line 2033, in test_slice_batchnorm_reshape_batchnorm
    check_layer_forward_withinput(net, x)
  File "/work/mxnet/tests/python/gpu/../unittest/test_gluon.py", line 1507, in check_layer_forward_withinput
    mx.test_utils.assert_almost_equal(x.grad.asnumpy(), x_hybrid.grad.asnumpy(), rtol=1e-5, atol=1e-6)
  File "/work/mxnet/python/mxnet/ndarray/ndarray.py", line 1980, in asnumpy
    ctypes.c_size_t(data.size)))
  File "/work/mxnet/python/mxnet/base.py", line 253, in check_call
    raise MXNetError(py_str(_LIB.MXGetLastError()))

mxnet.base.MXNetError: [04:48:16] src/operator/nn/./cudnn/cudnn_convolution-inl.h:875: 
Failed to find any forward convolution algorithm.  with workspace size of 1073741824 
bytes, please consider reducing batch/model size or increasing the workspace size
```

lebeg:
Possibly related to:
Failing test: test_gluon_gpu.test_slice_batchnorm: https://github.com/apache/incubator-mxnet/issues/12715

larroy:
I'm unsure this is a flaky test, I think it's a cuda / cudnn or CI environment problem. Could you reproduce?

piyushghai:
@mxnet-label-bot [flaky, Gluon]

lebeg:
Another consecutive run failed on master CI:
[...]

PR #12768, opened 2018-10-09
Disabled: test_gluon_gpu.test_slice_batchnorm_reshape_batchnorm

lebeg:
## Description ##

Disabled flaky test: test_gluon_gpu.test_slice_batchnorm_reshape_batchnorm. Tracked in https://github.com/apache/incubator-mxnet/issues/12767

### DIFF
--- a/tests/python/unittest/test_gluon.py
+++ b/tests/python/unittest/test_gluon.py
@@ -2007,6 +2007,7 @@ def test_reshape_batchnorm_reshape_batchnorm():
 
 
 @with_seed()
+@unittest.skip('Flaky test: https://github.com/apache/incubator-mxnet/issues/12767')
 def test_slice_batchnorm_reshape_batchnorm():
     class Net(gluon.HybridBlock):
         def __init__(self, shape, slice, **kwargs):


larroy:
See my comment in the issue, were you able to reproduce? Do you have some evidence to declare it as a flaky test?

marcoabreu:
Agree, it seems like a more general problem that's no related to a test being flaky but actually having a bug in our code. I don't think that disabling all cudnn-backed tests is the way to go

marcoabreu:
Basically all batchnorm tests are randomly failing, so don't think that we can disable a dozens of tests.

lebeg:
> Basically all batchnorm tests are randomly failing because of CuDNN

There is no evidence for this.

Even if this would be the case merging a PR that reduces failure rate appears to be the right thing to do to me.

ChaiBapchya:
Temporarily fixing the flaky test by skipping it, shouldn't cause any harm. Would request this to be merged until flakiness is resolved.

nswamy:
I debugged a similar test( a week ago, the issue does not arise(at least for the test I ran) when you run it standalone how many every times. It would crash only when you ran the entire test suite and fail in 1/10 times. 

There are a couple of possibilities(from my findings): 
1) There is a leak in the Nvidia drivers. 
2) There is a leak in the CUDA code that is getting aggregated(running all the tests) in the process and hence throwing the error. -- most likely this. 

I ran the all the GPU tests over the weekend(outside docker) and found that out of 100 tests, 10 times I get this error, I tend to believe there is a memory leak in the CUDA code.

nswamy:
The test I was debugging was test_gluon_gpu.test_slice_batchnorm(its disabled currently)

larroy:
I will bring it up with NVidia.

larroy:
https://github.com/apache/incubator-mxnet/pull/12804

Issue #12431, opened 2018-09-01
[R] use of mx.io.arrayiter completely crashes R environment


some-guy1:
Running the code below causes the entire R environment to terminate.  This code used to work on older versions of mxnet (older than 1.2.0).  It seems to be related to mx.io.arrayiter

## Environment info (Required)
R version 3.5.0 (2018-04-23)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows >= 8 x64 (build 9200)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] mxnet_1.3.0

loaded via a namespace (and not attached):
 [1] Rcpp_0.12.16       pillar_1.2.2       compiler_3.5.0     RColorBrewer_1.1-2 influenceR_0.1.0   plyr_1.8.4        
 [7] bindr_0.1.1        viridis_0.5.1      tools_3.5.0        digest_0.6.15      jsonlite_1.5       viridisLite_0.3.0 
[13] tibble_1.4.2       gtable_0.2.0       rgexf_0.15.3       pkgconfig_2.0.1    rlang_0.2.0        igraph_1.2.1      
[19] rstudioapi_0.7     yaml_2.1.19        bindrcpp_0.2.2     gridExtra_2.3      downloader_0.4     DiagrammeR_1.0.0  
[25] dplyr_0.7.4        stringr_1.3.1      htmlwidgets_1.2    hms_0.4.2          grid_3.5.0         glue_1.2.0        
[31] R6_2.2.2           Rook_1.1-1         XML_3.98-1.11      readr_1.1.1        purrr_0.2.4        tidyr_0.8.0       
[37] ggplot2_2.2.1      magrittr_1.5       codetools_0.2-15   scales_0.5.0       htmltools_0.3.6    assertthat_0.2.0  
[43] colorspace_1.3-2   brew_1.0-6         stringi_1.2.2      visNetwork_2.0.3   lazyeval_0.2.1     munsell_0.4.3  

## Build info (Required if built from source)
https://github.com/apache/incubator-mxnet/tree/master/R-package

```
cran <- getOption("repos")
[...]

PR #12776, opened 2018-10-10
R fix metric shape

jeremiedb:
Reshape label and predictions for regression use cases to handle more general cases where there are multiple output. Fix for #12431


piyushghai:
@ankkhedia @anirudhacharya Looping you guys in to review this! 
@jeremiedb Thank you for your fixing the issue! 

PR #12791, opened 2018-10-11
[MXNET-1093] Add python3 Docker images for each MXNet release

mbaijal:
## Description ##
@KellenSunderland @marcoabreu @gautamkmr 
This PR updates the docker/docker-python folder to add 8 new dockerfiles for python3. It also updates the automated docker release script to build and update the python 3 docker images similar to python 2. 
Python version used is 3.5 which is the default version on ubuntu 16.04. 

### Essentials ###

### Changes ###
- [x] Added 8 new dockerfiles for python 3 - cpu, cu80, cu90, cu92 with and without mkl. 
- [x] Updated the build script to also build, test and upload these new dockerfiles
- [x] Updated the readme.


docker/docker-python/Dockerfile.mxnet.python3.cpu
```diff
@@ -0,0 +1,28 @@
...
+# under the License.
+#
+# Dockerfile to build MXNet for CPU
+
+FROM ubuntu:16.04
+
+RUN apt-get update
+RUN apt-get install -y wget python3 gcc
+RUN wget https://bootstrap.pypa.io/get-pip.py
+RUN python3 get-pip.py
```
KellenSunderland: Looks good.  Super nit: I'd combine smaller commands like this into a single layer.  
```Dockerfile
RUN apt-get update && \
        apt-get install -y wget python3 gcc && \
        ...
```

docker/docker-python/Dockerfile.mxnet.python3.cpu
```diff
@@ -0,0 +1,28 @@
...
+
+RUN pip3 install mxnet
```
marcoabreu: Version has to be specifyable

docker/docker-python/README.md
```diff
@@ -12,13 +12,21 @@ It uses the appropriate pip binaries to build different docker images as -
 * gpu_cu80_mkl
 * gpu_cu92 
 * gpu_cu92_mkl
+* cpu_py3
```
marcoabreu: I'd prefer if python3 was the default.

docker/docker-python/README.md
```diff
@@ -39,3 +47,10 @@ Credentials can be provided in the following ways:
 * **Set Environment Variables:** Set the following environment variables which the script will pick up to login to dockerhub at runtime -
     * $MXNET_DOCKERHUB_PASSWORD
     * $MXNET_DOCKERHUB_USERNAME
+    
+
+### Using the Docker Images
+* The MXNet Python Docker images can be found here: https://hub.docker.com/r/mxnet/python/
+
+* Docker Pull Command: `docker pull mxnet/python:<image_tag>`
```
marcoabreu: Where can I find details about version tagging?

docker/docker-python/build_python_dockerfile.sh
```diff
@@ -117,10 +131,40 @@ docker_build_image "${mxnet_version}_gpu_cu92_mkl" "Dockerfile.mxnet.python.gpu.
 docker_test_image_gpu "${mxnet_version}_gpu_cu92_mkl"
 
 
+
+# Build and Test Python3 dockerfiles - CPU
+docker_build_image "${mxnet_version}_cpu_python3" "Dockerfile.mxnet.python3.cpu"
```
marcoabreu: I'd prefer if the builds and CPU tests ran parallelized to reduce the critical path

KellenSunderland: I'd be a little concerned about memory usage if we were to parallelize the builds.  I could see that this might make building the images difficult on a laptop or even a middle-tier cloud instance.

marcoabreu: I built 15 images for our docker cache in parallel on my laptop and didn't encounter any problems. I don't think it will be an issue. Also, this script is targeted for CD execution - individual compilation could be an additional mode, but we should optimise for maximum resource utilization here IMO

docker/docker-python/Dockerfile.mxnet.python3.cpu
mbaijal: Do you mean parameterize the MXNet version in the dockerfiles?

docker/docker-python/README.md
mbaijal: What details are you looking for? The tags can be looked up in the dockerhub link I have provided above - https://hub.docker.com/r/mxnet/python/tags/
Do you want me to add any other details to the README?

docker/docker-python/Dockerfile.mxnet.python3.cpu
mbaijal: @KellenSunderland @bhavinthaker Thank you, I agree. I will update the PR.

docker/docker-python/README.md
mbaijal: I would not want to change the docker images already uploaded for MXNet 1.3 since this might cause something to break for a user. This change to make python3 the default docker version cannot be immediate and should be discussed on the dev list. If agreed upon, this should take effect from the next release onwards. Wdyt?

docker/docker-python/Dockerfile.mxnet.python3.cpu
marcoabreu: Yes. We have to be able to create images for the different versions in a reproducible fashion.

docker/docker-python/README.md
marcoabreu: Very good points. Could you start the discussion?

docker/docker-python/README.md
marcoabreu: I'm only on my phone, so I might be missing some stuff. I'll leave it up to you :) sorry for the confusion

mbaijal:
Yea.

docker/docker-python/README.md
mbaijal: No worries, Thanks for the review.

docker/docker-python/Dockerfile.mxnet.python.cpu
```diff
@@ -19,10 +19,11 @@
...
-RUN apt-get install -y wget python gcc
-RUN wget https://bootstrap.pypa.io/get-pip.py
-RUN python get-pip.py
+RUN apt-get update && \
+    apt-get install -y wget python gcc && \
+    wget https://bootstrap.pypa.io/get-pip.py && \
+    python get-pip.py
 
-RUN pip install mxnet
+RUN pip install mxnet==$version
```
marcoabreu: I think we should separate the mxnet package installation from the dependency installation and rather do it during runtime. Otherwise, the package will be cached by our Docker cache and we won't be running the live version.

docker/docker-python/build_python_dockerfile.sh
```diff
@@ -79,48 +76,82 @@ docker_account_logout(){
...
+docker_generate_image_gpu "${mxnet_version}_gpu_cu90" "Dockerfile.mxnet.python.gpu.cu90" "python" &
+docker_generate_image_gpu "${mxnet_version}_gpu_cu90_mkl" "Dockerfile.mxnet.python.gpu.cu90.mkl" "python" &
+docker_tag_image_gpu "${mxnet_version}_gpu_cu90" "gpu" "python" &
+docker_generate_image_gpu "${mxnet_version}_gpu_cu80" "Dockerfile.mxnet.python.gpu.cu80" "python" &
+docker_generate_image_gpu "${mxnet_version}_gpu_cu80_mkl" "Dockerfile.mxnet.python.gpu.cu80.mkl" "python" &
+docker_generate_image_gpu "${mxnet_version}_gpu_cu92" "Dockerfile.mxnet.python.gpu.cu92" "python" &
+docker_generate_image_gpu "${mxnet_version}_gpu_cu92_mkl" "Dockerfile.mxnet.python.gpu.cu92.mkl" "python"
+
+echo "Waiting for MXNet Python2 Docker Images to Build"
+wait
```
marcoabreu: What exactly does this do?

docker/docker-python/build_python_dockerfile.sh
```diff
@@ -79,48 +76,82 @@ docker_account_logout(){
...
 
-docker_build_image "${mxnet_version}_gpu_cu92_mkl" "Dockerfile.mxnet.python.gpu.cu92.mkl"
-docker_test_image_gpu "${mxnet_version}_gpu_cu92_mkl"
+# Build and Test dockerfiles - CPU
+docker_generate_image_cpu "${mxnet_version}_cpu" "Dockerfile.mxnet.python.cpu" "python" &
+docker_generate_image_cpu "${mxnet_version}_cpu_mkl" "Dockerfile.mxnet.python.cpu.mkl" "python" &
+docker_tag_image_cpu "${mxnet_version}_cpu" "latest" "python" &
 
+#Build and Test dockerfiles - GPU
+docker_generate_image_gpu "${mxnet_version}_gpu_cu90" "Dockerfile.mxnet.python.gpu.cu90" "python" &
```
marcoabreu: Does this actually run all the commands in parallel?

docker/docker-python/Dockerfile.mxnet.python.cpu
mbaijal: This can be avoided by adding the `--no-cache` flag to the docker build command too, right?

docker/docker-python/build_python_dockerfile.sh
mbaijal: Yes I verified using `ps -ef` that builds/tests are running in parallel. However, I do also want to record the total runtime with and without parallelization.

docker/docker-python/build_python_dockerfile.sh
mbaijal: It waits for all the processes spawned as background processes (build, test) to complete before moving on to the next command.

docker/docker-python/Dockerfile.mxnet.python.cpu
marcoabreu: Right, but we want to use the cache. We just have to separate runtime dependencies from build time dependencies.

docker/docker-python/build_python_dockerfile.sh
marcoabreu: Excellent! Did you also test error propagation in that case?

docker/docker-python/README.md
mbaijal: Done. The decision is to make python3 default. I will make the change.

mbaijal: This has been discussed on dev-list and the decision is to switch to python3 images as default. This will be done in the next release to avoid disruption to the current customers.

docker/docker-python/build_python_dockerfile.sh
mbaijal: The following are runtimes on a p2.16x instance to build and test all docker images (python2 and python3)
Runtime with parallelization: 13mins
Runtime without parallelization: 29mins

marcoabreu: Awesome! What happens if there is an error in one of the builds (you can simulate that by just returning a non zero exit code in one of the RUN clauses)? Does it get propagated?

mbaijal: I checked that by hardcoding an incorrect mxnet version in one of the dockerfiles. While a non-zero error message is printed the logs, the script does not exit and completes with "success". 
I will fix this.

marcoabreu: Great, thanks

ankkhedia:
@marcoabreu It seems that your review comments has been addressed. Could you please take a look?

mbaijal:
@marcoabreu @sandeep-krishnamurthy Could either of you please merge this PR. Thanks!

docker/docker-python/build_python_dockerfile.sh
mbaijal: Fixed in the last commit by writing each build log to a file and then parsing these logs for a possible error message. If error found, script fails. The error message makes it easy to find out which tag failed.

marcoabreu:
Please wait with merging. I'm still discussing some things internally

docker/docker-python/build_python_dockerfile.sh
```diff
@@ -39,29 +40,34 @@ fi
...
+  rm -rf ${LOGDIR}
+fi
+mkdir ${LOGDIR}
 
-docker_tag_image(){
-    docker tag mxnet/python:${1} mxnet/python:${2}
-}
 
+# Functions
 docker_test_image_cpu(){
```
larroy: We can't have docker running like this because it leaves containers hanging around even when the parent process terminates. We introduce logic to avoid this kinds of problems in CI  https://github.com/apache/incubator-mxnet/pull/12381

My suggestion is to use one of this two options:
- Use https://github.com/apache/incubator-mxnet/blob/master/ci/build.py as it's done in https://github.com/apache/incubator-mxnet/blob/master/ci/docker_cache.py to run the tests safely, build the images in parallel. This is probably the fastest route that is safe but introduces some coupling between CI and CD.
- Abstract the cleanup logic in build.py in a module which is then used from CI and CD with the specific commands.
- Duplicate the logic from build.py and docker_cache.py for this task (copy & paste) and adjust for the specific case. A downside is that introduces similar logic to be maintained but not coupling.

If on doubt, we can choose option 1 and go from there.

larroy:
Thanks for this effort. I have some reservations about the implementation, see my previous comment about zombie containers and resource disposal in the code.

Also, can we use Python instead of shell to automate these steps? It's more robust, debuggable, flexible and easier to mantain, we don't want to introduce complex logic in shell scripts. 

All the automation on the project is done in Python. The code that is forking process to parallelize the build should be done as it's done in https://github.com/apache/incubator-mxnet/blob/master/ci/docker_cache.py or with a similar Python construct of your choice. Even if it's not critical to parallelize building of the images, if the code is well factored and building the container it's on its own function is very easy to parallelize.

ankkhedia:
@mbaijal Could you please look into comments provided by @lanking520 and @marcoabreu

kalyc:
Thanks for your contributions @mbaijal 

@larroy ,  @marcoabreu could you please help with reviewing this PR?

vandanavk:
@mbaijal could you check the CI build failure and add a comment explaining the changes you made/are making in response to Pedro and Marco's feedback?

@mxnet-label-bot update [CI, Docker, pr-awaiting-testing]

larroy:
Given that we are taking over this script, shall we close this?

larroy:
@perdasilva

perdasilva:
@larroy - let's close it. I'll open a new PR after my RFC goes through

larroy:
@perdasilva it's meghnas PR, @mbaijal can you close?  Thanks.

mbaijal:
I don't mind closing this but half of the code for this change (python2 and initial automation script) has already been merged in a previous PR and this PR contains some updates and fixes to it. This updated script was also used for the last two docker releases. In my opinion it might be better to merge this PR to complete those changes and then deprecate it when a new python script is available.

perdasilva:
@mbaijal imo, in this case, we should probably go fwd and merge. @larroy?

Roshrini:
@larroy Is it okay if we merge this PR? 
@nswamy @marcoabreu

anirudhacharya:

sandeep-krishnamurthy:
@larroy / @marcoabreu - I think we should take this PR forward. Can you please help take a look again?

marcoabreu:
Please hold on with this PR since we are currently refactoring that pipeline.

@perdasilva for further communication regarding this topic

larroy:
up to @perdasilva

perdasilva:
@marcoabreu lets talk about this issue today and come up with the solution. If the PR does what someone needs right now, I don't see any reason to hold this back. We can always refactor it later if need be...

perdasilva:
@mbaijal @sandeep-krishnamurthy @srochel  please merge this - thank you

mbaijal:
Thanks @perdasilva !

PR #12793, opened 2018-10-11
Fix mismatch shapes

### LINKED ISSUE #12676: Mismatched Shapes in Value Error

ChaiBapchya:
## Description ##
Fixes #12676 

Shapes were mismatched. Switched them.

### DIFF
--- a/python/mxnet/ndarray/ndarray.py
+++ b/python/mxnet/ndarray/ndarray.py
@@ -869,7 +869,7 @@ fixed-size items.
         source_array = np.asarray(source_array, dtype=self.dtype, order='C')
         if source_array.shape != self.shape:
             raise ValueError('Shape inconsistent: expected %s vs got %s'%(
-                str(self.shape), str(source_array.shape)))
+                str(source_array.shape))), str(self.shape)
         check_call(_LIB.MXNDArraySyncCopyFromCPU(
             self.handle,
             source_array.ctypes.data_as(ctypes.c_void_p),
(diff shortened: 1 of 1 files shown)


ChaiBapchya:
Yes, we are good to go. As mentioned in the issue #12676 it's a minor issue of printing error/exception where the shapes got mismatched. This shouldn't need a unit test since nothing to assert as such.

PR #12803, opened 2018-10-11
Fix Extremely Unbalanced Issue in split_and_load

zhanghang1989:
## Description ##
Existing implementation:
```python
import mxnet as mx
x = mx.nd.ones((38, 3, 4, 4))
ctx = [mx.gpu(i) for i in range(8)]
xs = mx.gluon.utils.split_and_load(x, ctx, even_split=False)

for xi in xs:
    print(xi.shape[0])
```
The output is:
```bash
4
4
4
4
4
4
4
10
```

### DIFF
--- a/python/mxnet/gluon/utils.py
+++ b/python/mxnet/gluon/utils.py
@@ -72,7 +72,7 @@ def split_data(data, num_slice, batch_axis=0, even_split=True):
             "uneven partitioning of data."%(
                 str(data.shape), num_slice, batch_axis, num_slice))
 
-    step = size // num_slice
+    step = size // num_slice + 1
     if batch_axis == 0:
         slices = [data[i*step:(i+1)*step] if i < num_slice - 1 else data[i*step:size]
                   for i in range(num_slice)]
--- a/python/mxnet/gluon/utils.py
+++ b/python/mxnet/gluon/utils.py
@@ -22,6 +22,7 @@ __all__ = ['split_data', 'split_and_load', 'clip_global_norm',
            'check_sha1', 'download']
 
 import os
+import math
 import hashlib
 import warnings
 import collections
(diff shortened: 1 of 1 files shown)


zhanghang1989:
FYI @eric-haibin-lin

marcoabreu:
Can we double check that this is properly covered by tests? I'm afraid this might be a behaviour change, so we might run into compatibility issues

sandeep-krishnamurthy:
Thanks, @zhanghang1989 for your contributions.

Is see the behavior when even_split=True and input is batch of 16 and there are 4 GPUs is changing, it is expected to load 4 samples per GPU. But, with this change it becomes - 5, 5, 5, 1

ankkhedia:
@zhanghang1989 ping!
Did you get a chance to look into @sandeep-krishnamurthy's comments above and could you also look into CI failures

zhanghang1989:
@ankkhedia I have made the changes according to @sandeep-krishnamurthy 's comments https://github.com/apache/incubator-mxnet/pull/12803/commits/688259d2e43f6eb35c6d73f8162b98baa35932c9. I will take a look at CI failure later. Thanks

kalyc:
@zhanghang1989 could you please take a look at the CI build failure whenever you have the time?

stu1130:
@zhanghang1989 ping again

anirudhacharya:
@sandeep-krishnamurthy are you sure it behaves the way you have outlined when `even_split=True`. I just tried and it distributes evenly -
@zhanghang1989 

```
import mxnet as mx
x = mx.nd.ones((36, 3, 4, 4))
ctx = [mx.gpu(i) for i in range(4)]
xs = mx.gluon.utils.split_and_load(x, ctx, even_split=True)

for xi in xs:
    print(xi.shape[0])
```
returns
```
9
9
9
9
```

python/mxnet/gluon/utils.py
```diff
@@ -72,7 +73,7 @@ def split_data(data, num_slice, batch_axis=0, even_split=True):
             "uneven partitioning of data."%(
                 str(data.shape), num_slice, batch_axis, num_slice))
 
-    step = size // num_slice
+    step = math.ceil(1.0 * size / num_slice)
```
anirudhacharya: I am not sure this fixes the problem, for instance when `size=34` and `num_slice=8`, the above calculation will return `step = 5`. How will 34 elements be split 8 ways with a `step = 5`.

Issue #12662, opened 2018-09-25
Memory leak when passing images of different dimensions with MXNET_CUDNN_AUTOTUNE_DEFAULT


Ishitori:
## Description
I have noticed, that if I use MXNET_CUDNN_AUTOTUNE_DEFAULT=1 with big image dimensions (1900x1900), then after a forward pass a lot of GPU memory got consumed and never released. Autotune gets Out of Memory exception, if I try to pass another image after the first one with also big, but different dimensions (1800x1800). 

The image dimensions are smaller, so my assumption is that since 1900x1900 got processed then 1800x1800 should also be processed, because it takes less memory. But it is actually not the case, because after the first image processing some of the GPU memory is not released.

The main question for me is why GPU memory is not released once the first image is processed? It seems like something is holding it. I think there is a memory leak or some sort of cache, which is never released.

## Environment info (Required)

```
----------Python Info----------
Version      : 3.6.4
Compiler     : GCC 7.2.0
Build        : ('default', 'Jan 16 2018 18:10:19')
Arch         : ('64bit', '')
------------Pip Info-----------
Version      : 9.0.1
Directory    : /home/ubuntu/anaconda3/lib/python3.6/site-packages/pip
----------MXNet Info-----------
/home/ubuntu/anaconda3/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprec$
ted. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
Version      : 1.3.0
Directory    : /home/ubuntu/anaconda3/lib/python3.6/site-packages/mxnet
Commit Hash   : b3be92f4a48bce62a5a8424271871c2f81c8f7f1
----------System Info----------
Platform     : Linux-4.4.0-1066-aws-x86_64-with-debian-stretch-sid
system       : Linux
node         : ip-172-31-22-61
release      : 4.4.0-1066-aws
version      : #76-Ubuntu SMP Thu Aug 16 16:21:21 UTC 2018
----------Hardware Info----------
machine      : x86_64
processor    : x86_64
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                32
On-line CPU(s) list:   0-31
Thread(s) per core:    2
Core(s) per socket:    16
Socket(s):             1
NUMA node(s):          1
[...]

PR #12804, opened 2018-10-12
CudnnFind() usage improvements

DickJC123:
## Description ##

This PR improves the MXNet's use of cudnnFind() to address a few issues:
    1. With the gluon imperative style, cudnnFind() is called during forward(), and so might have its timings perturbed by other GPU activity (including potentially other cudnnFind() calls).
    2. With some cuda drivers versions, care is needed to ensure that the large I/O and workspace cudaMallocs() performed by cudnnFind() are immediately released and available to MXNet.
    3. cudnnFind() makes both conv I/O and workspace allocations that must be covered by the GPU global memory headroom defined by MXNET_GPU_MEM_POOL_RESERVE.  Per issue #12662, large convolutions can result in out-of-memory errors, even when MXNet's storage allocator has free memory in its pool.

This PR addresses these issues, providing the following benefits:
    1. Consistent algo choice for a given convolution type in a model, both for instances in the same GPU and in other GPUs in a multi-GPU training setting.
    2. Consistent algo choice from run to run, based on eliminating sources of interference of the cudnnFind() timing process.
    3. Consistent model global memory footprint, both because of the consistent algo choice (algo's can have markedly different workspace requirements) and changes to MXNet's use of cudaMalloc.
    4. Increased training performance based on being able to consistently run with models that approach the GPU's full global memory footprint.
    5.  Adds a unittest for and solves issue #12662.

Now specifically, the PR makes the following changes/additions:
    1. Merges the cudnn algo registry separate Find() and Register() methods into a single FindOrElseRegister() call protected by a mutex.  Before, there was a window of time between when a thread found no algo entry and when it registered a new entry.  This window could allow a different thread to call cudnnFind() on the same convolution type, getting a different answer.  FindOrElseRegister() takes a callback that determines the algo via cudnnFind() and makes access to the cudnn algo registry effectively atomic.  The callbacks provided with cudnn_convolution and cudnn_deconvolution also grab the storage allocator's mutex to quiet the GPU activity from other threads during the running of cudnnFind().
    2. Changes MXNet's storage allocator to round large allocations to a multiple of 2MB.  This value ensures that small permanent allocations that might have shared a cuda memory allocation slab with a prior large allocation don't prevent the reclaiming of that slab area.
    3. Includes use of the MXNet storage allocator's Alloc() and DirectFree() API to coax a convolution's I/O storage needs away from MXNet's storage allocator prior to calling cudnnFind().

The first commit introduces a unittest that demonstrates issue #12662 in the current codebase.  In order to make the test adaptable to GPUs of varying global memory, a C-level API was brought out to the python API by introducing mx.context.gpu_memory_info(dev_id).  Follow-up commits supply the fix and the rest of the PR. 

Hopefully Interested parties: @KellenSunderland @piiswrong @eric-haibin-lin @zheng-da

### Essentials ###

### Changes ###

## Comments ##

The following command was run on an 8-GPU system to demonstrate the new consistency of global memory footprint:
```
python /opt/mxnet/example/image-classification/train_imagenet.py --gpu 0,1,2,3,4,5,6,7 --batch-size 256 --num-epochs 1 --data-train /data/iman-val-recordio-352/train.rec --disp-batches 10 --network inception-v3 --data-nthreads 32 --image-shape 3,299,299
```
Two nvidia-smi outputs are shown.  The first is before the PR (showing footprints in the range of 5GB to 8GB), while the second is after (showing consistent 5GB footprints with < 1% variation).

Before PR:
```
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 410.45                 Driver Version: 410.45                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  Tesla V100-SXM2...  On   | 00000000:06:00.0 Off |                    0 |
| N/A   58C    P0   247W / 300W |   6312MiB / 32480MiB |     86%      Default |
+-------------------------------+----------------------+----------------------+
|   1  Tesla V100-SXM2...  On   | 00000000:07:00.0 Off |                    0 |
| N/A   60C    P0   274W / 300W |   5542MiB / 32480MiB |     86%      Default |
+-------------------------------+----------------------+----------------------+
|   2  Tesla V100-SXM2...  On   | 00000000:0A:00.0 Off |                    0 |
| N/A   63C    P0   255W / 300W |   5152MiB / 32480MiB |     89%      Default |
+-------------------------------+----------------------+----------------------+
|   3  Tesla V100-SXM2...  On   | 00000000:0B:00.0 Off |                    0 |
| N/A   54C    P0   258W / 300W |   6730MiB / 32480MiB |     88%      Default |
+-------------------------------+----------------------+----------------------+
|   4  Tesla V100-SXM2...  On   | 00000000:85:00.0 Off |                    0 |
| N/A   57C    P0   222W / 300W |   5158MiB / 32480MiB |     87%      Default |
+-------------------------------+----------------------+----------------------+
|   5  Tesla V100-SXM2...  On   | 00000000:86:00.0 Off |                    0 |
| N/A   59C    P0   260W / 300W |   5168MiB / 32480MiB |     84%      Default |
+-------------------------------+----------------------+----------------------+
|   6  Tesla V100-SXM2...  On   | 00000000:89:00.0 Off |                    0 |
| N/A   63C    P0   266W / 300W |   5160MiB / 32480MiB |     88%      Default |
+-------------------------------+----------------------+----------------------+
|   7  Tesla V100-SXM2...  On   | 00000000:8A:00.0 Off |                    0 |
| N/A   58C    P0   255W / 300W |   8252MiB / 32480MiB |     90%      Default |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|    0     53319      C   python                                      6289MiB |
|    1     53319      C   python                                      5519MiB |
|    2     53319      C   python                                      5129MiB |
|    3     53319      C   python                                      6707MiB |
|    4     53319      C   python                                      5135MiB |
|    5     53319      C   python                                      5145MiB |
|    6     53319      C   python                                      5137MiB |
|    7     53319      C   python                                      8229MiB |
+-----------------------------------------------------------------------------+
```
After PR:
```
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 410.45                 Driver Version: 410.45                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  Tesla V100-SXM2...  On   | 00000000:06:00.0 Off |                    0 |
| N/A   55C    P0   271W / 300W |   5044MiB / 32480MiB |     85%      Default |
+-------------------------------+----------------------+----------------------+
|   1  Tesla V100-SXM2...  On   | 00000000:07:00.0 Off |                    0 |
| N/A   57C    P0   221W / 300W |   5042MiB / 32480MiB |     87%      Default |
+-------------------------------+----------------------+----------------------+
|   2  Tesla V100-SXM2...  On   | 00000000:0A:00.0 Off |                    0 |
| N/A   59C    P0   214W / 300W |   5042MiB / 32480MiB |     84%      Default |
+-------------------------------+----------------------+----------------------+
|   3  Tesla V100-SXM2...  On   | 00000000:0B:00.0 Off |                    0 |
| N/A   52C    P0   263W / 300W |   5040MiB / 32480MiB |     82%      Default |
+-------------------------------+----------------------+----------------------+
|   4  Tesla V100-SXM2...  On   | 00000000:85:00.0 Off |                    0 |
| N/A   55C    P0   252W / 300W |   5048MiB / 32480MiB |     82%      Default |
+-------------------------------+----------------------+----------------------+
|   5  Tesla V100-SXM2...  On   | 00000000:86:00.0 Off |                    0 |
| N/A   56C    P0   180W / 300W |   5056MiB / 32480MiB |     81%      Default |
+-------------------------------+----------------------+----------------------+
|   6  Tesla V100-SXM2...  On   | 00000000:89:00.0 Off |                    0 |
| N/A   59C    P0   244W / 300W |   5060MiB / 32480MiB |     79%      Default |
+-------------------------------+----------------------+----------------------+
|   7  Tesla V100-SXM2...  On   | 00000000:8A:00.0 Off |                    0 |
| N/A   54C    P0   255W / 300W |   5084MiB / 32480MiB |     83%      Default |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|    0     47478      C   python                                      5021MiB |
|    1     47478      C   python                                      5019MiB |
|    2     47478      C   python                                      5019MiB |
|    3     47478      C   python                                      5017MiB |
|    4     47478      C   python                                      5025MiB |
|    5     47478      C   python                                      5033MiB |
|    6     47478      C   python                                      5037MiB |
|    7     47478      C   python                                      5061MiB |
+-----------------------------------------------------------------------------+
```


tests/python/gpu/test_gluon_gpu.py
```diff
@@ -233,6 +236,47 @@ def test_symbol_block_fp16():
...
+    with net.name_scope():
+        net.add(nn.Conv2D(128, 3))
+        net.add(nn.LeakyReLU(0.1))
+        net.add(nn.Conv2D(largest_num_features, 3))
+        net.add(nn.LeakyReLU(0.1))
+        net.add(nn.Conv2D(1, 3))
+
+    net.hybridize()
+    net.initialize(mx.init.Normal(sigma=0.01), ctx=ctx)
+    mx.nd.waitall()
```
szha: since the in channels are not defined and the network didn't go through a forward, deferred initialization won't happen and no actual allocation is happening here.

eric-haibin-lin:
This is awesome! Is the change complete?

tests/python/gpu/test_gluon_gpu.py
DickJC123: I borrowed the approach from the mentioned issue, what about the lines:

https://github.com/apache/incubator-mxnet/blob/94614a545d0191cbbea95d944e470004eb5e9839/tests/python/gpu/test_gluon_gpu.py#L276-L277

DickJC123:
I'm still resolving issues with the CI test I introduced.  The rest of the PR is complete, but not yet pushed.

tests/python/gpu/test_gluon_gpu.py
szha: the first time it runs through this line of code, it will trigger actual allocation. I'm merely pointing out that the waitall is likely not needed.

DickJC123:
@KellenSunderland @piiswrong @eric-haibin-lin @zheng-da

It took a bit of playing with my newly introduced 'test_large_models' test to get it to fail on the correct CI nodes and not others.  The logs of commit 7bf8d51show the proper misbehavior present with the current codebase.  I've now committed the PR code, which demonstrates the issue is fixed.  See initial description for additional PR benefits.

As I look over the diffs now, I see the PR might look larger than it really is.  Much of the PR change involves moving most of SelectAlgo() code intact into a new callback function CuDNNAlgoSetter.

Also, I see now that I didn't mention introducing an environment variable MXNET_GPU_MEM_LARGE_ALLOC_ROUND_SIZE that defaults to the ideal value 2MB- my preference here rather than hardcoding a magic constant.

DickJC123:
To remotely debug the issues with my new test as run on the CI systems, I added stderr output that will be emitted on the same nosetests output line (even without the '-s').  Here shown during a failing run:

test_gluon_gpu.test_large_models ...  testing global memory of size 7988903936 ...  1200x1200  1190x1190  1180x1180  1170x1170 
ERROR

KellenSunderland:
Did some testing over the weekend and can verify this fixes a number of issues I've seen including our recent batch/slice test failures.  Will do a quick syntax pass tomorrow.

perl-package/AI-MXNetCAPI/mxnet.i
```diff
@@ -349,7 +349,7 @@ int MXGetGPUCount(int* out);
  * \param total_mem pointer to the integer holding total GPU memory
  * \return 0 when success, -1 when failure happens
  */
-int MXGetGPUMemoryInformation(int dev, int *out, int *out);
+int MXGetGPUMemoryInformation(int dev, uint64_t *out, uint64_t *out);
```
KellenSunderland: @szha Given the usage I'd consider this a non-breaking change, but it is technically changing types to an argument.  From a SemVer perspective are you ok with changing this in a minor update?

DickJC123: I also pondered this question and it's worth discussing.  As the MXGetGPUMemoryInformation interface stands now, it would only report global memory total and free bytes < 2MB correctly.  It's been many years since GPUs had so little global memory, so it's hard to imagine anyone out there using an interface so broken.  MXNet didn't wrap this C++ API in Python, although it was exposed in Perl I think.

KellenSunderland: Yeah I also grepped through the code and only see references to this API being used from Perl.

roywei:
@eric-haibin-lin @szha @zheng-da @anirudh2290 @sandeep-krishnamurthy @nswamy   could you help review? Thanks a lot!
This could fix #https://github.com/apache/incubator-mxnet/issues/12767 which is failing in multiple PRs

KellenSunderland:
Agree with @roywei, this PR will fix many issues and it would be great to get it merged soon.  Let's still do due diligence in terms of review as it's an important component.

tests/python/gpu/test_gluon_gpu.py
```diff
@@ -233,6 +236,45 @@ def test_symbol_block_fp16():
...
+    # The idea is to create models with large tensors of (say) 20% of the total memory.
+    # This in the past has given cudnnFind() trouble when it needed to allocate similar I/O's
+    # from the area carved out by the MXNET_GPU_MEM_POOL_RESERVE setting (by default 5%).
+    (free_mem_bytes, total_mem_bytes) = mx.context.gpu_memory_info(ctx.device_id)
+    start_size = tensor_size(0.20 * total_mem_bytes)
+    num_trials = 10
+    sys.stderr.write(' testing global memory of size {} ... '.format(total_mem_bytes))
+    sys.stderr.flush()
+    for i in range(num_trials):
+        sz = start_size - 10 * i
```
KellenSunderland: I'm not 100% sure but is there possibly a bug here?  start_size is roughly 20% of the available number of bytes on the GPU right?  It seems like we want to do a few trials and gradually increase I/O tensor sizes but we're subtracting -10 * i.  Is this what we want or do we want to start with 20% of the gpu's allocation and gradually work down?  I.e. something like?

```python
sz = start_size  * ( (10 - i)  / 10)
```

tests/python/gpu/test_gluon_gpu.py
```diff
@@ -233,6 +236,45 @@ def test_symbol_block_fp16():
...
+        (height, width) = (sz,sz)
+        sys.stderr.write(" {}x{} ".format(height,width))
+        sys.stderr.flush()
+        data_in = nd.random_uniform(low=0, high=255, shape=(1, 3, height, width),
```
KellenSunderland: Second question about this (and I may well be misunderstanding something here).  height and width are roughly both 20% of the number of bytes available on the GPU for i=0 here right?  Lets say we have a 12 GB card, we'd then ~2GB to allocate.  Wouldn't this result in a tensor much larger than 20% of the GPU space as an input because it would be ~2GB squared?  Also would it be 4x larger than intended as we're allocated float32s but we're basing that on the number of bytes available?

tests/python/gpu/test_gluon_gpu.py
DickJC123: We don't want to test the same size, since that would be a hit in the algo cache and cudnnFind() would not be re-run.  The idea is to test different sizes that are roughly the size as chosen initially.  I observed that with the progression of these models, the MXNet storage manager would slowly allocate up to 95% of memory by default (although much of it ends up in its free store).  At that point, if MXNet's storage manager is asked for a big allocation that doesn't match the size of something in the free store, it will trigger a beneficial ReleaseAll():
```
    if (free <= total * reserve_ / 100 || size > free - total * reserve_ / 100)
      ReleaseAll();
```
However, if instead cudnnFind is called at that point, the large allocations it wants to make for I/Os and workspace will cause an out of memory or 'no algo found' error, despite the fact that unused allocations are in the free store.  I'm using the existing Alloc/DirectFree API to possibly trigger a ReleaseAll() prior to calling cudnn within the algo-setting callback.

DickJC123: And to get back to your question more succinctly, there's nothing magic about the initial 1200x1200 starting point.  It was calculated based on the 8GB global memory of the CI machines.  I verified that the test initially failed, then passed, on 16GB and 32GB GPUs (with different sizes in each case).

KellenSunderland: Ah that makes total sense, you're just using dimensions to make sure we're not pulling the alg straight from the registry and bypassing the auto-tuning.  Gotcha, thanks Dick.

src/operator/nn/cudnn/cudnn_algoreg-inl.h
```diff
@@ -75,7 +81,7 @@ class CuDNNAlgoReg {
             bool add_to_weight,
             CuDNNAlgo<cudnnConvolutionFwdAlgo_t> *fwd,
             CuDNNAlgo<cudnnConvolutionBwdDataAlgo_t> *bwd,
-            CuDNNAlgo<cudnnConvolutionBwdFilterAlgo_t> *flt) {
+            CuDNNAlgo<cudnnConvolutionBwdFilterAlgo_t> *flt, AlgoSetter_t algo_setter) {
```
KellenSunderland: Would using a const ref for algo_setter save us a copy?

src/storage/pooled_storage_manager.h
```diff
@@ -57,6 +57,7 @@ class GPUPooledStorageManager final : public StorageManager {
   GPUPooledStorageManager() {
     reserve_ = dmlc::GetEnv("MXNET_GPU_MEM_POOL_RESERVE", 5);
     page_size_ = dmlc::GetEnv("MXNET_GPU_MEM_POOL_PAGE_SIZE", 4096);
+    large_alloc_round_size_ = dmlc::GetEnv("MXNET_GPU_MEM_LARGE_ALLOC_ROUND_SIZE", 2 * 1024 * 1024);
```
szha: I developed another rounding memory manager that uses a combination of exponential rounding size and linear rounding size and has shown good results. If adding rounding to GPU is desired by default, I think we can combine them into a single memory manager that offer these knobs.

let's address the memory pool change in a separate PR.

perl-package/AI-MXNetCAPI/mxnet.i
szha: I think @DickJC123 you meant 2GB? One alternative is to provide the new interface as MXGetGPUMemoryInformation64 and update the old interface in major version

src/operator/nn/cudnn/cudnn_convolution-inl.h
```diff
@@ -611,236 +612,274 @@ class CuDNNConvolutionOp {
...
-                                          filter_desc_,
-                                          CUDNN_CONVOLUTION_BWD_FILTER_SPECIFY_WORKSPACE_LIMIT,
-                                          workspace_byte,
-                                          &fastest_bwd_filt_algo));
-        back_algo_w_.Set(fastest_bwd_filt_algo, false);
+      if (i == nalgo) {
+        LOG(FATAL) << nalgo << " forward algorithms with minimum memory requirement "
+                   << min_memory_needs << " bytes have been tried. Workspace size is set to "
+                   << workspace_byte << " bytes, please consider reducing the batch/model size, "
+                   << "or increasing workspace size.";
```
szha: There are several messages that look alike. Consider wrapping it into a macro

python/mxnet/context.py
```diff
@@ -258,6 +258,30 @@ def num_gpus():
...
+    device_id : int, optional
+        The device id of the GPU device.
+
+    Raises
+    ------
+    Will raise an exception on any CUDA error.
+
+    Returns
+    -------
+    (free, total) : (int, int)
```
blac2kite: Minor: 'total' - is it referring to total used, total available or the total size of the physical GPU

src/operator/nn/cudnn/cudnn_algoreg-inl.h
blac2kite: Minor: Other parameters are on their own lines. Maybe algo_setter can go in its own line.

python/mxnet/context.py
DickJC123: I prefer to leave this as is.  Regarding int vs long, I'm not a Python wizard, but ints are 'plain integers' and longs have unlimited precision:
```
$ python
Python 2.7.12 (default, Dec  4 2017, 14:50:18) 
[GCC 5.4.0 20160609] on linux2
>>> import sys
>>> x  = sys.maxsize
>>> x
9223372036854775807
>>> type(x)
<type 'int'>
>>> y=2*x
>>> y
18446744073709551614L
>>> type(y)
<type 'long'>
>>> 
```
And unfortunately, there's not a real short answer to what 'total' memory means.  We're wrapping the cuda call cudaMemGetInfo(), and the NVIDIA documentation says:
```
Returns in *free and *total respectively, the free and total amount of memory available for allocation by the device in bytes.
```
Let's say you've got a GPU with published memory T.  The GPU driver puts some control structures like the page table in that memory, so call that driver overhead D.  Finally, your GPU may be driving a monitor, so a window manager is using the GPU with overhead W.  So what does the API return for 'total' in this scenario?  The answer is T - D.  The long answer then is: 'total' means the total memory available to both your MXNet process **and** other processes that may be using the GPU.  I don't know a way to suggest this succinctly without introducing more confusion.

eric-haibin-lin:
@DickJC123 could you address the build issue in CI?
```
src/c_api/c_api.cc:130:43: error: 'free_mem64' may be used uninitialized in this function [-Werror=maybe-uninitialized]

   *free_mem = static_cast<int>(free_mem64);

                                           ^

src/c_api/c_api.cc:131:45: error: 'total_mem64' may be used uninitialized in this function [-Werror=maybe-uninitialized]

   *total_mem = static_cast<int>(total_mem64);

```

DickJC123:
@eric-haibin-lin I've fixed the latest uninitialized var issue you pointed out and am passing CI.  Thanks to all for the careful review! @KellenSunderland has been following even the last tweeks and approves this PR.

KellenSunderland:
Yep, updates look good, did some testing and it fixes a few issues (including some CI failures we've been getting recently).

Thanks for the quick iteration @DickJC123.

PR #12847, opened 2018-10-17
[CI branch] script for building openblas from source

szha:
## Description ##
script for building openblas from source


szha:
@lanking520

Issue #12836, opened 2018-10-16
Running slow on Scala Benchmark on batch image input


lanking520:
Recently I wrote some code to test the performance of Scala benchmark Inference and found that it shows almost equal performance with single inference and batch inference. The core problems are the rebind on the input if you feed in a batch of images. This should be addressed to release the full strength of Scala batch inference.

lanking520:
@mxnet-label-bot, please add label [bug, scala, performance]

lanking520:
@kalyc No, it's allow the functionalities to do batch correctly, but we still not seen the performance improvement so please keep it open.

lanking520:
@mxnet-label-bot remove [bug]

lanking520:
This problem self-fixed. I add parallel execution in the Batch files, it is much faster now

PR #12848, opened 2018-10-17
Fix Batch input issue with Scala Benchmark

lanking520:
## Description ##
This PR is intended to solve the issue explained here:
https://github.com/apache/incubator-mxnet/issues/12836
Modification applied on Benchmark script as well as basic image classifier section.
@nswamy @piyushghai @yzhliu 

### Essentials ###

### DIFF
--- a/scala-package/core/src/main/scala/org/apache/mxnet/Shape.scala
+++ b/scala-package/core/src/main/scala/org/apache/mxnet/Shape.scala
@@ -34,7 +34,6 @@ class Shape(dims: Traversable[Int]) extends Serializable {
   def size: Int = shape.size
   def length: Int = shape.length
   def drop(dim: Int): Shape = new Shape(shape.drop(dim))
-  def dropBack(dim : Int) : Shape = new Shape(shape.dropRight(dim))
   def slice(from: Int, end: Int): Shape = new Shape(shape.slice(from, end))
   def product: Int = shape.product
   def head: Int = shape.head
--- a/scala-package/core/src/main/scala/org/apache/mxnet/Shape.scala
+++ b/scala-package/core/src/main/scala/org/apache/mxnet/Shape.scala
@@ -34,6 +34,7 @@ class Shape(dims: Traversable[Int]) extends Serializable {
   def size: Int = shape.size
   def length: Int = shape.length
   def drop(dim: Int): Shape = new Shape(shape.drop(dim))
+  def dropBack(dim : Int) : Shape = new Shape(shape.dropRight(dim))
   def slice(from: Int, end: Int): Shape = new Shape(shape.slice(from, end))
   def product: Int = shape.product
   def head: Int = shape.head
--- a/scala-package/examples/src/main/scala/org/apache/mxnetexamples/InferBase.scala
+++ b/scala-package/examples/src/main/scala/org/apache/mxnetexamples/InferBase.scala
@@ -21,7 +21,7 @@ import org.apache.mxnet._
 
 trait InferBase {
 
-  def loadModel(context: Array[Context]): Any
+  def loadModel(context: Array[Context], batchInference : Boolean): Any
   def loadSingleData(): Any
   def loadBatchFileList(batchSize: Int): List[Any]
   def loadInputBatch(source: Any): Any
--- a/scala-package/infer/src/main/scala/org/apache/mxnet/infer/ImageClassifier.scala
+++ b/scala-package/infer/src/main/scala/org/apache/mxnet/infer/ImageClassifier.scala
@@ -76,7 +76,7 @@ class ImageClassifier(modelPathPrefix: String,
                     topK: Option[Int] = None): IndexedSeq[IndexedSeq[(String, Float)]] = {
 
     val scaledImage = ImageClassifier.reshapeImage(inputImage, width, height)
-    val pixelsNDArray = ImageClassifier.bufferedImageToPixels(scaledImage, inputShape)
+    val pixelsNDArray = ImageClassifier.bufferedImageToPixels(scaledImage, inputShape.drop(0))
     inputImage.flush()
     scaledImage.flush()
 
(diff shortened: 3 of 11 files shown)


scala-package/examples/src/main/scala/org/apache/mxnetexamples/InferBase.scala
```diff
@@ -21,7 +21,7 @@ import org.apache.mxnet._
 
 trait InferBase {
 
-  def loadModel(context: Array[Context]): Any
+  def loadModel(context: Array[Context], batchInference : Boolean): Any
```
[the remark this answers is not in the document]
piyushghai: The examples still remain stand alone. The InferBase utility is only used, if one wants to measure the performance of a particular model in Scala using the inference APIs. 
That way they would not have to dig into the examples and add the metrics in each example.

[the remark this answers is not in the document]
nswamy: The problem I am pointing is all examples are overriding this trait. Adding metrics is not what is targeted in the examples, its to show the underlying APIs.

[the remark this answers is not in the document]
piyushghai: I see your point.

scala-package/core/src/main/scala/org/apache/mxnet/Shape.scala
```diff
@@ -34,6 +34,7 @@ class Shape(dims: Traversable[Int]) extends Serializable {
   def size: Int = shape.size
   def length: Int = shape.length
   def drop(dim: Int): Shape = new Shape(shape.drop(dim))
+  def dropBack(dim : Int) : Shape = new Shape(shape.dropRight(dim))
```
nswamy: I don't think you should introduce this public API, also the name is not intuitive.

lanking520: Is there any use case if the user would like to drop some dimension from back?

nswamy: suggest to let the user handle as required

lanking520: Ok, they can use `slice` as an alternative, will drop this

Roshrini:
@mxnet-label-bot [pr-awaiting-testing]

scala-package/infer/src/main/scala/org/apache/mxnet/infer/ImageClassifier.scala
```diff
@@ -182,7 +180,7 @@ object ImageClassifier {
     resizedImage.flush()
 
     // creating NDArray according to the input shape
-    val pixelsArray = NDArray.array(result, shape = inputImageShape)
+    val pixelsArray = NDArray.array(result, shape = new Shape(1 +: inputImageShape.toVector))
```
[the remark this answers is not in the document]
lanking520: @nswamy as we discussed offline, this is just a design decision. As an alternative, we will request the user to add a reshape if they would like to concatenate the Image NDArray altogether in one piece.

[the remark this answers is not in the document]
nswamy: The forward operation most often contains a axis for Batch, either you can do it here or outside the method just before you call forward. I think this is fine or if you guys think we should outside that is also fine -- you can use the NDArray.expand_dims() if it is a single dimension
The earlier version was wrong it was expecting the batch axis to be already there.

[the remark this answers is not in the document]
andrewfayres: I think it should be done outside. Currently it does exactly what the javadocs say it does. Takes a buffered image and turns it into an NDArray of shape inputShape. As it turns out this isn't actually what we wanted to be doing.

Also, this is a public api. As long as we have a reasonable alternative we should try to avoid changing it's behavior.

scala-package/infer/src/main/scala/org/apache/mxnet/infer/Predictor.scala
```diff
@@ -181,6 +181,7 @@ class Predictor(modelPathPrefix: String,
 
     // rebind with the new batchSize
     if (batchSize != inputBatchSize) {
+      logger.warn(s"Latency increased due to batchSize mismatch $batchSize vs $inputBatchSize")
```
nswamy: We don't want to warn, can you change this to Info?

scala-package/infer/src/main/scala/org/apache/mxnet/infer/ObjectDetector.scala
```diff
@@ -147,13 +150,16 @@ class ObjectDetector(modelPathPrefix: String,
...
-    val imageBatch = ListBuffer[NDArray]()
-    for (image <- inputBatch) {
-      val scaledImage = ImageClassifier.reshapeImage(image, width, height)
-      val pixelsNdarray = ImageClassifier.bufferedImageToPixels(scaledImage, inputShape)
-      imageBatch += pixelsNdarray
-    }
-    val op = NDArray.concatenate(imageBatch)
+    val inputBatchSeq = inputBatch.toIndexedSeq
+    val imageBatch = inputBatchSeq.indices.par.map(idx => {
+      val scaledImage = ImageClassifier.reshapeImage(inputBatchSeq(idx), width, height)
```
nswamy: we should parallelize this method using threads - you can create a JIRA for it and pick up whenever time permits

scala-package/infer/src/main/scala/org/apache/mxnet/infer/ImageClassifier.scala
```diff
@@ -147,9 +152,9 @@ object ImageClassifier {
     * returned by this method after the use.
     * </p>
     * @param resizedImage     BufferedImage to get pixels from
-    * @param inputImageShape  Input shape; for example for resnet it is (1,3,224,224).
+    * @param inputImageShape  Input shape; for example for resnet it is (3,224,224).
                               Should be same as inputDescriptor shape.
-    * @return                 NDArray pixels array
+    * @return                 NDArray pixels array with shape (1, 3, 224, 224) in NCHW format
```
nswamy: this does not return(1, anymore right?

scala-package/infer/src/main/scala/org/apache/mxnet/infer/ObjectDetector.scala
lanking520: `.par` already done the parallelism.

nswamy: oh cool, does it spawn threads?

Wei-1: `par` will spawn as many threads as the default `ForkJoinThreadPool` size if there is no other concurrent process.

Issue #12829, opened 2018-10-15
Incorrect docs link for NDArray


thomelane:
All links for `NDArray` in the Python documentation link to `mxnet.optimizer.NDArray` instead of `mxnet.ndarray.NDArray`. Could add confusion for users, since this is the wrong reference to NDArray. And it's only technically possible because NDArray is imported in `optimizer.py`.

* See https://mxnet.incubator.apache.org/api/python/ndarray/ndarray.html#mxnet.ndarray.save for an example.
* Under 'Parameters', 'data', click on 'NDArray'
* Goes (incorrectly) to https://mxnet.incubator.apache.org/api/python/optimization/optimization.html#mxnet.optimizer.NDArray

thomelane:
@mxnet-label-bot [Website, Doc]

lanking520:
@aaronmarkham can you take over from here? :)

aaronmarkham:
This seems related to this other issue (why can't Sphinx access the shorthand function calls):
https://github.com/apache/incubator-mxnet/issues/12318

Looking at the code section, it's clear that the erroneous links come from Sphinx's autodoc plugin. It is going through, importing things and checking the functions. 
https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/ndarray/utils.py#L235-L238
If it's linking to optimizer then maybe that's because the context that it is in at that point. 

I get the feeling that init / relative import stuff is messed up.  Help?

aaronmarkham:
@lanking520 noted that everything is fine in v1.3.0. 
I compared the docs build process and note that the optimizer folder is new in master where as in 1.3.0 it was just a optimizer.py file. 
We're seeing a bunch of errors in Sphinx:
```
/home/ubuntu/incubator-mxnet/python/mxnet/autograd.py:docstring of mxnet.autograd.mark_variables:None: WARNING: more than one target found for cross-reference u'NDArray': mxnet.optimizer.NDArray, mxnet.ndarray.NDArray
/home/ubuntu/incubator-mxnet/python/mxnet/autograd.py:docstring of mxnet.autograd.mark_variables:None: WARNING: more than one target found for cross-reference u'NDArray': mxnet.optimizer.NDArray, mxnet.ndarray.NDArray
[...]

PR #12871, opened 2018-10-18
fix broken Python IO API docs

### LINKED ISSUE #12854: Data Iterators documentation is broken
Many of the API references are not working - http://mxnet.incubator.apache.org/api/python/io/io.html#data-iterators

aaronmarkham:
## Description ##
Fixes #12854 
* This PR manually specifies members of the IO module so that the docs will render as expected. This is workaround in the docs to deal with a bug introduced in the Python code/structure since v1.3.0. See the comments for more info.

* This PR also fixes another issue that may or may not be related. Cross references to same-named entities like `name`, `shape`, or `type` are confusing Sphinx and it seems to just link to whatever it last dealt with that has the same name, and not the current module. To fix this you have to be very specific. Don't use `type`, use `np.type` if that's what you want. Otherwise you might end up with `mxnet.kvstore.KVStore.type`.

I mapped these to aliases, so the docs still have a shorthand reference. 
```
.. _name: mxnet.symbol.Symbol.name
.. _shape: mxnet.ndarray.NDArray.shape
```
So you will see me use `name_` in the docstring, and that will link up to the `_name` alias for `mxnet.symbol.Symbol.name` that is defined at the end of the documentation page's rst code.

**This is important for any future modules** - that they recognize this issue and make efforts to map the params and other elements.

### Essentials ###
- [x] Code is well-documented

### Preview
**To see the APIs populated:**
Go to the 1.3.0 docs: http://mxnet.incubator.apache.org/versions/1.3.0/api/python/io/io.html#data-iterators
Then switch to master. See how links are not there, and down in the API Reference none of those things are present.
Now go to my preview: http://34.201.8.176/versions/1.3.1/api/python/io/io.html

**To see the parameter mapping:**
It's broken on master and not even visible, but you can look at 1.3.0.
http://mxnet.incubator.apache.org/versions/1.3.0/api/python/io/io.html#mxnet.io.DataDesc.get_list
Note that the mappings pretty much work, but type seems messed up because I doubt it is intended to be from `kvstore`. Now look at my preview:
http://34.201.8.176/versions/1.3.1/api/python/io/io.html#mxnet.io.DataDesc.get_list

## Comments ##
Changes that have come in since v1.3.0 have caused some havoc with Sphinx. This could be a bug in the Python code or maybe Sphinx needs things to be a certain way to function properly.
* It can't distinguish between methods of the same name
* It can't use automodule to fetch all members automatically 
**This means a lot of the docs are currently broken!** I ask that some more people take a look because it could be a minor change in the Python versus having to go through and update every single bit of docs and doc strings in each module.

You can see in this PR what had to be done just for one module. Note the `recordio` hasn't been touched and Sphinx renders the docs just fine.
http://mxnet.incubator.apache.org/versions/1.3.0/api/python/io/io.html#data-iterators

### DIFF
--- a/docs/api/python/io/io.md
+++ b/docs/api/python/io/io.md
@@ -3,7 +3,7 @@
 ## Overview
 
 This document summarizes supported data formats and iterator APIs to read the
-data including
+data including:
 
 ```eval_rst
 .. autosummary::
@@ -68,8 +68,7 @@ A detailed tutorial is available at
 
 ## Helper classes and functions
 
-
-Data structures and other iterators provided in the ``mxnet.io`` packages.
+### Data structures and other iterators
 
 ```eval_rst
 .. autosummary::
@@ -83,7 +82,7 @@ Data structures and other iterators provided in the ``mxnet.io`` packages.
     io.MXDataIter
 ```
 
-Functions to read and write RecordIO files.
+### Functions to read and write RecordIO files
 
 ```eval_rst
 .. autosummary::
@@ -95,7 +94,7 @@ Functions to read and write RecordIO files.
     recordio.pack_img
 ```
 
-## Develop a new iterator
+## How to develop a new iterator
 
 Writing a new data iterator in Python is straightforward. Most MXNet
 training/inference programs accept an iterable object with ``provide_data``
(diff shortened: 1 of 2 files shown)


aaronmarkham:
@anirudhacharya 
@thomelane - this approach can probably fix #12829 LMK what you think... but also I'd like to know if there's not an easier way to fix what's going here...

Roshrini:
@mxnet-label-bot [pr-awaiting-review, Doc]

anirudhacharya:
@zhreshold can you please review this PR and merge it. Thanks

Issue #12868, opened 2018-10-18
mxnet.image.imdecode crash if input a empty buffer


frankfliu:
## Description

mxnet.image.imdecode() will crash and python process will die if pass a empty bytes buffer to the function.

This looks like a opencv bug, but we should do validation to avoid crash the process.

## Environment info (Required)

```
----------Python Info----------
('Version      :', '2.7.15')
('Compiler     :', 'GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)')
('Build        :', ('default', 'Oct  2 2018 12:50:38'))
('Arch         :', ('64bit', ''))
------------Pip Info-----------
('Version      :', '18.0')
('Directory    :', '/usr/local/lib/python2.7/site-packages/pip')
----------MXNet Info-----------
objc[37046]: Class CaptureDelegate is implemented in both /usr/local/opt/opencv/lib/libopencv_videoio.3.4.dylib (0x1179ed9d8) and /usr/local/lib/python2.7/site-packages/cv2/cv2.so (0x1257b02d0). One of the two will be used. Which one is undefined.
('Version      :', '1.3.1')
('Directory    :', '/Users/lufen/source/mxnet/python/mxnet')
Hashtag not found. Not installed from pre-built package.
----------System Info----------
('Platform     :', 'Darwin-16.7.0-x86_64-i386-64bit')
('system       :', 'Darwin')
('node         :', 'f45c89cb5707.ant.amazon.com')
('release      :', '16.7.0')
('version      :', 'Darwin Kernel Version 16.7.0: Thu Jun 21 20:07:39 PDT 2018; root:xnu-3789.73.14~1/RELEASE_X86_64')
----------Hardware Info----------
('machine      :', 'x86_64')
('processor    :', 'i386')
machdep.cpu.extfeatures: SYSCALL XD 1GBPAGE EM64T LAHF LZCNT PREFETCHW RDTSCP TSCI
machdep.cpu.leaf7_features: SMEP ERMS RDWRFSGS TSC_THREAD_OFFSET BMI1 AVX2 BMI2 INVPCID SMAP RDSEED ADX IPT FPU_CSDS
[...]

PR #12912, opened 2018-10-22
Add bytearray support back to imdecode (#12855, #12868)

frankfliu:
## Description ##
1. image.imdecode used to support bytearray, There is a regression introduced in: d4991a0e324b53d8d6e84decbd237ff6a5a3ab08. This commit trying to address the regression.

2. When input buffer is empty, OpenCV crash and Python process is killed. Add empty check to workaround this issue in OpenCV.

### Essentials ###

### Changes ###
- [X] Fixed backward compatibility issue regarding bytearray support.
- [X] Avoid OpenCV crash for empty input.

### DIFF
--- a/tests/python/unittest/test_image.py
+++ b/tests/python/unittest/test_image.py
@@ -92,6 +92,22 @@ class TestImage(unittest.TestCase):
             cv_image = cv2.imread(img)
             assert_almost_equal(image.asnumpy(), cv_image)
 
+    def test_imdecode_bytearray(self):
+        try:
+            import cv2
+        except ImportError:
+            return
+        for img in TestImage.IMAGES:
+            with open(img, 'rb') as fp:
+                str_image = bytearray(fp.read())
+                image = mx.image.imdecode(str_image, to_rgb=0)
+            cv_image = cv2.imread(img)
+            assert_almost_equal(image.asnumpy(), cv_image)
+
+    @raises(ValueError)
+    def test_imdecode_empty_buffer(self):
+        mx.image.imdecode(b'', to_rgb=0)
+
     def test_scale_down(self):
         assert mx.image.scale_down((640, 480), (720, 120)) == (640, 106)
         assert mx.image.scale_down((360, 1000), (480, 500)) == (360, 375)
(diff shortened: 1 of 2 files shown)


tests/python/unittest/test_image.py
```diff
@@ -92,6 +94,26 @@ def test_imdecode(self):
             cv_image = cv2.imread(img)
             assert_almost_equal(image.asnumpy(), cv_image)
 
+    def test_imdecode_bytearray(self):
```
zhreshold: I think the tests should target exception handling. 
Can you please also add a test that assertRaises if the buf is empty?

roywei:
@frankfliu  Thanks for the contribution!

tests/python/unittest/test_image.py
frankfliu: I added two test cases, 2nd one is testing empty buffer case: test_imdecode_empty_buffer

tests/python/unittest/test_image.py
```diff
@@ -92,6 +92,26 @@ def test_imdecode(self):
...
+        for img in TestImage.IMAGES:
+            with open(img, 'rb') as fp:
+                str_image = bytearray(fp.read())
+                image = mx.image.imdecode(str_image, to_rgb=0)
+            cv_image = cv2.imread(img)
+            assert_almost_equal(image.asnumpy(), cv_image)
+
+    @raises(ValueError)
+    def test_imdecode_empty_buffer(self):
+        try:
```
zhreshold: import cv2 is not necessary here right? Opencv is internally linked to mxnet library in C++ backend.

frankfliu: I actually don't know, I just copied from test case: test_imdecode(self)
And I have no idea why we catch ImportError exception.
Let me know if you want to remove the import.

zhreshold: I think it can be removed for test_imdecode_empty_buffer

frankfliu: OK, removed.

zhreshold:
@szha LGTM now, do you want another look?

szha:
@frankfliu thanks for the change. @yajiedesign @zhreshold thanks for the review.

PR #12913, opened 2018-10-22
fix the Scala paths issue for downloading script

lanking520:
## Description ##
Some of the downloading links are not valid anymore for the Scala package. Redirect them into more stable servers.
@nswamy @piyushghai

### Essentials ###

### DIFF
--- a/scala-package/examples/scripts/infer/imageclassifier/get_resnet_18_data.sh
+++ b/scala-package/examples/scripts/infer/imageclassifier/get_resnet_18_data.sh
@@ -37,5 +37,5 @@ if [ ! -f "$data_path" ]; then
   wget https://s3.us-east-2.amazonaws.com/scala-infer-models/resnet-18/resnet-18-symbol.json -P $data_path
   wget https://s3.us-east-2.amazonaws.com/scala-infer-models/resnet-18/resnet-18-0000.params -P $data_path
   wget https://s3.us-east-2.amazonaws.com/scala-infer-models/resnet-18/synset.txt -P $data_path
-  wget https://s3.amazonaws.com/model-server/inputs/kitten.jpg -P $image_path
+  wget https://s3.us-east-2.amazonaws.com/mxnet-scala/scala-example-ci/resnet152/kitten.jpg -P $image_path
 fi
--- a/scala-package/examples/scripts/infer/imageclassifier/get_resnet_data.sh
+++ b/scala-package/examples/scripts/infer/imageclassifier/get_resnet_data.sh
@@ -34,8 +34,8 @@ if [ ! -d "$image_path" ]; then
 fi
 
 if [ ! -f "$data_path" ]; then
-  wget http://data.mxnet.io/models/imagenet-11k/resnet-152/resnet-152-0000.params -P $data_path
-  wget http://data.mxnet.io/models/imagenet-11k/resnet-152/resnet-152-symbol.json -P $data_path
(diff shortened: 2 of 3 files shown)


scala-package/examples/scripts/infer/imageclassifier/get_resnet_18_data.sh
```diff
@@ -37,5 +37,5 @@ if [ ! -f "$data_path" ]; then
   wget https://s3.us-east-2.amazonaws.com/scala-infer-models/resnet-18/resnet-18-symbol.json -P $data_path
   wget https://s3.us-east-2.amazonaws.com/scala-infer-models/resnet-18/resnet-18-0000.params -P $data_path
   wget https://s3.us-east-2.amazonaws.com/scala-infer-models/resnet-18/synset.txt -P $data_path
-  wget https://s3.amazonaws.com/model-server/inputs/kitten.jpg -P $image_path
+  wget https://s3.us-east-2.amazonaws.com/mxnet-scala/scala-example-ci/resnet152/kitten.jpg -P $image_path
```
piyushghai: I believe you are using S3 bucket to keep the model artifacts.
But, should the bucket name be mxnet-scala or something else ?

lanking520: @piyushghai Previously when I set up with the CI, I created a bucket namely `mxnet-scala`. There are already `mxnet-model` or `mxnet-models`in the S3 bucket. I don't want to add additional folder inside to change something there.

piyushghai: My point is, if tomorrow there was a similar example for another language binding, say Java, then they would either have to : 
1. Create a new S3 bucket and host the same model files there.
OR
2. Point to an already existing centralized bucket.  
I don't think it's a scalable to follow point 1, which we seem to be doing here, because then we will end up having to duplicate hosting of files at other places as well. 

There are also other cases, where let's say one wants to replace a trained model with a new state of the art trained model, the fewer the number of places in which this change happens, the easier it will be right ?

lanking520: > My point is, if tomorrow there was a similar example for another language binding, say Java, then they would either have to :
> 
> 1. Create a new S3 bucket and host the same model files there.
>    OR
> 2. Point to an already existing centralized bucket.
>    I don't think it's a scalable to follow point 1, which we seem to be doing here, because then we will end up having to duplicate hosting of files at other places as well.
> 
> There are also other cases, where let's say one wants to replace a trained model with a new state of the art trained model, the fewer the number of places in which this change happens, the easier it will be right ?

But the fact is that some of the models with the same name cannot be used in Scala/Java. Such as Res152. Not all res152 models from MXNet Python can be successfully ported into MXNet Scala/Java especially those in data.mxnet.io. The cause behind it can be some diff on the precision (Float64) or input-type diffs. In that case, there has to be a Scala/Java model zoo somewhere to holds the code. Since Java is the child came from Scala, it should be fine for it to use the model directly from Scala.

piyushghai: I see your point there. But we might not be using the best practice here. 
I'll [reluctantly] approve this PR.

roywei:
@lanking520 Thanks for the contribution!

PR #12918, opened 2018-10-22
Update adversary attack generation example

ThomasDelteil:
## Description ##

Move adversarial example to Gluon, rewrite some comments

@thomelane  @safrooze

PR #12929, opened 2018-10-23
Update bilstm integer array sorting example

ThomasDelteil:
## Description ##
- Removed the symbolic implementation
- Created a Gluon implementation


anirudhacharya:
@ThomasDelteil why are examples being moved to using the Gluon API. I see that https://github.com/apache/incubator-mxnet/pull/12918 PR is also doing that. Won't it make sense if the example folder contains the symbolic implementation and the gluon code can go in the Straight Dope repository?

roywei:
@ThomasDelteil Thanks for the contribution!
I understand for most users, Gluon should be a better interface to MXNet, we should promote Gluon in examples. However, symbolic examples are still useful for some small use cases. We used pure symbolic mxnet in keras-mxnet project. So I would also vote for keeping symbolic examples in a separate folder instead of replacing it with Gluon.

ThomasDelteil:
We see a lot of user issues coming from trying to use symbolic rather than Gluon for the simplest tasks. I understand that symbolic is used for keras-mxnet, but that's an implementation detail, users should only use Gluon or keras-mxnet for implementing models, which is the focus of the example folders.

I think we should replace as many Symbolic examples as possible with Gluon ones, I am open, though very reluctant, to keep the symbolic examples in a `deprecated` folder for reference. I am afraid some people might end up there and try to use them rather than using the Gluon ones.

@safrooze what is your opinion on that?

anirudhacharya:
@ThomasDelteil what about "symbolic in the examples folder and gluon in the Straight Dope book"?

safrooze:
I agree with @ThomasDelteil. One problem I have observed in the community is the lack of clarity on what is MXNet's API and which examples should one use. Most people start by Googling for a solution to the problem they're working on and some arrive at the symbolic implementation. Then they look for supporting material and they also find Gluon. I think this is a very confusing experience for users.

example/vae/VAE_example.ipynb
```diff
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
```
szha: Could you clean up this change?
```suggestion
   "execution_count": 1,
```

PR #12955, opened 2018-10-24
[MXNET-1159]Added unit tests for making Resource Scope work in Java

piyushghai:
## Description ##
Add unit tests to ensure ResourceScope works in Java and to also demonstrate an example use-case.

It works by using try-with-resources technique in Java.

### Essentials ###

### Changes ###
- [x] Made Maven compiler plugin to use 1.8 in order to use Java Lambda expressions for try-with-resources in Java.

## Comments ##

### DIFF
--- a/scala-package/pom.xml
+++ b/scala-package/pom.xml
@@ -190,8 +190,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.3</version>
         <configuration>
-          <source>1.6</source>
-          <target>1.6</target>
+          <source>1.8</source>
+          <target>1.8</target>
           <encoding>UTF-8</encoding>
         </configuration>
       </plugin>
--- a/scala-package/core/src/test/java/org/apache/mxnet/javaapi/ResourceScopeTestSuite.java
+++ b/scala-package/core/src/test/java/org/apache/mxnet/javaapi/ResourceScopeTestSuite.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package org.apache.mxnet.javaapi;
+
+import org.apache.mxnet.NativeResourceRef;
+import org.apache.mxnet.ResourceScope;
+import org.junit.Test;
+
+import java.util.*;
+import java.util.concurrent.Callable;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class ResourceScopeTestSuite {
+
+    /**
+     * This is a placeholder class to test out whether NDArray References get collected or not when using
+     * try-with-resources in Java.
+     *
+     */
+    class TestNDArray  {
+        NDArray selfArray;
+
+        public TestNDArray(Context context, int[] shape) {
+            this.selfArray = NDArray.ones(context, shape);
+        }
+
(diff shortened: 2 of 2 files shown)


scala-package/pom.xml
```diff
@@ -190,8 +190,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.3</version>
         <configuration>
-          <source>1.6</source>
-          <target>1.6</target>
+          <source>1.8</source>
```
lanking520: Why change the version of Maven compiler plugin?

piyushghai: It's in the PR description : 
" Made Maven compiler plugin to use 1.8 in order to use Java Lambda expressions and try-with-resources in Java"

lanking520: > It's in the PR description :
> " Made Maven compiler plugin to use 1.8 in order to use Java Lambda expressions and try-with-resources in Java"

But does this mean we lose support for Java7 users? Or this is just a definition for Maven

piyushghai: I think it's just for Maven compilation. 

> > It's in the PR description :
> > " Made Maven compiler plugin to use 1.8 in order to use Java Lambda expressions and try-with-resources in Java"
> 
> But does this mean we lose support for Java7 users? Or this is just a definition for Maven

lanking520: After checking the https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html website here, this would lead to some problems for users using Java7. Here is what I have tried locally (with my JDK 1.8):
```
8c85904324fa:incubator-mxnet qingla$ javac -source 1.9
javac: invalid source release: 1.9
Usage: javac <options> <source files>
use -help for a list of possible options
8c85904324fa:incubator-mxnet qingla$ javac -source 1.8
javac: no source files
Usage: javac <options> <source files>
use -help for a list of possible options
8c85904324fa:incubator-mxnet qingla$ javac -source 1.7
javac: no source files
Usage: javac <options> <source files>
use -help for a list of possible options
```
Since we have not decided to fully abandon 1.7 users, this change may risk their usages

piyushghai: > After checking the https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html website here, this would lead to some problems for users using Java7. Here is what I have tried locally (with my JDK 1.8):
> 
> ```
> 8c85904324fa:incubator-mxnet qingla$ javac -source 1.9
> javac: invalid source release: 1.9
> Usage: javac <options> <source files>
> use -help for a list of possible options
> 8c85904324fa:incubator-mxnet qingla$ javac -source 1.8
> javac: no source files
> Usage: javac <options> <source files>
> use -help for a list of possible options
> 8c85904324fa:incubator-mxnet qingla$ javac -source 1.7
> javac: no source files
> Usage: javac <options> <source files>
> use -help for a list of possible options
> ```
> Since we have not decided to fully abandon 1.7 users, this change may risk their usages

We don't have the Java APIs available publicly right now, so we do not have Java users. 

I'm not sure what you are trying to indicate with the commands you pasted above. 

Also, here's Oracle's end of lifecycle roadmap link : https://www.oracle.com/technetwork/java/eol-135779.html
Oracle itself has stopped supporting Java 7 starting from April 2015.

nswamy: In the same page you linked look at the Note in the same page, its intended to use Java8 features and throw linkage errors, we haven't used any Java8 features yet but as we keep moving with a lot more Java APIs, we should move to Java8..I don't think it breaks for Java7 users, although worth checking..
I cannot find what the behavior of this setting is for version 3.3.0 of the plugin on their doc, may should check their github for [3.3 tag] (https://github.com/apache/maven-compiler-plugin/tree/maven-compiler-plugin-3.3/)

---
Note: Merely setting the target option does not guarantee that your code actually runs on a JRE with the specified version. The pitfall is unintended usage of APIs that only exist in later JREs which would make your code fail at runtime with a linkage error.
---
After verifying if it indeed breaks 1.7, I suggest to send out a email on the dev@ seek more opinions and revert if necessary.

nswamy: @piyushghai as discussed offline, could you please revert just this change and bring back only when required - I understand it doesn't impact but to be sure. You can start a discussion on dev@ about moving to Java8 and dropping support, many frameworks have moved to support Java8+

piyushghai: Okay sure, Will do it.

andrewfayres: Maybe I'm overlooking it but I don't see anything that we need Java 8 for. Try with resources was introduced in Java 7. The three lambda's that I see could easily be for loops.

I've got no issue with the idea of us moving to 8 and starting that discussion on the dev list. In the meantime I think we could move to 7 and rewrite the few lambdas.

Issue #8702, opened 2017-11-18
[DISCUSSION] Should we deprecate Makefile and only use CMake?


piiswrong:
1. Maintaining two build pipelines has been a lot of trouble.
2. We are planning to pull in dependencies and build/link them together with MXNet, similar to what we are doing with the pip package. CMake is better suited for this.

@cjolivier01 @yajiedesign @szha

yajiedesign:
+1 CMake is better

cjolivier01:
+1 for CMake

szha:
+1 for cmake. On specifics:
1. while we can provide the option to statically link dependencies, how to bring in dependencies is still the concern of a user. So the option of linking shared objects or specifying a search path should be supported by cmake. OpenCV and a number of other packages does it.
2. I haven't had enough experience in using cmake in mxnet to make a call on whether all use cases are supported in our cmake file. I know for sure that the makefile has served me well. So I'd suggest that during transition we only add new options to cmake while still maintaining makefile for the old options, until there's enough confidence in cmake.

cjolivier01:
I am not aware of any build option that CMake doesn’t support — I’ve been keeping it current.  At this point, I think it has more features than Makefile. If there is something missing, it’s an outlier and I haven’t run into it but it could be added quickly.

As for submodules, I think some we’d tend to want some of them to be part of a recursive clone, such as OpenMP — it’s small and they payoff for using it instead of gcc’s version is high.

piiswrong:
One potential problem is it  adds another dependency.
Its particularly painful for people using centos or redhat

cjolivier01:
One potential solution would be a script (python or bash) that fetches,
builds and installs CMake. An AMI that we are all familiar with does this.
It could be run on any platform if it’s python.

yajiedesign:
shared or static opencv depends on how user provide opencv.the opencv also use cmake build system and include "OpenCVConfig.cmake",this makes it easy for other cmake projects to refer to it.

yuewu001:
+1 cmake, especially for cross platform

tqchen:
CMake is great, and we cans till use the configuration based tricks https://github.com/dmlc/tvm/blob/master/CMakeLists.txt#L10  to rely on a config.cmake to customize the build option.

We might want to leave simple rules like lint to  Makefile (or as a python script) since it does not really help to bring them to cmake.

cjolivier01:
[...]

PR #12962, opened 2018-10-24
fix broken links and reorganize build from source page

aaronmarkham:
## Description ##
This PR fixes broken links and updates the build from source page.

## Preview
http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-12962/2/install/build_from_source.html

## Comments
The examples now use `cmake`. Reviewers please verify that the switches' names didn't change for things like opencv and cuda and cudnn between cmake and make.


roywei:
@aaronmarkham Thanks for the contribution!

docs/install/build_from_source.md
```diff
@@ -150,6 +140,32 @@ If you have a different processor you can still try to use MKL, but performance
...
+#### Install GPU Software
+
+If you want to run MXNet with GPUs, you must install [NVDIA CUDA and cuDNN](https://developer.nvidia.com/cuda-downloads).
+
+
+#### Install Optional Software
+
+These might be optional, but they're typically desirable as the extend or enhance MXNet's functionality.
+
+* [OpenCV](http://opencv.org/) - Image Loading and Augmentation. Each operation system has different packages and build from source options for OpenCV. Refer to your OS's link in the [Build Instructions by Operating System](#build-instructions-by-operating-system) section for further instructions.
```
roywei: nit: operating system

aaronmarkham: Thanks @roywei - fixed.

aaronmarkham:
@lebeg - Are the cmake flags in the examples alright? Should there be some others?

docs/install/build_from_source.md
```diff
@@ -183,53 +199,88 @@ nosetests --verbose tests/python/gpu/test_nccl.py
...
 ```
 
-* Build without using OpenCV:
+
+#### Recommended for Systems with NVIDIA GPUs and Intel CPUs
+* Build MXNet with `cmake` and install with MKL DNN, GPU, and OpenCV support:
 
 ```bash
-make USE_OPENCV=0
+cmake -j USE_CUDA=1 USE_CUDA_PATH=/usr/local/cuda USE_CUDNN=1 USE_MKLDNN=1
```
anirudhacharya: we usually give these default build options for user to use while building from source. Should we have a section describing different build options from here - https://github.com/apache/incubator-mxnet/blob/master/make/config.mk explaining what each of those do so that the user can make an informed personalized choice for himself? Provide the default options, but also add this section. Thoughts?

aaronmarkham: It's interesting that we tend to promote using MKLDNN for better performance and in the CMAKE instructions, they imply that it is turned on by default. 
However the config.mk has it turned off.
I thought it was supplied as a submodule and it is on by default. Is the config.mk out of date?

And yes, it should be mentioned what happens when you just run `make`.
Seems like right now you get:
opencv, openmp (that doesn't work on mac out the box), atlas || apple

Looking this config.mk file actually makes things even more unclear...

@anirudhacharya Can you propose the logic that is followed and the recommended flags per general situation?

nswamy:
I am wondering this if the build instructions are tested in CI to make sure we are aware of any breakage and users aren't surprised and frustrated.

nswamy:
can you please share a screenshot on how the install selection will look like?

docs/install/build_from_source.md
anirudhacharya: There is a conversation here - https://lists.apache.org/thread.html/ca3c858a712a66f4b3443dfb395d84e8fdf2c86e297bcd28a3a36ebd@%3Cdev.mxnet.apache.org%3E on including mkl-dnn in mxnet default build.

I think the recommended flags seem okay for now, the way it is. As and when it is decided to include MKL-DNN in the default build, then the recommended flags can also be updated.

What I was suggesting was to have a section in the documentation that describes the flags already provided in the config.mk file( each environment variable has a comment describing what it does). Please let me know if I need to help in any way here, and I would be glad to.

ankkhedia:
@nswamy ping!

lupesko:
@aaronmarkham have you tested the cmake flags in the doc?
Anything else the community can help with?

aaronmarkham:
I pulled the flags from the cmake PR. I haven't tested the variations. It would be great if the community provided tested/recommended cmake flags for the doc.

docs/install/build_from_source.md
lebeg: The flags are good

docs/install/build_from_source.md
```diff
@@ -1,93 +1,83 @@
...
+        * [with NCCL](#build-mxnet-with-nccl) (optional)
+        * [for C++](#build-mxnet-with-c++) (optional)
+        * [Usage Examples](#usage-examples)
+            * [systems with GPUs and Intel CPUs](#recommended-for-Systems-with-NVIDIA-GPUs-and-Intel-CPUs)
+            * [GPUs with non-Intel CPUs](#recommended-for-Systems-with-Intel-CPUs)
+            * [Intel CPUs](#recommended-for-Systems-with-Intel-CPUs)
+            * [non-Intel CPUs](#recommended-for-Systems-with-non-Intel-CPUs)
+2. [Install the language API binding(s)](#installing-mxnet-language-bindings) you would like to use for MXNet.
+MXNet's newest and most popular API is Gluon. Gluon is built into the Python binding. If Python isn't your preference, you still have more options. MXNet supports several other language APIs:
+    - [Python (includes Gluon)](../api/python/index.html)
```
nswamy: The flow does not seem right. We are taking about Installing API bindings and we redirect the users to API page, we should redirect them to the language install page instead.

nswamy:
I had offline discussion with @aaronmarkham and expressed a concern, I am noting here that we suggest to use `make` in some places, now we seem to be making this decision to advise our users to use CMake instead. I am not saying this is bad, but that we should make it consistent across.

For example Scala still uses `make`, if we start migrating to CMake and any new feature or flag is introduced in cmake files and forget to maintain in both the places, it will lead to inconsistency in performance/features, etc.,
Should we have discussion on dev@ and make a concerted effort to move to CMake ? WDYT?

nswamy:
@lebeg is that the command we run in CI?

lebeg:
@nswamy we had already discussions about this and moving to cmake exclusively is the way forward. At the time we are running both make and cmake in CI: [runtime_functions.sh](https://github.com/apache/incubator-mxnet/blob/master/ci/docker/runtime_functions.sh)

lebeg:
Here is a link to an old discussion: https://github.com/apache/incubator-mxnet/issues/8702

lebeg:
Discussion on [@dev](https://lists.apache.org/thread.html/14a3b8baf2abc7d482760d8c5dbfe20f680769782f1beb9b31120b62@%3Cdev.mxnet.apache.org%3E)

nswamy:
Thanks for sharing the links. I agree we should move to `cmake` still doesn't address the inconsistency that can arise, I am assuming you'll start a separate effort to deprecate Make.

Issue #12623, opened 2018-09-20
Feature Request NDArray.has_nan() method


zhanghang1989:
`ndarray.has_nan(NDArray)` and  its convenient form `NDArray.has_nan()`, which return `True` if there is `nan` in the data else return `False`. This is very helpful in debugging. 

Optimally has an option to output the per location prediction `ndarray.has_nan(x, gather=True)`, when `gather=False`, it should output a binary NDArray with the same shape as `x` and each location predicts the corresponding data is `nan` or not.

zhanghang1989:
@mxnet-label-bot [Feature Request, Operator]

eric-haibin-lin:
Similarly
NDArray.is_inf() https://www.tensorflow.org/api_docs/python/tf/is_inf

ChaiBapchya:
Just went through the Debug operators and looks like, there are quite of few of them
https://www.tensorflow.org/api_guides/python/control_flow_ops#Debugging_Operations
```
1. tf.is_finite
2. tf.is_inf
3. tf.is_nan
4. tf.verify_tensor_all_finite
5. tf.check_numerics
6. tf.add_check_numerics_ops
7. tf.Assert
8. tf.Print
```

Mostly I will work on that.

PR #12967, opened 2018-10-24
[MXNET-1173] Debug operators - isfinite and isinf

ChaiBapchya:
## Description ##
First two Debug operators. isinf was discussed in #12623 

### Essentials ###

### Changes ###
- [ ] Added unit tests
- [ ] Added functions in python front-end api

## Comments ##
- just the 2 of the total 6-7 operators that can be clubbed under "debug operators"

### DIFF
--- a/python/mxnet/ndarray/contrib.py
+++ b/python/mxnet/ndarray/contrib.py
@@ -19,6 +19,7 @@
 # pylint: disable=wildcard-import, unused-wildcard-import,redefined-outer-name
 """Contrib NDArray API of MXNet."""
 import math
+import numpy as np
 from ..context import current_context
 from ..random import uniform
 from ..base import _as_list
@@ -28,7 +29,7 @@ try:
 except ImportError:
     pass
 
-__all__ = ["rand_zipfian", "foreach", "while_loop", "cond"]
+__all__ = ["rand_zipfian", "foreach", "while_loop", "cond", "isinf", "isfinite"]
 
 # pylint: disable=line-too-long
 def rand_zipfian(true_classes, num_sampled, range_max, ctx=None):
@@ -460,3 +461,65 @@ def cond(pred, then_func, else_func):
         return then_func()
     else:
         return else_func()
+
+def isinf(data, ctx=None):
+    """Checks whether a given NDArray has infinity or not
+
+
+    Parameters
+    ----------
+    input : NDArray
+        An N-D NDArray.
+    ctx : Context
+        Device context of output. Default is current context.
+
+    Returns
+    -------
+    output: NDArray
+        The output NDarray with 1 and 0 indicating whether infinite or not.
+
+    Examples
+    --------
+    >>> data = mx.nd.array([np.inf, -np.inf, np.NINF, -1])
+    >>> output = mx.nd.contrib.isinf(data)
+    >>> output
+    [1. 1. 1. 0.]
+    <NDArray 4 @cpu(0)>
+    """
+    if ctx is None:
+        ctx = current_context()
+    # any(x in data for x in [np.inf, -np.inf])
+    # abs(data) == np.inf > TypeError: bad operand type for abs(): 'NDArray'
+    # [abs(x) == np.inf for x in data] > TypeError: bad operand type for abs(): 'NDArray'
+    return ndarray.NDArray.abs(data) == np.inf
+
+def isfinite(data, ctx=None):
+    """Checks whether a given NDArray has finit value or not
+
+
+    Parameters
+    ----------
+    input : NDArray
+        An N-D NDArray.
+    ctx : Context
+        Device context of output. Default is current context.
+
+    Returns
+    -------
+    output: NDArray
+        The output NDarray with 1 and 0 indicating whether finite or not.
+
+    Examples
+    --------
+    >>> data = mx.nd.array([np.inf, -np.inf, np.NINF, -1])
+    >>> output = mx.nd.contrib.isfinite(data)
+    >>> output
+    [0. 0. 0. 1.]
+    <NDArray 4 @cpu(0)>
+    """
+    if ctx is None:
+        ctx = current_context()
+    # any(x in data for x in [np.inf, -np.inf])
+    # abs(data) == np.inf > TypeError: bad operand type for abs(): 'NDArray'
+    # [abs(x) == np.inf for x in data] > TypeError: bad operand type for abs(): 'NDArray'
+    return ndarray.NDArray.abs(data) != np.inf
--- a/tests/python/unittest/test_ndarray.py
+++ b/tests/python/unittest/test_ndarray.py
@@ -1495,6 +1495,12 @@ def test_dlpack():
             mx.test_utils.assert_almost_equal(a_np, d_np)
             mx.test_utils.assert_almost_equal(a_np, e_np)
 
+@with_seed()
+def test_ndarray_is_inf():
+    shape = (3, 4, 2)
+    x = mx.nd.zeros(shape)
+    assert_almost_equal(mx.nd.contrib.isinf(x),False)
+
 if __name__ == '__main__':
     import nose
     nose.runmodule()


roywei:
@ChaiBapchya Thanks for the contribution!

python/mxnet/ndarray/contrib.py
```diff
@@ -460,3 +461,65 @@ def _to_python_scalar(inputs, type_, name):
         return then_func()
     else:
         return else_func()
+
+def isinf(data, ctx=None):
+    """Checks whether a given NDArray has infinity or not
```
anirudhacharya: the description should include the word element-wise. Something like, "Performs an element-wise check to determine if the NDArray contains an infinite element or not"

python/mxnet/ndarray/contrib.py
```diff
@@ -460,3 +461,65 @@ def _to_python_scalar(inputs, type_, name):
...
+    ----------
+    input : NDArray
+        An N-D NDArray.
+    ctx : Context
+        Device context of output. Default is current context.
+
+    Returns
+    -------
+    output: NDArray
+        The output NDarray with 1 and 0 indicating whether infinite or not.
```
anirudhacharya: This should specify that the output has the same shape as input, and that it will be 1 where the array element is equal to +/- infinity and is 0 in other places.

python/mxnet/ndarray/contrib.py
```diff
@@ -460,3 +461,65 @@ def _to_python_scalar(inputs, type_, name):
...
+    """
+    if ctx is None:
+        ctx = current_context()
+    # any(x in data for x in [np.inf, -np.inf])
+    # abs(data) == np.inf > TypeError: bad operand type for abs(): 'NDArray'
+    # [abs(x) == np.inf for x in data] > TypeError: bad operand type for abs(): 'NDArray'
+    return ndarray.NDArray.abs(data) == np.inf
+
+def isfinite(data, ctx=None):
+    """Checks whether a given NDArray has finit value or not
```
anirudhacharya: same comments as above.

tests/python/unittest/test_ndarray.py
```diff
@@ -1495,6 +1495,20 @@ def test_dlpack():
...
+    output = mx.nd.contrib.isinf(data)
+    expected_output = mx.nd.array([1,1,1,0])
+    assert_almost_equal(output.asnumpy(), expected_output.asnumpy())
+
+@with_seed()
+def test_ndarray_is_finite():
+    data = mx.nd.array([np.inf, -np.inf, np.NINF, -1])
+    output = mx.nd.contrib.isfinite(data)
+    expected_output = mx.nd.array([0,0,0,1])
+    assert_almost_equal(output.asnumpy(), expected_output.asnumpy())
```
anirudhacharya: make a check for `Nan` values

python/mxnet/ndarray/contrib.py
```diff
@@ -460,3 +461,65 @@ def _to_python_scalar(inputs, type_, name):
...
+    >>> output
+    [0. 0. 0. 1.]
+    <NDArray 4 @cpu(0)>
+    """
+    if ctx is None:
+        ctx = current_context()
+    # any(x in data for x in [np.inf, -np.inf])
+    # abs(data) == np.inf > TypeError: bad operand type for abs(): 'NDArray'
+    # [abs(x) == np.inf for x in data] > TypeError: bad operand type for abs(): 'NDArray'
+    return ndarray.NDArray.abs(data) != np.inf
```
anirudhacharya: this method should also check for `Nan` values and return 0 in places where the element == `Nan`

tests/python/unittest/test_ndarray.py
```diff
@@ -1495,6 +1495,20 @@ def test_dlpack():
             mx.test_utils.assert_almost_equal(a_np, d_np)
             mx.test_utils.assert_almost_equal(a_np, e_np)
 
+@with_seed()
+def test_ndarray_is_inf():
+    data = mx.nd.array([np.inf, -np.inf, np.NINF, -1])
```
anirudhacharya: randomize the shape and size of the array in this test case

tests/python/unittest/test_ndarray.py
```diff
@@ -1495,6 +1495,20 @@ def test_dlpack():
...
+    output = mx.nd.contrib.isinf(data)
+    expected_output = mx.nd.array([1,1,1,0])
```
anirudhacharya: do not hardcode the expected output. Calculate it using numpy.

python/mxnet/ndarray/contrib.py
ChaiBapchya: It returns 0 for NaN even now.

ChaiBapchya: for isinf it was returning 0 for NaN (the way it was written)
however for is_finite it wasn't returning 0
so now worked on it. Thanks
ultimately both isinf and isfinite return 0 for NaN

tests/python/unittest/test_ndarray.py
ChaiBapchya: Thanks for that.

python/mxnet/ndarray/contrib.py
```diff
@@ -28,7 +29,7 @@
 except ImportError:
     pass
 
-__all__ = ["rand_zipfian", "foreach", "while_loop", "cond"]
+__all__ = ["rand_zipfian", "foreach", "while_loop", "cond", "isinf", "isfinite"]
```
eric-haibin-lin: Could you also update https://github.com/apache/incubator-mxnet/blob/master/docs/api/python/ndarray/contrib.md#contrib ?

python/mxnet/ndarray/contrib.py
```diff
@@ -460,3 +461,96 @@ def _to_python_scalar(inputs, type_, name):
...
+    --------
+    >>> data = mx.nd.array([np.inf, -np.inf, np.NINF, -1])
+    >>> output = mx.nd.contrib.isinf(data)
+    >>> output
+    [1. 1. 1. 0.]
+    <NDArray 4 @cpu(0)>
+    """
+    if ctx is None:
+        ctx = current_context()
+    return ndarray.NDArray.abs(data) == np.inf
```
eric-haibin-lin: simply use `data.abs() == np.inf`?

python/mxnet/ndarray/contrib.py
```diff
@@ -460,3 +461,96 @@ def _to_python_scalar(inputs, type_, name):
...
+
+    Examples
+    --------
+    >>> data = mx.nd.array([np.inf, -np.inf, np.NINF, -1])
+    >>> output = mx.nd.contrib.isinf(data)
+    >>> output
+    [1. 1. 1. 0.]
+    <NDArray 4 @cpu(0)>
+    """
+    if ctx is None:
```
eric-haibin-lin: is ctx taking effect and tested? Why are we adding the ctx option to these ops?

python/mxnet/ndarray/contrib.py
ChaiBapchya: Sure. simpler. Thanks

python/mxnet/ndarray/contrib.py
ChaiBapchya: Done

python/mxnet/ndarray/contrib.py
```diff
@@ -460,3 +461,96 @@ def _to_python_scalar(inputs, type_, name):
...
+    >>> data = mx.nd.array([np.inf, -np.inf, np.NINF, -1])
+    >>> output = mx.nd.contrib.isfinite(data)
+    >>> output
+    [0. 0. 0. 1.]
+    <NDArray 4 @cpu(0)>
+    """
+    if ctx is None:
+        ctx = current_context()
+    is_data_not_nan = data == data
+    is_data_not_infinite = ndarray.NDArray.abs(data) != np.inf
```
apeforest: Same here: use data.abs()

python/mxnet/ndarray/contrib.py
```diff
@@ -460,3 +461,96 @@ def _to_python_scalar(inputs, type_, name):
...
+    Examples
+    --------
+    >>> data = mx.nd.array([np.nan, -1])
+    >>> output = mx.nd.contrib.isnan(data)
+    >>> output
+    [1. 0.]
+    <NDArray 2 @cpu(0)>
+    """
+    if ctx is None:
+        ctx = current_context()
```
apeforest: same here, why ctx is passed in while not used.

tests/python/unittest/test_ndarray.py
```diff
@@ -1495,6 +1495,45 @@ def test_dlpack():
...
+    data[0][0][1]=-np.inf
+    data[0][1][0]=np.nan
+    output = mx.nd.contrib.isinf(data)
+    expected_output = np.isinf(data.asnumpy())
+    assert_almost_equal(output.asnumpy(), expected_output.astype(int))
+    # astype since numpy functions default return type is boolean array instead of int
+
+@with_seed()
+def test_ndarray_is_finite():
+    random_dimensions = np.random.randint(3, 5)
```
apeforest: also add regular number in the data array

tests/python/unittest/test_ndarray.py
```diff
@@ -1495,6 +1495,45 @@ def test_dlpack():
             mx.test_utils.assert_almost_equal(a_np, d_np)
             mx.test_utils.assert_almost_equal(a_np, e_np)
 
+@with_seed()
+def test_ndarray_is_inf():
+    random_dimensions = np.random.randint(3, 5)
+    random_shape = [np.random.randint(2, 5) for i in range(random_dimensions)]
+    data = mxnet.test_utils.rand_ndarray(random_shape,'default')
+    data[0][0][0]=np.inf
```
apeforest: use create data with random dimension, but only set number in three dimensions?

tests/python/unittest/test_ndarray.py
```diff
@@ -1495,6 +1495,45 @@ def test_dlpack():
...
+    expected_output = np.isinf(data.asnumpy())
+    assert_almost_equal(output.asnumpy(), expected_output.astype(int))
+    # astype since numpy functions default return type is boolean array instead of int
+
+@with_seed()
+def test_ndarray_is_finite():
+    random_dimensions = np.random.randint(3, 5)
+    random_shape = [np.random.randint(2, 5) for i in range(random_dimensions)]
+    data = mxnet.test_utils.rand_ndarray(random_shape,'default')
+    data[0][0][0]=np.inf
```
apeforest: same here why only set three dimensions while you create data with random dimensions.

tests/python/unittest/test_ndarray.py
```diff
@@ -1495,6 +1495,45 @@ def test_dlpack():
...
+def test_ndarray_is_nan():
+    random_dimensions = np.random.randint(3, 5)
+    random_shape = [np.random.randint(2, 5) for i in range(random_dimensions)]
+    data = mxnet.test_utils.rand_ndarray(random_shape,'default')
+    data[0][0][0]=np.inf
+    data[0][0][1]=-np.inf
+    data[0][1][0]=np.nan
+    output = mx.nd.contrib.isnan(data)
+    expected_output = np.isnan(data.asnumpy())
+    assert_almost_equal(output.asnumpy(), expected_output.astype(int))
```
apeforest: this should be assert_equal not assert_almost_equal

tests/python/unittest/test_ndarray.py
```diff
@@ -1495,6 +1495,45 @@ def test_dlpack():
...
+def test_ndarray_is_finite():
+    random_dimensions = np.random.randint(3, 5)
+    random_shape = [np.random.randint(2, 5) for i in range(random_dimensions)]
+    data = mxnet.test_utils.rand_ndarray(random_shape,'default')
+    data[0][0][0]=np.inf
+    data[0][0][1]=-np.inf
+    data[0][1][0]=np.nan
+    output = mx.nd.contrib.isfinite(data)
+    expected_output = np.isfinite(data.asnumpy())
+    assert_almost_equal(output.asnumpy(), expected_output.astype(int))
```
apeforest: this should be assert_equal not assert_almost_equal

tests/python/unittest/test_ndarray.py
ChaiBapchya: This was after discussion with @anirudhacharya about how to verify that cases of infinite, negative infinite and NaN are handled.

python/mxnet/ndarray/contrib.py
ChaiBapchya: As an after thought, nothing context specific is being done. So removing it.

python/mxnet/ndarray/contrib.py
ChaiBapchya: As an after thought, nothing context specific is being done. So removing it.

tests/python/unittest/test_ndarray.py
ChaiBapchya: the function - `rand_ndarray` in `/python/mxnet/test_utils` is meant to do exactly that - initialize the array with random numbers

ChaiBapchya: Anyway, incorporated that nonetheless.

tests/python/unittest/test_ndarray.py
ChaiBapchya: Makes sense. Since I am passing the same data, it is necessary that it's exactly equal (not "almost")
Thanks for pointing that out.

tests/python/unittest/test_ndarray.py
anirudhacharya: the dimensions for setting of nan and np.inf values can be randomized too.

PR #12975, opened 2018-10-25
Implemented a regression unit test for #11793

asadoughi:
## Description ##
When using C++-based iterators, it's important that only a single batch is referenced at a time. Because C++ iterators are exposed to the Python code through a C API, there is no concept of reference counting. Hence, typically C++ iterators will deallocate a batch when next() is called on them. So, we need to make sure the Python code only references a single batch at a time, otherwise the Python code will attempt to access freed memory, resulting in either (a) garbage accuracy or (b) a segmentation fault.

The test passes with the latest mxnet build. I verified it failed on previous releases, such as mxnet==1.2.0.

### Essentials ###

### Changes ###
- No feature changes.

## Comments ##
We encountered an issue with garbage accuracy on an older MXNet release with our custom C++ iterators and saw that it was fixed in the latest master, but without a regression test. So, here I have implemented such a regression test.

### DIFF
--- a/tests/python/unittest/test_module.py
+++ b/tests/python/unittest/test_module.py
@@ -812,6 +812,65 @@ def test_forward_types():
     assert mod.predict(data1).shape == (1, 10)
 
 
+def test_reference_single_batch_during_fit():
+    """
+    When using C++-based iterators, it's important that only a single batch is referenced at a time. Because C++
+    iterators are exposed to the Python code through a C API, there is no concept of reference counting. Hence,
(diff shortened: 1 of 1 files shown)


leleamol:
@roywei I took a look but the code or changes are related to mxnet internal implementation and not in the cpp-package.
I lack the sufficient background in that functionality to review this code.

asadoughi:
Correct, there is no C++ code in this pull request. Perhaps @szha could take a look since @szha approved the related pull request https://github.com/apache/incubator-mxnet/pull/11793? Thanks!

ankkhedia:
@szha Could you please take a look?

PR #12926, opened 2018-10-23
parallelize NDArray::Copy<cpu, cpu> when data size is large

XiaotaoChen:
## Description ##
1. Parallelize NDArray::Copy<cpu, cpu> OP by OpenMP when data size bigger than MXNET_CPU_PARALLEL_COPY_SIZE.
2. introduce the environment variable named: MXNET_CPU_PARALLEL_COPY_SIZE (default 200,000) . When data size is bigger than this threshold, NDArray::Copy is implemented by OpenMP with the Recommended OMP Thread Count, otherwise implemented by default.

## Comments ##
We are optimizing [Sockeye](https://github.com/awslabs/sockeye)'s performance on CPU and found that the CopyCPU2CPU operation take too much time, probably accounting for 14% of the overall time (detail data as below). So we parallelized NDArray::Copy<cpu, cpu> operator. 

As far as i know, NDArray::Copy<cpu, cpu> is called in the front-end by those functions: **nd.copy(), nd.copyto(), nd.asnumpy(), nd.array(),nd.ones(), nd.zeros()** . **Those functions can gain 2-14 times speedup when data size is range from 20000 to 2e8.  And Sockeye's perf increased from 17.10 sent/sec to 19.19 sent/sec on CPU inference, CopyCPU2CPU time overhead dropped from 14.68% to 2.3%.**  the detail data as below.

the logs and scripts are here: [logs and scripts](https://github.com/XiaotaoChen/incubator-mxnet/tree/cxt-test/tests/cxt-test/parallel-copy)
@pengzhao-intel
### detail data

**hardware: skx-8180 single socket(28 cores)**

| functions\size | 20,000(shape:[20000,1]) |              |             | 200,000(shape:[20000,10]) |              |             | 2000,000(shape:[20000,100]) |              |           | 20,000,000(shape:[20000,1000]) |              |            | 200,000,000(shape:[20000,10000]) |              |            |
| -------------- | ----------------------- | ------------ | ----------- | ------------------------- | ------------ | ----------- | --------------------------- | ------------ | --------- | ------------------------------ | ------------ | ---------- | -------------------------------- | ------------ | ---------- |
| APIs           | default(us)             | parallel(us) | speedup     | default(us)               | parallel(us) | speedup     | default(us)                 | parallel(us) | speedup   | default(us)                    | parallel(us) | speedup    | default(us)                      | parallel(us) | speedup    |
[...]

PR #12997, opened 2018-10-28
A better take forward kernel for CPU

HyperZealot:
## Description ##
The same as title.

### Essentials ###

### Changes ###

## Comments ##
~5x improvement on CPU with this kernel

### DIFF
--- a/src/operator/tensor/indexing_op.cc
+++ b/src/operator/tensor/indexing_op.cc
@@ -28,6 +28,28 @@
 namespace mxnet {
 namespace op {
 
+template<bool clip = true>
+struct TakeCPU {
+  // assume that idx have been flattened to a 1-D tensor (N,)
+  // assume that out_data and in_data have been flattened to 2-D tensors, (N, M) and (K, M)
+  // M is the number of columns of in_data and out_data
+  // K is the number of rows of in_data
+  // i is the index of out_data
+  template<typename DType, typename IType>
+  MSHADOW_XINLINE static void Map(int i, DType* out_data, const DType* in_data,
+                                  const IType* idx, const size_t M, const int64_t K) {
+    int64_t j = static_cast<int64_t>(idx[i]);
+    if (clip) {
+      if (j <= 0) j = 0;
+      else if (j >= K) j = K - 1;
+    } else {
+      j = j % K;
+      j += (j < 0) ? K : 0;
+    }
+    std::memcpy(out_data + i * M, in_data + j * M, M * sizeof(DType));
+  }
+};
+
 /*
  * \brief returns true if all indices are between [min, max]
  * \param data_ptr the indices to check
@@ -48,6 +70,29 @@ bool CheckIndexOutOfBound(const DType* data_ptr, size_t data_size,
   return is_valid;
 }
 
+// Embedding forward implementation with dense weight
+template<>
+void EmbeddingOpForwardDnsImpl<cpu>(mshadow::Stream<cpu>* s,
+                                    const TBlob& data,
+                                    const TBlob& weight,
+                                    const OpReqType req,
+                                    const TBlob& output) {
+  using namespace mxnet_op;
+  const TShape& ishape = data.shape_;
+  const TShape& oshape = output.shape_;
+
+  MSHADOW_TYPE_SWITCH(output.type_flag_, DType, {
+    MSHADOW_TYPE_SWITCH(data.type_flag_, IType, {
+      Tensor<cpu, 1, IType> idx = data.get_with_shape<cpu, 1, IType>(
+        Shape1(ishape.ProdShape(0, ishape.ndim())), s);
+      Tensor<cpu, 2, DType> wmat = weight.get<cpu, 2, DType>(s);
+      Tensor<cpu, 2, DType> out = output.get_with_shape<cpu, 2, DType>(
+        Shape2(oshape.ProdShape(0, oshape.ndim()-1), oshape[oshape.ndim()-1]), s);
+      Kernel<TakeCPU<true>, cpu>::Launch(s, oshape.Size() / wmat.shape_[1], out.dptr_, wmat.dptr_,
+                                         idx.dptr_, wmat.shape_[1], wmat.shape_[0]);
+    });
+  });
+}
 
 template<>
 void SparseEmbeddingOpForwardRspImpl<cpu>(const OpContext& ctx,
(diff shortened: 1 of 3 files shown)


pengzhao-intel:
@HyperZealot Thanks for the efforts for this great improvements :)
Do you mind pointing out which model/test is tested with 5X  speedup?

HyperZealot:
@pengzhao-intel Hello, I have encountered this problem on my own model so I don't want to share that directly, but I can provide a small script to measure this, would you like to see that?

HyperZealot:
@pengzhao-intel 
script:
```Python
import mxnet as mx
import random
from mxnet.test_utils import rand_ndarray, check_speed

num_rows = 1000000
test_rows = 20000
num_cols = 512
shape = (num_rows, num_cols)
data = rand_ndarray(shape, stype='default')
print(data.dtype)
x = mx.sym.Variable('data')
idx = mx.sym.Variable('indices')
sym = mx.sym.take(a=x, indices=idx, axis=0)
trials = int(num_rows / test_rows)
total_time = 0.0
all_indices = [i for i in range(num_rows)]
random.shuffle(all_indices)
for i in range(trials):
    indices = mx.nd.array(all_indices[i * test_rows: (i + 1) * test_rows])
    total_time += check_speed(sym, location={'data': data, 'indices': indices}, ctx=mx.cpu(), N=100, typ="forward")
print(test_rows * num_cols * 4 * trials / total_time / 1024 / 1024 / 1024, "GB/s")
```
Before with mxnet-mkl release version: 2.0371343465121163 GB/s
After: 12.2486153514499 GB/s
~6x speedup with this new one

pengzhao-intel:
@HyperZealot thanks for your info and script. We also have an optimized version of `take` in local and plan to upstream to the official repo. 
@rongzha1 will try your script and compare the performance of both implementations.
I think we can combine two solutions together  :)

src/operator/tensor/indexing_op.cc
```diff
@@ -28,6 +28,28 @@
...
+struct TakeCPU {
+  // assume that idx have been flattened to a 1-D tensor (N,)
+  // assume that out_data and in_data have been flattened to 2-D tensors, (N, M) and (K, M)
+  // M is the number of columns of in_data and out_data
+  // K is the number of rows of in_data
+  // i is the index of out_data
+  template<typename DType, typename IType>
+  MSHADOW_XINLINE static void Map(int i, DType* out_data, const DType* in_data,
+                                  const IType* idx, const size_t M, const int64_t K) {
+    int64_t j = static_cast<int64_t>(idx[i]);
```
rongzha1: Can you tell why  use idx[i]?  Origin is idx[i/M] 
Thanks

rongzha1: origin is 
int j = static_cast<int>(idx[i/M]);

rongzha1: OK I see. But what if N%M != 0,  the remaining N%M part ?

src/operator/tensor/indexing_op.cc
```diff
@@ -48,6 +70,29 @@ bool CheckIndexOutOfBound(const DType* data_ptr, size_t data_size,
...
+  const TShape& oshape = output.shape_;
+
+  MSHADOW_TYPE_SWITCH(output.type_flag_, DType, {
+    MSHADOW_TYPE_SWITCH(data.type_flag_, IType, {
+      Tensor<cpu, 1, IType> idx = data.get_with_shape<cpu, 1, IType>(
+        Shape1(ishape.ProdShape(0, ishape.ndim())), s);
+      Tensor<cpu, 2, DType> wmat = weight.get<cpu, 2, DType>(s);
+      Tensor<cpu, 2, DType> out = output.get_with_shape<cpu, 2, DType>(
+        Shape2(oshape.ProdShape(0, oshape.ndim()-1), oshape[oshape.ndim()-1]), s);
+      Kernel<TakeCPU<true>, cpu>::Launch(s, oshape.Size() / wmat.shape_[1], out.dptr_, wmat.dptr_,
```
rongzha1: As above. What if   oshape.Size() % wmat.shape_[1]   != 0;   the remaining  oshape.Size() % wmat.shape_[1]  bit ?

src/operator/tensor/indexing_op.cc
```diff
@@ -227,6 +272,74 @@ void TakeOpForwardCsrImpl<cpu>(const TakeParam& params,
...
+  const TShape& oshape = outputs[take_::kOut].shape_;
+
+  Stream<cpu> *s = ctx.get_stream<cpu>();
+  const int actual_axis = param.axis + ((param.axis < 0) ? arrshape.ndim() : 0);
+
+  MSHADOW_TYPE_SWITCH(outputs[0].type_flag_, DType, {  // output data type
+    MSHADOW_TYPE_SWITCH(inputs[1].type_flag_, IType, {  // index data type
+      if (actual_axis == 0) {
+        if (param.mode == take_::kClip) {
+          Kernel<TakeCPU<true>, cpu>::Launch(s, idxshape.Size(),
```
rongzha1: Same as above

src/operator/tensor/indexing_op.cc
```diff
@@ -227,6 +272,74 @@ void TakeOpForwardCsrImpl<cpu>(const TakeParam& params,
...
+                                             outputs[take_::kOut].dptr<DType>(),
+                                             inputs[take_::kArr].dptr<DType>(),
+                                             inputs[take_::kIdx].dptr<IType>(),
+                                             oshape.Size()/idxshape.Size(), arrshape[0]);
+        } else {
+          Kernel<TakeCPU<false>, cpu>::Launch(s, idxshape.Size(),
```
rongzha1: same as above.  What if oshape.Size()%idxshape.Size() != 0?

HyperZealot:
@rongzha1 Hi, all shapes here are inferred so they are guaranteed to satisfy divisibility requirements.

ankkhedia:
@HyperZealot  Thanks for the awesome contribution!
@rongzha1 It seems all your comments have been addressed. Could you please review the PR again 

@mxnet-label-bot [pr-awaiting-review, operator]

rongzha1:
> @rongzha1 Hi, all shapes here are inferred so they are guaranteed to satisfy divisibility requirements.

Thanks for your reminder
There's still one question.  When you use memcpy, can you add   if(size> N) ? 
Some model M is very small, such as wide_deep we are doing optimization (https://github.com/apache/incubator-mxnet/tree/master/example/sparse/wide_deep) M is 8. If you use memcpy to copy 8 float every time, overhead is much big.

I test it use your script using our wide deep parameters 
num_rows = 800
test_rows = 8
num_cols = 61400


  | M=8
-- | --
for | 131.7215735239637 GB/s
memcpy | 58.68885431408931 GB/s

src/operator/tensor/indexing_op.cc
```diff
@@ -28,6 +28,28 @@
...
+                                  const IType* idx, const size_t M, const int64_t K) {
+    int64_t j = static_cast<int64_t>(idx[i]);
+    if (clip) {
+      if (j <= 0) j = 0;
+      else if (j >= K) j = K - 1;
+    } else {
+      j = j % K;
+      j += (j < 0) ? K : 0;
+    }
+    std::memcpy(out_data + i * M, in_data + j * M, M * sizeof(DType));
```
rongzha1: can you do some test to get a best some_size :
if( M > some_size) {
  std::memcpy(out_data + i * M, in_data + j * M, M * sizeof(DType))
} else {
      const int jM = j*M;
      const int iM = i*M;
      for (int n = 0; n < M; n++) {
        out_data[iM + n] = in_data[jM + n];
      }
}

Small M will have a big overhead

HyperZealot:
@rongzha1 From your code M should be num_cols(61400), that is a pretty big number.

rongzha1:
Thanks for your info.  I have asked admin about the server config: 
CPU: skylake 8180  (56 physical cores)
Memory bandwidth :
255GB/s = 2.66*12*8(2666MHz DDR4) 
and run STREAM:  (omp use all 56 core)
-------------------------------------------------------------
Function    Best Rate MB/s  Avg time     Min time     Max time
Copy:          129854.6     0.001265     0.001232     0.001322
Scale:         121861.0     0.001392     0.001313     0.001653
Add:           115944.8     0.002097     0.002070     0.002171
Triad:         126780.0     0.001911     0.001893     0.001948

maybe some config error, when I set OMP_NUM_THREADS=28 (half its total cores), performance even better.

HyperZealot:
@rongzha1 TBH that's also what I got on my side. my CPU has 4 physical cores, so when I'm using OMP_NUM_THREADS=4 I get the best performance (~30GB/s).

HyperZealot:
@rongzha1 BTW I don't think your machine is a representative example for most users'...

pengzhao-intel:
@HyperZealot FYI, This is a different scenario for the enterprise users who will use the high bin machines (servers, like SKX 6148, 8180) with multiple cores (20+) per socket and high bandwidth memory for their services, such as AWS cloud. So, we need to make the change more general to cover various cases.

Thanks for the improvements again :)

HyperZealot:
I don't think the workloads provided by @rongzha1 is suitable for determining the memory bandwidth, based on the following arguments:
0. For each input, the effective working set size is test_rows*num_cols*4 bytes. Here's the working set size for each input:
(1M, 20k, 512) = ~39MB
(1M, 20k, 8) = ~0.6MB
(800, 8, 61400) = ~1.90MB
1. the benchmark script runs 100 trials for each group of indices, so if the cache's size is greater than the working set size, then the source data can be totally loaded in cache after paying for compulsory misses during the 1st trial, then after 1st trial you're actually measuring the cache bandwidth.
2. Although the benchmark runs 100 trials for the same input, users usually on run once for each input so they always pay for the compulsory misses and the bottleneck is the memory speed.
3. With a bit of search I found Skylake 8180 has a 38.5MB L3 cache, so both (800, 8, 61400) and (1M, 20k, 8) workloads can totally be cached into the L3 cache, so the measurements performed on those cannot accurately compare the performance on increasing memory bandwidth consumption of different versions of code.
4. If you really want to showcase the effect of different versions on num_cols=8, maybe you can switch to a CPU with smaller cache or you can increase test_rows to make the working set larger than your L3 cache size.
5. I tested (50M, 1M, 8) (~30.5M working set size, definitely greater than my L3 cache size) on my own machine and got 7.90 GB/s for "for" version and 9.80 GB/s for "memcpy" version.

pengzhao-intel:
@HyperZealot I suggest to using an environment `MXNET_CPU_PARALLEL_COPY_SIZE` to decide when use `memcpy` or OMP copy as in #12926. The num_cols=8 case in here is just an indicator to cover different scenario.

HyperZealot:
@pengzhao-intel I think what @rongzha1 is suggesting here is to have a special case for smaller sizes?
I've shown that "memcpy" could give more bandwidth than 'for' even when num_cols is small so maybe such special handling is indeed unnecessary? For your comment, are you suggesting that we change the kernel launch to omp loop when num_cols is large?

HyperZealot:
@pengzhao-intel Thanks for your approval!

PR #13000, opened 2018-10-29
fix quantize_graph pass error when there're multiple outputs from a single node

ciyongch:
## Description ##
The PR is to fix an error of quantize_graph pass which can't handle multiple outputs (like split op followd by convolution or fullyconnected) correctly. The current implementation is only quantizing for the first outputs but not the rest ones. In fact, for the case with multiple outputs from a single node, it requires to quantize separately for each output but not only the first one.

Check the attachment for the symbols before and after the fix.
* symbols from current implementation: [original_split_quantize.pdf](https://github.com/apache/incubator-mxnet/files/2524495/original_split_quantize.pdf)
* symbols from fixed version: [fixed_split_quantize.pdf](https://github.com/apache/incubator-mxnet/files/2524494/fixed_split_quantize.pdf)

### Essentials ###

### Changes ###

## Comments ##

@pengzhao-intel @ZhennanQin

### DIFF
--- a/src/operator/quantization/quantize_graph_pass.cc
+++ b/src/operator/quantization/quantize_graph_pass.cc
@@ -131,7 +131,7 @@ Graph QuantizeGraph(Graph &&src) {
   // graph. Key is the currently visited graph's node pointer, and value is a copied node of the key
   // node. The existing key's value may be updated with the newly created quantize/dequantize op.
   std::unordered_map<Node*, NodePtr> mirror_map;
-  nnvm::NodeEntryMap<NodeEntry> multiple_outputs_entry_map;
+  nnvm::NodeEntryMap<NodeEntry> mirror_entry_map;
   DFSVisit(src.outputs, [&](const NodePtr& node) {
     NodePtr new_node = Node::Create();
     // If the currently visited node needs quantization, insert a quantize op node before the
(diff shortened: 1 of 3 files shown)


pengzhao-intel:
@reminisce Could you help take a review?

ankkhedia:
@pengzhao-intel Thanks for your contribution!
@reminisce Could you please help review the PR
@mxnet-label-bot [pr-awaiting-review, operator]

ciyongch:
thanks @lupesko .
This scenario is quite common in unroll RNN domain.

src/operator/quantization/quantize_graph_pass.cc
```diff
@@ -156,29 +158,54 @@ Graph QuantizeGraph(Graph &&src) {
...
         } else if (!NeedQuantize(e.node, excluded_nodes) &&
                    (mirror_node->op() == nullptr ||
-                    mirror_node->op()->name != "_contrib_quantize")) {
+                    mirror_node->op()->name != "_contrib_quantize") &&
+                    (!multiple_outputs_entry_map.count(e))) {
+          // When there're multiple entrys outgoing from a single node, need to add entry
+          // index (or output name) into quantize/min/max node to distinguish them.
+          // Or the output name is not ending with 'output', just put the output name here
+          // to better align with calibration phase. No need to change name to weights/bias.
+          std::string subfix = "";
```
apeforest: Do you mean suffix?

src/operator/quantization/quantize_graph_pass.cc
```diff
@@ -117,9 +117,10 @@ inline bool NeedQuantize(NodePtr node, const std::unordered_set<std::string>& ex
 }
 
 Graph QuantizeGraph(Graph &&src) {
-  static auto& quantized_op_map = Op::GetAttr<mxnet::FQuantizedOp>("FQuantizedOp");
-  static auto& need_requantize_map = Op::GetAttr<mxnet::FNeedRequantize>("FNeedRequantize");
-  static auto& avoid_quantize_input_map =
+  static const auto& flist_outputs = nnvm::Op::GetAttr<nnvm::FListOutputNames>("FListOutputNames");
```
apeforest: nit: output_list is a better

src/operator/quantization/quantize_graph_pass.cc
```diff
@@ -156,29 +158,54 @@ Graph QuantizeGraph(Graph &&src) {
           new_node->inputs.emplace_back(mirror_entry);
         } else if (!NeedQuantize(e.node, excluded_nodes) &&
                    (mirror_node->op() == nullptr ||
-                    mirror_node->op()->name != "_contrib_quantize")) {
+                    mirror_node->op()->name != "_contrib_quantize") &&
```
apeforest: Make _contrib_quantize a constant variable. Or even better this should be a function that returns op name

src/operator/quantization/quantize_graph_pass.cc
```diff
@@ -130,6 +131,7 @@ Graph QuantizeGraph(Graph &&src) {
   // graph. Key is the currently visited graph's node pointer, and value is a copied node of the key
   // node. The existing key's value may be updated with the newly created quantize/dequantize op.
   std::unordered_map<Node*, NodePtr> mirror_map;
+  nnvm::NodeEntryMap<NodeEntry> multiple_outputs_entry_map;
```
apeforest: I would prefer a generic way to handle both single and multiple outputs than treating them differently

src/operator/quantization/quantize_graph_pass.cc
```diff
@@ -215,12 +242,19 @@ Graph QuantizeGraph(Graph &&src) {
           min_index = num_outputs + 2 * e.index;
           max_index = num_outputs + 2 * e.index + 1;
         } else {
-          CHECK(mirror_node->op() != nullptr &&
-                mirror_node->op()->name == "_contrib_quantize")
+          CHECK((mirror_node->op() != nullptr &&
+                mirror_node->op()->name == "_contrib_quantize") ||
```
apeforest: Same here replace the op name by a function

src/operator/quantization/quantize_graph_pass.cc
ciyongch: Yes, will fix the typo :)

src/operator/quantization/quantize_graph_pass.cc
ciyongch: Just want to keep the same name as others in this file, like [#L326](https://github.com/apache/incubator-mxnet/blob/master/src/operator/quantization/quantize_graph_pass.cc#L326).
Anyway, this could be changed as you mentioned.

src/operator/quantization/quantize_graph_pass.cc
ciyongch: This is the convention naming in current MXNet, right?
For this case, the string name here is already interpretable as either a constant variable or a function.

src/operator/quantization/quantize_graph_pass.cc
ciyongch: Good point here.
This patch is trying to keep the mapping relation as before with minimal change. Actually it's able to handle both single and multiple outputs via their entry instead of current implementation, and thus no need to update the mirror_map's value to quantize/requantize op.
I will update the code to make it uniform for single and multiple outputs cases.

src/operator/quantization/quantize_graph_pass.cc
ciyongch: Same comments as above.

src/operator/quantization/quantize_graph_pass.cc
apeforest: Comparing to a plain string is error prone. What if someone mis-spelled by one char. It's going to be hard to debug.
In fact, I think a more rigorous way to check if an operator may be checking the operator itself: if (op ==  nnvm::Op::Get("_contrib_quantize"))

apeforest:
```suggestion
                    mirror_node->op() != nnvm::Op::Get("_contrib_quantize")
```

src/operator/quantization/quantize_graph_pass.cc
apeforest: See reply above

src/operator/quantization/quantize_graph_pass.cc
ciyongch: Yes, that will be better and more strict way compared with original one.
I'm thinking  about checking `entry` (like: `!mirror_entry_map.count(e)`) but not `mirror_node` here, since we stores the mapping between current entries and the entries outgoing from `_contrib_quantize` node via  `nnvm::NodeEntryMap<NodeEntry>` for both single output and multiple outputs. `mirror_node` can handle single output but not multiple outputs, while `entry` can handle both cases.
What do you think about this?

ciyongch: Thanks for the suggestion, using `entry` instead of `mirror_node` here could be a better choice.

ciyongch: Anyway, I will change all the others as you suggested.

apeforest: Checking `entry` as you suggested is fine. Checking op name with a string seems to be error-prone and requires many updates if we later change the op name

src/operator/quantization/quantize_graph_pass.cc
apeforest: If so, keeping it is fine with me as long as it is consistent.

src/operator/quantization/quantize_graph_pass.cc
apeforest: Using `entry` is fine with me as long as you are not only comparing the op name

pengzhao-intel:
@anirudh2290 could you help merge this PR?

ciyongch:
Thanks @apeforest @yuxihu for approving the changes. :)
Anyone could help to merge this PR?

TaoLv:
@ciyongch Could you rebase the code and re-trigger CI? I notice the last run of CI is about 6 days ago.

TaoLv:
@ciyongch Please take a look at the CI failure. Thanks.

ciyongch:
@TaoLv The failure was caused by unsupported quantized_concat in gpu, it's fixed now, please help to check and merge when it's done. :)

ciyongch:
@TaoLv Code is rebase to latest master, test case is updated, please help to take a review. :)

TaoLv:
@ciyongch Thanks for the contribution. @apeforest @yuxihu Thanks for the review and discussion. Now merging.

PR #13043, opened 2018-10-30
Fix test failure due to hybridize call in test_gluon_rnn.test_layer_fill_shape

yuxihu:
## Description ##
Previously, the value of bulk_size in struct BulkStatus of ThreadedEngine was not reset to its original one if there is exception thrown when executing StaticForward/DynamicForward for CachedOp. This causes another test (test_operator_gpu.test_incorrect_gpu) to fail if we add hybridize call in test_gluon_rnn.test_layer_fill_shape. We fix this by adding a try-catch block and reset the bulk_size in case of exception when executing CachedOp::StaticForward/DynamicForward. We also add the try-catch block in other similar places to prevent potential bugs.

### DIFF
--- a/src/imperative/cached_op.cc
+++ b/src/imperative/cached_op.cc
@@ -855,10 +855,15 @@ OpStatePtr CachedOp::Forward(
   int prev_bulk_size = Engine::Get()->set_bulk_size(config_.forward_bulk_size);
 
   OpStatePtr op_state;
-  if (config_.static_alloc) {
-    op_state = StaticForward(default_ctx, inputs, outputs);
-  } else {
-    op_state = DynamicForward(default_ctx, inputs, outputs);
+  try {
+    if (config_.static_alloc) {
+      op_state = StaticForward(default_ctx, inputs, outputs);
+    } else {
+      op_state = DynamicForward(default_ctx, inputs, outputs);
+    }
+  } catch (const dmlc::Error& e) {
+    Engine::Get()->set_bulk_size(prev_bulk_size);
+    throw e;
   }
 
   Engine::Get()->set_bulk_size(prev_bulk_size);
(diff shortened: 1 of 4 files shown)


ankkhedia:
@yuxihu Thanks for your contribution!
@mxnet-label-bot [pr-work-in-progress, Gluon]

kalyc:
Hi @yuxihu 
Thanks for your contribution
There appears to be build failure for your PR could you take a look and resolve them?

yuxihu:
@mxnet-label-bot [pr-awaiting-review, Backend, Gluon]

yuxihu:
@mxnet-label-bot update [pr-awaiting-review, Backend, Gluon]

src/imperative/imperative.cc
```diff
@@ -494,9 +494,14 @@ std::vector<NDArray*> Imperative::Backward(
   bool prev_training = set_is_training(is_train);
```
apeforest: Do we need reset this upon exception?

src/imperative/cached_op.cc
```diff
@@ -855,10 +855,15 @@ OpStatePtr CachedOp::Forward(
   int prev_bulk_size = Engine::Get()->set_bulk_size(config_.forward_bulk_size);
 
   OpStatePtr op_state;
-  if (config_.static_alloc) {
-    op_state = StaticForward(default_ctx, inputs, outputs);
-  } else {
-    op_state = DynamicForward(default_ctx, inputs, outputs);
+  try {
```
apeforest: This change looks good to me. One nit suggestion would be: is it cleaner by using a RAII class object to manage the bulk_size value instead of a local variable? Doing that we may also get rid of the try block. Not something necessary, but I would love to hear some comments on this.

src/imperative/imperative.cc
yuxihu: Although the lines are unrelated to the bug we are trying to fix here, it looks like we need to reset those as well.

src/imperative/cached_op.cc
yuxihu: Thanks for the suggestion. It can be done but try-catch also gives us more flexibility to do more things (like your comment above which we need to do more cleanup upon exception). Prefer to keep try-catch here.

PR #12992, opened 2018-10-26
[MXNET-1179] Enforce deterministic algorithms in convolution layers

apeforest:
## Description ##
Some of the CUDNN convolution algorithms are non-deterministic (see issue https://github.com/apache/incubator-mxnet/issues/11341). This PR adds an env variable to enforce determinism in the convolution operators. If set to true, only deterministic CUDNN algorithms will be used. If no deterministic algorithm is available, MXNet will error out.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] new environment variable added: MXNET_ENFORCE_DETERMINISM (default false)
- [x] CUDNN covolution, deconvolution, max pooling operators
## Comments ##
- still need to add a test to verify determinism.

PR #13049, opened 2018-10-30
Document the newly added env variable

apeforest:
## Description ##
Document the env variable: MXNET_ENFORCE_DETERMINISM added in PR: https://github.com/apache/incubator-mxnet/pull/12992


ankkhedia:
@anirudh2290 @sandeep-krishnamurthy Could this PR be merged if this looks good?

apeforest:
@samskalicky I am only aware of these cudnn algorithms that may cause non-terminism. However, I plan to have this env variable to cover all MXNet. In case there are other non-deterministic algorithms used, we can then use this env variable to screen them out.

PR #13066, opened 2018-10-31
Remove unused header from engine.h

apeforest:
## Description ##
There are very complicated dependencies in the current mxnet include folder. This is an initial step to reduce unnecessary dependency.


ankkhedia:
@apeforest Thanks for taking up the task simplify dependencies.

@mxnet-label-bot [pr-awaiting-review, backend]

PR #13094, opened 2018-11-02
Simplifications and some fun stuff for the MNIST Gluon tutorial

kohr-h:
## Description ##

This PR removes some of the (IMO) unnecessary complexity from the tutorial (one of the first that new users click on) by *always* using 1 GPU or 1 CPU. This makes the MLP training code much more straightforward (no splitting, no inner loop).
I also think that there was too much staccato of obvious comments in that part, while the truly non-obvious parts were left uncommented (e.g., what is `ag.record()`?). I tried to straighten up that part a bit, although I think some comments are still a bit on the obvious side.

Furthermore, I added a small section where some of the mislabeled (apparently it's one "l", need to fix that) images are plotted together with the predicted and true labels. I think that's more fun than just looking at an accuracy score.

I'm not finished with the whole tutorial yet, so it's WIP, but I'd like to know what you think before I put more work into this PR.

**Note:** I've included the 2 new images in the same directory as the tutorial for convenience, they would obviously need to go somewhere else.

### Essentials ###


ankkhedia:
@kohr-h Thanks for your contribiution!

@mxnet-label-bot [pr-work-in-progress, Gluon]

docs/tutorials/gluon/mnist.md
```diff
@@ -126,71 +125,116 @@ training scope which is defined by `autograd.record()`.
...
-        # Splits train data into multiple slices along batch_axis
-        # and copy each slice into a context.
-        data = gluon.utils.split_and_load(batch.data[0], ctx_list=ctx, batch_axis=0)
-        # Splits train labels into multiple slices along batch_axis
-        # and copy each slice into a context.
-        label = gluon.utils.split_and_load(batch.label[0], ctx_list=ctx, batch_axis=0)
-        outputs = []
-        # Inside training scope
+        # Possibly copy data and labels to the GPU
+        data = batch.data[0].copyto(ctx)
```
ThomasDelteil: use `as_in_context` instead of `copyto` please

docs/tutorials/gluon/mnist.md
```diff
@@ -126,71 +125,116 @@ training scope which is defined by `autograd.record()`.
...
-    # and copy each slice into a context.
-    label = gluon.utils.split_and_load(batch.label[0], ctx_list=ctx, batch_axis=0)
-    outputs = []
-    for x in data:
-        outputs.append(net(x))
-    # Updates internal evaluation
-    metric.update(label, outputs)
-print('validation acc: %s=%f'%metric.get())
+    # Possibly copy data and labels to the GPU
+    data = batch.data[0].copyto(ctx)
```
ThomasDelteil: please use `as_in_context`

docs/tutorials/gluon/mnist.md
```diff
@@ -126,71 +125,116 @@ training scope which is defined by `autograd.record()`.
...
 
+That said, a single numer only gives us very limited information on the performance of our neural network. It is always a good idea to actually look at the images on which the network performed poorly, and check for clues on how to improve the performance. We do that with the help of a small function that produces a list of the images where the network got it wrong, together with the predicted and true labels:
+
+```python
+def get_mislabelled(it):
+    """Return list of ``(input, pred_lbl, true_lbl)`` for mislabelled samples."""
+    mislabelled = []
+    it.reset()
+    for batch in it:
+        data = batch.data[0].copyto(ctx)
```
ThomasDelteil: please use `as_in_context` rather than `copyto`

docs/tutorials/gluon/mnist.md
```diff
@@ -126,71 +125,116 @@ training scope which is defined by `autograd.record()`.
...
+fig, axs = plt.subplots(ncols=sample_size)
+for ax, (img, pred, lbl) in zip(axs, wrong_val_sample):
+    fig.set_size_inches(18, 4)
+    fig.suptitle("Sample of wrong predictions in the validation set", fontsize=20)
+    ax.imshow(img[0], cmap="gray")
+    ax.set_title("Predicted: {}\nActual: {}".format(pred, lbl))
+    ax.xaxis.set_visible(False)
+    ax.yaxis.set_visible(False)
+```
+![png](./wrong_train.png)
```
ThomasDelteil: please host the images on https://github.com/dmlc/web-data in the corresponding path

kohr-h: Does that require a separate PR there?

kohr-h:
> Thanks for the much needed update @kohr-h, can I bother you to update the data loading to be pure Gluon rather than using the old DataIter API ? That means no need to reset the iterator between epoch and the data comes out of the iterator as a tuple `(data, label)` rather than a `batch.data` and `batch.label`

No problem. I've started doing that, more to come.

docs/tutorials/gluon/mnist.md
ThomasDelteil: Yes indeed, please tag me in and I'll take care of merging it. Try to follow the same path, `mxnet/doc/tutorials/gluon/mnist/your_image.png`

kohr-h: Okay, will do.

kalyc:
Thanks for your contribution @kohr-h 
Could you please update the PR as per discussions above?

kohr-h:
Thanks for the reminder @kalyc. I got too many balls in the air. Will get this done asap.

kohr-h:
I'm done with the rewording and restructuring of the example.

Note that I also made the CNN part more basic by moving from a custom `nn.Block` subclass to a `nn.Sequential` model. I don't think it added any value, and there are other tutorials that explain this in more detail.

Ready for review.

docs/tutorials/gluon/mnist.md
kohr-h: Done, see https://github.com/dmlc/web-data/pull/142

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,434 @@
...
+Since the MNIST dataset is relatively small, the `MNIST` class loads it into memory all at once, but for larger datasets like ImageNet, this would no longer be possible.
+The Gluon `Dataset` class from which `MNIST` derives supports both cases.
+In general, `Dataset` and `DataLoader` (which we will encounter next) are the machinery in MXNet that provides a stream of input data to be consumed by a training algorithm, typically in batches of multiple data entities at once for better efficiency.
+In this tutorial, we will configure the data loader to feed examples in batches of 100.
+
+An image batch is commonly represented as a 4-D array with shape `(batch_size, num_channels, height, width)`.
+This convention is denoted by "BCHW", and it is the default in MXNet.
```
ThomasDelteil: can you replace with `NCHW` ? Thanks!

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,434 @@
...
+For the MNIST dataset, each image has a size of 28x28 pixels and one color channel (grayscale), hence the shape of an input batch will be `(batch_size, 1, 28, 28)`.
 
-Image batches are commonly represented by a 4-D array with shape `(batch_size, num_channels, width, height)`. For the MNIST dataset, since the images are grayscale, there is only one color channel. Also, the images are 28x28 pixels, and so each image has width and height equal to 28. Therefore, the shape of input is `(batch_size, 1, 28, 28)`. Another important consideration is the order of input samples. When feeding training examples, it is critical that we don't feed samples with the same label in succession. Doing so can slow down training.
-Data iterators take care of this by randomly shuffling the inputs. Note that we only need to shuffle the training data. The order does not matter for test data.
+Another important consideration is the order of input samples.
+When feeding training examples, it is critical not feed samples with the same label in succession since doing so can slow down training.
```
ThomasDelteil: I would rather state it as: slow down the convergence of your model*
to make sure that users don't read it as actual slow down in throughput.

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,434 @@
...
+Data iterators take care of this issue by randomly shuffling the inputs.
```
ThomasDelteil: I would add a "(`DataLoader`)" after "Data iterators"

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,434 @@
...
+from __future__ import print_function  # only relevant for Python 2
 import mxnet as mx
-from mxnet import gluon
+from mxnet import nd, gluon, autograd
 from mxnet.gluon import nn
-from mxnet import autograd as ag
 ```
 
-### Define a network: Multilayer Perceptron
+### Defining a network: Multilayer Perceptron
```
ThomasDelteil: nit: "Multilayer Perceptron (MLP)"

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,434 @@
...
+The number *k*, also referred to as *hidden size*, corresponds to the number of neurons in the FC layer.
+An FC layer has two learnable parameters: the (*m x k*) weight matrix *W* and the (*1 x k*) bias vector *b*.
 
-MLPs consist of several fully connected layers. A fully connected layer or FC layer for short, is one where each neuron in the layer is connected to every neuron in its preceding layer. From a linear algebra perspective, an FC layer applies an [affine transform](https://en.wikipedia.org/wiki/Affine_transformation) to the *n x m* input matrix *X* and outputs a matrix *Y* of size *n x k*, where *k* is the number of neurons in the FC layer. *k* is also referred to as the hidden size. The output *Y* is computed according to the equation *Y = W X + b*. The FC layer has two learnable parameters, the *m x k* weight matrix *W* and the *m x 1* bias vector *b*.
+In an MLP, the outputs of FC layers are typically fed into an activation function that applies an elementwise nonlinearity.
```
ThomasDelteil: nit: In a MLP

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,434 @@
...
-The last fully connected layer often has its hidden size equal to the number of output classes in the dataset. Furthermore, these FC layers uses ReLU activation for performing an element-wise ReLU transformation on the FC layer output.
-
-To do this, we will use [Sequential layer](http://mxnet.io/api/python/gluon/gluon.html#mxnet.gluon.nn.Sequential) type. This is simply a linear stack of neural network layers. `nn.Dense` layers are nothing but the fully connected layers we discussed above.
+**Note**: using the `name_scope()` context manager is optional.
+It is, however, good practice since it uses a common prefix for the names of all layers generated in that scope, which can be very helpful during debugging.
 
 ```python
-# define network
-net = nn.Sequential()
+net = nn.Sequential('MLP')
```
ThomasDelteil: I would err on the side of replacing it with `nn.HybridSequential`, in order to get people used to use Hybrid blocks. non-hybrid blocks should only be used when they contained non-hybridizable blocks IMO.

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,434 @@
...
+Typically, the training is run until convergence, which means that further iterations will no longer lead to improvements of the loss function, and that the network has probably learned a good set of model parameters from the train data.
+For the purpose of this tutorial, we only loop 10 times over the entire dataset.
+One such pass over the data is usually called an *epoch*.
+
+The following steps are taken in each `epoch`:
+
+- Get a minibatch of `inputs` and `labels` from the `train_loader`.
+- Feed the `inputs` to the network, producing `outputs`.
+- Compute the minibatch `loss` value by comparing `outputs` to `labels`.
+- Backpropagate the gradients to update the network parameters by calling `loss.backward()`.
```
ThomasDelteil: This actually does not update the network parameters, a better wording would be:
- Use backpropagation to compute the gradients of the loss for each of the network parameters by calling `loss.backward()`
- Update the parameters of the network according to the optimizer rule with `trainer.step(batch_size=inputs.shape[0])`

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,434 @@
...
+num_epochs = 10
+
+for epoch in range(num_epochs):
+    for inputs, labels in train_loader:
+        # Possibly copy inputs and labels to the GPU
+        inputs = inputs.as_in_context(ctx)
+        labels = labels.as_in_context(ctx)
+
+        # The forward pass and the loss computation need to be wrapped
+        # in a `record()` scope to indicate that the results will
```
ThomasDelteil: The forward pass and the loss computation need to be wrapped in a `record()` scope to make sure the computational graph is recorded in order to automatically compute the gradients during the backward pass.

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,434 @@
...
-**Figure 3:** First conv + pooling layer in LeNet.
+The size of the neighborhood that a convolution layer considers for each pixel is usually referred to as *filter size* or *kernel size*.
+The array of weights -- which does not depend on the output pixel location, only on the position within such a neighborhood -- is called *filter* or *kernel*.
+Typical filter sizes range from *3 x 3* to *13 x 13*, which implies that a convolution layer has *far* fewer parameters than a dense layer.
 
-Now we train LeNet with similar hyper-parameters as before. Note that, if a GPU is available, we recommend using it. This greatly speeds up computation given that LeNet is more complex and compute-intensive than the previous multilayer perceptron. To do so, we only need to change `mx.cpu()` to `mx.gpu()` and MXNet takes care of the rest. Just like before, we'll stop training after 10 epochs.
+```python
+conv_layer = nn.Conv2D(kernel_size=(3, 3), channels=32, in_channels=16, activation='relu')
+print(conv_layer.params)
+# Output:
```
ThomasDelteil: Please use the following notation, and remove the commented lines:


     ```python
     conv_layer = nn.Conv2D(kernel_size=(3, 3), channels=32, in_channels=16, activation='relu')
     print(conv_layer.params)
     ```

     `Parameter conv0_weight (shape=(32, 16, 3, 3), dtype=<class 'numpy.float32'>)` 

     `Parameter conv0_bias (shape=(32,), dtype=<class 'numpy.float32'>)`

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,434 @@
...
+fig, axs = plt.subplots(ncols=sample_size)
+for ax, (img, pred, lbl) in zip(axs, wrong_val_sample):
+    fig.set_size_inches(18, 4)
+    fig.suptitle("Sample of wrong predictions in the validation set", fontsize=20)
+    ax.imshow(img[0], cmap="gray")
+    ax.set_title("Predicted: {}\nActual: {}".format(pred, lbl))
+    ax.xaxis.set_visible(False)
+    ax.yaxis.set_visible(False)
 ```
+![png](./wrong_train.png)
```
ThomasDelteil: Please add `<!--notebook-skip-line-->` after the image so that it does not appear in the downloaded notebook that is automatically generated.

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,434 @@
...
+The following code defines a CNN architecture called *LeNet*.
+The LeNet architecture is a popular network known to work well on digit classification tasks.
+We will use a version that differs slightly from the original in the usage of `tanh` activations instead of `sigmoid`.
 
 ```python
-# set the context on GPU is available otherwise CPU
-ctx = [mx.gpu() if mx.test_utils.list_gpus() else mx.cpu()]
-net.initialize(mx.init.Xavier(magnitude=2.24), ctx=ctx)
-trainer = gluon.Trainer(net.collect_params(), 'sgd', {'learning_rate': 0.03})
+lenet = nn.Sequential('LeNet')
```
ThomasDelteil: Same as above, I would err on the side of using a `HybridSequential` here, so that if someone wants to take this example and play around hybridization and export, they can do it transparently.

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,434 @@
...
-        # batch size of data to normalize the gradient by 1/batch_size.
-        trainer.step(batch.data[0].shape[0])
-    # Gets the evaluation result.
-    name, acc = metric.get()
-    # Reset evaluation result to initial state.
-    metric.reset()
-    print('training acc at epoch %d: %s=%f'%(i, name, acc))
+lenet.initialize(mx.init.Xavier(), ctx=ctx)
+lenet.summary(nd.zeros((1, 1, 28, 28), ctx=ctx))
+# Output:
```
ThomasDelteil: I guess here the simplest is to simply put all the summary in a
` ``` ` block, rather than commented lines. Unfortunately `<!--notebook-skip-line-->` only works for single line rather than blocks 😭 so there are no ideal solution

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,434 @@
...
 
-If all went well, we should see a higher accuracy metric for predictions made using LeNet. With CNN we should be able to correctly predict around 98% of all test images.
+If all went well, we should see a higher accuracy metric for predictions made using LeNet.
+With this CNN we should be able to correctly predict around 98% of all test images.
 
 ## Summary
 
-In this tutorial, we have learned how to use MXNet to solve a standard computer vision problem: classifying images of hand written digits. You have seen how to quickly and easily build, train and evaluate models such as MLP and CNN with MXNet Gluon package.
+In this tutorial, we demonstrated how to use MXNet to solve a standard computer vision problem: classifying images of hand-written digits.
+We showed how to quickly build, train and evaluate models such as MLPs and CNNs with the MXNet Gluon package.
```
ThomasDelteil: Please delete the two images that are still part of this PR, thanks!

docs/tutorials/gluon/mnist.md
kohr-h: Totally right.

docs/tutorials/gluon/mnist.md
kohr-h: Okay

docs/tutorials/gluon/mnist.md
kohr-h: Hm, I thought it kind of depends. If you say "Multilayer Perceptron" in your mind, then it starts with a spoken consonant, and the article should be "a". However, if you say "MLP", the "M" starts with a vowel, and the article should be "an".

Is there a clear rule for cases like this one?

docs/tutorials/gluon/mnist.md
kohr-h: Yes, that's better.

docs/tutorials/gluon/mnist.md
kohr-h: Oops, maybe I should have consulted the docs before :grin:

docs/tutorials/gluon/mnist.md
kohr-h: Asked by a non-native speaker as well :smile: 
Maybe a native speaker can jump in and sort this out? Otherwise I'll stick with the current version.

kohr-h:
I think I addressed all the comments from the review.

vandanavk:
@sandeep-krishnamurthy @stu1130 for review

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,443 @@
...
+net = nn.HybridSequential('MLP')
 with net.name_scope():
-    net.add(nn.Dense(128, activation='relu'))
-    net.add(nn.Dense(64, activation='relu'))
-    net.add(nn.Dense(10))
+    net.add(
+        nn.Flatten(),
+        nn.Dense(128, activation='relu'),
+        nn.Dense(64, activation='relu'),
+        nn.Dense(10, activation='relu')
```
sandeep-krishnamurthy: Why relu activation in the last layer?

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,443 @@
...
-trainer = gluon.Trainer(net.collect_params(), 'sgd', {'learning_rate': 0.03})
+lenet = nn.HybridSequential('LeNet')
+lenet.add(
+    nn.Conv2D(channels=20, kernel_size=(5, 5), activation='tanh'),
+    nn.MaxPool2D(pool_size=(2, 2), strides=(2, 2)),
+    nn.Conv2D(channels=50, kernel_size=(5, 5), activation='tanh'),
+    nn.MaxPool2D(pool_size=(2, 2), strides=(2, 2)),
+    nn.Flatten(),
+    nn.Dense(500, activation='tanh'),
+    nn.Dense(10, activation='tanh'),
```
sandeep-krishnamurthy: why tanh activation in last layer?

kohr-h:
> 1 question - Why activation function in last layer?

No real reason. I *think* I didn't change the original logic, but I can check again. If a different last activation is more reasonable for this tutorial, I can put it in, no problem.

kohr-h:
@sandeep-krishnamurthy I switched both nets to sigmoid activation. Note though that they perform worse with the same settings. I've tested a few learning rates, and the current ones improve things a bit. But maybe it's okay for the scope of the tutorial that the training isn't perfectly fine-tuned.

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,443 @@
...
+net = nn.HybridSequential(prefix='MLP_')
 with net.name_scope():
-    net.add(nn.Dense(128, activation='relu'))
-    net.add(nn.Dense(64, activation='relu'))
-    net.add(nn.Dense(10))
+    net.add(
+        nn.Flatten(),
+        nn.Dense(128, activation='relu'),
+        nn.Dense(64, activation='relu'),
+        nn.Dense(10, activation='sigmoid')
```
ThomasDelteil: There shouldn't be an activation here, soft max is taken care of by the loss

docs/tutorials/gluon/mnist.md
```diff
@@ -1,333 +1,443 @@
...
+lenet = nn.HybridSequential(prefix='LeNet_')
+with lenet.name_scope():
+    lenet.add(
+        nn.Conv2D(channels=20, kernel_size=(5, 5), activation='tanh'),
+        nn.MaxPool2D(pool_size=(2, 2), strides=(2, 2)),
+        nn.Conv2D(channels=50, kernel_size=(5, 5), activation='tanh'),
+        nn.MaxPool2D(pool_size=(2, 2), strides=(2, 2)),
+        nn.Flatten(),
+        nn.Dense(500, activation='tanh'),
+        nn.Dense(10, activation='sigmoid'),
```
ThomasDelteil: Same here there shouldn't be an activation, soft max is taken care of by the loss and need to be done manually at inference time

docs/tutorials/gluon/mnist.md
kohr-h: :man_facepalming: of course, that explains :sweat_smile: 
Update on the way.

kohr-h:
Okay, now we're in the right ballpark with the results. Thanks @ThomasDelteil for the catch.

PR #13095, opened 2018-11-02
[MXNET-1041] Add Java benchmark

lanking520:
## Description ##
Add Java benchmark for Object Detection:
```
bash run_java_inference_bm.sh --model-name ObjectDetection --num-runs 5 --model-path-prefix ../infer/models/resnet50_ssd/resnet50_ssd_model --input-image ../infer/images/dog.jpg --batchsize 5
```
result
```
[13:59:18] src/nnvm/legacy_json_util.cc:209: Loading symbol saved by previous version v0.11.0. Attempting to upgrade...
[13:59:18] src/nnvm/legacy_json_util.cc:217: Symbol successfully upgraded!

ObjectDetectionBatch_p99 4262.494227ms
ObjectDetectionBatch_p90 4262.494227ms
ObjectDetectionBatch_p50 3781.067256ms
ObjectDetectionBatch_average 3858.62ms
[13:59:38] src/nnvm/legacy_json_util.cc:209: Loading symbol saved by previous version v0.11.0. Attempting to upgrade...
[13:59:38] src/nnvm/legacy_json_util.cc:217: Symbol successfully upgraded!

ObjectDetection_p99 1020.063161ms
ObjectDetection_p90 1020.063161ms
ObjectDetection_p50 822.827342ms
ObjectDetection_average 861.85ms
```
@andrewfayres @nswamy @yzhliu @piyushghai @zachgk 
### Essentials ###


ankkhedia:
@lanking520 Thanks for your contribution
Could you please take a look into CI failures.


@mxnet-label-bot [pr-awaiting-review , Java]

scala-package/examples/src/main/java/org/apache/mxnetexamples/javaapi/benchmark/JavaBenchmark.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    public static void main(String[] args) {
+        if (args.length < 2) {
+            System.out.println("Please specify model name");
+            return;
+        }
+        String modelName = args[1];
+        InferBase model;
+        if (modelName.equals("ObjectDetection")) {
```
piyushghai: You can also use switch case in Java. In future if this file will be expanded to accommodate more benchmarking scripts, it'll be easier to read.

scala-package/examples/src/main/java/org/apache/mxnetexamples/javaapi/benchmark/JavaBenchmark.java
```diff
@@ -0,0 +1,114 @@
...
+            runBatch = true;
+            ObjectDetectionBenchmark inst = new ObjectDetectionBenchmark();
+            parse(inst, args);
+            model = inst;
+        } else {
+            System.out.println("Model name not found! " + modelName);
+            return;
+        }
+        List<Context> context = new ArrayList<Context>();
```
piyushghai: This thing can be abstracted out in a separate function, and in the main method you can just call that function : getContext()

scala-package/examples/src/main/java/org/apache/mxnetexamples/javaapi/benchmark/JavaBenchmark.java
```diff
@@ -0,0 +1,114 @@
...
+        if (runBatch) {
+            for (int i =0;i < model.numRun; i++) {
+                long currTime = System.nanoTime();
+                model.runBatchInference();
+                result[i] = System.nanoTime() - currTime;
+            }
+            printStatistics(result, modelName +"Batch");
+        }
+
+        ((ObjectDetectionBenchmark) model).batchSize = 1;
```
piyushghai: I don't see this variable being used in the runSingleInference code path. Why are you setting it here?

Also, this could be a potential ClassCastException in the future.

piyushghai: If you want to set class members from outside the class, use the Setter pattern. 
Do not access the members with the dot notation.

scala-package/examples/src/main/java/org/apache/mxnetexamples/javaapi/infer/objectdetector/SSDClassifierExample.java
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-package org.apache.mxnetexamples.infer.javapi.objectdetector;
+package org.apache.mxnetexamples.javaapi.infer.objectdetector;
```
andrewfayres: Why are we making this change? Curious if you just think that this structure is better or if something prompted it.

scala-package/infer/src/main/scala/org/apache/mxnet/infer/javaapi/ObjectDetector.scala
```diff
@@ -19,6 +19,8 @@ package org.apache.mxnet.infer.javaapi
 
 // scalastyle:off
 import java.awt.image.BufferedImage
+
+import org.apache.mxnet.javaapi.Shape
```
andrewfayres: Can be added to line 26

scala-package/examples/src/main/java/org/apache/mxnetexamples/javaapi/infer/objectdetector/SSDClassifierExample.java
lanking520: You are right. In this way users can shop the collections of Java APIs instead of doing:
`mxnetexamples.benchmark.javaapi.*`
and 
`mxnetexmaples.infer.javaapi.*`

What do you think?

lanking520: Same apllies to `org.apache.mxnet.javaapi.infer.*`, so a single `.javaapi` would bring all from core, infer and examples

scala-package/examples/src/main/java/org/apache/mxnetexamples/javaapi/benchmark/JavaBenchmark.java
```diff
@@ -0,0 +1,122 @@
...
+
+        long[] result = new long[model.numRun];
+        model.preProcessModel(context);
+        if (runBatch) {
+            for (int i =0;i < model.numRun; i++) {
+                long currTime = System.nanoTime();
+                model.runBatchInference();
+                result[i] = System.nanoTime() - currTime;
+            }
+            printStatistics(result, modelName +"Batch");
```
piyushghai: To be consistent with the Scala Benchmark runs, can you rename the second parameter to : "batch_inference" for batch, and "single_inference" for single inference calls? 
https://github.com/apache/incubator-mxnet/blob/master/scala-package/examples/src/main/scala/org/apache/mxnetexamples/benchmark/ScalaInferenceBenchmark.scala#L150

scala-package/examples/src/main/java/org/apache/mxnetexamples/javaapi/infer/objectdetector/SSDClassifierExample.java
lanking520: Agree on this change after offline disucssion

scala-package/examples/src/main/java/org/apache/mxnetexamples/javaapi/benchmark/JavaBenchmark.java
```diff
@@ -106,7 +106,7 @@ public static void main(String[] args) {
                 model.runBatchInference();
                 result[i] = System.nanoTime() - currTime;
             }
-            printStatistics(result, modelName +"Batch");
+            printStatistics(result, modelName +"batch_inference");
```
piyushghai: You can remove the modelName from the second parameter as well here. 
The Model Name will be aptly captured when reporting the results of the benchmark run as we'd know which script we are running it with :)

scala-package/infer/src/main/scala/org/apache/mxnet/infer/javaapi/ObjectDetector.scala
```diff
@@ -113,6 +113,14 @@ object ObjectDetector {
     org.apache.mxnet.infer.ImageClassifier.loadImageFromFile(inputImagePath)
   }
 
+  def reshapeImage(img : BufferedImage, newWidth: Int, newHeight: Int): BufferedImage = {
+    org.apache.mxnet.infer.ImageClassifier.reshapeImage(img, newWidth, newHeight)
```
andrewfayres: So I'm not sure whether or not we should add these to the public API for ObjectDetector. I made this mistake with loadImageFromFile earlier apparently because we don't yet have ImageClassifier in the Java API.

In the Scala version, the objectDetector benchmark and example call loadImageFromFile, reshapeImage, and bufferedImageToPixels directly from the ImageClassifier class. We're not able to do that from here yet but even if we could it seems odd to do so. Maybe we should move them to a utils class or something? What's everyone else think about how we should handle this?

lanking520: @andrewfayres I am expecting this: https://github.com/apache/incubator-mxnet/blob/master/scala-package/core/src/main/scala/org/apache/mxnet/Image.scala would eventually help and replace the current JavaIO we have.

nswamy:
consider adding RAM and GPU memory usage. 
Also your output should give more information on how many runs did it run, batch size, ..
Please drop the head and tail end of the runs, for example if you run 10 epochs you should drop 1 or 2 at the start and 1 or 2 epochs at the end.

lanking520:
> consider adding RAM and GPU memory usage.
> Also your output should give more information on how many runs did it run, batch size, ..
> Please drop the head and tail end of the runs, for example if you run 10 epochs you should drop 1 or 2 at the start and 1 or 2 epochs at the end.

@nswamy Add JIRA here: https://issues.apache.org/jira/browse/MXNET-1208 for memory usage monitoring

scala-package/examples/src/main/java/org/apache/mxnetexamples/javaapi/benchmark/JavaBenchmark.java
```diff
@@ -0,0 +1,135 @@
...
+        Arrays.sort(seq);
+        int k = (int) Math.ceil((seq.length - 1) * (p / 100.0));
+        return seq[k];
+    }
+
+    private static void printStatistics(long[] inferenceTimesRaw, String metricsPrefix)  {
+        long[] inferenceTimes = inferenceTimesRaw;
+        // remove head and tail
+        if (inferenceTimes.length > 2) {
+            inferenceTimes = Arrays.copyOfRange(inferenceTimesRaw,
```
piyushghai: A good logic to drop head and tail would be to say, check for a minimum length of the array ( let's say 10 or 20 or 50) and then drop the first 10% for head and last 10% for tail.

Hardcoding it to remove 1 item for head and 1 for tail might not prove to be statistically significant for a large array size.

lanking520: The reason to remove the first and last is due to the fact that: first GPU inference would introduce Pre-memory allocation and that would make the only first inference takes 10x times than the other. The last one also contains a similar problem with CPU.

nswamy: I agree with @piyushghai, i think it would make sense to drop max(x%, 1), wdyt ?

lanking520: > I agree with @piyushghai, i think it would make sense to drop max(x%, 1), wdyt ?
@piyushghai @nswamy Bare with me, any technical reason or proof of concept for doing this? Why dropping the first 10% and the last 10% would bring a better result? Why not 20%?(under the assumption that the time cost is a normal distribution, it would be better to drop the smallest 5% and the largest 5% so we fit 2 sigma)

nswamy: I don't have a statistical reason for x%, x depends on what you are measuring. I don't know if 10% is the right answer.
What i have a seen is that there is a warm up time(for example loading cache, etc.) and cool down period( evicting cache as an example) during the start and end during performance runs, so you tend to drop a certain percentage of it.

andrewfayres: I've witnessed the warmup period during training, I don't doubt there is some effect during inference but I'd be surprised if it's very prominent. If there is one, there's no reason that comes to mind as to why it would depend upon the number of inferences you're performing. Sure it could depend on the example pool size but given the same number of available examples, the warmup for 100 inferences should be the same as 100k. Removing max(x%) in both of these cases would remove the warmup in the 100 case and remove warmup + outliers in the 100k.

I don't see a reason to add rules in an attempt to account for things that we don't really understand. I would vote that we measure every run. We've got a P90, if the assumption about a warmup/cool down is true then it'll be in the 10% that gets trimmed there.

PR #13105, opened 2018-11-04
[MXNET-1158] JVM Memory Management Documentation

nswamy:
## Description ##
Add Documentation for JVM Memory management explaining the various options and its usage.

@andrewfayres @lanking520 @piyushghai @yzhliu

### DIFF
--- a/scala-package/examples/scripts/run_train_mnist.sh
+++ b/scala-package/examples/scripts/run_train_mnist.sh
@@ -35,7 +35,7 @@ fi
 
 MXNET_ROOT=$(cd "$(dirname $0)/../../.."; pwd)
 echo $MXNET_ROOT
-CLASS_PATH=$MXNET_ROOT/scala-package/assembly/$platform-$hw_type/target/*:$MXNET_ROOT/scala-package/examples/target/*:$MXNET_ROOT/scala-package/examples/target/classes/lib/*:$MXNET_ROOT/scala-package/infer/target/*
+CLASS_PATH=$MXNET_ROOT/scala-package/assembly/$platform-$hw_type/target/*:$MXNET_ROOT/scala-package/examples/target/*:$MXNET_ROOT/scala-package/examples/target/classes/lib/*
 
 # model dir
 DATA_PATH=$2
--- a/scala-package/examples/scripts/run_train_mnist.sh
+++ b/scala-package/examples/scripts/run_train_mnist.sh
@@ -19,15 +19,31 @@
 
 set -e
 
+hw_type=cpu
+if [[ $1 = gpu ]]
+then
+    hw_type=gpu
+fi
+
+platform=linux-x86_64
+
+if [[ $OSTYPE = [darwin]* ]]
+then
+    platform=osx-x86_64
+    hw_type=cpu
+fi
+
 MXNET_ROOT=$(cd "$(dirname $0)/../../.."; pwd)
 echo $MXNET_ROOT
-CLASS_PATH=$MXNET_ROOT/scala-package/assembly/linux-x86_64-cpu/target/*:$MXNET_ROOT/scala-package/examples/target/*:$MXNET_ROOT/scala-package/examples/target/classes/lib/*:$MXNET_ROOT/scala-package/infer/target/*
+CLASS_PATH=$MXNET_ROOT/scala-package/assembly/$platform-$hw_type/target/*:$MXNET_ROOT/scala-package/examples/target/*:$MXNET_ROOT/scala-package/examples/target/classes/lib/*:$MXNET_ROOT/scala-package/infer/target/*
 
 # model dir
 DATA_PATH=$2
 
-java -XX:+PrintGC -Xms256M -Xmx512M -Dmxnet.traceLeakedObjects=false -cp $CLASS_PATH \
-        org.apache.mxnetexamples.imclassification.TrainMnist \
-        --data-dir /home/ubuntu/mxnet_scala/scala-package/examples/mnist/ \
+java -XX:+PrintGC -Dmxnet.traceLeakedObjects=false -cp $CLASS_PATH \
+        org.apache.mxnetexamples.imclassification.TrainModel \
+        --data-dir $MXNET_ROOT/scala-package/examples/mnist/ \
+        --network mlp \
+        --num-layers 50 \
         --num-epochs 10000000 \
         --batch-size 1024
--- a/scala-package/memory-management.md
+++ b/scala-package/memory-management.md
@@ -1,13 +1,13 @@
 # JVM Memory Management
-The Scala and Java binding of Apache MXNet uses native memory(C++ Heap either in RAM or GPU memory) in most of the MXNet Scala objects such as NDArray, Symbol, Executor, KVStore, Data Iterators, etc.,. the Scala classes associated with them act as wrappers, 
-the operations on these objects are directed to the MXNet C++ backend via JNI for performance , so the bytes are also stored in the native heap for fast access.   
+The Scala and Java bindings of Apache MXNet uses native memory(C++ heap either in RAM or GPU memory) for most of the MXNet Scala objects such as NDArray, Symbol, Executor, KVStore, Data Iterators, etc.,. 
+The Scala classes associated with them act as wrappers, the operations on these objects are directed to the MXNet C++ backend via JNI for performance , so the bytes are also stored in the native heap for fast access.
 
-The JVM using the Garbage Collector only manages objects allocated in the JVM Heap and is not aware of the memory footprint of these objects in the native memory, hence allocation/deAllocation of the native memory has to be managed by MXNet Scala.  
-Allocating native memory is straight forward and is done during the construction of the object by a calling the associated C++ API through JNI, however since JVM languages do not have destructors, De-Allocation of these objects becomes problematic and has to explicitly de-allocated. 
-To make it easy, MXNet Scala provides a few modes of operation.
+The JVM using the Garbage Collector only manages objects allocated in the JVM Heap and is not aware of the memory footprint of these objects in the native memory, hence allocation/deallocation of the native memory has to be managed by MXNet Scala.
+Allocating native memory is straight forward and is done during the construction of the object by a calling the associated C++ API through JNI, However since JVM languages do not have destructors, deallocation of these objects needs to be done explicitly.
+To make it easy, MXNet Scala provides a few modes of operation, explained in detail below.
 
 ## Memory Management in Scala 
-### [ResourceScope.using](https://github.com/apache/incubator-mxnet/blob/master/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala#L106) (Recommended)
+### 1.  [ResourceScope.using](https://github.com/apache/incubator-mxnet/blob/master/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala#L106) (Recommended)
 `ResourceScope.using` provides the familiar Java try-with-resources primitive in Scala and also extends to automatically manage the memory of all the MXNet objects created in the code block (`body`) associated with it by tracking the allocations in a stack. 
 If an MXNet object or an Iterable containing MXNet objects is returned from the code-block, it is automatically excluded from de-allocation in the current scope and moved to 
 an outer scope if ResourceScope's are stacked.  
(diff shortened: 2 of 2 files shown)


scala-package/examples/scripts/run_train_mnist.sh
```diff
@@ -19,15 +19,31 @@
...
+if [[ $OSTYPE = [darwin]* ]]
+then
+    platform=osx-x86_64
+    hw_type=cpu
+fi
+
 MXNET_ROOT=$(cd "$(dirname $0)/../../.."; pwd)
 echo $MXNET_ROOT
-CLASS_PATH=$MXNET_ROOT/scala-package/assembly/linux-x86_64-cpu/target/*:$MXNET_ROOT/scala-package/examples/target/*:$MXNET_ROOT/scala-package/examples/target/classes/lib/*:$MXNET_ROOT/scala-package/infer/target/*
+CLASS_PATH=$MXNET_ROOT/scala-package/assembly/$platform-$hw_type/target/*:$MXNET_ROOT/scala-package/examples/target/*:$MXNET_ROOT/scala-package/examples/target/classes/lib/*:$MXNET_ROOT/scala-package/infer/target/*
```
lanking520: Could you also try to remove import infer CLASSPATH as well since we applied them in the assembly package as well?

scala-package/examples/scripts/run_train_mnist.sh
```diff
@@ -19,15 +19,31 @@
...
 
 # model dir
 DATA_PATH=$2
 
-java -XX:+PrintGC -Xms256M -Xmx512M -Dmxnet.traceLeakedObjects=false -cp $CLASS_PATH \
-        org.apache.mxnetexamples.imclassification.TrainMnist \
-        --data-dir /home/ubuntu/mxnet_scala/scala-package/examples/mnist/ \
+java -XX:+PrintGC -Dmxnet.traceLeakedObjects=false -cp $CLASS_PATH \
```
lanking520: Will the user be confused about the result they expect to see since this is a demo script to train their model

scala-package/memory-management.md
```diff
@@ -0,0 +1,117 @@
+# JVM Memory Management
+The Scala and Java binding of Apache MXNet uses native memory(C++ Heap either in RAM or GPU memory) in most of the MXNet Scala objects such as NDArray, Symbol, Executor, KVStore, Data Iterators, etc.,. the Scala classes associated with them act as wrappers, 
+the operations on these objects are directed to the MXNet C++ backend via JNI for performance , so the bytes are also stored in the native heap for fast access.   
+
+The JVM using the Garbage Collector only manages objects allocated in the JVM Heap and is not aware of the memory footprint of these objects in the native memory, hence allocation/deAllocation of the native memory has to be managed by MXNet Scala.  
+Allocating native memory is straight forward and is done during the construction of the object by a calling the associated C++ API through JNI, however since JVM languages do not have destructors, De-Allocation of these objects becomes problematic and has to explicitly de-allocated. 
+To make it easy, MXNet Scala provides a few modes of operation.
```
lanking520: Not finished? Which operations are supported?

scala-package/memory-management.md
```diff
@@ -0,0 +1,117 @@
...
+        val r3 = NDArray.ones(Shape(5, 6))
+        val r4 = NDArray.ones(Shape(7, 8))
+        (r3, r4)
+    }
+    r4
+}
+```
+In the example above, we have two ResourceScopes stacked together, 4 NDArrays `(r1, r2, r3, r4)` are created in the inner scope, the inner scope returns 
+`(r3, r4)`. The ResourceScope code recognizes that it should not de-allocate these objects and automatically moves `r3` and  `r4` to the outer scope. The outer scope 
+returns `r4` from its code-block, so ResourceScope.using removes this from its list of objects to be de-allocated. All other objects are automatically released(native memory) by calling the C++ Backend to free the memory.
```
lanking520: recommend: + and `r3` will be de-allocated from this level

scala-package/memory-management.md
```diff
@@ -0,0 +1,117 @@
...
+
+Apache MXNet uses [Phantom References](https://docs.oracle.com/javase/8/docs/api/java/lang/ref/PhantomReference.html) to track all MXNet Objects that has native memory associated with it. 
+When the Garbage Collector runs, GC identifies unreachable Scala/Java objects in the JVM Heap and finalizes them, 
+we take advantage of Garbage Collector which enqueues objects into a reference queue that are ready to be reclaimed, 
+at which point we do pre-mortem clean up by call the MXNet backend C++ API to free the native memory. 
+ 
+In this approach, you do not have to write any special code to have native memory cleaned up, however this approach solely depends on the Garbage collector to run and find unreachable objects.
+You can control the frequency of Garbage Collector by calling System.gc() at strategic points such as at the end of an epoch or at the end of a mini-batch in Training.
+
+This approach could be suitable for use-cases such as inference on CPUs and you have large amount of Memory(RAM) on your system.
```
lanking520: So in this case, I can call `system.gc()` inside my code in a certain timely manner and most of my `NDArray` would be de-allocated thanks to Phantom Reference?

scala-package/memory-management.md
```diff
@@ -0,0 +1,117 @@
...
+        val dataBatch = itr.next()
+        while(itr.next()) {
+           m.forward(dataBatch)
+           m.backward(dataBatch)
+           m.update()
+        }
+     }
+ }
+}
+
```
Wei-1: I think the indent here is a little weird?

scala-package/memory-management.md
```diff
@@ -0,0 +1,117 @@
...
+```
+or 
+```
+try {
+    ResourceScope scope = new ResourceScope()
+    NDArray r = NDArray.ones((Shape(2,2))
+} finally {
+   scope.close()
+}
+```
```
Wei-1: Maybe `scope.close()` miss a space in the front?

scala-package/memory-management.md
```diff
@@ -0,0 +1,117 @@
+# JVM Memory Management
+The Scala and Java binding of Apache MXNet uses native memory(C++ Heap either in RAM or GPU memory) in most of the MXNet Scala objects such as NDArray, Symbol, Executor, KVStore, Data Iterators, etc.,. the Scala classes associated with them act as wrappers,
```
lupesko: Phrasing nits:
```suggestion
The Scala and Java bindings of Apache MXNet uses native memory (C++ heap either in RAM or in GPU memory) for most of the MXNet Scala objects such as NDArray, Symbol, Executor, KVStore, Data Iterators, etc. The Scala classes associated with them act as wrappers, 
```

scala-package/memory-management.md
```diff
@@ -0,0 +1,117 @@
+# JVM Memory Management
+The Scala and Java binding of Apache MXNet uses native memory(C++ Heap either in RAM or GPU memory) in most of the MXNet Scala objects such as NDArray, Symbol, Executor, KVStore, Data Iterators, etc.,. the Scala classes associated with them act as wrappers, 
+the operations on these objects are directed to the MXNet C++ backend via JNI for performance , so the bytes are also stored in the native heap for fast access.   
+
+The JVM using the Garbage Collector only manages objects allocated in the JVM Heap and is not aware of the memory footprint of these objects in the native memory, hence allocation/deAllocation of the native memory has to be managed by MXNet Scala.
```
lupesko: Nit:
```suggestion
The JVM using the Garbage Collector only manages objects allocated in the JVM Heap and is not aware of the memory footprint of these objects in the native memory, hence allocation/deallocation of the native memory has to be managed by MXNet Scala.  
```

scala-package/memory-management.md
```diff
@@ -0,0 +1,117 @@
+# JVM Memory Management
+The Scala and Java binding of Apache MXNet uses native memory(C++ Heap either in RAM or GPU memory) in most of the MXNet Scala objects such as NDArray, Symbol, Executor, KVStore, Data Iterators, etc.,. the Scala classes associated with them act as wrappers, 
+the operations on these objects are directed to the MXNet C++ backend via JNI for performance , so the bytes are also stored in the native heap for fast access.   
+
+The JVM using the Garbage Collector only manages objects allocated in the JVM Heap and is not aware of the memory footprint of these objects in the native memory, hence allocation/deAllocation of the native memory has to be managed by MXNet Scala.  
+Allocating native memory is straight forward and is done during the construction of the object by a calling the associated C++ API through JNI, however since JVM languages do not have destructors, De-Allocation of these objects becomes problematic and has to explicitly de-allocated.
```
lupesko: Phrasing suggestions:
```suggestion
Allocating native memory is straightforward and is done during the construction of the object by a calling the associated C++ API through JNI. However, since JVM languages do not have destructors, deallocation of these objects needs to be done explicitly. 
```

scala-package/memory-management.md
lupesko:
```suggestion
To make it easy, MXNet Scala provides a few modes of operation, explained in detail below.
```

lupesko:
Thanks for the contribution @nswamy .
Pulling in @aaronmarkham to help with the doc review.

lanking520:
Hi @nswamy could you please address the change and let's get this PR forward

scala-package/examples/scripts/run_train_mnist.sh
piyushghai: Is there a reason why the min and max heap size has been removed from the run command ?

scala-package/memory-management.md
```diff
@@ -0,0 +1,117 @@
...
+```
+def showDispose(): Unit = {
+    val r = NDArray.ones(Shape (2, 2))
+    r.dispose()
+}
+```
+
+## Memory Management in Java
+Memory Management in MXNet Java is similar to Scala, We recommend to use [ResourceScope](https://github.com/apache/incubator-mxnet/blob/master/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala#L32) in a `try-with-resources` block or in a `try-finally` block.   
+Java 8 onwards supports [try-with-resource](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) where the resources declared in the try block is automatically closed.
```
piyushghai: I think the try-with-resource block is supported Java 7 onwards. 
Eg : http://tutorials.jenkov.com/java-exception-handling/try-with-resources.html

scala-package/examples/scripts/run_train_mnist.sh
nswamy: I used it for my testing, there is no need to put those restrictions. user can change it to whatever they like.

scala-package/memory-management.md
```diff
@@ -0,0 +1,120 @@
...
+}
+
+```  
+       
+### 2.  Using Phantom References (Recommended for some use cases)
+
+Apache MXNet uses [Phantom References](https://docs.oracle.com/javase/8/docs/api/java/lang/ref/PhantomReference.html) to track all MXNet Objects that has native memory associated with it. 
+When the Garbage Collector runs, GC identifies unreachable Scala/Java objects in the JVM Heap and finalizes them, 
+we take advantage of Garbage Collector which enqueues objects into a reference queue that are ready to be reclaimed, 
+at which point we do pre-mortem clean up by call the MXNet backend C++ API to free the native memory.
```
harshp8l: [nit] typo: pre-mortem clean up by calling*

scala-package/memory-management.md
```diff
@@ -0,0 +1,120 @@
...
+     }
+   }
+ }
+}
+
+```  
+       
+### 2.  Using Phantom References (Recommended for some use cases)
+
+Apache MXNet uses [Phantom References](https://docs.oracle.com/javase/8/docs/api/java/lang/ref/PhantomReference.html) to track all MXNet Objects that has native memory associated with it.
```
piyushghai: [nit] : 'have' instead of 'has'

scala-package/memory-management.md
```diff
@@ -0,0 +1,120 @@
...
+```
+def showDispose(): Unit = {
+    val r = NDArray.ones(Shape (2, 2))
+    r.dispose()
+}
+```
+
+## 3. Memory Management in Java
+Memory Management in MXNet Java is similar to Scala, We recommend to use [ResourceScope](https://github.com/apache/incubator-mxnet/blob/master/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala#L32) in a `try-with-resources` block or in a `try-finally` block.   
+Java 7 onwards supports [try-with-resource](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) where the resources declared in the try block is automatically closed.
```
harshp8l: [nit] typo: where the resources declared in the try block are*

scala-package/memory-management.md
```diff
@@ -0,0 +1,120 @@
...
+try {
+    ResourceScope scope = new ResourceScope()
+    NDArray test = NDArray.ones((Shape(2,2))
+} finally {
+    scope.close()
+}
+``` 
+**Note:**
+ResourceScope within a try block tracks all MXNet Native Object Allocations (NDArray, Symbol, Executor, etc.,) and deallocates at
+the end of the try block even the objects that are returned, ie., in the above even if `test` were to be returned the native memory associated
```
harshp8l: (Confusing sentence ...)
Can be cleared up ->
deallocates them* (or mention deallocates the objects specifically) at the end of the try block.
Then say, For example, .....
I think this would help make it clear

harshp8l:
Thanks, @nswamy this looks great. This was an informative doc - very worthwhile.

andrewfayres:
Overall the content looks good. I'll work to correct some of the minor grammatical stuff after I grab some dinner.

andrewfayres:
Instead of nitpicking at this I made some edits and pushed a new commit like we discussed.

ddavydenko:
@nswamy , do you think you can get to failures in CI in order to push this PR to completion?

zachgk:
I put some suggested edits in a PR at https://github.com/nswamy/incubator-mxnet/pull/3

nswamy:
Thanks all(@andrewfayres , @zachgk , @lupesko @lanking520 ) for the review and edits. This is ready to be merge, hopefully CI passes and we can merge.

PR #13142, opened 2018-11-06
[WIP] cudnnFind** API does not return deterministic algorithms

apeforest:
## Description ##
I have found that cudnnFind** API sometimes returns nondeterministic algorithm even though the return type is CUDNN_DETERMINISTIC. On the other hand, choosing cudnnGet** API does not have this issue. This PR is a WA before the root cause is confirmed. 

Require reviews by @DickJC123 @ptrendx Thanks!


anirudhacharya:
@mxnet-label-bot [pr-work-in-progress]

kalyc:
Thanks for your contribution @apeforest 
@DickJC123 @ptrendx could you please review the PR?

stu1130:
@DickJC123 @ptrendx ping here for review thanks

roywei:
@DickJC123 could you help take a look? Thanks a lot!

sandeep-krishnamurthy:
I think this issue is solved with another PR from @DickJC123 , can you please correct me if this is correct?

DickJC123:
This PR addresses a different type of determinism.  My PR improved the framework's ability to select the same conv algorithm within a model run and from one run to the next based on improvements in the use of cudnnFind().  This PR deals with making that selection only within the domain of 'deterministic algorithms.'  Some algos have unordered threads summing into a result value using atomic adds.  These algos do not always produce the same result from the same input data (i.e. they are non-deterministic).  This PR deals with limiting the cudnnFind() selection to only deterministic algos.

apeforest:
@DickJC123 I think this PR is meant to raise a question rather than to be merged. My question is that when the result.determinism flag is equal to `CUDNN_DETERMINISTIC` as in this [line](https://github.com/apache/incubator-mxnet/blob/7b787d3cf37a994e51a3420149c6bbd48ca09c8c/src/operator/nn/cudnn/cudnn_convolution-inl.h#L901), should the user always expect a deterministic outcome? From what we had observed, it does not. Hence, this PR is to ask if we need to check anything else to guarantee a deterministic convolution result. Thanks a lot!

DickJC123:
I have a possible explanation.  When cudnnFind measures the runtime of algos that have similar runtimes, the selection of which is fastest may differ from run to run.  Filtering the algorithms based on MXNET_ENFORCE_DETERMINISM=1 just throws out the non-deterministic algos.  However, the results from run to run may differ based on the non-deterministic choice of which deterministic algo to run.

pinaraws:
Is this still WIP @apeforest ?

piyushghai:
@apeforest Pinging again for updating this PR.

pinaraws:
@apeforest Gentle ping for updates

PR #13143, opened 2018-11-06
[MXNET-1206] Support NDArray indexing with None and Ellipsis

### LINKED ISSUE #13134: [Python]: Indexing with None and Ellipsis
The expectation of Python users familiar with NumPy is that, e.g., `array[None, :]` and `array[..., 0]` work, which is not the case currently with `NDArray`.

Both of these would be quite straightforward to add. I could do that, but I just want to make sure that it's not deliberately left out for some reason.

I'm working with a manual build from a91b364c490b163da424a1c2935652ea7b799050.

### LINKED ISSUE #13166: [Python] __getitem__ copies too often
## Description

While working on #13134 I noticed that `NDArray.__getitem__` does not really do a good job of determining whether or not a view instead of a copy can be returned.

For instance, wrapping any indexing object into a 1-element tuple should have no effect at all (at least if one wants to be consistent with NumPy), but the wrapping currently leads to a copy.
Or, indexing a 3D array as `array[0]` should be equivalent to `array[0, :, :]`, but the latter copies while the former does not.

Here are some examples:
```py
def getitem_returns_view(x, idx):
    tmp = x.copy()
    tmp[:] = 0
    y = tmp[idx]
    y[:] = 1
    return all(yi == xi for yi, xi in zip(y.reshape((-1)), tmp[idx].reshape((-1,))))

x = nd.zeros((3, 4, 5))

## Integers

# Okay
getitem_returns_view(x, 0)  # True
getitem_returns_view(x.asnumpy(), 0)  # True

# NOT okay
getitem_returns_view(x, (0,))  # False
getitem_returns_view(x.asnumpy(), (0,))  # True

# Same with 1 instead of 0

## Slices

# Okay
getitem_returns_view(x, slice(None))  # True
getitem_returns_view(x.asnumpy(), slice(None))  # True

# NOT okay
getitem_returns_view(x, (slice(None),))  # False
getitem_returns_view(x.asnumpy(), (slice(None),))  # True

# NOT okay
getitem_returns_view(x, (0, slice(None)))  # False
getitem_returns_view(x.asnumpy(), (0, slice(None)))  # True

# NOT okay
getitem_returns_view(x, (0, slice(None), slice(None)))  # False
getitem_returns_view(x.asnumpy(), (0, slice(None), slice(None)))  # True

# NOT at all okay
getitem_returns_view(x, (slice(None), slice(None), slice(None)))  # False
getitem_returns_view(x.asnumpy(), (slice(None), slice(None), slice(None)))  # True
```

In particular the last case `array[:, :, :]`, which is explicitly *everything*, should definitely not copy.

## Suggestion

Instead of the current way of special-casing integers, lists and slice objects, I think it would be more robust to always sanitize the indexing object, in particular:

- If it's not a tuple, wrap it in a 1-tuple.
- Expand it to length `array.ndim` by adding `slice(None)` to the right.

Then the subsequent logic can look at all the entries in the tuple and determine whether the result will be contiguous (is it correct that the backend does not yet support strided views?). That's not entirely trivial but certainly doable.

Of course, things get a little bit more involved if `None` (new axis) and `Ellipsis` (`...`) are allowed as well. But the code in https://github.com/apache/incubator-mxnet/pull/13143 already does the above sanitization for that case, it just doesn't always use it to avoid copies.

kohr-h:
## Description ##

This PR adds the capability of indexing `NDArray` with `None` to create new axes, and with `...` (`Ellipsis`) as a shorthand for an adequate number of `:` (`slice(None)`).

Closes #13134
Closes #13166

### Essentials ###

### Changes ###
- `NDArray.__getitem__` and `NDArray.__setitem__` now first expand `None` and `Ellipsis` entries in indices.
- `__setitem__` ignores `None` entries, which corresponds to the behavior of NumPy.
- If the indexed portion is contiguous, `__getitem__` now **always** returns a view.

## Comments ##
- I've noted that indexing is not as copy-free as it could be. For instance:
  ```py
  >>> x = nd.ones(3)
  >>> y = x[:2]  # same as y = x[slice(2)]
  >>> y[:] = 0
  >>> x  # as expected
  [0. 0. 1.]
  <NDArray 3 @cpu(0)>
  >>> y = x[(slice(2),)]  # Should be equivalent
  >>> y[:] = -17
  >>> x  # unchanged!
  [0. 0. 1.]
  <NDArray 3 @cpu(0)>
  ```
- Another nifty case is this one:
  ```py
  >>> x = nd.ones((2, 3))
  >>> y = x[:2]
  >>> y[:] = 0
  >>> x  # as expected
  [[0. 0. 0.]
   [1. 1. 1.]]
  <NDArray 2x3 @cpu(0)>
  >>> y = x[:2, :]
  >>> y[:] = -17
  >>> x  # unchanged!
  [[0. 0. 0.]
   [1. 1. 1.]]
  <NDArray 2x3 @cpu(0)>
  ```
I think that the implementation does not live up to the promise of the documentation:
```
Returns a sliced view of this array if the elements fetched are contiguous in memory;
otherwise, returns a newly created NDArray.
```
In particular, the "detection" of contiguousness is not good.

I'll make that an extra issue, but it's related to this PR, so I mention it here.

## Remaining TODOs
- [x] Fix error in `File "/work/mxnet/tests/python/unittest/test_dgl_graph.py", line 64, in check_compact` (from `test_uniform_sample`)
- [x] Fix failure in `File "/work/mxnet/tests/python/unittest/test_operator.py", line 4195, in test_empty_tensor` (from `test_tile`)
- [x] Fix failure in `File "/work/mxnet/tests/python/gpu/../unittest/test_operator.py", line 4290, in test_empty_indices` (from `test_one_hot`)
- [x] Fix linter errors/warnings


anirudhacharya:
@kohr-h please ping here once you think the PR is ready for review.

kohr-h:
Yes, will do. The whole thing is more complex than anticipated (as usual), in particular with advanced indexing. So it's still work-in-progress. But I'm determined to push it through.

vandanavk:
@mxnet-label-bot update[NDArray, pr-work-in-progress]

roywei:
@kohr-h Thanks for the contribution! Here are some resources if you need help:
1. you can take a look at [MXNet wiki](https://cwiki.apache.org/confluence/display/MXNET/Development) for developers to understand the design
2. Ask questions on [our forum](https://discuss.mxnet.io/) and discuss if you have any difficulty.
3. For API changes, we can discuss on [devlist and slack channel](https://mxnet.incubator.apache.org/community/mxnet_channels.html)

kohr-h:
Getting closer. Advanced indexing with `__getitem__` should work now, but I'll add test cases for that.

Roshrini:
@kohr-h Thanks for working on this! can you look into failing CI builds?

kohr-h:
> @kohr-h Thanks for working on this! can you look into failing CI builds?

@Roshrini Yes will do. I haven't had much time during Holiday and the next time will also be a bit on and off, but I'll see what I can do.

vandanavk:
@kohr-h Could add "Fixes https://github.com/apache/incubator-mxnet/issues/13166" in the PR description so that #13166 is closed too.

Also could you re-trigger CI

vandanavk:
@mxnet-label-bot update [pr-awaiting-testing, NDArray]

kohr-h:
This is almost finished. There are 2 failing tests for which I have no idea why they're failing and how to fix them. I'd appreciate some help.

Here are the stacktraces (ignore the `_gpu` part, they also fail on CPU):
```
FAIL: test_operator_gpu.test_tile
    self.test(*self.arg)
  File "/usr/lib/python3.6/site-packages/nose/util.py", line 620, in newfunc
    return func(*arg, **kw)
    orig_test(*args, **kwargs)
  File "/work/mxnet/tests/python/gpu/../unittest/test_operator.py", line 4246, in test_tile
    test_empty_tensor()
  File "/work/mxnet/tests/python/gpu/../unittest/test_operator.py", line 4195, in test_empty_tensor
    assert same(a_tiled, b_tiled)
AssertionError: 
common: INFO: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=2048974122 to reproduce.

FAIL: test_operator_gpu.test_one_hot
    self.test(*self.arg)
  File "/usr/lib/python3.6/site-packages/nose/util.py", line 620, in newfunc
    return func(*arg, **kw)
    orig_test(*args, **kwargs)
  File "/work/mxnet/tests/python/gpu/../unittest/test_operator.py", line 4306, in test_one_hot
    test_empty_indices()
  File "/work/mxnet/tests/python/gpu/../unittest/test_operator.py", line 4290, in test_empty_indices
    assert same(expected_array, mx_one_hot_array)
AssertionError: 
common: INFO: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=776351319 to reproduce.
```

Both tests fail because tensors with 0s in their shapes are handled incorrectly in the two failing functions `tile` and `one_hot`. I didn't change any of those, and the closest thing I changed (my only change in the backend) is [this one](https://github.com/apache/incubator-mxnet/pull/13143/files#diff-d5c9c7d539907caef01a08a74e16592b) in `slice_forward`. But I couldn't find a connection to the other ops.

karan6181:
@szha and @anirudh2290 Could you please have a took at this PR? Thank you!

abhinavs95:
@kohr-h There are 7 CI tests failing currently, could you have a look? Thanks!

kohr-h:
@abhinavs95 See [my comment](https://github.com/apache/incubator-mxnet/pull/13143#issuecomment-474256702) above. I'd love to fix them but I don't know why they're failing in the first place and could use some input. It's probably only about the (tiny) portion of C++ code in the changeset.

abhinavs95:
@kohr-h I understand, I thought I should provide you with an update since there are 7 tests failing now as opposed to 2 you mentioned in your comment.

@szha @anirudh2290 could you please have a look and see if you can help?

kohr-h:
@abhinavs95 Sure, thanks for keeping up the conversation.
What I was referring to as failures were 2 specific unittests, but they fail on 7 CIs (actually only on 6, the 7th is just linting which I'll take on when the code issues are sorted out).

piyushghai:
@kohr-h Did you get a chance to resolve the CI issues on this one ? :) This seems to be a super useful PR

kohr-h:
@piyushghai No but I'll have another stab at it tonight.

kohr-h:
This PR is now up-to-date with the latest 0-size tensor changes from #14661, and tests pass. Ready for review, finally.

reminisce:
@kohr-h Thank you for closing the gap of mxnet ndarray indexing with NumPy. Sorry for the delay. I will review the PR this or next week. :)

tests/python/unittest/test_ndarray.py
```diff
@@ -1292,7 +1358,9 @@ def test_setitem_autograd(np_array, index):
         try:
             with mx.autograd.record():
                 x[index] = y
-                assert False  # should not reach here
+                # TODO: figure out why `None` doesn't raise
```
reminisce: I guess it's because this case is equivalent to `y.copyto(x)`, where `x` is only passed into the autograd recorder as a pure output, while in other cases, `x` is passed as both an input and an output tensor. Therefore, it's fine for skipping this case.

tests/python/unittest/test_operator.py
```diff
@@ -4388,13 +4388,14 @@ def test_normal_case():
 
     def test_empty_tensor():
         shape = (2, 3, 0, 4)
-        a = np.array([], dtype=np.int32).reshape(shape)
-        b = mx.nd.array(a, ctx=default_context(), dtype=a.dtype)
-        reps = (2, 4, 6)
+        with mx.np_compat():
```
reminisce: nit: Instead of using the `with` statement, you can decorate this function with `mx.use_np_compat` to avoid indenting the code. This would introduce less difference between two code versions.

tests/python/unittest/test_operator.py
```diff
@@ -4484,13 +4485,15 @@ def test_normal_case(index_type=np.int32):
...
     def test_empty_indices():
         shape = (2, 0, 9, 3)
-        indices = np.array([]).reshape(shape)
-        depth = 10
-        mx_one_hot_array = mx.nd.one_hot(
-            mx.nd.array(indices, ctx=default_context(), dtype=np.int32),
-            depth=depth, dtype=np.int32).asnumpy()
-        expected_array = np.array([], dtype=np.int32).reshape(shape + (depth, ))
-        assert same(expected_array, mx_one_hot_array)
+        with mx.np_compat():
```
reminisce: Same as above.

python/mxnet/ndarray/ndarray.py
```diff
@@ -387,19 +388,15 @@ def __setstate__(self, state):
...
     def __setitem__(self, key, value):
         """x.__setitem__(i, y) <=> x[i]=y
 
-        Sets value to self[key]. This functions supports advanced indexing defined in the following reference with
-        some restrictions.
-
-        https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.indexing.html#combining-advanced-and-basic-indexing
+        Sets ``self[key]`` to ``value``.
 
-        - If key is a list type, only a list of integers is supported, e.g. key=[1, 2] is supported,
```
reminisce: We only supported list of integers when the `key` is a list because we find the behavior of nested lists are confusing. For example, `x[[1], [2]]` is equivalent to `x[[[1], [2]]]`. Has this implementation enable indexing with key as a nested list? If so, could you explain and document the behavior consistent with numpy?

abhinavs95:
@kohr-h Could you please address the review comments?

piyushghai:
@kohr-h Gentle ping ....

tests/python/unittest/test_ndarray.py
```diff
@@ -1208,22 +1266,24 @@ def test_getitem(np_array, index, is_scalar=False):
...
+                idx.asnumpy() if isinstance(idx, mx.nd.NDArray) else idx
+                for idx in index
+            )
 
         np_indexed_array = np_array[np_index]
         mx_array = mx.nd.array(np_array, dtype=np_array.dtype)
-        mx_indexed_array = mx_array[index]
+        try:
+            mx_indexed_array = mx_array[index]
+        except Exception as e:
```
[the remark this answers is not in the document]
kohr-h: No, it's a debugging leftover.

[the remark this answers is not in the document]
kohr-h: OTOH, there's more code like this in the test suite, like `assert <stmt>, "informative error text"`. It adds information to the plain failure report, so I'd prefer to leave it in.

tests/python/unittest/test_operator.py
kohr-h: But decorating the whole function means not testing the version without `np_compat`. Not sure if that's desirable.

python/mxnet/ndarray/ndarray.py
kohr-h: The whole topic of advanced indexing is deeply confusing, and I've heard people say that it's too advanced to be understandable (and useful).
Your second example, though, is deprecated and will change in the future (to be aware of here as well!). If wrapping the whole index in a sequence, a `tuple` must be used, not a `list`. Here's the warning from current Numpy:

    FutureWarning: Using a non-tuple sequence for multidimensional indexing is
    deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be
    interpreted as an array index, `arr[np.array(seq)]`, which will result either in an
    error or a different result.

Re: compatibility with Numpy: within the reach of the current unit tests, I think advanced indexing is 100 % compatible with Numpy's. So I guess we can just point to the Numpy documentation for a detailed explanation.
The only remaining difference is with respect to strided views, where Numpy has more cases (in basic indexing) where a view is returned. Getting on par there would require a much bigger effort (extend storage backend).

tests/python/unittest/test_ndarray.py
kohr-h: Okay, thanks for the pointer. I'll change the comment. There are a couple of cases where `None` is in the indexing tuple, and those raise as expected.

Roshrini:
@kohr-h Can you please revisit this PR again?

kohr-h:
I think I addressed all the review comments, and an in-depth review is still required I guess.

tests/python/unittest/test_operator.py
reminisce: Empty tensors are not supported anyway without numpy shape semantics. The original test put here was actually not desirable. I'm fine with your change here. It's just a matter of code tidiness.

reminisce:
Thanks for addressing the cr comments. Since the logic of integer indexing and first-axis contiguous slicing have been organized into a more general case handling flow, could you verify there is no performance regression in the situation where an output ndarray sharing the memory with the input ndarray? This kind of indexing is used a lot in Gluon dataset loaders and is critical to the training performance.

tests/python/unittest/test_operator.py
kohr-h: A-ha, so the old test was never guaranteed to pass? In that case, I'll use the decorator, of course.

kohr-h:
Sure, I understand that it's important to avoid performance regressions after such a change.

Is there any performance testing suite I could use? Otherwise I'd do some home-brew benchmarks.

reminisce:
You can run some simple performance tests involving integer indexing and first-axis slicing. For example, you can get the time of "splitting" a batch of images, say shape (256, 3, 224, 224) into four sub-batches, each of which shape is (64, 3, 224, 224).

reminisce:
@kohr-h Gentle ping. Any updates on the performance tests? This PR is useful in getting NumPy consistent behavior for ndarray indexing and aligns with our efforts of introducing numpy-like coding experience into mxnet. Would appreciate if we can merge this PR sooner than later.

kohr-h:
Hi, sorry, it's vacation time and in the past couple weeks and in the following weeks there was no/won't be any time for me to run the performance test. I'll get onto it when I'm back.

zoeygxy:
Hi, thank you for your work! I am from MXNet team and I'm currently working on indexing of mxnet.numpy.ndarray, the child class of NDArray class. The numpy.ndarray indexing is currently based on your version of ``ndarray.py``, and I have thus made some changes to this file on [my branch](https://github.com/zoeygxy/incubator-mxnet/commits/numpy_indexing):

* Added support of ``None`` in NDArray advanced slicing, so that all the test cases in your ``test_ndarray.py`` are passed. Example of supported case:

```
index = (slice(None), slice(3, 5), None, None, [2, 3], [3, 4])
```

* Wrapped some operators as NDArray class methods so that my implementation can take advantage of your exisitng code by inheritance.

You can still make changes when you are back and I will adapt my code to them. Thanks for your contribution!

sandeep-krishnamurthy:
FYI - @apeforest @access2rohit

reminisce:
@kohr-h Our work on mxnet.numpy highly depends on your work in this PR. We would like to see this PR get merged sooner than later. If you don't mind, I can resolve the conflicts in this PR and try to get it merged if it passes CI. We will also do the performance validation. Thanks for your great contribution.

kohr-h:
Hi @reminisce, I'm sorry for the delay. I will be on vacation for another two weeks, so if you want to move this PR ahead please feel free to resolve the conflicts and do the performance tests.
If you have a question I'll try to answer as best I can.

On August 2, 2019 4:35:17 PM GMT+08:00, reminisce <notifications@github.com> wrote:
>@kohr-h Our work on mxnet.numpy highly depends on your work in this PR.
>We would like to see this PR get merged sooner than later. If you don't
>mind, I can resolve the conflicts in this PR and try to get it merged
>if it passes CI. We will also do the performance validation. Thanks for
>your great contribution.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/apache/incubator-mxnet/pull/13143#issuecomment-517612386

-- 
Sent from my mobile phone.

reminisce:
@kohr-h Thanks for the reply. @zoeygxy did some benchmark and found that
1. Indexing operations that should return views are slower than before.
2. Indexing operations that should return newly allocated `NDArray`s are faster than before.

We can get this PR merged first and take care of the cases that return views in a later PR.

PS: benchmark results
[old_indexing.txt](https://github.com/apache/incubator-mxnet/files/3463396/old_indexing.txt)
[new_indexing.txt](https://github.com/apache/incubator-mxnet/files/3463397/new_indexing.txt)

reminisce:
@kohr-h I rebased your branch with master. After the CI passes, I will merge the PR.

I also commented out the test cases where `None` is used in advanced indices since it has not been supported yet. @zoeygxy has implemented this in [her branch](https://github.com/zoeygxy/incubator-mxnet/blob/numpy_indexing/python/mxnet/ndarray/ndarray.py) and will PR the change together with indexing support for `mxnet.numpy.ndarray`. In summary, there are two action items (by @zoeygxy) after this PR is merged:
1. Fix indexing performance regression for outputs being views.
2. Implement `None` support for advanced indices.

Please let us know if you have other concerns or suggestions. Thanks again for your contribution.

kohr-h:
Great to see this PR merged, thanks for the "last mile" effort @reminisce and @zoeygxy.

Regarding the performance differences, the trend you observed is somewhat to be expected. The checks performed in basic indexing are more comprehensive than before, and in more cases than before, views are being returned. Thus, the cases where views have always been returned are a bit slower because of extra overhead, but the cases that turned from copy to view should amortize easily.

The advanced indexing performance could have gone either way, so I'm happy about the improvement there.

To see the effect more clearly, I think it would make sense to test very small arrays (pure overhead) and very large arrays (purely copy vs. view).

How much are you worried about the performance hit for the view cases? If you really want to be on par with the old speed, you could re-introduce the same special-casing as in the old code, but IMO that would make the code more complex. My implementation strives to be as generic as possible and takes the small performance hit of manipulating tuples and slice objects a bit too much for simple cases.

reminisce:
@kohr-h Thanks for the response. Since returning views is a very frequent operation in many dataset loaders in MXNet, we have to keep the performance on a par with the previous version. @zoeygxy will keep working on improving the performance of the cases where views are returned. Although such special handling may lead code to look a little bit disorganized, the performance gain is still worth the effort because we don't want to introduce performance regression in training jobs.

DickJC123:
@kohr-h 
A `git show` of this PR's commit includes:
```
diff --git a/python/mxnet/test_utils.py b/python/mxnet/test_utils.py
index 2275f4d5d..89013f439 100644
--- a/python/mxnet/test_utils.py
+++ b/python/mxnet/test_utils.py
@@ -1074,7 +1074,11 @@ def check_symbolic_forward(sym, location, expected, rtol=1E-4, atol=None,
 
     executor = sym.bind(ctx=ctx, args=location, args_grad=args_grad_data, aux_states=aux_states)
     for g in executor.grad_arrays:
-        g[:] = 0
+        print(g.shape)
+        if g.ndim == 0:
+            g[()] = 0
+        else:
+            g[:] = 0
 
```
Please confirm that the line `print(g.shape)` was a bit of debugging code inadvertently left in, and so should be removed.

kohr-h:
@DickJC123 It's part of fe6336df4064f3086443d5abb7535fd549dd3a48 so I'll bounce the question to @reminisce.

kohr-h:
> @kohr-h Thanks for the response. Since returning views is a very frequent operation in many dataset loaders in MXNet, we have to keep the performance on a par with the previous version. @zoeygxy will keep working on improving the performance of the cases where views are returned. Although such special handling may lead code to look a little bit disorganized, the performance gain is still worth the effort because we don't want to introduce performance regression in training jobs.

@reminisce Makes sense as a trade-off.

zoeygxy:
> A `git show` of this PR's commit includes:
> 
> ```
> diff --git a/python/mxnet/test_utils.py b/python/mxnet/test_utils.py
> index 2275f4d5d..89013f439 100644
> --- a/python/mxnet/test_utils.py
> +++ b/python/mxnet/test_utils.py
> @@ -1074,7 +1074,11 @@ def check_symbolic_forward(sym, location, expected, rtol=1E-4, atol=None,
>  
>      executor = sym.bind(ctx=ctx, args=location, args_grad=args_grad_data, aux_states=aux_states)
>      for g in executor.grad_arrays:
> -        g[:] = 0
> +        print(g.shape)
> +        if g.ndim == 0:
> +            g[()] = 0
> +        else:
> +            g[:] = 0
>  
> ```
> 
> Please confirm that the line `print(g.shape)` was a bit of debugging code inadvertently left in, and so should be removed.

@DickJC123 I believe so. Thanks for pointing out! I will take care of it.

Issue #13161, opened 2018-11-07
Scaladoc build errors


zachgk:
## Description

When compiling the docs, there are a number of errors and warnings while generating the scaladocs.  Sample output is below:
```
Execute "cd /home/ubuntu/incubator-mxnet/docs/../scala-package; scaladoc `find . -type f -name "*.scala" | egrep "\/core|\/infer" | egrep -v "Suite"`; exit 0"
./core/src/main/scala/org/apache/mxnet/FeedForward.scala:23: error: object slf4j is not a member of package org
import org.slf4j.{Logger, LoggerFactory}
           ^
./core/src/main/scala/org/apache/mxnet/FeedForward.scala:61: error: not found: type Logger
  val logger: Logger = LoggerFactory.getLogger(classOf[FeedForward])
              ^
./core/src/main/scala/org/apache/mxnet/FeedForward.scala:61: error: not found: value LoggerFactory
  val logger: Logger = LoggerFactory.getLogger(classOf[FeedForward])
                       ^
./core/src/main/scala/org/apache/mxnet/FeedForward.scala:292: error: not found: type Logger
          logger: Logger, workLoadList: Seq[Float]): Unit = {
                  ^
./core/src/main/scala/org/apache/mxnet/FeedForward.scala:369: error: not found: type Logger
                  batchEndCallback: BatchEndCallback = null, logger: Logger = FeedForward.logger,
                                                                     ^
./core/src/main/scala/org/apache/mxnet/FeedForward.scala:328: error: not found: type Logger
          batchEndCallback: BatchEndCallback, logger: Logger,
                                                      ^
./core/src/main/scala/org/apache/mxnet/FeedForward.scala:464: error: not found: type Logger
  private val logger: Logger = LoggerFactory.getLogger(classOf[FeedForward])
                      ^
./core/src/main/scala/org/apache/mxnet/FeedForward.scala:464: error: not found: value LoggerFactory
  private val logger: Logger = LoggerFactory.getLogger(classOf[FeedForward])
                               ^
./core/src/main/scala/org/apache/mxnet/FeedForward.scala:542: error: not found: type Logger
    private var logger: Logger = FeedForward.logger
                        ^
./core/src/main/scala/org/apache/mxnet/FeedForward.scala:697: error: not found: type Logger
    def setLogger(logger: Logger): Builder = {
                          ^
./core/src/main/scala/org/apache/mxnet/IO.scala:23: error: object slf4j is not a member of package org
[...]

PR #13189, opened 2018-11-08
Fix scaladoc build errors

zachgk:
## Description ##
This PR fixes the scaladoc build errors and warnings noted in #13161. 

### Essentials ###

### Changes ###
- [x] Add classpath to scaladoc build fixing the dependency erros
- [x] Unrelated small fix for duplicate scalastyle plugin declaration in scala/core/pom file

## Comments ##

@lanking520 @andrewfayres @aaronmarkham @nswamy


harshp8l:

docs/mxdoc.py
```diff
@@ -109,7 +109,13 @@ def build_scala_docs(app):
     """build scala doc and then move the outdir"""
     scala_path = app.builder.srcdir + '/../scala-package'
     # scaldoc fails on some apis, so exit 0 to pass the check
-    _run_cmd('cd ' + scala_path + '; scaladoc `find . -type f -name "*.scala" | egrep \"\/core|\/infer\" | egrep -v \"Suite\"`; exit 0')
+    scala_doc_sources = 'find . -type f -name "*.scala" | egrep \"\.\/core|\.\/infer\" | egrep -v \"Suite\"'
+    scala_doc_classpath = ':'.join([
+        '`find native -name "*.jar" | grep "target/lib/" | tr "\n" ":" `',
+        '`find macros -name "*-SNAPSHOT.jar" | tr "\n" ":" `'
```
lanking520: Not sure why include macros SNAPSHOT?

zachgk: We only build the scaladocs for the core and infer modules. So, I am treating the macros build as a dependency (which is required to avoid some missing dependency errors). The other option is that we also build docs for the macros as well, but I think it is better to hide them since it is not really for the users

zachgk:
The current scaladoc build command and output are as follows (note that this is after https://github.com/apache/incubator-mxnet/pull/13210 as well):
```
ubuntu@ip-172-31-68-239:~/incubator-mxnet/scala-package$ scaladoc `find . -type f -name "*.scala" | egrep "\.\/core|\.\/infer" | egrep -v "Suite"` -classpath `find native -name "*.jar" | grep "target/lib/" | tr '\n' ':'`:`find macros -name "*-SNAPSHOT.jar"` -feature -deprecation
model contains 139 documentable templates
./core/src/main/scala/org/apache/mxnet/module/BaseModule.scala:368: warning: Could not find any member to link for "IOException".
  /**
  ^
one warning found
```

I have enabled the feature and deprecation flags which show additional warnings (which have also been fixed).  The fixed errors and warnings were:
- Missing dependencies fixed by adding target/lib to the scaladoc generation classpath
- Treat macro module as dependency for docs by adding it to classpath as well
- Remove "@throws" no longer in use
- Scaladoc interprets contents surrounded by [[ as a link and throws a warning if it can't be resolved.  It is used for NDArrays so I convert them to [ [ instead
- Various scala features such as implicit methods require a corresponding import from the scala.language package
- The "@deprecated" flag takes two parameters: a message and the version/date of deprecation (https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html)

Comments about the remaining IOException warning are in the source issue https://github.com/apache/incubator-mxnet/issues/13161.

lanking520:
@zachgk nice work!

aaronmarkham:
Restarting CI since there was a timeout in the last run.

zachgk:
@aaronmarkham I don't know about the clojure, but the classpath does need to be added to the java build step. Scala right now includes the dependencies and macros in the classpath. I think you might also need scala core and scala infer added to the java classpath.

nswamy:
thanks for fixing the warnings zack!, i resolved conflicts, will merge as soon as the CI finishes.

kalyc:

PR #13190, opened 2018-11-08
[WIP] catch floating number divide by zero

apeforest:
## Description ##
NaN errors could be due to divide-by-zero floating number in C++ backend. It is not caught by default. 
Turnining on the signal exception can report such errors when running python unit test.
However, this exception currently only works for Linux OS. Need more feedbacks on a better handling cross platforms.


src/engine/engine.cc
```diff
@@ -35,6 +55,18 @@ inline Engine* CreateEngine() {
   if (type == nullptr) type = "ThreadedEnginePerDevice";
   std::string stype = type;
 
+#ifdef _TEST_FLOAT_EXCEPTION
+  struct sigaction act;
+  memset(&act, '\0', sizeof(act));
+  act.sa_sigaction = &float_num_err_handler;
+  act.sa_flags = SA_SIGINFO;
+  if (sigaction(SIGFPE, &act, NULL) < 0) {
+    fprintf(stderr, "Cannot register signal handler\n");
```
yuxihu: Might be also good to log the errno in the message such that we can use it to see why it failed.

src/engine/engine.cc
```diff
@@ -23,10 +23,30 @@
  * \brief Implementation of engine.
  */
 #include <mxnet/engine.h>
+#include <fenv.h>
```
yuxihu: QQ: do we follow any rules (like alphabetic order of library name) to include libraries?

harshp8l:

stu1130:
@apeforest could you address the comments of @yuxihu Thanks!

src/engine/engine.cc
```diff
@@ -23,10 +23,30 @@
...
+    fprintf(stderr, ">>> Integer division by zero\n");
+    break;
+  case FPE_FLTOVF:
+    fprintf(stderr, ">>> Floating overflow trap\n");
+    break;
+  case FPE_FLTDIV:
+    fprintf(stderr, ">>> Floating/decimal division by zero\n");
+    break;
+  }
+  assert(0);
```
stu1130: why adding assert(0) here?

roywei:
@apeforest ping for update. Thanks!

sandeep-krishnamurthy:
@anirudh2290 - FYI

pinaraws:
@apeforest Any updates?

apeforest:
Don't have time to work on this now. Closing it.

PR #13107, opened 2018-11-04
Port of scala Image API to clojure

kedarbellare:
## Description ##

- This PR is for issue https://github.com/apache/incubator-mxnet/issues/13089
- It ports the Image API recently added (https://github.com/apache/incubator-mxnet/pull/12995) in the scala package to clojure package. It provides the ability to read an image ndarray from file, decode an image ndarray from input stream, resize/crop/pad an image ndarray and lastly to convert a ndarray back into an image.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [X] Changes are complete (i.e. I finished coding on this PR)
- [X] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. (_**Needs better documentation**_)
- [X] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [X] Clojure `image` module

PR #13207, opened 2018-11-09
[Scala][Clojure] Fix problem with some OSX not handling the cast on imDecode

gigasquid:
## Description ##

My OSX system was having a problem running the Scala function in the Image package `imDecode`.
When running the Scala unit tests it would error out with:

```
ImageSuite:
- Test load image
[19:13:09] src/io/image_io.cc:147: Decoding failed. Invalid image file.
2018-11-07 19:13:09,666 [ScalaTest-main-running-ImageSuite] [org.apache.mxnet.ImageSuite] [INFO] - OpenCV load image with shape: ()
2018-11-07 19:13:09,671 [ScalaTest-main-running-ImageSuite] [org.apache.mxnet.ImageSuite] [INFO] - OpenCV resize image with shape: (224,224,3)
- Test load image from Socket *** FAILED ***
  java.lang.IllegalArgumentException: requirement failed: image shape not Match!
  at scala.Predef$.require(Predef.scala:224)
  at org.apache.mxnet.ImageSuite$$anonfun$3.apply$mcV$sp(ImageSuite.scala:69)
  at org.apache.mxnet.ImageSuite$$anonfun$3.apply(ImageSuite.scala:64)
  at org.apache.mxnet.ImageSuite$$anonfun$3.apply(ImageSuite.scala:64)
```

After some investigation with @lanking520 , I found that adding an explicit conversion of the byte to an int before creating the NDArray solved the issue.

### Essentials ###

### Changes ###

Small tweak to the Scala function in `imDecode`

This is also relevant to https://github.com/apache/incubator-mxnet/pull/13107

### DIFF
--- a/scala-package/core/src/main/scala/org/apache/mxnet/Image.scala
+++ b/scala-package/core/src/main/scala/org/apache/mxnet/Image.scala
@@ -42,7 +42,7 @@ object Image {
   def imDecode(buf: Array[Byte], flag: Int,
                to_rgb: Boolean,
                out: Option[NDArray]): NDArray = {
-    val nd = NDArray.array(buf.map(_.toFloat), Shape(buf.length))
+    val nd = NDArray.array(buf.map( x => (x & 0xFF).toFloat), Shape(buf.length))
     val byteND = NDArray.api.cast(nd, "uint8")
     val args : ListBuffer[Any] = ListBuffer()
     val map : mutable.Map[String, Any] = mutable.Map()


marcoabreu:
I can remember quite a few issues where people reported the same error about apparently invalid image files. I'm not sure whether it was related to Scala, but this is a great catch! Thanks Carin.

lanking520:
@marcoabreu @gigasquid  I do think there might be an issue with a different operating system. The things that make me suspicious is that: In theory, a byte & 0xFF would only bring the positive result. The signed bit has been disabled for that. However, when I feed the positive float data and cast them as `uint8`. The converted float data from `uint8` contains negative values:

The flow is like this:

`byte` AND `0xFF` -> All positive float

 float cast `uint8` -> uint8

 uint8 cast float -> Contains negative value???

@frankfliu since he raised this concern during our offline conversation

gigasquid:
@lanking520 - I'm thinking that might be just a printing display thing? I can't see how a uint8 ndarray could have negative values. Do you know of a way to test that?

nswamy:
I haven't followed the entire conversation, but just a reminder that mxnet Scala supports only float32

gigasquid:
I'm going to go ahead and merge this since it seems like it fixes the issue. We'll just have to keep an eye open if there are any other problems in this area.

nswamy:
@marcoabreu this feature did not exist in 1.3.0, so No.

PR #13209, opened 2018-11-09
Allow dataloader iterator to be reused

zhreshold:
## Description ##
In gluon, DataLoader with very large ``num_workers`` can be slow at iteration initialization step, because forking a large number of processes can be expensive. 
This PR allow user to cache and reset to reuse the iterator by:

Updated:

```python

data_loader = mx.gluon.DataLoader(dataset, batch_size, num_workers=100)
it = iter(data_loader) # explicitly cache the iterator

for i in range(num_epochs):
    for batch in it:
        # training code
```

This PR don't affect the existing usage where each iterator is initialized and destroyed on-the-fly:
```python
data_loader = mx.gluon.DataLoader(dataset, batch_size, num_workers=100)
for i in range(epochs):
    for batch in data_loader:
        # training code
```
  

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/python/mxnet/gluon/data/dataloader.py
+++ b/python/mxnet/gluon/data/dataloader.py
@@ -268,9 +268,7 @@ class _MultiWorkerIter(object):
         self._fetcher.daemon = True
         self._fetcher.start()
 
-        # pre-fetch
-        for _ in range(2 * self._num_workers):
-            self._push_next()
+        self._reset()
 
     def __len__(self):
         return len(self._batch_sampler)
@@ -278,8 +276,9 @@ class _MultiWorkerIter(object):
     def __del__(self):
         self.shutdown()
 
-    def reset(self):
+    def _reset(self):
         """Reset iterator with multiprocessing workers alive."""
+        assert not self._shutdown, "call reset after shutdown is forbidden"
         # clear key queue
         removed_idx = set()
         while True:
@@ -333,6 +332,7 @@ class _MultiWorkerIter(object):
         return self.__next__()
 
     def __iter__(self):
+        self._reset()
         return self
 
     def shutdown(self):
--- a/python/mxnet/gluon/data/dataloader.py
+++ b/python/mxnet/gluon/data/dataloader.py
@@ -41,7 +41,6 @@ except ImportError:
 
 from . import sampler as _sampler
 from ... import nd, context
-from ...recordio import MXRecordIO
 
 if sys.platform == 'darwin' or sys.platform == 'win32':
     def rebuild_ndarray(*args):
@@ -164,29 +163,9 @@ def _as_in_context(data, ctx):
         return [_as_in_context(d, ctx) for d in data]
     return data
 
-def _recursive_fork_recordio(obj, depth, max_depth=1000):
-    """Recursively find instance of MXRecordIO and reset file handler.
-    This is required for MXRecordIO which holds a C pointer to a opened file after fork.
-    """
-    if depth >= max_depth:
-        return
-    if isinstance(obj, MXRecordIO):
-        obj.close()
-        obj.open()  # re-obtain file hanlder in new process
-    elif (hasattr(obj, '__dict__')):
-        for _, v in obj.__dict__.items():
-            _recursive_fork_recordio(v, depth + 1, max_depth)
 
 def worker_loop(dataset, key_queue, data_queue, batchify_fn):
     """Worker loop for multiprocessing DataLoader."""
-    # re-fork a new recordio handler in new process if applicable
-    # for a dataset with transform function, the depth of MXRecordIO is 1
-    # for a lazy transformer, the depth is 2
-    # for a user defined transformer, the depth is unknown, try a reasonable depth
-    # limit = sys.getrecursionlimit()
-    # max_recursion_depth = min(limit - 5, max(10, limit // 2))
-    # _recursive_fork_recordio(dataset, 0, max_recursion_depth)
-
     while True:
         idx, samples = key_queue.get()
         if idx is None:
--- a/python/mxnet/gluon/data/dataloader.py
+++ b/python/mxnet/gluon/data/dataloader.py
@@ -212,7 +212,28 @@ def fetcher_loop(data_queue, data_buffer, pin_memory=False, data_buffer_lock=Non
 
 
 class _MultiWorkerIter(object):
-    """Interal multi-worker iterator for DataLoader."""
+    """Interal multi-worker iterator for DataLoader.
+    It allow reset() to reuse iterator with all workers alive.
+
+    Parameters
+    ----------
+    num_workers : int, default 0
+        The number of multiprocessing workers to use for data preprocessing.
+    dataset : Dataset
+        Source dataset. Note that numpy and mxnet arrays can be directly used
+        as a Dataset.
+    batchify_fn : callable
+        Callback function to allow users to specify how to merge samples
+        into a batch.
+    batch_sampler : Sampler
+        A sampler that returns mini-batches. Do not specify batch_size,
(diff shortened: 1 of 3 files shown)


zhreshold:
@hetong007 @szha for review.

python/mxnet/gluon/data/dataloader.py
```diff
@@ -244,6 +270,35 @@ def __len__(self):
     def __del__(self):
         self.shutdown()
 
+    def reset(self):
```
szha: assert not shutdown

python/mxnet/gluon/data/dataloader.py
```diff
@@ -281,6 +335,58 @@ def shutdown(self):
...
+        If ``True``, the dataloader will copy NDArrays into pinned memory
+        before returning them. Copying from CPU pinned memory to GPU is faster
+        than from normal CPU memory.
+
+    """
+    def __init__(self, dataset, batchify_fn, batch_sampler, pin_memory=False):
+        self._dataset = dataset
+        self._batchify_fn = batchify_fn
+        self._batch_sampler = batch_sampler
+        self._iter = iter(self._batch_sampler)
```
szha: call reset here in case more states need to be maintained in reset()

python/mxnet/gluon/data/dataloader.py
leezu: Do we need to introduce a new reset method? Just moving this functionality into `__iter__` of this class would be cleaner I think

anirudhacharya:

leezu:
Thanks for your work @zhreshold . I think you can just remove the `reset()` method. The logic should instead be handled in `_MultiWorkerIter.__iter__` and `_SameProcessIter.__iter__`. That would simplify the use case of your example:

```
data_loader = mx.gluon.DataLoader(dataset, batch_size, num_workers=100)
it = iter(data_loader) # explicitly cache the iterator

for i in range(num_epochs):
    # it.reset() # Deleted. Not necessary anymore.
    for batch in it: # This calls it.__iter__ which resets the iterator
        # training code
```

Still does not affect the existing use-case:

```
data_loader = mx.gluon.DataLoader(dataset, batch_size, num_workers=100)
for i in range(epochs):
    for batch in data_loader:
        # training code
```

However, I wonder if there is a cleaner way to represent the process pool. The current solution may be a bit confusing to users. It is not very intuitive that the iterator must be cached for good performance. Explicitly specifying the resource (ie. process pool) used for the execution may help to clarify? Are you familiar with the `concurrent.futures` pools? I feel like we are reimplementing a slightly different way of this Python standard libray API here. It may be better to find abstractions that just rely on the Python standard API. It may also be easier to understand for new users? For example, the concept of a executor may be useful. For multiprocessing, there is https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor

The last paragraph is not a critic of your PR but I want to raise the point that a overhaul of the API may help improve clarity. We can introduce a dependency on the [backport of `concurrent.futures`](https://www.google.com/search?q=concurrent.futures+python+2&ie=utf-8&oe=utf-8&client=firefox-b-ab) to Py2 for Python 2 users (Python 2 is EOL soon). @szha

zhreshold:
@leezu Good point, using __iter__ is definitely much cleaner in such use cases. Let me go back to the original concerns of pool, and will update later.

leezu:
@zhreshold ok, great. Btw, I noticed you missed updating `_SameProcessIter` regarding `reset / __iter__` in your last commit. Thanks for your work on this, I think addressing speed concerns in the Gluon API is very important for a good user experience.

stu1130:
@zhreshold could you address the CI failure
@szha ping for the second round review

leezu:
@stu1130 I think @zhreshold still plans to make some changes (my understanding of https://github.com/apache/incubator-mxnet/pull/13209#issuecomment-437983209) If so, no need for review yet

zhreshold:
@leezu I currently do have an option to utilize process pools, and it should work elegantly without caching iterators explicitly. 
However, due to the backward compatibility issue of future module mentioned here https://pypi.org/project/futures/:
*"The ProcessPoolExecutor class has known (unfixable) problems on Python 2 and should not be relied on for mission critical work."* 

Therefore I am not including the update in this PR. Besides, the logic to protect RecordIO handles is moved to `recordio.py`. This PR can improve initialization time significantly (more than 5x) on average according to local benchmark.

Please have a second look @leezu @szha @hetong007

python/mxnet/gluon/data/dataloader.py
```diff
@@ -363,13 +449,8 @@ def __init__(self, dataset, batch_size=None, shuffle=False, sampler=None,
 
     def __iter__(self):
```
leezu: Would it make sense to add documentation about the API of returned `_SameProcessIter` / `_MultiWorkerIter`? Ie. that they will reset when `__iter__` is called on them.

python/mxnet/gluon/data/dataloader.py
```diff
@@ -203,7 +187,29 @@ def fetcher_loop(data_queue, data_buffer, pin_memory=False):
...
+        Callback function to allow users to specify how to merge samples
+        into a batch.
+    batch_sampler : Sampler
+        A sampler that returns mini-batches. Do not specify batch_size,
+        shuffle, sampler, and last_batch if batch_sampler is specified.
+    pin_memory : boolean, default False
+        If ``True``, the dataloader will copy NDArrays into pinned memory
+        before returning them. Copying from CPU pinned memory to GPU is faster
+        than from normal CPU memory.
+
```
eric-haibin-lin: Can we also document `worker_fn`?

python/mxnet/gluon/data/dataloader.py
```diff
@@ -267,20 +316,128 @@ def __next__(self):
                 return batch
 
     def next(self):
+        """Compatible portal for __next__ in python2.
+
+        Returns
+        -------
+        type
+            Description of returned object.
```
eric-haibin-lin: need to update doc

python/mxnet/gluon/data/dataloader.py
```diff
@@ -267,20 +316,128 @@ def __next__(self):
...
+        alive, thus save re-init time of forking processes.
+
+        Returns
+        -------
+        iterator
+            Iterator of self.
+
+        """
+        self._reset()
         return self
```
eric-haibin-lin: What happens if users used the dataloader and created two iterators? Does the 2nd iterator reset the state for the first one that is running? For example:
```
>>> a = [1,2,3]
>>> for x in a:
...     for y in a:
...         print(x,y)
...
(1, 1)
(1, 2)
(1, 3)
(2, 1)
(2, 2)
(2, 3)
(3, 1)
(3, 2)
(3, 3)
```

zhreshold: iterators from the same dataloader are two separate instances, they don't interfere with each other, however, for explicitly cached iterator, it will break

```
a = [1,2,3]
it = iter(a)
for x in it:
    for y in it:
        # this will not work

File: python/mxnet/gluon/data/dataloader.py (hunk 5 above)
Comment (zhreshold) [reply python/mxnet/gluon/data/dataloader.py]: will do

File: python/mxnet/gluon/data/dataloader.py [hunk 7]
```diff
@@ -203,7 +187,29 @@ def fetcher_loop(data_queue, data_buffer, pin_memory=False):
 
 
 class _MultiWorkerIter(object):
-    """Interal multi-worker iterator for DataLoader."""
+    """Interal multi-worker iterator for DataLoader.
```
anirudhacharya: Interal -> Internal

zhreshold:
Close this in favor of better solution: #13447

Issue #10923, opened 2018-05-13
some operators don't have unit tests.


zheng-da:
Our Python unit tests don't cover all operators. Recently, I found softmax activation was never tested and the test has been added in https://github.com/apache/incubator-mxnet/pull/10918. Currently, the backward computation of norm doesn't have GPU tests. I'm afraid more operators don't have unit tests.

eric-haibin-lin:
Also https://github.com/apache/incubator-mxnet/issues/10868

eric-haibin-lin:
We need to make a list tracking missing unit tests

anirudhacharya:
- [ ] check_numeric_gradient for FFT operator

anirudhacharya:
`initializer.py` has very less test coverage
- [ ] Orthogonal Initializer
- [ ] Bilinear Initializer
- [ ] MSRAPrelu Initializer
- [ ] Mixed Initializer

anirudhacharya:
Missing Gradient Checks - https://github.com/apache/incubator-mxnet/issues/13254

apeforest:
@mxnet-label-bot add [Good First Time]

anirudhacharya:
Missing Optimizer tests -
 - [ ] Large batch BSGD
 - [ ] Delay compensation asynchronous SGD
 - [ ] SGLD
 - [ ] AdaDelta
 - [ ] Adamax
 - [ ] Nadam

hanzhaogang:
What skills and knowledge are required to write the test code? How do we test operators? Checked the `test_subgraph_op.py` whihc I think should be a test for the subgraph operator in the `incubator-mxnet/tests/python/unittest/` package but still have no idea how to contribute.

cchung100m:
> Missing Optimizer tests -
> 
> * [ ]  Large batch BSGD
> * [ ]  Delay compensation asynchronous SGD
> * [x]  SGLD
> * [x]  AdaDelta
> * [x]  Adamax
> * [ ]  Nadam

Hi @anirudhacharya ,

It's been a while since the checked missing unit tests last year, and I checked Nadam has [test](https://github.com/apache/incubator-mxnet/blob/50495d7b948c813222597fb9ffaf45cfcea29d2e/tests/python/unittest/test_optimizer.py#L1061), I would appreciate if you can point me where can I find the latest list of missing unit tests. 

Many thanks,

anirudhacharya:
[...]

PR #13229, opened 2018-11-12
Add gauss err function operator

eric-haibin-lin:
## Description ##
Add error function operator (https://en.wikipedia.org/wiki/Error_function). 

@szha

### Essentials ###

### Changes ###

## Comments ##


anirudhacharya:

tests/python/unittest/test_operator.py
```diff
@@ -3496,6 +3496,10 @@ def test_special_functions_using_scipy():
     mathematical_core("gammaln", lambda x: mx.sym.gammaln(x), lambda x: scipy_special.gammaln(x),
                      lambda x: scipy_special.psi(x), 0.5, 0.5)
 
+    # erf
+    mathematical_core("erf", lambda x: mx.sym.erf(x), lambda x: scipy_special.erf(x),
```
anirudhacharya: the `mathematical_core` does not perform gradient checking. so these special functions are not completely tested for their gradient calculation. Is that okay?

tests/python/unittest/test_operator.py
```diff
@@ -3496,6 +3496,10 @@ def test_special_functions_using_scipy():
...
+                     lambda x: 2.0 / math.sqrt(math.pi) * math.exp(-(x ** 2)), 0.5, 0.5)
```
anirudhacharya: also in this method in line 3488 when importing scipy fails, we are skipping these tests. Shouldn't we be failing them by raising an error/exception?

szha: scipy is optional dep and CI is set up to test both with and without scipy.

szha: the same applies to sparse support

anirudhacharya: ok

tests/python/unittest/test_operator.py
szha: https://github.com/apache/incubator-mxnet/blob/65736917a63d8d04ee87f653ec2ffcf2be595dd1/tests/python/unittest/test_operator.py#L3480

anirudhacharya: yes, but `npout_grad` is calculated from the `backward_numpy_call` method that we pass to this function.
https://github.com/apache/incubator-mxnet/blob/65736917a63d8d04ee87f653ec2ffcf2be595dd1/tests/python/unittest/test_operator.py#L3473-L3474

with the above statement, the backward pass is getting tested. But that is not the same thing as doing "gradient checking", right? which will test for the gradient of the operator without using the `backward_numpy_call`

anirudhacharya: this is more a question than a review comment.

eric-haibin-lin: Good question. I think the numerical grad test cases for all mathematical_core ops are missing. Since there are many ops, we'd better track that as a separate task.

eric-haibin-lin: https://github.com/apache/incubator-mxnet/issues/13254

anirudhacharya: Thanks, I will also update it here - https://github.com/apache/incubator-mxnet/issues/10923

PR #11047, opened 2018-05-24
[MXNET-33] Enhance mkldnn pooling to support full convention

TaoLv:
## Description ##
Follow-up for https://github.com/apache/incubator-mxnet/pull/10021

Then gluon/squeezenet can benefit from this PR with below performance improvement.

Squeezenet1.1 inference on SKX8180:

BS | Before | After | Speed-up
-- | -- | -- | --|
1 | 114.891003 | 265.844219 | 2.313881958
2 | 155.3841 | 429.361312 | 2.76322553
4 | 182.232015 | 687.565177 | 3.773020767
8 | 203.664481 | 845.273893 | 4.15032552
16 | 221.299678 | 1110.591241 | 5.018494609
32 | 232.902484 | 1497.973739 | 6.431763686
64 | 232.521016 | 1464.001852 | 6.296213036
128 | 234.921866 | 1285.150863 | 5.470545952
256 | 243.123314 | 1152.503337 | 4.740406496

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #13260, opened 2018-11-14
Support full convention in quantized pooling

TaoLv:
## Description ##
1. Fix quantized pooling to support full convention
2. Enable quantized pooling in INT8 SqueezeNet example

### Essentials ###

### Changes ###

## Comments ##


TaoLv:
This PR depends on #11047 to past the new unit test.

kalyc:
@TaoLv thanks for your contribution - could you please take a look at the failed CI and re-trigger it?

kalyc:

TaoLv:
@kalyc Code was re-based and passed the CI. It's ready for review now.

pengzhao-intel:
@zheng-da @reminisce @szha @eric-haibin-lin @apeforest 
This is an important enhancement for quantization flow of MKL-DNN in r1.4.
Please help take a review

kalyc:

example/quantization/imagenet_gen_qsym_mkldnn.py
```diff
@@ -225,11 +225,11 @@ def save_params(fname, arg_params, aux_params, logger=None):
         rgb_mean = '123.68,116.779,103.939'
         rgb_std = '58.393, 57.12, 57.375'
         calib_layer = lambda name: name.endswith('_output')
-        excluded_sym_names += ['squeezenet0_flatten0_flatten0',
-                               'squeezenet0_pool0_fwd',
-                               'squeezenet0_pool1_fwd',
-                               'squeezenet0_pool2_fwd',
-                               'squeezenet0_pool3_fwd']
+        excluded_sym_names += ['squeezenet0_flatten0_flatten0']
+        #                       'squeezenet0_pool0_fwd',
```
apeforest: Remove commented lines

src/operator/quantization/quantized_pooling.cc
```diff
@@ -52,17 +52,30 @@ bool QuantizedPoolingShape(const nnvm::NodeAttrs& attrs,
...
   oshape[C] = dshape[C];
   if (param.global_pool) {
     oshape[H] = 1;
     oshape[W] = 1;
   } else {
-    oshape[H] = 1 + (dshape[H] + 2 * param.pad[0] - param.kernel[0]) /
-        param.stride[0];
-    oshape[W] = 1 + (dshape[W] + 2 * param.pad[1] - param.kernel[1]) /
-        param.stride[1];
+     if (param.pooling_convention == pool_enum::kValid) {
```
apeforest: nit: indentation mismatch

src/operator/quantization/quantized_pooling.cc
```diff
@@ -52,17 +52,30 @@ bool QuantizedPoolingShape(const nnvm::NodeAttrs& attrs,
...
+      oshape[2] = 1 +
```
apeforest: Use H/W instead of 2 and 3.

src/operator/quantization/quantized_pooling.cc
```diff
@@ -52,17 +52,30 @@ bool QuantizedPoolingShape(const nnvm::NodeAttrs& attrs,
...
+                  (dshape[2] + 2 * param.pad[0] - param.kernel[0]) /
+                      param.stride[0];
+      oshape[3] = 1 +
+                  (dshape[3] + 2 * param.pad[1] - param.kernel[1]) /
+                      param.stride[1];
+    } else {
+      oshape[2] = 1 + static_cast<int>(std::ceil(
```
apeforest: It might be good to write a function to do such calculation, with a flag pool_enum to distinguish the way of calculating the shape. kValid: using floor, otherwise, using ceil.

tests/python/quantization/test_quantization.py
```diff
@@ -214,7 +214,7 @@ def check_quantized_conv(data_shape, kernel, num_filter, pad, stride, no_bias, q
 
 @with_seed()
 def test_quantized_pooling():
-    def check_quantized_pooling(data_shape, kernel, pool_type, pad, stride, global_pool, qdtype):
+    def check_quantized_pooling(data_shape, kernel, pool_type, pad, stride, global_pool, qdtype, convention):
```
apeforest: I suggest we keep this test to test the default behavior without passing in convention. Add another test to pass in convention as an extra argument.

src/operator/quantization/quantized_pooling.cc
TaoLv: You are probably right. But this code snippet was copied from FP32 pooling infer shape here: https://github.com/apache/incubator-mxnet/blob/master/src/operator/nn/pooling.cc#L163

I try not to change the code of FP32 pooling in this PR and keep the code align between quantized pooling and FP32 pooling. So I would keep it as is and refactor them in another PR in the future. What do you think?

tests/python/quantization/test_quantization.py
TaoLv: Having another check function like `check_quantized_pooling_with_convention` may need much redundant code here. Do you think we can give the argument `convention` a default value here?

yuxihu: What is the default convention before this PR? By adding a new argument, you are testing "valid" and "full". Has the default case been covered with your change?

tests/python/quantization/test_quantization.py
```diff
@@ -244,7 +245,8 @@ def check_quantized_pooling(data_shape, kernel, pool_type, pad, stride, global_p
         quantized_pooling = mx.sym.contrib.quantized_pooling(data=qdata, min_data=min_data,
                                                                 max_data=max_data, kernel=kernel,
```
yuxihu: May be also good to fix the alignment from L246-L249.

tests/python/quantization/test_quantization.py
```diff
@@ -244,7 +245,8 @@ def check_quantized_pooling(data_shape, kernel, pool_type, pad, stride, global_p
...
                                                                 pad=pad, stride=stride, pool_type=pool_type,
-                                                                global_pool=global_pool)
+                                                                global_pool=global_pool,
+                                                                pooling_convention=convention)
```
yuxihu: same question here: what was the default value for pooling_convention? Make sure the current case is covered.

tests/python/quantization/test_quantization.py
apeforest: It's okay to only keep this test method. But As @yuxihu mentioned, you may want to test the backward compatibility of the operator without passing this argument.

TaoLv: The default convention for mxnet pooling is "valid".
https://github.com/apache/incubator-mxnet/blob/master/src/operator/nn/pooling-inl.h#L74

I will set the default value for argument "convention" to "valid" here and revert the change for L264-L267. So the behavior of these 4 test cases will be as before.

TaoLv: Sure.

tests/python/quantization/test_quantization.py
TaoLv: Fixed.

tests/python/quantization/test_quantization.py
TaoLv: Should be "valid". https://mxnet.incubator.apache.org/api/python/ndarray/contrib.html#mxnet.ndarray.contrib.quantized_pooling

TaoLv:
@marcoabreu I don't have any idea about the CI failure. Could you help to take a look? Thanks.

PR #13276, opened 2018-11-15
enabling test_dropout after fixing flaky issue

### LINKED ISSUE #12314: flaky test: test_operator.test_dropout
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/PR-12311/1/pipeline
```

FAIL: test_operator.test_dropout

  File "/usr/local/lib/python3.5/dist-packages/nose/case.py", line 198, in runTest

    self.test(*self.arg)

  File "/work/mxnet/tests/python/unittest/common.py", line 172, in test_new

    orig_test(*args, **kwargs)

  File "/work/mxnet/tests/python/unittest/test_operator.py", line 5844, in test_dropout

    check_dropout_ratio(0.0, shape)

  File "/work/mxnet/tests/python/unittest/test_operator.py", line 5788, in check_dropout_ratio

    assert exe.outputs[0].asnumpy().min() == min_value

AssertionError: 

common: INFO: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=119185068 to reproduce.

```

mseth10:
## Description ##
Enabled unit test for dropout. Fixed flaky behavior on MKLDNN CPU. Fixes #12314 .

### Essentials ###

### Changes ###
- [x] BernoulliGenerate function modified to pass constant positive seed to vslNewStream function of MKL across different threads.

## Comments ##
- Found this error while going through another code stub using vslNewStream function https://github.com/intel/mklnn/blob/master/src/Random.c#L32


src/operator/nn/dropout-inl.h
```diff
@@ -82,7 +82,7 @@ class DropoutOp {
   static void BernoulliGenerate(common::random::RandGenerator<cpu, DType> gen,
                                 int n, double p, int* r) {
     typename RandGenerator<xpu, DType>::Impl genImpl(&gen, 1);
-    const int seed = 17 + genImpl.rand() % 4096;  // NOLINT(runtime/threadsafe_fn)
+    const int seed = 17 + abs(genImpl.rand() % 4096);
     const int nthr = engine::OpenMP::Get()->GetRecommendedOMPThreadCount();
 #pragma omp parallel num_threads(nthr)
```
pengzhao-intel: I think it will be better to change this parallel approach to "#pragma omp paralle for" so it will be more readable than OMP section parallel.
Do you mind to change it?

pengzhao-intel:
@mseth10 Thanks for the fix :)

>  In the present code, different seed is passed to vslNewStream for different threads. Making the seeds constant fixes the flaky issue

Could you elaborate how do you make the seeds constant?

src/operator/nn/dropout-inl.h
```diff
@@ -92,7 +92,7 @@ class DropoutOp {
       const int my_amount = std::min(my_offset + avg_amount, n) - my_offset;
       if (my_amount > 0) {
         VSLStreamStatePtr stream;
-        vslNewStream(&stream, VSL_BRNG_MCG31, seed + my_offset);
+        vslNewStream(&stream, VSL_BRNG_MCG31, seed);
```
mseth10: @pengzhao-intel vslNewStream was passed (seed + my_offset) which is variable for different threads as my_offset depends on thread number. I have removed my_offset, so whatever seed value is generated, it's used to generate stream for each thread.

src/operator/nn/dropout-inl.h
mseth10: @pengzhao-intel Thanks for your feedback. I don't see why "#pragma omp parallel for" would be more readable in this case as there is no for loop involved. The job for each thread is defined based on its thread number. Introducing a for loop with "#pragma omp parallel for" would be redundant. Please correct me if I've misunderstood it.

mseth10: @apeforest vslNewStream expects a non-negative integer as seed by definition. genImpl.rand() generates negative numbers too, which sometimes leads to negative seed values being passed to vslNewStream. abs() helps keep that in check.

src/operator/nn/dropout-inl.h
mseth10: @apeforest By removing my_offset from vslNewStream and making the seed same for all threads, we are ensuring that the same stream is generated for each thread.
For each thread, my_offset acts an index for vslSkipAheadStream and viRngBernoulli. It tells vslSkipAheadStream which section of stream to use and viRngBernoulli which section of vector to populate.

src/operator/nn/dropout-inl.h
apeforest: Thanks for the explanation. In this case I would suggest we add an assert here to make sure seed is nonnegative

kalyc:
Hi @mseth10 thanks for your contribution!
Could you please add "Fixes <issue #>" in the PR description - so when this PR is closed, the issue will be closed automatically?

kalyc:

apeforest:
Please also add your name to CONTRIBUTORS.md if you have not already done so.

src/operator/nn/dropout-inl.h
mseth10: Added a check to make sure seed is non-negative.

pengzhao-intel:
@mseth10 thanks for the nice explanation in this PR.  Great work!

Issue #12732, opened 2018-10-03
TODO list for supporting dynamic shape


zheng-da:
MXNet doesn't support operators that don't allow static shape inference, such as unique and boolean_index. One possible solution is proposed in https://cwiki.apache.org/confluence/display/MXNET/Dynamic+shape

This task can be broken into multiple steps:
* support such an operator in the imperative execution. A possible modification in the imperative execution is shown here: https://github.com/apache/incubator-mxnet/pull/12400
* support such an operator in CachedOp. This step requires writing a new executor that runs operators in a computation graph one by one without shape inference and memory planning. We can start with executing operators synchronously, but eventually, we should execute them asynchronously.
* an optimization is to use the subgraph API to split the computation graph into subgraphs so that a subgraph either contain only operators allowing static shape inference or operators with dynamic shape.

vrakesh:
@zheng-da thanks for the detailed proposal on supporting dynamic shapes

@mxnet-label-bot [Feature request]

chinakook:
I think that a good solution to support **pseudo**-dynamic (in most cases it's enough) shape is to support infer_shape_partial in HybridBlock.

PR #12400, opened 2018-08-29
[WIP] support dynamic shape in the imperative mode.

zheng-da:
## Description ##
This is the first step of supporting dynamic shape in mxnet. As the first step, the PR only supports dynamic shape in the imperative mode. It creates a boolean_mask operator to test dynamic shape.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #13283, opened 2018-11-15
[MXNET-1215] Allow dynamic shape exists in imperative mode

junrushao1994:
## Description ##
It is the PR I took over from @zheng-da at #12400. All credits go to Da Zheng.

This PR relaxes the constraint on NDArrays that shape used to be pre-determined. For unittesting, this PR introduces an operator called `BooleanMask` in contrib, which is a practical use case that actually produces dynamic shape. Note that the support for boolean mask is very experimental, which only allows 2-d inputs, and 1-d mask, because it serves only the functionality of testing. We could improve its better in future PRs. 

This is an initial step for the roadmap #12732.

### Essentials ###

### Changes ###
- [x] Introducing new methods in NDArray allowing delayed assignment of their shapes.
- [x] Introducing boolean mask operator, which we believe will be useful in the future.
- [x] Add unit test.

## Comments ##
No comments.

## TODO ##
- [x] Refactor the boolean index operator
- [x] Write backward pass of the boolean index operator
- [x] Test coverage

### DIFF
--- a/include/mxnet/ndarray.h
+++ b/include/mxnet/ndarray.h
@@ -109,7 +109,7 @@ class NDArray {
    * \param ctx context of NDArray
    * \param dtype data type of this ndarray
    */
-  NDArray(Context ctx, int dtype = mshadow::default_type_flag) {
+  explicit NDArray(Context ctx, int dtype = mshadow::default_type_flag) {
     ptr_ = std::make_shared<Chunk>(TShape(mshadow::Shape1(0)), ctx, true, dtype);
     dtype_ = dtype;
     storage_type_ = kDefaultStorage;
@@ -984,7 +984,7 @@ class NDArray {
 #endif
       }
     }
-    /*! \brief initialize the shape and dtype for this Chunk, assuming it is not initialized before. */
+    /*! \brief initialize the shape and dtype, assuming it is not initialized before. */
     void Init(const TShape &shape, int dtype) {
       auto size = shape.Size();
       storage_shape = shape;
--- a/include/mxnet/ndarray.h
+++ b/include/mxnet/ndarray.h
@@ -103,7 +103,18 @@ class NDArray {
           bool delay_alloc = true, int dtype = mshadow::default_type_flag,
           std::vector<int> aux_types = {}, std::vector<TShape> aux_shapes = {},
           TShape storage_shape = TShape(mshadow::Shape1(0)));
-
+  /*!
+   * \brief constructs a new dynamic NDArray whose shape is unknown,
+   *        hence the NDArray is inherently lazily created
+   * \param ctx context of NDArray
+   * \param dtype data type of this ndarray
+   */
+  NDArray(Context ctx, int dtype = mshadow::default_type_flag) {
+    ptr_ = std::make_shared<Chunk>(TShape(mshadow::Shape1(0)), ctx, true, dtype);
+    dtype_ = dtype;
+    storage_type_ = kDefaultStorage;
+    entry_ = {nullptr, 0, 0};
+  }
   /*!
    * \brief constructing a static NDArray that shares data with TBlob
    *  Use with caution: allocate ONLY ONE NDArray for each TBlob,
@@ -157,7 +168,20 @@ class NDArray {
       : ptr_(std::make_shared<Chunk>(stype, data, aux_data, dev_id)), shape_(shape),
         dtype_(data.type_flag_), storage_type_(stype), entry_({nullptr, 0, 0}) {
   }
-
+  /*!
+   * \brief initialize the NDArray, assuming it is not assigned a meaningful shape before
+   * \param shape the shape of the NDArray
+   */
+  void Init(const TShape &shape) {
+    ptr_->Init(shape, this->dtype_);
+    this->shape_ = shape;
+  }
+  /*!
+   * \brief set the correct shape of NDArray directly from the storage_shape of its own chunk.
+   */
+  void SetShapeFromChunk() {
+    shape_ = ptr_->storage_shape;
+  }
   /*
    * This indicates whether an array is a view of another array (created by
    * reshape or slice). If an array is a view and the the data is stored in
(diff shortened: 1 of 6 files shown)


src/operator/contrib/boolean_mask.cc
```diff
@@ -0,0 +1,59 @@
...
+  CHECK_EQ(in_attrs->size(), 2);
+  CHECK_EQ(out_attrs->size(), 1);
+  for (size_t i = 0; i < out_attrs->size(); i++)
+    out_attrs->at(i) = kDefaultStorage;
+  *dispatch_mode = DispatchMode::kFComputeEx;
+  return true;
+}
+
+NNVM_REGISTER_OP(_contrib_BooleanMask)
+.describe(R"code(
```
ZhennanQin: Any description for this op?

src/operator/contrib/boolean_mask-inl.h
```diff
@@ -0,0 +1,91 @@
...
+
+struct BooleanMaskParam : public dmlc::Parameter<BooleanMaskParam> {
+  int axis;
+  DMLC_DECLARE_PARAMETER(BooleanMaskParam) {
+    DMLC_DECLARE_FIELD(axis).set_default(0)
+    .describe("An integer that represents the axis in NDArray to mask from.");
+  }
+};
+
+template <typename xpu>
```
ZhennanQin: This is a pure CPU specific implementation, this template keyword will mislead other developer this could be used for GPU as well.

src/imperative/imperative_utils.h
```diff
@@ -117,11 +117,14 @@ inline void SetShapeType(const Context& ctx,
   for (auto& i : outputs) {
     out_shapes.push_back(i->shape());
   }
-  CHECK(infershape.count(attrs.op))
-    << "Operator " << attrs.op->name << " is missing FInferShape attribute";
-  CHECK(infershape[attrs.op](attrs, &in_shapes, &out_shapes));
-  CHECK_EQ(out_shapes.size(), outputs.size());
-
+  const bool dynamic_shape_exists = [&]() {
```
ZhennanQin: Not a good name here. The major job here is shape inference, and check if dynamic shape exist.

include/mxnet/ndarray.h
```diff
@@ -103,7 +103,18 @@ class NDArray {
...
           TShape storage_shape = TShape(mshadow::Shape1(0)));
-
+  /*!
+   * \brief constructs a new dynamic NDArray whose shape is unknown,
+   *        hence the NDArray is inherently lazily created
+   * \param ctx context of NDArray
+   * \param dtype data type of this ndarray
+   */
+  explicit NDArray(Context ctx, int dtype = mshadow::default_type_flag) {
+    ptr_ = std::make_shared<Chunk>(TShape(mshadow::Shape1(0)), ctx, true, dtype);
```
ZhennanQin: Why not lazily create Chunk when NDArray::Init() is called? Then we don't need to add Chunk::Init() function.

src/imperative/imperative_utils.h
junrushao1994: @ZhennanQin Thanks for the comment! Could you kindly suggest me a better name?

src/operator/contrib/boolean_mask-inl.h
junrushao1994: True, I am going to refactor this piece of code :-)

src/imperative/imperative_utils.h
ZhennanQin: Why not moving shape inference out of this lambda function? Actually, I don't see any reason that using lambda function here.

junrushao1994: @ZhennanQin personal coding style though: in principle, having stuff defined and constant without uninitialized value helps reduce bugs. Meanwhile, compiler could optimize out the lambda function, so it doesn't introduce any overhead. But I admit that it does not look that straight-forward. So, if you don't like it, I could definitely change it back.

kalyc:
@junrushao1994 please take a look at the CI failure and re-trigger CI

include/mxnet/ndarray.h
ZhennanQin: Because TShape(mshadow::Shape1(0)) doesn't mean no shape. It may confuse other developer, and also cause Chunk shape mismatch with NDArray, which may have potential risk.

zheng-da: The main reason we want to create a chunk here is to create the var in the chunk. Originally, I wanted to allow async execution. Now we only allow sync execution in the imperative mode. We probably don't need to create a chunk here.

junrushao1994:
CC: @szha @yidawang @zheng-da

junrushao1994:
Have updated according to @zheng-da's comments

include/mxnet/ndarray.h
ZhennanQin: Why this PR is merged without approve? I guess this comment isn't addressed. @junrushao1994 @zheng-da @szha

junrushao1994: @ZhennanQin Sorry I was too hurry. Personally, I think it is okay just to leave it 0-d for now. In the long term, we could support 0-d tensors in a more systematic approach.

zheng-da: @ZhennanQin we discussed it and think it's ok to use 0-dim tensor for now. Actually, we are using 0-dim shape when the shape is unknown in other places, so it should be fine.

zheng-da: it's true that we should have commented it here as well.

ZhennanQin:
The explanation is OK for me. But it's better to post your offline discussion result in this page before merge, then external reviewer can get involved. Thanks.

junrushao1994:
@ZhennanQin I agree with every word with your comment. It is completely my fault pushing this PR too hard, mainly because I don't want to block upcoming work. I will definitely behave next time.

PR #13285, opened 2018-11-15
Force APT cache update before executing install

marcoabreu:
Within our Docker caching, the APT cache is being persisted as well. Due to the inconsistent use of ``apt update``, this might lead to the Docker cache being refreshed partially while still using a stale APT cache.

By running ``apt update`` before every ``apt install``, we ensure the repository is always in the latest state if that layer has to be refreshed.

the ``|| true`` is added due to stability reasons. APT is not 100% available and we had problems because of that. By swallowing that error, we increase the tolerance. If the package is **really** missing during the installation stage (because it's not in the APT cache and updating the cache failed due to server errors), we will throw the error anyway. If we still have a good cache with the proper package present even though the update failed, this means that the update wasn't actually necessary and that we can still proceed. We already employed this mechanism in some places. Now we're making everything equal to harden all install scripts.

### DIFF
--- a/ci/docker/install/ubuntu_llvm.sh
+++ b/ci/docker/install/ubuntu_llvm.sh
@@ -23,4 +23,5 @@ echo deb-src http://apt.llvm.org/xenial/ llvm-toolchain-xenial-5.0 main\
      >> /etc/apt/sources.list.d/llvm.list
 
 wget -O - http://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
-apt-get update && apt-get install -y --force-yes llvm-5.0
+apt-get update || true
+apt-get install -y --force-yes llvm-5.0
--- a/ci/docker/install/ubuntu_scala.sh
+++ b/ci/docker/install/ubuntu_scala.sh
@@ -24,8 +24,9 @@ set -ex
 cd "$(dirname "$0")"
 # install libraries for mxnet's scala package on ubuntu
 echo 'Installing Scala...'
+apt-get update || true
 apt-get install -y software-properties-common
-apt-get update
+apt-get update || true
 apt-get install -y openjdk-8-jdk
 apt-get install -y openjdk-8-jre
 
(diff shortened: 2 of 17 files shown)


ci/docker/install/ubuntu_llvm.sh
```diff
@@ -23,4 +23,5 @@ echo deb-src http://apt.llvm.org/xenial/ llvm-toolchain-xenial-5.0 main\
      >> /etc/apt/sources.list.d/llvm.list
 
 wget -O - http://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
-apt-get update && apt-get install -y --force-yes llvm-5.0
\ No newline at end of file
+apt-get update || true
+apt-get install -y --force-yes llvm-5.0
```
KellenSunderland: Nit: NL.

ci/docker/install/ubuntu_scala.sh
```diff
@@ -24,8 +24,9 @@ set -ex
 cd "$(dirname "$0")"
 # install libraries for mxnet's scala package on ubuntu
 echo 'Installing Scala...'
+apt-get update || true
 apt-get install -y software-properties-common
-apt-get update
+apt-get update || true
```
KellenSunderland: Just double checking.  Is the reason we do two updates here that software-properities-common adds some new repos for apt?

marcoabreu: correct

aaronmarkham:
I was wondering why we do all of this updating and fetching all of the time. Why not build docker images that have the deps pre-installed with pinned and tested versions?
Most of the time I see site build failures is because some dep goes offline. This wouldn't be a problem if we used pre-built docker images. I imagine there's some major trade-off there, but what's worse?

marcoabreu:
We do use prebuilt docker images (that's what our Docker cache is for). The problem is that sometimes dependencies change (aka the install scripts get updated). In that case, the cache has to be regenerated. 

This updating and fetching will only happen if somebody modifies the install script. In 99% of the cases though, we will skip the execution of all these scripts entirely.

aaronmarkham:
@marcoabreu I refer to these kinds of issues where we appear to download the same dependencies every build, and then are at risk of failure when those dependencies are unreachable: http://jenkins.mxnet-ci.amazon-ml.com/job/restricted-website-build/685/console
```
Could not find artifact codox:codox:jar:0.10.3 in central (https://repo1.maven.org/maven2/)
Could not transfer artifact codox:codox:jar:0.10.3 from/to clojars (https://repo.clojars.org/): Read timed out
Could not find artifact codox:codox:jar:0.10.3 in staging (https://repository.apache.org/content/repositories/staging)
```

PR #13290, opened 2018-11-15
A few operators on graphs stored as CSR

zheng-da:
## Description ##
We can store a graph as an CSR matrix. In this case, the value in the matrix is the edge id starting from 0. These operators run on such special CSR matrices. This is joint work with @HyperZealot and @aksnzhy 

### Essentials ###

### Changes ###

## Comments ##


zheng-da:
@eric-haibin-lin could you please review this PR?

kalyc:

src/operator/contrib/dgl_graph.cc
```diff
@@ -0,0 +1,466 @@
...
+
+static void GetSubgraph(const NDArray &csr_arr, const NDArray &varr,
+                        const NDArray &sub_csr, const NDArray *old_eids) {
+  const TBlob &data = varr.data();
+  int64_t num_vertices = csr_arr.shape()[0];
+  const size_t len = varr.shape()[0];
+  const dgl_id_t *vid_data = data.dptr<dgl_id_t>();
+  HashTableChecker def_check(vid_data, len);
+  // check if varr is sorted.
+  std::is_sorted(vid_data, vid_data + len);
```
eric-haibin-lin: It returns True if it is actually sorted. But the return value is not used anywhere. Is this intended?

src/operator/contrib/dgl_graph.cc
```diff
@@ -0,0 +1,466 @@
...
+Example::
+  x = [[ 1, 0, 0 ],
+       [ 0, 2, 0 ],
+       [ 0, 0, 3 ]]
+  u = [ 0, 0, 1, 1, 2, 2 ]
+  v = [ 0, 1, 1, 2, 0, 2 ]
+  edge_id(x, u, v) = [ 1, -1, 2, -1, -1, 3 ]
+
+The storage type of ``edge_id`` output depends on storage types of inputs
+  - quadratic(csr, default, default) = default
```
eric-haibin-lin: quadratic -> edge_id

src/operator/contrib/dgl_graph.cc
```diff
@@ -0,0 +1,466 @@
...
+.set_num_outputs(1)
+.set_attr<nnvm::FListInputNames>("FListInputNames",
+  [](const NodeAttrs& attrs) {
+    return std::vector<std::string>{"data", "u", "v"};
+  })
+.set_attr<nnvm::FInferShape>("FInferShape", EdgeIDShape)
+.set_attr<nnvm::FInferType>("FInferType", EdgeIDType)
+.set_attr<FInferStorageType>("FInferStorageType", EdgeIDStorageType)
+.set_attr<FComputeEx>("FComputeEx<cpu>", EdgeIDForwardEx<cpu>)
+.set_attr<nnvm::FInplaceOption>("FInplaceOption",
```
eric-haibin-lin: This is actually not used since the first input is sparse. Inplace option with sparse array is not enabled

src/operator/contrib/dgl_graph.cc
```diff
@@ -0,0 +1,466 @@
...
+  } else {
+    LogUnimplementedOp(attrs, ctx, inputs, req, outputs);
+  }
+}
+
+NNVM_REGISTER_OP(_contrib_edge_id)
+.describe(R"code(This operator implements the edge_id function for csr arrays,
+where output[i] = input[u[i], v[i]] if input[u[i], v[i]] is a non-zero element of input,
+otherwise output[i] will be -1. Both u and v should be 1D vectors.
+Example::
```
eric-haibin-lin: If csr.data contains zero values, would it be returned as 0 or -1? I don't see verification of "non-zero" when searching the value in csr.data

zheng-da: The CSR is a special CSR. Its value is edge Id, so it contains edge id 0. So when we search for edges, we only need to check if the location has a value.

eric-haibin-lin: Yes I am aware of that. That's why I think it's not appropriate to put `if input[u[i], v[i]] is a non-zero element of input` since it also returns elements that are zeros and stored in csr.

PR #13297, opened 2018-11-16
[MKLDNN]Add quantized concat

ZhennanQin:
## Description ##
This PR is to add quantized concat op and its MKLDNN implementation.

@pengzhao-intel @TaoLv 

### Essentials ###

### Changes ###

## Comments ##


src/operator/quantization/mkldnn/mkldnn_quantize_concat.cc
```diff
@@ -0,0 +1,117 @@
...
+                                         const std::vector<OpReqType>& req,
+                                         const std::vector<NDArray>& out_data) {
+  const ConcatParam& param_ = nnvm::get<ConcatParam>(attrs.parsed);
+  // Collect data and output min/max
+  std::vector<float> data_min(param_.num_args);
+  std::vector<float> data_max(param_.num_args);
+  float output_min = 0.f;
+  float output_max = 0.f;
+  for (int i = 0; i < param_.num_args; ++i) {
+    data_min[i] = in_data[param_.num_args + 2 * i].data().dptr<float>()[0];
```
larroy: would it be necessary to assert that in_data has the required size?

ZhennanQin: Thanks for review. The corresponding check is added.

kalyc:
Thanks for your contribution @ZhennanQin 

pengzhao-intel:
@larroy please help take a review again. This is an important OP for the quantization flow so we hope it can be merged before r1.4 code freeze.

pengzhao-intel:
@zheng-da @reminisce @szha @eric-haibin-lin @apeforest 
please help take a review :)

src/operator/quantization/quantized_concat.cc
```diff
@@ -0,0 +1,148 @@
...
+    }
+  }
+  TYPE_ASSIGN_CHECK(*out_type, 0, dtype);
+  TYPE_ASSIGN_CHECK(*out_type, 1, mshadow::kFloat32);
+  TYPE_ASSIGN_CHECK(*out_type, 2, mshadow::kFloat32);
+
+  return true;
+}
+
+NNVM_REGISTER_OP(_contrib_quantized_concat)
```
apeforest: Please add contrib operator under contrib module

src/operator/quantization/quantization_utils.h
```diff
@@ -31,6 +31,8 @@
 namespace mxnet {
 namespace op {
 
+static const size_t uint8_range = 255;
```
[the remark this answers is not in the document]
ZhennanQin: Usually we use all capital for macro defined constant, not for static constant. If MXNet prefer that, I can make the change.

src/operator/quantization/quantized_concat.cc
ZhennanQin: All quantized op has contrib prefix, but they all stay in quantization module. @reminisce, do you know the reason behind?

src/operator/quantization/quantization_utils.h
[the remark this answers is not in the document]
apeforest: Sorry for the confusion, MXNet follows the Google C++ Style Guide for const names. Please follow this: https://google.github.io/styleguide/cppguide.html#Constant_Names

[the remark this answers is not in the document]
ZhennanQin: @apeforest, no problem, I would make the change to follow Google C++ Style.

src/operator/nn/mkldnn/mkldnn_concat-inl.h
```diff
@@ -0,0 +1,108 @@
...
+  std::vector<mkldnn::primitive::at> data_mem;
+  std::shared_ptr<mkldnn::memory> out;
+
+ public:
+  mkldnn::concat::primitive_desc fwd_pd;
+
+  MKLDNNConcatFwd(
+      int concat_dim,
+      const std::vector<mkldnn::memory::primitive_desc> &data_md): fwd_pd(concat_dim, data_md) {
+    data.resize(data_md.size());
```
apeforest: nit: indentation

src/operator/quantization/quantized_concat.cc
```diff
@@ -0,0 +1,148 @@
...
+  using namespace mshadow;
+  const ConcatParam& param_ = nnvm::get<ConcatParam>(attrs.parsed);
+  CHECK_EQ(in_shape->size(), static_cast<size_t>(param_.num_args * 3));
+  CHECK_EQ(out_shape->size(), 3U);
+  TShape dshape;
+  index_t size = 0;
+  bool has_zero = false;
+  int axis = -1;
+  for (int i = 0; i < param_.num_args; ++i) {
+    TShape tmp = (*in_shape)[i];
```
apeforest: Maybe use a reference instead of calling the copy constructor here?

src/operator/quantization/quantized_concat.cc
```diff
@@ -0,0 +1,148 @@
...
+    if (tmp.ndim()) {
+      axis = CheckAxis(param_.dim, tmp.ndim());
+      has_zero = tmp[axis] == 0 || has_zero;
+      size += tmp[axis];
+      tmp[axis] = 0;
+      shape_assign(&dshape, tmp);
+    }
+  }
+
+  TShape tmp = (*out_shape)[0];
```
apeforest: Same here: reference seems to be more efficient.

src/operator/quantization/quantized_concat.cc
```diff
@@ -0,0 +1,148 @@
...
+  return std::vector<std::string>{"output", "min_output", "max_output"};
+})
+.set_attr<nnvm::FInferType>("FInferType", ConcatType)
+.set_attr<nnvm::FInferShape>("FInferShape", ConcatShape)
+.set_attr<std::string>("key_var_num_args", "num_args");
+
+NNVM_REGISTER_OP(Concat)
+.set_attr<FQuantizedOp>("FQuantizedOp", [](const NodeAttrs& attrs) {
+  nnvm::NodePtr node = nnvm::Node::Create();
+  node->attrs.op = Op::Get("_contrib_quantized_concat");
```
apeforest: It seems the Concat operator will only have the MKLDNN version. Is this intended?

src/operator/nn/mkldnn/mkldnn_concat-inl.h
ZhennanQin: That's the indentation followed Google C++ style. What's wrong here?

src/operator/quantization/mkldnn/mkldnn_quantized_concat.cc
```diff
@@ -0,0 +1,119 @@
...
+                                         const std::vector<NDArray>& in_data,
+                                         const std::vector<OpReqType>& req,
+                                         const std::vector<NDArray>& out_data) {
+  const ConcatParam& param_ = nnvm::get<ConcatParam>(attrs.parsed);
+  CHECK_EQ(in_data.size(), static_cast<size_t>(param_.num_args * 3));
+  CHECK_EQ(out_data.size(), 3U);
+  // Collect data and output min/max
+  std::vector<float> data_min(param_.num_args);
+  std::vector<float> data_max(param_.num_args);
+  float output_min = 0.f;
```
apeforest: If the in_data are always positive, the output_min will not be updated, right?

src/operator/quantization/mkldnn/mkldnn_quantized_concat.cc
```diff
@@ -0,0 +1,119 @@
...
+ */
+
+#if MXNET_USE_MKLDNN == 1
+#include "../../nn/mkldnn/mkldnn_concat-inl.h"
+#include "../quantization_utils.h"
+
+namespace mxnet {
+namespace op {
+
+namespace quantized_concat_enum {
```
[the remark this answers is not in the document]
ZhennanQin: This is quantized_concat specific enum, which shouldn't be put in concat-inl.h to avoid wrongly used in fp32 concat. And quantized_concat-inl.h doesn't exist, do you want me to create that file only containing this enum? From my opinion, it's a bit over structure.

src/operator/quantization/mkldnn/mkldnn_quantized_concat.cc
ZhennanQin: No, it won't be updated. But it doesn't matter, because we only care the result of max(abs(output_min, output_max)). If the case you mentioned happens, then output_max will dominate the result.

src/operator/quantization/quantized_concat.cc
ZhennanQin: Yes, all quantized op doesn't have default implementation for cpu. MKLDNN is the only cpu implementation. I guess that's why they all have `_contrib_` prefix.

src/operator/nn/mkldnn/mkldnn_concat-inl.h
```diff
@@ -0,0 +1,108 @@
...
+#include <utility>
+#include "../concat-inl.h"
+#include "./mkldnn_ops-inl.h"
+#include "./mkldnn_base-inl.h"
+
+namespace mxnet {
+namespace op {
+
+class MKLDNNConcatFwd {
+  std::shared_ptr<mkldnn::concat> fwd;
```
apeforest: If these are private members. Add identifier and move them after public members

src/operator/nn/mkldnn/mkldnn_concat-inl.h
```diff
@@ -0,0 +1,108 @@
...
+ public:
+  mkldnn::concat::primitive_desc fwd_pd;
+
+  MKLDNNConcatFwd(
+      int concat_dim,
+      const std::vector<mkldnn::memory::primitive_desc> &data_md): fwd_pd(concat_dim, data_md) {
+    data.resize(data_md.size());
+  }
+
+  void SetNewMem(const std::vector<const mkldnn::memory *> &in_data,
```
apeforest: Why move this implementation from .cc file to header here?

src/operator/nn/mkldnn/mkldnn_concat-inl.h
```diff
@@ -0,0 +1,108 @@
...
+    if (this->fwd == nullptr)
+      fwd.reset(new mkldnn::concat(fwd_pd, data_mem, *out));
+  }
+
+  const mkldnn::concat &GetFwd() const {
+    return *fwd;
+  }
+};
+
+static MKLDNNConcatFwd &GetConcatForward(
```
apeforest: Why is this method static?

src/operator/quantization/mkldnn/mkldnn_quantized_concat.cc
apeforest: Even so, I will prefer to initialize output_min to FLT_MAX. Otherwise, this piece code is still buggy.

src/operator/quantization/quantized_concat.cc
apeforest: Okay, I didn't notice that. Keep it as it is then.

src/operator/quantization/quantized_concat.cc
apeforest: What if MKLDNN is not on and user invoke this operator? Any error message will be given?

src/operator/quantization/quantized_concat.cc
```diff
@@ -0,0 +1,148 @@
...
+})
+.set_num_outputs(3)
+.set_attr_parser(ParamParser<ConcatParam>)
+.set_attr<nnvm::FListInputNames>("FListInputNames", [](const NodeAttrs& attrs) {
+  const ConcatParam& params = nnvm::get<ConcatParam>(attrs.parsed);
+  std::vector<std::string> ret;
+  for (int i = 0; i < params.num_args; ++i) {
+    ret.push_back(std::string("arg") + std::to_string(i));
+  }
+  for (int i = params.num_args; i < params.num_args * 2; ++i) {
```
apeforest: Not sure why we need this. And if user list the input names, are they supposed to see these extra input names? Do you have a test for this?

src/operator/nn/mkldnn/mkldnn_concat-inl.h
ZhennanQin: Because this class will be used in both fp32 concat and quantized concat, which means, this header  will be included from both mkldnn_concat.cc and mkldnn_quantized_concat.cc.

src/operator/quantization/quantized_concat.cc
ZhennanQin: Yes, we suppose so. Because they are a part of op visible inputs. All quantized op follow this design, like quantized_conv, quantized_pooling. What kind of test do you want? Just test the ListInputNames functionality? It's not the scope of this PR. It should be covered by existing unit test. And I don't see any op adding test for checking its ListInputNames.

src/operator/nn/mkldnn/mkldnn_concat-inl.h
ZhennanQin: No special reason, just because it's copied from .cc with `static` and it works well. Do you have any question for this?

src/operator/quantization/mkldnn/mkldnn_quantized_concat.cc
ZhennanQin: In quantized world, the maximum of output_min is 0.f, at least on mkldnn side. Any value higher than 0 doesn't have difference, but cause performance drop. Can you tell me what's the buggy?

src/operator/quantization/quantized_concat.cc
ZhennanQin: I'm not sure. It should follow framework default behavior. Quantized op are all supported by FComputeEx, so basically it should only be used when MKLDNN is on. One way to avoid this is to define quantized_concat as a mkldnn specific op, by declaring it inside `MXNET_USE_MKLDNN` macro. But we don't have such backend specific op before.  Do you have any suggestion?

src/operator/quantization/mkldnn/mkldnn_quantized_concat.cc
ZhennanQin: I have to say that's what it suppose to do. Just not as your suppose. I didn't say it's going to get the minimal value from outputs, but it's going to get the minimal value not greater than 0. The 'min' in variable name and comments are refer to the output name:'min_output'. That's the `quantized_concat` op defined. Do you want me to rename the variable to "output_non_positive_min" ?

src/operator/nn/mkldnn/mkldnn_concat-inl.h
ZhennanQin: Done.

src/operator/quantization/quantized_concat.cc
apeforest: I think we should issue an error message that the quantized op is not supported in non MKLDNN build.

src/operator/quantization/mkldnn/mkldnn_quantized_concat.cc
apeforest: Code with good readability should not easily cause confusion or misunderstanding. I think you already embedded two different logic in this one variable.

It will be much cleaner if you initialize output_min to FLT_MAX and add in line L60:
```suggestion
float output_min = FLT_MAX;
out_data[quantized_concat_enum::kMin].data().dptr<float>()[0] = std::min(output_min, 0.0)
```

src/operator/quantization/quantized_concat.cc
apeforest: This function should returns the input argument names when you run at python level:
`>>> concat.list_inputs()`
['data', 'bn_gamma', 'bn_beta', 'bn_moving_mean', 'bn_moving_var']
I am curious what your return result is since these "arg(i)_min" inputs are auto-generated.

You don't need to add unit test. But can you at least try to run a test to see what the output is for your operator?

src/operator/nn/mkldnn/mkldnn_concat-inl.h
apeforest: This is not a good argument. Just because other code used this does not justify it here. Bugs can be copied too.

src/operator/nn/mkldnn/mkldnn_concat-inl.h
apeforest: Then keeping the interface in the header should be sufficient. Why do you move all the implementation from .cc to this header instead?

src/operator/quantization/mkldnn/mkldnn_quantized_concat.cc
ZhennanQin: Firstly, `embedded two different logic in this one variable` doesn't mean buggy. Please choose your word carefully in case of confusing.

Secondly, your suggested code definitely works, but with slower performance.  And I don't think it has better readability. Please get rid of this piece of code if you agree that it's not a bug.

src/operator/nn/mkldnn/mkldnn_concat-inl.h
ZhennanQin: It's not an excuse, the real question is, what are you worried about? why can't this function be static?

src/operator/quantization/quantized_concat.cc
ZhennanQin: Yes, I agree with that. But that beyond the scope of this PR. Currently all quantized op has this issue. We need to create another PR to add error message from framework level for each op that don't have default implementation.

src/operator/nn/mkldnn/mkldnn_concat-inl.h
ZhennanQin: Why can't? I think It's MXNet style that put function body and its declaration inside header. eg. https://github.com/apache/incubator-mxnet/blob/master/src/operator/l2_normalization-inl.h

src/operator/quantization/quantized_concat.cc
apeforest: Can we have that PR first before we merge this? Knowing there is some limitation without issuing any clear message may create a bad user experience.

src/operator/quantization/quantization_utils.h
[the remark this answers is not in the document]
KellenSunderland: Are static const all upper case in Google Style?

src/operator/quantization/mkldnn/mkldnn_quantized_concat.cc
apeforest: Thanks for your explanation. It may not be a bug in this case but having a piece of confusing code is bug-prone. e.g. someone else copy paste these piece of logic in a similar operator but having a different assumption as yours.

A code with good readability should be self-explanatory. Otherwise, a detailed comment should be added. In this case, I would suggest that you either initiate output_min to FLT_MAX so you get the minimal values from data_min correctly and then compare it with 0.0 to get the non-positive one. Or, rename the variable and add comment about your logic. Personally, I prefer the former as it is more easily understood.

Regarding performance loss, coding design should follow the 80/20 rule where 80% performance is determined by 20% of the critical code. I think in this case we should not sacrifice readability for one of the 80% non-critical code.

src/operator/nn/mkldnn/mkldnn_concat-inl.h
apeforest: Please refer to https://github.com/apache/incubator-mxnet/blob/master/src/operator/nn/mkldnn/mkldnn_convolution-inl.h#L102

Could you please make sure all mkldnn operators follow the same design pattern?

src/operator/quantization/quantization_utils.h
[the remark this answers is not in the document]
ZhennanQin: @KellenSunderland  Good to see you! The line you seen is outdated. It's already changed to kUInt8Range.

src/operator/quantization/quantized_concat.cc
ZhennanQin: @apeforest  Build MXNet with `make  USE_OPENCV=1 USE_BLAS=openblas`, and run quantized model, below message is reported:

```
[10:54:08] src/executor/attach_op_execs_pass.cc:351: Neither FCompute nor FComputeEx registered _contrib_quantized_concat
[10:54:08] src/executor/attach_op_execs_pass.cc:351: Neither FCompute nor FComputeEx registered _contrib_quantized_pooling
[10:54:08] src/executor/attach_op_execs_pass.cc:351: Neither FCompute nor FComputeEx registered _contrib_quantized_conv
```
I think framework can handle this case properly.

src/operator/quantization/mkldnn/mkldnn_quantized_concat.cc
ZhennanQin: Comments are added, variable is renamed to reflect the real functionality.

src/operator/nn/mkldnn/mkldnn_concat-inl.h
ZhennanQin: Function body is moved to .cc file.

src/operator/nn/mkldnn/mkldnn_concat-inl.h
ZhennanQin: As discussed offline, this part will keep unchanged.

ZhennanQin:
@apeforest  All comments are addressed. Can you review again? Thanks a lot for keeping review round after round.

TaoLv:
Thanks for the contribution. Now merging.

Issue #13126, opened 2018-11-05
Can't break loop when using > 1 worker with DataLoader


ThomasDelteil:
## Description

When using more than 1 worker with a DataLoader and starting the iteration and breaking the loop straight away, the worker crash and throw an error.

## Environment info (Required)

```
MXNet 1.3.1
```

## Minimum reproducible example

```python

train_dataset = mx.gluon.data.vision.MNIST(train=True).transform_first(mx.gluon.data.vision.transforms.ToTensor())
train_data = mx.gluon.data.DataLoader(train_dataset, shuffle=True, last_batch='rollover', batch_size=100, num_workers=2)

for data, label in train_data:
    break
```
```
Exception in thread Thread-5:
Traceback (most recent call last):
  File "/home/ubuntu/anaconda3/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/ubuntu/anaconda3/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/home/ubuntu/anaconda3/lib/python3.6/site-packages/mxnet/gluon/data/dataloader.py", line 195, in fetcher_loop
    idx, batch = data_queue.get()
  File "/home/ubuntu/anaconda3/lib/python3.6/multiprocessing/queues.py", line 337, in get
    return _ForkingPickler.loads(res)
  File "/home/ubuntu/anaconda3/lib/python3.6/site-packages/mxnet/gluon/data/dataloader.py", line 57, in rebuild_ndarray
    fd = fd.detach()
  File "/home/ubuntu/anaconda3/lib/python3.6/multiprocessing/resource_sharer.py", line 58, in detach
    return reduction.recv_handle(conn)
  File "/home/ubuntu/anaconda3/lib/python3.6/multiprocessing/reduction.py", line 182, in recv_handle
    return recvfds(s, 1)[0]
  File "/home/ubuntu/anaconda3/lib/python3.6/multiprocessing/reduction.py", line 153, in recvfds
    msg, ancdata, flags, addr = sock.recvmsg(1, socket.CMSG_LEN(bytes_size))
ConnectionResetError: [Errno 104] Connection reset by peer

```

leleamol:
@ThomasDelteil Thanks for submitting this issue. I am labeling this issue so that MXNet Community members can help resolve it.

@mxnet-lebel-bot [Gluon, Data-loading]

harshp8l:
[...]

PR #13318, opened 2018-11-19
Improving multi-processing reliability for gluon DataLoader

YutingZhang:
I found some multi-processing-related issues in the Gluon  DataLoader.

 1) Each time a _MultiWorkerIter shuts down, it could leave some dangling processes. The shutdown mechanism could not guarantee that all worker processes can be terminated. As a result, after running for several epochs, more and more dangling processes will stay there.

  This problem barely happens during training. In this case, there is a decent time interval between the last-batch data prefetching and the _MultiWorkerIter's shutting down).
  But the problem frequently happens 1) when I stop the iter before the end of an epoch, and 2) when I use the DataLoader for a data loading service and load the data as fast as possible. In both cases, the time interval between the most recent data prefetching and the iter shutdown are short. I guess that the _MultiWorkerIter iter is unable to shut down properly during active data prefetching.

  To fix this, I explicitly terminate the worker processes inside the shutdown function.

  2) When loading data fast (still mostly during testing and data serving), there seems to be a risk of data racing. The data iter uses a _MultiWorkerIter to cache prefetched data, but the dict does not seem to be thread-safe for concurrent inserting and deleting elements. So occasionally, the data can be missing from the  dict.

  To prevent this, I use a scope lock to guard the dict access.

## Description ##

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/python/mxnet/gluon/data/dataloader.py
+++ b/python/mxnet/gluon/data/dataloader.py
@@ -189,7 +189,7 @@ def worker_loop(dataset, key_queue, data_queue, batchify_fn):
         batch = batchify_fn([dataset[i] for i in samples])
         data_queue.put((idx, batch))
 
-def fetcher_loop(data_queue, data_buffer, pin_memory=False):
+def fetcher_loop(data_queue, data_buffer, pin_memory=False, data_buffer_lock=None):
     """Fetcher loop for fetching data from queue and put in reorder dict."""
     while True:
         idx, batch = data_queue.get()
@@ -199,7 +199,11 @@ def fetcher_loop(data_queue, data_buffer, pin_memory=False):
             batch = _as_in_context(batch, context.cpu_pinned())
         else:
             batch = _as_in_context(batch, context.cpu())
-        data_buffer[idx] = batch
+        if data_buffer_lock is not None:
+            with data_buffer_lock:
+                data_buffer[idx] = batch
+        else:
+            data_buffer[idx] = batch
 
 
 class _MultiWorkerIter(object):
@@ -213,7 +217,10 @@ class _MultiWorkerIter(object):
         self._batch_sampler = batch_sampler
         self._key_queue = Queue()
         self._data_queue = Queue() if sys.version_info[0] <= 2 else SimpleQueue()
+
         self._data_buffer = {}
+        self._data_buffer_lock = threading.Lock()
+
         self._rcvd_idx = 0
         self._sent_idx = 0
         self._iter = iter(self._batch_sampler)
@@ -227,10 +234,11 @@ class _MultiWorkerIter(object):
             worker.daemon = True
             worker.start()
             workers.append(worker)
+        self._workers = workers
 
         self._fetcher = threading.Thread(
             target=fetcher_loop,
-            args=(self._data_queue, self._data_buffer, pin_memory))
+            args=(self._data_queue, self._data_buffer, pin_memory, self._data_buffer_lock))
         self._fetcher.daemon = True
         self._fetcher.start()
 
@@ -261,7 +269,8 @@ class _MultiWorkerIter(object):
 
         while True:
             if self._rcvd_idx in self._data_buffer:
-                batch = self._data_buffer.pop(self._rcvd_idx)
+                with self._data_buffer_lock:
+                    batch = self._data_buffer.pop(self._rcvd_idx)
                 self._rcvd_idx += 1
                 self._push_next()
                 return batch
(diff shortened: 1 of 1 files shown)


python/mxnet/gluon/data/dataloader.py
```diff
@@ -275,9 +284,24 @@ def __iter__(self):
...
+            self._fetcher.join()
+            # send shutdown signal to all worker processes
             for _ in range(self._num_workers):
                 self._key_queue.put((None, None))
-            self._data_queue.put((None, None))
+            # wait 0.5 sec for workers to shut down if necessary
+            for w in self._workers:
+                if w.is_alive():
+                    import time
+                    time.sleep(0.5)   # allow 0.5 sec to join
```
zhreshold: sleep in the main thread is not a good idea.

zhreshold:
@YutingZhang I think it's okay to terminate worker processes right after shutdown, but I don't understand why you mentioned "The shutdown mechanism could not guarantee that all worker processes can be terminated". 

In some cases, if worker has a propagated key queue(i.e., workers are busy), it's likely workers need longer time to exit, but the terminate signal (None, None) would make sure these daemon processes will quit when they finish their jobs, or get killed by the main process. 

So I am curious do you know what caused the dangling processes?

stu1130:
@mxnet-label-bot update [pr-awaiting-response, Gluon, Data-loading]

python/mxnet/gluon/data/dataloader.py
YutingZhang: I have two options:
1) simply remove this loop. Then, do not wait the workers to finish, and just kill them if they are alive. 
2) fork a new process to do the waiting and termination.

YutingZhang:
@zhreshold It was actually also a bit confusing to me, but that was what happened. 

One guess:

Is there any size limit or get-put sync of the `data_queue`? Is it possible that the worker got stuck at the `data_queue.put` ? The `fetch_loop` thread can get joined before the workers (this is possible in original code, and more likely in my PR), and then the `data_queue` is full can the `put` get stuck. Is there any such possibility? 

By the way, I tried to join the workers before sending `(None, None)` to the `fetch_loop` thread, but this can cause the `fetch_loop` to get stuck at `data_queue.get` (the requested data is not in the queue, and no worker will put the data). This can stuck the main thread, if we try to join the `fetch_loop` thread or leave a dangling thread otherwise.

ThomasDelteil:
Can you see if your implementation solves this bug? https://github.com/apache/incubator-mxnet/issues/13126

stu1130:
@mxnet-label-bot update [pr-awaiting-review, Gluon, Data-loading]

zhreshold:
@YutingZhang Oh yes, "Is it possible that the worker got stuck at the data_queue.put" this is possible. If you destroy the queue before workers exit, it may cause problems. 
So actually even though daemon processes got killed after main process exits, it's not guaranteeing the dangling processes will be freed while the main process is still running. 

Now I get your point, thanks @YutingZhang 

The solution is fairly simple, make sure shutdown is killing all workers and fetcher.

python/mxnet/gluon/data/dataloader.py
zhreshold: It's safe to go with 1.

YutingZhang:
> Can you see if your implementation solves this bug? #13126

@ThomasDelteil  Yes. I think it solves the problem. I met this type of problems before, but I could not remember the exact cause now (I did the fixes for my code a while ago). It possibly due to the incorrect ordering of shutting down the workers and the fether_loop.

I tried your example. I could replicate the problem, and the problem was solved by my fixes. How about to try it out at your side?

YutingZhang:
@zhreshold Right. If the `shutdown()` does not wait for process to join, the queue can be destroyed first.
As to the "daemon processes" thing, I think the whole program is the parent process of the workers. As long as the program runs, the deamon progresses will keep staying there. So for each epoch, there is a possiblity to add a few dangling processes.

python/mxnet/gluon/data/dataloader.py
YutingZhang: pushed with option 1

zhreshold:
This PR is good to merge. Before merging, I need to add a note that for statement (2), built-in structures are thread-safe for single operation, so it was actually good.

However, introducing lock here will actually protect if it goes to separate operations in the future and adding peace of mind. So I think the lock is good here. 

Thanks @YutingZhang for contribution.

YutingZhang:
@zhreshold Thank you for merging it. 
I could also remember that single operation of dict should be thread-safe. However, I did get weird problems as described in statement (2). So I started doubting that ...

YutingZhang:
Submitted a similar PR for gluoncv: https://github.com/dmlc/gluon-cv/pull/486

Issue #11796, opened 2018-07-18
Batch_dot does not support FP16 well


szhengac:
The `batch_dot` does not support FP16 well and can make training slower compared to using FP32. This is tested using [Transformer](https://github.com/dmlc/gluon-nlp/blob/master/scripts/nmt/train_transformer.py) model in Gluonnlp. This feature has been added in a [NVIDIA mxnet](https://docs.nvidia.com/deeplearning/dgx/mxnet-release-notes/rel_17.11.html#rel_17.11). So I think it is good to enable this in the master.

ptrendx:
@DickJC123

eric-haibin-lin:
I'm adding cublas strided gemm calls in mshadow https://github.com/dmlc/mshadow/pull/353

eric-haibin-lin:
merged

eric-haibin-lin:
Sorry about the revert. I found that it is better to implement fp16 ops in mxnet instead of in mshadow, since there are built in functionality to detect/enable tensorcore. I can make a PR in maybe two or three days. @sbodenstein are you using symbol or gluon to train transformer?

sbodenstein:
@eric-haibin-lin: we are using symbol to train transformer. That would be great to reenable this as soon as possible.

Is there any reason to not expose `linalg_gemm` and `linalg_gemm2` `float16`-support on GPU as well?

sbodenstein:
@eric-haibin-lin: any updates about this?

eric-haibin-lin:
Added in https://github.com/apache/incubator-mxnet/pull/13716

PR #13336, opened 2018-11-20
GEMM Tensor Core Support

sbodenstein:
This PR adds support for `float32` `linalg.gemm` and `linalg.gemm2` ops to make use of Tensor Cores via implicit casting. It is motivated by there being no way of doing `GEMM` with Tensor Core support in MXNet currently (`linalg.gemm` and `linalg.gemm2` don't support `float16`, and `batch_dot` is super slow for `float16` #11796). It is critical for applications such as Transformers to have a fast version of this operator.

A simple benchmark shows significant speedups on a V100:
```
def benchmark(dtype, dev, shape):
    ctx = mx.gpu()
    a = mx.nd.random.normal(0, 1, shape=shape, ctx=dev, dtype=dtype)
    b = mx.nd.random.normal(0, 1, shape = shape, ctx=dev, dtype=dtype)
    s = mx.symbol.linalg.gemm2(mx.sym.Variable("a"), mx.sym.Variable("b"), transpose_b=False, alpha=1.0)
    e = s.bind(dev, {"a":a,"b":b})
    
    # warmup
    e.forward()
    e.outputs[0].asnumpy()

    begin = time.time()
    for i in range(100):
        e.forward()
    e.outputs[0].asnumpy()

    end = time.time()
    print(end - begin)
    return None
shape = (64, 1024, 1024)
os.environ["MXNET_CUDA_TENSOR_OP_MATH_ALLOW_CONVERSION"] = "0"
benchmark("float32", mx.gpu(), shape)
os.environ["MXNET_CUDA_TENSOR_OP_MATH_ALLOW_CONVERSION"] = "1"
benchmark("float32", mx.gpu(), shape)
```
gives `1.153s` and `0.3957s` (so a factor 2.9x speedup). Its also around 50% faster letting cuBLAS do the casting than using MXNet ops to cast to `float16`, do GEMM in `float16`, and then back to `float32`.

@eric-haibin-lin

### DIFF
--- a/src/operator/linalg_impl.h
+++ b/src/operator/linalg_impl.h
@@ -315,8 +315,56 @@ void linalg_gemm<gpu, mshadow::half::half_t>(const Tensor<gpu, 2, mshadow::half:
                               &beta, C.dptr_, C.stride_, C.size(1) * C.stride_, A.size(0))) \
   }
 
-  LINALG_GPU_BATCH_GEMM(SgemmStridedBatched, float)
   LINALG_GPU_BATCH_GEMM(DgemmStridedBatched, double)
+
+  #if CUDA_VERSION < 9010
+  LINALG_GPU_BATCH_GEMM(SgemmStridedBatched, float)
+  #else
+    template <>
+    inline void linalg_batch_gemm<gpu, float>(const Tensor<gpu, 3, float>& A,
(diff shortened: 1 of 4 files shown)


stu1130:
@sbodenstein Great work!

PR #13339, opened 2018-11-20
CMake: Enable installation of cpp-package headers

MoritzMaxeiner:
This allows installing the cpp-package via CMake's install target:
- Installs the headers of the cpp-package
- Installs the missing nnvm headers (on which we depend)


stu1130:
Thanks for your contribution @MoritzMaxeiner

CMakeLists.txt
```diff
@@ -730,6 +730,7 @@ install(TARGETS ${MXNET_INSTALL_TARGETS}
 )
 
 install(DIRECTORY include/ DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+install(DIRECTORY 3rdparty/tvm/nnvm/include/ DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
```
larroy: Could you please add comments to where are we installing things by default? it's quite hidden in CMAKE documentation where they go. ( I had to dig a bit )

lebeg: Isn't mshadow needed as well?

MoritzMaxeiner:
> Does it work for you? I have some errors when I run install:

Yes, I use these changes[1][2] in our vcpkg fork[3] on Windows (7, 10, Server 2016) where I build and deploy a plugin (shared library) using the CPP package, linked against mxnet's shared library.
Since it seems you're using Linux, are you installing with root permissions? If not, you'd best try to use something like `-DCMAKE_INSTALL_PREFIX=~/.local`

[1] https://git.imp.fu-berlin.de/bioroboticslab/robofish/vcpkg/blob/master/ports/mxnet/0003-install-cpp-package.patch
[2] https://git.imp.fu-berlin.de/bioroboticslab/robofish/vcpkg/blob/master/ports/mxnet/0004-install-nnvm.patch
[3] https://git.imp.fu-berlin.de/bioroboticslab/robofish/vcpkg/tree/master/ports/mxnet

CMakeLists.txt
MoritzMaxeiner: That's a bit out of scope for what I intended with the PR, but sure.
You can find CMake's documentation [here](https://cmake.org/cmake/help/git-stage/command/install.html).

MoritzMaxeiner: As far as I could tell, no public headers include it. At least, my C++ project importing the full cpp-package header works without mshadow being installed.

MoritzMaxeiner: See a005e9c24630bf0df0fa5f0324420ee6a5f19ada

larroy:
It doesn't #13578

CMakeLists.txt
larroy: Thanks for that.

Roshrini:
@MoritzMaxeiner Can you retrigger the build for this PR? It seems to be approved

MoritzMaxeiner:
@Roshrini All checks seem to have passed, now.

larroy:
https://github.com/apache/incubator-mxnet/issues/13801

ChaiBapchya:
@larroy Since this PR is merged, does this fix the cpp build package issue #13801 that you mentioned? If it is fixed, we can update that issue and close it in that case.

PR #13344, opened 2018-11-20
Refactor Jenkinsfiles

marcoabreu:
This PR refactors our Jenkinsfiles that are used in the PR stage for the following reasons:

1. Work around issues with Jenkinsfiles being too large: https://issues.jenkins-ci.org/browse/JENKINS-37984
2. Improve the readability of the Jenkinsfiles. Now you can clearly tell which jobs are being run while implementation details are hidden in another file. Also, this allows code re-usage.
3. Increase the speed of the PR pipeline. We will now have multiple pipelines which run in parallel. This reduces the critical path since a stage no longer has to wait for the slowest job before being able to move forward.

So far, I have not refactored other pipelines. This would be done in a follow-up PR.


ci/jenkins/Jenkinsfile_edge
```diff
@@ -0,0 +1,56 @@
...
+// timeout in minutes
+max_time = 120
+
+node('mxnetlinux-cpu') {
+  // Loading the utilities requires a node context unfortunately
+  checkout scm
+  utils = load('ci/Jenkinsfile_utils.groovy')
+  custom_steps = load('ci/jenkins/Jenkins_steps.groovy')
+}
+utils.assign_node_labels(linux_cpu: 'mxnetlinux-cpu', linux_gpu: 'mxnetlinux-gpu', linux_gpu_p3: 'mxnetlinux-gpu-p3', windows_cpu: 'mxnetwindows-cpu', windows_gpu: 'mxnetwindows-gpu')
```
ThomasDelteil: as discussed offline with @marcoabreu, it would be nice to abstract away the need to have windows node labeling in test files that are testing centos or arm, etc.

marcoabreu: Done

ci/Jenkinsfile_utils.groovy
```diff
@@ -144,9 +157,78 @@ def docker_run(platform, function_name, use_nvidia, shared_mem = '500m') {
   sh command
 }
 
+// Allow publishing to GitHub with a custom context (the status shown under a PR)
+// Credit to https://plugins.jenkins.io/github
+def get_repo_url() {
+  checkout scm
+  sh "git config --get remote.origin.url > .git/remote-url"
```
larroy: Why not .execute()  writing a file doesn't seem the best way.

marcoabreu: There is no way to directly access shell output in Jenkins https://stackoverflow.com/questions/36507410/is-it-possible-to-capture-the-stdout-from-the-sh-dsl-command-in-the-pipeline

I could've used other methods, but that would then have triggered the security mechanisms of Jenkins due to the sandboxing mechanism. Execute (although I don't know exactly which function you're referring to) sounds like it would probably trigger that restriction.

PR #13350, opened 2018-11-21
Fix github status context if running inside a folder

marcoabreu:
This PR resolves problems with the status context if the job is deployed within a Jenkins folder. This then cuts off the job-name, resulting in all status updates overlapping.


marcoabreu:
I'm taking the liberty to merge this right now since all PRs are currently getting a wrong status report

PR #13362, opened 2018-11-21
Add NHWC layout support to Pooling (cuDNN only)

ptrendx:
## Description ##
This PR adds NHWC layout support to Pooling operator. Work mostly done by @DickJC123 

Training using TensorCores is more efficient when using NHWC data layout. Convolutions and BatchNorm operators in MXNet already have support for NHWC, while Pooling operator currently accepts only NCHW format. This PR introduces NHWC layout support for Pooling when using cuDNN.

This PR is part of upstreaming improvements available in NVIDIA's NGC MXNet container and was tested as part of it to train multiple convnets to expected accuracy. We also introduced additional tests checking consistency between NHWC and NCHW versions of pooling.

### Essentials ###

## Comments ##
- Support for channel last is done only when cuDNN is enabled (noted in the updated documentation)

### DIFF
--- a/python/mxnet/gluon/nn/conv_layers.py
+++ b/python/mxnet/gluon/nn/conv_layers.py
@@ -673,7 +673,7 @@ class Conv3DTranspose(_Conv):
 class _Pooling(HybridBlock):
     """Abstract class for different pooling layers."""
     def __init__(self, pool_size, strides, padding, ceil_mode, global_pool,
-                 pool_type, count_include_pad=None, **kwargs):
+                 pool_type, layout, count_include_pad=None, **kwargs):
         super(_Pooling, self).__init__(**kwargs)
         if strides is None:
             strides = pool_size
@@ -684,6 +684,7 @@ class _Pooling(HybridBlock):
         self._kwargs = {
             'kernel': pool_size, 'stride': strides, 'pad': padding,
             'global_pool': global_pool, 'pool_type': pool_type,
+            'layout': layout,
             'pooling_convention': 'full' if ceil_mode else 'valid'}
         if count_include_pad is not None:
             self._kwargs['count_include_pad'] = count_include_pad
@@ -695,7 +696,8 @@ class _Pooling(HybridBlock):
         return F.Pooling(x, name='fwd', **self._kwargs)
 
     def __repr__(self):
-        s = '{name}(size={kernel}, stride={stride}, padding={pad}, ceil_mode={ceil_mode})'
+        s = '{name}(size={kernel}, stride={stride}, padding={pad}, ceil_mode={ceil_mode}'
+        s += ', global_pool={global_pool}, pool_type={pool_type}, layout={layout})'
         return s.format(name=self.__class__.__name__,
                         ceil_mode=self._kwargs['pooling_convention'] == 'full',
                         **self._kwargs)
@@ -738,12 +740,13 @@ class MaxPool1D(_Pooling):
     """
     def __init__(self, pool_size=2, strides=None, padding=0, layout='NCW',
                  ceil_mode=False, **kwargs):
-        assert layout == 'NCW', "Only supports 'NCW' layout for now"
+        assert layout == 'NCW' or layout == 'NWC',\
+            "Only NCW and NWC layouts are valid for 1D"
         if isinstance(pool_size, numeric_types):
             pool_size = (pool_size,)
         assert len(pool_size) == 1, "pool_size must be a number or a list of 1 ints"
         super(MaxPool1D, self).__init__(
-            pool_size, strides, padding, ceil_mode, False, 'max', **kwargs)
+            pool_size, strides, padding, ceil_mode, False, 'max', layout, **kwargs)
 
 
 class MaxPool2D(_Pooling):
@@ -786,12 +789,13 @@ class MaxPool2D(_Pooling):
     """
     def __init__(self, pool_size=(2, 2), strides=None, padding=0, layout='NCHW',
                  ceil_mode=False, **kwargs):
-        assert layout == 'NCHW', "Only supports 'NCHW' layout for now"
+        assert layout == 'NCHW' or layout == 'NHWC',\
+            "Only NCHW and NHWC layouts are valid for 2D"
         if isinstance(pool_size, numeric_types):
             pool_size = (pool_size,)*2
         assert len(pool_size) == 2, "pool_size must be a number or a list of 2 ints"
         super(MaxPool2D, self).__init__(
-            pool_size, strides, padding, ceil_mode, False, 'max', **kwargs)
+            pool_size, strides, padding, ceil_mode, False, 'max', layout, **kwargs)
 
 
 class MaxPool3D(_Pooling):
@@ -836,12 +840,13 @@ class MaxPool3D(_Pooling):
     """
     def __init__(self, pool_size=(2, 2, 2), strides=None, padding=0,
                  ceil_mode=False, layout='NCDHW', **kwargs):
-        assert layout == 'NCDHW', "Only supports 'NCDHW' layout for now"
+        assert layout == 'NCDHW' or layout == 'NDHWC',\
+            "Only NCDHW and NDHWC layouts are valid for 3D"
         if isinstance(pool_size, numeric_types):
             pool_size = (pool_size,)*3
         assert len(pool_size) == 3, "pool_size must be a number or a list of 3 ints"
         super(MaxPool3D, self).__init__(
-            pool_size, strides, padding, ceil_mode, False, 'max', **kwargs)
+            pool_size, strides, padding, ceil_mode, False, 'max', layout, **kwargs)
 
 
 class AvgPool1D(_Pooling):
@@ -882,12 +887,14 @@ class AvgPool1D(_Pooling):
     """
     def __init__(self, pool_size=2, strides=None, padding=0, layout='NCW',
                  ceil_mode=False, count_include_pad=True, **kwargs):
-        assert layout == 'NCW', "Only supports 'NCW' layout for now"
+        assert layout == 'NCW' or layout == 'NWC',\
+            "Only NCW and NWC layouts are valid for 1D"
         if isinstance(pool_size, numeric_types):
             pool_size = (pool_size,)
         assert len(pool_size) == 1, "pool_size must be a number or a list of 1 ints"
         super(AvgPool1D, self).__init__(
-            pool_size, strides, padding, ceil_mode, False, 'avg', count_include_pad, **kwargs)
+            pool_size, strides, padding, ceil_mode, False, 'avg', layout, count_include_pad,
+            **kwargs)
 
 
 class AvgPool2D(_Pooling):
@@ -931,12 +938,14 @@ class AvgPool2D(_Pooling):
     """
     def __init__(self, pool_size=(2, 2), strides=None, padding=0,
                  ceil_mode=False, layout='NCHW', count_include_pad=True, **kwargs):
-        assert layout == 'NCHW', "Only supports 'NCHW' layout for now"
+        assert layout == 'NCHW' or layout == 'NHWC',\
+            "Only NCHW and NHWC layouts are valid for 2D"
         if isinstance(pool_size, numeric_types):
             pool_size = (pool_size,)*2
         assert len(pool_size) == 2, "pool_size must be a number or a list of 2 ints"
         super(AvgPool2D, self).__init__(
-            pool_size, strides, padding, ceil_mode, False, 'avg', count_include_pad, **kwargs)
+            pool_size, strides, padding, ceil_mode, False, 'avg', layout, count_include_pad,
+            **kwargs)
 
 
 class AvgPool3D(_Pooling):
@@ -982,12 +991,14 @@ class AvgPool3D(_Pooling):
     """
     def __init__(self, pool_size=(2, 2, 2), strides=None, padding=0,
                  ceil_mode=False, layout='NCDHW', count_include_pad=True, **kwargs):
-        assert layout == 'NCDHW', "Only supports 'NCDHW' layout for now"
+        assert layout == 'NCDHW' or layout == 'NDHWC',\
+            "Only NCDHW and NDHWC layouts are valid for 3D"
         if isinstance(pool_size, numeric_types):
             pool_size = (pool_size,)*3
         assert len(pool_size) == 3, "pool_size must be a number or a list of 3 ints"
         super(AvgPool3D, self).__init__(
-            pool_size, strides, padding, ceil_mode, False, 'avg', count_include_pad, **kwargs)
+            pool_size, strides, padding, ceil_mode, False, 'avg', layout, count_include_pad,
+            **kwargs)
 
 
 class GlobalMaxPool1D(_Pooling):
@@ -1011,9 +1022,10 @@ class GlobalMaxPool1D(_Pooling):
           when `layout` is `NCW`.
     """
     def __init__(self, layout='NCW', **kwargs):
-        assert layout == 'NCW', "Only supports 'NCW' layout for now"
+        assert layout == 'NCW' or layout == 'NWC',\
+            "Only NCW and NWC layouts are valid for 1D"
         super(GlobalMaxPool1D, self).__init__(
-            (1,), None, 0, True, True, 'max', **kwargs)
+            (1,), None, 0, True, True, 'max', layout, **kwargs)
 
 
 class GlobalMaxPool2D(_Pooling):
@@ -1038,9 +1050,10 @@ class GlobalMaxPool2D(_Pooling):
           `(batch_size, channels, 1, 1)` when `layout` is `NCHW`.
     """
     def __init__(self, layout='NCHW', **kwargs):
-        assert layout == 'NCHW', "Only supports 'NCHW' layout for now"
+        assert layout == 'NCHW' or layout == 'NHWC',\
+            "Only NCHW and NHWC layouts are valid for 2D"
         super(GlobalMaxPool2D, self).__init__(
-            (1, 1), None, 0, True, True, 'max', **kwargs)
+            (1, 1), None, 0, True, True, 'max', layout, **kwargs)
 
 
 class GlobalMaxPool3D(_Pooling):
@@ -1066,9 +1079,10 @@ class GlobalMaxPool3D(_Pooling):
           `(batch_size, channels, 1, 1, 1)` when `layout` is `NCDHW`.
     """
     def __init__(self, layout='NCDHW', **kwargs):
-        assert layout == 'NCDHW', "Only supports 'NCDHW' layout for now"
+        assert layout == 'NCDHW' or layout == 'NDHWC',\
+            "Only NCDHW and NDHWC layouts are valid for 3D"
         super(GlobalMaxPool3D, self).__init__(
-            (1, 1, 1), None, 0, True, True, 'max', **kwargs)
+            (1, 1, 1), None, 0, True, True, 'max', layout, **kwargs)
 
 
 class GlobalAvgPool1D(_Pooling):
@@ -1090,9 +1104,10 @@ class GlobalAvgPool1D(_Pooling):
         - **out**: 3D output tensor with shape `(batch_size, channels, 1)`.
     """
     def __init__(self, layout='NCW', **kwargs):
-        assert layout == 'NCW', "Only supports 'NCW' layout for now"
+        assert layout == 'NCW' or layout == 'NWC',\
+            "Only NCW and NWC layouts are valid for 1D"
         super(GlobalAvgPool1D, self).__init__(
-            (1,), None, 0, True, True, 'avg', **kwargs)
+            (1,), None, 0, True, True, 'avg', layout, **kwargs)
 
 
 class GlobalAvgPool2D(_Pooling):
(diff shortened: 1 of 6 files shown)


stu1130:
Thanks for your contribution @ptrendx 
Could you rebase to the latest master branch

KellenSunderland:
Thanks for the contribution guys.  FYI: I'm keeping an eye on this one and will do my best to get it past CI.  We had a few CI issues that this PR looks like it has been caught up in.  If you have a chance rebasing might help.

marcoabreu:
I took care of it :)

I actually just wanted to show up as contributor of this PR :P Just kidding, I hope everything is fine now.

lupesko:
Thanks for the contribution @DickJC123 and @ptrendx !
Adding @apeforest @yuxihu and @anirudh2290 to help review.

Vikas89:
@ptrendx can you please add more details in description explaining what is this PR doing and the motivation behind doing it(why is this done). Also, if you can please add any required background information, it would be helpful.

src/operator/nn/cudnn/cudnn_pooling-inl.h
```diff
@@ -73,15 +73,17 @@ class CuDNNPoolingOp {
     CUDNN_CALL(cudnnDestroyPoolingDescriptor(pooling_desc_));
   }
 
-  void Forward(const OpContext &ctx, const TBlob &in_data,
+  // Return boolean saying whether pooling configuration is supported.
+  bool Forward(const OpContext &ctx, const TBlob &in_data,
```
apeforest: Is this boolean return necessary? I don't find it in other operators. What happens if it returns false?

src/operator/nn/cudnn/cudnn_pooling-inl.h
```diff
@@ -113,9 +115,11 @@ class CuDNNPoolingOp {
     } else {
       LOG(FATAL) << "Only support 2D or 3D pooling";
     }
+    return true;
   }
 
-  void Backward(const OpContext &ctx, const TBlob &out_grad,
+  // Return boolean saying whether pooling configuration is supported
+  bool Backward(const OpContext &ctx, const TBlob &out_grad,
```
apeforest: Same here.

src/operator/nn/cudnn/cudnn_pooling-inl.h
```diff
@@ -165,55 +170,78 @@ class CuDNNPoolingOp {
     } else {
       LOG(FATAL) << "Only support 2D or 3D pooling";
     }
+    return true;
   }
 
  private:
-  inline void Init(mshadow::Stream<gpu> *s, const TBlob &in_data,
+  // Return boolean saying whether pooling configuration is supported
+  inline bool Init(mshadow::Stream<gpu> *s, const TBlob &in_data,
```
apeforest: Same here

src/operator/nn/cudnn/cudnn_pooling-inl.h
```diff
@@ -165,55 +170,78 @@ class CuDNNPoolingOp {
...
-                                            out.shape_[3]));
+                                            oshape_nchw[0],
+                                            oshape_nchw[1],
+                                            oshape_nchw[2],
+                                            oshape_nchw[3]));
+      int window_height = param_.global_pool ? dshape_nchw[2] : param_.kernel[0];
+      int window_width = param_.global_pool ? dshape_nchw[3] : param_.kernel[1];
+      // CuDNN v7.1.4 backprop kernel doesn't support window sizes 9 and above.
+      #if CUDNN_VERSION == 7104
+      is_supported = window_height <= 8 && window_width <= 8;
```
apeforest: Could you please add a reference to this?

TaoLv:
Have you ever tried to pass NHWC input into MKL-DNN pooling? Seems it's not well handled here.

TaoLv:
The new parameter `layout` will also be passed to quantized_pooling: https://github.com/apache/incubator-mxnet/blob/master/src/operator/quantization/quantized_pooling.cc

TaoLv:
Do we have any other operator exposing layout information to front user? Does NDArray know the layout information of itself?

python/mxnet/gluon/nn/conv_layers.py
```diff
@@ -673,7 +673,7 @@ def __init__(self, channels, kernel_size, strides=(1, 1, 1), padding=(0, 0, 0),
 class _Pooling(HybridBlock):
     """Abstract class for different pooling layers."""
     def __init__(self, pool_size, strides, padding, ceil_mode, global_pool,
-                 pool_type, count_include_pad=None, **kwargs):
+                 pool_type, layout, count_include_pad=None, **kwargs):
```
apeforest: Do we need to expose this option to user? Is there any use case?

src/operator/nn/pooling-inl.h
```diff
@@ -53,6 +53,7 @@ struct PoolingParam : public dmlc::Parameter<PoolingParam> {
   bool cudnn_off;
   dmlc::optional<int> p_value;
   dmlc::optional<bool> count_include_pad;
+  dmlc::optional<int> layout;
```
apeforest: Why only adding this option to pooling?

apeforest:
@ptrendx It would be great if you can add in the PR description why we need this option. Thanks!

python/mxnet/gluon/nn/conv_layers.py
ptrendx: I'm not sure I understand - that option was already exposed to users, e.g. see here: https://mxnet.incubator.apache.org/api/python/gluon/nn.html#mxnet.gluon.nn.MaxPool1D
It just was not passed up to base class and to C++ because pooling supported only 1 layout.

src/operator/nn/cudnn/cudnn_pooling-inl.h
ptrendx: Reference to this - do you mean the cudnn 7.1.4 bug? Not sure where should I put it, it's mentioned here, in fixed issues: https://docs.nvidia.com/deeplearning/sdk/cudnn-release-notes/rel_721.html#rel_721

src/operator/nn/pooling-inl.h
ptrendx: Convolutions already have it (layout parameter), batchnorm has notion of axis as well.

ptrendx:
@TaoLv 
Re MKLDNN: Per title of this PR this is enabling NHWC support only for cuDNN path. I don't know whether MKLDNN supports NHWC pooling.
Re quantized pooling: that file seems to assume NCHW layout. What would you like me to do with it?
Re layout parameter: Yes, other operators expose layout information (in fact, pooling already does as well) - see convolutions (e.g. here https://mxnet.incubator.apache.org/api/python/gluon/nn.html#mxnet.gluon.nn.Conv2D) and batchnorm (axis parameter). NDArray does not know its layout and for many use cases it would not make sense. That is why layout is exposed as an operator parameter.

ptrendx:
@apeforest @Vikas89 Not sure I understand the question. Are you asking why do we need NHWC support for pooling?

src/operator/nn/cudnn/cudnn_pooling-inl.h
apeforest: Yes, if it does not violate any nvidia license rules, it would be great to add the link here as comment.

src/operator/nn/pooling.cu
```diff
@@ -116,10 +118,12 @@ void PoolingGradCompute<gpu>(const nnvm::NodeAttrs& attrs,
...
         case pool_enum::kAvgPooling:
-          GetCuDNNPoolingOp<DType>(param).Backward(ctx, inputs[ograd_idx],
+          if (GetCuDNNPoolingOp<DType>(param).Backward(ctx, inputs[ograd_idx],
                                                    inputs[in_data_idx], inputs[out_data_idx],
-                                                   req[0], outputs[0]);
-          return;
+                                                   req[0], outputs[0])) {
+            return;
+          }
+          break;
```
sandeep-krishnamurthy: should we handle else case here?

tests/python/gpu/test_operator_gpu.py
```diff
@@ -608,6 +608,52 @@ def test_convolution_versions():
...
+
+        sym = mx.sym.Pooling(kernel=(3,3), pool_type='max', pooling_convention='full', name='pool')
+        check_consistency(sym, ctx_list)
+
+        sym = mx.sym.Pooling(kernel=(300,300), pool_type='max', global_pool=True, name='pool')
+        check_consistency(sym, ctx_list)
+
+
+@with_seed()
+@assert_raises_cudnn_not_satisfied(min_version='7.0.1')
```
sandeep-krishnamurthy: should it rather skip this test if cudnn version is < 7.0.1 ?

ptrendx:
@Vikas89  I took a stab at modifying PR description to include motivation and testing that we did.

src/operator/nn/pooling.cu
ptrendx: No - if Backward returns false, program will exit the switch statement and proceed to non-cuDNN implementation.

tests/python/gpu/test_operator_gpu.py
ptrendx: Hmm, skipping this test if cudnn version is not satisfied was more or less my understanding of what this decorator is doing. Is it wrong? What should I do here instead?

DickJC123:
@apeforest Re: bool return of Forward(), Backward(), etc.

As you know, MXNet has been moving to a more imperative style of processing.  Before that, all operators were subclasses of Operator, which defined Forward() and Backward() as returning void.  Now, operators like Pooling are their own class and can define Forward() and Backward() as they desire.  In addition, with every operator, we face the chore of selecting the best implementation (cudnn, cuda, mkldnn, etc.).  Rather than code the selection logic in a central place, I recommend each implementation be asked (in a favored order) whether it supports the operation, given its parameters.  With the cudnn convolution operator back in the "pre-imperative era", I added a 'Supports(..., param, ...) method.  Sadly, this tended to repeat a lot of the checks and analysis done by the actual Forward() and Backward() calls.  Rather than adding a repetitive Supports() method to Pooling, I realized how much cleaner it was to just have Forward() or Backward() return a flag indicating whether the function could be performed.  This allows the implementation selector to cleanly fall back to another implementation.

TaoLv:
> Prior to this PR, in MXNet, the Pooling operator did not have a parameter to specify the layout, forcing a transposition always to NCHW.

@DickJC123 Good to know that. Could you explain a bit more about when and where this kind of transposition happens?

With this new feature, what's the typical usage in user model? Does user need to specify layout in each layer and each operator?

marcoabreu:
I love that approach about the operator selection, dick!

DickJC123:
The forced transposition I referred to is done currently by the user in the python model.  With this new feature, the Transpose operators could get removed in favor of a Pooling with the new layout parameter.

I'm a bit sorry that NDArray's don't carry the layout information.  As a result, the layout has to travel along with the data ('out-of-band') and be passed into the operators along the way.  On the other hand, having multi-input operators deal with inconsistent layouts would be a headache.  Anyway, that's a discussion to be debated in its own thread.

TaoLv:
Sure. It's a problem of MXNet, not this PR. Do you mind documenting some where what does the output layout look like? Then user can know what they need to do before sending the output to next layer.

apeforest:
@DickJC123 Thanks for your detailed explanation about the operator selection logic. I like the way you used a boolean return type for `Forward()` and `Backward()` methods to choose the right operator implementation. Although it might be an elegant way, this method is not yet well received in the developer community. And all the other operators do not currently implement in this way. I am afraid that using it alone in this operator without a clear guideline may cause more confusion for prospect developers. Can you still follow the traditional way of declaring `Forward()` and `Backward()` function and maybe propose this in a separate thread to refactor the `Forward()` and `Backward()` with boolean return type?

TaoLv:
Tested this PR on CPU w/ and w/o MKL-DNN  and got error message as below.
w/o MKL-DNN:
```
  File "test_pool_nhwc.py", line 28, in <module>
    test_pooling_nhwc()
  File "test_pool_nhwc.py", line 21, in test_pooling_nhwc
    out_nhwc = mx.nd.Pooling(data=inp_nhwc, kernel=kernel, pool_type='max', global_pool=False, stride=stride, pad=pad, layout='NHWC')
  File "<string>", line 132, in Pooling
  File "/home/lvtao/Workspace/mxnet/python/mxnet/_ctypes/ndarray.py", line 92, in _imperative_invoke
    ctypes.byref(out_stypes)))
  File "/home/lvtao/Workspace/mxnet/python/mxnet/base.py", line 252, in check_call
mxnet.base.MXNetError: [12:35:52] src/operator/nn/./pooling-inl.h:167: Check failed: param_.layout.value() == kNCW || param_.layout.value() == kNCHW || param_.layout.value() == kNCDHW Need CuDNN for layout support
```
w/o MKL-DNN:
```
terminate called after throwing an instance of 'dmlc::Error'
  what():  [12:28:11] src/engine/./threaded_engine.h:380: std::exception
A fatal error occurred in asynchronous engine operation. If you do not know what caused this error, you can try set environment variable MXNET_ENGINE_TYPE to NaiveEngine and run with debugger (i.e. gdb). This will force all operations to be synchronous and backtrace will give you the series of calls that lead to this error. Remember to set MXNET_ENGINE_TYPE back to empty after debugging.

Stack trace returned 10 entries:
[bt] (0) /home/lvtao/Workspace/mxnet/python/mxnet/../../lib/libmxnet.so(dmlc::StackTrace()+0x42) [0x7f547889afb6]
[bt] (1) /home/lvtao/Workspace/mxnet/python/mxnet/../../lib/libmxnet.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x1b) [0x7f547889b233]
[bt] (2) /home/lvtao/Workspace/mxnet/python/mxnet/../../lib/libmxnet.so(mxnet::engine::ThreadedEngine::ExecuteOprBlock(mxnet::RunContext, mxnet::engine::OprBlock*)+0x66e) [0x7f547b7ad24c]
[bt] (3) /home/lvtao/Workspace/mxnet/python/mxnet/../../lib/libmxnet.so(void mxnet::engine::ThreadedEnginePerDevice::CPUWorker<(dmlc::ConcurrentQueueType)0>(mxnet::Context, mxnet::engine::ThreadedEnginePerDevice::ThreadWorkerBlock<(dmlc::ConcurrentQueueType)0>*, std::shared_ptr<dmlc::ManualEvent> const&)+0x8b) [0x7f547b7bfe9d]
[bt] (4) /home/lvtao/Workspace/mxnet/python/mxnet/../../lib/libmxnet.so(mxnet::engine::ThreadedEnginePerDevice::PushToExecute(mxnet::engine::OprBlock*, bool)::{lambda()#1}::operator()() const::{lambda(std::shared_ptr<dmlc::ManualEvent>)#1}::operator()(dmlc::ManualEvent) const+0x33) [0x7f547b7be22b]
[bt] (5) /home/lvtao/Workspace/mxnet/python/mxnet/../../lib/libmxnet.so(std::_Function_handler<void (std::shared_ptr<dmlc::ManualEvent>), mxnet::engine::ThreadedEnginePerDevice::PushToExecute(mxnet::engine::OprBlock*, bool)::{lambda()#1}::operator()() const::{lambda(std::shared_ptr<dmlc::ManualEvent>)#1}>::_M_invoke(std::_Any_data const&, std::shared_ptr<dmlc::ManualEvent>)+0x4a) [0x7f547b7c2751]
[bt] (6) /home/lvtao/Workspace/mxnet/python/mxnet/../../lib/libmxnet.so(std::function<void (std::shared_ptr<dmlc::ManualEvent>)>::operator()(std::shared_ptr<dmlc::ManualEvent>) const+0x5c) [0x7f547b7b6142]
[bt] (7) /home/lvtao/Workspace/mxnet/python/mxnet/../../lib/libmxnet.so(void std::_Bind_simple<std::function<void (std::shared_ptr<dmlc::ManualEvent>)> (std::shared_ptr<dmlc::ManualEvent>)>::_M_invoke<0ul>(std::_Index_tuple<0ul>)+0x56) [0x7f547b7b6036]
[bt] (8) /home/lvtao/Workspace/mxnet/python/mxnet/../../lib/libmxnet.so(std::_Bind_simple<std::function<void (std::shared_ptr<dmlc::ManualEvent>)> (std::shared_ptr<dmlc::ManualEvent>)>::operator()()+0x1b) [0x7f547b7b5ee1]
[bt] (9) /home/lvtao/Workspace/mxnet/python/mxnet/../../lib/libmxnet.so(std::thread::_Impl<std::_Bind_simple<std::function<void (std::shared_ptr<dmlc::ManualEvent>)> (std::shared_ptr<dmlc::ManualEvent>)> >::_M_run()+0x1c) [0x7f547b7b5e30]


Aborted (core dumped)
```

marcoabreu:
I would be prefer to leave the booleans in. If there is no harm in taking that approach, I'd rather go down that path and use this operator implementation as example for others rather than waiting for a saint who's going to refactor everything - because that won't happen.

Dicks designs often greatly improved the architecture or usability of mxnet, thus I really appreciate these things and definitely would not like to see these improvements blocked because some people can't adapt that fast.

This project had some sub-par design decisions that mainly involved choosing preprocessor statements instead of proper class design and abstraction. This PR is a good step towards the first direction and I think everybody should support this path.

I'm feeling strong for having this new method in as long as it doesn't have any drawbacks. If there are any, I'm happy to revisit my decision.

marcoabreu:
Thanks for the test Tao! Can you check why that has not been caught by our CI? What are we missing?

TaoLv:
Seems there is no negative test for this feature. Only GPU with certain versions of CUDNN will be tested.

marcoabreu:
Great catch! Do you mind adding a test that runs on both cpu and gpu?

I'm still having trouble to understand why we are exposing the Backend (whether it's cudnn or mkldnn) to the frontend and making conditional tests. The tests should work with any Backend (except if there is no valid implementation available). In an ideal world, we would just call the Operator and the mkldnn/cudnn magic would be done by the Backend. I think this new structure is a good step towards that.

Can we already add a generalized test that both works on CPU and GPU at this stage to catch Taos error?

apeforest:
@marcoabreu The harm of leaving this boolean return type *in this PR* is that it will cause confusion to developers and also leave unclear messages to MXNet users. E.g. when it returns false, what happens in the Python frontend, any clear message? If this is meant to signal the frontend that this operatator implementation is not supported on particular platform, we need a consistent log message (in fact, we already achieve this via a different way in https://github.com/apache/incubator-mxnet/blob/master/src/operator/nn/mkldnn/mkldnn_pooling.cc#L53). Falling back to the conventional `void` return type seems not affect any functionality of this PR. I would like this PR to mainly focusing on the new feature in Pooling only and we can have a separate PR to change `void` to `boolean` return type and invite more discussions there. I am open to any other suggestions and add @Vikas89 @yuxihu for their comments.

src/operator/nn/pooling-inl.h
```diff
@@ -92,6 +93,16 @@ struct PoolingParam : public dmlc::Parameter<PoolingParam> {
               "calculation. For example, with a 5*5 kernel on a 3*3 corner of a image,"
               "the sum of the 9 valid elements will be divided by 25 if this is set to true,"
               "or it will be divided by 9 if this is set to false. Defaults to true.");
+
+    DMLC_DECLARE_FIELD(layout)
```
yuxihu: Where is the enum for NWC defined? I see it is used in multiple places in other files like pooling.cc.

src/operator/nn/cudnn/cudnn_pooling-inl.h
```diff
@@ -165,55 +170,80 @@ class CuDNNPoolingOp {
...
                                              param_.global_pool ? 0 : param_.pad[0],
-                                             param_.global_ppol ? 0 : param_.pad[1],
+                                             param_.global_pool ? 0 : param_.pad[1],
                                              param_.global_pool ? 1 : param_.stride[0],
                                              param_.global_pool ? 1 : param_.stride[1]));
       #endif
     } else {
+      CHECK(param_.layout.value() == mshadow::kNCDHW ||
+            param_.layout.value() == mshadow::kNDHWC) << "Need 3D layout";
+      CHECK(param_.layout.value() == mshadow::kNCDHW) << "Only the NCDHW layout is supported.";
```
yuxihu: If only NCDHW layout is supported, it seems me that there is no need to have the above "Need 3D layout" check.

Besides that, I am also a little bit confused here. In conv_layers.py, it is mentioned that "Dimension ordering of data and weight. Only supports 'NCDHW' and 'NDHWC' (only with cuDNN) layouts for now." for all 3D pool classes. Why here we only support NCDHW? This is CuDNNPoolingOp class, right? Am I missing something?

marcoabreu:
@apeforest I expect that this PR is one of multiple. In the end, there will be a dispatcher in between the operators and the initial call to the operators. That dispatcher will evaluate the capabilities of every registered operator and then return a unified message. The frontends shouldn't even know about the fact that there may be multiple operators.

The way it was done in mkldnn is also not consistent. This evaluation has to come from a central component and the further processing as well. In this particular example you gave, the process would then fail to execute an operator although there might be other implementations available that actually support the requested operation.

I think you got valid points regarding what's going to happen with the return value and at which point this logic is finally handled. Dick might be happy to answer this.

ptrendx:
Closing in favor of PR #13749

PR #13364, opened 2018-11-21
[MXNET-1225] Always use config.mk in make install instructions

zachgk:
## Description ##
Arguments to make for the mxnet build can be passed either directly through the command line or through config.mk. The install instructions are inconsistent and use both options. However, some frontends such as scala require the arguments as well so they must be passed to both the initial make and the following "make scalapkg". Failure to pass the arguments to the scalapkg make can result in an improper configuration and slower performance. To resolve this, I am switching the install instructions to consistently use the config.mk because it is automatically used by all make commands which resolves the potential frontend problems.

### Essentials ###

@nswamy @lanking520 @yzhliu @andrewfayres @piyushghai

### DIFF
--- a/docs/install/build_from_source.md
+++ b/docs/install/build_from_source.md
@@ -2,6 +2,7 @@
 
 This document explains how to build MXNet from source code.
 
+**For Java/Scala/Clojure, please follow [this guide instead](./scala_setup.md)**
 
 ## Overview
 
@@ -27,7 +28,6 @@ MXNet's newest and most popular API is Gluon. Gluon is built into the Python bin
     - [Python (includes Gluon)](../api/python/index.html)
     - [C++](../api/c++/index.html)
     - [Clojure](../api/clojure/index.html)
-    - Java (coming soon)
     - [Julia](../api/julia/index.html)
     - [Perl](../api/perl/index.html)
     - [R](../api/r/index.html)
@@ -35,6 +35,7 @@ MXNet's newest and most popular API is Gluon. Gluon is built into the Python bin
     - [Java](../api/java/index.html)
 
 <hr>
+
 ## Build Instructions by Operating System
 
 Detailed instructions are provided per operating system. Each of these guides also covers how to install the specific [Language Bindings](#installing-mxnet-language-bindings) you require.
@@ -160,7 +161,7 @@ More information on turning these features on or off are found in the following 
 ## Build Configurations
 
 There is a configuration file for make,
-[`make/config.mk`](https://github.com/apache/incubator-mxnet/blob/master/make/config.mk), that contains all the compilation options. You can edit it and then run `make` or `cmake`. `cmake` is recommended for building MXNet (and is required to build with MKLDNN), however you may use `make` instead.
+[`make/config.mk`](https://github.com/apache/incubator-mxnet/blob/master/make/config.mk), that contains all the compilation options. You can edit it and then run `make` or `cmake`. `cmake` is recommended for building MXNet (and is required to build with MKLDNN), however you may use `make` instead. For building with Java/Scala/Clojure, only `make` is supported.
 
 <hr>
 
@@ -214,7 +215,7 @@ It is recommended to set environment variable NCCL_LAUNCH_MODE to PARALLEL when 
 For example, you can specify using all cores on Linux as follows:
 
 ```bash
-cmake -j $(nproc) .
+cmake -j$(nproc)
 ```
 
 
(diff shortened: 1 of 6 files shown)


zachgk:
@mxnet-label-bot add [Build, Doc, Installation]

Also @aaronmarkham

larroy:
Why are we documenting make instead of CMake? We want to completely switch over to CMake...

zachgk:
@larroy I am documenting make because as far as I know the scala package only works through make. If it works through cmake, is there anything blocking us from entirely switching all make commands to cmake throughout the install instructions?

Also, does cmake use config.mk as well? If it does, I think the same thing applies such that all cmake commands should use config.mk instead of passing arguments directly. I ran into a few cmake commands that did not use config.mk in the build_from_source.md, but I was unsure whether they could be switched or not.

larroy:
I'm not familiar with the compilation in the Scala package. And no CMake doesn't work with config.mk

Just wanted to make you aware of the CMake efforts.

zachgk:

@larroy The scala package compilation process currently requires a number of building and linking flags used to build the backend to be passed to the Java Native Interface as well. So, we currently call make and then make calls maven with those flags. Right now, we are working on improving the scala build so that it doesn't require those flags, but it doesn't seem to be trivial. Therefore, I added redundant flags to all cmake calls for now since it would at least result in the scala build always working.

lanking520:
Could you please raise a discussion on devlist?

andrewfayres:
I ran through this to test. Everything seemed to work fine except I had to specify USE_CUDA = 0.

ddavydenko:
@zachgk , @lanking520 , do you guys think it makes sense to merge this in?

docs/install/ubuntu_setup.md
```diff
@@ -165,23 +165,34 @@ If building on CPU and using OpenBLAS:
 ```bash
     git clone --recursive https://github.com/apache/incubator-mxnet.git
     cd incubator-mxnet
-    make -j $(nproc) USE_OPENCV=1 USE_BLAS=openblas
+    echo "USE_OPENCV = 1" >> ./config.mk
+    echo "USE_BLAS = openblas" >> ./config.mk
+    make -j $(nproc)
```
aaronmarkham: Can you make this cmake instead? That way we're consistent throughout?

docs/install/build_from_source.md
```diff
@@ -214,57 +214,79 @@ It is recommended to set environment variable NCCL_LAUNCH_MODE to PARALLEL when
...
 * Build MXNet with `cmake` and install with MKL DNN, GPU, and OpenCV support:
 
 ```bash
-cmake -j USE_CUDA=1 USE_CUDA_PATH=/usr/local/cuda USE_CUDNN=1 USE_MKLDNN=1
+echo "USE_OPENCV = 1" >> ./config.mk
+echo "USE_CUDA = 1" >> ./config.mk
+echo "USE_CUDA_PATH = /usr/local/cuda" >> ./config.mk
+echo "USE_CUDNN = 1" >> ./config.mk
+echo "USE_MKLDNN = 1" >> ./config.mk
+cmake -j $(nproc) USE_CUDA=1 USE_CUDA_PATH=/usr/local/cuda USE_CUDNN=1 USE_MKLDNN=1 .
```
larroy: this is not how cmake is invoked, and -j is not a valid cmake option.

larroy:
I would suggest to converge in using CMake, I would propose to use a thin wrapper over cmake so we can use a file to define options in this fashion:

https://github.com/apache/incubator-mxnet/blob/master/cmake/cmake_options.yml

This would then work in the same way as config.mk now. But parsed into a commandline as it's done here:

https://github.com/apache/incubator-mxnet/blob/master/dev_menu.py#L48

Also, some of the changes proposed here don't work at all, have they been tested locally?

docs/install/build_from_source.md
zachgk: @larroy, I was testing it on an EC2 and it seems to work because -j was added in cmake 3.12 (https://cmake.org/cmake/help/v3.12/release/3.12.html#cmake-3-12-release-notes). But, it may be better to do something compatible with older versions of cmake. What is the correct way that we should be invoking cmake?

zachgk:
@larroy @aaronmarkham 
I discussed the issue offline with lanking520 and andrewfayres. Right now, this PR is just a stopgap until we complete the final solution worked on in https://issues.apache.org/jira/browse/MXNET-1224. After that, we can just migrate all backend builds into cmake and that will be fine from the scala end.

In the meantime, we have a redundancy in our docs between the build_from_source instructions and the os specific files. Our current plan is to take advantage of it by leaving cmake in the build_from_source and make in the os specific. Then, we will tell all scala users that they must use the os specific instructions and not the build from source instructions. If they did build using cmake and the build_from_source instructions, they should "make clean" and remake it with the os specific instructions.

aaronmarkham:
@larroy Can you point me to more info on "dev menu"? I think I understand what you're suggesting, but I don't see anything on the wiki or in the repo, other than the code itself. 
Let me try to play back my understanding:
1) users only use cmake (so our instructions are simple and clear)
2) cmake uses the cmake_options.yml file
3) cmake or maybe dev_menu.py generates an artifact (config.mk) that make can use
4) scala's current make requirement is satisfied by this artifact

larroy:
@aaronmarkham I extended dev_menu.py with a CMake builder helper that uses cmake options from a file, as the pattern of copying config.mk was useful for several people:

You can try what I mean by using ./dev_menu.py and choosing "Local build"   check the CMake class inside that file:

https://github.com/apache/incubator-mxnet/blob/e0309f453175d066f8a147a3ff5afb01c575327f/dev_menu.py#L54

docs/install/build_from_source.md
larroy: check how is done in ci/docker/runtime_functions.sh usually:
```
mkdir -p build && cd build
cmake\
    -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
    -DCMAKE_C_COMPILER_LAUNCHER=ccache \
    -DUSE_CPP_PACKAGE=ON\
    -DUSE_CUDA=OFF\
    -DUSE_OPENMP=OFF\
    -DUSE_OPENCV=ON\
    -DCMAKE_BUILD_TYPE=Debug\
    -GNinja ..
ninja -v
```

I woud say let's not add options like -j which won't work in the CMake that comes with most distros. Also check the comment below regarding a possible wrapping of CMake in python to simplify builds as is done in dev_menu.py

Hope it helps, thanks.

lebeg:
I would like to mention that, unfortunately, the current option for algebra packages in CMake is [`BLAS`](https://github.com/apache/incubator-mxnet/blob/master/cmake/ChooseBlas.cmake#L18) and not [`USE_BLAS`](https://github.com/apache/incubator-mxnet/blob/master/Makefile#L173) as in make.

zachgk:
@larroy 
If we use the cmake_options approach, we will also need to configure cmake to work for scala similarly to how make works now. However, our plan is that we will actually drop make altogether in the near future and use mvn directly. So, I would rather just warn users for now instead of doing a temporary technological fix. Also, I corrected the cmake usages in the build_from_source file (they were broken in the old docs as well).

If you have no objections, I will also open a PR to move the rest of the install instructions to cmake once scala is ready.

docs/install/c_plus_plus.md
```diff
@@ -6,7 +6,11 @@ To enable C++ package, just add `USE_CPP_PACKAGE=1` in the [build from source](b
 For example to build MXNet with GPU support and the C++ package, OpenCV, and OpenBLAS, from the project root you would run:
 
 ```bash
-make -j USE_CPP_PACKAGE=1 USE_OPENCV=1 USE_BLAS=openblas USE_CUDA=1
+echo "USE_CPP_PACKAGE=1" >> ./config.mk
+echo "USE_OPENCV=1" >> ./config.mk
+echo "USE_BLAS=openblas" >> ./config.mk
+echo "USE_CUDA=1" >> ./config.mk
+make -j $(nproc)
```
larroy: USE_CPP_PACKAGE also works with CMake, can we use CMake here as well?

larroy:
@zachgk shall we leave Scala as it is with Make but move the others to CMake? that would be my preferred option.

zachgk:
I fixed the CPP instructions for now, but all of the other usages of make in the install docs could be reached on the path to installing scala (scala defers the engine build to the os specific pages)

docs/install/build_from_source.md
```diff
@@ -203,68 +204,74 @@ It is recommended to set environment variable NCCL_LAUNCH_MODE to PARALLEL when
...
 ### Usage Examples
 
-* `-j` runs multiple jobs against multi-core CPUs.
-
 For example, you can specify using all cores on Linux as follows:
 
 ```bash
-cmake -j$(nproc)
+cmake -GNinja .
+ninja -v
```
larroy: Should we better recommend an out of source build?  
mkdir build && cd build
cmake -GNinja ..
ninja -v

zachgk: Ok

docs/install/build_from_source.md
```diff
@@ -203,68 +204,82 @@ It is recommended to set environment variable NCCL_LAUNCH_MODE to PARALLEL when
...
 ```
 
 #### Recommended for Systems with NVIDIA GPUs
 * Build with both OpenBLAS, GPU, and OpenCV support:
 
 ```bash
-cmake -j BLAS=open USE_CUDA=1 USE_CUDA_PATH=/usr/local/cuda USE_CUDNN=1
+mkdir build && cd build
+cmake -DBLAS=open -DUSE_CUDA=1 -DUSE_CUDA_PATH=/usr/local/cuda -DUSE_CUDNN=1 -GNinja .
```
lanking520: What does `ninja` do?

zachgk: It is a build system (https://ninja-build.org/). I believe CMake generates the ninja configuration files and then ninja performs the actual build.

PR #13376, opened 2018-11-22
Fix incorrect delete in MXExecutorReshape exception handling

taliesinb:
MXExecutorReshape makes a very obvious and simple error in its cleanup code, which is that it deletes the location of the out pointer rather the pointed-to executor. It also does this deletion even if the out pointer was not set, so it could delete a nullptr or other garbage value.

### DIFF
--- a/src/c_api/c_api_executor.cc
+++ b/src/c_api/c_api_executor.cc
@@ -148,8 +148,6 @@ int MXExecutorBindEX(SymbolHandle symbol_handle,
                      NDArrayHandle *aux_states,
                      ExecutorHandle shared_exec,
                      ExecutorHandle *out) {
-  Executor* exec = nullptr;
-
   API_BEGIN();
   nnvm::Symbol *symb = static_cast<nnvm::Symbol*>(symbol_handle);
   Context ctx = Context::Create(static_cast<Context::DeviceType>(dev_type), dev_id);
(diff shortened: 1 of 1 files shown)


marcoabreu:
@mkolod @KellenSunderland

KellenSunderland:
Taking a look ...

KellenSunderland:
So I see we potentially are deleting memory we shouldn't whenever an error is encountered.  What I'm wondering is what would happen if an error is now encountered on line 569 for example?

stu1130:
Thanks for your contribution @taliesinb

taliesinb:
@KellenSunderland good point. I've moved the nullptr set up to just after API_BEGIN.

src/c_api/c_api_executor.cc
```diff
@@ -560,6 +560,7 @@ int MXExecutorReshape(int partial_shaping,
                       ExecutorHandle *out) {
   MXAPIThreadLocalEntry *ret = MXAPIThreadLocalStore::Get();
   API_BEGIN();
+  *out = nullptr; // ensure we can know whether to free executor on early abort
```
KellenSunderland: Small linting issue:
src/c_api/c_api_executor.cc:563:  At least two spaces is best between code and comments  [whitespace/comments] [2]
Category 'whitespace' errors found: 1

taliesinb: I've added the space, and am also creating a temporary pointer for the new executor as I see is done in other functions like MXExecutorGetOptimizedSymbol.

taliesinb:
@KellenSunderland Good to merge?

roywei:
@anirudh2290 @apeforest could you help take a look and merge if looks good? Thanks!

roywei:
@mxnet-label-bot add[C API, pr-awaiting-merge]

KellenSunderland:
@taliesinb Sorry for the delay, looks good.  Thanks for the fix!

PR #13408, opened 2018-11-26
Change docker login

marcoabreu:
This PR changes how we publish our Docker cache. This is required for security reasons.

### DIFF
--- a/ci/docker_cache.py
+++ b/ci/docker_cache.py
@@ -30,10 +30,12 @@ import argparse
 import sys
 import subprocess
 import json
+import time
 from typing import *
 import build as build_util
 
-
+DOCKERHUB_LOGIN_NUM_RETRIES = 5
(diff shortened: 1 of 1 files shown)


ci/docker_cache.py
```diff
@@ -217,7 +249,11 @@ def script_name() -> str:
     args = parser.parse_args()
 
     platforms = build_util.get_platforms()
-    return build_save_containers(platforms=platforms, registry=args.docker_registry, load_cache=True)
+    try:
```
larroy: Can we use with idiom? Python context? you can make it a separate PR

marcoabreu: Good idea, will do.

marcoabreu:
I wanted to avoid environment variables since these would have to be defined on the Jenkins master level and it gets quite messy.

marcoabreu:
Self-merge due to urgency. Docker cache is disabled right now.

PR #13427, opened 2018-11-27
Nit on string conversion for mxdoc.py

harshp8l:
## Description ##
f-strings in python3.6 are more performant, prefer using this over the .format() convention in the docs file

### Essentials ###
- [x] To my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### DIFF
--- a/docs/mxdoc.py
+++ b/docs/mxdoc.py
@@ -29,7 +29,7 @@ from six import StringIO
 from six.moves import configparser
 
 _BUILD_VER = os.getenv('BUILD_VER', 'default')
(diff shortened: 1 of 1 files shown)


harshp8l:
@mxnet-label-bot add [doc, pr-awaiting-review]

ChaiBapchya:
In such a scenario do we update the CI to use python3.6 or continue to use .format() conversion.
Logically it makes sense to upgrade to 3.6 (but might cause breaking changes) and might involve some work.

larroy:
I don't think we should break python3 backwards compatibility for this feature.

Issue #13421, opened 2018-11-27
MKL-DNN deconvolution runs into crash


TaoLv:
Note: Providing complete information in the most concise form is the best way to get help. This issue template serves as the checklist for essential information to most of the technical issues and bug reports. For non-technical issues and feature requests, feel free to present the information in what you believe is the best form.

For Q & A and discussion, please start a discussion thread at https://discuss.mxnet.io 

## Description
MKL-DNN deconvolution might run into crash since commit 91c536d2b3fe14fb84dff568a4a2ea240ea5ab31. I think it should be fixed before 1.4.0 code freezing.

## Environment info (Required)

```
What to do:
1. Download the diagnosis script from https://raw.githubusercontent.com/apache/incubator-mxnet/master/tools/diagnose.py
2. Run the script using `python diagnose.py` and paste its output here.

```

Package used (Python/R/Scala/Julia):
Python

For Scala user, please provide:
1. Java version: (`java -version`)
2. Maven version: (`mvn -version`)
3. Scala runtime if applicable: (`scala -version`)

For R user, please provide R `sessionInfo()`:

## Build info (Required if built from source)

Compiler (gcc/clang/mingw/visual studio): gcc 4.8.5

MXNet commit hash:
(Paste the output of `git rev-parse HEAD` here.)
91c536d2b3fe14fb84dff568a4a2ea240ea5ab31

Build config:
(Paste the content of config.mk, or the build command.)
make -j20 USE_OPENCV=1 USE_MKLDNN=1 USE_BLAS=mkl USE_PROFILER=1 DEBUG=1

## Error Message:
$ python deconv.py
Traceback (most recent call last):
  File "deconv.py", line 27, in <module>
    t = o.asnumpy()
  File "/home/lvtao/Workspace/mxnet-official/python/mxnet/ndarray/ndarray.py", line 1980, in asnumpy
    ctypes.c_size_t(data.size)))
  File "/home/lvtao/Workspace/mxnet-official/python/mxnet/base.py", line 252, in check_call
    raise MXNetError(py_str(_LIB.MXGetLastError()))
mxnet.base.MXNetError: [22:32:01] src/operator/nn/mkldnn/mkldnn_deconvolution.cc:270: Check failed: weight_mem->get_primitive_desc() == fwd_pd.weights_primitive_desc()

Stack trace returned 10 entries:
[...]

PR #11778, opened 2018-07-17
[MXNET-483] C++ tests for mkldnn convolution/deconvolution operator

azai91:
## Description ##
Add tests for MKLDNN convolution / deconvolution operator. Need to convert MKLDNN conv/deconv inputs if they are views (https://github.com/apache/incubator-mxnet/issues/12303).

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] Add tests for MKLDNN convolution / deconvolution operator
- [x] Reorder inputs of conv / deconv mkldnn inputs if MKLDNN and view.

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #13433, opened 2018-11-28
Fix deconvolution  / PR 13421

azai91:
## Description ##
PR to fix issues - https://github.com/apache/incubator-mxnet/issues/13421. Added unit test for deconvolution inference and reverted refactor change to mkldnn  deconvolution inference commmited in https://github.com/apache/incubator-mxnet/pull/11778.

### Essentials ###

### Changes ###
- [ ] refactor change to mkldnn  deconvolution inference. 

## Comments ##

### DIFF
--- a/tests/python/mkl/test_mkldnn.py
+++ b/tests/python/mkl/test_mkldnn.py
@@ -398,6 +398,24 @@ def test_softmax_with_large_inputs():
     softmax_forward(mx.nd.array([[[[-3.4e38,-3.4e38]]]]), np.array([1.0,1.0]))
     softmax_forward(mx.nd.array([[[[3.4e38,3.4e38]]]]), np.array([1.0,1.0]))
 
+def test_deconvolution_inference():
+    np.random.seed(12345)
+    num_filter = 256
+    num_group = 1
+    kernel = (3, 3)
+    pad = (1, 1)
+    shape = (1, 256, 200, 233)
+    x = mx.sym.Variable('x')
+    w = mx.sym.Variable('w')
+    y = mx.sym.Deconvolution(data=x, weight=w, num_filter=num_filter, num_group=num_group, kernel=kernel, no_bias=True, pad=pad)
+    exe = y.simple_bind(ctx=mx.cpu(), x=shape, grad_req='null')
+    exe.arg_arrays[0][:] = np.random.normal(size=exe.arg_arrays[0].shape)
+    exe.arg_arrays[1][:] = np.random.normal(size=exe.arg_arrays[1].shape)
+    for i in range(10):
+        exe.forward(is_train=False)
+        o = exe.outputs[0]
+        t = o.asnumpy()
+
 @with_seed()
 def test_non_mkldnn_fcomputeex():
     # test special case where MKLDNN formatted NDArray feeds into non-mkldnn fcomputeex operator
--- a/src/operator/nn/mkldnn/mkldnn_deconvolution.cc
+++ b/src/operator/nn/mkldnn/mkldnn_deconvolution.cc
@@ -262,12 +262,14 @@ void MKLDNNDeconvForward::SetDataHandle(const DeconvolutionParam& param,
     // For inference, we want to reorder the weight array so we don't need to
     // reorder data every time.
     if (weight.IsDefaultData()) {
+      weight_mem = GetWeights(weight, fwd_pd.weights_primitive_desc(), param.num_group);
       // We also need to modify the layout on the original weight array. The
       // data conversion happens after the weight array is used.
       const_cast<NDArray&>(weight).MKLDNNDataReorderAsync(fwd_pd.weights_primitive_desc());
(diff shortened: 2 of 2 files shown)


azai91:
@TaoLv @anirudh2290

vandanavk:
@mxnet-label-bot [MKLDNN, pr-awaiting-review]

tests/python/mkl/test_mkldnn.py
```diff
@@ -398,6 +398,24 @@ def softmax_forward(input_data, true_output):
     softmax_forward(mx.nd.array([[[[-3.4e38,-3.4e38]]]]), np.array([1.0,1.0]))
     softmax_forward(mx.nd.array([[[[3.4e38,3.4e38]]]]), np.array([1.0,1.0]))
 
+def test_deconvolution_inference():
+    np.random.seed(12345)
```
pengzhao-intel: set random seed @with_seed()

tests/python/mkl/test_mkldnn.py
```diff
@@ -398,6 +398,24 @@ def softmax_forward(input_data, true_output):
     softmax_forward(mx.nd.array([[[[-3.4e38,-3.4e38]]]]), np.array([1.0,1.0]))
     softmax_forward(mx.nd.array([[[[3.4e38,3.4e38]]]]), np.array([1.0,1.0]))
 
+@with_seed(12345)
```
TaoLv: No need to use a fixed seed here.

azai91:
@TaoLv I think you're referring to this test? https://github.com/apache/incubator-mxnet/blob/dc3648bd910092c2ded2d67b6fb66d48d91f55ce/tests/python/gpu/test_operator_gpu.py#L628

looking into

azai91:
the difference is that the unit test does not use filter length that is divisible by 8. When the the input shape has a filter size that is divisible by 8, then mkldnn reorders the data and thus the check fails. the gpu tests uses filter length of 8.

tests/python/mkl/test_mkldnn.py
anirudh2290: @TaoLv do you mean tests/python/unittest/test_operator.py. The tests in tests/python/gpu/test_operator_gpu.py are only meant to run with gpu context.
I see that there are tests in tests/python/unittest/test_operator.py for deconv but they are also skipped because of flakiness. Doesnt seem like we have any coverage on the Deconvolution operator

tests/python/mkl/test_mkldnn.py
```diff
@@ -398,6 +398,23 @@ def softmax_forward(input_data, true_output):
...
+    kernel = (3, 3)
+    pad = (1, 1)
+    shape = (1, 256, 200, 233)
+    x = mx.sym.Variable('x')
+    w = mx.sym.Variable('w')
+    y = mx.sym.Deconvolution(data=x, weight=w, num_filter=num_filter, num_group=num_group, kernel=kernel, no_bias=True, pad=pad)
+    exe = y.simple_bind(ctx=mx.cpu(), x=shape, grad_req='null')
+    exe.arg_arrays[0][:] = np.random.normal(size=exe.arg_arrays[0].shape)
+    exe.arg_arrays[1][:] = np.random.normal(size=exe.arg_arrays[1].shape)
+    for i in range(10):
```
anirudh2290: nit: do we need this 10 times.

tests/python/mkl/test_mkldnn.py
```diff
@@ -398,6 +398,23 @@ def softmax_forward(input_data, true_output):
...
+        exe.forward(is_train=False)
+        o = exe.outputs[0]
+        t = o.asnumpy()
```
anirudh2290: can we just do exe.outputs[0].wait_to_read()

azai91: you're right. this is only dependent on the shape and not values.

TaoLv:
> the difference is that the unit test does not use filter length that is divisible by 8. When the the input shape has a filter size that is divisible by 8, then mkldnn reorders the data and thus the check fails. the gpu tests uses filter length of 8.

Nice catch. Do you mind adding new test shapes to the tests in `test_operator_gpu.py`? Then there is no need for us to add and maintain a new test case for MKL-DNN specific.

juliusshufan:
@azai91 The previously encountered issue can't be reproduced on applying your PR on the MXNET repo.

azai91:
what do you mean? the test changes don't catch the issue on the master branch which currently has the bug?

juliusshufan:
I can also reproduce the issue on an internal test case I am tracking, this case involving deconv, and the test case can pass with your fix.

anirudh2290:
@TaoLv from @juliusshufan 's comment this PR fixed his issue.

TaoLv:
Thanks, @azai91 @anirudh2290 @juliusshufan . Now I will close the corresponding issue.

PR #13046, opened 2018-10-30
[MXNET-1155] Add scala packageTest utility

zachgk:
## Description ##
This PR creates a scala packageTest utility to run the full test suite against any mxnet jar.  This can be used to verify that the packages produced for release and as part of the ci/cd run correctly on various environments (16.04, 18.04, centOS, amazon linux, etc.).

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

@lanking520 @nswamy @andrewfayres @yzhliu

PR #13450, opened 2018-11-29
[MXNET-862] Basic maven jenkins pipeline

zachgk:
## Description ##
This is a ~~WIP~~ PR ~~to share progress~~ on the maven nightly jenkins pipeline. The goal is to create a Jenkins for continuous deployment that will build scala jars, test them on multiple different environments, and then deploy them to the apache snapshot repository. It can also provide some setup for building out additional continuous deployment efforts for other platforms such as pip.

~~Note that this PR depends on https://github.com/apache/incubator-mxnet/pull/13046 and is rebased on top of it. Those changes are displayed as part of the first commit in the PR.~~

@nswamy @lanking520 @andrewfayres @piyushghai @marcoabreu @yzhliu 

### Essentials ###

### Changes ###
- [x] Build pipeline for build, test, deploy


zachgk:
@mxnet-label-bot add [Java, Scala, CI, Maven]

marcoabreu:
Lets wait until after the release before we move forward with this PR as discussed on dev@ :)

ci/Jenkinsfile_utils.groovy
```diff
@@ -18,18 +18,23 @@
...
     try {
       // Make sure wait long enough for api.github.com request quota. Important: Don't increase the amount of
       // retries as this will increase the amount of requests and worsen the throttling
       timeout(time: 15, unit: 'MINUTES') {
         checkout scm
-        sh 'git clean -xdff'
+        if(clean == 1)
+          clean = 'git clean -xdf'
+        else if(clean == 2)
+          clean = 'git clean -xdff'
```
piyushghai: Is there a difference between the ```clean``` command in clean == 1 and clean == 2 ?

scala-package/examples/pom.xml
```diff
@@ -119,6 +119,14 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
+        <executions>
```
piyushghai: Would  the child poms not inherit this execution stage from the parent pom file ?

scala-package/packageTest/README.md
```diff
@@ -0,0 +1,50 @@
+# MXNet Scala Package Test
```
piyushghai: This README is super helpful! 
Thanks @zachgk 💯

scala-package/packageTest/Makefile
```diff
@@ -0,0 +1,92 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+SCALA_VERSION_PROFILE := 2.11
+SCALA_VERSION := 2.11.8
+MXNET_VERSION := 1.3.1-SNAPSHOT
```
piyushghai: Should this version remain hardcoded here ?

Makefile
```diff
@@ -650,11 +662,20 @@ scalarelease-perform:
...
 	(cd $(ROOTDIR)/scala-package; \
-		mvn deploy -Papache-release,$(SCALA_PKG_PROFILE),$(SCALA_VERSION_PROFILE) \-DskipTests=true -Dcxx="$(CXX)" \
+		mvn deploy $(MAVEN_ARGS) -Papache-release,$(SCALA_PKG_PROFILE),$(SCALA_VERSION_PROFILE) \-DskipTests=true -Dcxx="$(CXX)" \
 		    -Dbuild.platform="$(SCALA_PKG_PROFILE)" \
 			-Dcflags="$(CFLAGS)" -Dldflags="$(LDFLAGS)" \
 			-Dlddeps="$(LIB_DEP) $(ROOTDIR)/lib/libmxnet.a")
 
+scaladeploylocal:
+	(cd $(ROOTDIR)/scala-package; \
+		mvn deploy $(MAVEN_ARGS) -Papache-release,$(SCALA_PKG_PROFILE),$(SCALA_VERSION_PROFILE) \-DskipTests=true -Dcxx="$(CXX)" \
```
piyushghai: Is there a reason why you are skipping the tests here on the deploy stage ?

Makefile
```diff
@@ -454,6 +454,10 @@ else
 	CFLAGS += -DMXNET_USE_LIBJPEG_TURBO=0
 endif
 
+ifeq ($(CI), 1)
+	MAVEN_ARGS := -B
```
piyushghai: Curious to know what this ```-B``` flag does :)

scala-package/packageTest/Makefile
zachgk: I will investigate using [1.3.1-SNAPSHOT,) to see if it works. Otherwise, I don't know if there are other options.

ci/Jenkinsfile_utils.groovy
marcoabreu: Please leave this as it is.

The workspace always has to be considered empty and reproducible. If you need to persist something in between stages, use the stashing mechanism. With this change you are circumventing a problem - that you are not persisting state properly - and it will blow up as soon as we have more than one restricted slave and job one gets scheduled onto a different slave than job two

zachgk: This change is not to persist something in between stages. During maven deploy, it attempts to move all code into a temp directory, rebuild, and deploy from that temp directory. The rebuilding calls "git submodule update" which breaks if the .git directory has already been cleaned with "-ff" since the submodule configuration is deleted. I keep the default as "clean -xdff" so that keeping the .git directory must be explicitly chosen and does not affect any other code.

marcoabreu: I think I'd like to have a deeper look into that, so far I'm not following

Roshrini:
@zachgk Can you rebase this PR?
@mxnet-label-bot Update [pr-work-in-progress]

zachgk:
I am currently thinking that since these are relatively large commits, I would split them into three different PRs that will create respectively a basic pipeline, a pipeline with multi-environment testing, and the final pipeline integrating the static linking code. After rebasing, I reduced the amount of code as part of this PR (it has some of commit 2), so it should be roughly equal to the code from 20 days ago. This also no longer has code from the packageTest utility because that was merged into master.

@mxnet-label-bot update [CI, Java, Maven, Scala, pr-awaiting-review]

zachgk:
As part of the later PRs, the changes to the current files will include:

- Add new Docker containers specific to publishing and a clean dependency-less environment for testing so the updates to ubuntu_scala will be moved
- The scala deployment will be moved from scala-package/dev to ci/publish/scala
- It will use the static building script instead of the scala-package/dev/compile-mxnet-backend.sh script for building the backend

Note that this PR forms a working pipeline that will build, test, and then deploy the package for ubuntu 16.04

ci/publish/Jenkinsfile
```diff
@@ -0,0 +1,105 @@
...
+
+node('restricted-utility') {
+  // Loading the utilities requires a node context unfortunately
+  checkout scm
+  utils = load('ci/Jenkinsfile_utils.groovy')
+}
+utils.assign_node_labels(utility: 'restricted-utility', linux_cpu: 'restricted-mxnetlinux-cpu', linux_gpu: 'restricted-mxnetlinux-gpu', linux_gpu_p3: 'restricted-mxnetlinux-gpu-p3', windows_cpu: 'restricted-mxnetwindows-cpu', windows_gpu: 'restricted-mxnetwindows-gpu')
+
+def nodeMap = ['cpu': NODE_LINUX_CPU, 'gpu': NODE_LINUX_GPU]
+def scalaOSMap = ['cpu': 'linux-x86_64-cpu', 'gpu': 'linux-x86_64-gpu']
```
piyushghai: Can you add a comment here as to why we are not doing MacOS through Jenkins ?

zachgk: ok

ci/Jenkinsfile_utils.groovy
zachgk: The maven release plugin messes with the project SCM (git). It first tries to create a new tag for the release and a new commit updating the version number located in the POM file. Then, when performing the release, it will checkout the code to a temporary directory and perform the release (building, testing, and deploying) from that new clean directory. This will break if the .git directory has been pruned. You can see docs for the maven release plugin at http://maven.apache.org/maven-release/maven-release-plugin/examples/prepare-release.html

ci/Jenkinsfile_utils.groovy
```diff
@@ -18,18 +18,25 @@
...
       timeout(time: 15, unit: 'MINUTES') {
         checkout scm
-        sh 'git clean -xdff'
+        if(clean == 0)
+          cleanCmd = 'git status'
+        else if(clean == 1)
+          cleanCmd = 'git clean -xdf'
+        else if(clean == 2)
+          cleanCmd = 'git clean -xdff'
+        sh cleanCmd
```
marcoabreu: Could you elaborate why this is needed? This "super brutal clean" as I like to call it, was requested by @lebeg and he said it's important. Maybe you two could have a discussion there :)

ci/publish/Jenkinsfile
```diff
@@ -0,0 +1,106 @@
...
+  checkout scm
+  utils = load('ci/Jenkinsfile_utils.groovy')
+}
+utils.assign_node_labels(utility: 'restricted-utility', linux_cpu: 'restricted-mxnetlinux-cpu', linux_gpu: 'restricted-mxnetlinux-gpu', linux_gpu_p3: 'restricted-mxnetlinux-gpu-p3', windows_cpu: 'restricted-mxnetwindows-cpu', windows_gpu: 'restricted-mxnetwindows-gpu')
+
+// CPU and GPU. OSX nodes are not currently supported by Jenkins
+def nodeMap = ['cpu': NODE_LINUX_CPU, 'gpu': NODE_LINUX_GPU]
+def scalaOSMap = ['cpu': 'linux-x86_64-cpu', 'gpu': 'linux-x86_64-gpu']
+
+def wrapStep(nodeToRun, workspaceName, step) {
```
marcoabreu: I love this generator model. Great job!

ci/publish/Jenkinsfile
```diff
@@ -0,0 +1,106 @@
...
+  return {
+    node(nodeToRun) {
+      ws("workspace/${workspaceName}") {
+        env.MAVEN_OPTS = "-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn" // Clean up maven logs
```
marcoabreu: Would it be possible to move this logic into the runtime functions? We generally try to keep Jenkinsfile limited to process flow design while logic is in the runtime functions. Otherwise, you can't reproduce the individual steps using the docker wrapper because certain variables (like this env var) are set outside of the docker process.

scala-package/dev/buildkey.py
```diff
@@ -0,0 +1,151 @@
...
+    except botocore.exceptions.ClientError as client_error:
+        if client_error.response['Error']['Code'] == 'ResourceNotFoundException':
+            name = (secret_key_name if get_secret_value_response
+                    else secret_creds_name)
+            logging.exception("The requested secret %s was not found", name)
+        elif client_error.response['Error']['Code'] == 'InvalidRequestException':
+            logging.exception("The request was invalid due to:")
+        elif client_error.response['Error']['Code'] == 'InvalidParameterException':
+            logging.exception("The request had invalid params:")
+        else:
```
marcoabreu: I know this is the default template of AWS SecretsManager, but could we raise always (instead of the else case) to halt the execution? I have made the experience that letting the process continue to then fail later obfuscated the error message and confuses some people. 

In this particular case, the function would return None while the caller expects a tuple. This then throws a confusion message which makes it harder to debug. wdyt?

scala-package/dev/buildkey.py
```diff
@@ -0,0 +1,151 @@
...
+        secret = get_secret_value_response['SecretString']
+        secret_dict = json.loads(secret)
+        secret_key = get_secret_key_response['SecretString']
+        return secret_dict, secret_key
+
+
+def importASC(key, gpgPassphrase):
+    filename = os.path.join(KEY_PATH, "key.asc")
+    with open(filename, 'w') as f:
+        f.write(key)
```
marcoabreu: Are we making sure that this file is deleted when the process finishes? Or is it local to the container?

scala-package/dev/buildkey.py
```diff
@@ -0,0 +1,151 @@
...
+    subprocess.run(['gpg2', '--batch', '--yes',
+                    '--passphrase=\"{}\"'.format(gpgPassphrase),
+                    "--import", "{}".format(filename)])
```
marcoabreu: Ha! Pro tip - I currently have to write a COE because I made the same mistake  - don't do that or you will leak credentials :P 

Subprocess.run will print the command (in this case including the password) if the execution fails. This then writes the password to the log and you will leak credentials. Is it possible to pass in the password using STDIN or some other redirection method?

I can link you the ticket privately if you're interested in further details.

scala-package/dev/buildkey.py
```diff
@@ -0,0 +1,151 @@
...
+    return str(result.stdout)[2:-3]
+
+
+def masterPSW(password):
+    with open(os.path.join(KEY_PATH, "settings-security.xml"), "w") as f:
+        f.write("<settingsSecurity>\n <master>{}</master>\n</settingsSecurity>"
+                .format(password))
+
+
+def severPSW(username, password, gpgPassphrase):
```
marcoabreu: what does "sever" mean? 🤔

lanking520: The Apache server

scala-package/dev/buildkey.py
lanking520: @marcoabreu Currently, we use fake gpg key in here as it is not necessary to have the real one in the deploy stage. Problems will appear in the release stage as I would not do it with CI. But still it is very useful information, please keep us updated if you find a better solution to this.

scala-package/dev/buildkey.py
lanking520: Nice catch, same applies to the hidden `settings.xml` in `.m2` folder

ci/Jenkinsfile_utils.groovy
zachgk: Actually, I can remove this because we are no longer going to use the maven-release-plugin following updates as part of #13626

ci/publish/Jenkinsfile
zachgk: Makes sense. I'll move it

scala-package/dev/buildkey.py
zachgk: Yeah, this sounds like a good improvement. I'll apply it in docker_cache as well for the sake of consistency

scala-package/dev/buildkey.py
zachgk: It is put in the ~ directory, which I think that makes it local to the container. Either way, I will add some rm statements for extra security.

scala-package/dev/buildkey.py
zachgk: @lanking520 We still use this to encrypt the passwords through maven, so this is certainly a problem we need to address. 

I worked around this using a combination of passing data through stdin and passing data using the linux expect program (because maven doesn't work with stdin)

scala-package/dev/buildkey.py
zachgk: @lanking520 "server" was misspelled

lanking520:
@marcoabreu Could you please review again?

scala-package/dev/build.sh
```diff
@@ -0,0 +1,36 @@
...
+
+# Setup Environment Variables
+# MAVEN_PUBLISH_OS_TYPE: linux-x86_64-cpu|linux-x86_64-gpu|osx-x86_64-cpu
+# export MAVEN_PUBLISH_OS_TYPE=linux-x86_64-cpu
+
+bash scala-package/dev/compile-mxnet-backend.sh $MAVEN_PUBLISH_OS_TYPE ./
+
+# Scala steps to deploy
+make scalapkg CI=1
+make scalaunittest CI=1
```
marcoabreu: The filename lets me assume the task of this script is to build, but it seems like a lot of testing is happening as well.

Generally, we try to split compilation and testing first to make it cleaner, and second to use the most appropriate instance type. Since you also made a test script, could we move these tests over there to keep the compilation jobs as lightweight as possible?

scala-package/dev/deploy.sh
```diff
@@ -0,0 +1,53 @@
...
+set -ex
+
+# Setup Environment Variables
+# MAVEN_PUBLISH_OS_TYPE: linux-x86_64-cpu|linux-x86_64-gpu|osx-x86_64-cpu
+# export MAVEN_PUBLISH_OS_TYPE=linux-x86_64-cpu
+
+
+if [[ $MAVEN_PUBLISH_OS_TYPE == "linux-x86_64-cpu" ]];
+then
+    MAKE_FLAGS="USE_BLAS=openblas USE_DIST_KVSTORE=1 ENABLE_TESTCOVERAGE=1"
```
marcoabreu: Do we want to compile with test coverage annotations for production builds?

scala-package/dev/deploy.sh
```diff
@@ -0,0 +1,53 @@
...
+elif [[ $MAVEN_PUBLISH_OS_TYPE == "linux-x86_64-gpu" ]]
+then
+    MAKE_FLAGS="USE_OPENCV=1 USE_BLAS=openblas USE_CUDA=1 USE_CUDA_PATH=/usr/local/cuda USE_CUDNN=1 SCALA_ON_GPU=1 SCALA_TEST_ON_GPU=1 USE_DIST_KVSTORE=1 ENABLE_TESTCOVERAGE=1"
```
marcoabreu: Same here

PR #13452, opened 2018-11-29
Get the correct include path in pip package

apeforest:
## Description ##
The include path in pip package and MXNet source has different relative path. This PR is to get any of the correct one.

## Comments ##
- This feature is needed in Horovod MXNet


apeforest:
@yuxihu @zhreshold Please help to review again. Thanks!

vandanavk:

python/mxnet/libinfo.py
```diff
@@ -97,9 +97,16 @@ def find_include_path():
...
     curr_path = os.path.dirname(os.path.abspath(os.path.expanduser(__file__)))
     incl_path = os.path.join(curr_path, '../../include/')
-    if not os.path.isdir(incl_path):
-        raise RuntimeError('Cannot find the MXNet include path.\n')
-    return incl_path
+    if os.path.isdir(incl_path):
+        # include path if build from source
+        return incl_path
+    else:
+        incl_path = os.path.join(curr_path, '../include/')
```
zhreshold: does pip package really have a include directory?

apeforest: Yes, thanks to @szha who added it several days ago.

You may find it at: <mxnet_package_location>/lib/python3.5/site-packages/mxnet/include

yuxihu: '../include/' should be 'include/'.

python/mxnet/libinfo.py
```diff
@@ -97,9 +97,16 @@ def find_include_path():
 
     curr_path = os.path.dirname(os.path.abspath(os.path.expanduser(__file__)))
     incl_path = os.path.join(curr_path, '../../include/')
```
yuxihu: I think we should check pip path first. Most of the cases we should be able to find it there. And it might be the preferred path, as in Carl's horovod instruction, the include files are also copied under python/mxnet.

python/mxnet/libinfo.py
apeforest: Thanks! Good catch!

python/mxnet/libinfo.py
apeforest: Good suggestion. Updated

python/mxnet/libinfo.py
```diff
@@ -96,10 +96,17 @@ def find_include_path():
             logging.warning("MXNET_INCLUDE_PATH '%s' doesn't exist", incl_from_env)
 
     curr_path = os.path.dirname(os.path.abspath(os.path.expanduser(__file__)))
-    incl_path = os.path.join(curr_path, '../../include/')
-    if not os.path.isdir(incl_path):
-        raise RuntimeError('Cannot find the MXNet include path.\n')
-    return incl_path
+    # include path in pip package
+    incl_path = os.path.join(curr_path, './include/')
```
yuxihu: I think you do not need './' before 'include/'. But it is ok to keep it.

python/mxnet/libinfo.py
```diff
@@ -96,10 +96,17 @@ def find_include_path():
...
+    if os.path.isdir(incl_path):
+        return incl_path
+    else:
+        # include path if build from source
+        incl_path = os.path.join(curr_path, '../../include/')
+        if os.path.isdir(incl_path):
+            return incl_path
+        else:
+            raise RuntimeError('Cannot find the MXNet include path: ' + incl_path + '\n')
```
yuxihu: do we need to have both incl_path in the error msg? Something like:

    # include path in pip package
    pip_incl_path = os.path.join(curr_path, './include/')
    if os.path.isdir(pip_incl_path):
        return pip_incl_path
    else:
        # include path if build from source
        src_incl_path = os.path.join(curr_path, '../../include/')
        if os.path.isdir(src_incl_path):
            return src_incl_path
        else:
            raise RuntimeError('Cannot find the MXNet include path in either ' + pip_incl_path + ' or ' + src_incl_path + '\n')

python/mxnet/libinfo.py
apeforest: updated

python/mxnet/libinfo.py
apeforest: thanks. updated.

apeforest:

PR #13456, opened 2018-11-29
Improve CCache handling

marcoabreu:
This PR improves our ccache access. Particularly, we are now able to able to cache NVCC calls (which are super expensive) and also add caching to the compilation of submodules (especially ps-lites' submodule zeroMQ likes to hardcode its compilers).

Just to give some numbers:
- ARMv8 improved from 35s to 14s.
- Dependency compilation on cpu and gpu got recuced from 7 minutes to 10 seconds
- TODO: Expand

Full NVCC Ccache support for CMake is blocked by #13459

### DIFF
--- a/ci/docker/install/centos7_adduser.sh
+++ b/ci/docker/install/centos7_adduser.sh
@@ -34,4 +34,9 @@ then
     mkdir /work/mxnet
     mkdir /work/build
     chown -R jenkins_slave /work/
+
+    # Later on, we have to override the links because underlying build systems ignore our compiler settings. Thus,
+    # we have to give the process the proper permission to these files. This is hacky, but unfortunately 
+    # there's no better way to do this without patching all our submodules.
+    chown -R jenkins_slave /usr/local/bin
 fi
--- a/ci/docker/runtime_functions.sh
+++ b/ci/docker/runtime_functions.sh
@@ -39,32 +39,34 @@ clean_repo() {
 build_ccache_wrappers() {
     set -ex
 
-    rm -f cc
-    rm -f cxx
+    # Recommended by CCache: https://ccache.samba.org/manual.html#_run_modes
+    # Add to the beginning of path to ensure this redirection is picked up instead
+    # of the original ones. Especially CUDA/NVCC appends itself to the beginning of the
+    # path and thus this redirect is ignored. This change fixes this problem
+    # This hacky approach with symbolic links is required because underlying build
+    # systems of our submodules ignore our CMake settings. If they use Makefile,
+    # we can't influence them at all in general and NVCC also prefers to hardcode their
+    # compiler instead of respecting the settings. Thus, we take this brutal approach
+    # and just redirect everything of this installer has been called.
+    # In future, we could do these links during image build time of the container.
+    # But in the beginning, we'll make this opt-in. In future, loads of processes like
+    # the scala make step or numpy compilation and other pip package generations
+    # could be heavily sped up by using ccache as well.
+    export PATH=/usr/local/bin:$PATH
+    ln -s ccache /usr/local/bin/gcc
+    ln -s ccache /usr/local/bin/g++
+    ln -s ccache /usr/local/bin/g++-8
+    ln -s ccache /usr/local/bin/gcc-8
+    ln -s ccache /usr/local/bin/nvcc
+    ln -s ccache /usr/local/bin/clang++-3.9
+    ln -s ccache /usr/local/bin/clang-3.9
+    ln -s ccache /usr/local/bin/clang++-5.0
+    ln -s ccache /usr/local/bin/clang-5.0
+    ln -s ccache /usr/local/bin/clang++-6.0
+    ln -s ccache /usr/local/bin/clang-6.0
 
-    touch cc
-    touch cxx
-
-    if [ -z ${CC+x} ]; then
-        echo "No \$CC set, defaulting to gcc";
-        export CC=gcc
-    fi
-
-    if [ -z ${CXX+x} ]; then
-       echo "No \$CXX set, defaulting to g++";
-       export CXX=g++
-    fi
-
-    # this function is nessesary for cuda enabled make based builds, since nvcc needs just an executable for -ccbin
-
-    echo -e "#!/bin/sh\n/usr/local/bin/ccache ${CC} \"\$@\"\n" >> cc
-    echo -e "#!/bin/sh\n/usr/local/bin/ccache ${CXX} \"\$@\"\n" >> cxx
-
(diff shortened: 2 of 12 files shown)


marcoabreu:
CCache support for NVCC in CMake needs an upgrade. Tracked at https://github.com/apache/incubator-mxnet/issues/13459

vandanavk:
@mxnet-label-bot add [CI, pr-awaiting-review]

marcoabreu:
Ready for review. I will remove the Jenkinsfile changes when I'm done

ci/docker/install/centos7_adduser.sh
```diff
@@ -34,4 +34,9 @@ then
     mkdir /work/mxnet
     mkdir /work/build
     chown -R jenkins_slave /work/
+
+    # Later on, we have to override the links because underlying build systems ignore our compiler settings. Thus,
+    # we have to give the process the proper permission to these files. This is hacky, but unfortunately 
+    # there's no better way to do this without patching all our submodules.
+    chown -R jenkins_slave /usr/local/bin
```
larroy: Not clear, even with comment I don't understand why. Why can't we have some other folder added to the PATH?

ci/docker/runtime_functions.sh
```diff
@@ -39,32 +39,59 @@ clean_repo() {
...
+    ln -s ccache /tmp/ccache-redirects/g++
+    ln -s ccache /tmp/ccache-redirects/g++-8
+    ln -s ccache /tmp/ccache-redirects/nvcc
+    ln -s ccache /tmp/ccache-redirects/clang++-3.9
+    ln -s ccache /tmp/ccache-redirects/clang-3.9
+    ln -s ccache /tmp/ccache-redirects/clang++-5.0
+    ln -s ccache /tmp/ccache-redirects/clang-5.0
+    ln -s ccache /tmp/ccache-redirects/clang++-6.0
+    ln -s ccache /tmp/ccache-redirects/clang-6.0
+    ln -s ccache /usr/local/bin/gcc
```
larroy: Why don't you set some other folder in the path????

ci/docker/runtime_functions.sh
```diff
@@ -129,6 +158,7 @@ build_armv6() {
 
     # We do not need OpenMP, since most armv6 systems have only 1 core
 
+    build_ccache_wrappers
```
larroy: isn't this working already?

ci/docker/runtime_functions.sh
```diff
@@ -637,9 +685,11 @@ build_ubuntu_gpu_cmake_mkldnn() {
 build_ubuntu_gpu_cmake() {
     set -ex
     cd /work/build
+    build_ccache_wrappers
```
larroy: Why we need both?

larroy:
Nice change! thanks!

ci/docker/install/centos7_adduser.sh
marcoabreu: We do, that's why I created /tmp/ccache-redirects. The problem is that in some places (especially submodules), that paths have been hardcoded. That's why we have to do both :(

ci/docker/runtime_functions.sh
marcoabreu: I did. But sometimes (in submodules) /usr/local/bin/gcc is referenced directly. In that case, my folder /tmp/ccache-redirects would be ignored.

ci/docker/runtime_functions.sh
marcoabreu: Not for submodules unfortunately.

ci/docker/runtime_functions.sh
marcoabreu: Because NVCC in cmake below 3.10 doesn't redirect the calls to g++ to ccache. Thus, we have to do both. Also, submodules - again :(

marcoabreu:
Thank you :)

Sorry these things look really hacky. Unfortunately, a lot of the underlying build systems in our submodules are not adhering to best practices and often hardcode some things.

ci/docker/runtime_functions.sh
larroy: Which submodules? Shouldn't we better fix them? This seems hacky, even though is great that you got it to work.

larroy:
Thanks for your follow up. Can you list which submodules are the offenders so we can ponder if we should better fix the submodule or use the hacky solution? Thx.

ci/docker/runtime_functions.sh
marcoabreu: Ps-lite and ZeroMQ are the first ones that stuck to my mind. I didn't benchmark every single one.

In total, we spend about 5-8 minutes compiling submodules. After this PR it's 20 seconds.

marcoabreu:
Ps-lite and ZeroMQ are the first ones that stuck to my mind. I didn't benchmark every single one.

In total, we spend about 5-8 minutes compiling submodules. After this PR it's 20 seconds.

ci/docker/runtime_functions.sh
larroy: Have you tried export CXX and CC? I don't see any references to /usr/local/bin/gcc in our source tree.

marcoabreu: Yes we used that method previously but zeromq ignored it for example. Have you searched for references to "GCC"? Zeromq is downloaded dynamically during the build (it's a zip file), so a grep would not find it.

Well that's the recommended way how ccache writes it in their guide. In the end, you can't control all build systems

marcoabreu: https://ccache.samba.org/manual/latest.html#_run_modes

roywei:
@marcoabreu Thanks for the contribution, what do we need to change in PS-lite?
@mxnet-label-bot add[CI]

ci/docker/runtime_functions.sh
leezu: @larroy @marcoabreu based on the update to cmake 3.13, now `build_ccache_wrappers` is only needed due to some submodules not handling the `CMAKE_CXX_COMPILER_LAUNCHER` correctly?

larroy:
@josephevans please review, you said you got ccache to work with nvcc.

ci/docker/runtime_functions.sh
larroy: those submodules should use CC and CXX variables, otherwise they are broken. We would need to check.

marcoabreu: I'd recommend to verify. Some submodules used to ignore the standards. Cmake as well as make paths should be tested - sometimes, cmake and make get intermingled

PR #12182, opened 2018-08-15
[MXNET-698] Correct train-metric log to reflect epoch metric

vandanavk:
## Description ##
The log `self.logger.info('Epoch[%d] Train-%s=%f', epoch, name, val)` gives the user the impression that the metric printed is for the entire epoch, which is misleading. When `auto_reset=False`, Train-\<metric\> is the average for the entire epoch and when `auto_reset=True`, Train-\<metric\> reflects the metric value for the remaining batches in the epoch.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- Track epoch metric value separately
- Print batch numbers for which Speedometer prints metric value

**Before**

```
INFO:root:Epoch[0] Batch [100]  Speed: 45690.13 samples/sec     accuracy=0.772123
INFO:root:Epoch[0] Batch [200]  Speed: 50611.24 samples/sec     accuracy=0.898594
INFO:root:Epoch[0] Batch [300]  Speed: 54257.01 samples/sec     accuracy=0.921562
INFO:root:Epoch[0] Batch [400]  Speed: 57961.52 samples/sec     accuracy=0.934063
INFO:root:Epoch[0] Batch [500]  Speed: 44537.19 samples/sec     accuracy=0.931719
INFO:root:Epoch[0] Batch [600]  Speed: 48699.48 samples/sec     accuracy=0.935469
[...]

PR #13471, opened 2018-11-30
Optimization of metric evaluation

ptrendx:
## Description ##
Currently metrics are mostly evaluated on CPU using NumPy. Due to Python GIL, they are evaluated in single thread, sequentially, which may become a problem once number of used GPUs is large enough.

### Essentials ###

### Changes ###
- [x] Changed TopKAccuracy metric implementation from using numpy.argsort to numpy.argpartition, which from my experiments on NDArrays of shape (208,1000) and top_k=5 is ~4x faster 
- [x] Added global statistics to metrics. We noticed the problem with metrics speed after PR #12182 which introduced computing the metrics twice (once for giving the immediate values for accuracy and once for computing accuracy over the entire epoch). This is wasteful, since the calculations needed are exactly the same in this case. This PR introduces additional fields in EvalMetric class (global_sum_metric and global_num_inst), accessors for them (get_global and get_global_name_values) and function reset_local which performs reset only on non-global versions of statistics. It also modifies all current metrics to be able to use the global statistics. That way, while the code is backward compatible (one can still just use get/get_name_values/reset functions), it also eliminates overhead introduced by calculating statistics twice in fit function.

CC @vandanavk for comments.


vandanavk:
@mxnet-label-bot add [Metric, pr-awaiting-review]

Thanks @ptrendx, I'll have a look

python/mxnet/metric.py
```diff
@@ -91,6 +91,10 @@ def __init__(self, name, output_names=None,
         self.output_names = output_names
         self.label_names = label_names
         self._kwargs = kwargs
+        if "has_global_stats" in kwargs:
+            self._has_global_stats = kwargs["has_global_stats"]
+        else:
+            self._has_global_stats = False
```
vandanavk: `self._has_global_stats = kwargs.get("has_global_stats", False)` ?

python/mxnet/metric.py
```diff
@@ -667,18 +756,25 @@ def update(self, labels, preds):
...
-            self.sum_metric += self.metrics.fscore
+            self.sum_metric += fscore
+            self.global_sum_metric += fscore
             self.num_inst += 1
+            self.global_num_inst += 1
             self.metrics.reset_stats()
         else:
-            self.sum_metric = self.metrics.fscore * self.metrics.total_examples
+            self.sum_metric = fscore * self.metrics.total_examples
+            self.global_sum_metric = fscore * self.metrics.total_examples
```
vandanavk: `self.sum_metric = self.global_sum_metric =` ?

python/mxnet/metric.py
```diff
@@ -667,18 +756,25 @@ def update(self, labels, preds):
...
             self.num_inst = self.metrics.total_examples
+            self.global_num_inst = self.metrics.total_examples
```
vandanavk: similarly here

python/mxnet/metric.py
```diff
@@ -768,18 +865,25 @@ def update(self, labels, preds):
...
+            self.global_sum_metric += matthewscc
             self.num_inst += 1
+            self.global_num_inst += 1
             self._metrics.reset_stats()
         else:
-            self.sum_metric = self._metrics.matthewscc * self._metrics.total_examples
+            self.sum_metric = matthewscc * self._metrics.total_examples
+            self.global_sum_metric = matthewscc * self._metrics.total_examples
             self.num_inst = self._metrics.total_examples
+            self.global_num_inst = self._metrics.total_examples
```
vandanavk: similarly here

python/mxnet/metric.py
ptrendx: done (with pop to not keep the has_global_stats key in kwargs, to not screw with deserialization later.)

python/mxnet/metric.py
ptrendx: done

python/mxnet/metric.py
ptrendx: done.

python/mxnet/metric.py
ptrendx: done.

ptrendx:
I tested train_imagenet.py, tools/parse_log.py and  both auto_reset values for Speedometer.

eric-haibin-lin:
@ptrendx thanks for the PR. Do you mind elaborating a bit more how this PR avoids GIL/speeds up metric evaluation?

ptrendx:
It does not avoid GIL, I just do less work in Python -
 - topk evaluation using numpy argpartition vs argsort is faster (and you don't need to do full sort to get top k elements)
 - currently metrics are calculated twice during training to get both local and per-epoch results - this PR makes it so they are evaluated once, and the result is used for both per-batch and per epochs statistics.

lupesko:
Thanks for the contribution @ptrendx !
Adding @sandeep-krishnamurthy to help review and merge.

python/mxnet/metric.py
```diff
@@ -883,6 +983,17 @@ def get(self):
...
 
+    def get_global(self):
+        """Returns the current global evaluation result.
+
+        Returns
+        -------
+        Tuple of (str, float)
+            Representing name of the metric and evaluation result.
+        """
+        num = self.global_num_inst if self.global_num_inst > 0 else float('nan')
```
sandeep-krishnamurthy: return float('nan')?

python/mxnet/metric.py
```diff
@@ -164,6 +173,24 @@ def get(self):
...
+        values : list of float
+           Value of the evaluations.
+        """
+        if self._has_global_stats:
+            if self.global_num_inst == 0:
+                return (self.name, float('nan'))
+            else:
+                return (self.name, self.global_sum_metric / self.global_num_inst)
+        else:
+            return self.get()
```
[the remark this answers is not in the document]
ptrendx: I'm not sure if that is possible - doing this would break fully custom metrics (other than subclasses of `CustomMetric` class, for which I added support) that did not implement global stats.

python/mxnet/metric.py
```diff
@@ -325,6 +371,15 @@ def reset(self):
...
             pass
 
+    def reset_local(self):
+        """Resets the local portion of the internal evaluation results
+        to initial state."""
+        try:
+            for metric in self.metrics:
+                metric.reset_local()
+        except AttributeError:
+            pass
```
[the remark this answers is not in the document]
ptrendx: This is not added in this PR and I'm not sure myself why this is needed here (this function is basically a copy of the `reset` function but calls `reset_local` on children instead of `reset`).

python/mxnet/metric.py
```diff
@@ -423,8 +501,11 @@ def update(self, labels, preds):
 
             check_label_shapes(label, pred_label)
 
-            self.sum_metric += (pred_label == label).sum()
+            num_correct = (pred_label == label).sum()
+            self.sum_metric += num_correct
+            self.global_sum_metric += num_correct
```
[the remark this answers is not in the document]
ptrendx: They will be different when you call `reset_local` (which is called from `Speedometer` for example) - this will reset the local versions of metrics while keeping the global versions intact (which was the point of this PR - to enable computing both per-batch and per-epoch statistics using only single computation).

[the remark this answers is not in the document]
ptrendx: That said, this comment made me think about this again and I found a bug in how I handle global statistics in `F1` and `MCC` metrics - fix and tests incoming. Thank you!

python/mxnet/metric.py
```diff
@@ -487,7 +569,7 @@ def update(self, labels, preds):
 
         for label, pred_label in zip(labels, preds):
             assert(len(pred_label.shape) <= 2), 'Predictions should be no more than 2 dims'
-            pred_label = numpy.argsort(pred_label.asnumpy().astype('float32'), axis=1)
+            pred_label = numpy.argpartition(pred_label.asnumpy().astype('float32'), -self.top_k)
```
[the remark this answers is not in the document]
ptrendx: Sure, will do.

python/mxnet/metric.py
ptrendx: Oh, this change got here by accident - I will revert it. Good catch, thank you!

ptrendx: When I added tests, it turned out that changing this is actually necessary (I did change it slightly differently though, to match the `get` function from a base class) to avoid floating point division by 0 exception.

ptrendx:
I made fixes and added test for all metrics. @sandeep-krishnamurthy please review again.

PR #13479, opened 2018-11-30
[v1.4.x] Get the correct include path in pip package (#13452)

apeforest:
* add find_include_path API

* address reviewer comment

* change return type from list to string

* add unit test

* address reviewer comment

* address reviewer comment

* address reviewer comment

* address reviewer comment

* fix include path problem in pip package

* add comment

* fix lint error

* address reviewer comment

* address reviewer comment

## Description ##
This PR fixes a reporting API that returns the include header path in pip package. The earlier version only returns the include path when MXNet is built from source. The relative path is different if MXNet is built from pip.

This PR is needed by Horovod MXNet integration. The API only returns the include path so it should not impact MXNet performance or functionality.

### DIFF
--- a/python/mxnet/libinfo.py
+++ b/python/mxnet/libinfo.py
@@ -96,10 +96,18 @@ def find_include_path():
             logging.warning("MXNET_INCLUDE_PATH '%s' doesn't exist", incl_from_env)
 
     curr_path = os.path.dirname(os.path.abspath(os.path.expanduser(__file__)))
-    incl_path = os.path.join(curr_path, '../../include/')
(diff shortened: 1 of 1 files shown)


lebeg:
Hi @apeforest,

The description is a bit misleading, can you add some information about why this is needed?

Also, if the PR is towards the v1.4.x branch, it would be good if you could add (v1.4.x) to the title.

vandanavk:

apeforest:
@lebeg Thanks for your suggestion. I have updated the PR title and description.

PR #13496, opened 2018-11-30
Optimize C++ API

zhaoyao73:
Pass parameter with reference instead of value.
Add const as well as it is not changed.

## Description ##
Optimize some functions to pass reference instead of value. This could avoid unnecessary construction/destruction and copy.

### Essentials ###
-  tiny changes
- did a successful build
  Didn't find the API doc in mxnet.io/doxygen

## Comments ##
I am not sure I could run a sanity test with it or how to run a unit test/sanity test against the change. But pass value -> pass reference, should be transparent to C++, although not like C's pass value->pass pointer.

### DIFF
--- a/cpp-package/example/utils.h
+++ b/cpp-package/example/utils.h
@@ -42,7 +42,7 @@ bool check_datafiles(const std::vector<std::string> &data_files) {
   return true;
   }
 
-bool setDataIter(MXDataIter *iter , std::string useType,
+bool setDataIter(MXDataIter *iter , const std::string &useType,
               const std::vector<std::string> &data_files, int batch_size) {
     if (!check_datafiles(data_files))
         return false;
--- a/cpp-package/include/mxnet-cpp/operator.h
+++ b/cpp-package/include/mxnet-cpp/operator.h
@@ -86,7 +86,7 @@ class Operator {
   * \param symbol the input symbol
   * \return reference of self
   */
-  Operator &SetInput(const std::string &name, Symbol symbol);
+  Operator &SetInput(const std::string &name, const Symbol &symbol);
   /*!
   * \brief add an input symbol
   * \param symbol the input symbol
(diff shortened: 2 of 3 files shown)


vandanavk:
@mxnet-label-bot add [C++, pr-awaiting-review]

@leleamol for review

cpp-package/example/utils.h
```diff
@@ -42,7 +42,7 @@ bool check_datafiles(const std::vector<std::string> &data_files) {
   return true;
   }
 
-bool setDataIter(MXDataIter *iter , std::string useType,
+bool setDataIter(MXDataIter *iter , const std::string &useType,
               const std::vector<std::string> &data_files, int batch_size) {
```
stu1130:
```suggestion
               const std::vector<std::string> &data_files, const int batch_size) {
```
nit: could this be better?

zhaoyao73: no, you don't need to specify "const int" for a "int", as C or C++ passing value, even you changed the batch_size in the function(it basically it is a copy from original one), it doesn't change the original batch_size from the caller. So normally nobody will use const for a int or any primitive type.

zhaoyao73:
any idea why centos-cpu still not done?
Is it the only thing need to be done?

zhaoyao73:
@marcoabreu May I ask why the centos-cpu still not done?

marcoabreu:
Seems like it hasn't been kicked off. Please rebase to start another build

zhaoyao73:
do I need to do anything?

marcoabreu:
@zhaoyao73 I have kicked off a build on your behalf. Sorry for the inconveniences.

marcoabreu:
Yeah, just make a new commit and it will automatically start a build

larroy:

zhaoyao73:
weird, look at this PR it is still pr-awaiting-review, although I see @larrroy put the pr-awaiting-merge?

PR #12946, opened 2018-10-23
[WIP] Onnx export API optional args and additional operator support, Fixes #12682

Roshrini:
## Description ##
ONNX export api takes optional arguments out_label and out_shapes
Adding operators - LogisticRegressionOutput, MakeLoss, BlockGrad, _rminus_scalar, r_div_scalar, identity

https://github.com/apache/incubator-mxnet/issues/12925

WIP - Need to add:
- Test Cases for additional operators
- Test case for multiple output model
- Add comments

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #13511, opened 2018-12-02
Fix use-before-assignment in convert_dot

zhipeng-jia:
## Description ##
Fix use-before-assignment bug of input_node_a and input_node_b when trans_a/trans_b is False.


vandanavk:
@mxnet-label-bot add [ONNX, pr-awaiting-review]

@Roshrini for review. She had a fix for this in PR https://github.com/apache/incubator-mxnet/pull/12946

PR #13519, opened 2018-12-03
Fix exception handling api doc

anirudh2290:
## Description ##
There were some issues with MXNet users because of not being aware of waitall not rethrowing exceptions. It was documented as a limitation in exception handling doc here: https://mxnet.incubator.apache.org/architecture/exception_handling.html?highlight=exception#limitation but not in the API doc.
Adding it to the API doc.

### Essentials ###

### Changes ###
- [x] API doc change

### DIFF
--- a/python/mxnet/ndarray/ndarray.py
+++ b/python/mxnet/ndarray/ndarray.py
@@ -157,11 +157,11 @@ def waitall():
     """Wait for all async operations to finish in MXNet.
 
     This function is used for benchmarking only.
-    Rethrowing exceptions as part of mx.nd.waitall is not supported.
-    If your code has exceptions, waitall can cause silent failures.
-    Avoid waitall in your code/precede it with wait_to_read for the outputs
-    unless you are confident about your code not throwing an exception in
-    any scenario.
+    .. warning::
+    If your code has exceptions, `waitall` can cause silent failures.
+    For this reason you should avoid `waitall` in your code.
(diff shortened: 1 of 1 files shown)


python/mxnet/ndarray/ndarray.py
```diff
@@ -157,6 +157,11 @@ def waitall():
     """Wait for all async operations to finish in MXNet.
 
     This function is used for benchmarking only.
```
nswamy: Not many people will understand what Rethrowing exception means. I suggest the below:

. warning:: This API is for internal purpose  for benchmarking) , it does not do any error handling and can cause silent failures. Use it only if you are confident that your code error free, suggested to call `wait_to_read` on all outputs after `waitall`.

python/mxnet/ndarray/ndarray.py
```diff
@@ -157,6 +157,11 @@ def waitall():
...
+    Rethrowing exceptions as part of mx.nd.waitall is not supported.
```
aaronmarkham:
```suggestion
. warning:: 
    If your code has exceptions, `waitall` can cause silent failures.
    For this reason you should avoid `waitall` in your code. 
    Use it only if you are confident that your code is error free. 
    Then make sure you call `wait_to_read` on all outputs after `waitall`.
```

anirudh2290:
Thanks @aaronmarkham @nswamy . Updated the API doc.

PR #13529, opened 2018-12-04
Improve dev_menu usability, local build and virtualenv

larroy:
## Description ##

Adds a convenient `./dev_menu.py build` command which does a local build with the settings defined in cmake_options.yaml

Adds capability to create a local environment where built mxnet is installed for testing & other tasks.

### Essentials ###

### DIFF
--- a/ci/requirements_mxnet_virtualenv.txt
+++ b/ci/requirements_mxnet_virtualenv.txt
@@ -1,0 +1,6 @@
+opencv-python
+ipython
+matplotlib
+nose
+nose-timer
+mock
--- a/ci/requirements_mxnet_virtualenv.txt
+++ b/ci/requirements_mxnet_virtualenv.txt
@@ -1,6 +1,0 @@
-opencv-python
-ipython
-matplotlib
-nose
-nose-timer
-mock
--- a/dev_menu.py
+++ b/dev_menu.py
@@ -103,7 +103,7 @@ def create_virtualenv(venv_exe, pyexe, venv) -> None:
     # Activate virtualenv in this interpreter
     exec(open(activate_this_py).read(), dict(__file__=activate_this_py))
     check_call(['pip', 'install', '--upgrade','--force-reinstall', '-e', 'python'])
-    check_call(['pip', 'install', '-r', 'ci/requirements_mxnet_virtualenv.txt'])
+    check_call(['pip', 'install', '-r', 'tests/requirements.txt'])
 
 def create_virtualenv_default():
     create_virtualenv('virtualenv', DEFAULT_PYTHON, DEFAULT_PYENV)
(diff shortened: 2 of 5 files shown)


larroy:
@mxnet-label-bot  [pr-awaiting-review]

larroy:

aaronmarkham:
The menu is nice. 
Is defaulting to CUDA=ON the preferred way? I thought OFF was the default with make right now...
I was a little confused with the mixing of cmake_options.yaml and cmake_options.yml. Can we have just one reference?

dev_menu.py
```diff
@@ -128,6 +151,9 @@ def clip(x, mini, maxi):
 
 @retry((ValueError, RuntimeError), 3, delay_s = 0)
 def show_menu(items: List[str], header=None) -> int:
+    print()
+    print('-- MXNet dev menu --')
+    print()
```
jlcontreras: Nit: this would look better to me in a single line
```suggestion
    print('\n-- MXNet dev menu --\n')
```

larroy:
@aaronmarkham thanks for the feedback. We can default to cuda OFF. Where are the duplicated references from yml  yaml? this must have been a typing issue. Any additional suggestions?

ci/requirements_mxnet_virtualenv.txt
```diff
@@ -0,0 +1,6 @@
+opencv-python
```
marcoabreu: Are these requirements duplicated? I feel like they are pretty generic and should be in another place. could you check if there's an existing requirements.txt that we can use? otherwise, we have too many requirements files which makes us prone to miss one

larroy: There are no suitable file with such requirements other than ci/docker/install/ubuntu_tutorials.sh

marcoabreu: https://github.com/apache/incubator-mxnet/blob/master/tests/requirements.txt

marcoabreu: We do need opencv for the tests afaik

larroy: They are working right? Why should I change the requirements for unit tests on this PR?  it could introduce errors in CI like last time. I think that's better done in a separate PR. Is your concern addressed?

marcoabreu: We don't use that linked requirements.txt in our CI. Instead, we install OpenCV and all that as part of our docker setup. 

If it would introduce errors, we would see it as part of the PR validation. Please, add your requirements to the linked requirements file and delete the one you created since they are redundant. The linked file is exactly for the purpose of providing an environment to run tests, which is also the purpose of your new requirements file

larroy: This is going to have implications on other places, in qemu we are installing those with:

81:    check_call(['sudo', 'pip3', 'install', '-r', 'mxnet/tests/requirements.txt'])

Any additional requirement can take a lot of time to compile in QEMU, I don't want to add requirements to the test files that are not needed.

I'm against the change that you propose because this reason.

larroy:
@marcoabreu  I changed what you requested, can we merge please if you don't have any more concerns?

Issue #13342, opened 2018-11-20
inv_khatri_rao.OneInputMatrixTransposed segfaults on ARM


larroy:
## Description

inv_khatri_rao.OneInputMatrixTransposed segfaults on ARM

## Environment info (Required)

ARMv7 using QEMU

mxnet_unit_tests --gtest_filter="inv_khatri_rao.OneInputMatrixTransposed"

reproduction info is sent in a separate PR:  #13334

vdantu:
@mxnet-label-bot add [C++, Test, ARM, Bug]

larroy:
Theres a crash because lapack is no available.

369+>  MXNET_LAPACK_UNAVAILABLE(dposv)

Crash when accessing the stream producing the fatal error.

```
(gdb) bt
#0  0xb68de2e6 in std::ostream::sentry::sentry(std::ostream&) () from /usr/lib/arm-linux-gnueabihf/libstdc++.so.6
#1  0xb68de794 in std::basic_ostream<char, std::char_traits<char> >& std::__ostream_insert<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, c
har const*, int) () from /usr/lib/arm-linux-gnueabihf/libstdc++.so.6
#2  0xb68dea6c in std::basic_ostream<char, std::char_traits<char> >& std::operator<< <std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*)
 () from /usr/lib/arm-linux-gnueabihf/libstdc++.so.6
#3  0x013d7034 in mxnet_lapack_dposv(...) () at /work/mxnet/src/operator/contrib/../c_lapack_api.h:369
```

leleamol:
@mxnet-label-bot update [ARM, Bug, Test]

PR #13535, opened 2018-12-04
[MXNET-244] Work around likely compiler bug on nested inlines and temporary acces…

larroy:
…s to stream

## Description ##

This fixes a segfault in ARMv7 on throwing a fatal error on the lapack undefined functions. 

Addresses  #13342  
We don't crash as before, the test fails which is expected.

I think this is a bug in the compiler due to the temporary LogMessageFatal created on unwrapping the macro https://github.com/apache/incubator-mxnet/blob/master/src/operator/c_lapack_api.h#L369 and too many inlined functions. The segfault is inside the ostream in libc, could be related to the lifetime of the temporary variable with a reference to the stream which is a very corner case of temporary variable lifetime.

Having the functions not inlined solves the problem.

```
qemu@qemu:~/mxnet/build/tests$ ./mxnet_unit_tests --gtest_filter="inv_khatri_rao.OneInputMatrixTransposed"

Note: Google Test filter = inv_khatri_rao.OneInputMatrixTransposed
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from inv_khatri_rao
[ RUN      ] inv_khatri_rao.OneInputMatrixTransposed
unknown file: Failure
C++ exception with description "[13:27:04] /work/mxnet/src/operator/c_lapack_api.cc:42: MXNet build without lapack. Function dposv is not available.

Stack trace returned 10 entries:
[bt] (0) ./mxnet_unit_tests(dmlc::StackTrace()+0x40) [0x1291afc]
[bt] (1) ./mxnet_unit_tests(dmlc::LogMessageFatal::~LogMessageFatal()+0x40) [0x1291e2c]
[bt] (2) ./mxnet_unit_tests(mxnet_lapack_dposv(...)+0x88) [0x19466dc]
[bt] (3) ./mxnet_unit_tests(int MXNET_LAPACK_posv<double>(int, char, int, int, double*, int, double*, int)+0x58) [0x13d7070]
[bt] (4) ./mxnet_unit_tests(void mxnet::op::inv_khatri_rao<double>(mshadow::Tensor<mshadow::cpu, 2, double>, std::vector<mshadow::Tensor<mshadow::cpu, 2, double>, std::allocator<mshadow::Tensor<mshadow::cpu, 2, double> > > const&, bool)+0x394) [0x13da000]
[bt] (5) ./mxnet_unit_tests(mxnet::op::inv_khatri_rao_OneInputMatrixTransposed_Test::TestBody()+0x238) [0x13cf0e8]
[bt] (6) ./mxnet_unit_tests(void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*)+0x74) [0x1491b50]
[bt] (7) ./mxnet_unit_tests(void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*)+0x4c) [0x148bb10]
[bt] (8) ./mxnet_unit_tests(testing::Test::Run()+0xe8) [0x146d9f8]
[bt] (9) ./mxnet_unit_tests(testing::TestInfo::Run()+0x138) [0x146e310]

" thrown in the test body.
[  FAILED  ] inv_khatri_rao.OneInputMatrixTransposed (70 ms)
[----------] 1 test from inv_khatri_rao (74 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (90 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] inv_khatri_rao.OneInputMatrixTransposed

 1 FAILED TEST
```

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/src/operator/c_lapack_api.cc
+++ b/src/operator/c_lapack_api.cc
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "c_lapack_api.h"
+
+#if (MSHADOW_USE_MKL && MXNET_USE_LAPACK)
+#elif MXNET_USE_LAPACK
+#else
+  // Define compilable stubs.
+  #define MXNET_LAPACK_CWRAPPER1(func, dtype) \
+  int MXNET_LAPACK_##func(int matrix_layout, char uplo, int n, dtype* a, int lda) { \
+    LOG(FATAL) << "MXNet build without lapack. Function " << #func << " is not available."; \
+    return 1; \
+  }
+
+  #define MXNET_LAPACK_CWRAPPER2(func, dtype) \
+  int MXNET_LAPACK_##func(int matrix_layout, int m, int n, dtype* a, \
+                                 int lda, dtype* tau, dtype* work, int lwork) { \
+    LOG(FATAL) << "MXNet build without lapack. Function " << #func << " is not available."; \
+    return 1; \
+  }
+
+  #define MXNET_LAPACK_CWRAPPER3(func, dtype) \
+  int MXNET_LAPACK_##func(int matrix_layout, char uplo, int n, dtype *a, \
+                                 int lda, dtype *w, dtype *work, int lwork, \
+                                 int *iwork, int liwork) { \
+    LOG(FATAL) << "MXNet build without lapack. Function " << #func << " is not available."; \
+    return 1; \
+  }
+
+  #define MXNET_LAPACK_UNAVAILABLE(func) \
+  int mxnet_lapack_##func(...) { \
+    LOG(FATAL) << "MXNet build without lapack. Function " << #func << " is not available."; \
+    return 1; \
+  }
+  MXNET_LAPACK_CWRAPPER1(spotrf, float)
+  MXNET_LAPACK_CWRAPPER1(dpotrf, double)
+  MXNET_LAPACK_CWRAPPER1(spotri, float)
+  MXNET_LAPACK_CWRAPPER1(dpotri, double)
+
+  MXNET_LAPACK_UNAVAILABLE(sposv)
+  MXNET_LAPACK_UNAVAILABLE(dposv)
+
+  MXNET_LAPACK_CWRAPPER2(sgelqf, float)
+  MXNET_LAPACK_CWRAPPER2(dgelqf, double)
+  MXNET_LAPACK_CWRAPPER2(sorglq, float)
+  MXNET_LAPACK_CWRAPPER2(dorglq, double)
+
+  MXNET_LAPACK_CWRAPPER3(ssyevd, float)
+  MXNET_LAPACK_CWRAPPER3(dsyevd, double)
+#endif  // MSHADOW_USE_MKL == 0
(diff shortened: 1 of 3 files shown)


larroy:
@jlcontreras good point, but that was there already, I think its because lapack is used in some operators. Do you have a better suggestion?

jlcontreras:
Difficult to say, maybe common/ ?

larroy:
@jlcontreras do you think it makes sense to keep it in a different PR?

src/operator/c_lapack_api.h
```diff
@@ -375,7 +364,10 @@ inline void flip(int m, int n, DType *b, int ldb, DType *a, int lda) {
 
   MXNET_LAPACK_CWRAPPER3(ssyevd, float)
   MXNET_LAPACK_CWRAPPER3(dsyevd, double)
-
+  #undef MXNET_LAPACK_CWRAPPER1
```
lebeg: Why this is needed?

src/operator/c_lapack_api.h
```diff
@@ -325,41 +325,30 @@ inline void flip(int m, int n, DType *b, int ldb, DType *a, int lda) {
 #else
 
   // use pragma message instead of warning
+  /*
```
lebeg: Why commenting out? Maybe it should go to `c_lapack_api.cc`?

src/operator/c_lapack_api.h
larroy: to avoid being defined again in the implementation

src/operator/c_lapack_api.h
larroy: Good point

asmushetzel:
LGTM. Don't think that it makes sense to move the c_lapack_api to a different directory now. There are pros and cons about where it should sit, but no immediate benefit from moving.

larroy:
@szha @zhreshold

larroy:
@mxnet-label-bot  [pr-awaiting-merge]

src/operator/c_lapack_api.cc
```diff
@@ -0,0 +1,72 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "c_lapack_api.h"
+
+#if (MSHADOW_USE_MKL && MXNET_USE_LAPACK)
```
KellenSunderland: Why do we need the left hand side of this &&, MSHADOW_USE_MKL if we're testing against only the right hand side in the next line?  For what cases would line 22 alone be true and line 23 + line 22 be false?

larroy: So it's the same logic in the header in line 153:  https://github.com/apache/incubator-mxnet/pull/13535/files#diff-3600054dbf1ab219ca274dc6f21869b5R153

KellenSunderland: I don't see that it's the same, maybe I'm missing something.  Here's what doesn't make sense to me:

```cpp
#if (MSHADOW_USE_MKL && MXNET_USE_LAPACK)
#elif MXNET_USE_LAPACK
#else 
do_something...
```
Looking at this with a simple truth table:

Assume MXNET_USE_LAPACK is true, Assume MSHADOW_USE_MKL true -> Do nothing
Assume MXNET_USE_LAPACK is true, Assume MSHADOW_USE_MKL false -> Do nothing
Assume MXNET_USE_LAPACK is false, Assume MSHADOW_USE_MKL true -> Do something
Assume MXNET_USE_LAPACK is false, Assume MSHADOW_USE_MKL false -> Do something

So no matter what we set MSHADOW_USE_MKL to it won't affect the control flow right? (or am I missing something really basic which is totally possible).

larroy: My point is that it's matching the define flow from the header file. But since it's causing confusion and I think your analysis is right, I can change it to read
```
 #if MXNET_USE_LAPACK
  ....
 #endif 

```
From your commend I understand that this would be your preference, am I correct?

KellenSunderland: Right, I just think it better describes what's going on. Thanks Pedro.

KellenSunderland:
Any idea why the CI failed?  Looks like a timeout right?

larroy:
@szha @KellenSunderland @tqchen  could we please merge this? it's needed for fixing crashes on some scenarios. Thank you.

PR #13543, opened 2018-12-05
Chi_square_check for discrete distribution fix

ChaiBapchya:
## Description ##
check for bucket instead of the bucket index

It was incorrectly checking for index
```
for i in range(len(buckets)):
    obs_freq[i] = (sample_bucket_ids == i).sum()
```

It should be checking for the value at the index

Bucketing for discrete and continuous distributions is different and hence has to be handled differently.

Moreover this change passes the verify_generator tests for both discrete distribution (randint) as well as continuous ones (normal, uniform, ... etc)

### DIFF
--- a/python/mxnet/test_utils.py
+++ b/python/mxnet/test_utils.py
@@ -1911,12 +1911,15 @@ def chi_square_check(generator, buckets, probs, nsamples=1000000):
     if continuous_dist:
         sample_bucket_ids = np.searchsorted(buckets_npy, samples, side='right')
     else:
-        sample_bucket_ids = samples
+        sample_bucket_ids = np.array(samples)
     if continuous_dist:
         sample_bucket_ids = sample_bucket_ids // 2
     obs_freq = np.zeros(shape=len(buckets), dtype=np.int)
     for i, _ in enumerate(buckets):
-        obs_freq[i] = (sample_bucket_ids == buckets[i]).sum()
+        if continuous_dist:
+            obs_freq[i] = (sample_bucket_ids == i).sum()
+        else:
+            obs_freq[i] = (sample_bucket_ids == buckets[i]).sum()
     _, p = ss.chisquare(f_obs=obs_freq, f_exp=expected_freq)
     return p, obs_freq, expected_freq
 
--- a/python/mxnet/test_utils.py
+++ b/python/mxnet/test_utils.py
@@ -1916,7 +1916,7 @@ def chi_square_check(generator, buckets, probs, nsamples=1000000):
         sample_bucket_ids = sample_bucket_ids // 2
     obs_freq = np.zeros(shape=len(buckets), dtype=np.int)
     for i, _ in enumerate(buckets):
-        obs_freq[i] = (sample_bucket_ids == buckets[i]).astype('uint8').sum()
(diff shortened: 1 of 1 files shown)


python/mxnet/test_utils.py
```diff
@@ -1915,8 +1915,8 @@ def chi_square_check(generator, buckets, probs, nsamples=1000000):
     if continuous_dist:
         sample_bucket_ids = sample_bucket_ids // 2
     obs_freq = np.zeros(shape=len(buckets), dtype=np.int)
-    for i in range(len(buckets)):
-        obs_freq[i] = (sample_bucket_ids == i).sum()
+    for i, _ in enumerate(buckets):
```
yuxihu: Looks like you do not need to change this line. You just need i.

python/mxnet/test_utils.py
```diff
@@ -1915,8 +1915,8 @@ def chi_square_check(generator, buckets, probs, nsamples=1000000):
...
+        obs_freq[i] = (sample_bucket_ids == buckets[i]).astype('uint8').sum()
```
yuxihu: Are you sure this work? (sample_bucket_ids == buckets[i]) will return a bool object. you might hit AttributeError here. Please double check.

ChaiBapchya: hence I tried to convert it to int so as to take sum

python/mxnet/test_utils.py
ChaiBapchya: nope, it fails on CI saying use_enumerate instead of range(Len())
Hence I had to put a separate commit c989695 for that

python/mxnet/test_utils.py
ChaiBapchya: but even that doesn't work that way. hence using count instead.

python/mxnet/test_utils.py
```diff
@@ -1915,8 +1915,8 @@ def chi_square_check(generator, buckets, probs, nsamples=1000000):
     if continuous_dist:
         sample_bucket_ids = sample_bucket_ids // 2
     obs_freq = np.zeros(shape=len(buckets), dtype=np.int)
-    for i in range(len(buckets)):
-        obs_freq[i] = (sample_bucket_ids == i).sum()
+    for i, _ in enumerate(buckets):
+        obs_freq[i] = (sample_bucket_ids == buckets[i]).sum()
```
yuxihu: This does not work. You might want to test it locally and make sure it passes.

python/mxnet/test_utils.py
```diff
@@ -1911,12 +1911,15 @@ def chi_square_check(generator, buckets, probs, nsamples=1000000):
...
     if continuous_dist:
         sample_bucket_ids = sample_bucket_ids // 2
     obs_freq = np.zeros(shape=len(buckets), dtype=np.int)
-    for i in range(len(buckets)):
-        obs_freq[i] = (sample_bucket_ids == i).sum()
+    for i, _ in enumerate(buckets):
+        if continuous_dist:
+            obs_freq[i] = (sample_bucket_ids == i).sum()
+        else:
+            obs_freq[i] = (sample_bucket_ids == buckets[i]).sum()
```
yuxihu: it will still fail if it goes to the 'else' branch, right?

Issue #13385, opened 2018-11-23
[Clojure] - Turn examples into integration tests


gigasquid:
Currently we have a good set of examples in the Clojure package but it would be nice to be able to run them as integration/ regression tests whenever we are making changes or getting ready for a release.

Some thought needs to be put into the best way to do this, but one idea is to make them runnable with `lein test` somehow and then have a script that runs through all the directories in the examples and run them.

gigasquid:
This is now in place!  Thanks @hellonico

PR #13554, opened 2018-12-06
#13385 [Clojure] - Turn examples into integration tests

hellonico:
!!! This is work in progress. 

Would:
- the committed script 
- adding a bare minimum deftest inside each example

be ok?

### DIFF
--- a/contrib/clojure-package/examples/cnn-text-classification/src/cnn_text_classification/classifier.clj
+++ b/contrib/clojure-package/examples/cnn-text-classification/src/cnn_text_classification/classifier.clj
@@ -16,7 +16,9 @@
 ;;
 
 (ns cnn-text-classification.classifier
-  (:require [cnn-text-classification.data-helper :as data-helper]
+  (:require [clojure.java.io :as io]
+            [clojure.java.shell :refer [sh]]
+            [cnn-text-classification.data-helper :as data-helper]
             [org.apache.clojure-mxnet.eval-metric :as eval-metric]
             [org.apache.clojure-mxnet.io :as mx-io]
             [org.apache.clojure-mxnet.module :as m]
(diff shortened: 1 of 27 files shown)


gigasquid:
Thanks so much @hellonico for tackling this. This will be a big improvement to stability and regression of the package as we develop it.  The way you are approaching it with a minimum test and then the script seems like the right direction. 😸

hellonico:
The following examples are turned into integration tests:

SuperPinkicious:examples niko$ find . -name test 
./gan/test
./visualization/test
./neural-style/test
./tutorial/test
./multi-label/test
./module/test
./cnn-text-classification/test
./imclassification/test
./profiler/test
./rnn/test

rnn has a training section and a test trained network section, but the training section is very, very, very slow 
I am wondering if it should still be added or not? It largely depends on the CI setup at this stage, but on my dev machine I can not get anything in a reasonable amount of time.

gigasquid:
@hellonico - Awesome work!
Regarding the rnn example - I think limiting it to just the test trained network is sufficient and will keep the time down to a somewhat reasonable level.

hellonico:
Good morning ! 
Great. Just squashed the commits ... and rebased on master.

gigasquid:
Great! I’ll take a look on Monday. Thanks!

Have a good rest of the weekend
:)

Carin

hellonico:
> ci/jenkins/mxnet-validation/sanity — Job failed 

Any idea on how to avoid the (de facto not very possible) license check on json files ?

hellonico:
Finally ... 

[nohup.txt](https://github.com/apache/incubator-mxnet/files/2661882/nohup.txt)

gigasquid:
Awesome work! I think we can get the CI to pass the RAT check if you put the files having trouble:

`mxnet/contrib/clojure-package/examples/imclassification/test/test-symbol.json.ref`
 `mxnet/contrib/clojure-package/examples/profiler/test/profile-matmul-20iter.json.ref`

 in here https://github.com/apache/incubator-mxnet/blob/master/tests/nightly/apache_rat_license_check/rat-excludes

It gets called from here https://github.com/apache/incubator-mxnet/blob/master/ci/docker/runtime_functions.sh#L996

@marcoabreu Do you have thoughts on how these integration tests should be integrated in the CI? Should it be nightly since they run pretty long? Should we merge this first and then open another one for integrating into the CI?

marcoabreu:
Yeah nightly would be a good place. I'll leave it up to you whether we do it in this PR or a separate one. But I think a separate one is a good idea

hellonico:
- Rebased master (monday 10th commits) 
- Updated json files to be ignored in the nightly checks 

Ready to be merged !

Let's find out how much time it takes to run the suite. When the models are already there, those tests run in about 10-15 minutes on slow machine.

gigasquid:
Thanks for taking this on and making it happen 💯 

I'm going to go ahead a merge this and then we can work on adding it to the nightly tests in another PR and see how it runs there.

PR #13559, opened 2018-12-06
fix for opencv4

### LINKED ISSUE #13288: OpenCV 4.0 Compatibility
OpenCV 4 is happening soon https://github.com/opencv/opencv/wiki/ChangeLog#version400. The big changes include the switch to C++11 and API changes.

If you have experience trying it out with MXNet, please do share your findings here. Thank you.

daleydeng:
## Description ##
Fix compatibility for opencv 4 with flags started with cv namespace.

@szha: resolves #13288

### DIFF
--- a/src/io/image_aug_default.cc
+++ b/src/io/image_aug_default.cc
@@ -479,14 +479,14 @@ class DefaultImageAugmenter : public ImageAugmenter {
         }
         if (i == 1) {
           // contrast
-          cvtColor(res, temp_, CV_RGB2GRAY);
+          cvtColor(res, temp_, cv::COLOR_RGB2GRAY);
           float gray_mean = cv::mean(temp_)[0];
           res.convertTo(res, -1, alpha_c, (1 - alpha_c) * gray_mean);
         }
         if (i == 2) {
           // saturation
-          cvtColor(res, temp_, CV_RGB2GRAY);
-          cvtColor(temp_, temp_, CV_GRAY2BGR);
+          cvtColor(res, temp_, cv::COLOR_RGB2GRAY);
+          cvtColor(temp_, temp_, cv::COLOR_GRAY2BGR);
           cv::addWeighted(res, alpha_s, temp_, 1 - alpha_s, 0.0, res);
         }
       }
@@ -495,7 +495,7 @@ class DefaultImageAugmenter : public ImageAugmenter {
     // color space augmentation
     if (param_.random_h != 0 || param_.random_s != 0 || param_.random_l != 0) {
       std::uniform_real_distribution<float> rand_uniform(0, 1);
-      cvtColor(res, res, CV_BGR2HLS);
+      cvtColor(res, res, cv::COLOR_BGR2HLS);
       // use an approximation of gaussian distribution to reduce extreme value
       float rh = rand_uniform(*prnd); rh += 4 * rand_uniform(*prnd); rh = rh / 5;
       float rs = rand_uniform(*prnd); rs += 4 * rand_uniform(*prnd); rs = rs / 5;
@@ -515,7 +515,7 @@ class DefaultImageAugmenter : public ImageAugmenter {
           }
         }
       }
-      cvtColor(res, res, CV_HLS2BGR);
+      cvtColor(res, res, cv::COLOR_HLS2BGR);
     }
 
     // pca noise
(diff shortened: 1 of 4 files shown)


larroy:
I checked the docs from opencv the older flags are enums, are we sure the ones in the namespace are compatible it was not clear to me from a quick look to the documentation even though CI agrees.

szha:
@larroy the flags are used in conjunction with the opencv function calls, so even if the value changes, as long as the input/output remains the same, it should still work.

zhreshold:
LGTM, but not sure if every single ENUM is covered in this PR, is this process still conducted by eyes?

daleydeng:
i have compiled successfully. you can use grep to found every instances of cv flags

Roshrini:
@larroy @zhreshold Is this PR good to go?

wkcn:
The current source of MXNet is not compatible with OpenCV4, and this PR fixes the issue.

In Arch Linux, The package name of OpenCV4 has been changed to `opencv4` rather than `opencv`, so I change the Makefile L138~L145:
```Makefile
# setup opencv
ifeq ($(USE_OPENCV), 1)
	CFLAGS += -DMXNET_USE_OPENCV=1 $(shell pkg-config --cflags opencv4)
	LDFLAGS += $(filter-out -lopencv_ts, $(shell pkg-config --libs opencv4))
	BIN += bin/im2rec
else
	CFLAGS+= -DMXNET_USE_OPENCV=0
endif
```

wkcn:
@szha Yes. 
However, In Arch Linux, ‘pkg-config --cflags opencv’ will raise the error ‘package opencv not found’ after installing OpenCV4.
I don’t know whether does the error also occur on other operating systems like Ubuntu.

gigasquid:
I was just trying to setup a new mac and now `brew install opencv` gives you 4.0 with no way (through brew) to get opencv3 :(

So this PR is pretty important to tested/merged

gigasquid:
on OSX  - I cannot do a make without errors due to `package opencv not found’. 

```Package opencv was not found in the pkg-config search path.
Perhaps you should add the directory containing `opencv.pc'
to the PKG_CONFIG_PATH environment variable
No package 'opencv' found
```

and 

```src/imperative/../profiler/./profiler.h:1186:8: warning: lambda capture 'this' is not used [-Wunused-lambda-capture]
      [this](OprExecStat *stat) {}, name_.c_str(), dev_type_, dev_id_,
       ^
In file included from src/io/image_aug_default.cc:31:
src/io/./image_augmenter.h:31:10: fatal error: 'opencv2/opencv.hpp' file not found
#include <opencv2/opencv.hpp>
```

This is because as @wkcn said above the pkg name has changed to `opencv4` from `opencv`

```
ls -l /usr/local/lib/pkgconfig
opencv4.pc -> ../../Cellar/opencv/4.0.1/lib/pkgconfig/opencv4.pc
```

If I make the changes to the Makefile that @wkcn suggested to change it to opencv4, it will work - however we would also need to change all the CI files to bump opencv as well as coordinate changing building of nightly distributions
cc @marcoabreu @lanking520 

Maybe we could put a switch in the Makefile or another config to specify what version of opencv you have? That way we could make the upgrade path easier.

Could the makefile somehow detect what version of opencv you have?

chinakook:
Manjaro(based on Arch) system should be supported because it's widely used and opencv 4 is the default version on Manjaro/Arch.

marcoabreu:
Yeah I'd be in favour if you upgrade to OpenCV4 in our Ci while you're at it.

lanking520:
Do we still have OpenCV 2 and 3 support if we switch the namespace? Since CI now are using CV2 and publish using CV3

vandanavk:
what are the next steps on this PR?

ankkhedia:
@daleydeng Could you please take a look into the comments above?

wkcn:
Thanks for your contribution! @daleydeng 

@szha @zhreshold @anirudh2290 
Hi! Is it available to merge this PR? Thanks!

zhreshold:
I see there's comment not addressed yet?

vandanavk:
@mxnet-label-bot update [pr-awaiting-response]

szha:
The issues in this PR are being fixed in #14313

wkcn:
Sorry that I submited another PR, which supports multiple versions of OpenCV.
https://github.com/apache/incubator-mxnet/pull/14313

Thanks for your contribution!

PR #13574, opened 2018-12-07
Replaced rand_r with std:: random generation

lebeg:
## Description ##

`rand_r` is not thread-safe and is not available on Android. Part of bigger https://github.com/apache/incubator-mxnet/pull/11148

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/src/operator/rnn_impl.h
+++ b/src/operator/rnn_impl.h
@@ -34,6 +34,7 @@
 #include <vector>
 #include <string>
 #include <utility>
+#include <random>
 #include "./math.h"
 #include "./math_functions-inl.h"
 #include "./operator_common.h"
@@ -149,7 +150,6 @@ void LstmForwardTraining(DType* ws,
   const int r_size = D * T * N * H * 6;
   const int y_offset = T * N * H * 5;
   const int cell_size = N * H;
-  unsigned int seed_ = 17 + rand() % 4096;  // NOLINT(runtime/threadsafe_fn)
   int idx = 0;  // state & cell state's idx;
   const int omp_threads = mxnet::engine::OpenMP::Get()->GetRecommendedOMPThreadCount();
   for (int i = 0; i < L; ++i) {
@@ -176,13 +176,17 @@ void LstmForwardTraining(DType* ws,
       if (dropout > 0.0f) {
         #pragma omp parallel for num_threads(omp_threads)
         for (int j = 0; j < T * N * H * D; j++) {
-          int rand_data = rand_r(&seed_);
+          static thread_local std::random_device device;
+          static thread_local std::default_random_engine generator(device());
+          static thread_local std::uniform_int_distribution<int> distribution;
+          static thread_local auto dice = std::bind(distribution, generator);
+          int rand_data = dice();
           if (static_cast<float>(rand_data % 1000) < static_cast<float>(1000 * dropout)) {
             dropout_random[i * T * N * H * D + j] = 0;
             y.dptr_[j] = 0;
           } else {
             dropout_random[i * T * N * H * D + j] = 1.0f - dropout;
-            y.dptr_[j] =  y.dptr_[j] / (1.0f - dropout);
+            y.dptr_[j] = y.dptr_[j] / (1.0f - dropout);
           }
         }
       }
@@ -994,7 +998,6 @@ void GruForwardTraining(DType* ws,
   DType* bx_l = bx;
   DType* bh_l = bh;
   DType* y_tmp = x_ptr;
-  unsigned int seed_ = 17 + rand() % 4096;  // NOLINT(runtime/threadsafe_fn)
   for (int l = 0; l < L; l++) {
     if (l != 0) {
       y_tmp = y_l;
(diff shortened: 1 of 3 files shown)


src/operator/rnn_impl.h
```diff
@@ -176,13 +176,17 @@ void LstmForwardTraining(DType* ws,
       if (dropout > 0.0f) {
         #pragma omp parallel for num_threads(omp_threads)
         for (int j = 0; j < T * N * H * D; j++) {
-          int rand_data = rand_r(&seed_);
+          static thread_local std::random_device device;
```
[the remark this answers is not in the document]
larroy: I see.

tests/cpp/include/test_ndarray_utils.h
```diff
@@ -54,9 +55,11 @@ inline unsigned gen_rand_seed() {
...
 inline float RandFloat() {
-  static unsigned seed = gen_rand_seed();
-  double v = rand_r(&seed) * 1.0 / RAND_MAX;
-  return static_cast<float>(v);
+  static thread_local std::random_device device;
+  static thread_local std::default_random_engine generator(device());
+  static thread_local std::uniform_real_distribution<float> distribution;
+  static thread_local auto dice = std::bind(distribution, generator);
+  return dice();
 }
```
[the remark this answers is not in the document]
lebeg: The problem is file is in `tests/`.

Roshrini:
@perdasilva @larroy Is your concern answered? Is this PR good to go?

Roshrini:
@mxnet-label-bot Update [pr-awaiting-response]

anirudhacharya:
@lebeg @larroy @perdasilva ping for update

lebeg:
I think this PR is good to merge, can somebody do this?

lebeg:
@mxnet-label-bot Update [pr-awaiting-merge]

sandeep-krishnamurthy:
@apeforest - Can you please take a look at this PR?
Thanks @lebeg for this PR.

tests/cpp/include/test_ndarray_utils.h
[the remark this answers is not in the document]
perdasilva: There's no facepalm emoticon, so I had thumbs-up

lebeg:
Can anybody merge this?

lanking520:
@apeforest Could you please take a look?

src/operator/rnn_impl.h
eric-haibin-lin: This should probably addressed at the framework level by providing APIs to get rand numbers. We should not expect developers who implement operators to handle thread local variables

lebeg:
> we have random number generator as operator resources

@szha can you add a bit more information on this? Do you mean there is a random number generator operator?

src/operator/rnn_impl.h
lebeg: @eric-haibin-lin In general, I don't see why a thread local variable is an issue - there is a separate generator for every thread and nothing needs to be done additionally to ensure thread safety. Otherwise locking needs to be in place.

I could add a generate random number function to the framework API, but wouldn't change the implementation. What do you think?

lebeg: And in case of 1 function for all there will be complications with setting the seed if needed as well.

szha:
@lebeg I recommend looking at the dropout implementation (in `src/operator/nn/dropout-inl.h`)

TaoLv:
Looks like the change is extracted from #11148 .  Reviewed the code before and related comment and response are here: https://github.com/apache/incubator-mxnet/pull/11148/files#r210174235

lebeg:
@szha @TaoLv thanks, quoting here again for reference:
> Well, the problem is not about the seed itself, but with usage of `rand_r` and the fact that it's multithreaded. As far as I can tell in dropout-inl.h#L85 the seed is set outside of OpenMP threads and no random number gets generated with it afterwards, so it should be fine.

yzhliu:
@lebeg Can we port it to https://github.com/apache/incubator-mxnet/blob/master/include/mxnet/random_generator.h ? It is designed exactly for non-threadsafe rand number generators (in which cuda device random api is not thread-safe). I guess this is what @szha meant.

Ref:
https://github.com/apache/incubator-mxnet/blob/master/include/mxnet/random_generator.h
https://github.com/apache/incubator-mxnet/blob/master/src/resource.cc#L298
https://github.com/apache/incubator-mxnet/blob/master/src/operator/random/sampler.h

yzhliu:
Usually using `thread_local` rand generator is not a good idea, not only in the sense of system design, but it could be invalid.

Say if you use 4 threads to fill an array A[4], each Gaussian random generator fills one entry with its head element - `A[0] = gen1[0], A[1] = gen2[0], A[2] = gen3[0], A[3] = gen4[0]` - this array does not follow Gaussian distribution at all. Pseudorandom sequence has to be long enough in order to satisfy the distribution property.

Also it will be great to let users set their global seed, as some (maybe not-well-designed) algorithms are sensitive to seed number. This means all random generators within the system should take same input seed as their constructor argument. (I know currently this rule is not well followed in mxnet though.)

This is why `RandGenerator` is implemented so complicated.

src/operator/rnn_impl.h
yzhliu: suggest to refer to `RandGenerator<xpu>` as I mentioned below. It handles thread-safe problem without locking. The implement is somewhat tricky though.

lebeg:
Thank you @yzhliu for your input! Now it's indeed much clearer to me. I will take a look at the RandGenerator implementation and modify the change accordingly.

vandanavk:
@mxnet-label-bot update [pr-awaiting-response, Operator]

karan6181:
@lebeg Did you get a chance to look at the RandGenerator implementation ? Can you please make the relevant code changes for that? Thank you!

lebeg:
Closing this for now, no capacity to work on this right now.

PR #13575, opened 2018-12-07
Fix the bug of BidirectionalCell

BeyonderXX:
## Description ##
I did hybridize( ) and pass "valid_length" to the unroll( ) function of BidirectionalCell, then returned AssertionError in line 79. Because symbol.split( ) return a symbol but not a symbol list. Result in the length of inputs is not equal to parameter "length"  when call unroll( )  to compute r_outputs and r_states. Symbol support for converting to list through list( ) API, so that use list( ) function insted of _as_list( ).

### Essentials

* [x]  Changes are complete (i.e. I finished coding on this PR)
* [x]  All changes have test coverage:
-  Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
 * [x]  To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
python/mxnet/gluon/rnn/rnn_cell.py
- [line 1055] [line 1069]add _reverse_sequences( ) to calculate "reversed_inputs" and "reversed_r_outputs" 
- [line 1055] use list( ) instead of _as_list( )

tests/python/unittest/test_gluon_rnn.py
- [line 603] add test_bidirectional_unroll_valid_length( ) to test the unroll( ) of BidirectionalCell when pass 
valid_length to it.

### DIFF
--- a/tests/python/unittest/test_gluon_rnn.py
+++ b/tests/python/unittest/test_gluon_rnn.py
@@ -243,6 +243,34 @@ def test_bidirectional():
     assert outs == [(10, 200), (10, 200), (10, 200)]
 
 
+def test_bidirectional_unroll_valid_length():
+    # Test BidirectionalCell.
+    # In 1.3.1 version, after hybridize( ), BidirectionalCell would failed when pass valid_length to unroll( ).
+    class BiLSTM(gluon.nn.HybridBlock):
+        def __init__(self, rnn_size, time_step, **kwargs):
+            super(BiLSTM, self).__init__(**kwargs)
+            self.time_step = time_step
+            with self.name_scope():
+                self.bi_lstm = gluon.rnn.BidirectionalCell(
+                    gluon.rnn.LSTMCell(rnn_size, prefix='rnn_l0_'),
+                    gluon.rnn.LSTMCell(rnn_size, prefix='rnn_r0_'),
+                    output_prefix='lstm_bi_')
+
+        def hybrid_forward(self, F, inputs, valid_len):
+            outputs, states = self.bi_lstm.unroll(self.time_step, inputs, valid_length=valid_len,
+                                                  layout='TNC', merge_outputs=True)
+            return outputs, states
+
+    rnn_size, time_step = 100, 3
+    net = BiLSTM(rnn_size, time_step)
+    net.initialize()
+    net.hybridize()
+    inputs_data = mx.nd.random.uniform(shape=(3, 10, 50))
+    valid_len = mx.nd.array(range(1, 11))
+    outputs, _ = net(inputs_data, valid_len)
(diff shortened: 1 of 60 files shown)


szha:
@larroy _as_list would return a wrong list as it would wrap the single grouped symbol inside, instead of `length` number of tensors for each time-step, which is what the output should be.

szha:
@BeyonderXX thanks for the fix. Could you add a test with the proper assertions?

BeyonderXX:
> @BeyonderXX thanks for the fix. Could you add a test with the proper assertions?

I‘m glad to. Do you mean add assertion to this code or provide a test case with assertion? If the latter case, is there any sample test?

tests/python/unittest/test_gluon_rnn.py
```diff
@@ -243,6 +243,34 @@ def test_bidirectional():
...
+            self.time_step = time_step
+            with self.name_scope():
+                self.bi_lstm = gluon.rnn.BidirectionalCell(
+                    gluon.rnn.LSTMCell(rnn_size, prefix='rnn_l0_'),
+                    gluon.rnn.LSTMCell(rnn_size, prefix='rnn_r0_'),
+                    output_prefix='lstm_bi_')
+
+        def hybrid_forward(self, F, inputs, valid_len):
+            outputs, states = self.bi_lstm.unroll(self.time_step, inputs, valid_length=valid_len,
+                                                  layout='TNC', merge_outputs='True')
```
szha: I think merge_outputs should be boolean or None

BeyonderXX: Thank you so much!

BeyonderXX:
@szha I just added one test case named "test_bidirectional_unroll_valid_length" to test_gluon_rnn. While jenkins detect another test called "test_layer_bidirectional " which returned error in various plateform while using cpu. Is this OK or raise by my code in BidirectionalCell ? But jenkins checked pass at first version. 
This is error info,
"test_gluon_rnn.test_bidirectional_unroll_valid_length ... ok (0.0268s)

test_gluon_rnn.test_layer_bidirectional ... *** Error in `python3.6': munmap_chunk(): invalid pointer: 0x00007ff25c1f3140 ***
"

BeyonderXX:
@szha I reproduced the issue in windows. Might unroll( ) would raise null pointer when pass valid_length after hybridize( ).

BeyonderXX:
@szha  I figured out the issue, because I passed a wrong parameter to unroll( ) which led to out of bounds. I have fixed the problem and rewrited my code in BidirectionalCell.  So sorry to disturb you abruptly. Thanks!

szha:
@BeyonderXX thanks for the fix!

Issue #12255, opened 2018-08-20
Pretty high cpu load when import mxnet


fighting-liu:
When i import mxnet in 8 processes simultaneously, all cpu resources will be used and the program stagnates for almost 5 minutes.

It works fine for mxnet1.1 but failed for mxnet1.2 and mxnet1.3
Following is the sample code

```
import multiprocessing

def mxnet_worker():
    print 'before import'
    import mxnet 
    print 'after import'

read_process = [multiprocessing.Process(target=mxnet_worker) for i in range(8)]
for p in read_process:
    p.daemon = True
    p.start()
```

Any solution to this for mxnet1.2 and mxnet1.3? Thanks.

vdantu:
@fighting-liu : Just to understand your problem a little further, could you give details of your environment? This is given in the issue template. Also, what was the issue? Was the "after import" being printed after 5 minutes? How long did it take for 1.1 and also, how long does it take for one process import'ing mxnet with .12?  System/environment information would be useful to help you further.

For reference of the issue template: https://github.com/apache/incubator-mxnet/blob/master/.github/ISSUE_TEMPLATE.md

@mxnet-label-bot : [Python, Question]

fighting-liu:
@vdantu   Thanks for your attention.

**1. Environment info**
```
----------System Info----------
('Platform     :', 'Linux-3.10.0-693.17.1.el7.x86_64-x86_64-with-centos-7.4.1708-Core')
('system       :', 'Linux')
('node         :', 'train2st.prod.yiran.com')
('release      :', '3.10.0-693.17.1.el7.x86_64')
('version      :', '#1 SMP Thu Jan 25 20:13:58 UTC 2018')
----------Hardware Info----------
('machine      :', 'x86_64')
('processor    :', 'x86_64')
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                56
On-line CPU(s) list:   0-55
Thread(s) per core:    2
Core(s) per socket:    14
Socket(s):             2
NUMA node(s):          2
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 79
Model name:            Intel(R) Xeon(R) CPU E5-2680 v4 @ 2.40GHz
Stepping:              1
CPU MHz:               2401.000
CPU max MHz:           2401.0000
CPU min MHz:           1200.0000
BogoMIPS:              4800.04
Virtualization:        VT-x
L1d cache:             32K
[...]

Issue #10560, opened 2018-04-16
It gets blocked when executing python binding during installation from source 


lionel92:
Note: Providing complete information in the most concise form is the best way to get help. This issue template serves as the checklist for essential information to most of the technical issues and bug reports. For non-technical issues and feature requests, feel free to present the information in what you believe is the best form.

For Q & A and discussion, please start a discussion thread at https://discuss.mxnet.io 

## Description
Build mxnet from source successfully. However, it always gets blocked when executing python binding.

## Environment info (Required)
Centos Red Hat 4.8.5-16
GCC 4.8.5
```
What to do:
1. Download the diagnosis script from https://raw.githubusercontent.com/apache/incubator-mxnet/master/tools/diagnose.py
2. Run the script using `python diagnose.py` and paste its output here.

```----------Python Info----------
('Version      :', '2.7.5')
('Compiler     :', 'GCC 4.8.5 20150623 (Red Hat 4.8.5-16)')
('Build        :', ('default', 'Aug  4 2017 00:39:18'))
('Arch         :', ('64bit', 'ELF'))
------------Pip Info-----------
('Version      :', '10.0.0')
('Directory    :', '/home/linliu/test_env/lib/python2.7/site-packages/pip')
----------MXNet Info-----------
No MXNet installed.
----------System Info----------
('Platform     :', 'Linux-3.10.0-693.el7.x86_64-x86_64-with-centos-7.4.1708-Core')
('system       :', 'Linux')
('node         :', 'mlt-bdw035')
('release      :', '3.10.0-693.el7.x86_64')
('version      :', '#1 SMP Tue Aug 22 21:09:27 UTC 2017')
----------Hardware Info----------
('machine      :', 'x86_64')
('processor    :', 'x86_64')
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                44
On-line CPU(s) list:   0-43
Thread(s) per core:    1
Core(s) per socket:    22
Socket(s):             2
NUMA node(s):          1
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 79
Model name:            Intel(R) Xeon(R) CPU E5-2699 v4 @ 2.20GHz
Stepping:              1
CPU MHz:               1204.500
CPU max MHz:           3600.0000
CPU min MHz:           1200.0000
BogoMIPS:              4389.75
Virtualization:        VT-x
L1d cache:             32K
L1i cache:             32K
[...]

PR #8686, opened 2017-11-16
Kernel operator tuning

cjolivier01:
@piiswrong 

## Description ##
Automatic OMP operator tuning based upon kernel operation workload.
Determines "weight" of a unary or binary kernel op and then uses this to determine if OMP should be used, given # of iterations required and # threads to perform the job.

Correct decision accuracy is tested in gtest OMP_TUNING test suite by comparing with OMP, without OMP, and Auto times.

For example:
* AWS c4.8xlarge (36 vCores):
  Success rate for type float: 0.90278
  Success rate for type double: 0.88889
  Success rate for type mshadow::half::half_t: 0.83333
  Success rate for type unsigned char: 0.86111
  Success rate for type int: 0.95833
  Success rate for type long: 0.88889

* desktop: 12-core (6 real CPU cores + hyperthreading)
  Success rate for type float: 0.79167
  Success rate for type double: 0.75000
  Success rate for type unsigned char: 0.72222
  Success rate for type int: 0.94444
  Success rate for type long: 1.00000

A sample output from OMP_TUNING tests including staticstical data:
[tune_all.txt](https://github.com/apache/incubator-mxnet/files/1450885/tune_all.txt)

Currently autotuned kernel operators (tuning at startup takes a total of ~ 3ms):

mxnet::op::PopulateFullIdxRspKernel
mxnet::op::mxnet_op::set_to_int<0>
mxnet::op::mshadow_op::smooth_l1_gradient
mxnet::op::mshadow_op::smooth_l1_loss
mxnet::op::mshadow_op::eq
mxnet::op::mshadow_op::ne
mxnet::op::mshadow_op::le
mxnet::op::mshadow_op::lt
mxnet::op::mshadow_op::hypot_grad_right
mxnet::op::mshadow_op::hypot_grad_left
mxnet::op::mshadow_op::hypot
mxnet::op::mshadow_op::arctanh_grad
mxnet::op::mshadow_op::arctan_grad
mxnet::op::mshadow_op::cosh
mxnet::op::mshadow_op::rpower
mxnet::op::mshadow_op::minimum
mxnet::op::mshadow_op::arctan
mxnet::op::mshadow_op::reciprocal_square_root
mxnet::op::mshadow_op::rminus
mxnet::op::mshadow_op::arccosh_grad
mxnet::op::mshadow_op::square_root_grad
mxnet::op::mshadow_op::arctanh
mxnet::op::mshadow_op::floor
mxnet::op::mshadow_op::cosh_grad
[...]

PR #13602, opened 2018-12-10
Fix for import mxnet taking long time if multiple process launched

Vikas89:
In case there are many cores(72 cores as in c5.18xl), doing import mxnet in multiple processes take very long time. Details here: #12255

One of the reason we have OMP tuning code which iterates to find OMP tune overhead. We are reducing this iteration count to reduce the overehead of tuning code.
Also, We added an environment variable where users can set the number of cores that should be used to determine tuning.
Reducing number of cores for tuning also helps in reduction of number of iterations.

Successfully ran operator tuning unit test.(OMP_TUNING*)
Ran below test code on c5.18xl with 72 cores to see if the import mxnet is faster

```
import os
import time
import multiprocessing
from os import getpid

def mxnet_worker():
    print("before import: pid:{}".format(getpid()))
    st_time = time.time()
    import mxnet
    end_time = time.time()
    print("after import: pid:{} time:{}".format(getpid(), end_time - st_time))

read_process = [multiprocessing.Process(target=mxnet_worker) for i in range(30)]
from threading import Thread
i=0
#while i<32:
#    t1 = Thread(target=mxnet_worker)
#    t1.start()
#    i=i+1
for p in read_process:
    p.daemon = True
#    time.sleep(3)
    p.start()
time.sleep(100000)

```
Loading 30 mxnet processes takes 41 seconds with no change in num_cores for tuning.

Ideally num_cores should be 3
When I run above code, after setting the environ variable export `MXNET_USE_NUM_CORES_OPERATOR_TUNING=3`
the 30 mxnet processes finishs loading in 2.2 seconds.

## Description ##

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/docs/faq/env_var.md
+++ b/docs/faq/env_var.md
@@ -222,3 +222,18 @@ Settings for More GPU Parallelism
 - Set ```MXNET_GPU_WORKER_NTHREADS``` to a larger number (e.g., 2)
   - To reduce memory usage, consider setting ```MXNET_EXEC_NUM_TEMP```.
   - This might not speed things up, especially for image applications, because GPU is usually fully utilized even with serialized jobs.
+
+Settings for controlling OMP tuning
+---------------------------------
+- Set ```MXNET_USE_OPERATOR_TUNING=0``` to disable Operator tuning code which decides whether to use OMP or not for operator
+  - 
+   * Values: String representation of MXNET_ENABLE_OPERATOR_TUNING environment variable
+   *            0=disable all
+   *            1=enable all
+   *            float32, float16, float32=list of types to enable, and disable those not listed
+   * refer : https://github.com/apache/incubator-mxnet/blob/master/src/operator/operator_tune-inl.h#L444
+
+- Set ```MXNET_USE_NUM_CORES_OPERATOR_TUNING``` to define num_cores to be used by operator tuning code.
+  - This reduces operator tuning overhead when there are multiple instances of mxnet running in the system and we know that
+    each mxnet will take only partial num_cores available with system. 
+  - refer: https://github.com/apache/incubator-mxnet/pull/13602
--- a/docs/faq/env_var.md
+++ b/docs/faq/env_var.md
@@ -226,12 +226,11 @@ Settings for More GPU Parallelism
 Settings for controlling OMP tuning
 ---------------------------------
 - Set ```MXNET_USE_OPERATOR_TUNING=0``` to disable Operator tuning code which decides whether to use OMP or not for operator
-  - 
-   * Values: String representation of MXNET_ENABLE_OPERATOR_TUNING environment variable
-   *            0=disable all
-   *            1=enable all
-   *            float32, float16, float32=list of types to enable, and disable those not listed
-   * refer : https://github.com/apache/incubator-mxnet/blob/master/src/operator/operator_tune-inl.h#L444
+   - Values: String representation of MXNET_ENABLE_OPERATOR_TUNING environment variable
+   -            0=disable all
+   -            1=enable all
+   -            float32, float16, float32=list of types to enable, and disable those not listed
+   - refer : https://github.com/apache/incubator-mxnet/blob/master/src/operator/operator_tune-inl.h#L444
 
 - Set ```MXNET_USE_NUM_CORES_OPERATOR_TUNING``` to define num_cores to be used by operator tuning code.
   - This reduces operator tuning overhead when there are multiple instances of mxnet running in the system and we know that
--- a/src/operator/operator_tune-inl.h
+++ b/src/operator/operator_tune-inl.h
@@ -358,7 +358,6 @@ class OperatorTune : public OperatorTuneByType<DType> {
     // so take an average across all core counts
     const auto max_cores_default = static_cast<size_t>(omp_get_num_procs()) >> 1;
     const auto max_cores = dmlc::GetEnv("MXNET_USE_NUM_CORES_OPERATOR_TUNING", max_cores_default);
-    LOG(INFO) << "max_cores for tuning:" << max_cores;
     if (max_cores >= 2) {
       std::vector<duration_t> core_times;
       // Take care of any OMP lazy-init with a throwaway call
@@ -380,7 +379,7 @@ class OperatorTune : public OperatorTuneByType<DType> {
     }
     return INT_MAX;  // If only one core, then never use OMP (say the overhead is huge)
   }
-  
+
   /*!
    * \brief Some string utility functions that aren't specific to tuning
    */
--- a/src/operator/operator_tune-inl.h
+++ b/src/operator/operator_tune-inl.h
@@ -56,7 +56,7 @@ namespace op {
 #endif
 #endif  // MXNET_NO_INLINE
 
-#define OUTSIDE_COUNT_SHIFT    9
+#define OUTSIDE_COUNT_SHIFT  3
 
 namespace tune {
 
@@ -356,7 +356,9 @@ class OperatorTune : public OperatorTuneByType<DType> {
   static duration_t GetOMPLoopOverhead() {
     // It was found empirically that OMP times was not heavily tied to number of cores,
     // so take an average across all core counts
-    const auto max_cores = static_cast<size_t>(omp_get_num_procs()) >> 1;
+    const auto max_cores_default = static_cast<size_t>(omp_get_num_procs()) >> 1;
+    const auto max_cores = dmlc::GetEnv("MXNET_USE_NUM_CORES_OPERATOR_TUNING", max_cores_default);
+    LOG(INFO) << "max_cores for tuning:" << max_cores;
     if (max_cores >= 2) {
       std::vector<duration_t> core_times;
       // Take care of any OMP lazy-init with a throwaway call
@@ -378,7 +380,7 @@ class OperatorTune : public OperatorTuneByType<DType> {
     }
     return INT_MAX;  // If only one core, then never use OMP (say the overhead is huge)
   }
-
+  
   /*!
    * \brief Some string utility functions that aren't specific to tuning
    */


src/operator/operator_tune-inl.h
```diff
@@ -356,7 +356,9 @@ class OperatorTune : public OperatorTuneByType<DType> {
   static duration_t GetOMPLoopOverhead() {
     // It was found empirically that OMP times was not heavily tied to number of cores,
     // so take an average across all core counts
-    const auto max_cores = static_cast<size_t>(omp_get_num_procs()) >> 1;
+    const auto max_cores_default = static_cast<size_t>(omp_get_num_procs()) >> 1;
```
marcoabreu: That's just diving it by two, but according to your statement and the comment, this process does not benefit of many codes. Should we maybe instead use max (4, num_cores) as default?

Otherwise, the default behaviour would not change and users would still face the same challenges. I'm not a big fan of having to know a "secret" environment variable. Of course, we can have this variable to offer users the possibility to set it specifically, but generally we should choose a good default

src/operator/operator_tune-inl.h
```diff
@@ -56,7 +56,7 @@ namespace op {
 #endif
 #endif  // MXNET_NO_INLINE
 
-#define OUTSIDE_COUNT_SHIFT    9
+#define OUTSIDE_COUNT_SHIFT  3
```
marcoabreu: Could you elaborate on this? Maybe add a comment in the code why you made that change

Vikas89: The tuning code is trying to run a trivial code(int x+=y) and then average the timing stats over all the runs. 2^OUTSIDE_COUNT_SHIFT is one factor in determining number of iterations.
if OUTSIDE_COUNT_SHIFT 9 , it will run the trivial code these many time:
for 2^outside_count_shift
  for wl_count times, see here: // https://github.com/apache/incubator-mxnet/blob/master/src/operator/operator_tune-inl.h#L339 
    run_trivial_code

return total_time_taken_to_run_all_iteration / 2^outside_count_shift

Earlier the outside loop was running for 2^9 times, not it will be running for 2^3 times and averaging after that. It's just one way to reduce total number of iterations used to determine operator tuning.

src/operator/operator_tune-inl.h
Vikas89: "Should we maybe instead use min (3, num_cores) as default" 
One way to make this process fast was reducing num of iterations it was taking to determine if OMP should be used or not.
The reduction in outside loop count reduces the overhead of tuning code significantly. 
I tested specifically with 72 cores and and it took around 40 sec to launch 30 mxnet processes which is much better than existing behavior.
I am not sure about if we use 3 cores, that would be right behavior because OMP badness can increase with increase in num_cores. So tuning code can say OMP is good as it was tested on 3 cores, but the real code may behave differently.

src/operator/operator_tune-inl.h
marcoabreu: Ah I see, thanks a lot! How big is the time difference in between the old and the new setting?

I would imagine that 2^9 greatly benefits from caching while 2^3 is too trivial to benefit of the Caches and thus might influence the measurement in a way which doesn't reflect the reality (I assume we run a lot on caches).

src/operator/operator_tune-inl.h
marcoabreu: I see. To clarify, these 3 cores are just for the tuning stage, right?

@cjolivier01 @TaoLv @pengzhao-intel could you maybe provide some guidance here? I think you guys already made a lot of experience with open mp

src/operator/operator_tune-inl.h
Vikas89: "How big is the time difference in between the old and the new setting?"
On c5.18xl with 72 cores, launching 30 mxnet processes,
It almost never finishes, I gave up after 10 minutes, some benchmark says it took them 1500 seconds. 
With new change it takes ~41 seconds to launch 30 processes.

src/operator/operator_tune-inl.h
Vikas89: "these 3 cores are just for the tuning stage" - that is correct.

src/operator/operator_tune-inl.h
marcoabreu: And if you limit the number of used cores using the environment variable?
Basically a comparison of all combinations of limiting the cpu cores and the iterations

src/operator/operator_tune-inl.h
```diff
@@ -356,7 +356,8 @@ class OperatorTune : public OperatorTuneByType<DType> {
   static duration_t GetOMPLoopOverhead() {
     // It was found empirically that OMP times was not heavily tied to number of cores,
     // so take an average across all core counts
-    const auto max_cores = static_cast<size_t>(omp_get_num_procs()) >> 1;
+    const auto max_cores_default = static_cast<size_t>(omp_get_num_procs()) >> 1;
+    const auto max_cores = dmlc::GetEnv("MXNET_USE_NUM_CORES_OPERATOR_TUNING", max_cores_default);
```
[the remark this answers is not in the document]
pengzhao-intel: The env variable is documented in [here](https://github.com/apache/incubator-mxnet/blob/master/docs/faq/env_var.md).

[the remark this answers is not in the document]
TaoLv: +1 for document.

Vikas89:
> Still not very sure about the mechanism of operator tuning and how it contributes to other operator computation. Do we have any performance comparison between w/ and w/o operator tuning? Previously, I set USE_OPERATOR_TUNING=0 when built MXNet to avoid import hang. Followed the tips here: #10560 (comment)
Good point. The operator tuning code was introduced by this PR: https://github.com/apache/incubator-mxnet/pull/8686 , I don't see any numbers for performance comparison with/without tuning code.  I think you should start a discussion in dev thread and may be we would able to find some relevant historical info and then decide whether operator tuning should be enabled as default or not.

Regarding this PR, it is optimizing for the cases where OPERATOR_TUNING is enabled by default and unblocking 4.0 release.

src/operator/operator_tune-inl.h
```diff
@@ -56,7 +56,7 @@ namespace op {
 #endif
 #endif  // MXNET_NO_INLINE
 
-#define OUTSIDE_COUNT_SHIFT    9
```
anirudh2290: does changing this impact the IsOMPFaster selection faster in operator_tune.h. Do we need to tweak WORKLOAD_COUNT_SHIFT too ?

roywei:
@mxnet-label-bot add[Environment Variables, Operator]

pengzhao-intel:
Based on 10560's [comment](https://github.com/apache/incubator-mxnet/issues/10560#issuecomment-381514559), "It sometimes block executing in Ubuntu and always block executing in Windows" and several related issues, including import hang, are reported.

Does anyone can help verify the funtionality of this feature? 
Maybe set it off by default. Any idea?

anirudh2290:
I am wondering if this change is needed. For example, we are disabling openmp for child processes, then why cant we also skip tuning for the child processes ?

Vikas89:
@anirudh2290 
 >I am wondering if this change is needed. For example, we are disabling openmp for child processes, then why can't we also skip tuning for the child processes ?

if the option to enable OMP remains there, I think this is needed. If someone enables the option, this makes the tuning faster. 

If we remove OMP tuning at all, then this will automatically be removed. But removing it or disabling it will require bigger discussion and historical reason to keep it.

src/operator/operator_tune-inl.h
Vikas89: Workload_count_shift is currently 11, which means workload count will be 2048. 
this means that operation is done for 2048 times. This number can be made 10, but IsOMPFaster doesn't look like bottleneck for the related issue. It is the function which is calculating OMP overhead which is causing the problem.

anirudh2290:
for child processes we disable openmp : https://github.com/apache/incubator-mxnet/blob/master/src/initialize.cc#L66 . This means that number of omp threads will be 1 for all operators executed in the child process. Tuning doesnt help much then.

Vikas89:
@anirudh2290  
I started a discussion in dev list for disabling openmp tuning by default, just to make sure we are not missing any proven benefit. We would go either way, disable it completely or continue with this PR, depending on input from community.

anirudh2290:
To summarize an offline discussion with @Vikas89 . The auto tuning feature has performance benefits which has been documented in https://github.com/apache/incubator-mxnet/pull/8686 in the attached txt. 
[tune_all.txt](https://github.com/apache/incubator-mxnet/files/2674360/tune_all.txt)

As we can see for various operator and different inputs the Auto tune selects whether to use OMP or not. For close to 90% of the tests it makes the right selection. 

Also @Vikas89 tried to remove tuning for child process since we are disabling openmp for child process, but since the tuning gets triggered during static variable initialization as part of process startup, changes to the fork handlers are not reflected in the tuning code. So we decided to stick to the existing implementation to reduce the number of iterations for omp overhead calculation.

Vikas89:
[tuning-perf-results.txt](https://github.com/apache/incubator-mxnet/files/2677149/tuning-perf-results.txt) - For test results
@anirudh2290 ran test ./build/tests/cpp/mxnet_unit_tests --gtest_filter=OMP_TUNING.EvaluateTuneTestFloat --perf

anirudh2290:
Thanks results look good , 93 out of 96 tests do correct selection

anirudh2290:
@pengzhao-intel Can you please point to the performance issues related to OMP tuning.

pengzhao-intel:
https://github.com/apache/incubator-mxnet/issues/12255#issuecomment-444319311

@anirudh2290 this is a good summary of the issue, especially in Intel Xeon Phi system where 72 physical cores and 72X4 logic cores are available.

anirudh2290:
@pengzhao-intel yes i saw that issue. that issue checks for omp overhead serially so it launches different parallel section (with 2,..18 threads) serially. So I think for 36 cores omp runtime would try to reuse the already launched threads and not launch 170 threads. This can still be a problem when we fork the process into many subprocesses and we tried to disable operator tuning subprocesses but it wasn't trivial. Therefore the solution implemented here would be a good intermediate solution for a reasonable number of processes  forked. We should revisit the long term solution to disable tuning in forked processes though.

docs/faq/env_var.md
```diff
@@ -222,3 +222,17 @@ Settings for More GPU Parallelism
...
   - This might not speed things up, especially for image applications, because GPU is usually fully utilized even with serialized jobs.
+
+Settings for controlling OMP tuning
+---------------------------------
+- Set ```MXNET_USE_OPERATOR_TUNING=0``` to disable Operator tuning code which decides whether to use OMP or not for operator
+   - Values: String representation of MXNET_ENABLE_OPERATOR_TUNING environment variable
+   -            0=disable all
+   -            1=enable all
+   -            float32, float16, float32=list of types to enable, and disable those not listed
+   - refer : https://github.com/apache/incubator-mxnet/blob/master/src/operator/operator_tune-inl.h#L444
```
TaoLv: Not sure it's a good choice to put code link here. Once `operator_tune-inl.h` is changed, probably we need revise the line number here to avoid confusion.

Vikas89: Ah , I forgot to add that diff where I listed all the data type. I will create a separate PR to correct this.

PR #13612, opened 2018-12-11
add pos_weight for SigmoidBinaryCrossEntropyLoss

eureka7mt:
## Description ##
Add pos_weight for SigmoidBinaryCrossEntropyLoss.
A value `pos_weights > 1` decreases the false negative count, hence increasing the recall.
Conversely setting `pos_weights < 1` decreases the false positive count and increases the precision.
This can be seen from the fact that `pos_weight` is introduced as a multiplicative coefficient for the positive targets term in the loss expression:
    label * -log(sigmoid(pred)) * pos_weight + (1 - label) * -log(1 - sigmoid(pred))

It's adopted from [tensorflow's implementation](https://github.com/tensorflow/tensorflow/blob/r1.12/tensorflow/python/ops/nn_impl.py#L187)

### Essentials ###

## Comments ##

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -211,6 +211,7 @@ List of Contributors
 * [Harsh Patel](https://github.com/harshp8l)
 * [Xiao Wang](https://github.com/BeyonderXX)
 * [Piyush Ghai](https://github.com/piyushghai)
+* [Dang Trung Kien](https://github.com/kiendang)
 * [Zach Boldyga](https://github.com/zboldyga)
 * [Gordon Reid](https://github.com/gordon1992)
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -214,8 +214,8 @@ List of Contributors
 * [Dang Trung Kien](https://github.com/kiendang)
 * [Zach Boldyga](https://github.com/zboldyga)
 * [Gordon Reid](https://github.com/gordon1992)
-
 * [Ming Yang](http://ufoym.com)
+* [Neo Chien](https://github.com/cchung100m)
 
 Label Bot
 ---------
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -216,6 +216,7 @@ List of Contributors
 * [Gordon Reid](https://github.com/gordon1992)
 * [Ming Yang](http://ufoym.com)
 * [Neo Chien](https://github.com/cchung100m)
+* [Wujie Zhou](https://github.com/eureka7mt)
 
 Label Bot
 ---------
--- a/python/mxnet/gluon/loss.py
+++ b/python/mxnet/gluon/loss.py
@@ -222,7 +222,7 @@ class SigmoidBinaryCrossEntropyLoss(Loss):
           to the same shape as pred. For example, if pred has shape (64, 10)
           and you want to weigh each sample in the batch separately,
           sample_weight should have shape (64, 1).
-        - **pos_weight**:a weighting tensor of positive examples.Must be a vector with length
+        - **pos_weight**: a weighting tensor of positive examples.Must be a vector with length
           equal to the number of classes.For example, if pred has shape (64, 10)
           pos_weight should have shape (1, 10).
 
--- a/python/mxnet/gluon/loss.py
+++ b/python/mxnet/gluon/loss.py
@@ -243,9 +243,16 @@ class SigmoidBinaryCrossEntropyLoss(Loss):
             else:
                 # We use the stable formula: x - x * z + (1 + z * pos_weight - z) * (log(1 + exp(-abs(x))) + max(-x, 0))
                 log_weight = 1 + (pos_weight - 1) * label
-                loss = pred - pred*label + log_weight*(F.Activation(-F.abs(pred), act_type='softrelu') + F.relu(-pred))
+                loss = pred - pred*label + log_weight * (
+                           F.Activation(-F.abs(pred), act_type='softrelu') + F.relu(-pred))
         else:
-            loss = -(F.log(pred+1e-12)*label*pos_weight + F.log(1.-pred+1e-12)*(1.-label))
+            eps = 1e-12
+            if pos_weight == 1:
+                loss = -(F.log(pred + eps) * label +\
+                         F.log(1. - pred + eps) * (1. - label))
+            else:
+                loss = -(F.log(pred + eps) * label * pos_weight +\
+                         F.log(1. - pred + eps) * (1. - label))
         loss = _apply_weighting(F, loss, self._weight, sample_weight)
         return F.mean(loss, axis=self._batch_axis, exclude=True)
 
--- a/python/mxnet/gluon/loss.py
+++ b/python/mxnet/gluon/loss.py
@@ -234,10 +234,10 @@ class SigmoidBinaryCrossEntropyLoss(Loss):
         super(SigmoidBinaryCrossEntropyLoss, self).__init__(weight, batch_axis, **kwargs)
         self._from_sigmoid = from_sigmoid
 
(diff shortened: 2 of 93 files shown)


roywei:
@eureka7mt Thanks for the contribution, could you add a unit test for this case?

roywei:
@mxnet-label-bot add[Gluon, pr-awaiting-review]

python/mxnet/gluon/loss.py
```diff
@@ -194,6 +194,14 @@ class SigmoidBinaryCrossEntropyLoss(Loss):
         L = - \sum_i {label}_i * \log({pred}_i) +
             (1 - {label}_i) * \log(1 - {pred}_i)
 
+    A value `pos_weights > 1` decreases the false negative count, hence increasing
```
anirudhacharya: please update the formulas above with the new `pos_weight` parameter. The documentation should specify how the formula changes when `from_sigmoid` is True and False and where in the formula `pos_weight` is used.

python/mxnet/gluon/loss.py
```diff
@@ -227,13 +238,20 @@ def __init__(self, from_sigmoid=False, weight=None, batch_axis=0, **kwargs):
         super(SigmoidBinaryCrossEntropyLoss, self).__init__(weight, batch_axis, **kwargs)
         self._from_sigmoid = from_sigmoid
 
-    def hybrid_forward(self, F, pred, label, sample_weight=None):
+    def hybrid_forward(self, F, pred, label, sample_weight=None, pos_weight=None):
         label = _reshape_like(F, label, pred)
         if not self._from_sigmoid:
             # We use the stable formula: max(x, 0) - x * z + log(1 + exp(-abs(x)))
-            loss = F.relu(pred) - pred * label + F.Activation(-F.abs(pred), act_type='softrelu')
+            if pos_weight is None:
```
anirudhacharya: can default `pos_weight` be set to 1 since it is a multiplicative factor and update the formulas accordingly. Then we can do away with this `if else` statement.

eureka7mt: Well.Set pos_weight=1 is OK.I just want to write it In this format like 'sample_weight=None'.

tests/python/unittest/test_loss.py
```diff
@@ -420,6 +420,37 @@ def test_poisson_nllloss_mod():
             initializer=mx.init.Normal(sigma=0.1), eval_metric=mx.metric.Loss(),
             optimizer='adam')
     assert mod.score(data_iter, eval_metric=mx.metric.Loss())[0][1] < 0.05
+    
+@with_seed()
+def test_bce_loss_with_pos_weight():
+    #Suppose it's a multi-label classification
+    N = 20
```
anirudhacharya: please fetch a random number here for N

stu1130:
@eureka7mt Could you fix the Trailing whitespace issue?

vandanavk:
@eureka7mt could you re-trigger the CI?

ankkhedia:
@eureka7mt Could you please look into the CI failures?

anirudhacharya:

python/mxnet/gluon/loss.py
```diff
@@ -227,13 +234,14 @@ def __init__(self, from_sigmoid=False, weight=None, batch_axis=0, **kwargs):
...
         self._from_sigmoid = from_sigmoid
 
-    def hybrid_forward(self, F, pred, label, sample_weight=None):
+    def hybrid_forward(self, F, pred, label, pos_weight=1, sample_weight=None):
         label = _reshape_like(F, label, pred)
         if not self._from_sigmoid:
             # We use the stable formula: max(x, 0) - x * z + log(1 + exp(-abs(x)))
-            loss = F.relu(pred) - pred * label + F.Activation(-F.abs(pred), act_type='softrelu')
+            log_weight = 1 + (pos_weight - 1) * label
+            loss = pred - pred*label + log_weight*(F.Activation(-F.abs(pred), act_type='softrelu') + F.relu(-pred))
```
wkcn: Thanks for your contribution!
I am checking this formula.

wkcn: Right.

python/mxnet/gluon/loss.py
```diff
@@ -227,13 +234,14 @@ def __init__(self, from_sigmoid=False, weight=None, batch_axis=0, **kwargs):
         super(SigmoidBinaryCrossEntropyLoss, self).__init__(weight, batch_axis, **kwargs)
         self._from_sigmoid = from_sigmoid
 
-    def hybrid_forward(self, F, pred, label, sample_weight=None):
+    def hybrid_forward(self, F, pred, label, pos_weight=1, sample_weight=None):
         label = _reshape_like(F, label, pred)
         if not self._from_sigmoid:
             # We use the stable formula: max(x, 0) - x * z + log(1 + exp(-abs(x)))
```
wkcn: pos_weight will drop the performance when it is equal to 1.
Could you please add the following if-else statement?
```python
            if pos_weight == 1:
                # We use the stable formula: max(x, 0) - x * z + log(1 + exp(-abs(x)))
                loss = F.relu(pred) - pred * label + F.Activation(-F.abs(pred), act_type='softrelu')
            else:
                # We use the stable formula: x - x * z + (1 + z * pos_weight - z) * (log(1 + exp(-abs(x))) + max(-x, 0))
                log_weight = 1 + (pos_weight - 1) * label
                loss = pred - pred*label + log_weight*(F.Activation(-F.abs(pred), act_type='softrelu') + F.relu(-pred))
```

Thanks!

wkcn: Thanks for the contribution!
By the way, don't forget to add your name to CONTRIBUTORS.md

CONTRIBUTORS.md
```diff
@@ -20,52 +20,6 @@ Contributors of Apache MXNet (incubating)
 MXNet has been developed by a community of people who are interested in large-scale machine learning and deep learning.
 Everyone is more than welcomed to is a great way to make the project better and more accessible to more users.
 
-Committers
```
szha: List is outdated and doesn't seem necessary to be maintained.

eureka7mt: I see,I made a mistake when changing it using notepad++.I will fix it

eureka7mt: @wkcn I just worked for another thing and didn't find that the if-else statement wasn't added in the new version.Thanks for your correction.

eureka7mt:
Add the if-else statement make an error.Though the default value of pos_weight is set to be 1,the pos_weight is an (1,N) NDArray.

wkcn:
@eureka7mt 
I see. Since pos_weight is a tensor, it is better to  default pos_weight=None as you wrote before. I will update it.

eureka7mt:
@wkcn It could be a scalar for classifying a single class.But for multi-class and multi-label classifying,it should be a tensor.Because in this situation,the number of positive examples and negative examples isn't same for each class.
And in pytorch,it's also defined as a tensor.See the [pytorch docs](https://pytorch.org/docs/master/nn.html?highlight=binary_cross_entropy_with_logits#torch.nn.functional.binary_cross_entropy_with_logits).

wkcn:
I change the order of SigmoidBinaryCrossEntropyLoss inputs from `(self, F, pred, label, pos_weight=None, sample_weight=None)` to `(self, F, pred, label, sample_weight=None, pos_weight=None)`, since we need the compatibilty for other projects

wkcn:
Sorry that I trigger the sanity problem.
Could someone please help me solve it? Thanks!

eureka7mt:
Maybe the broadcast_mul isn't necessary.I think that a NDArray * a NDArray will do broadcast_mul automatically.

wkcn:
@eureka7mt I think we may pass Symbol into SigmoidBinaryCrossEntropyLoss.

wkcn:
The PR has been merged.
Thanks for your contribution: )

PR #13621, opened 2018-12-12
[MXNET-1251] Basic configuration to do static-linking

lanking520:
## Description ##
For Ubuntu 14.04 base build to install all dependencies.
@szha @zachgk 

### Essentials ###


marcoabreu:
Any specific reason we aim for Ubuntu 14? Ubuntu 18 has already been released

lanking520:
@marcoabreu As far as I know. `Cent OS 7`, `Amazon Linux` all share the same `GLIBC` version with `Ubuntu 14.04` and `Ubuntu 16.04` has a higher `GLIBC` version. To make it compatible, we should use 14.04 as the base OS

ci/publish/ubuntu_build_base.sh
```diff
@@ -0,0 +1,53 @@
...
+    pandoc \
+    python3-pip \
+    automake \
+    pkg-config
+update-alternatives \
+ --install /usr/bin/gcc gcc /usr/bin/gcc-4.8 50 \
+ --slave /usr/bin/g++ g++ /usr/bin/g++-4.8
+update-alternatives --config gcc
+# Java specific
+apt-get install -y openjdk-8-jdk
```
zachgk: I recommend using the maven3 ppa (sudo add-apt-repository ppa:andrei-pozolotin/maven3) and combining all of the apt-get installs into one call. Also just run one apt-get update after all of the ppa additions and use "set -ex"

marcoabreu:
Yeah but then we have to maintain two systems. RHEL is known to always be behind due to their strict support, but that shouldn't limit us for support of other operating systems. Would be great if you could separate that

lanking520:
@marcoabreu This is the publish configuration that works for all different platforms and this will not be used in all the rest CIs for testing. The optimal goal is using one configuration to create a `pip` or `maven` package that works for all platforms. We do have a plan to upgrade this into `16.04` once we have a test pipeline set up in Jenkins. This script is just to inherit the legacy of current publish pipeline.

About the RHEL and Debian diff. Firstly the published stuff is all the same no matter on RHEL or Debian and can be run on either. Introducing a different publish pipeline for these two in this case is not necessary as it introduce more maintainance nightmare.

marcoabreu:
Thanks for the insight, Qing! Can the binaries you create with this script in that form be loaded into ubuntu16/18 and other systems that have later versions? If yes, then I'm fine with that approach :)

lanking520:
> Thanks for the insight, Qing! Can the binaries you create with this script in that form be loaded into ubuntu16/18 and other systems that have later versions? If yes, then I'm fine with that approach :)

The answer is yes. In fact, @frankfliu has already working on producing the Dockerfile from this script that works for 16.04 with the publish pipeline. However it depends on the setup of the `test-publish` step to be ready and then we can test them through the pipeline instead of dive into different OS VMs to test manually.

roywei:
@mxnet-label-bot  add[Build, pr-awaiting-merge]

PR #13648, opened 2018-12-14
scripts for building libmxnet binary and wheel

szha:
## Description ##
scripts for building libmxnet binary and wheel

### Essentials ###

### Changes ###
- [x] add script for making all dependencies
- [x] tools for building pip package
- [x] build scripts for lib and wheel


roywei:
@mxnet-label-bot add[Build, pr-awaiting-review]

tools/build/build_lib.sh
```diff
@@ -0,0 +1,80 @@
...
+fi
+
+git clone --recursive https://github.com/apache/incubator-mxnet mxnet-build
+
+>&2 echo "Now building mxnet modules..."
+cp $make_config mxnet-build/config.mk
+
+cd mxnet-build
+
+$MAKE DEPS_PATH=$DEPS_PATH DMLCCORE
```
lanking520: Please specify the `$MAKE`

byronyi:
This is insensible. 

Why do you hard-code git clone to mxnet-build when you already have a source tree to start with? 

What if I try to maintain a fork and doesn't want to do nested checkout to the upstream?

szha:
@byronyi I imagine you can replace the clone command with a symlink to the parent folder

szha:
@byronyi we also didn't expect creating a wheel to be a prerequisite of submitting a PR. If there's any specific issue we can help with, feel free to create an issue.

byronyi:
Symlink does the trick. I will take a closer look and see how to fix this.

And you’re right, building a wheel is not necessary before a PR. But it’s also makes little value for us to send out a PR without first trying it out, which usually requires a wheel package in our in-house environment. I thought it’s a common requirement for certain users, but I could be wrong.

It would be best if we could simply produce nightly packages in the same way you produce one and release it on PyPI. Currently I’m not sure how to do that.

szha:
@byronyi I agree that simplifying it would add value, and a fix is definitely more than welcome. This PR is a simple migration of code from a separate repo and hence the current structure.

byronyi:
Sorry, symlink does not appear to work.

Issue #11315, opened 2018-06-15
Support Double precision type in Scala


nswamy:
Currently MXNet-Scala APIs support only float32, this would cause accuracy issues when customers port models between developed on Python to Scala. This needs to be fixed

lupesko:
Thanks for reporting @nswamy - but is this a bug or a feature request?

nswamy:
This is both a bug and feature request, since the accuracy will drop using Float32 on a model vs Float64

lanking520:
Let's combine the conversation in here so we can kill dupes in issues: https://github.com/apache/incubator-mxnet/issues/10338

piyushghai:
@mxnet-label-bot add [Feature request]

Issue #10338, opened 2018-03-30
NDArray saved in Python cannot be loaded in Scala


vascokk:
## Description
I am trying to convert a numpy array into an ndarray and save it. I intend to use the ndaray file in Scala, but getting src/ndarray/ndarray.cc:672: Check failed: fi->Read(data) Invalid NDArray file format.

## Environment info (Required)
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              4
On-line CPU(s) list: 0-3
Thread(s) per core:  1
Core(s) per socket:  4
Socket(s):           1
NUMA node(s):        1
Vendor ID:           GenuineIntel
CPU family:          6
Model:               60
Model name:          Intel(R) Core(TM) i5-4570 CPU @ 3.20GHz
Stepping:            3
CPU MHz:             3192.584
CPU max MHz:         3600.0000
CPU min MHz:         800.0000
BogoMIPS:            6385.16
Virtualization:      VT-x
L1d cache:           32K
L1i cache:           32K
L2 cache:            256K
L3 cache:            6144K
NUMA node0 CPU(s):   0-3
Flags:               fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt aes xsave avx f16c rdrand lahf_lm abm cpuid_fault epb invpcid_single pti retpoline tpr_shadow vnmi flexpriority ept vpid fsgsbase tsc_adjust bmi1 hle avx2 smep bmi2 erms invpcid rtm xsaveopt dtherm ida arat pln pts
----------Python Info----------
('Version      :', '2.7.14')
('Compiler     :', 'GCC 7.2.0')
('Build        :', ('default', 'Sep 23 2017 22:06:14'))
('Arch         :', ('64bit', ''))
------------Pip Info-----------
('Version      :', '9.0.3')
('Directory    :', '/usr/local/lib/python2.7/dist-packages/pip')
----------MXNet Info-----------
[...]

PR #13678, opened 2018-12-18
[MXNET-1260] Float64 DType computation support in Scala/Java

piyushghai:
## Description ##
This PR introduces Float64/Double data type support in NDArrays in Scala. Currently we only allow precision upto Float32 in Scala as a result of which there are issues when one tries to load a model trained using float64 (in another language binding).

This also fixes two long standing issues : fixes https://github.com/apache/incubator-mxnet/issues/11315 & fixes https://github.com/apache/incubator-mxnet/issues/10338

### Essentials ###

## Comments ##
- Need to complete the F64 support in other classes as well, and then test out and compare training of a model using float32 and float64. The comparison would be in terms of the precision of the loss, accuracy of the trained model, memory occupied by the model during training process.

### DIFF
--- a/scala-package/core/src/main/scala/org/apache/mxnet/MX_PRIMITIVES.scala
+++ b/scala-package/core/src/main/scala/org/apache/mxnet/MX_PRIMITIVES.scala
@@ -31,9 +31,11 @@ object MX_PRIMITIVES {
     * more supported primitives in the future. Currently Float and Double
     * * are supported.
     */
-  trait MX_NUMBER_LIKE {
+  trait MX_NUMBER_LIKE extends Ordered[MX_NUMBER_LIKE]{
 
     def toString: String
+
+    def unary_- : MX_NUMBER_LIKE
   }
 
   /**
@@ -43,12 +45,15 @@ object MX_PRIMITIVES {
   class MX_FLOAT(val data: Float) extends MX_NUMBER_LIKE {
 
     override def toString: String = data.toString
+
+    override def unary_- : MX_NUMBER_LIKE = new MX_FLOAT(data.unary_-)
+
+    override def compare(that: MX_NUMBER_LIKE): Int = this.data.compareTo(that.asInstanceOf[MX_FLOAT].data)
   }
 
   implicit def FloatToMX_Float(d : Float): MX_FLOAT = new MX_FLOAT(d)
 
-
-  implicit def MX_FLoatToFloat(d: MX_FLOAT) : Float = d.data
+  implicit def MX_FloatToFloat(d: MX_FLOAT) : Float = d.data
 
   /**
     * Mimics Double in Scala.
@@ -57,12 +62,14 @@ object MX_PRIMITIVES {
   class MX_Double(val data: Double) extends MX_NUMBER_LIKE {
 
     override def toString: String = data.toString
+
+    override def unary_- : MX_NUMBER_LIKE = new MX_Double(data.unary_-)
+
+    override def compare(that: MX_NUMBER_LIKE): Int = this.data.compareTo(that.asInstanceOf[MX_Double].data)
   }
 
   implicit def DoubleToMX_Double(d : Double): MX_Double = new MX_Double(d)
 
-
   implicit def MX_DoubleToDouble(d: MX_Double) : Double = d.data
-
 
 }
--- a/scala-package/core/src/main/scala/org/apache/mxnet/MX_PRIMITIVES.scala
+++ b/scala-package/core/src/main/scala/org/apache/mxnet/MX_PRIMITIVES.scala
@@ -31,24 +31,24 @@ object MX_PRIMITIVES {
     * more supported primitives in the future. Currently Float and Double
     * * are supported.
     */
-  trait MX_NUMBER_LIKE extends Ordered[MX_NUMBER_LIKE]{
+  trait MX_PRIMITIVE_TYPE extends Ordered[MX_PRIMITIVE_TYPE]{
 
     def toString: String
 
-    def unary_- : MX_NUMBER_LIKE
+    def unary_- : MX_PRIMITIVE_TYPE
   }
 
   /**
     * Mimics Float in Scala.
     * @param data
     */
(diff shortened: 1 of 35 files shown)


piyushghai:
@andrewfayres @lanking520 @zachgk

piyushghai:
@lanking520 I have already added the Scala tests in NDArraySuite.scala class. It's in this commit : https://github.com/apache/incubator-mxnet/pull/13678/commits/5529f94796269d3ba94d74e24c0f98c4dd43b0df

scala-package/core/src/main/scala/org/apache/mxnet/MX_PRIMITIVES.scala
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+  * This defines the basic primitives we can use in Scala for mathematical
+  * computations in NDArrays. This gives us a flexibility to expand to
+  * more supported primitives in the future. Currently Float and Double
+  * are supported.
+  */
+object MX_PRIMITIVES {
+
+  /**
```
lanking520: duplicated comments

scala-package/core/src/main/scala/org/apache/mxnet/NDArray.scala
```diff
@@ -262,16 +263,32 @@ object NDArray extends NDArrayBase {
...
+  }
+
+
   // Perform power operator
   def power(lhs: NDArray, rhs: NDArray): NDArray = {
     NDArray.genericNDArrayFunctionInvoke("_power", Seq(lhs, rhs))
   }
 
-  def power(lhs: NDArray, rhs: Float): NDArray = {
+  def power(lhs: NDArray, rhs: MX_PRIMITIVE_TYPE): NDArray = {
```
lanking520: curious what will be the input from the user perspective? Will this be `typeT` or simply the name as `MX_PRIMITIVE_TYPE`? Will users be able to get information what kind of types should be used here. If a new data type (for example `int8`) is not supported currently, will the user be able to get the correct error messages?

piyushghai: The input from the user perspective will be either a ```Double``` or a ```Float```. 
The users in the scala docs would see a link to MX_PRIMITIVE_TYPE for support, which defines the primitives as Float or a Double.

If an incompatible datatype is passed, the IDE/compiler would throw an error because it would not be able to resolve the appropriate implicit conversion for MX_PRIMITIVE_TYPE.

scala-package/core/src/main/scala/org/apache/mxnet/MX_PRIMITIVES.scala
piyushghai: Sure. Will remove them.

scala-package/core/src/main/scala/org/apache/mxnet/NDArray.scala
piyushghai: As discussed offline, I'll add comments on the methods changed in NDArray.

piyushghai:

piyushghai:
@mxnet-label-bot Add [pr-awaiting-review]

scala-package/core/src/main/scala/org/apache/mxnet/javaapi/NDArray.scala
```diff
@@ -172,6 +189,18 @@ object NDArray extends NDArrayBase {
...
     sourceArr.asScala.map(ele => Float.unbox(ele)).toArray, shape, ctx)
 
+  /**
+    * Create a new NDArray that copies content from source_array.
+    * @param sourceArr Source data (list of Doubles) to create NDArray from.
+    * @param shape shape of the NDArray
+    * @param ctx The context of the NDArray, default to current default context.
+    * @return The created NDArray.
+    */
+  def arrayWithDouble(sourceArr: java.util.List[java.lang.Double], shape: Shape,
```
andrewfayres: Does anything prevent us from using a java.util.Collection instead of a list? If we can do that it'll also let people use the same method with Sets, queues, etc.

scala-package/core/src/test/scala/org/apache/mxnet/IOSuite.scala
```diff
@@ -303,5 +303,32 @@ class IOSuite extends FunSuite with BeforeAndAfterAll {
     assert(dataDesc(0).layout == Layout.NTC)
     assert(labelDesc(0).dtype == DType.Int32)
     assert(labelDesc(0).layout == Layout.NT)
+
+
+    // Test with passing Float64 hardcoded as Dtype of data
+    val dataIter4 = new NDArrayIter(
+      IO.initDataDesc(data, false, "data", DType.Float64, Layout.NTC),
+      IO.initDataDesc(label, false, "label", DType.Int32, Layout.NT),
```
andrewfayres: Any reason for the label to not also have DType.Float64? I know people don't really use that for labels but we added the capability so should test it as well.

scala-package/core/src/main/scala/org/apache/mxnet/javaapi/NDArray.scala
piyushghai: That seems a good suggestion. Basically we should use Iterable as a super class for this method, rather than a subclass. 

I can take it up in a separate PR other than this. 
I created a backlog JIRA item to track this : https://issues.apache.org/jira/browse/MXNET-1293

scala-package/core/src/test/scala/org/apache/mxnet/IOSuite.scala
piyushghai: The default DType in DataDesc is Float32. In this test, I'm trying to show here that not just Float64, but any other DType would also work with Data Descriptors :) 

And yes, for labels, people don't require higher precision datatypes.

lanking520:
@piyushghai please Satisfy the lint god (╯‵□′)╯︵┻━┻

lanking520:
@piyushghai please Satisfy the lint god (‵□′)

PR #13682, opened 2018-12-19
Add a docs build (with error checking) option to the dev_menu.py script 

aaronmarkham:
## Description ##
This PR extends the `dev_menu.py` script to give it a docs build option. 🔥 
It leverages the existing `deploy_docs` function from CI that is basically a straight call to `make docs SPHINXOPTS=-W`. This will check your build for docs errors, so this makes it just a little more easier to test your dev work before making a PR.

As with any `make docs` run, make sure you update `/docs/settings.ini` to choose which doc sets to run, otherwise you're running them all! 😲 

## Comments
Unrelated to this change is that while getting the script to function, I noticed the script has a couple of deviations in requirements from a standard user installation of MXNet. These aren't in our user setup scripts:

1. `yaml` Python module (pip install)
2. `ccache` (apt install)

So to run this script you need these! 

I tried using a docker variation for giggles and ran into security issues at one point, so there are probably some additional config steps to make the docker options work for a regular user.

It also requires Python 3. It will fail to run on Python 2 (around line 75). I think that since we still support Python 2 the script should be updated accordingly. Again, this has nothing to do with this feature I'm adding. I'm just commenting on the current situation with the script that is already in the repo.

## Experimental
I experimented with adding an option to run the full website build since that would be pretty cool to offer for easy dev & testing, but I put that off after seeing that it's not quite setup to handle a bunch of parameters. Sure, it could be done, but I didn't want to make so many changes without discussing the best approach. I had a --website option going that took in the 5 params needed, but then found I'd need to rewrite/extend the commands functions, so I didn't. Saving this kind of update for later since I'd also like to make it easy to pick which doc set(s) to run (python/c/scala/java/clojure/r/etc). Ping me if anyone wants to help with that.

### DIFF
--- a/dev_menu.py
+++ b/dev_menu.py
@@ -115,6 +115,8 @@ COMMANDS = OrderedDict([
         CMake(),
         create_virtualenv_default,
     ]),
+    ('[Website and docs build] Will build to docs/_build/html/',
+        "ci/docker/runtime_functions.sh deploy_docs"),
     ('[Docker] sanity_check. Check for linting and code formatting.',
         "ci/build.py --platform ubuntu_cpu /work/runtime_functions.sh sanity_check"),
     ('[Docker] Python3 CPU unittests',
@@ -225,4 +227,3 @@ def main():
 
 if __name__ == '__main__':
     sys.exit(main())
-


aaronmarkham:
@larroy - added this feature per your suggestion. Also, I have some comments about the script in this PR's description. In particular, I'd want to sync up the dependencies and make sure those are in the setup scripts... yaml is fine, but do we need to require ccache? I'm ok either way... just not sure if a basic build option really needs that.

Roshrini:
@larroy @nswamy Can you review this PR?

PR #13564, opened 2018-12-06
Enhancement of split op with indices option

HyperZealot:
## Description ##
Support split by arbitrary ranges through addition of indices option.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] Indices option for split op
- [x] Unit tests

PR #13687, opened 2018-12-19
split_v2 operator

HyperZealot:
## Description ##
New version of split operator to match the behavior of [numpy.split](https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.split.html)

### Essentials ###

### Changes ###
- [x] New Split_v2 operator
- [x] Unit tests

### DIFF
--- a/python/mxnet/ndarray/ndarray.py
+++ b/python/mxnet/ndarray/ndarray.py
@@ -47,7 +47,7 @@ __all__ = ["NDArray", "concatenate", "_DTYPE_NP_TO_MX", "_DTYPE_MX_TO_NP", "_GRA
            "imdecode", "lesser", "lesser_equal", "logical_and", "logical_or", "logical_xor",
            "maximum", "minimum", "moveaxis", "modulo", "multiply", "not_equal", "onehot_encode",
            "power", "subtract", "true_divide", "waitall", "_new_empty_handle", "histogram",
-           "to_dlpack_for_read", "to_dlpack_for_write", "from_dlpack"]
+           "split_v2", "to_dlpack_for_read", "to_dlpack_for_write", "from_dlpack"]
 
 _STORAGE_TYPE_UNDEFINED = -1
 _STORAGE_TYPE_DEFAULT = 0
@@ -1132,6 +1132,14 @@ fixed-size items.
         this array as data.
         """
         return op.split(self, *args, **kwargs)
+
+    def split_v2(self, *args, **kwargs):
+        """Convenience fluent method for :py:func:`split_v2`.
+
+        The arguments are the same as for :py:func:`split_v2`, with
+        this array as data.
+        """
+        return split_v2(self, *args, **kwargs)
 
     def slice(self, *args, **kwargs):
         """Convenience fluent method for :py:func:`slice`.
@@ -3901,6 +3909,12 @@ def histogram(a, bins=10, range=None):
         Values outside the range are ignored. The first element of the range must be less than or
         equal to the second. range affects the automatic bin computation as well, the range will
         be equally divided by the number of bins.
+
+    Returns
+    -------
+    NDArray
+        A created array.
+
     """
 
     # pylint: disable= no-member, protected-access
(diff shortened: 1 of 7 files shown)


Roshrini:
@HyperZealot Thanks for working on this.

python/mxnet/ndarray/ndarray.py
```diff
@@ -1133,6 +1133,14 @@ def split(self, *args, **kwargs):
         """
         return op.split(self, *args, **kwargs)
 
+    def split_v2(self, *args, **kwargs):
+        """Convenience fluent method for :py:func:`split_v2`.
+
+        The arguments are the same as for :py:func:`split_v2`, with
+        this array as data.
+        """
+        return split_v2(self, *args, **kwargs)
```
zheng-da: call itself?

python/mxnet/symbol/symbol.py
```diff
@@ -2967,4 +2980,44 @@ def histogram(a, bins=10, range=None, **kwargs):
...
+    """
+    indices = []
+    sections = 0
+    if isinstance(indices_or_sections, int):
+        sections = indices_or_sections
+    elif isinstance(indices_or_sections, tuple):
+        indices = [0] + list(indices_or_sections)
+    else:
+        raise ValueError('indices_or_sections must either int or tuple of ints')
+    return _internal._split_v2(ary, indices, axis, squeeze_axis, sections)
```
zheng-da: why the symbol implementation and ndarray implementation are different? the ndarray impl always computes indices and here it may pass the number of sections directly.

src/operator/tensor/matrix_op-inl.h
```diff
@@ -2520,6 +2520,323 @@ void SpaceToDepthOpForward(const nnvm::NodeAttrs& attrs,
...
+              " **Note** that setting `squeeze_axis` to ``true`` removes axis with length 1"
+              " only along the `axis` which it is split."
+              " Also `squeeze_axis` can be set to ``true``"
+              " only if ``input.shape[axis] == num_outputs``.");
+    DMLC_DECLARE_FIELD(sections).set_default(0)
+    .describe("Number of sections if equally splitted. Default to 0 which means split by indices.");
+  }
+};  // struct SplitParam
+
+inline TShape GetSplitIndices(const TShape& ishape, int axis, int sections) {
```
zheng-da: maybe return `Tuple` may make more sense.

HyperZealot: TShape is also a Tuple: https://github.com/dmlc/tvm/blob/master/nnvm/include/nnvm/tuple.h#L325. But maybe I can also switch to nnvm::Tuple<int>.

HyperZealot:
@apeforest Please also avoid v2 of the same question as well, I think ur question was asked by @rongzha1 already.
This version is to support split of an array at arbitrary points along as axis while the previous split only supported equal splits along an axis, to match numpy behavior of the same operator.
On the other hand, there was some API-breaking changes to cpp-package in my first attempt to modify the original split operator [here](https://github.com/apache/incubator-mxnet/pull/13564) so that's why I'm moving to creation of a new operator.

Roshrini:
@zheng-da Can you take a look again to see if your comments are addressed? Thanks

src/operator/tensor/matrix_op-inl.h
```diff
@@ -2520,6 +2520,323 @@ void SpaceToDepthOpForward(const nnvm::NodeAttrs& attrs,
...
+                                  const DType *in_data, DType** out_data, const size_t* indices,
+                                  const size_t num_sections, const size_t axis_size,
+                                  const size_t trailing_size) {
+    size_t idx = i / trailing_size % axis_size;
+    size_t target = 0;
+    for (size_t section = 0; section < num_sections; target = section++) {
+      if (indices[section] > idx) {
+        break;
+      }
+    }
```
zheng-da: another way to write this is:
```c++
for (size_t section = 0; section < num_sections && indices[section] <= idx; target = section++);
```

src/operator/tensor/matrix_op-inl.h
```diff
@@ -2520,6 +2520,323 @@ void SpaceToDepthOpForward(const nnvm::NodeAttrs& attrs,
...
+    DType* target_data = out_data[target];
+    const size_t mid_idx = idx - indices[target];
+    const size_t head_idx = i / (trailing_size * axis_size);
+    const size_t tail_idx = i % trailing_size;
+    const size_t section_size = indices[target + 1] - indices[target];
+    const size_t target_idx =
+      head_idx * trailing_size * section_size + mid_idx * trailing_size + tail_idx;
+    target_data[target_idx] = in_data[i];
```
zheng-da: I'm a little concerned about this kernel. It takes a lot of computation to copy an element from the original array to a destination array. At least we should copy the entire row unless we split in the last dimension.

src/operator/tensor/matrix_op-inl.h
```diff
@@ -2520,6 +2520,323 @@ void SpaceToDepthOpForward(const nnvm::NodeAttrs& attrs,
...
+  }
+};
+
+struct ConcatenateKernel {
+  /*!
+   * \brief Map function for split operator indices option
```
zheng-da: This comment is confusing.

HyperZealot: Done.

src/operator/tensor/matrix_op-inl.h
HyperZealot: done

src/operator/tensor/matrix_op-inl.h
HyperZealot: I doubt if memcpy is optimal for all cases actually, and splitting the kernel into 2 versions would also make the code harder to maintain and manage, what do you think if we treat that as a further optimization and do it in a follow-up PR so that the users of DGL can enjoy this new op ASAP?

Issue #13658, opened 2018-12-16
Converting MX array to DLPack crashes when MX array goes out-of-scope


jermainewang:
## Description
Converting MX NDArray to DLPack, then to other framework's DLPack-compatible NDArray causes memory corruption when the origin MX NDArray goes out-of-scope.

## Environment info (Required)

```
----------Python Info----------
Version      : 3.5.2
Compiler     : GCC 5.4.0 20160609
Build        : ('default', 'Nov 12 2018 13:43:14')
Arch         : ('64bit', '')
------------Pip Info-----------
Version      : 18.1
Directory    : /usr/local/lib/python3.5/dist-packages/pip
----------MXNet Info-----------
Version      : 1.4.0
Directory    : /usr/local/lib/python3.5/dist-packages/mxnet
Commit Hash   : 1f73c5d9d308a690b57ea1b474d2ba99ca06c476
----------System Info----------
Platform     : Linux-4.19.4-arch1-1-ARCH-x86_64-with-Ubuntu-16.04-xenial
system       : Linux
node         : 17d02f89890e
release      : 4.19.4-arch1-1-ARCH
version      : #1 SMP PREEMPT Fri Nov 23 09:06:58 UTC 2018
----------Hardware Info----------
machine      : x86_64
processor    : x86_64
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                8
On-line CPU(s) list:   0-7
Thread(s) per core:    2
Core(s) per socket:    4
Socket(s):             1
NUMA node(s):          1
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 62
Model name:            Intel(R) Xeon(R) CPU E5-1620 v2 @ 3.70GHz
Stepping:              4
CPU MHz:               1812.064
CPU max MHz:           3900.0000
CPU min MHz:           1200.0000
BogoMIPS:              7384.55
Virtualization:        VT-x
Hypervisor vendor:     vertical
Virtualization type:   full
L1d cache:             32K
L1i cache:             32K
L2 cache:              256K
L3 cache:              10240K
NUMA node0 CPU(s):     0-7
[...]

PR #13698, opened 2018-12-20
Fix NDArray ToDLPack Bug

wkcn:
## Description ##
Fix Issue #13658 
The variable `shape` in `DLTensor` is pointed to the `TShape` instance in `tblob_` in `NDArray`.
However, the `NDArray` instance in `NDArrayDLManager` is different from the original `NDArray` instance.

Test Case written by @jermainewang 
```python
import mxnet as mx
import numpy as np
import tvm

def foo():
  x = mx.nd.array([0, 5], dtype='int64')
  dl = x.to_dlpack_for_read()
  return tvm.nd.from_dlpack(dl)

for i in range(10):
  y = foo()
  y.asnumpy()
```

Command:
```bash
for i in {0..100}; do echo $i && python3 t.py || break ; done
```

The test is passed more than 300 times on Ubuntu 16.04.

### Essentials ###
- [x] Tiny Changes. The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)

### Changes ###
- [x] `Fix NDArray::ToDLPack`

## Comments ##

### DIFF
--- a/src/ndarray/ndarray.cc
+++ b/src/ndarray/ndarray.cc
@@ -330,11 +330,10 @@ struct NDArrayDLManager {
 };
 
 DLManagedTensor* NDArray::ToDLPack() const {
+  CHECK(!is_none()) << "NDArray is not initialized";
   NDArrayDLManager* dlmanager(new NDArrayDLManager);
   dlmanager->handle = *this;
-  if (!is_none()) {
-    dlmanager->tensor.dl_tensor = dlmanager->handle.data().dltensor();
-  }
+  dlmanager->tensor.dl_tensor = dlmanager->handle.data().dltensor();
   dlmanager->tensor.manager_ctx = dlmanager;
   dlmanager->tensor.deleter = [](DLManagedTensor* dlmanager){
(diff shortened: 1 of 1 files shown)


src/ndarray/ndarray.cc
```diff
@@ -330,11 +330,10 @@ struct NDArrayDLManager {
 };
 
 DLManagedTensor* NDArray::ToDLPack() const {
+  CHECK(!is_none()) << "NDArray is not initialized";
```
jermainewang: Just to confirm. Such check should never fail as `ToDLPack` will only be called after `wait_to_read` or `wait_to_write` so the NDArray should always be initialized.

wkcn: Yes. But when users use C++ API directly, they may forget to call WaitToRead or WaitToWrite before calling ToDLPack.

jermainewang:
LGTM. Thanks @wkcn .

PR #13708, opened 2018-12-21
Fixing a symlink issue with R install

lupesko:
## Description ##
Fixing a symlink issue with R install, which prevents the R install instructions from working as expected.
Verified that the fix is working.

### DIFF
--- a/docs/install/index.md
+++ b/docs/install/index.md
@@ -697,13 +697,13 @@ To run MXNet you also should have OpenCV and OpenBLAS installed. You may install
 
 ```bash
 brew install opencv
-brew install openblas@0.3.1
+brew install openblas
 ```
 
-Add a soft link to the OpenBLAS installation. This example links the 0.3.1 version:
+To ensure MXNet R package runs with the version of OpenBLAS installed, create a symbolic link as follows:
 
 ```bash
-ln -sf /usr/local/opt/openblas/lib/libopenblasp-r0.3.* /usr/local/opt/openblas/lib/libopenblasp-r0.3.1.dylib
+ln -sf /usr/local/opt/openblas/lib/libopenblas.dylib /usr/local/opt/openblas/lib/libopenblasp-r0.3.1.dylib
 ```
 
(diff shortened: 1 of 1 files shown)


lupesko:
@ankkhedia @anirudhacharya - please take a look.

lupesko:
@aaronmarkham you may want to merge this one...

docs/install/index.md
```diff
@@ -703,7 +703,7 @@ brew install openblas@0.3.1
 Add a soft link to the OpenBLAS installation. This example links the 0.3.1 version:
 
 ```bash
-ln -sf /usr/local/opt/openblas/lib/libopenblasp-r0.3.* /usr/local/opt/openblas/lib/libopenblasp-r0.3.1.dylib
+ln -sf /usr/local/opt/openblas/lib/libopenblasp-r0.3.*.dylib /usr/local/opt/openblas/lib/libopenblasp-r0.3.1.dylib
```
nswamy: @ankkhedia @anirudhacharya why openblas@0.3.1? there are bug fixes/performance improvements post this version https://github.com/xianyi/OpenBLAS/releases

Usually you create symlink without the version to a specific version something like this. if there are multiple 0.3* version, the command fails
```
ln -sf /usr/local/opt/openblas/lib/libopenblasp-r0.3.1.dylib /usr/local/opt/openblas/lib/libopenblasp.dylib
``

Roshrini:
@mxnet-label-bot Add [R, pr-awaiting-response]

sandeep-krishnamurthy:
@ankkhedia @anirudhacharya - can you please take a look at Naveen's comment?

docs/install/index.md
anirudhacharya: it picks up the BLAS version from the machine where the R-package was built. We explored not hard-linking to a particular BLAS version, but that is not trivial. Will revisit it again. Either ways, this PR need not be blocked because of that right?

aaronmarkham:
@lupesko is this ready to ship?

PR #13725, opened 2018-12-24
disable error checking when building old versions

aaronmarkham:
## Description ##
This PR fixes a bug in the options settings for the docs build for the website publishing job in CI. If you ran master first the error checking would get turned on and then never turn off for the older versions.

## Testing
This will fail now:
./build_all_version.sh "master;v1.3.x" "master;1.3.1"
So the workaround is to run master last, but with this PR, it will work generating different versions in any order.

### DIFF
--- a/docs/build_version_doc/build_all_version.sh
+++ b/docs/build_version_doc/build_all_version.sh
@@ -122,6 +122,8 @@ function checkout () {
     git pull
     # master gets warnings as errors for Sphinx builds
     OPTS="-W"
+  else
+    OPTS=
   fi
   git submodule update --init --recursive
   cd ..


docs/build_version_doc/build_all_version.sh
```diff
@@ -122,6 +122,8 @@ function checkout () {
     git pull
     # master gets warnings as errors for Sphinx builds
     OPTS="-W"
+  else
```
sandeep-krishnamurthy: OPTS is already empty (set at line 47). Why is this required? If tag is master OPTS is set to "-w" otherwise it is left as is which is empty.

aaronmarkham: Once master gets called OPTS is set, but when another version is called it should be unset.

PR #13745, opened 2018-12-29
Fix the order of error term's operands

wlbksy:
Error term should be "actual minus predicted".

This PR should be no op, since all related error terms are either squared or transformed into absolute value.

### DIFF
--- a/python/mxnet/gluon/loss.py
+++ b/python/mxnet/gluon/loss.py
@@ -101,7 +101,7 @@ class Loss(HybridBlock):
 class L2Loss(Loss):
     r"""Calculates the mean squared error between `pred` and `label`.
 
-    .. math:: L = \frac{1}{2} \sum_i \vert {pred}_i - {label}_i \vert^2.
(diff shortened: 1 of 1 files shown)


python/mxnet/gluon/loss.py
```diff
@@ -101,7 +101,7 @@ def hybrid_forward(self, F, x, *args, **kwargs):
 class L2Loss(Loss):
     r"""Calculates the mean squared error between `pred` and `label`.
```
vishaalkapoor: by the same reasoning, this should be swapped as `label` is the subject.

vishaalkapoor: same elsewhere.

wlbksy: Thanks, good finding!

vishaalkapoor: Thanks!

vishaalkapoor:
LGTM!

Issue #13598, opened 2018-12-10
More fine-grained operator implementation dispatch & memory planning flow 


eric-haibin-lin:
## Existing Execution Flow
```
g = graph()
shapes = g.infer_shape()
types = g.infer_type()
storage_types, dispatch_modes = g.infer_storage_type()
memory_plan = nnvm::plan_memory() // which calls node.finplace_option(node.attr)
for node in g:
  fcompute = get_fcompute(node)
  fcompute(x)
```
### The drawback of the existing flow
- the selection of MKL/CPU/GPU/CUDNN implementation happens after graph attribute inference and memory planning, **memory planning is thus not aware of the implementation that will be used for execution in the future, which may result in sub-optimal result**. For example, the memory inplace option may vary depending on the accelerator backend (the new version of CUDNN enables x/dx inplace for _backward_conv).
- some sparse operator need to access dtype/shape information to decide which implementation to invoke for execution, and whether to perform fallback. This information is not yet exposed in the existing infer storage type interface. 

## Alternative Flow
op implementations
```
CovolutionComputeCUDNN(const nnvm::NodeAttrs& attrs,
                       const OpContext& ctx,
                       const std::vector<TBlob>& inputs,
                       const std::vector<OpReqType>& req,
                       const std::vector<TBlob>& outputs) {
  // 1st CUDNN implementation goes here 
}

CovolutionComputeMKL(const nnvm::NodeAttrs& attrs,
                       const OpContext& ctx,
                       const std::vector<NDArray>& inputs,
                       const std::vector<OpReqType>& req,
                       const std::vector<NDArray>& outputs) {
  // MKL implementation goes here 
}

CovolutionComputeImplGPU(const nnvm::NodeAttrs& attrs,
                       const OpContext& ctx,
                       const std::vector<TBlob>& inputs,
                       const std::vector<OpReqType>& req,
                       const std::vector<TBlob>& outputs) {
  // GPU implementation goes here 
}

CovolutionComputeImplCPU(const nnvm::NodeAttrs& attrs,
                       const OpContext& ctx,
                       const std::vector<TBlob>& inputs,
                       const std::vector<OpReqType>& req,
                       const std::vector<TBlob>& outputs) {
  // CPU implementation goes here 
}
```
new finferstorage_ex interface
```
void  FInferStorageTypeEx(const std::vector<TShape>& in_shapes,
                          const std::vector<int>& in_types,
                          const std::vector<int>& in_stype,
[...]

PR #13749, opened 2018-12-31
Add NHWC layout support to Pooling (cpu, gpu cuda, gpu cuDNN)

DickJC123:
## Description ##
This is a continuation of PR #13362 started by @ptrendx titled "Add NHWC layout support to Pooling (cuDNN only)".  We're in agreement that I'm the best one to push this work forward to a point of final acceptance.

Based on reviewer comments, I've expanded the support of the newly introduced 'layout' parameter to include cpu and cuda implementations, not just cudnn.  Also 3D support in cuDNN and graceful fallback for MKLDNN.  I've re-enabled and expanded the testing of test_operator_gpu.py:test_pooling_versions.  I understand and avoid the earlier flakiness of that test, which was due to inclusion of the pooling_v1 operator in some problem domains where its definition differs.  The initial commit here is identical to the last one made by @ptrendx for PR #13362.  Follow-up commits will add this new functionality and additional description.

### Essentials ###

### Changes ###

## Comments ##


anirudhacharya:

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "3rdparty/mshadow"]
 	path = 3rdparty/mshadow
-	url = https://github.com/dmlc/mshadow.git
+	url = https://github.com/DickJC123/mshadow.git
```
apeforest: temp reminder to update the mshadow submodule. It would be great if you could also refer the mshadow PR here.

DickJC123: Don't merge this PR until the mshadow PR has been accepted:
https://github.com/dmlc/mshadow/pull/368.  The new mshadow functionality is tested in https://github.com/apache/incubator-mxnet/pull/13857.

sandeep-krishnamurthy:
@apeforest - Can you please help take a look back at this PR?

DickJC123:
I'll be updating the mshadow repo and SHA, after which this PR is ready for merging.

DickJC123:
@apeforest I've incorporated the newest mshadow (with now-merged https://github.com/dmlc/mshadow/pull/368), so this PR is ready for final review.

pengzhao-intel:
@DickJC123 It's great to see the improvements and thank you to cover the MKLDNN backend as well :)
Do you have the performance data to show how much benefit from the new data format?

I am very open to see the more data format are supported by MXNet to max the perforamnce on different HW. Regarding the memory format switching, AFAIK, several approaches can be used as below.
- OP level switch, as this PR. The advantage is we can control the format in the fine level and maybe get the best performance with tuning; however, it will increase the complexity for the user to transpose the format for each OP.
- Model level switch, like Keras, the format can be specified by .json file with `"image_data_format": "channels_last"/"channels_first"`. The advantage is that it is quick easy for the user to switch the data format and the model can keep the constants for the different format.
- Mix OP and Model level switch

Currently, the TF fully supports NHWC and partially for NCHW, like what we are doing for MXNet now. And TF is going to resolve the legacy issue by switching transparently.
https://www.tensorflow.org/guide/performance/overview#data_formats

> The brief history of these two formats is that TensorFlow started by using NHWC because it was a little faster on CPUs. In the long term, we are working on tools to auto rewrite graphs to make switching between the formats transparent and take advantages of micro optimizations where a GPU op may be faster using NHWC than the normally most efficient NCHW.

For a short term, I am fine with this PR and I don't want to block the improvement.
For a long term, could you share your plan and big picture to the community for the supports of two data format? Maybe a RFC or design proposal is prefer to send @dev.

Thanks a lot and feel free to correct me :)

DickJC123:
Thanks for your detailed thoughts.  I view this PR as filling in a missing piece in the MXNet support for NHWC (with Convolution and Batchnorm already supporting it).  The performance benefit of having end-to-end support of NHWC (thereby avoiding transposes) was demonstrated for Volta hardware in https://mlperf.org/results/.  We're offering this benefit to upstream with this PR (along with others to come).  Once we complete the NHWC API support with this PR, I look forward to joining the discussion on how to make experimenting with layout easier for the user or done automatically.

DickJC123:
PR https://github.com/apache/incubator-mxnet/pull/13857, which includes a test of the new mshadow float32->float16 cast-with-rounding (as relied on by this PR's test), has been merged.  Still waiting on final review for this PR.

src/operator/nn/pool_utils.h
```diff
@@ -98,14 +98,16 @@ struct lp_grad<DType, 1> {
 template<typename DType>
 struct lp_grad<DType, 2> {
   static MSHADOW_XINLINE DType Map(const DType grad, const DType in_data, const DType out_data) {
-    return grad * in_data / out_data;
+    // Avoid nan result if both grad and out_data are 0.
+    return (grad == DType(0.0)) ? DType(0.0) : grad * in_data / out_data;
```
apeforest: Why not also check out_data == DType(0.0) here if it can be zero, to avoid overflow?

src/operator/nn/pool_utils.h
```diff
@@ -98,14 +98,16 @@ struct lp_grad<DType, 1> {
...
   }
 };
 
 template<typename DType>
 struct lp_grad<DType, 3> {
   static MSHADOW_XINLINE DType Map(const DType grad, const DType in_data, const DType out_data) {
-    return grad * in_data * in_data / (out_data * out_data);
+    // Avoid nan result if both grad and out_data are 0.
+    return (grad == DType(0.0)) ? DType(0.0) : grad * in_data * in_data / (out_data * out_data);
```
apeforest: Why not also check out_data == DType(0.0) here if it can be zero, to avoid overflow?

DickJC123: I introduced this to quiet some nans that were causing test failures.  I'm reviewing the specifics to see if a better solution exists as suggested in your comment, so stay tuned.

TaoLv:
Maybe it's out of the scope of this PR, but I would suggest to remove layout parameters from all operator APIs in the next major release. It's framework that should take care of the performance of model. We cannot ask users to complicate their models to get some performance benefit from one specific hardware or backend. The change may be nightmare if they want to change back to other hardwares or backends.

src/operator/nn/pooling.cc
```diff
@@ -421,11 +463,16 @@ NNVM_REGISTER_OP(_backward_Pooling)
 .set_attr<nnvm::FInplaceOption>(
     "FInplaceOption",
     [](const NodeAttrs &attrs) {
-#if MXNET_USE_CUDNN == 1
-  return std::vector<std::pair<int, int> >();
-#else
-  return std::vector<std::pair<int, int> >{{1, 0}};
+#if MXNET_USE_MKLDNN == 1 && MXNET_USE_CUDA == 0 && MXNET_USE_CUDNN == 0
```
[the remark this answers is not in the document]
DickJC123: So unfortunately, all enabled backend operator implementations must live with the same FInPlaceOptions.  I voiced concern about this restriction to @eric-haibin-lin and he graciously put some thought into: https://github.com/apache/incubator-mxnet/issues/13598. The issue has seen no activity in a month, but I'd like to see it reawakened.

Until then, the code you highlight enforces that the most conservatinve "least common denominator" FInPlaceOptions (i.e. none) are in effect for the enabled Pooling implementations. Since an operator cannot rely on FInPlace actually occurring, there should be no functional regression. And perf-wise, a user must understand that at this point enabling multiple back-ends may have some performance consequences due to the FInPlaceOption implementation.

tests/python/gpu/test_operator_gpu.py
```diff
@@ -608,6 +608,72 @@ def test_convolution_versions():
 
 
 @with_seed()
+def test_pooling_with_convention():
```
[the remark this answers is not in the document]
DickJC123: As part of finalizing this PR, I greatly expanded the coverage of test_pooling_versions(), to the point where it basically repeats the test cases covered here.  In response to your comment, I expanded test_pooling_versions() to include some new subtleties (like whether default parameters are explicitly passed or not) to match what was done in test_pooling_with_convention().  Given that, I'm now removing test_pooling_with_convention().

src/operator/nn/pooling.cc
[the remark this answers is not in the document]
DickJC123: Sorry for any confusion.  Perhaps my comments in the code for FInplaceOption speak to the issue more directly:
```
  // Since this routine is not provided the cpu/gpu context info, only in the case
  // where CUDA and CUDNN implementations are not available can we be sure the MKLDNN
  // implementation will be employed.  The MKLDNN FInplaceOptions are not compatible
  // with the other (i.e. cpu, cuda and cudnn) implementations.
```

DickJC123:
I agree with @TaoLv that a new approach to layout must wait until the next major (API-breaking) release.  I look forward to the discussions of how to free users from coding layout considerations into their models and how to allow the framework to make the best backend-specific layout choices.

I've attempted to address reviewer comments to-date.  Requesting further input, e.g. from those additionally listed reviewers @zheng-da @szha @zhreshold.  Thanks in advance!

src/operator/nn/pool_utils.h
DickJC123: I've pushed my solution to your comment in commit https://github.com/apache/incubator-mxnet/pull/13749/commits/098bc49f1d288ea9f2b64453aefcc1537ca5254e.

The checking of grad == 0.0 that you highlighted only succeeded because of the quirks of our check_consistency() routine in test_utils.py, which uses the symbol forward() output as the gradient.  Per your suggestion, I'm now using a check of out_data == 0 as the more general way of quieting the test failures.

The test failures I was seeing often occurred in float16 lp-3 pooling.  By example, take the case where a pool window of 2 has identical inputs 2^-9 and 2^-9.  The forward output for this case is the cube root of (2^-9)^3 + (2^-9)^3.  If this is calculated in float16, the 2^-27 terms underflow to 0 and the output is 0.  The backward output is then grad * 2^-9 * 2^-9 / (0 * 0) = +inf (or nan if grad is also 0).  When performed in float32, no underflow occurs in the forward op, and +infs are avoided in the backward op.

My conclusion: float16 is ill-equipped to perform the forward pooling operation for lp-2 and lp-3.  Part of my solution here thus involves promoting the calculation to be in float32 for cpu and mxnet cuda implementations of float16-i/o pooling.  This is consistent with other operators like float16-i/o Convolution and Batchnorm, which perform internal calculations in float32.  I've run test_pooling_versions() thousands of times with no failures in this mode.

src/operator/nn/cudnn/cudnn_pooling-inl.h
```diff
@@ -125,7 +126,8 @@ class CuDNNPoolingOp {
     CHECK_EQ(s->dnn_handle_ownership_, mshadow::Stream<gpu>::OwnHandle);
     typename DataType<DType>::ScaleType alpha = 1.0f;
     typename DataType<DType>::ScaleType beta = 0.0f;
-    this->Init(s, in_data, out_data);
+    if (!this->Init(s, in_data, out_data))
+      LOG(FATAL) << "CuDNN Pooling invoked with unsupported parameters.";
```
apeforest: nit: is the official name cuDNN or CuDNN?

src/operator/nn/cudnn/cudnn_pooling-inl.h
```diff
@@ -167,102 +169,215 @@ class CuDNNPoolingOp {
...
                                             dtype_,
-                                            out.shape_[0],
-                                            out.shape_[1],
-                                            out.shape_[2],
-                                            out.shape_[3]));
+                                            oshape_nchw[0],
+                                            oshape_nchw[1],
+                                            oshape_nchw[2],
+                                            oshape_nchw[3]));
+      int window_height = param_.global_pool ? dshape_nchw[2] : param_.kernel[0];
```
apeforest: nit: use the name `kernel_height` and `kernel_width` to be consistent across this operator.

src/operator/nn/cudnn/cudnn_pooling-inl.h
DickJC123: Official name is cuDNN, even at the beginning of a sentence apparently.  Fixed.

src/operator/nn/cudnn/cudnn_pooling-inl.h
DickJC123: Done.

src/operator/nn/pooling.cc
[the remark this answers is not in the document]
pengzhao-intel: Thanks for the explanation. I understand the current limitation in here. 
My concern is the performance regression in the CPU side when _mxnet-cuXXmkl_ binary is used which actually is the widely used binary like sockeye, gluonNLP, gluonCV, etc. 
Before we resolved issue #13598 (btw, we're working on this now）, I think we need to keep this line **AS-IS**.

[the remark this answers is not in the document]
eric-haibin-lin: mxnet-cuXX-mkl pip binary statically links cudnn and will not be affected. @szha correct me if I am wrong. 

The potentially affected cases are:
- build mxnet with mkldnn and cuda, but without cudnn 
- build mxnet with neither mkldnn nor cuda.

As this operator is used in backward for training, I see the two cases above quite rare. Therefore, I do not think this is a blocking issue for this PR. 

Nonetheless, I suggest @DickJC123 create a github issue with this change so that people are aware and can be tracked later, and preferably run some performance testing to have a rough idea how big an impact it is making for the above two cases.

Let's continue discussions on  #13598 to avoid such problems in the future.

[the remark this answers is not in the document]
DickJC123: I originally set out to create a test case (and separate issue) to demonstrate the need for the  restrictive #if clause I put around the MKLDNN FInplaceOption.  I was unable to do so and in the process I came to realize that I misunderstood the meaning behind the {1,0} option.  In fact, it allows the framework to share the backward '1' input (the unused workspace gradient) with the backward '0' output (the pooling input gradient) where the additional MKLDNN workspace is present.  This sharing is compatible with CUDA and cuDNN Pooling implementations, which do not use a workspace or its gradient!  I'm therefore reverting back to a simple code guard around the MKLDNN FinplaceOption- one whose logic mirrors the other MKLDNN/no-MKLDNN switches of behavior within nn/pooling.cc.  No new issue is required.  With my last commit I've also expanded the pooling tests to cover cases that have the inplace tensor sharing, which requires the pooling input to be the same size as the output (and also then the workspace).

src/operator/nn/pooling.cc
```diff
@@ -421,11 +463,16 @@ NNVM_REGISTER_OP(_backward_Pooling)
...
+  // Since this routine is not provided the cpu/gpu context info, only in the case
+  // where CUDA and CUDNN implementations are not available can we be sure the MKLDNN
+  // implementation will be employed.  The MKLDNN FInplaceOptions are not compatible
+  // with the other (i.e. cpu, cuda and cudnn) implementations.
```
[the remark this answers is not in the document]
DickJC123: Done.

src/operator/nn/pooling.cc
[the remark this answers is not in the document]
eric-haibin-lin: Good to know MKLDNN impl is not affected

python/mxnet/gluon/nn/conv_layers.py
```diff
@@ -716,7 +718,8 @@ class MaxPool1D(_Pooling):
         If padding is non-zero, then the input is implicitly
         zero-padded on both sides for padding number of points.
     layout : str, default 'NCW'
-        Dimension ordering of data and weight. Only supports 'NCW' layout for now.
+        Dimension ordering of data and weight. Only supports 'NCW' and 'NWC'
+        (only with cuDNN) layouts for now.
```
szha: it's a bit ambiguous whether it's the class that requires cudnn, or it's the second layout.

DickJC123: Glad you pointed out this ambiguity.  I expanded the scope of this work mid-PR and provided cpu and cuda implementations of the 'transposed' layout pooling, making these descriptions incorrect.  I've changed the text now to be simpler and correct.  Ditto for Symbol's Pooling description.

src/operator/nn/cudnn/cudnn_pooling-inl.h
```diff
@@ -167,102 +169,215 @@ class CuDNNPoolingOp {
...
+                                            oshape_nchw[1],
+                                            oshape_nchw[2],
+                                            oshape_nchw[3]));
+      int kernel_height = param_.global_pool ? dshape_nchw[2] : param_.kernel[0];
+      int kernel_width = param_.global_pool ? dshape_nchw[3] : param_.kernel[1];
+      // CuDNN v7.1.4 backprop kernel doesn't support kernel sizes 9 and above.
+      // For reference see Fixed Issues section in
+      // https://docs.nvidia.com/deeplearning/sdk/cudnn-release-notes/rel_721.html#rel_721
+      #if CUDNN_VERSION == 7104
+      is_supported = kernel_height <= 8 && kernel_width <= 8;
```
szha: a CHECK here is probably more informative.

src/operator/nn/cudnn/cudnn_pooling-inl.h
```diff
@@ -81,7 +81,8 @@ class CuDNNPoolingOp {
     CHECK_EQ(s->dnn_handle_ownership_, mshadow::Stream<gpu>::OwnHandle);
     typename DataType<DType>::ScaleType alpha = 1.0f;
     typename DataType<DType>::ScaleType beta = 0.0f;
-    this->Init(s, in_data, out_data);
+    if (!this->Init(s, in_data, out_data))
+      LOG(FATAL) << "cuDNN Pooling invoked with unsupported parameters.";
```
szha: e.g. otherwise it's hard to figure out where the problem is with this error message.

src/operator/nn/mkldnn/mkldnn_pooling-inl.h
```diff
@@ -104,7 +104,8 @@ class MKLDNNPoolingBwd {
 inline bool SupportMKLDNNPooling(const PoolingParam &param) {
   return param.kernel.ndim() == 2 &&
          (param.pool_type == pool_enum::kMaxPooling ||
-          param.pool_type == pool_enum::kAvgPooling);
+          param.pool_type == pool_enum::kAvgPooling) &&
+         (!param.layout.has_value() || param.layout.value() == mshadow::kNCHW);
```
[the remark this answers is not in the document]
szha: @TaoLv are you tracking this separately?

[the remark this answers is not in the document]
TaoLv: Yes. This change looks good to me for now. Two possible follow-ups:
1. if layout parameter will be removed in the next major release, possibly we will not support NHWC.
2. otherwise, if needed, we will extend MKL-DNN integration to support NHWC. Currently I don't see any necessity for it.

Do you think I can put the suggestion for removing layout into https://github.com/apache/incubator-mxnet/issues/9686?

src/operator/nn/cudnn/cudnn_pooling-inl.h
DickJC123: We don't want to error-out here via CHECK() in this Supports() function, since there is a graceful fallback to the CUDA implementation.

src/operator/nn/cudnn/cudnn_pooling-inl.h
DickJC123: Thanks for this tip.  Updated per your suggestion.

DickJC123: Any other issues you'd like me to look at?

PR #13766, opened 2019-01-03
Fixed mailing list addresses that were swapped (doc change only)

ifeherva:
## Description ##
The user and dev mailing lists were swapped in the docs. This PR fixes it. There is no code or any functional change.


aaronmarkham:
I retriggered the failed test - let's let that pass and we can merge this change.

And yes, it would be nice if we could limit the CI process on docs changes. It does seem, however, that this is a non-trivial task. Very happy to discuss and figure it out though!

aaronmarkham:
Docker was down yesterday, so that's why the auth fail... I re-triggered it again... and it hung for four hours. Let's see if it can get through the rest now!

PR #13794, opened 2019-01-08
[MXNET-1263] Unit Tests for Java Predictor and Object Detector APIs

piyushghai:
## Description ##
Added unit tests for Java Predictor API and Object Detector API.

### Essentials ###

@lanking520 @andrewfayres @zachgk

### DIFF
--- a/scala-package/core/pom.xml
+++ b/scala-package/core/pom.xml
@@ -13,6 +13,10 @@
   <artifactId>mxnet-core</artifactId>
   <name>MXNet Scala Package - Core</name>
 
+  <properties>
+    <skipTests>false</skipTests>
+  </properties>
+
   <build>
     <plugins>
       <plugin>
@@ -112,16 +116,6 @@
           <argLine>
             -Djava.library.path=${project.parent.basedir}/native/target \
             -Dlog4j.configuration=file://${project.basedir}/src/test/resources/log4j.properties
-          </argLine>
-        </configuration>
-      </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-surefire-plugin</artifactId>
-        <version>2.22.0</version>
-        <configuration>
-          <argLine>
-            -Djava.library.path=${project.parent.basedir}/native/target
           </argLine>
         </configuration>
       </plugin>
--- a/scala-package/core/pom.xml
+++ b/scala-package/core/pom.xml
@@ -14,7 +14,7 @@
   <name>MXNet Scala Package - Core</name>
 
   <properties>
-    <skipTests>false</skipTests>
+    <skipJavaTests>false</skipJavaTests>
   </properties>
 
   <build>
--- a/scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorOutputTest.java
+++ b/scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorOutputTest.java
@@ -17,23 +17,29 @@
 
 package org.apache.mxnet.infer.javaapi;
 
+import org.junit.Assert;
 import org.junit.Test;
 
 public class ObjectDetectorOutputTest {
+
+    private String predictedClassName = "lion";
+
+    private float delta = 0.00001f;
 
     @Test
     public void testConstructor() {
 
         float[] arr = new float[]{0f, 1f, 2f, 3f, 4f};
 
-        ObjectDetectorOutput odOutput = new ObjectDetectorOutput("simba", arr);
+        ObjectDetectorOutput odOutput = new ObjectDetectorOutput(predictedClassName, arr);
 
-        assert (odOutput.getClassName().equals("simba"));
-        assert (odOutput.getProbability() == 0);
-        assert (odOutput.getXMin() == 1);
-        assert (odOutput.getXMax() == 2);
-        assert (odOutput.getYMin() == 3);
-        assert (odOutput.getYMax() == 4);
+        Assert.assertEquals(odOutput.getClassName(), predictedClassName);
+        Assert.assertEquals("Threshold not matching", odOutput.getProbability(), 0f, delta);
+        Assert.assertEquals("Threshold not matching", odOutput.getXMin(), 1f, delta);
+        Assert.assertEquals("Threshold not matching", odOutput.getXMax(), 2f, delta);
+        Assert.assertEquals("Threshold not matching", odOutput.getYMin(), 3f, delta);
+        Assert.assertEquals("Threshold not matching", odOutput.getYMax(), 4f, delta);
+
     }
 
     @Test (expected = ArrayIndexOutOfBoundsException.class)
@@ -41,8 +47,13 @@ public class ObjectDetectorOutputTest {
 
         float[] arr = new float[]{0f, 1f};
 
-        ObjectDetectorOutput odOutput = new ObjectDetectorOutput("simba", arr);
+        ObjectDetectorOutput odOutput = new ObjectDetectorOutput(predictedClassName, arr);
 
-        odOutput.getYMax();
+        Assert.assertEquals(odOutput.getClassName(), predictedClassName);
+        Assert.assertEquals("Threshold not matching", odOutput.getProbability(), 0f, delta);
+        Assert.assertEquals("Threshold not matching", odOutput.getXMin(), 1f, delta);
+
+        // This is where exception will be thrown
+        odOutput.getXMax();
     }
 }
--- a/scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorOutputTest.java
+++ b/scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorOutputTest.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.mxnet.infer.javaapi;
+
+import org.junit.Test;
+
+public class ObjectDetectorOutputTest {
+
+    @Test
+    public void testConstructor() {
+
+        float[] arr = new float[]{0f, 1f, 2f, 3f, 4f};
+
+        ObjectDetectorOutput odOutput = new ObjectDetectorOutput("simba", arr);
+
+        assert (odOutput.getClassName().equals("simba"));
+        assert (odOutput.getProbability() == 0);
+        assert (odOutput.getXMin() == 1);
+        assert (odOutput.getXMax() == 2);
+        assert (odOutput.getYMin() == 3);
+        assert (odOutput.getYMax() == 4);
+    }
+
+    @Test (expected = ArrayIndexOutOfBoundsException.class)
+    public void testIncompleteArgsConstructor() {
+
+        float[] arr = new float[]{0f, 1f};
+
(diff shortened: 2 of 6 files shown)


scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorOutputTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Test
+    public void testConstructor() {
+
+        float[] arr = new float[]{0f, 1f, 2f, 3f, 4f};
+
+        ObjectDetectorOutput odOutput = new ObjectDetectorOutput("simba", arr);
+
+        assert (odOutput.getClassName().equals("simba"));
+        assert (odOutput.getProbability() == 0);
```
vishaalkapoor: assertEquals would give a clearer error message.

vishaalkapoor: Two comments
1) You shouldn't compare floats with equality. The recommended way is to use thresholds. https://howtodoinjava.com/java/basics/correctly-compare-float-double/
2) Use assertEquals with a delta to do this for you. Additionally, if there is inequality, assertequals will provide a more useful error message.

scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorOutputTest.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Test (expected = ArrayIndexOutOfBoundsException.class)
+    public void testIncompleteArgsConstructor() {
+
+        float[] arr = new float[]{0f, 1f};
+
+        ObjectDetectorOutput odOutput = new ObjectDetectorOutput("simba", arr);
+
+        odOutput.getYMax();
```
vishaalkapoor: If you want to be pedantic, you could also check that probability is 0 and getXmin are also set as expected. :)

scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import java.awt.image.BufferedImage;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ObjectDetectorTest {
+
+    List<DataDesc> inputDesc;
```
vishaalkapoor: private or private final for the member variables?

scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+    List<List<ObjectDetectorOutput>> result;
+
+    ObjectDetector objectDetector;
+
+    @Before
+    public void setUp() {
+
+        inputDesc = new ArrayList<>();
+        inputDesc.add(new DataDesc("", new Shape(new int[]{1, 3, 512, 512}), DType.Float32(), Layout.NCHW()));
```
vishaalkapoor: new Shape(int[] ...) is the same "dummy" shape throughout. the details aren't important so you can use a member variable or a dummy creation method to remove the details

inputDesc.add(new DataDec("", testShape(), ...)

scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorTest.java
```diff
@@ -0,0 +1,87 @@
...
+        inputImage = new BufferedImage(512, 512, BufferedImage.TYPE_INT_RGB);
```
vishaalkapoor: Also it would be worthwhile to extract any globals like 512 into private static finals so they can be recognized easily as the same thing, and descriptively.

scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorTest.java
```diff
@@ -0,0 +1,87 @@
...
+        result.get(0).add(new ObjectDetectorOutput("simbaa", new float[]{}));
+    }
+
+    @Test
+    public void testObjectDetectorWithInputImage() {
+
+        Mockito.when(objectDetector.imageObjectDetect(inputImage, 5)).thenReturn(result);
+        List<List<ObjectDetectorOutput>> actualResult = objectDetector.imageObjectDetect(inputImage, 5);
+        Mockito.verify(objectDetector, Mockito.times(1)).imageObjectDetect(inputImage, 5);
+        Assert.assertEquals(result, actualResult);
```
vishaalkapoor: nit - result and actualResult are ambiguous. if you want, I'd recommend using something like expectedResult, actualResult or expectedResult, result.

scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorTest.java
```diff
@@ -0,0 +1,87 @@
...
+    @Before
+    public void setUp() {
+
+        inputDesc = new ArrayList<>();
+        inputDesc.add(new DataDesc("", new Shape(new int[]{1, 3, 512, 512}), DType.Float32(), Layout.NCHW()));
+        inputImage = new BufferedImage(512, 512, BufferedImage.TYPE_INT_RGB);
+        objectDetector = Mockito.mock(ObjectDetector.class);
+        result = new ArrayList<>();
+        result.add(new ArrayList<ObjectDetectorOutput>());
+        result.get(0).add(new ObjectDetectorOutput("simbaa", new float[]{}));
```
vishaalkapoor: why simbaa? is this supposed to be Search IMages By Appearance? (simba)?

scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorTest.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Test
+    public void testObjectDetectorWithInputImage() {
+
+        Mockito.when(objectDetector.imageObjectDetect(inputImage, 5)).thenReturn(result);
+        List<List<ObjectDetectorOutput>> actualResult = objectDetector.imageObjectDetect(inputImage, 5);
```
vishaalkapoor: Is it possible to use real data here instead of Mockito? Mockito was made for situations where real objects would be difficult or impossible. This seems like a situation where we can use a real object and test code paths. I do realize there is an argument that if you use a real object, you're not isolating the component your testing, but on the flip side, using a mock is not as likely to catch real errors. There is an argument in favor of preferring real objects to mocks here if you're curious:
https://testing.googleblog.com/2013/05/testing-on-toilet-dont-overuse-mocks.html

scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorOutputTest.java
piyushghai: Thanks. Done in 41f33716118c7cad36605daad820cc97a2e46201

scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorOutputTest.java
piyushghai: Good point. I am now using assertEquals with delta. 
41f33716118c7cad36605daad820cc97a2e46201

scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorTest.java
piyushghai: Done. 41f33716118c7cad36605daad820cc97a2e46201

scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorTest.java
piyushghai: Done. 41f33716118c7cad36605daad820cc97a2e46201

scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorTest.java
piyushghai: Done. 
41f33716118c7cad36605daad820cc97a2e46201

scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorTest.java
piyushghai: That was just a random string for UTs. 
Modified it :) 
41f33716118c7cad36605daad820cc97a2e46201

scala-package/infer/src/test/java/org/apache/mxnet/infer/javaapi/ObjectDetectorTest.java
piyushghai: > This test is mocking out the called method on the next line, right? Shouldn't it be mocking out the method called at
> 
> [incubator-mxnet/scala-package/infer/src/main/scala/org/apache/mxnet/infer/javaapi/ObjectDetector.scala](https://github.com/apache/incubator-mxnet/blob/96439e6e4889b9c6e2e00ad38b6d379837dc43e1/scala-package/infer/src/main/scala/org/apache/mxnet/infer/javaapi/ObjectDetector.scala#L68)
> Line 68 in [96439e6](/apache/incubator-mxnet/commit/96439e6e4889b9c6e2e00ad38b6d379837dc43e1)
>  val ret = objDetector.imageObjectDetect(inputImage, Some(topK)) 
> But on a bigger note, these tests are on a very thin translation layer so the Mockito tests test very little. Is it possible to use real data here instead of Mockito? It is possible that imageObjectDetect can be changed so as to be impossible to return "result" without violating any types. If so, this test would not be testing a true code path.
> 
> Generally, there's an argument on real objects vs Mocks that I'll quote: https://testing.googleblog.com/2013/05/testing-on-toilet-dont-overuse-mocks.html
> 
> Mocks were made for situations where using real objects would be difficult, undesirable, or impossible, e.g. network calls. I do realize there is an argument that if you use a real object, you're not isolating the component your testing, but on the flip side, using a mock is not as likely to catch real errors; and if the underlying code is changed, the test above should still run unchanged. As for performance, I wouldn't imagine a big impact.
> 
> Let me know what you think.

Yes you have a valid point that it should be mocking out the next line. The unit tests for ```val ret = objDetector.imageObjectDetect(inputImage, Some(topK)) ``` (Scala code) already exist and can be found here : https://github.com/apache/incubator-mxnet/blob/master/scala-package/infer/src/test/scala/org/apache/mxnet/infer/ObjectDetectorSuite.scala 


As for using Mockito v/s real data, I would prefer using a proper integration test Suite, with real world data, and leave the unit tests to check the code paths and branches with mocked data. 

The integ test for SSD in Java can be found here : https://github.com/apache/incubator-mxnet/blob/master/scala-package/mxnet-demo/java-demo/src/main/java/mxnet/ObjectDetection.java

vishaalkapoor: Nice! for the integration test.

Though this integration test has a different purpose of coverage. For example, it won't cover each of the methods you have. And the current Mock tests in isolation leave a gap between the tests.

Ideally there would be more coverage in the unit tests as they are not in the form to catch the most common kind of bug that will be introduced, and that is unexpected input of the same shape - because you're prescribing that input with Mockito.

But I don't think I can block as there are integration tests and reasonable unit tests. 

LGTM. Thanks for making those changes!

vishaalkapoor:
LGTM

scala-package/core/pom.xml
```diff
@@ -13,6 +13,10 @@
   <artifactId>mxnet-core</artifactId>
   <name>MXNet Scala Package - Core</name>
 
+  <properties>
+    <skipTests>false</skipTests>
```
zachgk: I think the skipTests property is already used to refer to skipping the scala integrationtests. Maybe rename this to runsJavaTests or something

piyushghai: Good point. Done in 0be5765d6d65412b85f8479c33f01ad68713fa69

scala-package/pom.xml
```diff
@@ -228,8 +229,12 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
-        <version>2.19</version>
+        <version>2.22.0</version>
```
lanking520: Any reason do the version change?

PR #13795, opened 2019-01-08
add extra header file to export

apeforest:
## Description ##
Export this header file for Horovod to throw and catch exception elegantly in MXNet.
Horovod PR: https://github.com/uber/horovod/pull/542


plugin/opencv/cv_api.cc
```diff
@@ -26,9 +26,9 @@
 #include <dmlc/base.h>
 #include <mxnet/base.h>
 #include <mxnet/ndarray.h>
+#include <mxnet/c_api_common.h>
```
yuxihu: QQ: do we have guideline regarding the order of putting include files? Shall we follow alphabetic order of the filename? If so, please move this include to the right place here and all other places in this PR.

apeforest: I don't find a guideline for order. I was following the order of full path header followed by relative path header

include/mxnet/c_api_common.h
```diff
@@ -22,8 +22,8 @@
  * \file c_api_error.h
  * \brief Error handling for C API.
  */
-#ifndef MXNET_C_API_C_API_COMMON_H_
-#define MXNET_C_API_C_API_COMMON_H_
+#ifndef MXNET_C_API_COMMON_H_
+#define MXNET_C_API_COMMON_H_
 
 #include <dmlc/base.h>
 #include <dmlc/logging.h>
```
eric-haibin-lin: Do you need all these functions in the public header files, including `MXAPIThreadLocalEntry`, `CopyAttr`, etc?

eric-haibin-lin: Do you mind pointing out how horovod will use these APIs?

apeforest:
@eric-haibin-lin @szha The PR that requires this change is https://github.com/ctcyang/horovod/pull/19. The reason we need this header is that we are using `check_call` in Horovod to check the return status [here](https://github.com/ctcyang/horovod/pull/19/files#diff-6626ba53afb59b3f75cb017aa6259fd1R260). We need a way to catch the exception thrown from MXNet C API like [this](https://github.com/ctcyang/horovod/pull/19/files#diff-79b78539e15db1d75d73029490eed708R325). However, the `API_BEGIN()` and `API_END()` are defined in `c_api_common.h` which is the one I need to export.

include/mxnet/c_api_common.h
apeforest: @eric-haibin-lin I agree `MXAPIThreadLocalEntry`, `CopyAtt` are not need in error handling. I don't know the historical reason of adding them to this header file in the first place. My change was following the rationale of making least change to existing structure.

I could extract the error handling functions from `c_api_common.h` into `c_api_error.h` and only export `c_api_error.h` to include. Does that sound like a better approach to you?

eric-haibin-lin:
I agree that MXNet should expose the c apis for error handling. Exposing API_BEGIN/API_END/API_END_HANDLE_ERROR sounds reasonable. We probably should prepend "MX_" in these macro. WDYT @apeforest @szha

szha:
Let's limit the exposure to only what's required.

apeforest:
@eric-haibin-lin @szha Per your suggestion, I have stripped out the error handling APIs into a separate header file and export that one only. The remaining code in MXNet modules remain unchanged. Please review this change again at your earliest convenience. Thanks a lot!

src/c_api/c_api_common.h
```diff
@@ -29,37 +29,35 @@
...
-#define API_END_HANDLE_ERROR(Finalize) } catch(dmlc::Error &_except_) { Finalize; on_exit_api(); return MXAPIHandleException(_except_); } on_exit_api(); return 0; // NOLINT(*)
+#ifndef API_BEGIN
+#define API_BEGIN MX_API_BEGIN
+#endif
 
 /*!
- * \brief Set the last error message needed by C API
- * \param msg The error message to set.
+ * \brief every function starts with API_BEGIN();
+ * and finishes with API_END() or API_END_HANDLE_ERROR
```
yuxihu: can we consolidate the comments into one paragraph? It is kind of duplicate now. Maybe moving L46/L47 and L54-56 to after L39.

apeforest: updated

include/mxnet/c_api_error.h
```diff
@@ -0,0 +1,63 @@
...
+ * \file c_api_error.h
+ * \brief Error handling for C API.
+ */
+#ifndef MXNET_C_API_ERROR_H_
+#define MXNET_C_API_ERROR_H_
+
+#include <dmlc/base.h>
+#include <dmlc/logging.h>
+#include <mxnet/c_api.h>
+#include <mxnet/base.h>
```
szha: are all the headers needed?

apeforest: good catch. removed unnecessary headers

PR #13807, opened 2019-01-09
[MXNET-1180] Java Image API

lanking520:
## Description ##
As title described, add Scala Image API to Java...
@zachgk @piyushghai @andrewfayres 

### Essentials ###

### DIFF
--- a/scala-package/core/src/main/scala/org/apache/mxnet/Image.scala
+++ b/scala-package/core/src/main/scala/org/apache/mxnet/Image.scala
@@ -37,7 +37,7 @@ object Image {
     * @param flag   Convert decoded image to grayscale (0) or color (1).
     * @param to_rgb Whether to convert decoded image
     *               to mxnet's default RGB format (instead of opencv's default BGR).
-    * @return NDArray in HWC format with DType uint8
+    * @return NDArray in HWC format with DType [[DType.UInt8]]
     */
   def imDecode(buf: Array[Byte], flag: Int,
                to_rgb: Boolean,
@@ -56,7 +56,7 @@ object Image {
   /**
     * Same imageDecode with InputStream
     * @param inputStream the inputStream of the image
-    * @return NDArray in HWC format with DType uint8
+    * @return NDArray in HWC format with DType [[DType.UInt8]]
     */
   def imDecode(inputStream: InputStream, flag: Int = 1,
                to_rgb: Boolean = true,
@@ -78,7 +78,7 @@ object Image {
     * @param flag     Convert decoded image to grayscale (0) or color (1).
     * @param to_rgb   Whether to convert decoded image to mxnet's default RGB format
     *                 (instead of opencv's default BGR).
-    * @return org.apache.mxnet.NDArray in HWC format with DType uint8
+    * @return org.apache.mxnet.NDArray in HWC format with DType [[DType.UInt8]]
     */
   def imRead(filename: String, flag: Option[Int] = None,
              to_rgb: Option[Boolean] = None,
--- a/scala-package/core/src/main/scala/org/apache/mxnet/Image.scala
+++ b/scala-package/core/src/main/scala/org/apache/mxnet/Image.scala
@@ -37,7 +37,7 @@ object Image {
     * @param flag   Convert decoded image to grayscale (0) or color (1).
     * @param to_rgb Whether to convert decoded image
     *               to mxnet's default RGB format (instead of opencv's default BGR).
-    * @return NDArray in HWC format
+    * @return NDArray in HWC format with DType uint8
     */
   def imDecode(buf: Array[Byte], flag: Int,
                to_rgb: Boolean,
@@ -56,7 +56,7 @@ object Image {
   /**
     * Same imageDecode with InputStream
     * @param inputStream the inputStream of the image
-    * @return NDArray in HWC format
+    * @return NDArray in HWC format with DType uint8
     */
   def imDecode(inputStream: InputStream, flag: Int = 1,
                to_rgb: Boolean = true,
@@ -78,7 +78,7 @@ object Image {
     * @param flag     Convert decoded image to grayscale (0) or color (1).
     * @param to_rgb   Whether to convert decoded image to mxnet's default RGB format
     *                 (instead of opencv's default BGR).
-    * @return org.apache.mxnet.NDArray in HWC format
+    * @return org.apache.mxnet.NDArray in HWC format with DType uint8
     */
   def imRead(filename: String, flag: Option[Int] = None,
              to_rgb: Option[Boolean] = None,
--- a/scala-package/core/src/main/scala/org/apache/mxnet/javaapi/Image.scala
+++ b/scala-package/core/src/main/scala/org/apache/mxnet/javaapi/Image.scala
@@ -35,13 +35,24 @@ object Image {
     org.apache.mxnet.Image.imDecode(buf, flag, toRGB, None)
   }
 
+  def imDecode(buf: Array[Byte]): NDArray = {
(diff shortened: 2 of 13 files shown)


anirudhacharya:
@mxnet-label-bot add [pr-awaiting-review, Java]

scala-package/core/src/main/scala/org/apache/mxnet/javaapi/Image.scala
```diff
@@ -0,0 +1,95 @@
...
+    * Decode image with OpenCV.
+    * Note: return image in RGB by default, instead of OpenCV's default BGR.
+    * @param buf    Buffer containing binary encoded image
+    * @param flag   Convert decoded image to grayscale (0) or color (1).
+    * @param toRGB Whether to convert decoded image
+    *               to mxnet's default RGB format (instead of opencv's default BGR).
+    * @return NDArray in HWC format
+    */
+  def imDecode(buf: Array[Byte], flag: Int, toRGB: Boolean): NDArray = {
+    org.apache.mxnet.Image.imDecode(buf, flag, toRGB, None)
```
sandeep-krishnamurthy: Will this directly call imdecode operator (C++) or is there a logic at Scala level?

lanking520: This will call Scala Image API which will call C++ level imdecode

scala-package/core/src/main/scala/org/apache/mxnet/Image.scala
```diff
@@ -56,7 +56,7 @@ object Image {
   /**
     * Same imageDecode with InputStream
     * @param inputStream the inputStream of the image
-    * @return NDArray in HWC format
+    * @return NDArray in HWC format with DType uint8
```
piyushghai: Can you refer to the DType Enum : Dtype.UInt8 to be clear here ?

scala-package/examples/pom.xml
```diff
@@ -128,5 +129,11 @@
       <artifactId>slf4j-simple</artifactId>
       <version>1.7.5</version>
     </dependency>
+    <dependency>
```
zachgk: Since we repeat this dependency in infer, move both to the main pom.xml

scala-package/core/src/test/java/org/apache/mxnet/javaapi/ImageTest.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @BeforeClass
+    public static void downloadFile() throws Exception {
+        String tempDirPath = System.getProperty("java.io.tmpdir");
+        imLocation = tempDirPath + "/inputImages/Pug-Cookie.jpg";
+        try {
+            downloadUrl("https://s3.amazonaws.com/model-server/inputs/Pug-Cookie.jpg",
+                    imLocation, 3);
+        } catch (Exception e) {
```
piyushghai: If you are re-throwing the same exception, then you don't need to catch it, as your method signature already says that it'll throw an ex.

scala-package/examples/src/test/java/org/apache/mxnetexamples/javaapi/infer/predictor/PredictorExampleTest.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Test
+    public void testPredictor(){
+        PredictorExample example = new PredictorExample();
+        String[] args = new String[]{
+                "--model-path-prefix", modelPathPrefix,
+                "--input-image", inputImagePath
+        };
+        example.main(args);
```
piyushghai: Is the test to simply run the example code ?

lanking520: Yes, the example test is to verify the example is still working...

scala-package/core/src/main/scala/org/apache/mxnet/javaapi/Image.scala
```diff
@@ -0,0 +1,95 @@
...
+  /**
+    * Resize image with OpenCV.
+    * @param src     source image in NDArray
+    * @param w       Width of resized image.
+    * @param h       Height of resized image.
+    * @param interp  Interpolation method (default=cv2.INTER_LINEAR).
+    * @return org.apache.mxnet.NDArray
+    */
+  def imResize(src: NDArray, w: Int, h: Int, interp: Integer): NDArray = {
+    val interpVal = if (interp == null) None else Some(interp.intValue())
```
zachgk: You can just use Option(interp.intValue())

scala-package/examples/pom.xml
```diff
@@ -15,6 +15,7 @@
 
   <properties>
     <skipTests>true</skipTests>
+    <skipJavaTests>${skipTests}</skipJavaTests>
```
zachgk: What is this for? I don't think it is valid in maven to set one property based on another property and it is just set to false in modules which have java tests to run

scala-package/core/src/main/scala/org/apache/mxnet/javaapi/Image.scala
```diff
@@ -0,0 +1,95 @@
...
+  def imDecode(buf: Array[Byte], flag: Int, toRGB: Boolean): NDArray = {
+    org.apache.mxnet.Image.imDecode(buf, flag, toRGB, None)
+  }
+
+  /**
+    * Same imageDecode with InputStream
+    * @param inputStream the inputStream of the image
+    * @return NDArray in HWC format with DType [[DType.UInt8]]
+    */
+  def imDecode(inputStream: InputStream, flag: Int = 1, toRGB: Boolean = true): NDArray = {
```
piyushghai: Are we consciously hiding the ```out``` NDArray ( the last parameter in the corresponding Scala API) here  ?

scala-package/examples/pom.xml
lanking520: This keeps control of a Junit test is going to run or not, by default it is set to be false. In fact, this is working...

scala-package/core/src/main/scala/org/apache/mxnet/javaapi/Image.scala
lanking520: We can still expose this to users but essentially didn't find any help...

scala-package/core/src/main/scala/org/apache/mxnet/javaapi/Image.scala
lanking520: This will cause a null pointer exception, since users may send null in

lanking520:
@sandeep-krishnamurthy could you please give a review?

scala-package/core/src/main/scala/org/apache/mxnet/javaapi/Image.scala
yzhliu: In my understanding, default args defined in Scala does not help Java users
https://lampwww.epfl.ch/~michelou/scala/using-scala-from-java.html

so as same issues below. correct me if I'm wrong.

lanking520: @yzhliu Yes, this is right. The point to have default args here is just be as a reference for Java users if they happened to see the source code for more information

yzhliu: then I suggest to make it more java, say, have
```scala
def imDecode(inputStream: InputStream, flag: Int, toRGB: Boolean)
def imDecode(inputStream: InputStream, flag: Int)
def imDecode(inputStream: InputStream)
```

PR #13823, opened 2019-01-09
[MXNET-1177]Adding Scala Demo to be run as a part of Nightly CI

piyushghai:
## Description ##
As title.

### Essentials ###

@lanking520 @zachgk @andrewfayres @marcoabreu


scala-package/mxnet-demo/scala-demo/pom.xml
```diff
@@ -9,6 +9,13 @@
     <name>MXNet Scala Demo</name>
     <packaging>pom</packaging>
 
+    <repositories>
```
nswamy: this could be misleading to users, when they modify the demo code in master and try to run, but you are changing it to always pick from snapshot repo. Can you selectively only do this only for CI test?

piyushghai: Good point. Done in : 1cbc59d8f87dff3fa3996916c18b95aeb215200c

scala-package/mxnet-demo/scala-demo/Makefile
```diff
@@ -44,11 +43,14 @@ endif
 scalademo:
 	(mvn package -Dmxnet.profile=$(SCALA_PKG_PROFILE) \
 		-Dmxnet.scalaprofile=$(SCALA_VERSION_PROFILE) \
-		-Dmxnet.version=$(MXNET_VERSION) \
+		-Dscala.version=$(SCALA_VERSION))
+
+scalacidemo:
```
andrewfayres: Two comments:
1.) (Nitpick) I'm not crazy about this name, mainly because scalac is the compiler and this is easily read as "scalac i demo".
2.) This is being done a bit differently than how the java-demo one is done. First, the Java one only has javademo and javaclean. The nightly uses make javademo instead of introducing a javacidemo. Also, the javademo makefile uses mvn install instead of mvn package. Are you aware of anything that keeps the two from being consistent with each other?

scala-package/mxnet-demo/scala-demo/Makefile
```diff
@@ -17,7 +17,6 @@
 
 SCALA_VERSION_PROFILE := 2.11
 SCALA_VERSION := 2.11.8
-MXNET_VERSION := 1.3.0
```
andrewfayres: Anything preventing us from also making this change in the java-demo? It'd be good to keep them as in sync as possible.

scala-package/mxnet-demo/scala-demo/Makefile
piyushghai: I think the differences in java-demo and scala-demo projects are only because they were developed independent of each other and over two different periods of time. 

I'll take your suggestion here and make both of them consistent because essentially they are just showing how to setup a sample project with jars pulled in from Maven. 

Will standardize java-demo, scala-demo in this PR itself.

scala-package/mxnet-demo/scala-demo/Makefile
piyushghai: Yep, referring to my reply above, I'll make both the "demos" in sync with each other and follow the same pattern.

lanking520: +1 for make them Sync. And due to the advantage of static build, user would likely to have less problems with SNAPSHOT or future versions

anirudhacharya:
@mxnet-label-bot add [pr-awaiting-review, Scala]

ci/docker/runtime_functions.sh
```diff
@@ -1227,11 +1227,18 @@ nightly_tutorial_test_ubuntu_python2_gpu() {
...
+    make java_ci_demo
     ./bin/java_sample.sh
     ./bin/run_od.sh
 }
 
+nightly_scala_demo_test_cpu() {
+    set -ex
+    cd /work/mxnet/scala-package/mxnet-demo/scala-demo
+    make scala_ci_demo
+    bash bin/demo.sh
```
zachgk: Sync up script filenames and calling method (bash x.sh vs ./x.sh) as well

scala-package/mxnet-demo/java-demo/Makefile
```diff
@@ -42,10 +41,12 @@ endif
 
 javademo:
 	(mvn install dependency:copy-dependencies package -Dmxnet.profile=$(SCALA_PKG_PROFILE) \
-		-Dmxnet.scalaprofile=$(SCALA_VERSION_PROFILE) \
-		-Dmxnet.version=$(MXNET_VERSION))
+		-Dmxnet.scalaprofile=$(SCALA_VERSION_PROFILE))
+
+java_ci_demo:
+	(mvn -Pci-nightly install dependency:copy-dependencies package -Dmxnet.profile=$(SCALA_PKG_PROFILE) \
```
zachgk: Can this not be just "mvn install"? Install will automatically build package because it is after it in the lifecycle and the copy-dependencies should occur automatically because it was added to the package phase.

piyushghai: As far as I remember when doing this, copy-dependencies does not occur automatically. 
You either add it in the ```<build>``` in the pom file eg : https://github.com/apache/incubator-mxnet/blob/master/scala-package/mxnet-demo/scala-demo/pom.xml#L52, or you can do it via the maven command as shown here.

ci/docker/runtime_functions.sh
piyushghai: Syncing up the script filenames would be a bigger change, as we will have to clean up in places where the demo tutorials is being referred. eg : README, Website etc. I don't want to bring in any other side effects here

I can change the calling method to bash though in the nightly runs.

scala-package/mxnet-demo/java-demo/Makefile
zachgk: Adding it in the build in the pom file should be the proper way of doing it. Otherwise, it makes it difficult if you do other maven commands such as test, install, or deploy because all the commands would need to work around copy-dependencies. In fact, we should probably try to remove all Makefiles throughout scala and use the maven commands alone.

PR #13834, opened 2019-01-10
Reduce maven verbosity which is filling up the logs of the builds

larroy:
## Description ##

Maven is too verbose and is producing tons of output while downloading dependencies etc. This is a known issue. This PR reduces maven verbosity in the invocations from the Makefile so the logs are not overflowed.

https://blogs.itemis.com/en/in-a-nutshell-removing-artifact-messages-from-maven-log-output

### Essentials ###

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -614,13 +614,13 @@ scalaclean:
 	(cd $(ROOTDIR)/scala-package && mvn clean)
 
 scalapkg:
-	(cd $(ROOTDIR)/scala-package && mvn install -DskipTests)
+	(cd $(ROOTDIR)/scala-package && mvn install -q -DskipTests)
 
 scalainstall:
-	(cd $(ROOTDIR)/scala-package && mvn install)
+	(cd $(ROOTDIR)/scala-package && mvn install -q)
 
 scalaunittest:
-	(cd $(ROOTDIR)/scala-package && mvn install)
+	(cd $(ROOTDIR)/scala-package && mvn install -q)
 
 scalaintegrationtest:
 	(cd $(ROOTDIR)/scala-package && mvn integration-test -DskipTests=false)


Makefile
```diff
@@ -614,13 +614,13 @@ scalaclean:
 	(cd $(ROOTDIR)/scala-package && mvn clean)
 
 scalapkg:
-	(cd $(ROOTDIR)/scala-package && mvn install -DskipTests)
+	(cd $(ROOTDIR)/scala-package && mvn install -q -DskipTests)
```
frankfliu: This change will make it hard to troubleshoot PR failure.
What's the limitation of the log out?
Can we redirect maven output the a log file, and make the log file as artifact to download?

larroy: We don't store artifacts right now. I didn't find a way to suppress the log output less than with -q

Won't a failure be printed in any case? On a failing PR the -q can be removed for diagnosis anyway.

frankfliu: I assume, developer has tested locally with maven already. If there is a failure in CI, in most cases, error message is not enough to troubleshoot a PR failure.

Switching -q flag to debug on CI isn't a right approach, and it's very time consuming effort.

anirudhacharya:

Makefile
larroy: Can you put examples of what you describe or provide an alternative? Why removing -q on a PR wouldn't be enough to troubleshoot any failures?  That PR would repro the error with full verbosity

sandeep-krishnamurthy:
@frankfliu - Can you please take a look at this PR and response by @larroy

Makefile
lanking520: Still, adding -q would made developer hard to debug the code if there is any failures in middle. This PR is also stale since all CIs has changed from using Makefile to use `mvn` command directly. As Jenkins didn't have a restriction on the lines of logs to print, we should have as much as possible

PR #13855, opened 2019-01-11
Fix symbolic example for rand_zipfian function

ChaiBapchya:
## Description ##
`mx.sym.contrib.rand_zipfian` function example fixed
It contained NDArray specific example for a Symbolic API

### DIFF
--- a/python/mxnet/symbol/contrib.py
+++ b/python/mxnet/symbol/contrib.py
@@ -73,17 +73,8 @@ def rand_zipfian(true_classes, num_sampled, range_max):
 
     Examples
     --------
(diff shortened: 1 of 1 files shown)


anirudhacharya:

KellenSunderland:
LGTM but I can't tell what's fixed.  Were the values reported in the docs wrong?

ChaiBapchya:
The example before had NDArray. I replaced it with a corresponding Symbolic example

stu1130:
@KellenSunderland I think what he fixed is just doc. The symbol API should use smybol for demonstration rather than NDArray. And we won't get intermediate result with Symbol.

PR #13865, opened 2019-01-12
Modifying clojure CNN text classification example

kedarbellare:
## Description ##

- This PR modifies the clojure CNN text classification example. It tries to come close to the python-based example and also fixes some bugs.
- It can be used with pretrained embeddings such as `:glove` or `:word2vec`. It can also be used without any pretrained embedding in which case an embedding is learned from training data.
- There was an issue with using pretrained embeddings where a new random vector (uniformly sampled) was generated for the same word in the vocabulary if it appeared multiple times across the corpus. This PR fixes that issue by generating a random vector just once per out-of-vocabulary (OOV) word. This also helps avoid the memory `GC limit exceeded` issue.
- Finally, it can be used with `word2vec` embeddings which required some tweaks in the way the model was being loaded. To optimize memory usage, either `max-vectors` can be loaded or a `vocab` can be passed in so that only a subset of embeddings are loaded.

### Essentials ###

### Changes ###
- [X] CNN text classification (including docs)

### Reviewers ###

@gigasquid

### DIFF
--- a/contrib/clojure-package/examples/cnn-text-classification/README.md
+++ b/contrib/clojure-package/examples/cnn-text-classification/README.md
@@ -3,19 +3,19 @@
 An example of text classification using CNN
 
 To use you must download the MR polarity dataset and put it in the path specified in the mr-dataset-path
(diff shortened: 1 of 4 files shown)


gigasquid:
Thanks for your work in improving this. I'm looking forward to taking it for a spin 😸

Issue #12649, opened 2018-09-24
Sequence operators should allow different type for sequence length


szha:
Currently, Sequence operators (i.e. Sequence{Last,Mask,Reverse}) require sequence lengths and input data to have the same type, which doesn't seem necessary.

stephenrawls:
FYI I created a PR for this issue: https://github.com/apache/incubator-mxnet/pull/13880

szha:
It's now possible to use different type on index thanks to @stephenrawls's contribution.

PR #13880, opened 2019-01-15
MXNET-1295 Adding integer index support to Sequence* family of operators.

stephenrawls:
## Description ##
Adding ability to use int32 arrays, or any castable-to-int type, as
the sequence_length array to SequenceMask, SequenceLast, and
SequenceReverse. Previously these operaters all requred sequence_length
to be the same data type as the input array.

See MxNet Jira ticket here:
  https://issues.apache.org/jira/browse/MXNET-1295

See also GitHub issues here:
   https://github.com/apache/incubator-mxnet/issues/12649
   https://github.com/dmlc/gluon-nlp/issues/346

## Test Coverage ##
I have tested this works by building locally and running all the Sequence* operators in python with a float32 input array and an int32 sequence_length array. I confirmed things work correct.

However when I looked at the unit tests it was not immediately clear how to add an appropriate test, because the current tests all use the `check_symbolic_forward()` function, which takes a single dtype for all arguments and doesn't allow clients to specify different dtypes for different input arguments.

### Essentials ###
- [??] All changes have test coverage:

### DIFF
--- a/python/mxnet/test_utils.py
+++ b/python/mxnet/test_utils.py
@@ -654,8 +654,8 @@ def _parse_location(sym, location, ctx, dtype=default_dtype()):
                              % (str(set(sym.list_arguments())), str(set(location.keys()))))
     else:
         location = {k: v for k, v in zip(sym.list_arguments(), location)}
-    location = {k: mx.nd.array(v, ctx=ctx, dtype=v.dtype if dtype == "asnumpy" else dtype) if isinstance(v, np.ndarray) \
-               else v for k, v in location.items()}
+    location = {k: mx.nd.array(v, ctx=ctx, dtype=v.dtype if dtype == "asnumpy" else dtype) \
+               if isinstance(v, np.ndarray) else v for k, v in location.items()}
     return location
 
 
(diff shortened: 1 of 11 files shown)


szha:
Thanks for contributing this, @stephenrawls. Could you add some test cases for using different input types? (tests/python/unittest/test_operator.py)

stephenrawls:
@szha Okay I added unit tests. As I alluded to in my original PR, it wasn't exactly clear the best way to do this, because the testing code seems to have a lot of assumptions that all argument types will be the same type, and that the only possible argument types are floats.

For example, there are `assert dtype in (np.float16, np.float32, np.float64)` sprinkled all throughout the unit test framework, and I don't know what other assumptions I'll break if I change any of that.

Also, the unit tests are relying on test_utils.py to convert all input arrays to a single input type, e.g. here is an example:

```
    location = {k: mx.nd.array(v, ctx=ctx, dtype=dtype) if isinstance(v, np.ndarray) \
               else v for k, v in location.items()}
```

I ended up modifying the logic so that I could control the dtype of each array independently by just creating the original numpy array to be the data type that I want, and telling the code to keep the numpy dtype when converting to mxnet.ndarray:

```
    location = {k: mx.nd.array(v, ctx=ctx, dtype=v.dtype if dtype == "asnumpy" else dtype) if isinstance(v, np.ndarray) \
               else v for k, v in location.items()}
```

If you want to use the new behavior, you have to opt-in and pass "asnumpy" as the dtype.

Take a look, and if you prefer a different approach just let me know.

Best,
Stephen

szha:
@stephenrawls thanks! the approach sounds reasonable.

szha:
@stephenrawls thanks for the contribution!

PR #13921, opened 2019-01-17
Version switching user experience improvements

aaronmarkham:
## Description ##
This PR fixes some usability problems with the version selector. You tend to lose your place when you use the search feature or when browsing the APIs looking for stuff. If you switch versions, you don't always end up where you were or get redirected to the home page. 

**EDIT**: It also fixes a search bug for developers that try to use the search feature in their dev versions of the website.

## Preview for Search
When you search for something on the website in an older version, don't find it, then switch to a newer version, you are redirected to the root of the website. 

For example:
1. Search for "taco": http://mxnet.incubator.apache.org/versions/1.0.0/search.html?q=taco&check_keywords=yes&area=default
No 🌮 ! What if master has 🌮 ? 
2. Use the versions dropdown and change to master.
You lost your search! 😭 and still no 🌮 .

Now try to search for 🌮 on the preview:
1. http://34.201.8.176/versions/1.3.1/search.html?q=taco
No 🌮 . 
2. Change to master. 
Still no 🌮 but 🌤 you didn't get bounced to the home page.

## Preview for Anchors
It seems to work now when the page is loaded anew. However, I noticed that using in-page anchors the page didn't provide a new `window.location.hash` to the function in `docversion.js` that updates the version dropdown's links, so the when switching versions you don't load the last anchor you were on. 

1. Go look at ndarray docs: http://34.201.8.176/versions/1.3.1/api/python/ndarray/ndarray.html
2. Scroll down and click the link for `ndarray.shape`.
3. You jump down to the docs for shape. But you decide you don't like it, and you want to know if master has more features. Your current location is now http://34.201.8.176/versions/1.3.1/api/python/ndarray/ndarray.html#mxnet.ndarray.NDArray.shape
4. Use the version dropdown to change to Master.
5. Witness 👀  that you are now on: http://34.201.8.176/versions/master/api/python/ndarray/ndarray.html#mxnet.ndarray.NDArray.shape

If you try this flow in production you see that it is kind of broken.

## Comments
The redirects update and overview of intentions and test links are on the wiki: https://cwiki.apache.org/confluence/display/MXNET/Redirects+on+the+Website

The search feature on the website is buggy. Today it shows results in an old version that shouldn't be there. For example "allreduce" shouldn't appear <v1.2.1, but it does today. I'm not trying to fix Sphinx's search in this PR.

### DIFF
--- a/docs/_static/js/docversion.js
+++ b/docs/_static/js/docversion.js
@@ -18,16 +18,23 @@
  */
 
 /* Set the version of the website */
-function setVersion(){
-        let doc = window.location.pathname.match(/^\/(api\/.*)$/) || window.location.pathname.match(/^\/versions\/[^*]+\/(api\/.*)$/);
+function setVersion(anchor){
+        if (arguments.length==0) {
+            anchor = window.location.hash
+        };
+        console.log('anchor: ', anchor, arguments.length, window.location.hash, window.location.search);
+        //let doc = window.location.pathname.match(/^\/(api\/.*)$/) || window.location.pathname.match(/^\/versions\/[^*]+\/(api\/.*)$/);
+        let doc = window.location.pathname.match(/^\/versions\/[^\/]+\/([^*]+.*)$/);
         if (doc) {
             if (document.getElementById('dropdown-menu-position-anchor-version')) {
                     versionNav = $('#dropdown-menu-position-anchor-version a.main-nav-link');
                     $(versionNav).each( function( index, el ) {
                             currLink = $( el ).attr('href');
-                            version = currLink.match(/\/versions\/([0-9.master]+)\//);
+                            version = currLink.match(/\/versions\/([^\/]+)\//);
+                            console.log(version);
                             if (version) {
-                                    versionedDoc = '/versions/' + version[1] + '/' + doc[1] + (window.location.hash || '');
+                                    versionedDoc = '/versions/' + version[1] + '/' + doc[1] + (anchor || '') + (window.location.search || '');
+                                    console.log(versionedDoc);
                                     $( el ).attr('href', versionedDoc);
                             }
                     });
@@ -40,5 +47,5 @@ $(document).ready(function () {
 });
 
 $('a.reference.internal').click(function(){
-    setVersion();
+    setVersion($(this).attr("href"));
 });
--- a/docs/_static/js/docversion.js
+++ b/docs/_static/js/docversion.js
@@ -22,8 +22,6 @@ function setVersion(anchor){
         if (arguments.length==0) {
             anchor = window.location.hash
         };
-        console.log('anchor: ', anchor, arguments.length, window.location.hash, window.location.search);
-        //let doc = window.location.pathname.match(/^\/(api\/.*)$/) || window.location.pathname.match(/^\/versions\/[^*]+\/(api\/.*)$/);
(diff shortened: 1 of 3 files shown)


aaronmarkham:
This is a rehash of https://github.com/apache/incubator-mxnet/pull/12689 - but this time the redirects to master or other versions seem to be working well. @safrooze can you take a look and let me know what you think? @ddavydenko this uses your regex's. Thank you! Please take a look at the implementation.

docs/_static/js/docversion.js
```diff
@@ -18,16 +18,23 @@
  */
 
 /* Set the version of the website */
-function setVersion(){
-        let doc = window.location.pathname.match(/^\/(api\/.*)$/) || window.location.pathname.match(/^\/versions\/[^*]+\/(api\/.*)$/);
+function setVersion(anchor){
+        if (arguments.length==0) {
+            anchor = window.location.hash
+        };
+        console.log('anchor: ', anchor, arguments.length, window.location.hash, window.location.search);
```
ddavydenko: I suggest to remove this debugging statement from production code.

docs/_static/js/docversion.js
```diff
@@ -18,16 +18,23 @@
...
+        //let doc = window.location.pathname.match(/^\/(api\/.*)$/) || window.location.pathname.match(/^\/versions\/[^*]+\/(api\/.*)$/);
+        let doc = window.location.pathname.match(/^\/versions\/[^\/]+\/([^*]+.*)$/);
         if (doc) {
             if (document.getElementById('dropdown-menu-position-anchor-version')) {
                     versionNav = $('#dropdown-menu-position-anchor-version a.main-nav-link');
                     $(versionNav).each( function( index, el ) {
                             currLink = $( el ).attr('href');
-                            version = currLink.match(/\/versions\/([0-9.master]+)\//);
+                            version = currLink.match(/\/versions\/([^\/]+)\//);
+                            console.log(version);
```
ddavydenko: I suggest to remove this debugging statement from production code.

docs/_static/js/docversion.js
```diff
@@ -18,16 +18,23 @@
...
                             if (version) {
-                                    versionedDoc = '/versions/' + version[1] + '/' + doc[1] + (window.location.hash || '');
+                                    versionedDoc = '/versions/' + version[1] + '/' + doc[1] + (anchor || '') + (window.location.search || '');
+                                    console.log(versionedDoc);
```
ddavydenko: I suggest to remove this debugging statement from production code.

docs/_static/js/docversion.js
aaronmarkham: Thanks, I removed all of the debug statements.

ChaiBapchya:
nitpick: Can we get rid of the "searching" after it has searched and displays the output
screen shot 2019-01-17 at 4 48 18 pm

aaronmarkham:
@ChaiBapchya - IDK, maybe I can fix that. 
I just fixed a bug in the search where if you tried to click on a search result from the results shown in the search bubble, it would never connect because dev servers don't have https (usually). I made it look at the current protocol instead and use that. Now using and debugging search should be easier on dev machines. (I am trying to figure out why search is messed up right now, but if this passes CI, we should probably just merge it and get whatever bug fixes I might come up with later.)

ChaiBapchya:
Yes, this sounds good.

IvyBazan:
LGTM!

PR #13922, opened 2019-01-17
Support populating errors back to MXNet engine in callback

yuxihu:
This PR adds an optional dmlc::Error* argument in MXNet engine callback functions. The callers can leverage this argument to populate errors back to MXNet engine through callback such that the errors can be handled properly by MXNet engine. One such [use case](https://github.com/ctcyang/horovod/pull/24) is to populate the errors detected in Horovod back to MXNet engine for Hovorod and MXNet integration. This change does not affect existing use cases.

### DIFF
--- a/include/mxnet/engine.h
+++ b/include/mxnet/engine.h
@@ -74,15 +74,15 @@ class CallbackOnComplete {
  public:
   // use implicit copy and assign
   /*! \brief involve the callback */
-  inline void operator()(const char* error_msg = nullptr) const {
-    (*callback_)(engine_, param_, error_msg);
+  inline void operator()(const dmlc::Error* error = nullptr) const {
+    (*callback_)(engine_, param_, error);
   }
 
  private:
   /*! \brief engine can see content of callback */
   friend class ::mxnet::Engine;
   /*! \brief the real callback */
-  void (*callback_)(Engine *, void *, const char *);
+  void (*callback_)(Engine *, void *, const dmlc::Error *);
   /*! \brief the engine class passed to callback */
   Engine* engine_;
   /*! \brief the parameter set on callback */
@@ -275,7 +275,7 @@ class MXNET_API Engine {
    * \param param the paramter passed to callback.
    */
   inline CallbackOnComplete CreateCallback(
-      void (*callback)(Engine *, void *, const char *), void *param) {
+      void (*callback)(Engine *, void *, const dmlc::Error *), void *param) {
     CallbackOnComplete ret;
     ret.callback_ = callback;
     ret.engine_ = this;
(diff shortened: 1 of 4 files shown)


include/mxnet/engine.h
```diff
@@ -74,15 +74,15 @@ class CallbackOnComplete {
  public:
   // use implicit copy and assign
   /*! \brief involve the callback */
-  inline void operator()() const {
-    (*callback_)(engine_, param_);
+  inline void operator()(const char* error_msg = nullptr) const {
```
apeforest: Instead of passing in a const string of error message, would it make sense to pass in a `error` struct which contains error message and error code in the fields? This leaves more room for proper error handling based on error code.

yuxihu: The error code is not universally defined across different libraries. In the Horovod case, the error types are mostly Horovod specific. We convert all those to dmlc::Error which MXNet can catch.

apeforest: Can we define an error data structure and export it? So all libraries that consume libmxnet will use this error data structure to pass data back.

yuxihu: This sounds like a good idea to make it extendable in the future. We can use the error message for now.

apeforest: sgtm

stu1130:
Nice work @yuxihu 

include/mxnet/engine.h
yuxihu: Use dmlc::Error instead of char*. We can change dmlc::Error if more information need to be passed.

yuxihu:
@eric-haibin-lin please take a look.

src/engine/threaded_engine.cc
```diff
@@ -478,10 +478,14 @@ inline void ThreadedEngine::ThrowException(ThreadedVar* threaded_var) {
...
 
-void ThreadedEngine::OnCompleteStatic(
-    Engine *engine, void *opr_block_) {
+void ThreadedEngine::OnCompleteStatic(Engine *engine, void *opr_block_,
+                                      const dmlc::Error* error) {
   OprBlock *opr_block = static_cast<OprBlock*>(opr_block_);
   ThreadedOpr *threaded_opr = opr_block->opr;
+  if (error != nullptr) {
+    auto ex_p = std::make_exception_ptr(*error);
+    threaded_opr->opr_exception = std::make_shared<std::exception_ptr>(ex_p);
```
larroy: Isn't exception_ptr already reference counted? Do we need to wrap it again in make_shared?

yuxihu: Good catch. I will change related sites by removing the redundant shared_ptr in a separate PR.

PR #13930, opened 2019-01-18
Add batchify transformer to help end to end models

sandeep-krishnamurthy:
## Description ##
Add new transformer in Gluon "Batchify". This transformer joins/batchify a list of tensors of shape (C x H x W) into a single tensor of shape (N x C x H x W).

This will be helpful when users have created a fused end to end model (transformers + network), saved the graph and this end to end model is deployed in production for "single request inference" or "mini-batch inference". The raw input can be of different shapes, till the input reaches resize transformation, if running a mini batch inference, then input will be a list of NDArrays. After Resize, these individual NDArrays will batched with this transformer.

### Essentials ###

### Changes ###
- Add new transformer Batchify to take list of NDArrays and batchify them in to 1 NDArray

@stu1130

### DIFF
--- a/python/mxnet/gluon/data/vision/transforms.py
+++ b/python/mxnet/gluon/data/vision/transforms.py
@@ -487,10 +487,13 @@ class RandomLighting(HybridBlock):
         return F.image.random_lighting(x, self._alpha)
 
 
-class ListToBatch(HybridBlock):
+class Batchify(HybridBlock):
     """Joins a list of tensors of shape (C x H x W) into a single 
     tensor of shape (N x C x H x W) where N is the number of input
-    tensors.
+    tensors. 
+
+    If the input is a single tensor of shape (C x H x W) it is bathchified
+    to a tensor of shape (1 x C x H x W).
 
     This transformer is useful when transformation pipeline is fused into
     neural network graph resulting in single model/graph. When running 
@@ -501,7 +504,7 @@ class ListToBatch(HybridBlock):
 
     For example, a typical graph can look like below:
 
-    Imdecode -> Resize -> ListToBatch -> ToTensor -> Normalize -> Network
+    Imdecode -> Resize -> Batchify -> ToTensor -> Normalize -> Network
     
     Parameters
     ----------
@@ -517,7 +520,7 @@ class ListToBatch(HybridBlock):
     
     Examples
     --------
-    >>> transformer = transforms.ListToBatch()
+    >>> transformer = transforms.Batchify()
     >>> image1 = mx.nd.random.uniform(0, 1, shape=(3, 28, 28))
     >>> image2 = mx.nd.random.uniform(0, 1, shape=(3, 28, 28))
     >>> image3 = mx.nd.random.uniform(0, 1, shape=(3, 28, 28))
--- a/python/mxnet/gluon/data/vision/transforms.py
+++ b/python/mxnet/gluon/data/vision/transforms.py
@@ -488,15 +488,15 @@ class RandomLighting(HybridBlock):
 
 
 class Batchify(HybridBlock):
-    """Joins a list of tensors of shape (C x H x W) into a single 
+    """Joins a list of tensors of shape (C x H x W) into a single
     tensor of shape (N x C x H x W) where N is the number of input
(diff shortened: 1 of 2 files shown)


sandeep-krishnamurthy:
> This seems to be the concern of batchify function in dataloader.

Added the description explaining the usage. Basically this will be helpful during mini batch inference on an end to end model i.e., a fused model (transformer + network).

Input can be different shapes. Resize transformer is also part of the network.
Example model looks like - Resize -> Batchify -> ToTensor -> Normalize -> Network
Example mini batch inference looks like
```
im1 = imread(...)
im2 = imread(...)
Outputs = executor.forward(data0=im1, data1=im2)
```

stu1130:
Great work! @sandeep-krishnamurthy

python/mxnet/gluon/data/vision/transforms.py
```diff
@@ -485,3 +485,52 @@ def __init__(self, alpha):
...
+    This transformer is useful when transformation pipeline is fused into
+    neural network graph resulting in single model/graph. When running
+    mini batch inference on such graph, raw input tensors, before transformations,
+    can be of different shapes, hence, cannot be batched into single tensor. Hence,
+    input to the graph will be list of Tensors that can be batched after Resize
+    transformation.
+
+    For example, a typical graph can look like below:
+
+    Imdecode -> Resize -> Batchify -> ToTensor -> Normalize -> Network
```
zhreshold: This argument does not sound right, images after resize has layout HWC, which violate the input expectation of CHW with this Batchify block

szha: More importantly, dataset.transform takes care of only sample-level transformation. It needs to be clear which class is responsible for stacking the samples. This is currently done in dataloader.

sandeep-krishnamurthy: @zhreshold - Thanks for pointing it out. Functionally, input can be either HWC or CHW it gets converted to NHWC or NCHW respectively. Will update the code comment.

sandeep-krishnamurthy: @szha - The way I am thinking is - this transformer is used when user is exporting a model for inference and he wants to fuse transformations as part of the network. This transformer is not applicable for training / validation which all works as existing today.
So to summarize, the idea here is:
1. Training does not changing anything from what it is today - dataset and dataloader
2. When user wants to export a model transformations (Resize, Batchify, ToTensor, Normalize) is concatenated to network block and exported.

sandeep-krishnamurthy: I agree, but, thinking what is more intuitive and ease of use.

szha: Introducing this would only make it confusing to people who are already using the dataloader in python. If the goal is to make it available to other users, one simple approach would be to wrap this stack operator in the logic of exporting data preprocessing pipeline. @zhreshold

tests/python/unittest/test_gluon_data_vision.py
```diff
@@ -83,6 +83,25 @@ def test_transformer():
     transform(mx.nd.ones((245, 480, 3), dtype='uint8')).wait_to_read()
 
 
+@with_seed()
+def test_batchify():
+    transformer = transforms.Batchify()
+    im1 = nd.random.uniform(0, 1, (3, 28, 28))
+    im2 = nd.random.uniform(0, 1, (3, 28, 28))
+    im3 = nd.random.uniform(0, 1, (3, 28, 28))
```
vandanavk: Could you add a test with inputs of different sizes too? That may serve as an example for users as well

ankkhedia:
@sandeep-krishnamurthy Could you please look into the review comments above?

tests/python/unittest/test_gluon_data_vision.py
anirudhacharya: can the batch sizes be randomized?

sandeep-krishnamurthy:
Closing this PR because having is transformer is more confusing than useful for users. As suggested above in reviews, such transformer should be internally managed by respective export/import APIs.

PR #13937, opened 2019-01-20
Julia: add windows-cpu build

iblis17:
- [x] Julia v0.7
- [x] Julia v1.0


iblis17:
well, why errors in powershell script are not considered as build failure? 
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fwindows-cpu/detail/PR-13937/2/pipeline/82#step-151-log-22

marcoabreu:
Yeah Powershell is not very smart unfortunately :( Have a look at how we do it here: https://github.com/apache/incubator-mxnet/blob/e8a2b8b9fdafaccbf65397cec142fffcae2289b7/ci/windows/test_py3_cpu.ps1#L27

marcoabreu:
Btw here's the RFC to add native error handling to Powershell. Feel free to give it a +1 https://github.com/PowerShell/PowerShell-RFC/pull/88

ci/windows/test_jl07_cpu.ps1
```diff
@@ -0,0 +1,45 @@
...
+
+# rm -rf
+Remove-item -Recurse -Force -ErrorAction Ignore C:\julia07
+
+# mkdir
+New-item -ItemType Directory C:\julia07
+
+# Download most recent Julia Windows binary
+[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]::Tls12
+(New-Object System.Net.WebClient).DownloadFile($env:JULIA_URL, "C:\julia07\julia-binary.exe")
```
marcoabreu: Please note that C:\julia07 is a shared folder that will be accessed by multiple jobs in parallel. This can cause concurrency problems.

Instead, I'd recommend to download into the current workspace since that one guarantees exclusive access. In parallel, please create an issue with a feature request so we install Julia in our Windows slaves.

iblis17: Ah, okay. I filed an issue: https://github.com/apache/incubator-mxnet/issues/13941

iblis17:
This PR is ready for review.

marcoabreu:
Could you elaborate that's the deal with

```
powershell.exe : [ Info: Path `C:\jenkins_slave\workspace\ut-julia07-cpu\julia-depot\dev\MXNet` exists and looks like the correct package, using existing path instead of cloning

At C:\jenkins_slave\workspace\ut-julia07-cpu@tmp\durable-9d0a0ccf\powershellWrapper.ps1:3 char:1

+ & powershell -NoProfile -NonInteractive -ExecutionPolicy Bypass -Comm ...

+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

    + CategoryInfo          : NotSpecified: ([ Info: Path `C...tead of cloning    :String) [], RemoteException

    + FullyQualifiedErrorId : NativeCommandError
```

It's flagged as an error, doesn't seem like one and isn't handled like one.

iblis17:
Still not sure which line causes that message.
Will get a windows machine to figure out it later.

ci/windows/test_jl10_cpu.ps1
```diff
@@ -0,0 +1,54 @@
...
+
+# Run installer silently, output to C:\julia10\julia
+Start-Process -Wait "julia-binary.exe" -ArgumentList "/S /D=$JULIA_DIR"
+if (! $?) { Throw ("Error on installing Julia") }
+
+& $JULIA -e "using InteractiveUtils; versioninfo()"
+
+# workaround a bug of Julia 0.7
+# use PackageSpec(name = "MXNet", path = "...") if we drop the 0.7 support
+Copy-Item -Path .\julia -Recurse -Destination "$env:JULIA_DEPOT_PATH\dev\MXNet"
```
marcoabreu: It seems like it's this line @iblis17

iblis17: screenshot

I found the julia function `Pkg.develop` cause that error message, but it's intended.

I'm going to find another way to install MXNet package locally.

iblis17: Seems powershell will capture `stderr` and raise `NativeCommandError`, since Julia generate the logger output to `stderr`. (FYI: https://superuser.com/questions/1253297)

```
PS C:\Users\iblis\tmp\incubator-mxnet\ci\windows> echo '@info "test"' | & $JULIA
julia.exe : [ Info: test
位於 線路:1 字元:23
+ echo '@info "test"' | & $JULIA
+                       ~~~~~~~~
    + CategoryInfo          : NotSpecified: ([ Info: test:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
```

`@info "test"` is a valid Julia logger syntax.
```julia
julia> @info "test"
[ Info: test 

```

The solution is redirecting them to `stdout`.

ci/windows/test_jl07_cpu.ps1
```diff
@@ -0,0 +1,52 @@
...
+
+$src='
+    using Pkg
+    Pkg.activate(".\\julia")
+    Pkg.build()
+    Pkg.test()
+'
+
+$src > .\ci-build.jl
+& $JULIA .\ci-build.jl 2>&1 | %{ "$_" }
```
marcoabreu: Could you add a short comment why we do that redirect (maybe just a summary of what you explained previously)>

iblis17: ah, sure

lebeg:
Looks good, only 4 and 5 minutes for the testing parts

PR #13940, opened 2019-01-20
Add back R tests and fix typo around R and perl tests

marcoabreu:
This PR adds back the R tests that have been accidentally removed in my previous refactor. Sorry about that!!!

Also, I had a typo that mixed up Perl and R tests. Fixed that as well...

@jeremiedb @anirudhacharya @hetong007

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -596,7 +596,7 @@ rpkg:
 	cp -rf 3rdparty/dmlc-core/include/* R-package/inst/include/
 	cp -rf 3rdparty/tvm/nnvm/include/* R-package/inst/include
 	Rscript -e "if(!require(devtools)){install.packages('devtools', repo = 'https://cloud.r-project.org/')}"
-	Rscript -e "if(!require(devtools)||packageVersion('roxygen2') < '6.1.1'){install.packages('roxygen2', repo = 'https://cloud.r-project.org/')}"
+	Rscript -e "if(!require(roxygen2)||packageVersion('roxygen2') < '6.1.1'){install.packages('roxygen2', repo = 'https://cloud.r-project.org/')}"
 	Rscript -e "library(devtools); library(methods); options(repos=c(CRAN='https://cloud.r-project.org/')); install_deps(pkg='R-package', dependencies = TRUE)"
 	cp R-package/dummy.NAMESPACE R-package/NAMESPACE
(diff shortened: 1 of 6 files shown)


larroy:
LGTM

marcoabreu:
Seems like something isn't right:

```
** installing vignettes

** testing if installed package can be loaded

* DONE (devtools)



The downloaded source packages are in

	'/tmp/RtmppoYF6h/downloaded_packages'

Warning message:

In library(package, lib.loc = lib.loc, character.only = TRUE, logical.return = TRUE,  :

  there is no package called 'devtools'

Rscript -e "if(!require(devtools)||packageVersion('roxygen2') < '6.1.1'){install.packages('roxygen2', repo = 'https://cloud.r-project.org/')}"

Loading required package: devtools

Error in packageVersion("roxygen2") : package 'roxygen2' not found

Execution halted

Makefile:589: recipe for target 'rpkg' failed

make: *** [rpkg] Error 1

```

Does somebody know from the top of their head what has to be done? Otherwise, I'm happy to have a look myself.

hetong007:
I'm not sure what exactly caused the error. 

The error means R cannot find the dependencies. If they were just installed (seems so from the log), then I suggest to look for a mismatch in the environment, for instance: do the command which installed the dependencies, and the following `Rscript -e xxx` refer to the same R binary?

Makefile
```diff
@@ -596,7 +596,7 @@ rpkg:
 	cp -rf 3rdparty/dmlc-core/include/* R-package/inst/include/
 	cp -rf 3rdparty/tvm/nnvm/include/* R-package/inst/include
 	Rscript -e "if(!require(devtools)){install.packages('devtools', repo = 'https://cloud.r-project.org/')}"
-	Rscript -e "if(!require(devtools)||packageVersion('roxygen2') < '6.1.1'){install.packages('roxygen2', repo = 'https://cloud.r-project.org/')}"
+	Rscript -e "if(!require(roxygen2)||packageVersion('roxygen2') < '6.1.1'){install.packages('roxygen2', repo = 'https://cloud.r-project.org/')}"
```
marcoabreu: @jeremiedb could you double-check this please? You recently added this line and I assumed that it was a copy&paste leftover.

The result was that the script tried to query for the version of roxygen2 even if it was not installed, thus leading to a "null pointer exception".

Issue #13959, opened 2019-01-22
website redirect loop on error pages


aaronmarkham:
## Description
The 404 error page is triggered when going to a missing page, but there's a redirect loop.

https://mxnet.incubator.apache.org/error/404.html

PR #13963, opened 2019-01-23
Fix website error pages

aaronmarkham:
## Description ##
Fixes https://github.com/apache/incubator-mxnet/issues/13959
This PR creates new error pages, uses some better regex's in the `.htaccess` file, and it provides better artifact handling for the build process.

### Changes ###
- [x] Changed the `ErrorDocument` for 404 errors to a local file instead of using a hard-coded link to the error page in production. This is partly why we ended up with a redirect loop. It also had the drawback of giving a 302 redirect header instead of 404. Bots 🤖 don't like that, plus it was probably fooling the broken link checker. 
- [x] Generalized the redirects for pages that should be current (thanks @ddavydenko!).
- [x] Deleted unnecessary artifacts (index.html was from an old workaround and is not used)
- [x] Created two new html error pages that don't use relative links to assets as this fails when being loaded dynamically for a 404 message.
- [x] Updated the Sphinx script to copy these error pages when doing a local build
- [x] Moved the `mxnet.css` artifact copy to the Sphinx script function to better consolidate artifact and build updates; also this help make sure that any updates to css in your dev version will get applied during a local build AND a full website publish

## Preview

Aside from the [redirect test links](https://cwiki.apache.org/confluence/display/MXNET/Redirects+on+the+Website), you can try:
* The new [404 error page](http://34.201.8.176/versions/local_error/error/404.html)
* The new [API error page](http://34.201.8.176/versions/local_error/error/api.html)
🌮  Break things with a random URL like: http://34.201.8.176/versions/local_error/tacos
❓ Look for an API doc that doesn't exist, like Clojure for v0.11.0: http://34.201.8.176/versions/0.11.0/api/clojure/index.html

Note that I redirect 🌮 requests or anything really in an API subfolder's that is designated in the .htaccess file to the API error, rather than the 404 error. This way we can say, well, maybe it does exist, so try a different version. 

## Comments
Unfortunately, for some reason `ErrorDocument` converts the environment variable for `default_version` just fine, but then just outputs the URL as text to the browser rather that forcing a redirect. So it has to stay hard-coded to `master`. (That's fine anyway as I can't imagine why an error page could be the latest at all times.)


docs/build_version_doc/artifacts/404.html
```diff
@@ -0,0 +1,50 @@
...
+
+</style>
+  </head>
+  <body>
+    <div>
+      <a href="/" id="logo" ><img width="200"  src="https://raw.githubusercontent.com/dmlc/web-data/master/mxnet/image/mxnet_logo.png"></a>
+    </div>
+    <div>
+    <span id="page-does-not-exist"></span><h1>Page Does Not Exist</h1>
+    <p>If you’re here that means you requested a page that doesn’t exist. Sorry about that! Maybe try the search box to find what you’re looking for, or navigate to the <a href="https://mxnet.incubator.apache.org/index.html">Home Page</a>. Also, make sure you’re looking in the correct version, as some features may only be available in <a href="https://github.com/apache/incubator-mxnet/releases">newer versions</a> or the <a href="https://mxnet.incubator.apache.org/versions/master">master branch</a>.</p>
```
marcoabreu: Relative link maybe?

aaronmarkham: Yes, I fixed that. Thanks.

PR #13549, opened 2018-12-05
Runtime feature detection

larroy:
## Description ##

It would be good to have runtime feature detection, for example to know if we were compiled with OPENCV, LAPACK, some tests fail when MXNet is not compiled with some feature or another.

This is a proposal for a runtime feature detection mechanism that would expose how the library was compiled and what features are available.

It also grooms some of the feature flags which were not uniformly set in base.h

See related discussion in dev:

https://lists.apache.org/thread.html/19fe1da67551f9a199ec85dbb9c894e65c205d5fdd6764509e6e41ed@%3Cdev.mxnet.apache.org%3E

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #13964, opened 2019-01-23
Addresses comments in runtime feature discovery API

larroy:
## Description ##
Returns a struct of LibFeature and reduces to a single API call querying of compile-time features.

Followup on API comments on https://github.com/apache/incubator-mxnet/pull/13549

@szha 

### Essentials ###


szha:
@larroy thanks! will take a look shortly.

include/mxnet/mxruntime.h
```diff
@@ -0,0 +1,194 @@
...
+
+#include <string>
+#include "dmlc/base.h"
+#include "mshadow/base.h"
+
+/*!
+ *\brief whether to use opencv support
+ */
+#ifndef MXNET_USE_OPENCV
+#define MXNET_USE_OPENCV 1
```
marcoabreu: Shouldn't this be 0?

include/mxnet/mxruntime.h
```diff
@@ -0,0 +1,194 @@
...
+#endif
+
+
+
+namespace mxnet {
+namespace features {
+// Check compile flags such as CMakeLists.txt
+
+/// Compile time features
+// ATTENTION: When changing this enum, match the strings in mxruntime.cc!!!
```
marcoabreu: Could we have a sanity check script that ensures that they are in sync? It could then throw an error if they are out of sync.

marcoabreu: Or is that what tests/cpp/misc/runtime_test.cc is supposed to do?

tests/cpp/misc/runtime_test.cc
```diff
@@ -0,0 +1,32 @@
...
+
+#include <gtest/gtest.h>
+#include <mxnet/mxruntime.h>
+
+using namespace mxnet;
+using namespace std;
+using namespace mxnet::features;
+
+/*
+ * Test that different Context have different hash values
```
marcoabreu: I'm not sure, but does that comment match the implementation? Maybe you could provide a second paragraph where you explain what's the purpose of this test and what people should look out for if it fails

tests/python/unittest/test_runtime.py
```diff
@@ -0,0 +1,40 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+import mxnet as mx
+import sys
+from mxnet.runtime import *
+from mxnet.base import MXNetError
+from nose.tools import *
+
+def test_runtime_features():
```
marcoabreu: Great idea!

tests/python/unittest/test_runtime.py
```diff
@@ -0,0 +1,40 @@
...
+    for f in Feature:
```
marcoabreu: Maybe assert that ```Feature``` is not empty?

tests/python/unittest/test_runtime.py
```diff
@@ -0,0 +1,40 @@
...
+        res = has_feature_index(f.value)
+        ok_(type(res) is bool)
+    for f in features_enabled():
+        ok_(type(f) is Feature)
+    ok_(type(features_available()) is list)
```
marcoabreu: Maybe assert that ```Feature``` and ```features_enabled``` are not empty?

include/mxnet/mxruntime.h
```diff
@@ -0,0 +1,194 @@
+        /*
```
szha: space. also why not name it with mx prefix? I don't think this convention is used anywhere else.

include/mxnet/c_api.h
```diff
@@ -208,6 +208,24 @@ MXNET_DLL const char *MXGetLastError();
...
 // Part 0: Global State setups
 //-------------------------------------
+
+/*!
+ * \brief Check if a feature is enabled in the runtime
+ * \param feature to check mxruntime.h
+ * \param out set to true if the feature is enabled, false otherwise
+ * \return 0 when success, -1 when failure happens.
+ */
+MXNET_DLL int MXRuntimeHasFeature(const mx_uint feature, bool *out);
```
szha: why not have one interface that returns the complete enum and the values for each of them? doesn't seem necessary to have a capi for testing individual feature given that it can be cached in the frontend.

python/mxnet/runtime.py
```diff
@@ -0,0 +1,82 @@
...
+    feature_list_sz = ctypes.c_size_t()
+    check_call(_LIB.MXRuntimeFeatureList(ctypes.byref(feature_list_sz), ctypes.byref(feature_list)))
+    feature_names = []
+    for i in range(feature_list_sz.value):
+        feature_names.append(py_str(feature_list[i]))
+    return feature_names
+
+Feature = enum.Enum('Feature', {name: index for index, name in enumerate(_feature_names_available())})
+
+def features_available():
```
szha: a single api that lists all features and their values seems easier to maintain and it wouldn't compromise the usability.

include/mxnet/mxruntime.h
larroy: Yes the test checks that they have the same number of elements, I think is enough to prevent them going out of sync. I think is enough.

tests/cpp/misc/runtime_test.cc
larroy: paste error

include/mxnet/c_api.h
larroy: Agree, I can return the enum values, but they are also implicit on the index of the strings. One of the reasons why I like having a function to test if the feature is present is that is much simpler and faster than fiddling with strings and it doesn't allocate any memory. If we remove this function then we should have two functions, features available and features enabled. I think they are functionally equivalent, there has to be some part of the code that test the feature flags anyway, exposed or not.

python/mxnet/runtime.py
larroy: See comment above, there's features available and enabled. These are two sets where enabled is a subset of available.

larroy:
Seems CI status is fully missing.

include/mxnet/mxruntime.h
larroy: there is another files in the project named runtime.h so there's a clash with the includes. Naming it runtime.h was prevented by the IDE. Do you have a better suggestion? I don't have a strong opinion on this.

include/mxnet/c_api.h
marcoabreu: As far as I understand, this function is not intended for usage within high performance loops, so I would rather prefer usability instead of optimizing for a few bytes of allocation. I'm also in favour of a dictionary with the function name as key and the status as boolean value. For HPC cases, you could still cache the results if really necessary

larroy:
Didn't work out to put it in libinfo, causes installation dependencies.
I would suggest to leave it as it is.

https://github.com/larroy/mxnet/tree/feature_discovery_libinfo

include/mxnet/c_api.h
szha: This function doesn’t have to be fast.

szha:
What installation dependencies are you referring to? Libinfo is responsible for the information on libmxnet hence the suggestion.

include/mxnet/c_api.h
larroy: I don't like two functions returning strings, there's some type checking, verification happening if you have to test the enum explicitly that you don't get with two sets of strings.

larroy:
@szha I did the changes that you requested, but doesn't work due to libinfo.py being used before loading the library, is a rabbit hole, see this branch: https://github.com/larroy/mxnet/tree/feature_discovery_libinfo

I discarded that idea. Can we provide a path to completion?

python/mxnet/runtime.py
szha: What's wrong with having a single method to return the complete list of features in string, the mask number, and a binary state indicator for each feature on whether a feature is enabled or not?

szha:
@larroy I will need more information to offer advice on how to make it work. Alternatively, I can fork the branch and try to make it work. Let me know what you prefer.

python/mxnet/runtime.py
larroy: ah ok I understood you wanted too different API calls. That sounds reasonable to me.

larroy: To return a list of features, mask number etc, that would be array of struct. We would need either another call to first get the number of features so we can reserve an array of struct of the propper size in python using ctypes. Another option is that this array of struct is initialized statically somewhere like LibraryInitializer or as a singleton, which is always a bit of an anti pattern for testing / mocking, or asuming there's only one instance of somthing. Third option is on the heap but that opens the door for leaks. What would in your opinion be the best choice?

szha: I think having a call that gets the number first is fine. This should make it an idempotent function.

szha:
since the information is fixed after compilation, libinfo sounds more accurate. there's no need to maintain the consistency in naming between module name of python frontend and the file name in the backend.

On the API design, yes, I feel strongly about the change, hence the change request. The reason is that the library information doesn't change after it's compiled, and since it's loaded only once, there's no futher need to query the backend, especially given that ctypes isn't exactly fast. Thus, having an API that queries all feature status at the beginning, and make that information available should be the goal. To achieve that goal, a single method with complete information should be the way to go.

larroy:
Thanks for the merge @szha

PR #13973, opened 2019-01-23
Update README.md to include new bindings

ThomasDelteil:
## Description ##
Simply updating the readme to add the new bindings supports 

@aaronmarkham


aaronmarkham:
Excellent! But what about Perl?

Issue #13442, opened 2018-11-28
[Clojure] Port ResourceScope to the Clojure Package


gigasquid:
It would be nice to port `ResourceScope` to the Clojure package to have a standard way to manage memory among the JVM languages 

https://github.com/apache/incubator-mxnet/blob/master/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala

Completing this would involving the correct interop to call the functions and use it in an idiomatic Clojure way. It would also involve creating unit tests along the lines of https://github.com/apache/incubator-mxnet/blob/master/scala-package/core/src/test/scala/org/apache/mxnet/ResourceScopeSuite.scala

@nswamy Is available as a resource from the Scala package side to provide an overview / insight on the design and limitations

gigasquid:
working on this one

gigasquid:
merged!

PR #13993, opened 2019-01-25
[Clojure] Add resource scope to clojure package

gigasquid:
## Description ##
Ports ResourceScope from the Scala package and introduces `using` macro to automatically close any resource in the enclosing forms.

https://github.com/apache/incubator-mxnet/issues/13442

and see https://cwiki.apache.org/confluence/display/MXNET/JVM+Memory+Management for more context

### Essentials ###

## Changes

Also added the resource scope to the imclassification example.

New runs to the example do not show any warnings of memory leaks of this nature:
`WARN  org.apache.mxnet.WarnIfNotDisposed: LEAK: [one-time warning] ...`


contrib/clojure-package/test/org/apache/clojure_mxnet/resource_scope_test.clj
```diff
@@ -0,0 +1,65 @@
...
+            [org.apache.clojure-mxnet.symbol :as sym]
+            [org.apache.clojure-mxnet.resource-scope :as resource-scope]
+            [clojure.test :refer :all]))
+
+(deftest test-resource-scope-with-ndarray
+  (let [x (ndarray/ones [2 2])
+        return-val (resource-scope/using
+                    (def temp-x (ndarray/ones [3 1]))
+                    (def temp-y (ndarray/ones [3 1]))
+                    (let [z {:just-a-test (def temp-z (ndarray/ones [3 3]))}
```
benkamphaus: Is there a more natural use case that the nested `def` forms that still illustrates the behavior and is amenable to testing?

gigasquid: I agree the def are weird - do you think an atom with a map of ndarrays under test that could be `swap` ed in would be better?

benkamphaus: Hmm. I'd say yes _cautiously_. Tradeoff IMO: it would look a bit more idiomatic, but also put more indirection in the tests.

gigasquid: Thanks for the feedback and review @benkamphaus - I updated the test. They are an improvement. I feel like it could be further improved as well, but I can't think of an immediate way :)

contrib/clojure-package/test/org/apache/clojure_mxnet/resource_scope_test.clj
```diff
@@ -0,0 +1,65 @@
...
+                    (let [temp-x (ndarray/ones [3 1])]
+                      (swap! native-resources assoc :temp-x temp-x)
+                     (resource-scope/using
+                      (let [temp-y (ndarray/ones [3 1])]
+                        (swap! native-resources assoc :temp-y temp-y)))))]
+    (is (true? (.isDisposed (:temp-y @native-resources))))
+    (is (true? (.isDisposed (:temp-x @native-resources))))
+    (is (false? (.isDisposed x)))))
+
+(deftest test-resource-scope-with-sym
```
kedarbellare: can you also add a unit test for nested with sym?

contrib/clojure-package/examples/imclassification/src/imclassification/train_mnist.clj
```diff
@@ -97,24 +76,39 @@
...
-      ))))
+        )))))
 
 (defn -main [& args]
   (let [[dev dev-num] args
         devs (if (= dev ":gpu")
                (mapv #(context/gpu %) (range (Integer/parseInt (or dev-num "1"))))
                (mapv #(context/cpu %) (range (Integer/parseInt (or dev-num "1")))))]
-    (start devs)))
+    (resource-scope/using (start devs))))
```
kedarbellare: having the whole thing encapsulated in one `resource-scope` seems weird. is it possible to wrap `using` around the individual pieces (e.g. `mnist-iter`, `fit`, `fit-params`) so that most of the code remains the same?

contrib/clojure-package/src/org/apache/clojure_mxnet/resource_scope.clj
```diff
@@ -0,0 +1,25 @@
...
+;; See the License for the specific language governing permissions and
+;; limitations under the License.
+;;
+
+(ns org.apache.clojure-mxnet.resource-scope
+  (:require [org.apache.clojure-mxnet.util :as util])
+  (:import (org.apache.mxnet ResourceScope)))
+
+(defmacro using
+  "Uses a Resource Scope for all forms. This is a way to manage all Native Resources like NDArray and Symbol - it will deallocate all Native Resources by calling close on them automatically. It will not call close on Native Resources returned from the form"
```
kedarbellare: nit: split into multiple lines?

kedarbellare: also is there a way of adding examples inside the docstring so that they show up on the clojure api docs?

contrib/clojure-package/test/org/apache/clojure_mxnet/resource_scope_test.clj
```diff
@@ -0,0 +1,65 @@
...
+  (:require [org.apache.clojure-mxnet.ndarray :as ndarray]
+            [org.apache.clojure-mxnet.symbol :as sym]
+            [org.apache.clojure-mxnet.resource-scope :as resource-scope]
+            [clojure.test :refer :all]))
+
+
+(deftest test-resource-scope-with-ndarray
+  (let [native-resources (atom {})
+        x (ndarray/ones [2 2])
+        return-val (resource-scope/using
```
kedarbellare: i have a wild idea. let me know if i'm way off here but `using` gives the impression that its form should be `(using x (do y))`. instead i think `using` the way you've defined it, is very much like a `let` IMHO. can the `return-val` be:
```clojure
(resource-scope/let 
  [temp-x (ndarray/ones [3 1])
   temp-y (ndarray/ones [3 1])]
  (swap! native-resources assoc :temp-x temp-x)
  (swap! native-resources assoc :temp-y temp-y)
  (ndarray/+ temp-x 1))
```

contrib/clojure-package/examples/imclassification/src/imclassification/train_mnist.clj
gigasquid: that's a good idea - thanks

contrib/clojure-package/test/org/apache/clojure_mxnet/resource_scope_test.clj
gigasquid: I like where you are going - I think adding an alias for `let` is a good idea. It's also like a `do` though because any native resources created in the forms will be closed. What do you think about adding aliases for `resource-scope/let` and `resource-scope/do` while keeping `resource-scope/using` to keep with the Scala api?

kedarbellare: that sounds great!

contrib/clojure-package/test/org/apache/clojure_mxnet/resource_scope_test.clj
kedarbellare: one more unit test ideas (similar in spirit to scala api): many `ndarray/ones` within a `vector` and only the first one is returned after a transformation (e.g. `(+ (first v) 1)`)

contrib/clojure-package/src/org/apache/clojure_mxnet/resource_scope.clj
```diff
@@ -0,0 +1,25 @@
...
+  [& forms]
+  `(ResourceScope/using (new ResourceScope) (util/forms->scala-fn ~@forms)))
```
kedarbellare: this may be a stupid question but is there any value in reusing a previously defined `ResourceScope` object? why does the scala API allow a `scope` to be passed in?

gigasquid: That's a good question. From looking at the Scala api I would say that it would enable finer grained control - you could manually add or remove native resources from a scope and then manually call close on it. But I don't think that we need to expose that on the Clojure side, unless you have a different view on it.

kedarbellare: i don't see a big need for it -- at least for now -- but i was wondering whether the motivation was something like `https://www.tensorflow.org/api_docs/python/tf/variable_scope` (e.g. reusing symbols).

gigasquid: I don't know - maybe @nswamy knows more about the context behind the design? ^

gigasquid:
I had some stuff come up - but plan to finish this up sometime this week :)

contrib/clojure-package/test/org/apache/clojure_mxnet/resource_scope_test.clj
gigasquid: well `do` and `let` are special forms - so we need another name... I'm thing `resource-scope/with-do` and `resource-scope/with-let`

benkamphaus: I def. think `with-` is the right choice to convey resource cleanup.

contrib/clojure-package/test/org/apache/clojure_mxnet/resource_scope_test.clj
gigasquid: interesting the result here - I put test cases in, but if a whole vector has been assigned to a let binding and you return the first, the whole vector is *not* disposed. You have to call copy on the first or not assign the whole vector to a let to have it disposed

gigasquid: There are some edge cases still I think there - but on the whole, it's an improvement for mem management of native resources.

gigasquid:
I think all the feedback is addressed. Feel free to take another look. If there are no more changes needed, I will merge 😸

contrib/clojure-package/test/org/apache/clojure_mxnet/resource_scope_test.clj
```diff
@@ -0,0 +1,151 @@
...
+;; then they are disposed as usual
+(deftest test-list-creation-with-return-a-copy-of-first
+  (let [native-resources (atom [])
+        return-val (resource-scope/using
+                    (first (repeatedly 5 #(do
+                                            (let [x (ndarray/ones [3 1])]
+                                              (swap! native-resources conj x)
+                                              x)))))]
+    (is (false? (ndarray/is-disposed return-val)))
+    (is (every? true? (mapv ndarray/is-disposed (:temp-ndarrays @native-resources))))))
```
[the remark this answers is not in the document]
gigasquid: If fixed some other problems with laziness in the tests with `repeatedly`. Evidently the weird behavior in the dispose of the first of the collection was just my fault on the tests. Now everything works as expected 🎆

contrib/clojure-package/examples/imclassification/src/imclassification/train_mnist.clj
```diff
@@ -96,18 +75,33 @@
     (do
       (println "Starting Training of MNIST ....")
       (println "Running with context devices of" devs)
-      (let [_mod (m/module (get-symbol) {:contexts devs})]
-        (m/fit _mod {:train-data train-data
-                    :eval-data test-data
+      (resource-scope/with-let [_mod (m/module (get-symbol) {:contexts devs})]
+        (-> _mod
+            (m/fit {:train-data (mx-io/mnist-iter {:image (str data-dir "train-images-idx3-ubyte")
```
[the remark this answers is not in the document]
gigasquid: It will work if they are `defs` outside, but you will still get the undisposed ndarray warning. I refactored to move it out to functions which does work and looks better too :)

PR #14028, opened 2019-01-30
Cmake blas

edisongustavo:
## Description ##

Eases the pain of linking with OpenBLAS using cmake.

This PR adds support to use the provided `OpenBLASConfig.cmake`.

### Justification for the change

The standard way of linking with dependencies in CMake is via the `find_package()` mechanism ([docs](https://cmake.org/cmake/help/latest/command/find_package.html)). This mechanism provides 2 ways of finding the dependencies:
- Config mode (uses a file named `OpenBLASConfig.cmake` (or `openblas-config.cmake`), which are provided by the dependency itself)
- Module mode (uses a file named `FindOpenBLAS.cmake`, which is provided by either cmake or the project compiling the dependency, mxnet in this case)

The preferred way to find a dependency is if they provide their "Config" file, since the authors of the dependency are the ones who know best about the structure of their code and how they should be linked.

When I tried to compile MXNet on Windows it didn't work well. My setup was:

- Visual studio 2017.
- OpenBLAS package from the SourceForge binaries package, as specified in the [documentation to build MXNet on Windows from source](http://mxnet.incubator.apache.org/versions/master/install/windows_setup.html#build-from-source).

It failed because this binary package is compiled with a Visual Studio version prior to 2015. So I would get the error `unresolved external symbol __imp____iob_func`. The reasons are explained here: https://stackoverflow.com/questions/30412951/unresolved-external-symbol-imp-fprintf-and-imp-iob-func-sdl2

Then I tried to use the [conda-forge package](https://anaconda.org/conda-forge/openblas). This almost worked.

On further inspection of the [FindOpenBLAS.cmake](https://github.com/apache/incubator-mxnet/blob/master/cmake/Modules/FindOpenBLAS.cmake) and the conda-forge package, I noticed that MXNet didn't use the provided "cmake Config mode" files (`OpenBLASConfig.cmake`). It tried to find the library all by itself, which is an "anti-pattern" when using cmake. So I thought that this could be an opportunity to be able to contribute to the project by improving this module.

### Essentials ###

### Changes ###

- [ ] Add , tests, (and when applicable, API doc)

I have tested this by compiling it on Windows and Linux.

To provide OpenBLAS I have tried:

On Linux:
- compiled OpenBLAS from source with cmake and installed it
- compiled OpenBLAS from source with make and installed it

On Windows:
- used the compiled OpenBLAS package from conda-forge: https://anaconda.org/conda-forge/openblas

## Comments ##


vandanavk:
@mxnet-label-bot add [pr-awaiting-review, Build]

@marcoabreu for review

edisongustavo:
This is still a WIP, I will fix it as soon as possible.

cmake/Modules/FindOpenBLAS.cmake
```diff
@@ -14,78 +14,168 @@
...
+# 2 - Use pkg-config
+# 3 - Find the files manually
+#
+# At each step that was just described, in order to guarantee that the library was found correctly,
+# this script tries to compile this simple program:
+#
+# #include <cblas.h>
+# void main() { cblas_sasum(0, (float*)0, 0); }
+#
+# If this simple program can't be compiled with the found include directories provided by the step,
```
ChaiBapchya: How about this - 
"If this simple program can't be compiled with the OpenBLAS library discovered in the previous step,"
does this sound simpler?

cmake/Modules/FindOpenBLAS.cmake
```diff
@@ -14,78 +14,168 @@
...
+# then it assumes the step failed and moves on to the next step.
+#
+# To control the behavior of each step, these settings can be set for each step:
+# - For step 1:
```
ChaiBapchya: missing anything?

cmake/Modules/FindOpenBLAS.cmake
```diff
@@ -14,78 +14,168 @@
...
+# - For step 3 you can set the environment variable OpenBLAS_HOME with the installed directory of
```
ChaiBapchya: for step 2?

cmake/Modules/FindOpenBLAS.cmake
edisongustavo: This looks better, I will change the text. Thanks for the suggestion.

cmake/Modules/FindOpenBLAS.cmake
edisongustavo: Yes, that slipped. Oooops

cmake/Modules/FindOpenBLAS.cmake
edisongustavo: I will revisit these.

ci/docker/install/android_arm64_openblas.sh
```diff
@@ -25,8 +25,41 @@ pushd .
 git clone https://github.com/xianyi/OpenBLAS.git
 cd OpenBLAS
 make -j$(nproc) TARGET=ARMV8 ARM_SOFTFP_ABI=1 HOSTCC=gcc NOFORTRAN=1 libs
-# Can't be run (utility not compiled for the target platform)
-#make install
-cp *.h /usr/include
-cp libopenblas.a /usr/local/lib
+
+# Ideally, a simple `make install` would do the job. However, OpenBLAS fails when running `make
```
lebeg: Actually, you can set the [`PREFIX`](https://github.com/apache/incubator-mxnet/blob/master/ci/docker/install/arm_openblas.sh#L26) instead.

ci/docker/Dockerfile.build.android_armv8
```diff
@@ -73,7 +73,7 @@ ENV CXX=${CROSS_ROOT}/bin/${CROSS_TRIPLE}-clang++
 # Build ARM dependencies.
 COPY install/android_arm64_openblas.sh /work/
 RUN /work/android_arm64_openblas.sh
-ENV CPLUS_INCLUDE_PATH /work/deps/OpenBLAS
+ENV OpenBLAS_HOME=/work/deps/OpenBLAS
```
perdasilva: is this change also relevant for the other arm builds?

ci/docker/install/android_arm64_openblas.sh
edisongustavo: Yes, but the problem is that the utility `getarch` which is invoked by the `Makefile` fails to execute. That would be another rabbit hole that I didn't want to get into.

ci/docker/Dockerfile.build.android_armv8
edisongustavo: > Did you figure out if this affects anything downstream?

What do you mean by that? If any other image is using this docker image? I didn't check that, but it shouldn't affect any downstreams, unless they're depending on this, which would be depending on an undocumented hack.

> and is it worth moving this change to a different PR for PR atomicity, etc.

The problem is that this change is **required** for this PR to be merged. Without this change it won't work.

perdasilva: Yup - I saw that. All good! My bad, I didn't read to the end before commenting =P

edisongustavo:
Good points @aaronmarkham! I will take a look at the docs to see what can be updated.

anirudhacharya:
@edisongustavo ping to update the PR

edisongustavo:
@anirudhacharya I will get back to this asap, probably tomorrow.

karan6181:
@edisongustavo Thank you for your contributions! Can you please provide some updates in this PR? Thanks!

abhinavs95:
@mxnet-label-bot update [Build, pr-awaiting-merge]

abhinavs95:
@edisongustavo Could you please provide an update on the PR? Thanks

edisongustavo:
@abhinavs95 Yes, I will do that tomorrow.

piyushghai:
@edisongustavo Gentle ping...

Roshrini:
@edisongustavo Can you address Aaron's comments so that we can move forward with this PR? Thanks

roywei:

edisongustavo:
Hello all, sorry for taking so long to answer. Things have been very hectic at lately.

Anyway, I've looked at the docs and the place that I see that could use an update is this part here (https://mxnet.incubator.apache.org/versions/master/install/windows_setup.html#build-from-source):

> 1. Download the [OpenBlas](https://sourceforge.net/projects/openblas/files/v0.2.19/OpenBLAS-v0.2.19-Win64-int32.zip/download) package. Later versions of OpenBLAS are available, but you would need to build from source. v0.2.19 is the most recent version that ships with binaries. Contributions of more recent binaries would be appreciated.

The problem is that the version of the OpenBLAS package compiled there is using Visual Studio 2008, which is not compatible with Visual Studio 2017.

I have looked at all releases at the OpenBLAS sourceforge page and didn't find any releases with binaries on it.

So the problem is: **"find compiled binaries of OpenBLAS (for the Visual Studio 2017) available for download on the web"**.

A solution I could think of is to use the compiled package from [conda-forge](https://conda-forge.org/): https://anaconda.org/conda-forge/openblas

In that page, the closest version `0.2.19` that have Windows releases is `0.2.20`: `win-64/openblas-0.2.20-vc14_8.tar.bz2`. The newer versions all have Windows releases, so we could just use point to use version `0.3.6` instead.

The rewritten doc would look like this:

```markdown
6. Download the [OpenBlas](https://anaconda.org/conda-forge/openblas/0.2.20/download/win-64/openblas-0.2.20-vc14_8.tar.bz2) package.
7. Untar the file. There will be a directory inside named `Library`, rename it to ```OpenBLAS``` and put it under `C:\utils`. You can place the unzipped files and folders in another directory if you wish.
```

edisongustavo:
People, I accidentally deleted my `incubator-mxnet` fork while I was cleaning up my GitHub account...

I've recreated the PR here: https://github.com/apache/incubator-mxnet/pull/14871

aaronmarkham:
Closing in favor of #14871.

PR #13906, opened 2019-01-16
[MXNET-703] Update onnx-tensorrt for int8/fp16 support

KellenSunderland:
## Description ##
Update onnx-tensorrt with new support for different data types, including (importantly for v100) fp16.  This will not fully enable int8/fp16 inference yet, but lays the groundwork for further updates.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

PR #14040, opened 2019-01-31
Reformat of TensorRT to use subgraph API

Caenorst:
## Description ##
This PR modify TensorRT usage by relying on the Subgraph API, the backend is named 'TensorRT'
### Essentials ###
- Need an update in documentation for general use cases

### Changes ###
- TensorRT now just rely on the Subgraph API, and a contrib function for loading weights
- Add FP16 support, computation in FP16 forced by default
- Graph partitioning have changed so if the same input of a subgraph is used multiple times in the subgraph then it stay as a single node in the subgraph

## Comments ##
- Now that it's using the Subgraph API, you can use TensorRT with both module API and symbolic API
- The way weights are loaded in the TensorRT node is a little bit dirty, we will do a follow up PR (including one on NNVM) to natively have a node attributes own subgraph parameters including a setter on the c api


src/operator/subgraph/tensorrt/nnvm_to_onnx.cc
```diff
@@ -411,6 +383,49 @@ void ConvertElementwiseAdd(NodeProto* node_proto, const NodeAttrs& /*attrs*/,
...
+    case mshadow::kInt8:
+      return TensorProto_DataType_INT8;
+    case mshadow::kInt64:
+      return TensorProto_DataType_INT64;
+    default:
+      return TensorProto_DataType_UNDEFINED;
+  }
+}
+
+inline std::string StringDType(int dtype) {
```
KellenSunderland: Is this still needed, don't see any references to it.

src/operator/subgraph/tensorrt/onnx_to_tensorrt.cc
```diff
@@ -127,16 +128,15 @@ nvinfer1::ICudaEngine* onnxToTrtCtx(
...
       throw dmlc::Error("Cannot parse ONNX into TensorRT Engine");
   }
-
-  bool fp16 = trt_builder->platformHasFastFp16();
-
+  if (dmlc::GetEnv("MXNET_TENSORRT_USE_FP16", true)) {
+    if (trt_builder->platformHasFastFp16()) {
+      trt_builder->setFp16Mode(true);
+    } else {
+      LOG(INFO) << "WARNING: TensorRT can't use fp16 on this plateform";
```
KellenSunderland: plateform -> platform

KellenSunderland: Also we're logging INFO level logs but have WARNING in the message.  I'd remove the warning from the message and set log level to warning.  (this is a common issue in our codebase).

src/operator/subgraph/tensorrt/tensorrt-inl.h
```diff
@@ -0,0 +1,217 @@
...
+
+  bool SelectInput(const nnvm::Node &n, const nnvm::Node &new_node) override {
+    if (new_node.is_variable()) {
+      if (withWeightsOps.count(n.op()->name)) {
+        return n.inputs[0].node->attrs.name != new_node.attrs.name;
+      } else {
+        return false;
+      }
+    }
+    if (isTRTCompatible(new_node))
```
KellenSunderland: Can be simplified to 
```C++
return isTRTCompatible(new_node);
```

src/operator/subgraph/tensorrt/tensorrt-inl.h
```diff
@@ -0,0 +1,217 @@
...
+                              std::vector<nnvm::NodeEntry*>* output_entries) const override {
+    std::vector<nnvm::NodeEntry>& outputs = subgraph_node->attrs.subgraphs[0]->outputs;
+    TRTParam& _params = nnvm::get<TRTParam>(subgraph_node->attrs.parsed);
+    for (int i = 0; i < outputs.size(); i++) {
+      auto& o = outputs[i];
+      for (auto& e : *output_entries) {
+        if (o.index == e->index && o.node.get() == e->node.get()) {
+          e->index = i;
+          e->node = subgraph_node;
+          // TODO(cfujitsang): For futur support this would fail
```
KellenSunderland: futur -> future.

src/operator/subgraph/tensorrt/tensorrt-inl.h
```diff
@@ -0,0 +1,217 @@
...
+};
+
+class TensorrtProperty : public SubgraphProperty {
+ public:
+  static SubgraphPropertyPtr Create() {
+    return std::make_shared<TensorrtProperty>();
+  }
+
+  nnvm::NodePtr CreateSubgraphNode(const nnvm::Symbol &sym,
+                                   const int subgraph_id = 0) const override {
```
KellenSunderland: If possible can we avoid default values in overriding functions.

KellenSunderland:
ONNX-Tensorrt might fail to build as it adds a new default target that requires a header not in a standard search path.  There's a few ways to work around the problem, we could just build the library and not that tool in CI, or we could include the header folder location in the search path like so: https://github.com/apache/incubator-mxnet/pull/13906/files#diff-56133c25b5a238b76f54c0928f05a8e6

It should allow the TensorRT build to pass CI if we make that change in this PR.

Caenorst:
Seems that my modification in CutGraphInputs is breaking some of the default attributes inference functions, I think my modification make sense (duplicating inputs in the subgraph appear to me like a bug). I can modify `src/operator/subgraph/common.h` accordingly if you agree with my modification on CutGraphInputs.

KellenSunderland:
CI should be in reasonably good shape now.  Looks like there's some linting issues in a few headers:
```
=====269/271 cpp-header files passed check=====
src/operator/subgraph/tensorrt/nnvm_to_onnx-inl.h: 1 Errors of 1 Categories map={'build': 1}
src/operator/subgraph/tensorrt/tensorrt-inl.h: 4 Errors of 2 Categories map={'whitespace': 1, 'build': 3}
```

reminisce:
Great to see this is happening. I have two high-level comments:
1. If you use the subgraph API, there should be no needs to add specific Python functions (e.g. `init_tensorrt_params`) to use TensorRT as a backend inference engine. I think everything can be handled in backend to support Module and simple_bind APIs in the frontend.
2. Have you considered adopting TensorRT C++ APIs to convert MXNet subgraph IR to TensorRT IR? This would allow you to get rid of dependency of protobuf, onnx, and onnx-trt. I used to find building MXNet with those third-party libs is a painful process, especially on edge devices. In addition, simply relying on the TensorRT C++ APIs allows we to extend operator coverage promptly once new release of TensorRT is out. I did this while integrating TensorRT with TVM, which shares the same graph IR as in MXNet.
https://github.com/reminisce/tvm/blob/subgraph_integration/src/contrib/subgraph/tensorrt_executor.cc

Caenorst:
@reminisce
1) If you know how to do it I'd like your help, I asked the question on the dev list and have discussed with Kellen Sunderland and Zheng Da, and we haven't find an easy solution. The final objective later would be to modify function that do parameters initialization.

2) so far on our side, there is no plan to integrate TensorRT in MxNet without ONNX. It's easier for us to rely on onnx-tensorrt for reusability

reminisce:
@Caenorst 
Just want to clarify, I'm not blocking this PR. We can think through about these comments and make incremental changes afterwards.

The point is adding zero overhead in terms of user experience of using TensorRT in MXNet. I don't think you need to pass the param data explicitly beforehand. They can all be collected at the first forward call for each subgraph. You can take a look how this is done in [TVM](https://github.com/reminisce/tvm/blob/subgraph_integration/src/contrib/subgraph/tensorrt_executor.cc#L918). On the high level, in the forward function of a tensorrt subgraph op, you have a list of inputs which contains both input data and params. You just need to differentiate param names from input data names. This can be achieved by analyzing whether a data node's `NodeEntry` is an operator's input data or param.

Regarding the dependence on protobuf, onnx, and onnx-tensorrt, here is my two cents, I personally don't think introducing so many third-party dependencies is a good idea, because it results in an extra layer of complexity in terms of implementation, maintenance, build, and deployment, and the conversion process is not manageable by this community. If there is a direct solution of creating a bridge from nnvm to tensorrt, why not use it?

Caenorst:
@reminisce
I completely get your points, we thought about loading the weights from inputs, the problem is that you can't mix context in a single graph and as far as I know you can't delete the weights from the graph, so it will force to have the twice the weights in GPU memory, which may become a problem at some point. Ideally if we can somehow at least deallocate the memory of the weights once copied inside the TensorRT engine that would be nice, I haven't tried to see if it is doable tho

reminisce:
> @reminisce
> you can't mix context in a single graph
What do you mean "mix context"? We only have one context which gpu in this case.

Regarding releasing the weight params after they are copied to tensorrt engines, I think it's doable by removing the corresponding `NDArray` from `data_entry_` in `GraphExecutor`.

Caenorst:
@reminisce 
Yes, what I mean is that if weights for TensorRT node are on CPU while the rest of the graph is on GPU, it would be less of a problem. Also an argument for onnx-tensorrt is that there is more Ops supported with plugins implemented (slice, some activation, resize with nearest interpolation...)

reminisce:
> weights for TensorRT node are on CPU while the rest of the graph is on GPU.

This is not true. When binding completes, you have all the weights on GPU.

reminisce:
> Also an argument for onnx-tensorrt is that there is more Ops supported with plugins implemented (slice, some activation, resize with nearest interpolation...)

Are you saying that there are operators supported by TensorRT but creating them is not available through the C++ APIs? Have those operators been added to the `unconditionalTRTop`? The TRT C++ API also supports plugin interface.

ankkhedia:
@Caenorst Could you look fix merge conflicts and retrigger CI?
@reminisce Could you please help @Caenorst with his questions?

Caenorst:
> > weights for TensorRT node are on CPU while the rest of the graph is on GPU.
> 
> This is not true. When binding completes, you have all the weights on GPU.

Yes, that's what I meant, weights have to be copied from CPU to GPU, so they have to be originally on CPU. Actually they have to be copied first in the ONNX model

karan6181:
@KellenSunderland and @anirudh2290 could you please review this PR again. 

Thanks !

KellenSunderland:
@karan6181 Taking a look this week.

KellenSunderland:
Hey @Caenorst  Sorry can you rebase this once more?  There's no reported conflicts from Github but for some reason when running in CI it's not picking up some changes which means the R tests will fail (they're trying to use an old domain).

abhinavs95:
@Caenorst Thank you for making the changes. @KellenSunderland is this good to go?

abhinavs95:
@mxnet-label-bot add [pr-awaiting-testing]

KellenSunderland:
LGTM, did some testing on a home machine.

KellenSunderland:
Are you able to reproduce this locally @Caenorst?  I've run through quite a few times now and haven't hit the issue.

KellenSunderland:
Scratch that, able to reproduce now.  Digging a bit further in ...

KellenSunderland:
Ok I see what's happening.  Looks like we've done some refactoring to help prevent shutdown engine crashes but still have some fixed to apply before it's working 100%.  I believe this PR should solve the issue we're seeing in the test: https://github.com/apache/incubator-mxnet/pull/14591  Would you be able to test with that commit applied?

KellenSunderland:
Ok issue that was tripping up our tests has been fixed.  Would you be able to do another rebase (sorry) and then we should be able to merge.

roywei:
@Caenorst thanks for the contribution, gentle ping to rebase so we can merge this PR.

Caenorst:
Rebase done. Sorry for the delay.

KellenSunderland:
No worries about the delay.  My team's been testing the TRT integration out and are seeing better than expected speedups so far.  Let's merge this change.  I'm adding a todo for myself to update all documentation relating to the API change in our TRT tutorials.  We've also found a few minor bugs we'll hope to PR soon.

ThomasDelteil:
@KellenSunderland @Caenorst , the tutorial for tensorrt is out of the date and I have had a few user asking us why it's not working anymore https://mxnet.incubator.apache.org/versions/master/tutorials/tensorrt/inference_with_trt.html
What is the current recommended way to use tensorRT ?
Thanks!

Caenorst:
@ThomasDelteil the PR for updating it is on its way: https://mxnet.incubator.apache.org/versions/master/tutorials/tensorrt/inference_with_trt.html

Meanwhile here is the new way to use TensorRT:
```
sym, arg_params, aux_params = mx.model.load_checkpoint('my_net', epochs)
trt_sym = sym.get_backend_symbol('TensorRT')
remaining_arg_params, remaining_aux_params = mx.contrib.tensorrt.init_tensorrt_params(trt_sym, arg_params, aux_params)
```
And then you can use `trt_sym` as your usual symbol, no more `tensorrt_bind` or `MXNET_USE_TENSORRT` environment variable. Also now FP16 computation is by default if you want to disable it use the function `mx.contrib.tensorrt.set_use_fp16(False)`

ThomasDelteil:
Thanks @Caenorst for the quick update!

KellenSunderland:
Yes, sorry let me update that PR to target the 1.5 branch.

Issue #14045, opened 2019-02-01
Flaky Test: Clojure CPU org.apache.clojure-mxnet.operator-test


Chancebair:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-cpu/detail/master/281/pipeline

```
Testing org.apache.clojure-mxnet.operator-test

FAIL in (test-concat) (operator_test.clj:72)
expected: (= out1 ret)
  actual: (not (= #object[org.apache.mxnet.NDArray 0x30bac2a0 "[\n [2.6332547E-28,4.5834E-41]\n [1.12572034E24,1.5341626E28]\n [2.9223691E-27,4.5834E-41]\n [1.12572034E24,4.4632523E30]\n [NaN,4.5828E-41]\n]\n<NDArray (5,2) cpu(0) float32>"] #object[org.apache.mxnet.NDArray 0x189d74e4 "[\n [2.6332547E-28,4.5834E-41]\n [1.12572034E24,1.5341626E28]\n [2.9223691E-27,4.5834E-41]\n [1.12572034E24,4.4632523E30]\n [NaN,4.5828E-41]\n]\n<NDArray (5,2) cpu(0) float32>"]))

FAIL in (test-concat) (operator_test.clj:80)
expected: (= grads (mapv (fn* [p1__8963#] (ndarray/+ p1__8963# 1)) np-grads))
  actual: (not (= (#object[org.apache.mxnet.NDArray 0x3a5537e6 "[\n [1.0,1.0]\n [1.12572034E24,1.5341626E28]\n]\n<NDArray (2,2) cpu(0) float32>"] #object[org.apache.mxnet.NDArray 0x51574c53 "[\n [1.0,1.0]\n [1.12572034E24,4.4632523E30]\n [NaN,1.0]\n]\n<NDArray (3,2) cpu(0) float32>"]) [#object[org.apache.mxnet.NDArray 0x5009b104 "[\n [1.0,1.0]\n [1.12572034E24,1.5341626E28]\n]\n<NDArray (2,2) cpu(0) float32>"] #object[org.apache.mxnet.NDArray 0x17de104d "[\n [1.0,1.0]\n [1.12572034E24,4.4632523E30]\n [NaN,1.0]\n]\n<NDArray (3,2) cpu(0) float32>"]]))

Embedded symbol: {
  "nodes": [
    {
      "op": "null", 
      "name": "data", 
      "inputs": []
[...]

PR #14049, opened 2019-02-01
[Clojure] Fix for flaky concat test failures

gigasquid:
Fix for issue https://github.com/apache/incubator-mxnet/issues/14045

Ran new concat test 10000 times with no failures

Also ran the whole Operator Suite 1000 times with no failures


vandanavk:
@mxnet-label-bot add [pr-awaiting-review, Clojure]

PR #14053, opened 2019-02-02
in-place reshape ops

### LINKED ISSUE #13998: expand_dims() makes copy instead of simply reshaping
Hi,

Not sure if this is expected behavior or a bug, but I noticed somewhat surprisingly that all `expand_dims()` calls in some code we were writing were making full copies of the underlying array we were expanding. In our case, the only reason to expand was to be able to broadcast arithmetic operations correctly, and it is unacceptable to make a copy. Our current work around is to use reshape directly.

See example code which shows my point:
```
import mxnet as mx

a = mx.nd.array([0])

reshaped = a.reshape(1,1)
reshaped[0,0] = 5
print(a) # prints 5, because reshaped is a wrapper over a                                                                                                                 

expanded = a.expand_dims(1)
expanded[0,0] = 10
print(a) # prints 5, because expanded is a *copy* of a 
```

I personally feel like expand_dims() should have the same behavior as reshape() here, but just be a more fluent way of reshaping an array in a particular way. Therefore I feel like it's a bug for it to make a copy.

However I don't have the full context, and maybe there's a reason expand_dims needs to make a copy? Please let me know--and if not I can look into making a PR to get rid of spurious copy.

Thanks,
Stephen

szha:
## Description ##
make NDArray fluent methods for expand_dims/flatten/squeeze in-place reshapes.

### Essentials ###

### Changes ###
- [x] make NDArray fluent methods for expand_dims/flatten/squeeze in-place reshapes.

## Comments ##
- resolves #13998

### DIFF
--- a/python/mxnet/ndarray/ndarray.py
+++ b/python/mxnet/ndarray/ndarray.py
@@ -1259,8 +1259,14 @@ fixed-size items.
         """
         return op.sign(self, *args, **kwargs)
 
-    def flatten(self):
-        """Returns a flattened **view** of this array without altering any data.
+    def flatten(self, inplace=False):
+        """Flatten this array without altering any data.
+
+        Parameters
+        ----------
+        inplace : bool, default False
+            If True, this method returns a **view** of this array
+            that shares data with this array. Otherwise, a copy is returned.
 
         Returns
         -------
@@ -1271,7 +1277,8 @@ fixed-size items.
         Examples
         --------
         >>> x = mx.nd.arange(30).reshape(5,2,3)
-        >>> y = x.flatten()
+        >>> y = x.flatten(inplace=True)
+        >>> z = x.flatten()
         >>> y.shape
         (5, 6)
         >>> y[0].asnumpy()
@@ -1280,8 +1287,10 @@ fixed-size items.
         >>> x[0].asnumpy()
         array([[-1., -1., -1.],
                [-1., -1., -1.]], dtype=float32)
+        >>> z[0].asnumpy()
+        array([0., 1., 2., 3., 4., 5.], dtype=float32)
         """
-        return self.reshape((0, -1))
+        return op.flatten(self) if not inplace else self.reshape((0, -1))
 
     def shape_array(self, *args, **kwargs):
         """Convenience fluent method for :py:func:`shape_array`.
@@ -1299,8 +1308,8 @@ fixed-size items.
         """
         return op.size_array(self, *args, **kwargs)
 
-    def expand_dims(self, axis):
-        """Returns a **view** of this array with additional dimension without altering any data.
+    def expand_dims(self, axis, inplace=False):
+        """Adds an additional dimension to the current array without altering any data.
 
         Parameters
         ----------
@@ -1308,6 +1317,9 @@ fixed-size items.
             Position where new axis is to be inserted.
             Suppose that the input NDArray's dimension is ndim,
             the range of the inserted axis is [-ndim, ndim].
+        inplace : bool, default False
+            If True, this method returns a **view** of this array
+            that shares data with this array. Otherwise, a copy is returned.
 
         Returns
         -------
@@ -1319,7 +1331,8 @@ fixed-size items.
         Examples
         --------
         >>> x = mx.nd.arange(6).reshape(2,3)
-        >>> y = x.expand_dims(1)
+        >>> y = x.expand_dims(1, inplace=True)
+        >>> z = x.expand_dims(1)
         >>> y.shape
         (2, 1, 3)
         >>> y[0].asnumpy()
@@ -1328,14 +1341,19 @@ fixed-size items.
         >>> x.asnumpy()
         array([[-1., -1., -1.],
                [-1., -1., -1.]], dtype=float32)
+        >>> z[0].asnumpy()
+        array([[0., 1., 2.]], dtype=float32)
         """
-        new_shape = list(self.shape)
-        assert -len(new_shape) <= axis <= len(new_shape), \
-                "axis {} is out of range for {}d array".format(axis, len(new_shape))
-        if axis < 0:
-            axis += len(new_shape)
-        new_shape.insert(axis, 1)
-        return self.reshape(new_shape)
+        if not inplace:
+            return op.expand_dims(self, axis=axis)
+        else:
+            new_shape = list(self.shape)
+            assert -len(new_shape)-1 <= axis <= len(new_shape), \
+                    "axis {} is out of range for {}d array".format(axis, len(new_shape))
+            if axis < 0:
+                axis += len(new_shape) + 1
+            new_shape.insert(axis, 1)
+            return self.reshape(new_shape)
 
     def tile(self, *args, **kwargs):
         """Convenience fluent method for :py:func:`tile`.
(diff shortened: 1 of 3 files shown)


ZhennanQin:
Have you tried this with mkldnn enabled? E.g reshape the output of mkldnn convolution?

szha:
@ZhennanQin I didn't test that specifically based on the assumption that the backend implementation should not cause existing APIs such as `NDArray.reshape` or `NDArray.shape` to break.

stephenrawls:
@szha -- Thanks for putting in this patch!

I have a couple questions about the operator `expand_dims`, which I understand you are actually changing the python code to *not* use that operator, but I think still has relevance.

(1) I originally discovered this because I was using the C API to call MXImperativeInvoke() using the expand_dims operator, and I noticed it was causing a copy. This fix only effects the Python version when operating on ndarrays, not users of the expand_dims operator.

I see that there is a path in the operator that checks if it is an in-place operation. Presumably it uses that path if I pass an output array that is the same NDArrayHandle as the input array? But what if I still need the original input array handle, and I want to create a new output array handle with the expanded dim but still not make a copy?

(2) In the issue I created you commented that: "For symbol (and thus the hybridized version), since in-place identity is possible it should not matter".

Can you talk a little more about that? I assume you mean that in this case:
```
x_expanded = x.expand_dims(1)
y = x_expanded + foo
```
The engine can figure out that x is not needed again, and can thus turn the expand_dims(1) into an in-place operation that doesn't make a copy?

I'm not very familiar with how this part of the code works, so what happens if you had code that looked like this?
```
x_expanded = x.expand_dims(1)
y = x_expanded + foo
z = 2 * x
```
i.e. the code still makes a reference to the original x, and thus presumably the engine can't decide to use the in-place version of expand_dims in that case, right? So I guess my question is -- Does the ability for the Syblolic / hybridized engine to elide the copy depend on the code not referencing the un-expanded version of the array after calling expand_dims()? If so, it seems like there will still be some use cases where an unexpected copy is happening.

szha:
@stephenrawls

> This fix only effects the Python version when operating on ndarrays, not users of the expand_dims operator.

Right. The reshape operators always return a copy. This is because in imperative mode there is no complete computation graph to be analyzed and see whether a copy is needed or not.

> I see that there is a path in the operator that checks if it is an in-place operation. Presumably it uses that path if I pass an output array that is the same NDArrayHandle as the input array?

Right, it is done through `output=original_array` argument to the operators. It does not work for autograd (i.e. Gluon's training mode) as autograd doesn't support in-place operation.

> But what if I still need the original input array handle, and I want to create a new output array handle with the expanded dim but still not make a copy?

As long as it's not used in the context of autograd, you can save the original handle before in-place operation. Though the output of in-place op shares the space with the input, it's still a separate handle.

> For symbol (and thus the hybridized version), since in-place identity is possible it should not matter.

To elaborate, in symbolic mode, the backend can see the complete computation graph, and as a result it can analyze the graph (through node coloring) and figure out whether nodes can share the same space. `FInplaceIdentity` is the graph node attribute that expresses whether this should be allowed in order for compiler to plan memory. For details see [3rdparty/tvm/nnvm/src/pass/plan_memory.cc](https://github.com/dmlc/tvm/blob/0f053c82a747b4dcdf49570ec87c17e0067b7439/nnvm/src/pass/plan_memory.cc).

Note that such memory plan happens before the engine/scheduler work happens, and it happens for both symbolic executor and hybridized Gluon HybridBlock.

stephenrawls:
@szha -- Thanks for the detailed explanation, that helps a lot.

So just for clarity, I see that in the operator definition for expand_dims, it sets the FInPlaceIdentity attribute to true:
https://github.com/apache/incubator-mxnet/blob/45d1a1e6ff9c58cfc75f72651c1bf671ac7f1885/src/operator/tensor/matrix_op.cc#L400-L419

I didn't see anything in the plan_memory.cc file that depended on autograd / training.

So should I take it that the code is smart enough to not allocate new memory for the expand_dims operator in symbolic mode, i.e. not to make a copy, even during training when using autograd?

If so, then I think I understand it, and I see what you mean about only needing to fix this for the Python ndarray use case.

Thanks again!
Stephen

vandanavk:
@mxnet-label-bot add [pr-awaiting-review, NDArray]

szha:
@stephenrawls actually, autograd comes into play in pure imperative mode. MXNet's autograd is trace based and does not support in-place operation such as `input = mx.nd.expand_dims(input, ..., output=input)`. You're actually right in that it doesn't help the case where you write it in imperative mode in C++. In order to make use of the graph pass for saving memory, you need to use Symbol API to express the model and use symbolic execution (or its counter part for Gluon, which is CachedOp).

TaoLv:
Does it mean that with these changes python API will have different behavior compared with other frontend languages?

szha:
@TaoLv Other language bindings don't seem to have the fluent methods for reshaping NDArray.

szha:
The current implementation would break backward compatibility for in-place update.

szha:
I plan on adding an `inplace` argument to these fluent methods to allow users to specify whether in-place operation should be used.

ankkhedia:
@reminisce Could you please help review the PR?

szha:
I haven't finished this PR yet as I haven't got around to add the in-place option.

ankkhedia:

ankkhedia:

python/mxnet/ndarray/ndarray.py
```diff
@@ -1701,13 +1765,41 @@ def softmin(self, *args, **kwargs):
...
+            return op.squeeze(self, axis=axis)
+        else:
+            new_shape = list(self.shape)
+            if isinstance(axis, int):
+                axis = [axis]
+            if axis:
+                assert len(axis) == len(set(axis)), \
+                    "axis {} contains duplicate which is not allowed.".format(axis)
+                for i in axis:
+                    assert 0 <= i < len(new_shape), "axis {} is out of range.".format(i)
```
reminisce: Negative axes should be supported.

szha:
@reminisce good catch. I was relying on existing tests before adding the inplace option but forgot to add tests for them afterwards.

szha:
Added tests

szha:
@wkcn good point on symbol and HybridBlock. I'm not sure if a warning should be added as it can get very verbose.

szha:
Nonetheless it should not throw an error when using inplace with hybridize. Let me see how to best deal with it.

karan6181:
@szha Thank you for your contributions! Is this PR work in progress?

abhinavs95:
@szha is this PR ready to merge?

piyushghai:
@szha Is this good to merge ?

szha:
Thanks for all the reviews. I plan to add a dummy option in the symbol version of the reshape ops so that it won't throw error when people use this option in the context of Gluon

roywei:
@mxnet-label-bot add[pr-work-in-progress]

roywei:
@mxnet-label-bot remove [pr-awaiting-response]

karan6181:
@szha Did you get a chance to add a dummy option in the symbol version? Thanks!

piyushghai:
@szha @wkcn This PR is ready to be merged I guess ? 

@mxnet-label-bot Update [NDArray, pr-awaiting-merge]

szha:
@piyushghai thanks for pinging. I wanted to hold onto this for the 1.5 release code freeze.

abhinavs95:
@szha is this good to go now?

PR #14063, opened 2019-02-04
fix test_stn

szha:
## Description ##
fix test_stn display setting which likely has been broken by dependency change.


szha:
@apache/mxnet-committers your attention is appreciated as the problem fixed in this PR is blocking the CI builds for other PRs.

KellenSunderland:
LGTM, thanks for fixing this @szha.

PR #14097, opened 2019-02-08
Relaxing type requirements for slice_like op

ptrendx:
## Description ##
Currently the type of the second argument in `slice_like` op (`shape_from` ) needs to be the same as the first argument, even though only its (second argument's) shape is used. This PR relaxes that so `slice_like` accepts any type for the `shape_from` argument.

### Essentials ###

### DIFF
--- a/tests/python/unittest/test_operator.py
+++ b/tests/python/unittest/test_operator.py
@@ -2516,6 +2516,20 @@ def test_slice_like():
             assert_allclose(xgrad1.asnumpy(), mx.nd.zeros_like(xgrad1).asnumpy())
 
 @with_seed()
+def test_slice_like_different_types():
+    x = [[  1.,   2.,   3.,   4.],
+         [  5.,   6.,   7.,   8.],
+         [  9.,  10.,  11.,  12.]]
+
+    y = [[  0.,   0.,   0.],
+         [  0.,   0.,   0.]]
+
+    x = mx.nd.array(x)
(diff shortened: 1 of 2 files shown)


vandanavk:

tests/python/unittest/test_operator.py
```diff
@@ -2515,6 +2515,20 @@ def test_slice_like():
             assert_allclose(xx, xgrad.asnumpy())
             assert_allclose(xgrad1.asnumpy(), mx.nd.zeros_like(xgrad1).asnumpy())
 
+@with_seed()
+def test_slice_like_different_types():
+    x = [[  1.,   2.,   3.,   4.],
```
ChaiBapchya: Though this example looks good, I am afraid it is specific. Do you mind using randomized shape, size and NDarray initalization to ensure this is generic?

eric-haibin-lin:
@zhreshold

tests/python/unittest/test_operator.py
ptrendx: Those parameters do not matter for this test (and they are already tested in the dedicated test of `slice_like` op functionality) - the values for `x` and `y` were taken from the example from `slice_like` documentation. The only thing that matters is the type of `y` in this test. Before this PR this code would not work - it would throw an exception during type inference. With this PR, this code works (as it should).

ChaiBapchya: Alright. Makes sense. Thanks for the explanation.

PR #14123, opened 2019-02-12
Add int8 data loader

ZhennanQin:
@reminisce @zheng-da @szha @KellenSunderland @pengzhao-intel @TaoLv 
PR major change:
* introduce int8 data loader. This can be used for quantized network with rgb_mean != (0, 0, 0).

### Essentials ###

### Changes ###

## Comments ##


ankkhedia:
@ZhennanQin Thanks for your contribution!

ankkhedia:
@ZhennanQin Thanks for your contribution!

ZhennanQin:
@anirudh2290 Thanks for reviewing. That idea was in my head, but I was a bit worried about letting quantize_v2 accept int8 dtypes looks a bit weird, and would not be accepted by community.  I have to say, that way is good for user experience as user can use single int8 model for both fp32 and int8 data loader. If you think it's doable, then I will refactor towards it.

anirudh2290:
Yes I would prefer that and I think it would be doable. Would need to take care of inplace optimization if input and output is int8.

ZhennanQin:
@anirudh2290 I tried to add FInplaceOption and FInplaceIdentity to quantize_v2 to optimize the case that input and ouput is int8. But failed with this check in plan memory pass:
https://github.com/dmlc/tvm/blob/master/nnvm/src/pass/plan_memory.cc#L227
```
sid_in >= 0 &&
```

`export MXNET_MEM_PLAN_VERBOSE_LOGGING=1` provides below logging:
```
[14:02:23] src/executor/../common/exec_utils.h:300: node 0 var
[14:02:23] src/executor/../common/exec_utils.h:302: node 1 _contrib_quantize_v2
[14:02:23] src/executor/../common/exec_utils.h:306: 		input 0: [1,3,224,224] (147 KB) -> var (-1)
[14:02:23] src/executor/../common/exec_utils.h:312: 		output 1: [1,3,224,224] (147 KB) -> group 0
[14:02:23] src/executor/../common/exec_utils.h:312: 		output 2: [1] (0 KB) -> group 2
[14:02:23] src/executor/../common/exec_utils.h:312: 		output 3: [1] (0 KB) -> group 1
```
It seems that, data is an variable of this network, so its `sid_in` is -1, making it can't be inplaced with  quantize_v2 output. Is there any reason we can't inplace var with op output?

example/quantization/imagenet_inference.py
```diff
@@ -192,24 +210,52 @@ def benchmark_score(symbol_file, ctx, batch_size, num_batches, logger=None):
...
+                                              batch_size=batch_size,
+                                              data_shape=data_shape,
+                                              label_name=label_name,
+                                              rand_crop=False,
+                                              rand_mirror=False,
+                                              shuffle=args.shuffle_dataset,
+                                              shuffle_chunk_seed=args.shuffle_chunk_seed,
+                                              seed=args.shuffle_seed,
+                                              **combine_mean_std)
+        else:  #int8
```
zhreshold: should reorder the types and let float32 be the fallback default type?

ZhennanQin: Sounds reasonable. Will change the order later.

anirudh2290:
apologies for the delay,  I made an attempt on my end to make quantize use inplace optimization when in dtype is also int8. I couldnt find a way to save state in InferType to reuse it within function registered for Inplace optimization. Here is my attempt: https://github.com/anirudh2290/mxnet/commit/cc6bf9e96d982a31bb6bf953d76b2d6ce3525f22 . Also, I am not sure why input var cant be inplaced and would like to reproduce this issue at my end. Would you be able to share your current code for the inplace optimization for quantization op. If you are short on time, I think we can proceed with the quantization implementation without inplace optimization.

ZhennanQin:
@anirudh2290 No need to apologies, you're one of the most active reviewers I've seen in MXNet project:)
The patch letting quantize_v2 accept int8 is ready at https://github.com/ZhennanQin/incubator-mxnet/pull/5. You can apply it based on this PR. Alternately,  you can pick whole patch with int8 data loader at https://github.com/ZhennanQin/incubator-mxnet/pull/6.

`I couldnt find a way to save state in InferType to reuse it within function registered for Inplace optimization.` Yes, that's the major problem when apply inplace optimization. But luckily, there's a strict dtype check in memory planning pass, that allows memory inplace happen only when input and output has same dtype. So registering FInplaceOption and FInplaceIdentity to quantize_v2 is enough.

anirudh2290:
Thanks @ZhennanQin ! Good to know that `PlanMemory` pass checks for the dtypes. will try out your patch later today.

src/operator/quantization/quantize_graph_pass.cc
```diff
@@ -156,6 +157,10 @@ Graph QuantizeGraph(Graph &&src) {
         if (avoid_quantize_input_map.count(node->op()) &&
             avoid_quantize_input_map[node->op()](node->attrs, i)) {
           new_node->inputs.emplace_back(mirror_entry);
+        // If network will directly accept quantized data, simply add data as input.
+        } else if (use_quantized_data_layer && e.node->is_variable() &&
+                   e.node->attrs.name == "data") {
```
reminisce: This is a very strong assumption on the name of input variable node. One solution is allowing users to specify the names of inputs and pass them as a graph attribute. By default, it can be called `data` if not provided. Need to document the usage well.

ZhennanQin: Agree, this can be removed if  quantize_v2 can directly accept int8.

anirudh2290:
I guess there are good reasons for not allowing inplace operations for inputs. One reason could be that the data allocation for inputs is controlled by the user instead of engine and it should not be mutated. We can argue that identity operations should still be supported, but this means that the memory cannot be reused if identity is followed by an inplace operation. Also, the reference count for reused input memory will never drop to 0. This means that it cannot be recycled and reused in the computation graph. 

According to me, even if we cannot do inplace optimization we should still add support in the operator for int8 to make the user experience easier.

ZhennanQin:
@anirudh2290 Thanks for having a look on this. I agree with your point that user experience is important. Then I will add operator support without inplace optimization. In future, we can implement another pass to delete such redundant op to get it optimized.

ZhennanQin:
PR is updated to address all comments received. Please review again. Thanks. @anirudh2290 @zhreshold @reminisce

anirudh2290:
Looks like other comments have been addressed. Merging now.

PR #14130, opened 2019-02-12
Refine runtime feature discovery python API and add documentation to …

larroy:
…Python API docs

## Description ##
see title

@aaronmarkham 

### Essentials ###

### Changes ###
- [x]  Decode feature name to string since it comes as a buffer
- [x]  Add __repr__ as beautifier

### DIFF
--- a/docs/api/python/libinfo/libinfo.md
+++ b/docs/api/python/libinfo/libinfo.md
@@ -10,7 +10,7 @@ The libinfo functionality allows to check for compile-time features supported by
 
 ### Example usage
 
-```python
+```
 In [1]: import mxnet as mx
 
 In [2]: import mxnet.runtime
--- a/docs/api/python/libinfo/libinfo.md
+++ b/docs/api/python/libinfo/libinfo.md
@@ -15,8 +15,14 @@ In [1]: import mxnet as mx
 
 In [2]: import mxnet.runtime
 
-In [3]: mx.runtime.libinfo_features()
-Out[3]:
+In [3]: mxnet.runtime.is_enabled('DEBUG')
+Out[3]: True
+
+In [4]: mxnet.runtime.is_enabled('CUDA')
+Out[4]: False
+
+In [5]: mx.runtime.libinfo_features()
+Out[5]:
 [✔ CUDA,
  ✔ CUDNN,
  ✔ NCCL,
--- a/docs/api/python/libinfo/libinfo.md
+++ b/docs/api/python/libinfo/libinfo.md
@@ -12,47 +12,22 @@ The libinfo functionality allows to check for compile-time features supported by
 
 ```
 In [1]: import mxnet as mx
+   ...: import mxnet.runtime
+   ...: fs = mx.runtime.Features()
 
-In [2]: import mxnet.runtime
+In [2]: fs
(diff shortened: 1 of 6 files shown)


docs/api/python/libinfo/libinfo.md
```diff
@@ -0,0 +1,70 @@
...
+ ✔ BLAS_APPLE,
+ ✔ LAPACK,
+ ✔ MKLDNN,
+ ✖ OPENCV,
+ ✔ CAFFE,
+ ✔ PROFILER,
+ ✔ DIST_KVSTORE,
+ ✔ CXX14,
+ ✔ SIGNAL_HANDLER,
+ ✖ DEBUG]
```
iblis17: :heart: this

ankkhedia:
@larroy Thanks for your contribution!
@mxnet-label-bot add [pr-awaiting-review, Doc]

docs/api/python/libinfo/libinfo.md
larroy: I copied from you :-)

python/mxnet/runtime.py
```diff
@@ -28,21 +29,48 @@ class LibFeature(ctypes.Structure):
...
+    :param x: str The name of a valid feature as string for example 'CUDA'
+
+    Returns
+    -------
+    :return: bool True if it's enabled, False if it's disabled, RuntimeError if the feature is not known
+    """
+    feature_dict = {f.name: f.enabled for f in libinfo_features()}
+    if tocheck not in feature_dict:
+        raise RuntimeError("Feature '{}' is unknown, known features are: {}".format(tocheck, list(feature_dict.keys())))
+    return feature_dict[tocheck]
```
szha: if a dictionary is easier to operate, should the libinfo_features return a dictionary?

larroy: I prefer to keep ordering as in a list. Going from list to dict is oneliner.

szha: is the order information useful? I always thought of it as a set. also, by defining your own data structure class, you can define more pythonic behaviors such as `__contains__` and `__getitem__` to make it easier for feature check.

larroy: They are grouped if you see the enum definition as some features are related, semantically is a set as you describe. For using __contains__ as you propose we would have to wrap the list in a containing class. I don't see a lot of value on complicating this further TBH. Could you provide an example of what you have in mind if you feel strongly about it?

szha: I don't feel stongly about the change. What I had in mind was something like:
```python
>>> features = mx.runtime.libinfo_features()
>>> 'NCCL' in features
True
>>> 'nccl' in features
True
>>> nccl_idx = 2;
>>> nccl_idx in features
True
>>> features['NCCL']
✔ NCCL
>>> features['NCCL'].index
2
>>> features['NCCL'].enabled
True
```

larroy: I see, that looks nice. On the upside repeated calls are more efficient than my proposal. On the downside I think the primary case which is to check if a feature is enabled is less ergonomic due to the syntactic sugar. I think a simple function with a single arg is always clearer.  A good point is that it should be case insensitive. I'm a bit torn.

apeforest: Just curious is the index information useful to user? Why do we even need to expose index?

tests/python/unittest/test_runtime.py
```diff
@@ -29,6 +29,17 @@ def test_libinfo_features():
     ok_(type(features) is list)
     ok_(len(features) > 0)
 
+def test_is_enabled():
```
apeforest: Can this test run in a docker image with CUDA on the machine without GPU?

python/mxnet/runtime.py
larroy: @apeforest I just removed it.

tests/python/unittest/test_runtime.py
larroy: Not sure I get what's the intention.

larroy:
I made the suggested changed, please have another look and merge if you are good with it, thanks.

docs/api/python/libinfo/libinfo.md
```diff
@@ -0,0 +1,51 @@
...
+
+In [6]:
+```
+
+
+```eval_rst
+.. autosummary::
+    :nosignatures:
+
+    LibFeature
```
szha: outdated.

larroy:
Thanks for the merge and the review.

PR #14153, opened 2019-02-14
Fix shape inference pass

ptrendx:
## Description ##

This PR fixes shape inference, which currently results in wrong result for some cases.

An example problematic case (as it works in current version of MXNet, before applying this change):
If I do
```
import mxnet as mx

data = mx.sym.Variable('data', shape=(1,0,512,512))
weight = mx.sym.Variable('weight')
cdata = mx.sym.cast(data, dtype='float16')
cweight = mx.sym.cast(weight, dtype='float16')
test = mx.sym.Convolution(data=cdata, weight=cweight,layout='NCHW', pad=(3, 3), num_filter=64, stride=(2, 2), no_bias=True, dilate=(1, 1), kernel=(7, 7), num_group=1)

print(test.infer_shape_partial())
```
I get expected result:
```
([(1, 0, 512, 512), (64, 0, 7, 7)], [(1, 64, 256, 256)], [])
```
but when I change H and W dimensions in the shape to 0s,
```
import mxnet as mx

data = mx.sym.Variable('data', shape=(1,0,0,0))
weight = mx.sym.Variable('weight')
cdata = mx.sym.cast(data, dtype='float16')
cweight = mx.sym.cast(weight, dtype='float16')
test = mx.sym.Convolution(data=cdata, weight=cweight,layout='NCHW', pad=(3, 3), num_filter=64, stride=(2, 2), no_bias=True, dilate=(1, 1), kernel=(7, 7), num_group=1)

print(test.infer_shape_partial())
```
I get
```
([(1, 0, 0, 0), ()], [(1, 64, 0, 0)], [])
```
so the shape of the weight changed to `()`.

### Essentials ###

### Changes ###
- [x] Change to the way passes are done so that both forward and backward inference is performed every time - I'm not sure if this is necessary - @eric-haibin-lin, thoughts? 
- [x] Change to the way shape inference works. Currently a shape contributes 1 to the `num_unknown` if it has at least 1 zero. After the change number of 0 elements is added to the `num_unknown` - that way shape inference pass does not end prematurely if only some of the elements of shape were deduced.

### DIFF
--- a/src/executor/infer_graph_attr_pass.cc
+++ b/src/executor/infer_graph_attr_pass.cc
@@ -383,7 +383,7 @@ nnvm::Graph InferShapeAttr(nnvm::Graph &&ret,
                            const DispatchMode default_mode_val = DispatchMode::kUndefined) {
   using nnvm::IndexedGraph;
   using nnvm::Op;
-  using AttrType = nnvm::TShape;
+  using AttrType = mxnet::TShape;
   using FInferType = nnvm::FInferShape;
   using AttrVector = std::vector<AttrType>;
   using NodeAttrVector = std::vector<DispatchMode>;
--- a/src/executor/infer_graph_attr_pass.cc
+++ b/src/executor/infer_graph_attr_pass.cc
@@ -384,7 +384,7 @@ nnvm::Graph InferShapeAttr(nnvm::Graph &&ret,
   using nnvm::IndexedGraph;
   using nnvm::Op;
   using AttrType = mxnet::TShape;
-  using FInferType = nnvm::FInferShape;
+  using FInferType = mxnet::FInferShape;
   using AttrVector = std::vector<AttrType>;
   using NodeAttrVector = std::vector<DispatchMode>;
   using dmlc::any;
--- a/src/executor/infer_graph_attr_pass.cc
+++ b/src/executor/infer_graph_attr_pass.cc
@@ -68,6 +68,26 @@ bool ApplyOpInferAttr<int, FInferStorageType>(const nnvm::Graph& g,
  * shape/type inference functions'. The nnvm InferAttr will be deprecated
  * in the future. Please use interfaces InferShape, InferType, and InferStorageType
  * to call this function.
+ *
+ * \param ret graph used for attribute inference
+ * \param emmpty_val empty value of the attribute
+ * \param infer_name name of the function used for attribute inference
+ * \param input_name name of the attribute in the graph used to store the
+ *                   input data for attribute inference
+ * \param attr_key_name name of the attribute used for inference for variable nodes
+ * \param attr_name name of the inferred attribute
+ * \param unknown_name name of the attribute storing number of entries
+ *                     impossible to infer
+ * \param fis_none function returning true for not fully inferred values
+ * \param fdefault default function used for inference if the node does not
+ *                 provide its own implementation.
+ * \param bwd_identity_assign whether the attributes of forward NDArray and backward
+ *                            NDArray have to be the same. False only for storage
+ *                            type inference
+ * \param dispatch_mode_name name of the dispatch mode attribute on the node. Used for
+ *                           storage type inference
+ * \param default_mode_val default value of the dispatch mode attribute on the node. Used
+ *                         for storage type inference
  */
 template<typename AttrType, typename FInferType, typename IsNone, typename FDefault>
 nnvm::Graph InferAttr(nnvm::Graph &&ret,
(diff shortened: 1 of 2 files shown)


ankkhedia:
@ptrendx Thanks for your contribution! Could you please look into CI failures?

@mxnet-label-bot add [pr-awaiting-review , python]

src/executor/infer_graph_attr_pass.cc
```diff
@@ -282,26 +285,28 @@ nnvm::Graph InferAttr(nnvm::Graph &&ret,
...
+  size_t num_unknown = static_cast<size_t>(-1);  // Infinity
+
+  if (scalars_only) {
+    size_t num_unknown_dispatch_mode = dispatch_mode_name ? node_end - node_start : 0;
+    size_t num_unknown_entry_attr = entry_end - entry_start;
+    num_unknown = num_unknown_entry_attr + num_unknown_dispatch_mode;
+  }
   int i = 0;
   do {
-    if (i % 2 == 0) {
```
eric-haibin-lin: why is `if (i % 2 == 0)` removed?

ptrendx: As I've written in the description `Change to the way passes are done so that both forward and backward inference is performed every time - I'm not sure if this is necessary - @eric-haibin-lin, thoughts?` - I'm not sure if this is necessary. I did remove it because it seemed like a bug - let's say you only have the information on the output instead of input, I was worried that then the inference pass would stop after just going forward, which would not infer anything. As I think about this more though, I think that even on the forward side the last op will actually decrease number of unknowns and so the `i = 1` step will be invoked. 
If that is true, then this `if` statement is correct there and I will put it back.

eric-haibin-lin: If the forward pass doesn't infer anything, then the backward pass won't either. I think the original code is fine

src/executor/infer_graph_attr_pass.cc
```diff
@@ -337,7 +342,19 @@ nnvm::Graph InferShape(nnvm::Graph&& graph,
       "FInferShape", "shape_inputs", "shape_attr_key",
       "shape", "shape_num_unknown_nodes",
       [](const nnvm::TShape& s) { return s.ndim() == 0 || s.Size() == 0; },
-      nullptr, true, nullptr);
+      [](const nnvm::TShape& s) {
+        if (s.ndim() == 0) {
```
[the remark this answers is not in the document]
ptrendx: done

src/executor/infer_graph_attr_pass.cc
ptrendx: done

src/executor/infer_graph_attr_pass.cc
```diff
@@ -69,7 +69,8 @@ bool ApplyOpInferAttr<int, FInferStorageType>(const nnvm::Graph& g,
  * in the future. Please use interfaces InferShape, InferType, and InferStorageType
  * to call this function.
  */
-template<typename AttrType, typename FInferType, typename IsNone, typename FDefault>
+template<typename AttrType, typename FInferType, typename IsNone,
```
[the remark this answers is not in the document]
ptrendx: I added descriptions to params in the InferAttr function, but I was not really sure about some of them - could you look at the descriptions I made and check they are correct (as well as suggest what I should write in the ones I marked with `TODO`)?

src/executor/infer_graph_attr_pass.cc
```diff
@@ -68,8 +68,28 @@ bool ApplyOpInferAttr<int, FInferStorageType>(const nnvm::Graph& g,
...
+ * \param attr_name name of the inferred attribute
+ * \param unknown_name name of the attribute storing number of entries
+ *                     impossible to infer
+ * \param fis_none function returning true for not fully inferred values
+ * \param fnum_unknown function returning how many elements are unknown in
+ *                     partially inferred value of the attribute
+ * \param fdefault default function used for inference if the node does not
+ *                 provide its own implementation.
+ * \param scalars_only whether the inferred attribute may be only a scalar
+ * \param bwd_identity_assign TODO
```
reminisce: Only equal to `false` in inferring storage types. It means whether the attributes of forward ndarray and backward ndarray are necessarily the same.

src/executor/infer_graph_attr_pass.cc
```diff
@@ -68,8 +68,28 @@ bool ApplyOpInferAttr<int, FInferStorageType>(const nnvm::Graph& g,
...
+ * \param dispatch_mode_name TODO
```
eric-haibin-lin: Name of the dispatch mode attribute on the node. Used for storage type inference.

src/executor/infer_graph_attr_pass.cc
```diff
@@ -68,8 +68,28 @@ bool ApplyOpInferAttr<int, FInferStorageType>(const nnvm::Graph& g,
...
+ * param default_mode_val TODO
```
eric-haibin-lin: Default value of the dispatch mode attribute on the node. Used for storage type inference.

ptrendx:
@junrushao1994 I wanted to rebase this PR on top of the current master and I'm confused - in PR #14193 you made InferShapeAttr as a standalone function, but then in #14270 you changed InferShape to use generic InferAttr again. Was this intentional? Which one should I use then?

junrushao1994:
@ptrendx Sorry it is a mistake. I am actually working on two projects, dynamic shape and numpy operators. In dynamic shape, I should make infer shape pass standalone, but in numpy operator everything is just fine. I think these two projects interfere a bit.

ptrendx:
Ok, so I should move it back to InferShapeAttr and make my changes there then?

junrushao1994:
@ptrendx Yes, please. I am so sorry for the inconvenience.

ptrendx:
@junrushao1994 @reminisce I think this PR is good to go. Do you have any other comments?

PR #14154, opened 2019-02-14
Improving the run time of the tests which use assert_almost_equal OR check_numeric_gradient functions

drivanov:
## Description ##
- The analog of `numpy.allclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False)` is implemented as a MxNet operator:
```
    mx.nd.contrib.allclose(a, b, rtol, atol, equal_nan)
```
- For now, besides the unit test `test_allclose_function()`, this method is used only in
```
def assert_almost_equal(a, b, rtol=None, atol=None, names=('a', 'b'), equal_nan=False)
```
where parameter **a** or/and **b** could be defined also as  `mx.nd.array`(s):
```
    Parameters
    ----------
    a : np.ndarray or mx.nd.array
    b : np.ndarray or mx.nd.array
```
- When calling `assert_almost_equal`, no more `asnumpy()` conversion needed. It will be done automatically (in `assert_almost_equal`), if 
     (a) **a** or **b** has no attribute "context" OR 
     (b) these attributes are different.

- The elimination of `asnumpy()` conversions and the use of `mx.nd.contrib.allclose(...)` for `mx.nd.array`'s allows to achieve 5-7x speedup for GPU tests that use long arrays.

- The MxNet operator `mx.nd.contrib.approx_gradient(...)`, for calculation of **one** OR **all** coordinates of the numerical approximation of the gradient vector was implemented. This operator is called in the `numeric_grad(...)`, which is called by widely used `check_numeric_gradient(...)`
 
- New parameter  `use_batch` which was added to
```
    def check_numeric_gradient(... use_batch=False):
```
allows to calculate the approximation of gradient in a _batch mode_. 
- Unfortunately our current implementation of _batch mode_ **DOES NOT** allow to use it for all operations. For instance, it is impossible to use it   
    (a) for operation, which are similar to matrix multiplications;  
    (b) when the shape of the output of the operation is NOT the same as the shapes of its input parameters;   
    (c) some other cases;   

Currently we are using the _batch mode_ **only** when we call
```
def test_elemwise_binary_ops():
        . . .
        if skip_gradient_check is not True:
            check_numeric_gradient(test, location,
                                   grad_stype_dict=grad_stypes, use_batch=True)
```
- For _elemwise_ operations the _batch mode_ used in `check_numeric_gradient(...)` gives significant (3.5-5x on CPU and 6-10x on GPU) speedup. 

- Even when we cannot use  _batch mode_, we still see the run time improvement associated with the usage of new MxNet operators:
```
     mx.nd.contrib.approx_gradient(...)
     mx.nd.contrib.allclose(...)
```
There are two causes of this improvement:
    (a) now the number of `asnumpy()` conversions, used inside `check_numeric_gradient(...)`, is significantly less than before;
    (b) more calculations which are made on GPUs.  

- In general, we see 1.58x improvement in runtime for the test suite `L0_self_test`:
```
Now:      Ran 520 tests in  994.935s
Before:   Ran 520 tests in 1574.341s 
```

### Essentials ###

## Comments ##
Perhaps, it makes sense:
- to make possible to use (OR just start to use)  _batch mode_ for other operations;
- eliminate `asnumpy()` conversions used when `assert_almost_equal` function is called.

Unfortunately, it's not an easy task, because `assert_almost_equal` is used in many different places. But, it should make the Python code cleaner and some tests will run much faster, especially, those, where both parameters **a** and **b** are the  `mx.nd.array`'s


ankkhedia:
@drivanov Thanks for your contribution. Could you please look into CI failures?

@mxnet-label-bot add [pr-awaiting-testing, test]

drivanov:
I think that fixed previously found problems. 
The problem in  `test_operator_gpu.py:test_laop` was NOT reproducible due to nondeterministic behavior we see in that test. One of the cause of the non-determinism was eliminated. I am not sure that it was the last one, but at least, this test should become more predictable.

drivanov:
Started to use improved version of assert_almost_equal

DickJC123:
From my comparison of the runtime of the centos-gpu CI for this PR vs. others, it appears that this saves over 10 minutes!  I'd think that would be a huge motivation to get this incorporated for the sake of developers and to the lower costs of maintaining the CI system.  @marcoabreu @larroy @szha

szha:
Thanks and great work! Happy to take a look after test passes.

src/operator/contrib/allclose_op-inl.h
```diff
@@ -0,0 +1,138 @@
...
+
+using namespace mshadow_op::isnan_typed;
+
+template<int req>
+struct allclose_forward {
+  template<typename DType>
+  MSHADOW_XINLINE static void Map(int i, int *out_data, const DType* in_a, const DType* in_b,
+                                  const float rtol, const float atol, bool equal_nan) {
+      const DType a = in_a[i], b = in_b[i];
+      const bool val = (IsNan(a) || IsNan(b))? equal_nan :
```
larroy: Could we make this more readable? What's wrong with if-else? this is an abuse of the ternary operator in my opinion.

src/operator/contrib/allclose_op-inl.h
```diff
@@ -0,0 +1,138 @@
...
+  const TBlob& in_array1 = inputs[1];
+
+  const AllCloseParam& param = nnvm::get<AllCloseParam>(attrs.parsed);
+
+  auto s = ctx.get_stream<xpu>();
+  const int num_items = in_array0.Size();
+
+  // Get length of the additional memory (which is used only by DeviceReduce::Min(...) on gpu)
+  const size_t extraStorageBytes = GetAdditionalMemorySize<xpu>(num_items);
+  const size_t workspace_total_bytes_ = num_items * sizeof(float) + extraStorageBytes;
```
larroy: why hardcoded float? Can't we use type_flag? will it work if the type of the tensor is not float?

drivanov: Actually, it was a good catch!

We don't need to use `type_flag` or even `float` here. The correct fix would be
```
  const size_t workspace_total_bytes_ = num_items * sizeof(int) + extraStorageBytes;
```

In my initial implementation `out_data`, which is used in 
```
  MSHADOW_XINLINE static void Map(int i, int *out_data, const DType* in_a, const DType* in_b,
```
was a pointer to `float`. But when I was implementing the possibility for  `in_a, in_b` also be **not-a-numbers**, I 've changed the type of pointer to `int *out_data`.

Perhaps, we could use even `char *out_data`, but I am not sure if 
```
  cub::DeviceReduce::Min(workspaceMemory + num_items, extraStorageBytes,
                         workspaceMemory, outPntr, num_items);
```
would allow for `workspaceMemory` be a `char *`

wkcn:
Thanks for your contribution!

Why are there so much changed files in this PR?

drivanov:
> Thanks for your contribution!
> 
> Why are there so much changed files in this PR?

1. I added 6 C++ files, with the implementations of 2 new MxNet operators 
2. I changed 12-13 Python files where these operators were used (directly or indirectly). Most changes here are related to the removal of `asnumpy()` conversions.

wkcn:
Is it available to rebase this PR? It is too many changed files showed in Github to review.
Thanks!

drivanov:
To simplify the merge I will close current PR and will create two separate PRs for allclose and approx_gradient.

Issue #13468, opened 2018-11-29
Running the mxnet scala jar in the interpreter throws an error


aaronmarkham:
## Description
If you try following the docs for Scala --> NDArray, it uses the interpreter. The latest jar doesn't work in the interpreter. 

You can recreate this by running:
```
scala -classpath ../scala-package/assembly/linux-x86_64-cpu/target/mxnet-full_2.11-linux-x86_64-cpu-1.4.0-SNAPSHOT.jar 
```

Then paste in one of the examples:
```
import org.apache.mxnet._
// all-zero array of dimension 100x50
val a = NDArray.zeros(100, 50)
// all-one array of dimension 256x32x128x1
val b = NDArray.ones(256, 32, 128, 1)
// initialize array with contents, you can specify dimensions of array using Shape parameter while creating array.
val c = NDArray.array(Array(1, 2, 3, 4, 5, 6), shape = Shape(2, 3))
```
## Errors
First one you see:
```
scala> import org.apache.mxnet._
import org.apache.mxnet._

scala> val x = NDArray.zeros(100, 200)
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
java.lang.NoSuchMethodError: scala.Predef$.byteArrayOps([B)Lscala/collection/mutable/ArrayOps;
  at org.apache.mxnet.NDArrayInternal$$anonfun$12.apply(NDArray.scala:1189)
```
Then you also get:
```
scala> val a = NDArray.zeros(100, 50)
java.lang.NoSuchMethodError: scala.Predef$.byteArrayOps([B)Lscala/collection/mutable/ArrayOps;
```

aaronmarkham:
@mxnet-label-bot Add [Scala]

ddavydenko:
@mxnet-label-bot add [Bug]

aaronmarkham:
Same errors...
Also this one:
```
scala> val c = NDArray.array(Array(1, 2, 3, 4, 5, 6), shape = Shape(2, 3))
java.lang.NoSuchMethodError: scala.Predef$.byteArrayOps([B)Lscala/collection/mutable/ArrayOps;
```

larroy:
[...]

PR #14169, opened 2019-02-14
[MXNET-1305] Scala interpreter instructions

zachgk:
## Description ##
Add instructions to the scala_setup page for using the interpreter and link to in the only example featuring the interpreter. It also adds a note that we only support scala 2.11. This should address the problem seen in #13468.

@lanking520 @piyushghai @andrewfayres @aaronmarkham 

### Essentials ###

### DIFF
--- a/docs/install/scala_setup.md
+++ b/docs/install/scala_setup.md
@@ -22,6 +22,7 @@
 The following instructions are provided for macOS and Ubuntu. Windows is not yet available.
 
 **Note:** If you use IntelliJ or a similar IDE, you may want to follow the [MXNet-Scala on IntelliJ tutorial](../tutorials/scala/mxnet_scala_on_intellij.html) instead of these instructions.
+**Note:** Currently, we only support scala 2.11
 
 <hr>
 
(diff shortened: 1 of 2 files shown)


docs/install/scala_setup.md
```diff
@@ -116,6 +117,33 @@ mvn install
...
+## Interpreter
+
+To run the scala interpreter, first download and install scala 2.11.x (run `scala -version` to make sure you have the right version installed.**
+
+### Installing the Interpreter
+
+**Ubuntu***
+
+```
+sudo apt-get install scala
```
aaronmarkham: Seems like this could be more specific. Like maybe:
```
wget www.scala-lang.org/files/archive/scala-2.11.7.deb
sudo dpkg -i scala-2.11.7.deb
```
Or whatever minor version is recommended.

ankkhedia:
@zachgk Thanks for the cotribution!
@mxnet-label-bot add [pr-awaiting-review, Scala]

@lanking520 @piyushghai @andrewfayres 
Could you please help review this PR?

docs/install/scala_setup.md
zachgk: It should be fine. Scala uses a slightly different versioning scheme (https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html) so with version x.y.z, changes to either x or y can be breaking but changes to z are safe. Apt-get installs 2.11.x on both 16.04 and 18.04 so I think it is better to give the simpler command since it is valid.

PR #12189, opened 2018-08-15
Drop dmlc-core commits that break MXNet build

hcho3:
## Description ##
Partially reverts apache/incubator-mxnet#12129 by dropping commits dmlc/dmlc-core#409 and dmlc/dmlc-core#399.

These commits by themselves are fine, but they tigger double free memory error inside CSVIter. In #12139, the CSVIter test crashed on CentOS. The memory error has been inside CSVIter but remained undetected until dmlc/dmlc-core#409 and dmlc/dmlc-core#399 were added. See [address_sanitizer_log.txt](https://github.com/apache/incubator-mxnet/files/2291834/address_sanitizer_log.txt) for a detailed log of double-free error.

Note that the dmlc-core submodule will now track the `mxnet-stable` branch of dmlc/dmlc-core.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] Revert dmlc-core to last working point
- [x] Selectively add dmlc-core commits so that the double-free memory error in CSVIter is not triggered
- [x] Enabled the disabled test test_CSVIter (disabled in #12139)

## Comments ##
- This is a temporary measure. We'll need to come back to CSVIter and address memory error. Address Sanitizer reveals double free issues in CSVIter.

Issue #12139, opened 2018-08-12
Broken test test_io.test_CSVIter


larroy:
## Description

```
test_io.test_CSVIter ... *** Error in `python3.6': double free or corruption (fasttop): 0x0000000002b74390 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x81499)[0x7feff3eec499]
/lib64/libstdc++.so.6(_ZNSs9_M_mutateEmmm+0x1be)[0x7fefdfc56dce]
/lib64/libstdc++.so.6(_ZNSs12_M_leak_hardEv+0x26)[0x7fefdfc56e06]
/work/mxnet/python/mxnet/../../lib/libmxnet.so(_ZN4dmlc4data9CSVParserIjiE10ParseBlockEPKcS4_PNS0_17RowBlockContainerIjiEE+0x2a8)[0x7fefbcbe7a48]
/work/mxnet/python/mxnet/../../lib/libmxnet.so(+0x3386e6d)[0x7fefbcbb4e6d]
/lib64/libgomp.so.1(+0x163c5)[0x7fefdb11f3c5]
/lib64/libpthread.so.0(+0x7e25)[0x7feff4948e25]
/lib64/libc.so.6(clone+0x6d)[0x7feff3f69bad]
======= Memory map: ========
```

http://jenkins.mxnet-ci.amazon-ml.com/blue/rest/organizations/jenkins/pipelines/incubator-mxnet/branches/master/runs/1398/nodes/926/steps/1123/log/?start=0

lanking520:
@mxnet-label-bot please label this as [build, CI]

marcoabreu:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/master/1397/pipeline
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/master/1398/pipeline/
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/master/1399/pipeline/
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/master/1400/pipeline/

haojin2:
Fix is merged, and the related test has been re-enabled now. @sandeep-krishnamurthy @nswamy Can you help with closing the issue?

PR #14173, opened 2019-02-15
MXNet AMP (automatic mixed precision)

ptrendx:
## Description ##

Whis is a Work in Progress PR for AMP (automatic mixed precision) support for MXNet, similar to pyTorch version found in https://github.com/NVIDIA/apex.

This PR relies on multiple other PRs and bug fixes, listed in Comments section.

Dynamic loss scaling part done by @Caenorst (commits were squashed for easier rebasing).

### Essentials ###

### Changes ###
- [x] Auditor that enables/disables operations to be done in FP16 automatically. It is implemented via patching MXNet functions in mxnet.symbol and mxnet.ndarray to insert casts to FP16/FP32 where necessary.
- [x] Operator `amp_cast` and `amp_multicast` that handle casting between FP16/FP32 when necessary and do not change other types. They are optimized to not do anything if the input is already in the proper type.
- [x] Dynamic loss scaling and supporting operators for checking gradients for infs/NaNs and skipping update step if such value is encountered.

## Comments ##
- This PR relies on multiple other PRs/bug fixes:
  - [x] #14153 
  - [x] https://github.com/dmlc/tvm/pull/2572 - once this is done, need to change submodule to point to dmlc/tvm again
  - [x] #14097 
  - [x] https://github.com/dmlc/dmlc-core/issues/503
  - [x] #12139 which was masked by #12189, but which prevents updating to newer dmlc-core
  - [x] #14209 (without this fix SSD from GluonCV does not work)
  - [x] #14325 
  - [x] #14326
  - [x] #14377

FYI @eric-haibin-lin @szha


.gitmodules
```diff
@@ -25,7 +25,7 @@
 	url = https://github.com/dmlc/cub
 [submodule "3rdparty/tvm"]
```
eric-haibin-lin: If we need to upgrade to the latest commit of tvm (which depends on it own version of dmlc-core), it is necessary to upgrade the dmlc-core submodule in mxnet?

ptrendx: a. any nnvm header that is included in MXNet .cc file will use the MXNet's dmlc-core
b. I'm not sure if the make of NNVM is done using its own dmlc-core - it is possible that the env is set up so all components use the same (MXNet's) dmlc-core. Otherwise you could end up with some binary incompatibilities (e.g. if dmlc-core changes some struct and it is passed around between NNVM and MXNet you don't want it to be interpreted differently).

ankkhedia:
@ptrendx Thanks for the contribution!

@mxnet-label-bot  add [pr-work-in-progress ]

python/mxnet/amp/amp.py
```diff
@@ -0,0 +1,248 @@
...
+from . import lists
+from ..gluon import trainer
+from .. import optimizer as opt
+from .loss_scaler import *
+
+def _cast_symbol_NDArray(s, dtype):
+    if isinstance(s, Symbol):
+        return symbol.amp_cast(s, dtype=dtype)
+    elif isinstance(s, NDArray):
+        if s.dtype != dtype and (s.dtype == np.float16 or s.dtype == np.float32):
```
ZhennanQin: Is this check not enough for avoid inserting cast with same dtype? That is, why we need to add identity to cast op?

python/mxnet/amp/amp.py
```diff
@@ -0,0 +1,248 @@
...
+            return ndarray.amp_cast(s, dtype=dtype)
```
ZhennanQin: Why not insert normal cast op instead?

python/mxnet/amp/amp.py
```diff
@@ -0,0 +1,248 @@
...
+        return s
+
+def _wrap_symbol_functions(module):
+    def _ndarray_wrapper(f, target_dtype, cond_arg=None):
+        def _new_fun(*args, **kwargs):
+            if cond_arg is not None:
+                if (cond_arg[0] not in kwargs or
+                        kwargs[cond_arg[0]] not in cond_arg[1]):
+                    return f(*args, **kwargs)
+            new_args = list(map(lambda x: _cast_symbol_NDArray(x, target_dtype), args))
```
ZhennanQin: If an fp16 output is used for multiple fp32 operators, can we cast the fp16 only once?

.gitmodules
szha: tvm now comes with a [notice](https://github.com/dmlc/tvm/blob/master/NOTICE) file, so we'd need to add that to mxnet's notice too per apache license.

python/mxnet/amp/amp.py
ptrendx: For NDArray (so fully imperative) it is enough. But for symbolic there is no type information at this point.

python/mxnet/amp/amp.py
ptrendx: Because we only want to cast between fp16/fp32 and not e.g. int to fp32, since that would break a lot of ops. So the amp_cast operator passes along any non-floating point data.

python/mxnet/amp/amp.py
ptrendx: That would require a graph pass approach instead of simple function substitution.

pengzhao-intel:
@ptrendx @eric-haibin-lin could you share the total picture (or methodology/sw stack) of AMP integration? I'd like to understand in the high level before going to details.

python/mxnet/amp/amp.py
pengzhao-intel: This will degrade the performance a lot so I suggest using the graph pass.

ptrendx: Adding a graph pass may be the next step for performance tuning but it definitely is not necessary for the feature to be useful. I tested all the models from GluonCV and in none of them I saw a need for it (the unnecessary casts typically take <1% of the total time).

pengzhao-intel: Sounds good :)  Could you share your data and the plan as my comments in the below?

ptrendx:
@pengzhao-intel The goal for AMP is to enable users to get the benefits of mixed precision training (so mixed fp16/fp32 dtypes) with as little code changes as possible. NVIDIA had a set of guidelines for a long time how to do mixed precision training, but it turned out that implementing them is too hard for typical users. That is the reason of making AMP (automatic mixed precision) tool that would automatically apply the guidelines to the code to make mixed precision training as easy as possible.

To give an example (I will push the tutorial to this PR soon as well), the typical training loop in Gluon looks like this:
```
net = get_network()
trainer = mx.gluon.Trainer(...)
for data in dataloader:
    with autograd.record(True):
        out = net(data)
        l = loss(out)
        autograd.backward(l)
     trainer.step()
```
With AMP, the same loop would look like this:
```
amp.init()
net = get_network()
trainer = mx.gluon.Trainer(...)
amp.init_trainer(trainer)
for data in dataloader:
    with autograd.record(True):
        out = net(data)
        l = loss(out)
        with amp.scale_loss(l, trainer) as scaled_loss:
            autograd.backward(scaled_loss)
     trainer.step()
```

The methodology of AMP integration is pretty simple:
 - we wrap the functions in `mx.sym` and `mx.ndarray` (as well as the `Symbol` and `NDArray` class members) to insert appropriate casts
 - since we don't have dtype information at this point for the symbolic graph we insert a special `amp_cast` and `amp_multicast` functions there (for cast to fp16/fp32 and casting to widest type respectively) that mark themselves as as invisible as possible (inplace as well as inplaceidentity) in order to be able to just exit early if the dtypes on both ends of the op end up being the same and input shares the memory with the output.
 - we give utilities to scale the loss (by default dynamically, but I will also add ability to do static loss scaling, since dynamic loss scaling is slightly costly)

We tested all (except for the newest pose estimation) of the networks from GluonCV and for all of them we saw between few to over 100% speedups running on V100, with typical speedup being over 50%, depending on what the relative cost of convolution/gemms are compared to the rest of the network.

For nearly all of the networks the procedure was basically applying the 3 lines I showed earlier. The only problem was with detection networks that GluonCV first runs on the CPU to generate anchors and lack of FP16 GEMM on CPU makes it impossible (@pengzhao-intel it would be great if you could close that gap). For those networks I needed to change the script to generate anchors on the GPU and then copy them to the CPU for Gluon data loader to consume.

python/mxnet/amp/lists/symbol.py
```diff
@@ -0,0 +1,175 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# coding: utf-8
+"""Lists of functions whitelisted/blacklisted for automatic mixed precision in symbol API."""
+
+FP16_FUNCS = [
```
eric-haibin-lin: There are operators neither whitelisted nor blacklisted. Can users expect those operators to work with AMP? If a new operator is contributed to MXNet from the community, can we have some kind of general guideline (either documentation here or external resources) on when an operator should be added to one of these lists?

python/mxnet/amp/loss_scaler.py
```diff
@@ -0,0 +1,77 @@
...
+    Properties
+    ----------
+    loss_scale : float
+        The current loss scale
+    """
+    def __init__(self):
+        self._loss_scale = 2.**16
+        self._next_loss_scale = self._loss_scale
+        self._max_loss_scale = 2.**24
+        self._scale_seq_len = 2000
```
eric-haibin-lin: Is there any time an user wants to tweak seq_len for specific models? I am just curious because the fairseq impl has those options

python/mxnet/amp/loss_scaler.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def wait_and_update(self):
+        """Wait for the results of overflow checking and update the loss scale."""
+        if self._wait_for_outputs:
+            self._has_overflow = not bool(self.output.asnumpy())
+            self._loss_scale = self._next_loss_scale
+            if self._has_overflow:
+                self._next_loss_scale = self._loss_scale / 2.
+                self._unskipped = 0
+                print("_HAS_OVERFLOW")
```
eric-haibin-lin: Maybe add a verbosity/logger option to amp.init instead of always using print. Otherwise if users use `logging.info` for other parts of the program, `print` and `logging.info` may produce interleaving logs in the terminal

src/operator/contrib/all_finite.cc
```diff
@@ -0,0 +1,166 @@
...
+                         const std::vector<TBlob> &inputs,
+                         const std::vector<OpReqType> &req,
+                         const std::vector<TBlob> &outputs) {
+  using namespace mxnet_op;
+  Stream<cpu>* s = ctx.get_stream<cpu>();
+  const AllFiniteParam& op_param = nnvm::get<AllFiniteParam>(attrs.parsed);
+  Tensor<cpu, 2, float> out = outputs[0].FlatTo2D<cpu, float>(s);
+  if (op_param.init_output)
+    out = 1.;
+  MSHADOW_REAL_TYPE_SWITCH(inputs[0].type_flag_, DType, {
```
eric-haibin-lin: nit: indentation

python/mxnet/amp/amp.py
```diff
@@ -0,0 +1,262 @@
...
+            args = tuple(new_args)
+            kwargs = {k: _cast_symbol_NDArray(v, target_dtype) for k, v in kwargs.items()}
+            return f(*args, **kwargs)
+        _new_fun.__name__ = f.__name__
+        _new_fun.__module__ = f.__module__
+        _new_fun.__doc__ = f.__doc__
+        return _new_fun
+
+    def _symbol_wrapper(f, target_dtype, cond_arg=None):
```
eric-haibin-lin: What is the expected workflow for deploying models trained with AMP? 

FP16 is not well supported on CPU and I assume we do not want to include amp cast/multi-cast in the exported symbol.json file. For users who want to do deployment on GPUs with fp16 support with c++, exporting the symbols with casts is fine. For users who want to deploy models on CPU, are they expected to export only when `amp.init()` is not called?

pengzhao-intel:
Thanks @ptrendx @eric-haibin-lin this is no FP16 GEMM from Intel MKL/MKL-DNN till now so I think we have to fall back to FP32 GEMM or simulated FP16 GEMM (maybe from `mshadow` or other BLAS library).

The current design is good for APEX.  I am thinking about a general solution for mixed precision training, such as INT8, FP16, BF16, FP32. The different HW supports for the different data type, like INT8 in CPU and small GPU chip and BF16 on CPU and FP16 on GPU. So we need more flexibility for data type change or other library integration.

We are going to support BF16 and the ideal situation is the user can switch FP16 and BF16 based on the devices transparently.

We are looking into the changes and will give back more details later.

python/mxnet/amp/lists/symbol.py
ptrendx: Operators that are neither white- nor blacklisted are considered to be neutral to dtype and so they are just handled by normal type inference in the framework (if the input was fp16 then they will be in fp16, if the input was fp32 then it will be fp32).

python/mxnet/amp/loss_scaler.py
ptrendx: I don't think there is any guidance on what it should be besides "it should not happen too often", so that if it goes up and then immediately hits overflow, not many of the batches are lost.

python/mxnet/amp/loss_scaler.py
ptrendx: Sure, that was just for debug.

python/mxnet/amp/amp.py
ptrendx: All the weights are fp32 and the model itself does not know anything about fp16, so user can import and export fp32 models.

ptrendx:
> Sharing for sharing the speedup results. Are those benchmarked with multiple GPUs with kvstore?

There are multiple potential bottlenecks that can be hit during the multiGPU run and for AMP benchmarks I used single GPU. The problems I hit when doing multi GPU tests, which limited speedup achieved:
 - IO bottleneck (e.g. SSD - with AMP with synthetic data I was getting over 300 imgs/s on 4 GPUs, with real data using Gluon dataloader I was gettting a little over 200)
 - bottleneck of pushing ops to the engine form a single Python thread (FasterRCNN and MaskRCNN were the biggest offenders here, it should not be a problem once we go the Horovod route)
 - SyncBN usage - some models in GluonCV use SyncBN which has very slow implementation (a bunch of mshadow kernels) - in some cases (e.g. FCN) changing from SyncBN to no SyncBN was giving 50% speedup. Also I believe current SyncBN implementation is not actually working with multiprocess launch (Horovod, dist kvstore).

pengzhao-intel:
> We tested all (except for the newest pose estimation) of the networks from GluonCV and for all of them we saw between few to over 100% speedups running on V100, with typical speedup being over 50%, depending on what the relative cost of convolution/gemms are compared to the rest of the network.

@ptrendx  TensorCore with FP16 provides the 8X speedup against FP32.  
What is the gap between 2X in MXNet with 8X from TensorCore?

python/mxnet/amp/amp.py
ptrendx: Hmmm, actually I would need to double check it. If you just save parameters from the Gluon model then they all will be in fp32, but I'm not sure about saving the entire symbol from the network built while AMP was active.

ptrendx: The workaround of saving params only and then building the network again without AMP with those parameters and exporting symbol would work, but it is definitely not ideal.

ptrendx:
@pengzhao-intel 8x is possible to get for huge GEMMs, but DL training has a lot of other operations that limit this speedup, most notably all of memory bandwidth limited operations like batchnorm, relu, add etc. Those operations also speed up from using fp16 but only up to 2x by making the data 2x smaller.
What is more is NCHW is not ideal format for TensorCores since they need the data in NHWC format, which means that all convolutions are preceded with NCHW to NHWC transpose kernel which limits speedup as well. I would like to introduce automatic conversion to NHWC as a graph pass as part of AMP later, but this requires more consideration in order to guarantee that the exported parameters are what user would expect.

src/operator/contrib/all_finite.cc
ptrendx: I don't see any problem, could you be more specific here?

python/mxnet/amp/loss_scaler.py
ptrendx: Done.

ptrendx:
@marcoabreu Why is clang-tidy complaining about those lines - I did not change them and on master it does not produce errors:
```
clang-tidy-6.0 -header-filter=^/work/build/.* -p=/work/build /work/mxnet/src/operator/tensor/elemwise_unary_op_basic.cc

/work/mxnet/src/operator/tensor/elemwise_unary_op_basic.cc:363:11: error: use emplace_back instead of push_back [modernize-use-emplace,-warnings-as-errors]

      lhs.push_back(nnvm::NodeEntry{ng, 0, 0});

          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~      ~~

          emplace_back(

/work/mxnet/src/operator/tensor/elemwise_unary_op_basic.cc:502:11: error: use emplace_back instead of push_back [modernize-use-emplace,-warnings-as-errors]

      lhs.push_back(nnvm::NodeEntry{ng, 0, 0});

          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~      ~~

          emplace_back(


```

docs/tutorials/amp/amp_tutorial.md
```diff
@@ -0,0 +1,266 @@
...
+For some models that may be enough to start training in mixed precision, but the full FP16 recipe recommends using dynamic loss scaling to guard against over- and underflows of FP16 values. Therefore, as a next step, we create a trainer and initialize it with support for AMP's dynamic loss scaling. Currently, support for dynamic loss scaling is limited to trainers created with `update_on_kvstore=False` option, and so we add it to our trainer initialization.
+
+
+```python
+trainer = gluon.Trainer(
+    net.collect_params(), 'sgd',
+    {'learning_rate': lr, 'wd': wd, 'momentum': momentum},
+    update_on_kvstore=False)
+
+amp.init_trainer(trainer)
```
ZhennanQin: As you said, FP16 is an option of lower precision, thus AMP should be able to extend to support other lower precision, like BF16. From API level, I think it's well designed and a small change is enough to cover other low precision support:
```
amp.init_trainer(trainer, target_dtype)
```

python/mxnet/amp/amp.py
```diff
@@ -0,0 +1,262 @@
...
+from ..gluon import trainer
+from .. import optimizer as opt
+from .loss_scaler import LossScaler
+
+def _cast_symbol_NDArray(s, dtype):
+    if isinstance(s, Symbol):
+        return symbol.amp_cast(s, dtype=dtype)
+    elif isinstance(s, NDArray):
+        if (s.dtype != dtype and
+                (s.dtype == np.float16 or s.dtype == np.float32) and
```
ZhennanQin: How about make a list of all float types supported by AMP, and then we can check with the list:
```
float_types = (np.float16, np.float32) # we can extend this list if we have other float types support
if s.dtype in float_types:
```

python/mxnet/amp/amp.py
```diff
@@ -0,0 +1,262 @@
...
+                if isinstance(arg, (Symbol, NDArray)):
+                    symbols.append((args, i, arg))
+                    is_symbol = is_symbol or isinstance(arg, Symbol)
+            for k, arg in kwargs.items():
+                if isinstance(arg, (Symbol, NDArray)):
+                    symbols.append((kwargs, k, arg))
+                    is_symbol = is_symbol or isinstance(arg, Symbol)
+            if not is_symbol:
+                # NDArray case
+                widest_type = np.float16
```
ZhennanQin: Considering to support other low precision types, it's better to use `target_dtype` from `init_trainer` instead of hard code `np.float16`

python/mxnet/amp/lists/symbol.py
ZhennanQin: New fp32 operators are introduced into MXNet frequently, and the author may not know `FP32_FUNCS` exists and didn't add it into this list, then the new added operators would be neither white- nor blacklisted. When they're used with AMP, this behavior may try to run it in low precision mode, causing unexpected result. So maybe we can create `FP16_FUNCS` and `FP16_FP32_FUNCS` here, making non-listed operators are FP32 only.

src/operator/tensor/amp_cast.cc
```diff
@@ -0,0 +1,150 @@
...
+#include "./amp_cast.h"
+
+namespace mxnet {
+namespace op {
+
+DMLC_REGISTER_PARAMETER(AMPCastParam);
+DMLC_REGISTER_PARAMETER(AMPMultiCastParam);
+
+NNVM_REGISTER_OP(amp_cast)
+.describe(R"code(Cast function between FP16/FP32 used by AMP.
```
ZhennanQin: How about change FP16 to low precision float?

src/operator/tensor/amp_cast.h
```diff
@@ -0,0 +1,165 @@
...
+    TYPE_ASSIGN_CHECK(*out_attrs, 0, (*in_attrs)[0]);
+  }
+  return (*in_attrs)[0] != -1;
+}
+
+inline bool AMPMultiCastType(const nnvm::NodeAttrs& attrs,
+                        std::vector<int> *in_attrs,
+                        std::vector<int> *out_attrs) {
+  using mshadow::kFloat32;
+  using mshadow::kFloat16;
```
ZhennanQin: Not hard code. It's better to be detected from in_dtypes.

src/operator/tensor/amp_cast.h
```diff
@@ -0,0 +1,165 @@
...
+    DMLC_DECLARE_FIELD(num_outputs)
+    .describe("Number of input/output pairs to be casted to the widest type.");
+  }
+};
+
+inline bool AMPCastType(const nnvm::NodeAttrs& attrs,
+                        std::vector<int> *in_attrs,
+                        std::vector<int> *out_attrs) {
+  using mshadow::kFloat32;
+  using mshadow::kFloat16;
```
ZhennanQin: Not hard code. It's better to be detected from param.dtype.

python/mxnet/amp/lists/symbol.py
```diff
@@ -0,0 +1,177 @@
...
+    'Convolution',
+    'Deconvolution',
+    'FullyConnected',
+    'RNN',
+    ]
+
+FP32_FUNCS = [
```
anirudh2290: Can we add  support in the API for updating the lists or using custom lists for the user ? 
User may want to experiment with certain ops being run in FP16 vs FP32 to see accuracy changes.

ptrendx: Makes sense, I will add this today.

ptrendx: Done.

python/mxnet/amp/amp.py
eric-haibin-lin: That's right. Requiring exporting without AMP does not sound ideal. The best way is to do this kind of graph rewriting in the backend and export the original one

eric-haibin-lin:
dmlc-core has now been fixed and updated :)

python/mxnet/amp/amp.py
```diff
@@ -0,0 +1,303 @@
...
+def scale_loss(loss, optimizer_or_trainer):
+    assert optimizer_or_trainer._amp_loss_scaler is not None, \
+        'Loss scaler is not initialized, did you forget to call amp.init_trainer()?'
+    optimizer_or_trainer._scale = 1. / optimizer_or_trainer._amp_loss_scaler.loss_scale
+    if isinstance(loss, (list, tuple)):
+        yield [l * optimizer_or_trainer._amp_loss_scaler.loss_scale for l in loss]
+    else:
+        yield optimizer_or_trainer._amp_loss_scaler.loss_scale * loss
+
+def init(fp16_list=None, conditional_fp32_list=None, fp32_list=None):
```
anirudh2290: Thanks for adding this support !
can we change the interface to look something like this : 
```
def init(target_dtype="float16", original_dtype="float32", target_precision_ops=None, original_precision_ops=None, conditional_original_precision_ops=None)
```

This way we can add bf16 support for target_dtype and also add support for other high precision dtypes. WDYT?

Also is there any reason to not include the widest_type_list ?

TaoLv:
Looks like it's a brand new feature and I see many files are changed for that. Is it possible for you to post a RFC or a proposal for this PR? @ptrendx

python/mxnet/amp/amp.py
ptrendx: Added a graph pass that eliminates amp casts when saving the model.

python/mxnet/amp/amp.py
ptrendx: Ok.

python/mxnet/amp/amp.py
ptrendx: Ok.

src/operator/tensor/amp_cast.cc
ptrendx: Ok.

src/operator/tensor/amp_cast.h
ptrendx: That is not really possible though since you need to know what is the relationship between those types (which one is wider).

src/operator/tensor/amp_cast.h
ptrendx: This is again not really possible to avoid specifying the types here explicitly - you need to be able to figure out whether the type is floating point or not.

docs/tutorials/amp/amp_tutorial.md
ptrendx: I added the target_dtype to amp.init() instead, which does the wrapping..

ptrendx:
@TaoLv I made a design proposal for it [here](https://docs.google.com/document/d/1sQzMoPEwux0WXSWirY07us1POD_6Y8pLYq--b9Fvd1o/edit?usp=sharing)

python/mxnet/contrib/amp/amp.py
```diff
@@ -0,0 +1,342 @@
...
+            if cond_arg is not None:
+                if (cond_arg[0] not in kwargs or
+                        kwargs[cond_arg[0]] not in cond_arg[1]):
+                    return f(*args, **kwargs)
+            sym = f(*args, **kwargs)
+            inputs = sym.get_children()
+            aux = sym.list_auxiliary_states()
+            inputs = list(map(lambda x: _cast_symbol_NDArray(x, target_dtype)
+                              if x.name not in aux else x, inputs))
+            atomic_sym = sym._gen_atomic_symbol()
```
anirudh2290: why is this needed ?

python/mxnet/contrib/amp/amp.py
```diff
@@ -0,0 +1,342 @@
...
+    fp32_ops : list of string
+        Override the list of functions casted to FP32. Entries in this list
+        are names of the functions casted to FP32.
+    """
+    global _amp_initialized
+    global _loss_scaler
+    if not _amp_initialized:
+        _amp_initialized = True
+        logging.info("Using AMP")
+        target_dtype = np.dtype(target_dtype)
```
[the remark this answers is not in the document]
ptrendx: Ok

python/mxnet/contrib/amp/amp.py
ptrendx: Basically this allows us to cast inputs to the op that were not specified by the user (e.g. you make a convolution via symbolic API, you don't need to pass weights and biases to it, MXNet will generate them). So what we do here is create a symbol using the original function (which will create all the other children), then take those children and cast them. MXNet (nor NNVM) does not allow, however, to manipulate a symbol children after it was created. So what we do is we create a new, atomic symbol (atomic means it does not have any children set yet, but has all the params the same as the symbol used to create it), and populate that symbol with casted inputs.

python/mxnet/contrib/amp/lists/symbol.py
```diff
@@ -0,0 +1,599 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# coding: utf-8
+"""Lists of functions whitelisted/blacklisted for automatic mixed precision in symbol API."""
+
+# Functions that should be cast to lower precision
+TARGET_DTYPE_FUNCS = [
```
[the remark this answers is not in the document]
ptrendx: In the previous comment you said "change the lists to have target_dtype_list, target_dtype_fp32_list." and that is why I did that change. I can revert it - either way is fine by me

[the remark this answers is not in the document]
anirudh2290: Apologies for not being clear. I meant <target_dtype>_fp32_list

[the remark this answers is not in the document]
ptrendx: ok, will change

python/mxnet/contrib/amp/amp.py
anirudh2290: when pretrained models are loaded under amp.init will it silently fail ?

ptrendx: Pretrained as in symbol? That will not do anything (it needs your PR).
Pretrained as in load_parameters? That will work as expected (I was using it for all my experiments with GluonCV where e.g. SSD uses pretrained RN50 backbone).

python/mxnet/model.py
```diff
@@ -391,7 +391,7 @@ def _train_multi_device(symbol, ctx, arg_names, param_names, aux_names,
     # end of all epochs
 
 
-def save_checkpoint(prefix, epoch, symbol, arg_params, aux_params):
+def save_checkpoint(prefix, epoch, symbol, arg_params, aux_params, remove_amp_cast=True):
```
eric-haibin-lin: Please document remove_amp_cast

python/mxnet/model.py
```diff
@@ -905,7 +905,7 @@ def fit(self, X, y=None, eval_data=None, eval_metric='acc',
                             sym_gen=self.sym_gen)
 
 
-    def save(self, prefix, epoch=None):
+    def save(self, prefix, epoch=None, remove_amp_cast=True):
```
eric-haibin-lin: Please document remove_amp_cast

python/mxnet/symbol/symbol.py
```diff
@@ -1275,7 +1275,7 @@ def debug_str(self):
             self.handle, ctypes.byref(debug_str)))
         return py_str(debug_str.value)
 
-    def save(self, fname):
+    def save(self, fname, remove_amp_cast=True):
```
eric-haibin-lin: Please document remove_amp_cast

python/mxnet/model.py
ptrendx: Done

python/mxnet/model.py
ptrendx: Done

python/mxnet/symbol/symbol.py
ptrendx: Done

Issue #13265, opened 2018-11-14
MXNet C++Interface reasoning leads to CPU memory leak


chenyujing1234:
I am using the c++ interface of mxnet to carry out the reasoning of mtcnn algorithm for face detection.
Environment: Ubuntu16.04.1 + GPU cuda9.0 + MXNet1.3.0
I use the interface: MXPredCreate MXPredReshape MXPredForward MXPredFree...
When I used a lot of pictures for stress testing (running for a long time), I found that my process occupied more and more CPU memory, and eventually it occupied all the memory.
My process was forced to kill by the system.
Later, I wrote the program To Test:
1) If only do MXPredCreate, then MXPredFree, and MXPredCreate gives different width and height.
After 100 thousand cycles, memory will increase by 1G
2) If you do MXPredCreate, and then MXPredReshape to change the width and height, so that the continuous cycle, found that memory leaks very quickly, less than half an hour to leak 4G.
Do you want to invite MXNet's main developers to find this problem? Please track it down for me. It has been bothering me for almost a week.

zachgk:
@mxnet-label-bot add [C++, Memory, Bug, Pending Requester Info]

Thanks for raising the issue @chenyujing1234. Can you provide a bit more info so we can better help?

## Environment info (Required)

```
What to do:
1. Download the diagnosis script from https://raw.githubusercontent.com/apache/incubator-mxnet/master/tools/diagnose.py
2. Run the script using `python diagnose.py` and paste its output here.

```

## Build info (If built from source)

Compiler (gcc/clang/mingw/visual studio):

MXNet commit hash:
(Paste the output of `git rev-parse HEAD` here.)

Build config:
(Paste the content of config.mk, or the build command.)

## Minimum reproducible example
Can you share the code for the program that you used to test the memory leak?

leleamol:
@mxnet-label-bot update [Bug, C API, Memory, Pending Requester Info]

chenyujing1234:
You can get my test code from this address. TKS
https://pan.baidu.com/s/19Wtd_Cf1BGF-2MS3le1mYA

chenyujing1234:
> @mxnet-label-bot update [Bug, C API, Memory, Pending Requester Info]
You can get my test code from this address. TKS
https://pan.baidu.com/s/19Wtd_Cf1BGF-2MS3le1mYA

chenyujing1234:
[...]

PR #312, opened 2015-10-18
Fix destructor ordering of ObjectPool singleton objects

pluskid:
fix #309

PR #14176, opened 2019-02-15
uncomment memory pool free

arcadiaphy:
## Description ##
When detecting memory leaks on c++ inference code using ASAN, I have found that almost all of the leaks in ASAN reports come from unreleased memory in object pool. The [free code](https://github.com/apache/incubator-mxnet/blob/master/src/common/object_pool.h#L135) is deliberately commented out to avoid program crash from accessing too early destructed objects in global singletons.

The main problem of object pool is fixed in #312, maybe there are still some underlying issues. I have re-added the free operation, and experienced no problems in several weeks' usage.

I think the correct way is to just let problems happen, then we can fix them to approach leak-free codes.

Since issues like #13265 are reported, some ASAN tests are suggested to be added in CI to assure absolutely no leaks in c++ interface.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/cpp-package/example/alexnet.cpp
+++ b/cpp-package/example/alexnet.cpp
@@ -319,6 +319,7 @@ int main(int argc, char const *argv[]) {
   }
   /*don't foget to release the executor*/
   delete exec;
+  delete opt;
   MXNotifyShutdown();
   return 0;
 }
--- a/cpp-package/example/charRNN.cpp
+++ b/cpp-package/example/charRNN.cpp
@@ -500,6 +500,9 @@ void train(const std::string file, int batch_size, int max_epoch, int start_epoc
     std::string filepath = prefix + "-" + std::to_string(epoch) + ".params";
     SaveCheckpoint(filepath, RNN, exe);
   }
+
+  delete exe;
+  delete opt;
 }
 
 /*The original example, rnn_cell_demo.py, uses default Xavier as initalizer, which relies on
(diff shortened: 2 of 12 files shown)


marcoabreu:
Hi,

thanks a lot for your input. We already have an ASAN setup ready in CI, but it is set as non blocking:

https://github.com/apache/incubator-mxnet/blob/master/ci/docker/runtime_functions.sh#L409
https://github.com/apache/incubator-mxnet/blob/master/ci/docker/runtime_functions.sh#L1021

I'd recommend to enable the blocking for memory leaks and then incrementally work in your PR to address these leaks until CI is green. What do you think?

marcoabreu:
@KellenSunderland fyi

KellenSunderland:
Yep, agree with the assessment here.  If we can ensure properly shutdown we can re-enable these frees and remove these ASAN reports.  I also wonder if we can't use smart pointers to manage state of the pools so that we don't have to be so careful when shutting down threadpools / engine threads.

ankkhedia:
@arcadiaphy Thanks for the contribution!

@mxnet-label-bot  add [pr-awaiting-review, C API]

ankkhedia:
@leleamol Ping for review!

arcadiaphy:
@marcoabreu Yep, let me have a look on the ASAN tests.

arcadiaphy:
@marcoabreu @KellenSunderland I have played with ASAN test of mlp_cpu example, there are roughly two kinds of memory leaks:

1. When using threaded engine, the memory release is depended upon workers in engine, and thus asynchronous. When program exits, the destruction of engine will not wait for all operation to finish, leading to leaked memory. This kind of leak is not very serious since it only happens on exit, and is more like ungraceful program exit.
2. True memory leak that happens elsewhere.

To fully address the 1st leak is hard, many memories are hidden in global/thread_local singleton, so it's related to enforce a correct destruction order on static variables (the engine is also a static variable) and make engine to wait for unfinished operations. Actually, I don't find any mechanisms concerning static variable destruction order in mxnet, so I wonder why there is no problem in usage.

A easy workaround is to use naive engine in ASAN tests, if leaks still exists, then it's definitely the 2nd leak and we must fix them.

PR #14192, opened 2019-02-18
[MXNET-1324] Add NaiveRunGraph to imperative utils

junrushao1994:
## Description ##
This PR introduces another way of executing a graph, `NaiveRunGraph`, which runs the operators one by another. This is part of effort of enabling dynamic shapes inside a CahcedOp. Since the output shape of operators coundn't be determined ahead of time, we have to run these operators one at a time.

### Essentials ###

### Changes ###
- [x] Add a `NaiveForward` pass.

## Comments ##
- Since it is only used in dynamic shape, this won't cause performance regression.


junrushao1994:
@szha @zheng-da Could you help review? Thanks a lot!

roywei:
@mxnet-label-bot add[Backend, pr-awaiting-review]

junrushao1994:
@szha Hey I added a unit test for this PR :-)

src/imperative/cached_op.cc
```diff
@@ -262,6 +262,32 @@ std::vector<nnvm::NodeEntry> CachedOp::Gradient(
...
+  nnvm::Graph& g = state.info.fwd_graph;
+  ShapeVector shape_inputs;
+  shape_inputs.reserve(inputs.size());
+  for (auto input : inputs) {
+    shape_inputs.emplace_back(input->shape());
+  }
+  bool contain_dynamic_shape = false;
+  CheckAndInferShape(&g, std::move(shape_inputs), true,
+                     {0, 0}, {0, 0},
+                     &contain_dynamic_shape);
```
zheng-da: please add comments about why we can use shape inference to detect dynamic shape operators.

junrushao1994: Sure!

vandanavk:
@zheng-da is this PR good to go?

junrushao1994:
Rebased to the master because of the dramatic change to TShape.

junrushao1994:
@szha @zheng-da Hey could you do another round of review since the CI is going to pass?

src/imperative/cached_op.cc
```diff
@@ -839,6 +867,70 @@ OpStatePtr CachedOp::DynamicForward(
...
+  // build the indexed graph
+  nnvm::Graph& g = runtime.info.fwd_graph;
+  const auto& idx = g.indexed_graph();
+  const size_t num_inputs = idx.input_nodes().size();
+  const size_t num_entries = idx.num_node_entries();
+  std::vector<uint32_t> ref_count = g.GetAttr<std::vector<uint32_t> >(
+    recording ? "full_ref_count" : "forward_ref_count");
+  // construct `arrays`
+  runtime.buff.resize(num_entries);
+  std::vector<NDArray*> arrays;
```
szha: would it make sense to refactor the array construction into a function? the hope is to keep the convention visible (i.e. buf, inputs, outputs)

junrushao1994: Sure!

junrushao1994:
@szha @zheng-da hey could you guys do another round of review?

src/imperative/imperative_utils.cc
```diff
@@ -22,6 +22,106 @@
...
+  std::vector<NDArray*> ndinputs;
+  ndinputs.reserve(num_inputs);
+  for (const auto& j : node.inputs) {
+    size_t eid = idx.entry_id(j);
+    ndinputs.emplace_back(arrays[eid]);
+  }
+  return ndinputs;
+}
+
+inline std::vector<NDArray*> NodeOutputs(const nnvm::IndexedGraph& idx,
```
larroy: Why inline?

src/imperative/imperative_utils.cc
```diff
@@ -22,6 +22,106 @@
...
+  std::vector<OpReqType> req;
+  req.reserve(num_outputs);
+  for (size_t j = 0; j < num_outputs; ++j) {
+    size_t eid = idx.entry_id(node_idx, j);
+    req.push_back(array_reqs[eid]);
+  }
+  return req;
+}
+
+inline void InvokeOperator(const nnvm::IndexedGraph& idx,
```
larroy: Why inline?

junrushao1994:
@larroy thanks for asking! So could you elaborate what’s the problem of inline?

larroy:
You are inlining a big function, this will cause code bloat at no benefit. Let the compiler do the inline for you in the majority of the cases. I removed the inline anyway so it's no longer in master. 

See also: https://github.com/apache/incubator-mxnet/issues/15182

src/imperative/imperative_utils.cc
larroy: Also this function should be static or in the anonymous namespace as it's not used outside this unit.

larroy:
Since this commit might cause a regression, I would kindly request if you can elaborate a bit about the differences between running operators one at a time vs the previous behaviour.

src/imperative/cached_op.cc
```diff
@@ -262,6 +262,35 @@ std::vector<nnvm::NodeEntry> CachedOp::Gradient(
...
+
+  nnvm::Graph& g = state.info.fwd_graph;
+  ShapeVector shape_inputs;
+  shape_inputs.reserve(inputs.size());
+  for (auto input : inputs) {
+    shape_inputs.emplace_back(input->shape());
+  }
+  // We leverage the shape inference pass to detect whether dynamic shape exists.
+  // If so, the pass will fail with `contain_dynamic_shape = true`,
+  // This method is only called once, so the overhead is negligible.
```
larroy: Were you able to verify this? I'm afraid this could cause slowdown.

junrushao1994:
> Since this commit might cause a regression, I would kindly request if you can elaborate a bit about the differences between running operators one at a time vs the previous behaviour.

I agree. This is part of [this](https://cwiki.apache.org/confluence/display/MXNET/Dynamic+shape) by @zheng-da, but I believe it will be super helpful if
1) the changes are reflected or elaborated in our docs so that the behavior is clearly understood.
2) suppress the newly added naive run graph pass if other flags (like static_alloc) are on.

src/imperative/cached_op.cc
junrushao1994: If infer shape fails, it means there is probably an op of dynamic shape (np.unique, boolean mask).

MXNet didn't support this kind of op before, because of the limitation of our system. Because of these lines, we now can support it in Gluon blocks.

I would suggest to clearly mention the behavior in our docs that if infer shape fails, the code will go to the slow path (naive run graph, etc).

PR #14195, opened 2019-02-18
[Clojure] Add methods based on NDArrayAPI/SymbolAPI

kedarbellare:
## Description ##

- This PR generates new methods based on the newer NDArrayAPI and SymbolAPI.
- It uses the scala's `Base/_LIB` with direct access to JNI in order to generate the ndarray and symbol methods along with named arguments and documentation. Although the current PR doesn't add specs these can be added over time.

### Essentials ###

### Changes ###
- [x] NDArrayAPI methods, tests, (and when applicable, API doc)
- [x] SymbolAPI methods, tests, (and when applicable, API doc)

### Reviewers ###

@gigasquid


roywei:
@mxnet-label-bot [Clojure]

gigasquid:
Thanks for looking into this. It will be interesting to see how it comes out. The new API might be more difficult to interop with than the regular data structures of the current one.

kedarbellare:
@gigasquid it does seem like the new API is somewhat more cumbersome in terms of interop. the main benefit i see is the type checking but while in scala you can call the functions with named optional arguments (e.g. `Dropout(ndarray, p = Some(0.25), out = Some(outputArray))`) it's not so straightforward to have this in clojure.

gigasquid:
@kedarbellare Thanks so much for looking into this. I thought it might be the case that it might not be as nice to use. That's totally fine. We can stick to the other API for now and if for some reason we need to move off of it (like the Scala package deprecates it) , we can think of a different strategy, like moving a level down for NDArrary and Symbol and using the Scala Generator functions to generate our own functions directly using the JNI calls instead.

For example - getting all the backend functions by calling the JNI
https://github.com/apache/incubator-mxnet/blob/master/scala-package/macros/src/main/scala/org/apache/mxnet/GeneratorBase.scala#L63

and these to create the functions
https://github.com/apache/incubator-mxnet/blob/master/scala-package/macros/src/main/scala/org/apache/mxnet/GeneratorBase.scala#L73

Let me know what you think.

kedarbellare:
Makes sense. I'm working on unit tests for NDArray API right now which should give an idea of how we could start using it. I won't put too much effort into doing the same for Symbol API. It'll be great if we can start using JNI directly but I don't have much experience on that front.

gigasquid:
@kedarbellare Thanks for tackling the toughest part of the whole Clojure package 💯  I think you have made some really nice progress and we can definitely leverage some of this work.

Since we are starting a new ndarray api, we have a nice opportunity to improve upon the current api and make it more useful for Clojure users. I agree with your assessment that the typing makes things better, but all the positional arguments and options make it more cumbersome. This is due to the philosophical differences between Clojure and Scala, where Clojure loves data structures and Scala loves types :) I'm thinking we should take a step back and think about what *we* (the clojure package) would want and then try to shape what you have into it. After looking at what you have, I think there is a path forward with maybe generating only the arity with all the parameters and then taking a data structure of vectors or maps to get there.

I think it would be nice to outline what that API should look like with a couple examples from the NDArray and Symbol API before we code it up too much. Since you have experience and good feeling for what is possible now with your work, are you up for helping to draft up the future look of the API on the wiki?

kedarbellare:
Yep. I wholeheartedly agree about the philosophical differences between Scala and Clojure and I completely see the value in doing what's best for the Clojure package. I can try to take a stab at creating an outline so that we can add it to the wiki and ask for comments. Note, however, that I mostly work on this project during the weekends or when I have free time so I'll try my best to put something together soon (over the next week or so). Please let me know if you want to have something out there ASAP, in which case it may not make sense for me to work on it.

gigasquid:
@kedarbellare Definitely no rush - whatever pace works well for you. I appreciate your efforts and your contributions to improve the Clojure project.

I don't have rights to create accounts on the wiki so I'll ask out on dev for access for you.

Thanks again!

gigasquid:
@kedarbellare You should have access now - ping me if you have any trouble 😸

vandanavk:
@mxnet-label-bot update [Clojure, pr-work-in-progress]

kedarbellare:
@gigasquid i've improved the api generation for both ndarrays and symbols. PTAL thanks!

gigasquid:
Awesome work!!! I look forward to reviewing this 😸

contrib/clojure-package/test/org/apache/clojure_mxnet/symbol_api_test.clj
```diff
@@ -0,0 +1,61 @@
...
+                                      (sym/list-arguments)))))
+
+(deftest test-infer-type
+  (let [data (sym/variable "data")
+        f32data (sym-api/cast {:data data :dtype "float32"})
+        fc1 (sym-api/fully-connected {:data f32data :num-hidden 128 :name"fc1"})
+        mlp (sym-api/softmax-output {:data fc1 :name"softmax"})
+        [arg out aux] (sym/infer-type mlp {:data dtype/FLOAT64})]
+    (is (= [dtype/FLOAT64 dtype/FLOAT32 dtype/FLOAT32 dtype/FLOAT32] (util/buffer->vec arg)))
+    (is (= [dtype/FLOAT32 (util/buffer->vec out)]))
```
Chouffe: Shouldn't it be:
```
(is (= [dtype/FLOAT32] (util/buffer->vec out)))
```

contrib/clojure-package/src/dev/generator.clj
```diff
@@ -56,37 +60,138 @@
...
-(def symbol-public-no-default (->> symbol-public
-                                   (filter #(not (re-find #"org\$apache\$mxnet" (str (:name %)))))
-                                   (filter #(not (re-find #"\$default" (str (:name %)))))))
+;;;;;;; Common operations
+
+(def libinfo (Base/_LIB))
+(def op-names
+  (let [l ($ ListBuffer/empty)]
+    (do (.mxListAllOpNames libinfo l)
+        (filter #(not (or (= "Custom" %)
```
Chouffe: Should we use `remove` instead of `filter #(not ...)`

contrib/clojure-package/src/dev/generator.clj
```diff
@@ -56,37 +60,138 @@
...
+        {:params []}
+        pnames)
+       :params))
+
+(defn get-public-no-default-methods [obj]
+  (->> (r/reflect obj)
+       :members
+       (map #(into {} %))
+       (filter #(-> % :flags :public))
+       (filter #(not (re-find #"org\$apache\$mxnet" (str (:name %)))))
```
Chouffe: `remove`? instead of `filter #not`?

contrib/clojure-package/src/dev/generator.clj
```diff
@@ -56,37 +60,138 @@
...
+       (filter #(not (re-find #"\$default" (str (:name %)))))))
+
+(defn get-public-to-gen-methods [public-to-hand-gen public-no-default]
+  (remove #(contains? (->> public-to-hand-gen
```
Chouffe: should we use a a `let` binding to avoid recomputing the set over and over?
```clojure
(let [public-to-hand-gen-names
       (->> public-to-hand-gen
               (mapv :name)
               (mapv str)
               (set))] ...)
```
Also, we can probably use the fact that a set is a function
```clojure
(remove #(-> % :name str public-to-hand-gen-names) public-no-default)
```

contrib/clojure-package/src/dev/generator.clj
```diff
@@ -314,16 +387,171 @@
...
+  (let [{:keys [fn-name fn-description args]} (gen-op-info op-name)
+        params (mapv (fn [{:keys [name] :as opts}]
+                       (assoc opts :sym (symbol name)))
+                     (conj args {:name "out"
+                                 :type "NDArray-or-Symbol"
+                                 :optional? true
+                                 :description "Output array (optional)"}))
+        doc (gen-ndarray-api-doc fn-description params)
+        opt-params (filter :optional? params)
+        req-params (filter (comp not :optional?) params)
```
Chouffe: `(remove :optional? params)` instead of `filter not`?

contrib/clojure-package/test/org/apache/clojure_mxnet/symbol_api_test.clj
gigasquid: good catch @Chouffe !

gigasquid:
@kedarbellare - I think the next step after this is to try to get https://github.com/apache/incubator-mxnet/blob/master/contrib/clojure-package/test/org/apache/clojure_mxnet/conv_test.clj to use the new symbol-api. It will probably shake out some further refinements.

It certainly is a much nicer api to use and would love to get it to the point were we could deprecate the old symbol and ndarray way of doing things 😸

gigasquid:
I'm also not sure why the unix-gpu keeps failing from CI, might want to try to rebase on master and see if that helps

gigasquid:
Oh @kedarbellare - one more thing... Would you mind marking the new api namespaces as `Experimental` ? That way users know that it is in active work and might change and not be stable yet.

Thanks!

kedarbellare:
@Chouffe @gigasquid : thanks for the detailed feedback! i'll take care of these issues soon (hopefully by end of the week). 

@gigasquid : where should i mark the namespaces as `Experimental`? in the docs at the top or in the individual docs of the functions or somewhere else?

gigasquid:
@kedarbellare marking it as Experimental in the namespace docstring of contrib/clojure-package/src/org/apache/clojure_mxnet/ndarray_api.clj - or a comment should be fine. And in the corresponding symbol-api namespace. Thanks!

Chouffe:
We should probably start rewriting some of the examples with this once it gets merged. 
I can help with this @kedarbellare @gigasquid :)

contrib/clojure-package/src/dev/generator.clj
Chouffe: #nitpick 
We can even use a `xform` here I guess - not convinced it is more readable though
```clojure
(let [public-to-hand-gen-names
       (into #{} (mapv (comp str :name) public-to-hand-gen)] ...)
```

kedarbellare: done! agree that it looks much nicer :)

kedarbellare:
@gigasquid @Chouffe i've updated the PR with the following:

1. addressed cleanup comments
2. tagged the new apis as `"Experimental"`
3. modified the `conv_test.clj` to use the new `symbol-api` and fixed a small bug that i discovered during that
4. tweaked the `coerce-param` for `symbol-api` (especially for the `attr` param)
5. rebased to the latest upstream/master to see if that fixes the unix-gpu CI failure

PTAL thanks!

PR #14200, opened 2019-02-19
Bulked op segments to allow Variable nodes

DickJC123:
## Description ##
Background: Operators are bulked into segments that define synchronization points between the cpu-based worker threads and the often-GPU-based operator execution.  While the default size of the segments is set to 15, in practice the segment size is much smaller due to the current restriction that 'Variable' nodes (e.g. weight inputs to Convolution) terminate the segment formation.  This is an unnecessary restriction that can be removed to increase segment size and improve performance.

This PR changes the operator bulking code to allow Variable nodes to be part of the segment, but does not count them toward the limit, which is set currently by the environment variable MXNET_EXEC_BULK_EXEC_MAX_NODE_TRAIN (default = 15).  Larger segments are desirable to improve both CPU and GPU efficiency.  However, there are practical considerations that keep full forward- and backward-pass bulking from being the optimal.  For example, generated gradients of the backward pass must wait until the end of a segment before gradient reduction can get started in a multi-GPU training scenario.  Because of the possible different optima in the forward and backward pass, this PR adds 2 additional environment variable 'knobs':

MXNET_EXEC_BULK_EXEC_MAX_NODE_TRAIN_FWD - sets the maximum size of bulked operator segments in the forward pass.  If unset, then MXNET_EXEC_BULK_EXEC_MAX_NODE_TRAIN is used, or if that is similarly unset, the default of 15.

MXNET_EXEC_BULK_EXEC_MAX_NODE_TRAIN_BWD - sets the maximum size of bulked operator segments in the backward pass.  If unset, then MXNET_EXEC_BULK_EXEC_MAX_NODE_TRAIN is used, or if that is similarly unset, the default of 15.

While this PR is being reviewed, I will make a further post detailing the performance gains that result.  Also due is a commit that adds documentation for the additional environment variables.

### Essentials ###
- [X] Changes are complete (i.e. I finished coding on this PR.  Documentation update still pending.) 
- [X] All changes have test coverage: In effect the entire CI now runs with the larger segments.

### Changes ###

## Comments ##


ptrendx:
Just to clarify: this is bulking support for symbolic execution. Bulking in Gluon was handled in (already merged) PR #13890

roywei:
@mxnet-jenkins add[Backend, pr-awaiting-review]

DickJC123:
I measured the perf gains of this PR under 2 scenarios.

The first scenario was a run across 8 Volta GPUs of a mixed precision NHWC Resnet50 v1b (also with horovod and DALI in NVIDIA's MXNet container).  To simulate upstream MXNet with it's current bulking, I set the new-PR's bulking to 2.  I reasoned that a residual unit has a forward sequence Conv-BN-Relu-Conv-BN-Add-Relu, which would have 4 segments (due to the Variable inputs to the Conv's and BN's) amongst 7 nodes, for an average nodes/segment of 1.75.  The speed-up measured from MXNET_EXEC_BULK_EXEC_MAX_NODE_TRAIN=2 to MXNET_EXEC_BULK_EXEC_MAX_NODE_TRAIN=15 was:

```
batchsize  32: 3.8% speedup
batchsize  64: 5.1% speedup
batchsize 128: 1.9% speedup
batchsize 256: 1.2% speedup
```

A second scenario was a run across 8 Volta GPUs of a mixed precision NCHW Inception-v3 (with DALI in NVIDIA's MXNet container).   The speed-up measured from MXNET_EXEC_BULK_EXEC_MAX_NODE_TRAIN=2 to MXNET_EXEC_BULK_EXEC_MAX_NODE_TRAIN=15 was:

```
batchsize  32: 6.3% speedup
batchsize  64: 7.4% speedup
batchsize 128: 3.8% speedup
batchsize 256: 2.1% speedup
```

DickJC123:
The main change of this PR is in graph_executor.cc in an area often touched by @piiswrong @eric-haibin-lin @KellenSunderland.  Not sure if any of them would like to weigh in.

roywei:
@mxnet-label-bot add[Backend]

docs/faq/env_var.md
```diff
@@ -115,7 +115,13 @@ $env:MXNET_STORAGE_FALLBACK_LOG_VERBOSE=0
   - If set to `1`, during training MXNet executes the computation graph as several subgraphs in bulk mode.
 * MXNET_EXEC_BULK_EXEC_MAX_NODE_TRAIN
   - Values: Int ```(default=15)```
-  - The maximum number of nodes in the subgraph executed in bulk during training(not inference). Setting this to a larger number may reduce the degree of parallelism for multi-GPU training.
+  - The maximum number of nodes in the subgraph executed in bulk during training (not inference). Setting this to a larger number may reduce the degree of parallelism for multi-GPU training.
```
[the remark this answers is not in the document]
DickJC123: Not really.  The thought was that forward() might be bulked to a very large extent, while backward might want smaller segments if that allows gradient reduction to start sooner in a multi-GPU scenario.

Also, I prefer consistency and the forward_bulk_size and backward_bulk_size members of CachedOpConfig were added 8 months ago, seemingly as an invite to expand the functionality to support bulk size separately for forward() and backward().

To add some data to the discussion, I've created a unittest that I'll be adding to this PR that tests for the perf increase of bulking.  It uses a symbolic model comprising a 1000 small GPU kernels.  An adapted variant of that test produced the following stats:
```
fwd & bwd bulk seg size = 0,    runtime = 32.9 msecs
fwd & bwd bulk seg size = 1,    runtime = 33.7 msecs
fwd & bwd bulk seg size = 2,    runtime = 24.8 msecs
fwd & bwd bulk seg size = 3,    runtime = 22.1 msecs
fwd & bwd bulk seg size = 4,    runtime = 20.7 msecs
fwd & bwd bulk seg size = 5,    runtime = 19.8 msecs
fwd & bwd bulk seg size = 6,    runtime = 19.6 msecs
fwd & bwd bulk seg size = 7,    runtime = 19.1 msecs
fwd & bwd bulk seg size = 8,    runtime = 18.8 msecs
fwd & bwd bulk seg size = 9,    runtime = 18.8 msecs
fwd & bwd bulk seg size = 10,   runtime = 18.5 msecs
fwd & bwd bulk seg size = 11,   runtime = 18.6 msecs
fwd & bwd bulk seg size = 12,   runtime = 18.1 msecs
fwd & bwd bulk seg size = 13,   runtime = 18.2 msecs
fwd & bwd bulk seg size = 14,   runtime = 17.9 msecs
fwd & bwd bulk seg size = 15,   runtime = 17.8 msecs
fwd & bwd bulk seg size = 16,   runtime = 17.6 msecs
fwd & bwd bulk seg size = 17,   runtime = 17.8 msecs
fwd & bwd bulk seg size = 18,   runtime = 18.0 msecs
fwd & bwd bulk seg size = 19,   runtime = 17.6 msecs
fwd & bwd bulk seg size = 20,   runtime = 18.1 msecs
fwd & bwd bulk seg size = 21,   runtime = 18.1 msecs
fwd & bwd bulk seg size = 22,   runtime = 17.7 msecs
fwd & bwd bulk seg size = 23,   runtime = 17.5 msecs
fwd & bwd bulk seg size = 24,   runtime = 16.2 msecs
```
It shows that there's a huge benefit for this launch-bound model of going from bulk_size=1 to bulk_size=2, with diminishing returns in the bulk_size=15 range.  On the backward() pass of RN50, there are 2 conv weight gradients generated for every 7 nodes or so.  That might suggest a small bulking for backward, although it really depends on whether the parameter server function is off the cycle-time critical path.

Bottom line, if we don't have these knobs, it's hard to know if we've tuned the platform correctly for a given model.  I like to think of the environment variables as split into two categories: every day ones, and those that one can safely ignore that are there for the experts to play with.

src/imperative/cached_op.h
```diff
@@ -53,10 +53,12 @@ struct CachedOpConfig : public dmlc::Parameter<CachedOpConfig> {
     .set_default(2)
     .describe("Maximum number of operators that can be inlined.");
     DMLC_DECLARE_FIELD(forward_bulk_size)
-    .set_default(dmlc::GetEnv("MXNET_EXEC_BULK_EXEC_MAX_NODE_TRAIN", 15))
```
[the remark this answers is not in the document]
DickJC123: Please review the latest commits, which address this valid concern about consistency.  I also consolidated all op-bulking env var references to a central place and added timing-based tests for the perf impact of bulking.  I'm happy with the PR now (assuming it passes CI).  Anyone else you want to pull into the review @eric-haibin-lin?

DickJC123:
@eric-haibin-lin Thanks for your comments thus far.  The PR is back in a good state for you to complete your review, although it appears testing on Windows has been bypassed .  Happy to respond to further comments from you or from others you may wish to pull into the review.

src/executor/graph_executor.cc
```diff
@@ -1211,63 +1212,53 @@ void GraphExecutor::InitOpSegs() {
...
-  size_t num_nodes_threshold = dmlc::GetEnv("MXNET_EXEC_BULK_EXEC_MAX_NODE_TRAIN", 15);
+  // The maximum number of nodes in a segment executed in bulk (excluding variables) in fwd pass.
+  size_t segment_num_nodes_threshold_fwd = Imperative::BulkExecMaxNodeTrainFwd();
+  // The maximum number of nodes in a segment executed in bulk (excluding variables) in bwd pass.
+  size_t segment_num_nodes_threshold_bwd = Imperative::BulkExecMaxNodeTrainBwd();
 
   // create forward segments for training
   size_t topo_start = 0;
+  size_t segment_node_count = 0;
   for (size_t nid = 0; nid < num_forward_nodes_; nid++) {
```
eric-haibin-lin: The inner loops for forward and backward bulking are very similar. Do you think it's better to refactor them as functions?

DickJC123: I think having a single function to bulk forward and backward is a good idea, such as:
```
void GraphExecutor::BulkTrainingOpSegsOverRange(size_t from_node, size_t up_to_node, size_t max_nodes_per_segment) { ... }
```
The only issue to work out is if there's a good reason why in forward we have:
```
bool ignore_node = node->is_variable();
```
while in backward we have the more complicated:
```
bool ignore_node = node->is_variable() || op_node.skip_exec_node || op_node.exec == nullptr; 
```
Do you know by any chance?  The goal would be to unify these two as much as possible.

eric-haibin-lin: Should be fine to always use 
```
bool ignore_node = node->is_variable() || op_node.skip_exec_node || op_node.exec == nullptr; 
```
`skip_exec_node` defaults to False. Is it set to True if the node should be skipped during execution. This is usually for elemwise_add op during backward pass. It is set when the elemwise_add op is skipped, and the req for the corresponding entry at the previous op is set to `kAddTo`. For forward pass, `skip_exec_node` should always be False. 
https://github.com/apache/incubator-mxnet/blob/992c3c0dd90c0723de6934e826a49bad6569eeac/src/executor/exec_pass.h#L162

It should also be ok to check `op_node.exec` for forward ops, as non variable nodes are assigned with exec: https://github.com/apache/incubator-mxnet/blob/master/src/executor/graph_executor.cc#L1089-L1094

DickJC123: Your comments were very helpful.  Based on the encouragement, I consolidated the routines for bulking, going a little farther than what you suggested by including the forward inference bulking as well.  CI is offline right now.

Issue #9543, opened 2018-01-24
Variable Length Support for cuDNN RNN


sbodenstein:
We want to add support for variable-length sequences to the cuDNN RNN operator, as we cannot 'fake' support via masking for LSTM (the cuDNN operator does not return a history of cell states) and bidirectional RNNs. Also, there should be some speedups for batches of sequences of very different lengths (can avoid bucketing).

The annoying part is that cuDNN requires that the sequences be sorted by sequence length, and also packed as a contiguous block of memory. PyTorch deals with this by introducing a [pack_padded_sequence](http://pytorch.org/docs/master/nn.html#torch.nn.utils.rnn.pack_padded_sequence) operator, which returns a `PackedSequence` object that the user can feed to cuDNN RNN. MXNet probably (?) doesn't want to take this route, and hide these details from the user (other backends eg upcoming MKL for `symbol.rnn` might anyways not require this) and just pack/unpack inside `symbol.RNN`. So some questions: 
- is any work being done on this already? (so I don't duplicate)
- @DickJC123: what is the best way of doing the sorted packing/unpacking efficiently? Are there any plans to add this utility function to cuDNN? 

@sxjscience, @piiswrong, @szha

sbodenstein:
Any comments on this?

lupesko:
Awesome initiative @sbodenstein 
Bumping again to @piiswrong 

@sandeep-krishnamurthy please label:
- RNN
- Feature

piiswrong:
I don't think any one is working on this. This can be added as an option of sym.RNN. when use_mask=True, RNN can take an extra argument.

ping @DickJC123 again

DickJC123:
I've been remiss at commenting because frankly new cuDNN features are generally advertised only upon general availability.  The cuDNN team is continually improving the feature set, and it's clear to us that sorting and packing of sequences is a customer pain point.  As such, we are working on feature improvements in this area for one of our future releases.  I'll point out that cudNN v7.1 is newly available for download, although this minor release has no new features to address variable-length sequences.  If there's more detail you are able to share about your application, whether publicly or privately, that would be very helpful.

sbodenstein:
So it seems as though we should wait a little bit before jumping in to add support. It seems as though Intel will have their RNN operator ready this month as well, so it might be a good time to revisit this then.

whamza15:
Any progress on this?

DickJC123:
[...]

PR #14208, opened 2019-02-19
Add support for fast variable-length LSTM

stephenrawls:
## Description ##
This is very much a work in progress (WIP).

I am trying to add support for variable length LSTMs in MxNet, and need help with someone to help me complete the PR.

This PR currently adds support for variable length sequences using the built-in support of that by cuDNN. It currently does not add CPU support but probably should.

I had to make a few changes to get it working that definitely need to change before we think about merging:
  1) I changed the way gluon handles kwargs to model blocks. Originally I did this because I thought I needed to pass in `use_sequence_length=True, sequence_length=sequence_length` the way the current SequenceLast/SequenceReverse/etc operators do. Now I think I probably don't need this since I moved the use_sequence_length parameter to the constructor.   (Although perhaps we don't need this at all? it seems clunky to me, but I was trying to follow existing conventions).

  2) Currently I hard-coded the `F.RNN` call in rnn_layer.py to look like: `.RNN(inputs, params, states[0], states[1], sequence_length, ...` because this was giving me python syntax errors:  `.RNN(inputs, params, *states, sequence_length,...`

  3) optimizer.py had a bunch of imports that were failing for me (e.g. `multi_sgd_update`). Are they not failing for everyone else? Not sure why they just failed for me, but I had to remove to get my test script to run.

  4) I had to modify imperative_utils.h to temporarily remove the check that all ndarrays are on the same context, because cudnn expects the sequence length array to live on the CPu.

Still need to add unit tests to test this feature. currently I am using the following test script:
https://gist.github.com/stephenrawls/8a4f4cb941f0c3f32bf5a8b3b12d32c0

Unfortunately it looks like the test I have is failing. Not sure why might need help from cudnn guys to help track down.

I have tested with cudnn v7.4.2.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/python/mxnet/gluon/rnn/rnn_layer.py
+++ b/python/mxnet/gluon/rnn/rnn_layer.py
@@ -37,7 +37,7 @@ class _RNNLayer(HybridBlock):
                  i2h_bias_initializer, h2h_bias_initializer,
                  mode, projection_size, h2r_weight_initializer,
                  lstm_state_clip_min, lstm_state_clip_max, lstm_state_clip_nan,
-                 dtype, **kwargs):
+                 dtype, use_sequence_length=False, **kwargs):
         super(_RNNLayer, self).__init__(**kwargs)
         assert layout in ('TNC', 'NTC'), \
             "Invalid layout %s; must be one of ['TNC' or 'NTC']"%layout
@@ -58,6 +58,7 @@ class _RNNLayer(HybridBlock):
         self._lstm_state_clip_max = lstm_state_clip_max
         self._lstm_state_clip_nan = lstm_state_clip_nan
         self._dtype = dtype
+        self._use_sequence_length = use_sequence_length
 
         self._gates = {'rnn_relu': 1, 'rnn_tanh': 1, 'lstm': 4, 'gru': 3}[mode]
 
@@ -219,7 +220,7 @@ class _RNNLayer(HybridBlock):
             states.append(func(name='%sh0_%d'%(self.prefix, i), **info))
         return states
 
-    def hybrid_forward(self, F, inputs, states=None, **kwargs):
+    def hybrid_forward(self, F, inputs, states=None, sequence_length=None, **kwargs):
         if F is ndarray:
             batch_size = inputs.shape[self._layout.find('N')]
         skip_states = states is None
@@ -236,12 +237,12 @@ class _RNNLayer(HybridBlock):
                     raise ValueError(
                         "Invalid recurrent state shape. Expecting %s, got %s."%(
                             str(info['shape']), str(state.shape)))
-        out = self._forward_kernel(F, inputs, states, **kwargs)
+        out = self._forward_kernel(F, inputs, states, sequence_length, **kwargs)
 
         # out is (output, state)
         return out[0] if skip_states else out
 
-    def _forward_kernel(self, F, inputs, states, **kwargs):
+    def _forward_kernel(self, F, inputs, states, sequence_length, **kwargs):
         """ forward using CUDNN or CPU kenrel"""
         if self._layout == 'NTC':
             inputs = F.swapaxes(inputs, dim1=0, dim2=1)
@@ -261,13 +262,22 @@ class _RNNLayer(HybridBlock):
 
         params = F._internal._rnn_param_concat(*params, dim=0)
 
-        rnn = F.RNN(inputs, params, *states, state_size=self._hidden_size,
-                    projection_size=self._projection_size,
-                    num_layers=self._num_layers, bidirectional=self._dir == 2,
-                    p=self._dropout, state_outputs=True, mode=self._mode,
-                    lstm_state_clip_min=self._lstm_state_clip_min,
-                    lstm_state_clip_max=self._lstm_state_clip_max,
-                    lstm_state_clip_nan=self._lstm_state_clip_nan)
+        if self._use_sequence_length:
+            rnn = F.RNN(inputs, params, states[0], states[1], sequence_length, use_sequence_length=self._use_sequence_length,
+                        state_size=self._hidden_size, projection_size=self._projection_size,
+                        num_layers=self._num_layers, bidirectional=self._dir == 2,
+                        p=self._dropout, state_outputs=True, mode=self._mode,
+                        lstm_state_clip_min=self._lstm_state_clip_min,
+                        lstm_state_clip_max=self._lstm_state_clip_max,
+                        lstm_state_clip_nan=self._lstm_state_clip_nan)
+        else:
+            rnn = F.RNN(inputs, params, *states, state_size=self._hidden_size,
+                        projection_size=self._projection_size,
+                        num_layers=self._num_layers, bidirectional=self._dir == 2,
+                        p=self._dropout, state_outputs=True, mode=self._mode,
+                        lstm_state_clip_min=self._lstm_state_clip_min,
+                        lstm_state_clip_max=self._lstm_state_clip_max,
+                        lstm_state_clip_nan=self._lstm_state_clip_nan)
 
         if self._mode == 'lstm':
             outputs, states = rnn[0], [rnn[1], rnn[2]]
--- a/python/mxnet/gluon/rnn/rnn_layer.py
+++ b/python/mxnet/gluon/rnn/rnn_layer.py
@@ -220,7 +220,7 @@ class _RNNLayer(HybridBlock):
             states.append(func(name='%sh0_%d'%(self.prefix, i), **info))
         return states
 
-    def hybrid_forward(self, F, inputs, states=None, sequence_length=None, **kwargs):
+    def hybrid_forward(self, F, inputs, sequence_length=None, states=None, **kwargs):
         if F is ndarray:
             batch_size = inputs.shape[self._layout.find('N')]
         skip_states = states is None
@@ -237,12 +237,12 @@ class _RNNLayer(HybridBlock):
                     raise ValueError(
                         "Invalid recurrent state shape. Expecting %s, got %s."%(
                             str(info['shape']), str(state.shape)))
-        out = self._forward_kernel(F, inputs, states, sequence_length, **kwargs)
+        out = self._forward_kernel(F, inputs, sequence_length, states, **kwargs)
 
         # out is (output, state)
         return out[0] if skip_states else out
 
-    def _forward_kernel(self, F, inputs, states, sequence_length, **kwargs):
+    def _forward_kernel(self, F, inputs, sequence_length, states, **kwargs):
         """ forward using CUDNN or CPU kenrel"""
         if self._layout == 'NTC':
             inputs = F.swapaxes(inputs, dim1=0, dim2=1)
@@ -263,21 +263,18 @@ class _RNNLayer(HybridBlock):
         params = F._internal._rnn_param_concat(*params, dim=0)
 
         if self._use_sequence_length:
-            rnn = F.RNN(inputs, params, states[0], states[1], sequence_length, use_sequence_length=self._use_sequence_length,
-                        state_size=self._hidden_size, projection_size=self._projection_size,
-                        num_layers=self._num_layers, bidirectional=self._dir == 2,
-                        p=self._dropout, state_outputs=True, mode=self._mode,
-                        lstm_state_clip_min=self._lstm_state_clip_min,
-                        lstm_state_clip_max=self._lstm_state_clip_max,
-                        lstm_state_clip_nan=self._lstm_state_clip_nan)
+            rnn_args = states + [sequence_length]
         else:
-            rnn = F.RNN(inputs, params, *states, state_size=self._hidden_size,
-                        projection_size=self._projection_size,
-                        num_layers=self._num_layers, bidirectional=self._dir == 2,
-                        p=self._dropout, state_outputs=True, mode=self._mode,
-                        lstm_state_clip_min=self._lstm_state_clip_min,
-                        lstm_state_clip_max=self._lstm_state_clip_max,
-                        lstm_state_clip_nan=self._lstm_state_clip_nan)
+            rnn_args = states
+
+        rnn = F.RNN(inputs, params, *rnn_args, use_sequence_length=self._use_sequence_length,
+                    state_size=self._hidden_size, projection_size=self._projection_size,
+                    num_layers=self._num_layers, bidirectional=self._dir == 2,
+                    p=self._dropout, state_outputs=True, mode=self._mode,
+                    lstm_state_clip_min=self._lstm_state_clip_min,
+                    lstm_state_clip_max=self._lstm_state_clip_max,
+                    lstm_state_clip_nan=self._lstm_state_clip_nan)
+
 
         if self._mode == 'lstm':
             outputs, states = rnn[0], [rnn[1], rnn[2]]
--- a/python/mxnet/gluon/rnn/rnn_layer.py
+++ b/python/mxnet/gluon/rnn/rnn_layer.py
@@ -220,7 +220,7 @@ class _RNNLayer(HybridBlock):
             states.append(func(name='%sh0_%d'%(self.prefix, i), **info))
         return states
 
-    def hybrid_forward(self, F, inputs, states=None, sequence_length=None, **kwargs):
+    def hybrid_forward(self, F, inputs, sequence_length=None, states=None, **kwargs):
         if F is ndarray:
             batch_size = inputs.shape[self._layout.find('N')]
         skip_states = states is None
--- a/python/mxnet/gluon/rnn/rnn_layer.py
+++ b/python/mxnet/gluon/rnn/rnn_layer.py
@@ -223,7 +223,7 @@ class _RNNLayer(HybridBlock):
     def __call__(self, inputs, states=None, sequence_length=None, **kwargs):
         self.skip_states = states is None
         if states is None:
-            if type(inputs) is ndarray.NDArray:
+            if isinstance(inputs, ndarray.NDArray):
                 batch_size = inputs.shape[self._layout.find('N')]
                 states = self.begin_state(batch_size, ctx=inputs.context, dtype=inputs.dtype)
             else:
--- a/python/mxnet/gluon/rnn/rnn_layer.py
+++ b/python/mxnet/gluon/rnn/rnn_layer.py
@@ -220,7 +220,7 @@ class _RNNLayer(HybridBlock):
             states.append(func(name='%sh0_%d'%(self.prefix, i), **info))
         return states
 
-    def hybrid_forward(self, F, inputs, sequence_length=None, states=None, **kwargs):
+    def hybrid_forward(self, F, inputs, states=None, sequence_length=None, **kwargs):
         if F is ndarray:
             batch_size = inputs.shape[self._layout.find('N')]
         skip_states = states is None
--- a/python/mxnet/gluon/rnn/rnn_layer.py
+++ b/python/mxnet/gluon/rnn/rnn_layer.py
@@ -252,7 +252,7 @@ class _RNNLayer(HybridBlock):
         # out is (output, state)
         return out[0] if self.skip_states else out
 
-    def _forward_kernel(self, F, inputs, sequence_length, states, **kwargs):
+    def _forward_kernel(self, F, inputs, states, sequence_length, **kwargs):
         """ forward using CUDNN or CPU kenrel"""
         if self._layout == 'NTC':
             inputs = F.swapaxes(inputs, dim1=0, dim2=1)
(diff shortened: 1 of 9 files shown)


ptrendx:
```
3\. optimizer.py had a bunch of imports that were failing for me (e.g. `multi_sgd_update`). Are they not failing for everyone else? Not sure why they just failed for me, but I had to remove to get my test script to run.
```
This is very strange - are you sure you are importing the right version of .so? Those functions are not present in 1.3.1 or 1.4.0 but should be present on master.
```
4\. I had to modify imperative_utils.h to temporarily remove the check that all ndarrays are on the same context, because cudnn expects the sequence length array to live on the CPu.
```
This is going to be a bigger issue since I don't think NNVM currently allows you to do inputs on different contexts. I think @sandeep-krishnamurthy was going to look at it at some point for handling of nvJPEG.

stephenrawls:
Sorry will fix shortly. Was a snafu when trying to clean up my code before pushing to GitHub. Thanks for the catch!

stephenrawls:
@szha maybe this deserves another PR but I also noticed that cuDNN supports both sequence-major and batch-major sequences:

From: https://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html#cudnnRNNForwardInferenceEx
```
With unpacked layout, both sequence major (i.e. time major) and batch major are supported
```

Currently gluon is supporting batch-major sequences by calling swapaxes:
https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/gluon/rnn/rnn_layer.py#L241-L242

This ends up calling mshadow's swapaxis function, which I *assume* must be doing a copy (?) but to be honest I couldn't totally follow what the code is doing at a quick glance:
https://github.com/dmlc/mshadow/blob/3dc80815d965b56b9a975dc27229361955bf66fe/mshadow/extension/swapaxis.h 

If it *is* doing a copy for swapaxes, it would be preferable for Gluon to just let cudnn handle the batch-major layout directly w/o calling swapaxes.

szha:
@stephenrawls swapaxis is doing an actual transpose

TaoLv:
@stephenrawls Still cannot get the whole picture of this PR. Can you post a proposal for this task with below information? Then we can think about how to support it on CPU side. Thank you.

- What does the front end API look like? What's the typical usage of the API?
- What's done in Gluon blocks/hybridblocks and sym.RNN to support that?
- What's done inside cudnn?
- From the title, only LSTM is supported?

stephenrawls:
Hi @TaoLv ,

Basically my motivation is:  we want to use stacked bidirectional LSTM's with variable sequence length. Currently this is slow in MxNet because to do it right you either have to (1) Use the LSTMCell and unroll which doesn't take advantage of cuDNN on the GPU; (2) Use cuDNN one layer at a time, and do a lot of reversing of your output to pass into the backward-direction lstm, so you can be sure that padding doesn't effect the result.

This all seemed a bit silly since cuDNN directly provides support for variable length stacked bidirectional LSTMs efficiently.

Beyond my direct use case, for the community I guess it would probably be nice if:
  - All RNN types are supported.  (This happens already via cuDNN even though I am focussed on LSTM)
  - Both CPU and GPU are supported

I guess the API I have is currently like this:

```
lstm = mx.gluon.LSTM(..normal params..., use_sequence_length=True)
lstm(input, sequence_length)
```

I am not tied to this api however. For example if we want to not have a `use_sequence_length` param in initialization then I'm happy to remove it. I only put it there in the first place to be similar to existing Sequence* family of operators.

As far as what changes need to happen on the symbol side, I am less sure, I primarily use gluon. Happy for folks to offer suggestions.

Thanks,
Stephen

stephenrawls:
P.S. There was a GitHub issue for this last year but it sort of morphed into other things and this particular piece didn't get accomplished;

https://github.com/apache/incubator-mxnet/issues/9543

drivanov:
@stephenrawls: With following simple changes in your `test_varlength_lstm.py`:
```
ctx = mx.gpu()
X = mx.nd.random.randn(timesteps, batch_size, nchannels).as_in_context(ctx)
valid_lengths = mx.nd.array([10,10,10,7,6]).as_in_context(ctx) # [10,10,10,10,6] seems to work fine
```
you don't need to modify `imperative_utils.h`

stephenrawls:
@drivanov  -- yes if I move the sequence length array to the gpu then I don't have to change the logic the verifies everything is in the same context. The reason I am keeping the sequence length array on the cpu even though the data is on the gpu is because the cudnn api is that the sequence length array is on cpu, i.e. host memory.

If I pass in the sequence length array on GPU memory then I would have to do a GPU -> CPU copy inside of the lstm operator, right?

stephenrawls:
Latest push from me now gets variable length lstms working correctly in cudnn. Turns out that in addition to setting the layout to UNPACKED when creating input descriptors, we also need to call a function to enable unpacked padded LSTMs prior to calculating the operator's workspace size.

Now  the forward pass works correctly and my rudimentary tests pass.

Still need help figuring out what to do with passing sequence_length input on the CPU. Any ideas?

roywei:
@mxnet-label-bot add[RNN, Backend, Gluon, pr-work-in-progress]

szha:
@stephenrawls for now you can do a `CHECK(!use_seq_len)` in CPU kernel. @pengzhao-intel @TaoLv will see how to add it for CPU kernels.

stephenrawls:
@szha  okay I will do that and update PR shortly.

My question though (sorry for not being clear) is about the GPU kernel. Currently the cuDNN api requires the sequence_length array to be in host memory, so I am passing the sequence_length array on the CPU even though the data array is on the GPU.

Currently the imperative_utils.h requires that all ndarrays passed to an operator are on the same context. To get around this I just commented out that check so I could at least test that my changes were working.

My question is--(1) What's the right way to modify the context check in imperative_utils.h? (2) Is there something similar that happens on the symbol api side that would have to change?

The alternative is to just pass the sequence_length array in on the GPU, but this would require an extra GPU -> CPU data transfer inside the operator that I would prefer to avoid.

szha:
@stephenrawls the length array is generally small so I think it's fine to do a copy inside the operator. @reminisce @junrushao1994 are looking into how to refurbish the backend so that such cases can be better supported.

ptrendx:
To answer your question `(2) Is there something similar that happens on the symbol api side that would have to change?`: yes, place device graph pass will just error out saying that it is impossible to mix devices.

I had a similar problem in aggregate SGD case, where I wanted to put learning rates and weight decay rates as CPU side NDArray. I ended up passing them as a Tuple argument to the op in Python.

stephenrawls:
@szha Okay I re-worked the code to take in sequence_length on the GPU and copy to the CPU inside the kernel. I added a unit test to test the functionality and it appears to be working.

For some reason I am getting errors when trying to pass an int32 sequence_length array. Not sure why, because I changed infer_shape() to allow it. Can look into it later but if anyone has suggestions please let me know.

@ptrendx -- I figured out the weird import error and reverted those changes to optimizer.py in my PR. Also as mentioned above, since I am now doing an extra GPU->CPU copy in the kernel code, I have also reverted the change to imperative_utils.h.

@szha  -- I think all that is remaining in this PR is to add the `CHECK(!use_seq_len)` for the CPU case, to figure out why int32 sequence_length is not working, and any cleanup / modifications you think are appropriate. Also--I added a unit test for gluon, but should I also add one for the Symbol api code, or is it sufficient to test this functionality in Gluon?

pinaraws:
@szha Could you please respond to questions posted by @stephenrawls ?

piyushghai:
@stephenrawls Can you rebase with the latest master to resolve the merge conflicts ?

stephenrawls:
FYI I am looking at this. I have my changes rebased against the latest changes from master now, and updated to fit the new pattern. I just need to build and test. Since things have changed significantly since before I just need to re-setup my build environment, should have an update in a day or so.

python/mxnet/gluon/rnn/rnn_layer.py
```diff
@@ -214,7 +215,7 @@ def begin_state(self, batch_size=0, func=ndarray.zeros, **kwargs):
             states.append(func(name='%sh0_%d'%(self.prefix, i), **info))
         return states
 
-    def hybrid_forward(self, F, inputs, states=None, **kwargs):
+    def hybrid_forward(self, F, inputs, sequence_length=None, states=None, **kwargs):
```
szha: for compatibility for positional argument, maybe move the new argument after states?

stephenrawls:
@szha finally got around to rebasing my PR and pushing the update.

The centos-gpu build is failing. The new var-length unit test is failing with this error:
```
ERROR: test_gluon_gpu.test_layer_bidirectional_varseqlength
    self.test(*self.arg)
    orig_test(*args, **kwargs)
  File "/work/mxnet/tests/python/gpu/../unittest/common.py", line 112, in test_new
    assert_raises((MXNetError, RuntimeError), orig_test, *args, **kwargs)
  File "/usr/lib64/python3.6/unittest/case.py", line 750, in assertRaises
    return context.handle('assertRaises', args, kwargs)
  File "/usr/lib64/python3.6/unittest/case.py", line 195, in handle
    callable_obj(*args, **kwargs)
  File "/work/mxnet/tests/python/gpu/test_gluon_gpu.py", line 290, in test_layer_bidirectional_varseqlength
    check_layer_bidirectional_varseqlen(7, 5)
  File "/work/mxnet/tests/python/gpu/test_gluon_gpu.py", line 267, in check_layer_bidirectional_varseqlen
    net_output = net(data, sequence_length).asnumpy()
  File "/work/mxnet/python/mxnet/gluon/block.py", line 540, in __call__
    out = self.forward(*args)
  File "/work/mxnet/python/mxnet/gluon/block.py", line 912, in forward
    self._deferred_infer_shape(x, *args)
  File "/work/mxnet/python/mxnet/gluon/block.py", line 793, in _deferred_infer_shape
    raise ValueError(error_msg)
ValueError: Deferred initialization failed because shape cannot be inferred. Error in operator lstm_rnn0: Shape inconsistent, Provided = [10], inferred shape=(2,10,7)
infer_shape error. Arguments:
  data0: (11, 10, 5)
  data1: (10,)
```

I'm not sure why yet, because when I run the unit test code on my own machine with my own build it works.

Also the unix-gpu build is failing for the unit test `test_operator.test_custom_op_exc` which seems unrelated to my PR.

I'll see if I can figure out why the centos-gpu build is failing (even though it works for me on my own machine) tomorrow. But if you have any ideas / advice please let me know.

python/mxnet/gluon/rnn/rnn_layer.py
stephenrawls: I think `sequence_length` actually has to come before `states`, because it is a positional argument provided by the user, whereas `states` is an automatic positional argument provided by gluon, since gluon automatically adds a block's learnable parameters as positional arguments at the end of  `hybrid_forward()` functions' argument list.

The reason my personal unit tests were passing was because I was using a version of my code that hadn't switched the order of these arguments. I reverted this change so hopefully the centsos-gpu tests will pass now.

stephenrawls: Nevermind I misunderstood earlier.

Actually the `states` array is *possibly* a user-provided parameter, but if they don't provide it then the code will fill in a default value.

I was having trouble getting this to work:
```
model(input, sequence_length=sequence_length)
```
Because the code kept on complaining about an unknown parameter `sequence_length`. I think this must be related to how gluon handles keyword arguments for a model's forward pass.

The solution I found was to over-ride the `__call__()` method for the rnn layer which is happy to take keyword arguments, and then I can use that to impose an order for arguments that I pass to the `hybrid_forward()` function. To get this to work I also needed to move the logic of setting a default value for `states` into the `__call__()` function.

szha: right, hybrid_forward doesn't support keyword arguments.

stephenrawls:
@szha  as far as I can tell, my gluon RNN & LSTM code is now working correctly, but the LSTM & RNN tests are failing in the Symbol API unit tests. Not sure why yet.

stephenrawls:
@szha Ok looks like the tests are finally passing.  (3 are still pending as of me writing this, but centos-gpu and centos-cpu have both passed, and others look mostly done).

Can you please review and let me know if anything needs to be cleaned up before this can be merged?

(As a reminder: this PR only implements variable-length sequence support for the GPU using cudnn. Presumably a separate PR needs to add CPU support).

tests/python/gpu/test_gluon_gpu.py
```diff
@@ -225,6 +226,54 @@ def forward(self, inpt):
...
+    for k in weights:
+        net_params[k].set_data(weights[k])
+        ref_net_params[k.replace('l0', 'l0l0').replace('r0', 'r0l0')].set_data(weights[k])
+
+
+    batch_size = 10
+    num_timesteps = 11
+    data = mx.random.uniform(shape=(num_timesteps, batch_size, in_size))
+
+    # TODO: figure out why int32 doesn't work here
```
szha: is this still the case that int32 doesn't work? I thought the kernel already uses a separate IType

stephenrawls: the kernel does work with an arbitrary IType. I don't recall now why the unit test failed to work, I think it was something about some other unit test utility function expecting a float type, not the rnn kernel.

szha: it would be great to test this out later, though not urgent

src/operator/rnn.cc
```diff
@@ -106,25 +128,33 @@ static bool RNNShape(const nnvm::NodeAttrs& attrs,
...
-    CHECK_EQ(in_type->size(), 4U);
-  } else {
-    CHECK_EQ(in_type->size(), 3U);
-  }
+
+  CHECK_EQ(in_type->size(), GetNumInputArguments(param_));
+
+  size_t seq_len_input_idx = rnn_enum::kSequenceLength;
+  if (param_.mode != rnn_enum::kLstm)
+    --seq_len_input_idx;
```
szha: nit: either use single line conditional statement or use curly braces

src/operator/rnn-inl.h
```diff
@@ -1362,7 +1427,9 @@ class RNNOp {
   std::vector<cudnnTensorDescriptor_t> x_desc_vec_, y_desc_vec_, dx_desc_vec_, dy_desc_vec_;
   #if USE_CUDNN_LSTM_PROJ
   cudnnRNNDataDescriptor_t x_data_desc_, y_data_desc_, dx_data_desc_, dy_data_desc_;
+  DType padding_fill_ = 0;
   #endif
+  std::vector<IType> sequence_length_array_cpu;
```
szha: instead of allocating in the operator, try use the shared temp resource for this.

request: https://github.com/apache/incubator-mxnet/blob/master/src/operator/tensor/indexing_op.cc#L530-L533
usage: https://github.com/apache/incubator-mxnet/blob/master/src/operator/tensor/indexing_op.cc#L365
get_host_space: https://github.com/apache/incubator-mxnet/blob/master/include/mxnet/resource.h#L143

src/operator/rnn.cc
stephenrawls: Done. I think I got all of them.

src/operator/rnn-inl.h
stephenrawls: Done.

tests/python/gpu/test_gluon_gpu.py
stephenrawls: I'll get to the bottom of this in my next PR after this is merged.

(For next PR I'm planning to add support for NTC layout without transposing, which is supported by cuDNN, however currently gluon transposes NTC -> TNC:
https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/gluon/rnn/rnn_layer.py#L246-L247)

stephenrawls:
@szha  -- ok I fixed the issues I was having with the linter / unit tests.

The unix-gpu test is failing but it looks entirely unrelated to this PR:
```
Run LeNet-5 inference comparison between MXNet and TensorRT. ... [03:47:12] /work/mxnet/src/operator/subgraph/build_subgraph.cc:683: start to execute partition graph.
FAIL
FAIL: Run LeNet-5 inference comparison between MXNet and TensorRT.
  File "/usr/local/lib/python3.6/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/work/mxnet/tests/python/tensorrt/test_tensorrt_lenet5.py", line 100, in test_tensorrt_inference
    MXNet = %f, TensorRT = %f""" % (mx_pct, trt_pct)
AssertionError: Diff. between MXNet & TensorRT accuracy too high:
           MXNet = 99.020000, TensorRT = 99.030000
```

Also the unix-cpu test is failing, and it also seems unrelated:

```
FAIL: test_operator.test_custom_op_exc
...
mxnet.base.MXNetError: [03:36:22] src/operator/tensor/./dot-inl.h:1237: Check failed: L[!Ta].Size() == R[Tb].Size() (2 vs. 1) : dot shape error: [4,2] X [1,4]
…
```

stephenrawls:
rebased against mainline (requiring a force-push), which now gets all unit tests passing.

src/operator/rnn-inl.h
```diff
@@ -28,6 +28,7 @@
 
 #define MXNET_USE_CUDNN_RNN MXNET_USE_CUDNN == 1 && CUDNN_MAJOR >= 5
 #define USE_CUDNN_LSTM_PROJ MXNET_USE_CUDNN == 1 && CUDNN_VERSION >= 7200
+#define USE_VAR_SEQ_LENGTH MXNET_USE_CUDNN == 1 && CUDNN_VERSION >= 7200
```
szha: seems unnecessary for the duplicate

stephenrawls: Fair enough. At first I added it since I was putting things that had nothing to do with LSTM_PROJ under this #ifdef guard.

Maybe I'll just rename it to something that makes sense for both uses.

tests/python/gpu/test_gluon_gpu.py
```diff
@@ -225,6 +226,54 @@ def forward(self, inpt):
...
+    sequence_length = nd.random.randint(1, num_timesteps+1, shape=(batch_size)).astype("float")
+
+    net_output = net(data, sequence_length=sequence_length).asnumpy()
+    ref_net_output = ref_net(data, sequence_length).asnumpy()
+    sequence_length_np = sequence_length.asnumpy().astype("int32")
+
+    # Only compare the valid sections for each batch entry
+    for b in range(batch_size):
+        assert_allclose(net_output[:sequence_length_np[b], b], ref_net_output[:sequence_length_np[b], b])
```
szha: this doesn't seem to test if the length-based masking is working properly, because the reference implementation also relies on sequence length feature. consider using LSTMCell as reference instead.

src/imperative/imperative_utils.h
```diff
@@ -67,6 +67,7 @@ inline Context GetContext(const nnvm::NodeAttrs& attrs,
   Context ctx;
   if (inputs.size()) {
     ctx = inputs[0]->ctx();
+
```
szha: revert?

tests/python/gpu/test_gluon_gpu.py
stephenrawls: The reference net is not using the sequence length feature of cudnn, because use_sequence_length defaults to false.

The reference net is actually manually implementing the variable sequence length support by using two LSTMs for forward/backward direction and manually handling reversing them and concatenating the forward/backward directions. That is, it is doing it a slower way than via cudnn, but in a way we know should produce correct results.

src/imperative/imperative_utils.h
stephenrawls: good catch will do

tests/python/gpu/test_gluon_gpu.py
szha: I see. I mistook the use_sequence_length flag to be in rnn op. Still, whether the returned state is of the right step or not is not tested, which is also an important aspect of variable length RNN support. It may be hard to test it using RNN layer as reference.

stephenrawls: Ah, good point. I am only testing the returned output, not the return state.

I guess I could just loop over the batch elements one-by-one, passing them each in turn to the reference lstm. That way each input is correctly sized and I can easily grab the right return state.

Should I do that now or in the follow-on PR?

stephenrawls:
All unit tests except for the following passed:
```
AssertionError: Diff. between MXNet & TensorRT accuracy too high:
           MXNet = 99.150000, TensorRT = 99.140000
```

So I pushed a new minor change to trigger unit tests to run again and get this flaky unit test to pass.

tests/python/gpu/test_gluon_gpu.py
szha: I'm fine for it to be in a follow-up PR.

Issue #14157, opened 2019-02-14
Inconsistent handling for nan 


eric-haibin-lin:
```
>>> np.maximum([1],[np.nan])
array([nan])
>>> np.maximum([np.nan],[1])
array([nan])

>>> a = mx.nd.array([np.nan],ctx=mx.gpu()); b=mx.nd.array([3],ctx=mx.gpu())
>>> mx.nd.maximum(a,b)

[3.]
<NDArray 1 @gpu(0)>
>>> mx.nd.maximum(b,a)

[nan]
<NDArray 1 @gpu(0)>
```

ChaiBapchya:
Similarly found it for `mx.nd.minimum`
```
>>> mx.nd.minimum(b,a)

[nan]
<NDArray 1 @cpu(0)>
>>> mx.nd.minimum(a,b)

[3.]
<NDArray 1 @cpu(0)>
```

Issue #11115, opened 2018-05-31
ReLU Clips NaNs to Zero


imgemp:
## Description
In tracking down a bug, I found that the relu in mxnet.ndarray clips NaN’s to zero. Ideally, I would think the relu should propagate NaN’s so that bugs are not “covered up” for downstream operations.

## Environment info (Required)
mxnet version ‘1.2.0’
numpy version ‘1.14.3’

```
----------Python Info----------
Version      : 3.6.5
Compiler     : GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)
Build        : ('default', 'Apr 26 2018 08:42:37')
Arch         : ('64bit', '')
------------Pip Info-----------
Version      : 10.0.1
Directory    : /anaconda3/envs/mxnet/lib/python3.6/site-packages/pip
----------MXNet Info-----------
Version      : 1.2.0
Directory    : /anaconda3/envs/mxnet/lib/python3.6/site-packages/mxnet
Commit Hash   : b011eccd17d660b2f1ecf901e0db0007706df35c
----------System Info----------
Platform     : Darwin-16.7.0-x86_64-i386-64bit
system       : Darwin
node         : 8c85904326b5.ant.amazon.com
release      : 16.7.0
version      : Darwin Kernel Version 16.7.0: Tue Jan 30 11:27:06 PST 2018; root:xnu-3789.73.11~1/RELEASE_X86_64
----------Hardware Info----------
machine      : x86_64
processor    : i386
b'machdep.cpu.extfeatures: SYSCALL XD 1GBPAGE EM64T LAHF LZCNT PREFETCHW RDTSCP TSCI'
b'machdep.cpu.leaf7_features: SMEP ERMS RDWRFSGS TSC_THREAD_OFFSET BMI1 HLE AVX2 BMI2 INVPCID RTM SMAP RDSEED ADX IPT SGX FPU_CSDS MPX CLFSOPT'
[...]

PR #14262, opened 2019-02-27
Fix NaN value comparisons in relu, max and min ops

anirudhacharya:
## Description ##
Fix NaN comparisons in relu, max and min ops

Fixes https://github.com/apache/incubator-mxnet/issues/14157 
Fixes https://github.com/apache/incubator-mxnet/issues/11115

### Essentials ###

### Changes ###
- [x] change the way max, min and relu operators function to make sure nan values are handled properly. And added tests.

@anirudh2290 @apeforest

### DIFF
--- a/tests/python/unittest/test_ndarray.py
+++ b/tests/python/unittest/test_ndarray.py
@@ -1550,6 +1550,27 @@ def test_ndarray_is_nan():
     np.testing.assert_equal(output.asnumpy(), expected_output.astype(int))
     # astype since numpy functions default return type is boolean array instead of int
 
+@with_seed()
+def test_ndarray_nan_comparison():
+    random_dimensions = np.random.randint(2, 5)
+    random_shape = [np.random.randint(2, 5) for i in range(random_dimensions)]
+    data1 = mxnet.test_utils.rand_ndarray(random_shape,'default')
+    data2 = mxnet.test_utils.rand_ndarray(random_shape,'default')
+    data1[1][0] = np.NaN
+    data2[0][0] = np.NaN
+
+    nd_max = mx.nd.maximum(data1, data2)
+    np_max = np.maximum(data1.asnumpy(), data2.asnumpy())
+    np.testing.assert_equal(nd_max.asnumpy(), np_max)
+
+    nd_min = mx.nd.minimum(data1, data2)
+    np_min = np.minimum(data1.asnumpy(), data2.asnumpy())
+    np.testing.assert_equal(nd_min.asnumpy(), np_min)
+
+    nd_relu = mx.nd.relu(data1)
+    np_relu = np.maximum(data1.asnumpy(), 0)
+    np.testing.assert_equal(nd_relu.asnumpy(), np_relu)
+
 if __name__ == '__main__':
     import nose
     nose.runmodule()
--- a/tests/python/unittest/test_ndarray.py
+++ b/tests/python/unittest/test_ndarray.py
@@ -1571,6 +1571,14 @@ def test_ndarray_nan_comparison():
     np_relu = np.maximum(data1.asnumpy(), 0)
     np.testing.assert_equal(nd_relu.asnumpy(), np_relu)
 
+    data1.attach_grad()
+    with mx.autograd.record():
+        y = mx.nd.relu(data1)
+    y.backward()
+    data1_grad = data1.grad.asnumpy()
+    for i in (np.isnan(data1_grad))[1][0].flatten():
+        assert i == True
+
 if __name__ == '__main__':
     import nose
     nose.runmodule()
--- a/src/operator/mshadow_op.h
+++ b/src/operator/mshadow_op.h
@@ -780,7 +780,17 @@ namespace isnan_typed {
 
 MXNET_UNARY_MATH_OP_NC(relu, isnan_typed::IsNan(a) || (a > DType(0)) ? a : DType(0));
 
-MXNET_UNARY_MATH_OP_NC(relu_grad, a > DType(0) ? DType(1) : DType(0));
+/*! \brief used for computing gradient of relu operator */
+struct relu_grad : public mxnet_op::tunable {
+  template<typename DType>
+  MSHADOW_XINLINE static DType Map(DType a) {
+    if (isnan_typed::IsNan(a)) {
+      return a;
+    } else {
+      return a > DType(0) ? DType(1) : DType(0);
+    }
+  }
+};
 
 /*! \brief used for computing binary operator maximum */
 struct maximum : public mxnet_op::tunable {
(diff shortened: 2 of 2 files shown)


anirudhacharya:
@apeforest we need it because operations like this  ``mx.nd.relu(np.NaN*nd.ones(1), out)`` returns `[0.]` whereas activation functions are expected to propagate the `NaN` values. 

And `maximum` and `minimum` operators have inconsistent behavior w.r.t `NaN` values. 
```
>>> a = np.NaN*nd.ones(1)
>>> b = nd.zeros(1)

>>> nd.maximum(a,b)
[0.]
<NDArray 1 @cpu(0)>

>>> nd.maximum(b,a)
[nan]
<NDArray 1 @cpu(0)>
```

szha:
@anirudhacharya I'm not sure if it answers the question. Why do we need to start to support nan values, especially given the extra handling required for nan.

anirudhacharya:
@szha I ideally do not want `relu` operator to clip `NaN` values to zero, especially when I am trying to debug a model. 
And with regards to `maximum` and `minimum` it is not about 'starting to support' `nan` values but to fix inconsistent handling of `nan` values.

Pytorch's `relu` behavior -
```
>>> import torch
>>> import torch.nn as nn
>>> m = nn.ReLU()
>>> input = np.NaN * torch.ones(1)
>>> out = m(input)
>>> out
tensor([nan])
```

Also I found a related issue here - https://github.com/apache/incubator-mxnet/issues/14157

anirudhacharya:

szha:
@anirudhacharya thanks for the explanation. should relu grad deal with nan in a special way?

anirudhacharya:
@szha yes I think the relu grad should also be handled in a special way, thanks for pointing it out. 

Currently relu grad at `nan` returns a `0` by evaluating this expression 
```
MXNET_UNARY_MATH_OP_NC(relu_grad, a > DType(0) ? DType(1) : DType(0));
```
But `max(NaN, 0)` evaluates to `NaN` and that should translate to a relu grad value of `1` and not `0`. I will make the changes to fix it.

FYI - Here is an in depth conversation on `NaN` handling - https://github.com/JuliaLang/julia/issues/7866

adrianloy:
Nice PR! I also had a bug in my model, and because of relu activations removing NaNs it took me much longer to realize there is a bug. Behaviour should definitely be changed!

tests/python/unittest/test_ndarray.py
```diff
@@ -1550,6 +1550,27 @@ def test_ndarray_is_nan():
...
+    np.testing.assert_equal(nd_max.asnumpy(), np_max)
+
+    nd_min = mx.nd.minimum(data1, data2)
+    np_min = np.minimum(data1.asnumpy(), data2.asnumpy())
+    np.testing.assert_equal(nd_min.asnumpy(), np_min)
+
+    nd_relu = mx.nd.relu(data1)
+    np_relu = np.maximum(data1.asnumpy(), 0)
+    np.testing.assert_equal(nd_relu.asnumpy(), np_relu)
+
```
anirudh2290: can you also add a test for relu grad

anirudhacharya: done

szha:
@anirudhacharya I'm not sure if relu grad should act like that. As a sanity check, consider if nan is larger than or smaller than 0.

wkcn:
Could we add a new operator to check whether there are nan? It is used when debug.
When nan appears in the model, it always means that the model fails, and the output and gradient are not reliable.

anirudhacharya:
> @anirudhacharya I'm not sure if relu grad should act like that. As a sanity check, consider if nan is larger than or smaller than 0.

nan compared to any number is always `False`. `nan > 0` -> `False` and `nan < 0` -> `False`

Ref - https://stackoverflow.com/questions/49011370/nan-propagation-and-ieee-754-standard/49040225

But there are languages and libraries which consider nan to be greater than any number even np.inf.

szha:
@anirudhacharya it's not about comparison. nan is not in the domain of the function.

anirudhacharya:
@szha What you say makes sense, nan is not a number and hence not in the realm of comparison. So any occurrence either forward or backward will have to be propagated. 

Maybe pytorch is doing it wrong, but just for comparison's sake pytorch seems to treat the gradient of relu @ NaN as equal to `1` - 
```
>>> import torch
>>> import numpy as np
>>> a = np.NaN * torch.ones(1)
>>> a.requires_grad_(True)
tensor([nan], requires_grad=True)
>>> m = torch.nn.ReLU()
>>> out = m(a)
>>> out.backward()
>>> a.grad
tensor([1.])
```

My main motivation when I first made changes to the relu forward behavior was that the operator silently clipping NaN values was very misleading while trying to build or debug models.

I am open to suggestions on how relu gradient should behave, it would seem there is no single consensus on this and each community/library decide things for themselves .

anirudhacharya:
> Could we add a new operator to check whether there are nan? It is used when debug.
> When nan appears in the model, it always means that the model fails, and the output and gradient are not reliable.

I think you are looking for this - http://mxnet.incubator.apache.org/api/python/ndarray/contrib.html?highlight=isnan#mxnet.ndarray.contrib.isnan

anirudhacharya:
I modilied relu grad to also propagate `NaN` values. As discussed above, since `NaN` does not exist in the domain of the function, it can also not be mapped to any element in the range of the function, hence the output is also `NaN`.

szha:
@anirudhacharya one last thing, could you measure the performance before and after this change? This change is nonetheless necessary, still it would better if we could anticipate any performance change from this. Thanks.

anirudhacharya:
#### Run Mode: Before --> After ( time in ms)

'Whole CPU run: '    -  0.843163 -->  0.864071
'Forward CPU run: ' -  0.016467 --> 0.043115
'Whole GPU run: '    -  0.460900 --> 0.480667
'Forward GPU run: ' -  0.058783 --> 0.059333

script used

```python
import mxnet as mx                                                                                                                                                    
import numpy as np
from mxnet.test_utils import check_speed

ctx = mx.cpu()
#ctx = mx.gpu(0)

sample_data = mx.nd.ones((3, 500, 500), ctx=ctx)
sample_data[0] = -1. 
sample_data[1] = np.NaN
sample = mx.sym.Variable("sample")
relu_sym = mx.sym.relu(data=sample)

print("Whole CPU run: ", check_speed(relu_sym, location={"sample": sample_data}, ctx=ctx, N=int(1e5), typ="whole"))
print("Forward CPU run: ", check_speed(relu_sym, location={"sample": sample_data}, ctx=ctx, N=int(1e5), typ="forward"))
#print("Whole GPU run: ", check_speed(relu_sym, location={"sample": sample_data}, ctx=ctx, N=int(1e5), typ="whole"))
#print("Forward GPU run: ", check_speed(relu_sym, location={"sample": sample_data}, ctx=ctx, N=int(1e5), typ="forward"))
```

PR #14113, opened 2019-02-11
Enhance subgraph API

ZhennanQin:
@reminisce @zheng-da @apeforest @szha  @pengzhao-intel
PR Major change:

* Introduce `SubgraphSelectorV2` that accepts `BiDirectedNode` instead of `nnvm::Node` to get full context of visited node. `BiDirectedNode` is double linked version of  `nnvm::Node`, which can list the output consumer of node. This is very useful to help make selection which has output requirement. Wrapper class `SubgraphSelectorV2Bridge` is added to convert `SubgraphSelector` to `SubgraphSelectorV2`, which allows developer to continue using old `SubgraphSelector`. So this feature is fully back-compatible.

* Introduce `kAdjust` mode for `SubgraphProperty` to allow adjusting network parameters instead of create new node. New API `AdjustSubgraphNode()` is added to  `SubgraphProperty` for this, which can adjust selected nodes attributes. This is useful when we don't want to create new subgraph node, but to do some modification on selected nodes. Quantized scale alignment optimization is pending on this feature.

* Change the behavior of PartitionGraph to create single subgraph node for nodes isolating with each other. https://cwiki.apache.org/confluence/display/MXNET/Enable+Operator+Level+Parallelism+under+Subgraph is depending on this change.

* Add new subgraph pass "quantized scale alignment optimization" as an example for new subgraph API.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
[...]

PR #14277, opened 2019-02-28
Enhance PartitionGraph

ZhennanQin:
## Description ##
Extracted from https://github.com/apache/incubator-mxnet/pull/14113. This PR covers:

* Add `inference_only` attr support when `SubgraphProperty` is created to indicate that this pass should be used for inference only.
* Allow registering multiple subgraph pass within same backend names.
* Refactor the way to run PartitionGraph in simple bind stage to ensure any graph nodes reordering can be correctly handled.

This PR is the full version of https://github.com/apache/incubator-mxnet/pull/14276.
@xinyu-intel @pengzhao-intel @TaoLv @reminisce @zheng-da

### Essentials ###

### Changes ###

## Comments ##


anirudhacharya:

src/operator/subgraph/subgraph_property.h
```diff
@@ -200,7 +197,7 @@ typedef dmlc::ThreadLocalStore<std::unordered_map<std::string, std::unordered_se
 
 #define MXNET_REGISTER_SUBGRAPH_PROPERTY(Name, SubgraphPropertyType) \
   static DMLC_ATTRIBUTE_UNUSED auto __make_ ## SubgraphPropertyType ## _ ## Name ## __ = \
-    SubgraphPropertyRegistry::Get()->__REGISTER_OR_GET__(#Name, &SubgraphPropertyType::Create)
+    SubgraphPropertyRegistry::Get()->__REGISTER__(#Name, &SubgraphPropertyType::Create)
```
reminisce: Any reason of changing this? I remember there was a discussion in the first subgraph API PR and using `__REGISTER__` may result in terminating the program.
https://github.com/apache/incubator-mxnet/pull/12157#discussion_r211713108

src/operator/subgraph/mkldnn/mkldnn_subgraph_property.cc
```diff
@@ -0,0 +1,34 @@
...
+#if MXNET_USE_MKLDNN == 1
+
+#include "mkldnn_conv_property.h"
+#include "mkldnn_post_quantize_conv_property.h"
+
+namespace mxnet {
+namespace op {
+
+MXNET_REGISTER_SUBGRAPH_PROPERTY(MKLDNN, SgMKLDNNConvProperty);
+MXNET_REGISTER_SUBGRAPH_PROPERTY(MKLDNN_POST_QUANTIZE, SgMKLDNNConvPostQuantizeProperty);
```
anirudh2290: Can you add a test for this PR. If I understand correctly, the intention is to register multiple properties, under simple backend name. Why are we using MKLDNN and MKLDNN_POST_QUANTIZE here then ? Shouldn't the name be MKLDNN for both ?

src/operator/subgraph/subgraph_property.h
ZhennanQin: Because in this PR,  we allow to register same backend multiple times, and also possible to register same property multiple times. `__REGISTER__` won't cause terminating anymore.

src/operator/subgraph/mkldnn/mkldnn_subgraph_property.cc
ZhennanQin: Yes, this PR is to support register multiple properties under same backend name. They could be all registered under `MKLDNN`. But we decided to separate them because, `MKLDNN` is to cover all subgraph pass that will be used in symbol bind stage, that is, it is designed to optimize common model into MKLDNN specific model. the optimization is suitable for most of networks, and user can use it easily with `export MXNET_SUBGRAPH_BACKEND=MKLDNN `.
`MKLDNN_POST_QUANTIZE` is for optimizing quantized model. it should be used in quantization flow only, by inserting `qsym = qsym.get_backend_symbol('MKLDNN_POST_QUANTIZE')` explicitly.

src/operator/subgraph/subgraph_property.h
anirudh2290: Removing the check on line 181: https://github.com/apache/incubator-mxnet/pull/14277/files#diff-4db2c685581a8f376abff8d6ac908e1cL181 will not prevent a subgraph property from registering more than once. I think the intention was to create only one shared pointer to one subgraph property object and use it everywhere. We should stick to doing this. One way is to modify the subgraph registry to create unordered_map<string, map<string, unoredered_map<string, createfunc>> . Then you register each subgraph property with top level name and second level name to search within the inner map.

ZhennanQin: Basically, the main change here is, replace `std::unordered_map<std::string, SubgraphPropertyCreateFn>` with `std::unordered_map<std::string, std::vector<SubgraphPropertyCreateFn>>`  as the type of prop_fn_map_. 
Using `std::vector` instead of `unoredered_map`  as the inner type because, we need to keep `SubgraphPropertyCreateFn` in a user defined order, while unoredered_map can't guarantee it.

anirudh2290: @ZhennanQin Here is an example: https://github.com/anirudh2290/mxnet/tree/fix_bind . I created an additional.h which has a new subgraph property (test) and selector. It is included from multiple source files and thus there are multiple subgraph properties in prop_fn["test"]. Partition Graph is also called twice instead of once. Every unique subgraph property should register only once.

ZhennanQin: I understand your point.  As quantized FC is merged into master. I rebased this PR to show what we needed here:
https://github.com/apache/incubator-mxnet/blob/f3b5e980581c94c9d36b5a9f02cbb923be9ba7bd/src/operator/subgraph/mkldnn/mkldnn_subgraph_property.cc#L30-L34

In our design, we will register same backend name many times with different subgraph properties.  Property will run inthe order they registered. 

It's possible to register unique subgraph property many times for different backends or even for unique backend. Below registration is possible in future:

```
MXNET_REGISTER_SUBGRAPH_PROPERTY(MKLDNN, SgMKLDNNConvProperty);
MXNET_REGISTER_SUBGRAPH_PROPERTY(MKLDNN, SgMKLDNNFCProperty);
MXNET_REGISTER_SUBGRAPH_PROPERTY(MKLDNN_POST_QUANTIZE, SgMKLDNNConvProperty);
MXNET_REGISTER_SUBGRAPH_PROPERTY(MKLDNN_POST_QUANTIZE, SgMKLDNNFCProperty);
MXNET_REGISTER_SUBGRAPH_PROPERTY(MKLDNN_POST_QUANTIZE, SgMKLDNNConvPostQuantizeProperty);
MXNET_REGISTER_SUBGRAPH_PROPERTY(MKLDNN_POST_QUANTIZE, SgMKLDNNFCPostQuantizeProperty);
```
As you seen, `SgMKLDNNConvProperty` is registered twice for different backend names.

Also, it's possible to have this in future:
```
MXNET_REGISTER_SUBGRAPH_PROPERTY(MKLDNN, SgMKLDNNConvProperty);
MXNET_REGISTER_SUBGRAPH_PROPERTY(MKLDNN, SgMKLDNNFCProperty);
MXNET_REGISTER_SUBGRAPH_PROPERTY(MKLDNN, SgMKLDNNConversionProperty);
MXNET_REGISTER_SUBGRAPH_PROPERTY(MKLDNN, SgMKLDNNConvProperty);
```
We need to register `SgMKLDNNConvProperty` twice within `MKLDNN`, and we need to run each property as the order registered. The reason might be, `SgMKLDNNConversionProperty` will do further optimization on fused conv, and this conversion might expose more opportunities for `SgMKLDNNConvProperty`, so we need to run `SgMKLDNNConvProperty` again.

As backend implements more and more subgraph pass, it's highly possible to have above requirement. We need to design for it in advance.

anirudh2290: In my code, I use MXNET_REGISTER_SUBGRAPH_PROPERTY only once here: https://github.com/anirudh2290/mxnet/blob/fix_bind/src/operator/subgraph/additional.h#L66 . But include in two different source files( https://github.com/anirudh2290/mxnet/blob/fix_bind/src/operator/subgraph/default_subgraph_property.cc#L23 , https://github.com/anirudh2290/mxnet/blob/fix_bind/src/operator/subgraph/mkldnn/mkldnn_subgraph_property.cc#L24) which will be part of different translation units. As you can see, it will create two properties and run PartitionGraph twice. This is not intended and different from your example where you have explicitly called MXNET_REGISTER_SUBGRAPH_PROPERTY twice with the same name and property.

ZhennanQin: @anirudh2290 Putting MXNET_REGISTER_SUBGRAPH_PROPERTY in header file is not a good usecase. If developer include that header twice, then I think he just wants to register that properties twice for same backend. I admited this may cause misleading if developer doesn't know the details behind MXNET_REGISTER_SUBGRAPH_PROPERTY . But I have no idea how to avoid this. Do you have any good suggestion? Please consider the requirement I listed in previous comment.

anirudh2290: We can have a different macro for registering property versus registering backend. For example:
```
MXNET_REGISTER_SUBGRAPH_PROPERTY(DefaultSubgraphProperty)
MXNET_REGISTER_BACKEND(default, DefaultSubgraphProperty, DefaultSubgraphProperty)
```

I don't mind sticking to the limitation of including MXNET_REGISTER_SUBGRAPH_PROPERTY to only from .cc files but we need to document this well otherwise this can lead to difficult to trace bugs.

ZhennanQin: @anirudh2290 I've consider the way you suggested, including other formats like,
```
MXNET_REGISTER_BACKEND(default, {DefaultSubgraphProperty, DefaultSubgraphProperty})
```
or
```
MXNET_REGISTER_BACKEND(default, DefaultSubgraphProperty, 0)
MXNET_REGISTER_BACKEND(default, DefaultSubgraphProperty, 1)
```
But I give them up because of different reasons. The first problem is, `DefaultSubgraphProperty` is only the name of class, but not a object or function. So it can't be inserted in a vector. The correct line is,
```
MXNET_REGISTER_BACKEND(default, {DefaultSubgraphProperty::Create(), DefaultSubgraphProperty::Create()})
```
But it's not elegant.

For your suggestion,
```
MXNET_REGISTER_BACKEND(default, DefaultSubgraphProperty, DefaultSubgraphProperty)
```
It should work if we define `MXNET_REGISTER_BACKEND` with 3 arguments. But it's not scalable because if we want to register more property for this backend, we need to write more `MXNET_REGISTER_BACKEND` implementation with more number of arguments.

For
```
MXNET_REGISTER_BACKEND(default, DefaultSubgraphProperty, 0)
MXNET_REGISTER_BACKEND(default, DefaultSubgraphProperty, 1)
```
This can help to avoid register property twice unexceptly from header as it defines the execution order. But this will make code a bit ugly.

Personally, I prefer the current implementation although we need to use it carefully, like not putting `MXNET_REGISTER_SUBGRAPH_PROPERTY` in header. But I don't mind to change to other idea I listed above. Then you can make the final decision now, and I will follow it.

anirudh2290: I think I am okay with the REGISTER only in cc limitation as long as it is well documented here: http://mxnet.incubator.apache.org/versions/master/tutorials/c++/subgraphAPI.html?highlight=subgraph#subgraph-api . 

Also, we need to make sure that we call Create only once for each property even for this use case:
```
MXNET_REGISTER_SUBGRAPH_PROPERTY(default, DefaultSubgraphProperty)
MXNET_REGISTER_SUBGRAPH_PROPERTY(default, DefaultSubgraphProperty)
```
This can be done by adding a map of subgraph property name to the corresponding subgraph property pointer. For this passing the property name along with the create fn is required when calling __REGISTER__. The create gets called and returned subgraph property ptr gets added in __REGISTER__ itself. We can then reuse the already added subgraph property ptr in CreateSubgraphProperty.

ZhennanQin: Code is refactored as your comments. Please review again. Thanks.

src/operator/subgraph/subgraph_property.h
```diff
@@ -205,9 +204,14 @@ class SubgraphPropertyRegistry {
...
 
+#define DECLEARPROPERTY1(NAME, SubgraphPropertyType, X) \
+  DMLC_ATTRIBUTE_UNUSED auto __make_##SubgraphPropertyType##_##Name##_##X##__
+#define DECLEARPROPERTY(NAME, SubgraphPropertyType, X) \
+  DECLEARPROPERTY1(NAME, SubgraphPropertyType, X)
+
 #define MXNET_REGISTER_SUBGRAPH_PROPERTY(Name, SubgraphPropertyType) \
-  static DMLC_ATTRIBUTE_UNUSED auto __make_ ## SubgraphPropertyType ## _ ## Name ## __ = \
-    SubgraphPropertyRegistry::Get()->__REGISTER_OR_GET__(#Name, &SubgraphPropertyType::Create)
+  static DECLEARPROPERTY(Name, SubgraphPropertyType, __LINE__) =     \
```
anirudh2290: nit: did you mean DECLAREPROPERTY ?

src/operator/subgraph/mkldnn/mkldnn_fc_property.h
```diff
@@ -120,19 +120,19 @@ class SgMKLDNNFCSelector : public SubgraphSelector {
...
-    } else {
-      LOG(INFO) << "Start to execute MKLDNN FullyConnected optimization pass.";
-    }
   }
 
   static SubgraphPropertyPtr Create() {
-    return std::make_shared<SgMKLDNNFCProperty>();
+    static const std::string &name = "MKLDNN FullyConnected optimization pass";
+    if (dmlc::GetEnv("MXNET_DISABLE_MKLDNN_FC_OPT", 0)) {
+      LOG(INFO) << name << "is disabled.";
```
anirudh2290: nit space: " is disabled."

src/operator/subgraph/mkldnn/mkldnn_fc_property.h
```diff
@@ -120,19 +120,19 @@ class SgMKLDNNFCSelector : public SubgraphSelector {
...
-      LOG(INFO) << "MKLDNN FullyConnected optimization pass is disabled.";
-    } else {
-      LOG(INFO) << "Start to execute MKLDNN FullyConnected optimization pass.";
-    }
   }
 
   static SubgraphPropertyPtr Create() {
-    return std::make_shared<SgMKLDNNFCProperty>();
+    static const std::string &name = "MKLDNN FullyConnected optimization pass";
+    if (dmlc::GetEnv("MXNET_DISABLE_MKLDNN_FC_OPT", 0)) {
```
anirudh2290: can you document all these variables been documented in env var docs.

pengzhao-intel:
@anirudh2290 thanks a lot for the review :)

@reminisce could you help to review again?

src/operator/subgraph/subgraph_property.h
ZhennanQin: Fixed

src/operator/subgraph/mkldnn/mkldnn_fc_property.h
ZhennanQin: Fixed

src/operator/subgraph/mkldnn/mkldnn_fc_property.h
ZhennanQin: Added

src/operator/subgraph/subgraph_property.h
```diff
@@ -205,9 +204,14 @@ class SubgraphPropertyRegistry {
 typedef dmlc::ThreadLocalStore<std::unordered_map<std::string, std::unordered_set<std::string>>>
   SubgraphPropertyOpNameSet;
 
+#define DECLAREPROPERTY1(NAME, SubgraphPropertyType, X) \
```
reminisce: DECLAREPROPERTY1 -> DECLARE_PROPERTY_EX

src/operator/subgraph/subgraph_property.h
```diff
@@ -205,9 +204,14 @@ class SubgraphPropertyRegistry {
...
+  DMLC_ATTRIBUTE_UNUSED auto __make_##SubgraphPropertyType##_##Name##_##X##__
+#define DECLAREPROPERTY(NAME, SubgraphPropertyType, X) \
```
reminisce: DECLAREPROPERTY -> DECLARE_PROPERTY

src/executor/graph_executor.cc
```diff
@@ -1500,91 +1473,220 @@ static nnvm::Symbol PartitionGraph(const nnvm::Symbol& src,
...
+  std::vector<mxnet::op::SubgraphPropertyPtr> properties =
+      op::SubgraphPropertyRegistry::Get()->CreateSubgraphProperty(prop_name);
+  for (auto subgraph_prop : properties) {
+    if (subgraph_prop->HasAttr("inference_only") &&
+        subgraph_prop->GetAttr<bool>("inference_only") == true) {
+      if (need_grad) {
+        auto full_name = subgraph_prop->HasAttr("property_name")
+                             ? subgraph_prop->GetAttr<std::string>("property_name")
+                             : prop_name;
+        LOG(INFO) << "Skip subgraph " << full_name << " as it requires `grad_req=null`.";
```
reminisce: "skip subgraph" -> "skip partitioning graph with subgraph property " ...

ZhennanQin:
CI is blocked by https://github.com/apache/incubator-mxnet/issues/14507 and should be fixed https://github.com/apache/incubator-mxnet/pull/14504.

PR #11254, opened 2018-06-13
add blocklist

piiswrong:
## Description ##
(Brief description on what this PR is about)

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #14284, opened 2019-02-28
Allow gluon blocks to register lists of blocks as attributes

ifeherva:
## Description ##
Currently gluon Blocks (and Hybridblocks) register attributes' parameters if they are of type Parameter or Block. However, in certain cases one wants to add a list of blocks (or parameters) whose size is only known in runtime.

I am proposing here a small improvement, please advise if this idea even makes sense or is already covered by a similar functionality.

### DIFF
--- a/python/mxnet/gluon/block.py
+++ b/python/mxnet/gluon/block.py
@@ -205,6 +205,9 @@ class Block(object):
                 "If you want to share parameters between blocks, please set " \
                 "'params' at Block construction instead."
             self._reg_params[name] = value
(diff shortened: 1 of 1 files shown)


ThomasDelteil:
See https://github.com/apache/incubator-mxnet/pull/11254 for a list-like container that supports setting / appending items while registering children, which is not currently taken into account in your proposed change.

anirudhacharya:

wkcn:
The PR is good.
I do not know whether there is a conflict between this PR and HybridConcurrent.

For example, 
```python
a = [Block1(), Block2()]
HybridConcurrent(a)
```

PR #14188, opened 2019-02-18
Disables  test_random.test_exponential_generator

perdasilva:
## Description ##
Disables  test_random.test_exponential_generator due to isse  #14187 

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

PR #13498, opened 2018-12-01
Fix flaky test test_random:test_randint_generator

ChaiBapchya:
## Description ##
Fixes #13446 flaky test test_random:test_randint_generator 
Uses the PR #13543 to get rid of the fix seed and make the test robust
### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [X] Changes are complete (i.e. I finished coding on this PR)
- [X] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [X] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

## Comments ##
- Created the alpha variable to ensure consistency
- Scipy uses 0.01 for discrete distribution here - https://github.com/scipy/scipy/blob/9f12af697763fb5f9767d5cb1280ce62456a3974/scipy/stats/tests/test_discrete_basic.py#L45
Hence for testing `randint` (discrete uniform distribution) specifically passed alpha=0.01
- Due to random factor lacking robustness had to set fixed seed to prevent CI from failing

PR #14287, opened 2019-02-28
FIX: flaky test exponential generator

access2rohit:
## Description ##
change success_rate to 0.20 for test_exponential_generator since 0.25 much stricter and fails random exponential generator tests.

### Essentials ###

## Testing ##
Command:
ubuntu@ip-172-31-82-110 ~/Workspace/mxnet/tests/python/unittest (master) $ MXNET_TEST_COUNT=10000 nosetests --logging-level=DEBUG --verbose -s test_random.py:test_exponential_generator
Output:
[DEBUG] 10000 of 10000: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=291062812 to reproduce.
ok

Ran 1 test in 46911.498s

OK

### DIFF
--- a/tests/python/unittest/test_random.py
+++ b/tests/python/unittest/test_random.py
@@ -560,12 +560,12 @@ def test_exponential_generator():
         for scale in [0.1, 1.0]:
             buckets, probs = gen_buckets_probs_with_ppf(lambda x: ss.expon.ppf(x, loc=0, scale=scale), 5)
             generator_mx = lambda x: mx.nd.random.exponential(scale, shape=x, ctx=ctx, dtype=dtype).asnumpy()
-            verify_generator(generator=generator_mx, buckets=buckets, probs=probs)
+            verify_generator(generator=generator_mx, buckets=buckets, probs=probs, success_rate=0.20)
             generator_mx_same_seed = \
                 lambda x: np.concatenate(
                     [mx.nd.random.exponential(scale, shape=x // 10, ctx=ctx, dtype=dtype).asnumpy()
                      for _ in range(10)])
-            verify_generator(generator=generator_mx_same_seed, buckets=buckets, probs=probs)
+            verify_generator(generator=generator_mx_same_seed, buckets=buckets, probs=probs, success_rate=0.20)
 
 @with_seed()
 def test_poisson_generator():


access2rohit:
@anirudh2290 @eric-haibin-lin @apeforest  Can you please review ?

python/mxnet/test_utils.py
```diff
@@ -1933,7 +1933,7 @@ def chi_square_check(generator, buckets, probs, nsamples=1000000):
     _, p = ss.chisquare(f_obs=obs_freq, f_exp=expected_freq)
     return p, obs_freq, expected_freq
 
-def verify_generator(generator, buckets, probs, nsamples=1000000, nrepeat=5, success_rate=0.25, alpha=0.05):
+def verify_generator(generator, buckets, probs, nsamples=1000000, nrepeat=5, success_rate=0.15, alpha=0.05):
```
yuxihu: why did we change the value from 0.15 to 0.25?

mseth10: Actually it's changed from 0.25 to 0.15, but I have the same question. Why this change?

apeforest: Same question. The comment in this method says "25%" success rate. Why do we change it?

access2rohit: @yuxihu @mseth10 @apeforest 
It was changed in this PR: https://github.com/apache/incubator-mxnet/pull/13498/files
Even though It is mentioned in the comments that success rate should be 25%. There isn't any data to support why it should be kep at 25%. It could have been a typo too back then. SO, reverting it to original value seemed more logical.

access2rohit: My concern is due to increasing the strinctness of this check for all generators it might caouse them to become flaky. I can also explicitly relax the criteria for only this generator and we can observe the behaviour of others and make changes accordingly. Setting it to a fixed value for all the generators doesn't seem like a good idea since all the random generators are different.

yuxihu: It looks like @ChaiBapchya increased it from 0.15 to 0.25 in the PR in order to fix another flaky test. The question now: if you change back to 0.15, will the other test become flaky again?

access2rohit: @yuxihu : Actually No. He in fact made the check more strict. I am simply relaxing it. Therefore it shouldn't make another test flaky. But to make our tests better what I am suggesting is that i let it stay at 0.25 as default and explicitly pass 0.15 for exponential generator check. If others don't fail maybe only this one needed adjustment. Do you think this approach is better ?

anirudh2290: I think its fine to revert it to 0.25 to reduce flakiness. it wont make the other test flaky since this change is decreasing the failure chances.

anirudhacharya:

python/mxnet/test_utils.py
yuxihu: @access2rohit I like your idea of passing 0.15 for exponential generator check.

apeforest: @access2rohit Please make the change as you commented: make 0.25 as default and pass in 0.15 for exponential generator check.

access2rohit: @yuxihu @apeforest @mseth10 @anirudh2290 : I have updated successrate to 20% which is less stricter but not too loose a check. Currently running it 10k times and will update the PR once they succeed

access2rohit: All tests passed with success_rate=0.20

access2rohit:
@anirudh2290 @apeforest can you review as well ?

access2rohit:
@yuxihu : It still hasn't been merged. #14188

PR #14290, opened 2019-02-28
Optimize NMS

ptrendx:
## Description ##
Current Box NMS forward path is very slow on the GPU, because it launches potentially thousands very tiny kernels with exposed launch latency. This PR introduces a new kernel for this slow part of the NMS op, reducing the time significantly.

On single GPU, training Faster RCNN from GluonCV model zoo, the time to perform this portion of NMS improved by 50x, from 100 ms to 2.1 ms.

### Essentials ###

### Changes ###
- [x] GPU path of Box NMS forward op changed from previous nms_impl kernel to the new optimized kernel 

## Comments ##
- While this was the biggest problem with NMS op, there are other ones, now exposed because the biggest offender got significantly faster. Namely: using thrust for sorting introduces syncs, allocations and deallocations in the op; there are 2 Slice operations that in the case of FRCNN take together ~18ms (which is 20% of the end to end time in FP16 training). I don't have time currently to try to fix them though :-(.  @zhreshold FYI


src/operator/contrib/bounding_box-inl.cuh
```diff
@@ -57,6 +60,226 @@ int FilterScores(mshadow::Tensor<gpu, 1, DType> out_scores,
...
+              mshadow::Tensor<gpu, 1, int32_t>* sorted_index,
+              mshadow::Tensor<gpu, 1, int32_t>* batch_start,
+              mshadow::Tensor<gpu, 3, DType>* buffer,
+              mshadow::Tensor<gpu, 1, DType>* areas,
+              int num_elem, int width_elem,
+              int coord_start, int id_index,
+              float threshold, bool force_suppress,
+              int in_format) {
+  using namespace mxnet_op;
+  constexpr int THRESHOLD = 1024;
```
zhreshold: one question, why launch bound set to 512 while threshold is 1024?

zhreshold:
@pengzhao-intel @ZhennanQin This is the fix I mentioned to fix nms perf on GPU

src/operator/contrib/bounding_box-inl.cuh
ptrendx: To have bigger register count available and to increase the useful work done by the threads, since I need to keep all the threads around anyway for syncthreads calls. Each thread processes threshold / block_size elements.

zhreshold:
lgtm, I will leave it open for 24hr before merging this in case others have more comments

ptrendx:
Just as an update - I trained FasterRCNN from GluonCV to completion with this new kernel and got the same accuracy as reported in GluonCV (37/57.8/39.9 mAP).

zhreshold:
Awesome, I am merging this now

PR #14313, opened 2019-03-04
compatibility with opencv4

wkcn:
## Description ##
To compatible with multiple versions of OpenCV, define the old OpenCV macros to support OpenCV4.

In OpenCV4, `opencv2/imgproc/imgproc_c.h` and `opencv2/videoio/videoio_c.h` define these macros, however I couldn't include them since it would bring redefinition problems.

### Essentials ###

### Changes ###
- [x] Add a new file `src/io/opencv_compatibility.h` to define the OpenCV macros.
- [x] Modify `Makefile` to find the path of `opencv4`

## Comments ##
There is a similar PR https://github.com/apache/incubator-mxnet/pull/13559, Thank @daleydeng

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,9 +147,11 @@ endif
 # setup opencv
 ifeq ($(USE_OPENCV), 1)
 	CFLAGS += -DMXNET_USE_OPENCV=1
-	LDFLAGS += -lopencv_core -lopencv_imgcodecs -lopencv_imgproc
 	ifneq ($(USE_OPENCV_PATH), )
 		CFLAGS += -I$(USE_OPENCV_PATH)
+		ifneq ($(USE_OPENCV_LIB_PATH), )
+			LDFLAGS += -L$(USE_OPENCV_LIB_PATH)
+		endif
 	else
 		ifeq ("$(shell pkg-config --exists opencv4; echo $$?)", "0")
 			OPENCV_LIB = opencv4
@@ -157,7 +159,10 @@ ifeq ($(USE_OPENCV), 1)
 			OPENCV_LIB = opencv
 		endif
 		CFLAGS += $(shell pkg-config --cflags $(OPENCV_LIB))
+		LDFLAGS += $(shell pkg-config --libs-only-L $(OPENCV_LIB))
 	endif
+	LDFLAGS += -lopencv_core -lopencv_imgcodecs -lopencv_imgproc
+
 	BIN += bin/im2rec
 else
 	CFLAGS += -DMXNET_USE_OPENCV=0
--- a/Makefile
+++ b/Makefile
@@ -129,7 +129,7 @@ ifdef CAFFE_PATH
 endif
 
 ifndef LINT_LANG
-	LINT_LANG="all"
+	LINT_LANG = "all"
 endif
 
 ifeq ($(USE_MKLDNN), 1)
--- a/Makefile
+++ b/Makefile
@@ -147,10 +147,16 @@ endif
 # setup opencv
 ifeq ($(USE_OPENCV), 1)
 	CFLAGS += -DMXNET_USE_OPENCV=1
-	ifneq ($(USE_OPENCV_PATH), )
-		CFLAGS += -I$(USE_OPENCV_PATH)
-		ifneq ($(USE_OPENCV_LIB_PATH), )
-			LDFLAGS += -L$(USE_OPENCV_LIB_PATH)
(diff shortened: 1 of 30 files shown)


Makefile
```diff
@@ -146,11 +146,18 @@ endif
 
 # setup opencv
 ifeq ($(USE_OPENCV), 1)
-	CFLAGS += -DMXNET_USE_OPENCV=1 $(shell pkg-config --cflags opencv)
-	LDFLAGS += $(filter-out -lopencv_ts, $(shell pkg-config --libs opencv))
+	ifeq ("$(shell pkg-config --exists opencv4; echo $$?)", "0")
```
szha: there needs to be a way for user of make to override whatever is found by pkg-config.

szha: not a requirement but nice to have. libturbojpeg has two variables for this purpose: https://github.com/apache/incubator-mxnet/blob/master/Makefile#L460-L461
Also, many of what's returned in pkg-config of opencv are not necessary for mxnet. The only necessary modules in opencv for mxnet are core, imgcodecs, and imgproc. As a result, I turn off most of the build options for opencv in mxnet pip package: https://github.com/apache/incubator-mxnet/blob/master/tools/dependencies/opencv.sh#L39

wkcn: I see. I will modify it. Thanks!

wkcn: OpenCV2 doesn't have libopencv_imgcodecs.so, and I do not know how to determine whether libopencv_imgcodecs.so exists or the version of OpenCV is 2.

Makefile
```diff
@@ -146,11 +146,26 @@ endif
 
 # setup opencv
 ifeq ($(USE_OPENCV), 1)
-	CFLAGS += -DMXNET_USE_OPENCV=1 $(shell pkg-config --cflags opencv)
-	LDFLAGS += $(filter-out -lopencv_ts, $(shell pkg-config --libs opencv))
+	CFLAGS += -DMXNET_USE_OPENCV=1
+	ifneq ($(USE_OPENCV_PATH), )
```
szha: Thanks for the change. Remember to update the make configs in `make/` too :)

anirudhacharya:

szha:
Seems like this time the CI is reporting an actual problem. @wkcn could you take a look?

wkcn:
@szha Yes. OpenCV2 does not have libopencv_imgcodecs.so, but OpenCV3/4 need it.
I do not know how to address the problem.
Maybe we should add a new flag OPENCV_VERSION.

wkcn:
I do not know what OpenCV shared library we did not link. It raises undefined reference in unit-cpu CI.

szha:
@wkcn it might be that the system has an opencv2 installation while the build is statically linking opencv3.

wkcn:
@szha I see. I used wrong linking order. I will fix it. Thank you!

dbsxdbsx:
As far as I know, when I tried to build mxnet on win10 with cuda 10.1. It failed with compling error.
```
incomplete type is not allowed	mxnet	c:\opencv\build\include\opencv2\core\cvstd_wrapper.hpp	45	

```

wkcn:
@dbsxdbsx Could you please open an issue and provide more error message?
I only fix the compatibility on Linux in this PR.

dbsxdbsx:
@wkcn, thanks. I posted it #14385.

Makefile
```diff
@@ -146,11 +146,33 @@ endif
 
 # setup opencv
 ifeq ($(USE_OPENCV), 1)
-	CFLAGS += -DMXNET_USE_OPENCV=1 $(shell pkg-config --cflags opencv)
-	LDFLAGS += $(filter-out -lopencv_ts, $(shell pkg-config --libs opencv))
+	CFLAGS += -DMXNET_USE_OPENCV=1
+	ifneq ($(USE_OPENCV_INC_PATH), NONE)
```
reminisce: This can be problematic when `USE_OPENCV_INC_PATH` is not defined in `config.mk`, because it's an empty string at the time. In that situation, no correct paths to headers and libs of opencv are actually added to `CFLAGS` and `LDFLAGS`.

szha: @reminisce good catch. @wkcn how about we add a default value for empty value here to help with cases where people already have old config.mk without the new variables?

wkcn: @reminisce @szha Hi! I have fixed it in the PR https://github.com/apache/incubator-mxnet/pull/14424

Issue #14253, opened 2019-02-25
[RFC] Introducing NumPy-compatible coding experience into MXNet


reminisce:
## Motivation
Today deep learning scientists spend majority of their time on data processing, debugging tensor algorithms, and tuning model parameters, instead of architecting models from scratch by themselves as a result from the abundant pre-trained models existing in many deep learning model zoos. This has highlighted the usability of tensor APIs as a key factor for a framework to be widely adopted.

MXNet was firstly designed with the focus on memory efficiency, computation throughput and scalability. The usability problems begin to show up nowadays when more and more models demonstrate dynamic natures, e.g. unknown-shape tensors before runtime, control flow depending on a runtime result, etc. Here we highlight the most frequent complaints about usability from users.
- Scalar tensors (aka zero-dim tensors) are not supported. For example, given `a = [0, 1, 2]`, `a[1]` will generate an `NDArray` of shape `(1,)`, instead of `()` as in NumPy.
- Zero-size tensor is not supported. For example, a tensor of shape `(0, 16, 256)` cannot be passed to an operator, because our system currently treats 0, the first dimension size, as unknown, rather than a concrete number.
- Many operators' signatures and functionality are not NumPy compatible, e.g. `nd.dot` vs. `np.dot`, `nd.concatenate` vs. `np.concatenate`, etc.
- Many NumPy operators are missing. See the [reference link](https://github.com/apache/incubator-mxnet/issues?q=is%3Aissue+numpy+label%3ANumpy) to GitHub issues.
- Operators whose outputs' shapes can only be determined at runtime are not supported, e.g. `data[data < 0]` cannot run.
- Diverged programming experience due to the separation of imperative and symbolic operators registered under `mxnet.ndarray` and `mxnet.symbol`.
- Control flow operators are hard to use. Users have to understand the complicated signatures of control flow operators, instead of writing native Python code using `for`, `while`, `if/else`, etc.
For example, we have learned (in a hard way) that it does not make a lot of sense to ask users to write code like the following to perform a cumulative sum.
```python
def sum(state, i):
    s = state + data[i]
    return s, [s, i + 1]

def sum_cond(state, i):
    return i < 4
    
out, state = F.contrib.while_loop(sum_cond, sum, [F.zeros((1)), F.zeros((1))],
                                  max_iterations=5)
```
Instead, users should be able to just write native Python code as the following and if required, let the framework serialize it into a computation graph for optimization and deployment.
```python
data = np.arange(5)
[...]

PR #14315, opened 2019-03-04
[numpy] Shape support scalar tensor

reminisce:
## Description ##
This PR made necessary changes in the shape-related infra code to support scalars, aka, zero-dim tensors. 
See this [RFC](https://github.com/apache/incubator-mxnet/issues/14253) for more information.

Implemented `np.sum` to verify the correctness with the following script.

After merging this PR, we can get started to make all the existing infer shape functions accommodate the new definition of empty shapes. Implementation of NumPy operators can be carried out in parallel.

Note that before we change all the existing infer functions to accommodate the new empty shape definition, the CI will fail without surprise. When we work on specific operators, we can test those locally to verify the correctness and submit PRs. After all the infer functions are revised, the CI should pass eventually. Then, we can merge the dev branch to the master.

```python
import mxnet as mx
from mxnet import numpy as np


shape = (2, 2)
ret = np.sum(mx.nd.ones(shape))
print("Imperative invoke result==============")
print(ret)

data = mx.sym.var('data', shape=shape)
ret = mx.sym.numpy.sum(data)
exe = ret.simple_bind(ctx=mx.cpu(), data=(2, 2))
exe.forward(data=mx.nd.ones(shape))
print("Symbol bind result===============")
print(exe.outputs[0])

print("CachedOp result===============")
func = mx.nd.CachedOp(ret)
print(func(mx.nd.ones(shape)))
```
```
Imperative invoke result==============

4.0
<NDArray  @cpu(0)>
Symbol bind result===============

4.0
<NDArray  @cpu(0)>
CachedOp result===============

4.0
<NDArray  @cpu(0)>
```
@junrushao1994 @szha @eric-haibin-lin @zheng-da @yzhliu

### DIFF
--- a/include/mxnet/tensor_blob.h
+++ b/include/mxnet/tensor_blob.h
@@ -198,7 +198,6 @@ class TBlob {
       << "Expected: " << type_flag_ << " v.s. given " << mshadow::DataType<DType>::kFlag;
     return mshadow::Tensor<Device, 2, DType>(static_cast<DType*>(dptr_),
                                              shape_.FlatTo2D(),
-                                             shape_[shape_.ndim() - 1],
                                              stream);
   }
   /*!
--- a/include/mxnet/tuple.h
+++ b/include/mxnet/tuple.h
@@ -346,6 +346,7 @@ class Tuple {
   ValueType* data_heap_{nullptr};
   // internal function to change the dimension
   inline void SetDim(int ndim) {
+    CHECK_GE(ndim, -1) << "ndim cannot be less than -1, received " << ndim;
     if (ndim > kStackCache &&
         ndim > num_heap_allocated_) {
       delete [] data_heap_;
--- a/include/mxnet/tuple.h
+++ b/include/mxnet/tuple.h
@@ -17,7 +17,7 @@
  * under the License.
  */
 /*!
- *  Copyright (c) 2016 by Contributors
+ *  Copyright (c) 2019 by Contributors
  * \file mxnet/tuple.h
  * \brief Data structure Tuple and TShape to store dynamic sized shapes.
  */
@@ -39,10 +39,13 @@ namespace mxnet {
 
 /*!
  * \brief A dynamic sized array data structure that is optimized for storing
- *        small number of elements with same type.
+ * small number of elements with same type.
  *
  *  Data will be stored in stack when number of elements is small.
  *  It is suitable to hold shape of Tensor.
+ *
+ *  The ndim of a valid tuple is an integer in range [0, inf).
+ *  ndim = 0 means the tuple is empty.
  *
  * \tparam ValueType The type of data stored inside tuple.
  * \sa TShape
@@ -61,7 +64,11 @@ class Tuple {
    * \param s the source tuple
    */
   inline Tuple(const Tuple<ValueType>& s) {
-    this->assign(s.begin(), s.end());
+    if (s.ndim() == -1) {
+      this->SetDim(-1);
+    } else {
+      this->assign(s.begin(), s.end());
+    }
   }
   /*!
    * \brief constructor from initializer list
@@ -106,6 +113,7 @@ class Tuple {
   inline void assign(RandomAccessIterator begin,
                      RandomAccessIterator end) {
     this->SetDim(end - begin);
+    CHECK_GE(ndim(), 0);
     std::copy(begin, end, this->begin());
   }
   /*!
@@ -124,7 +132,11 @@ class Tuple {
    * \return reference of self
    */
   inline Tuple<ValueType>& operator=(const Tuple<ValueType>& src) {
-    this->assign(src.begin(), src.end());
+    if (src.ndim() == -1) {
+      this->SetDim(-1);
+    } else {
+      this->assign(src.begin(), src.end());
+    }
     return *this;
   }
   /*!
@@ -151,6 +163,7 @@ class Tuple {
    */
   inline bool operator==(const Tuple<ValueType> &s) const {
     if (ndim_ != s.ndim_) return false;
+    if (ndim() == -1) return true;
     return std::equal(begin(), end(), s.begin());
   }
   /*!
@@ -177,7 +190,7 @@ class Tuple {
     return ndim_ <= kStackCache ? (data_stack_ + ndim_): (data_heap_ + ndim_);
   }
   /*! \return number of dimension of the tuple */
-  inline uint32_t ndim() const {
+  inline int ndim() const {
     return ndim_;
   }
   /*!
@@ -185,7 +198,8 @@ class Tuple {
    * \param i dimension index
    * \return the corresponding dimension size
    */
-  inline ValueType& operator[](size_t i) {
+  inline ValueType& operator[](int i) {
+    CHECK(i >= 0 && i < ndim());
     return begin()[i];
   }
   /*!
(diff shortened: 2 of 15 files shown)


junrushao1994:
Awesome work! Let’s work on this

anirudhacharya:

include/mxnet/tuple.h
```diff
@@ -106,7 +113,10 @@ class Tuple {
   inline void assign(RandomAccessIterator begin,
                      RandomAccessIterator end) {
     this->SetDim(end - begin);
-    std::copy(begin, end, this->begin());
+    CHECK_GE(ndim(), 0);
+    if (ndim() > 0) {
```
junrushao1994: Don't need check `ndim() > 0` actually

include/mxnet/tuple.h
```diff
@@ -186,6 +201,7 @@ class Tuple {
    * \return the corresponding dimension size
    */
   inline ValueType& operator[](size_t i) {
```
junrushao1994: Let's make `int i` instead

include/mxnet/tuple.h
```diff
@@ -186,6 +201,7 @@ class Tuple {
...
+    CHECK_GT(ndim(), 0);
```
junrushao1994: Don't check this, but let's check 0 <= i < ndim

include/mxnet/tuple.h
```diff
@@ -194,6 +210,7 @@ class Tuple {
    * \return the corresponding dimension size
    */
   inline const ValueType& operator[](size_t i) const {
+    CHECK_GT(ndim(), 0);
```
junrushao1994: Same as above

include/mxnet/tuple.h
```diff
@@ -220,6 +237,7 @@ class Tuple {
    * \return the ostream
    */
   friend std::ostream &operator<<(std::ostream &os, const Tuple<ValueType> &t) {
+    if (t.ndim() == -1) return os;
```
junrushao1994: Why not we print something like "?"?

src/operator/operator_common.h
```diff
@@ -159,16 +169,16 @@ inline std::string type_string(const int& x) {
  * \return whether x and y are compatible.
  */
 inline bool shape_assign(mxnet::TShape *y, const mxnet::TShape& x) {
-  if (y->ndim() == 0) {
+  if (y->ndim() < 0) {
```
junrushao1994: Let's say `== -1` instead of `< 0`

include/mxnet/tensor_blob.h
```diff
@@ -196,6 +196,10 @@ class TBlob {
     CHECK(mshadow::DataType<DType>::kFlag == type_flag_)
       << "TBlob.get_with_shape: data type do not match specified type."
       << "Expected: " << type_flag_ << " v.s. given " << mshadow::DataType<DType>::kFlag;
+    // if scalar, handle it with special care
+    if (ndim() == 0) {
+      return this->get_with_shape<Device, 2, DType>(mshadow::Shape2(1, 1), stream);
```
szha: why is this needed? should the `shape_.FlatTo2D()` handle it, given that the next call already passes `dptr_`?

reminisce: Because this line would crash.
```cpp
shape_[shape_.ndim() - 1],
```

include/mxnet/tuple.h
reminisce: Good catch. I was fixing another assignment problem and just added the check everywhere. Will remove this.

include/mxnet/tuple.h
reminisce: How about making it more specific: `UNKNOWN_SHAPE`?

include/mxnet/tensor_blob.h
szha: Is the stride needed?

szha: Without providing the stride, it already looks up the size of this dimension as stride:
`static const int  kSubdim = dimension - 1;`

include/mxnet/tuple.h
```diff
@@ -404,9 +452,11 @@ class TShape : public Tuple<dim_t> {
   }
   /*! \return total number of elements in the shape */
   inline size_t Size() const {
+    if (this->ndim() == -1) return 0;
```
szha: Should this throw an error instead?

szha: The hope is to easily identify any illegal use of undefined shapes. what do you think?

reminisce: Agree. It doesn't make sense to get a size value when shape is unknown. Will change that.

include/mxnet/tuple.h
```diff
@@ -220,6 +235,10 @@ class Tuple {
    * \return the ostream
    */
   friend std::ostream &operator<<(std::ostream &os, const Tuple<ValueType> &t) {
+    if (t.ndim() == -1) {
```
szha: should there be an explicit method for checking whether shape is known?

reminisce: By the current convention, it's defined in `operator_common.h` as `shape_is_known`, previously known as `shape_is_none`. For this case specifically, we should not use that function, because we want to just check `ndim`. For this release, I suggest we not move it to keep the convention as is. For the next major release, we can move the util function closer to where the object is defined.

szha:
Thanks for addressing my concerns

include/mxnet/tuple.h
```diff
@@ -316,48 +335,74 @@ class Tuple {
...
   /*!
    * constructor to construct a shape with all 1.
+   * TODO(junwu): The value should default to -1. Need to keep 1 for now
+   * for backward compatibility. Change it to -1 in the future when we can
+   * break backward compatibility.
    * \param ndim the number of dimension
+   * \param value the dimension size for all dims
    */
-  inline TShape(uint32_t ndim) {  // NOLINT(*)
+  inline TShape(int ndim, int value = 1) {  // NOLINT(*)
```
junrushao1994: Should we make `value = -1` now, or in the future? I mean, this PR is targeted at the dev branch, so should be good anyway

include/mxnet/tuple.h
```diff
@@ -404,9 +453,11 @@ class TShape : public Tuple<dim_t> {
   }
   /*! \return total number of elements in the shape */
   inline size_t Size() const {
+    CHECK_GE(this->ndim(), 0) << "Shape is unknown.";
```
junrushao1994: Is this 0 or -1?

include/mxnet/tuple.h
reminisce: I made it -1 before, but it failed at other places where 1 is expected. To avoid resolving conflicts with the master branch too early, we can defer this to the time when we want to merge the whole changes back to master.

junrushao1994: Sounds good

include/mxnet/tuple.h
reminisce: It's CHECK_GE, so should be 0. :)

junrushao1994: I see. I misread just now. Suggest using `CHECK_NE` to be more specific

reminisce: I think `>= 0` is a safer check than `!= -1` to rule out the cases when ndim is mistakenly messed up ending with a negative integer other than -1.

junrushao1994: When it is a negative number other than -1, you should immediately throw an error, instead of silently allow this to happen.

reminisce: Makes sense. I will add this check wherever possible.

PR #14352, opened 2019-03-06
Optimize NMS part 2

ptrendx:
## Description ##
This PR changes the `batch_start` calculation in the BoxNMSForward op to the custom kernel, much faster than the mshadow generated one. In MaskRCNN model it changes the runtime of that part from 20 ms to 2 us, speeding up the single GPU training by 20% in fp16 mode.

### Essentials ###

## Comments ##
- I'm pretty sure that on a CPU path a simple for loop would be much better than the mshadow generated kernel as well, but since I did not have experimental data, I did not change it. FYI @zhreshold

### DIFF
--- a/src/operator/contrib/bounding_box-inl.cuh
+++ b/src/operator/contrib/bounding_box-inl.cuh
@@ -280,6 +280,45 @@ void NMSApply(mshadow::Stream<gpu> *s,
   }
 }
 
+__launch_bounds__(512)
+__global__ void nms_calculate_batch_start_kernel(int32_t * batch_start,
+                                                 int32_t * valid_batch_id,
+                                                 size_t N,
+                                                 int num_batch) {
+  size_t tid = blockIdx.x * blockDim.x + threadIdx.x;
+  if (tid < N) {
+    const int32_t previous = tid > 0 ? __ldg(valid_batch_id + tid - 1) : -1;
+    const int32_t my = __ldg(valid_batch_id + tid);
+    if (my > previous) {
+      for (int32_t current = previous + 1; current <= my; ++current) {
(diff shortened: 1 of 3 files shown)


vandanavk:

src/operator/contrib/bounding_box-inl.cuh
```diff
@@ -280,6 +280,45 @@ void NMSApply(mshadow::Stream<gpu> *s,
...
 }
 
+__launch_bounds__(512)
+__global__ void nms_calculate_batch_start_kernel(int32_t * batch_start,
+                                                 int32_t * valid_batch_id,
+                                                 size_t N,
+                                                 int num_batch) {
+  size_t tid = blockIdx.x * blockDim.x + threadIdx.x;
+  if (tid < N) {
+    const int32_t previous = tid > 0 ? __ldg(valid_batch_id + tid - 1) : -1;
```
arcadiaphy: Using __ldg intrinsics will fail to compile on some earlier cuda archetectures.

ptrendx: It will fail on sm 3.0 and earlier (so Fermi and the first Kepler). I can put ifdef there, but do we care about those?

arcadiaphy: In Makefile, sm 30 is in KNOWN_CUDA_ARCHS.
[https://github.com/apache/incubator-mxnet/blob/master/Makefile#L385](https://github.com/apache/incubator-mxnet/blob/master/Makefile#L385)

ptrendx: Then we do ;-). I will introduce the guard, thanks!

PR #14361, opened 2019-03-07
[numpy] Add a global switch to turn on/off numpy compatibility

reminisce:
This PR is implemented based upon the discussion with @eric-haibin-lin. It provides two APIs and NumPy-compatible NDArray indexing behavior, i.e. `y = x[i]`.

1. `set_numpy_comp(enable)`
This is a thread-safe utility function for users to enable or disable NumPy-compatible behaviors in MXNet. This will generally affect the behavior of the following two operations:
    1. NDArray indexing. When `enable=False`, NDArray indexing always returns a tensor with `ndim >= 1`, which is the default behavior of MXNet kept for backward compatibility. When `enable=True`, NDArray indexing will return a tensor with `ndim >= 0`, which is consistent with NumPy's indexing behavior. The result tensor with `ndim = 0` is actually a scalar whose shape is `()`. For example:

    >>> from mxnet.base import set_numpy_comp
    >>> x = mx.nd.arange(6)
    >>> x
    [0. 1. 2. 3. 4. 5.]
    <NDArray 6 @cpu(0)>
    >>> x[1]
    [1.]
    <NDArray 1 @cpu(0)>
    >>> set_numpy_comp(enable=True)
    >>> x[1]
    1.0
    <NDArray  @cpu(0)>
    >>> set_numpy_comp(enable=False)
    >>> x[1]
    [1.]
    <NDArray 1 @cpu(0)>

    
    2. NDArray's convenience fluent methods. When `enable=True`, the convenience fluent methods will dispatch the calls to NumPy operators if implemented in MXNet. For example, given an NDArray `data`, `data.sum()` will call `mxnet.ndarray.sum(data)` when `enable=False` (default behavior), and mxnet.numpy.sum(data)` when `enable=True`.

2. `_is_numpy_comp()`
This is a thread-safe utility function for checking whether NumPy-compatibility has been enabled or disabled. This is implemented for developers to use. Users are not expected to call this function.

@junrushao1994 @eric-haibin-lin @szha @zheng-da @yzhliu

### DIFF
--- a/include/mxnet/tensor_blob.h
+++ b/include/mxnet/tensor_blob.h
@@ -198,7 +198,6 @@ class TBlob {
       << "Expected: " << type_flag_ << " v.s. given " << mshadow::DataType<DType>::kFlag;
     return mshadow::Tensor<Device, 2, DType>(static_cast<DType*>(dptr_),
                                              shape_.FlatTo2D(),
-                                             shape_[shape_.ndim() - 1],
                                              stream);
   }
   /*!
--- a/include/mxnet/tuple.h
+++ b/include/mxnet/tuple.h
@@ -17,7 +17,7 @@
  * under the License.
  */
 /*!
- *  Copyright (c) 2016 by Contributors
+ *  Copyright (c) 2019 by Contributors
  * \file mxnet/tuple.h
  * \brief Data structure Tuple and TShape to store dynamic sized shapes.
  */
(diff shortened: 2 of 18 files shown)


wkcn:
In my opinion, we can not use the switch. Instead, We can add a typecast from 0-shape NDArray to a scalar.

reminisce:
@wkcn Can you elaborate why we cannot use a switch to guarantee backward compatibility and allow users to opt in new numpy compatible behavior?

wkcn:
when set_numpy_comp(true), the output of some reduction operators is always a scalsr, breaking the construct of static graph and the deployment.

wkcn:
Sorry that I clicked by mistake.

wkcn:
when set_numpy_comp(True) is called, the output of some reduction operators Is always a scalar, which break the construct of computational graph, and not benefit for deployment.

I will write an example.

abhinavs95:
@reminisce Could you please rebase to resolve merge conflicts and fix the CI failures? Thanks

abhinavs95:
@mxnet-label-bot update [Numpy, pr-awaiting-response]

PR #14363, opened 2019-03-08
Support multi-threading for Custom Operator

wkcn:
## Description ##
Hi! I found that there was single-thread to execute custom operator. It drops the performance.
Is it a better way to support custom operator in multi-thread?

Test Case:
```python
import mxnet as mx
import time

class TestOP(mx.operator.CustomOp):
    def __init__(self):
        super(TestOP, self).__init__()
        print("sleep 5s")
        time.sleep(5)
        print("sleep finished")
        out_data[0][:] = in_data[0]
        pass

@mx.operator.register("test_op")
class TestOPProp(mx.operator.CustomOpProp):
    def __init__(self):
        super(TestOPProp, self).__init__()
    def list_arguments(self):
        return ['x']
    def list_outputs(self):
        return ['y']
        return in_shape, in_shape
        return TestOP()

a = mx.nd.array([1,2,3])
tic = time.time()
b = mx.nd.Custom(a, op_type='test_op')
c = mx.nd.Custom(a, op_type='test_op')
d = mx.nd.Custom(a, op_type='test_op')
mx.nd.waitall()
print(time.time() - tic)
```

Cost Time:
Before: 15s
After: 5s

### Essentials ###

### Changes ###
- [x] Support multi-threading for Custom Operator
- [x] Add environmental variable `MXNET_CUSTOM_OP_NUM_THREADS`

## Comments ##

### DIFF
--- a/src/operator/custom/custom-inl.h
+++ b/src/operator/custom/custom-inl.h
@@ -129,6 +129,9 @@ class CustomOperator {
           ctx.run_ctx.ctx, vars, vars2, FnProperty::kNormal, 0,
           "CustomOperator");
     });
+    // increase num_threads if there is not enough threads to execute custom operator
+    if (q_.size() > num_free_threads)
+      CreateThreads(q_.size() - num_free_threads);
     cv_.notify_all();
   }
 
@@ -146,35 +149,45 @@ class CustomOperator {
   static CustomOperator* Get();
 
  private:
-  CustomOperator() {
+  CustomOperator() : num_free_threads(0) {
     destructing_ = false;
     naive_engine_ = true;
     if (std::string("NaiveEngine") != dmlc::GetEnv("MXNET_ENGINE_TYPE", std::string())) {
       naive_engine_ = false;
-      const int num_threads = dmlc::GetEnv("MXNET_CUSTOM_OP_NUM_THREADS", 4);
-      for (int i = 0; i < num_threads; ++i) {
-        workers_.emplace_back(std::thread(
-          [&]() {
-            std::unique_lock<std::mutex> lock(mutex_);
-            while (!q_.empty() || !destructing_) {
-              cv_.wait(lock, [&] {return !q_.empty() || destructing_;});
-              while (!q_.empty()) {
-                auto fn = q_.front();
-                q_.pop();
-                lock.unlock();
-                fn();
-                lock.lock();
-              }
-            }
-        }));
+      SetNumThreads(1);
+    }
+  }
+  void ThreadTarget() {
+    std::unique_lock<std::mutex> lock(mutex_);
+    while (!q_.empty() || !destructing_) {
+      cv_.wait(lock, [&] {return !q_.empty() || destructing_;});
+      while (!q_.empty()) {
+        --num_free_threads;
+        auto fn = q_.front();
+        q_.pop();
+        lock.unlock();
+        fn();
+        ++num_free_threads;
+        lock.lock();
       }
     }
+  }
+  void SetNumThreads(int num_threads) {
+    num_threads = std::min(dmlc::GetEnv("MXNET_CUSTOM_OP_NUM_THREADS", 16), num_threads);
+    for (int i = workers_.size(); i < num_threads; ++i) {
+      workers_.emplace_back(std::thread([this]{this->ThreadTarget();}));
+      ++num_free_threads;
+    }
+  }
+  void CreateThreads(int num_new_threads) {
+    SetNumThreads(workers_.size() + num_new_threads);
   }
   std::mutex mutex_;
   std::map<std::string, CustomOpPropCreator> registry_;
   // async worker
   std::condition_variable cv_;
   std::vector<std::thread> workers_;
+  std::atomic<uint32_t> num_free_threads;
   std::queue<std::function<void(void)> > q_;
   bool naive_engine_;
   bool destructing_;
--- a/src/operator/custom/custom-inl.h
+++ b/src/operator/custom/custom-inl.h
@@ -31,6 +31,7 @@
 #include <mxnet/operator.h>
 #include <mxnet/c_api.h>
 #include <mxnet/imperative.h>
+#include <algorithm>
 #include <map>
 #include <vector>
 #include <string>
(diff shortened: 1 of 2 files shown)


wkcn:
I will check whehter it is correct to use the lock.

src/operator/custom/custom-inl.h
```diff
@@ -139,38 +142,52 @@ class CustomOperator {
...
+        q_.pop();
+        lock.unlock();
+        fn();
+        ++num_free_threads;
+        lock.lock();
+      }
+    }
+  }
+  void SetNumThreads(int num_threads) {
+    num_threads = std::min(dmlc::GetEnv("MXNET_CUSTOM_OP_NUM_THREADS", 16), num_threads);
```
eric-haibin-lin: I see `SetNumThreads (1) ` is called. Is MXNET_CUSTOM_OP_NUM_THREADS going to affect the result in anyway?

src/operator/custom/custom-inl.h
```diff
@@ -129,6 +129,9 @@ class CustomOperator {
           ctx.run_ctx.ctx, vars, vars2, FnProperty::kNormal, 0,
           "CustomOperator");
     });
+    // increase num_threads if there is not enough threads to execute custom operator
+    if (q_.size() > num_free_threads)
```
eric-haibin-lin: I'm not sure if this is a good behavior. If the number of ops to execute is too many, it keeps creating threads and the number of threads may keep going up unbounded.

src/operator/custom/custom-inl.h
wkcn: MXNET_CUSTOM_OP_NUM_THREADS is the maximum number of threads for custom operator.

src/operator/custom/custom-inl.h
wkcn: MXNET_CUSTOM_OP_NUM_THREADS is the upbound number of threads.

vandanavk:

src/operator/custom/custom-inl.h
```diff
@@ -139,38 +143,51 @@ class CustomOperator {
       destructing_ = true;
       cv_.notify_all();
     }
-    worker_.join();
+    for (auto &worker : workers_)
+      worker.join();
```
anirudh2290: Have you evaluated this when the custom op is part of a bigger graph and if there is any performance impact ? Since the CustomOperator is static its lifetime is till the end of the program and destructor of customoperator gets called at the end. This means there is one thread that is waiting on the condition variable while other 11 threads tryign to obtain the lock and in a blocked state. Since these are idle threads, I am not sure if the impact will be significant but good to verify. Will also help us come up with a good default for MXNET_CUSTOM_OP_NUM_THREADS

src/operator/custom/custom-inl.h
```diff
@@ -139,38 +143,51 @@ class CustomOperator {
...
     }
   }
+  void ThreadTarget() {
+    std::unique_lock<std::mutex> lock(mutex_);
+    while (!q_.empty() || !destructing_) {
+      cv_.wait(lock, [&] {return !q_.empty() || destructing_;});
+      while (!q_.empty()) {
+        --num_free_threads;
+        auto fn = q_.front();
+        q_.pop();
```
anirudh2290: Nice ! Popping early will prevent the race condition.

src/operator/custom/custom-inl.h
wkcn: Thank you!

Sorry that I do not have any server with multiple GPUs to evaluate a big graph.

In this PR, the number of threads will increase when threads are not enough, and the maximum number id MXNET_CUSTOM_OP_NUM_THREADS.

There are always threads which get the lock and execute the operator function, so I think the idle threads do not drop the performance.

anirudh2290: Yes it's fine during op execution but even after custom op execution there is one thread waiting on CV and other 11 trying to acquire lock.

wkcn: I see. 
I ever thought that decrease the number of threads when threads are idle, however it is difficult to estimate the number.

anirudh2290: can you run the example/reinforcement-learning/dqn  which includes a custom op on CPU to check for performance.

wkcn: IT may be not available to check for performance on only CPU，since there Is only a computational stream usually. I will find a server with multiple GPUs and check on it.

anirudh2290: Yes this change is mainly for performance improvements on a GPU machine, but we should be careful not to impact cpu performance for inference. Otherwise we should keep the default small.

wkcn: Sorry that I met the problem when running the RL demo:
```bash
(gdb) bt                                                                                                                                    
#0  0x00007ffff7e1ff73 in free () at /usr/lib/libc.so.6                                                                                     
#1  0x00007fffbf2c9ad1 in ALEInterface::welcomeMessage[abi:cxx11]() () at /usr/lib/python2.7/site-packages/ale_python_interface/libale_c.so 
#2  0x00007fffbf2c9d0a in ALEInterface::ALEInterface() () at /usr/lib/python2.7/site-packages/ale_python_interface/libale_c.so              
#3  0x00007fffbf2c7fbb in ALE_new () at /usr/lib/python2.7/site-packages/ale_python_interface/libale_c.so                                   
#4  0x00007ffff6d236d0 in ffi_call_unix64 () at /usr/lib/libffi.so.6                                                                        
#5  0x00007ffff6d230a0 in ffi_call () at /usr/lib/libffi.so.6                                                                               
#6  0x00007ffff6cc9d77 in _ctypes_callproc () at /usr/lib/python2.7/lib-dynload/_ctypes.so                                                 
#7  0x00007ffff6cc37d0 in  () at /usr/lib/python2.7/lib-dynload/_ctypes.so                                                                 
#8  0x00007ffff7c8cb43 in PyObject_Call () at /usr/lib/libpython2.7.so.1.0                                                                  
#9  0x00007ffff7c1c10e in PyEval_EvalFrameEx () at /usr/lib/libpython2.7.so.1.0                                                            
#10 0x00007ffff7c156ba in PyEval_EvalCodeEx () at /usr/lib/libpython2.7.so.1.0             
```

wkcn: I'm checking it.

wkcn: In my experiment, the performance does not drop in DQN example. The FPS keeps 1500 on my laptop with only CPU i7-7500U(2c4t).

wkcn:
@eric-haibin-lin @szha Hi! Is it available to merge this PR?

RogerChern:
Great work!

PR #14372, opened 2019-03-08
Memory fixes. Resolves #10867, and resolves #14080

### LINKED ISSUE #10867: Scala Module API resize is leaking memory on the native size.
## Description

Create and bind a MXNet Module with batch size N+1 and proceed to loop and pass DataBatches to it that require the Module to resize before performing the forward pass. Monitor the system resources (With htop, nvidia-smi, jvmtop) and you will notice the used system memory in htop will start to grow, but not the jvm heap size (the system memory usages grows beyond the set max JVM heap size) or GPU memory usage. This will continue until your system runs out of memory and there is a crash or the JVM is killed clearing all of the leaked used system memory with it.

```
Version      : 3.4.3
Compiler     : GCC 4.8.4
Build        : ('default', 'Nov 17 2016 01:08:31')
Arch         : ('64bit', 'ELF')
Version      : 10.0.1
Directory    : /usr/local/lib/python3.4/dist-packages/pip
Version      : 1.1.0
Directory    : /home/paperspace/src/mxnet/python/mxnet
Hashtag not found. Not installed from pre-built package.
Platform     : Linux-4.4.0-31-generic-x86_64-with-Ubuntu-14.04-trusty
system       : Linux
node         : psplnlbg
release      : 4.4.0-31-generic
version      : #50~14.04.1-Ubuntu SMP Wed Jul 13 01:07:32 UTC 2016
machine      : x86_64
processor    : x86_64
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                8
On-line CPU(s) list:   0-7
Thread(s) per core:    1
Core(s) per socket:    8
Socket(s):             1
NUMA node(s):          1
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 79
Stepping:              1
CPU MHz:               2600.072
BogoMIPS:              5200.14
Hypervisor vendor:     Microsoft
Virtualization type:   full
L1d cache:             32K
L1i cache:             32K
L2 cache:              256K
L3 cache:              10240K
NUMA node0 CPU(s):     0-7
Setting timeout: 10
Timing for FashionMNIST: https://apache-mxnet.s3-accelerate.dualstack.amazonaws.com/gluon/dataset/fashion-mnist/train-labels-idx1-ubyte.gz, DNS: 0.0884 sec, LOAD: 0.4182 sec.
Timing for MXNet: https://github.com/apache/incubator-mxnet, DNS: 0.0131 sec, LOAD: 0.4660 sec.
Timing for Gluon Tutorial(cn): https://zh.gluon.ai, DNS: 0.0957 sec, LOAD: 0.6263 sec.
Timing for PYPI: https://pypi.python.org/pypi/pip, DNS: 0.0099 sec, LOAD: 0.3049 sec.
Timing for Gluon Tutorial(en): http://gluon.mxnet.io, DNS: 0.0134 sec, LOAD: 0.6473 sec.
Timing for Conda: https://repo.continuum.io/pkgs/free/, DNS: 0.0126 sec, LOAD: 0.0343 sec.

```

Package used (Python/R/Scala/Julia):
Scala. This seems to be specific to the Scala API. Can not reproduce in Python. 

For Scala user, please provide:
1. Java version: (`java -version`)
java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)

2. Maven version: (`mvn -version`)
Apache Maven 3.0.5
Maven home: /usr/share/maven
Java version: 1.8.0_131, vendor: Oracle Corporation
Java home: /usr/lib/jvm/java-8-oracle/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.4.0-31-generic", arch: "amd64", family: "unix"

3. Scala runtime if applicable: (`scala -version`)
2.11.11

## Build info (Required if built from source)

Compiler (gcc/clang/mingw/visual studio): GCC 4.8.4

MXNet commit hash:
07a83a0325a3d782513a04f47d711710972cb144

Build config:
make  -j $(nproc) USE_OPENCV=1 USE_BLAS=openblas USE_CUDA=1 USE_CUDA_PATH=/usr/local/cuda USE_CUDNN=1
make scalainstall -j $(nproc) USE_OPENCV=1 USE_BLAS=openblas USE_CUDA=1 USE_CUDA_PATH=/usr/local/cuda USE_CUDNN=1

## Error Message:
None

link to simple Scala project/code to reproduce issue
https://github.com/jessebrizzi/MXNet-Bug/blob/master/scala/TestBug.scala

## Steps to reproduce

1. Pull [this project](https://github.com/jessebrizzi/MXNet-Bug) and run it with SBT to test this issue.   

1. Avoid passing in DataBatch that requires resize by padding smaller DataBatch with 0's. 
2. Not using the Scala API

### LINKED ISSUE #11926: segfault in native code while trying to use CustomOp
I'm trying to use CustomOp to create a Constant, as it has been suggested in #8428.
As soon as I define that my CustomOp has no inputs, it fails with a segfault, and I can't find a workaround.

## Environment 
- Linux setup using  https://raw.githubusercontent.com/apache/incubator-mxnet/1.2.1/ci/docker/install/ubuntu_core.sh (actually in a Docker)
- MXNet 1.2.1, scala 2.11.11

## Code
```
class ConstantOp(value: NDArray) extends CustomOp {

  def forward(isTrain: Boolean, req: Array[String], inData: Array[NDArray], outData: Array[NDArray], aux: Array[NDArray]): Unit = {
    val data = value.copyTo(outData(0).context)
    this.assign(outData(0), req(0), data)
    data.dispose()
  }

  def backward(req: Array[String], outGrad: Array[NDArray], inData: Array[NDArray], outData: Array[NDArray], inGrad: Array[NDArray], aux: Array[NDArray]): Unit = {
    throw new Exception(s"Backward not supported by Constant")
  }
}

class ConstantOpProp(needTopGrad: Boolean = false) extends CustomOpProp(needTopGrad) {

  override def listArguments(): Array[String] = Array()

  override def listOutputs(): Array[String] = Array("output")

  override def inferShape(inShape: Array[Shape]): (Array[Shape], Array[Shape], Array[Shape]) = {
    val data = NDArray.deserialize(this.kwargs("value").toCharArray.map(_.toByte))
    (Array(), Array(data.shape), null)
  }

  override def inferType(inType: Array[DType]): (Array[DType], Array[DType], Array[DType]) = {
    val data = NDArray.deserialize(this.kwargs("value").toCharArray.map(_.toByte))
    (Array(), Array(data.dtype), null)
  }

  override def createOperator(ctx: String, inShapes: Array[Array[Int]],
                              inDtypes: Array[Int]): CustomOp = {
    // hacky stuff to workaround the declaration using String
    val data = NDArray.deserialize(this.kwargs("value").toCharArray.map(_.toByte))
    new ConstantOp(data)
  }
}


object TestConst {
  Operator.register("constant", new ConstantOpProp())

  val value = NDArray.array(Array(1f), Shape(1))
  val const = Symbol.Custom("constant")()(
    kwargs = Map(
      "op_type" -> "constant",
      // hacky thing to workaround the fact CustomOpProp uses Map[String, String] internally for kwargs
      "value" -> String.copyValueOf(value.serialize().map(_.toChar))
    ))

  val a = Symbol.Variable("a")
  val symbol = a + const
  val e = symbol.bind(Context.defaultCtx, Map(
    "a" -> NDArray.array(Array(10f), Shape(1)))
  )

  e.forward()

  println("outputs=" + e.outputs.mkString(", "))
}
```

## Error:
```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f37ec25e7a9, pid=14172, tid=0x00007f379cbf7700
#
# JRE version: OpenJDK Runtime Environment (8.0_151-b12) (build 1.8.0_151-8u151-b12-0ubuntu0.16.04.2-b12)
# Java VM: OpenJDK 64-Bit Server VM (25.151-b12 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# V  [libjvm.so+0x6797a9]
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /app/hs_err_pid14172.log
```
in the log
```
[...]

Stack: [0x00007f6b28ad3000,0x00007f6b28bd4000],  sp=0x00007f6b28bcecd0,  free space=1007k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
V  [libjvm.so+0x6797a9]
C  [mxnet-scala+0x455dd4]  Java_org_apache_mxnet_LibInfo_mxCustomOpRegister::{lambda(char const*, int, char const**, char const**, MXCallbackList*)#1}::operator()(char const*, int, char const**, char const**, MXCallbackList*) const::{lambda(int, int*, unsigned int**, void*)#5}::_FUN(int, {lambda(char const*, int, char const**, char const**, MXCallbackList*)#1}, unsigned int*, unsigned int**)+0x454
C  [mxnet-scala+0x6acf4c]  mxnet::op::custom::InferShape(nnvm::NodeAttrs const&, std::vector<nnvm::TShape, std::allocator<nnvm::TShape> >*, std::vector<nnvm::TShape, std::allocator<nnvm::TShape> >*)+0x2dc

Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
j  org.apache.mxnet.LibInfo.mxExecutorBindEX(JIII[Ljava/lang/String;[I[II[J[J[I[JJLorg/apache/mxnet/Base$RefLong;)I+0
j  org.apache.mxnet.Symbol.bindHelper(Lorg/apache/mxnet/Context;Lscala/collection/Seq;Lscala/collection/Iterable;Lscala/collection/Iterable;Lscala/collection/Iterable;Lscala/collection/Iterable;Lscala/collection/immutable/Map;Lorg/apache/mxnet/Executor;)Lorg/apache/mxnet/Executor;+767
j  org.apache.mxnet.Symbol.bind(Lorg/apache/mxnet/Context;Lscala/collection/immutable/Map;)Lorg/apache/mxnet/Executor;+38

[...]
```

### side-note
As you see in the code, I'm obliged to hack NDArrays into strings to transmit the data. That's because CustomOp implementation defines Map[String, String] for kwargs, whereas Symbol.Custom allows Map[String, Any]. It leads to very strange things where we actually have, at runtime, non-string objects behind java String references. But they aren't castable anyway because of the type system. Weird
A change of the def in CustomOp would be welcome.

### LINKED ISSUE #14080: Adam Optimizer Memory Leak in Scala
## Description
Memory leak issue while using Adam optimizer with MXNet Scala Bindings. Running the code below will keep consuming more and more memory till you run out.

## Steps to Reproduce
```scala
// Simple MLP network
def mlpNetwork(): Symbol = {
    val input = Symbol.Variable("data")
    val label = Symbol.Variable("label")
    val fc1 = Symbol.FullyConnected(name = "fc1")()(Map("data" -> input, "num_hidden" -> 128))
    val act1 = Symbol.Activation(name = "relu")()(Map("data" -> fc1, "act_type" -> "relu"))
    val fc2 = Symbol.FullyConnected(name = "fc2")()(Map("data" -> act1, "num_hidden" -> 1))
    val loss = Symbol.LinearRegressionOutput(name="loss")()(Map("data" -> fc2, "label" -> label))
    loss
  }

def getNDArrayIter: NDArrayIter = {
    val f = NDArray.zeros(100, 20, 20)
    val l = NDArray.zeros(100, 1)
    val data = Array(f)
    val labels = Array(l)
    val batchSize = 10
    val iter = new NDArrayIter(data, labels, batchSize)
    iter
  }

val net = mlpNetwork()
val iter = getNDArrayIter()
val optimizer = new Adam(0.001f, 0.9f, 0.999f, 1e-8f, 1 - 1e-8f, 0f, 10f, null);
val init = new Normal(0.01f);
val model = FeedForward.newBuilder(modelSpec)
                .setContext(Array(Context.gpu(0)))
                .setInitializer(init)
                .setNumEpoch(100000)
                .setOptimizer(optimizer)
                .setTrainData(iter)
                .setEvalData(iter)
                .build();

```

## Issue

The issue is (I think) some temporary NDArrays are not getting disposed in Adam optimizer when using `disposeDepsExcept`.

The places exactly where the memory leak occurs is in 3 locations where the method `disposeDepsExcept` is used in Adam's `update` method.

## Temporary Fix

Replace all the 3 lines that use `disposeDepsExcept` in `update` method of `Adam.scala` by explicitly disposing the temporary NDArrays that were created as shown below

Instead of the 3 following lines in `Adam.scala`
```scala
val meanT = (beta1t * mean + (1.0 - beta1t) * resdGrad)
      .disposeDepsExcept(mean, resdGrad)

val varianceT = (beta2 * variance + (1.0f - beta2) * resdGrad * resdGrad)
      .disposeDepsExcept(variance, resdGrad)

val step = (learningRate * meanT / (NDArray.sqrt(varianceT) + epsilon))
      .disposeDepsExcept(meanT, varianceT)
```

Replace it by:
```scala
val beta1Mean = beta1 * mean
    val beta1ResGrad = (1.0 - beta1t) * resdGrad
    val meanT = beta1Mean + beta1ResGrad
    // dipose temp NDArrays
    betaMean.dispose()
    betaResGrad.dispose()

val beta2Variance = beta2 * variance
    val beta2ResGrad = (1.0f - beta2) * resdGrad
    val beta2ResGradSquare = beta2ResGrad * resdGrad
    val varianceT = beta2Variance + beta2ResGradSquare
    // dipose temp NDArrays
    beta2Variance.dispose()
    beta2ResGrad.dispose()
    beta2ResGradSquare.dispose()


    val lrMeanT = learningRate * meanT
    val sqrtVar = NDArray.sqrt(varianceT)
    val sqrtVarPlusEpsilon = sqrtVar + epsilon
    val step = lrMeanT / sqrtVarPlusEpsilon
    // dipose temp NDArrays
    lrMeanT.dispose()
    sqrtVar.dispose()
    sqrtVarPlusEpsilon.dispose()

```

The above changes fixes things for now, but for some reason `disposeDepsExcept` is not doing its job in this case.

```
Version      : 3.7.1
Compiler     : GCC 7.3.0
Build        : ('default', 'Dec 14 2018 19:28:38')
Arch         : ('64bit', '')
Version      : 18.1
Directory    : /home/satya/anaconda3/lib/python3.7/site-packages/pip
Version      : 1.3.1
Directory    : /home/satya/Documents/workspace/mxnet_1.3.x/python/mxnet
Hashtag not found. Not installed from pre-built package.
Platform     : Linux-4.4.0-141-generic-x86_64-with-debian-stretch-sid
system       : Linux
node         : DS5
release      : 4.4.0-141-generic
version      : #167-Ubuntu SMP Wed Dec 5 10:40:15 UTC 2018
Setting timeout: 10
Timing for MXNet: https://github.com/apache/incubator-mxnet, DNS: 0.0405 sec, LOAD: 0.6186 sec.
Timing for Gluon Tutorial(en): http://gluon.mxnet.io, DNS: 0.1403 sec, LOAD: 0.4726 sec.
Timing for Gluon Tutorial(cn): https://zh.gluon.ai, DNS: 0.2418 sec, LOAD: 0.4049 sec.
Timing for FashionMNIST: https://apache-mxnet.s3-accelerate.dualstack.amazonaws.com/gluon/dataset/fashion-mnist/train-labels-idx1-ubyte.gz, DNS: 0.0445 sec, LOAD: 0.1894 sec.
Timing for PYPI: https://pypi.python.org/pypi/pip, DNS: 0.0779 sec, LOAD: 0.2447 sec.
Timing for Conda: https://repo.continuum.io/pkgs/free/, DNS: 0.0409 sec, LOAD: 0.0746 sec.
```

Package used (Python/R/Scala/Julia): Scala

For Scala user, please provide:
1. Java version: 1.8.0_201
2. Maven version: 3.6.0
3. Scala runtime if applicable: 2.11.6

## Build info (Required if built from source)

Compiler (gcc/clang/mingw/visual studio): gcc

MXNet commit hash: 96b4b6ef3c60c63644a7c4d672109b97561b839d

andrewfayres:
## Description ##
There were a few memory leaks in executor and some instances of where ResourceScope was causing problems. This should fix those known issues.

To elaborate on the ResourceScope issue: there are some times when classes such as Executor creates other NativeResources (usually NDArrays) after the Executor has been instantiated. The executor then has a dependency on that new NativeResource. Since these happen at different times, they could potentially be created at different ResourceScope levels. The result is an unstable state where exiting the scope would cause a crash. Fixed this by adding a method to NativeResource which allows moving a resource to the same scope as another. This allows the parent class to control the scope of their dependencies.

Resolves #10867, and resolves #14080.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
+++ b/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
@@ -47,7 +47,6 @@ class ResourceScope extends AutoCloseable {
     * the associated`'org.apache.mxnet.NativeResource.close()` method
     */
   override def close(): Unit = {
-    print("in close\n")
     ResourceScope.removeFromThreadLocal(this)
     resourceQ.foreach(resource => if (resource != null) resource.dispose(false) )
     resourceQ.clear()
@@ -111,11 +110,6 @@ object ResourceScope {
       scope
     } else new ResourceScope()
 
-    if (scope != null) {
-      print("in non null using\n")
-    } else {
-      print("in null using\n")
-    }
     @inline def resourceInGeneric(g: scala.collection.Iterable[_]) = {
       g.foreach( n =>
         n match {
@@ -140,12 +134,14 @@ object ResourceScope {
 
     try {
       val ret = body
-       ret match {
+      if (scope == null) {
+        ret match {
           // don't de-allocate if returning any collection that contains NativeResource.
-        case resInGeneric: scala.collection.Iterable[_] => resourceInGeneric(resInGeneric)
-        case nRes: NativeResource => curScope.moveToOuterScope(nRes)
-        case ndRet: NDArrayFuncReturn => ndRet.arr.foreach( nd => curScope.moveToOuterScope(nd) )
-        case _ => // do nothing
+          case resInGeneric: scala.collection.Iterable[_] => resourceInGeneric(resInGeneric)
+          case nRes: NativeResource => curScope.moveToOuterScope(nRes)
+          case ndRet: NDArrayFuncReturn => ndRet.arr.foreach(nd => curScope.moveToOuterScope(nd))
+          case _ => // do nothing
+        }
       }
       ret
     } catch {
@@ -175,6 +171,16 @@ object ResourceScope {
         } finally {
           throw toThrow
         }
+      }
+    }
+  }
+
+  private[mxnet] def usingIfScopeExists[A](scope: Option[ResourceScope])(body: => A): A = {
+    if (scope == None) {
+      body
+    } else {
+      ResourceScope.using(scope.get){
+        body
       }
     }
   }
--- a/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
+++ b/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
@@ -137,7 +137,7 @@ object ResourceScope {
           // don't de-allocate if returning any collection that contains NativeResource.
         case resInGeneric: scala.collection.Iterable[_] => resourceInGeneric(resInGeneric)
         case nRes: NativeResource => curScope.moveToOuterScope(nRes)
-        case ndRet: NDArrayFuncReturn => ndRet.arr.foreach(nd => curScope.moveToOuterScope(nd))
+        case ndRet: NDArrayFuncReturn => ndRet.arr.foreach( nd => curScope.moveToOuterScope(nd) )
         case _ => // do nothing
       }
       ret
@@ -147,9 +147,7 @@ object ResourceScope {
         null.asInstanceOf[A] // we'll throw in finally
     } finally {
       var toThrow: Throwable = retThrowable
-      if (retThrowable eq null) {
-        curScope.close
-      }
+      if (retThrowable eq null) curScope.close
       else {
         try {
           curScope.close
@@ -194,7 +192,6 @@ object ResourceScope {
     * @param r ResourceScope to remove
     */
   private[mxnet] def removeFromThreadLocal(r: ResourceScope): Unit = {
-    // threadLocalScopes.get() -= r
     threadLocalScopes.get().remove(threadLocalScopes.get().lastIndexOf(r))
   }
 
--- a/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
+++ b/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
@@ -134,14 +134,12 @@ object ResourceScope {
 
     try {
       val ret = body
-      if (scope == null) {
-        ret match {
+       ret match {
           // don't de-allocate if returning any collection that contains NativeResource.
-          case resInGeneric: scala.collection.Iterable[_] => resourceInGeneric(resInGeneric)
-          case nRes: NativeResource => curScope.moveToOuterScope(nRes)
-          case ndRet: NDArrayFuncReturn => ndRet.arr.foreach(nd => curScope.moveToOuterScope(nd))
-          case _ => // do nothing
-        }
+        case resInGeneric: scala.collection.Iterable[_] => resourceInGeneric(resInGeneric)
+        case nRes: NativeResource => curScope.moveToOuterScope(nRes)
+        case ndRet: NDArrayFuncReturn => ndRet.arr.foreach(nd => curScope.moveToOuterScope(nd))
+        case _ => // do nothing
       }
       ret
     } catch {
--- a/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
+++ b/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
@@ -182,6 +182,11 @@ object ResourceScope {
     threadLocalScopes.get() -= r
   }
 
+  private[mxnet] def getIndexOfScope(resourceScope: ResourceScope): Int = {
+    val scopes = threadLocalScopes.get()
+    scopes.indexOf(resourceScope)
+  }
+
   /**
     * Get the latest Scope in the stack
     * @return
--- a/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
+++ b/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
@@ -47,6 +47,7 @@ class ResourceScope extends AutoCloseable {
     * the associated`'org.apache.mxnet.NativeResource.close()` method
     */
   override def close(): Unit = {
+    print("in close\n")
     ResourceScope.removeFromThreadLocal(this)
     resourceQ.foreach(resource => if (resource != null) resource.dispose(false) )
     resourceQ.clear()
@@ -105,8 +106,16 @@ object ResourceScope {
   // TODO: we should move to the Scala util's Using method when we move to Scala 2.13
   def using[A](scope: ResourceScope = null)(body: => A): A = {
 
-    val curScope = if (scope != null) scope else new ResourceScope()
+    val curScope = if (scope != null) {
+      ResourceScope.addToThreadLocal(scope)
+      scope
+    } else new ResourceScope()
 
+    if (scope != null) {
+      print("in non null using\n")
+    } else {
+      print("in null using\n")
+    }
     @inline def resourceInGeneric(g: scala.collection.Iterable[_]) = {
       g.foreach( n =>
         n match {
@@ -146,11 +155,19 @@ object ResourceScope {
     } finally {
       var toThrow: Throwable = retThrowable
       if (retThrowable eq null) {
-        if (scope == null) curScope.close()
+        if (scope == null) {
+          curScope.close()
+        } else {
+          ResourceScope.removeFromThreadLocal(scope)
+        }
       }
       else {
         try {
-          if (scope == null) curScope.close
+          if (scope == null) {
+            curScope.close
+          } else {
+            ResourceScope.removeFromThreadLocal(scope)
+          }
         } catch {
           case closeThrowable: Throwable =>
             if (NonFatal(retThrowable) && !NonFatal(closeThrowable)) toThrow = closeThrowable
@@ -181,7 +198,8 @@ object ResourceScope {
     * @param r ResourceScope to remove
     */
   private[mxnet] def removeFromThreadLocal(r: ResourceScope): Unit = {
-    threadLocalScopes.get() -= r
+    // threadLocalScopes.get() -= r
+    threadLocalScopes.get().remove(threadLocalScopes.get().lastIndexOf(r))
   }
 
   /**
--- a/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
+++ b/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
@@ -48,8 +48,10 @@ class ResourceScope extends AutoCloseable {
     */
   override def close(): Unit = {
     ResourceScope.removeFromThreadLocal(this)
-    resourceQ.foreach(resource => if (resource != null) resource.dispose(false) )
-    resourceQ.clear()
+    if (!ResourceScope.threadLocalScopes.get().contains(this)) {
+      resourceQ.foreach(resource => if (resource != null) resource.dispose(false))
+      resourceQ.clear()
+    }
   }
 
   /**
@@ -105,10 +107,7 @@ object ResourceScope {
   // TODO: we should move to the Scala util's Using method when we move to Scala 2.13
   def using[A](scope: ResourceScope = null)(body: => A): A = {
 
-    val curScope = if (scope != null) {
-      ResourceScope.addToThreadLocal(scope)
-      scope
-    } else new ResourceScope()
+    val curScope = if (scope != null) scope else new ResourceScope()
 
     @inline def resourceInGeneric(g: scala.collection.Iterable[_]) = {
       g.foreach( n =>
@@ -149,19 +148,11 @@ object ResourceScope {
     } finally {
       var toThrow: Throwable = retThrowable
       if (retThrowable eq null) {
-        if (scope == null) {
-          curScope.close()
-        } else {
-          ResourceScope.removeFromThreadLocal(scope)
-        }
+        curScope.close
       }
       else {
         try {
-          if (scope == null) {
-            curScope.close
-          } else {
-            ResourceScope.removeFromThreadLocal(scope)
-          }
+          curScope.close
         } catch {
           case closeThrowable: Throwable =>
             if (NonFatal(retThrowable) && !NonFatal(closeThrowable)) toThrow = closeThrowable
@@ -177,6 +168,7 @@ object ResourceScope {
     if (scope == None) {
       body
     } else {
+      ResourceScope.addToThreadLocal(scope.get)
       ResourceScope.using(scope.get){
         body
       }
--- a/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
+++ b/scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
@@ -145,10 +145,12 @@ object ResourceScope {
         null.asInstanceOf[A] // we'll throw in finally
     } finally {
       var toThrow: Throwable = retThrowable
-      if (retThrowable eq null) curScope.close()
+      if (retThrowable eq null) {
+        if (scope == null) curScope.close()
+      }
       else {
         try {
-          curScope.close
+          if (scope == null) curScope.close
         } catch {
           case closeThrowable: Throwable =>
             if (NonFatal(retThrowable) && !NonFatal(closeThrowable)) toThrow = closeThrowable
@@ -180,11 +182,6 @@ object ResourceScope {
     */
   private[mxnet] def removeFromThreadLocal(r: ResourceScope): Unit = {
     threadLocalScopes.get() -= r
-  }
-
-  private[mxnet] def getIndexOfScope(resourceScope: ResourceScope): Int = {
-    val scopes = threadLocalScopes.get()
-    scopes.indexOf(resourceScope)
   }
 
   /**
--- a/scala-package/core/src/test/scala/org/apache/mxnet/ResourceScopeSuite.scala
+++ b/scala-package/core/src/test/scala/org/apache/mxnet/ResourceScopeSuite.scala
@@ -102,12 +102,13 @@ class ResourceScopeSuite extends FunSuite with BeforeAndAfterAll with Matchers {
   }
 
   /**
-    * Goes multiple scopes deep to make sure we can move across multiple levels correctly
+    * Tests passing a scope to using and creating new resources within.
     */
   test("test moving scope of native resource to scope of another") {
     var a: TestNativeResource = null
     var b: TestNativeResource = null
     var c: TestNativeResource = null
+    var d: TestNativeResource = null
 
     ResourceScope.using() {
       a = new TestNativeResource()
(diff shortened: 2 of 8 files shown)


vandanavk:
@mxnet-label-bot add [Memory, Scala, pr-awaiting-review]

andrewfayres:
NeuralStyle on gpu seems to be failing. I'll have to dig into why. I also think I've thought of a better way to deal with the interdependency of symbol/executor than what is being done right now so I'll update that as well.

nswamy:
If I understood correct(from the description), the problem is Executor creates some NDArrays that gets disposed before the executor?
I am wondering how are these NDArrays getting released before the destruction of the executor?

andrewfayres:
@nswamy This is because there is no guarantee that they are created at the same ResourceScope level. For example, the Executor gets created and returned to the user. The user then opens up a ResourceScope to do inference. In this scope, the executor might end up creating new resources for itself. Those resources are then in the scope level of the user's ResourceScope and get disposed when that closes. This leaves the Executor without the dependencies it created.

The solution is to have objects like executor put any dependencies that they create into the same ResourceScope as themselves. This way they are all coupled together. I thought about using moveToOuterScope but there's no way to know how many levels of scope could be between the two.

nswamy:
@andrewfayres, thanks for taking care of this, Its a really nice find :)

I find the design of executor less than desirable(where it creates states in a method dynamically) after the 
the object is created. 
Anyway, here is what we discussed offline:

* `moveToScopeOf` needs cognizant use of it after the creation of the resources
* calling `moveToScopeOf` on a large amount of objects will have performance impact.

My suggestion is to create a couple methods in ResourceScope.
ResourceScope.setScope(scope: ResourceScope)
ResourceScope.resetScope() 
We can document these methods to be used when you want to hold NativeResources as state variables inside another NativeResource, in this case the Executor. 

~~Explore to automatically detect(possibly reflection) creation of NativeResource within another NativeResource. 
Ideal would have been to pass the scope in which to create to a native resource, unfortunately that would require changing every API of NDArray, Symbol, etc., since most of them are exposed as static methods.~~

I am open to hear any other interesting ideas that is less intrusive.

andrewfayres:
I came up with a different solution which we (@nswamy and I) discussed offline. Essentially, ResourceScope.using already has an optional parameter for an existing scope. A few changes minor changes should let us reuse that.

This approach should be more consistent and intuitive. I'll update the PR.

lanking520:
Any updates?

andrewfayres:
I was on call last week and didn't have time to pick this back up unfortunately. I started working on it again yesterday and finished the code changes. I'm going to do testing today and if all goes well I'll have an update this afternoon.

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
```diff
@@ -140,12 +134,14 @@ object ResourceScope {
 
     try {
       val ret = body
-       ret match {
+      if (scope == null) {
```
nswamy: why do we need to this only when a scope is not passed? these deallocate resources in a collection when using an existing scope.

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
```diff
@@ -179,6 +175,16 @@ object ResourceScope {
     }
   }
 
+  private[mxnet] def usingIfScopeExists[A](scope: Option[ResourceScope])(body: => A): A = {
```
nswamy: why do you need this new method?

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
andrewfayres: Removed it before I saw your comment but I'll explain what I was doing anyway.

These keep the return from being deallocated. When explicitly passing in a scope that you want all new native resources to be in, this code removes it from that scope and puts it in whatever happens to be the scope which is 2nd to last. I was considering making that not happen when you pass in a scope (because presumably you want resources to all be at the scope level you specified) but changed my mind.

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
andrewfayres: Yeah so originally I wanted to use the .using method directly but ran into some problems. There are times when we need to add new native resources to the same scope of the parent but there's no guarantee that the parent is actually in an existing scope.

When this happens it leads to a few issues. First, the parent is in None scope which we cannot pass to ResourceScope.using. Second, we still want to execute the body and allocate all the new resources but don't want to want to make a new ResourceScope because all those new resources will disappear with the new scope.

Alternatives that I thought of were: 1.) to have the caller check whether or not it was in a scope and handle it appropriately. This is ugly and puts the onus on the callers in multiple places. 2.) Changing the using method to work with a None scope. I opted not to do this because it complicates that method and I believe would require changing the method parameters which I didn't want to do. 3.) Changing the default scope to be something other than None. This is probably a reasonable solution. Maybe we have some kind of base scope or something similar. That's likely to be a fairly significant change in both the design and behavior of this class.

andrewfayres:
Found a new bug last night where many of the optimizers cause a seg fault when used with FeedForward in 1.4.0. Surprisingly, no one has filed an issue for this so I went ahead and did so. It's fixed in this PR.

Resolves #14498

andrewfayres:
CI seems to be experiencing unrelated failures. I'll restart this later once they appear to be resolved.

nswamy:
I am surprised that the tests did not catch the seg-faults, I ran the code through several examples for many days and did not see the seg-faults.

nswamy:
Feedforward is to be deprecated and Module should be used instead, we never deprecated that module, may be we should consider deprecating that.

scala-package/core/src/test/scala/org/apache/mxnet/ResourceScopeSuite.scala
```diff
@@ -101,6 +101,38 @@ class ResourceScopeSuite extends FunSuite with BeforeAndAfterAll with Matchers {
...
+        b = new TestNativeResource()
+        ResourceScope.usingIfScopeExists(a.scope) {
+          c = new TestNativeResource()
+          ResourceScope.using() {
+            d = new TestNativeResource()
+          }
+          assert(d.isDisposed == true)
+        }
+        assert(b.isDisposed == false)
+        assert(c.isDisposed == false)
```
nswamy: you should also check if (c.scope == a.scope)

nswamy:
summary of offline discussion:
* The method is private, this needs to be re-evaluate if we can do without adding scope twice into threadlocal. (ResourceScope:L162)
* There is a bug in using method that closes the passed in scope(it should not), we will track a JIRA to track and resolve.

scala-package/core/src/test/scala/org/apache/mxnet/ResourceScopeSuite.scala
andrewfayres: Good suggestions. I've added it.

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
```diff
@@ -179,7 +192,7 @@ object ResourceScope {
     * @param r ResourceScope to remove
     */
   private[mxnet] def removeFromThreadLocal(r: ResourceScope): Unit = {
-    threadLocalScopes.get() -= r
+    threadLocalScopes.get().remove(threadLocalScopes.get().lastIndexOf(r))
```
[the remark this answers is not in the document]
andrewfayres: We discussed this offline so I'm resolving.

Summary of our discussion:
This is essentially being more explicit but is potentially slower. We don't expect many ResourceScopes in the queue so the difference should be minimal. Also, if things work as is assumed by the previous implementation, r will always be in the last index and will be the first element tried.

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
```diff
@@ -160,6 +162,17 @@ object ResourceScope {
     }
   }
 
+  private[mxnet] def usingIfScopeExists[A](scope: Option[ResourceScope])(body: => A): A = {
+    if (scope == None) {
+      body
+    } else {
+      ResourceScope.addToThreadLocal(scope.get)
```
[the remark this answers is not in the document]
andrewfayres: Discussed this offline so resolving.

Summary of discussion:
There current implementation of ResourceScope always uses the scope on the top of the stack as the current scope. So to get around this we are pushing the supplied scope onto the stack, running the code, then popping the scope off the stack. The alternative is to redesign ResourceScope so that it keeps track of it's current scope and it could be set. I'm creating a jira ticket to look into that possibility.

scala-package/core/src/main/scala/org/apache/mxnet/ResourceScope.scala
```diff
@@ -48,8 +48,10 @@ class ResourceScope extends AutoCloseable {
     */
   override def close(): Unit = {
     ResourceScope.removeFromThreadLocal(this)
-    resourceQ.foreach(resource => if (resource != null) resource.dispose(false) )
-    resourceQ.clear()
+    if (!ResourceScope.threadLocalScopes.get().contains(this)) {
```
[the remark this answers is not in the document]
andrewfayres: Discussed offline so resolving.

Summary of offline discussion:
The private UsingIfExists method adds an existing scope onto the stack. We don't want to clear that scope when it's removed from the stack and this if should make sure we don't.

nswamy:
Thanks for nailing down this issue, great Job 💯 Please link the JIRA we discussed for fixing the scope closing if passed

andrewfayres:
Link to Jira for looking into & fixing the ResourceScope.using closing scopes that are passed in.

https://issues.apache.org/jira/browse/MXNET-1376

PR #14377, opened 2019-03-09
Correct update count with Gluon trainer and update_on_kvstore=False

### LINKED ISSUE #12713: distributed kvstore bug in MXNet
I'm using distributed kvstore with Gluon trainer. I found the two following bugs:

1. Initializing `trainer = gluon.Trainer(update_on_kvstore=True)` doesn't work. Inspecting `trainer._update_on_kvstore` shows that the value is still set to `False`. 

2. When distributed kvstore is used, by default `gluon.Trainer` doesn't work with `mx.optimizer.LRScheduler` if a worker has more than 1 GPU. To be more specific, the trainer updates once per GPU, the `LRScheduler` object is shared across GPUs and get a wrong update count. 

This means one cannot train imagenet classification using resnet with gluon trainer.

### LINKED ISSUE #13752: Adam, AdaMax and FTML cannot be used with Trainer(update_on_kv=False)
These optimizers adjust the learning rate by the number of optimization steps, which is problematic if `update_on_kv=False` is set and multiple GPUs share the same optimizer object (leading to incorrect count of optimization steps). For example: https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/optimizer/optimizer.py#L1093

cc @szha @sandeep-krishnamurthy

ptrendx:
## Description ##
This PR fixes the update count when Gluon trainer is created with multiple devices and `update_on_kvstore=False`.
Fixes #13752, fixes #12713

@eric-haibin-lin This slightly hacky but should work for all cases. Thoughts?

### Essentials ###

## Comments ##
- The real fix here should be moving the _index_update_count to Updater (which is per device) from optimizer (which is common to all updaters), but that would require breaking API change. FYI @szha

### DIFF
--- a/python/mxnet/optimizer/optimizer.py
+++ b/python/mxnet/optimizer/optimizer.py
@@ -106,7 +106,8 @@ class Optimizer(object):
         self.wd_mult = {}
         self.begin_num_update = begin_num_update
         self.num_update = begin_num_update
-        self._index_update_count = {}
+        self._all_index_update_counts = {0 : {}}
+        self._index_update_count = self._all_index_update_counts[0]
         self.clip_gradient = clip_gradient
         self.multi_precision = multi_precision
         self.aggregate_num = 0
@@ -379,6 +380,11 @@ class Optimizer(object):
                 if name in attr and '__wd_mult__' in attr[name]:
                     self.wd_mult[name] = float(attr[name]['__wd_mult__'])
         self.wd_mult.update(args_wd_mult)
+
+    def set_current_context(self, ctx):
+        if ctx not in self._all_index_update_counts:
+            self._all_index_update_counts[ctx] = {}
+        self._index_update_count = self._all_index_update_counts[ctx]
 
     def _update_count(self, index):
         """Updates num_update.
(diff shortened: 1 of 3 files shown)


karan6181:
@mxnet-label-bot add [Bug, Gluon, Optimizer, pr-awaiting-review]

karan6181:
@ptrendx Thank you for the contribution!

@eric-haibin-lin Can you please review this PR?

python/mxnet/optimizer/optimizer.py
```diff
@@ -380,6 +381,11 @@ def set_wd_mult(self, args_wd_mult):
                     self.wd_mult[name] = float(attr[name]['__wd_mult__'])
         self.wd_mult.update(args_wd_mult)
 
+    def set_current_context(self, ctx):
+        if ctx not in self._all_index_update_counts:
```
szha: rename to _set_current_context, and document the type of ctx?

ptrendx: Will do

ptrendx: Done.

PR #14338, opened 2019-03-06
Bypass ThreadedEngine in test_convolution_multiple_streams.

DickJC123:
## Description ##
Per discussion with @szha, this PR stabilizes the CI by eliminating errors in test_operator_gpu.py:test_convolution_multiple_streams that began with the merging of the subsequent PR https://github.com/apache/incubator-mxnet/pull/14223.  The segfaults and python double-free aborts currently seen in this test only appear when run with MXNET_ENGINE_TYPE=ThreadedEngine (which uses the ThreadedEnginePooled class).  It is hoped that @arcadiaphy will submit a follow-up PR to enable his improvements to run under the ThreadedEngine.

The test being modified was introduced by https://github.com/apache/incubator-mxnet/pull/14006.

This a the first step to resolving https://github.com/apache/incubator-mxnet/issues/14329.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [X ] Changes are complete (i.e. I finished coding on this PR)
- [X ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [X ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

Issue #14329, opened 2019-03-05
[Flaky] flaky test in test_operator_gpu.test_convolution_multiple_streams


szha:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-14324/3/pipeline#step-682-log-1293

@DickJC123

piyushghai:
@mxnet-label-bot Add [Test, Flaky]

DickJC123:
Looking at this now.  The problem occurs during shutdown of the system when MXNET_ENGINE_TYPE=ThreadedEngine (so the ThreadedEnginePooled).  I'll first try to determine if it's related to the dual stream work, or whether it's just that the dual-stream PR, in expanding the meager testing of the ThreadedEnginePooled, uncovered an issue with it.

DickJC123:
I believe I've isolated the problem.  I started by checking out the commit that introduced the dual stream facility and test (5f32f32e8).  I could run the test 20 times without failure.  I then examined behavior under the other 19 commits to master that were made since then.  The problem behavior first appears with:
```
053ffc7f8  2019-03-03  Wang Jiajun            fix memory-related issues to enable ASAN tests (#14223)
```
I ran the test_convolution_multiple_streams test with the engine list pruned down to just 'ThreadedEngine' (so ThreadedEnginePooled) for speed of testing.  It failed with the `/usr/bin/python3': double free or corruption (out)` abort at the 6th iteration.  I then backed up my testing to the previous commit:
```
0e23a18da  2019-03-02  Yuxi Hu                add symbolic link to mkldnn header files in include (#14300)
```
...and saw no failures after 40 repetitions of the full test_convolution_multiple_streams test. I even left the mshadow and dmlc-core versions the same as the 053ffc7f8 commit, which advances their versions.

In inspecting the change introduced by 053ffc7f8, I see changes to threaded_engine.cc as well as many free/delete operations added.  From this I conclude that 053ffc7f8 is the likely culprit and is not compatible with the ThreadedEnginePooled.  @szha @eric-haibin-lin @ptrendx

junrushao1994:
http://jenkins.mxnet-ci.amazon-ml.com/blue/rest/organizations/jenkins/pipelines/mxnet-validation/pipelines/unix-gpu/branches/PR-14192/runs/5/nodes/273/log/?start=0

======================================================================
[...]

PR #14223, opened 2019-02-21
fix memory-related issues to enable ASAN tests

arcadiaphy:
## Description ##
Continuing the discussion in #14176, this PR fixes memory-related issues detected by ASAN:
1. DeleteVar operations are still executed in engine shutdown phase to cleanup memory
2. Mark DeleteOperator as DeleteVar
3. Uncomment free operation in object pool
4. Free unreleased pointers of executer and optimizer in cpp examples
5. Fix heap overflow in classification example

Currently [ASAN tests](https://github.com/apache/incubator-mxnet/blob/master/ci/docker/runtime_functions.sh#L1032) is non blocking, after this PR, the checks are green.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

Issue #309, opened 2015-10-17
uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument


pluskid:
While testing on a simple MNIST example on the Julia binding, I noticed the following error while the program finish running.
```
libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument

signal (6): Abort trap: 6
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
```
Before trying to dig out what is happening here, is this kind of error look familiar to anyone? Here are some more information:

* It only happens under Mac OS X. I also see this error on the unittest in travis OS X node, but locally unittest will not trigger this error. I have never see this error under Linux.
* It occurs only when the program finished running.

tqchen:
Might due to module unloading and de-allocation of things after module destruction. I am not sure exactly what it could be though.

pluskid:
Hmm... that could be possible. But even if I disabled all calls to `MX***Free` destructors, I still get the same error. Is there anything else that should be called other than `MXNotifyShutdown` to properly shutdown everything?

tqchen:
No, there is no need to call other things, MXNotifyShutdown is only a hint for the engine, which is also not needed, if the thing is reproducible , can you check where the lock exactly is?

pluskid:
OK, I managed to get some back trace:

```
* thread #1: tid = 0x38acd5, 0x00007fff8b9e80ae libsystem_kernel.dylib`__pthread_kill + 10, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fff8b9e80ae libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff95517500 libsystem_pthread.dylib`pthread_kill + 90
    frame #2: 0x00007fff9081137b libsystem_c.dylib`abort + 129
    frame #3: 0x00007fff8f77ff81 libc++abi.dylib`abort_message + 257
    frame #4: 0x00007fff8f7a5a2f libc++abi.dylib`default_terminate_handler() + 243
    frame #5: 0x00007fff9879e0a2 libobjc.A.dylib`_objc_terminate() + 124
    frame #6: 0x00007fff8f7a319e libc++abi.dylib`std::__terminate(void (*)()) + 8
    frame #7: 0x00007fff8f7a3213 libc++abi.dylib`std::terminate() + 51
[...]

PR #14382, opened 2019-03-10
fix engine crash in shutdown phase

arcadiaphy:
## Description ##
@DickJC123 has reported the engine crash in #14329 which is related to my early PR on ASAN fix (#14223), this PR tries to fix the crash and restores the CI bypass in #14338.

After reproducing the crash in docker, I found it happens when trying to destruct resources using `Storage` singleton in engine shutdown phase. So I added shared_ptr in threaded engine to force `Storage` to destruct after engine cleanup.

Thus, the crash has two properties:
1. It doesn't happen all the time and is hard to reproduce because the destruction order of static variable is almost undefined in c++, which causes many similar bugs, #309 for example. For the time being, we can only add patches of shared_ptr when crash happens.
2. It happens in `test_convolution_multiple_streams` because sub-process is used to run tests, leading to 6 engine shutdown phases in one single test.

In order to verify the fix, 100 times of `test_convolution_multiple_streams` is executed without crash.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/src/engine/threaded_engine.h
+++ b/src/engine/threaded_engine.h
@@ -38,6 +38,7 @@
 #include <mutex>
 #include <string>
 #include <thread>
+#include <mxnet/storage.h>
 #include "./engine_impl.h"
 #include "../profiler/profiler.h"
 #include "./openmp.h"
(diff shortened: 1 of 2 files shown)


wkcn:
The PR has been merged.
Thanks for your contribution!

Issue #13234, opened 2018-11-12
waitforall() hides errors


samskalicky:
mx.nd.waitforall() doesnt check for errors and rethrow any that have occurred users will get an error when reading the output data. But I ran into an issue where I was gathering performance data, measuring elapsed time over an inference via model.forward() and mx.nd.waitforall() but I had hit an error and it was being hidden by waitforall(). I would have gotten the error had I actually read the output data, though. 

For this issue, we should fix waitforall() to have the ability to check for errors on each of the outputs.

anirudh2290:
this is by design and has been documented in the PR: https://github.com/apache/incubator-mxnet/pull/9681 . The waitforall is generally used for synchronization and benchmarking purposes. If you want to throw the exception you can call wait to read on the variables before or after the waitall call. Adding waitall support requires a global exception ptr and complicates the implementation. To workaround you should read the the output data before or after caliing waitall.

lupesko:
@mxnet-label-bot add [Feature Request]

lupesko:
@mxnet-label-bot add [NDArray]

PR #14397, opened 2019-03-12
Add exception handling support for waitall

anirudh2290:
## Description ##
Add exception handling support for waitall
Fixes: https://github.com/apache/incubator-mxnet/issues/13234, Fixes: #14426 

### Essentials ###

### Changes ###
- [x] added exception handling support for waitall
- [x] added tests
- [x] modified documentation

## Comments ##


anirudh2290:
We had earlier decided this based on complication associated with adding global exception_ptr and having to reset all the exception_ptrs corresponding to vars and ops.Another, reason was the performance impact. I think we overestimated the difficulty since the decision to used shared_ptr<exception_ptr> enables us to modify the exception_ptrs associated with vars and ops by just setting it  to nullptr. Also, there would be performance implication if there is an exception in the code and this can be more pronounced for a tool supposed to used only for benchmarking. But customers seems to be using it outside benchmarking and the performance impact will only happen when there is an exception thrown so it should be acceptable.

karan6181:
@mxnet-label-bot add [Exception Handling, pr-work-in-progress]

apeforest:
This is good stuff. waitall() is needed when we want to synchronize on multiple tensors.

src/engine/threaded_engine.h
```diff
@@ -177,7 +177,11 @@ class ThreadedVar final
   static std::atomic<std::size_t> counter;
   ~ThreadedVar() { LOG(INFO) << __func__ << " " << --counter; }
 #endif  // ENGINE_DEBUG
-  /*! \brief exception_ptr associated with the ThreadedVar */
+  /*!
+   * \brief exception_ptr associated with the ThreadedOpr
+   * cannot modify state of exception object since dereferencing
+   * exception_ptr is undefined behavior. Using shared_ptr to hold
```
yuxihu: Look likes [exception_ptr](https://en.cppreference.com/w/cpp/error/exception_ptr) is already a shared-ownership smart pointer. Would that be sufficient for our use cases, instead of wrapping it with shared_ptr?

src/engine/threaded_engine.h
```diff
@@ -432,6 +440,9 @@ class ThreadedEngine : public Engine {
   };
   /*! thread local store for bulk */
   typedef dmlc::ThreadLocalStore<BulkStatus> BulkStatusStore;
+  /*! shared_ptr to exception_ptr, used for exception handling */
+  typedef std::shared_ptr<std::exception_ptr> ExceptionRef;
```
yuxihu: can we also use ExceptionRef in other locations where we use std::shared_ptr<std::exception_ptr>？

docs/architecture/exception_handling.md
```diff
@@ -123,6 +123,3 @@ except mx.base.MXNetError as ex:
 d.asnumpy()
 ```
 
-### Limitation
```
yuxihu: shall we call out the potential performance impact of using waitall?

anirudh2290: yes good point, i can add that.

src/engine/threaded_engine.h
anirudh2290: sure

src/engine/threaded_engine.h
anirudh2290: the purpose is to update the object that exception ptr points to and dereferencing an exception_ptr is undefined behavior. So wrapping it in shared_ptr is another way.

anirudh2290:
this is ready for review

anirudh2290:
The CI stage is complete but shows pending. Any idea @marcoabreu @lebeg

src/engine/threaded_engine.cc
```diff
@@ -428,6 +447,14 @@ inline void ThreadedEngine::OnComplete(ThreadedOpr* threaded_opr) {
   for (auto&& i : threaded_opr->mutable_vars) {
     if (threaded_opr->opr_exception && *threaded_opr->opr_exception) {
       i->var_exception = threaded_opr->opr_exception;
+      // add current operator exceptions to global exceptions if not already
+      // added
+      auto it = std::find(global_exception_refs_.begin(),
```
yuxihu: L452-L457 are used in three places. Can we make it a function?

src/resource.cc
```diff
@@ -190,12 +190,14 @@ class ResourceManagerImpl : public ResourceManager {
     cpu_rand_->Seed(seed);
     cpu_parallel_rand_->Seed(seed);
 #if MXNET_USE_CUDA
-    gpu_rand_.Get(ctx.dev_id, [ctx, seed, this]() {
-      return new ResourceRandom<gpu>(ctx, seed);
-    })->Seed(seed);
-    gpu_parallel_rand_.Get(ctx.dev_id, [ctx, seed, this]() {
-      return new ResourceParallelRandom<gpu>(ctx, gpu_native_rand_copy_, seed);
-    })->Seed(seed);
+    if (ctx != Context::CPU()) {
```
yuxihu: Was this a bug? Shall we check ctx == Context::GPU()? There are other non-GPU context.

src/engine/threaded_engine.cc
anirudh2290: Thanks for the review ! I have moved it into a function.

src/resource.cc
anirudh2290: yes this is a bug. I have changed the check for GPU context.

src/engine/threaded_engine.cc
```diff
@@ -415,6 +415,25 @@ void ThreadedEngine::WaitForAll() {
   finished_cv_.wait(lock, [this]() {
       return pending_.load() == 0 || kill_.load();
     });
+  std::exception_ptr tmp;
+  if (!global_exception_refs_.empty()) {
+    // iterate through all exception refs
+    for (auto itr = global_exception_refs_.begin();
```
larroy: shall we use range for with const reference? is much less noisy.

src/engine/threaded_engine.h
```diff
@@ -60,6 +60,9 @@ namespace engine {
 // Forward declarations
 struct ThreadedOpr;
 
+/*! shared_ptr to exception_ptr, used for exception handling */
+typedef std::shared_ptr<std::exception_ptr> ExceptionRef;
```
larroy: Do we need to wrap it in a shared_ptr? exception_ptr has already shared ptr semantics according to cppreference.

larroy: Also, the name Ref to me is confusing, why not Ptr? why add a suffix of the type at all?

src/engine/threaded_engine.cc
```diff
@@ -415,6 +415,25 @@ void ThreadedEngine::WaitForAll() {
...
+         itr != global_exception_refs_.end(); ++itr) {
+      const ExceptionRef& ptr = *itr;
+      // the first exception will be saved to be rethrown later
+      if (*ptr != nullptr && !tmp) {
```
larroy: can be evaluated in bool context, so less noise.

src/engine/threaded_engine.cc
```diff
@@ -415,6 +415,25 @@ void ThreadedEngine::WaitForAll() {
   finished_cv_.wait(lock, [this]() {
       return pending_.load() == 0 || kill_.load();
     });
+  std::exception_ptr tmp;
```
larroy: Shall this code be wrapped in a function?

tests/python/unittest/test_exc_handling.py
```diff
@@ -77,46 +86,83 @@ def gluon(exec_wait=True):
...
-    except MXNetError:
-        caught = True
-    assert caught, "No exception thrown"
-    try:
-        b = mx.nd.random.normal(0, -1, (2, 2)).copyto(default_context())
-        b.wait_to_read()
-    except MXNetError:
-        caught = True
-    assert caught, "No exception thrown"
+    def multiple_waits(waitall=False):
```
larroy: does it make sense to use "@raises"?  maybe it would be easier to read.

https://nose.readthedocs.io/en/latest/testing_tools.html

At least a small comment explaining the test approach for future readers and that we expect exception to be thrown, is that the intent?

src/engine/threaded_engine.h
anirudh2290: exception_ptr cannot be dereferenced , so we cannot update the exception object it is pointing to or make it nullptr. Since this is a requirement for us we wrapped it in a shared_ptr. Used ref to make it consistent with other places in MXNet.

tests/python/unittest/test_exc_handling.py
anirudh2290: I have added comments. Intention is to test multiple wait_to_reads and waitalls for vars in same scope.

src/engine/threaded_engine.cc
anirudh2290: changed

src/engine/threaded_engine.cc
anirudh2290: changed

src/engine/threaded_engine.cc
anirudh2290: Currently it is used only once so it is fine to not use a function.

src/engine/threaded_engine.cc
anirudh2290: Thanks for the review!

src/engine/threaded_engine.cc
larroy: Maybe then a better variable name than tmp? ex_to_rethrow?

src/engine/threaded_engine.cc
```diff
@@ -415,6 +415,23 @@ void ThreadedEngine::WaitForAll() {
   finished_cv_.wait(lock, [this]() {
       return pending_.load() == 0 || kill_.load();
     });
+  std::exception_ptr exception_to_rethrow = nullptr;
+  if (!global_exception_refs_.empty()) {
+    // iterate through all exception refs
+    for (const auto& global_exception_ref : global_exception_refs_) {
+      // the first exception will be saved to be rethrown later
```
access2rohit: What is the order of exceptions stored in the "global_exception_refs_" ? If we are throwing the first one then is it the innermost in the stack that causes all other exceptions or the outermost ? If its outermost then it might not give correct idea about what was the root cause

anirudh2290: @access2rohit the order of the exceptions will be maintained exception thrown first will be rethrown first.

piyushghai:
@anirudh2290 Is this PR good to merge now ?

wkcn:

ThomasDelteil:
This PR likely broke the profiler tutorial, http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/NightlyTestsForBinaries/detail/master/275/pipeline/
@anirudh2290 any idea why? https://mxnet.incubator.apache.org/versions/master/tutorials/python/profiler.html

anirudh2290:
checking this now.

Issue #14391, opened 2019-03-11
as_in_context() does not copy the gradient to the new context


ThomasDelteil:
I think it should copy the gradient as well automatically?

```python
ctx = mx.cpu()
a = mx.nd.ones((10,10), ctx=ctx)
a.attach_grad()
print(a.context)
print(a.grad.context)
ctx = mx.gpu()
a = a.as_in_context(ctx)
print(a.context)
print(a.grad.context)
```

```
cpu(0)
cpu(0)
gpu(0)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-16-deab22190603> in <module>()
      5 ctx = mx.gpu()
      6 a = a.as_in_context(ctx)
----> 7 print(a.context, a.grad.context)

AttributeError: 'NoneType' object has no attribute 'context'
```

wkcn:
`as_in_context` will return a new copied NDArray object, but it doesn't copy the gradient.
If you need to copy the gradient, `a.grad.as_in_context(ctx)` is okay.

wkcn:
I think it Is not necessary to copy the gradient automatically, because the gradient is not used usually, it may be useless to copy automatically.

ThomasDelteil:
I disagree I think it would make more sense to copy the gradient as well as the tensor. It doesn't make sense for them to be on different contexts?
Also a variable that had gradient attached after being moved does not have a gradient attached anymore, that seems clearly a bug to me?

wkcn:
You are right. Let me think about a lazy method to return the gradient.

wkcn:
Is it good to define the function def as_in_context(ctx, ignore_grad=True)?
Because sometime we do not need to copy the gradient, and I do not how to copy the gradient lazily.

ThomasDelteil:
If we are worried about breaking APIs, I think we could start with a `as_in_context(ctx, copy_grad=False)` to avoid a breaking API change but for MXNet 2.0 I would suggest to move to `copy_grad=True` which makes more sense to me.

wkcn:
I'm worry `copy_grad=True` will drop the performance. Assume that it takes 1s to copy the data of a large tensor, it will takes extra 1s to copy the gradient.

ThomasDelteil:
In what case would you not want to copy gradients that has been attached?
Most as in context happened on data without gradient, I only discovered the
bug on some niche use case that requires data with gradient attached.

If the data does not have gradient attached the runtime would be the same
[...]

PR #14427, opened 2019-03-14
Copy the gradient to the new context when as_in_context() is called

wkcn:
## Description ##
Hi! there.
Fixed: https://github.com/apache/incubator-mxnet/issues/14391
Now, ndarray.as_in_context() copies the gradient to the new context.

### Essentials ###

### Changes ###
- [x] Add a new arguments `copy_grad` in `mx.nd.as_in_context` and `mx.nd.copyto`
- [x] Add unit-test in `test_ndarray.py`

## Comments ##

### DIFF
--- a/python/mxnet/ndarray/ndarray.py
+++ b/python/mxnet/ndarray/ndarray.py
@@ -2047,7 +2047,7 @@ fixed-size items.
         self.copyto(res)
         return res
 
-    def copyto(self, other):
+    def copyto(self, other, copy_grad=True):
         """Copies the value of this array to another array.
 
         If ``other`` is a ``NDArray`` object, then ``other.shape`` and
@@ -2061,6 +2061,8 @@ fixed-size items.
         ----------
         other : NDArray or Context
             The destination array or context.
+        copy_grad : bool
+            Default `True`. Whether to copy the gradient if it exists.
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


python/mxnet/ndarray/ndarray.py
```diff
@@ -2086,12 +2088,17 @@ def copyto(self, other):
...
         elif isinstance(other, Context):
             hret = NDArray(_new_alloc_handle(self.shape, other, True, self.dtype))
-            return _internal._copyto(self, out=hret)
+            data = _internal._copyto(self, out=hret)
         else:
             raise TypeError('copyto does not support type ' + str(type(other)))
+        if copy_grad:
+            grad = self.grad
+            if grad is not None:
+                setattr(data, '_old_grad', grad.copyto(other, copy_grad=False))
```
szha: This seems rather obscure as users aren't aware that this field exists. If user explicitly chose to do the copy, I think it's ok to set it in `.grad` attribute

szha: What's less clear to me is if copy_grad is true, the new grad array should be marked as the gradient of the new data array in the target context. Again, this seems to implicit.

wkcn: There is a case:
```python
a = mx.nd.array([1,2,3],ctx=mx.cpu())
a.attach_grad()
b = a.as_in_context()
print(b.grad) # return a gradient which is equal to a.grad
b.attach_grad()
print(b.grad) # return a new gradient
```
I need the attribute `_old_grad` to store the gradient.

wkcn:
@szha There is a discussion about whether to copy the gradient automatically.
https://github.com/apache/incubator-mxnet/issues/14391

PR #14431, opened 2019-03-14
Add post initialization callback for Gluon Parameter

yuxihu:
In Gluon, we defer initialization for some parameters until their shapes are known after the first forward pass. This makes it difficult for us to sync parameters among workers when we conduct distributed training using MXNet + Horovod with different random seeds for workers. More details can be found in this [Horovod issue](https://github.com/horovod/horovod/issues/895). 

In this PR, we add a post_init_callback to Gluon Parameter which will be called after the parameter is initialized. With this callback, parameters can be synced among workers once they are initialized. The typical usage of this callback in Horovod will be like this:
```
random.seed(hvd.local_rank())
data = ..
model = mx.gluon.nn.Dense(10)
model.initialize(post_init_callback=hvd.broadcast_parameters)
```

### DIFF
--- a/python/mxnet/gluon/block.py
+++ b/python/mxnet/gluon/block.py
@@ -483,7 +483,7 @@ class Block(object):
         return self
 
     def initialize(self, init=initializer.Uniform(), ctx=None, verbose=False,
-                   force_reinit=False):
+                   force_reinit=False, post_init_callback=None):
         """Initializes :py:class:`Parameter` s of this :py:class:`Block` and its children.
         Equivalent to ``block.collect_params().initialize(...)``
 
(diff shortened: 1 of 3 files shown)


yuxihu:
@eric-haibin-lin @apeforest @ctcyang please help review

yuxihu:
> why not do the synchronization after you know the initialization is complete? adding callback is generally not recommended as it's obscure to use.

It can be done like that. Users need to broadcast parameters in the Gluon train loop after the first forward pass. Based on my offline discussion with @eric-haibin-lin, we do not think that's a very good user experience. In addition, with the upcoming Gluon fit API, users may not need to write the trail loop anymore. In that case, we still need a way to sync the parameters.

szha:
@yuxihu I think this is a better fit in the fit API (excuse the pun), since @roywei's design has ways to add a plugin.

yuxihu:
@szha @eric-haibin-lin and I had a offline discussion and agreed that this design is not an optimal way to go. We will try to solve the issue in Horovod itself. Will close this PR once the new idea is validated.

@mxnet-label-bot update [pr-working-in-progress]

Issue #14406, opened 2019-03-13
resnet cpp-package test is broken


anirudh2290:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-14397/5/pipeline

after adding waitall support the resnet example is failing with cudamalloc out of memory error.

wkcn:
I wonder the memory of GPU in CI.
The input shape is (50,3,224,224), which may triggers OOM.

wkcn:
In addition, the model in cpp-package seems to be not convergent.

anirudh2290:
I think its running on a p3.8xlarge which should be sufficient to run this test. @marcoabreu can you confirm.

anirudh2290:
> In addition, the model in cpp-package seems to be not convergent.

yes i observed that too.

wkcn:
Since the input shape of ResNet is (3, 224, 224), so I resized the MNIST image (1, 28, 28) to (3, 224, 224).

marcoabreu:
We run on a g3.8xlarge

wkcn:
Changing batch size to a smaller value will address the OOM issue.

leleamol:
@marcoabreu  There are no changes to the alexnet.cpp or cpp-package recently. 
Are there any changes to underlying cuda or mxnet implementation.

These test were passing before.

ddavydenko:
Did infra that these tests are run on have changed recently? It seems that the test would be running fine on p3.8xl but would fail on g3.8x (legacy hardware)... @marcoabreu

anirudh2290:
as i said this happened in waitall change. waitall earlier used to hide exceptions, but with the PR: https://github.com/apache/incubator-mxnet/pull/14397 it is thrown. These problems would have been there from before but surfacing now.

leleamol:
I tried these examples with the recent code change in "WaitAll()" on p2.16x instances and c5.18x instances. I did not see the crash.

However,  we still need to add missing exception handling in the example so that we can prevent the crashes due to unhandled exceptions.

anirudh2290:
hi @leleamol . to reproduce you will have to use g3.8xlarge. I was able to reproduce on a g3.8xlarge.

wkcn:
Could someone please look the GPU memory used by the model?

anirudh2290:
[...]

PR #14441, opened 2019-03-15
[MXNET-1357] Fix the cpp-examples to add exception handling

leleamol:
## Description ##
The scope of this PR is to add missing exception handling to the examples. In future, if the underlying MXNet code throws any exception, the example will catch this exception and fail gracefully instead of crashing due to unhandled exception.

### Essentials ###
- [y] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [y] Changes are complete (i.e. I finished coding on this PR)
- [y] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###

## Comments ##

### DIFF
--- a/cpp-package/example/alexnet.cpp
+++ b/cpp-package/example/alexnet.cpp
@@ -228,6 +228,7 @@ int main(int argc, char const *argv[]) {
   }
 #endif
 
+  TRY
   /*net symbol*/
   auto Net = AlexnetSymbol(10);
 
@@ -352,5 +353,6 @@ int main(int argc, char const *argv[]) {
   delete exec;
   delete opt;
   MXNotifyShutdown();
+  CATCH
   return 0;
 }
--- a/cpp-package/example/utils.h
+++ b/cpp-package/example/utils.h
@@ -27,6 +27,15 @@
 
 using namespace mxnet::cpp;
 
+#define TRY \
+  try {
+#define CATCH \
+  } catch(...) { \
+    LG << "Status: FAIL";\
+    LG << "With Error: " << MXGetLastError(); \
+    return 1; \
+  }
+
 bool isFileExists(const std::string &filename) {
   std::ifstream fhandle(filename.c_str());
   return fhandle.good();
--- a/cpp-package/example/utils.h
+++ b/cpp-package/example/utils.h
@@ -30,9 +30,12 @@ using namespace mxnet::cpp;
 #define TRY \
   try {
 #define CATCH \
-  } catch(...) { \
+  } catch(dmlc::Error &err) { \
     LG << "Status: FAIL";\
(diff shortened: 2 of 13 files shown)


leleamol:
@anirudh2290

leleamol:
#14406

cpp-package/example/alexnet.cpp
```diff
@@ -228,6 +228,7 @@ int main(int argc, char const *argv[]) {
   }
 #endif
 
+  TRY
```
anirudh2290: can we just add exception handling to cpp-package the way it is done in python. Every C API call is inside a check_call function which checks the return code and throws corresponding frontend exception.

karan6181:
@mxnet-label-bot add [C++, Example, Exception Handling, pr-awaiting-response ]

leleamol:
@mxnet-jenkins  The unix-cpu and unix-gpu build failures are not related to the changes in PR.
Can we restart these builds?

cpp-package/example/alexnet.cpp
leleamol: It is certainly possible to add exception handling in cpp-package API itself.  However, it will require careful design and we may require to change the cpp-package API as well. 
We will certainly consider this enhancement as a part of C++ API enhancement project.

cpp-package/example/utils.h
```diff
@@ -27,6 +27,15 @@
 
 using namespace mxnet::cpp;
 
+#define TRY \
+  try {
+#define CATCH \
+  } catch(...) { \
+    LG << "Status: FAIL";\
```
anirudh2290: catch only dmlc::Error. For non backend exceptions MXGetLastError will be empty.

leleamol: Fixed.

cpp-package/example/utils.h
```diff
@@ -27,6 +27,18 @@
...
 using namespace mxnet::cpp;
 
+#define TRY \
+  try {
+#define CATCH \
+  } catch(dmlc::Error &err) { \
+    LG << "Status: FAIL";\
+    LG << "With Error: " << MXGetLastError(); \
+    return 1; \
+  } catch(...) { \
```
anirudh2290: i think you should only have dmlc::Error catch and remove the catch all. If there is any other exception thrown it could be because of some MXNet backend showing non dmlc::Errors (which is very unlikely) or some other exception thrown from the example code because of user or programming error. Users should be able to see what these exceptions are without having to modify utils.

piyushghai:
@anirudh2290 Are you comments addressed now in this PR ?

PR #14451, opened 2019-03-16
fix custom operation in fork

### LINKED ISSUE #14396: mx.nd.Custom not working in subprocess
mx.nd.Custom gets stuck at subprocess. 

The following code to replicate the error is from https://github.com/wkcn/MobulaOP/issues/40#issuecomment-471803878

```python
from concurrent import futures

import mxnet as mx
import sys

class AdditionOP(mx.operator.CustomOp):
    def __init__(self):
        super(AdditionOP, self).__init__()
        out_data[0][:] = in_data[0] + in_data[1]
        in_grad[0][:] = out_grad[0]
        in_grad[1][:] = out_grad[0]

@mx.operator.register("AdditionOP")
class AdditionOPProp(mx.operator.CustomOpProp):
    def __init__(self):
        super(AdditionOPProp, self).__init__()
    def list_arguments(self):
        return ['a', 'b']
    def list_outputs(self):
        return ['output']
        return in_shape, [in_shape[0]]
        return AdditionOP()

def foo():
    a = mx.nd.array([1, 2, 3])
    b = mx.nd.array([4, 5, 6])

    a.attach_grad()
    b.attach_grad()

    print("REC")
    with mx.autograd.record():
        c = mx.nd.Custom(a, b, op_type='AdditionOP')

    dc = mx.nd.array([7, 8, 9])
    c.backward(dc)

    print('Okay :-)')
    print('a + b = c \n {} + {} = {}'.format(a.asnumpy(), b.asnumpy(), c.asnumpy()))

def main():
    ex = futures.ProcessPoolExecutor(1)
    r = ex.submit(foo)
    r.result()

if __name__ == '__main__':
    main()
```

`asnumpy` gets stuck due to `mx.nd.Custom`

arcadiaphy:
## Description ##
Fix #14396. Update pthread_atfork function to properly fork on custom operator.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/src/initialize.cc
+++ b/src/initialize.cc
@@ -26,6 +26,7 @@
 #include <dmlc/logging.h>
 #include <mxnet/engine.h>
 #include "./engine/openmp.h"
+#include "./operator/custom/custom-inl.h"
 #if MXNET_USE_OPENCV
 #include <opencv2/opencv.hpp>
 #endif  // MXNET_USE_OPENCV
@@ -53,11 +54,14 @@ class LibraryInitializer {
 
 // disable openmp for multithreaded workers
 #ifndef _WIN32
+    using op::custom::CustomOperator;
     pthread_atfork(
       []() {
+        CustomOperator::Get()->Stop();
         Engine::Get()->Stop();
       },
       []() {
+        CustomOperator::Get()->Start();
         Engine::Get()->Start();
       },
       []() {
@@ -71,6 +75,7 @@ class LibraryInitializer {
 #endif  // MXNET_USE_OPENCV
         engine::OpenMP::Get()->set_enabled(false);
         Engine::Get()->Start();
+        CustomOperator::Get()->Start();
       });
 #endif
   }
--- a/src/initialize.cc
+++ b/src/initialize.cc
@@ -61,8 +61,8 @@ class LibraryInitializer {
         Engine::Get()->Stop();
       },
       []() {
-        CustomOperator::Get()->Start();
         Engine::Get()->Start();
+        CustomOperator::Get()->Start();
       },
       []() {
         // Conservative thread management for multiprocess workers
--- a/docs/tutorials/gluon/customop.md
+++ b/docs/tutorials/gluon/customop.md
@@ -216,11 +216,12 @@ print(y)
 ```
 
 ## Using custom operators with fork
-In Linux systems, the default method in multiprocessing to create process is by using fork. If there are unfinished async custom operations when forking, the program will be blocked because of python GIL.
+In Linux systems, the default method in multiprocessing to create process is by using fork. If there are unfinished async custom operations when forking, the program will be blocked because of python GIL. Always use sync calls like `wait_to_read` or `waitall` before calling fork.
 
 ```
 x = mx.nd.array([0, 1, 2, 3])
 y = mx.nd.Custom(x, op_type='sigmoid')
+# unfinished async sigmoid operation will cause blocking
 os.fork()
 ```
 
@@ -229,6 +230,7 @@ Correctly handling this will make mxnet depend upon libpython, so the workaround
 ```
 x = mx.nd.array([0, 1, 2, 3])
 y = mx.nd.Custom(x, op_type='sigmoid')
+# force execution by reading y
 print(y.asnumpy())
 os.fork()
 ```
--- a/docs/tutorials/gluon/customop.md
+++ b/docs/tutorials/gluon/customop.md
@@ -30,6 +30,7 @@ Custom operator in python is easy to develop and good for prototyping, but may h
 import numpy as np
 import mxnet as mx
 from mxnet import gluon, autograd
+import os
 ```
 
 ## Parameter-less operators
@@ -214,5 +215,22 @@ y = dense(x)
 print(y)
 ```
 
+## Using custom operators with fork
+In Linux systems, the default method in multiprocessing to create process is by using fork. If there are unfinished async custom operations when forking, the program will be blocked because of python GIL.
+
+```
+x = mx.nd.array([0, 1, 2, 3])
+y = mx.nd.Custom(x, op_type='sigmoid')
+os.fork()
+```
+
+Correctly handling this will make mxnet depend upon libpython, so the workaround now is to ensure that all custom operations are executed before forking process.
+
+```
+x = mx.nd.array([0, 1, 2, 3])
+y = mx.nd.Custom(x, op_type='sigmoid')
+print(y.asnumpy())
+os.fork()
+```
 
 <!-- INSERT SOURCE DOWNLOAD BUTTONS -->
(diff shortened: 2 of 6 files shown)


arcadiaphy:
@wkcn For the two questions:
1. Yes, each process has its independent threads. Fork only duplicates the caller thread, so we need to make sure all locking primitives are in valid states and restart the threads in child process. The easiest way is to restart CustomOperator when fork happens just like Engine does.
2. There is no fork on windows, so python use spawn method to create new process. I have no windows machine so I can only test on unix with:
```
import multiprocessing as mp
mp.set_start_method('spawn')
```
It seems like python re-import mxnet in child process when spawning, so the bug doesn't exist even without the fix.

arcadiaphy:
I find another strange bug, the custom operator hangs even without fork.
```
import mxnet as mx

class AdditionOP(mx.operator.CustomOp):
    def __init__(self):
        super(AdditionOP, self).__init__()
        out_data[0][:] = in_data[0] + in_data[1]
        in_grad[0][:] = out_grad[0]
        in_grad[1][:] = out_grad[0]

@mx.operator.register("AdditionOP")
class AdditionOPProp(mx.operator.CustomOpProp):
    def __init__(self):
        super(AdditionOPProp, self).__init__()
    def list_arguments(self):
        return ['a', 'b']
    def list_outputs(self):
        return ['output']
        return in_shape, [in_shape[0]]
        return AdditionOP()

def foo():
    a = mx.nd.array([1, 2, 3])
    b = mx.nd.array([4, 5, 6])
    c = mx.nd.Custom(a, b, op_type='AdditionOP')
    # mx.nd.waitall()

def main():
    foo()

if __name__ == '__main__':
    main()
```

I test the script on ubuntu, it hangs when calling python CFUNCTYPE function in c++. If you add `mx.nd.waitall()` to wait all async operations, then the bug disappears.

anirudh2290:
@arcadiaphy waitall silently hides exceptions, it doesnt rethrow the exception yet. This should be fixed by #14397 . what happenswhen you add c.wait_to_read() in foo . does it still hang ?

arcadiaphy:
@anirudh2290 No hangs after adding c.wait_to_read(). It only happens in calling CFUNCTYPE when destructing CustomOperator or running pre-fork Stop function.

karan6181:
@mxnet-label-bot add [Bug, Operator]

src/initialize.cc
```diff
@@ -53,11 +54,14 @@ class LibraryInitializer {
 
 // disable openmp for multithreaded workers
 #ifndef _WIN32
+    using op::custom::CustomOperator;
     pthread_atfork(
       []() {
+        CustomOperator::Get()->Stop();
         Engine::Get()->Stop();
```
anirudh2290: can we call Engine stop before customop stop. This will ensure that all ops pushed to engine have executed before stopping engine.

anirudh2290: I take that back. Maybe that won't work since Stopping engine before custom op means there may be ops that havent been pushed to engine by custom op but it has stopped. Another option is to add waitall after the worker joined in CustomOp stop to make sure all pushed ops have completed.

arcadiaphy: The second option doesn't work either. When it goes into Stop function, the custom worker thread has already frozen on python function, leading to the blocking of worker join. I cannot think of a simple and graceful way to complete the unfinished custom operation other than manually add waiting command in the code.

anirudh2290: do you know why it is frozen on worker join ? The worker should just push the operator to engine and return.

arcadiaphy: It's very easy to understand the first situation. The destruction of static variable on engine exit happens when python interpreter has been shutdown, so something bad happens when you call CFUNCTYPE function at that time.
I push a new commit to move custom Stop function to MXNotifyShutdown, forcing custom function called before python shutdown, so the first situation is all right.

arcadiaphy: The second situation of forking is so mysterious, I don't have a clue now. The custom Stop function is registered in prepare fork handler, everything should be OK at that time, but the  worker thread just freezes.

arcadiaphy:
I write a simple example to show the program freeze, which comes from calling CFUNCTYPE in thread when preparing fork.
[fork example](https://github.com/apache/incubator-mxnet/files/2977114/fork.tar.gz)

YutingZhang:
@arcadiaphy Does this also related to https://discuss.mxnet.io/t/custom-operator-cause-stuck-in-multicard-train/1452
Basically, a CustomOP could randomly get stuck in a multi-GPU setting. I also observed the same problem in my current experiment.

anirudh2290:
I think this may be related to https://docs.python.org/3/c-api/init.html#non-python-created-threads . 
Quoting from the doc:


> If you need to call Python code from these threads (often this will be part of a callback API provided by the aforementioned third-party library), you must first register these threads with the interpreter by creating a thread state data structure, then acquiring the GIL, and finally storing their thread state pointer, before you can start using the Python/C API. When you are done, you should reset the thread state pointer, release the GIL, and finally free the thread state data structure.

We don't have any such handling for callbacks in custom ops.

arcadiaphy:
@anirudh2290 The program blocks because of GIL, additional handling should be added to correctly run callbacks.
```
#include <iostream>
#include <thread>
#include <pthread.h>
#include <Python.h>
#include "./lib.h"

callback *fun_python;

void set_python_callback(callback *func) {
  fun_python = func;
}

class Foo {
public:
  Foo() {
    pthread_atfork([]() {
      // call fun_python successfully
      std::cout << fun_python(5) << std::endl;
      // call fun_python failed in thread
      Py_BEGIN_ALLOW_THREADS
      std::thread t([]() { std::cout << fun_python(10) << std::endl;});
      t.join();
      Py_END_ALLOW_THREADS
    }, NULL, NULL);
  }
};

static Foo f;
```
But adding this will make mxnet link against libpython, do we really want to do this?

arcadiaphy:
@YutingZhang I think the problem you mentioned is related to GIL, correctly handling GIL in c++ extension is a tricky problem, easy to cause deadlock. Now we have no such handling at all.

anirudh2290:
@arcadiaphy yes I don't see other option apart from depending on libpython if we have to fix these issues w.r.t custom op. What do others think ?

arcadiaphy:
@anirudh2290 If we don't want to depend on libpython, then this partial fix will work fine if you don't try to fork when having unfinished custom operation.

anirudh2290:
@arcadiaphy I think we can move forward with this PR as this is an improvement from what we have today. Can you document this as a limitation of using custom op with forking here:  http://mxnet.incubator.apache.org/versions/master/tutorials/gluon/customop.html , 
http://mxnet.incubator.apache.org/versions/master/faq/new_op.html. I think custom op may cause issues for multi-gpu scenario because of not handling the callbacks correctly, but I haven't found a reproducible step yet. Once found I will open an issue.

arcadiaphy:
@anirudh2290 Add some docs on custom op.

arcadiaphy:
@anirudh2290 Everything OK now.

PR #14487, opened 2019-03-21
[numpy] Fix unit tests after introducing numpy compatible shapes

reminisce:
1. Fixed a portion of failing tests after introducing numpy compatible shapes.
2. Added a thread-safe switch to turn on/off numpy compatibility. By default, it's off and existing tests should not be affected.

@junrushao1994 @szha @eric-haibin-lin @zheng-da @yzhliu

### DIFF
--- a/python/mxnet/rnn/rnn_cell.py
+++ b/python/mxnet/rnn/rnn_cell.py
@@ -78,6 +78,7 @@ def _normalize_sequence(length, inputs, layout, merge, in_layout=None):
 class RNNParams(object):
     """Container for holding variables.
     Used by RNN cells for parameter sharing between cells.
+
     Parameters
     ----------
     prefix : str
@@ -90,6 +91,7 @@ class RNNParams(object):
 
     def get(self, name, **kwargs):
         """Get the variable given a name if one exists or create a new one if missing.
+
         Parameters
         ----------
         name : str
@@ -105,6 +107,7 @@ class RNNParams(object):
 
 class BaseRNNCell(object):
     """Abstract base class for RNN cells
+
     Parameters
     ----------
     prefix : str, optional
@@ -137,12 +140,14 @@ class BaseRNNCell(object):
 
     def __call__(self, inputs, states):
         """Unroll the RNN for one time step.
+
         Parameters
         ----------
         inputs : sym.Variable
             input symbol, 2D, batch * num_units
         states : list of sym.Variable
             RNN state from previous step or the output of begin_state().
+
         Returns
         -------
         output : Symbol
@@ -153,6 +158,7 @@ class BaseRNNCell(object):
             The type of this symbol is same as the output of begin_state().
             This can be used as input state to the next time step
             of this RNN.
+
         See Also
         --------
         begin_state: This function can provide the states for the first time step.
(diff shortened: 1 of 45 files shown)


python/mxnet/rnn/rnn_cell.py
```diff
@@ -431,8 +431,8 @@ def __init__(self, num_hidden, prefix='lstm_', params=None, forget_bias=1.0):
 
     @property
     def state_info(self):
-        return [{'shape': (0, self._num_hidden), '__layout__': 'NC'},
-                {'shape': (0, self._num_hidden), '__layout__': 'NC'}]
+        return [{'shape': (-1, self._num_hidden), '__layout__': 'NC'},
```
reminisce: I will revert all the changes in this file since we have a global switch to control numpy compatibility now.

src/c_api/c_api_common.h
```diff
@@ -91,7 +91,7 @@ struct MXAPIThreadLocalEntry {
     data->resize(shapes.size());
     size_t size = 0;
     for (const auto& s : shapes) {
-      CHECK_GE(s.ndim(), 0);
+      if (s.ndim() > 0);
```
zheng-da: what is this?

src/operator/batch_norm_v1-inl.h
```diff
@@ -261,7 +261,7 @@ class BatchNormV1Prop : public OperatorProperty {
     using namespace mshadow;
     CHECK_EQ(in_shape->size(), 3U) << "Input:[data, gamma, beta]";
     const mxnet::TShape &dshape = in_shape->at(0);
-    if (!shape_is_known(dshape)) return false;
+    if (dshape.ndim() == -1) return false;
```
zheng-da: i don't understand this modification. ndim == -1 means unknown shape for numpy convention. ndim == 0 means unknown shape for mxnet convention. Why do we only need to check for numpy convention?

src/operator/contrib/multi_proposal-inl.h
```diff
@@ -108,7 +108,7 @@ class MultiProposalProp : public OperatorProperty {
     using namespace mshadow;
     CHECK_EQ(in_shape->size(), 3) << "Input:[cls_prob, bbox_pred, im_info]";
     const mxnet::TShape &dshape = in_shape->at(proposal::kClsProb);
-    if (!mxnet::op::shape_is_none(dshape)) return false;
+    if (mxnet::op::shape_is_none(dshape)) return false;
```
zheng-da: this simply reverses the condition check?

src/c_api/c_api_common.h
reminisce: `s.ndim()` can be -1 if the shape is unknown. In that situation, it should not contribute to the calculation of the total size.

src/operator/contrib/multi_proposal-inl.h
reminisce: Yes, it was done incorrectly in a previous PR.

python/mxnet/rnn/rnn_cell.py
```diff
@@ -78,7 +78,6 @@ def _normalize_sequence(length, inputs, layout, merge, in_layout=None):
 class RNNParams(object):
     """Container for holding variables.
     Used by RNN cells for parameter sharing between cells.
-
```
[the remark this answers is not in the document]
reminisce: Done.

src/c_api/c_api_common.h
eric-haibin-lin: why not remove this line of code?

PR #12850, opened 2018-10-17
Updated nvidia docker base image to CUDA 10.0

lebeg:
## Description ##

PR requires newly deployed AMIs with updated drivers.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #14513, opened 2019-03-25
Update base CUDA image for CI to v10.0 cuDNN 7.3.1

perdasilva:
## Description ##

 I believe CI isn't reporting some GPU test issues because both the version of cudnn in the environment [that we set](https://github.com/apache/incubator-mxnet/blob/master/ci/docker/runtime_functions.sh#L815) and the cudnn version [set by the image](https://gitlab.com/nvidia/cuda/blob/ubuntu16.04/9.1/devel/cudnn7/Dockerfile#L5) are lower than what is required for some tests [eg](https://github.com/apache/incubator-mxnet/blame/master/tests/python/gpu/test_gluon_gpu.py#L85) by the test. 

So, as it is in the example linked above, the test just ensures an error is raised. I don't know if (m)any of the cudnn functions are being tested right now. Maybe there are bugs slipping through.

This PR bumps the base image for GPU instances to CUDA v10.0. Furthermore, it updates the test functions to use the cuDNN version variable given by the environment, and only use 7.0.3 by default, if nothing is set.

### DIFF
--- a/ci/docker/Dockerfile.build.ubuntu_base_gpu
+++ b/ci/docker/Dockerfile.build.ubuntu_base_gpu
@@ -19,11 +19,16 @@
 # Dockerfile to run the MXNet Installation Tests on Ubuntu 16.04
 # This should run in an empty docker with ubuntu and cuda.
 
-FROM nvidia/cuda:9.1-cudnn7-devel
+FROM nvidia/cuda:10.0-devel-ubuntu16.04
+
+ENV CUDNN_VERSION=7.3.1.20
 
 WORKDIR /work/deps
 
 RUN apt-get update && apt-get -y install sudo
+
+COPY install/ubuntu_cudnn.sh /work/
+RUN /work/ubuntu_cudnn.sh
 
 ARG USER_ID=0
 ARG GROUP_ID=0
--- a/ci/docker/Dockerfile.build.ubuntu_build_cuda
+++ b/ci/docker/Dockerfile.build.ubuntu_build_cuda
@@ -21,7 +21,9 @@
 # package generation, requiring the actual CUDA library to be
 # present
 
-FROM nvidia/cuda:9.1-cudnn7-devel
+FROM nvidia/cuda:10.0-devel-ubuntu16.04
+
+ENV CUDNN_VERSION=7.3.1.20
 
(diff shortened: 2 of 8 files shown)


lebeg:
I can suggest to try to update to 10 instead, like here: https://github.com/apache/incubator-mxnet/pull/12850

abhinavs95:
@mxnet-label-bot add[pr-work-in-progress, Test, CI]

perdasilva:
I'm disabling the failing tests until we can figure out how to fix them. I think it's important to apply these changes to CI asap so that we don't have errors slipping through.

larroy:
Could you add more info in the description of the issue, like pointers to line numbers? Is not clear to me what is the problem.

larroy:
Who uses the variable "CUDNN_VERSION" ? when cuda is loaded?

lebeg:
> Who uses the variable "CUDNN_VERSION" ? when cuda is loaded?

Look [here](https://gitlab.com/nvidia/cuda/blob/ubuntu16.04/10.0/devel/cudnn7/Dockerfile#L5)

perdasilva:
@larroy I've updated the PR description with the links and examples. Sorry about the confusion.

perdasilva:
trying cuda 10.1, out of curiosity

perdasilva:
Trying to put the failing tests on P3 instances, which should have a higher computer and maybe more functionality available to them.

perdasilva:
I've split out the CUDNN_VERSION environment variable commit to its own PR (#14595) to see if it would get passed CI as it currently is.

perdasilva:
Updated the CI images to be based off *cuda10-base* and I'm manually installing a lower version of cudnn to test the theory that maybe we aren't compatible with cuDNN 7.5

perdasilva:
Re-running CI after rebase. If it's good, please merge. @stu1130 will explore fixing any issues that are brought by using cudnn 7.5 (#14652) and thereafter bring CI to 10.1.

perdasilva:
@mxnet-label-bot remove[pr-work-in-progress]

KellenSunderland:
Nice work @perdasilva @lebeg @marcoabreu

PR #14530, opened 2019-03-26
[WIP] (Do not merge) Removed conflicting GitHub status update

lebeg:
## Description ##

### Essentials ###

### Changes ###

## Comments ##


abhinavs95:
@lebeg Could you please provide a description for the PR?

abhinavs95:

ci/Jenkinsfile_utils.groovy
```diff
@@ -186,6 +186,42 @@ def update_github_commit_status(state, message) {
     context = get_github_context()
     echo "context=${context}"
 
+    echo "sending the github status 1..."
```
jlcontreras: Not familiar with groovy, but wouldn't a loop work here? Something like

```suggestion
for (int i = 0; i <3; i++) {
   echo String.format("sending the github status %d...", i+1)
   
   step([
      $class: 'GitHubCommitStatusSetter',
      reposSource: [$class: "ManuallyEnteredRepositorySource", url: repoUrl],
      contextSource: [$class: "ManuallyEnteredCommitContextSource", context: context],
      commitShaSource: [$class: "ManuallyEnteredShaSource", sha: commitSha],
      statusBackrefSource: [$class: "ManuallyEnteredBackrefSource", backref: "${env.RUN_DISPLAY_URL}"],
      errorHandlers: [[$class: 'ShallowAnyErrorHandler']],
      statusResultSource: [
        $class: 'ConditionalStatusResultSource',
        results: [[$class: "AnyBuildResult", message: message, state: state]]
      ]
    ])
    if(i < 2){
        sleep 1
    }  
}
```

ci/Jenkinsfile_utils.groovy
```diff
@@ -186,18 +186,27 @@ def update_github_commit_status(state, message) {
...
-      commitShaSource: [$class: "ManuallyEnteredShaSource", sha: commitSha],
-      statusBackrefSource: [$class: "ManuallyEnteredBackrefSource", backref: "${env.RUN_DISPLAY_URL}"],
-      errorHandlers: [[$class: 'ShallowAnyErrorHandler']],
-      statusResultSource: [
-        $class: 'ConditionalStatusResultSource',
-        results: [[$class: "AnyBuildResult", message: message, state: state]]
-      ]
-    ])
+    // a few attempts need to be made: https://github.com/apache/incubator-mxnet/issues/11654
+    for (int attempt = 1; attempt < 4; attempt++) {
```
vishaalkapoor: nit, use <= 3. it's easier to mentally parse at a glance. same with <= 2.

ci/Jenkinsfile_utils.groovy
```diff
@@ -186,18 +186,27 @@ def update_github_commit_status(state, message) {
...
+      echo "Sending GitHub status attempt ${attempt}..."
+
+      step([
```
vishaalkapoor: I don't understand few points here 
- will this step be repeated if it's successful? 
- is the operation idempotent? 
- is the backoff of 1 second x 3 empirically effective
- is there a more robust way to handle this planned? 
Thanks!

vishaalkapoor:
Thanks for working on this @lebeg - some comments added :)

ci/Jenkinsfile_utils.groovy
lebeg: @vishaalkapoor I mentioned the answers in the description, but here they are again explicitly:

> (why) will this step be repeated if it's successful? is there a way to see if the step succeeded and not iterate?

There is no known to me way to determine if step was a success. In case of a real failure the step just does nothing - the list of repositories it needs to update is empty.

> is the operation idempotent (in the event you cannot determine if the step succeeded and are forced to do the step several times)?

I needed to look in Wikipedia what [Idempotence](https://en.wikipedia.org/wiki/Idempotence) actually is and found out that yes, the operation is idempotent.

> What kind of outages is the backoff strategy aiming to mitigate and what is the duration and cause of those outages? 

The repository object can not be fetched from a [cache](https://github.com/jenkinsci/github-plugin/blob/26ac9b0d1c4702f5d0940201a7e8deb19d5a42c6/src/main/java/com/cloudbees/jenkins/GitHubRepositoryName.java#L163-L167) for some reason by [GitHubCommitStatusSetter](https://github.com/jenkinsci/github-plugin/blob/497f43251e3bfc130f125f75e45f73e06894361d/src/main/java/org/jenkinsci/plugins/github/status/GitHubCommitStatusSetter.java). This happens once in a while and became a significant problem because we have split the pipelines. 

> Is the backoff of 1 second x 3 empirically effective? Have you considered different strategies? 

We are aiming to test that on all the PR's. For the test job it didn't fail at all. If this simple strategy will not be enough we might look into other.

> is there a more robust way to handle this planned?

Potentially [another](https://wiki.jenkins.io/display/JENKINS/GitHub+Integration+Plugin) plugin might be used when it will be mature enough. Currently it doesn't work since it can not be configured properly.

ci/Jenkinsfile_utils.groovy
lebeg: Ok, changed as you suggested

ci/Jenkinsfile_utils.groovy
vishaalkapoor: Apologies for missing the first point which was addressed in the comments, but you didn't adderss backoff or idempotence :) If the back-off strategy isn't right, you may see the issue reappear with some low frequency. 

I'm good with committing because it will temporarily fix the issue at least (LGTM)

A suggestion if it's not in the comments already :) Can you repro with a minimal scenario and figure out some probabilities (to justify 4 back-offs and a 1 second delay)?

lebeg: > Can you repro with a minimal scenario and figure out some probabilities

There is a certain problem with that. The test job doesn't catch the failures, but I can see a metric in CloudWatch that this is happening 1-2 times per hour. Once this PR get merged it should have an immediate effect on the metric. If not - the strategy would need to be reconsidered.

ci/Jenkinsfile_utils.groovy
```diff
@@ -186,18 +186,27 @@ def update_github_commit_status(state, message) {
...
-      commitShaSource: [$class: "ManuallyEnteredShaSource", sha: commitSha],
-      statusBackrefSource: [$class: "ManuallyEnteredBackrefSource", backref: "${env.RUN_DISPLAY_URL}"],
-      errorHandlers: [[$class: 'ShallowAnyErrorHandler']],
-      statusResultSource: [
-        $class: 'ConditionalStatusResultSource',
-        results: [[$class: "AnyBuildResult", message: message, state: state]]
-      ]
-    ])
+    // a few attempts need to be made: https://github.com/apache/incubator-mxnet/issues/11654
+    for (int attempt = 1; attempt <= 3; attempt++) {
```
lanking520: It seemed not to be an attempt, it will run three times no matter what if the first time succeeded or not. Please correct me if I understand wrongly.

lebeg: @lanking520 yes, you understand correctly. Do you suggest to rename the variable?

lanking520: @lebeg Please check the attempt if it successful by checking the result (status:200).

lebeg: @lanking520 the return code of what you suggest to check?

perdasilva: @lanking520 the problem is that this call doesn't return a value to check against. The strategy here is really to just hit the github endpoint three times (instead of once) as a mitigation for the github status problem. We realise it's not elegant, and that it doesn't _solve_ the problem. But this is just a stop-gap measure to improve the the situation for the developers until we can finally fix the issue.

lebeg: Maybe I'll mention again here the reasons mentioned in the description of this PR for clarity:
> * Using a different Jenkins plugin GitHub Integration Plugin has been considered, but unfortunately after installation there was not way to configure it - the options were missing from the configuration screen
> * All other plugins required explicit credentials passing as arguments
> * No way of asking for the commit status for verification has been found to check the status after the update
> * No way for checking the list of repositories that need to be updated was found (and whether it's empty) - all what happens, happens internally in the plugin

PR #14214, opened 2019-02-20
Add infer_type_partial

ptrendx:
## Description ##
Add `mx.sym.infer_type_partial` similar to `mx.sym.infer_shape_partial`.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

PR #14532, opened 2019-03-26
Nightly test fix: add filter to warnings

ThomasDelteil:
## Description ##

nightly test fails because of warnings that have been introduced. However there are no ways to not get the warnings at the moment. Adding filters to ignore them.

That's a bad solution, but I think it's worst to simply ignore warnings in the nightly tests. We need to either remove the warnings that have been introduced in : https://github.com/apache/incubator-mxnet/pull/14214, suppress the warnings in the imports / onnx load API or update the API to specify the type at model loading time.

### DIFF
--- a/docs/tutorials/gluon/hybrid.md
+++ b/docs/tutorials/gluon/hybrid.md
@@ -157,6 +157,7 @@ to gluon with `SymbolBlock`:
 import warnings
 
 with warnings.catch_warnings():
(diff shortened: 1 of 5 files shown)


abhinavs95:
@mxnet-label-bot add [Test, pr-awaiting-review]

Ishitori:
I run it on my EC2 instance and all failed tutorials are working now.

PR #14535, opened 2019-03-26
[DOC] Updated install instructions for mac

larroy:
## Description ##
Updates simple install instructions for mac. Build from source.

### Essentials ###

### DIFF
--- a/cmake/cmake_options.yml
+++ b/cmake/cmake_options.yml
@@ -46,7 +46,8 @@ USE_SIGNAL_HANDLER: "ON" # Print stack traces on segfaults.
 USE_TENSORRT: "OFF" # Enable infeference optimization with TensorRT.
 USE_ASAN: "OFF" # Enable Clang/GCC ASAN sanitizers.
 ENABLE_TESTCOVERAGE: "OFF" # Enable compilation with test coverage metric output
-CMAKE_BUILD_TYPE: "Debug"
+USE_INT64_TENSOR_SIZE: "OFF" # Use int64_t to represent the total number of elements in a tensor
+CMAKE_BUILD_TYPE: "Debug" # Debug | Release | RelWithDebInfo | MinSizeRel
 CMAKE_CUDA_COMPILER_LAUNCHER: "ccache"
 CMAKE_C_COMPILER_LAUNCHER: "ccache"
 CMAKE_CXX_COMPILER_LAUNCHER: "ccache"
--- a/cmake/cmake_options_osx.yml
+++ b/cmake/cmake_options_osx.yml
@@ -31,7 +31,6 @@ USE_MKLDNN: "ON" # Use MKLDNN variant of MKL (if MKL found) IF USE_MKL_IF_AVAILA
 USE_OPERATOR_TUNING: "ON" # Enable auto-tuning of operators IF NOT MSVC
 USE_GPERFTOOLS: "ON" # Build with GPerfTools support (if found)
 USE_JEMALLOC: "ON" # Build with Jemalloc support
-USE_PROFILER: "ON" # Build with Profiler support
 USE_DIST_KVSTORE: "OFF" # Build with DIST_KVSTORE support
 USE_PLUGINS_WARPCTC: "OFF" # Use WARPCTC Plugins
 USE_PLUGIN_CAFFE: "OFF" # Use Caffe Plugin
@@ -47,7 +46,8 @@ USE_SIGNAL_HANDLER: "ON" # Print stack traces on segfaults.
 USE_TENSORRT: "OFF" # Enable infeference optimization with TensorRT.
 USE_ASAN: "OFF" # Enable Clang/GCC ASAN sanitizers.
 ENABLE_TESTCOVERAGE: "OFF" # Enable compilation with test coverage metric output
-CMAKE_BUILD_TYPE: "Debug"
+USE_INT64_TENSOR_SIZE: "OFF" # Use int64_t to represent the total number of elements in a tensor
+CMAKE_BUILD_TYPE: "Debug" # Debug | Release | RelWithDebInfo | MinSizeRel
 CMAKE_CUDA_COMPILER_LAUNCHER: "ccache"
 CMAKE_C_COMPILER_LAUNCHER: "ccache"
 CMAKE_CXX_COMPILER_LAUNCHER: "ccache"
--- a/cmake/cmake_options_osx.yml
+++ b/cmake/cmake_options_osx.yml
@@ -1,0 +1,53 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+--- # CMake configuration
+USE_CUDA: "OFF" # Build with CUDA support
+USE_OLDCMAKECUDA: "OFF" # Build with old cmake cuda
+USE_NCCL: "OFF" # Use NVidia NCCL with CUDA
+USE_OPENCV: "OFF" # Build with OpenCV support
+USE_OPENMP: "OFF" # Build with Openmp support
+USE_CUDNN: "ON" # Build with cudnn support) # one could set CUDNN_ROOT for search path
+USE_SSE: "ON" # Build with x86 SSE instruction support IF NOT ARM
+USE_F16C: "ON" # Build with x86 F16C instruction support) # autodetects support if "ON"
+USE_LAPACK: "ON" # Build with lapack support
+USE_MKL_IF_AVAILABLE: "ON" # Use MKL if found
+USE_MKLML_MKL: "ON" # Use MKLDNN variant of MKL (if MKL found) IF USE_MKL_IF_AVAILABLE AND (NOT APPLE)
+USE_MKLDNN: "ON" # Use MKLDNN variant of MKL (if MKL found) IF USE_MKL_IF_AVAILABLE AND (NOT APPLE)
+USE_OPERATOR_TUNING: "ON" # Enable auto-tuning of operators IF NOT MSVC
+USE_GPERFTOOLS: "ON" # Build with GPerfTools support (if found)
+USE_JEMALLOC: "ON" # Build with Jemalloc support
+USE_PROFILER: "ON" # Build with Profiler support
+USE_DIST_KVSTORE: "OFF" # Build with DIST_KVSTORE support
+USE_PLUGINS_WARPCTC: "OFF" # Use WARPCTC Plugins
+USE_PLUGIN_CAFFE: "OFF" # Use Caffe Plugin
+USE_CPP_PACKAGE: "OFF" # Build C++ Package
+USE_MXNET_LIB_NAMING: "ON" # Use MXNet library naming conventions.
+USE_GPROF: "OFF" # Compile with gprof (profiling) flag
+USE_CXX14_IF_AVAILABLE: "OFF" # Build with C++14 if the compiler supports it
+USE_VTUNE: "OFF" # Enable use of Intel Amplifier XE (VTune)) # one could set VTUNE_ROOT for search path
+ENABLE_CUDA_RTC: "ON" # Build with CUDA runtime compilation support
+BUILD_CPP_EXAMPLES: "ON" # Build cpp examples
+INSTALL_EXAMPLES: "OFF" # Install the example source files.
+USE_SIGNAL_HANDLER: "ON" # Print stack traces on segfaults.
+USE_TENSORRT: "OFF" # Enable infeference optimization with TensorRT.
+USE_ASAN: "OFF" # Enable Clang/GCC ASAN sanitizers.
+ENABLE_TESTCOVERAGE: "OFF" # Enable compilation with test coverage metric output
+CMAKE_BUILD_TYPE: "Debug"
+CMAKE_CUDA_COMPILER_LAUNCHER: "ccache"
+CMAKE_C_COMPILER_LAUNCHER: "ccache"
+CMAKE_CXX_COMPILER_LAUNCHER: "ccache"
--- a/docs/install/osx_setup.md
+++ b/docs/install/osx_setup.md
@@ -104,10 +104,21 @@ After you have installed the dependencies, pull the MXNet source code from Git a
 The file called ```osx.mk``` has the configuration required for building MXNet on OS X. First copy ```make/osx.mk``` into ```config.mk```, which is used by the ```make``` command:
 
 ```bash
-    git clone --recursive https://github.com/apache/incubator-mxnet ~/mxnet
-    cd ~/mxnet
-    cp cmake/cmake_options_osx.yml cmake_options.yml
-    ./dev_menu.py menu 1
+git clone --recursive https://github.com/apache/incubator-mxnet ~/mxnet
+cd ~/mxnet
+mkdir build && cd build
+cmake \
+    -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+    -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+    -DUSE_MKL_IF_AVAILABLE=O \
+    -DUSE_CPP_PACKAGE=ON \
+    -DUSE_CUDA=OFF \
+    -DUSE_OPENMP=OFF \
+    -DUSE_OPENCV=ON \
+    -DUSE_SIGNAL_HANDLER=ON \
+    -DCMAKE_BUILD_TYPE=Debug \
+    -GNinja ..
+ninja
 ```
 
 To build with MKLDNN
@@ -145,10 +156,11 @@ We have installed MXNet core library. Next, we will install MXNet interface pack
 To install the MXNet Python binding navigate to the root of the MXNet folder then run the following:
 
 ```bash
-$ cd python
-$ pip install -e .
+virtualenv -p`which python3` py3_venv
+source py3_venv/bin/activate
+pip install -e python
 ```
-
+First we create a [virtualenv](https://docs.python-guide.org/dev/virtualenvs/#lower-level-virtualenv) to isolate this installation from our global environment.
 Note that the `-e` flag is optional. It is equivalent to `--editable` and means that if you edit the source files, these changes will be reflected in the package installed.
 
 ## Install the MXNet Package for R
--- a/docs/install/osx_setup.md
+++ b/docs/install/osx_setup.md
@@ -114,7 +114,7 @@ cmake \
     -DUSE_OPENMP=OFF \
     -DUSE_OPENCV=ON \
     -DUSE_SIGNAL_HANDLER=ON \
-    -DCMAKE_BUILD_TYPE=Debug \
+    -DCMAKE_BUILD_TYPE=RelWithDebInfo\
     -GNinja ..
 ninja
 popd
--- a/docs/install/osx_setup.md
+++ b/docs/install/osx_setup.md
@@ -101,17 +101,15 @@ Install the dependencies, required for MXNet, with the following commands:
 ### Build MXNet Shared Library
 After you have installed the dependencies, pull the MXNet source code from Git and build MXNet to produce an MXNet library called ```libmxnet.so```. You can clone the repository as described in the following code block, or you may try the [download links](download.md) for your desired MXNet version.
 
-The file called ```osx.mk``` has the configuration required for building MXNet on OS X. First copy ```make/osx.mk``` into ```config.mk```, which is used by the ```make``` command:
-
 ```bash
 git clone --recursive https://github.com/apache/incubator-mxnet ~/mxnet
-cd ~/mxnet
+cd ~/mxnet && pushd .
 mkdir build && cd build
 cmake \
     -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
     -DCMAKE_C_COMPILER_LAUNCHER=ccache \
-    -DUSE_MKL_IF_AVAILABLE=O \
-    -DUSE_CPP_PACKAGE=ON \
+    -DUSE_MKL_IF_AVAILABLE=OFF \
+    -DUSE_MKLDNN=ON \
     -DUSE_CUDA=OFF \
     -DUSE_OPENMP=OFF \
     -DUSE_OPENCV=ON \
@@ -119,6 +117,7 @@ cmake \
     -DCMAKE_BUILD_TYPE=Debug \
     -GNinja ..
 ninja
+popd
 ```
 
 To build with MKLDNN
@@ -156,12 +155,15 @@ We have installed MXNet core library. Next, we will install MXNet interface pack
 To install the MXNet Python binding navigate to the root of the MXNet folder then run the following:
 
 ```bash
-virtualenv -p`which python3` py3_venv
-source py3_venv/bin/activate
+virtualenv -p`which python3` mxnet_py3
+source mxnet_py3/bin/activate
 pip install -e python
 ```
 First we create a [virtualenv](https://docs.python-guide.org/dev/virtualenvs/#lower-level-virtualenv) to isolate this installation from our global environment.
 Note that the `-e` flag is optional. It is equivalent to `--editable` and means that if you edit the source files, these changes will be reflected in the package installed.
+
+If you don't want to use virtualenv, a simple `pip install -e python` or `pip3 install -e python`
+will suffice.
 
 ## Install the MXNet Package for R
 You have 2 options:
--- a/docs/install/osx_setup.md
+++ b/docs/install/osx_setup.md
@@ -77,6 +77,10 @@ Install the dependencies, required for MXNet, with the following commands:
 
 ```bash
 	brew update
+    brew install python3
+	brew install cmake
+	brew install ninja
+	brew install ccache
 	brew install pkg-config
 	brew install graphviz
 	brew install openblas
@@ -91,7 +95,7 @@ Install the dependencies, required for MXNet, with the following commands:
 	# For visualization of network graphs
 	pip install graphviz==0.8.4
 	# Jupyter notebook
-	pip install jupyter
+	pip3 install jupyter
 ```
 
 ### Build MXNet Shared Library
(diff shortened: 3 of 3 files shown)


abhinavs95:
@mxnet-label-bot add [Doc, Installation, pr-awaiting-review]

apeforest:
I treid to follow this doc to use cmake on my Mac, but got the following error message:

```
-- Could NOT find MKL (missing: MKL_INCLUDE_DIR MKL_RT_LIBRARY) 
-- Found OpenBLAS libraries: /usr/local/opt/openblas/lib/libopenblas.dylib
-- Found OpenBLAS include: /System/Library/Frameworks/vecLib.framework/Headers
-- Could NOT find Gperftools (missing: GPERFTOOLS_LIBRARIES GPERFTOOLS_INCLUDE_DIR) 
-- Found PkgConfig: /usr/local/bin/pkg-config (found version "0.29.2") 
-- Could NOT find Jemalloc (missing: JEMALLOC_LIBRARY JEMALLOC_INCLUDE_DIR) 
-- Found OpenCV: /usr/local (found version "3.4.3") found components:  core highgui imgproc imgcodecs 
--  OpenCV_LIBS=opencv_core;opencv_highgui;opencv_imgproc;opencv_imgcodecs
-- OpenCV 3.4.3 found (/usr/local/share/OpenCV)
CMake Error at /usr/local/Cellar/cmake/3.13.0/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES)
Call Stack (most recent call first):
  /usr/local/Cellar/cmake/3.13.0/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/Cellar/cmake/3.13.0/share/cmake/Modules/FindOpenMP.cmake:473 (find_package_handle_standard_args)
  CMakeLists.txt:408 (find_package)


-- Configuring incomplete, errors occurred!
See also "/Users/lnyuan/work/mxnet/build/CMakeFiles/CMakeOutput.log".
  File "./dev_menu.py", line 242, in <module>
    sys.exit(main())
  File "./dev_menu.py", line 238, in main
    globals()[args.command](args)
  File "./dev_menu.py", line 197, in use_menu_ui
    handle_commands(COMMANDS[command_list[choice]])
  File "./dev_menu.py", line 183, in handle_commands
    handle_commands(cmd)
  File "./dev_menu.py", line 187, in handle_commands
    cmds()
  File "./dev_menu.py", line 92, in __call__
    check_call(cmd_lst)
  File "/usr/local/Cellar/python/3.7.2_2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/subprocess.py", line 347, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['cmake', '-DUSE_CUDA=OFF', '-DUSE_OLDCMAKECUDA=OFF', '-DUSE_NCCL=OFF', '-DUSE_OPENCV=ON', '-DUSE_OPENMP=ON', '-DUSE_CUDNN=ON', '-DUSE_SSE=ON', '-DUSE_F16C=ON', '-DUSE_LAPACK=ON', '-DUSE_MKL_IF_AVAILABLE=ON', '-DUSE_MKLML_MKL=ON', '-DUSE_MKLDNN=ON', '-DUSE_OPERATOR_TUNING=ON', '-DUSE_GPERFTOOLS=ON', '-DUSE_JEMALLOC=ON', '-DUSE_PROFILER=ON', '-DUSE_DIST_KVSTORE=OFF', '-DUSE_PLUGINS_WARPCTC=OFF', '-DUSE_PLUGIN_CAFFE=OFF', '-DUSE_CPP_PACKAGE=OFF', '-DUSE_MXNET_LIB_NAMING=ON', '-DUSE_GPROF=OFF', '-DUSE_CXX14_IF_AVAILABLE=OFF', '-DUSE_VTUNE=OFF', '-DENABLE_CUDA_RTC=ON', '-DBUILD_CPP_EXAMPLES=ON', '-DINSTALL_EXAMPLES=OFF', '-DUSE_SIGNAL_HANDLER=ON', '-DUSE_TENSORRT=OFF', '-DUSE_ASAN=OFF', '-DENABLE_TESTCOVERAGE=OFF', '-DCMAKE_BUILD_TYPE=Debug', '-DCMAKE_CUDA_COMPILER_LAUNCHER=ccache', '-DCMAKE_C_COMPILER_LAUNCHER=ccache', '-DCMAKE_CXX_COMPILER_LAUNCHER=ccache', '-GNinja', '..']' returned non-zero exit status 1.
```

larroy:
@apeforest would you prefer something like ./dev_menu.py build?  You can always use CMake to build. this is just a convenience script on top which makes it easier.

apeforest:
IMHO, I prefer the plain `cmake` or `make` command for building the software with default flags set in a config file. Any Python utility script can be a nice thing for individual developer's preference but not used as a general instruction.

larroy:
@apeforest can you review again?

docs/install/osx_setup.md
```diff
@@ -100,14 +104,21 @@ After you have installed the dependencies, pull the MXNet source code from Git a
...
-    echo "USE_BLAS = openblas" >> ./config.mk
-    echo "ADD_CFLAGS += -I/usr/local/opt/openblas/include" >> ./config.mk
-    echo "ADD_LDFLAGS += -L/usr/local/opt/openblas/lib" >> ./config.mk
-    echo "ADD_LDFLAGS += -L/usr/local/lib/graphviz/" >> ./config.mk
-    make -j$(sysctl -n hw.ncpu)
+git clone --recursive https://github.com/apache/incubator-mxnet ~/mxnet
+cd ~/mxnet
+mkdir build && cd build
+cmake \
+    -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
```
apeforest: While `ccache` is really tool for speeding up the compilation time, I still feel we should just provide the minimum required steps for users to build mxnet. Can we suggest these performance tools in a separate doc/section?

docs/install/osx_setup.md
```diff
@@ -146,10 +157,11 @@ We have installed MXNet core library. Next, we will install MXNet interface pack
 To install the MXNet Python binding navigate to the root of the MXNet folder then run the following:
 
 ```bash
-$ cd python
-$ pip install -e .
+virtualenv -p`which python3` py3_venv
```
apeforest: Same for `virtualenv`. This has nothing to do with MXNet. We should let users to choose their own tool (such as conda) instead.

docs/install/osx_setup.md
```diff
@@ -146,10 +157,11 @@ We have installed MXNet core library. Next, we will install MXNet interface pack
...
+source py3_venv/bin/activate
+pip install -e python
 ```
-
+First we create a [virtualenv](https://docs.python-guide.org/dev/virtualenvs/#lower-level-virtualenv) to isolate this installation from our global environment.
```
apeforest: Same for virtualenv. This has nothing to do with MXNet. We should let users to choose their own tool (such as conda) instead.

docs/install/osx_setup.md
```diff
@@ -100,14 +104,21 @@ After you have installed the dependencies, pull the MXNet source code from Git a
...
-    echo "ADD_CFLAGS += -I/usr/local/opt/openblas/include" >> ./config.mk
-    echo "ADD_LDFLAGS += -L/usr/local/opt/openblas/lib" >> ./config.mk
-    echo "ADD_LDFLAGS += -L/usr/local/lib/graphviz/" >> ./config.mk
-    make -j$(sysctl -n hw.ncpu)
+git clone --recursive https://github.com/apache/incubator-mxnet ~/mxnet
+cd ~/mxnet
+mkdir build && cd build
+cmake \
+    -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+    -DCMAKE_C_COMPILER_LAUNCHER=ccache \
```
apeforest: Can we make this options default in the `cmake_options_osx.yml` so users can run the simplest `cmake` to build MXNet just like we are running `make`

docs/install/osx_setup.md
larroy: Im afraid that if we don't specify either one users will override the system global installation of mxnet, it's good hygiene to scope it to a virtualenv.

docs/install/osx_setup.md
larroy: cmake_options_osx.yml is used by dev_menu.py not by cmake.  I can remove ccache if it bothers you so much.

Roshrini:
@aaronmarkham Can you please help review this?

docs/install/osx_setup.md
```diff
@@ -100,14 +104,21 @@ After you have installed the dependencies, pull the MXNet source code from Git a
...
+    -DUSE_MKL_IF_AVAILABLE=O \
+    -DUSE_CPP_PACKAGE=ON \
```
apeforest: I think this flag is off by default in config.mk. Can we also disable it here?

docs/install/osx_setup.md
apeforest: I think it's a general practice for users to install any package not related to mxnet. But this is non blocking to me. @aaronmarkham please advise.

docs/install/osx_setup.md
apeforest: Got error `CMake Error: The source directory "/Users/lnyuan/work/mxnet/build" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.` following this instruction.

docs/install/osx_setup.md
```diff
@@ -100,14 +104,21 @@ After you have installed the dependencies, pull the MXNet source code from Git a
 The file called ```osx.mk``` has the configuration required for building MXNet on OS X. First copy ```make/osx.mk``` into ```config.mk```, which is used by the ```make``` command:
```
apeforest: This will be obsolete if we use cmake, right?

docs/install/osx_setup.md
aaronmarkham: It depends where it's going. When we have a detailed guide, it is fine to add recommended practices as new users may not know these details like a seasoned Python developer might.
When we have a very succinct section like the install page, I'd leave out these details.
So, I'd leave this in here since osx_setup.md is meant to be a detailed guide.

I do suggest another name though.  Like `mxnet_p36`, which is used as the names of the Conda environments on the DLAMI. Speaking of Conda, it would be good to mention this as an option too. It's ok to not document those steps as you're providing the virtualenv steps already. It's nice to let people know that mxnet will work fine in a Conda environment if that's what they want to do.

docs/install/osx_setup.md
```diff
@@ -77,6 +77,10 @@ Install the dependencies, required for MXNet, with the following commands:
 
 ```bash
 	brew update
+    brew install python3
```
aaronmarkham: Ah the implications of this. Let's do it, but I think we need to socialize how many tutorials are going to break and how many other things need to change.

docs/install/osx_setup.md
aaronmarkham: I'm all for speed. I think we should include it. 
Unless there's a lot of complexity added... does the user have to setup ccache or is it a seamless installation during the dependencies setup?

docs/install/osx_setup.md
larroy: I haven't encountered any problems. Are the tutorials in CI? I think compat issues between py2 / py3 are very easy to diagnose by users on their own.

docs/install/osx_setup.md
larroy: renamed to mxnet_py3

anirudhacharya:
@apeforest can you please take a look at this PR again.

vandanavk:
adding @marcoabreu @perdasilva for review too

docs/install/osx_setup.md
apeforest: I also prefer just use python in the instructon. Also, if user use virtualenv, this will not work.

docs/install/osx_setup.md
apeforest: If that's the case, we should make this default rather than letting users to specify. I have found this instruction too complicated to use for me. Copy and paste multiple lines does not work well.

docs/install/osx_setup.md
```diff
@@ -89,25 +92,31 @@ Install the dependencies, required for MXNet, with the following commands:
...
-    echo "ADD_LDFLAGS += -L/usr/local/opt/openblas/lib" >> ./config.mk
-    echo "ADD_LDFLAGS += -L/usr/local/lib/graphviz/" >> ./config.mk
-    make -j$(sysctl -n hw.ncpu)
+git clone --recursive https://github.com/apache/incubator-mxnet ~/mxnet
+cd ~/mxnet && pushd .
+mkdir build && cd build
+cmake \
+    -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+    -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+    -DUSE_MKL_IF_AVAILABLE=OFF \
```
apeforest: Can we make this single line. Comparing to the Makefile instruction, this is too complicated.

karan6181:
@larroy Could you please address the review comments? Thanks!

abhinavs95:
@larroy Could you have a look at the latest review comments? Thanks

docs/install/osx_setup.md
larroy: the lines are escaped, so it can be copied and pasted no problem.

larroy: Screen Shot 2019-06-05 at 3 47 53 PM

larroy:
Any more changes requested?

vandanavk:
@larroy could you rebase this PR?

@apeforest @aaronmarkham is this PR good to go?

larroy:
@vandanavk rebased

aaronmarkham:
I'm hesitant to push the python3 thing through. It'll start a wave of issues. If python3 is on someone's plate and they'll support the blast radius from this, then go for it. Otherwise, let's keep the instructions working with python2 as that seems to be the current default for most things in this project.

larroy:
@aaronmarkham what do you suggest? to change python3 commands to just python ?  I thought there was a discussion about dropping python2 soon but I wasn't able to find it now.

larroy:
please suggest path forward: 

Shall python3 mention be removed?
Shall we close this PR as it doesn't advance?
@apeforest @aaronmarkham ?

aaronmarkham:
@larroy The last discussion I think I saw on Python 2 deprecation mentioned dropping it by Jan 1. I don't remember seeing a formal vote though... either way, that's no clear indication that someone can lend a hand if this causes people issues.
So I suggest two options for moving forward.
1) Change it to Python 2.
2) Have both Python 2 and Python 3 instructions (more awesome route)

karan6181:
@larroy Could you please address @aaronmarkham comments? thanks!

larroy:
@aaronmarkham as per dev@ discussion, Python 2 will be deprecated soon. I would recommend only referring to Python3 in documentations and examples. This would be an incremental effort. Referring to Python2 in docs now is counterproductive.

aaronmarkham:
If we're doing Python 3 then let's do it all the way. I'm ok with this change. @apeforest I like to be specific, so using `pip3` is better than being vague and having people accidentally use Python 2 or pip when Python 2 is still default in so many installations.

piyushghai:
@apeforest Is this PR good to go now ?

cmake/cmake_options_osx.yml
```diff
@@ -0,0 +1,53 @@
...
+BUILD_CPP_EXAMPLES: "ON" # Build cpp examples
+INSTALL_EXAMPLES: "OFF" # Install the example source files.
+USE_SIGNAL_HANDLER: "ON" # Print stack traces on segfaults.
+USE_TENSORRT: "OFF" # Enable infeference optimization with TensorRT.
+USE_ASAN: "OFF" # Enable Clang/GCC ASAN sanitizers.
+ENABLE_TESTCOVERAGE: "OFF" # Enable compilation with test coverage metric output
+CMAKE_BUILD_TYPE: "Debug"
+CMAKE_CUDA_COMPILER_LAUNCHER: "ccache"
+CMAKE_C_COMPILER_LAUNCHER: "ccache"
+CMAKE_CXX_COMPILER_LAUNCHER: "ccache"
```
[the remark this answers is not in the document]
larroy: sure

docs/install/osx_setup.md
```diff
@@ -91,23 +95,29 @@ Install the dependencies, required for MXNet, with the following commands:
 	# For visualization of network graphs
 	pip install graphviz==0.8.4
```
apeforest: If we use pip3, we should change to pip3 all over the doc. Also, as I said this will break for people who uses virtualenv, as I always use pip in my virtualenv. It's up to @aaronmarkham 's call.

docs/install/osx_setup.md
apeforest: Should this be python instead of python3? Also it seems the community has not reached an agreement on retiring python2 sypport yet.

docs/install/osx_setup.md
marcoabreu: These should be bare minimum steps. Requiring ccache is an advanced thing and I'd prefer to not have it in the basic instructions.

docs/install/osx_setup.md
```diff
@@ -145,12 +155,16 @@ We have installed MXNet core library. Next, we will install MXNet interface pack
 To install the MXNet Python binding navigate to the root of the MXNet folder then run the following:
 
 ```bash
-$ cd python
-$ pip install -e .
+virtualenv -p`which python3` mxnet_py3
```
marcoabreu: While I generally agree that virtualenv is a good practice, it adds another dependency.

These guides are for the very basic user to get something running. I think this PR overcomplicates a lot of things. Adding the cmake compilation? Fine! Everything else? Please leave it as it is.

cmake/cmake_options_osx.yml
```diff
@@ -0,0 +1,53 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
apeforest: Instead of having this file, can we add a CMakeList_osx.txt instead? I think this adding another layer on top of CMakeList.txt is unnecessary.

cmake/cmake_options.yml
```diff
@@ -46,7 +46,8 @@ USE_SIGNAL_HANDLER: "ON" # Print stack traces on segfaults.
 USE_TENSORRT: "OFF" # Enable infeference optimization with TensorRT.
 USE_ASAN: "OFF" # Enable Clang/GCC ASAN sanitizers.
 ENABLE_TESTCOVERAGE: "OFF" # Enable compilation with test coverage metric output
-CMAKE_BUILD_TYPE: "Debug"
+USE_INT64_TENSOR_SIZE: "OFF" # Use int64_t to represent the total number of elements in a tensor
+CMAKE_BUILD_TYPE: "Debug" # Debug | Release | RelWithDebInfo | MinSizeRel
```
apeforest: The danger of this default option is that users are not aware they are building a Debug instead of Release.

docs/install/osx_setup.md
```diff
@@ -97,17 +101,23 @@ Install the dependencies, required for MXNet, with the following commands:
...
+    -DUSE_MKL_IF_AVAILABLE=OFF \
+    -DUSE_MKLDNN=ON \
+    -DUSE_CUDA=OFF \
+    -DUSE_OPENMP=OFF \
+    -DUSE_OPENCV=ON \
+    -DUSE_SIGNAL_HANDLER=ON \
+    -DCMAKE_BUILD_TYPE=Debug \
+    -GNinja ..
+ninja
+popd
```
apeforest: Why need `pushd .` and `popd`?

docs/install/osx_setup.md
```diff
@@ -97,17 +101,23 @@ Install the dependencies, required for MXNet, with the following commands:
...
-    echo "ADD_LDFLAGS += -L/usr/local/lib/graphviz/" >> ./config.mk
-    make -j$(sysctl -n hw.ncpu)
+git clone --recursive https://github.com/apache/incubator-mxnet ~/mxnet
+cd ~/mxnet && pushd .
+mkdir -p build && cd build
+cmake \
+    -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+    -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+    -DUSE_MKL_IF_AVAILABLE=OFF \
+    -DUSE_MKLDNN=ON \
```
apeforest: I feel all these flags are unnecessary. We already have a CMakeList.txt and isn't that the place for user to switch on/off these flags?

docs/install/osx_setup.md
marcoabreu: @pengzhao-intel are these the optimal compile flags? Especially with regard to omp and mkl?

Also, why are we so verbose here? Are our defaults not good enough?

docs/install/osx_setup.md
marcoabreu: Usually users shouldn't be touching cmakelists.txt

docs/install/osx_setup.md
```diff
@@ -97,17 +101,23 @@ Install the dependencies, required for MXNet, with the following commands:
...
+    -DUSE_CUDA=OFF \
+    -DUSE_OPENMP=OFF \
+    -DUSE_OPENCV=ON \
+    -DUSE_SIGNAL_HANDLER=ON \
+    -DCMAKE_BUILD_TYPE=Debug \
```
marcoabreu: Why are we making a debug build? Wouldn't a release build be a bit faster and improve the first impression? Or maybe release with symbols.

docs/install/osx_setup.md
```diff
@@ -145,12 +155,16 @@ We have installed MXNet core library. Next, we will install MXNet interface pack
...
 
 ```bash
-$ cd python
-$ pip install -e .
+virtualenv -p`which python3` mxnet_py3
+source mxnet_py3/bin/activate
+pip install -e python
 ```
-
+First we create a [virtualenv](https://docs.python-guide.org/dev/virtualenvs/#lower-level-virtualenv) to isolate this installation from our global environment.
```
apeforest: This should not be required for installing mxnet.

docs/install/osx_setup.md
larroy: @apeforest that's the way CMake works. I'm not sure what exactly are you suggesting. Feedback should be concrete, timely and actionable to be useful.

docs/install/osx_setup.md
larroy: This is a good point.

docs/install/osx_setup.md
larroy: This is a good practice not to polute global environments.

cmake/cmake_options.yml
larroy: This file is only used for the dev_menu.py script build. If you call cmake explicitly it doesn't affect the cmake command.

larroy: I don't care about Debug or Release, I assume that if someone is building they want to develop or diagnose a problem. For developers I think most of the time Debug is the build you want.

cmake/cmake_options_osx.yml
larroy: That's not how CMake works unfortunately.

docs/install/osx_setup.md
larroy: Removed

larroy:
This PR was opened 5 months ago, adding comments ad infinitum after a few rounds of CR and existing approvals is not a good use of time for volunteers trying to contribute to this project.

docs/install/osx_setup.md
larroy: Isn't it obvious? to return back to the root directory. Common practice in shell scripts.

larroy:
@apeforest you already approved this PR before. Why are you requesting additional changes again after some time? Is this to frustrate contributions? I find this way of reviewing code inconsistent.

apeforest:
My earlier approval was based on your assumption that python2 is no longer supported by mxnet. However, reading the latest discussion on devlist made me realize it is not the case. 

If so, such change in your PR:
`brew install python3` would seem not in sync with mxnet community decision.

Also, this man page is crucial to any new and existing mxnet users, it's never too much to be cautious. I have also found the dev_menu.py usage very confusing. What's the benefit of adding another layer of yaml file on top of exiting CMakeList.txt? It has recently caused some issue for a new developer who build mxnet with this script without realizing the build type is set to Debug by default.

In my opinion, we should teach users to change CMakeList.txt and build mxnet for different platforms.

docs/install/osx_setup.md
apeforest: Agree. ccache is not a must and it's an advanced compiler tool that should be up to users to choose.
To @aaronmarkham 's question, yes it does require user to set up `ccache` and it's not seamless installation.

docs/install/osx_setup.md
```diff
@@ -77,6 +77,10 @@ Install the dependencies, required for MXNet, with the following commands:
 
 ```bash
 	brew update
+    brew install python3
+	brew install cmake
+	brew install ninja
```
apeforest: Same thing for ninja. We can just add a paragraph referring users to ninja and ccache without selling these to them in a combo package with mxnet.

cmake/cmake_options_osx.yml
apeforest: That's interesting: https://preshing.com/20170511/how-to-build-a-cmake-based-project/

I have been successfully configure CMakeList.txt in this project to build without relying on your script. I have found it much simpler. To build mxnet, I only need the following three lines:

```
mkdir -p build
cd build
cmake -DCMAKE_BUILD_TYPE=Release ..
make
```

larroy:
https://www.python.org/doc/sunset-python-2/

larroy:
I changed the build to release. Is there any other concrete changes you are requesting so we can get done with this? This has been open for 6 months.

cmake/cmake_options_osx.yml
larroy: Please propose or suggest concrete changes. I don't think changing cmakelist is a good idea. Haven't seen that in other projects, if you know better please propose something.

docs/install/osx_setup.md
larroy: I think this is the fastest and most reliable way to build. We have discussed this before.

cmake/cmake_options_osx.yml
apeforest: We should probably follow what pytorch builds: https://github.com/pytorch/pytorch with a proper setup.py

The current way of dev_menu.py seems like a hacky solution.

docs/install/osx_setup.md
apeforest: A build process should be as simple as possible. Ideally, after a few times a user can build it without having to find this page and do copy-paste. I've found it difficult to remember all these flags and it's really frustrating to developers to open this page every time when they want to build mxnet from source.

apeforest:
1) If this PR goes in, users will have no information of building mxnet with make on MacOS. So this is a very critical change.

2) The current way of building cmake is hacky and not well designed. Most other cmake projects either use native `cmake` command to build with very few required user flags or use a setup.py like pytorch

3) We should design a proper way of cmake build script before we update the man page. These two changes should happen in two seperate PRs rather than mixed together.

docs/install/osx_setup.md
apeforest: The best install instruction is for users to build mxnet out of box.

larroy:
Wasn't hacky a very bad slang according to the politburo?  Maybe you can show how to make it better. For me it works for building quickly and reproducing test results.

larroy:
Anyway seems that this PR has some fundamental flaws that are far beyond my intellect to comprehend, so I will go ahead and close it. Maybe @apeforest can propose something better. Thank you for your time and for the reviews it has been quite a learning experience.

larroy:
Anyway seems that this PR has some fundamental flaws that are far beyond my intellect to comprehend, so I will go ahead and close it. Maybe @apeforest can propose something better. Thank you for your time and for the reviews it has been quite a learning experience.

PR #14536, opened 2019-03-26
[numpy] Fix R package errors after using numpy-compatible shape definition

hetong007:
## Description ##
Fix the R-package for numpy-compatible shape definition.

@szha @reminisce


abhinavs95:
@mxnet-label-bot add [Numpy, R, pr-awaiting-review]

Issue #14454, opened 2019-03-18
[numpy] Fix test failures after using new shape semantics for zero-dim/size tensors


reminisce:
**Branch:** https://github.com/apache/incubator-mxnet/tree/numpy

## CPU withou MKLDNN

**Folder:** [tests/python/unittest](https://github.com/apache/incubator-mxnet/tree/numpy/tests/python/unittest)

| Errors/Failures (CPU) |   Assignee |  Status |
|-------------------------|------------------------------|--------|
|  test_contrib_control_flow.test_contrib_rnn  |     @zheng-da       | Fixed by  https://github.com/apache/incubator-mxnet/pull/14520       |
| test_dynamic_shape.test_dynamic_shape  |    @junrushao1994        |    Fixed by https://github.com/apache/incubator-mxnet/pull/14538     |
| test_gluon_contrib.test_convrnn  |           |     Resolved    |
| test_gluon_contrib.test_convlstm  |           |     Resolved    |
| test_gluon_contrib.test_convgru  |           |    Resolved     |
| test_gluon_contrib.test_conv_fill_shape  |           |  Resolved       |
| test_gluon_contrib.test_lstmp  |           |   Resolved      |
| test_gluon_contrib.test_vardrop  |           |    Resolved     |
| test_gluon_contrib.test_contrib_unroll  |   @junrushao1994          |    Resolved     |
| test_gluon_data.test_recordimage_dataset | @szha  | Resolved |
| test_gluon_model_zoo.test_models  |           |    Resolved     |
| test_gluon.test_deconv | @reminisce   | Resolved |
| test_gluon.test_symbol_block  |           |   Resolved      |
| test_gluon.test_deferred_init  |           |     Resolved    |
| test_gluon.test_hybrid_stale_cache  |           |   Resolved      |
| test_gluon.test_fill_shape_deferred  |           |    Resolved     |
| test_gluon.test_dtype  |           |    Resolved     |
| test_gluon.test_fill_shape_load  |           |    Resolved     |
| test_gluon.test_inline  |           |    Resolved     |
| test_gluon.test_req  |           |  Resolved       |
| test_gluon.test_save_load  |           |   Resolved      |
| test_gluon.test_symbol_block_save_load  |           |   Resolved      |
| test_gluon.test_hybrid_multi_context  |           |   Resolved      |
| test_gluon.test_hook  |           |  Resolved       |
| test_gluon.test_summary  |           |   Resolved      |
[...]

PR #14538, opened 2019-03-26
[numpy] Fix test_dynamic_shape.test_dynamic_shape

junrushao1994:
#14454

@reminisce hey could you take a look?

### DIFF
--- a/src/imperative/imperative_utils.cc
+++ b/src/imperative/imperative_utils.cc
@@ -19,6 +19,7 @@
 
 #include "./imperative_utils.h"
 #include "./cached_op.h"
+#include "../operator/operator_common.h"
 
 namespace mxnet {
 namespace imperative {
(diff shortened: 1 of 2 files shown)


src/imperative/imperative_utils.cc
```diff
@@ -186,7 +186,7 @@ void NaiveRunGraph(
       Imperative::Get()->InvokeOp(ctx, node.source->attrs, ndinputs, ndoutputs,
                                   req, dispatch_mode, state);
       for (size_t j = 0; j < ndoutputs.size(); ++j) {
-        if (ndoutputs[j]->shape().ndim() == 0) {
+        if (ndoutputs[j]->shape().ndim() == -1) {
```
reminisce: nit: Better to use `shape_is_known` for readability. More precisely, it should be `shape_is_known` instead of just `ndim_is_known`, right?

junrushao1994: Good idea!

PR #14550, opened 2019-03-28
Do not touch GPU 0 during ReleaseAll

ptrendx:
## Description ##
Currently, during call to ReleaseAll (either in pooled allocator destructor or when the memory is full and the allocator needs to release cached allocations) the memory handle is not fully populated (only ptr and size are populated, not context information), which makes the DirectFreeNoLock method to switch to GPU 0 (since 0 is the default constructed Context's dev_id). In the case of multi process training (e.g. with Horovod) this could produce Out of Memory errors due to context creation on GPU 0, or it could outright fail, crashing the application, if the exclusive mode is set on GPU 0.

This PR fixes that by adding `initial_context_` parameter to pooled storage managers and populating the handle with it in ReleaseAll method.

@apeforest @yuxihu FYI

### Essentials ###

### DIFF
--- a/src/storage/pooled_storage_manager.h
+++ b/src/storage/pooled_storage_manager.h
@@ -54,10 +54,11 @@ class GPUPooledStorageManager final : public StorageManager {
   /*!
    * \brief Default constructor.
    */
-  GPUPooledStorageManager() {
+  GPUPooledStorageManager(int dev_id) {
     reserve_ = dmlc::GetEnv("MXNET_GPU_MEM_POOL_RESERVE", 5);
     page_size_ = dmlc::GetEnv("MXNET_GPU_MEM_POOL_PAGE_SIZE", 4096);
     large_alloc_round_size_ = dmlc::GetEnv("MXNET_GPU_MEM_LARGE_ALLOC_ROUND_SIZE", 2 * 1024 * 1024);
+    dev_id_ = dev_id;
     if (large_alloc_round_size_ <= 0) {
       LOG(FATAL) << "MXNET_GPU_MEM_LARGE_ALLOC_ROUND_SIZE cannot be set to a value <= 0, found: "
                  << large_alloc_round_size_;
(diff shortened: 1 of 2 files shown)


src/storage/pooled_storage_manager.h
```diff
@@ -54,10 +54,11 @@ class GPUPooledStorageManager final : public StorageManager {
   /*!
    * \brief Default constructor.
    */
-  GPUPooledStorageManager() {
+  GPUPooledStorageManager(int dev_id) {
```
lebeg: Maybe pass the context instead? And call it creation_ or initial_context? And maybe add a short explanation to the parameters documentation?

ptrendx: Done.

abhinavs95:
@mxnet-label-bot add [Backend, pr-awaiting-review]

src/storage/pooled_storage_manager.h
```diff
@@ -123,6 +126,8 @@ class GPUPooledStorageManager final : public StorageManager {
   int reserve_;
   // number of devices
   const size_t NDEV = 32;
+  // context used by this Storage Manager
+  Context initial_context_;
```
yuxihu: One more comment: make it const?

src/storage/pooled_storage_manager.h
```diff
@@ -290,6 +299,8 @@ class GPUPooledRoundedStorageManager final : public StorageManager {
   size_t cut_off_;
   // percentage of reserved memory
   int reserve_;
+  // context used by this Storage Manager
+  Context initial_context_;
```
yuxihu: Same as above

src/storage/pooled_storage_manager.h
ptrendx: Done.

src/storage/pooled_storage_manager.h
ptrendx: Done.

ptrendx:
The CI failure seems completely unrelated.

yuxihu:
@mxnet-label-bot update [Backend, pr-awaiting-merge]

wkcn:
The PR has been merged. Thank you for the fix!

PR #14554, opened 2019-03-28
[numpy] Fix cpp-package build after using new shape definition

reminisce:
Fixed problem of building cpp-package after using the numpy compatible shape definition.
@junrushao1994 @zheng-da


junrushao1994:
Just to confirm, this issue is to caused by `mx_uint`, which causes comparison between unsigned and signed, right?

reminisce:
@junrushao1994 The issue is caused by `mx_uint`, but the comparison between unsigned and signed is not the main problem this PR intends to address. The PR fixes the interface compatibility at compile time as well as well auto-generation of operator interfaces in cpp-package after using the new shape/tuple data type.

PR #14557, opened 2019-03-29
[Numpy] fix test_operator_gpu.test_upsampling_bilinear_with_type

junrushao1994:
Hey @reminisce yet another out-of-bound error

### DIFF
--- a/src/operator/nn/deconvolution-inl.h
+++ b/src/operator/nn/deconvolution-inl.h
@@ -122,7 +122,7 @@ struct DeconvolutionParam : public dmlc::Parameter<DeconvolutionParam> {
     // Modified by Li.bs
     // Use tag to control the calculation of pad
     bool bCal = false;
-    if (!mxnet::op::shape_is_none(target_shape)) {
+    if (target_shape.ndim() != 0) {
       for (index_t i = 0; i < target_shape.ndim(); i++) {
         if (target_shape[i] != 0) bCal = true;
       }
--- a/src/operator/nn/deconvolution-inl.h
+++ b/src/operator/nn/deconvolution-inl.h
@@ -122,7 +122,7 @@ struct DeconvolutionParam : public dmlc::Parameter<DeconvolutionParam> {
     // Modified by Li.bs
     // Use tag to control the calculation of pad
     bool bCal = false;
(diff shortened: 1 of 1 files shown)


src/operator/nn/deconvolution-inl.h
```diff
@@ -122,7 +122,7 @@ struct DeconvolutionParam : public dmlc::Parameter<DeconvolutionParam> {
     // Modified by Li.bs
     // Use tag to control the calculation of pad
     bool bCal = false;
-    if (target_shape.ndim() != 0) {
```
reminisce: `target_shape` is default to `mxnet::TShape(0)`, so the original code is correct.

src/operator/nn/deconvolution-inl.h
```diff
@@ -143,9 +143,9 @@ struct DeconvolutionParam : public dmlc::Parameter<DeconvolutionParam> {
         }
       }
     } else {
-      for (size_t i = 0; i < ndim; i++) {
-        o_pad[i] = pad[i];
-        o_adj[i] = adj[i];
+      for (int i = 0; i < (int) ndim; i++) {
+        o_pad[i] = i < pad.ndim() ? pad[i] : 0;
```
reminisce: According the [initialization function](https://github.com/apache/incubator-mxnet/blob/master/src/operator/nn/deconvolution.cc#L335), `pad.ndim()` should be equal to `ndim` here. Do you understand why `i` can be bigger than `pad.ndim()`?

abhinavs95:
@mxnet-label-bot add [numpy]

src/operator/nn/deconvolution-inl.h
junrushao1994: Correct. Revert this accordingly

src/operator/nn/deconvolution-inl.h
junrushao1994: The upsampling operator customized DeconvolutionParam [here](https://github.com/apache/incubator-mxnet/blob/master/src/operator/nn/upsampling-inl.h#L170-L188). It is kinda weird but I believe adj is not used in this case.

junrushao1994:
@reminisce Hey could you review again? Thanks!

Issue #14569, opened 2019-03-29
performance degradation in model inference from 1.3.1 to 1.4.0


apeforest:
There seems to be a regression on resnet-18 model inference time (When running on GPU) post this PR, this was caught in MMS. nightly runs, the changes in this PR seem to be causing this issue.

**Setup**

We use MMS docker images to run load tests, we can start a local container using the following command.

```bash
nvidia-docker run --name mms_benchmark_gpu -p 8080:8080 -p 8081:8081  -itd awsdeeplearningteam/mxnet-model-server:nightly-mxnet-gpu
```

for building MXNet opencv 3.2 and CUDA 9.2 were used. 

Load testing was done using locust, to install locust
```bash
pip install locust
```

# Download Test image
```bash
curl -O https://s3.amazonaws.com/model-server/inputs/kitten.jpg
```
The locust script for load testing
```python
# test_resnet_!8.py
from locust import HttpLocust, TaskSet, task, TaskSequence, seq_task
import urllib
import os
data = None
with open(os.path.join(os.getcwd(),'kitten.jpg'), 'rb') as data:
    data = data.read()

class PredictionTasks(TaskSet):
    @task
    def inference(self):
        self.client.post("/predictions/resnet-18", data=data,headers={'Content-Type': 'image/jpeg'})

class Prediction(HttpLocust):
    task_set = PredictionTasks
    min_wait = 100
    max_wait = 100
```

**Running Load test**

Registering and loading model

```bash
# Register and load resnet-18 model archive
 curl -X POST 127.0.0.1:8081/models?url=https://s3.amazonaws.com/model-server/model_archive_1.0/resnet-18.mar
```

Start a single worker and run latency test

```bash
Start worker and latency test
$ curl -X PUT 'http://127.0.0.1:8081/models/resnet-18?min_worker=1&synchronous=true'
$ locust -f test_resnet_18.py Prediction --host=http://127.0.0.1:8080 --no-web  -c 1 -r 1 -t 20s --only-summary
```

**To change mxnet version/build in docker image,** 

**NOTE** By default recent pip version is pulled.

```bash
# Go into docker image
nvidia-docker exec -u root -it mms_benchmark_gpu bash
$  pip uninstall mxnet-cu92mkl
$ pip install <new-build>.whl
ctrl + p + q to quit docker image

# Destroy existing worker, and create new worker, this loads in newly installed mxnet
[...]

Issue #14563, opened 2019-03-29
Performance decrease with pypi's mxnet==1.4.0 (Mac)


fhieber:
I am experiencing significant slowdowns with the latest pypi release of `mxnet==1.4.0` on MacOs when running transformer training with Sockeye:

```
mxnet==1.3.1
[INFO:sockeye.training] Checkpoint [1]	Updates=1000 Epoch=1 Samples=16000 Time-cost=13.735 Updates/sec=72.805

Mxnet==1.4.0
[INFO:sockeye.training] Checkpoint [1]	Updates=1000 Epoch=1 Samples=16000 Time-cost=20.084 Updates/sec=49.791

mxnet-mkl==1.3.1
[INFO:sockeye.training] Checkpoint [1]	Updates=1000 Epoch=1 Samples=16000 Time-cost=15.200 Updates/sec=65.791

mxnet-mkl==1.4.0
[INFO:sockeye.training] Checkpoint [1]	Updates=1000 Epoch=1 Samples=16000 Time-cost=23.078 Updates/sec=43.331
```

Interestingly, this difference in performance does **not** exist on a Linux machine (with or without GPU). That is, I don't see any difference in speed between Linux' `mxnet-cu92==1.3.1` and `mxnet-cu92==1.4.0.post0`, or `mxnet==1.3.1` and `mxnet==1.4.0.post0`:

Note that the Linux packages all install a `1.4.0.post0` version that doesnt' seem to exist for MacOs. Is there a reason for that?

apeforest:
@fhieber Can you post a script to reproduce this? I am investigating one of the performance issue https://github.com/apache/incubator-mxnet/issues/14496 introduced in 1.4.0. They could be due to the same rootcause.  Thanks!

vdantu:
@mxnet-label-bot add [performance, question, pending requester info]

fhieber:
@apeforest thanks for investigating! I cannot provide a small script at the moment that reproduces the issue, I just observed the speed difference running various system tests with Sockeye. If you want to run them for yourself you can do the following:

```bash
git clone https://github.com/awslabs/sockeye.git
cd sockeye
pip install -r requirements/requirements.dev.txt
pip install -r requirements/requirements.txt
pytest test/system -k "Sort:transformer:transformer"
```
The speed difference will be visible directly from the first log messages of the form `Speed: 1004.51 samples/sec 10366.54 tokens/sec 62.78 updates/sec`.

[...]

Issue #14496, opened 2019-03-21
performance degradation from 1.3.1 to 1.4.0


samskalicky:
There appears to be some performance degradation between the 1.3.1 and 1.4.0 releases. So far we know of imdecode and tranpose operators having reduced performance.

We've tracked the responsible PRs for these operators to:
Transpose:
https://github.com/dmlc/mshadow/pull/359/files
https://github.com/apache/incubator-mxnet/pull/11742

Imdecode:
https://github.com/apache/incubator-mxnet/pull/8757

Im using the cat.jpg from here as input data:
https://github.com/dmlc/web-data/blob/master/mxnet/doc/tutorials/python/predict_image/cat.jpg

Heres the current performance benchmarking script for imdecode:

```
import time
import mxnet as mx
from mxnet import image as img
import numpy as np

flag=1
to_rgb=True
out=None

dtimes = []
for i in range(1000):
    buf = open("cat.jpg", 'rb').read()

    start = time.time()
    img.imdecode(buf, flag, to_rgb, out)
    decode = (time.time() - start)*1000
    dtimes.append(decode)

print('mxnet version: %s' % mx.__version__)
print('decode:---------')
print('p50: %4.2f ms' % np.percentile(dtimes,50))
print('p90: %4.2f ms' % np.percentile(dtimes,90))
print('p99: %4.2f ms' % np.percentile(dtimes,99))
```

And here are the performance results using mxnet-cu90mkl for 1.4.0 and 1.3.1:

```
mxnet version: 1.4.0
decode:--------- 
p50: 13.75 ms 
p90: 15.74 ms 
p99: 19.89 ms 

mxnet version: 1.3.1 
decode:--------- 
p50: 12.82 ms 
p90: 13.59 ms 
p99: 13.88 ms 
```

Setting the flag = 1 + 128 (instead of just 1) as an argument to imdecode results in the following results: 

```
mxnet version: 1.4.0 
load:--------- 
p50: 0.10 ms 
p90: 0.16 ms 
p99: 0.18 ms 
decode:--------- 
p50: 13.27 ms 
p90: 13.83 ms 
p99: 15.63 ms 
```

So there is some additional work thats going on that makes the imdecode take longer. This is a "feature" of the v1.4 release and changed the defaults which is where some performance degradation is happening at least in the imdecode function. 

Heres the current performance benchmarking script for transpose:

```
import mxnet as mx
import time
import numpy as np

sizes = [10, 50, 100,200,500]
[...]

Issue #13928, opened 2019-01-18
MXNet 1.5.0 is slower than 1.3.0 when intputs are variant


wkcn:
## Description
Hi! I have an experiment about Object Counting, which needs variant inputs.
I write the code with Gluon, and hybridize the model with `static_alloc=True`
I found there is obvious difference between MXNet 1.5.0 and MXNet 1.3.0, and I checked it on two servers.

~~I think the method of memory allocation for Gluon may be changed after MXNet 1.3.0.~~

Thanks!

Update:
When there are dilated Convolutional layers in the model, and the input size is variational, the performance will drop.
I think it may be related to one of the two PRs: #11742 #12722

## Environment info (Required)
OS: ubuntu 14.04
GPU: Tesla M40 x 4

## Minimum reproducible example
I write a minimum reproducible example without dataset.
[Code](https://gist.githubusercontent.com/wkcn/69f0f6d2ca467816dc481a00c225104f/raw/2899896f42a920ff0fde5ff93b9a16d16aec507f/test_fcn_for_mxnet.py)

- Performance for test code [a fully convolutional model(vgg16 without FC layer), variant inputs]:
MXNet 1.5.0: 10 images / sec
MXNet 1.3.0: 40+ images / sec

The performances are the same when input shape is fixed.

Input shape: (9, 3, 300\~512, 300\~512) in NCHW order

Package used (Python/R/Scala/Julia):
Python 2.7.12, 3.7.1

MXNet is installed by pip:
```
# MXNet 1.5.0
pip install mxnet-cu80 --pre
# MXNet 1.3.0
pip install mxnet-cu80==1.3.0
```

## Steps to reproduce
Download [the test code](https://gist.githubusercontent.com/wkcn/69f0f6d2ca467816dc481a00c225104f/raw/2899896f42a920ff0fde5ff93b9a16d16aec507f/test_fcn_for_mxnet.py).
Run the test code in different version (1.3.0 and 1.5.0) of MXNet.

## Performance

I test several versions of MXNet.

version|performance
----------|----------------
1.4.0b20181207|slow
1.3.1b20181101|slow
1.3.1b20181010|slow
1.3.1b20181004|fast
1.3.1b20181001|fast

Some pre-build versions don't  support CUDA9.0, so I cound't test it.
The performance drops during 20181004 to 20181010.

If changing the dilation of dilated conv to 1, the performance will be normal.
It seems the problem occurs in dilated conv.

piyushghai:
[...]

PR #14570, opened 2019-03-29
add a compiler flag to use int64 as tensor size

apeforest:
## Description ##
This PR is to fix the performance degradation reported in https://github.com/apache/incubator-mxnet/issues/14496.

The performance degradation is introduced by PR https://github.com/apache/incubator-mxnet/pull/11742. I have verified the performance degradation in one of the operators transpose using a script below:

```
import mxnet as mx
import time
import numpy as np

sizes = [10, 50, 100,200,500]
iters = [10000,1000,500,200,20]
times = []
for size in range(len(sizes)):
    data = []
    s = sizes[size]
    print(s)
    for i in range(iters[size]):
        x = mx.nd.ones((s,s,s))
        mx.nd.waitall()
        start = time.time()
        y = mx.nd.transpose(x,(2,0,1))
        mx.nd.waitall()
        data.append((time.time() - start)*1000)
        #print(data[-1])                                                                                                                                                                            
    times.append(data)

print('mxnet version: %s' % mx.__version__)
for s in range(len(sizes)):
    print('--------------------')
    print('size: %s' % str(sizes[s]))
    print('p50: %4.2f ms' % np.percentile(times[s],50))
    print('p90: %4.2f ms' % np.percentile(times[s],90))
    print('p99: %4.2f ms' % np.percentile(times[s],99))
```

By changing the `index_t` type from `int64_t` to `int32_t` can consistently change the 50-percentile runtime of tranpose of size 500 from 5000ms to 2400ms. 

By changing the data type in the operator (https://github.com/dmlc/mshadow/blob/master/mshadow/extension/transpose.h#L70) alone, can also reduce the 50-percentile runtime of size 500 to 2600ms.

I thereforce come to the conclusion that the performance degradation is caused by the runtime of  integer arithmetic operations between 32-bit integers and 64-bit integers.

To further experiment with the arithmetic operations alone, I tested using a small program [here](https://github.com/apeforest/doraemon/blob/master/perf32vs64.cc). The runtime results are shown below:

```
result = 49995000
Add 32-bit time in ms 1359
result = 49995000
Add 64-bit time in ms 1971
result = 349965000
Add Mul 32-bit time in ms 1196
result = 349965000
Add Mul 64-bit time in ms 3477
result = 7137858
Add Div 32-bit time in ms 2878
result = 7137858
Add Div 64-bit time in ms 8499
```

I can think of three solutions to this problem:
(1) Add a compilation flag to choose data types for tensor size (This PR)
(2) Add an environment variable to choose data type for tensor size at runtime
(3) Choose data type for tensor size at runtime based on the size of the tensor

Given the expression template used in mshadow for operators, either (2) or (3) requries a significant change in the mshadow library. (1) can be used as a quick solution to fix performance degradation reported in several issues: https://github.com/apache/incubator-mxnet/issues/14496, https://github.com/apache/incubator-mxnet/issues/13928, https://github.com/apache/incubator-mxnet/issues/14563, https://github.com/apache/incubator-mxnet/issues/14569

Any other suggestion is appreciated.

This PR also depends on https://github.com/dmlc/mshadow/pull/371

### Essentials ###

### Changes ###

## Comments ##


apeforest:
@pengzhao-intel @TaoLv Please help to review. Thanks!

abhinavs95:
@mxnet-label-bot add [pr-work-in-progress, backend]

TaoLv:
Do we need to change CI to cover this new flag? Do we need to add it to the runtime feature set? @larroy

wkcn:
Thank you for the quick fix!

I think we can use the solution (3): 'Choose data type for tensor size at runtime based on the size of the tensor' in the performance bottleneck firstly.

pengzhao-intel:
Agree with adding the new flag to switch to int64 when the user explicitly know they want to use the large array, like DGL. For the normal case, int32 is enough w/o performance issue.

@TaoLv do we need to switch off MKLDNN when int64 is enabled before MKL-DNN 1.0?

@wkcn would you mind mentioning the advantage of 3)?

wkcn:
@pengzhao-intel Yes. For users, it is confusing that there will be two versions of mxnet: mxnet and mxnet-largetensor.  In other words, users need to rebuild the source, when they use large tensor. It is too complex.

```c++
```

apeforest:
@wkcn Thanks for the suggestion. This macro is good for certain simple operators but I found it difficult to apply to all operators given the different architecture of operator implementation. E.g. transpose operator is currently using the expression template pattern in mshadow. It seems a big surgery to apply this macro to operators in mshadow. Besides, int64_t is used as type of data members in the tensor datastructure. This approach will not directly change the internal methods in tensor or requires static cast. Any other thoughts on this?

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "3rdparty/mshadow"]
 	path = 3rdparty/mshadow
-	url = https://github.com/dmlc/mshadow.git
+	url = https://github.com/apeforest/mshadow.git
```
samskalicky: I see why you did this for testing, but do we want to check this is to master and depend on your fork?

make/config.mk
```diff
@@ -214,6 +214,10 @@ EXTRA_OPERATORS =
 # Create C++ interface package
 USE_CPP_PACKAGE = 0
 
+# Use int64_t type for tensor size.
+# This will cause performance degradation reported in issue #14496
```
samskalicky: Maybe add something to say when to use it too, like "Set to 1 for large tensor sizes, when tensors will have large shapes such as for graphs" or something...?

.gitmodules
apeforest: Yes, after mshadow PR is merged, I will update this here accordingly.

make/config.mk
apeforest: updated.

make/config.mk
```diff
@@ -216,6 +216,7 @@ USE_CPP_PACKAGE = 0
 
 # Use int64_t type for tensor size.
 # This will cause performance degradation reported in issue #14496
+# Set to 1 for large tensor size greater than 2 billion
```
samskalicky: not to be too picky, but what does this mean "tensor size greater than 2 billion"? Does this mean that if you have more than 2 billion elements in a single dimension you need this? or 2 billion elements in the whole tensor?

include/mxnet/tensor_blob.h
```diff
@@ -442,7 +442,7 @@ class FieldEntry<mxnet::TShape>
         throw dmlc::ParamError(os.str());
     }
     if (enforce_nonzero_) {
-      for (mxnet::index_t i = 0; i < v.ndim(); ++i) {
+      for (int i = 0; i < v.ndim(); ++i) {
```
larroy: why int here?

include/mxnet/tensor_blob.h
```diff
@@ -456,7 +456,7 @@ class FieldEntry<mxnet::TShape>
     this->enforce_nonzero_ = true;
     return this->self();
   }
-  inline FieldEntry<mxnet::TShape> &set_expect_ndim(mxnet::index_t ndim) {
+  inline FieldEntry<mxnet::TShape> &set_expect_ndim(int ndim) {
```
larroy: why int? I know dimensions can't be too high but wouldn't it make sense to have a typedef for it?

include/mxnet/tensor_blob.h
apeforest: dimensions will not be hight and `int` is used across many template for dim such as https://github.com/dmlc/mshadow/blob/95ebe0f109ae021d0d66e2a627ccfc55c3253b55/mshadow/tensor.h#L43

include/mxnet/tensor_blob.h
apeforest: Prefer `int` to defined type for better readability. `unsigned int` is discouraged to use as data type. See https://google.github.io/styleguide/cppguide.html#Integer_Types

Makefile
```diff
@@ -188,6 +188,11 @@ ifeq ($(USE_OPERATOR_TUNING), 1)
 	CFLAGS += -DMXNET_USE_OPERATOR_TUNING=1
 endif
 
+ifeq ($(USE_INT64_TENSOR_SIZE), 1)
+   CFLAGS += -DMSHADOW_INT64_TENSOR_SIZE=1
```
anirudh2290: Since USE_INT64_TENSOR_SIZE has been set by default to 0 this is a breaking change. Are we going to add an exception for semver for the next minor release ?

apeforest: This does not break API but will impact cases for large tensors. How should we handle this cases?

anirudh2290: Ya probably doesn't fall under semver. We should still let users like dgl know about the change and how they can build for large tensor support.

include/mxnet/tensor_blob.h
larroy: I get your point. I'm not convinced that the dimensions type should be signed, as it makes no sense. Agree that the int width should be enough. I personally would make it unsigned.

samskalicky:
@larroy the dimension can be -1 if its unknown, or if -1 refers to the last dimension. So we do have to use signed types.

include/mxnet/tensor_blob.h
```diff
@@ -223,11 +223,11 @@ class TBlob {
    * \param idx the dimension count from the highest dimensin
    * \return the size
    */
-  inline index_t size(index_t idx) const {
+  inline size_t size(index_t idx) const {
```
eric-haibin-lin: This will break shapes with negative numbers? @reminisce

reminisce: As another effort in parallel, we need to use `-1` to represent unknown dim size to support zero-size tensors. So the type has to be kept as a signed integer.

larroy:
@reminisce why aren't we using ndim=1 for scalars?

reminisce:
@larroy Tensors with `ndim=1` are not scalars, but 1D tensors, aka vectors.

make/config.mk
apeforest: updated

include/mxnet/tensor_blob.h
apeforest: updated

.gitmodules
eric-haibin-lin: https://github.com/dmlc/mshadow/pull/371 has been merged

apeforest:
Added build with this new compiler flag on to CI and added a nightly test for large tensor per @TaoLv suggestion. @larroy @marcoabreu Please help to review.

apeforest:
@reminisce replaced nnvm::Tuple with mxnet::Tuple as we discussed in dev list. Please help to review. Thanks!

larroy:
Thanks a lot for the detailed report and analysis @apeforest and @samskalicky . From the data you guys have provided I'm missing a disassembly dump of the small loop that is supposed to be slow. Could you guys provide this with "objdump -d" or similar? I find it surprising that the degradation is only due to data widening. I suspect the cause is more the memory access than the wider arithmetic operation itself. I think having the additional data point of the assembly would help reach a more solid conclusion.

larroy:
To be honest, while I understand the rationale of the solution proposed. I think it would be better if we could avoid having yet another build flavour if possible.  Could we for example, narrow the computation at runtime in case the number of dimensions doesn't overflow? Also see my previous comment above to see if we can work with memory access to make it almost as fast with the wider type in this particular computation. If I understood correctly the regression is only in transposing? the other operators are fine? We could try to find a solution only for this operator then instead of adding a new build flavour.

include/mxnet/tensor_blob.h
larroy: Thanks @reminisce  and @apeforest . Let's resolve this comment then.

samskalicky:
@larroy we dont have any data that guarantees only the transpose operator is affected. In reality it could be all mshadow implemented operators that are affected. We dont have performance tests for each operator that we could run and verify. There are efforts to look into putting together some operator performance tests, but i dont think we should wait on this fix for that. 

This PR is important to restore "balance to the force" until we can work on a better solution. Since new functionality for large tensors was already added and is being used for the Deep Graph Library effort we need to provide a way for them to use the large tensor feature. At the same time, we need to fix the performance degradation. So this is fix (adding the flag) the common middle ground between the two.

Makefile
yuxihu: Will the MXNet pip packages be built with large tensor support disabled after this PR is merged? If so, for users who want large tensor support, do they have to build MXNet from source?

larroy:
@samskalicky @apeforest makes sense. I just wanted to raise this point of view. If you guys think this is a tradeof we should do now at the cost of adding an additional build flavour and we revisit later once we have per operator performance metrics, I'm fine with that.

ci/docker/runtime_functions.sh
```diff
@@ -755,6 +755,35 @@ build_ubuntu_gpu_cmake() {
...
+        USE_CPP_PACKAGE=1             \
+        USE_INT64_TENSOR_SIZE=1       \
+        USE_SIGNAL_HANDLER=1          \
+        -j$(nproc)
+}
+
+build_ubuntu_gpu_large_tensor() {
+    set -ex
+    build_ccache_wrappers
+    make  \
```
larroy: why not cmake?

ci/docker/runtime_functions.sh
```diff
@@ -755,6 +755,35 @@ build_ubuntu_gpu_cmake() {
     ninja -v
 }
 
+build_ubuntu_cpu_large_tensor() {
+    set -ex
+    build_ccache_wrappers
+    export CC="ccache gcc"
```
larroy: build_ccache_wrappers should be enough. My PR also fixes that function: https://github.com/apache/incubator-mxnet/pull/14631

Would be good to merge then no need to export the CC and CXX variables.

CMakeLists.txt
```diff
@@ -295,6 +296,13 @@ else()
   add_definitions(-DMXNET_USE_NCCL=0)
 endif()
 
+if (USE_INT64_TENSOR_SIZE)
```
larroy: Can you also add this to the runtime feature detection code please?

https://mxnet.incubator.apache.org/api/python/libinfo/libinfo.html

ci/docker/runtime_functions.sh
apeforest: I saw both `make` and `cmake` used in runtime_functions.sh such as https://github.com/apache/incubator-mxnet/blob/master/ci/docker/runtime_functions.sh#L358

Sholud we only use `cmake` in the future?

include/mxnet/tensor_blob.h
larroy: Adding to this, the type is signed ok I understand that. I still think we should specify the width of the type, say int32_t

Makefile
apeforest: Yes, users who need large tensor support will have to build MXNet from source.

yuxihu: I think we need to document this. Maybe on the build from source page?

ci/docker/runtime_functions.sh
larroy: We have all agreed to use CMake, so we should not introduce new make builds unless there's a problem with CMake, this is my understanding.

CMakeLists.txt
apeforest: Added.

Makefile
apeforest: It's documented in the config.mk file: https://github.com/apache/incubator-mxnet/pull/14570/files#diff-ea86416d0f764a60d29337663f7a1767R220

ci/docker/runtime_functions.sh
apeforest: Sure, updated the process to cmake. I think we may need a seperate PR to update all runtime_functions to cmake otherwise it may cause confusion to future developers.

ci/docker/runtime_functions.sh
apeforest: No longer needed after I change to using cmake

apeforest:
@TaoLv Per @marcoabreu suggestion, I moved the tests from CI to nightly to ease the load of CI pipeline. Please let me know if this is okay with you. Thanks!

TaoLv:
@apeforest Thank you. That sounds reasonable. But I remember I couldn't find the nightly test report for large tensor before. Do you know where we can read it once this PR is merged?

apeforest:
> @apeforest Thank you. That sounds reasonable. But I remember I couldn't find the nightly test report for large tensor before. Do you know where we can read it once this PR is merged?

Tao, I think the nightly tests are here: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/NightlyTests/activity

@marcoabreu  Could you please confirm? Thanks!

apeforest:
@TaoLv Do you have any more concerns for this PR? Thanks!

apeforest:
> Are we checking for narrowing / overflow since we have this additional wider type for dimensions?

This PR itself is not narrowing/widening the data type for dimension. It simply adds a compiler flag to control this.

apeforest:
@eric-haibin-lin @yuxihu @samskalicky @anirudh2290 @reminisce  Please let me know if your concerns have been addressed. Thanks!

marcoabreu:
LGTM from a CI perspective

TaoLv:
Thank you for the fix, @apeforest. Now it's approved. Just copy some comments from the mshadow PR here: int64 data type only affects the definition of some variables and the total size of a NDArray. We won't pass any NDArray with dim[x] > INT32_MAX into a operator. So it will not break the existing definition of BLAS and MKL-DNN APIs.

apeforest:
@marcoabreu @TaoLv @anirudh2290 could you please help to merge it if it’s good to ship? Thanks

tests/nightly/JenkinsfileForBinaries
```diff
@@ -59,6 +78,22 @@ core_logic: {
         }
       }
     },
+    'Test Large Tensor Size: CPU': {
+      node(NODE_LINUX_CPU) {
+        ws('workspace/large_tensor-cpu') {
+            utils.unpack_and_init('cpu_int64', mx_cmake_lib)
```
edisongustavo: I guess this should have been `ubuntu_cpu_int64`.

The nightly tests are failing because no one is publishing this stash: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/NightlyTestsForBinaries/detail/master/292/pipeline

As you can see from "Build / CPU: USE_INT64_TENSOR_SIZE", it publishes `ubuntu_cpu_int64`.

Issue #14522, opened 2019-03-26
mx.nd.Custom conflicts with memory management


YutingZhang:
When training/running a large neural network with CustomOP, mxnet can get stuck. My speculation is that if memory management (e.g., releasing/reallocating GPU memory, raising an "out of memory error") is needed while running CustomOP, mxnet can get stuck. 

A minimum piece code to show that CustomOP can deadlock with memory management ("out-of-memory in this case"):
As expected, main(1) should work. 
main(100) should give "out of GPU memory" error. However, it just got stuck.

The real-world problem I met is not just about inability to give an "out of memory" error. It seems MxNet can release/reallocate memory dynamically, but this probably also deadlock with CustomOP, so my program, which could fit into GPU memory, can also get stuck

```python
import mxnet as mx


class MyMulMax(mx.operator.CustomOp):

    def __init__(self):
        super().__init__()

    def forward(self, is_train, req, in_data, out_data, aux):
        a, b = in_data[0:2]
        c = mx.nd.batch_dot(a, b)
        d = mx.nd.max(c, axis=-1, keepdims=True)
        self.assign(out_data[0], req[0], d)

    def backward(self, req, out_grad, in_data, out_data, in_grad, aux):
        self.assign(in_grad[0], req[0], 0)
        self.assign(in_grad[1], req[1], 0)


@mx.operator.register("MyMulMax")
class MyMulMaxProp(mx.operator.CustomOpProp):

    def __init__(self):
        super().__init__()

    def list_arguments(self):
        return ['a', 'b']

    def list_outputs(self):
        return ['d']

    def infer_shape(self, in_shape):
        return in_shape, [list(in_shape[0][:-1] + [1])]

    def create_operator(self, ctx, shapes, dtypes):
        return MyMulMax()


def main(n):
    with mx.Context(mx.gpu(0)):
        a = mx.nd.random.uniform(shape=(n, 6000, 1))
        b = mx.nd.random.uniform(shape=(n, 1, 7000))
        d = mx.nd.Custom(a, b, op_type="MyMulMax")
        d_np = d.asnumpy()


if __name__ == "__main__":
    main(1)
    print("DONE -- 1")
    main(100)
    print("DONE -- 2")
```

Tested with nightly build of mxnet-cu90mkl with Python3.6

YutingZhang:
@arcadiaphy @wkcn @anirudh2290 Interested in looking into this?

wkcn:
[...]

PR #13932, opened 2019-01-18
[v1.4.x] Support populating errors back to MXNet engine in callback

yuxihu:
This PR adds an optional dmlc::Error* argument in MXNet engine callback functions. The callers can leverage this argument to populate errors back to MXNet engine through callback such that the errors can be handled properly by MXNet engine. It is needed to populate the errors detected in Horovod back to MXNet engine for Hovorod and MXNet integration (see [PR](https://github.com/ctcyang/horovod/pull/24)). This change does not affect existing use cases.

PR #14575, opened 2019-03-31
fix custom exception handling

arcadiaphy:
## Description ##
This PR fixes Problem 1 in #14522:

> 1. The exception thrown in custom thread cannot be caught in main thread, causing program crash.

It stores exception of custom op and re-throws in sync function.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/src/engine/threaded_engine.cc
+++ b/src/engine/threaded_engine.cc
@@ -31,6 +31,7 @@
 #include <utility>
 #include "./threaded_engine.h"
 #include "../common/cuda_utils.h"
+#include "../operator/custom/custom-inl.h"
 
 namespace mxnet {
 namespace engine {
@@ -373,10 +374,12 @@ void ThreadedEngine::DeleteVariable(SyncFn delete_fn,
 }
 
 void ThreadedEngine::WaitForVar(VarHandle var) {
+  using mxnet::op::custom::CustomOperator;
   BulkFlush();
   ThreadedVar* threaded_var = ThreadedVar::CastFromBase(var);
   if (threaded_var->ready_to_read()) {
     ThrowException(threaded_var);
+    CustomOperator::Get()->ThrowException();
     return;
   }
   if (engine_info_) {
@@ -407,6 +410,7 @@ void ThreadedEngine::WaitForVar(VarHandle var) {
   }
 
   ThrowException(threaded_var);
+  CustomOperator::Get()->ThrowException();
 }
 
 void ThreadedEngine::WaitForAll() {
--- a/src/engine/threaded_engine.cc
+++ b/src/engine/threaded_engine.cc
@@ -31,7 +31,6 @@
 #include <utility>
 #include "./threaded_engine.h"
 #include "../common/cuda_utils.h"
-#include "../operator/custom/custom-inl.h"
 
 namespace mxnet {
 namespace engine {
@@ -374,12 +373,10 @@ void ThreadedEngine::DeleteVariable(SyncFn delete_fn,
 }
 
 void ThreadedEngine::WaitForVar(VarHandle var) {
-  using mxnet::op::custom::CustomOperator;
   BulkFlush();
   ThreadedVar* threaded_var = ThreadedVar::CastFromBase(var);
   if (threaded_var->ready_to_read()) {
     ThrowException(threaded_var);
-    CustomOperator::Get()->ThrowException();
     return;
   }
   if (engine_info_) {
@@ -410,7 +407,6 @@ void ThreadedEngine::WaitForVar(VarHandle var) {
   }
 
   ThrowException(threaded_var);
-  CustomOperator::Get()->ThrowException();
 }
 
 void ThreadedEngine::WaitForAll() {
--- a/src/operator/custom/custom-inl.h
+++ b/src/operator/custom/custom-inl.h
@@ -48,6 +48,11 @@ namespace mxnet {
 namespace op {
 namespace custom {
 
+struct CustomTask {
+  std::function<void(void)> fn;
+  mxnet::engine::CallbackOnComplete on_complete;
+};
+
 class CustomOperator {
  public:
   void Register(const std::string &op_type, CustomOpPropCreator creator) {
@@ -92,7 +97,7 @@ class CustomOperator {
       return;
     }
     std::unique_lock<std::mutex> lock(mutex_);
-    q_.push([=]() mutable {
+    q_.push({[=]() mutable {
       bool prev_recording = Imperative::Get()->set_is_recording(recording);
       bool prev_training = Imperative::Get()->set_is_training(training);
 
@@ -129,7 +134,7 @@ class CustomOperator {
           },
           ctx.run_ctx.ctx, vars, vars2, FnProperty::kNormal, 0,
           "CustomOperator");
-    });
+    }, ctx.async_on_complete});
     // increase num_threads if there is not enough threads to execute custom operator
     if (q_.size() > num_free_threads)
       CreateThreads(q_.size() - num_free_threads);
@@ -145,6 +150,7 @@ class CustomOperator {
     num_free_threads = 0;
     destructing_ = false;
     naive_engine_ = true;
+    exception_ = nullptr;
     if (std::string("NaiveEngine") != dmlc::GetEnv("MXNET_ENGINE_TYPE", std::string())) {
       naive_engine_ = false;
     }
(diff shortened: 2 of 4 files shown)


src/operator/custom/custom-inl.h
```diff
@@ -172,10 +186,16 @@ class CustomOperator {
...
       while (!q_.empty()) {
         --num_free_threads;
-        auto fn = q_.front();
+        auto task = q_.front();
         q_.pop();
         lock.unlock();
-        fn();
+        try {
+          task.fn();
+        } catch (dmlc::Error& e) {
```
larroy: What about other types of exceptions? Could we add a comment in the code clarifying this?

tests/python/unittest/test_operator.py
```diff
@@ -5237,6 +5238,17 @@ def custom_add():
         p.join(5)
         assert not p.is_alive(), "deadlock may exist in custom operator"
 
+    # test except handling
+    # see https://github.com/apache/incubator-mxnet/pull/14575
+    def custom_add_exc():
```
larroy: Can we add a comment that an exception is expected due to shapes I assume?

piyushghai:
@mxnet-label-bot Add [Backend, Operator]

arcadiaphy:
@anirudh2290 I've simplified the exception catching according to your suggestion.
Agreed that the situation 2 is really tricky, because for `ExecType::kAsync` op, the on_complete callback is async to op computation, and right now there are no mechanism to ensure on_complete not skipped when exception happens.

src/operator/custom/custom-inl.h
arcadiaphy: What other types? I think the only valid exception is dmlc::Error, other exception means wrong code.

tests/python/unittest/test_operator.py
arcadiaphy: Sure, I'll add it.

src/engine/threaded_engine.cc
```diff
@@ -373,10 +374,12 @@ void ThreadedEngine::DeleteVariable(SyncFn delete_fn,
 }
 
 void ThreadedEngine::WaitForVar(VarHandle var) {
+  using mxnet::op::custom::CustomOperator;
   BulkFlush();
   ThreadedVar* threaded_var = ThreadedVar::CastFromBase(var);
   if (threaded_var->ready_to_read()) {
     ThrowException(threaded_var);
+    CustomOperator::Get()->ThrowException();
```
anirudh2290: Lets remove this. ThreadedEngine should not depend on custom operator code.

src/operator/custom/custom-inl.h
```diff
@@ -96,7 +96,14 @@ class CustomOperator {
...
 
-      func();
+      try {
+        func();
+      } catch (dmlc::Error& e) {
+        exception_ =
+            std::make_shared<std::exception_ptr>(std::current_exception());
+        ctx.async_on_complete();
+        return;
+      }
```
anirudh2290: I think we can solve both 1 and 2 this way: After func is called do wait_to_read on all elements in arrs. Then catch and save. Remove lines 104 and 105. In PushSync, check if exception is set and rethrow exception. Also catch it and call async_on_complete in pushsync. and return.

Something like the following:

```
Engine::Get()->PushSync(
     [=](RunContext rctx) {
    try {
    if (exception_) {
       std::rethrow_exception(exception_);
    }
    } catch(dmlc::Error& err) {
         ctx.async_on_complete(&err);
         return;
    }
 }
```

Thanks to this support added for horovod: https://github.com/apache/incubator-mxnet/pull/13932 we may be able to leverage this to call async_on_complete with the error.

arcadiaphy: Adding wait_to_read in custom op can solve 1&2, and PushSync can be removed since the computation is already finished.

anirudh2290: we probably still need PushSync for the Sparse ndarray updates.

anirudh2290: we still need ExecType::kAsync. Custom operator is still async and when push is called it just pushes it into its custom op worker queue for execution later. Async will ensure that the threaded_engine_pooled and threaded_engine_per_device treat it as a special case and execute immediately instead of pushing the work again to one of the engine worker thread queue. Pushing to engine worker thread queue is unnecessary for custom op.

arcadiaphy: After testing, ExecType::kAsync is really needed. Adding wait_to_read in engine worker thread will cause deadlock.
But PushSync can be removed and works well.

anirudh2290: we probably still need it for sparse. since for sparse we are updating chunk it is a write option. WaitToRead may not be enough.

arcadiaphy: Yes, I also add WaitToWrite to make sure there's no left out exceptions.

arcadiaphy:
@anirudh2290 I'll test the example you've mentioned.

Let me mention another problem here:
Adding WaitToRead in custom thread will block the thread, which will cause deadlock. Suppose we only have one custom thread and custom op A is dependent upon custom B (A->B), then deadlock will happen due to not enough working custom thread.

```
import mxnet as mx

class A(mx.operator.CustomOp):
    def __init__(self):
        super(A, self).__init__()
        out_data[0][:] = mx.nd.Custom(in_data[0], in_data[1], op_type='B')
        in_grad[0][:] = out_grad[0]
        in_grad[1][:] = out_grad[0]

@mx.operator.register("A")
class AProp(mx.operator.CustomOpProp):
    def __init__(self):
        super(AProp, self).__init__()
    def list_arguments(self):
        return ['a', 'b']
    def list_outputs(self):
        return ['output']
        return in_shape, [in_shape[0]]
        return A()

class B(mx.operator.CustomOp):
    def __init__(self):
        super(B, self).__init__()
        out_data[0][:] = in_data[0] + in_data[1]
        in_grad[0][:] = out_grad[0]
        in_grad[1][:] = out_grad[0]

@mx.operator.register("B")
class BProp(mx.operator.CustomOpProp):
    def __init__(self):
        super(BProp, self).__init__()
    def list_arguments(self):
        return ['a', 'b']
    def list_outputs(self):
        return ['output']
        return in_shape, [in_shape[0]]
        return B()

def main():
    a = mx.nd.array([1, 2])
    b = mx.nd.array([4, 5])
    c = mx.nd.Custom(a, b, op_type='A')
    print c.asnumpy()

if __name__ == '__main__':
    main()
```

Run the code with:
```
MXNET_CUSTOM_OP_NUM_THREADS=1 python reproduce.py
```

Even if you have n threads, we can still deadlock the execution with n+1 custom op dependency chain: A->B->C...

wkcn:
@arcadiaphy Custom Operator creates new thread when in need now, and MXNET_CUSTOM_OP_NUM_THREADS is the maximum number of threads.

Removing this line https://github.com/apache/incubator-mxnet/blob/master/src/operator/custom/custom-inl.h#L185 to support unlimited number of threads.

anirudh2290:
@arcadiaphy Yes good point, the code you linked will cause a deadlock. this is also true if custom op forward or backward has asnumpy() or wait_to_read at the end even before applying your change, when you change MXNET_CUSTOM_OP_NUM_THREADS=1. Even before the custom op multi threaded change this was the case. 

Also, I am concerned about the performance because of the blocking call during op execution. I think we have reached a point where there is no option but to special case "CustomOperator" in ExecuteOprBlock by checking whether op is CustomOperator and if so, execute the operator instead of skipping (https://github.com/apache/incubator-mxnet/blob/master/src/engine/threaded_engine.h#L377) . I was trying to avoid this to keep the engine agnostic of the op its executing but this seems like the least impactful solution currently.

arcadiaphy:
@anirudh2290 If the engine code remains unchanged, I think enabling unlimited number of custom threads when needed is a quick fix.

arcadiaphy:
Also for unlimited threads, theoretically, there should be no performance impact.

anirudh2290:
I think we should make the modification in engine. Keeping that env variable is fine as Users may want to limit the custom op threads, so that customers don't run into situation where there are too many custom ops in their model and lot of threads are spawned and thread spawning overhead being high. Also, performance wise adding the waittoread and waittowrite can impact performance: every thread is now pushing all ops, waiting it to complete and then exiting as compared to custom op threads earlier which would just push op to engine and exit. Custom op threads were lightweight earlier and we should keep it that way.

arcadiaphy:
@anirudh2290 How about leave this PR temporarily and create a new PR to modify the engine code?

anirudh2290:
Yes I am fine with that

arcadiaphy:
close due to issue fixed in #14693

Issue #12932, opened 2018-10-23
Probability Distributions Support


thomelane:
## Probability Distributions Support

Would be great to have out-of-the-box support for distributions, similar to functionality provided by [TensorFlow Probability](https://www.tensorflow.org/probability) and [PyTorch Distributions](https://pytorch.org/docs/stable/distributions.html). My current use case is for Reinforcement Learning algorithms that learn stochastic action policies (i.e. learn parameters of a distribution from which actions are sampled), and I update these parameters using the likelihood.

MXNet would ideally have methods on each type of distribution for calculating:

* probability density,
* log probability of a data sample given the distribution,
* entropy of distribution,
* kl divergence of distribution with another distribution,
* sampling from distribution (mostly already implemented).

And would support a variety of distributions including:

* Categorical
* MultivariateNormal
* Bernoulli
* Beta
* Dirichlet
* Exponential
* Gamma
* Poisson
* Uniform

[MXFusion](https://github.com/amzn/MXFusion) is a related project but doesn't have the functionality mentioned above. And it would be ideal to have this as a submodule of the MXNet package.

thomelane:
@mxnet-label-bot [Feature Request, Gluon, Operators]

anirudhacharya:
Some of them are here - https://mxnet.incubator.apache.org/api/python/ndarray/random.html?highlight=mxnet.ndarray.random#random-distribution-generator

KL Loss - https://mxnet.incubator.apache.org/api/python/gluon/loss.html?highlight=kl#mxnet.gluon.loss.KLDivLoss

But yes we do need a probability submodule within mxnet

thomelane:
Cheers @anirudhacharya, but the functionality I'm referring to is not covered by those references.

I'm talking about functionality beyond just sampling from distributions: calculations of probability density, log probability of a data sample given the distribution, entropy of distribution, etc. And as far as I'm aware the `KL Loss` in MXNet only works with samples, rather than the theoretical KL Divergence between distributions, so this is also insufficient for certain use cases.

You can't back propagate gradients through samples, so that's why it's important to have such formulas (e.g. log probability) implemented.

I can see a single case of probability being returned by [`mxnet.ndarray.random.multinomial`](https://mxnet.incubator.apache.org/api/python/ndarray/random.html?highlight=mxnet.ndarray.random#mxnet.ndarray.random.multinomial) but this is only for the sampled data point, and not calculated for an arbitrary data point which is required.

[...]

PR #14579, opened 2019-04-01
PDF operators for each distribution

david-seiler:
## Description ##
This PR adds operators for computing the densities of samples drawn from any of the various distributions defined in operator/random, as well as their gradients.  There are lots of changes to test_random.py to test each PDF alongside its distribution; aside from that, the patch should be entirely stand-alone.  See pdf_op.cc for more-detailed description strings.

### Essentials ###

### DIFF
--- a/src/operator/random/pdf_op.cc
+++ b/src/operator/random/pdf_op.cc
@@ -1,0 +1,319 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * Copyright (c) 2018 by Contributors
+ * \file pdf_op.cc
+ * \brief CPU-operators for computing the pdf of random distributions. 
+ */
+
+#include "./pdf_op.h"
+
+namespace mxnet {
+namespace op {
+
+DMLC_REGISTER_PARAMETER(PdfParam);
+
+#define MXNET_OPERATOR_REGISTER_PDF(distr, pdffunc, num_parms, \
+                                    parm_name_1, parm_name_2, \
+                                    parm_desc_1, parm_desc_2, \
+                                    description, vectorparms) \
+  NNVM_REGISTER_OP(_random_pdf_##distr) \
+  .add_alias("random_pdf_" #distr) \
+  .describe(description()+std::string(ADD_FILELINE)) \
+  .set_num_inputs(num_parms+1) \
+  .set_num_outputs(1) \
+  .set_attr_parser(ParamParser<PdfParam>) \
+  .set_attr<nnvm::FListInputNames>("FListInputNames", \
+    [](const NodeAttrs& attrs) { \
+      std::vector<std::string> v = {"sample", parm_name_1, parm_name_2}; \
+      v.resize(num_parms+1); \
+      return v; \
+    }) \
+  .set_attr<mxnet::FInferShape>("FInferShape", PdfOpShape<vectorparms>) \
+  .set_attr<nnvm::FInferType>("FInferType", ElemwiseType<num_parms+1, 1>) \
+  .set_attr<FCompute>("FCompute<cpu>", PdfOpForward<cpu, pdffunc, num_parms, vectorparms>) \
+  .set_attr<nnvm::FGradient>("FGradient", ElemwiseGradUseInOut{"_backward_pdf_" #distr}) \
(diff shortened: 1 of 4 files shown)


src/operator/random/pdf_op.h
```diff
@@ -0,0 +1,594 @@
...
+        }
+    }
+  }
+};
+
+struct PdfParam : public dmlc::Parameter<PdfParam> {
+  bool is_log;
+  DMLC_DECLARE_PARAMETER(PdfParam) {
+    DMLC_DECLARE_FIELD(is_log).set_default(false)
+    .describe("Whether to compute the log PDF or not.");
```
asmushetzel: May be that can be phrased a bit better.

src/operator/random/pdf_op.h
```diff
@@ -0,0 +1,594 @@
...
+MSHADOW_XINLINE mshadow::half::half_t ceph_psi(mshadow::half::half_t val) {
+    return special_functions::cephes::psi<float>(val);
+}
+
+template<bool logpdf>
+struct PDF_Uniform {
+  template<typename DType, typename IType1, typename IType2>
+  MSHADOW_XINLINE static void Map(int start, int length, int sample_size, int index,
+                                  DType *out, IType1 *sample, IType2 *lower, IType2 *upper) {
+    const DType l(lower[index]), h(upper[index]);
```
asmushetzel: Providing "index" to this and all similar functions below is redundant. According to the processing in LaunchExWrapper it always hold that index == start/sample_size. Still we may want to keep "index" as an explicit parameter for clarity. So o.k. with either choice.

asmushetzel:
Should be noted that this PR allows to compute log-pdf and pdf. And that it also adds pdf of Dirichlet distribution which does not yet exist in MXNet. 
For consistency, we should then add also a sampler for Dirichlet-distributions after this PR gets merged.

asmushetzel:
This addresses several requests from issue #12932

src/operator/random/pdf_op.cc
```diff
@@ -0,0 +1,292 @@
...
+
+    sample = [[1,2,3,4],
+              [1,2,3,4]]
+    random_pdf_generalized_negative_binomial(sample=sample, alpha=[1, 0.6666], mu=[1, 1.5]) =
+        [[0.25       0.125      0.0625     0.03125   ]
+         [0.26517063 0.16573331 0.09667706 0.05437994]]
+)code");
+}
+
+inline std::string dirichlet_desc() {
```
anirudhacharya: I don't think we have a function to draw samples from a dirichlet distribution? this only computes the pdf of the distribution?

marcoabreu:
Is this PR pointing against the 1.3 branch on purpose?

piyushghai:
@david-seiler Seems like you want to add a particular feature to a previous version of MXNet. 
The correct process to do this would be : 

- Raise a PR against the master branch. 
- After that PR is merged into master, cherry-pick the same commit in another PR raised against v1.3.x (Or a different version) branch. 

@mxnet-label-bot Add [pr-awaiting-review, operator]

david-seiler:
> @david-seiler Seems like you want to add a particular feature to a previous version of MXNet.

I started with 1.3.x because it's what we're using internally (for now); I'll close this in favor of a new PR against master once I've updated to build against master and addressed the rest of the review comments.

david-seiler:
This PR now correctly targets master -- still from a branch named v1.3.x on my side, since I couldn't see how to reset that, but that branch now follows apache:master with this one commit on top.  Sorry for the confusion, but it seems better on net than throwing away the existing conversation.

src/operator/random/pdf_op.cc
david-seiler: Updated the commit message to clarify that, unlike the other distributions, we don't have a Dirichlet sampler yet.

src/operator/random/pdf_op.h
david-seiler: It's gone.

src/operator/random/pdf_op.h
david-seiler: It's now "If set, compute the density of the log-probability instead of the probability."

src/operator/random/pdf_op.cc
```diff
@@ -0,0 +1,292 @@
...
+Given *lam* of dimension *n* and *sample* of dimension *n+1*, produces
+output of dimension *n+1* such that each *n*-dimensional index *i* in
+the output array holds the PDFs of the samples at index *i* in *sample*,
+parameterized by the value of *lam* at index *i*.
+
+Examples::
+
+  random_pdf_exponential(sample=[[1, 2, 3]], lam=[1]) =
+      [[0.36787945 0.13533528 0.04978707]]
+
```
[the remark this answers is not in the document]
david-seiler: Added second examples for the exponential and Dirichlet.

src/operator/random/pdf_op.cc
```diff
@@ -0,0 +1,292 @@
...
+inline std::string uniform_desc() {
+  return std::string(R"code(Computes the value of the PDF of *sample* of
+uniform distributions on the intervals given by *[low,high)*.
+
+Given *low* and *high* each of dimension *n* and *sample* of dimension *n+1*,
+produces output of dimension *n+1* such that each *n*-dimensional index *i*
+in the output array holds the PDFs of the samples at index *i* in *sample*,
+parameterized by the values of *low* and *high* at index *i*.
+
```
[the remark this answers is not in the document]
david-seiler: Called that out explicitly.

david-seiler:
The tests in Jenkins have been stuck ever single I switched this PR to master, and I haven't found a way to unstick them, so please join me in a new PR of the same code: https://github.com/apache/incubator-mxnet/pull/14617

All the comments from this PR should be addressed there.

PR #14587, opened 2019-04-01
[MXNET-1344, 1346][FIT API] Retrieve Batch size and Logging verbose support for Gluon fit() API

karan6181:
## Description ##
This PR infers the batch size from` gluon.data.DataLoader` in fit() API so that the end user doesn't need to provide a batch size during fit() call. This PR also adds the verbose option in logging handler where the end user can control the granularity of logging information.

I haven't removed the `batch_size=None` option in fit() API for now. This will be addressed in separate PR as there will be modification in Integration, Unit test, and tutorial which uses fit() API with `batch_size` option. (DONE)

JIRA epic for FIT-API: https://issues.apache.org/jira/projects/MXNET/issues/MXNET-1333

### Essentials ###
- [x] The PR title starts with [MXNET-1344](https://issues.apache.org/jira/browse/MXNET-1344), [MXNET-1346](https://issues.apache.org/jira/browse/MXNET-1346)

## Comments ##

@roywei @nswamy @abhinavs95

### DIFF
--- a/python/mxnet/gluon/estimator/estimator.py
+++ b/python/mxnet/gluon/estimator/estimator.py
@@ -21,7 +21,8 @@
 
 import copy
 import warnings
-
+from mxnet import nd
+import numpy as np
 from .event_handler import LoggingHandler
 from ... import gluon, autograd
 from ...context import Context, cpu, gpu, num_gpus
@@ -173,7 +174,7 @@ class Estimator(object):
         label = gluon.utils.split_and_load(label, ctx_list=ctx, batch_axis=0)
         return data, label
 
-    def infer_data_info(self, data):
+    def _infer_data_info(self, data):
         """Retrieve the data information such as batch size,
         Number of batches, and total number of samples
 
@@ -192,16 +193,22 @@ class Estimator(object):
             Batch size
         """
         if isinstance(data, gluon.data.DataLoader):
-            num_batches = len(data)
-            total_samples = len(data._dataset)
+            if isinstance(data._dataset, gluon.data.ArrayDataset):
+                total_samples = data._dataset._data[0].shape[0]
+            elif isinstance(data._dataset, nd.ndarray.NDArray):
+                total_samples = data._dataset.shape[0]
+            else:
+                total_samples = len(data._dataset)
+
             if total_samples == 0:
                 raise ValueError("DataLoader is Empty. Please refer to gluon.data.DataLoader "
-                                 "for more detail on how to contruct a DataLoader")
+                                 "for more detail on how to construct a DataLoader")
             for batch in data:
                 batch_size = batch[0].shape[0]
                 break
         else:
             raise ValueError("Please provide the data as gluon.data.DataLoader")
+        num_batches = int(np.ceil(total_samples / batch_size))
         return num_batches, total_samples, batch_size
 
     def evaluate(self,
@@ -268,7 +275,7 @@ class Estimator(object):
         """
 
         self.epochs = epochs
-        num_batches, total_samples, batch_size = self.infer_data_info(train_data)
+        num_batches, total_samples, batch_size = self._infer_data_info(train_data)
 
         if isinstance(self.context, list):
             if batch_size < len(self.context):
--- a/python/mxnet/gluon/estimator/estimator.py
+++ b/python/mxnet/gluon/estimator/estimator.py
@@ -238,7 +238,7 @@ class Estimator(object):
                 not any(isinstance(handler, LoggingHandler) for handler in event_handlers):
             event_handlers.append(LoggingHandler())
             warnings.warn("No Event Handler specified, default `LoggingHandler()` "
-                          "is used with verbose=LoggingHandler.LOG_VERBOSITY_PER_BATCH. "
+                          "is used with verbose=LoggingHandler.LOG_VERBOSITY_PER_EPOCH. "
                           "Please look at gluon.estimator.event_handler for more detail.")
 
         train_begin, epoch_begin, batch_begin, \
--- a/python/mxnet/gluon/estimator/estimator.py
+++ b/python/mxnet/gluon/estimator/estimator.py
@@ -236,8 +236,8 @@ class Estimator(object):
                 not any(isinstance(handler, LoggingHandler) for handler in event_handlers):
             event_handlers.append(LoggingHandler())
             warnings.warn("No Event Handler specified, default `LoggingHandler()` "
-                          "is used with verbose=1. Please look at gluon.estimator.event_handler"
-                          "for more detail.")
+                          "is used with verbose=LoggingHandler.ONLY_EPOCH. Please "
+                          "look at gluon.estimator.event_handler for more detail.")
 
         train_begin, epoch_begin, batch_begin, \
         batch_end, epoch_end, train_end = self._categorize_handlers(event_handlers)
--- a/python/mxnet/gluon/estimator/estimator.py
+++ b/python/mxnet/gluon/estimator/estimator.py
@@ -127,7 +127,6 @@ class Estimator(object):
             raise ValueError("context must be a Context or a list of Context, "
                              "refer to mxnet.Context:{}".format(context))
 
-
         # initialize the network
         self.initializer = initializer
         if self.initializer:
@@ -135,7 +134,7 @@ class Estimator(object):
                 # if already initialized, re-init with user specified initializer
                 warnings.warn("Network already initialized, re-initializing with %s. "
                               "You don't need to pass initializer if you already "
-                              "initialized your net."% type(self.initializer).__name__)
+                              "initialized your net." % type(self.initializer).__name__)
                 self.net.initialize(init=self.initializer, ctx=self.context, force_reinit=True)
             else:
                 # initialize with user specified initializer
@@ -241,16 +240,35 @@ class Estimator(object):
             from a data batch and load into contexts(devices)
         """
 
-
         self.epochs = epochs
-        if not batch_size:
-            batch_size = 32 * len(self.context)
+        if isinstance(train_data, gluon.data.DataLoader):
+            num_batches = len(train_data)
+            total_samples = len(train_data._dataset)
+            if total_samples == 0:
+                raise ValueError("DataLoader is Empty. Please refer to gluon.data.DataLoader "
+                                 "for more detail")
+            for dt, lb in train_data:
+                batch_size = dt.shape[0]
+                break
+        else:
+            raise ValueError("Please provide a data as gluon.data.DataLoader")
+
+        if isinstance(self.context, list):
+            if batch_size < len(self.context):
+                raise ValueError("The batch size value is small in comparison to the provided"
+                                 "CPU/GPU context. Please provide the batch size value(power of 2) "
+                                 "greater than the number of GPUs in your system")
+
+        self.train_stats['batch_size'] = batch_size
 
         event_handlers = event_handlers or []
         # provide default logging handler
         if not event_handlers or \
                 not any(isinstance(handler, LoggingHandler) for handler in event_handlers):
-            event_handlers.append(LoggingHandler(self))
+            event_handlers.append(LoggingHandler(self, verbose=1))
+            warnings.warn("No Event Handler specified, default logging handler "
+                          "is used with verbose=1. Please see gluon.estimator.event_handler"
+                          "for more detail.")
 
         # training begin
         for handler in event_handlers:
@@ -303,11 +321,14 @@ class Estimator(object):
                     loss_metric.update(0, [l for l in loss])
                     self.train_stats['batch_' + loss_metric.name] = loss_metric.get()[1]
 
+                # last batch size may be different from the rest
+                if i == num_batches - 1:
+                    batch_size = total_samples - (batch_size * i)
+                    completed_samples = total_samples
+                else:
+                    completed_samples = batch_size * (i + 1)
+
                 try:
-                    completed_samples = len(train_data._dataset) if i == len(train_data._dataset) - 1 \
-                                        else batch_size * (i + 1)
-                    # We need to check if this is the last batch in the current epoch and select
-                    # the value to print appropriately
                     self.train_stats['step'] = "{}/{}".format(completed_samples, len(train_data._dataset))
                 except AttributeError:
                     self.train_stats['step'] = i
@@ -333,6 +354,9 @@ class Estimator(object):
 
             if self.stop_training:
                 break
+
+            # Reset batch size since last batch size may be different
+            batch_size = self.train_stats['batch_size']
 
         # train end
         for handler in event_handlers:
--- a/python/mxnet/gluon/estimator/estimator.py
+++ b/python/mxnet/gluon/estimator/estimator.py
@@ -284,7 +284,7 @@ class Estimator(object):
         # provide default logging handler
         if not event_handlers or \
                 not any(isinstance(handler, LoggingHandler) for handler in event_handlers):
-            event_handlers.append(LoggingHandler(self, verbose=1))
+            event_handlers.append(LoggingHandler(self))
             warnings.warn("No Event Handler specified, default `LoggingHandler()` "
                           "is used with verbose=1. Please look at gluon.estimator.event_handler"
                           "for more detail.")
--- a/python/mxnet/gluon/estimator/estimator.py
+++ b/python/mxnet/gluon/estimator/estimator.py
@@ -203,7 +203,6 @@ class Estimator(object):
     def fit(self, train_data,
             val_data=None,
             epochs=1,
-            batch_size=None,
             event_handlers=None,
             batch_fn=None):
         """Main training loop
--- a/python/mxnet/gluon/estimator/estimator.py
+++ b/python/mxnet/gluon/estimator/estimator.py
@@ -247,11 +247,11 @@ class Estimator(object):
             if total_samples == 0:
                 raise ValueError("DataLoader is Empty. Please refer to gluon.data.DataLoader "
                                  "for more detail")
-            for dt, lb in train_data:
+            for dt, _ in train_data:
                 batch_size = dt.shape[0]
                 break
         else:
-            raise ValueError("Please provide a data as gluon.data.DataLoader")
+            raise ValueError("Please provide the data as gluon.data.DataLoader")
 
         if isinstance(self.context, list):
             if batch_size < len(self.context):
--- a/python/mxnet/gluon/estimator/estimator.py
+++ b/python/mxnet/gluon/estimator/estimator.py
@@ -21,8 +21,6 @@
 
 import copy
 import warnings
-from mxnet import nd
-import numpy as np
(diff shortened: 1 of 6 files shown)


python/mxnet/gluon/estimator/estimator.py
```diff
@@ -241,16 +240,35 @@ def fit(self, train_data,
...
         self.epochs = epochs
-        if not batch_size:
-            batch_size = 32 * len(self.context)
+        if isinstance(train_data, gluon.data.DataLoader):
+            num_batches = len(train_data)
+            total_samples = len(train_data._dataset)
+            if total_samples == 0:
+                raise ValueError("DataLoader is Empty. Please refer to gluon.data.DataLoader "
+                                 "for more detail")
+            for dt, _ in train_data:
```
abhinavs95: Can this be condensed to 1 line? How about `np.ceil(len(train_data._dataset)/len(train_data))`

tests/python/unittest/test_gluon_estimator.py
```diff
@@ -275,3 +276,45 @@ def test_context():
...
+    label = mx.nd.random.randint(low=0, high=2, shape=(0,))
+    batch_size = 2
+    data_arr = mx.gluon.data.dataset.ArrayDataset(data, label)
+    data_loader = mx.gluon.data.DataLoader(data_arr, batch_size=batch_size)
+    est = estimator.Estimator(net=net, loss=loss, context=ctx)
+    with assert_raises(ValueError):
+        est.fit(train_data=data_loader)
+
+    # Batch size less than context
+    ctx = [mx.gpu(0), mx.gpu(1), mx.gpu(2), mx.gpu(3)]
```
abhinavs95: Are we sure we have 4 GPUs for unit tests? This might fail otherwise

python/mxnet/gluon/estimator/estimator.py
```diff
@@ -334,6 +355,9 @@ def fit(self, train_data,
             if self.stop_training:
                 break
 
+            # Reset batch size since last batch size may be different
+            batch_size = self.train_stats['batch_size']
```
roywei: here batch size is more like a hyper parameter, we don't want to change it for the last batch

python/mxnet/gluon/estimator/event_handler.py
```diff
@@ -76,14 +76,20 @@ class LoggingHandler(EventHandler):
         file name to save the logs
     file_location: str
         file location to save the logs
+    verbose: int, default 0
```
roywei: nit: default is 1

python/mxnet/gluon/estimator/estimator.py
roywei: this will error out if  train data is a tuple of 3 or more, get it from batch[0].shape[0], it's safer to assume first of batch is the data, batch[1], batch[2] can be label1 label2 for multi label classifications.

piyushghai:
Thanks for your contribution @karan6181. 

@mxnet-label-bot Add [Gluon, pr-awaiting-review]

python/mxnet/gluon/estimator/estimator.py
```diff
@@ -241,16 +240,35 @@ def fit(self, train_data,
...
+            for dt, _ in train_data:
+                batch_size = dt.shape[0]
+                break
+        else:
+            raise ValueError("Please provide the data as gluon.data.DataLoader")
+
+        if isinstance(self.context, list):
+            if batch_size < len(self.context):
+                raise ValueError("The batch size value is small in comparison to the provided"
+                                 "CPU/GPU context. Please provide the batch size value(power of 2) "
```
piyushghai: Should the batch size always be a power of 2 ?

python/mxnet/gluon/estimator/event_handler.py
```diff
@@ -98,29 +104,33 @@ def train_end(self):
         pass
 
     def batch_begin(self):
-        self.batch_start = time.time()
+        if self.verbose == 2:
```
piyushghai: Can we replace the hard coding with a constant?
It'll help in improving the readability as well as be less error prone if this needs further changes in the future.

python/mxnet/gluon/estimator/estimator.py
```diff
@@ -241,16 +240,35 @@ def fit(self, train_data,
...
-
         self.epochs = epochs
-        if not batch_size:
-            batch_size = 32 * len(self.context)
+        if isinstance(train_data, gluon.data.DataLoader):
+            num_batches = len(train_data)
+            total_samples = len(train_data._dataset)
+            if total_samples == 0:
+                raise ValueError("DataLoader is Empty. Please refer to gluon.data.DataLoader "
+                                 "for more detail")
```
roywei: nit: for more detail on how to contruct a dataloader?

python/mxnet/gluon/estimator/estimator.py
roywei: nit: better message, something like: batch size is too small to be split and loaded into the list of devices you provided in context

python/mxnet/gluon/estimator/estimator.py
```diff
@@ -241,16 +240,35 @@ def fit(self, train_data,
...
+        self.train_stats['batch_size'] = batch_size
 
         event_handlers = event_handlers or []
         # provide default logging handler
         if not event_handlers or \
                 not any(isinstance(handler, LoggingHandler) for handler in event_handlers):
-            event_handlers.append(LoggingHandler(self))
+            event_handlers.append(LoggingHandler(self, verbose=1))
+            warnings.warn("No Event Handler specified, default logging handler "
+                          "is used with verbose=1. Please see gluon.estimator.event_handler"
```
roywei: nit: specify LoggingHandler here

tests/python/unittest/test_gluon_estimator.py
piyushghai: This can be re-written as : ctx = [mx.gpu(i) for i in range(mx.context.num_gpus())]

tests/python/unittest/test_gluon_estimator.py
```diff
@@ -275,3 +276,45 @@ def test_context():
...
+    data_arr = mx.gluon.data.dataset.ArrayDataset(data, label)
+    data_loader = mx.gluon.data.DataLoader(data_arr, batch_size=batch_size)
+    est = estimator.Estimator(net=net, loss=loss, context=ctx)
+    with assert_raises(ValueError):
+        est.fit(train_data=data_loader)
+
+    # Correct batch size verification
+    for dt, lb in data_loader:
+        assert batch_size == dt.shape[0]
+        break
```
piyushghai: If you are doing it for just the first element in the data_loader, then no need to loop through it.

python/mxnet/gluon/estimator/estimator.py
karan6181: That won't work. Here is an example:

Total number of samples: 270 (`len(train_data._dataset)`)
Batch size(In dataLoader): 32
Number of batches from dataLoader: 9 (`len(train_data)`) [32 * 8 + 14]

`np.ceil(len(train_data._dataset)/len(train_data))`: 30.0 (Correct batch size is 32)

tests/python/unittest/test_gluon_estimator.py
karan6181: This specific test checks the total context should be less than the batch_size to equally split and loaded the data. It doesn't execute the forward/backward pass since the assert statement is at the start of fit() method. 

What could be the best way to execute this specific test case? Should it be only run on GPU machine by checking it as `mx.context.num_gpus()`?

python/mxnet/gluon/estimator/estimator.py
karan6181: Yes. I have modified the code structure in this commit id [5f94478](https://github.com/apache/incubator-mxnet/pull/14587/commits/5f9447892bfc453963fe6765f171efb556a12649). I am maintaining a separate variable as `current_batch_size`.

python/mxnet/gluon/estimator/event_handler.py
karan6181: Done. Thanks!

python/mxnet/gluon/estimator/estimator.py
karan6181: True. modified the code. [5f94478](https://github.com/apache/incubator-mxnet/pull/14587/commits/5f9447892bfc453963fe6765f171efb556a12649)

python/mxnet/gluon/estimator/estimator.py
karan6181: Not really. It's more of related to mapping of virtual to physical processor.

python/mxnet/gluon/estimator/estimator.py
karan6181: Done. Thanks!

python/mxnet/gluon/estimator/estimator.py
karan6181: Done. Thanks!

python/mxnet/gluon/estimator/estimator.py
karan6181: Done. Thanks!

tests/python/unittest/test_gluon_estimator.py
karan6181: This specific testcase doesn't require the actual GPUs to execute, which means, the aim of this test case is to check that the batch size should be greater than the total number of context to fully utilize the system. The code doesn't execute the forward/backward pass since the assert statement is at the start of fit() method.

should this test be run only on multi-gpu context environment?

tests/python/unittest/test_gluon_estimator.py
karan6181: We can't get the correct batch_size value by accessing the instance variable. So iterating through the data once is one of the solution to infer batch size.

tests/python/unittest/test_gluon_estimator.py
piyushghai: I believe so, or you can do something like this : 
```ctx = [mx.cpu(0), mx.cpu(0), mx.cpu(0), mx.cpu(0)]```

That would make it GPU agnostic and would just check for your test case to validate whether ```batch_size > len(ctx)``` 

WDYT ?

File: python/mxnet/gluon/estimator/event_handler.py [hunk 10]
```diff
@@ -76,14 +76,20 @@ class LoggingHandler(EventHandler):
         file name to save the logs
     file_location: str
         file location to save the logs
+    verbose: int, default 1
+        Limit the display level of training progress
+        verbose=0: display nothing(silent)
```
nswamy: then why have this eventHandler?

python/mxnet/gluon/estimator/estimator.py
```diff
@@ -175,14 +173,45 @@ def _batch_fn(self, batch, ctx, is_iterator=False):
...
+        num_batches: int
+            Number of batches the data is divided into
+        total_samples: int
+            Total Number of samples
+        batch_size: int
+            Batch size
+        """
+        if isinstance(data, gluon.data.DataLoader):
+            num_batches = len(data)
+            total_samples = len(data._dataset)
```
nswamy: this could be an expensive operation.

python/mxnet/gluon/estimator/estimator.py
abhinavs95: yes that won't work, thanks for clarifying.

python/mxnet/gluon/estimator/estimator.py
karan6181: If the DataLoader created using `MXNet.gluon.data.dataset.ArrayDataset` then the `ArrayDataset` has the instance variable `_length` which infers the length of an actual data. Another way to infer the total number of sample is to extract the `shape` information from the data.

tests/python/unittest/test_gluon_estimator.py
abhinavs95: The net is initialized when creating the estimator object which happens before fit. The test will fail there if at least 4 GPUs are not available. 

I think checking the number of GPUs first acc your suggestion should fix it, or we can move the test to nightly where GPUs are available

tests/python/unittest/test_gluon_estimator.py
```diff
@@ -275,3 +278,44 @@ def test_context():
...
+        est.fit(train_data=data_loader)
+
+    # Batch size less than context
+    ctx = [mx.gpu(i) for i in range(4)]
+    data = mx.nd.random.uniform(shape=(num_samples, 3, 28, 28))
+    label = mx.nd.random.randint(low=0, high=2, shape=(num_samples,))
+    batch_size = 2
+    data_arr = mx.gluon.data.dataset.ArrayDataset(data, label)
+    data_loader = mx.gluon.data.DataLoader(data_arr, batch_size=batch_size)
+    est = estimator.Estimator(net=net, loss=loss, context=ctx)
```
abhinavs95: A new net needs to be created here, otherwise it will use the one created before and the test won't be accurate.

tests/python/unittest/test_gluon_estimator.py
```diff
@@ -260,7 +251,8 @@ def test_context():
                     loss=loss,
                     metrics=metrics)
     # input list of context
-    ctx = [mx.gpu(0), mx.gpu(1)]
+    ctx = [mx.cpu() for _ in range(2)]
```
roywei: change this ctx=[ list of gpu ctxs] if num_gpus >1 else [mx.cpu()]

karan6181: Done. Thanks!

python/mxnet/gluon/estimator/estimator.py
```diff
@@ -241,16 +274,23 @@ def fit(self, train_data,
             from a data batch and load into contexts(devices)
         """
 
-
         self.epochs = epochs
-        if not batch_size:
-            batch_size = 32 * len(self.context)
+        num_batches, total_samples, batch_size = self._infer_data_info(train_data)
+
+        if isinstance(self.context, list):
```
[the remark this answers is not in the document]
karan6181: No longer relevant. Modified the designed.

python/mxnet/gluon/estimator/estimator.py
```diff
@@ -175,14 +174,51 @@ def _batch_fn(self, batch, ctx, is_iterator=False):
...
+            Number of batches the data is divided into
+        total_samples: int
+            Total Number of samples
+        batch_size: int
+            Batch size
+        """
+        if isinstance(data, gluon.data.DataLoader):
+            if isinstance(data._dataset, gluon.data.ArrayDataset):
+                total_samples = data._dataset._data[0].shape[0]
+            elif isinstance(data._dataset, nd.ndarray.NDArray):
```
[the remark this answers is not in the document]
karan6181: No longer relevant. Modified the designed.

python/mxnet/gluon/estimator/estimator.py
```diff
@@ -175,14 +174,51 @@ def _batch_fn(self, batch, ctx, is_iterator=False):
...
+        -------
+        num_batches: int
+            Number of batches the data is divided into
+        total_samples: int
+            Total Number of samples
+        batch_size: int
+            Batch size
+        """
+        if isinstance(data, gluon.data.DataLoader):
+            if isinstance(data._dataset, gluon.data.ArrayDataset):
```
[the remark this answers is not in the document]
karan6181: No longer relevant. Modified the designed.

tests/python/unittest/test_gluon_estimator.py
karan6181: Yes, fixed in all the test. Thanks

python/mxnet/gluon/estimator/event_handler.py
karan6181: Removed the verbose `0` option as it makes no sense adding it. Thanks for your input

python/mxnet/gluon/estimator/event_handler.py
karan6181: As per the discussion, I have added it as a constant so the end user can directly pass that as a parameter instead of number. Could you please have a look again? Thanks!

python/mxnet/gluon/estimator/estimator.py
```diff
@@ -261,6 +254,8 @@ def fit(self, train_data,
         for epoch in range(self.max_epoch):
             # epoch begin
             self.current_epoch = epoch
+            # Number of samples trained after every batch
+            completed_samples = 0
```
nswamy: isn't this just batch_size?

python/mxnet/gluon/estimator/estimator.py
```diff
@@ -309,12 +303,14 @@ def fit(self, train_data,
                     name, value = loss_metric.get()
                     self.train_stats['train_' + name] = value
 
+                completed_samples += batch_size
+
                 self.batch_idx = i
                 # record trained samples v.s. total samples if using Gluon DataLoader
                 if isinstance(train_data, gluon.data.DataLoader):
-                    self.samples = "{}/{}".format(self.batch_size * (i + 1), len(train_data._dataset))
+                    self.samples = "{}/{}".format(completed_samples, len(train_data._dataset))
```
nswamy: processed_samples

python/mxnet/gluon/estimator/event_handler.py
```diff
@@ -85,14 +85,26 @@ class LoggingHandler(EventHandler):
...
     file_location: str
         file location to save the logs
+    verbose: int, default ONLY_EPOCH
+        Limit the granularity of metrics displayed during training process
+        verbose=ONLY_EPOCH: display metrics every epoch
+        verbose=BATCH_WITH_EPOCH: display metrics every batch and epoch
     """
 
-    def __init__(self, file_name=None, file_location=None):
+    ONLY_EPOCH = 1
```
nswamy: nit: rename
LOG_PER_EPOCH
LOG_PER_BATCH

python/mxnet/gluon/estimator/estimator.py
```diff
@@ -208,16 +203,15 @@ def evaluate(self,
     def fit(self, train_data,
             val_data=None,
             epochs=1,
-            batch_size=None,
             event_handlers=None,
             batch_fn=None):
         """Main training loop
```
nswamy: Can you add what Fit does here and also note that the batch_size is inferred from the dataloader's batch_size.

karan6181: Done. Thanks!

python/mxnet/gluon/estimator/event_handler.py
karan6181: Done. Thanks!

python/mxnet/gluon/estimator/estimator.py
karan6181: Done. Thanks!

python/mxnet/gluon/estimator/estimator.py
karan6181: This variable stores the number of processed samples which is used in LoggingHandler for printing to console.

PR #14591, opened 2019-04-02
Reference engine from chunk via weak pointer

KellenSunderland:
## Description ##
This PR fixes an issue we're seeing in TRT integration which causes a crash on shutdown.  I believe the issue should be present under other usages as well.  This PR tracks a weak reference on the engine to ensure we don't try to schedule work on the engine after it has been shutdown (for example cleaning up memory from NDChunks that are still in scope).

### Essentials ###

### DIFF
--- a/include/mxnet/ndarray.h
+++ b/include/mxnet/ndarray.h
@@ -850,15 +850,20 @@ class NDArray {
     mxnet::ShapeVector aux_shapes;
     /*! \brief Reference to the storage to ensure proper destruct order */
     std::shared_ptr<Storage> storage_ref_;
+    /*! \brief Reference to the engine to ensure we cleanup without calling a destructed engine */
+    std::weak_ptr<Engine> engine_ref_;
 
-    /*! \brief default cosntructor */
+
+    /*! \brief default constructor */
     Chunk() : static_data(true), delay_alloc(false),
-              storage_ref_(Storage::_GetSharedRef()) {}
+              storage_ref_(Storage::_GetSharedRef()),
+              engine_ref_(Engine::_GetSharedRef()) {}
 
     /*! \brief construct a new chunk */
     Chunk(mxnet::TShape shape, Context ctx_, bool delay_alloc_, int dtype)
         : static_data(false), delay_alloc(true), ctx(ctx_),
-          storage_ref_(Storage::_GetSharedRef()) {
+          storage_ref_(Storage::_GetSharedRef()),
+          engine_ref_(Engine::_GetSharedRef()) {
       storage_shape = shape;
       if (shape_is_known(storage_shape)) {
         shandle.size = shape.Size() * mshadow::mshadow_sizeof(dtype);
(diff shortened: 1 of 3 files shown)


haohuanw:
:fireworks:

piyushghai:
Thanks for your contribution @KellenSunderland . 

@mxnet-label-bot Add [Numpy, pr-awaiting-review]

KellenSunderland:
Thanks @piyushghai.  Not immediately clear to me why the numpy label is attached.  Does this PR have a Numpy related aspect to it?

piyushghai:
Aah. My bad. I was adding ndarray, and I ended up typing numpy instead :/ 

@mxnet-label-bot Update[ndarray, pr-awaiting-review]

KellenSunderland:
@piyushghai Gotcha!  Thanks for the quick responses on the PR.

haohuanw:
LGTM

KellenSunderland:
@anirudhacharya @eric-haibin-lin 

Could one of you guys do a quick review (it's a pretty small change).

KellenSunderland:
@anirudh2290 Good point.  Took me a while to think of how we can test this.  Reproducing the issue is only possible when we have static de-allocating.  A bit tricky to test without modifying the code to make non-static scoped.

I do have an example of a test crashes on my machine without the fix.  It'll mark as passed but crash the test process (which will prevent regressions in CI I hope).  Stepping through I can see it's taking the branches you'd expect w and w/o the fix.  I.e. with the fix it skips the engine delete.

```
Note: Google Test filter = Engine.stop_cleanup
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from Engine
[ RUN      ] Engine.stop_cleanup
[       OK ] Engine.stop_cleanup (1 ms)
[----------] 1 test from Engine (1 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (1 ms total)
[  PASSED  ] 1 test.

Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)

```

tests/cpp/engine/threaded_engine_test.cc
```diff
@@ -122,6 +122,15 @@ double EvaluateWorloads(const std::vector<Workload>& workloads,
...
 
+TEST(Engine, stop_cleanup) {
+    static std::unique_ptr<mxnet::NDArray> ndArray;
+    {
+        auto engine = mxnet::Engine::_GetSharedRef();
+        ndArray = std::make_unique<mxnet::NDArray>(mxnet::Context::CPU());
+        engine->Stop();
+    }
+}
+
```
anirudh2290: @KellenSunderland Thanks for the nice work on the tests. I understand this may be difficult to test, and I am okay with the crash for verification. Would it be easy to pinpoint where it crashed when run with other tests ? Otherwise better to move it to a different source file.

KellenSunderland: It would be a little difficult in fact.  Let's move it to it's own test suite to hopefully make it a little more obvious for future contributors.

KellenSunderland:
@anirudh2290 Moved the test into its own fixture, should be ready for another look.

PR #14593, opened 2019-04-02
Set _scale in Trainer using Optimizer rescale_grad

yuxihu:
Currently `_scale` is set based on the value of rescale_grad in optimizer_params when creating Trainer. However, users can use optimizer_params to create an Optimizer object and pass this object to Trainer. In this case, `_scale` may not be correctly set since optimizer_params is None.

In this PR, we set `_scale` by reading the rescale_grad value from optimizer after optimizer is initialized in Trainer.


piyushghai:
That's a good catch @yuxihu.

@mxnet-label-bot Add [Gluon, pr-awaiting-review]

yuxihu:
@eric-haibin-lin @apeforest Please help review.

Issue #13909, opened 2019-01-16
gluon.utils.split_and_load(even_split=False) fails if num of contexts > num of data


Ishitori:
## Description
Sometimes, it is hard to predict how much data left in the dataloader by the time last batch comes. When doing multigpu training with `last_batch=keep`, it could happen that the number of items left in the last batch is smaller than numbers of gpus. In that case `gluon.utils.split_and_load` throws an exception `ValueError: Too many slices for data with shape ...`.

It would be great if it could work transparently. I would expect that if the parameter `even_split` on DataLoader is set to `False`, then exception shouldn't happen: the data should be distributed in a way that some arrays are empty, and later on the calculation of forward and backward passes with empty arrays are just silently ignored.

## Environment info (Required)

```
----------Python Info----------
Version      : 3.6.4
Compiler     : GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)
Build        : ('default', 'Jan 16 2018 12:04:33')
Arch         : ('64bit', '')
------------Pip Info-----------
Version      : 18.0
Directory    : /Users/sssokolo/anaconda3/lib/python3.6/site-packages/pip
----------MXNet Info-----------
/Users/sssokolo/anaconda3/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
Version      : 1.5.0
Directory    : /Users/sssokolo/anaconda3/lib/python3.6/site-packages/mxnet
Commit Hash   : fd34dc5f847192dfd522555afdf13be1eb67b72b
----------System Info----------
Platform     : Darwin-16.7.0-x86_64-i386-64bit
system       : Darwin
node         : 8c859074eea0
release      : 16.7.0
version      : Darwin Kernel Version 16.7.0: Sun Oct 28 22:30:19 PDT 2018; root:xnu-3789.73.27~1/RELEASE_X86_64
----------Hardware Info----------
machine      : x86_64
processor    : i386
b'machdep.cpu.extfeatures: SYSCALL XD 1GBPAGE EM64T LAHF LZCNT PREFETCHW RDTSCP TSCI'
[...]

PR #14607, opened 2019-04-03
split_and_load can now handle num_ctx > num_data. Github Issue #13909

mightydeveloper:
## Description ##
Handles Issue #13909
when last batch size is smaller than the number of contexts, 
the previous utility function` gluon.utils.split_and_load` throws an exception `ValueError: Too many slices for data with shape ....`
However, we can just put one data per context and ignore the remaining contexts.
This integrates nicely with the given reproducing example code. (User does not need to modify the code)
```
for d, l in zip(data, label):
    with autograd.record():
        out = net(d)
        losses.append(loss_fn(out, l))
for loss in losses:
    loss.backward()
```
where `data` and `label` will only output data that exists in the first few of the given context.
the forward / backward pass is only done in the contexts where needed. (remaining contexts does not need to do a fake forward/backward pass)
My concern is people can make mistakes when calculating mean of the losses.

So it would be nice if we add this behavior to the documentation. (`split_and_load` can output a list that has a size less than number of contexts when even_split=False)

### Essentials ###
-> I believe this PR is just tiny change

### DIFF
--- a/python/mxnet/gluon/utils.py
+++ b/python/mxnet/gluon/utils.py
@@ -63,10 +63,6 @@ def split_data(data, num_slice, batch_axis=0, even_split=True):
         Return value is a list even if `num_slice` is 1.
     """
     size = data.shape[batch_axis]
-    if size < num_slice:
-        raise ValueError(
-            "Too many slices for data with shape %s. Arguments are " \
-            "num_slice=%d and batch_axis=%d."%(str(data.shape), num_slice, batch_axis))
     if even_split and size % num_slice != 0:
         raise ValueError(
             "data with shape %s cannot be evenly split into %d slices along axis %d. " \
@@ -75,6 +71,12 @@ def split_data(data, num_slice, batch_axis=0, even_split=True):
                 str(data.shape), num_slice, batch_axis, num_slice))
 
     step = size // num_slice
+
+    # If size < num_slice, make fewer slices
+    if not even_split and size < num_slice:
+        step = 1
+        num_slice = size
+
     if batch_axis == 0:
         slices = [data[i*step:(i+1)*step] if i < num_slice - 1 else data[i*step:size]
                   for i in range(num_slice)]


mightydeveloper:
Just an FYI, I made a misleading comment on the code previously, so I fixed it.

wkcn:
I am not sure how the weights update for the context without inpus when backward?

mightydeveloper:
@wkcn 
In the example above, `losses` only contains loss terms that are forwarded by real data samples. (meaning that unnecessary(possibly fake) context loss are not appended from previous for loop.)
So I believe when we do `loss.backward()`, only the contexts that are appended from previous for loop will run and calculate gradients for previously marked variables from previous for loop. 

For example, suppose we have 3 examples left from the dataset and have 5 GPU contexts.
We would only call `losses.append(loss_fn(out, l))` 3 times, marking variables in 0, 1, 2 GPU contexts.
So when we call 
```
for loss in losses:
    loss.backward()
```
only gradients for GPU 0, 1, 2 will be calculated and when we eventually call `trainer.step()`, the weights will be updated.
Does this answer help you? (I might have misunderstood your concern)

+ Also, I noticed that I have an indentation error in my original PR request description so I just edited it.

wkcn:
@mightydeveloper 
Thank you for the detailed explanation!
Yes, only gradients for GPU 0, 1, 2 will be calculated.

However, it seems that `trainer.step()` will update the weights by using all gradients on 5 GPU contexts.
The gradients on GPU 3, 4 may be not zero-tensors.

mightydeveloper:
> However, it seems that `trainer.step()` will update the weights by using all gradients on 5 GPU contexts.
> The gradients on GPU 3, 4 may be not zero-tensors.

Thanks for checking and providing relevant code link!

So... If I want to zero out gradients, I guess I should either
1. call `trainer._params.zero_grad()` 
2. or  just `net.collect_params().zero_grad()` 
(Assuming that I initialized `trainer` with `Trainer(net.collect_params(), ...`)
3. or maybe call `trainer.step(ignore_stale_grad=True)` instead? 

Would all three of them work? which one would be better?

wkcn:
calling trainer.step(ignore_stale_grad=True) is okay : )

piyushghai:
@mightydeveloper Thanks for your contributions. 

@mxnet-label-bot Add [Gluon]

wkcn:

PR #12821, opened 2018-10-14
[MXNET-1095] [WIP] Support higher order gradient for cos,sin,negative,mul

sxjscience:
## Description ##
For context, please refer https://github.com/apache/incubator-mxnet/issues/10002

Support higher order gradient for cos, sin, negative, mul. I'll add the test later. For now, you may use the following code to test the correctness:

```python
import mxnet.ndarray as nd
from mxnet import autograd
from mxnet.test_utils import assert_almost_equal

x = nd.array([3.0])
x.attach_grad()
with autograd.record():
    y = nd.sin(x)
    y_grad = autograd.grad(y, x, create_graph=True, retain_graph=True)[0]
    z = y_grad * y_grad # Should be cos(x)^2
z.backward()
print(x.grad) # Should be - 2 * cos(x) * sin(x)
gt_grad = - 2 * nd.cos(x) * nd.sin(x)
assert_almost_equal(x.grad.asnumpy(), gt_grad.asnumpy())
```

```
[0.2794155]
<NDArray 1 @cpu(0)>
```

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] higher-order gradient of sin,cos,negative,mul, tests

## Comments ##
- Need to support more OPs.

PR #14095, opened 2019-02-08
Simplify creation of NodeEntry instances and use emplace_back

larroy:
## Description ##

Reduce copying of shared_ptr. In a microbenchmark is 10% faster. But the main motivation is simplifying creation of nnvm graph on operators.

    Optimize move semantics of NodeEntry
    https://github.com/dmlc/tvm/pull/2576
    Making copies of shared_ptr is more expensive than moving.
    This PR reduces lock contention by using move semantics in NNVM nodes
    making also more convenient to construct NodeEntry classes in the code
    due to the added ctors.
    
    Update NDarray with NodeEntry constructors and refine initializer lists.
    
    Sync gradient.cc with tvm

imperative.cc will be addressed in a different refactor.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

PR #14613, opened 2019-04-04
[MXNET-978] Second order gradient support for some unary operators

apeforest:
## Description ##
This PR is to support higher order gradient in some basic unary operators. Thanks to @sxjscience, this PR is a continued work of https://github.com/apache/incubator-mxnet/pull/12821.

This PR adds additional support for relu operator and fixed the issue of negative operator support in https://github.com/apache/incubator-mxnet/pull/12821. In addition, I added unit tests for to test higher order gradient.
### Essentials ###

### Changes ###
- [X] higher order gradient for a set of unary operators
- [X] unit test to test higher order gradient


src/imperative/imperative.cc
```diff
@@ -347,8 +347,9 @@ std::vector<NDArray*> Imperative::Backward(
       x_reqs.push_back(info.grad_req);
       info.fresh_out_grad = true;
     }
-    CHECK_GT(xs.size(), 0)
-        << "There are no inputs in computation graph that require gradients.";
+    if (xs.empty()) {
```
larroy: Why change from CHECK to warning?

apeforest: Current backward operation requires an operator must have at least one inputs, because the gradient of a constants is always zero. However, the second order of some operators such as relu is actually gradient of a constant (ones or zeros). Therefore we need to support gradient for constant operators.

piyushghai:
Thanks for your contributions @apeforest. Can you also look into the CI failures ? 

@mxnet-label-bot Add [Operator, pr-awaiting-review, Backend].

apeforest:
@sxjscience Please help to review

sxjscience:
Looks good. However, the test has not passed and we need to check what caused the error. Also, we should later add a testing utility function that checks the higher gradient by numerical differentiation.

Roshrini:
@apeforest Thanks for working on this. Can you look into build failures?

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -0,0 +1,89 @@
...
+import numpy as np
+from mxnet import gluon, nd, autograd
+from mxnet.test_utils import assert_almost_equal
+from tests.python.unittest.common import with_seed
+
+
+@with_seed()
+def test_elemwise_mul():
+    x = nd.array([1, 2, 3])
+    y = nd.zeros(3)
```
larroy: Do we need this y?

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -0,0 +1,89 @@
...
+    expect_grad = -nd.sin(x)
+    check_second_order_unary(x, sin, expect_grad)
+
+
+@with_seed()
+def test_cos():
+    def cos(x):
+        return nd.cos(x)
+
+    x = nd.array([1, 2, 3])
```
anirudhacharya: can we randomize the test arrays with [random_arrays](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/test_utils.py#L84) and [rand_shape_2d](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/test_utils.py#L402)

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -0,0 +1,89 @@
...
+    expect_grad = -nd.cos(x)
+    check_second_order_unary(x, cos, expect_grad)
+
+
+@with_seed()
+def test_negative():
+    def negative(x):
+        return nd.negative(x)
+
+    x = nd.array([1, 2, 3])
```
anirudhacharya: same as above and for rest of the tests

vandanavk:
@apeforest could you have a look at the CI failures

src/imperative/imperative.cc
larroy: I think we should dive deeper into this one.  Does it produce the warning (or early the failure) for some of the test cases?  

In the original code I think the intention is to get if there's any input nodes which have gradient attached, I understand your explanation but what I don't see is where would we store the gradient for such constants, is because grad_req of the constant is kNullOp?  the constant is just another node right?

tests/python/unittest/test_higher_order_grad.py
larroy: I think for second order not using random inputs helps reason about the gradient result...

src/operator/tensor/elemwise_binary_op_basic.cc
```diff
@@ -233,7 +233,17 @@ The storage type of ``elemwise_mul`` output depends on storage types of inputs
...
 .add_alias("_mul").add_alias("_Mul")
-.set_attr<nnvm::FGradient>("FGradient", ElemwiseGradUseIn{"_backward_mul"});
+.set_attr<nnvm::FGradient>("FGradient",
+  [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
+    auto lhs_grad = MakeNode("elemwise_mul", n->attrs.name + "_backward_lhs",
+                             {ograds[0], n->inputs[1]}, nullptr, &n);
+    auto rhs_grad = MakeNode("elemwise_mul", n->attrs.name + "_backward_rhs",
+                             {ograds[0], n->inputs[0]}, nullptr, &n);
+    std::vector<nnvm::NodeEntry> ret;
+    ret.emplace_back(nnvm::NodeEntry{lhs_grad, 0, 0});
```
larroy: Can we simplify as per https://github.com/apache/incubator-mxnet/pull/14095 ?

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -83,7 +83,18 @@ The storage type of ``relu`` output depends upon the input storage type:
...
+    auto zero_node = MakeNode("zeros_like", n->attrs.name + "_backward",
+      {n->inputs[0]}, nullptr, &n);
+    auto x_grad = MakeNode("_greater", n->attrs.name + "_mid_x_grad",
+      {n->inputs[0], nnvm::NodeEntry{zero_node, 0, 0}}, nullptr, &n);
+    auto in_grad = MakeNode("elemwise_mul", n->attrs.name + "_backward",
+      {ograds[0], nnvm::NodeEntry{x_grad, 0 , 0}}, nullptr, &n);
+    std::vector<nnvm::NodeEntry> ret;
+    ret.emplace_back(nnvm::NodeEntry{in_grad, 0, 0});
+    return ret;
+  });
```
larroy: We should measure if this causes regressions as we discussed, otherwise we should add FGradient to _backward_relu

src/imperative/imperative.cc
apeforest: The rootcause is when we do second order gradient of the negative(x) operator. The backward graph of this does not require any input and therefore will trigger this condition. I think if I remove the test for negative(x) then we do not need to modify this.

apeforest:
@sxjscience After discussing with @kshitij12345, we think it will cause less impact to exist first order gradient by adding a `FGradient` attribute to the backward operator. Please see https://github.com/apache/incubator-mxnet/pull/14992

Therefore I also refactored change in this PR. Now it passed all CI unit tests. Please review again. Thanks!

apeforest:
@anirudhacharya added random_arrays as suggested. Please review again. Thanks

apeforest:
@kshitij12345 Please help to review.

kshitij12345:
@apeforest Thank You. Only thing that bothered me was inconsistent use of Value Initialisation and List Initialisation. Should prefer List Initialisation
https://stackoverflow.com/questions/18222926/why-is-list-initialization-using-curly-braces-better-than-the-alternatives

Otherwise LGTM.

src/operator/tensor/elemwise_unary_op_trig.cc
```diff
@@ -63,7 +82,27 @@ The storage type of ``cos`` output is always dense
...
+                                      {n->inputs[1]}, nullptr, &n);
+      auto grad_grad_x = MakeNode("negative", n->attrs.name + "_backward_grad_grad",
+                                  {nnvm::NodeEntry(grad_grad_x_mid)}, nullptr, &n);
+      std::vector<nnvm::NodeEntry> ret;
+      // for the backward of the _backward_cos node
+      // first input is the ograd and second input is x (because ElemwiseUseIn)
+      ret.emplace_back(MakeNode("elemwise_mul", n->attrs.name + "_backward_grad_grad",
+                                {ograds[0], grad_x}, nullptr, &n));
+      ret.emplace_back(MakeNode("elemwise_mul", n->attrs.name + "_backward_grad_grad_in",
+                                {ograds[0], nnvm::NodeEntry(grad_grad_x)}, nullptr, &n));
```
larroy: you can move the shared_ptrs inside as you are not using them anymore.

apeforest: isn't NodeEntry constructor already doing the std::move? I think having this variable outside increase readability.

apeforest: Isn't the NodeEntry constructor already doing the std::move of shared_ptr? I thought declaring a separate variable in this scope increases readability.

apeforest:
@kshitij12345 Change to list initialization as suggested. Thanks

src/operator/tensor/elemwise_unary_op_trig.cc
larroy: The argument is by value, so it needs to be an rvalue if you want it moved two times, otherwise is copied, then moved. So is best:


      ret.emplace_back(MakeNode("elemwise_mul", n->attrs.name + "_backward_grad_grad_in",
                                {ograds[0], nnvm::NodeEntry(std::move(grad_grad_x))}, nullptr, &n));


Having a shared ptr by value then move into a class seem to be the accepted idiom nowadays. Since it allows both move and copy semantics, depending if the caller passes an rvalue or not.

apeforest:
@kshitij12345 @sxjscience @anirudhacharya  friendly ping for final review please:)

src/operator/tensor/elemwise_unary_op_trig.cc
```diff
@@ -46,7 +46,26 @@ The storage type of ``sin`` output depends upon the input storage type:
...
+.set_attr<nnvm::FGradient>("FGradient",
+    [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
+      // f(x) = sin(x)
+      // f'(x) = cos(x)
+      // f''(x) = -sin(x)
+      auto grad_x = nnvm::NodeEntry{n};
+      auto grad_grad_x_mid = MakeNode("sin", n->attrs.name + "_mid_grad_grad",
+                                      {n->inputs[1]}, nullptr, &n);
+      auto grad_grad_x = MakeNode("negative", n->attrs.name + "_backward_grad_grad",
+                                  {nnvm::NodeEntry(grad_grad_x_mid)}, nullptr, &n);
```
kshitij12345: List initialisation. Otherwise LGTM.

src/operator/tensor/elemwise_unary_op_trig.cc
```diff
@@ -63,7 +82,27 @@ The storage type of ``cos`` output is always dense
...
+.set_attr<nnvm::FGradient>("FGradient",
+    [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
+      // f(x) = cos(x)
+      // f'(x) = -sin(x)
+      // f''(x) = -cos(x)
+      auto grad_x = nnvm::NodeEntry{n};
+      auto grad_grad_x_mid = MakeNode("cos", n->attrs.name + "_mid_grad_grad",
+                                      {n->inputs[1]}, nullptr, &n);
+      auto grad_grad_x = MakeNode("negative", n->attrs.name + "_backward_grad_grad",
+                                  {nnvm::NodeEntry(grad_grad_x_mid)}, nullptr, &n);
```
kshitij12345: Ditto.

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -15,13 +15,52 @@
...
+def test_sin():
+    def sin(x):
+        return nd.sin(x)
+
+    def grad_grad_op(x):
+        return -nd.sin(x)
+
+    arrays = random_arrays((2, 2), (2, 3), (4, 5, 2), (3, 1, 4, 5))
+    for array in arrays:
+        check_second_order_unary(array, sin, grad_grad_op)
```
anirudhacharya: I think `check_second_order_unary` function should be moved to `python/mxnet/test_utils.py`

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -15,13 +15,52 @@
...
+
+@with_seed()
+def test_relu():
+    def relu(x):
+        return nd.relu(x)
+
+    def grad_grad_op(x):
+        return nd.zeros_like(x)
+
+    arrays = random_arrays((2, 2), (2, 3), (4, 5, 2), (3, 1, 4, 5))
```
anirudhacharya: shouldn't we test for 1-d arrays?

also there is this to randomize the shape of an array - https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/test_utils.py#L418

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -85,8 +85,20 @@ The storage type of ``relu`` output depends upon the input storage type:
 )code" ADD_FILELINE)
 .set_attr<nnvm::FGradient>("FGradient", ElemwiseGradUseOut{"_backward_relu"});
 
-MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU(_backward_relu,
-                                               unary_bwd<mshadow_op::relu_grad>);
+MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU(_backward_relu, unary_bwd<mshadow_op::relu_grad>)
+.set_attr<nnvm::FGradient>("FGradient",
+    [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
+      std::vector<nnvm::NodeEntry> ret;
+      // f(x) -> f = relu
```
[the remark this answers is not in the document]
apeforest: This is a hidden operator so user do not see this.

src/operator/tensor/elemwise_unary_op_trig.cc
```diff
@@ -46,7 +46,26 @@ The storage type of ``sin`` output depends upon the input storage type:
 )code" ADD_FILELINE)
 .set_attr<nnvm::FGradient>("FGradient", ElemwiseGradUseIn{ "_backward_sin" });
 
-MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_sin, unary_bwd<mshadow_op::sin_grad>);
+MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_sin, unary_bwd<mshadow_op::sin_grad>)
+.set_attr<nnvm::FGradient>("FGradient",
+    [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
+      // f(x) = sin(x)
```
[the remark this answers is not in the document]
apeforest: This is a hidden operator so user do not see this.

src/operator/tensor/elemwise_unary_op_trig.cc
apeforest: updated.

src/operator/tensor/elemwise_unary_op_trig.cc
apeforest: updated

tests/python/unittest/test_higher_order_grad.py
apeforest: It's only used in this test. If we add a different test file then it makes sense as you suggested.

tests/python/unittest/test_higher_order_grad.py
apeforest: thanks for the suggestion. updated.

apeforest:
@kshitij12345 gentle ping again :) please approve it there is no other concern. thanks

kshitij12345:
@apeforest As mentioned in the #14992 for `log` , I guess the check will fail with given script.
```Python
import numpy
import math

x.attach_grad()
with autograd.record():
  y = nd.log(x)
  y_grad = autograd.grad(y, x, head_grads= nd.ones_like(y) * 0.5, create_graph=True, retain_graph=True)[0]
y_grad.backward(nd.ones_like(y_grad) * 0.6)

numpy.testing.assert_allclose(x.grad.asnumpy() , ( grad_grad_op(x) * 0.5 * 0.6).asnumpy(), rtol=1e-7, atol=1e-7)
```

As the `grad` from upper layer is not preserved for `sin` and `cos`.

apeforest:
@kshitij12345 Thanks for catching this bug. I have updated my sin and cos implementation with comments to help better understanding the mathematics behind second order gradient calculation. Please help to review again.

apeforest:
@kshitij12345 could you please take a look at the PR again. Thanks

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -85,8 +85,23 @@ The storage type of ``relu`` output depends upon the input storage type:
...
+      std::vector<nnvm::NodeEntry> ret;
+      // ograds[0]: d^2L/dx^2
+      // inputs[0]: dL/dy
+      // inputs[1]: y
+      // f(x) -> relu(x)
+      // f'(x) = 1 if x > 0 else 0
+      // f''(x) = 0
+      auto gx = nnvm::NodeEntry{n};  // f'(x)
+      ret.emplace_back(MakeNode("elemwise_mul", n->attrs.name + "_backward_grad_grad",
+                                {ograds[0], gx}, nullptr, &n));
```
kshitij12345: `gx` is actually `f'(x) * head_grads/output_gradient`.  
It should actually only be `f'(x)`.

Explanation :  gx = `f'(x) * head_grads`.
Therefore, `gx w.r.t. f'(x) = head_grads`,
Similarly `gx w.r.t. head_grads = f'(x)`.

kshitij12345:
@apeforest

Could you update the `check_second_order_unary` as per

https://github.com/apache/incubator-mxnet/blob/37ce3b87268a8154f5c0ad97ce2522478038ee06/tests/python/unittest/test_higher_order_grad.py#L76-L103

This covers check for gradient of the first input argument as well. Have tested a similar Pytorch Script which works ( code in PR #15120 ).

However do note that for PR #15120 , 
https://github.com/apache/incubator-mxnet/blob/37ce3b87268a8154f5c0ad97ce2522478038ee06/tests/python/unittest/test_higher_order_grad.py#L102

Assertion fails with `head_grads.grad.asnumpy()` being all `0's`. 

Please check to see if it works for you.
Thank You.

apeforest:
@kshitij12345 I updated the test locally and it also fails with sin(x) and cos(x) as well with the same reason of log(x). I have not pushed it to the remote because it then requires your PR for log(x) merged first. Otherwise the test will always fail. Still trying to understand what's the meaning of head_grads.grad here.

apeforest:
Ping @aidan-plenert-macdonald for review

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -15,13 +15,55 @@
...
+    def cos(x):
+        return nd.cos(x)
+
+    def grad_grad_op(x):
+        return -nd.cos(x)
+
+    for dim in range(1, 5):
+        shape = rand_shape_nd(dim)
+        array = random_arrays(shape)
+        check_second_order_unary(array, cos, grad_grad_op)
```
aidan-plenert-macdonald: Can these `check_second_order_unary` checks be changes to Nth order?

src/operator/tensor/elemwise_unary_op_trig.cc
```diff
@@ -46,7 +46,33 @@ The storage type of ``sin`` output depends upon the input storage type:
...
+    [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
+      // ograds[0]: d^2L/dx^2
+      // inputs[0]: dL/dy
+      // inputs[1]: x (ElemwiseUseIn)
+      // f(x) = sin(x)
+      // f'(x) = cos(x)
+      // f''(x) = -sin(x)
+      auto x_grad = MakeNode("cos", n->attrs.name + "_x_grad",
+                             {n->inputs[1]}, nullptr, &n);
+      auto x_grad_grad = MakeNode("negative", n->attrs.name + "_x_grad_grad",
```
aidan-plenert-macdonald: Although it appears that this permit's Nth order derivatives, the `_grad_grad` thing concerns me as though this is simple registering a 2nd order grad and calling that good. Assuming this does support Nth order, it worries me that someone else trying to copy this would "cheat" by only having 2 gradients.

Ideally the testing would catch this.

larroy: as long as the nodes are themselves differentiable, then it would support additional differentiation. For some complex functions having just 2nd gradient should be good. In this case I would say it's N times differentiable. the _x_grad_grad is just the name of the node as far as I understand it. Why does it concern you?

src/operator/tensor/elemwise_unary_op_basic.cc
apeforest: Updated

src/operator/tensor/elemwise_unary_op_trig.cc
apeforest: I renamed it. Also this indicates the gradient is on top of previous gradient. User can still apply this recursively for nth order gradient.

tests/python/unittest/test_higher_order_grad.py
apeforest: This PR is only to support second order gradient. I can add a test later.

apeforest:
To summarize from discussion in https://github.com/apache/incubator-mxnet/pull/15120, I think in MXNet we calculate input gradients based on the variables specified in the autograd.grad() API. In this case y_grad, even if someone manually attach a grad attribute to it will not be assigned gradient values during the second pass of `backward()`. Only the second order gradient x.grad is preserved. This behavior is different from PyTorch, but I think it is not a bug of MXNet.

kshitij12345:
Thank you very much for the clarification. 

>  I think in MXNet we calculate input gradients based on the variables specified in the autograd.grad() API. In this case y_grad, even if a grad attribute is attached to it, no gradient values will be assigned during the second pass of backward(). Only the second order gradient x.grad is assigned. 

Btw, I was wondering if above should be documented? Mostly probably users won't end up in this situation but still. Your thoughts?

apeforest:
> BTW, I was wondering if above should be documented? Mostly probably users won't end up in this situation but still. Your thoughts?

Yes, I think we should document this in the autograd API. What do you think?

apeforest:
@kshitij12345 Do you think we can merge this PR now? If so, could you please approve it? thanks!

kshitij12345:
LGTM

PR #14615, opened 2019-04-04
Add MXEnginePushAsync and MXEnginePushSync C APIs

yuxihu:
std::function definition is not compatible between GCC [4.x](https://github.com/gcc-mirror/gcc/blob/gcc-4_8_4-release/libstdc++-v3/include/std/functional#L2069) and GCC [5.x](https://github.com/gcc-mirror/gcc/blob/gcc-5_4_0-release/libstdc++-v3/include/std/functional#L1854)+. The incompatibility caused [segmentation fault](https://github.com/horovod/horovod/issues/884) when a third-party library compiled with GCC 5+ uses the PushAsync/PushSync interfaces in engine (MXNet pip package is compiled using GCC 4).

To avoid the above incompatibility issue, this PR addes MXEnginePushAsync/MXEnginePushSync C APIs. These two APIs take function pointer instead of std::function as the engine execution function.


yuxihu:
@eric-haibin-lin @szha @apeforest @ctcyang Please help review

include/mxnet/engine.h
```diff
@@ -120,6 +120,10 @@ class MXNET_API Engine {
   typedef std::function<void(RunContext)> SyncFn;
   /*! \brief Asynchronous operation to pass to engine. */
   typedef std::function<void(RunContext, CallbackOnComplete)> AsyncFn;
+  /*! \brief Synchronous operation (function pointer) to pass to engine. */
+  typedef void (*SyncFnPtr)(RunContext, const std::shared_ptr<void>&);
```
eric-haibin-lin: why const reference for the shared_ptr?

include/mxnet/engine.h
```diff
@@ -206,6 +210,43 @@ class MXNET_API Engine {
...
+   * \param exec_ctx Execution context.
+   * \param const_vars The variables that current operation will use but not
+   *                   mutate.
+   * \param mutable_vars The variables that current operation will mutate.
+   * \param prop Property of the function.
+   * \param priority Priority of the action, as hint to the engine.
+   * \param opr_name The operator name.
+   * \param wait Whether this is a WaitForVar operation.
+   */
+  void PushAsyncPtr(AsyncFnPtr exec_fn_ptr, const std::shared_ptr<void>& param,
```
eric-haibin-lin: why const reference for the shared_ptr?

eric-haibin-lin:
Could you share what you tested to validate the change?

yuxihu:
> Could you share what you tested to validate the change?

Here is the [PR](https://github.com/horovod/horovod/pull/985) in Horovod that adopts the new API. I built MXNet with GCC 4.8.4 and Horovod with GCC 5.4.0. I ran the unit tests, MNIST example, ResNet50 example in Horovod and there are no segmentation fault.

include/mxnet/engine.h
yuxihu: I wanted to use reference to avoid unnecessary copy of shared_ptr which is an expensive operation. I did some simple experiment and observed that passing shared_ptr by value can be 20x slower compared to passing by reference. In addition, in the implementation of PushAsyncPtr/PushSyncPtr functions, when we create a [lambda function](https://github.com/apache/incubator-mxnet/pull/14615/files#diff-86dd2a82be9dffd682cbdd5c13490ac9R76), we captures the param, which is a copy operation. This is why I decide to use reference here. Then using const to prevent it from being modified.

include/mxnet/engine.h
yuxihu: See above

wkcn:
It is better to use the pure C type in the parameters list.
Because different compilers (such as gcc, clang, msvc) may have different implementations of STL containers (std::vector, std::shared_ptr, etc.), it will cause the ABI compatible problem too.

yuxihu:
> It is better to use the pure C type in the parameters list.
> Because different compilers (such as gcc, clang, msvc) may have different implementations of STL containers (std::vector, std::shared_ptr, etc.), it will cause the ABI compatible problem too.

This is a good point. But we may not be able to use the pure C type here. We are pushing execution into another thread and the lifetime of the params need to be managed properly. shared_ptr is a straightforward choice. I am not aware of any ABI compatibility issue for shared_ptr as of now. Please suggest if you find any.

piyushghai:
Thanks for your contributions @yuxihu. 
@mxnet-label-bot Add [Backend]

src/engine/engine.cc
```diff
@@ -67,4 +67,26 @@ Engine* Engine::Get() {
   static Engine *inst = _GetSharedRef().get();
   return inst;
 }
+
+void Engine::PushAsyncPtr(AsyncFnPtr exec_fn_ptr, const std::shared_ptr<void>& param,
+                          Context exec_ctx, std::vector<VarHandle> const& const_vars,
+                          std::vector<VarHandle> const& mutable_vars,
+                          FnProperty prop, int priority,
+                          const char* opr_name, bool wait) {
+  auto exec_fn = [exec_fn_ptr, param](RunContext rctx,
```
anirudh2290: can we do  a by ref capture here.

yuxihu: I think we cannot do a ref capture here. This is where we want to do a copy of the shared_ptr which increases its ref count. When the exec_fn_ptr is called in another thread, we need to make sure param is still valid

wkcn:
@yuxihu Could we add a callback function, which will be called to release the resource when the execution is finished?

yuxihu:
> @yuxihu Could we add a callback function, which will be called to release the resource when the execution is finished?
> There is an example in TVM project: https://github.com/dmlc/tvm/blob/master/src/runtime/c_runtime_api.cc#L443

This sounds like a better option. Let me try it.

yuxihu:
@wkcn Changed based on your suggestion. It worked. Please review again.

include/mxnet/engine.h
```diff
@@ -206,6 +212,45 @@ class MXNET_API Engine {
...
+   * \param const_vars The variables that current operation will use but not
+   *                   mutate.
+   * \param mutable_vars The variables that current operation will mutate.
+   * \param prop Property of the function.
+   * \param priority Priority of the action, as hint to the engine.
+   * \param opr_name The operator name.
+   */
+  void PushSyncPtr(SyncFnPtr exec_fn_ptr, void* param, FnPtrParamDeleter del,
+                   Context exec_ctx, std::vector<VarHandle> const& const_vars,
+                   std::vector<VarHandle> const& mutable_vars,
```
Vikas89: mutable and const ?

include/mxnet/engine.h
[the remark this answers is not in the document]
yuxihu: Got rid of vector and shared_ptr from the new API interfaces.

szha:
@junrushao1994 would you mind take a look at this PR?

include/mxnet/engine.h
yuxihu: The parameters have been changed.

yuxihu:
> Is there a plan to revisit the direct call of MXNet engine from horovod. I am not very familiar with horovod but when I look at horovod tensorflow code it registers an op for broadcast and allreduce. Why couldn't we do this for MXNet also ? Horovod directly plugging into MXNet engine code means that we have to provide backward compatibility guarantees like C API or frontend API, which didnt apply to MXNet backend earlier and can cause maintenance overhead.

There is no plan to rewrite the Horovod-MXNet integration code.

FYI, Horovod-Tensorflow integration recently encountered the same GCC compatibility issue on tensorflow nightly build. The new APIs added here is not specific to Horovod use case. It can also be used by other third-party libraries.

anirudh2290:
@yuxihu these APIs weren't supposed to be directly called by third-party libraries AFAIK. MXNet provides no guarantees for backward compatibility or semver for these APIs in the backend. I think we are changing this now with horovod. Better would be to register the op and invoke op on horovod side or for other 3rd party libs for that matter.

junrushao1994:
Private APIs in MXNet Engine are not designed in the way that they interact with external libraries, but C APIs are. I am not super familiar with Horovod side, but is it possible to do the integration through C API instead of private API?

yuxihu:
> @yuxihu these APIs weren't supposed to be directly called by third-party libraries AFAIK. MXNet provides no guarantees for backward compatibility or semver for these APIs in the backend. I think we are changing this now with horovod. Better would be to register the op and invoke op on horovod side or for other 3rd party libs for that matter.

I am not familiar with how custom op is implemented in MXNet so I am not sure if the custom op approach fits into Horovod use cases at this moment. Even if it fits, it is a completely different design and will require significant amount of change. Solving this GCC incompatibility issue is high priority for our users. To me, evaluating and potentially using custom op is a longer term plan which we can discuss more offline.

junrushao1994:
@yuxihu If a third-party library would call MXNet, it should go through C API. A more correct way to do this is to add a C API pushing stuff to the engine.

yuxihu:
> @yuxihu If a third-party library would call MXNet, it should go through C API. A more correct way to do this is to add a C API pushing stuff to the engine.

I will see if it is possible to convert these APIs to C APIs.

yuxihu:
> Private APIs in MXNet Engine are not designed in the way that they interact with external libraries, but C APIs are. I am not super familiar with Horovod side, but is it possible to do the integration through C API instead of private API?



> Private APIs in MXNet Engine are not designed in the way that they interact with external libraries, but C APIs are. I am not super familiar with Horovod side, but is it possible to do the integration through C API instead of private API?

Horovod already calls PushAsync these days. I was not involved in the original design of integrating Horovod and MXNet. I am not sure how the decision was made. This PR just replace PushAsync with PushAsyncPtr to avoid GCC incompatibility issue.

junrushao1994:
> Horovod already calls PushAsync these days. I was not involved in the original design of integrating Horovod and MXNet so I do not know how the decision was made.

So it is our opportunity to make it more correct, right?

> This PR just replaces PushAsync with PushAsyncPtr to avoid GCC incompatibility issue.

As you mentioned, it only avoids incompatibility issue in a certain compiler (GCC 4.x and 5.x) in some test environments (some Linux distribution?), which seems ad-hoc, right? Adopting MXNet's more principled approach would be better for future maintenance.

To make it clear, I would say it could be simple to define `PushAsyncPtr` as a C API, which internally uses `Engine::Get()->PushXXX`. This should be simple sed/awk refactoring.

yuxihu:
> > Horovod already calls PushAsync these days. I was not involved in the original design of integrating Horovod and MXNet so I do not know how the decision was made.
> 
> So it is our opportunity to make it more correct, right?

I agree with you C API is cleaner way of doing it.

> > This PR just replaces PushAsync with PushAsyncPtr to avoid GCC incompatibility issue.
> 
> As you mentioned, it only avoids incompatibility issue in a certain compiler (GCC 4.x and 5.x) in some test environments (some Linux distribution?), which seems ad-hoc, right? Adopting MXNet's more principled approach would be better for future maintenance.

It is not ad-hoc. Our MXNet pip packages are built with GCC 4.8.5. Most of recent OSes like Ubuntu 16.04 are GCC 5+. This is big usability problem for our users. The compatibility is not only between 4 and 5. It should be between 4 and 5+.

> To make it clear, I would say it could be simple to define `PushAsyncPtr` as a C API, which internally uses `Engine::Get()->PushXXX`. This should be simple sed/awk refactoring.

I got your point. But this also means we may need to expose data structures defined in Engine at C API level. Let's try it out.

anirudh2290:
I wasn't suggesting a custom op but a standard nnvm op like mxnet backend ops. This can then be invoked using MXImperativeInvokeOpEx. Since this may require considerable change I am also okay with adding C API for engine invoke directly ( though I am not aware of historical reason as to why it was not added yet. ). Also, depending on the complexity I am okay with moving forward with the current solution (to fix user issue) and follow up PRs on MXNet and horovod to fix this.

yuxihu:
> I wasn't suggesting a custom op but a standard nnvm op like mxnet backend ops. This can then be invoked using MXImperativeInvokeOpEx. Since this may require considerable change I am also okay with adding C API for engine invoke directly ( though I am not aware of historical reason as to why it was not added yet. ). Also, depending on the complexity I am okay with moving forward with the current solution (to fix user issue) and follow up PRs on MXNet and horovod to fix this.

Changed to C APIs. Let's discuss the op approach you suggested offline.

yuxihu:
@junrushao1994 Changed to C API. Please help review.

include/mxnet/c_api.h
```diff
@@ -2541,6 +2553,51 @@ MXNET_DLL int MXNDArrayGetSharedMemHandle(NDArrayHandle handle, int* shared_pid,
...
+  * \param prop_handle Property of the function.
+  * \param priority Priority of the action, as hint to the engine.
+  * \param opr_name The operation name.
+  */
+MXNET_DLL int MXEnginePushSync(EngineSyncFunc sync_func, void* func_param,
+                               EngineFuncParamDeleter deleter, ContextHandle ctx_handle,
+                               EngineVarHandle const_vars_handle, size_t num_const_vars,
+                               EngineVarHandle mutable_vars_handle, size_t num_mutable_vars,
+                               EngineFnPropertyHandle prop_handle, int priority,
+                               const char* opr_name, bool wait);
```
anirudh2290: wait is only used by wait_to_read, can we set default to false. Also, can we have default values for priority, deleter, prop_handle.

include/mxnet/c_api.h
```diff
@@ -2541,6 +2553,51 @@ MXNET_DLL int MXNDArrayGetSharedMemHandle(NDArrayHandle handle, int* shared_pid,
...
+                                EngineVarHandle const_vars_handle, size_t num_const_vars,
+                                EngineVarHandle mutable_vars_handle, size_t num_mutable_vars,
+                                EngineFnPropertyHandle prop_handle, int priority,
+                                const char* opr_name, bool wait);
+
+/*!
+  * \brief Push a synchronous operation to the engine.
+  * \param sync_func Execution function that executes the operation.
+  * \param func_param The parameter set on calling sync_func, can be NULL.
+  * \param deleter The callback to free func_param, can be NULL.
```
anirudh2290: Is this being used in horovod ?

yuxihu: Yes. It is needed to manage the lifetime of func_param in multi-threads environment.

include/mxnet/c_api.h
yuxihu: will add defaults.

src/c_api/c_api.cc
```diff
@@ -1401,3 +1401,78 @@ int MXNDArrayCreateFromSharedMem(int shared_pid, int shared_id, const mx_uint *s
...
+  auto mutable_vars = static_cast<VarHandle*>(mutable_vars_handle);
+
+  Engine::AsyncFn exec_fn;
+  if (deleter == nullptr) {
+    exec_fn = [async_func, func_param](RunContext rctx,
+                                       CallbackOnComplete on_complete) {
+      async_func(&rctx, &on_complete, func_param);
+    };
+  } else {
+    // Wrap func_param in a shared_ptr with del as deleter such that delelter
```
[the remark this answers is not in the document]
yuxihu: Done

yuxihu:
> Please add a test which pushes an op using this C API, to ensure we don't have regressions in the future.

Addressed your comments and added tests. Please review and merge if it looks good to you. Thanks.

yuxihu:
@mxnet-label-bot update [Backend, pr-awaiting-merge]

include/mxnet/c_api.h
```diff
@@ -2541,6 +2553,51 @@ MXNET_DLL int MXNDArrayGetSharedMemHandle(NDArrayHandle handle, int* shared_pid,
...
+  * \param num_const_vars The number of const_vars.
+  * \param mutable_vars_handle The variables that current operation will mutate.
+  * \param num_mutable_vars The number of mutable_vars.
+  * \param prop_handle Property of the function.
+  * \param priority Priority of the action, as hint to the engine.
+  * \param opr_name The operation name.
+  */
+MXNET_DLL int MXEnginePushSync(EngineSyncFunc sync_func, void* func_param,
+                               EngineFuncParamDeleter deleter, ContextHandle ctx_handle,
+                               EngineVarHandle const_vars_handle, size_t num_const_vars,
```
eric-haibin-lin: use signed integers whenever possible.

src/c_api/c_api.cc
```diff
@@ -1401,3 +1401,84 @@ int MXNDArrayCreateFromSharedMem(int shared_pid, int shared_id, const mx_uint *s
   *out = new NDArray(shared_pid, shared_id, mxnet::TShape(shape, shape + ndim), dtype);
   API_END();
 }
+
+typedef Engine::VarHandle VarHandle;
+typedef Engine::CallbackOnComplete CallbackOnComplete;
+
+int MXEnginePushAsync(EngineAsyncFunc async_func, void* func_param,
+                      EngineFuncParamDeleter deleter, ContextHandle ctx_handle,
+                      EngineVarHandle const_vars_handle, size_t num_const_vars,
```
eric-haibin-lin: same here: signed

include/mxnet/c_api.h
wkcn: size_t is the type of vector<T>::size(), although the size maybe not overflow in signed integer.

yuxihu: Since we denotes number of the vars here, would 0 be sufficient for special cases? Giving -1 to the number of vars does not make much sense. In addition, it may be error prone as we use it along with the var_handle to create a vector in this function. If users unintentionally give wrong negative values, it can potentially corrupt the memory. I prefer to use unsigned integer here.

eric-haibin-lin: I'm ok with `size_t` here because it is a standard type. Regarding negative values, I'll quote from google c++ style: 

> You should not use the unsigned integer types such as uint32_t, unless there is a valid reason such as representing a bit pattern rather than a number, or you need defined overflow modulo 2^N. In particular, do not use unsigned types to say a number will never be negative. Instead, use assertions for this.

yuxihu: Let me change to in to be consistent with other C APIs. I will add a check to make sure non-negative values are used internally.

yuxihu:
@eric-haibin-lin The PR is ready to merge. Thanks.

Issue #14574, opened 2019-03-31
Feature Request: 'arange_like/enumerate' operator


fhieber:
It would be quite useful to have an operator that can create ranges based on an existing 'shape_like' symbol, similar to the ops `zeros_like`, `ones_like`, and other `*_like`.

For example, I would like to enumerate the positions of a symbol with dynamic shape:
```python
x = mx.sym.Variable('x')
r = mx.sym.arange_like(x, start=0)
r.eval(x=mx.nd.ones((4,))) # should give: [0, 1, 2, 3]
r.eval(x=mx.nd.ones((2,))) # should give: [0, 1]
r2 = mx.sym.arange_like(x, start=5)
r.eval(x=mx.nd.ones((2, 2))) # should give: [[5, 6], [7, 8]]
```

Another name could be 'enumerate'.

vdantu:
@mxnet-label-bot add [operator, feature]

fhieber:
Resolved by #15400

PR #14638, opened 2019-04-07
[MXNET-1382] Add the index_array operator

nickguletskii:
## Description ##
This pull request implements `index_array`, an operator that returns an array of indexes of the input array.

For an input array with shape `(d_1, d_2, ..., d_n)`, `index_array` returns a `(d_1, d_2, ..., d_n, n)` array `idx`, where `idx[i_1, i_2, ..., i_n, :] = [i_1, i_2, ..., i_n]`.

Additionally, when the parameter `axes` is specified, `idx` will be a `(d_1, d_2, ..., d_n, m)` array where `m` is the length of `axes`, and the following
equality will hold: `idx[i_1, i_2, ..., i_n, j] = i_{axes[j]}`.

### Examples: ###
    x = mx.nd.ones((3, 2))

    mx.nd.contrib.index_array(x) = [[[0 0]
                                     [0 1]]

                                    [[1 0]
                                     [1 1]]

                                    [[2 0]
                                     [2 1]]]

    x = mx.nd.ones((3, 2, 2))

    mx.nd.contrib.index_array(x, axes=(1, 0)) = [[[[0 0]
                                                   [0 0]]

                                                  [[1 0]
                                                   [1 0]]]

                                                 [[[0 1]
                                                   [0 1]]

                                                  [[1 1]
                                                   [1 1]]]

### Motivation ###
This operator can be used to generate meshgrids for tensors without knowing their exact shapes during construction. For instance, this operator can be used to make a makeshift prior box generator for anchor-based computer vision models:

    feature_map = F.ones((8, 128, 128, 256)) # N x H x W x C, no shape information when using the Symbol API.
    prior_box_stride = 16
    box_size=[8, 8]

    template = F.squeeze(F.slice_axis(feature_map, begin=0, end=1, axis=-1), axis=-1) # N x H x W
    box_centres = F.contrib.index_array(template, axes=(-2, -1, -2, -1)).astype("float32") # N x H x W x 4
    box_centres = F.broadcast_mul(box_centres, F.array([prior_box_stride]).reshape((1, 1, 1, 1))) # N x H x W x 4
    corner_offsets = F.array(box_size).reshape((1, 1, 1, 2))
    corner_offsets = F.concat(-corner_offsets/2, corner_offsets/2, dim=-1)
    box_corners = F.broadcast_plus(box_centres, corner_offsets)

Also, this operator can be applied to implement positional encodings for sequence processing, e.g.:

    sequence_embeddings = F.ones((65, 8, 256)) # T x N x C, no shape information when using the Symbol API.
    template = sequence_embeddings.reshape((0, 0, -1, 2)) # T x N x C -> T x N x (C/2) x 2
    pos, i = F.split(
        F.contrib.index_array(template, axes=(0, 2)).astype("float32"), # T x N x (C/2) x 2 x 2
        axis=-1,
        num_outputs=2,
        squeeze_axis=True
    ) # T x N x (C/2) x 2 and T x N x (C/2) x 2
    base = F.ones((1, 1, 1, 1)) * 10000
    dmodel = F.slice_axis(F.shape_array(sequence_embeddings), begin=-1, end=None, axis=0)
    dmodel = dmodel.reshape((1, 1, 1, 1)).astype("float32")
    tmp = F.broadcast_div(pos, F.broadcast_power(base, F.broadcast_div(2 * i,  dmodel))) # T x N x (C/2) x 2
    sin_input, cos_input = F.split(tmp, axis=-1, num_outputs=2, squeeze_axis=True) # T x N x (C/2) and T x N x (C/2)
    positional_encoding = F.stack(F.sin(sin_input), F.cos(cos_input), axis=-1).reshape((0, 0, -3)) # T x N x C

I've also encountered situations where this operator would have been useful for some indexing tricks.

### Essentials ###

### Changes ###
- [x] The IndexArray operator (for the CPU)
- [x] The IndexArray operator (for the GPU)
- [x] Tests for the CPU implementation
- [x] Tests for the GPU implementation
- [x] Entries on the Python API reference pages

## Comments ##
- This operator always returns kInt64.
- The new tests in `test_operator_gpu.py` are exactly the same as in `test_operator.py`. I couldn't find any evidence that `test_operator.py` ever gets called with a GPU `default_context`, so I copied the tests into `test_operator_gpu.py` to make sure that the GPU implementation works too.

### DIFF
--- a/src/operator/contrib/index_array-inl.h
+++ b/src/operator/contrib/index_array-inl.h
@@ -1,0 +1,110 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef MXNET_OPERATOR_CONTRIB_INDEX_ARRAY_INL_H_
+#define MXNET_OPERATOR_CONTRIB_INDEX_ARRAY_INL_H_
+
+#include <vector>
+#include <utility>
+#include "../mshadow_op.h"
+#include "../tensor/init_op.h"
+
+namespace mxnet {
+namespace op {
+
+namespace index_array_enum {
+enum IndexArrayOpInputs {kIn};
+enum IndexArrayOpOutputs {kOut};
+enum IndexArrayOpResource {kTempSpace};
+}  // namespace index_array_enum
+
+template<int req>
+struct IndexArrayKernel {
+  MSHADOW_XINLINE static void Map(size_t i,
+                                  int64_t* out_data,
+                                  const uint32_t n,
+                                  const int64_t* workspace) {
+    for (uint32_t j = 0; j < n; j++) {
+      int64_t upper = workspace[2 * j];
+      int64_t lower = workspace[2 * j + 1];
+      KERNEL_ASSIGN(out_data[i * n + j], req, (i % upper) / lower);
+    }
+  }
+};
+
+template<int req>
+struct IndexArrayDefaultKernel {
+  MSHADOW_XINLINE static void Map(size_t i,
+                                  int64_t* out_data,
+                                  const uint32_t ndim,
+                                  const dim_t* shape) {
+    int64_t index = i;
+    for (uint32_t j = ndim; j-- > 0;) {
+      KERNEL_ASSIGN(out_data[i * ndim + j], req, index % shape[j]);
+      index /= shape[j];
+    }
+  }
+};
+
+inline std::vector<int64_t> IndexArrayComputeIndexProducts(const TShape &inshape) {
+  const uint32_t ndim = inshape.ndim();
+
+  std::vector<int64_t> index_products(ndim + 1);
+
+  index_products[ndim] = 1;
+
+  for (uint32_t i = ndim; i-- > 0;) {
+    index_products[i] = index_products[i + 1] * inshape[i];
+  }
+
+  return index_products;
+}
+
+inline void IndexArrayBuildSelectedAxesWorkspace(const TShape &axes,
+                                                 const std::vector<int64_t> &index_products,
+                                                 int64_t* workspace,
+                                                 const uint32_t ndim) {
+  for (uint32_t i = 0; i < axes.ndim(); i++) {
+    // Make sure that the axis is between 0 and ndim.
+    const dim_t axis = ((axes[i] % ndim) + ndim) % ndim;
+
+    workspace[2 * i] = index_products[axis];
+    workspace[2 * i + 1] = index_products[axis + 1];
+  }
+}
+
+template<typename xpu>
+void IndexArrayForward(const nnvm::NodeAttrs &attrs,
+                       const OpContext &ctx,
+                       const std::vector<TBlob> &inputs,
+                       const std::vector<OpReqType> &req,
+                       const std::vector<TBlob> &outputs);
+
+struct IndexArrayParam : public dmlc::Parameter<IndexArrayParam> {
+  dmlc::optional<mxnet::TShape> axes;
+  DMLC_DECLARE_PARAMETER(IndexArrayParam) {
+    DMLC_DECLARE_FIELD(axes).set_default(dmlc::optional<mxnet::TShape>())
+      .describe("The axes to include in the index array. Supports negative values.");
+  }
+};  // struct IndexArrayParam
+
+}  // namespace op
+}  // namespace mxnet
+
+#endif  // MXNET_OPERATOR_CONTRIB_INDEX_ARRAY_INL_H_
--- a/src/operator/contrib/index_array-inl.h
+++ b/src/operator/contrib/index_array-inl.h
@@ -36,11 +36,11 @@ enum IndexArrayOpResource {kTempSpace};
 
 template<int req>
 struct IndexArrayKernel {
-  MSHADOW_XINLINE static void Map(size_t i,
+  MSHADOW_XINLINE static void Map(int i,
                                   int64_t* out_data,
-                                  const uint32_t n,
+                                  const int n,
                                   const int64_t* workspace) {
-    for (uint32_t j = 0; j < n; j++) {
+    for (int j = 0; j < n; j++) {
       int64_t upper = workspace[2 * j];
       int64_t lower = workspace[2 * j + 1];
       KERNEL_ASSIGN(out_data[i * n + j], req, (i % upper) / lower);
@@ -50,12 +50,12 @@ struct IndexArrayKernel {
 
 template<int req>
 struct IndexArrayDefaultKernel {
-  MSHADOW_XINLINE static void Map(size_t i,
+  MSHADOW_XINLINE static void Map(int i,
                                   int64_t* out_data,
-                                  const uint32_t ndim,
+                                  const int ndim,
                                   const dim_t* shape) {
     int64_t index = i;
-    for (uint32_t j = ndim; j-- > 0;) {
+    for (int j = ndim - 1; j >= 0; j--) {
       KERNEL_ASSIGN(out_data[i * ndim + j], req, index % shape[j]);
       index /= shape[j];
     }
@@ -63,13 +63,13 @@ struct IndexArrayDefaultKernel {
 };
 
 inline std::vector<int64_t> IndexArrayComputeIndexProducts(const TShape &inshape) {
-  const uint32_t ndim = inshape.ndim();
+  const int ndim = inshape.ndim();
 
-  std::vector<int64_t> index_products(ndim + 1);
+  std::vector<int64_t> index_products(static_cast<size_t>(ndim + 1));
 
   index_products[ndim] = 1;
 
-  for (uint32_t i = ndim; i-- > 0;) {
+  for (int i = ndim - 1; i >= 0; i--) {
     index_products[i] = index_products[i + 1] * inshape[i];
   }
 
@@ -79,8 +79,8 @@ inline std::vector<int64_t> IndexArrayComputeIndexProducts(const TShape &inshape
 inline void IndexArrayBuildSelectedAxesWorkspace(const TShape &axes,
                                                  const std::vector<int64_t> &index_products,
                                                  int64_t* workspace,
-                                                 const uint32_t ndim) {
-  for (uint32_t i = 0; i < axes.ndim(); i++) {
+                                                 const int ndim) {
+  for (int i = 0; i < axes.ndim(); i++) {
     // Make sure that the axis is between 0 and ndim.
     const dim_t axis = ((axes[i] % ndim) + ndim) % ndim;
 
--- a/src/operator/contrib/index_array-inl.h
+++ b/src/operator/contrib/index_array-inl.h
@@ -76,13 +76,13 @@ inline std::vector<int64_t> IndexArrayComputeIndexProducts(const TShape &inshape
   return index_products;
 }
 
-inline void IndexArrayBuildSelectedAxesWorkspace(const TShape &axes,
+inline void IndexArrayBuildSelectedAxesWorkspace(const mxnet::Tuple<int> &axes,
                                                  const std::vector<int64_t> &index_products,
                                                  int64_t* workspace,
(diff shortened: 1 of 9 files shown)


piyushghai:
Thanks for your contributions @nickguletskii. Can you also look into the CI failures ? 
@mxnet-label-bot Add [Opeartor, pr-awaiting-review]

piyushghai:
@mxnet-label-bot Update [Operator, pr-awaiting-review]

nickguletskii:
The CI failure seems to be unrelated. Here's the relevant part of the log:

```
[2019-04-03 23:57:30   ERROR] Cuda initialization failure with error 35. Please check cuda installation:  http://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html.

terminate called after throwing an instance of 'std::runtime_error'

  what():  Failed to create object
```

If I am not mistaken, this error code corresponds to the following error message: "CUDA driver version is insufficient for CUDA runtime version". Could this be caused by the recent update to the `nvidia/cuda:10.0-cudnn7-devel` Docker tag? The [TensorRT build](https://github.com/apache/incubator-mxnet/blob/master/ci/docker/Dockerfile.build.ubuntu_gpu_tensorrt#L21) uses `nvidia/cuda:10.0-cudnn7-devel` while other builds use `nvidia/cuda:9.1-cudnn7-devel` (e.g. the [ubuntu_gpu build](https://github.com/apache/incubator-mxnet/blob/dab4ffcd2adb81db4a639c8db7e43c74165b296f/ci/docker/Dockerfile.build.ubuntu_gpu#L21)).

fhieber:
Thanks for implementing this! This seems similar to what I requested in #14574, correct?

Any reason this should be in the contrib namespace?

nickguletskii:
@fhieber In a way, yes. The behaviour you have described in #14574 can be emulated by applying `reshape`, `index_array`, `broadcast_add`, and finally, `reshape_list`. I was thinking of implementing `arange_like` instead of `index_array`, but I don't find the `arange` operator itself versatile enough and I didn't want to make `arange_like` too different from `arange`.

There's no particular reason why I've placed `index_array` into the contrib namespace other than the fact that it's a new operator. I'd be happy to update this pull request and move `index_array` out of the contrib namespace if the community deems it desirable.

fhieber:
Thanks @nickguletskii , I agree, your operator seems more versatile, thanks for working on it! 

`contrib` is fine with me, I just asked out of curiosity on whether the preferred or standard way of introducing new operators is to always place in `contrib` first.

nickguletskii:
@piyushghai  Sorry to bother you, but should I rebase this pull request on master now that the failing job (test_unix_python3_tensorrt_gpu) has been disabled (#14642), or should I just leave it as it is and wait for the review?

piyushghai:
@nickguletskii You should rebase with the latest master branch.  That'll enable you to push again, allowing the CI to pass on your PR. 
The reviews will happen on it in the meanwhile :)

nickguletskii:
Huh, this is weird, I've updated this PR to be compatible with the recently merged support for zero-dim and zero-shape arrays, and now the CI for Windows CPU tests are failing. However, when I try to run the tests locally (on a 64 bit Windows machine), the tests pass just fine. I'll make further attempts to reproduce this issue on my local machine tomorrow. I suspect that I've screwed up with memory access somewhere because the test runner crashes without providing any useful debugging information.

nickguletskii:
The previous CI failures on Windows were caused by indexing C-style arrays using ints. I'm not sure whether this was just undefined behaviour or a bug in the older versions of MSVC++, but it did not happen with VC++ 16.0. I fixed the issue by using ptrdiff_t to index C-style arrays.

nickguletskii:
@szha @reminisce This PR should be ready for a review now. The CI failure seems to be unrelated (the failing test is test_operator_gpu.test_convolution_multiple_streams).

vandanavk:
@szha @reminisce @fhieber Is this PR good to go?

src/operator/contrib/index_array.cc
```diff
@@ -0,0 +1,171 @@
...
+                                    return std::vector<std::string>{ "output" };
+                                  })
+.set_attr_parser(ParamParser<IndexArrayParam>)
+.set_attr<mxnet::FInferShape>("FInferShape", [](const nnvm::NodeAttrs &attrs,
+                                                mxnet::ShapeVector *in_shape,
+                                                mxnet::ShapeVector *out_shape) {
+  const IndexArrayParam &param = nnvm::get<IndexArrayParam>(attrs.parsed);
+  CHECK_EQ(in_shape->size(), 1U);
+  CHECK_EQ(out_shape->size(), 1U);
+  mxnet::TShape inshape = in_shape->at(index_array_enum::kIn);
```
reminisce: 1. `inshape` needs to be checked whether its `ndim` is known (use `ndim_is_known` function). If not, should return false if it's unknown.
2. `inshape` can be a reference of `in_shape-at(0)`.

src/operator/contrib/index_array.cc
```diff
@@ -0,0 +1,171 @@
...
+  for (int i = 0; i < inshape.ndim(); i++) {
+    oshape[i] = inshape[i];
+  }
+  if (param.axes.has_value()) {
+    oshape[inshape.ndim()] = param.axes.value().ndim();
+  } else {
+    oshape[inshape.ndim()] = inshape.ndim();
+  }
+  out_shape->clear();
+  out_shape->push_back(oshape);
```
reminisce: Please use `SHAPE_ASSIGN_CHECK` for shape assignment safety.

src/operator/contrib/index_array.cc
```diff
@@ -0,0 +1,171 @@
...
+  return true;
```
reminisce: Should return `shape_is_known(oshape)`.

src/operator/contrib/index_array-inl.h
```diff
@@ -0,0 +1,110 @@
...
+  for (int i = 0; i < axes.ndim(); i++) {
+    // Make sure that the axis is between 0 and ndim.
+    const int axis = ((axes[i] % ndim) + ndim) % ndim;
+
+    workspace[ptrdiff_t(2) * ptrdiff_t(i)] = index_products[axis];
+    workspace[ptrdiff_t(2) * ptrdiff_t(i) + ptrdiff_t(1)] = index_products[axis + 1];
+  }
+}
+
+template<typename xpu>
```
reminisce: Since you have separate implementations for cpu and gpu, it seems that defining a template and specializing it in .cc and .cu is unnecessary. You can just define different implementations in .cc and .cu without introducing templates.

tests/python/gpu/test_operator_gpu.py
```diff
@@ -2130,6 +2130,68 @@ def test_bilinear_sampler_versions():
                 if req_dict['grid'] is 'write':
                     assert_almost_equal(exe.grad_dict['grid'].asnumpy(), exe_list[ref_idx].grad_dict['grid'].asnumpy(), rtol=1e-3, atol=1e-5)
 
+@with_seed()
```
reminisce: The tests defined in `test_operator.py` will be run with `ctx=mx.gpu()` in CI. There is no need to duplicate test code in this file.

tests/python/unittest/test_operator.py
```diff
@@ -8054,6 +8054,68 @@ def test_image_normalize():
     # check backward using finite difference
     check_numeric_gradient(img_norm_sym, [data_in_4d], atol=0.001)
 
+@with_seed()
```
reminisce: Could you define a function called `test_index_array` that holds all the test functions you defined here? It's easier and cleaner to maintain just one test function for one operator.

tests/python/unittest/test_operator.py
```diff
@@ -8054,6 +8054,68 @@ def test_image_normalize():
...
+        input_array = np.ones(shape)
+        mgrid = np.mgrid[tuple(slice(0, x) for x in shape)]
+        expected = np.stack(mgrid, axis=-1)
+
+        check_symbolic_forward(index_array, [input_array], [expected])
+        check_symbolic_backward(index_array, [input_array], [np.ones(expected.shape)], [np.zeros_like(input_array)])
+
+@with_seed()
+def test_index_array_default_zero_dim():
+    with mx.np_compat(active=True):
```
reminisce: If the whole function is expected to be numpy compatible, you can use a decorator `mx.use_np_compat` to wrap the whole function. It would be easier for us to just delete the decorator in the next major release than putting additional efforts of adjusting indentation.

tests/python/unittest/test_operator.py
```diff
@@ -8054,6 +8054,68 @@ def test_image_normalize():
...
+
+        input_array = np.ones(())
+        expected = np.zeros((0,))
+
+        check_symbolic_forward(index_array, [input_array], [expected])
+        check_symbolic_backward(index_array, [input_array], [np.ones(expected.shape)], [np.zeros_like(input_array)])
+
+@with_seed()
+def test_index_array_default_zero_size():
+    with mx.np_compat(active=True):
```
reminisce: Same as above, please use the decorator `mx.use_np_compat`.

tests/python/unittest/test_operator.py
```diff
@@ -8054,6 +8054,68 @@ def test_image_normalize():
...
+        input_array = np.ones(shape)
+        mgrid = np.mgrid[tuple(slice(0, x) for x in shape)]
+        expected = np.stack(mgrid, axis=-1)[..., axes]
+
+        check_symbolic_forward(index_array, [input_array], [expected])
+        check_symbolic_backward(index_array, [input_array], [np.ones(expected.shape)], [np.zeros_like(input_array)])
+
+@with_seed()
+def test_index_array_select_axes_zero_size():
+    with mx.np_compat(active=True):
```
reminisce: Same as above.

src/operator/contrib/index_array-inl.h
nickguletskii: Sorry, I don't think templates can be avoided here without manual name mangling, i.e. calling the GPU implementation `IndexArrayForwardGpu`, since both versions of the function are linked into the final library. For the sake of completeness, I've tried compiling without the templates, and received the following error:

```
/work/mxnet/src/operator/contrib/index_array.cu:31: multiple definition of `mxnet::op::IndexArrayForward(nnvm::NodeAttrs const&, mxnet::OpContext const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&, std::vector<mxnet::OpReqType, std::allocator<mxnet::OpReqType> > const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&)'
build/src/operator/contrib/index_array.o:/work/mxnet/src/operator/contrib/index_array.cc:30: first defined here
```

Should I still remove the templates and manually mangle the names? It seems that template specialization is heavily utilised by other operators, especially when they are defined through `CreateOp`.

src/operator/contrib/index_array.cc
nickguletskii: Thank you, I've fixed this in 5343f133b2e147add8033aa7c3d7496e5fc62f54 and 55e07ba350cf001023042968f096b5f4e6240964.

src/operator/contrib/index_array.cc
nickguletskii: Thank you, I didn't know about `SHAPE_ASSIGN_CHECK`. Fixed in ca819abf92fa9b3205e9921a2b519f27c90adfa8.

src/operator/contrib/index_array.cc
nickguletskii: Thanks, I've fixed this in 5343f133b2e147add8033aa7c3d7496e5fc62f54.

tests/python/gpu/test_operator_gpu.py
nickguletskii: Sorry, I've removed these redundant tests in 3cdfcf582cb522b88938128c5c8025b1562267e5.

tests/python/unittest/test_operator.py
nickguletskii: Sure! The relevant commit is 3a430daaba4ea04bae90e2326e66febfef70a00f.

tests/python/unittest/test_operator.py
nickguletskii: Thanks, I've updated the subtests to use `@mx.use_np_compat` in 139567b9e5d1cf909b69355e0aeb162e38eb5ed6.

src/operator/contrib/index_array-inl.h
reminisce: Thanks for trying. I think manually naming forward functions for different contexts is fine, since the two implementation versions have nothing to share. Template function specialization used like in this case is actually not a preferred practice in writing C++ code. See here. http://www.gotw.ca/publications/mill17.htm

nickguletskii: I've removed the use of template specialization in 5f7f8a9512e21165467ef2ab00f3d57cd1ae394d, although I am not quite sure that the article that you've linked to applies in this case, since there's no overloading going on here. The functions share the same signature, and unless the processing unit is explicitly specified, the compiler will complain that it can't deduce the template parameter. Regardless, I am happy with either solution, since it's unlikely that `IndexArrayForward` will ever be used anywhere else.

karan6181:
@nickguletskii Could you please re-trigger the CI again? 
@szha Could you please review this PR? Thanks!

PR #14641, opened 2019-04-08
[MKLDNN]Improve quantizeV2 and dequantize latency

ZhennanQin:
## Description ##
Change quantizeV2 and dequantize to stateful op to improve small batch size performance.

@pengzhao-intel  @TaoLv  @reminisce @zheng-da @anirudh2290 

### Essentials ###

### Changes ###

## Comments ##


piyushghai:
@ZhennanQin Thanks for your contributions. Can you look at the CI failures ? 
@mxnet-label-bot Add [Quantization, pr-work-in-progress]

glingyan:
@piyushghai ping for any further comments?

src/operator/quantization/quantize_v2.cc
```diff
@@ -47,6 +47,22 @@ static bool QuantizeV2StorageType(const nnvm::NodeAttrs& attrs, const int dev_ma
   return true;
 }
 
+static OpStatePtr CreateQuantizeV2State(const nnvm::NodeAttrs& attrs, Context ctx,
+                                        const std::vector<TShape>& in_shapes,
+                                        const std::vector<int>& in_types) {
+  OpStatePtr state;
+  if (ctx.dev_type == kGPU) {
+    state = OpStatePtr::Create<QuantizeV2Operator<gpu>>(attrs);
```
anirudh2290: why are we changing the gpu op to be stateful to ? seems like that doesnt help much.

src/operator/quantization/mkldnn/mkldnn_dequantize-inl.h
```diff
@@ -26,82 +26,104 @@
...
+class SgMKLDNNDequantizeOperator {
+ public:
+  explicit SgMKLDNNDequantizeOperator(const nnvm::NodeAttrs &attrs)
+      : param_(nnvm::get<DequantizeParam>(attrs.parsed)) {}
 
+  void Forward(const OpContext &ctx, const std::vector<NDArray> &inputs,
+               const std::vector<OpReqType> &req, const std::vector<NDArray> &outputs);
+
+ private:
+  bool initalized_{false};
```
anirudh2290: nit: initialized

src/operator/quantization/mkldnn/mkldnn_quantize_v2-inl.h
```diff
@@ -34,99 +34,37 @@
...
+               const std::vector<OpReqType> &req, const std::vector<NDArray> &outputs);
 
-  primitive_attr attr;
-  const int mask = 0;
-  std::vector<float> scales = {scale};
-  attr.set_output_scales(mask, scales);
-  attr.set_int_output_round_mode(round_nearest);
-  mkldnn::engine cpu_engine = mxnet::CpuEngine::Get()->get_engine();
+ private:
+  bool initalized_{false};
```
anirudh2290: nit: initialized

src/operator/quantization/mkldnn/mkldnn_quantize_v2-inl.h
```diff
@@ -137,21 +75,101 @@ static void MKLDNNQuantizeV2Compute(const nnvm::NodeAttrs& attrs, const OpContex
...
+      float real_range = MaxAbs(data_min, data_max);
+      quantized_range = kInt8Range;
+      *outputs[1].data().dptr<float>() = -real_range;
+      *outputs[2].data().dptr<float>() = real_range;
     } else {
       LOG(FATAL) << "mkldnn quantize op only supports int8 and uint8 as output type";
     }
+
+    if (initalized_ && (cached_data_min_ != data_min || cached_data_max_ != data_max))
+      initalized_ = false;
```
anirudh2290: shouldnt this float comparison happen with an epsilon value

src/operator/quantization/mkldnn/mkldnn_dequantize-inl.h
```diff
@@ -26,82 +26,104 @@
...
-    i_dims[i] = static_cast<int>(in_buffer.shape()[i]);
-  }
-  mkldnn::memory::format i_fmt = static_cast<mkldnn::memory::format>(i_desc.data.format);
-  if (i_fmt == mkldnn::memory::format::nhwc) {
-    // For 4d tensor, nchw is the default format
-    i_fmt = mkldnn::memory::format::nchw;
+  float data_min = *inputs[1].data().dptr<float>();
+  float data_max = *inputs[2].data().dptr<float>();
+
+  if (initalized_ && (cached_data_min_ != data_min || cached_data_max_ != data_max))
```
anirudh2290: same as above

src/operator/quantization/quantize_v2.cc
ZhennanQin: You're right. Making gpu op stateful doesn't bring any help to gpu performance. This is because of the limitation of FW, that is, if an op registered `FCreateOpState`, then it will be treated as stateful op for all backends. `FComputeEx` and `FCompute` will be ignored. Only `FStatefulCompute` and `FStatefulComputeEx` are used. To implement `FStatefulCompute<gpu>`, we have to create a state for it.

src/operator/quantization/mkldnn/mkldnn_quantize_v2-inl.h
ZhennanQin: epsilon isn't necessary for this float comparison, because `cached_data_min_` is directly assigned from last used `data_min`, so `cached_data_min_` and `data_min` should be bit-wise identity if they are the same. Also, even if this check failed, the only bad thing is this op will be re-initialized and spent extra time to re-create mkldnn primitive. The output accuracy can still guarantee.

src/operator/quantization/mkldnn/mkldnn_dequantize-inl.h
ZhennanQin: Seems I fail to spell this word long time ago. Thanks for pointing this.

src/operator/quantization/mkldnn/mkldnn_quantize_v2-inl.h
anirudh2290: i see. reinitializing has its own cost associated with it and may impact perf, so if it doesn't change beyond a particular epsilon we may not need to reinitialize.

src/operator/quantization/quantize_v2.cc
anirudh2290: > This is because of the limitation of FW, that is, if an op registered FCreateOpState, then it will be treated as stateful op for all backends. FComputeEx and FCompute will be ignored. 

maybe i am missing something but seems like this should be fixed in the FW. 

also you could work around this by adding FCreateOpState inside the #if MXNET_USE_MKLDNN == 1 check, seems like neither non-mkldnn cpu or gpu implementation requires stateful op.

ZhennanQin: Your suggestion will have problem to build mkldnn and cuda together, just as the case of mxnet_cu92mkl. for mxnet_cu92mkl, `MXNET_USE_MKLDNN == 1` is enabled, but end-user may use GPU only.

anirudh2290: yes. makes sense to keep the way it is. hopefully this wont cause any perf impact.

src/operator/quantization/mkldnn/mkldnn_quantize_v2-inl.h
ZhennanQin: Yes. It's an optimization we can consider to add. But I have to say, adding epsilon comparison can only benefit some rare case, but slightly hurt the performance of static graph with calibration, which is the most use case for this op. Also, if epsilon isn't chose well, accuracy issue may introduce and hard to debug. So I suggest to add this when we face an case that really need it.

anirudh2290: by static graph with calibration do you mean the case where param.calib.min_range and param.max_range has been set ? I think we already have the information, so we can decide to check epsilon only if its not set. I agree though that it may not be easy to set the epsilon value correctly. makes sense to revisit this later if needed.

ZhennanQin: This check is moved to the case where param.calib.min_range and param.max_range has not been set. So even if we add epsilon comparison someday, op with calibration won't be slow down. Thanks for this suggestion.

src/operator/quantization/mkldnn/mkldnn_dequantize-inl.h
ZhennanQin: As above.

pengzhao-intel:
Merging now.

PR #14657, opened 2019-04-09
[Numpy] fix invalid ndarray dispose

yzhliu:


reminisce:
Thanks for the fix. Now all the tests but Perl's have passed.

PR #14661, opened 2019-04-10
[numpy] Support zero-dim and zero-size tensors in MXNet

reminisce:
## Description ##
This PR provides the infrastructure of supporting zero-dim and zero-size tensors as the first outcome of the initiative of introducing NumPy compatible coding experience into MXNet (see this [RFC](https://github.com/apache/incubator-mxnet/issues/14253)). Great thanks to many folks who have contributed design, implementation and code review to this feature. The credits should go to all of them. Sorry if I missed anyone below. It would be impossible to make so many things correct scattered all over the codebase with just a couple of hands.
- C++ backend and Python: @junrushao1994 @eric-haibin-lin @szha @zheng-da @wkcn @mli 
- MKLDNN: @pengzhao-intel @TaoLv @lihaofd @ZhennanQin
- Scala: @yzhliu 
- R: @hetong007 
- Perl: @sergeykolychev 

### What are zero-dim and zero-size tensors?
- Zero-dim tensors are scalars with shape equal to `()`.
- A zero-size tensor is the one whose shape has at least one dimension size as `0`. For example, `np.ones((1, 0, 3))` would generate a zero-size tensor with shape equal to `(1, 0, 3)`.

### Why are they important?
Mathematically speaking, their presence keeps the completeness and consistency of the all the tensor operations. For example, given `x = mx.nd.array([1, 2, 3])`, `x[0]` should return `1`, instead of `[1]`, and `x[0:0]` should return `[]` with shape equal to `(0,)`. Zero-size tensors can also be convenient in easing the code logic as placeholders for accumulations or aggregations.

**I find [this thread](https://discuss.mxnet.io/t/rank-0-arrays-in-mxnet-aka-pi-is-wrong/108) provides very good insights on the importance of zero-dim tensors.**

### How are they implemented?
In the backend (C++), we use `ndim = -1` to represent unknown shapes and `dim_size = -1` for unknown dim sizes. Before, they were represented by `0`.

### Is backward compatibility guaranteed in this PR?
Yes, the backward compatibility is guaranteed by default. That means `0` still represents unknown `ndim` or dim size in any frontend language bindings. It's just converted to `-1` in the infer shape logic of the backend.

### How to enable zero-dim or zero-size tensors?
Since we are committed to keep backward compatibility, we provided APIs for users to decide whether to opt in for this NumPy compatibility. Users can call `mx.set_np_compat(True)` to opt in and `mx.set_np_compat(False)` to opt out. Or in a safer way, use the `with` statement. To turn on this:
```python
 with mx.np_compat(active=True):
     # A scalar tensor's shape is `()`, whose `ndim` is `0`.
     scalar = mx.nd.ones(shape=())
     assert scalar.shape == ()
 
     # In NumPy compatible mode, 0 in a shape means that dimension contains zero elements.
     data = mx.sym.var("data", shape=(0, 2, 3))
     ret = mx.sym.sin(data)
     arg_shapes, out_shapes, _ = ret.infer_shape()
     assert arg_shapes[0] == (0, 2, 3)
     assert out_shapes[0] == (0, 2, 3)
 
     # -1 means unknown shape dimension size in the new NumPy-compatible shape definition
     data = mx.sym.var("data", shape=(-1, 2, 3))
     ret = mx.sym.sin(data)
     arg_shapes, out_shapes, _ = ret.infer_shape_partial()
     assert arg_shapes[0] == (-1, 2, 3)
     assert out_shapes[0] == (-1, 2, 3)
 
     # When a shape is completely unknown in NumPy-compatible mode, it is
     # represented as `None` in Python.
     data = mx.sym.var("data")
     ret = mx.sym.sin(data)
     arg_shapes, out_shapes, _ = ret.infer_shape_partial()
     assert arg_shapes[0] is None
     assert out_shapes[0] is None
```
or to disable this:
```python
 with mx.np_compat(active=False):
     # 0 means unknown shape dimension size in the legacy shape definition.
     data = mx.sym.var("data", shape=(0, 2, 3))
     ret = mx.sym.sin(data)
     arg_shapes, out_shapes, _ = ret.infer_shape_partial()
     assert arg_shapes[0] == (0, 2, 3)
     assert out_shapes[0] == (0, 2, 3)
 
     # When a shape is completely unknown in the legacy mode (default), its ndim is
     # equal to 0 and it is represented as `()` in Python.
     data = mx.sym.var("data")
     ret = mx.sym.sin(data)
     arg_shapes, out_shapes, _ = ret.infer_shape_partial()
     assert arg_shapes[0] == ()
     assert out_shapes[0] == ()
```
### Does this mean that every existing operator should support zero-dim or zero-size tensors in this PR?
Please note that the existing operators were implemented when these two types of tensors were not supported in MXNet. Some strong assumptions may have been made in their implementation and hence, lead to errors when NumPy compatibility is turned on. This PR only provides the infrastructure of supporting zero-dim and zero-size tensors in the backend. It does not guarantee that every existing operator would deal with zero-dim/size tensors correctly as in NumPy. As discussed in the [RFC](https://github.com/apache/incubator-mxnet/issues/14253), we are going to implement NumPy operators under `mxnet.numpy` that would fully support these two types of tensors.


wkcn:
Great work! Do we need to review this PR?

reminisce:
@hetong007 @yzhliu @sergeykolychev Could you please help review the changes in R, Scala, and Perl respectively? We would like to merge this PR sooner than later since our subsequent work depends on this and it's becoming harder to rebase.

szha:
@jeremiedb @lanking520 could you help out on this for reviews too? Thanks!

yzhliu:
LGTM

include/mxnet/tuple.h
```diff
@@ -390,7 +443,11 @@ class TShape : public Tuple<dim_t> {
    * \return self.
    */
   inline TShape& operator=(const Tuple<dim_t>& src) {
-    this->assign(src.begin(), src.end());
+    if (src.ndim() == -1) {
+      this->SetDim(-1);
+    } else {
```
szha: should `ndim_` be maintained?

include/mxnet/tuple.h
```diff
@@ -316,48 +340,72 @@ class Tuple {
...
   /*!
    * \brief copy constructor of TShape
    * \param s source shape.
    */
   inline TShape(const Tuple<dim_t>& s) { // NOLINT(*)
-    this->assign(s.begin(), s.end());
+    if (s.ndim() == -1) {
+      this->SetDim(-1);
+    } else {
+      this->assign(s.begin(), s.end());
```
szha: should `ndim_` be maintained?

include/mxnet/tuple.h
```diff
@@ -417,9 +476,11 @@ class TShape : public Tuple<dim_t> {
    * \param dimend end dimension
    */
   inline size_t ProdShape(int dimstart, int dimend) const {
+    CHECK_GE(this->ndim(), 0) << "Shape is unknown.";
     dim_t num = 1;
     const dim_t *d = this->data();
     for (int i = dimstart; i < dimend; ++i) {
```
szha: should there be boundary check for the index?

include/mxnet/tuple.h
```diff
@@ -475,11 +536,12 @@ class TShape : public Tuple<dim_t> {
    */
   inline mshadow::Shape<2> FlatTo2D(void) const {
     mshadow::Shape<2> s;
-    if (ndim() == 0) return mshadow::Shape2(0, 0);
+    CHECK_GE(ndim(), 0);
```
szha: error message

szha: or should we use ndim_is_known for the check?

python/mxnet/base.py
```diff
@@ -26,11 +26,11 @@
 import sys
 import inspect
 import platform
-import numpy as np
+import numpy as _np
 
 from . import libinfo
 
-__all__ = ['MXNetError']
+__all__ = ['MXNetError', 'is_np_compat', 'set_np_compat', 'enable_np_compat', 'disable_np_compat']
```
szha: I'd suggest we have one scope (e.g. `mx.np_compat(active=True)`) instead of two.
Also, shall we use the scope to implement a decorator so that it's easier to mark the functions where the flag can be turned on and off? (e.g. `@mx.np_compat(active=True)def foo():`, or you may need another name for the decorator)

scala-package/native/src/main/native/org_apache_mxnet_native_c_api.cc
```diff
@@ -892,6 +892,119 @@ JNIEXPORT jint JNICALL Java_org_apache_mxnet_LibInfo_mxExecutorBackward
   return ret;
 }
 
+JNIEXPORT jint JNICALL Java_org_apache_mxnet_LibInfo_mxExecutorReshape
```
lanking520: Is that possible to partially migrate the implementation to the Scala side for maintenance? It seemed too complicated for this method.

include/mxnet/tuple.h
reminisce: `assign` function will handle that.

include/mxnet/tuple.h
reminisce: Same as above.

scala-package/core/src/main/scala/org/apache/mxnet/NumpyScope.scala
```diff
@@ -0,0 +1,58 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.mxnet
+
+import org.apache.mxnet.Base._
+
+object NumpyScope {
```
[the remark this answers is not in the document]
reminisce: @yzhliu

scala-package/native/src/main/native/org_apache_mxnet_native_c_api.cc
reminisce: @yzhliu

python/mxnet/base.py
reminisce: One scope is a good suggestion.

The decorator may have to have a different name than `np_compat` since the `__init__` function would have a different signature. As of now, I don't see the strong need of such a decorator to move forward. How about we defer this till the time we really need it?

szha: When providing APIs of Gluon that are implemented with the numpy semantics, using a decorator would be easier to use than a scope as it doesn't require the indentation. It also means minimum amount of change for adapting to the next major version when we fully commit to numpy semantics. I'd recommend having the from the beginning.

jeremiedb:
@hetong007 Were you seing a need for adding integrating a `set_np_compat` option in the R-package? Otherwise, LGTM.

TaoLv:
Reviewed changes for mkldnn files. @juliusshufan you might want to have a performance validation on this PR.

python/mxnet/base.py
reminisce: That was needed when we wanted to dispatch fluent methods to numpy ops. But since we have decided to make a dedicated ndarray data type for numpy ops, such decorator may not be needed. We must clear out the thoughts for such use cases before adding the api. It cannot recalled once out. Let's discuss offline.

hetong007:
> @hetong007 Were you seing a need for adding integrating a `set_np_compat` option in the R-package? Otherwise, LGTM.

Would you elaborate more on this need? I think the current R API maps nd.array to native R array so this PR has minimum effect on R-related uses.

scala-package/native/src/main/native/org_apache_mxnet_native_c_api.cc
yzhliu: as discussed offline with @lanking520 , the code before and after `MXExecutorReshape` are for converting Java object to C++ object (e.g., string) and vice versa. Thus have to use jni to do so.

sergeykolychev:
@reminisce Please remove all changes related to Perl from this PR, if you use new Ex api calls, perl will not be affected by it and changes are not necessary (and bogus actually).

reminisce:
@sergeykolychev Just want to confirm, you want me remove all the change **you and I** made in perl-package, including those *.pm, mxnet.i, mxnet_typemaps.i? Thanks.

jeremiedb:
@hetong007 Yes I think you're right, misunderstanding on my part. All good for the R-pacage then.

sergeykolychev:
@reminisce I am sorry I did not realize that you have changed the .pm files also.
With these changes I believe perl API is good and you don't need to remove anything from your PR.
Let's see if the jenkins will confirm my theory.

reminisce:
@sergeykolychev Thanks for the clarification. The CI was able to pass after I modified *.pm files. Now it's failing due to rebasing with master. I will make it pass again. Thanks again for your work and being supportive and responsive to the initiative. :)

sergeykolychev:
@reminisce you're welcome! :-)

reminisce:
@anirudh2290 That's because np compat scope is a thread-local variable. You need to scope the forward function where numpy compatibility is expected. Scoping the main thread does not spread the compatibility to the worker thread. The following change should work.

```python
@mx.use_np_compat
```

anirudh2290:
Thanks for the decorator! I think customers may not understand the internal workings of custom op and the dispatch to a different thread workflow, and run into this pitfall. Can we synchronize the state of is_numpy_compat when pushing the op to custom op thread. Custom op is one area where customers use numpy operations and I feel there may be users running into this.

reminisce:
@anirudh2290 That's a good point from users' perspective. But this may not be a trivial implementation by propagating the `np_compat` from main thread to worker thread as potentially there are other pitfalls. I discussed this @eric-haibin-lin, maybe we can attach a compatibility state to the generated custom op. This can become a follow-up implementation in another PR. This PR does not need to be blocked by this.

include/mxnet/tuple.h
reminisce: Done

python/mxnet/base.py
reminisce: Done.

scala-package/core/src/main/scala/org/apache/mxnet/NumpyScope.scala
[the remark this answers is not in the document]
reminisce: Added.

include/mxnet/tuple.h
reminisce: Done.

cpp-package/include/mxnet-cpp/symbol.hpp
```diff
@@ -200,40 +200,40 @@ inline void Symbol::InferShape(
   arg_ind_ptr.push_back(arg_shape_data.size());
 
   mx_uint in_shape_size;
-  const mx_uint *in_shape_ndim;
-  const mx_uint **in_shape_data;
+  const int*in_shape_ndim;
```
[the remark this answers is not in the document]
reminisce: Fixed.

szha:
@anirudh2290 good points on custom op. Given the size of this PR and the fact that users need to opt-in to be affected by this change, I'd recommend we address these points in a follow-up PR, so that @reminisce won't have to keep full-time resolving conflicts :)

anirudh2290:
I am okay with making custom op changes later. just to be sure, the the is_numpy_compat thread local variable is set in the main thread, and is expected to be accessed only from main thread ? For example, this wont cause any issue the multi threaded inference interface provided by frontends like Scala ?

reminisce:
@anirudh2290 Thanks for the review. In practice,  `np_compat` needs to be set in the same thread as the one invoking shape inference functions, because that's where zero-dim and zero-size shapes are treated as unknown. In your example, the custom op's forward function invokes `mx.nd.ones` whose shape inference function was executed on the worker thread, and hence leads to failure. We are going to move all ops towards the direction of being numpy compatible, and that should include the custom op, where the forward/backward functions should always be scoped by the numpy-compatible state, and this problem can be resolved without asking users to use the decorator.

I'm not familiar with how scala multi-threaded inference is implemented, but as long as the np_compat is set in the same thread as the one invoking shape inference functions, the result should be as expected.

src/operator/tensor/matrix_op-inl.h
```diff
@@ -206,13 +209,16 @@ inline bool ReshapeShape(const nnvm::NodeAttrs& attrs,
       oshape[inf_idx] = dshape.Size() / oshape.Size();
     }
   } else {
-    return (*out_attrs)[0].ndim() && ReverseReshapeInferShape(&(*in_attrs)[0], (*out_attrs)[0]);
+    return shape_is_known((*out_attrs)[0])
+           && ReverseReshapeInferShape(&(*in_attrs)[0], (*out_attrs)[0]);
   }
   ReverseReshapeInferShape(&dshape, oshape);
+#if 0
```
leezu: @reminisce is this `#if 0` intentional? Will it be removed eventually?

PR #14673, opened 2019-04-11
[BUGFIX] fix ELU function will appear nan when calculating the gradient

fierceX:
## Description ##
fix ELU function will appear `nan` when calculating the gradient

e.g:
```python
elu = nn.ELU()
elu.initialize()
x = nd.ones((2, 3))
x.attach_grad()
x[0] = 100
x[1] = -1
with autograd.record():
    y = elu(x)

y.backward()
print(y)
print(x.grad)
```
output is 
```
[[100.         100.         100.        ]
 [ -0.63212055  -0.63212055  -0.63212055]]
<NDArray 2x3 @cpu(0)>

[[       nan        nan        nan]
 [0.36787945 0.36787945 0.36787945]]
<NDArray 2x3 @cpu(0)>
```
After experiments, it was found that the `where` and the `exp` were used together and that this bug would occur

e.g:

```python
x = nd.ones((2, 3))
x.attach_grad()
x[0] = 100
x[1] = 5

with autograd.record():
    y = nd.where(x > 0, x, nd.exp(x))
y.backward()
print(y)
print(x.grad)
```
output:
```
[[100. 100. 100.]
 [  5.   5.   5.]]
<NDArray 2x3 @cpu(0)>

[[nan nan nan]
 [ 1.  1.  1.]]
<NDArray 2x3 @cpu(0)>
```
When the `exp` calculation appears `inf`, even if `where` does not select the value, but the gradient will still be `nan`,so this PR does not completely fix the problem, but based on this, modified the ELU calculation method so that it does not appear `nan`

### DIFF
--- a/python/mxnet/gluon/nn/activations.py
+++ b/python/mxnet/gluon/nn/activations.py
@@ -158,7 +158,7 @@ class ELU(HybridBlock):
         self._alpha = alpha
 
     def hybrid_forward(self, F, x):
-         F.LeakyReLU(x, act_type='elu', slope=self._alpha)
+         return F.LeakyReLU(x, act_type='elu', slope=self._alpha)
 
 
 class SELU(HybridBlock):
--- a/python/mxnet/gluon/nn/activations.py
+++ b/python/mxnet/gluon/nn/activations.py
@@ -158,8 +158,7 @@ class ELU(HybridBlock):
         self._alpha = alpha
 
     def hybrid_forward(self, F, x):
-        _x = F.where(x < 0, x, F.zeros_like(x))
-        return F.where(x > 0, x, self._alpha * (F.exp(_x) - 1.0))
+         F.LeakyReLU(x, act_type='elu', slope=self._alpha)
 
 
 class SELU(HybridBlock):
--- a/python/mxnet/gluon/nn/activations.py
+++ b/python/mxnet/gluon/nn/activations.py
@@ -158,7 +158,7 @@ class ELU(HybridBlock):
         self._alpha = alpha
 
     def hybrid_forward(self, F, x):
-         return F.LeakyReLU(x, act_type='elu', slope=self._alpha)
+        return F.LeakyReLU(x, act_type='elu', slope=self._alpha)
 
 
 class SELU(HybridBlock):
--- a/python/mxnet/gluon/nn/activations.py
+++ b/python/mxnet/gluon/nn/activations.py
@@ -158,7 +158,8 @@ class ELU(HybridBlock):
         self._alpha = alpha
 
     def hybrid_forward(self, F, x):
-        return F.where(x > 0, x, self._alpha * (F.exp(x) - 1.0))
+        _x = F.where(x < 0, x, F.zeros_like(x))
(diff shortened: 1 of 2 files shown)


python/mxnet/gluon/nn/activations.py
```diff
@@ -158,7 +158,8 @@ def __init__(self, alpha=1.0, **kwargs):
         self._alpha = alpha
 
     def hybrid_forward(self, F, x):
-        return F.where(x > 0, x, self._alpha * (F.exp(x) - 1.0))
+        _x = F.where(x < 0, x, F.zeros_like(x))
+        return F.where(x > 0, x, self._alpha * (F.exp(_x) - 1.0))
```
haojin2: Thanks for your contribution!
Actually we do have ELU implemented in the backend so you can call it directly
Please use:
```Python
return F.LeakyReLU(x, act_type='elu', slope=self._alpha)
```
here instead and also you want to change: https://github.com/apache/incubator-mxnet/blob/master/tests/python/unittest/test_gluon.py#L1183 to:
```Python
return mx.nd.expm1(x) if x <= 0.0 else x
```
so that the test would still pass after this change.

fierceX: This bug does not affect the forward calculation, but occurs when the gradient is calculated backwards.

haojin2: @fierceX Yes I understand, and I'm presenting the recommended way to fix this issue to you in the above comment.

haojin2:
@fierceX You also need to change tests/python/unittest/test_gluon.py Line 1183 to:
```Python
return mx.nd.expm1(x) if x <= 0.0 else x
```
so that the test could pass, thanks!

python/mxnet/gluon/nn/activations.py
```diff
@@ -158,7 +158,7 @@ def __init__(self, alpha=1.0, **kwargs):
         self._alpha = alpha
 
     def hybrid_forward(self, F, x):
-        return F.where(x > 0, x, self._alpha * (F.exp(x) - 1.0))
+         F.LeakyReLU(x, act_type='elu', slope=self._alpha)
```
haojin2: Seems like you have an extra whitespace here, please get rid of that and this is good to go.

fierceX: Sorry, less `return`.

python/mxnet/gluon/nn/activations.py
```diff
@@ -158,7 +158,7 @@ def __init__(self, alpha=1.0, **kwargs):
         self._alpha = alpha
 
     def hybrid_forward(self, F, x):
-        return F.where(x > 0, x, self._alpha * (F.exp(x) - 1.0))
+         return F.LeakyReLU(x, act_type='elu', slope=self._alpha)
```
haojin2: Nit: The extra space before `return`

fierceX:
Note: There should still be a bug in the `where` operator.

haojin2:
@fierceX We'll surely investigate that.

haojin2:
@fierceX Can you rebase and do a force push?

haojin2:
@fierceX Merged, thanks for your contribution!

PR #14677, opened 2019-04-11
Fix profiler check

anirudh2290:
## Description ##
After adding the waitall PR the nightlies started failing and was reported here : https://github.com/apache/incubator-mxnet/pull/14397#issuecomment-481821129 
This was because waitall rethrowing exceptions that were hidden earlier.
Seems like operator-= for ProfileCounter is called during frees which checks if the amount to be freed is greater than alloced. Added a check before operator is called. 

More permanent fix would be to understand why memory pool counters are incorrect when free is called but this PR should stop the profiler nightly test failures.

@ThomasDelteil 

### Essentials ###

### Changes ###

## Comments ##


tests/python/unittest/test_exc_handling.py
```diff
@@ -165,6 +166,46 @@ def test_multiple_waitalls():
     assert caught, "No exception thrown"
     mx.nd.waitall()
 
+@with_seed()
+def test_exc_profiler():
```
ThomasDelteil: seems a bit heavy for a unittest, I think passing some nd.ones() through a simple dense layer would do the same right?

anirudh2290: I have simplified it. Please take a look

anirudh2290:
@yuxihu Good point I didnt investigate whether handle is getting double freed. I will check the same, though this PR doesn't have to block on it.

yuxihu:
> @yuxihu Good point I didnt investigate whether handle is getting double freed. I will check the same, though this PR doesn't have to block on it.

Yes, it is good to have the logic implemented in this PR anyway. The investigation can be done separately.

Issue #8968, opened 2017-12-06
1.0.0 crash with Google perf tools


juliensimon:
Hi,

1.0.0 dumps core at loading time when built with Google perf tools.

I'm using a freshly booted Deep Learning AMI, Ubuntu Linux - 2.4_Oct2017 (ami-dca37ea5).

Python 2.7.12 (default, Nov 20 2017, 18:23:56)
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import mxnet
src/tcmalloc.cc:278] Attempt to free invalid pointer 0x1d9cc60
Aborted (core dumped)

Steps to reproduce:
- sudo apt-get install libgoogle-perftools-dev
- clone a fresh MXNet repo
- set USE_GPERFTOOLS = 1
- build and install

The same build without libgoogle-perftools-dev installed works fine.

eric-haibin-lin:
@cjolivier01

cjolivier01:
@eric-haibin-lin

eric-haibin-lin:
@bhavinthaker @mbaijal should we also cover this in CI build tests?

cjolivier01:
It may be that it's not gperftools that's the problem, but that there's a bug and the normal malloc library doesn't make any issue of trying to free what it sees as an invalid pointer

bhavinthaker:
Hi Haibin: Yes, we should cover these tests in the CI. However I would request you or any team member to help contribute to the unit tests or create a backlog item (issue/Jira) so that any of the community members can work on it.

Hi Chris: Your theory could be valid. I remember Solaris being permissive but HPUX being strict for the same code in C. Such problems were typically found with strict checks using compiler flags/tools. Could you please investigate the problem and let us know your findings?

Do we have a unit-test in CI with Gperf enabled? If not, please add one. This should have been enforced by code-reviewers of the PR that enabled gperftools.

bhavinthaker:
Let's try a coverity run too: @rahul003

juliensimon:
Hello,

FWIW, I ran into this issue again a couple of days when building master.

KellenSunderland:
Maybe we could build and test with gperf, clang, and a few sanitizers?

rahul003:
Related issues
https://github.com/apache/incubator-mxnet/issues/9096
https://github.com/apache/incubator-mxnet/issues/9098

PR #14711, opened 2019-04-16
Disable USE_GPERFTOOLS by default

ZhennanQin:
## Description ##
Because of https://github.com/apache/incubator-mxnet/issues/8968, mxnet will random crash after USE_GPERFTOOLS = 1. Some end-user may get trouble if gperf-tools is  pre-installed. So before fix this, we'd better to disable it by default.

@pengzhao-intel @szha

### Essentials ###

### Changes ###

## Comments ##


pengzhao-intel:
@szha @eric-haibin-lin is there any reason that the gperf is ON by default?
Our customer encountered the crash w/ gperf ON, so we prefer to make it OFF by default.

pengzhao-intel:
cc @lebeg

pengzhao-intel:
Merging now since the BOSS agree :)

anirudh2290:
> I guess it was decided by an individual as there is significant performance benefit in using alternative malloc. It wasn't thoroughly evaluated and I actually noticed similar issues with tcmalloc, so I'm actually surprised that this is on by default in cmake.

An "individual" by himself/herself cannot make decisions in an apache incubating project.

cjolivier01:
Interesting theory szha, although also incorrect.  But disable if you want.  Jenalloc is faster most of the time anyway.

szha:
@cjolivier01 thanks for stepping in. I'd like to hear what you would like clairfy and what option you recommend?

PR #14716, opened 2019-04-17
Add unpooled gpu memory type

vladoovtcharov:
Adds option to configure mxnet to use unpooled gpu memory


roywei:
@eric-haibin-lin @apeforest could you help take a look? thanks!

Issue #14244, opened 2019-02-24
Feature request: share memory between numpy.array and mxnet.ndarray


SunDoge:
Last time I did a benchmark and @szha pointed out [that](https://gist.github.com/SunDoge/59a8ff336703b45be30b46dc3ee8b4ab#gistcomment-2841120)

> MXNet made the choice of not doing a zero-copy for numpy arrays, but instead making a copy of the numpy data. 

This is a safe choice but sometimes it can be a performance problem. I check the C API and find no function to share data from outside. Is it possible for MXNet to provide such api?

junrushao1994:
@szha will lead us to implement this feature. I will take a look this weekend.

junrushao1994:
Hey, I did some research this weekend and am convinced it might be possible. So is it acceptable if we assume we are using coupes (instead of cython) @szha

szha:
@junrushao1994 could you elaborate on the consideration?

junrushao1994:
The point is that we need transfer/share ownership of numpy's ndarray to mxnet's C++ backend, because we cannot guarantee the frontend object to exist forever.

Therefore, we need a customized deleter of MXNet's NDArray, aka calling something `Py_DECREF` of the numpy's ndarray object from the C++ backend - It is possible to implement the deleter via ctypes or cython, and then pass it to something roughly like follows this chunk of code.

https://github.com/apache/incubator-mxnet/blob/0f88f61379bd5f59fff6b825be1507d020bf2b7e/include/mxnet/ndarray.h#L131-L148

Per private discussion with @reminisce, his concern is how this could be compatible with the executor and memory planning, in which the executor may take over the ownership. I am not sure about this.

wkcn:
For the sharing from numpy to NDArray, we should use ctypes or weakref module add the inference of numpy object, and decrease the inference through NDArray::deleter.

For the sharing from NDArray to numpy, I think we can add a deleter attribute for numpy object.

junrushao1994:
Agreed with @wkcn

junrushao1994:
I prototyped a version that supports zero copy from numpy to DLManagedTensor in dlpack, but it turns out that MXNet hasn’t support dlpack 0.2 yet...

wkcn:
We can update the submodule dlpack.

junrushao1994:
@wkcn I opened a PR about this to https://github.com/dmlc/dlpack/pull/38.

junrushao1994:
[...]

PR #14733, opened 2019-04-18
[MXNET-1398] Enable zero-copy from numpy to MXNet NDArray

junrushao1994:
## Description ##
This PR allows users to convert numpy's NDArray without copying content data. This is achieved by converting numpy's NDArray to `DLManagedTensor`, which is introduced in https://github.com/dmlc/dlpack/pull/38. We expect that this would help users reduce major performance bottleneck when this conversion is conducted frequently.

See also: #14244.

### Essentials ###

### Changes ###
- [x] Rewrite `NDArray::FromDLPack` to make it resource safe.
- [x] Add DLManagedTensor and its deleter via ctypes that allows C++ side deletes it.

## Comments ##
Nothing.

### DIFF
--- a/include/mxnet/c_api.h
+++ b/include/mxnet/c_api.h
@@ -825,22 +825,6 @@ MXNET_DLL int MXNDArrayFromDLPack(DLManagedTensorHandle dlpack,
                                   NDArrayHandle *out_handle);
 
 /*!
-* \brief Create a NDArray backed by a dlpack tensor.
-*
-* This allows us to create a NDArray using the memory
-* allocated by an external deep learning framework
-* that is DLPack compatible.
-*
-* The memory is retained until the NDArray went out of scope.
-*
-* \param dlpack the pointer of the input DLManagedTensor
-* \param out_handle pointer holder to get pointer of NDArray
-* \return 0 when success, -1 when failure happens
-*/
-MXNET_DLL int MXNDArrayFromDLManagedTensor(DLManagedTensorHandle dlpack,
-                                           NDArrayHandle *out_handle);
-
-/*!
  * \brief Delete a dlpack tensor
  * \param dlpack the pointer of the input DLManagedTensor
  * \return 0 when success, -1 when failure happens
--- a/include/mxnet/c_api.h
+++ b/include/mxnet/c_api.h
@@ -823,6 +823,23 @@ MXNET_DLL int MXNDArrayToDLPack(NDArrayHandle handle,
 */
 MXNET_DLL int MXNDArrayFromDLPack(DLManagedTensorHandle dlpack,
                                   NDArrayHandle *out_handle);
+
+/*!
+* \brief Create a NDArray backed by a dlpack tensor.
+*
+* This allows us to create a NDArray using the memory
+* allocated by an external deep learning framework
+* that is DLPack compatible.
+*
+* The memory is retained until the NDArray went out of scope.
+*
+* \param dlpack the pointer of the input DLManagedTensor
+* \param out_handle pointer holder to get pointer of NDArray
+* \return 0 when success, -1 when failure happens
+*/
+MXNET_DLL int MXNDArrayFromDLManagedTensor(DLManagedTensorHandle dlpack,
+                                           NDArrayHandle *out_handle);
+
 /*!
  * \brief Delete a dlpack tensor
  * \param dlpack the pointer of the input DLManagedTensor
--- a/python/mxnet/ndarray/ndarray.py
+++ b/python/mxnet/ndarray/ndarray.py
@@ -4146,7 +4146,7 @@ class DLTensor(ctypes.Structure):
                 ("byte_offset", ctypes.c_uint64)]
 
 class DLManagedTensor(ctypes.Structure):
-  pass
+    pass
 
 
 DeleterFunc = ctypes.CFUNCTYPE(None, ctypes.POINTER(DLManagedTensor))
@@ -4164,7 +4164,47 @@ def dl_managed_tensor_deleter(dl_managed_tensor_handle):
     ctypes.pythonapi.Py_DecRef(pyobj)
 
 
-def from_numpy(array):
+def from_numpy(array, zero_copy=True):
+    """Returns an MXNet's NDArray backed by Numpy's ndarray.
+
+    Parameters
+    ----------
+    array: numpy.ndarray
+        input data
+
+    zero_copy: bool
+        Whether we use DLPack's zero-copy conversion to convert to MXNet's NDArray.
+        This is only available for c-contiguous arrays, i.e. array.flags[C_CONTIGUOUS] == True.
+
+    Returns
+    -------
+    NDArray
+        a NDArray backed by a dlpack tensor
+
+    Examples
+    --------
+    >>> x = mx.nd.ones((2,3))
+    >>> y = mx.nd.to_dlpack_for_read(x)
+    >>> type(y)
+    <class 'PyCapsule'>
+    >>> z = mx.nd.from_dlpack(y)
+    >>> type(z)
+    <class 'mxnet.ndarray.ndarray.NDArray'>
+    >>> z
+    [[ 1.  1.  1.]
+     [ 1.  1.  1.]]
+    <NDArray 2x3 @cpu(0)>
+
+    >>> w = mx.nd.to_dlpack_for_write(x)
+    >>> type(w)
+    <class 'PyCapsule'>
+    >>> u = mx.nd.from_dlpack(w)
+    >>> u += 1
+    >>> x
+    [[2. 2. 2.]
+     [2. 2. 2.]]
+    <NDArray 2x3 @cpu(0)>
+    """
 
     def make_manager_ctx(obj):
         pyobj = ctypes.py_object(obj)
@@ -4190,10 +4230,13 @@ def from_numpy(array):
         c_obj.deleter = dl_managed_tensor_deleter
         return c_obj
 
-    assert array.flags['C_CONTIGUOUS'], "We only support c-contiguous numpy arrays"
+    if not zero_copy:
+        return NDArray(handle=handle)
+
+    assert array.flags['C_CONTIGUOUS'], "Only c-contiguous arrays are supported for zero-copy"
     c_obj = make_dl_managed_tensor(array)
     address = ctypes.addressof(c_obj)
     address = ctypes.cast(address, ctypes.c_void_p)
     handle = NDArrayHandle()
     check_call(_LIB.MXNDArrayFromDLManagedTensor(address, ctypes.byref(handle)))
-    return NDArray(handle=handle)
+    return NDArray(array)
--- a/python/mxnet/ndarray/ndarray.py
+++ b/python/mxnet/ndarray/ndarray.py
@@ -4164,12 +4164,12 @@ def dl_managed_tensor_deleter(dl_managed_tensor_handle):
     ctypes.pythonapi.Py_DecRef(pyobj)
 
 
-def from_numpy(array, zero_copy=True):
+def from_numpy(ndarray, zero_copy=True):
     """Returns an MXNet's NDArray backed by Numpy's ndarray.
 
     Parameters
     ----------
-    array: numpy.ndarray
+    ndarray: numpy.ndarray
         input data
 
     zero_copy: bool
@@ -4181,29 +4181,6 @@ def from_numpy(array, zero_copy=True):
     NDArray
         a NDArray backed by a dlpack tensor
 
-    Examples
-    --------
-    >>> x = mx.nd.ones((2,3))
-    >>> y = mx.nd.to_dlpack_for_read(x)
-    >>> type(y)
-    <class 'PyCapsule'>
-    >>> z = mx.nd.from_dlpack(y)
-    >>> type(z)
-    <class 'mxnet.ndarray.ndarray.NDArray'>
-    >>> z
-    [[ 1.  1.  1.]
-     [ 1.  1.  1.]]
-    <NDArray 2x3 @cpu(0)>
-
-    >>> w = mx.nd.to_dlpack_for_write(x)
-    >>> type(w)
-    <class 'PyCapsule'>
-    >>> u = mx.nd.from_dlpack(w)
-    >>> u += 1
-    >>> x
-    [[2. 2. 2.]
-     [2. 2. 2.]]
-    <NDArray 2x3 @cpu(0)>
     """
 
     def make_manager_ctx(obj):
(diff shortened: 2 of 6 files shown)


junrushao1994:
@reminisce @wkcn Could you help review? Thanks!

python/mxnet/ndarray/ndarray.py
```diff
@@ -4115,3 +4115,105 @@ def from_dlpack(dlpack):
...
+        This is only available for c-contiguous arrays, i.e. array.flags[C_CONTIGUOUS] == True.
+
+    Returns
+    -------
+    NDArray
+        a NDArray backed by a dlpack tensor
+
+    """
+
+    def make_manager_ctx(obj):
```
reminisce: If this function is not user facing, better to make it private, i.e. `_make_managger_ctx`.

python/mxnet/ndarray/ndarray.py
```diff
@@ -4115,3 +4115,105 @@ def from_dlpack(dlpack):
...
+
+    def make_dl_managed_tensor(array):
+        c_obj = DLManagedTensor()
+        c_obj.dl_tensor = make_dl_tensor(array)
+        c_obj.manager_ctx = make_manager_ctx(array)
+        c_obj.deleter = dl_managed_tensor_deleter
+        return c_obj
+
+    if not zero_copy:
+        return array(ndarray)
```
reminisce: Should we pass `dtype=ndarray.dtype`?

python/mxnet/ndarray/ndarray.py
```diff
@@ -4115,3 +4115,105 @@ def from_dlpack(dlpack):
...
+
+    assert ndarray.flags['C_CONTIGUOUS'], "Only c-contiguous arrays are supported for zero-copy"
```
reminisce: Better to raise instead of assert.

python/mxnet/ndarray/ndarray.py
```diff
@@ -4115,3 +4115,105 @@ def from_dlpack(dlpack):
...
+
+    """
+
+    def make_manager_ctx(obj):
+        pyobj = ctypes.py_object(obj)
+        void_p = ctypes.c_void_p.from_buffer(pyobj)
+        ctypes.pythonapi.Py_IncRef(pyobj)
+        return void_p
+
+    def make_dl_tensor(array):
```
reminisce: Same as above.

python/mxnet/ndarray/ndarray.py
```diff
@@ -4115,3 +4115,105 @@ def from_dlpack(dlpack):
...
+        dl_tensor.data = array.ctypes.data_as(ctypes.c_void_p)
+        dl_tensor.ctx = DLContext(1, 0)
+        dl_tensor.ndim = array.ndim
+        dl_tensor.dtype = DLDataType.TYPE_MAP[str(array.dtype)]
+        dl_tensor.shape = array.ctypes.shape_as(ctypes.c_int64)
+        dl_tensor.strides = None
+        dl_tensor.byte_offset = 0
+        return dl_tensor
+
+    def make_dl_managed_tensor(array):
```
reminisce: Same as above.

tests/python/unittest/test_ndarray.py
```diff
@@ -1653,6 +1653,14 @@ def test_ndarray_nan_comparison():
     for i in (np.isnan(data1_grad))[1][0].flatten():
         assert i == True
 
+
+def test_zero_from_numpy():
+    print("test_zero_from_numpy")
```
reminisce: Remove `print`

tests/python/unittest/test_ndarray.py
```diff
@@ -1653,6 +1653,14 @@ def test_ndarray_nan_comparison():
     for i in (np.isnan(data1_grad))[1][0].flatten():
         assert i == True
 
+
+def test_zero_from_numpy():
```
reminisce: Can you add more test cases including
1. scalar tensors.
2. zero-size tensors.
3. `zero_copy=False`.
4. Perturbing the elements of numpy ndarray will equivalently affect mxnet NDArray, and vice versa.
5. Deleting either one of them should not obliterate the content of the other one when `zero_copy=True` and `False`.

python/mxnet/ndarray/ndarray.py
```diff
@@ -4115,3 +4115,105 @@ def from_dlpack(dlpack):
...
+    def make_dl_tensor(array):
+        dl_tensor = DLTensor()
+        dl_tensor.data = array.ctypes.data_as(ctypes.c_void_p)
+        dl_tensor.ctx = DLContext(1, 0)
+        dl_tensor.ndim = array.ndim
+        dl_tensor.dtype = DLDataType.TYPE_MAP[str(array.dtype)]
+        dl_tensor.shape = array.ctypes.shape_as(ctypes.c_int64)
+        dl_tensor.strides = None
+        dl_tensor.byte_offset = 0
+        return dl_tensor
```
reminisce: Do you need to increase the ref count of array?

python/mxnet/ndarray/ndarray.py
junrushao1994: There are asserts everywhere in MXNet's python package...Anyway, I will change it

python/mxnet/ndarray/ndarray.py
junrushao1994: No, this is rather a temporary stack variable

reminisce: Oh, I meant where incref for numpy object happened. NVM, just saw line 4189, missed reading that before.

python/mxnet/ndarray/ndarray.py
```diff
@@ -4115,3 +4115,105 @@ def from_dlpack(dlpack):
...
+        void_p = ctypes.c_void_p.from_buffer(pyobj)
+        ctypes.pythonapi.Py_IncRef(pyobj)
+        return void_p
+
+    def make_dl_tensor(array):
+        dl_tensor = DLTensor()
+        dl_tensor.data = array.ctypes.data_as(ctypes.c_void_p)
+        dl_tensor.ctx = DLContext(1, 0)
+        dl_tensor.ndim = array.ndim
+        dl_tensor.dtype = DLDataType.TYPE_MAP[str(array.dtype)]
```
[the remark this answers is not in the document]
junrushao1994: Good catch!

include/mxnet/c_api.h
```diff
@@ -823,6 +823,23 @@ MXNET_DLL int MXNDArrayToDLPack(NDArrayHandle handle,
...
+* allocated by an external deep learning framework
+* that is DLPack compatible.
+*
+* The memory is retained until the NDArray went out of scope.
+*
+* \param dlpack the pointer of the input DLManagedTensor
+* \param out_handle pointer holder to get pointer of NDArray
+* \return 0 when success, -1 when failure happens
+*/
+MXNET_DLL int MXNDArrayFromDLManagedTensor(DLManagedTensorHandle dlpack,
```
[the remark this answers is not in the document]
junrushao1994: The most important reason that I didn't call the original API is its overly-strong underlying assumption made me scared. Correct me if I was wrong, but I feel that it asks the DLManagedTensor to be alive throughout the entire lifetime of the external object. To be more specific, in the implementation of `NDArray::FromDLPack`, the deleter would assume the argument `tensor` always points to a valid address - this doesn't usually hold, especially for that in the frontend user may delete the numpy's ndarray. I was not aware why it is initially designed like this.

src/ndarray/ndarray.cc
```diff
@@ -365,6 +365,18 @@ NDArray NDArray::FromDLPack(const DLManagedTensor* tensor) {
   return NDArray(TBlob(dl_tensor), dl_tensor.ctx.device_id, deleter);
 }
 
+NDArray NDArray::FromDLManagedTensor(const DLManagedTensor* tensor) {
```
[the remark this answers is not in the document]
junrushao1994: No. The `FromDLManagerTensor ` does not require the address `tensor` points to to be alive.

include/mxnet/c_api.h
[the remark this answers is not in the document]
wkcn: I wrote the function FromDLPack, and it is my mistake. Could you please replace FromDLPack with FromDLManagedTensor, but keep the function name FromDLPack? Thank you!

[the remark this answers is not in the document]
junrushao1994: @wkcn No worries! I will try. Thanks!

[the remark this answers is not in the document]
junrushao1994: I tried several hours unifying these two implementations, but not able to resolve some core dumps in the unittest `test_ndarray.test_dlpack` because `manager_ctx` points to an invalid address (something like 0x6). Maybe we can do it later.

tests/python/unittest/test_ndarray.py
junrushao1994: Done

tests/python/unittest/test_ndarray.py
```diff
@@ -1653,6 +1653,14 @@ def test_ndarray_nan_comparison():
...
+    print("test_zero_from_numpy")
+    np_array = np.array([[1, 2], [3, 4], [5, 6]], dtype="float32")
+    mx_array = mx.nd.from_numpy(np_array)
+    print(mx_array)
```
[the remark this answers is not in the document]
junrushao1994: Done. Thanks :-)

junrushao1994:
@reminisce Hey I have already addressed your comments. Could you review again? Thanks!

include/mxnet/c_api.h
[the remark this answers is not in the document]
junrushao1994: Done per discussion with @wkcn. Now we unify `FromDLPack ` and `FromDLManagedTensor `. Thank you so much!

wkcn:
Merged. Thank you!

python/mxnet/ndarray/ndarray.py
```diff
@@ -4115,3 +4115,108 @@ def from_dlpack(dlpack):
...
+        c_obj.manager_ctx = _make_manager_ctx(array)
+        c_obj.deleter = dl_managed_tensor_deleter
+        return c_obj
+
+    if not zero_copy:
+        return array(ndarray, dtype=ndarray.dtype)
+
+    if not ndarray.flags['C_CONTIGUOUS']:
+        raise ValueError("Only c-contiguous arrays are supported for zero-copy")
+    c_obj = _make_dl_managed_tensor(ndarray)
```
szha: when doing zero-copy, should the ownership of the data pointer be transferred to the new ndarray? do we need to update the writable flag? what happens when the original data is updated by numpy?

junrushao1994: @szha They share the ownership - which means numpy's modification is transparent to mxnet and vice versa. As requested by @reminisce in his review, I added a testcase for this transparency.

szha: how does it work when using the asynchronous engine? since numpy calls happen immediately, allowing shared ownership may give unexpected results. consider the following case:
```
a = np.array(...)
b = nd.from_numpy(a, zero_copy=True)
c = nd.expensive_op(b, in_place=True)
d = np.inplace_op(a) # this is called when c hasn't finished yet
```
given the asynchronous execution, it's hard to tell whether people should expect d to be based on the input before or after `expensive_op`

junrushao1994: (We actually have such discussion long time before, but it's good to make things open and achievable to the community)

My understanding is that your concern is a general question in multi-threading (see the code below). In any multi-threaded system, we may expect this issue to exist -- and MXNet has multi-threaded engine, so this issue exists, right?

We have discussion that how about the content of array is changed somewhere inside the system but user don't know that, should we set up the `WRITEABLE` flag to be true to prevent users from writing, or should we completely disallow users to access this numpy array?

The question, in the high-level, is all about trade-off. Restricting the freedom of users may make things less error-prone, while giving users full freedom may make the system more powerful.

On one hand, in my humble option, restricting the freedom of users, like setting up `WRITEABLE` flag, does not completely resolve this issue, because users can still encounter pitfalls when they read the array using the numpy side API. The only helps in this case is to completely disable users from read or write the array, which grant them on access only the numpy side - This is more like guaranteeing safety by disallowing users to create threads/processes.

On the other hand, giving users freedom enables them to do more things, and does not necessarily mean that they will definitely make a mistake, because we will encourage to use mxnet's numpy compatibility, which is on the way.

Anyway, I am open to any discussion and open to possible changes setting up several flags on the numpy array.

szha: The API should have clear semantics regardless of users’ knowledge on how MXNet execution works. If you worry about flexibility I’d recommend making it an option to not disable writable flag.

junrushao1994: No. The point is that writable flag doesn’t address the issue at all.

szha: I think completely transferring the ownership from numpy and disabling access to the original array is preferred. If we are to expose numpy array for joint use, it should be done in asnumpy as another zero-copy creation feature, so that the wait_to_read can be done by us

wkcn: Maybe we can add a new bool argument for `from_numpy`. The argument is `True` by default, and it disables access to the original NumPy array. When users know the asychronous risk, they can set this arugument to `False`, in order to access to the original NumPy array.

szha: Once we add a zero-copy option to asnumpy where the ownership is completely transferred back to numpy, I think there wouldn't be any need for co-ownership.

wkcn: Consider the following case:
```python
# the variable `a` is a `mx.nd.NDArray` object
a = mx.nd.array([1,2,3])
b = a.asnumpy(zero_copy=True)
#  users call `asnumpy(zero_copy=True)` twice
c = a.asnumpy(zero_copy=True)
c[:] = 3
print(a)
print(b)

a[:] = 5
a.wait_to_read()
print(b)
print(c)
```
It's necessary to keep co-ownership.

szha: If zero-copy is on, then `a` should not be usable after the `asnumpy` call. The point is to let the  framework deal with all synchronization, and co-ownership doesn't allow that. And for the same case, you can simply copy `c` from `b` (the new numpy array) instead of `a` (the old ndarray), so it's not necessary to keep co-ownership.

szha: The hook in the context seems to refer to our own subclass of numpy array. Creating a new subclass seems like overkill.

leezu: For reference, @szha's suggestion was addressed in https://github.com/apache/incubator-mxnet/pull/14948

PR #14703, opened 2019-04-15
Set idx2name for Optimizer object

yuxihu:
Currently when we use `module.fit()`, we can pass a str or an Optimizer object as the [optimizer](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/module/base_module.py#L440). When we pass an Optimizer object that is created outside before calling `module.fit()`, we are not checking and setting `optimizer.idx2name` which is used to mask out entries in update for weight decay in [module.init_optimizer](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/module/module.py#L522) method. If `idx2name` is not set when creating the optimizer object outside, it can impact the convergence behavior as weight decay is applied to all parameters in this case.

This PR checks if `idx2name` is set if an Optimizer object is given. If not, we are setting it to the proper values. Fixes #9511 as @wkcn pointed out.

PR #14772, opened 2019-04-22
[v1.4.x] Set idx2name for Optimizer object (#14703)

yuxihu:
Cherry-pick the change that fixes issue affecting model convergence when using Module API to v1.4.x branch.


yuxihu:
@junrushao1994 @szha Please help review and merge. Thanks.

PR #14726, opened 2019-04-17
[MXNET-1377] Add static-dependencies licenses

lanking520:
## Description ##
This PR is to add licences for all static-dependencies. These licenese are not applicable to user who chose to build from source with all depdendencies dynamically-linked. But static-linking are required to place it under the generated package (e.g pip wheel, maven jars).
@zachgk @szha 

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

PR #14778, opened 2019-04-23
Updates python setup.py for recent license changes

perdasilva:
## Description ##
A recent PR #14726 has altered the way the static build deals with licenses. This has broken the `setup.py` script for building the release pip packages. This PR fixes this problem.

I've re-added some of the removed licenses, this should be confirmed to be OK by the submitters of the aforementioned PR. @lanking520 could you please take a look?

### DIFF
--- a/tools/pip/setup.py
+++ b/tools/pip/setup.py
@@ -66,12 +66,8 @@ shutil.copytree(os.path.join(CURRENT_DIR, 'mxnet-build/3rdparty/dmlc-core/tracke
 shutil.copy(LIB_PATH[0], os.path.join(CURRENT_DIR, 'mxnet'))
 
 # copy license and notice
-shutil.copy(os.path.join(CURRENT_DIR, 'mxnet-build/LICENSE'),
-            os.path.join(CURRENT_DIR, 'mxnet/LICENSE'))
-shutil.copy(os.path.join(CURRENT_DIR, 'mxnet-build/DISCLAIMER'),
-            os.path.join(CURRENT_DIR, 'mxnet/DISCLAIMER'))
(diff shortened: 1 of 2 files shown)


tools/staticbuild/build_lib.sh
```diff
@@ -34,6 +34,7 @@ $MAKE DEPS_PATH=$DEPS_PATH $PWD/3rdparty/tvm/nnvm/lib/libnnvm.a
 $MAKE DEPS_PATH=$DEPS_PATH PSLITE
 
 if [[ $VARIANT == *mkl ]]; then
+    MKLDNN_LICENSE='license.txt'
```
zachgk: These licenses are included as part of the source LICENSE. There is no need to re-add them

tools/pip/setup.py
```diff
@@ -66,12 +66,8 @@ def has_ext_modules(self):
 shutil.copy(LIB_PATH[0], os.path.join(CURRENT_DIR, 'mxnet'))
 
 # copy license and notice
-shutil.copy(os.path.join(CURRENT_DIR, 'mxnet-build/LICENSE'),
-            os.path.join(CURRENT_DIR, 'mxnet/LICENSE'))
-shutil.copy(os.path.join(CURRENT_DIR, 'mxnet-build/DISCLAIMER'),
```
zachgk: We do not currently add this DISCLAIMER. Can you add it to the build similarly to https://github.com/apache/incubator-mxnet/pull/14726/files#diff-2c9b07f53d24caa1ac38f1eb9c371118R70?

perdasilva:
@zachgk I've addressed your comments ^^ thanks for taking a look!

lanking520:
@szha

Issue #14619, opened 2019-04-04
[Discussion] 1.5.0 Roadmap


szha:
Let's start a discussion here about the roadmap towards 1.5.0. We are looking for:

- New features that are useful to your research and development.
- Improvements and patches to existing features.

If you have any item that you'd like to propose to have in the roadmap, please do:

- Create (or locate existing) issue/pull request for the item, note the issue/pull request number.
- Comment in this issue: 1) the above issue number, 2) one sentence of what the item is about and why it's useful to you.
- Indicate whether you'd be willing to help out on the item.
- Share the ETA if you're driving the item and have an guesstimate on when it will be done.

cc @apache/mxnet-committers

szha:
The changes since 1.4.0 release that are already merged in the master branch will be included in the 1.5.0 release. The list can be found at: https://github.com/apache/incubator-mxnet/compare/v1.4.x...master?expand=1

eric-haibin-lin:
Hi everyone, I've created v1.5.x branch here: https://github.com/apache/incubator-mxnet/tree/v1.5.x 
Before we have an agreement on the timeline and features, I will synchronize this branch with the master branch periodically. Once we have decided the code freeze day, we will only cherry-pick required changes/features to the branch then.

anirudh2290:
Thanks for starting this! I would like to include exception handling fixes: #14397 (@anirudh2290), #14433(@anirudh2290) , #14575 (@arcadiaphy). These three should be merged by end of next week hopefully. Conversion of FP32 models to mixed precision models (#14584) (Should be in by May first week tentatively). In addition, I have some changes to profiler to visualize gpu memory pooling and help make better decisions on the env variable choice. It is currently in a branch (https://github.com/anirudh2290/mxnet/tree/memory_profiler_poc2) and intend to open a PR soon (next week).

pengzhao-intel:
MKLDNN Quantization PR


sum | #14614
-- | --
relu | #14604
refactor requantize | #14608
conv + activation | WIP
cache op | WIP
SSD COCO model | WIP

zboldyga:
Some users pointed out useful features around matrix inversions, determinants, log determinants. I propose to add some small features to make these calculations easier: https://issues.apache.org/jira/projects/MXNET/issues/MXNET-1350?filter=allissues .

https://github.com/apache/incubator-mxnet/issues/14360

[...]

Issue #10002, opened 2018-03-06
_backward_Convolution is non-differentiable


lightingghost:
As I saw mxnet has `autograd` package to support high order gradients, I tried to implement wgan-gp with mxnet. But I got an error of 

```
mxnet.base.MXNetError: [08:32:17] C:\projects\mxnet-distro-win\mxnet-build\nnvm\src\pass\gradient.cc:187: Operator _backward_Convolution is non-differentiable because it didn't register FGradient attribute.
```

It seems convolution operator still does not support higher order gradients?

sxjscience:
Supporting higher order gradients is on the agenda. Refer to the discussion in https://github.com/apache/incubator-mxnet/issues/5699. Need to ping Eric @piiswrong

sxjscience:
I find the issue is similar as https://github.com/apache/incubator-mxnet/issues/9979 . Let's move the discussion there.

lightingghost:
@sxjscience I think #9979 is a question rather than a feature request. In his case, he should use `autograd.grad` and set `create_graph=True` to solve the problem. However, in this case, it seems second order derivative is not implemented for some operators?

piiswrong:
2nd order gradient is only implemented for a few operators like * and exp.

aidan-plenert-macdonald:
I looked at the code and it seems like the problem is that the current implementation is kind of only doing symbolic math down one level. For example, [sin](https://github.com/apache/incubator-mxnet/blob/master/src/operator/tensor/elemwise_unary_op_trig.cc#L32),

```c++
// sin                                                                                                                                                                                                      
MXNET_OPERATOR_REGISTER_UNARY_WITH_RSP(sin, cpu, mshadow_op::sin)
MXNET_ADD_SPARSE_OP_ALIAS(sin)
.describe(R"code(Computes the element-wise sine of the input array.                                                                                                                                         
                                                                                                                                                                                                            
The input should be in radians (:math:`2\pi` rad equals 360 degrees).                                                                                                                                       
                                                                                                                                                                                                            
.. math::                                                                                                                                                                                                   
   sin([0, \pi/4, \pi/2]) = [0, 0.707, 1]                                                                                                                                                                   
                                                                                                                                                                                                            
The storage type of ``sin`` output depends upon the input storage type:                                                                                                                                     
                                                                                                                                                                                                            
   - sin(default) = default                                                                                                                                                                                 
   - sin(row_sparse) = row_sparse                                                                                                                                                                           
[...]

PR #14779, opened 2019-04-23
[MXNET-978] Fully connected, higher order grad

### LINKED ISSUE #14991: Second order gradient wrt inputs, expected behaviour.
What would be the expected behaviour of this code?

It tries to calculate the gradient of a function using the gradient wrt of the inputs of the first gradient.

```
def test_ag_grad():
    x = mx.nd.ones((3,3))
    y = mx.nd.ones((3,3))
    x.attach_grad()
    y.attach_grad()
    with mx.autograd.record():
        z = x + y
        x_grad_y_grad = mx.autograd.grad(z, [x,y], create_graph=True, retain_graph=True)
        print(x_grad_y_grad)
        first_grad = nd.concat(*[x.reshape(-1) for x in x_grad_y_grad], dim=0)
        fg_f = 2 * first_grad
        second_grad = mx.autograd.grad(fg_f, [x,y], retain_graph=True)
```

larroy:
## Description ##
Add backward for fully connected backward.

This is not strictly correct if X or W depend on each other. After discussing with the community this a valid assumption that we can make (that the weights don't depend on the inputs).

This PR adds functionality required to get second order gradients of models using FullyConnected operator such as simple MLP.

https://github.com/apache/incubator-mxnet/issues/10002
https://github.com/apache/incubator-mxnet/issues/14619
Resolves #14991 

### Essentials ###


larroy:
@mxnet-label-bot add [Autograd, Backend, pr-work-in-progress, pr-awaiting-review]

apeforest:
Can you add unit test?

src/operator/nn/fully_connected-inl.h
```diff
@@ -237,6 +237,7 @@ void FullyConnectedGradCompute(const nnvm::NodeAttrs& attrs,
   }
 }
 
+
```
apeforest: why blank line here?

src/operator/nn/fully_connected.cc
```diff
@@ -1,4 +1,4 @@
-/*
+  /*
```
apeforest: indent

src/operator/nn/fully_connected.cc
```diff
@@ -165,6 +165,7 @@ static bool FullyConnectedType(const nnvm::NodeAttrs& attrs,
       attrs, in_type, out_type, -1);
 }
 
+
```
apeforest: why new line?

src/operator/nn/fully_connected.cc
```diff
@@ -176,6 +177,16 @@ struct FullyConnectedGrad {
   }
 };
 
+
+std::vector<nnvm::NodeEntry> FullyConnectedBackwardGrad(const nnvm::NodePtr& n,
+      const std::vector<nnvm::NodeEntry>& ograds) {
```
apeforest: indent

src/operator/nn/fully_connected.cc
```diff
@@ -310,6 +321,7 @@ If ``no_bias`` is set to be true, then the ``bias`` term is ignored.
 .add_argument("bias", "NDArray-or-Symbol", "Bias parameter.")
 .add_arguments(FullyConnectedParam::__FIELDS__());
 
+
```
apeforest: remove new line

src/operator/nn/fully_connected.cc
```diff
@@ -333,5 +346,6 @@ NNVM_REGISTER_OP(_backward_FullyConnected)
 #endif
 .set_attr<FCompute>("FCompute<cpu>", FullyConnectedGradCompute<cpu>);
 
+
```
apeforest: remove new line

apeforest:
@larroy Thanks for the contribution. Could you please add unit test and change the title if it's ready for review?

src/operator/nn/fully_connected.cc
larroy: style guide says it's ok 1 or 2 blank lines around functions, more consistent with python as we use two blank lines between functions according to pep8.

src/operator/nn/fully_connected.cc
larroy: for readability?

src/operator/nn/fully_connected.cc
apeforest: I know pep8 has this rule for python, but does it also apply to C++?

apeforest:
Why are there changes in submodules?

src/operator/nn/fully_connected.cc
larroy: I think is nice, and google style allows up to two newlines, so I don't know why are we discussing this. If it makes you so unhappy, I remove the newline. For me it  helps readability.

larroy:
Ignore submodules, I guess I was out of date. I will sync them on the next commits.

larroy:
@apeforest @kshitij12345

larroy:

larroy:
Working on fixing this, added the differential calculus developments for review and reference:
image-b8b4fdcb-128a-462c-909d-4438a3639c3a

tests/python/unittest/test_gluon.py
```diff
@@ -915,6 +915,24 @@ def test_sequential_warning():
         assert len(w) == 1
 
 
+@with_seed()
+def test_dense_backward():
```
kshitij12345: I guess rename the `test_dense_double_backward` or similar?

tests/python/unittest/test_gluon.py
```diff
@@ -915,6 +915,24 @@ def test_sequential_warning():
...
+    x = nd.array([[1,2,3,400]])
+    net = gluon.nn.Sequential()
+    with net.name_scope():
+        net.add(gluon.nn.Dense(1, in_units=x.shape[1]))
+    net.initialize(mx.initializer.Constant(.5))
+    params = [p.data() for p in net.collect_params().values()]
+    x.attach_grad()
+    with ag.record():
+        y = net.forward(x)
+        y_grad = ag.grad(y, x, create_graph=True, retain_graph=True)[0]
```
kshitij12345: As per our discussion in #15120 , this should be `x_grad`.

tests/python/unittest/test_gluon.py
apeforest: Can we also move this test to test_higher_order.py so it's easier to track higher order supported operators?

larroy: Where is this file?

larroy:
@apeforest @kshitij12345 is this good to merge?

larroy:
Don't merge yet. I need to validate the result I have some doubts now.

apeforest:
@larroy Could you please paste your PyTorch script (and result) here for comparison purpose?

apeforest:
@kshitij12345 We initially thought the second order gradient for FullyConnected (which is a simple linear transformation in MXNet) is constant zeros because X and W are usually independant tensors. Is this correct? Please suggest. Thanks.

larroy:
```
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
"""Description"""

__author__ = 'Pedro Larroy'
__version__ = '0.1'

import os
import sys
import subprocess

def main():
    import torch
    import torch.autograd as ag
    model = torch.nn.Sequential(
        torch.nn.Linear(3, 1))
    params = [x for x in model.parameters()]
    params[0].data = torch.FloatTensor([[1,2,3]])
    #x = torch.rand(5, 3)
    x = torch.FloatTensor([[0.2131, 0.5449, 0.9910],
       [0.1600, 0.1665, 0.1387],
       [0.6242, 0.0409, 0.0663],
       [0.6590, 0.9822, 0.3108],
       [0.8566, 0.3848, 0.8385]])
    x.requires_grad_(True)
    y=torch.sum(model(x))

    #w = torch.FloatTensor([1,2,3])
    #w.requires_grad_(True)
    #with torch.no_grad():
    #    model[0].weight.copy_(w)
    w=list(model.parameters())[0]
    b=list(model.parameters())[1]
    print(f"x: {x}")
    print(f"w: {w}")
    print(f"b: {b}")
    w_grad = ag.grad(y, w, create_graph=True, retain_graph=True)[0]
    w_grad_grad = ag.grad(torch.sum(w_grad), w, allow_unused=True)
    print(f"w_grad_grad {w_grad_grad}")


    params = [p for p in model.parameters()]
    params_grad = ag.grad(y, params, create_graph=True, retain_graph=True)
    print(f"params_grad {params_grad}")

    reduced_params_grad = torch.sum(torch.stack([torch.sum(p) for p in params_grad]))
    print(reduced_params_grad)
    # ? need scalar input
    params_grad_grad = ag.grad(reduced_params_grad, params, allow_unused=True)
    print(f"params_grad_grad {params_grad_grad}")
    return 0


    params_grad_flat = torch.cat([x.view(-1) for x in params_grad]).unsqueeze(0).requires_grad_(True)
    params_grad_flat_norm2 = params_grad_flat @ params_grad_flat.t()
    params_grad_grad = ag.grad(params_grad_flat_norm2, params, allow_unused=True)
    #params_grad_grad = ag.grad(params_grad_flat_norm2, params)
    w_grad = ag.grad(y, w, create_graph=True, retain_graph=True)[0].squeeze(0)
    xsum = torch.sum(x,0)
    print(f"torch.sum(x,0): {xsum}")
    print(f"w_grad: {w_grad}")
    w_grad_sum = torch.sum(w_grad, 0)
    print(f"sum(w_grad, 0) {w_grad_sum}")
    assert torch.all(torch.eq(w_grad, xsum))
    x_grad_w_grad = ag.grad(w_grad[1], x, grad_outputs=torch.FloatTensor([1,2,3]), allow_unused=True)[0]
    print(f"x_grad_w_grad: {x_grad_w_grad}")
    x_grad = ag.grad(y, x, create_graph=True, retain_graph=True)[0]
    print(f"x_grad: {x_grad}")
    x_grad_w_grad = ag.grad(x_grad[0], w, grad_outputs=torch.FloatTensor([1,2,1]), allow_unused=True)
    #x_grad_w_grad = ag.grad(x_grad[0][0], w, grad_outputs=torch.FloatTensor(1), allow_unused=True)
    print(f"x_grad_w_grad: {x_grad_w_grad}")
    w_grad_grad = ag.grad(torch.sum(w), w, allow_unused=True)
    print(f"w_grad_grad {w_grad_grad}")
    x_grad_grad = ag.grad(torch.sum(x), )

    return 0

if __name__ == '__main__':
    sys.exit(main())
```

larroy:
second order gradient in torch seems to be zeros as well. Need to verify the same in this PR.

larroy:
I think this is good to go @apeforest

kshitij12345:
Even I believe it should be zeros of X and W.
This is the computation that leads me to believe it.
image

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -129,6 +131,44 @@ def check_second_order_unary(x, op, grad_grad_op):
...
+
+
+@with_seed()
+def test_dense_backward():
+    import mxnet.autograd as ag
+    import mxnet.ndarray as nd
+    for x in RandomShapes(5):
+        net = gluon.nn.Sequential()
+        with net.name_scope():
+            #net.add(gluon.nn.Dense(1, in_units=x.shape[1]))
```
apeforest: please remove commented line

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -129,6 +131,44 @@ def check_second_order_unary(x, op, grad_grad_op):
...
+        shape = rand_shape_nd(self.curdim)
+        print(shape)
+        x = nd.random.normal(shape=shape)
+        self.curdim += 1
+        return x
+
+
+@with_seed()
+def test_dense_backward():
+    import mxnet.autograd as ag
```
apeforest: Move this import to file head

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -129,6 +131,44 @@ def check_second_order_unary(x, op, grad_grad_op):
...
+            raise StopIteration
+        shape = rand_shape_nd(self.curdim)
+        print(shape)
+        x = nd.random.normal(shape=shape)
+        self.curdim += 1
+        return x
+
+
+@with_seed()
+def test_dense_backward():
```
apeforest: Can we add a test for FullyConnected operator itself?

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -129,6 +131,44 @@ def check_second_order_unary(x, op, grad_grad_op):
...
+    for x in RandomShapes(5):
+        net = gluon.nn.Sequential()
+        with net.name_scope():
+            #net.add(gluon.nn.Dense(1, in_units=x.shape[1]))
+            net.add(gluon.nn.Dense(1))
+        net.initialize(mxnet.initializer.Constant(.5))
+        x.attach_grad()
+        with ag.record():
+            y = net.forward(x)
+            x_grad = ag.grad(y, x, create_graph=True, retain_graph=True)[0]
```
apeforest: please explicit set argument keyword like other tests for clarity

sxjscience: We'd better do something like the following to check the correctness.
```python
        with ag.record():
            y = net.forward(x)
            x_grad = ag.grad(y, x, create_graph=True, retain_graph=True)[0]
            z = (random_multiplier * x_grad).sum()
        z.backward()
```
Then we check whether `x.grad` is correct.

apeforest:
@larroy Please update your PR description.

tests/python/unittest/test_higher_order_grad.py
larroy: Dense calls FullyConnected during forward, do you suggest to use the symbolic API?

apeforest: Since your change in C++ is directly on the FullyConnected operator, wouldn't it make sense to test on it directly?

larroy: Makes sense, but for convenience I prefer to test with Gluon. I don't think is worth the complexity and extra work now. Let me know if you disagree.

larroy:
@apeforest @kshitij12345 @sxjscience thanks for the review. Is this good to go?

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -18,8 +18,14 @@
 
 import math
 from mxnet import nd, autograd
-from mxnet.test_utils import assert_almost_equal, random_arrays, rand_shape_nd
+from mxnet.test_utils import assert_almost_equal, random_arrays, rand_shape_nd, same
 from common import with_seed
+import mxnet.autograd as ag
+import mxnet.ndarray as nd
```
apeforest: many of these imports are duplicated. Could you please keep the required ones?

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -18,8 +18,14 @@
...
+from mxnet import gluon
+import mxnet
+from nose.tools import ok_
```
apeforest: this module should be imported before mxnet module

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -129,6 +135,83 @@ def check_second_order_unary(x, op, grad_grad_op):
...
+
+        with ag.record():
+            y = net.forward(x)
+            x_grad = ag.grad(heads=y, variables=x, create_graph=True, retain_graph=True)[0]
+            random_multiplier = nd.random.uniform_like(x_grad)
+            z = (random_multiplier * x_grad).sum()
+        z.backward()
+        same(x.grad, nd.zeros(4))
+
+        with ag.record():
```
apeforest: What are you trying to test here?

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -129,6 +135,83 @@ def check_second_order_unary(x, op, grad_grad_op):
...
+
+        w = list(net.collect_params().values())[0].data()
+        with ag.record():
+            y = net.forward(x)
+            w_grad_0 = ag.grad(heads=y, variables=w, create_graph=True, retain_graph=True)[0]
+        w_grad_0.backward()
+        w_grad_grad_0 = w.grad
+
+        x = x + nd.ones_like(x) * 0.01
+        with ag.record():
```
apeforest: What are you trying to test here? Maybe break the `with` blocks in this method with multiple methods so we know the purpose of each test?

tests/python/unittest/test_higher_order_grad.py
apeforest: The reason I was suggesting to test FullyConnected operator itself is then you could reuse the `check_second_order_unary` utility methods like all the other operators do. Having to write ad hoc test for each operator is not scalable and error prone.

sxjscience:
@larroy You may also refer to Line-7 of Algorithm 1 in https://arxiv.org/pdf/1704.00028.pdf . When implementing the gradient penalty, we should compute something like:

tests/python/unittest/test_higher_order_grad.py
larroy: numerical gradient, changing w and verifying that the first gradient changes but not the second one.

larroy:
Thanks a lot for your help @sxjscience. I will refine this PR in light of your feedback.

larroy:
Adding development for review:
IMG_20190711_123355__01

larroy:
The development is only for flattened inputs.

larroy:
Let me know if you have questions about notation or the development or see some mistakes.

tests/python/unittest/test_higher_order_grad.py
larroy: I believe you, but can you elaborate?

src/operator/nn/fully_connected-inl.h
```diff
@@ -1,4 +1,4 @@
-/*
+  /*
```
apeforest: indent

larroy: thanks, often happens to me due to vim bindings.

larroy:
From my side is just missing the bias term, please help review the approach.

larroy:
@sxjscience @kshitij12345 please help review.

src/operator/nn/fully_connected-inl.h
```diff
@@ -47,7 +48,23 @@ namespace fullc {
 enum FullyConnectedOpInputs {kData, kWeight, kBias};
 enum FullyConnectedOpResource {kTempSpace};
 enum FullyConnectedOpOutputs {kOut};
-}  // fullc
+enum FullyConnectedGradGradOutputs {
+  k_o_y_grad,
```
apeforest: please use the same camelCase to be consistent.

src/operator/nn/fully_connected-inl.h
```diff
@@ -91,21 +140,14 @@ void FCForward(const OpContext &ctx, const FullyConnectedParam &param,
   CHECK_EQ(s->blas_handle_ownership_, Stream<xpu>::OwnHandle)
       << "Must init CuBLAS handle in stream";
 #endif  // __CUDACC__
-  const mxnet::TShape& ishape = in_data[fullc::kData].shape_;
```
apeforest: Please don't refactor this as it is not related to this PR.

src/operator/nn/fully_connected-inl.h
```diff
@@ -126,6 +168,7 @@ void FCForward(const OpContext &ctx, const FullyConnectedParam &param,
   }
 }
 
+
```
apeforest: remove blank line

src/operator/nn/fully_connected-inl.h
```diff
@@ -134,59 +177,52 @@ void FCBackward(const OpContext &ctx, const FullyConnectedParam &param,
   using namespace mshadow::expr;
   // TODO(bing): check the BLAS Handle, be careful
   //  maybe need blas handle from context
-  Stream<xpu> *s = ctx.get_stream<xpu>();
-  const mxnet::TShape& ishape = in_data[fullc::kData].shape_;
-  const mxnet::TShape& oshape = out_grad[fullc::kOut].shape_;
-
-  Tensor<xpu, 2, DType> wmat = in_data[fullc::kWeight].get<xpu, 2, DType>(s);
-  Tensor<xpu, 2, DType> data, grad, gdata;
+  Stream<xpu> *stream = ctx.get_stream<xpu>();
```
apeforest: Please do not refactor in this PR.

src/operator/nn/fully_connected-inl.h
larroy: It's related I created a function that does that same thing because it's going to use in the two places. It's code duplication.

src/operator/nn/fully_connected-inl.h
```diff
@@ -249,6 +285,99 @@ void FullyConnectedGradCompute(const nnvm::NodeAttrs& attrs,
...
+  linalg_gemm(o_y, o_w_grad, x_grad_grad, false, false, stream);
+  linalg_gemm(o_y, o_x_grad, w_grad_grad, true, false, stream);
+  if (! param.no_bias) {
+    //  TODO(larroy)
+  }
+}
+
+
+template<typename xpu>
+void FullyConnectedGradGradDtypeDispatch(
```
apeforest: nit: DType?

src/operator/nn/fully_connected.cc
```diff
@@ -1,4 +1,4 @@
-/*
+    /*
```
apeforest: ??

src/operator/nn/fully_connected.cc
```diff
@@ -176,11 +176,23 @@ struct FullyConnectedGrad {
...
+  std::vector<nnvm::NodeEntry> operator()(const nnvm::NodePtr& n,
+                                          const std::vector<nnvm::NodeEntry>& ograds) const {
+    std::vector<nnvm::NodeEntry> heads(ograds.begin(), ograds.end());
+    heads.push_back(n->inputs[0]);  // o_y : head gradient of the output y
+    return MakeGradNode(op_name, n, heads, n->attrs.dict);
+  }
+};
+
+
+static bool FCStorageType(const nnvm::NodeAttrs& attrs,
```
apeforest: Why remove inline?

src/operator/nn/fully_connected.cc
```diff
@@ -210,11 +222,11 @@ inline static bool FCStorageType(const nnvm::NodeAttrs& attrs,
   return dispatched;
 }
 
-inline static bool BackwardFCStorageType(const nnvm::NodeAttrs& attrs,
-                                         const int dev_mask,
-                                         DispatchMode* dispatch_mode,
-                                         std::vector<int> *in_attrs,
-                                         std::vector<int> *out_attrs) {
+static bool BackwardFCStorageType(const nnvm::NodeAttrs& attrs,
```
apeforest: Why remove inline?

src/operator/nn/fully_connected.cc
```diff
@@ -240,6 +252,8 @@ inline static bool BackwardFCStorageType(const nnvm::NodeAttrs& attrs,
   return dispatched;
 }
 
+
```
apeforest: ??

src/operator/nn/fully_connected-inl.h
larroy: replied above.

src/operator/nn/fully_connected.cc
```diff
@@ -331,5 +346,32 @@ NNVM_REGISTER_OP(_backward_FullyConnected)
 #endif
 .set_attr<FCompute>("FCompute<cpu>", FullyConnectedGradCompute<cpu>);
 
+
```
apeforest: ??

src/operator/nn/fully_connected.cc
```diff
@@ -331,5 +346,32 @@ NNVM_REGISTER_OP(_backward_FullyConnected)
...
+// Higher order gradient for fully connected
```
apeforest: This is only second order as it cannot support higher order without using computation graph.

src/operator/nn/fully_connected.cc
```diff
@@ -331,5 +346,32 @@ NNVM_REGISTER_OP(_backward_FullyConnected)
...
+  const FullyConnectedParam& params = nnvm::get<FullyConnectedParam>(attrs.parsed);
+  return params.no_bias ? 3 : 4;
+})
+.set_num_outputs([](const NodeAttrs& attrs) {
+  return 3;
+})
+.set_attr<nnvm::TIsBackward>("TIsBackward", true)
+.set_attr_parser(ParamParser<FullyConnectedParam>)
+.set_attr<FCompute>("FCompute<cpu>", FullyConnectedGradGradDtypeDispatch<cpu>);
+
```
apeforest: remove extra line

tests/python/unittest/test_gluon.py
```diff
@@ -21,7 +21,7 @@
 import mxnet as mx
 from mxnet import gluon
 from mxnet.gluon import nn
-from mxnet.test_utils import assert_almost_equal
+from mxnet.test_utils import assert_almost_equal, same
```
apeforest: why need?

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -18,8 +18,17 @@
 
 import math
 from mxnet import nd, autograd
-from mxnet.test_utils import assert_almost_equal, random_arrays, rand_shape_nd
+from mxnet.test_utils import assert_almost_equal, random_arrays, rand_shape_nd, same
 from common import with_seed
+import mxnet.autograd as ag
```
apeforest: already imported

tests/python/unittest/test_higher_order_grad.py
apeforest: See https://github.com/apache/incubator-mxnet/pull/14779/files#diff-5ab121a4f290c534d444064283a6e35fR20

src/operator/nn/fully_connected.cc
larroy: sorry, vim bindings, often happens by accident.

tests/python/unittest/test_higher_order_grad.py
apeforest: https://www.python.org/dev/peps/pep-0008/#imports

Imports should be grouped in the following order:
Standard library imports.
Related third party imports.
Local application/library specific imports.
You should put a blank line between each group of imports.

src/operator/nn/fully_connected.cc
larroy: I have explained this before countless times. This is just code bloat, the function has several calls and a loop inside, inline does nothing good here. The compiler knows when to inline.

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -173,6 +182,87 @@ def check_second_order_unary(x, op, grad_grad_op):
     # Validate the gradients.
     assert_almost_equal(expected_grad_grad, x.grad.asnumpy())
 
+def arange_shape_like(y):
+    shape = y.shape
+    nelems = reduce(mul, shape)
+    x = nd.arange(nelems).reshape(shape)
+    return x
+
+class RandomShapes(object):
```
apeforest: I feel this class is not necessary.

We already have:

```
shape = rand_shape_nd(dim)
```

larroy:
Thanks for the comments about the whitespace. Is the calculation correct? Let's focus on the important stuff first, then do a round of beautification once we know the implementation is correct and the testing strategy is sound.

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -173,6 +182,87 @@ def check_second_order_unary(x, op, grad_grad_op):
...
+
+
+@with_seed()
+def test_dense_backward_flatten():
+    for x in RandomShapes(4,2):
+        hidden = random.randrange(1, 4)
+        net = gluon.nn.Sequential()
+        with net.name_scope():
+            net.add(gluon.nn.Dense(hidden, flatten=True))
+        net.initialize(mxnet.initializer.Constant(.5))
```
apeforest: use random initialize?

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -173,6 +182,87 @@ def check_second_order_unary(x, op, grad_grad_op):
...
+        x.attach_grad()
+        with ag.record():
+            y = net.forward(x)
+            o_y = arange_shape_like(y)  # head gradient of y
+            params = [p.data() for p in net.collect_params().values()]
+            w = params[0]
+            b = params[1]
+
+            # print(params)
```
apeforest: remove commented code

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -173,6 +182,87 @@ def check_second_order_unary(x, op, grad_grad_op):
...
+            x_grad = ag.grad(heads=y, variables=x, head_grads=o_y,
+                             create_graph=True, retain_graph=True)[0]
+            o_x_grad = arange_shape_like(x_grad)
+            #x_grad.attach_grad()
```
apeforest: remove commented code

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -173,6 +182,87 @@ def check_second_order_unary(x, op, grad_grad_op):
...
+@with_seed()
+def test_dense_backward_flatten():
+    for x in RandomShapes(4,2):
+        hidden = random.randrange(1, 4)
+        net = gluon.nn.Sequential()
+        with net.name_scope():
+            net.add(gluon.nn.Dense(hidden, flatten=True))
+        net.initialize(mxnet.initializer.Constant(.5))
+        x.attach_grad()
+        with ag.record():
```
apeforest: Could you reuse the `check_second_order_unary(x, op, grad_grad_op)` method as all the other operators do in this file?

src/operator/nn/fully_connected.cc
larroy: right, my bad

tests/python/unittest/test_gluon.py
larroy: same? I'm using it in the test.

larroy: To check that ndarray are equal. Is that not correct?

tests/python/unittest/test_higher_order_grad.py
larroy: I can change before the final merge. Now it helps visually inspect the arrays when debugging.

tests/python/unittest/test_higher_order_grad.py
larroy: How? Is not unary...  could you propose an example?

larroy:
Thanks for the review. Is the approach correct in your opinion?

tests/python/unittest/test_higher_order_grad.py
larroy: its an import alias, but ok.

tests/python/unittest/test_higher_order_grad.py
larroy: Is it such a big problem in a unit test? I find it useful to generate inputs.

src/operator/nn/fully_connected.cc
larroy: I have explained this before.

src/operator/nn/fully_connected-inl.h
larroy: It's quite unreadable for this purpose, but ok.

src/operator/nn/fully_connected-inl.h
larroy: I will separate it.

src/operator/nn/fully_connected-inl.h
```diff
@@ -47,7 +48,24 @@ namespace fullc {
...
 enum FullyConnectedOpResource {kTempSpace};
 enum FullyConnectedOpOutputs {kOut};
-}  // fullc
+enum FullyConnectedGradGradOutputs {
+  kOyGrad,
+  kXGradGrad,
+  kWGradGrad,
+  kBGradGrad
+};
+enum Inputs {
```
apeforest: A more specific name?

src/operator/nn/fully_connected-inl.h
```diff
@@ -47,7 +48,24 @@ namespace fullc {
 enum FullyConnectedOpInputs {kData, kWeight, kBias};
 enum FullyConnectedOpResource {kTempSpace};
 enum FullyConnectedOpOutputs {kOut};
-}  // fullc
+enum FullyConnectedGradGradOutputs {
```
apeforest: nit: Can we keep the same code style as previous enums

src/operator/nn/fully_connected-inl.h
```diff
@@ -249,6 +285,114 @@ void FullyConnectedGradCompute(const nnvm::NodeAttrs& attrs,
...
+    o_w_grad = FlattenAs2DTail<xpu, DType>(inputs[kOwGrad], ctx);
+    o_y = inputs[o_y_idx].get<xpu, 2, DType>(stream);
+    x_grad_grad = FlattenAs2DTail<xpu, DType>(outputs[kXGradGrad], ctx);
+    w_grad_grad = FlattenAs2DTail<xpu, DType>(outputs[kWGradGrad], ctx);
+  }
+  linalg_gemm(o_y, o_w_grad, x_grad_grad, false, false, stream);
+  linalg_gemm(o_y, o_x_grad, w_grad_grad, true, false, stream);
+  // 3rd order not supported
+  Fill(stream, o_y_grad_blob, kWriteTo, static_cast<DType>(0));
+  /* TODO(larroy) bias is not supported yet as there's no bias input to backward. Bias grad grad is
```
apeforest: @sxjscience Could you please review if this is correct?

tests/python/unittest/test_gluon.py
apeforest: But there is no change in the test in this file.

src/operator/nn/fully_connected-inl.h
larroy: Please suggest better ones in these circumstances. It would help, this is the best I could came up with and I'm not a native speaker.

src/operator/nn/fully_connected-inl.h
larroy: Meaning single line?  That's another $50 of cost on CI run for a whitespace change...

src/operator/nn/fully_connected-inl.h
```diff
@@ -249,6 +285,114 @@ void FullyConnectedGradCompute(const nnvm::NodeAttrs& attrs,
...
+    w_grad_grad = FlattenAs2DHead<xpu, DType>(outputs[kWGradGrad], ctx);
+  } else {
+    o_x_grad = FlattenAs2DTail<xpu, DType>(inputs[kOxGrad], ctx);
+    o_w_grad = FlattenAs2DTail<xpu, DType>(inputs[kOwGrad], ctx);
+    o_y = inputs[o_y_idx].get<xpu, 2, DType>(stream);
+    x_grad_grad = FlattenAs2DTail<xpu, DType>(outputs[kXGradGrad], ctx);
+    w_grad_grad = FlattenAs2DTail<xpu, DType>(outputs[kWGradGrad], ctx);
+  }
+  linalg_gemm(o_y, o_w_grad, x_grad_grad, false, false, stream);
+  linalg_gemm(o_y, o_x_grad, w_grad_grad, true, false, stream);
```
sxjscience: It seems that we need to add the `req` flag in linalg_gemm. Because the computing the gradient may trigger the `addTo` optimization.

src/operator/nn/fully_connected.cc
apeforest: Can you provide reference and document it somewhere. It's not clear why we need to remove inline here especially it's not related to this feature at all.

src/operator/nn/fully_connected.cc
apeforest: Please add reference. According to this article, the compiler can automatically detect whether inline is needed. https://www.geeksforgeeks.org/inline-functions-cpp/

Moreover, if this does not affect the feature you are developing in this PR, please not change.

tests/python/unittest/test_higher_order_grad.py
apeforest: Please change to random if you finished debugging this.

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -210,6 +217,168 @@ def check_second_order_unary(x, op, grad_grad_op, rtol=None, atol=None):
...
+                                   create_graph=True, retain_graph=True)[0]
+            o_x_grad = arange_shape_like(x_grad)
+            w_grad_grad = autograd.grad(heads=x_grad, variables=w,
+                                        head_grads=o_x_grad, create_graph=False)[0]
+            w_grad = autograd.grad(heads=y, variables=w, head_grads=o_y,
+                                   create_graph=True, retain_graph=True)[0]
+            o_w_grad = arange_shape_like(w_grad)
+            x_grad_grad = autograd.grad(heads=w_grad, variables=x,
+                                        head_grads=o_w_grad, create_graph=False)[0]
+        #print("x_grad_shape:\n{}".format(x_grad.shape));
```
apeforest: remove these comments please.

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -210,6 +217,168 @@ def check_second_order_unary(x, op, grad_grad_op, rtol=None, atol=None):
...
+            net.add(gluon.nn.Dense(hidden, flatten=True))
+        net.initialize(mxnet.initializer.Constant(.5))
+        x.attach_grad()
+        with autograd.record():
+            y = net.forward(x)
+            o_y = arange_shape_like(y)  # head gradient of y
+            params = [p.data() for p in net.collect_params().values()]
+            w = params[0]
+            b = params[1]
+            print("Checking y ({}) = x({}) * w^T({}) + b({})".format(y.shape, x.shape, w.shape, b.shape))
```
apeforest: remove print

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -210,6 +217,168 @@ def check_second_order_unary(x, op, grad_grad_op, rtol=None, atol=None):
...
+
+
+def flatten2d_left(x):
+    s_0 = reduce(mul, x.shape[:-1])
+    s_1 = x.shape[-1]
+    return x.reshape((s_0, s_1))
+
+
+@with_seed()
+def test_dense_backward_flatten():
```
apeforest: Can we create a generic method to test gluon blocks such as `test_gluon_block(data, weight, block, expected)` just like we are doing with check_second_order_unary? It will be easier to verify if the test is correct.

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -210,6 +217,168 @@ def check_second_order_unary(x, op, grad_grad_op, rtol=None, atol=None):
...
 
 
+def arange_shape_like(y):
+    shape = y.shape
+    nelems = reduce(mul, shape)
+    x = nd.arange(nelems).reshape(shape)
+    return x
+
+
+class NDArrayGenerator(object):
```
apeforest: Can we use existing test util instead of creating new ones?

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -210,6 +217,168 @@ def check_second_order_unary(x, op, grad_grad_op, rtol=None, atol=None):
...
+
+    def __next__(self):
+        if self.curdim > self.dim:
+            raise StopIteration
+        x = NDArrayGenerator.gen(self.curdim)
+        self.curdim += 1
+        return x
+
+
+def flatten2d_right(x):
```
apeforest: Can we use existing test util instead of creating new ones?

tests/python/unittest/test_higher_order_grad.py
larroy: Output is ignored by nosetests unless specifically enabled. It's ok to leave print in unit tests. Check nosetest for more detail.

tests/python/unittest/test_higher_order_grad.py
larroy: ok

tests/python/unittest/test_higher_order_grad.py
larroy: Why do we need this generic method? Seems to go against YAGNI. Where is the return on investment here?

tests/python/unittest/test_higher_order_grad.py
larroy: I don't see the issue.

tests/python/unittest/test_higher_order_grad.py
apeforest: > Output is ignored by nosetests unless specifically enabled. It's ok to leave print in unit tests. Check nosetest for more detail.

Reference? For most mxnet python test, not adding print statement has been a general practice.

tests/python/unittest/test_higher_order_grad.py
apeforest: It does not follow good coding practice and bloat code size.

larroy: Can we agree to remove the commented out code and then merge? or there's something else that you would want it changed? I have limited bandwidth to keep iterating in this PR.

PR #14767, opened 2019-04-22
Use DEFAULT macro in C APIs

yuxihu:
This PR changes the way of specifying the default values in C APIs by using the DEFAULT macro. Otherwise, it will fail compilation with C code. We also remove the duplicate declaration of MXInvokeCachedOp API.

PR #14789, opened 2019-04-24
[v1.4.x] Use DEFAULT macro in C APIs (#14767)

yuxihu:
Cherry-pick the change that fixes the way of specifying default values in C APIs. Otherwise, it will fail compilation with C code.


yuxihu:
@junrushao1994 @szha Please help review and merge. Thanks!

PR #14805, opened 2019-04-25
Fix memory leak when executing monitor callback in GraphExecutor

### LINKED ISSUE #10692: Executor's set_monitor_callback leaks memory in high-level languages.
## Description
Executor's set_monitor_callback leaks memory in high-level languages.

OSX, pip installed MXNet.

```python
import mxnet as mx
count = 0

def mon_callback(*args, **kwargs):
    global count
    count = count + 1

def main():
    while True:
        process_request()[0].wait_to_read()

def process_request():
    x = mx.sym.Variable('x')
    y = mx.sym.FullyConnected(x, num_hidden=1024)
    for i in range(0, 4):
        y = x + 1
    executor = y.bind(mx.cpu(), {'x': mx.nd.zeros((1024, ))})
    executor.set_monitor_callback(mon_callback)
    return executor.forward(is_train=False)

if __name__ == "__main__":
    main()
```

## Comments
*  Looks like the leak is due to this implementation: https://github.com/apache/incubator-mxnet/blob/master/src/executor/graph_executor.cc#L1461 which copies output ndarrays and then passes back raw pointers to them.  This might make sense for native languages, but if the language accessing the callback is python for example it becomes difficult to properly delete this memory.

yuxihu:
Fixes #10692

We create a copy of input/output NDArray to pass into the monitor callback function in GraphExecutor. However, we did not free the memory for the NDArray copy which caused memory leak. This PR fixes this issue.

### DIFF
--- a/src/executor/graph_executor.cc
+++ b/src/executor/graph_executor.cc
@@ -1255,14 +1255,16 @@ void GraphExecutor::ExecuteMonInputCallback(size_t nid) {
   CHECK_EQ(opnode.exec->in_array.size(), input_names.size());
   for (size_t i = 0; i < opnode.exec->in_array.size(); ++i) {
     if (node->inputs[i].node->is_variable()) {
-    // Monitor variable
-    NDArray *cpy = new NDArray(opnode.exec->in_array[i]);
-    std::string name = node->inputs[i].node->attrs.name;
-    this->monitor_callback_(name.c_str(), reinterpret_cast<void*>(cpy));
+      // Monitor variable
+      NDArray *cpy = new NDArray(opnode.exec->in_array[i]);
+      std::string name = node->inputs[i].node->attrs.name;
+      this->monitor_callback_(name.c_str(), reinterpret_cast<void*>(cpy));
+      delete cpy;
     }
     NDArray *cpy = new NDArray(opnode.exec->in_array[i]);
     std::string name = inode.source->attrs.name + "_" + input_names[i];
     this->monitor_callback_(name.c_str(), reinterpret_cast<void*>(cpy));
+    delete cpy;
   }
 }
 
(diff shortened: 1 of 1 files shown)


yuxihu:

yuxihu:
Fix no longer needed. The memory is freed at higher level language after the NDArray is used.

wkcn:
Hi, @yuxihu.

Could you please tell me where the memory of NDArray is freed?

I met a similar problem. I found the function [`AllocateNDArrayCopy`](https://github.com/apache/incubator-mxnet/blob/master/src/operator/custom/custom.cc#L46) in Custom Operator allocates new NDArray, but I do not know where the NDArray is released.

Thank you!

yuxihu:
> Hi, @yuxihu.
> 
> Could you please tell me where the memory of NDArray is freed?
> 
> I met a similar problem. I found the function [`AllocateNDArrayCopy`](https://github.com/apache/incubator-mxnet/blob/master/src/operator/custom/custom.cc#L46) in Custom Operator allocates new NDArray, but I do not know where the NDArray is released.
> 
> Thank you!

@wkcn It turned out that the higher level language (e.g. Python) needs to free the memory for the issue I was trying to fix with this PR. More information can be found [here](https://github.com/apache/incubator-mxnet/issues/10692#issuecomment-487698584). We create a (Python) NDArray with the raw pointer created before calling the callback. The Python NDArray will delete the pointer once it is out of scope. This is how we do it in [Monitor](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/monitor.py#L70) class. The memory is freed from [here](https://github.com/apache/incubator-mxnet/blob/225f71f744ac5e7bd29868b6d3ba0e4fe2527c43/python/mxnet/_ctypes/ndarray.py#L51).

wkcn:
Thank you!

PR #14588, opened 2019-04-01
[Dependency Update]Update the numpy version

stu1130:
## Description ##
* Numpy version less and equal to 1.16.0 has been found with security vulnerability. Detailed can be find in this link: https://nvd.nist.gov/vuln/detail/CVE-2019-6446#vulnCurrentDescriptionTitle.
Upgrade the numpy version both in setup.py and ci part to mitigate security vulnerability
* Update the dmlc-core commit to use python3 linter

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

## Comments ##
disable several pylint comparison-with-callable on event_handler @roywei please check it

PR #14807, opened 2019-04-26
Upgrade Pylint version to 2.3.1

stu1130:
## Description ##
* This PR is required to upgrade the numpy to 1.16.0(#14588) since Pylint have false alarms on some of Numpy functions but the patch is only available with pylint >= 2.0
1. upgrade the default `make pylint`  to use python3 pylint instead of python2 since pylint 2.0 only support python2 [link](https://github.com/PyCQA/pylint/issues/1763)
2. suppressed most of the pylint error for now and raised a issue to track #14851
3. fixes consider-using-in & useless-return

### Essentials ###

## Comments ##
@vandanavk

### DIFF
--- a/ci/docker/install/ubuntu_publish.sh
+++ b/ci/docker/install/ubuntu_publish.sh
@@ -66,5 +66,5 @@ python2 get-pip.py
 
 apt-get remove -y python3-urllib3
 
-pip2 install nose cpplint==1.3.0 pylint==1.9.4 'numpy<=1.15.2,>=1.8.2' nose-timer 'requests<2.19.0,>=2.18.4' h5py==2.8.0rc1 scipy==1.0.1 boto3
+pip2 install nose cpplint==1.3.0 'numpy<=1.15.2,>=1.8.2' nose-timer 'requests<2.19.0,>=2.18.4' h5py==2.8.0rc1 scipy==1.0.1 boto3
 pip3 install nose cpplint==1.3.0 pylint==2.3.1 'numpy<=1.15.2,>=1.8.2' nose-timer 'requests<2.19.0,>=2.18.4' h5py==2.8.0rc1 scipy==1.0.1 boto3
--- a/ci/docker/install/ubuntu_publish.sh
+++ b/ci/docker/install/ubuntu_publish.sh
@@ -66,5 +66,5 @@ python2 get-pip.py
 
 apt-get remove -y python3-urllib3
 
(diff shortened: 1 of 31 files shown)


ci/docker/install/ubuntu_publish.sh
```diff
@@ -66,5 +66,5 @@ python2 get-pip.py
 
 apt-get remove -y python3-urllib3
 
-pip2 install nose cpplint==1.3.0 pylint==1.9.3 'numpy<=1.15.2,>=1.8.2' nose-timer 'requests<2.19.0,>=2.18.4' h5py==2.8.0rc1 scipy==1.0.1 boto3
-pip3 install nose cpplint==1.3.0 pylint==2.1.1 'numpy<=1.15.2,>=1.8.2' nose-timer 'requests<2.19.0,>=2.18.4' h5py==2.8.0rc1 scipy==1.0.1 boto3
+pip2 install nose cpplint==1.3.0 pylint==1.9.4 'numpy<=1.15.2,>=1.8.2' nose-timer 'requests<2.19.0,>=2.18.4' h5py==2.8.0rc1 scipy==1.0.1 boto3
```
marcoabreu: We can get rid of pylint in py2 then, right?

stu1130: Good catch!

roywei:
@mxnet-label-bot add [CI, Python]

sandeep-krishnamurthy:
Waiting for CI.

python/mxnet/gluon/trainer.py
```diff
@@ -16,7 +16,7 @@
 # under the License.
 
 # coding: utf-8
-# pylint: disable=line-too-long
+# pylint: disable=line-too-long, no-else-raise
```
eric-haibin-lin: Is there any plan to fix the newly added pylint suppression? Can we avoid adding disables at global scope (e.g. move the disable to the corresponding python block?).

stu1130: @eric-haibin-lin thanks. I will raise a issue to capture what we need to fix & move the disable to the block

marcoabreu: +1

Instead of adding the surpressions, I'd prefer if the points are actually being addressed - or a specific disable if there's really no sense in implementing the suggestion.

stu1130: Since it's related to Numpy version update due to [security vulnerability issue](https://nvd.nist.gov/vuln/detail/CVE-2019-6446#vulnCurrentDescriptionTitle), I would prefer to use the issue #14851 to track and fix it later.
Personally, I would prefer to suppress unnecessary-pass & no-else-raise for better readability.

eric-haibin-lin:
Thanks. Comments addressed

PR #14816, opened 2019-04-26
support auto mode for BatchNorm use_global_stats 

YutingZhang:
This PR is to make BatchNorm support "auto" mode for `use_global_stats`. When `autograd.is_training()==True`, it `use_global_stats=False`; otherwise, `use_global_stats=True`.

### DIFF
--- a/python/mxnet/gluon/nn/basic_layers.py
+++ b/python/mxnet/gluon/nn/basic_layers.py
@@ -353,8 +353,21 @@ class BatchNorm(HybridBlock):
         super(BatchNorm, self).cast(dtype)
 
     def hybrid_forward(self, F, x, gamma, beta, running_mean, running_var):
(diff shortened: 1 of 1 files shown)


wkcn:
Could you please add the related document?

YutingZhang:
> Could you please add the related document?

Sure. Done. Thanks for the review!

YutingZhang:
> This would not take effect when hybridized.

Why?
When `hybrid_forward` is called with the context of `autograd.is_training()`, this will work. And when `hybridize()` is not called again, the block's behavior is not expected to change. Is it not it the expected behavior?

YutingZhang:
Based on the offline discussion with @szha, an implementation at the operator-level will be preferable. I will use this pull request as a temporary hack for my project and send a feature request for an operator-level implementation.

PR #14855, opened 2019-05-01
[WIP] Jetson nano & TX installation updates

aaronmarkham:
## Description ##
The Jetson instructions didn't work for me on the Nano, so I worked through a bunch of the errors and came up with these fixes.

The goal was to get around the really long compile time locally on the device and to use cross-compilation. 
I also thought it would be nice to use the Java API for inference-only applications.

So this PR gets you MXNet v1.4.1:
* [Pre-compiled binary for Jetson arch](https://s3.us-east-2.amazonaws.com/mxnet-public/install/jetson/1.4.1/libmxnet.so)
* [A Python wheel for Jetson arch](https://s3.us-east-2.amazonaws.com/mxnet-public/install/jetson/1.4.1/mxnet-1.4.1-py3-none-any.whl)
* Instructions on how to roll your own via Docker or manually

## Gotchas 
* There's no mention of the `MSHADOW_USE_PASCAL` setting in the CI cross-compile route. This probably needs to be addressed, but I'm not sure of the impact. This PR ignores it just as the current setup that uses Docker/CI does.

## Java Support Excluded (updated)
Thanks @lanking520 and @zachgk for helping on the Java stuff. And @larroy for general Jetson help.

I used a workaround to get v1.4.1 for Java, but it doesn't work on the Nano due to some hard-coded build flags that are for Intel chipsets. Removing the flag manually resulted in other errors asking for configuration. I don't know enough about cross-compiling to sort out these issues. Combine that with it being really hard to reproduce for v1.4.1, I'm letting this go for this PR. Maybe someone else can look at getting the Java API to work with 1.5.0 on the Nano.


aaronmarkham:
arm8 fails because it is looking for OpenCV. I turned that on by default because it is a requirement for the Java API to build. I think I should check arm8's install steps and add opencv to it.

ci/docker/runtime_functions.sh
```diff
@@ -24,8 +24,8 @@ set -ex
 
 NOSE_COVERAGE_ARGUMENTS="--with-coverage --cover-inclusive --cover-xml --cover-branches --cover-package=mxnet"
 NOSE_TIMER_ARGUMENTS="--with-timer --timer-ok 1 --timer-warning 15 --timer-filter warning,error"
-CI_CUDA_COMPUTE_CAPABILITIES="-gencode=arch=compute_52,code=sm_52 -gencode=arch=compute_70,code=sm_70"
-CI_CMAKE_CUDA_ARCH_BIN="52,70"
+CI_CUDA_COMPUTE_CAPABILITIES="-gencode=arch=compute_52,code=sm_52 -gencode arch=compute_53,code=sm_53 -gencode=arch=compute_70,code=sm_70"
+CI_CMAKE_CUDA_ARCH_BIN="52,53,70"
```
KellenSunderland: I don't think we need to change this line (unless we're running Jetson hardware in CI).

make/crosscompile.jetson.mk
```diff
@@ -71,15 +71,21 @@ USE_CUDA = 1
 
 # add the path to CUDA library to link and compile flag
 # if you have already add them to environment variable, leave it as NONE
-# USE_CUDA_PATH = /usr/local/cuda
-USE_CUDA_PATH = /usr/local/cuda-9.0/targets/aarch64-linux
+USE_CUDA_PATH = /usr/local/cuda
+# USE_CUDA_PATH = /usr/local/cuda-9.0/targets/aarch64-linux
```
KellenSunderland: let's just delete this line if it's not needed.

KellenSunderland:
Couple nits but looks good.  Thanks for updating this to support nano's @aaronmarkham

anirudhacharya:

docs/install/install-jetson.md
```diff
@@ -0,0 +1,272 @@
...
+This creates the required `.jar` file to use in your Java or Scala projects.
+
+## Conclusion and Next Steps
+
+You are now ready to run MXNet on your NVIDIA Jetson TX2 or Nano device.
+You can verify your MXNet Python installation with the following:
+
+```python
+import mxnet
+mxnet.__version__
```
lebeg: usually it's good to make 2 tensors on the gpu and add them (to make sure CUDA related stuff was built correctly)

make/crosscompile.jetson.mk
```diff
@@ -71,15 +71,20 @@ USE_CUDA = 1
...
 
 # whether to enable CUDA runtime compilation
 ENABLE_CUDA_RTC = 0
 
 # whether use CuDNN R3 library
 USE_CUDNN = 1
 
+# set flags for jetson boards
+CUDA_ARCH := -gencode arch=compute_53,code=sm_53 \
+             -gencode arch=compute_62,code=sm_62
```
lebeg: do we have 6.2 on Jetsons?

docs/install/install-jetson.md
```diff
@@ -0,0 +1,272 @@
...
+                        libatlas-base-dev \
+                        libopencv-dev \
+                        python-pip
+
+sudo pip install --upgrade \
+                        pip \
+                        setuptools
+
+sudo pip install \
+                        graphviz==0.8.4 \
```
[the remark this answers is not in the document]
aaronmarkham: This is for your local install vs doing it with docker.

docs/install/install-jetson.md
aaronmarkham: Good idea! I'll add that in. I was able to verify this myself.

make/crosscompile.jetson.mk
aaronmarkham: Yeah, this is what the Jetson Nano docs specify.

ci/docker/install/arm8_core.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
larroy: I'm confused, why do we need opencv on the host. We are cross compiling. don't you want opencv on the target instead?

https://airs.com/ian/configure/configure_5.html
https://landley.net/writing/docs/cross-compiling.html

ci/docker/install/arm8_python.sh
```diff
@@ -0,0 +1,36 @@
...
+export DEBIAN_FRONTEND=noninteractive
+
+apt-get -y install python-pip
+
+pip install --upgrade \
+                        pip \
+                        setuptools
+
+pip install \
+                        graphviz \
```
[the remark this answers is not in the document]
aaronmarkham: @lebeg Looking at this again... I added an [issue about this](https://github.com/apache/incubator-mxnet/issues/14918) last week. I ask why we need both the system version of graphviz and the python version. It appears that they offer different features and ways to invoke, but it isn't clear to me what we need. I'm going with the assumption that `mxnet.viz` need one or both of these.
You can't pin a system version - you can do some scripting checks to look for packages and try to get the right one but there's no >= feature for apt from what I can tell. So that one is unpinned, but the pip one I can pin.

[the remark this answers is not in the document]
lebeg: Strange, but probably ok.

ci/docker/install/arm8_core.sh
aaronmarkham: When I tried to get the Java API compiled on the Nano, it complained that I hadn't compiled MXNet with OpenCV. 
This error went away after I added OpenCV, but I have to say it added other issues to CI's deps.

aaronmarkham: Considering that I needed to pull the Java API support because of cross-compile issues, I could unwind this OpenCV dep. However, one of my main use cases for the Nano is to plug in a camera, so I'm pretty sure an MXNet build with OpenCV is a good default for Nano devices.

aaronmarkham:
I removed the scala setup step since this PR isn't going to get the Scala/Java stuff working yet. It'll focus on Python only.

ci/docker/install/arm8_core.sh
larroy: Is not a problem, I'm not saying you should unwind it. My question is that since it's installing in the host, I don't understand how it affects the target. Just a question not blocking the PR at all.

aaronmarkham:
This PR is hung up on a CI issue. It's arm pipelines are failing with not being able to `apt-get update`. Does anyone know why this would be the case?

lebeg:
Sometimes the docker cache is outdated and contains not anymore valid ppa links. Actually the docker cache should be rebuilt by [this](http://jenkins.mxnet-ci.amazon-ml.com/job/restricted-docker-cache-refresh/job/master/) job, so it's worth just a try on retriggering the PR verification.

larroy:
I sometimes have docker issues as well, I think the way we use it triggers some bugs in caching. I suggest to do the following which has worked for me when I had problems:

* docker pull the base images manually, ex. ubuntu
* Fully erase the docker cache.
* Don't use cache-from in your build.py script and build the container locally.

```
docker pull ubuntu:16.04
docker rm $(docker ps -a -q)
docker rmi $(docker images -a -q)
```

This is of course if you think your problems are related to docker caching.

aaronmarkham:
Rebasing to see if maybe the cache is fixed now...

aaronmarkham:
Still stuck. I'm not sure what to do now.

dwSun:
Thanks for your work.

But I have problems with your pre-built wheel.:
after installed your wheel with pip, I got this error:
```sh
david@TinyNuke:~$ python3 mnist_cnn.py 
  File "mnist_cnn.py", line 2, in <module>
    import mxnet as mx
  File "/home/david/.local/lib/python3.6/site-packages/mxnet/__init__.py", line 24, in <module>
    from .context import Context, current_context, cpu, gpu, cpu_pinned
  File "/home/david/.local/lib/python3.6/site-packages/mxnet/context.py", line 24, in <module>
    from .base import classproperty, with_metaclass, _MXClassPropertyMetaClass
  File "/home/david/.local/lib/python3.6/site-packages/mxnet/base.py", line 213, in <module>
    _LIB = _load_lib()
  File "/home/david/.local/lib/python3.6/site-packages/mxnet/base.py", line 204, in _load_lib
    lib = ctypes.CDLL(lib_path[0], ctypes.RTLD_LOCAL)
  File "/usr/lib/python3.6/ctypes/__init__.py", line 348, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: /home/david/.local/lib/python3.6/site-packages/mxnet/libmxnet.so: cannot open shared object file: No such file or directory

```

Then I figured out your wheel is packed with a x86-64 arch libmxnet.so:
```sh
david@TinyNuke:~$ ls -lah /home/david/.local/lib/python3.6/site-packages/mxnet/libmxnet.so
-rwxrwxr-x 1 david david 364M 6月   7 22:22 /home/david/.local/lib/python3.6/site-packages/mxnet/libmxnet.so
david@TinyNuke:~$ file /home/david/.local/lib/python3.6/site-packages/mxnet/libmxnet.so
/home/david/.local/lib/python3.6/site-packages/mxnet/libmxnet.so: ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, BuildID[sha1]=5b07574bc684f6acdef5ba509889ad9c8298d157, with debug_info, not stripped
d
```
Local dynamic library on jetson nano looked like this:
```sh
file /usr/local/cuda/lib64/libcudart.so.10.0.166
/usr/local/cuda/lib64/libcudart.so.10.0.166: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, stripped
```

larroy:
If you want we can sit together for a couple of hours and figure this out.

aaronmarkham:
@dwSun you're right, I need to swap out the .so file in that wheel. I have one that works in the main comment of this PR or just click here. https://s3.us-east-2.amazonaws.com/mxnet-public/install/jetson/1.4.1/libmxnet.so

dwSun:
@aaronmarkham Thanks, your .so file works perfectly on nano device.

aaronmarkham:
Closing in favor of #15117  - although at some point nano should be added to CI.

PR #14874, opened 2019-05-03
[MXNET-1399] multiclass-mcc metric enhancements

tlby:
## Description ##
multiclass-mcc metric enhancements

### Essentials ###

### Changes ###
- [x] Rename metric from "PCC" to "MMCC" because though the math is derived
   from Pearson CC, it's utility is as a multiclass extension of Mathews CC.
- [x] Harden mx.metric.MMCC.update to more variations of input format, similar to
   mx.metric.Accuracy.update.

## Comments ##
The name change from `mx.metric.PCC()` to `mx.metric.MMCC()` for this metric has not yet been included in any releases or tagged branches, and for anyone following master the code update would be trivial.

### DIFF
--- a/python/mxnet/metric.py
+++ b/python/mxnet/metric.py
@@ -860,7 +860,7 @@ class MCC(EvalMetric):
 
     .. note::
 
-        This version of MCC only supports binary classification.  See MMCC.
+        This version of MCC only supports binary classification.  See PCC.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


tlby:
verified that the additional checks in mx.metric.MMCC.update() are not a bottleneck.
[bench.py](https://github.com/apache/incubator-mxnet/files/3143902/bench.py.txt)

tlby:
Oh, I should clarify this here; I discovered that use cases like the [fastText example from gluon-nlp](https://github.com/dmlc/gluon-nlp/blob/b079babf096e2a82bb901c7e16efe14183f7709f/scripts/text_classification/fasttext_word_ngram.py#L154) expect metrics to be comfortable taking predictions already `.argmax()` resolved.  `mx.metric.Accuracy` has a special case for that and this patch makes mx.metric.MMCC flexible in the same way.

szha:
@tlby thanks for the patch! Since PCC has already been merged for almost a month, there might be people who already depend on PCC. If there is, then renaming PCC to MMCC will likely break them. Since PCC hasn't made into a release yet, it's technically not breaking a released API, but with that in mind, do you feel necessary to rename the class?

tlby:
I don't think it's absolutely necessary, but I definitely regret the name choice because while the implementation came from a discrete derivation of the Pearson correlation coefficient, all a user cares about is that it's a multiclass extension of Matthews correlation coefficient.  When it comes to using these metrics Pearson is for continuous and Matthews is for discrete predictions.  This metric is discrete so I feel it was misnamed.  I later discovered that the formula I used is equivalent to one [mentioned in Wikipedia](https://en.wikipedia.org/wiki/Matthews_correlation_coefficient#Multiclass_case) as an extension of Matthews.

Perhaps we should leave "pcc" as an alias for the sake of reverse compatibility?

anirudhacharya:

karan6181:
@szha Could you please review it again? Thanks!

piyushghai:
@szha Gentle ping...

vandanavk:
Is this PR good to go?

szha:
@karan6181 @piyushghai @vandanavk the class rename needs to be reverted

karan6181:
@szha I think @tlby addressed your review comments by renaming the class. Could you please review it again? Thanks!

szha:
Thanks @tlby! The only thing left is that we need test cases to cover both code paths.

piyushghai:
@szha Is this PR good to merge now ?

PR #13668, opened 2018-12-18
Update MKL-DNN to v0.18 release (was: fix the Dense layer issue)

TaoLv:
## Description ##
Issue: #13451

Since the case provided in the issue needs huge memory, I'm not sure if it's appropriate to put it into unit test.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #14899, opened 2019-05-06
Add mkldnn_version.h to pip package

yuxihu:
Per changes in #13668, we generate a new header file `mkldnn_version.h` during MXNet compilation based on the template file [mkldnn_version.h.in](https://github.com/intel/mkl-dnn/blob/7de7e5d02bf687f971e7668963649728356e0c20/include/mkldnn_version.h.in). This dynamically generated header file needs to be added to the MXNet pip package to prevent compilation failure for building Horovod with MKLDNN enabled MXNet pip package.

### DIFF
--- a/tools/pip/setup.py
+++ b/tools/pip/setup.py
@@ -152,6 +152,8 @@ if variant.endswith('MKL'):
         package_data['mxnet'].append('mxnet/libmkldnn.so.0')
     shutil.copytree(os.path.join(CURRENT_DIR, 'mxnet-build/3rdparty/mkldnn/include'),
                     os.path.join(CURRENT_DIR, 'mxnet/include/mkldnn'))
+    shutil.copy(os.path.join(CURRENT_DIR, 'mxnet-build/3rdparty/mkldnn/build/include/mkldnn_version.h'),
+                os.path.join(CURRENT_DIR, 'mxnet/include/mkldnn'))
 if platform.system() == 'Linux':
     shutil.copy(os.path.join(os.path.dirname(LIB_PATH[0]), 'libgfortran.so.3'), os.path.join(CURRENT_DIR, 'mxnet'))
     package_data['mxnet'].append('mxnet/libgfortran.so.3')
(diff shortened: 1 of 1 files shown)


yuxihu:
@szha @pengzhao-intel @TaoLv Please help review.

tools/pip/setup.py
```diff
@@ -152,6 +152,8 @@ def has_ext_modules(self):
         package_data['mxnet'].append('mxnet/libmkldnn.so.0')
     shutil.copytree(os.path.join(CURRENT_DIR, 'mxnet-build/3rdparty/mkldnn/include'),
```
szha: I thought the whole mkldnn include folder should already have been included here. Should we just change this to copy from the build folder instead?

yuxihu: `mkldnn_version.h` is generated and placed under `build/include` folder. Other header files are still in this folder. So we have to copy from two different places.

szha: I just verified that the rest of the headers are copied there too. Let's just use the folder.

yuxihu: I did not see the other headers are copied in my local environment. That's why I added separately.

```
root@8c384f679438:/build/mxnet-build/3rdparty/mkldnn/build/include# ls
mkldnn_version.h
```

szha:
```
% ls 3rdparty/mkldnn/build/install/include
i_malloc.h                mkl_blas.h                mkl_direct_blas.h         mkl_direct_call.h         mkl_direct_types.h        mkl_dnn_types.h           mkl_lapacke.h             mkl_trans.h               mkl_version.h             mkl_vml_defines.h         mkl_vml_types.h           mkl_vsl_defines.h         mkl_vsl_types.h
mkl.h                     mkl_cblas.h               mkl_direct_blas_kernels.h mkl_direct_lapack.h       mkl_dnn.h                 mkl_lapack.h              mkl_service.h             mkl_types.h               mkl_vml.h                 mkl_vml_functions.h       mkl_vsl.h                 mkl_vsl_functions.h
```
The one in the install folder has both

yuxihu: ok, changed to copy from build/install/include folder.

TaoLv:
Thank you for the fix @yuxihu. Merging now.

Issue #14125, opened 2019-02-12
FP16 support for topK 


eric-haibin-lin:
TopK doesn't support fp16 inputs.

stephenrawls:
+1

This is annoying for us because we need to use topk in our CRF Viterbi decoding during inference (for k-best Viterbi).

Because topk doesn't support float16 this means we have to convert from float16 to float32 before doing k-best Viterbi decoding. Since this is the most time consuming part of inference, we would prefer to be able to benefit from float16 operations here.

eric-haibin-lin:
related issues: https://github.com/apache/incubator-mxnet/issues/12705 https://github.com/apache/incubator-mxnet/issues/11156 https://github.com/apache/incubator-mxnet/issues/11061

Issue #12705, opened 2018-09-29
Fp16 support for top_k operator 


eric-haibin-lin:
```
mxnet.base.MXNetError: [05:47:56] src/operator/tensor/./ordering_op-inl.h:535: This operation does not support float16

Stack trace returned 10 entries:
[bt] (0) /home/ubuntu/batchdot/python/mxnet/../../lib/libmxnet.so(dmlc::StackTrace[abi:cxx11]()+0x5b) [0x7fb560c3385b]
[bt] (1) /home/ubuntu/batchdot/python/mxnet/../../lib/libmxnet.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x28) [0x7fb560c343c8]
[bt] (2) /home/ubuntu/batchdot/python/mxnet/../../lib/libmxnet.so(void mxnet::op::TopK<mshadow::gpu>(nnvm::NodeAttrs const&, mxnet::OpContext const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&, std::vector<mxnet::OpReqType, std::allocator<mxnet::OpReqType> > const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&)+0x275) [0x7fb564a47565]
[...]

Issue #11156, opened 2018-06-05
mx.nd.topk does not work with ndarray of type float16


hetong007:
Maybe there are some operators don't work well on some data types, but I just encountered the `float16`-`topk()` pair, didn't run a full scan of all possibly broken pairs.

The following is a few lines of code to reproduce it, including the error message:

```
Python 2.7.12 (default, Dec  4 2017, 14:50:18)
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import mxnet as mx
>>> a = mx.nd.array([1,2,3])
>>> a.topk()

[2.]
<NDArray 1 @cpu(0)>
>>> a.astype('float32').topk()

[2.]
<NDArray 1 @cpu(0)>
>>> a.astype('float16').max()

[3.]
<NDArray 1 @cpu(0)>
>>> a.astype('float16').topk()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ubuntu/incubator-mxnet/python/mxnet/ndarray/ndarray.py", line 189, in __repr__
    return '\n%s\n<%s %s @%s>' % (str(self.asnumpy()),
  File "/home/ubuntu/incubator-mxnet/python/mxnet/ndarray/ndarray.py", line 1894, in asnumpy
    ctypes.c_size_t(data.size)))
  File "/home/ubuntu/incubator-mxnet/python/mxnet/base.py", line 210, in check_call
    raise MXNetError(py_str(_LIB.MXGetLastError()))
mxnet.base.MXNetError: [18:53:33] include/mxnet/././tensor_blob.h:203: Check failed: mshadow::DataType<DType>::kFlag == type_flag_ TBlob.get_with_shape: data type do not match specified type.Expected: 2 v.s. given 0

Stack trace returned 10 entries:
[bt] (0) /home/ubuntu/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(dmlc::StackTrace[abi:cxx11]()+0x5b) [0x7f96f88955fb]
[bt] (1) /home/ubuntu/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x28) [0x7f96f8896168]
[bt] (2) /home/ubuntu/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(float* mxnet::TBlob::dptr<float>() const+0xfa) [0x7f96f890d2ba]
[...]

PR #14912, opened 2019-05-07
Add fp16 and fp64 support for topk

anirudhacharya:
## Description ##
Add fp16 and fp64 support for topk operator. Required for certain machine translation tasks( and other NLP related tasks).

### Essentials ###

### Changes ###
 - add fp16 and fp64 support for topk operator
 - Fixes https://github.com/apache/incubator-mxnet/issues/14125 https://github.com/apache/incubator-mxnet/issues/11156 https://github.com/apache/incubator-mxnet/issues/12705

For review - @anirudh2290 @apeforest

### DIFF
--- a/src/operator/tensor/ordering_op-inl.h
+++ b/src/operator/tensor/ordering_op-inl.h
@@ -563,13 +563,13 @@ void TopK(const nnvm::NodeAttrs& attrs,
           const std::vector<TBlob>& outputs) {
   const TopKParam& param = nnvm::get<TopKParam>(attrs.parsed);
   if (param.ret_typ == topk_enum::kReturnIndices || param.ret_typ == topk_enum::kReturnBoth) {
-    MXNET_NO_FLOAT16_TYPE_SWITCH(inputs[0].type_flag_, DType, {
+    MSHADOW_REAL_TYPE_SWITCH(inputs[0].type_flag_, DType, {
       MSHADOW_TYPE_SWITCH(param.dtype, IDType, {
         TopKImpl<xpu, DType, IDType>(ctx.run_ctx, ctx.requested[0], req, inputs[0], outputs, param);
       })
     });
   } else {
-    MXNET_NO_FLOAT16_TYPE_SWITCH(inputs[0].type_flag_, DType, {
+    MSHADOW_REAL_TYPE_SWITCH(inputs[0].type_flag_, DType, {
       TopKImpl<xpu, DType, index_t>(ctx.run_ctx, ctx.requested[0], req, inputs[0], outputs, param);
     });
   }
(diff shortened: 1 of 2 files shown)


vandanavk:
@mxnet-label-bot add [pr-awaiting-review, Operator]

src/operator/tensor/ordering_op-inl.h
```diff
@@ -401,8 +401,6 @@ void TopKImpl(const RunContext &ctx,
     mxnet::op::SortByKeyWorkspaceSize<int, int, xpu>(src.Size()));
   temp_size = std::max(temp_size,
     mxnet::op::SortByKeyWorkspaceSize<int, DType, xpu>(src.Size()));
-  temp_size = std::max(temp_size,
-    mxnet::op::SortByKeyWorkspaceSize<DType, int, xpu>(src.Size()));
```
anirudh2290: have you tried other dtypes like int64 ?

anirudhacharya: topk supports only floating types for now

anirudhacharya: yes, ran these tests successfully on my unix machine - https://github.com/apache/incubator-mxnet/blob/master/tests/python/unittest/test_ndarray.py#L799 it runs as part of `test_ndarray.py:test_order`

anirudhacharya:
@marcoabreu @perdasilva any idea why the build fails for only windows gpu and passes for the rest?

perdasilva:
I've had a look and also reached out to Anton. I have no idea what the problem here could be. I wonder if it has something to do with the environment. On Windows we seem to be using CUDA v9.2 and driver v398.75. But I don't see how this could really be an issue...=S

anirudhacharya:
Thank you @perdasilva . @lebeg can you please help out here.

src/operator/tensor/ordering_op-inl.h
sxjscience: This line should be useful because we need to do `mxnet::op::SortByKey(dat, ind, is_ascend, &sort_work);`.

anirudhacharya: can't it can be done even now? this change passed all the unit tests, btw.

sxjscience: Removing this line may break some corner cases uncovered by the existing test. Since we need to call `SortByKey<DType, int>(dat, ind, ...)`, `SortByKey<int, DType>(batch_id, dat, ...)` and `SortByKey<int, int>(batch_id, ind, ...)`, we should make sure that the temporary storage has enough size for all cases.

anirudhacharya: > may break some corner cases uncovered by the existing test

it does not break any existing tests. I checked that locally. And the tests passed the CI too.

> we should make sure that the temporary storage has enough size for all cases.

I thought about this when making the change. Is there a use-case/unit test you could point to which would break with this change.

abhinavs95:
@anirudh2290 @sxjscience Could you see if your comments are addressed? Thanks.

piyushghai:
@anirudh2290 Bouncing for a review...

vandanavk:
Is this PR good to go?

PR #14977, opened 2019-05-17
Add an utility for operator benchmarks

sandeep-krishnamurthy:
## Description ##

Add a tool to easily run operator benchmarks. This tool will be useful for MXNet developers, power users who wants to know more about individual operator performance. This tool can be integrated with CI/CD, nightly tests to catch any operator performance regression. Proposal on cwiki - https://cwiki.apache.org/confluence/display/MXNET/MXNet+Operator+Benchmarks

1. This is the first PR to setup all basic required utilities, overall structure for this operator benchmarking utility, and overall skeleton for the tool.
2. I have added <s> 3 NDArray operator (add, sub, mul) and 1 Gluon (Conv2D) operator benchmark </s> <s> 19 NDArray operator benchmark tests </s> 80 operator benchmark tests - code and results markdown file autogenerated by the tool, to showcase how it can be used. 
3. More operators will be covered in following PRs. With this PR, we will be setting up basic structure of other contributors to pitch in.
4. Added README explaining how to use the utility.

### Essentials ###

### Changes ###
- Utility to do operator benchmarks

## Comments ##

@apeforest @nswamy @access2rohit @pengzhao-intel @zha0q1


benchmark/opperf/README.md
```diff
@@ -0,0 +1,120 @@
...
+<!--- Unless required by applicable law or agreed to in writing, -->
+<!--- software distributed under the License is distributed on an -->
+<!--- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY -->
+<!--- KIND, either express or implied.  See the License for the -->
+<!--- specific language governing permissions and limitations -->
+<!--- under the License. -->
+
+# MXNet Operator Performance Benchmarks
+
+A Python utility for benchmarking and profiling individual MXNet operator execution.
```
szha: By "profiling", do you mean the tool would provide insight into the implementation of operator? (Examples include speed of each kernel, memory allocation, memory locality, scheduling such as stream synchronization)?

benchmark/opperf/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Prerequisites
+
+This utility uses MXNet profiler under the hood to fetch compute and memory metrics. Hence, you need to build MXNet with `USE_PROFILER=1` flag.
+
+Make sure to build the flavor of MXNet, for example - with/without MKL, with CUDA 9 or 10.1 etc., on which you would like to measure operator performance.
+
+## Usecase 1 - Run benchmarks for all the operators
+
+Below command runs all the MXNet operators (NDArray and Gluon) benchmarks with default inputs and saves the final result as JSON in the provided file.
```
szha: Gluon blocks are implemented with operators but blocks themselves are not usually considered operators. Measuring the actual operators would be more helpful.

benchmark/opperf/nn_operations/__init__.py
```diff
@@ -0,0 +1,16 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
szha: This module seems less urgently needed than measurement of the actual operators, since the execution is asynchronous.

benchmark/opperf/README.md
```diff
@@ -0,0 +1,120 @@
...
+2. Computation time
+
+See design proposal document for more details - https://cwiki.apache.org/confluence/display/MXNET/MXNet+Operator+Benchmarks 
+
+# TODO
+
+All contributions are welcome. Below is the list of desired features:
+
+1. Cover all MXNet operators.
+2. Enhance MXNet profiler with additional APIs to programmatically fetch and process profiler data.
```
szha: What do you mean by "process"? To what end?

benchmark/opperf/README.md
```diff
@@ -0,0 +1,120 @@
...
+3. Integration with CI/CD system to run operator benchmarks for PR builds, nightly builds.
+4. Dashboards and other modes of presentation of results for analyzing and planning tasks such as operator performance improvements.
+5. Integration with tools such as [Hypothesis](https://hypothesis.readthedocs.io/en/latest/) for randomized input generation for profiling to identify bottlenecks in operators.
```
szha: Could you elaborate on how random data helps identify relevant bottlenecks? Usually, models come with some specific input sizes and the speed shouldn't be affected by the data, but only by the input sizes (except in the rare cases of nan where interrupts are involved).

benchmark/opperf/README.md
sandeep-krishnamurthy: Current target is to provide timing of - forward, backward kernels and memory allocated

benchmark/opperf/README.md
sandeep-krishnamurthy: I agree, for majority of operators, I am using NDArray interface. Using Gluon NN blocks for operations such as - Conv2D, Pooling, L1Loss etc. to mimic the way MXNet users use these functionalities. Do you recommend moving everything to NDArray interface?

benchmark/opperf/README.md
sandeep-krishnamurthy: I meant, with current profiler we need to dump results to console or a file to fetch results, however, for use cases like - I profile my code, get the profile result, may be save it in a markdown file, Just fetch parts of the profiler data and save to a JSOn for an upstream system like a nightly benchmark job etc.

benchmark/opperf/README.md
sandeep-krishnamurthy: Here I referred to randomness in tensor shapes. Will update the text to clarify the motivation.

sandeep-krishnamurthy:
> Thanks for initiating this. Besides the questions below, I have some high-level questions and comments:
> 
> * why is it necessary to implement all the calls by hand? this approach seems rather inefficient. is there any way to implement this more concisely?
Good point. I did think about it and was discussed by other community members on the proposal doc. Below is my thought process.

Code per operator is something like below:
```
add_res = run_performance_test(nd.add, run_backward=True, dtype=dtype, ctx=ctx,
                                   inputs=[{"lhs": (1024, 1024),
                                            "rhs": (1024, 1024)},
                                           {"lhs": (10000, 10),
                                            "rhs": (10000, 10)},
                                           {"lhs": (10000, 1),
                                            "rhs": (10000, 100)}],
                                   warmup=warmup, runs=runs)
``` 
run_performance_test function will provide all the necessary tools to do a benchmark and get results. User is expected to specify 2 things - operator name, inputs for the operator. Providing inputs based on what needs to be tested is a crucial part and is made explicit. Each operator can have different criteria that needs to be in performance tests. Ex: broadcasting shapes are necessary for say arithmetic operators, small / large tensors etc.

Hence, if we automate fully for example a solution like - Automatically fetch all operators registered, fetch inputs required, understand the input meanings (Ex: lhs shape is same or broadcastable to rhs shape). Such concise thing may help in having less code, but, may hide too many details and make it hard to generally use this tool, integrate it with systems like PR/Nightly benchmark dashboarding etc.

Having said that, there are certainly few rooms of improvements - like all binary operators like add,sub,mul have similar expectations and can be made concise. But, I thought, it may lead to over engineering a simple utility required to easily run benchmark test for an operator. 

> * what happens when people implement new operators? must they implement profiling logics here too?
Yes. Not the profiling logic, but, something like below when they provide name of the operator, and specify inputs to use.

```python
add_res = run_performance_test(nd.add, run_backward=True, dtype=dtype, ctx=ctx,
                                   inputs=[{"lhs": (1024, 1024),
                                            "rhs": (1024, 1024)}],
                                   warmup=warmup, runs=runs)
```
> * I see the PR is marked as complete despite the many TODOs in the code. I don't think the code can be checked in in this state.
Motivation of this PR as described in the description is to setup the base infrastructure for a tool to easily do operator benchmarks. Add 4 operators benchmarks in the PR setting up everything required for any community members to help it out to get more than 200 operators we have in MXNet covered. Also, reason for adding TODOs is to cover the overall skeleton and clearly put out the idea of what we want to cover next so that roadmap is clearly set.

benchmark/opperf/README.md
sandeep-krishnamurthy: Moving everything to focus on the building blocks i.e., just using NDArray for now.

benchmark/opperf/nn_operations/__init__.py
sandeep-krishnamurthy: Agreed. Making it all ndarray operators only for this initial work.

benchmark/opperf/README.md
```diff
@@ -0,0 +1,174 @@
...
+               'inputs': {'lhs': (1024, 1024), 'rhs': (1024, 1024)}},
+             ]
+}
+```
+
+## Usecase 3 - Run benchmarks for specific operator
+For example, you want to run benchmarks for `nd.add` operator in MXNet, you just run the following python script.
+
+```
+#! /usr/bin/python
```
larroy: extra space on shebang?

benchmark/opperf/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Prerequisites
+
+This utility uses MXNet profiler under the hood to fetch compute and memory metrics. Hence, you need to build MXNet with `USE_PROFILER=1` flag.
+
+Make sure to build the flavor of MXNet, for example - with/without MKL, with CUDA 9 or 10.1 etc., on which you would like to measure operator performance.
+
+## Usecase 1 - Run benchmarks for all the operators
+
+Below command runs all the MXNet operators (NDArray) benchmarks with default inputs and saves the final result as JSON in the given file.
```
larroy: why not yaml? is more legible.

benchmark/opperf/README.md
```diff
@@ -0,0 +1,174 @@
...
+2. **ctx** : `cpu` or `gpu`. By default, cpu on CPU machine, gpu(0) on GPU machine. You can override and set the global context for all operator benchmarks. Example: --ctx gpu(2).
+
+3. **dtype** : By default, `float32`. You can override and set the global dtype for all operator benchmarks. Example: --dtype float64.
+
+## Usecase 2 - Run benchmarks for all the operators in a specific category
+
+For example, you want to run benchmarks for all NDArray Arithmetic Operators, you just run the following python script.
+
+```
+#! /usr/bin/python
```
larroy: space?

benchmark/opperf/README.md
```diff
@@ -0,0 +1,174 @@
...
+          'avg_time_backward_broadcast_add': 5.3841,
+          'inputs': {'lhs': (1024, 1024), 'rhs': (1024, 1024)}}]}
+
+```
+
+## Usecase 3.1 - Run benchmarks for group of operators with same input
+For example, you want to run benchmarks for `nd.add`, `nd.sub` operator in MXNet, with the same set of inputs. You just run the following python script.
+
+```
+#! /usr/bin/python
```
larroy: space

benchmark/opperf/README.md
sandeep-krishnamurthy: Using JSON for easy parsing across other upstream systems such as nightly benchmarks, PR validation.

benchmark/opperf/README.md
sandeep-krishnamurthy: fixed

benchmark/opperf/README.md
sandeep-krishnamurthy: fixed

benchmark/opperf/README.md
sandeep-krishnamurthy: fixed

sandeep-krishnamurthy:
Latest updates:
1. Add more operator benchmarks - ~19 tensor operations are added till now.
2. Support to export result as json/md
3. Add autogenerated results markdown file
4. Move everything to only test core low level operators, i.e., NDArray operators only, no Gluon.
5. Support to send a batch of operators for testing with same inputs.
Example - Below we run all 5 arithmetic operators performance tests in one call.
```
benchmark_res = run_performance_test([nd.add, nd.subtract, nd.multiply, nd.divide, nd.modulo], run_backward=True,
                                         dtype=dtype, ctx=ctx,
                                         inputs=[{"lhs": (1024, 1024),
                                                  "rhs": (1024, 1024)},
                                                 {"lhs": (10000, 10),
                                                  "rhs": (10000, 10)},
                                                 {"lhs": (10000, 1),
                                                  "rhs": (10000, 100)}],
                                         warmup=warmup, runs=runs)
```
6. Updates and clarifications in README

ping - @szha @apeforest @nswamy @access2rohit @Zha0q1 for review.

benchmark/opperf/README.md
larroy: they are homomorphic anyway.

larroy:
@sandeep-krishnamurthy I think it would be good to provide some type hints in this code, specially in the functions. Since we seem to move MXNet to python3 anyways, this should make the code much easier to read and maintain. What do you think?

larroy:
@szha This is a valid point that we have raised internally before, the question is that this could be implemented on top of this. To me looks like a good level of abstraction and this could be a layer on top, I think that's the fundamental question that we need to ask. Operator benchmark would be very beneficial, it's also possible to get it done in stages. Rome wasn't built in a day.

benchmark/opperf/README.md
sandeep-krishnamurthy: Agree. We can add it to todo list? Currently, JSON and MD is supported, will work on Yaml next.

sandeep-krishnamurthy:
> @sandeep-krishnamurthy I think it would be good to provide some type hints in this code, specially in the functions. Since we seem to move MXNet to python3 anyways, this should make the code much easier to read and maintain. What do you think?

This is a nice addition. Will add it to the main user facing functions to start with.

szha:
I just had a discussion w/ @sandeep-krishnamurthy and we agreed that: 1. the ideal state should be the automated approach that uses the data types from op registrty to generate default values for performance testing, and also allow users to override and test for specific settings. 2. current approach won't help towards that goal since the automated approach would render the current approach in this PR mostly obsolete.

szha:
The next step I suggested is to start with the simplest case such as elementwise operators and build automated approach towards that.

sandeep-krishnamurthy:
Thanks @szha for the discussion and putting the summary here. I wanted to add few more points we discussed in the context of this PR:
1. We agreed that the current functionality in this PR is helpful for existing operators. For purpose like PR builds, nightly performance test CI, quick comparison tools for developers like working on Int64 etc. For which one of your suggestion was to maintain a separate repository / move it to mxnet jenkins or CI repository.
2. Things not favorable in this PR is that it will be an additional burden for developers adding new operators or doing breaking changes in existing operators(ex: parameters change, namespace change) to come and make modifications in this utility which we all agree is a maintenance overhead.
3. We also discussed with automated solution, we cannot handle every operator and every new use case in a fully automated way. So it would still be a combination of automated (TODO) + explicit tests (currently in this PR).

sandeep-krishnamurthy:
> I just had a discussion w/ @sandeep-krishnamurthy and we agreed that: 1. the ideal state should be the automated approach that uses the data types from op registry to generate default values for performance testing, and also allow users to override and test for specific settings. 2. current approach won't help towards that goal since the automated approach would render the current approach in this PR mostly obsolete.

Also, to clarify, the current approach does not block the evolution of this tool towards automated approach. Because, at the end of the day, we still need to have utilities to run an operator, capture results, present it in consumable format. The way I was seeing it was - when we automate few of the operator benchmarks, we just delete them from the current explicit calls till then we still have it.

sandeep-krishnamurthy:
Updates:

1. Add functionality to query MXNet registry for operator.
2. Add an example of running 20 binary_broadcast operators without explicit operator calling.
See https://github.com/apache/incubator-mxnet/pull/14977/files#diff-bbbeb0f4b497dc86cf7f3e2e9713dc64 and https://github.com/apache/incubator-mxnet/pull/14977/files#diff-a5217faf4046473dec2565d1ee07236d

Below are the observations:
1. I have added the required utilities, to query MXNet registry for all registered operators.
2. With 20 operators, we see that whatever we have in this PR as of today can evolve to suggested ideal automated way.
3. Automated way to query the operator and running benchmarks cannot work for all the operators. Operators will have different inputs and input criteria. Example: Input shapes for broadcast_add cannot work for elementwise_add and so on. 
    - This is different than just auto-generating Python API docs, because here, we need to execute so we need to know semantics of the parameters. Ex: should param1 match shape of param2.
    - At minimum, we need to set rules to generate inputs.
4. We will have a few operators like elementwise, broadcast - unary/binary operators almost covered with automated way, few complex operators like convolution, pooling, where rules for inputs are explicit.

Having, seen how this utility can be evolved. This PR having all the basic utilities and >20 operator benchmarks in automated way and few more in explicit way. @szha - Do you see any other blocking concerns?

szha:
@sandeep-krishnamurthy thanks for the update for making the efforts. I will try to review this soon

benchmark/opperf/tensor_operations/binary_broadcast_operators.py
```diff
@@ -62,6 +62,9 @@ def run_mx_binary_operators_benchmarks(ctx=mx.cpu(), dtype='float32', warmup=10,
                                           warmup=warmup, runs=runs)
         mx_binary_op_results += cur_op_res
 
-    # Prepare combined results for Arithmetic operators
+    # Prepare combined results for Binary Broadcast operators
     mx_binary_op_results = merge_map_list(mx_binary_op_results)
     return mx_binary_op_results
+
+
+print(run_mx_binary_broadcast_operators_benchmarks())
```
larroy: If this is a module and not an executable, should we use logging instead of print?

sandeep-krishnamurthy: Yes. It should all be logging. I had put it as todo. Working on it:
1. logging
2. Type hints for user facing APIs

sandeep-krishnamurthy: Using logging now. All prints are removed.

benchmark/opperf/utils/op_registry_utils.py
```diff
@@ -0,0 +1,193 @@
...
+    plist = ctypes.POINTER(ctypes.c_char_p)()
+    size = ctypes.c_uint()
+
+    check_call(_LIB.MXListAllOpNames(ctypes.byref(size),
+                                     ctypes.byref(plist)))
+    mx_operators = {}
+    operators_with_backward = []
+
+    # Prepare master list of all operators.
+    for i in range(size.value):
```
larroy: Does it make sense to expose this function in C API?

If no, could we at least separate the concerns in two functions? Get the op names and then do the filtering by names? the logic would be clearer and easier to maintain and more testable than mixing all together. Ideally the op selection is a separate pure function that operates in this list of strings (and has no Ctypes dep)

You can also have a third function that assembles the operator handles.

I would simplify this line as:

operator_names = [py_str(plist[i]) for i in range(size.value)]

benchmark/opperf/utils/op_registry_utils.py
```diff
@@ -0,0 +1,193 @@
...
+    ----------
+    op_name: str
+        Name of operator to get handle for.
+    """
+    op_handle = OpHandle()
+    check_call(_LIB.NNGetOpHandle(c_str(op_name), ctypes.byref(op_handle)))
+    return op_handle
+
+
+def _get_op_arguments(op_name, op_handle):
```
larroy: nice

benchmark/opperf/utils/op_registry_utils.py
```diff
@@ -0,0 +1,193 @@
...
+        ctypes.byref(ret_type)))
+
+    narg = int(num_args.value)
+    arg_names = [py_str(arg_names[i]) for i in range(narg)]
+    arg_types = [py_str(arg_types[i]) for i in range(narg)]
+
+    return narg, arg_names, arg_types
+
+
+def _set_op_arguments(mx_operators):
```
larroy: I would put op assembling together in a single function and use a light class or named tuple instead of difficult to maintain deeply nested dictionaries.

benchmark/opperf/utils/op_registry_utils.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    for op_name in mx_operators:
+        op_handle = _get_op_handles(op_name)
+        narg, arg_names, arg_types = _get_op_arguments(op_name, op_handle)
+        mx_operators[op_name]["params"] = {"narg": narg,
+                                           "arg_names": arg_names,
+                                           "arg_types": arg_types}
+
+
+def _get_all_mxnet_operators():
```
larroy: see comment above

benchmark/opperf/utils/op_registry_utils.py
```diff
@@ -0,0 +1,193 @@
...
+    binary_broadcast_mx_operators = {}
+    for op_name, op_params in mx_operators.items():
+        if op_name.startswith("broadcast_") and op_params["params"]["narg"] == 2 and \
+                "lhs" in op_params["params"]["arg_names"] and \
+                "rhs" in op_params["params"]["arg_names"]:
+            binary_broadcast_mx_operators[op_name] = mx_operators[op_name]
+    return binary_broadcast_mx_operators
+
+
+def get_all_element_wise_binary_operators():
```
larroy: can we rename to "elemwise" for consistency with the C source files?

benchmark/opperf/utils/ndarray_utils.py
```diff
@@ -0,0 +1,95 @@
...
+    :param attach_grad: To attach a gradient for the Tensor. Default is True.
+    :return: MXNet NDArray Tensor.
+    """
+    if isinstance(in_tensor, tuple):
+        tensor = initializer(ctx=ctx, shape=in_tensor, dtype=dtype)
+    elif isinstance(in_tensor, np.ndarray):
+        tensor = nd.array(in_tensor, ctx=ctx, dtype=dtype)
+    elif isinstance(in_tensor, mx.ndarray):
+        tensor = in_tensor.as_in_context(ctx=ctx).astype(dtype=dtype)
+
```
larroy: can we remove ws?

larroy:
https://github.com/apache/incubator-mxnet/issues/15069

benchmark/opperf/utils/op_registry_utils.py
sandeep-krishnamurthy: 1. This logic is very specific to this utility so C API at this time may not be necessary. Also, we have C APIs to get list of operator names, then a C API to get by operator name the details of the operator params etc. so that should helps majority of users.
2. Refactored the helper code to split the responsibility. Thanks.

benchmark/opperf/utils/op_registry_utils.py
sandeep-krishnamurthy: Done

benchmark/opperf/utils/ndarray_utils.py
sandeep-krishnamurthy: done

benchmark/opperf/utils/op_registry_utils.py
sandeep-krishnamurthy: Getting a op handle by op name, getting a op arguments by op handle are all independent unit of work hence kept it separate. This assembling function just gets op handle, op argument and prepares a dictionary. I agree we could have a class / named tuple. But, to keep the utility simple, flexible and hackable kept it this way. For now, we have only 3 attributes, name, has_backward, params. I prefer to keep it this way for now and revisit this later if necessary. Let me know if you still strongly think otherwise.

benchmark/opperf/utils/op_registry_utils.py
sandeep-krishnamurthy: see above.

sandeep-krishnamurthy:
Updates:
1. Addressed @larroy comments.
2. Add support for automated unary operator benchmarks. (~54 operators)
3. Delete all TODO files to avoid confusion and add one readme with all todos.

@larroy @szha - Can you take a look? Thanks!

szha:
Impressive. Seems like the automation may have helped productivity. Sorry for not getting back to this earlier (been busy with events and releases). I will take a look this week. Thanks.

benchmark/opperf/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+# TODO
+
+All contributions are welcome. Below is the list of desired features:
+
+1. Cover all MXNet operators.
+2. Enhance MXNet profiler with additional APIs to programmatically fetch and process profiler data.
+3. Integration with CI/CD system to run operator benchmarks for PR builds, nightly builds.
+4. Dashboards and other modes of presentation of results for analyzing and planning tasks such as operator performance improvements.
+5. Integration with tools such as [Hypothesis](https://hypothesis.readthedocs.io/en/latest/) for randomized Tensor Shape generation for profiling to identify bottlenecks in the operators.
```
szha: Still not sure about this. Generating random input doesn't seem to require a new library. Either numpy.random or mxnet.random should do.

sandeep-krishnamurthy: Had missed to update this part. Yes numpy.random works.

sandeep-krishnamurthy:
> Could you include utility for listing the operators that are not covered, instead of writing TODO list in code?

https://github.com/apache/incubator-mxnet/pull/14977/files#diff-d7bc2931851dce319a4523bc3bb10ac7
I have a README listing things to do. I am updating it as things are progressing.

sandeep-krishnamurthy:
> Could you include utility for listing the operators that are not covered, instead of writing TODO list in code?

I have a readme file maintaining it - https://github.com/apache/incubator-mxnet/pull/14977/files#diff-d7bc2931851dce319a4523bc3bb10ac7

sandeep-krishnamurthy:
Updates:
1. Add support for Random Sampling operators (~16)
Total operators covered => ~100

szha:
That should be maintained automaticall instead of manually in a doc.

sandeep-krishnamurthy:
> That should be maintained automaticall instead of manually in a doc.

I understand, I was planning to do that later as I did not see that as blocker in phase 1. I can do that.

Are we good to go post that?

sandeep-krishnamurthy:
Updates:
1. Cover automated tests for reduction operators [~10].
2. Benchmarks for operators - pooling, conv, activations [~10]
Total operators covered so far ~130
3. Add utility function to give back list of operator names not covered in benchmarks.

@larroy @szha @apeforest - Can you please take a look at this PR? I would like to work towards getting this merged as I believe in the current state, this PR can add value to users and developers of MXNet. I will continue further operator coverage incrementally in new PRs.

benchmark/opperf/nd_operations/README.md
```diff
@@ -0,0 +1,107 @@
...
+<!--- specific language governing permissions and limitations -->
+<!--- under the License. -->
+
+# TODO - Operators to add to this Benchmark Utility
+
+Below are list of operators to be added to the benchmarking utility:
+
+**NN Operations**
+
+MXNet NDArray NN Operators
```
szha: Since this is now automated, is this doc needed anymore? Docs get stale quickly

sandeep-krishnamurthy: Agree. Missed to remove this. Updated with autogenerated list of operators not covered from the utility

benchmark/opperf/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+```
+python incubator-mxnet/benchmark/opperf/opperf.py --output-format json --output-file mxnet_operator_benchmark_results.json
+```
+
+**Other Supported Options:**
+
+1. **output-format** : `json` or `md` for markdown file output.
+
+2. **ctx** : `cpu` or `gpu`. By default, cpu on CPU machine, gpu(0) on GPU machine. You can override and set the global context for all operator benchmarks. Example: --ctx gpu(2).
```
apeforest: What if there are multiple GPUs? Does the profiler generate results per device?

benchmark/opperf/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+3. **dtype** : By default, `float32`. You can override and set the global dtype for all operator benchmarks. Example: --dtype float64.
+
+## Usecase 2 - Run benchmarks for all the operators in a specific category
+
+For example, you want to run benchmarks for all NDArray Broadcast Binary Operators, Ex: broadcast_add, broadcast_mod, broadcast_pow etc., You just run the following python script.
+
+```
+#!/usr/bin/python
+from benchmark.opperf.tensor_operations.binary_broadcast_operators import run_mx_binary_broadcast_operators_benchmarks
```
apeforest: nit: do we need the `_mx_` since its for mxnet only?

benchmark/opperf/nd_operations/nn_basic_operators.py
```diff
@@ -0,0 +1,83 @@
...
+"""
+
+
+def run_nn_basic_operators_benchmarks(ctx=mx.cpu(), dtype='float32', warmup=10, runs=50):
+    # FullyConnnected operator benchmarks
+    fc_benchmark_res = run_performance_test([nd.FullyConnected],
+                                            run_backward=True,
+                                            dtype=dtype,
+                                            ctx=ctx,
+                                            inputs=[{"data": (32, 3, 256, 256),
```
apeforest: Would it be more convenient to define this shape as a global constant?

benchmark/opperf/nd_operations/nn_basic_operators.py
```diff
@@ -0,0 +1,83 @@
...
+                                                          "p": 0.5,
+                                                          "mode": "always"}],
+                                                 warmup=warmup,
+                                                 runs=runs)
+    # BatchNorm benchmarks
+    batchnorm_benchmark_res = run_performance_test([nd.BatchNorm],
+                                                   run_backward=True,
+                                                   dtype=dtype,
+                                                   ctx=ctx,
+                                                   inputs=[{"data": (32, 3, 256, 256),
```
apeforest: Same for all these hyperparameters. It would be less error prone to define global constants

benchmark/opperf/nd_operations/nn_conv_operators.py
```diff
@@ -0,0 +1,139 @@
...
+                                                                      "kernel": 3,
+                                                                      "pool_type": pool_type,
+                                                                      "global_pool": global_pool,
+                                                                      "stride": 1,
+                                                                      "pad": 1,
+                                                                      "layout": 'NCW'}
+                                                                     ],
+                                                             warmup=warmup,
+                                                             runs=runs)
+            for pool2d_data in [(32, 3, 256, 256), (32, 3, 64, 64)]:
```
apeforest: Why benchmark for two different input shapes?

benchmark/opperf/nd_operations/random_sampling_operators.py
```diff
@@ -0,0 +1,61 @@
...
+
+"""
+
+import mxnet as mx
+
+from benchmark.opperf.utils.benchmark_utils import run_op_benchmarks
+from benchmark.opperf.utils.op_registry_utils import get_all_random_sampling_operators
+
+
+def run_mx_random_sampling_operators_benchmarks(ctx=mx.cpu(), dtype='float32', warmup=10, runs=50):
```
apeforest: For benchmarking random operators, do we want to set a fixed seed?

benchmark/opperf/opperf.py
```diff
@@ -0,0 +1,133 @@
...
+        device_id = int(ctx[4:-1])
+        return mx.gpu(device_id)
+
+
+if __name__ == '__main__':
+    # CLI Parser
+
+    # 1. GET USER INPUTS
+    parser = argparse.ArgumentParser(
+        description='Run all the MXNet operators (NDArray) benchmarks')
```
apeforest: nit: keep indentation consistent

benchmark/opperf/opperf.py
```diff
@@ -0,0 +1,133 @@
...
+
+    parser.add_argument('--ctx', type=str, default='cpu',
+                        help='Global context to run all benchmarks. By default, cpu on a '
+                             'CPU machine, gpu(0) on a GPU machine. '
+                             'Valid Inputs - cpu, gpu, gpu(0), gpu(1)...')
+    parser.add_argument('--dtype', type=str, default='float32', help='DType (Precision) to run benchmarks. By default, '
```
apeforest: not supporting int32, int64?

benchmark/opperf/opperf.py
```diff
@@ -0,0 +1,133 @@
...
+                                                                     'float32. Valid Inputs - float32, float64.')
+    parser.add_argument('--output-format', type=str, default='json',
```
apeforest: Do we need to provide markdown format? Can we just let user use https://www.npmjs.com/package/json2md if they need md format?

benchmark/opperf/nd_operations/nn_basic_operators.py
apeforest: Now I see you have a benchmark/opperf/rules/default_params.py. Why not just use those DEFAULT_SHAPE?

benchmark/opperf/README.md
sandeep-krishnamurthy: Since this single operator benchmarks, it runs only on one device and profiler output is only for that device.

benchmark/opperf/README.md
sandeep-krishnamurthy: May be not, just kept it this way just in case will spin off this tool outside MXNet and extend to other framework for comparison.

benchmark/opperf/nd_operations/nn_conv_operators.py
sandeep-krishnamurthy: Observe behavior in case of small and large tensors

benchmark/opperf/opperf.py
sandeep-krishnamurthy: done

benchmark/opperf/opperf.py
sandeep-krishnamurthy: It is supported. All DTypes supported by MXNet is supported.

benchmark/opperf/opperf.py
sandeep-krishnamurthy: Yes. I thought, usually we want to run this utility and generate a kind of report easy to view / consume manually hence added.

benchmark/opperf/nd_operations/random_sampling_operators.py
szha: is there any case where the random operator changes its runtime based on seed?

szha:
Thanks for making the changes. The automated information from this PR would be very useful as a dashboard in our wiki (along with the public nightly test results for example).

benchmark/opperf/nd_operations/nn_basic_operators.py
larroy: why global?

benchmark/opperf/nd_operations/nn_basic_operators.py
larroy: I think the shapes we should autodiscover actually

Issue #14978, opened 2019-05-17
Flaky Test: test_tensorrt_lenet5.test_tensorrt_inference


haojin2:
Happening for 2 independent PRs.
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-14959/9/pipeline
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-14830/19/pipeline

asitstands:
Another failure is here. http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-10951/33/pipeline

apeforest:
one more: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-15041/5/pipeline/317

perdasilva:
I've submitted a PR to fix it: #15014

apeforest:
https://github.com/apache/incubator-mxnet/pull/15014 has fixed this issue. Closing it.

anirudh2290:
Reopening, still encountered this here: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-15118/13/pipeline

PR #14992, opened 2019-05-18
[MXNET-978] Support higher order gradient for `log`.

kshitij12345:
## Description ##
With reference to #14613,  #10002 , this PR intends to add support for higher order gradient for `log` { and ideally for `log2, log10` } as well.

Tests are based totally on #14613 

### Essentials ###
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/browse/MXNET-978) created (except PRs with tiny changes)

### Changes ###
- [x] higher order gradient for a `log`.
- [x] unit test for the same.

### DIFF
--- a/src/operator/tensor/elemwise_unary_op_basic.cc
+++ b/src/operator/tensor/elemwise_unary_op_basic.cc
@@ -1069,13 +1069,72 @@ The storage type of ``log2`` output is always dense
 .set_attr<nnvm::FGradient>("FGradient", ElemwiseGradUseIn{"_backward_log2"});
 
 MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_log,
-                                                  unary_bwd<mshadow_op::log_grad>);
+                                                  unary_bwd<mshadow_op::log_grad>)
+.set_attr<nnvm::FGradient>("FGradient",
+  [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
+    // For g(x) -> g = log
+    // g''(x) = -1 * (g'(x) * g'(x))
+    auto gx = nnvm::NodeEntry{n, 0, 0};
+    auto ggx_mid = MakeNode("elemwise_mul", n->attrs.name + "_backward_mid_grad_grad",
+                            {gx, gx}, nullptr, &n);
+    auto ggx = MakeNode("negative", n->attrs.name + "_backward_grad_grad",
+                        {nnvm::NodeEntry{ggx_mid, 0, 0}}, nullptr, &n);
+
+    std::vector<nnvm::NodeEntry> ret;
+
+    ret.emplace_back(MakeNode("elemwise_mul", n->attrs.name + "_backward_grad_grad",
+                             {ograds[0], gx}, nullptr, &n));
+    ret.emplace_back(MakeNode("elemwise_mul", n->attrs.name + "_backward_grad_grad_inp",
+                             {ograds[0], nnvm::NodeEntry{ggx, 0, 0}}, nullptr, &n));
+    return ret;
+  });
 
 MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_log10,
-                                                  unary_bwd<mshadow_op::log10_grad>);
+                                                  unary_bwd<mshadow_op::log10_grad>)
+.set_attr<nnvm::FGradient>("FGradient",
+  [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
+    // For g(x) -> g = log10
+    // g'(x) = 1 / (log(10) * x)
+    // g''(x) = -1 * (g'(x) * 1/x)
+    auto gx = nnvm::NodeEntry{n, 0, 0};
+    auto g_lx = MakeNode("reciprocal", n->attrs.name + "_backward_log_grad",
+                            {n->inputs[1]}, nullptr, &n);
+    auto ggx_mid = MakeNode("elemwise_mul", n->attrs.name + "_backward_mid_grad_grad",
+                            {gx, nnvm::NodeEntry{g_lx, 0, 0}}, nullptr, &n);
+    auto ggx = MakeNode("negative", n->attrs.name + "_backward_grad_grad",
+                        {nnvm::NodeEntry{ggx_mid, 0, 0}}, nullptr, &n);
+
+    std::vector<nnvm::NodeEntry> ret;
+
+    ret.emplace_back(MakeNode("elemwise_mul", n->attrs.name + "_backward_grad_grad",
+                             {ograds[0], gx}, nullptr, &n));
+    ret.emplace_back(MakeNode("elemwise_mul", n->attrs.name + "_backward_grad_grad_inp",
+                             {ograds[0], nnvm::NodeEntry{ggx, 0, 0}}, nullptr, &n));
+    return ret;
+  });
 
 MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_log2,
-                                                  unary_bwd<mshadow_op::log2_grad>);
+                                                  unary_bwd<mshadow_op::log2_grad>)
+.set_attr<nnvm::FGradient>("FGradient",
+  [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
+    // For g(x) -> g = log2
+    // g'(x) = 1 / (log(2) * x)
+    // g''(x) = -1 * (g'(x) * 1/x)
+    auto gx = nnvm::NodeEntry{n, 0, 0};
+    auto g_lx = MakeNode("reciprocal", n->attrs.name + "_backward_log_grad",
+                            {n->inputs[1]}, nullptr, &n);
+    auto ggx_mid = MakeNode("elemwise_mul", n->attrs.name + "_backward_mid_grad_grad",
+                            {gx, nnvm::NodeEntry{g_lx, 0, 0}}, nullptr, &n);
+    auto ggx = MakeNode("negative", n->attrs.name + "_backward_grad_grad",
+                        {nnvm::NodeEntry{ggx_mid, 0, 0}}, nullptr, &n);
+
+    std::vector<nnvm::NodeEntry> ret;
+    ret.emplace_back(MakeNode("elemwise_mul", n->attrs.name + "_backward_grad_grad",
+                             {ograds[0], gx}, nullptr, &n));
+    ret.emplace_back(MakeNode("elemwise_mul", n->attrs.name + "_backward_grad_grad_inp",
+                             {ograds[0], nnvm::NodeEntry{ggx, 0, 0}}, nullptr, &n));
+    return ret;
+  });
 
 // log1p
 MXNET_OPERATOR_REGISTER_UNARY_WITH_RSP_CSR(log1p, cpu, mshadow_op::log1p)
--- a/src/operator/tensor/elemwise_unary_op_basic.cc
+++ b/src/operator/tensor/elemwise_unary_op_basic.cc
@@ -1074,18 +1074,18 @@ MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_log,
   [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
     // For g(x) -> g = log
     // g''(x) = -1 * (g'(x) * g'(x))
-    auto gx = nnvm::NodeEntry{n, 0, 0};
+    auto gx = nnvm::NodeEntry{n};
     auto ggx_mid = MakeNode("elemwise_mul", n->attrs.name + "_backward_mid_grad_grad",
                             {gx, gx}, nullptr, &n);
     auto ggx = MakeNode("negative", n->attrs.name + "_backward_grad_grad",
-                        {nnvm::NodeEntry{ggx_mid, 0, 0}}, nullptr, &n);
+                        {nnvm::NodeEntry{ggx_mid}}, nullptr, &n);
 
     std::vector<nnvm::NodeEntry> ret;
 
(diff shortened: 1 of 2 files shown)


src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -1016,7 +1016,28 @@ The storage type of ``log2`` output is always dense
 .set_attr<nnvm::FGradient>("FGradient", ElemwiseGradUseIn{"_backward_log2"});
 
 MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_log,
-                                                  unary_bwd<mshadow_op::log_grad>);
+                                                  unary_bwd<mshadow_op::log_grad>)
```
kshitij12345: I don't know much about this library but,

I believe it would be better to have gradients defined for  existing backward, instead of a differentiable gradient (relying on  autograd machinery) at least on `ops`  where backward is not trivial. It will allow to use existing optimised fused kernels and make sure there is no regression in the backward.

Note: `log` is relatively trivial (single reciprocal). But maybe we may see a performance regression for `sigmoid`, if we do it by relying on autograd machinery instead of the existing `_backward_sigmoid`.

kshitij12345:
I don't know much about this library but,

I believe it would be better to have gradients defined for  existing backward, instead of a differentiable gradient (relying on  autograd machinery) at least on `ops`  where backward is not trivial. It will allow to use existing optimised fused kernels and make sure there is no regression in the backward.

Note: `log` is relatively trivial (single reciprocal). But maybe we may see a performance regression for `sigmoid`, if we do it by relying on autograd machinery instead of the existing `_backward_sigmoid`.

kshitij12345:
Can anyone please point how I can have `1/log(2.0), 1/log(10.0)` multiplied with gradient for `log2, log10`.

pinaraws:
@mxnet-label-bot add[Operator, pr-awaiting-review]

apeforest:
@kshitij12345 Thanks for your contribution. I agree with you it would be better to have gradients defined for existing backward operators. 

I do not fully understand your question of 1/log(2.0), 1/log(10.0) multiplied with gradient for log2, log 10. Could you please elaborate?

kshitij12345:
> I do not fully understand your question of 1/log(2.0), 1/log(10.0) multiplied with gradient for log2, log 10. Could you please elaborate?

Reading it again, I phrased it poorly. Sorry. So actually, plan was to update `gradient` for `log2`, which would be `1/(log(2.0)  * x)`, for which I would have required a `log(2.0)`. So how to get that, is scalar multiplication allowed ? or `ones_like` followed by `fill`.  

**Note**: This is not need for this PR. But curious to know.

Thank You.

kshitij12345:
@larroy @apeforest Have updated as per #14095 .
Please review.

kshitij12345:
Failure looks irrelevant to the PR.

http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-14992/7/pipeline#step-767-log-360

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -1069,13 +1069,72 @@ The storage type of ``log2`` output is always dense
...
+.set_attr<nnvm::FGradient>("FGradient",
+  [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
+    // For g(x) -> g = log
+    // g''(x) = -1 * (g'(x) * g'(x))
+    auto gx = nnvm::NodeEntry{n, 0, 0};
```
larroy: You are welcome to simplify calls to NodeEntry as per:

https://github.com/apache/incubator-mxnet/pull/14095

Just call nnvm::NodeEntry{n}

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -1069,13 +1069,72 @@ The storage type of ``log2`` output is always dense
...
+    auto ggx_mid = MakeNode("elemwise_mul", n->attrs.name + "_backward_mid_grad_grad",
+                            {gx, gx}, nullptr, &n);
+    auto ggx = MakeNode("negative", n->attrs.name + "_backward_grad_grad",
+                        {nnvm::NodeEntry{ggx_mid, 0, 0}}, nullptr, &n);
+
+    std::vector<nnvm::NodeEntry> ret;
+
+    ret.emplace_back(MakeNode("elemwise_mul", n->attrs.name + "_backward_grad_grad",
+                             {ograds[0], gx}, nullptr, &n));
+    ret.emplace_back(MakeNode("elemwise_mul", n->attrs.name + "_backward_grad_grad_inp",
```
larroy: Why are we returning two gradients, isn't it an unary function with just one input?

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -1069,13 +1069,72 @@ The storage type of ``log2`` output is always dense
...
+    auto g_lx = MakeNode("reciprocal", n->attrs.name + "_backward_log_grad",
+                            {n->inputs[1]}, nullptr, &n);
+    auto ggx_mid = MakeNode("elemwise_mul", n->attrs.name + "_backward_mid_grad_grad",
+                            {gx, nnvm::NodeEntry{g_lx, 0, 0}}, nullptr, &n);
+    auto ggx = MakeNode("negative", n->attrs.name + "_backward_grad_grad",
+                        {nnvm::NodeEntry{ggx_mid, 0, 0}}, nullptr, &n);
+
+    std::vector<nnvm::NodeEntry> ret;
+
+    ret.emplace_back(MakeNode("elemwise_mul", n->attrs.name + "_backward_grad_grad",
```
larroy: same comment as above.

src/operator/tensor/elemwise_unary_op_basic.cc
kshitij12345: `log` is unary, but the `_backward_log` isn't. 
https://github.com/apache/incubator-mxnet/blob/016556553ced0c9c9400efec2f08a3c39cbb00ba/src/operator/tensor/elemwise_unary_op_basic.cc#L1071

It takes as input the output gradient and input to log.
https://github.com/apache/incubator-mxnet/blob/016556553ced0c9c9400efec2f08a3c39cbb00ba/src/operator/tensor/elemwise_unary_op_basic.cc#L1045

src/operator/tensor/elemwise_unary_op_basic.cc
kshitij12345: Oh. Missed that. Thank You.

apeforest:
> > I do not fully understand your question of 1/log(2.0), 1/log(10.0) multiplied with gradient for log2, log 10. Could you please elaborate?
> 
> Reading it again, I phrased it poorly. Sorry. So actually, plan was to update `gradient` for `log2`, which would be `1/(log(2.0) * x)`, for which I would have required a `log(2.0)`. So how to get that, is scalar multiplication allowed ? or `ones_like` followed by `fill`.
> 
> **Note**: This is not needed for this PR. But curious to know.
> 
> Thank You.

Can you just add an elemwise_mul operator node between them?

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -1069,13 +1069,73 @@ The storage type of ``log2`` output is always dense
 .set_attr<nnvm::FGradient>("FGradient", ElemwiseGradUseIn{"_backward_log2"});
 
 MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_log,
-                                                  unary_bwd<mshadow_op::log_grad>);
+                                                  unary_bwd<mshadow_op::log_grad>)
+.set_attr<nnvm::FGradient>("FGradient",
+  [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
+    // For g(x) -> g = log
```
apeforest: nit: It's very nice to see a comment here. The g(x) is actually a function of x. It might be easily confused with the variable gx two lines below. Maybe use `f(x)` in the comment here?

apeforest:
@kshitij12345 The CI failure in unix-GPU was due to a flaky test for TensorRT: https://github.com/apache/incubator-mxnet/issues/14978

The issue has been fixed by https://github.com/apache/incubator-mxnet/pull/15014. Please re-trigger CI again. Thanks!

src/operator/tensor/elemwise_unary_op_basic.cc
kshitij12345: Ah sure . That makes sense. Thank You.

larroy:
Lgtm

larroy:
"I believe it would be better to have gradients defined for existing backward, instead of a differentiable gradient (relying on autograd machinery) at least on ops where backward is not trivial. It will allow to use existing optimised fused kernels and make sure there is no regression in the backward."

I think you are right, we have discussed this before. I think we rushed merging this PR. We should do a check for regressions.

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -1069,13 +1069,73 @@ The storage type of ``log2`` output is always dense
...
 
 MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_log10,
-                                                  unary_bwd<mshadow_op::log10_grad>);
+                                                  unary_bwd<mshadow_op::log10_grad>)
+.set_attr<nnvm::FGradient>("FGradient",
+  [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
+    // For f(x) -> f = log10
+    // f'(x) = 1 / (log(10) * x)
+    // f''(x) = -1 * (f'(x) * 1/x)
+    auto gx = nnvm::NodeEntry{n, 0, 0};
```
larroy: Why don't we follow the same pattern as in the natural logarithm?

kshitij12345: For natural `log`,
we have with us in `gradient` function, `gx` i.e. `1/x` as well as `x`.
Since, second derivative of log is `-(gx * gx)` = `-1/(x^2)`. We use the pattern.

Considering `log2` (similar case for `log10`)
we have with us, `gx` i.e. `1/(log(2) * x)` as well as `x`.
Since second derivative is `-1/(log(2) * x * x)`
which we get in the code using `negative(gx * reciprocal(x))`, where `gx=1/(log(2) * x`.
Another way to get that will be `negative(gx * gx * log(2.0))`.

kshitij12345: @larroy Thanks for pointing this, going through this again made me realise that there is a problem with the implementation of `log`.

kshitij12345:
@apeforest @larroy 

https://github.com/kshitij12345/incubator-mxnet/blob/7b343d1fcde73b61322985580080333d9eee9e82/src/operator/tensor/elemwise_unary_op_basic.cc#L1077-L1079

We multiply  `gx * gx` where `gx = ograd * f'(x)`, getting `ograd^2 * f'(x)^2`, however we want only `ograd * f'(x)^2` which can be achieved in a similar fashion to the implementation of `_backward_log10/2`.

I have validated the expected results.
```python
import numpy
import math

x.attach_grad()
with autograd.record():
  y = nd.log(x)
  y_grad = autograd.grad(y, x, head_grads= nd.ones_like(y) * 0.5, create_graph=True, retain_graph=True)[0]
y_grad.backward(nd.ones_like(y_grad) * 0.6)

numpy.testing.assert_allclose(x.grad.asnumpy() , ( grad_grad_op(x) * 0.5 * 0.6).asnumpy(), rtol=1e-7, atol=1e-7)
```
Which fails with current code.
**Should make a new PR, or add commits in this PR itself?**

Have confirmed the behaviour with Pytorch as well.
```python
import torch
import numpy
import math

x = torch.randn(2,3)
x.requires_grad = True

y = torch.log(x)
y_grad = torch.autograd.grad(y, x, grad_outputs=torch.ones_like(y) * 0.5, create_graph=True, retain_graph=True)[0]
y_grad.backward(torch.ones_like(y_grad) * 0.6)

numpy.testing.assert_allclose(x.grad.detach().numpy() , ( grad_grad_op(x) * 0.5 * 0.6).detach().numpy(), rtol=1e-7, atol=1e-7)
```

apeforest:
Hi @kshitij12345 sorry, we missed that. We should have reviewed it more carefully. Please submit another PR to fix this issue. I will also update https://github.com/apache/incubator-mxnet/pull/14613 accordingly. Thanks!

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -1069,13 +1069,73 @@ The storage type of ``log2`` output is always dense
...
+                            {n->inputs[1]}, nullptr, &n);
+    auto ggx_mid = MakeNode("elemwise_mul", n->attrs.name + "_backward_mid_grad_grad",
+                            {gx, nnvm::NodeEntry{g_lx}}, nullptr, &n);
+    auto ggx = MakeNode("negative", n->attrs.name + "_backward_grad_grad",
+                        {nnvm::NodeEntry{ggx_mid}}, nullptr, &n);
+
+    std::vector<nnvm::NodeEntry> ret;
+
+    ret.emplace_back(MakeNode("elemwise_mul", n->attrs.name + "_backward_grad_grad",
+                             {ograds[0], gx}, nullptr, &n));
```
apeforest: Shouldn't this be `{ograds[0], g_lx}` instead? Isn't `dL/dy_grad = d^2L/dx^2 * f'(x)`?

kshitij12345: Yes, it should. Thanks. I have updated this change and added relevant test in the new PR #15120 .
Actually I am having trouble exactly at this part as the `grad` value is not being updated. More info in #15120

kshitij12345:
Have created a new PR for the same. #15120 Please review

Issue #14954, opened 2019-05-15
 broadcast_to() behavior change after new 0D and uninitialized shape support added?


DickJC123:
The documentation for broadcast_to(..., shape=<output_shape>,...) suggests that '0' can appear as a placeholder that means 'keep the same dimension as the input' for the given dimension.  However, the C++ code requires -1 for this.  Which is correct?  Are we changing the behavior?  @reminisce @KellenSunderland 

Documentation:
```
Broadcasting is allowed on axes with size 1, such as from (2,1,3,1) to (2,8,3,9). Elements will be duplicated on the broadcasted axes.  For example:

broadcast_to([[1,2,3]], shape=(2,3)) = [[ 1.,  2.,  3.],
                                        [ 1.,  2.,  3.]])

The dimension which you do not want to change can also be kept as 0 which means copy the original value. So with shape=(2,0), we will obtain the same result as in the above example.

Defined in src/operator/tensor/broadcast_reduce_op_value.cc:L262
```
But in the C++ code, we have:

https://github.com/apache/incubator-mxnet/blob/1eba37a8c3cb3a64efa0b52e79a3af7a6e7e5a57/src/operator/tensor/broadcast_reduce_op.h#L380-L401

For Q & A and discussion, please start a discussion thread at https://discuss.mxnet.io 

## Description
(Brief description of the problem in no more than 2 sentences.)

## Environment info (Required)

```
What to do:
1. Download the diagnosis script from https://raw.githubusercontent.com/apache/incubator-mxnet/master/tools/diagnose.py
2. Run the script using `python diagnose.py` and paste its output here.

```

Package used (Python/R/Scala/Julia):
(I'm using ...)

For Scala user, please provide:
1. Java version: (`java -version`)
2. Maven version: (`mvn -version`)
3. Scala runtime if applicable: (`scala -version`)

For R user, please provide R `sessionInfo()`:

## Build info (Required if built from source)

Compiler (gcc/clang/mingw/visual studio):

MXNet commit hash:
(Paste the output of `git rev-parse HEAD` here.)

Build config:
(Paste the content of config.mk, or the build command.)

## Error Message:
(Paste the complete error message, including stack trace.)

## Minimum reproducible example
(If you are using your own code, please provide a short script that reproduces the error. Otherwise, please provide link to the existing example.)

## Steps to reproduce
(Paste the commands you ran that produced the error.)

1.
2.

## What have you tried to solve it?

1.
2.

[...]

PR #14998, opened 2019-05-20
Revert the change on broadcast_to param shape

reminisce:
## Description ##
This PR reverts the mistaken change on `broadcast_to`'s param shape introduced by https://github.com/apache/incubator-mxnet/pull/14661. There was no unit test for the param shape with `0` as dimension size so this error was not caught. Such case has been added in unit tests to ensure correctness.

Thank @DickJC123 for reporting this in https://github.com/apache/incubator-mxnet/issues/14954.

### Essentials ###

### DIFF
--- a/src/operator/tensor/broadcast_reduce_op.h
+++ b/src/operator/tensor/broadcast_reduce_op.h
@@ -379,7 +379,7 @@ inline bool BroadcastAxesShape(const nnvm::NodeAttrs& attrs,
 
 inline bool BroadcastToShape(const nnvm::NodeAttrs& attrs,
                              mxnet::ShapeVector *in_attrs,
-                            mxnet::ShapeVector *out_attrs) {
+                             mxnet::ShapeVector *out_attrs) {
   CHECK_EQ(in_attrs->size(), 1U);
   CHECK_EQ(out_attrs->size(), 1U);
   mxnet::TShape& ishape = (*in_attrs)[0];
@@ -389,7 +389,7 @@ inline bool BroadcastToShape(const nnvm::NodeAttrs& attrs,
     << "Operand of shape " << ishape << " cannot be broadcasted to " << param.shape;
   mxnet::TShape oshape = param.shape;
   for (int i = 0; i < ishape.ndim(); ++i) {
-    if (oshape[i] != -1) {
+    if (oshape[i] != 0) {
       CHECK(ishape[i] == oshape[i] || ishape[i] == 1)
         << "Array cannot be broadcasted from " << ishape << " to " << param.shape;
     } else {
--- a/tests/python/unittest/test_operator.py
+++ b/tests/python/unittest/test_operator.py
@@ -2535,19 +2535,27 @@ def test_broadcast():
         size = tuple([shape[ele] for ele in axis])
         for ele in axis:
             shape[ele] = 1
+        target_shape_with_zero = list(target_shape)
+        for idx in range(len(target_shape_with_zero)):
+            if idx not in axis:
+                target_shape_with_zero[idx] = 0
+                break
+
         a = mx.symbol.Variable('a')
         sym_bcast_axis = mx.symbol.broadcast_axis(a, axis=axis, size=size)
         sym_bcast_to = mx.symbol.broadcast_to(a, shape=tuple(target_shape))
+        sym_bcast_to_with_zero = mx.symbol.broadcast_to(a, shape=tuple(target_shape_with_zero))
         sym_bcast_like = mx.symbol.broadcast_like(a, sym_bcast_to)
+
         def test_broadcasting_ele(sym_bcast):
             dat_npy = np.random.rand(*shape)
             groundtruth = dat_npy
             grad_nd = mx.nd.empty(shape)
             outgrad_npy = np.random.rand(*target_shape)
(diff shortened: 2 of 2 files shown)


reminisce:
@DickJC123 Could you please give a review? Thanks.

DickJC123:
Trying to pop my to-do stack.  Will look at this by EOD tomorrow.

DickJC123:
Thank you for your patience as I dealt with other MXNet fires.  I still feel I want to think about the Shape serialization to files.  But in the meantime, let me ask about the issue that started it all, and that you're addressing with this PR:

Isn't the real problem that the Python output shape parameter of broadcast_to() isn't being converted correctly by the parameter parser?  In the default mode (so not numpy compatibility), a shape dimension size of 0 should be converted to -1 in the backend, which is what the broadcast_to() infer shape routine is already (i.e. correctly) looking for?  In numpy compatibility mode, the user should use -1 for the output shape in Python to denote unknown.

reminisce:
@DickJC123 Thanks for the review. For this specific problem, it actually has nothing to do with numpy compatibility mode. The document says that `0` just means the output dimension size is the same as the corresponding input dimension size; it's not related to any judgement on whether the shape dimension size should be 0 or unknown. So per document, `0` should never be changed to `-1` in whatever situation, but it was changed by a mistake. This is the same as MXNet `reshape` operator where the shape parameter has `0`, `-1`, `-2`, `-3`, and `-4` for different shape inference maneuvers.

omarroberts40:
Stop
    On Thursday, May 23, 2019, 8:29:55 PM PDT, reminisce <notifications@github.com> wrote:  
 
 
@DickJC123 Thanks for the review. For this specific problem, it actually has nothing to do with numpy compatibility mode. The document says that 0 just means the output dimension size is the same as the corresponding input dimension size; it's not related to any judgement on whether the shape dimension size should be 0 or unknown. So per document, 0 should never be changed to -1 in whatever situation, but it was changed by a mistake. This is the same as MXNet reshape operator where the shape parameter has 0, -1, -2, -3, and -4 for different shape inference maneuvers.

fhieber:
I would appreciate a timely merge of this fix. Thank you!

DickJC123:
I'm close to giving this the thumbs up (see the end of my comments).  I think a PR relating to Shape serialization (as we have discussed) should be separate from this one because of the possible extended discussion, given the 1.5 release deadline and @fhieber's eagerness for this fix.

On this PR, let me try one more time with my argument.  Consider numpy broadcast_to() behavior:
```
Python 3.5.2 (default, Nov 12 2018, 13:43:14) 
[GCC 5.4.0 20160609] on linux
>>> import numpy as np
>>> x = np.array([1,2,3])
>>> np.broadcast_to(x, (3,3))
array([[1, 2, 3],
       [1, 2, 3],
       [1, 2, 3]])
>>> np.broadcast_to(x, (0,3))
array([], shape=(0, 3), dtype=int64)
>>> 
```
If we say that MXNet's broadcast_to() uses 0 to mean 'match input', regardless of compatibility mode, then we could never duplicate numpy's behavior.  However, if we think of 0 to mean 'unknown' in non-compatibility mode, and we say that 'unknown' really means 'unknown until inputs are bound, then match input size', we'd retain numpy functionality:

In numpy compatibility mode, MXNet could broadcast to a shape with a zero size as in:
 mx.sym.broadcast_to(x, (0,3))  and boadcast_to() an input-matching shape as in mx.sym.broadcast_to(x, (-1,3)) [this being an extension to numpy capability].

And speaking of which, is there a reason you chose -1 for unknown size in numpy compatibility mode? Why not None as in (None, 3) for the example above?  If you chose None, then even MXNet reshape functionality could be considered a compatible superset of Numpy's (in compatibility mode):

Numpy reshape:
0    = 0 size
-1   = calculate dim so total elements match

MXNet not in numpy-compatible mode:
0    = output size matches input [in contrast to numpy definition]
-1   = calculate dim so total elements match
-2 -> -4  other MXNet-specific definitions

MXNet numpy-compatible mode:
None    = unknown size (until binding), then output size matches input
0     = 0 size [like numpy]
-1   = calculate dim so total elements match [like numpy]
-2 -> -4  other MXNet-specific definitions

Having typed this up, I see one problem- what value holds 'None' size in the backend, if -1 might have a meaning as it does in reshape?  It could be defined to have a non-conflicting value as in `enum ShapeSizeConstants { kUnknownSize = MIN_INT }` and tested via:
```
inline bool dim_size_is_known(const dim_t dim_size) {  return dim_size != kUnknownSize; }
```

Anyway thanks for reading this far.  Unless this really resonates with you, I'll leave it by saying *I support this PR as it stands*- this corrects many of the Sockeye test failures I was seeing (the rest related to Shape serialization, not broadcast_to()).

reminisce:
@DickJC123 Thanks for the deep analysis. I can see the confusion of this numpy compatibility flag. We decided to change it to another name called `np_shape` in this PR: https://github.com/apache/incubator-mxnet/pull/15063

Please know that this flag is designed for only affecting the incoming MXNet-NumPy operators which are not in the master branch for now. Legacy operators are NOT expected to comply with NumPy when this flag is set `True`. We are not going to announce any numpy related design/implementation in 1.5 release notes.

DickJC123:
Thanks for your explanation.  Again, this PR LGTM.

PR #14963, opened 2019-05-15
Add matrix inversion operator in linalg

arcadiaphy:
As title.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #15007, opened 2019-05-20
Add matrix determinant operator in linalg

arcadiaphy:
## Description ##
The second PR on linalg enhancement. Three operators regarding matrix determinant is added: det, logdet, slogdet.

Something worth mentioning in this PR:

1. Matrix inversion code is refactored a little bit to remove duplicated workspace query in getri, and make it more readable and easy to use in det operators.
2. Log(det(x)) in logdet follows the log operator, the grad is passed backwards even when det(x) < 0.
3. Sign in slogdet follows the sign operator, the grad on it is ignored in backward pass since it's not properly defined.
4. The determinant calculation uses [LU factorization with partial pivoting](https://en.wikipedia.org/wiki/LU_decomposition#Computing_the_determinant).
5. The grad of determinant is derived from [Jacobi's formula](https://en.wikipedia.org/wiki/Jacobi%27s_formula), which has a pretty friendly closed form solution for numerical computing when input matrix A is invertible. The non-invertible case is not easy to implement since it involves adjugate matrix. In tensorflow, this case is ignored; while pytorch uses SVD to compute the grad. In this PR, it's left for future work, and now as a temporary method, no grad is passed backwards when det = 0. My inclination is to re-use LU instead of SVD for non-invertible case since it's already calculated.

### Essentials ###

### Changes ###

## Comments ##


pinaraws:
@mxnet-label-bot add[Operator, pr-work-in-progress]

arcadiaphy:
@eric-haibin-lin @reminisce @apeforest @anirudh2290 
Is anyone free to review my enhancement on linalg package? Also my last PR #14963 on matrix inversion is not fully reviewed, so please help review that part too. Thanks!

arcadiaphy:
The following is the gradient of added operators:

Matrix inversion



Matrix determinant (A is invertible)



Matrix log determinant and signed log determinant (A is invertible)

szha:
cc @asmushetzel

src/operator/tensor/la_op.cc
```diff
@@ -939,5 +939,153 @@ NNVM_REGISTER_OP(_backward_linalg_inverse)
...
+  *out* = *det(A)*
+
+If *n>2*, *det* is performed separately on the trailing two dimensions
+for all inputs (batch mode).
+
+.. note:: The operator supports float32 and float64 data types only.
+
+Examples::
+
+   // Single matrix inversion
```
reminisce: I don't think `//` is needed here since this is not a piece of c++ code.
inversion -> determinant

src/operator/tensor/la_op.cc
```diff
@@ -939,5 +939,153 @@ NNVM_REGISTER_OP(_backward_linalg_inverse)
...
+   A = [[1., 4.], [2., 3.]]
+   det(A) = [-5.]
+
+   // Batch matrix inversion
```
reminisce: Same as above.

src/operator/tensor/la_op.h
```diff
@@ -413,6 +413,50 @@ inline bool InverseShape(const nnvm::NodeAttrs& attrs,
...
 }
 
+// Shape inference function for det functions in linalg
+template<int onum>
+inline bool DetShape(const nnvm::NodeAttrs& attrs,
+                     mxnet::ShapeVector* in_attrs,
+                     mxnet::ShapeVector* out_attrs) {
+  CHECK_EQ(in_attrs->size(), 1);
+  CHECK_EQ(out_attrs->size(), onum + 2);
+  const mxnet::TShape& in = (*in_attrs)[0];
```
reminisce: Please add this line of code here:
```cpp
if (!ndim_is_known(in)) return false;
```

src/operator/tensor/la_op.h
```diff
@@ -413,6 +413,50 @@ inline bool InverseShape(const nnvm::NodeAttrs& attrs,
...
+    out = mxnet::TShape(1, 1);
+  } else {
+    out = mxnet::TShape(in.begin(), in.end() - 2);
+  }
+  for (int i = 0; i < onum; ++i) {
+    SHAPE_ASSIGN_CHECK(*out_attrs, i, out); /* sign or det or logdet */
+  }
+  SHAPE_ASSIGN_CHECK(*out_attrs, onum, in); /* LU */
+  SHAPE_ASSIGN_CHECK(*out_attrs, onum + 1, mxnet::TShape(in.begin(), in.end() - 1)); /* pivot */
+  return true;
```
reminisce: replace this with
```cpp
return shape_is_known(in);
```

src/operator/tensor/la_op.h
```diff
@@ -413,6 +413,50 @@ inline bool InverseShape(const nnvm::NodeAttrs& attrs,
...
+template<int onum>
+inline bool DetType(const nnvm::NodeAttrs& attrs,
+                    std::vector<int>* in_type,
+                    std::vector<int>* out_type) {
+  using namespace mshadow;
+  CHECK_EQ(in_type->size(), 1U);
+  int dtype = (*in_type)[0];
+  CHECK_NE(dtype, -1) << "Input must have specified type";
+
+  out_type->clear();
```
reminisce: Please use `TYPE_ASSIGN_CHECK` for every output type assignment.

src/operator/tensor/la_op.h
```diff
@@ -413,6 +413,50 @@ inline bool InverseShape(const nnvm::NodeAttrs& attrs,
...
+
+// Type inference function for det functions in linalg
+template<int onum>
+inline bool DetType(const nnvm::NodeAttrs& attrs,
+                    std::vector<int>* in_type,
+                    std::vector<int>* out_type) {
+  using namespace mshadow;
+  CHECK_EQ(in_type->size(), 1U);
+  int dtype = (*in_type)[0];
+  CHECK_NE(dtype, -1) << "Input must have specified type";
```
reminisce: change this to
```
if (dtype == -1) return false;
```

src/operator/tensor/la_op.h
```diff
@@ -413,6 +413,50 @@ inline bool InverseShape(const nnvm::NodeAttrs& attrs,
...
+}
+
+// Type inference function for det functions in linalg
+template<int onum>
+inline bool DetType(const nnvm::NodeAttrs& attrs,
+                    std::vector<int>* in_type,
+                    std::vector<int>* out_type) {
+  using namespace mshadow;
+  CHECK_EQ(in_type->size(), 1U);
+  int dtype = (*in_type)[0];
```
reminisce: const int dtype

src/operator/tensor/la_op.h
```diff
@@ -413,6 +413,50 @@ inline bool InverseShape(const nnvm::NodeAttrs& attrs,
...
+  CHECK_NE(dtype, -1) << "Input must have specified type";
+
```
reminisce: If only fp32/64 are supported, please add one line here for checking whether dtype is equal to kFloat32/kFloat64.

arcadiaphy: All updated.

arcadiaphy:
Regarding this point:

> The grad of determinant is derived from Jacobi's formula, which has a pretty friendly closed form solution for numerical computing when input matrix A is invertible. The non-invertible case is not easy to implement since it involves adjugate matrix. In tensorflow, this case is ignored; while pytorch uses SVD to compute the grad. In this PR, it's left for future work, and now as a temporary method, no grad is passed backwards when det = 0. My inclination is to re-use LU instead of SVD for non-invertible case since it's already calculated.

I've looked into pytorch code more carefully, I think their implementation is wrong. A simple example to show this:

```
In [1]: import torch

In [2]: x = torch.autograd.Variable(torch.tensor([[1., 2.], [2., 4.]]), requires_grad=True)

In [3]: y = x.det()

In [4]: y.backward(torch.ones_like(y))

In [5]: x.grad
Out[5]:
tensor([[0., 0.],
        [0., 0.]])
```

Since in actual computing it's very hard to hit upon det == 0 using float, is it really necessary to implement non-invertible case?

Also I haven't thought of a good method now, any suggestions are welcomed.

asmushetzel:
@mseeger

src/operator/linalg.h
```diff
@@ -195,50 +195,67 @@ int linalg_syevd_workspace_query(const Tensor<xpu, 2, DType>& A,
...
-                          Stream<xpu> *s = 0);
+                          const mxnet::OpContext& ctx);
+
+//////////////////////////////// DET ////////////////////////////////////////////
+
+// CPU/GPU-versions of helper functions to compute matrix determinant
+// Compute matrix inversion with LU and pivot using temp workspace,
+// the result stores back to LU
+template<typename xpu, typename DType>
+void linalg_batch_det_helper(const Tensor<xpu, 3, DType>& LU,
```
asmushetzel: Misleading comment and naming. It looks to me that this helper is not needed to compute the determinant, it is just used for the backward pass and it assumes that the determinant is actually computed and available.

src/operator/tensor/la_op.cc
```diff
@@ -939,5 +939,153 @@ NNVM_REGISTER_OP(_backward_linalg_inverse)
 .set_attr<nnvm::TIsBackward>("TIsBackward", true)
 .set_attr<FCompute>("FCompute<cpu>", LaOpBackward<cpu, 2, 2, 2, 1, inverse_backward>);
 
+NNVM_REGISTER_OP(_linalg_det)
```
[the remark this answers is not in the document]
arcadiaphy: Good point. If we only want to keep only one method, then the `slogdet` is the best choice. 

Since the community is now working on providing numpy experience in mxnet, how about we follow the numpy.linalg package, and keep `det` and `slogdet`.

src/operator/tensor/la_op-inl.h
```diff
@@ -458,14 +458,90 @@ struct inverse {
...
-    Tensor<xpu, 1, DType> work = ctx.requested[0]
-      .get_space_typed<xpu, 1, DType>(Shape1(lwork), s);
     // Since inverse(A) = trans(inverse(trans(A))), so we don't need to transpose
     // A even if we are using the col-major version of getrf and getri routines.
-    linalg_batch_inverse(A, B, work, s);
+    linalg_batch_inverse(A, B, ctx);
+  }
+};
+
+// partial pivoting LU decomposition: A = PLU, so det(A) = det(P)det(L)det(U)
```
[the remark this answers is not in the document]
arcadiaphy: I want to explain the computation method in det, I'll change the comment to make it more clearly.

src/operator/tensor/la_op.h
```diff
@@ -406,10 +406,72 @@ inline bool InverseShape(const nnvm::NodeAttrs& attrs,
...
+  if (!ndim_is_known(in)) return false;
   const int ndim(in.ndim());
   CHECK_GE(ndim, 2) << "Input A's dimension must be >= 2";
   CHECK_EQ(in[ndim-2], in[ndim-1]) << "Input A's last two dimension must be equal";
   SHAPE_ASSIGN_CHECK(*out_attrs, 0, in);
+  return shape_is_known(in);
+}
+
+// Type inference function for linalg_inverse
+inline bool InverseType(const nnvm::NodeAttrs& attrs,
```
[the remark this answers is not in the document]
arcadiaphy: It's separate PRs, I'll remove the changes not related to this PR.

src/operator/tensor/la_op.cc
```diff
@@ -939,5 +939,153 @@ NNVM_REGISTER_OP(_backward_linalg_inverse)
...
+   logabsdet = [1.609438, -inf, 1.609438]
+)code" ADD_FILELINE)
+.set_num_inputs(1)
+.set_num_outputs(4)
+.set_attr<nnvm::FListInputNames>("FListInputNames", [](const NodeAttrs& attrs)
+  { return std::vector<std::string>{"A"}; })
+.set_attr<nnvm::FNumVisibleOutputs>("FNumVisibleOutputs", [](const NodeAttrs& attrs) {
+  return 2; })
+.set_attr<mxnet::FInferShape>("FInferShape", DetShape<2>)
+.set_attr<nnvm::FInferType>("FInferType", DetType<2>)
```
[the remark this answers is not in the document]
arcadiaphy: If canonical type inference is used, then pivot tensor has different sizes for float and double. I need to calculate how much DType space is enough to hold pivot integers like what's done in workspace estimate. I think just use int is more clear and simple here.

src/operator/tensor/la_op.cc
```diff
@@ -920,7 +920,7 @@ Examples::
 .set_attr<nnvm::FListInputNames>("FListInputNames", [](const NodeAttrs& attrs)
   { return std::vector<std::string>{"A"}; } )
 .set_attr<mxnet::FInferShape>("FInferShape", InverseShape)
-.set_attr<nnvm::FInferType>("FInferType", ElemwiseType<1, 1>)
+.set_attr<nnvm::FInferType>("FInferType", InverseType)
```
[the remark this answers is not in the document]
arcadiaphy: I want to add type checks in layer setup. Now many size or type checks in linalg are carried out in forward process, which is not the common practice in mxnet.

But this change is not related to this PR, I'll remove it temporarily.

src/operator/tensor/la_op-inl.h
```diff
@@ -458,14 +458,90 @@ struct inverse {
...
+    Stream<xpu> *s = ctx.get_stream<xpu>();
+    Tensor<xpu, 1, DType> sign = ctx.requested[0]
+      .get_space_typed<xpu, 1, DType>(logdet.shape_, s);
+    Copy(LU, A, s);
+    linalg_batch_getrf(LU, pivot, false, s);
+    using namespace mxnet_op;
+    using namespace mshadow::expr;
+    Kernel<SignedLogDet, xpu>::Launch(s, pivot.size(0), pivot.size(1), pivot.dptr_,
+                                      LU.dptr_, sign.dptr_, logdet.dptr_);
+    const_cast<Tensor<xpu, 1, DType>&>(logdet) = F<mshadow_op::log>(sign) + logdet;
```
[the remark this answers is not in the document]
arcadiaphy: I'll remove `logdet`.

src/operator/tensor/la_op-inl.h
```diff
@@ -825,6 +901,100 @@ struct inverse_backward {
   }
 };
 
+// Here we set grad to zero if det = 0 as a temporary method
```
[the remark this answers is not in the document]
arcadiaphy: I'm not very sure how to handle det = 0 case. If it's ok to use zero here, I'll change the comment and add a note in operator docs.

src/operator/tensor/la_op.cc
```diff
@@ -939,5 +939,153 @@ NNVM_REGISTER_OP(_backward_linalg_inverse)
...
+  *out* = *log(det(A))*
+
+If *n>2*, *logdet* is performed separately on the trailing two dimensions
+for all inputs (batch mode).
+
+.. note:: The operator supports float32 and float64 data types only.
+
+Examples::
+
+   Single matrix inversion
```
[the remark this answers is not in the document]
arcadiaphy: Updated.

src/operator/linalg.h
arcadiaphy: Updated.

arcadiaphy:
Updates:

1. Only operators `det` and `slogdet` are kept in this PR.
2. Zero gradient when det = 0 is intended behavior now and explained in operator docs.

src/operator/tensor/la_op.cc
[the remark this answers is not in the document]
mseeger: Hello,

I developed the linalg operators with @asmushetzel.

I feel it is not very elegant to introduce a lot of MXNet operators, which are essentially just done by sticking existing operators together. It would be a lot cleaner just to provide Python functions for this (using F in {mx.nd, mx.sym} as first arg).

Note we ourselves made this mistake with sumlogdiag, which is ugly and should not be there, really (we could be excused, since diag wasn't there back then).

For example, if you really want logdet, you get it by a LQ decomp (linalg.geqlf), followed by log.sum over the diagonal of L, we have ops for all of this. In fact, you probably want log(abs(det)), because the determinant could be negative. You could return the sign as well.

I don't understand also why a det(.) op is needed, given there is logdet(.) with sign. You can get one from the other. Also, det(.) for large matrices is prone to over or underflow anyway.

It is also somewhat dangerous to offer such operators, because they end up recomputing the underlying factorizations every time. For example, to evaluate the log likelihood of a Gaussian, you need logdet and backsubstitution. You compute the Cholesky decomp. once, then use it twice. With your logdet, I cannot do that. It computes something inside, but does not return it.

Finally, I also find it dangerous to offer inverse. The matrix inverse is almost never needed, but people who lack numerical maths knowledge use it. They should not, it leads to bad code. They should use matrix factorizations, like Cholesky or LQ (i.e. QR), or SVD.

So, if you really want to do something useful, think about a set of Python functions for derived operators. An example:

def linalg_logdet_from_chol(F, lmat):
   return F.sum(F.log(F.abs(F.diag(lmat))))

# If you really want:
def linalg_logdet(F, amat):
   return linalg_logdet_from_chol(F, F.potrf(amat))

mseeger:
Hello,

I developed the linalg operators with @asmushetzel.

I feel it is not very elegant to introduce a lot of MXNet operators, which are essentially just done by sticking existing operators together. It would be a lot cleaner just to provide Python functions for this (using F in {mx.nd, mx.sym} as first arg).

Note we ourselves made this mistake with sumlogdiag, which is ugly and should not be there, really (we could be excused, since diag wasn't there back then).

For example, if you really want logdet, you get it by a LQ decomp (linalg.geqlf), followed by log.sum over the diagonal of L, we have ops for all of this. In fact, you probably want log(abs(det)), because the determinant could be negative. You could return the sign as well.

I don't understand also why a det(.) op is needed, given there is logdet(.) with sign. You can get one from the other. Also, det(.) for large matrices is prone to over or underflow anyway.

It is also somewhat dangerous to offer such operators, because they end up recomputing the underlying factorizations every time. For example, to evaluate the log likelihood of a Gaussian, you need logdet and backsubstitution. You compute the Cholesky decomp. once, then use it twice. With your logdet, I cannot do that. It computes something inside, but does not return it.

Finally, I also find it dangerous to offer inverse. The matrix inverse is almost never needed, but people who lack numerical maths knowledge use it. They should not, it leads to bad code. They should use matrix factorizations, like Cholesky or LQ (i.e. QR), or SVD.

So, if you really want to do something useful, think about a set of Python functions for derived operators. An example:

def linalg_logdet_from_chol(F, lmat):
   return F.sum(F.log(F.abs(F.diag(lmat))))

# If you really want:
def linalg_logdet(F, amat):
   return linalg_logdet_from_chol(F, F.potrf(amat))

arcadiaphy:
@mseeger I think many ordinary users are not experts in linear algebra, so providing a set of basic linear blas/lapack routines and some advanced operators at the same time is the best way for all levels of users. My reason to improve linalg package comes from my daily usage and some users' feature requests. I agree that using basic operators is more efficient, but sometimes easy usage is more important and it doesn't matter if the computation is not fast enough as long as I've quickly finished the implementation.

Also for using python to create some derived operators, actually it's what I want to do in the first place but is restricted by the weak supports of mxnet on registering backward function. Using fine-grained operators to mimic high level operation is all right for forward pass, but the backward pass will be terrible because the combined simple gradient computation is split into backward passes of basic operators. Not the only way to overwrite backward function in mxnet is Custom operator, which is not elegant. In pytorch, it's more user-and-developer-friendly to do this, and the following is the backward of matrix inverse:
```
- name: inverse(Tensor self)
  self: -at::matmul(result.transpose(-2, -1), at::matmul(grad, result.transpose(-2, -1)))
```

I think a good linalg package is like scipy.linalg, in which all routines of blas and lapack are exposed, and there are also a lot of high level functions.

piyushghai:
@arcadiaphy What's the path forward on this PR ? :)

arcadiaphy:
@piyushghai Two paths forward for linalg:

1. Use low level operators to implement determinant in python, but it's not very easy to override backward function.

2. Providing low level and high level operators in c++ at the same time.

This PR is for the 2nd option. If it's merged, the third part of linalg improvement  is to add `svd` and `pinverse`.

asmushetzel:
Tend to agree with @arcadiaphy concerning adding these operators though conceptually you can do the same more efficient by explicitly handling matrix factorization etc on your own. Providing a more seemless experience when porting models from one framework to the other is important for adoption. No objections from my side concerning merging this PR.

arcadiaphy:
@szha How about we merge this PR and move forward?

karan6181:
@szha Is this PR good to go for merge? Thanks!

piyushghai:
@szha Gentle ping for review.

reminisce:
We would like to merge this PR so that we can push forward adding more ops to match `np.linalg`. If no objection, I will merge this PR tomorrow. Thanks for all the comments and contribution.

Issue #14421, opened 2019-03-13
Updating mxnet from 1.0.0, networks give different outputs


jmerkow:
I am working in a production environment, where have some networks implemented in mxnet 1.0.0. I am working updating our systems and trying to push to the latest mxnet (1.4.x as of now) but when we upgrade our networks produce different outputs.
We are using symbols saved to json files and arg/aux_params stored in .params files. these were all produced by mxnet 1.0.0 or earlier.

When using the latest mxnet (or 1.4.x) we are getting different outputs for the same inputs, with our saved models. I have been trying to use git bisect or slowly upgrading versions to figure out where this breaking change occurred but there are issues with your git history and/or some strange (compiler??) incompatibilities which prevent getting a clean checkout/build for nearly all of the intermediate versions…

These are VERY different outputs, im rounding to about the 5 decimal point, or higher, so these aren’t numerical differences.
And we are using modules, not gluons (part of the point is upgrade and move towards gluons)

Does anyone have any idea what could be causing this? Or what to look into to solve it?

andrewfayres:
Do you have a sample model which can be shared to reproduce this? There is a backwards compatibility checker for models as part of the nightly pipeline. It trains models on earlier releases and checks for consistency on the latest builds. It's possible there is an edge case which is being missed.

@mxnet-label-bot add [Pending Requestor Info, Bug]

jmerkow:
Could I run the tool and pass the output instead?

chinakook:
How do you use your model to inference? If you are using mx.module, please make sure the parameter for_training is set to be 0.

piyushghai:
@jmerkow The tool does save and load the same weights and checks the forward pass on the model. 
It does not check in the intermediate layers in the model graph.

jmerkow:
Ok I am working on this again.

I've checked that the parameter files are the same. i.e. something like this:
```
## in a docker with mx1.4.0 loaded
np_arg_params = {key: value.asnumpy() for key, value in arg_params.items()}
np.save('mx140.npy', np_arg_params)
...

## in another docker with mx100 loaded
import numpy as np
np_arg_params2 = np.load('../140/mx140.npy')[()]
assert set(np_arg_params2.keys()).difference(np_arg_params.keys()) == set()
assert set(np_arg_params.keys()).difference(np_arg_params2.keys()) == set()
assert all((np_arg_params[k]-np_arg_params2[k]).sum() ==0 for k in np_arg_params.keys())
```
[...]

PR #9730, opened 2018-02-07
Check padding size for global pooling

TaoLv:
## Description ##
Check padding size for global pooling. #9714 

### Essentials ###
- [x] Passed code style checking (`make lint`)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- Check padding size for pooling, cudnn pooling and pooling_v1
- add python test case for global pooling with user padding size and stride size

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #15026, opened 2019-05-21
[MXNET-14421] Make global pooling backwards compatible

jmerkow:
## Description ##
There was a backward compatibility breaking change (#9730) to pooling where pad is explicitly overwritten causing global pooling to produce different results. This causes issues in any networks trained before this was changed as any layers after a global pooling layer will not get different input and therefore provide incorrect results.
To be clear, it is mathematically correct to have padding == 0.  The PR will not effect anytime the layer is used without setting pad (the default is zero), it will only effect times when global_pool=True and pad is non-zero. It also adds a warning that adding padding with global_pool=True leads to incorrect results and should not be done.  _This PR is to provide backward compatibility only._

Issue: #14421

### Essentials ###
- [x] The PR title starts with [MXNET-14421], where #14421 refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)

### Changes ###
- [x] Re-allow padding values to be passed when global_pool=true in pooling layers instead of forcing them to zero.
- [x] Added a warning that this isn't good to do.

## Comments ##
This is to allow backward compatibility only.  It should not affect anyone unless they have explicitly set global_pool=True and padding to non-zero.  This is for any networks trained before the change was made to work, and should not adversely affect any users.

### DIFF
--- a/src/operator/nn/pooling-inl.h
+++ b/src/operator/nn/pooling-inl.h
@@ -178,6 +178,11 @@ template<typename xpu, typename DType>
 class PoolingOp {
  public:
   void Init(PoolingParam p) {
+    if (p.global_pool && (p.pad[0] ||  p.pad[1]) ) {
+        LOG(WARNING) << "Using global pooling with padding leads to incorrect results"
+                     << "and is only supported for backwards compatibility.\n"
+                     << "If you are training a new network it is recommended that you set padding to 0.";
+    }
     this->param_ = p;
   }
 
--- a/src/operator/nn/pooling-inl.h
+++ b/src/operator/nn/pooling-inl.h
@@ -178,11 +178,17 @@ template<typename xpu, typename DType>
 class PoolingOp {
  public:
   void Init(PoolingParam p) {
-    if (p.global_pool && (p.pad[0] ||  p.pad[1])) {
-        LOG(WARNING) << "Using global pooling with padding leads to incorrect results "
-                     << "and is only supported for backwards compatibility.\n"
-                     << "If you are training a new network it is recommended "
-                     << "that you set padding to 0.";
+    if (p.global_pool) {
+        has_padding = 0;
+        for (index_t i = 0; i < p.pad.ndim(); i++) {
+            has_padding += p.pad[i];
+        }
+        if (has_padding) {
+            LOG(WARNING) << "Using global pooling with padding leads to incorrect results "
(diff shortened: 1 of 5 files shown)


src/operator/nn/pooling-inl.h
```diff
@@ -178,6 +178,12 @@ template<typename xpu, typename DType>
 class PoolingOp {
  public:
   void Init(PoolingParam p) {
+    if (p.global_pool && (p.pad[0] ||  p.pad[1])) {
```
jmerkow: I don't think this is the best place for this.  Can someone suggest a better place? @piiswrong

karan6181:
@mxnet-label-bot add [Operator, C++, pr-awaiting-review]

leleamol:
The C++ label is meant for API defined in cpp-package.  Removing the label as the issue is not related to those API.
@mxnet-label-bot  remove [C++]

piyushghai:
@TaoLv Can you take a look at this PR from @jmerkow ?

jmerkow:
Im not sure this should be merged, but it could provide assistance to those who have trained with the previous behavior and need a work around.

larroy:
@jmerkow could you look into the CI failures?

sandeep-krishnamurthy:
Thank you @jmerkow for all the effort in debugging the root cause and getting this PR. This PR and the issue is a great reference for people facing similar issue. As you rightly mentioned, I think we cannot merge this PR because that would be like backward compatibility of a bug. We should probably have this workaround suggested to users instead and encourage to upgrade to latest version of MXNet.

jmerkow:
I agree probably not worth merging it’s an edge case, but hopefully if someone comes across this issue they’ll find this and see the PR as a work around. We’ve started the process of retraining or deprecating networks effected by this bug (a long process...) so well eventually be able to upgrade.

sandeep-krishnamurthy:
> I agree probably not worth merging it’s an edge case, but hopefully if someone comes across this issue they’ll find this and see the PR as a work around. We’ve started the process of retraining or deprecating networks effected by this bug (a long process...) so well eventually be able to upgrade.

Thank you. Closing the PR and the issue.

Sorry, it is unfortunate you people have to retrain a lot of models due to this bug. Keep us updated and cut us if you find any other issues.

Issue #15029, opened 2019-05-22
mx.nd.random.shuffle crashes in the master branch


zheng-da:
After compiling the latest master branch (commit 5854b98) of MXNet, it seems it can't run the following simple code:
```python
import mxnet as mx
arr = mx.nd.arange(0, 153431, dtype='int64')
arr.wait_to_read()
arr = mx.nd.random.shuffle(arr)
arr.wait_to_read()
```
The code will result in segfault.

If MXNet is compiled with USE_INT64_TENSOR_SIZE = 1, the code above works fine.

The simple code should work with int32 tensor size.

apeforest:
The coredump does not happen on MacOS with 16GB memory. It occurrs on ubuntu 16.04

apeforest:
Here is the stack trace from GDB
```
#52347 0x00007fff9fcb71e9 in std::uniform_int_distribution<int>::operator()<std::mersenne_twister_engine<unsigned long, 32ul, 624ul, 397ul, 31ul, 2567483615ul, 11ul, 4294967295ul, 7ul, 2636928640ul, 15ul, 4022730752ul, 18ul, 1812433253ul> > (this=0x7fff2a5f4c80, __urng=...)
    at /usr/include/c++/5/bits/uniform_int_dist.h:165
#52348 0x00007fff9fcae86b in mxnet::op::(anonymous namespace)::<lambda(mxnet::index_t)>::operator()(mxnet::index_t) const (__closure=0x7fff2a5f61e0,
    n=-1) at src/operator/random/shuffle_op.cc:50
#52349 0x00007fff9fcb66e3 in __gnu_parallel::__parallel_random_shuffle_drs<long*, void mxnet::op::(anonymous namespace)::Shuffle1D<long, std::mersenne_twister_engine<unsigned long, 32ul, 624ul, 397ul, 31ul, 2567483615ul, 11ul, 4294967295ul, 7ul, 2636928640ul, 15ul, 4022730752ul, 18ul, 1812433253ul> >(long*, int, std::mersenne_twister_engine<unsigned long, 32ul, 624ul, 397ul, 31ul, 2567483615ul, 11ul, 4294967295ul, 7ul, 2636928640ul, 15ul, 4022730752ul, 18ul, 1812433253ul>*)::{lambda(int)#1}> () at /usr/include/c++/5/parallel/random_shuffle.h:384
[...]

PR #15041, opened 2019-05-22
Fix crash in random.shuffle operator

apeforest:
## Description ##
This PR fix https://github.com/apache/incubator-mxnet/issues/15029
The rootcause of the problem is when NDArray is 1-d and the platform is GNU Linux, the backend implementation uses `__gnu_parallel:random_shuffle()` See: https://github.com/apache/incubator-mxnet/blob/master/src/operator/random/shuffle_op.cc#L53. This explains why crash did not happen in MacOS.

The random_shuffle template defined in gcc (https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/parallel/random_shuffle.h#L384) is passing in `std::numeric_limits<uint32_t>::max()` to the __rng() function to generate random seed. Therefore, in our rng() function (https://github.com/apache/incubator-mxnet/blob/master/src/operator/random/shuffle_op.cc#L49) we cannot use our self defined data type `index_t` and have to use `uint32_t` to avoid integer overflow.

Why this bug was not detected in our unit test? Because our unit test of shuffle only tests 2-D shape, which did not use parallel_shuffle in backend.

### Essentials ###


zheng-da:
I think you should fix ShuffleND as well.

apeforest:
@zheng-da ShuffleND does not use  `__gnu_parallel:random_shuffle()` so there is no issue.

zheng-da:
thanks for clarifying.

karan6181:
@mxnet-label-bot add [Operator, pr-awaiting-merge]

apeforest:
@larroy yes, it's the internal implementation of random_shuffle that caused overflow. See the line: (https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/parallel/random_shuffle.h#L384

I already addeded this in the PR description

larroy:
Thanks, this was not entirely clear, was a range error?

PR #15044, opened 2019-05-22
[WIP] Added fused slice sum operator

nvchai:
## Description ##
This PR adds a fused slice sum operator that fuses the slice operation with a sum operation. The code is a work in progress as it only implements the forward op for now and includes a dummy backward op implementation. When the slice_sum operator is used for half precision tensors, it is required that the last dimension be a multiple of 2 in order to use the vectorized half type. 

### Essentials ###

### DIFF
--- a/src/operator/tensor/slice_sum-inl.h
+++ b/src/operator/tensor/slice_sum-inl.h
@@ -1,0 +1,203 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef MXNET_OPERATOR_TENSOR_SLICE_SUM_INL_H_
+#define MXNET_OPERATOR_TENSOR_SLICE_SUM_INL_H_
+
+#include <vector>
+#include <algorithm>
+#include <utility>
+#include "../mxnet_op.h"
+#include "./broadcast_reduce_op.h"
+#include "./init_op.h"
+#include "../../common/static_array.h"
+
+namespace mxnet {
+namespace op {
+
+struct SliceSumParam : public dmlc::Parameter<SliceSumParam> {
+  int begin, end, axis;
+  DMLC_DECLARE_PARAMETER(SliceSumParam) {
+    DMLC_DECLARE_FIELD(begin)
+    .describe("starting indices for the slice operation, supports negative indices.");
+    DMLC_DECLARE_FIELD(end)
+    .describe("ending indices for the slice operation, supports negative indices.");
+    DMLC_DECLARE_FIELD(axis)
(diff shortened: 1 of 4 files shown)


src/operator/tensor/slice_sum-inl.h
```diff
@@ -0,0 +1,203 @@
...
+  }
+};
+
+inline bool SliceSumOpShape(const nnvm::NodeAttrs& attrs,
+                         mxnet::ShapeVector* in_attrs,
+                         mxnet::ShapeVector* out_attrs) {
+  CHECK_EQ(in_attrs->size(), 2U);
+  CHECK_EQ(out_attrs->size(), 1U);
+  const mxnet::TShape& dshape = (*in_attrs)[0];
+  if (dshape.ndim() == 0) return false;
```
ptrendx: shape_is_none?

src/operator/tensor/slice_sum-inl.h
```diff
@@ -0,0 +1,203 @@
...
+  mxnet::TShape oshape = dshape;
+
+  SHAPE_ASSIGN_CHECK(*out_attrs, 0, oshape);
+  return !shape_is_none(dshape) && !shape_is_none(oshape);
```
ptrendx: Shape inference function should return true only if all of the shapes are known. You have 2 inputs but check shape of only 1 of them here.

src/operator/tensor/slice_sum-inl.h
```diff
@@ -0,0 +1,203 @@
...
+          Kernel<SliceSumKernel<ndim, req_type>, xpu>::Launch(s, num_elements,
+              out_data.dptr<DType>(), in1_data.dptr<DType>(), in2_data.dptr<DType>(), 
+              in1_strides, in2_strides, in2_offsets);
+        })
+      })
+  })
+}
+
+template<typename xpu>
+void SliceSumOpForwardEx(const nnvm::NodeAttrs& attrs,
```
ptrendx: nit: ending a name with Ex suggests that this function supports sparse inputs, which I don't think it does. Could you change it to just SliceSymOpForward?

src/operator/tensor/slice_sum.cc
```diff
@@ -0,0 +1,59 @@
...
+ * \brief Fused implementation of slice and sum operation
+ */
+#include "./slice_sum-inl.h"
+
+namespace mxnet {
+namespace op {
+
+DMLC_REGISTER_PARAMETER(SliceSumParam);
+
+NNVM_REGISTER_OP(_slice_sum)
```
ptrendx: Do you intend to introduce a graph pass to automatically use this op when applicable? Otherwise starting its name with "_" makes it hidden from users.

src/operator/tensor/slice_sum-inl.h
nvchai: Fixed in new commit.

src/operator/tensor/slice_sum-inl.h
nvchai: Fixed in new commit.

src/operator/tensor/slice_sum.cc
nvchai: Not planned for now. I changed the op name to slice_sum.

karan6181:
@mxnet-label-bot add [Operator, C++, pr-work-in-progress]

nvchai:
Yes, I agree the op is too specific right now. There are two approaches to solve this problem. @ptrendx  and I are investigating a more generic operator fusion solution that will make this PR reduntant. Alternatively, if the fusion approach does not work out, I can try to make this op more generic and maybe add a graph optimization pass too.

pengzhao-intel:
> Yes, I agree the op is too specific right now. There are two approaches to solve this problem. @ptrendx and I are investigating a more generic operator fusion solution that will make this PR reduntant. Alternatively, if the fusion approach does not work out, I can try to make this op more generic and maybe add a graph optimization pass too.

I think a graph pass should be fine @apeforest

roywei:
@nvchai Hi, could you rebase this PR? thanks!

ptrendx:
This PR is no longer needed due to pointwise fusion.

PR #15048, opened 2019-05-22
[MXNET-1408] Adding test to verify Large Tensor Support for ravel and unravel

access2rohit:
## Description ##
New Ops: ravel_multi_index, unravel_index

### Essentials ###
- [x] The PR title starts with [MXNET-1408], where #15048 1408 refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)

### Changes ###

## Comments ##

### DIFF
--- a/tests/nightly/test_large_array.py
+++ b/tests/nightly/test_large_array.py
@@ -15,9 +15,9 @@
 # specific language governing permissions and limitations
 # under the License.
 
-import mxnet as mx
 import numpy as np
-from mxnet.test_utils import rand_ndarray, assert_almost_equal
+import mxnet as mx
+from mxnet.test_utils import rand_ndarray, assert_almost_equal, rand_coord_2d
 from mxnet import gluon, nd
 from tests.python.unittest.common import with_seed
 
@@ -74,7 +74,7 @@ def test_ndarray_random_randint():
     # check if randint can generate value greater than 2**32 (large)
     low_large_value = 2**32
     high_large_value = 2**34
-    a = nd.random.randint(low_large_value,high_large_value, dtype=np.int64)
+    a = nd.random.randint(low_large_value, high_large_value, dtype=np.int64)
     low = mx.nd.array([low_large_value], dtype='int64')
     high = mx.nd.array([high_large_value], dtype='int64')
     assert a.__gt__(low) and a.__lt__(high)
@@ -128,13 +128,6 @@ def test_clip():
     a = nd.arange(0, LARGE_X * SMALL_Y).reshape(LARGE_X, SMALL_Y)
     res = nd.clip(a, a_min=100, a_max=1000)
     assert np.sum(res[-1].asnumpy() == 1000) == a.shape[1]
-
-
-def test_take():
-    a = nd.ones(shape=(LARGE_X, SMALL_Y))
-    idx = nd.arange(LARGE_X-1000, LARGE_X)
-    res = nd.take(a, idx)
-    assert np.sum(res[-1].asnumpy() == 1) == res.shape[1]
 
 
 def test_split():
@@ -215,9 +208,9 @@ def test_where():
 
 
 def test_pick():
-    a = mx.nd.ones(shape=(256*35, 1024*1024))
-    b = mx.nd.ones(shape=(256*35,))
-    res = mx.nd.pick(a,b)
+    a = mx.nd.ones(shape=(256 * 35, 1024 * 1024))
+    b = mx.nd.ones(shape=(256 * 35, ))
+    res = mx.nd.pick(a, b)
     assert res.shape == b.shape
 
 
@@ -252,11 +245,9 @@ def test_spacetodepth():
     output = mx.nd.space_to_depth(data, 2)
     assert_almost_equal(output.asnumpy(), expected, atol=1e-3, rtol=1e-3)
 
-
+@with_seed()
 def test_diag():
-    h = np.random.randint(2,9)
-    w = np.random.randint(2,9)
-    a_np = np.random.random((LARGE_X, 64)).astype(np.float32)
+    a_np = np.random.random((LARGE_X, SMALL_Y)).astype(np.float32)
     a = mx.nd.array(a_np)
 
     # k == 0
@@ -274,9 +265,31 @@ def test_diag():
     assert_almost_equal(r.asnumpy(), np.diag(a_np, k=k))
 
     # random k
-    k = np.random.randint(-min(LARGE_X, 64) + 1, min(h, w))
+    k = np.random.randint(-min(LARGE_X, SMALL_Y) + 1, min(LARGE_X, SMALL_Y))
     r = mx.nd.diag(a, k=k)
     assert_almost_equal(r.asnumpy(), np.diag(a_np, k=k))
+
+
+@with_seed()
+def test_ravel_multi_index():
+    x1, y1 = rand_coord_2d((LARGE_X - 100), LARGE_X, 10, SMALL_Y)
+    x2, y2 = rand_coord_2d((LARGE_X - 200), LARGE_X, 9, SMALL_Y)
+    x3, y3 = rand_coord_2d((LARGE_X - 300), LARGE_X, 8, SMALL_Y)
+    indices_2d = [[x1, x2, x3], [y1, y2, y3]]
+    idx = mx.nd.ravel_multi_index(mx.nd.array(indices_2d, dtype=np.int64), shape=(LARGE_X, SMALL_Y))
+    idx_numpy = np.ravel_multi_index(indices_2d, (LARGE_X, SMALL_Y))
+    assert np.sum(1 for i in range(idx.size) if idx[i] == idx_numpy[i]) == 3
+
+
+@with_seed()
+def test_unravel_index():
+    x1, y1 = rand_coord_2d((LARGE_X - 100), LARGE_X, 10, SMALL_Y)
+    x2, y2 = rand_coord_2d((LARGE_X - 200), LARGE_X, 9, SMALL_Y)
+    x3, y3 = rand_coord_2d((LARGE_X - 300), LARGE_X, 8, SMALL_Y)
+    original_2d_indices = [[x1, x2, x3], [y1, y2, y3]]
+    idx_numpy = np.ravel_multi_index(original_2d_indices, (LARGE_X, SMALL_Y))
+    indices_2d = mx.nd.unravel_index(mx.nd.array(idx_numpy, dtype=np.int64), shape=(LARGE_X, SMALL_Y))
+    assert (indices_2d.asnumpy() == np.array(original_2d_indices)).all()
 
 
 if __name__ == '__main__':
--- a/python/mxnet/test_utils.py
+++ b/python/mxnet/test_utils.py
@@ -419,6 +419,12 @@ def rand_shape_nd(num_dim, dim=10):
     return tuple(rnd.randint(1, dim+1, size=num_dim))
 
 
+def rand_coord_2d(x_low, x_high, y_low, y_high):
+    x = np.random.randint(x_low, x_high, dtype=np.int64)
+    y = np.random.randint(y_low, y_high, dtype=np.int64)
+    return x, y
+
+
 def np_reduce(dat, axis, keepdims, numpy_reduce_func):
     """Compatible reduce for old version of NumPy.
 


access2rohit:

access2rohit:
@apeforest : Please review

tests/nightly/test_large_array.py
```diff
@@ -279,6 +279,19 @@ def test_diag():
     assert_almost_equal(r.asnumpy(), np.diag(a_np, k=k))
 
 
+def test_ravel_and_unravel():
```
apeforest: break it into two tests

tests/nightly/test_large_array.py
```diff
@@ -279,6 +279,19 @@ def test_diag():
...
+    idxs_2d = [[LARGE_X-1,LARGE_X-100,6],[SMALL_Y-1,SMALL_Y-10,1]]
+    idx = mx.nd.array(1)
+    def test_ravel_multi_index():
+        idx = mx.nd.ravel_multi_index(mx.nd.array(idxs_2d, dtype=np.int64), shape=(LARGE_X,SMALL_Y))
+        idx_numpy = np.ravel_multi_index(idxs_2d, (LARGE_X,SMALL_Y))
+        assert np.sum(1 for i in range(idx.size) if idx[i] == idx_numpy[i]) == 3
+
+    def test_unravel_index():
+        original_2d_idxs = mx.nd.unravel_index(mx.nd.array(idx, dtype=np.int64), shape=(LARGE_X,SMALL_Y))
+        assert (original_2d_idxs.asnumpy() == np.array(idxs_2d)).all() == True
```
apeforest: remove `== True`

tests/nightly/test_large_array.py
access2rohit: Its more readable and understandable this way. Also I have split it into 2 functions inside.

tests/nightly/test_large_array.py
access2rohit: ok

tests/nightly/test_large_array.py
```diff
@@ -279,6 +279,19 @@ def test_diag():
     assert_almost_equal(r.asnumpy(), np.diag(a_np, k=k))
 
 
+def test_ravel_and_unravel():
+    idxs_2d = [[LARGE_X-1,LARGE_X-100,6],[SMALL_Y-1,SMALL_Y-10,1]]
```
marcoabreu: Please use C++ with mocking of the memory allocation instead of actually allocating memory.

Although this is nightly, I don't think it's necessary for these kind of tests to actually allocate real memory.

marcoabreu: @perdasilva

apeforest: We can add C++ unit test to mock the memory allocation. However, I thought the nightly test is  used for end-to-end testing purpose to mimic the real deployment situation, or is it not?

marcoabreu: They are to some extent, but in same cases the cost vs return ratio is not that good. While smoke tests are certainly necessary in some cases, resource intensive ones on the other hand can usually be abstracted away. If we assume that memory access on the system side is properly working and only mock this very laster layer, this result should be pretty much as meaningful as running it end to end.

access2rohit: @marcoabreu : I didn't understand your last comment. What did you mean by "very laster layer" ? All these tests are single operator tests. This tests I combined because its easier to test when combined together

access2rohit: because ravel and unravel were reverse of each other

access2rohit: @marcoabreu ??

marcoabreu: When I'm talking about layers in this case, I'm talking about the physical execution on a machine and not about machine learning operator layers.

Specifically in this case I'm talking about mocking memory accesses. Read and writes would access a virtual array array that doesn't have the full capacity on physical memory but instead consists of procedually generated data.

In various languages there's the concept of Iterators https://jeffknupp.com/blog/2013/04/07/improve-your-python-yield-and-generators-explained/ where a function can be invoked that will deliver a stream of entries. If you would call ```.toList()``` that would then create a physically mapped array (which we wouldn't want). Instead, you could call ```.get(1243952)``` which would iterate the array but not allocate for the full size but only for that particular entry. So support writes, you could have custom lookups that would allow setting certain values in a sparse fashion. The iterator function could be something like
```
float64 getValue(int64 index) 
{
    dictValue = _internalDict.get(index);
    if (dictValue != null)
        return dictValue;
    else
        return index; //Replace with generator function 
}

setValue(int64 index, float64 value) 
{
    _internalDict.set(index, value)
}
```

With this concept you'd be able to test infinitely big arrays as long as you only use sparse data to test (the size of the dict would depend on the number of specifically set entries). Since the purpose of these large array tests is not to test dense data (it wouldn't make a difference between sparse and dense besides performance I assume), this type of generator model would allow us to execute these tests in an efficient way.

The only thing we'd have to mock at this point are the getters and setters that try to read/write from/to physical memory. The calls would be redirected to the pseudo-code above.


I'm aware that we do not support this type of behaviour out of the box, but I think that we're getting to a point where we should take a step back and reconsider the approaches we would like to do in such resource-heavy cases. Of course, we wouldn't be testing the physical memory access with my method, but I think that we can assume that memory accesses are properly working or otherwise we'd notice in quite a few other cases.

marcoabreu: Sorry for my belated response. I have copied my comment to dev@ to bring this topic to a wider audience.

apeforest: Hi @marcoabreu, thanks a lot for your suggestion and it's a very good learning for me too. The large tensor support is crucial to DGL (Deep Graph Library) that is using MXNet and planning to release with support of large tensors. Without this test, we cannot make this feature production ready.

While I appreciate the direction of testing you suggested and definitely would like to explore it further, I am also afraid that not adding tests before we implement the approach you suggested will block the release of DGL and other MXNet projects that depend on large tensors as well.

Can we have a seperate project to address the test efficiency (we may need some expertise from CI team for this project) while moving on with the large tensor project? Please let me know if you have other concerns or advice. Thank you

access2rohit: @marcoabreu : We have to test cases where we have to allocate large tensors on actual physical memory. That is the whole purpose of adding these tests. Without which we cannot support or even guarantee large dense tensor support on MXNet. 

The purpose of these tests are not at all to test only Sparse. Currently we are not addressing performance issues with large dense arrays but trying to support them. From your reply it seems like you are under the impression that everything should work fine if tensor size >2^32 elements(Please correct me if I am wrong). But that's not the case and the operator simply segfaults or gives unpredictable behavior. That is why is necessary to test them with such large arrays/tensors.

access2rohit: @marcoabreu : Let me know if you have other unanswered queries that will help you understand the use case better and unblock this PR.

apeforest: nit: space between operators and operands

tests/nightly/test_large_array.py
```diff
@@ -279,6 +279,19 @@ def test_diag():
...
+    idx = mx.nd.array(1)
+    def test_ravel_multi_index():
+        idx = mx.nd.ravel_multi_index(mx.nd.array(idxs_2d, dtype=np.int64), shape=(LARGE_X,SMALL_Y))
+        idx_numpy = np.ravel_multi_index(idxs_2d, (LARGE_X,SMALL_Y))
```
apeforest: nit: space. btw, do you see any pylint warning out of these?

tests/nightly/test_large_array.py
```diff
@@ -279,6 +279,19 @@ def test_diag():
...
+        assert np.sum(1 for i in range(idx.size) if idx[i] == idx_numpy[i]) == 3
+
+    def test_unravel_index():
+        original_2d_idxs = mx.nd.unravel_index(mx.nd.array(idx, dtype=np.int64), shape=(LARGE_X,SMALL_Y))
```
apeforest: same: add space

tests/nightly/test_large_array.py
apeforest: nit: do you mean `indices_2d`?

access2rohit: yes . Though `make pylint` gives this output.

ubuntu@ip-172-31-82-110 ~/mxnet3 (master) $ make pylint
Makefile:345: WARNING: Significant performance increases can be achieved by installing and enabling gperftools or jemalloc development packages
python3 -m pylint --rcfile=/home/ubuntu/mxnet3/ci/other/pylintrc --ignore-patterns=".*\.so$,.*\.dll$,.*\.dylib$" python/mxnet tools/caffe_converter/*.py
Using config file /home/ubuntu/mxnet3/ci/other/pylintrc

------------------------------------
Your code has been rated at 10.00/10


I will still fix it. Dunno why pylint isn't catching it.

marcoabreu: Hey,

I understand the importance of large tensors and I'm really looking forward to them as well. Don't get me wrong, I'm not underestimating the work that comes with supporting them.

We already had a history of issues that were caused by too large allocations. Thus, for the sake of keeping CI stable (independent on whether we're running on per-PR or nightly) and to adhere to a high test standard, I'd prefer if we would not delay adressing the test strategy.

I think there's a misunderstanding. With my proposed approach, you would still be able to make a full test of a single operator. But instead of randomly generating the input tensor, you would use a generator model that gives you predictable and procedually generatable input tensor instead of a large and randomly generated one.

The idea here would then be to give this procedually generated input tensor to the operator and since it's a predicutable input, you can also have a predictable output that you can verify. On the way, the operator will run as usual, it's just that the source of the data is not physical memory but generated values. 

The only case that wouldn't be supported is the chaining of operators since that would require storing the intermediary values. 

The idea here is that (in whatever fashion you'd like) we generate fake input and output tensors that return predictable and valid data, but doesn't actually consume memory on the RAM. From MXNets perspective, it shouldn't matter where the real values is coming from. For the sake of an operator, it's important that it actually retrieves proper values and that would be the case here.

access2rohit: @marcoabreu  I still don't understand how will I know that on actual large tensor after memory allocation would it work as expected or not. Procedurally generating the values will indeed give me random values on random locations that are indexed at locations > 2^32. But the input should come from RAM. it does matter. It needs to be tested on actual physical memory so there are no surprises. If nighly tests are not feasible do we have a weekly test where we can run this ?

access2rohit: got it .... mxnet makefile doesn't do pylint on tests/nightly and tests/python/unittest ... added manually to local repo. Thanks for the advice !

tests/nightly/test_large_array.py
```diff
@@ -15,8 +15,8 @@
 # specific language governing permissions and limitations
 # under the License.
 
-import mxnet as mx
 import numpy as np
+import mxnet as mx
```
access2rohit: Suggested reordering of imports by pylint

tests/nightly/test_large_array.py
```diff
@@ -130,13 +130,6 @@ def test_clip():
     assert np.sum(res[-1].asnumpy() == 1000) == a.shape[1]
 
 
-def test_take():
```
access2rohit: This was duplicate

tests/nightly/test_large_array.py
```diff
@@ -279,6 +272,22 @@ def test_diag():
     assert_almost_equal(r.asnumpy(), np.diag(a_np, k=k))
 
 
+def test_ravel_multi_index():
+    indices_2d = [[LARGE_X-1, LARGE_X-100, 6], [SMALL_Y-1, SMALL_Y-10, 1]]
+    idx = mx.nd.ravel_multi_index(mx.nd.array(indices_2d, dtype=np.int64), shape=(LARGE_X, SMALL_Y))
+    idx_numpy = np.ravel_multi_index(indices_2d, (LARGE_X, SMALL_Y))
+    assert np.sum(1 for i in range(idx.size) if idx[i] == idx_numpy[i]) == 3
+
+
```
apeforest: remove extra line

access2rohit: done :)

tests/nightly/test_large_array.py
```diff
@@ -254,8 +247,8 @@ def numpy_space_to_depth(x, blocksize):
 
 
 def test_diag():
-    h = np.random.randint(2,9)
-    w = np.random.randint(2,9)
+    h = np.random.randint(2, 9)
```
apeforest: I do not see the necessity of creating h and w variables here.

marcoabreu: If you decide to move forward with them, please add the with seed annotation to the test

access2rohit: @apeforest Thats how the unittest for diagonal operator is implemented. 
@marcoabreu Good point will add `@with_seed()`

tests/nightly/test_large_array.py
```diff
@@ -274,11 +265,26 @@ def test_diag():
...
 
     # random k
-    k = np.random.randint(-min(LARGE_X, 64) + 1, min(h, w))
+    k = np.random.randint(-min(LARGE_X, SMALL_Y) + 1, min(LARGE_X, SMALL_Y))
     r = mx.nd.diag(a, k=k)
     assert_almost_equal(r.asnumpy(), np.diag(a_np, k=k))
 
 
+def test_ravel_multi_index():
+    indices_2d = [[LARGE_X-1, LARGE_X-100, 6], [SMALL_Y-1, SMALL_Y-10, 1]]
```
apeforest: can we generate the random indices instead?

tests/nightly/test_large_array.py
```diff
@@ -274,11 +265,26 @@ def test_diag():
...
+    idx = mx.nd.ravel_multi_index(mx.nd.array(indices_2d, dtype=np.int64), shape=(LARGE_X, SMALL_Y))
+    idx_numpy = np.ravel_multi_index(indices_2d, (LARGE_X, SMALL_Y))
+    assert np.sum(1 for i in range(idx.size) if idx[i] == idx_numpy[i]) == 3
+
+
+def test_unravel_index():
+    original_2d_indices = [[LARGE_X-1, LARGE_X-100, 6], [SMALL_Y-1, SMALL_Y-10, 1]]
```
apeforest: same here. can we generate random indices? you can pass in the 2d_array as input so you only keep one copy

tests/nightly/test_large_array.py
```diff
@@ -216,8 +209,8 @@ def test_where():
 
 def test_pick():
     a = mx.nd.ones(shape=(256*35, 1024*1024))
-    b = mx.nd.ones(shape=(256*35,))
-    res = mx.nd.pick(a,b)
+    b = mx.nd.ones(shape=(256*35, ))
```
apeforest: nit: space around * operator

tests/nightly/test_large_array.py
access2rohit: The input is not 2d_array, we pass indices as input. I can do following:
`
low_large_value = 2**32
high_large_value = 2**34
a = nd.random.randint(low_large_value, high_large_value, dtype=np.int64)
`

tests/nightly/test_large_array.py
```diff
@@ -274,11 +265,40 @@ def test_diag():
     assert_almost_equal(r.asnumpy(), np.diag(a_np, k=k))
 
     # random k
-    k = np.random.randint(-min(LARGE_X, 64) + 1, min(h, w))
+    k = np.random.randint(-min(LARGE_X, SMALL_Y) + 1, min(LARGE_X, SMALL_Y))
     r = mx.nd.diag(a, k=k)
     assert_almost_equal(r.asnumpy(), np.diag(a_np, k=k))
 
 
+def random_2d_coordinates(x_low, x_high, y_low, y_high):
```
apeforest: I just realized there is existing utility that does what you want to achieve here. Please use https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/test_utils.py#L418

apeforest: It seems you are re-use the existing utility method: https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/test_utils.py#L410

access2rohit: it is different and we need one value that to very large for tests case to be successful not any values between (1, very large value)

access2rohit: will move this to test_utils.py

PR #15061, opened 2019-05-24
ignore auto-generated file 'nnvm.cc' under amalgamation

zlargon:
## Description ##

ignore auto-generated file 'nnvm.cc' when under amalgamation

```bash
cd amalgamation
make
```


zlargon:
Duplicated pull request #15060

PR #15064, opened 2019-05-24
[Fit API] Update fit method and LoggingHandler

abhinavs95:
## Description ##
- moved train loop outside `fit()` into a new `train_epoch()` method. This allows tasks with custom train loop to use Estimator by simply overriding the `train_epoch()` method.
- update precision for learning rate in LoggingHandler to accurately display low values
- fix reading optimizer name from trainer object in LoggingHandler

### Essentials ###
- [x] Code is well-documented

@roywei @nswamy

### DIFF
--- a/python/mxnet/gluon/contrib/estimator/estimator.py
+++ b/python/mxnet/gluon/contrib/estimator/estimator.py
@@ -193,6 +193,51 @@ class Estimator(object):
                 self.val_metrics.append(val_metric)
         return self.train_metrics, self.val_metrics
 
+    def train_epoch(self,
+              train_data,
+              estimator_ref,
+              batch_begin,
+              batch_end,
+              batch_axis=0):
+        """Train model on train data for 1 epoch
+
+         Parameters
+         ----------
+         train_data : DataLoader
+             Train data loader with data and labels.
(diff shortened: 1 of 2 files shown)


karan6181:
@mxnet-label-bot add [Gluon, pr-awaiting-review]

piyushghai:
@abhinavs95 Can you look into the review comments ?

vandanavk:
@mxnet-label-bot update [Gluon, pr-awaiting-response]

abhinavs95:
@szha For my use case, I want perform certain operations (like gradient clipping) in between batch-begin and batch-end which cannot be captured by event handlers. Also my model is multi-input, multi-output so I also have modified the forward pass and loss computation. To use estimator for cases like these having the train method out of fit() is necessary so that the user can inject their custom code. 

Also this does not break the basic functionality of estimator as the default train method is used if the user does not specify their own.

szha:
@abhinavs95 `train_epoch`'s granularity is too coarse to ensure that registered handlers work correctly.

abhinavs95:
Closing this for now, will reopen when I get a chance to work on it.

PR #15066, opened 2019-05-24
fix imresize interp docs

### LINKED ISSUE #14993: The discription of function mxnet.image.imresize is wrong.
I find it wrong when I also write my psnr code.

'''
        2: Area-based (resampling using pixel area relation). It may be a
        preferred method for image decimation, as it gives moire-free
        results. But when the image is zoomed, it is similar to the Nearest
        Neighbors method. (used by default).
        3: Bicubic interpolation over 4x4 pixel neighborhood.
'''
should be
'''
        2: Bicubic interpolation over 4x4 pixel neighborhood.
        3: Area-based (resampling using pixel area relation). It may be a
        preferred method for image decimation, as it gives moire-free
        results. But when the image is zoomed, it is similar to the Nearest
        Neighbors method. (used by default).
        
'''
you can see this page
https://docs.opencv.org/master/da/d54/group__imgproc__transform.html
'''
import cv2
print(cv2.INTER_CUBIC)
'''
you will see 
2

aaronmarkham:
## Description ##
Fixes #14993

I just flipped the order of the interp param docs according to what @zhouhang95 noted in #14993.

### DIFF
--- a/python/mxnet/image/image.py
+++ b/python/mxnet/image/image.py
@@ -102,11 +102,11 @@ def imresize(src, w, h, *args, **kwargs):
         Possible values:
         0: Nearest Neighbors Interpolation.
         1: Bilinear interpolation.
-        2: Area-based (resampling using pixel area relation). It may be a
+        2: Bicubic interpolation over 4x4 pixel neighborhood.
+        3: Area-based (resampling using pixel area relation). It may be a
         preferred method for image decimation, as it gives moire-free
         results. But when the image is zoomed, it is similar to the Nearest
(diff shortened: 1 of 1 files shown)


karan6181:
@mxnet-label-bot add [Doc, pr-awaiting-review]

aaronmarkham:
@zhouhang95 Does this PR look good to you? I need a couple of approvals before merging it.

zhouhang95:
My PR is right. MXNet use opencv as its image process lib. The arg is  in accordance with the opencv arg.

PR #15097, opened 2019-05-29
[BUGFIX] fix unknown parameter shapes when np_shape is turned on.

zheng-da:
## Description ##
When np_shape is turned on, the unknown parameter shapes should be -1 instead of 0. So when a parameter is created but isn't initialized, the dimension of unknown size is -1 when np_shape is turned on, and is 0, otherwise.

### Essentials ###

### DIFF
--- a/python/mxnet/gluon/parameter.py
+++ b/python/mxnet/gluon/parameter.py
@@ -160,6 +160,9 @@ class Parameter(object):
         if self._shape is None:
             return None
         elif is_np_shape():
+            # Parameters shouldn't be zero-size. If one of its dimension is 0,
+            # it means the parameter isn't initialized. In the NumPy semantics,
+            # the unknown dimension should be marked with -1.
             return tuple(i if i != 0 else -1 for i in self._shape)
         else:
             return self._shape
--- a/python/mxnet/gluon/parameter.py
+++ b/python/mxnet/gluon/parameter.py
@@ -157,6 +157,11 @@ class Parameter(object):
 
     @property
     def shape(self):
+        """The shape of the parameter.
+
+        By default, an unknown dimension size is 0. However, when the NumPy semantic
+        is turned on, unknown dimension size is -1.
+        """
         if self._shape is None:
             return None
         elif is_np_shape():
--- a/python/mxnet/gluon/parameter.py
+++ b/python/mxnet/gluon/parameter.py
@@ -157,7 +157,9 @@ class Parameter(object):
 
     @property
     def shape(self):
-        if is_np_shape():
+        if self._shape is None:
+            return None
+        elif is_np_shape():
             return tuple(i if i != 0 else -1 for i in self._shape)
         else:
             return self._shape
--- a/python/mxnet/gluon/parameter.py
+++ b/python/mxnet/gluon/parameter.py
@@ -31,6 +31,7 @@ from .. import symbol, ndarray, initializer, context
 from ..context import Context, cpu
 from .. import autograd
 from .utils import _indent, _brief_print_list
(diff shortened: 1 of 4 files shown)


python/mxnet/gluon/parameter.py
```diff
@@ -156,7 +157,10 @@ def grad_req(self, req):
 
     @property
     def shape(self):
-        return self._shape
+        if is_np_shape():
+            return tuple(i if i != 0 else -1 for i in self._shape)
```
reminisce: I think you also need to handle `self._shape is None`.

tests/python/unittest/test_gluon.py
```diff
@@ -2726,6 +2726,23 @@ def hybrid_forward(self, F, x):
...
+@with_seed()
+def test_np_shape_parameters():
+    class Foo(gluon.Block):
+        def __init__(self, **kwargs):
+            super(Foo, self).__init__(**kwargs)
+            self.dense = gluon.nn.Dense(16)
+        def forward(self, x):
+            return self.dense(x)
+
+    mx.set_np_shape(True)
```
reminisce: It would be safer to use:
```python
with mx.np_shape(True):
    # test code
```

reminisce:
I think the following two places checking whether the shape is known should be modified too.
https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/gluon/parameter.py#L272
https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/gluon/parameter.py#L383

reminisce:
I'm worried about this line
https://github.com/apache/incubator-mxnet/blob/5fc4fc53df74f276aafa51208142e657e9cfe42d/python/mxnet/gluon/parameter.py#L168

Users may notice that we started to use `-1` to represent unknown dim size in np shape semantics and some of them in practice may use shapes like `(-1, 10)` to create parameters for deferred initialization, instead of `(0, 10)`. In those cases, we need to support that unknown dim size can be either `0` or `-1`, so this line should become
```python
all(j in (-1, 0, i) for i, j in zip(new_shape, self._shape))
```
when `is_np_shape = True`.

@szha What do you think?

abhinavs95:
@mxnet-label-bot add [Gluon]

szha:
@reminisce I had similar concern though I thought the API doc should do the job. another option is to expose an internal interface for infer_shape to use, instead of piggybacking on the `.shape` attribute

roywei:
@reminisce @sza @zheng-da  what's the next step here? There is nothing to prevent users from using `-1` as unknown dim. so maybe state clearly in comment or API doc and note experimental?

reminisce:
I think it would be safer to note `np_shape` as experimental and still under development.

zheng-da:
where should we document `-1` is unknown dim and is experimental?

python/mxnet/gluon/parameter.py
```diff
@@ -156,7 +157,12 @@ def grad_req(self, req):
 
     @property
     def shape(self):
-        return self._shape
+        if self._shape is None:
+            return None
+        elif is_np_shape():
+            return tuple(i if i != 0 else -1 for i in self._shape)
```
roywei: a simple comment about -1

python/mxnet/gluon/utils.py
```diff
@@ -412,3 +413,17 @@ def __enter__(self):
 
     def __exit__(self, ptype, value, trace):
         self.detach()
+
+def shape_is_known(shape):
+    """Check whether a shape is completely known w/ or w/o np semantics."""
```
roywei: maybe a reference to np shape for more details.

Issue #15084, opened 2019-05-28
MXNet GPU MKLDNN Build failure


roywei:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-15082/2/pipeline

This is currently blocking all PRs in CI.

jlcontreras:
Thanks for pointing it out! We are having a look to try and identify the problem. Any ideas what might have caused it?

frankfliu:
@mxnet-label-bot add [mkldnn, build, CI]

aaronmarkham:
> Thanks for pointing it out! We are having a look to try and identify the problem. Any ideas what might have caused it?

I we look at the trend, there's a cut off point. Should these commits be rolled back?
http://jenkins.mxnet-ci.amazon-ml.com/job/mxnet-validation/job/unix-gpu/job/master/

roywei:
see detailed discussion in  https://github.com/apache/incubator-mxnet/pull/15099
We need to keep this open until root cause found.

As next step maybe try remove some build flags and see if it helps.

anirudh2290:
I tried again with 

```
make DEV=1 ENABLE_TESTCOVERAGE=1 USE_CPP_PACKAGE=1 USE_BLAS=openblas USE_CUDA=1 USE_CUDA_PATH=/usr/local/cuda USE_CUDNN=1 'CUDA_ARCH=-gencode=arch=compute_52,code=sm_52 -gencode=arch=compute_70,code=sm_70' USE_SIGNAL_HANDLER=1 -j$(nproc)
```

and now I am not able to reproduce the bug :). Would be really helpful to me if you can try the same @roywei .

In the meanwhile I am looking more into this, seeing if something in the archive utility could have caused this.

anirudh2290:
This seems to be related to the archive utility ar bug: https://sourceware.org/bugzilla/show_bug.cgi?id=14625 . It has been fixed in a later version of archive utility in this patch : https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;h=e6cc316af931911da20249e19f9342e5cf8aeeff.  I am trying to reproduce the issue and have also downloaded the binutils source for the version where the issue is fixed. It is very likely that the mxnet archive crossed the 4G limit. If I am able to confirm this, this will require the CI to use a newer version of the archive utility.

anirudh2290:
[...]

PR #15033, opened 2019-05-22
Improve FC perf when no_bias=False

anirudh2290:
## Description ##
Improve FullyConnected op performance when no_bias is set to False. Fixes: https://github.com/apache/incubator-mxnet/issues/14938

Perf Results:
```
Before: 

GPU version: 
Flatten = T, Bias = T, Parameters count = 50000560 . Forward time: 0.4266066551208496, Backward time: 0.2575654983520508, Total time: 0.6841721534729004 
Flatten = T, Bias = F, Parameters count = 50000510 . Forward time: 0.42148637771606445, Backward time: 0.2561371326446533, Total time: 0.6776235103607178 
Flatten = F, Bias = T, Parameters count = 5005060 . Forward time: 0.39539504051208496, Backward time: 1.700193166732788, Total time: 2.095588207244873 
Flatten = F, Bias = F, Parameters count = 5005010 . Forward time: 0.3380405902862549, Backward time: 0.44127631187438965, Total time: 0.7793169021606445 

CPU version: 
Flatten = T, Bias = T, Parameters count = 50000560 . Forward time: 3.4699082374572754, Backward time: 7.634012699127197, Total time: 11.103920936584473 
Flatten = T, Bias = F, Parameters count = 50000510 . Forward time: 3.4408316612243652, Backward time: 6.942991733551025, Total time: 10.38382339477539 
Flatten = F, Bias = T, Parameters count = 5005060 . Forward time: 6.389216899871826, Backward time: 11.683183908462524, Total time: 18.07240080833435 
Flatten = F, Bias = F, Parameters count = 5005010 . Forward time: 5.604649543762207, Backward time: 7.6611716747283936, Total time: 13.2658212184906 


After: 

GPU version: 
Flatten = T, Bias = T, Parameters count = 50000560 . Forward time: 0.42462778091430664, Backward time: 0.25749874114990234, Total time: 0.682126522064209 
Flatten = T, Bias = F, Parameters count = 50000510 . Forward time: 0.42075681686401367, Backward time: 0.2557032108306885, Total time: 0.6764600276947021 
Flatten = F, Bias = T, Parameters count = 5005060 . Forward time: 0.3945605754852295, Backward time: 0.4475700855255127, Total time: 0.8421306610107422 
Flatten = F, Bias = F, Parameters count = 5005010 . Forward time: 0.3443477153778076, Backward time: 0.44979381561279297, Total time: 0.7941415309906006 

CPU version: 
[...]

PR #15099, opened 2019-05-29
Revert "Improve FC perf when no_bias=False"

roywei:
Reverts apache/incubator-mxnet#15033
@stu1130  and I are running a binary search on the commits that may cause https://github.com/apache/incubator-mxnet/issues/15084

This seems to be the one. The reason causing CI build failure is still unknown, as this only happens with CI enviroments, builidng locally with GPU + MKLDNN is fine on master.

I m creating this PR to test if CI check can pass

### DIFF
--- a/src/operator/nn/fully_connected-inl.h
+++ b/src/operator/nn/fully_connected-inl.h
@@ -36,7 +36,6 @@
 #include "../elemwise_op_common.h"
 #include "../linalg.h"
 #include "../../common/utils.h"
-#include "../tensor/broadcast_reduce_op.h"
 
 namespace mxnet {
 namespace op {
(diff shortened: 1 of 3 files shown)


szha:
cc @anirudh2290

anirudh2290:
Really weird! CI check passed even on my PR. So if CI check doesn't fail, does that tell much ?

szha:
We certainly need to find the root cause on why CI wasn't reliable in this case. (e.g. could it have been just a cache issue?)

larroy:
The root cause seems to be not enough shared memory in the docker container used in CI.

Using time ci/build.py --docker-registry mxnetci --platform ubuntu_build_cuda --docker-build-retries 3 --shm-size 8G /work/runtime_functions.sh build_ubuntu_gpu_mkldnn

fixes the build.

We have some files which are very slow to compile and produce big binaries, I think is excess of inline.

larroy:
https://github.com/apache/incubator-mxnet/pull/15100

szha:
Hmm, if this was required, how did the original PR pass CI check?

anirudh2290:
I was able to reproduce this with make with on my local machine with command: 
```
make DEV=1 ENABLE_TESTCOVERAGE=1 USE_CPP_PACKAGE=1 USE_BLAS=openblas USE_CUDA=1 USE_CUDA_PATH=/usr/local/cuda USE_CUDNN=1 'CUDA_ARCH=-gencode=arch=compute_52,code=sm_52 -gencode=arch=compute_70,code=sm_70' USE_SIGNAL_HANDLER=1 -j$(nproc)
```
Note: I was able to reproduce only with a clean build after make clean. for example, after building on top of previous commit i wasn't able to reproduce this.

roywei:
@stu1130 has tried with ~2G and failed. It also seems 8G in #15100 failed.
I just tried 1000m passed
```
ci/build.py --docker-registry mxnetci --platform ubuntu_build_cuda --docker-build-retries 3 --shm-size 10000m /work/runtime_functions.sh build_ubuntu_gpu_mkldnn
```

We have tried with 500m before https://github.com/apache/incubator-mxnet/pull/15033, everything works fine. Everything after this commit failed with 500m. 
For local tests, also able to reproduce now with @anirudh2290 's command.
So this can't be only shared memory related. 

1. why https://github.com/apache/incubator-mxnet/pull/15033 code change caused CI requiring more space? (it changed requiring additional temp space to not limit to MKLDNN only, but should not affect build?)
2. why CI check passed for initial PR?
3. how to fix local build?

I would prefer to merge this and keep https://github.com/apache/incubator-mxnet/issues/15084 open until we find root cause.

anirudh2290:
@roywei sounds good to me.

aaronmarkham:
Is there are next step for those of use with PRs that failed CI due to this issue?

anirudh2290:
@aaronmarkham Rebase and retrigger CI and it should be fine.

larroy:
We need to investigate what's going on, I thought it passed for me at 8G but then my PR validation failed. Could be a linker bug or an offset linking bigger than 2^32, it's still not 100% clear to me.

larroy:
I'd suggest strace dump of linker

larroy:
Just failed with 16G for me.

Issue #14584, opened 2019-04-01
Conversion from FP32 to Mixed Precision Models


anirudh2290:
## API Addition

Users want to bring a FP32 model and convert it to a Mixed precision model to run inference on it. They want to use the model zoo to convert pretrained models in Python and other frontends. They can do this with gluon models today by casting the inputs and the blocks but the same cannot be done for symbolic models (json and params). Proposing to add an API to convert FP32 models to FP16.

Considering the recent AMP work in progress here: https://github.com/apache/incubator-mxnet/pull/14173, I think we should add a conversion API to FP16 model under AMP namespace:
```
amp.convert_model(sym, arg_params, aux_params, target_dtype="float16", 
                  target_precision_ops=None, original_precision_ops=None,
                  widest_precision_ops=None, excluded_sym_names=None)
```
With the `target_precision_ops`, `original_precision_ops ` and `widest_precision_ops`, users should be able to override the default in the amp lists.

## Backend Changes

Additionally, Add a NNVM pass for the backend. This would by default use the amp lists for FP16, FP32 and widest type casts to use FP16 or FP32 inputs.
This pass will perform graph traversal and adding amp_cast and amp_multicast layers for FP16 and FP32 ops. 

Planning to start working on the POC unless someone is already working on this.

@ptrendx @DickJC123 @pengzhao-intel @ZhennanQin @eric-haibin-lin @Caenorst

EDIT: Proposal posted on dev list: https://cwiki.apache.org/confluence/display/MXNET/Conversion+from+FP32+to+Mixed+Precision+Models

ZhennanQin:
Hi @anirudh2290, It's really good to have AMP supported for symbolic models.

One thing I want to mention is, we want to let AMP support all kinds of low precision float, not only `FP16`. For the API itself, I think it's OK as target_dtype and op list are all configurable. For the NNVM pass implementation, we wish to avoid hard code `FP16` and make it easy extend to support other low precision float, e.g. `BF16`.

anirudh2290:
@ZhennanQin Thanks for the suggestion! Yes the plan is tho keep the NNVM pass target_dtype agnostic and allow for easy extension for BF16. In future, expectation is  amp_cast and amp_multicast ops should support BF16 apart from FP16 and there should be corresponding lists like these for (https://github.com/apache/incubator-mxnet/pull/14173/files#diff-b79bfa3e02355c43ca5b195ef67172a5R21) BF16 too.

[...]

PR #15118, opened 2019-06-01
Conversion from FP32 model to Mixed Precision model

anirudh2290:
## Description ##

Users want to bring a FP32 model, to convert it to a mixed precision model to run inference on it. This work leverages the existing work already done by @ptrendx, @Caenorst with AMP and tries to provide users with conversion APIs to convert their symbolic model or gluon model to a mixed precision model. This also adds the necessary C APIs, so that similar support for conversion APIs can be added in other frontends.

Thanks to all involved in prior discussions, suggestions and design review of the project (sincere apologies if I missed someone):
@ptrendx, @rahul003, Sudipta Sengupta (AWS), @bhagatindia (AWS), Wei Xiao (AWS), @Vikas89, @lupesko, @pengzhao-intel , @ZhennanQin 

 
## API Additions (Python) ##

1. Converting a symbolic model:
```
convert_model(sym, arg_params, aux_params, target_dtype="float16",
              target_dtype_ops=None, fp32_ops=None,
              conditional_fp32_ops=None, excluded_sym_names=None,
              cast_optional_params=False)
```

2. Converting a gluon model (hybrid block):
```
convert_hybrid_block(block, target_dtype="float16", target_dtype_ops=None,
                     fp32_ops=None, conditional_fp32_ops=None, excluded_sym_names=None,
                     ctx=mx.gpu(0), cast_optional_params=False)
```

3. Converting a symbol:
```
convert_symbol(sym, target_dtype="float16", target_dtype_ops=None,
               fp32_ops=None, conditional_fp32_ops=None,
               excluded_sym_names=None, data_names=None, cast_optional_params=False)
```

### Module API (executor_group.py) ###
1. Added support for retrieving input_types from the symbol and updating the input_types dict used in simple_bind.

### Gluon API (parameter.py) ###
1. Refactored code to support loading params from param_dict as well as file. This is used in the convert_hybrid_block API.

### Test Utils API (test_utils.py) ###
1. Copied download_model from example/image-classification/common/modelzoo.py to test_utils since this is a common use case and used in the tests

### AMP Tests ###
1. Moved the AMP tests to gpu directory since wanted to limit AMP tests to one file and wanted to add some additional tests to it which work only on GPU.

## Additions ##
1. Added APIs mentioned above in python/mxnet/contrib/amp/amp.py.
2. C API for symbol conversion.
2. Added nnvm pass in src/nnvm/low_precision_pass.cc
3. Added example, tutorials and tests.

Fixes : https://github.com/apache/incubator-mxnet/issues/14584
Doc: https://tinyurl.com/y42kx9hl

Other Flaky Tests/Bug Fixes:
1. Attempts to fix: https://github.com/apache/incubator-mxnet/issues/14978

### Essentials ###

### Changes ###

## Comments ##


anirudh2290:
cc @ptrendx @ZhennanQin @pengzhao-intel

docs/tutorials/amp/amp_tutorial.md
```diff
@@ -249,6 +251,46 @@ for epoch in range(1):
 
 We got 60% speed increase from 3 additional lines of code!
 
+## Inference with AMP
+
+To do inference with mixed precision for a trained model in FP32, you can use the conversion APIs: `amp.convert_model` for symbolic model and `amp.convert_hybrid_block` for gluon models. The conversion APIs will take the FP32 model as input and will return a mixed precision model, which can be used to run inference. Below, we demonstrate for a gluon model and a symbolic model: 1. Conversion from FP32 model to mixed precision model 2. Run inference on the mixed precision model.
+
+```python
+with mx.Context(mx.gpu(0)):
```
ptrendx: Maybe do
```suggestion
with ctx = mx.Context(mx.gpu(0)):
```
and then use this `ctx` instead of `mx.gpu()`?

src/c_api/c_api_symbolic.cc
```diff
@@ -810,6 +810,156 @@ int MXQuantizeSymbol(SymbolHandle sym_handle,
...
+  g.attrs["conditional_fp32_ops"] =
+      std::make_shared<nnvm::any>(std::move(conditional_fp32_ops));
+  g.attrs["excluded_syms"] =
+      std::make_shared<nnvm::any>(std::move(excluded_syms));
+  g.attrs["target_dtype"] = std::make_shared<nnvm::any>(target_dt);
+
+  g = ApplyPass(std::move(g), "ReducePrecision");
+  // Need to run type inference since it is possible that inferred
+  // type of some inputs has changed
+  g = mxnet::exec::InferType(std::move(g), std::move(arg_types), "");
```
ptrendx: If you already did InferType before on the original graph then I guess you should remove the dtype AttrVector from the graph, right?
Also, I'm not sure if indexed graph is recreated after your pass.

docs/tutorials/amp/amp_tutorial.md
anirudh2290: thanks for pointing this. I will change the ctx used inside to mx.current_context() since that is the correct way to use this context manager.

src/c_api/c_api_symbolic.cc
anirudh2290: By passing "" as the third argument, I ask the InferType to ignore existing dtype attrs in the symbol. arg_types is the input_types passed to trigger the infer pass. This may or may not include the param types.

> Also, I'm not sure if indexed graph is recreated after your pass.

I am not sure I understand. The input_nodes shouldn't be affected by the AMP pass.

piyushghai:
@mxnet-label-bot Add [pr-awaiting-review, Mixed Precision, FP32]

src/c_api/c_api_symbolic.cc
anirudh2290: To add more to this, the indexed_graph after applying graph pass will still have same input_nodes with the same ordering after the NNVM pass is applied. This is because the ordering of the inputs will be preserved by the NNVM pass.

anirudh2290:
@Vikas89 @mseth10 @larroy @access2rohit @stu1130 @rondogency @Zha0q1 @samskalicky @apeforest @yuxihu @sandeep-krishnamurthy @nswamy Please help review.

pengzhao-intel:
@anirudh2290 thanks for great works and we are working on BF16 enabling in local. 
So I am asking @ZhennanQin to take a review and see if both solutions are aligned.

src/nnvm/low_precision_pass.cc
```diff
@@ -0,0 +1,265 @@
...
+          AddCastNode(e, suffix, mirror_entry, "float16",
+                      &mirror_target_dtype_map, new_node);
+        }
+      }
+    } else if (!node->is_variable() &&
+               widest_dtype_ops.count(node->op()->name) > 0 &&
+               excluded_syms.count(node->attrs.name) == 0) {
+      CHECK(node->inputs.size() > 0)
+          << "Please check the symbol. node name: " << node->attrs.name
+          << "op name " << node->op()->name << " has no inputs";
```
samskalicky: If this check fails, what does it mean? And what should the user do if it fails? The message says "check the name" but what are they supposed to check for, and why did this fail?

Can you improve the error message and add some more comments to explain to devs what they should do if this check fails?

tests/python/gpu/test_contrib_amp.py
```diff
@@ -0,0 +1,255 @@
...
+
+    if ret1 != set():
+        warnings.warn("Operators " + str(ret1) + " do not exist in AMP lists (in "
+                       "python/mxnet/contrib/amp/lists/symbol.py) - please add them. "
+                       """Please follow these guidelines for choosing a proper list:
+                       - if your operator is not to be used in a computational graph
+                         (e.g. image manipulation operators, optimizers) or does not have
+                         inputs, put it in FP16_FP32_FUNCS list,
+                       - if your operator requires FP32 inputs or is not safe to use with lower
+                         precision, put it in FP32_FUNCS list,
```
samskalicky: Based on the discussion on the dev list didnt we decide not to have an explicit list for FP32_FUNCS? And instead assume anything not specified would go in this category? Otherwise this will be a maintenance nightmare and make adding new operators confusing.

tests/python/tensorrt/test_tensorrt_lenet5.py
```diff
@@ -46,7 +46,7 @@ def run_inference(sym, arg_params, aux_params, mnist, all_test_labels, batch_siz
     # Get this value from all_test_labels
     # Also get classes from the dataset
     num_ex = 10000
-    all_preds = np.zeros([num_ex, 10])
+    all_preds = np.zeros([num_ex, 10], dtype=np.float32)
```
samskalicky: isnt this the default? Why do we need to explicitly specify fp32?

src/nnvm/low_precision_pass.cc
anirudh2290: They should check the node name to see that the input is part of the symbol or not. If not maybe something went wrong with symbol construction. I can add more comment to check if something went wrong with symbol construction, but not much more information I can provide here.

src/c_api/c_api_symbolic.cc
```diff
@@ -810,6 +810,156 @@ int MXQuantizeSymbol(SymbolHandle sym_handle,
   API_END_HANDLE_ERROR(delete s);
 }
 
+int MXReducePrecisionSymbol(SymbolHandle sym_handle,
```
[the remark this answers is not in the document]
anirudh2290: i will add a comment here.

tests/python/gpu/test_contrib_amp.py
anirudh2290: i think i just moved this to a different file. @ptrendx  owns this part and can add more info here.

tests/python/tensorrt/test_tensorrt_lenet5.py
anirudh2290: yes by default it is np.float64.

samskalicky: Why do we need to specify the type here? Arent we just going to override later with the outputs:

```
all_preds[offset:offset+extent, :] = executor.outputs[0].asnumpy()[:extent]
```

samskalicky:
Does this only work with GPU? It looks like you removed the test from the CPU unittests directory.

tests/python/tensorrt/test_tensorrt_lenet5.py
anirudh2290: No it wont come from outputs
```
>>> x = np.zeros((2, 3))
>>> y = np.zeros((2, 3), dtype=np.float32)
>>> x[:] = y
>>> x
array([[ 0.,  0.,  0.],
       [ 0.,  0.,  0.]])
>>> x.dtype
dtype('float64')
```

anirudh2290:
@samskalicky the conversion itself will work but the inference is likely not to work because operator support for FP16 not being great on CPU.  Thanks for pointing though, I should still add a conversion test for convert_hybrid_block for cpu.

python/mxnet/contrib/amp/amp.py
```diff
@@ -342,3 +349,320 @@ def unscale(optimizer_or_trainer):
...
+            assert name in aux_names
+            arg_dict['aux:%s'%name] = param._reduce()
+            if name in attr_dict and "__dtype__" in attr_dict[name]:
+                if attr_dict[name]["__dtype__"] != "-1":
+                    typ = _DTYPE_MX_TO_NP[int(attr_dict[name]["__dtype__"])]
+                    arg_dict['aux:%s'%name] = arg_dict['aux:%s'%name].astype(typ)
+
+    # Create a symbolblock and cast the params to the dtypes based
+    # on the dtype information from the converted_symbol
+    ret = SymbolBlock(converted_sym, inputs)
```
larroy: just curious, why are we using SymbolBlock here?

python/mxnet/gluon/parameter.py
```diff
@@ -919,23 +919,49 @@ def load(self, filename, ctx=None, allow_missing=False,
                 assert name.startswith(restore_prefix), \
                     "restore_prefix is '%s' but Parameters name '%s' does not start " \
                     "with '%s'"%(restore_prefix, name, restore_prefix)
-        lprefix = len(restore_prefix)
         ndarray_load = ndarray.load(filename)
+        self.load_dict(ndarray_load, ctx, allow_missing,
+                       ignore_extra, restore_prefix, filename, cast_dtype)
+
+    def load_dict(self, param_dict, ctx=None, allow_missing=False,
```
larroy: Is the change related to this PR?

python/mxnet/module/executor_group.py
```diff
@@ -651,6 +652,20 @@ def _bind_ith_exec(self, i, data_shapes, label_shapes, shared_group):
             input_shapes.update(dict(label_shapes))
 
         input_types = {x.name: x.dtype for x in data_shapes}
+        attr_dict = self.symbol.attr_dict()
+
+        for sym_name in self.symbol.list_arguments():
```
larroy: Could this introduce regressions? 

Can we fold the nested ifs using and or better a function that encapsulates the criteria and makes this easier to read?

python/mxnet/test_utils.py
```diff
@@ -1521,6 +1522,72 @@ def download(url, fname=None, dirname=None, overwrite=False, retries=5):
     logging.info("downloaded %s into %s successfully", url, fname)
     return fname
 
+def download_model(model_name, dst_dir='./', meta_info=None):
```
larroy: should this be a separate PR?

python/mxnet/test_utils.py
```diff
@@ -2073,6 +2140,22 @@ def compare_optimizer(opt1, opt2, shape, dtype, w_stype='default', g_stype='defa
         compare_ndarray_tuple(state1, state2, rtol=rtol, atol=atol)
     assert_almost_equal(w1.asnumpy(), w2.asnumpy(), rtol=rtol, atol=atol)
 
+
+def compare_symbol_structure(sym1, sym2):
```
larroy: could be give it a better name? compare is not very clear. Maybe same ops or something like that?

python/mxnet/contrib/amp/amp.py
```diff
@@ -342,3 +349,320 @@ def unscale(optimizer_or_trainer):
...
+    target_dtype : str or numpy, optional defaults to float16
+        currently only supports float16. The target dtype indicates to add cast layers
+        when possible so that lower precision computation can be leveraged.
+    target_dtype_ops : list of strs, optional
+        Override the list of operator names casted to the target_dtype.
+        If None, uses the framework's default list to be casted to target_dtype.
+    fp32_ops : list of strs, optional
+        Override the list of operator names casted to FP32.
+        If None, uses the framework's default list to be casted to FP32.
+    conditional_fp32_ops : list of (string, string, list of string), optional
```
larroy: does it make sense to use a class here?  the constraints on the type seem a bit complex to me that it might be worth defining a helper class or named tuple. Just something to consider.

src/c_api/c_api_symbolic.cc
```diff
@@ -810,6 +810,156 @@ int MXQuantizeSymbol(SymbolHandle sym_handle,
...
+                            const mx_uint num_model_params,
+                            const char **target_dtype_op_names,
+                            const char **fp32_op_names,
+                            const char **widest_dtype_op_names,
+                            const char **conditional_fp32_op_names,
+                            const char **excluded_symbols,
+                            const char **param_names,
+                            const char **param_vals,
+                            const char **model_param_names,
+                            const char **arg_names) {
```
larroy: I would consider using namespace std or using some of the data structures like string etc to make it less noisy.

src/c_api/c_api_symbolic.cc
```diff
@@ -810,6 +810,156 @@ int MXQuantizeSymbol(SymbolHandle sym_handle,
...
+  nnvm::Graph g = Symbol2Graph(*sym);
+  std::unordered_set<std::string> target_dtype_ops;
+  std::unordered_set<std::string> fp32_ops;
+  std::unordered_set<std::string> widest_dtype_ops;
+  std::unordered_set<std::string> excluded_syms;
+  std::unordered_set<std::string> model_params;
+  std::unordered_map<std::string,
+                     std::unordered_map<std::string,
+                                        std::vector<std::string>>> conditional_fp32_ops;
+  int target_dt = *target_dtype;
```
larroy: do we need this variable?

src/c_api/c_api_symbolic.cc
```diff
@@ -810,6 +810,156 @@ int MXQuantizeSymbol(SymbolHandle sym_handle,
...
+  for (size_t i = 0; i < num_target_dtype_op_names; ++i) {
+    target_dtype_ops.emplace(target_dtype_op_names[i]);
+  }
+  for (size_t i = 0; i < num_fp32_op_names; ++i) {
+    fp32_ops.emplace(fp32_op_names[i]);
+  }
+  for (size_t i = 0; i < num_widest_dtype_op_names; ++i) {
+    widest_dtype_ops.emplace(widest_dtype_op_names[i]);
+  }
+  for (size_t i = 0; i < num_excluded_symbols; ++i) {
```
larroy: Does it make sense to use at or assert that the num_ variables matches num elements?

src/c_api/c_api_symbolic.cc
```diff
@@ -810,6 +810,156 @@ int MXQuantizeSymbol(SymbolHandle sym_handle,
...
+  g = ApplyPass(std::move(g), "ReducePrecision");
+  // Need to run type inference since it is possible that inferred
+  // type of some inputs has changed
+  g = mxnet::exec::InferType(std::move(g), std::move(arg_types), "");
+  const nnvm::DTypeVector &inferred_dtypes =
+      g.GetAttr<nnvm::DTypeVector>("dtype");
+  const nnvm::IndexedGraph &idx = g.indexed_graph();
+  const std::string dtype_keyword = "__dtype__";
+
+  for (uint32_t nid : idx.input_nodes()) {
```
larroy: I would separate this into a function, also, can node be const?

src/c_api/c_api_symbolic.cc
```diff
@@ -810,6 +810,156 @@ int MXQuantizeSymbol(SymbolHandle sym_handle,
...
+                            const char **target_dtype_op_names,
+                            const char **fp32_op_names,
+                            const char **widest_dtype_op_names,
+                            const char **conditional_fp32_op_names,
+                            const char **excluded_symbols,
+                            const char **param_names,
+                            const char **param_vals,
+                            const char **model_param_names,
+                            const char **arg_names) {
+  nnvm::Symbol *s = new nnvm::Symbol();
```
larroy: can we improve those variable names? s and sym? more semantic.

src/c_api/c_api_symbolic.cc
```diff
@@ -810,6 +810,156 @@ int MXQuantizeSymbol(SymbolHandle sym_handle,
...
+        node_without_dtype_map[node->attrs.name] =
+            inferred_dtypes[idx.entry_id(nid, 0)];
+      }
+    }
+  }
+
+  s->outputs = g.outputs;
+  *ret_sym_handle = s;
+  nnvm::Symbol *ret_sym = static_cast<nnvm::Symbol *>(*ret_sym_handle);
+  auto args = ret_sym->ListInputs(nnvm::Symbol::kAll);
```
larroy: I would type it for readability, also const?

src/c_api/c_api_symbolic.cc
```diff
@@ -810,6 +810,156 @@ int MXQuantizeSymbol(SymbolHandle sym_handle,
...
+      if (inferred_dtypes[idx.entry_id(nid, 0)] == -1) {
+        node_without_dtype_map[node->attrs.name] = 0;
+      } else {
+        node_without_dtype_map[node->attrs.name] =
+            inferred_dtypes[idx.entry_id(nid, 0)];
+      }
+    }
+  }
+
+  s->outputs = g.outputs;
```
larroy: Can this be separated on its own function?

src/nnvm/low_precision_pass.cc
```diff
@@ -0,0 +1,265 @@
...
+ * under the License.
+ */
+
+/*!
+ *  Copyright (c) 2016 by Contributors
+ * \file low_precision_pass.cc
+ * \brief Return new graph with amp_cast and amp_multicast operators added wherever required
+ */
+
+#include <nnvm/node.h>
```
larroy: please don't use diamonds for non system includes.

src/nnvm/low_precision_pass.cc
```diff
@@ -0,0 +1,265 @@
...
+NodePtr InsertNode(std::string op_name, std::string node_name, NodePtr current,
+                   NodeEntry previous) {
+    NodePtr node = CreateNode(op_name, node_name);
+    node->inputs.emplace_back(previous);
+    current->inputs.emplace_back(NodeEntry{node, 0, 0});
+    return node;
+}
+
+// get suffix for a node entry so that it can be used for amp_cast/amp_multicast node name
+std::string GetSuffix(const nnvm::NodeEntry &e,
```
larroy: can we rename e to node_entry?

src/nnvm/low_precision_pass.cc
```diff
@@ -0,0 +1,265 @@
...
+void AddCastNode(const nnvm::NodeEntry &e, const std::string &suffix,
+                 const nnvm::NodeEntry &input, const std::string dtype,
+                 nnvm::NodeEntryMap<NodeEntry> *mirror_entry_map,
+                 NodePtr curr_node) {
+  NodePtr cast_node =
+      InsertNode("amp_cast", e.node->attrs.name + suffix + "_amp_cast_" + dtype,
+                 curr_node, input);
+  cast_node->attrs.dict["dtype"] = dtype;
+  cast_node->op()->attr_parser(&(cast_node->attrs));
+  (*mirror_entry_map)[e] = NodeEntry{cast_node, 0, e.version};
```
larroy: microopt: you can use std::move(cast_node) since it's not used any longer. Same below.

src/nnvm/low_precision_pass.cc
```diff
@@ -0,0 +1,265 @@
...
+
+// add amp_multicast node between curr_node and inputs
+void AddMultiCastNode(const std::vector<NodeEntry> &inputs,
+                      const std::string &node_name,
+                      const std::unordered_map<Node *, NodePtr> &mirror_map,
+                      NodePtr curr_node) {
+    NodePtr node = CreateNode("amp_multicast",
+                              inputs[0].node->attrs.name + node_name + "_amp_multicast");
+    for (size_t i = 0; i < inputs.size(); ++i) {
+    const auto &e = inputs[i];
```
larroy: indent is foobar

src/nnvm/low_precision_pass.cc
```diff
@@ -0,0 +1,265 @@
...
+    NodePtr mirror_node = mirror_map.at(e.node.get());
+    NodeEntry mirror_entry = NodeEntry{mirror_node, e.index, e.version};
```
larroy: you can move

src/nnvm/low_precision_pass.cc
```diff
@@ -0,0 +1,265 @@
...
+      << "Only float16 target_dtype is supported yet";
+
+  // Additional data structures to share common cast node inputs among different nodes
+  std::unordered_map<Node *, NodePtr> mirror_map;
+  nnvm::NodeEntryMap<NodeEntry> mirror_fp32_map;
+  nnvm::NodeEntryMap<NodeEntry> mirror_target_dtype_map;
+
+  // Visit nodes in a topologically sorted order
+  DFSVisit(src.outputs, [&](const NodePtr &node) {
+    NodePtr new_node = Node::Create();
```
larroy: can't you pass node directly to Create()?  it's forwarding args so it should call the copy ctor.

src/nnvm/low_precision_pass.cc
```diff
@@ -0,0 +1,265 @@
...
+     * (to target_dtype) after its inputs
+     * 3. for nodes which need to run in widest dtype among its inputs, add
+     * amp_multicast operators between op and its inputs
+     * 4. for nodes which need to run in FP32 mode, based on a specific condition,
+     * check the condition, and if true add amp_cast (to fp32) after its inputs
+     * 4. for other nodes, create copy node and add it to the mirror_map
+     */
+    if (!node->is_variable() && fp32_ops.count(node->op()->name) > 0 &&
+        excluded_syms.count(node->attrs.name) == 0) {
+      for (size_t i = 0; i < node->inputs.size(); ++i) {
```
larroy: can you use range for here since you are not using the index?

src/nnvm/low_precision_pass.cc
```diff
@@ -0,0 +1,265 @@
...
+          NodeEntry mirror_entry = NodeEntry{mirror_node, e.index, e.version};
+          std::string suffix = GetSuffix(e, mirror_map);
+          AddCastNode(e, suffix, mirror_entry, "float32", &mirror_fp32_map,
+                      new_node);
+        }
+      }
+    } else if (!node->is_variable() &&
+               target_dtype_ops.count(node->op()->name) > 0 &&
+               excluded_syms.count(node->attrs.name) == 0) {
+      for (size_t i = 0; i < node->inputs.size(); ++i) {
```
larroy: same as above

python/mxnet/contrib/amp/amp.py
larroy: I also don't understand this parameter.

src/nnvm/low_precision_pass.cc
```diff
@@ -0,0 +1,265 @@
...
+        NodeEntry mirror_entry = NodeEntry{mirror_node, e.index, e.version};
+        new_node->inputs.emplace_back(mirror_entry);
+      }
+    }
+    mirror_map[node.get()] = std::move(new_node);
+  });
+
+  std::vector<NodeEntry> outputs;
+  for (const auto& e : src.outputs) {
+      outputs.emplace_back(NodeEntry{mirror_map.at(e.node.get()), e.index, e.version});
```
larroy: You can use directly the constructor arguments instead of copy constructing.

src/nnvm/low_precision_pass.cc
```diff
@@ -0,0 +1,265 @@
...
+          AddCastNode(e, suffix, mirror_entry, "float32", &mirror_fp32_map,
+                      new_node);
+        }
+      }
+    } else {
+      for (size_t i = 0; i < node->inputs.size(); ++i) {
+        const auto &e = node->inputs[i];
+        NodePtr mirror_node = mirror_map.at(e.node.get());
+        NodeEntry mirror_entry = NodeEntry{mirror_node, e.index, e.version};
+        new_node->inputs.emplace_back(mirror_entry);
```
larroy: move

tests/python/gpu/test_contrib_amp.py
```diff
@@ -0,0 +1,255 @@
...
+    ret = [item for item, count in collections.Counter(t).items() if count > 1]
+    assert ret == [], "Elements " + str(ret) + " exist in more than 1 AMP list."
+
+    # Check the coverage
+    py_str = lambda x: x.decode('utf-8')
+
+    plist = ctypes.POINTER(ctypes.c_char_p)()
+    size = ctypes.c_uint()
+
+    mx.base._LIB.MXListAllOpNames(ctypes.byref(size),
```
larroy: Don't we have a python binding for this?

python/mxnet/module/executor_group.py
anirudh2290: this is actually improving AMP support for the module api. For example after amp_cast nodes are added dtypes of params won't be able to be inferred easily, so before we provide a default we check the symbol itself if it has a dtype and provide that info to simple_bind. I can refactor to make it easier to read and also think if I can add some test for this.

python/mxnet/contrib/amp/amp.py
anirudh2290: Given a symbol and params, SymbolBlock is supposed to be used to load the model and run inference on it. Other Hybrid blocks are more for the hybrid mode, where you run imperative mode on the first run and cache ops and run forward on cached symbol from the second pass.

python/mxnet/gluon/parameter.py
anirudh2290: yes it is related to the PR, it is used in the convert_hybrid_block

python/mxnet/test_utils.py
anirudh2290: no this is needed for this PR so I added as part of this PR. I thought this was small enough to be included in this PR.

src/c_api/c_api_symbolic.cc
anirudh2290: We need this here to extend the support for bfloat16.

src/c_api/c_api_symbolic.cc
anirudh2290: Since this is C API we cannot use string

python/mxnet/contrib/amp/amp.py
anirudh2290: I think class may not be needed here. There is no state that needs to be maintained and acted upon multiple times and through lifetime of a program. Most users will be calling these conversion APIs once.

src/c_api/c_api_symbolic.cc
larroy: I meant "using namespace std" below, not what you understood.

src/nnvm/low_precision_pass.cc
anirudh2290: this has been used at many places in the codebase, for 3rd party libraries. The link also explains the difference but doesnt explicitly say what is the best practice. In MXNet we use #include <> for all system headers, 3rd party headers etc. Maybe I am missing something, can you elaborate  or provide more links here.

tests/python/gpu/test_contrib_amp.py
anirudh2290: no i don't think we have python binding for this.

src/c_api/c_api_symbolic.cc
anirudh2290: i think these C APIs will be mostly used by developers from the frontend, and we just expect the len of excluded_symbols to be passed correctly. We don't have any such checks for any other C APIs, and I think it is an acceptable tradeoff to make.

python/mxnet/contrib/amp/amp.py
```diff
@@ -342,3 +349,320 @@ def unscale(optimizer_or_trainer):
...
+        If None, uses the framework's default list of widest_precision_ops.
+    conditional_fp32_ops : list of (string, string, list of string)
+        Override the list of operators to be casted to FP32.
+        The format of the list is
+        (name of the function, name of the parameter,
+         list of values of the parameter that make the operator to be casted to
+        fp32)
+    excluded_sym_names : list of strs
+        A list of strings that represent the names of symbols that users want to exclude
+        from being quantized.
```
ZhennanQin: I'm not sure if low precision is a kind of quantization, maybe change `quantized` to `low precision`?

ZhennanQin:
Thanks @anirudh2290 providing this fantastic feature. I see that the ability to extend bfloat16 support is well considered. Here's an question from API design:
```
def convert_hybrid_block(block, target_dtype="float16", target_dtype_ops=None,
                         fp32_ops=None, conditional_fp32_ops=None,
                         excluded_sym_names=None, ctx=gpu(0)):
```
This API is to convert a gluon hybrid block to low precision, and it accepts `excluded_sym_names` to exclude some certain layer. My question is, since gluon layers don't have layer names, then what would you expect user to use this parameter?

python/mxnet/contrib/amp/amp.py
anirudh2290: sure will change this.

anirudh2290:
@ZhennanQin Thank you for the review ! I used excluded_sym_names since there is a use case where customer import a symbol and params using SymbolBlock and convert it to a  mixed precision block. In this case, the advanced user may want to exclude certain sym names and test. I agree this kwarg wont be used by majority of the users but its still kept for the advanced user. Since convert_symbol already supports it, there is hardly any effort from implementation point of view.

src/c_api/c_api_symbolic.cc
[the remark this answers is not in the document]
anirudh2290: There is a comment in .h file

python/mxnet/module/executor_group.py
anirudh2290: i have added a test for this.

src/c_api/c_api_symbolic.cc
anirudh2290: I would prefer keeping it consistent with rest of the codebase

src/nnvm/low_precision_pass.cc
anirudh2290: not sure about this. can you elaborate with code example.

anirudh2290:
@ptrendx @pengzhao-intel @samskalicky @larroy @ZhennanQin Thank you for your review ! I have addressed your comments.

src/nnvm/low_precision_pass.cc
larroy: it's not used correctly, and it's in the process of being fixed.  If you use diamonds it adds to the search path the standard include folders (the ones from the standard library provided by the compiler). Using quotes just searches in the include path of the project as specified in the build.

If you were to do make install for example of another mxnet version and have nnvm headers in /usr/include/nnvm you would be including these wrong headers first instead of the project (your own). Hence this is a bad practice that should be corrected, that's why I suggested not to do it when introducing new files, as the older ones need to be refactored accordingly.

src/nnvm/low_precision_pass.cc
larroy: Why not? 

```
NodePtr new_node = Node::Create(*node);  

```

tests/python/gpu/test_contrib_amp.py
larroy: I think we should, as we are using the same for operator benchmarks done by @sandeep-krishnamurthy

src/nnvm/low_precision_pass.cc
anirudh2290: yes you are right. my editor was getting confused by num of arguments probably because of std::forward. i have fixed this now.

tests/python/gpu/test_contrib_amp.py
anirudh2290: sure, thats a different discussion outside the scope of this PR. Feel free to propose.

src/nnvm/low_precision_pass.cc
anirudh2290: I think that would be expected behavior. First you would provide -I for the path for CFLAGS, which would be searched before sysdirs, and if you intentionally left that out, then you know what you are doing and got what was expected.  Why is it a bad practice ?

src/c_api/c_api_symbolic.cc
```diff
@@ -810,6 +810,191 @@ int MXQuantizeSymbol(SymbolHandle sym_handle,
...
+  nnvm::Symbol *result_sym = new nnvm::Symbol();
+  API_BEGIN();
+  nnvm::Symbol *sym = static_cast<nnvm::Symbol *>(sym_handle);
+  nnvm::Graph g = Symbol2Graph(*sym);
+  std::unordered_set<std::string> target_dtype_ops;
+  std::unordered_set<std::string> fp32_ops;
+  std::unordered_set<std::string> widest_dtype_ops;
+  std::unordered_set<std::string> excluded_syms;
+  std::unordered_set<std::string> model_params;
+  std::unordered_map<std::string,
```
larroy: Suggest to add comment on what this thing represents, or maybe a typedef with a doc. It would help with code maintainability.

src/nnvm/low_precision_pass.cc
anirudh2290: The reason I asked "Why it is a bad practice?" because you called it not a good practice, and bad practice twice, with no justification for the same. 

The links that you provide talk nothing about "bad practice" or "not a good practice", only what they mean.

I think you missed reading the article you sent.

> There are a number of command-line options you can use to add additional directories to the search path. The most commonly-used option is -Idir, which causes dir to be searched after the current directory (for the quote form of the directive) and ahead of the standard system directories. You can specify multiple -I options on the command line, in which case the directories are searched in left-to-right order.


> I tried to explain above without success, please refer to the compiler documentation:
>  If you need more clarification let's talk in person as seems the converstation is stuck.

I don't like your tone. It was certainly not warranted here. Its an easy way to drive away new contributors.

larroy: I think this is a misunderstanding, I wanted to talk about this f2f to clarify the order of includes and not discuss further on this PR.  My apologies if my comment didn't come across in a nice way.


 
I was referring to this part of the documentation I sent:

> include <file>
   This variant is used for system header files


> #include "file"
This variant is used for header files of your own program. 

My understanding was that if you do "make install" and have nnvm includes in the system header, it would be used first before the headers of the own program. But according to the doc seems that since we are using -I this won't be the case. 

I don't think there's a need to get angry about this, it's better to learn together.

anirudh2290:
From an offline review done by sudipta@, feedback was provided that it is important to for users to be able to obtain models with params casted wherever possible. After additional discussion with @ptrendx , we decided to add additional graph pass, which would go through all inputs of amp_cast and amp_multicast to infer the dtypes of the input nodes wherever possible. I have added the support for the same in the recent commits.

PR #15120, opened 2019-06-01
[bug] fix higher grad log 

kshitij12345:
There is a bug in the implementation for higher order gradient of `log`.

https://github.com/apache/incubator-mxnet/blob/7b343d1fcde73b61322985580080333d9eee9e82/src/operator/tensor/elemwise_unary_op_basic.cc#L1077-L1079

We multiply  `gx * gx` where `gx = ograd * f'(x)`, getting `ograd^2 * f'(x)^2`, however we want only `ograd * f'(x)^2` which can be achieved in a similar fashion to the implementation of `_backward_log10/2`.

I have validated the expected results for grad on `x`.
```python
import numpy
import math

op = lambda x: nd.log(x)
grad_op = lambda x: 1/x

head_grads = nd.random.normal(0,1,(3,3))
head_grad_grads = nd.random.normal(0,1,(3,3))

x.attach_grad()
head_grads.attach_grad()


with autograd.record():
  y = nd.cos(x)
  y_grad = autograd.grad(y, x, head_grads=head_grads, create_graph=True, retain_graph=True)[0]
y_grad.backward(head_grad_grads)


expected_grad_grad_x = (grad_grad_op(x) * head_grad_grads * head_grads).asnumpy()
expected_head_grad = (grad_op(x) * head_grad_grads).asnumpy()

numpy.testing.assert_allclose(x.grad.asnumpy(), expected_grad_grad_x, rtol=1e-7, atol=1e-7)
numpy.testing.assert_allclose(head_grads.grad.asnumpy(), expected_head_grad, rtol=1e-7, atol=1e-7)
```
Which fails with current code.

Have confirmed the behaviour with Pytorch as well.
```python
import torch
import numpy
import math

op = lambda x: torch.log(x)
grad_op = lambda x: 1/x

x = torch.randn(2,3)
head_grads = torch.randn(2,3)
head_grad_grads = torch.randn(2,3)
x.requires_grad = True
head_grads.requires_grad = True

y = op(x)
y_grad = torch.autograd.grad(y, x, grad_outputs= head_grads, create_graph=True, retain_graph=True)[0]
y_grad.backward(head_grad_grads)

expected_grad_grad_x = (grad_grad_op(x) * head_grad_grads * head_grads).detach().numpy()
expected_head_grad = (grad_op(x) * head_grad_grads).detach().numpy()

numpy.testing.assert_allclose(x.grad.detach().numpy(), expected_grad_grad_x, rtol=1e-7, atol=1e-7)
numpy.testing.assert_allclose(head_grads.grad.detach().numpy(), expected_head_grad, rtol=1e-7, atol=1e-7)```

### CODE REVIEW

**Comment by kshitij12345** [comment]
@apeforest @larroy please review

File: src/operator/tensor/elemwise_unary_op_basic.cc [hunk 1]
```diff
@@ -1074,16 +1074,19 @@ MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_log,
...
-                            {gx, gx}, nullptr, &n);
+                            {gx_mul_head_grads, nnvm::NodeEntry{g_lx}}, nullptr, &n);
     auto ggx = MakeNode("negative", n->attrs.name + "_backward_grad_grad",
                         {nnvm::NodeEntry{ggx_mid}}, nullptr, &n);
 
     std::vector<nnvm::NodeEntry> ret;
 
     ret.emplace_back(MakeNode("elemwise_mul", n->attrs.name + "_backward_grad_grad",
-                             {ograds[0], gx}, nullptr, &n));
+                             {ograds[0], nnvm::NodeEntry{g_lx}}, nullptr, &n));
```
kshitij12345: I am having trouble with `head_grads.grad` which is being returned as `0's` while I expect it to be this.
Please help.

piyushghai:
Thanks for your contributions @kshitij12345 

larroy:
Hi @kshitij12345 thanks for looking into this.

I think we need to clarify what exactly we have in the first parameter of FGradient "node". We were a bit puzzled with @apeforest looking at your PR. I validated the results with the tests but I think I tried only one log, don't remember which base. But the result seemed correct to me, I guess I missed this problem.

Why do you say that node is ograd*f'(x)?     the node argument I understand is the node to calculate the gradient for, in this case we are calculating the gradient of the backward of the log. So are you saying that by chain rule, the node is ograd(of log) * d (log(x)) / dx  = ograd * reciprocal?

Would be great if we could add this to the documentation, either to the FGradient typedef or to new_op. Otherwise I always have to dig through the code to refresh this. I think is poorly documented and tricky.

src/operator/tensor/elemwise_unary_op_basic.cc
larroy: Hi. What do you mean by head_grads.grad? NodeEntry doesn't have a grad field. Could you clarify?

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -27,52 +27,79 @@ def test_log():
...
+    y_grad.backward(head_grad_grads)
+
+    # Compute expected values.
+    expected_grad_grad = grad_grad_x.asnumpy() * head_grad_grads.asnumpy() * \
+        head_grads.asnumpy()
+    expected_heads_grad = grad_x.asnumpy()
+
+    # Validate the gradients.
+    assert_almost_equal(expected_grad_grad, x.grad.asnumpy())
+    assert_almost_equal(expected_heads_grad, head_grads.grad.asnumpy())
```
larroy: Now I understand your question, i don't think anything is updating head_grads here. Why do you want to set the head gradients manually? To verify your fix?

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -1074,16 +1074,19 @@ MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_log,
   [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
     // For f(x) -> f = log
     // f''(x) = -1 * (f'(x) * f'(x))
-    auto gx = nnvm::NodeEntry{n};
+    auto gx_mul_head_grads = nnvm::NodeEntry{n};  // f'(x) * head_grads
+    auto head_grads = nnvm::NodeEntry{n->inputs[0]};
+    auto g_lx = MakeNode("reciprocal", n->attrs.name + "_backward_log_grad",
```
larroy: can we add a comment about the inputs?  it would help reason about the code.  Are the inputs 0: ograd .1: f(x) . ?

src/operator/tensor/elemwise_unary_op_basic.cc
apeforest: Still looking into this. The first output should be the gradient of y_grad. However, the `head_grads.grad` does not get the value. I suspect the returned value from this function is dropped in the gradient calculation in imperative.cc. I will look more into this. Stay tuned.

tests/python/unittest/test_higher_order_grad.py
larroy: can you try 

`y_grad_grad = autograd.grad(y_grad, x, ..., create_graph = False...)[0]`


and in validation
`assert_almost_equal(expected_heads_grad, y_grad_grad.asnumpy())`

src/operator/tensor/elemwise_unary_op_basic.cc
kshitij12345: Sorry for the confusion. I forgot to add the line from the test file. 
Sure waiting to know what you find.

src/operator/tensor/elemwise_unary_op_basic.cc
kshitij12345: Sure thing.

tests/python/unittest/test_higher_order_grad.py
kshitij12345: Yeah to verify the fix. 
I expected `y_grad.backward(head_grad_grads)` to update the `head_grads.grad` similar to the Pytorch Script from the description.

Thanks for the suggestion,
I will surely try that.

kshitij12345:
> So are you saying that by chain rule, the node is ograd(of log) * d (log(x)) / dx = ograd * reciprocal?

Following the unary_bwd

https://github.com/apache/incubator-mxnet/blob/b0be6c50ae686a59c340c56b9c11493a5c84ff72/src/operator/tensor/elemwise_unary_op.h#L400-L402

which leads here.

https://github.com/apache/incubator-mxnet/blob/2fd4720342f87be80f280176ffbecb8f37e49b73/src/operator/mxnet_op.h#L520-L538

which fuses the `output_grad * func_grad` where `func_grad` our current case is `d (log(x))/ dx` operation.

Thus the node is multiplication of `ograd * d(log(x)) / dx`, which also implies that for every second order gradient function we have to do some operation to get only `func_grad` ( used for the gradient of `ograd`).

Also I had observed some weird behaviour for functions wrapped with `ElemGradUseOut`. Will post about it later in the day.

> Would be great if we could add this to the documentation, either to the FGradient typedef or to new_op. Otherwise I always have to dig through the code to refresh this. I think is poorly documented and tricky.

Yes it is. True there should be some documentation regarding the same.

apeforest:
@kshitij12345 I have some question about the equation `expected_head_grad = (grad_op(x) * head_grad_grads).asnumpy()` in your test.

My understanding from the chain rule:

```
Given y =f(x)
dL/dx = dL/dy * dy/dx -->  this is the first forward pass. Note dL/dy as y_grad, we get dL/dx (note as x_grad)

Now we rewrite the above the equation:

input0: y_grad
input1: x
output: x_grad = y_grad * f'(x)

Another backward pass for this would be:
dL/d y_grad = dL/d x_grad * f'(x)
dL/dx = dL/d x_grad * y_grad * f''(x)

What is the meaning of dL/d y_grad? Are we treating y_grad as another input variable here?
```

Thanks for your clarification.

apeforest:
I did some more probing. I think the reason that head_grads.grad being all zeros is the variable head_grads was not specified during the second backward pass. I updated the test as follows, but got an assertion error:

```
grad_grad_op = lambda x: -nd.sin(x)
grad_op = lambda x: nd.cos(x)

x = nd.array([1, 2, 3])
head_grads = nd.ones_like(x) * 0.5
head_grads_grads = nd.ones_like(x) * 0.6

x.attach_grad()
head_grads.attach_grad()

with autograd.record():
    y = nd.sin(x)
    xgrad = autograd.grad(heads=y, variables=x, head_grads=head_grads, create_graph=True, retain_graph=True)[0]
print('xgrad = {} exp_xgrad = {}'.format(xgrad.asnumpy(), grad_op(x) * head_grads))
grad_grad_out = autograd.grad(heads=xgrad, variables=[x, head_grads], head_grads=head_grads_grads, create_graph=False, retain_graph=True)
print(len(grad_grad_out))
```

It complains that the variable head_grads is unreachable from the output which I think is reasonable. This again comes back to my question above, what is the mathematical meaning of head_grads.grad and why do we need this value?

tests/python/unittest/test_higher_order_grad.py
larroy: y_grad.backward(head_grad_grads) indicate that head_grad_grads are the head gradients, it will update all the input variables which have attached gradient, in this case head_grad_grads is not an input to the graph, so your problem that the grad doesn't get updated is expected:

https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/ndarray/ndarray.py#L2188
https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/autograd.py#L270

larroy:
@apeforest see my reply above, head_grads is not in the graph so it's not updated during backward

apeforest:
@larroy Yes, I agree with your reply. Also, I don't understand the meaning (or need) to return dL/dy_grad. @kshitij12345 Please comment. Thanks

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -27,52 +27,79 @@ def test_log():
...
+    head_grad_grads = nd.random.normal(shape=x.shape)
+    head_grads.attach_grad()
+
+    # Perform compute.
     with autograd.record():
         y = op(x)
-        y_grad = autograd.grad(y, x, create_graph=True, retain_graph=True)[0]
-    y_grad.backward()
-    assert_almost_equal(expect_grad_grad.asnumpy(), x.grad.asnumpy())
+        y_grad = autograd.grad(y, x, head_grads=head_grads,
```
apeforest: This variable is actually dL/dx, maybe rename it to `xgrad` for better readability?

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -27,52 +27,79 @@ def test_log():
...
-def check_second_order_unary(x, op, grad_grad_op):
+def check_second_order_unary(x, op, grad_op, grad_grad_op):
     x = nd.array(x)
-    expect_grad_grad = grad_grad_op(x)
+    grad_x = grad_op(x)
+    grad_grad_x = grad_grad_op(x)
     x.attach_grad()
+
+    # Manual head_grads.
+    head_grads = nd.random.normal(shape=x.shape)
```
apeforest: Rename this to `y_grad` as it is dL/dy?

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -27,52 +27,79 @@ def test_log():
...
+    head_grad_grads = nd.random.normal(shape=x.shape)
```
apeforest: I still don't understand what this variable is mathematically...

larroy:
> @kshitij12345 I have some question about the equation `expected_head_grad = (grad_op(x) * head_grad_grads).asnumpy()` in your test.
> 
> My understanding from the chain rule:
> 
> ```
> Given y =f(x)
> dL/dx = dL/dy * dy/dx -->  this is the first forward pass. Let dL/dy be y_grad, we get dL/dx (noted as x_grad)
> 
> Now we rewrite the above the equation:
> 
> input0: y_grad
> input1: x
> output: x_grad = y_grad * f'(x)
> 
> Another backward pass for this would be:
> dL/d y_grad = dL/d x_grad * f'(x)
> dL/dx = dL/d x_grad * y_grad * f''(x)
> ```
> 
> What is the meaning of dL/d y_grad? Are we treating y_grad as another input variable here?
> 
> Many thanks for your clarification.



> @kshitij12345 I have some question about the equation `expected_head_grad = (grad_op(x) * head_grad_grads).asnumpy()` in your test.
> 
> My understanding from the chain rule:
> 
> ```
> Given y =f(x)
> dL/dx = dL/dy * dy/dx -->  this is the first forward pass. Let dL/dy be y_grad, we get dL/dx (noted as x_grad)
> 
> Now we rewrite the above the equation:
> 
> input0: y_grad
> input1: x
> output: x_grad = y_grad * f'(x)
> 
> Another backward pass for this would be:
> dL/d y_grad = dL/d x_grad * f'(x)
> dL/dx = dL/d x_grad * y_grad * f''(x)
> ```
> 
> What is the meaning of dL/d y_grad? Are we treating y_grad as another input variable here?
> 
> Many thanks for your clarification.

I think the introduction of L (loss) is confusing here. As per back accumulation of gradients and chain rule we always have the incoming gradient (also called head gradient or output gradient). So the second backward pass should calculate:
equation

I'm thinking that maybe the problem is that we should not reuse the head gradient from the first gradient in the second gradient. Shouldn't the two head gradients be independent variables?

<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="350pt" height="274pt" viewBox="0.00 0.00 350.00 273.80">
<g id="graph0" class="graph" transform="scale(1 1) rotate(0) translate(4 269.7972)">
<title>G</title>
<polygon fill="#ffffff" stroke="transparent" points="-4,4 -4,-269.7972 346,-269.7972 346,4 -4,4"></polygon>
<g id="clust1" class="cluster">
<title>cluster_1</title>
<polygon fill="#f0ffff" stroke="#f0ffff" points="8,-8 8,-257.7972 78,-257.7972 78,-8 8,-8"></polygon>
<text text-anchor="middle" x="43" y="-241.1972" font-family="Times,serif" font-size="14.00" fill="#000000">Fwd</text>
</g>
<g id="clust2" class="cluster">
<title>cluster_2</title>
<polygon fill="#e0eeee" stroke="#e0eeee" points="86,-8 86,-257.7972 178,-257.7972 178,-8 86,-8"></polygon>
<text text-anchor="middle" x="132" y="-241.1972" font-family="Times,serif" font-size="14.00" fill="#000000">bwd</text>
</g>
<g id="clust3" class="cluster">
<title>cluster_3</title>
<polygon fill="#c1cdcd" stroke="#c1cdcd" points="186,-8 186,-257.7972 334,-257.7972 334,-8 186,-8"></polygon>
<text text-anchor="middle" x="260" y="-241.1972" font-family="Times,serif" font-size="14.00" fill="#000000">bwd</text>
</g>

<g id="node1" class="node">
<title>x</title>
<ellipse fill="none" stroke="#000000" cx="43" cy="-206.9972" rx="27" ry="18"></ellipse>
<text text-anchor="middle" x="43" y="-202.7972" font-family="Times,serif" font-size="14.00" fill="#000000">x</text>
</g>

<g id="node2" class="node">
<title>log</title>
<polygon fill="none" stroke="#000000" points="61,-138.4986 25,-138.4986 25,-102.4986 61,-102.4986 61,-138.4986"></polygon>
<text text-anchor="middle" x="43" y="-116.2986" font-family="Times,serif" font-size="14.00" fill="#000000">log</text>
</g>

<g id="edge1" class="edge">
<title>x-&gt;log</title>
<path fill="none" stroke="#000000" d="M43,-188.6531C43,-177.1075 43,-161.97 43,-148.8942"></path>
<polygon fill="#000000" stroke="#000000" points="46.5001,-148.5196 43,-138.5197 39.5001,-148.5197 46.5001,-148.5196"></polygon>
</g>

<g id="node5" class="node">
<title>log_bwd</title>
<polygon fill="none" stroke="#000000" points="161.4972,-152.9958 96.5028,-152.9958 96.5028,-88.0014 161.4972,-88.0014 161.4972,-152.9958"></polygon>
<text text-anchor="middle" x="129" y="-116.2986" font-family="Times,serif" font-size="14.00" fill="#000000">log_bwd</text>
</g>

<g id="edge4" class="edge">
<title>x-&gt;log_bwd</title>
<path fill="none" stroke="#000000" d="M57.9801,-191.9303C66.665,-183.195 78.051,-171.743 89.1549,-160.5747"></path>
<polygon fill="#000000" stroke="#000000" points="91.8112,-162.8672 96.3798,-153.3079 86.8471,-157.9317 91.8112,-162.8672"></polygon>
</g>

<g id="node8" class="node">
<title>log_bwd_bwd</title>
<ellipse fill="none" stroke="#000000" cx="260" cy="-120.4986" rx="66.082" ry="18"></ellipse>
<text text-anchor="middle" x="260" y="-116.2986" font-family="Times,serif" font-size="14.00" fill="#000000">log_bwd_bwd</text>
</g>

<g id="edge7" class="edge">
<title>x-&gt;log_bwd_bwd</title>
<path fill="none" stroke="#000000" d="M65.1273,-196.3801C70.6013,-193.8659 76.4799,-191.2597 82,-188.9972 125.7079,-171.0828 137.9627,-170.086 182,-152.9972 192.7894,-148.8103 204.3363,-144.1177 215.1493,-139.6286"></path>
<polygon fill="#000000" stroke="#000000" points="216.5469,-142.838 224.4251,-135.7539 213.8487,-136.3788 216.5469,-142.838"></polygon>
</g>

<g id="node3" class="node">
<title>y</title>
<ellipse fill="none" stroke="#000000" cx="43" cy="-34" rx="27" ry="18"></ellipse>
<text text-anchor="middle" x="43" y="-29.8" font-family="Times,serif" font-size="14.00" fill="#000000">y</text>
</g>

<g id="edge2" class="edge">
<title>log-&gt;y</title>
<path fill="none" stroke="#000000" d="M43,-102.1545C43,-90.6089 43,-75.4714 43,-62.3956"></path>
<polygon fill="#000000" stroke="#000000" points="46.5001,-62.021 43,-52.0211 39.5001,-62.0211 46.5001,-62.021"></polygon>
</g>

<g id="node4" class="node">
<title>ograd</title>
<ellipse fill="none" stroke="#000000" cx="129" cy="-206.9972" rx="33.0469" ry="18"></ellipse>
<text text-anchor="middle" x="129" y="-202.7972" font-family="Times,serif" font-size="14.00" fill="#000000">ograd</text>
</g>

<g id="edge3" class="edge">
<title>ograd-&gt;log_bwd</title>
<path fill="none" stroke="#000000" d="M129,-188.6531C129,-181.1585 129,-172.1505 129,-163.1753"></path>
<polygon fill="#000000" stroke="#000000" points="132.5001,-163.1383 129,-153.1384 125.5001,-163.1384 132.5001,-163.1383"></polygon>
</g>

<g id="node6" class="node">
<title>x_grad</title>
<ellipse fill="none" stroke="#000000" cx="132" cy="-34" rx="37.7044" ry="18"></ellipse>
<text text-anchor="middle" x="132" y="-29.8" font-family="Times,serif" font-size="14.00" fill="#000000">x_grad</text>
</g>

<g id="edge5" class="edge">
<title>log_bwd-&gt;x_grad</title>
<path fill="none" stroke="#000000" d="M130.1342,-87.7972C130.4249,-79.4144 130.7352,-70.4687 131.0175,-62.327"></path>
<polygon fill="#000000" stroke="#000000" points="134.5199,-62.3175 131.3687,-52.2021 127.5241,-62.0748 134.5199,-62.3175"></polygon>
</g>

<g id="node7" class="node">
<title>ograd2</title>
<ellipse fill="none" stroke="#000000" cx="260" cy="-206.9972" rx="37.7044" ry="18"></ellipse>
<text text-anchor="middle" x="260" y="-202.7972" font-family="Times,serif" font-size="14.00" fill="#000000">ograd2</text>
</g>

<g id="edge6" class="edge">
<title>ograd2-&gt;log_bwd_bwd</title>
<path fill="none" stroke="#000000" d="M260,-188.6531C260,-177.1075 260,-161.97 260,-148.8942"></path>
<polygon fill="#000000" stroke="#000000" points="263.5001,-148.5196 260,-138.5197 256.5001,-148.5197 263.5001,-148.5196"></polygon>
</g>

<g id="node9" class="node">
<title>x_grad_grad</title>
<ellipse fill="none" stroke="#000000" cx="260" cy="-34" rx="59.6781" ry="18"></ellipse>
<text text-anchor="middle" x="260" y="-29.8" font-family="Times,serif" font-size="14.00" fill="#000000">x_grad_grad</text>
</g>

<g id="edge8" class="edge">
<title>log_bwd_bwd-&gt;x_grad_grad</title>
<path fill="none" stroke="#000000" d="M260,-102.1545C260,-90.6089 260,-75.4714 260,-62.3956"></path>
<polygon fill="#000000" stroke="#000000" points="263.5001,-62.021 260,-52.0211 256.5001,-62.0211 263.5001,-62.021"></polygon>
</g>
</g>
</svg>

kshitij12345:
> @kshitij12345 I have some question about the equation `expected_head_grad = (grad_op(x) * head_grad_grads).asnumpy()` in your test.
> 
> My understanding from the chain rule:
> 
> ```
> Given y =f(x)
> dL/dx = dL/dy * dy/dx -->  this is the first backward pass. Let dL/dy be y_grad, we get dL/dx (noted as x_grad)
> 
> Now we rewrite the above the equation:
> 
> input0: y_grad
> input1: x
> output: x_grad = y_grad * f'(x)
> 
> Another backward pass for this would be:
> dL/d y_grad = dL/d x_grad * f'(x)
> dL/dx = dL/d x_grad * y_grad * f''(x)
> ```
> 
> What is the meaning of dL/d y_grad? Are we treating y_grad as another input variable here?
> 
> Many thanks for your clarification.

I was following on the basis of this graph that I had in my mind.

second_backward


Even I am not sure about the mathematical meaning of it. Also in usual scenario it would essentially be an intermediate node. Like in the picture.

However since we are returning it might as well test for it. And in case of the current test, `head_grads` as depicted by `dL/dy` is an input leaf node in the second graph. So I expect that we will be able to accumulate gradients in it (similar to Pytorch).

I hope it makes sense. Thank You.

tests/python/unittest/test_higher_order_grad.py
kshitij12345: Oh yes. Will do that.

tests/python/unittest/test_higher_order_grad.py
kshitij12345: Sure.

tests/python/unittest/test_higher_order_grad.py
kshitij12345: `head_grads` is just the input node in the graph for `x_grad`. 
`head_grad_grads` is just to check the validity of the chain rule/backprop.

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -27,52 +27,79 @@ def test_log():
...
-    assert_almost_equal(expect_grad_grad.asnumpy(), x.grad.asnumpy())
+        y_grad = autograd.grad(y, x, head_grads=head_grads,
+                               create_graph=True, retain_graph=True)[0]
+
+    y_grad.backward(head_grad_grads)
+
+    # Compute expected values.
+    expected_grad_grad = grad_grad_x.asnumpy() * head_grad_grads.asnumpy() * \
+        head_grads.asnumpy()
+    expected_heads_grad = grad_x.asnumpy()
```
kshitij12345: Should be `grad_x.asnumpy() * head_grad_grads.asnumpy()`

apeforest:
@kshitij12345 The computation graph for the second backward pass makes sense to me. As you can see there is only one output from the graph, that is x_grad_grad. It is not clear to me where the output `dL/dy_grad` is

tests/python/unittest/test_higher_order_grad.py
kshitij12345: We are checking gradients for `head_grads` (not `head_grad_grads`), which is used to compute `x_grad`, so I believe we should accumulate some gradient in `head_grads`

kshitij12345:
> @kshitij12345 The computation graph for the second backward pass makes sense to me. As you can see there is only one output from the graph, that is x_grad_grad. It is not clear to me where the output `dL/dy_grad` is

Actually I haven't shown the other graph for `ograd_grad`. If you take a look at the `x_grad`, it is computed using `ograd` and `x`. So `x_grad` will change not only with `x` but also with `ograd`. 

If you focus only on `*` (multiplication node) while computing `x_grad`.

d2f29aff-35f8-491e-ab2a-81646f825224

apeforest:
> ```python
> x_grad = x_grad_mid * y_grad # Note this part.
> ```

I don't think this part is correct.

Note: your x_grad_mid is dL/dx.
So x_grad_mid = y_grad * grad_op(x)

Do you meant to write x_grad = grad_op(x) * y_grad?

apeforest:
@kshitij12345 I think it's because of the design of backward computation graph in MXNet. In C++ implementation, when you specify variables=x, it will compute gradients for the input variables.

As in your case 2:
```
x_grad = autograd.grad(heads=y, variables=x, head_grads=y_grad, create_graph=True, retain_graph=True)[0]
```
If you perform another backward on x_grad as `x_grad.backward(out_grad=head_grads_grads)`, y_grad is not listed as input variable and therefore it's gradient is zero

As in your case 1:
```
x_grad = x_grad_mid * y_grad # Note
x_grad.backward(out_grad=head_grad_grads)
```
You implicitly made y_grad an input variable when calling backward on x_grad. And that is why you will get values in y_grad.grad.

I replaced the `backward()` method with an explicit `autograd.grad()` call, which should call the same C++ backend function and result is different.

case 1.1: if I do the following, I again don't get any values for y_grad because the output only contains one gradient variable
```
out_grad = autograd.grad(heads=x_grad, variables=x, head_grads=head_grad_grads, create_graph=False, retain_graph=False)
print(out_grad[0])   # values equals to expected_grad_grad
```

case 1.2: I explicitly set y_grad as input variable, I then get the expected result as in your case 1
```
out_grad = autograd.grad(heads=x_grad, variables=[x, y_grad], head_grads=head_grad_grads, create_graph=False, retain_graph=False)
print(out_grad[0])   # value equals to expected_grad_grad
print(out_grad[1])   # value equals to expected_heads_grad
```

At this point, I am not sure if this is a bug because the backward API is designed differently from PyTorch. If y_grad is not specified as part of the input variables that need to perform gradient on, it will not get values assigned even if you write `y_grad.attach_grad()` to it. This seems to be consistent from API spec. Also, given that the value `y_grad` does not have real useful values, I also don't feel the necessity to store it. Please let me know if this makes sense. Thanks a lot for your careful drawing and insightful discussion.

apeforest:
As a follow up, I just dumped out the computation graph in case 2. Indeed, the node that used to calculate y_grad.grad is not even in the final symbolic graph because there is no input dependency.

kshitij12345:
Oh thank you very much for explaining what is happening. I get it now. 
It makes sense, as passing argument as `head_grads` doesn't make any explicit computation on the graph.

`y_grad` may not have any useful meaning as of now, however I believe that we should test for the value using case 1.2, just to verify and catch any mistakes there. Rather than the same thing biting us in the future. Your thoughts? @larroy @apeforest 

Also it would be really great if you can tell me how to get the dump of the graph that would be really helpful.

Thank you again for your time and efforts.

tests/python/unittest/test_higher_order_grad.py
larroy: the default behaviour in python vs mxnet is different with respect accumulation of gradients (pytorch: add) mxnet: write. Having said that, I still don't understand why do you expect gradient accumulation in head_grads.

larroy:
@apeforest can you paste your dump of the graph? and share where did you dump it? I was working on a utility to dump the graph.

apeforest:
> Also it would be really great if you can tell me how to get the dump of the graph that would be really helpful.

@kshitij12345 There is a utility function `inline void LogMemoryPlan(const nnvm::Graph& g)` in exec_utils.h that can dump out the computation graph. However it only works in symbol mode currently and you need to manually copy paste it to imperative.cc to dump out graph in NDArray APIs. Besides, you also need to comment out the `const auto& vstorage = g.GetAttr<nnvm::StorageVector>("storage_id")` line otherwise you will get a core dump.

In one word, it's not an easy to use debugging utility and @larroy is working to improve it and create a PR soon. Stay tuned :)

apeforest:
> `y_grad` may not have any useful meaning as of now, however I believe that we should test for the value using case 1.2, just to verify and catch any mistakes there. Rather than the same thing biting us in the future. Your thoughts? @larroy @apeforest

I don't have a clear answer to how to test this value now. Definitely not through a python unit test because the value is not getting output as NDArray. Besides, as I commented earlier, that computation node was even in nnvm graph maybe it was later deleted by graph optimization because it is a dangling node with no output.

tests/python/unittest/test_higher_order_grad.py
kshitij12345: Oh. I just expected it to have gradients ( by accumulation or writing ), as it is / its value is used while computing the `x_grad`. But from your and @apeforest's explanation, I kinda understand the behaviour better. 
Thank You for digging in and explaining.

kshitij12345:
@apeforest , Thank You for explaining how to get the dump of the graph. Waiting the PR which simplifies that.

apeforest:
> @apeforest , Thank You for explaining how to get the dump of the graph. Waiting the PR which simplifies that.
> 
> Also could you tell me if it is possible to print the Node and it's value (for the array) via C++ and if yes then how .

We don't yet have a utility function to do that. What do you mean the value of the node? A node only represent a computation in the graph. The values are passed in/out when the graph is traversed and each node invokes the FCompute/FGradient functions registered. I think we can only track values in the corresponding NDArray outputs. Please let me know if I misunderstand your question.

kshitij12345:
Oops, sorry for confusion. I meant the value of a NDArray.
`auto gx = nnvm::NodeEntry{n};`, can I get/print the value of the NDArray corresponding to that node.

Thank You.

larroy:
Right now is very difficult to print the value of the NDArray. I will work on a utility to dump the graph this week, first the graph itself, and if possible the NDArray values.

apeforest:
There is no available utility to print out value from NDArray. I used to write a for loop to iterate the `dptr` pointer and dump out values. Besides, we need extra synchronization handling if the NDArray is stored in GPU. I agree with @larroy that this is something very important for debugging MXNet backend. I have created an issue https://github.com/apache/incubator-mxnet/issues/15198 to ask for contributions.

tests/python/unittest/test_higher_order_grad.py
apeforest: Thanks for the clarification.

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -66,13 +66,26 @@ def grad_grad_op(x):
...
+    y_grad = nd.random.normal(shape=x.shape)
+    head_grad_grads = nd.random.normal(shape=x.shape)
+
+    # Perform compute.
     with autograd.record():
         y = op(x)
-        y_grad = autograd.grad(y, x, create_graph=True, retain_graph=True)[0]
-    y_grad.backward()
-    assert_almost_equal(expect_grad_grad.asnumpy(), x.grad.asnumpy())
+        x_grad = autograd.grad(y, x, head_grads=y_grad,
```
apeforest: Can you explicitly specify the argument as 

```suggestion
        x_grad = autograd.grad(heads=y, variables=x, head_grads=y_grad, create_graph=True, retain_graph=True)
```

I think this will make it easier to understand.

apeforest:
@kshitij12345 Could you please rebase and retrigger CI again? Thanks!

apeforest:
@kshitij12345 still one GPU test failed. I looked at the log and don't find it related to your change. Could you please rebase and trigger CI one more time. The master branch was broken last weekend. Thanks again.

kshitij12345:
@apeforest Note that the pending job has succeeded but for some reason it isn't updated here.

apeforest:
@kshitij12345 There was some issue with CI recently. Could you please re-trigger it one more time? Sorry for the inconvenience.

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -1090,68 +1090,84 @@ MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_log,
...
 
 MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_log2,
                                                   unary_bwd<mshadow_op::log2_grad>)
 .set_attr<nnvm::FGradient>("FGradient",
   [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
-    // For f(x) -> f = log2
+    // ograds[0]: dL/dxgrad
+    // inputs[0]: dL/dy
+    // inputs[1]: x
+    // f(x) = y = log10(x)
```
apeforest:
```suggestion
    // f(x) = y = log2(x)
```

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -1090,68 +1090,84 @@ MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_log,
...
     // f'(x) = 1 / (log(10) * x)
     // f''(x) = -1 * (f'(x) * 1/x)
-    auto gx = nnvm::NodeEntry{n, 0, 0};
-    auto g_lx = MakeNode("reciprocal", n->attrs.name + "_backward_log_grad",
+    auto dydx_mul_dldy = nnvm::NodeEntry{n};  // f'(x) * head_grads
+    auto dydx = MakeNode("elemwise_div", n->attrs.name + "_dydx",
```
larroy: Doesn't elemwise_div require two inputs?

kshitij12345: Oh, it does. However I guess, this thing is skipped in the test computation graph and hence we don't see the error.

Will fix it. However we should somehow find a way to test for the same.

PR #15121, opened 2019-06-01
[clojure] clojurify function names in image.clj namespace

Chouffe:
## Description ##

The `image` namespace is using some functions that are not named the `clojure way`. It is an attempt to clojurify these functions.
The namespaces are supposed to be imported qualified and having the suffix `-image` in the following functions is redundant
* `resize-image`
* `decode-image`
* `read-image`

One should be able to call `image/resize`, `image/decode`, `image/read` instead.

The function `to-image` is misleading as we dont know what is turned into an image. I suggest to name it `ndarray->image` instead.

* `deprecated` is added in metadata
* `deprecated` is added in docstrings

### DIFF
--- a/contrib/clojure-package/src/org/apache/clojure_mxnet/image.clj
+++ b/contrib/clojure-package/src/org/apache/clojure_mxnet/image.clj
@@ -63,10 +63,20 @@
   ([input-stream]
    (decode-image input-stream {})))
 
+(s/def ::color #{:grayscale :color})
+(s/def ::decode-image-opts-2 (s/keys :opt-un [::color ::to-rgb ::output]))
+
+(defn- color->int
+  [color-flag]
+  (case color-flag
+    :grayscale 0
+    :color 1))
+
 (defn decode
   "Decodes an image from an input stream with OpenCV.
     `input-stream`: `InputStream` - Contains the binary encoded image
-    `color-flag`: 0 or 1 - Convert decoded image to grayscale (0) or color (1)
+    `color`: keyword in `#{:color :grayscale}` - Convert decoded image to
+             grayscale or color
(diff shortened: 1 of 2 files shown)


contrib/clojure-package/src/org/apache/clojure_mxnet/image.clj
```diff
@@ -60,14 +63,38 @@
   ([input-stream]
    (decode-image input-stream {})))
 
+(defn decode
+  "Decodes an image from an input stream with OpenCV.
+    `input-stream`: `InputStream` - Contains the binary encoded image
+    `color-flag`: 0 or 1 - Convert decoded image to grayscale (0) or color (1)
```
kedarbellare: one thing i'm wondering is whether `color-flag` should be specified as `:grayscale` or `:color` and we handle the conversion to 0, 1 in this function?

kedarbellare:
@Chouffe : you need to fix some conflicts because i merged your other PR

Chouffe:
> lgtm! let me merge this.
> 
> as an aside, are there other places in examples where `*-image` functions are used? if yes, can you replace those usages as well?

Will do! Thanks for reviewing so promptly @kedarbellare

contrib/clojure-package/src/org/apache/clojure_mxnet/image.clj
Chouffe: Very good point! Let me fix this then.

Chouffe: I had to use another name for the keyword `:color` instead of `:color-flag`. I am open to better name for that. The problem is that it is not possible to have two different specs with the same name.

Chouffe:
Should be ready to :ship: now @kedarbellare

PR #15124, opened 2019-06-01
[MXNET-1294] Priority-based parameter propagation for improved data parallel training throughput

anandj91:
## Description ##
https://issues.apache.org/jira/browse/MXNET-1294

During data parallel training , the parameters of each layer is synchronized through parameter server after slicing them into smaller slices and synchronizing them independent to each other based on their priority. Currently priority of a layer is set as its negative index.

For more details, refer https://anandj.in/wp-content/uploads/sysml.pdf.

### Essentials ###

### Changes ###

## Comments ##


piyushghai:
Thanks for your contribution @anandj91. Can you look into the CI failures ? 
@mxnet-label-bot Add [pr-awaiting-review, Backend].

anandj91:
Is there a way to rerun the test cases without doing a git push?

piyushghai:
> Is there a way to rerun the test cases without doing a git push?

You could push an empty commit. 
```git commit -m "Empty commit" --allow-empty```

File: src/kvstore/kvstore_dist.h [hunk 1]
```diff
@@ -55,7 +55,7 @@ class KVStoreDist : public KVStoreLocal {
           ps::kWorkerGroup + ps::kServerGroup + ps::kScheduler);
       }
     }
-    bigarray_bound_ = dmlc::GetEnv("MXNET_KVSTORE_BIGARRAY_BOUND", 1000 * 1000);
+    bigarray_bound_ = dmlc::GetEnv("MXNET_KVSTORE_BIGARRAY_BOUND", 40 * 1000);
```
eric-haibin-lin: Changing this also affects the Encoding logic for row sparse arrays, and compressed arrays.

src/kvstore/kvstore_dist.h
```diff
@@ -544,31 +592,25 @@ class KVStoreDist : public KVStoreLocal {
...
-        // send it to a single random picked server
-        int server = (key * 9973) % num_servers;
-        ps::Key ps_key = krs[server].begin() + key;
-        CHECK_LT(ps_key, krs[server].end());
+      /**
+       * Round-Robin key assignment
+       */
+      int64_t params = pskv_size;
+      int64_t slice_bound = bigarray_bound_ * num_bytes;
+      static ps::Key server = 0;
```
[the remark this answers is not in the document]
anandj91: the server variable is static. so it is incremented on every invocation of EncodeDefaultKey function. this does not cause overloading server 0.

src/kvstore/kvstore_dist.h
```diff
@@ -412,17 +439,38 @@ class KVStoreDist : public KVStoreLocal {
...
-              cmd, [cb]() { cb(); });
+
+          auto *counter = new std::atomic<int>(pskv.keys.size());
+          int len = 0;
+          for (size_t i = 0; i < pskv.keys.size(); i++) {
+            auto ks = new ps::SArray<ps::Key>(std::move(pskv.keys.segment(i, i+1)));
+            auto vs = new ps::SArray<char>(std::move(vals.segment(len, len+pskv.lens[i])));
+            auto ls = new ps::SArray<int>(std::move(pskv.lens.segment(i, i+1)));
+            CHECK_NOTNULL(ps_worker_)->ZPush(*ks, *vs, *ls,
+                    static_cast<int>(RequestType::kDefaultPushPull),
```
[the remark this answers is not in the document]
anandj91: I was also concerned about the overhead of introducing many NDArrays. for models like VGG, this could go as high as hundreds of thousand with the current slice size. Is there a way to define dependency between operations without creating ndarrays?

vandanavk:
@pinaraws @anirudhacharya for review

anandj91:
Modified the code to address the review comments. Sorry for the delay.
I have added a new flag in the config.mk to enable priority based update (USE_PRIORITY_UPDATE). the flag is disabled by default. @eric-haibin-lin Can you please tell me on how to add a new test case to build the code with this flag turned on and run the dist_kvstore test cases?

roywei:
@anandj91 gentle ping, any update on this PR?

anandj91:
@roywei The current implementation uses multiple `ThreadVar` to specify dependency between push and pull between slices. After some benchmarking on large models like VGG-19, I found that this causes large overhead and the training performance reduce to 50%.

Instead I'm planning to introduce a new API for pushpull which combines push and pull of one slice. I had an offline discussion with @eric-haibin-lin and he is fine with this approach.

anandj91:
This PR is waiting on https://github.com/apache/incubator-mxnet/pull/15559 for the PushPull API in KVStore.

anandj91:
I'm facing some design level challenges to properly implement Priority based update (P3) on top of PushPull API. MXNet does a simple load balancing before pushing or pulling key-values by splitting NDArrays equally to the parameter servers. P3 requires a round-robin style parameter distribution which means slicing a large NDArray into thousands of smaller ones. Much more granular than current default distribution strategy and each PS would get more than one slice.

With the way mxnet and ps-lite designed right now, ps-lite assumes a single ZPush/ZPull/ZPushPull belongs to a single layer/NDArray. It also assumes that one slice only belong to one PS. These assumption need to be broken for implementing P3. What I have done right now is to add round-robin (RR) distribution strategy along with the default one and use a boolean flag to switch between these two. When user chooses to use RR, KVStore consider each slice as separate key-value pair. Otherwise fallback to the default mode.

src/kvstore/kvstore_dist.h
```diff
@@ -188,11 +191,19 @@ class KVStoreDist : public KVStoreLocal {
   void InitImpl(const std::vector<int>& keys,
                 const std::vector<NDArray>& values) override {
     CheckUnique(keys);
+
     for (size_t i = 0; i < keys.size(); ++i) {
       comm_->Init(keys[i], values[i].storage_type(), values[i].shape(), values[i].dtype());
+      if (slice_threshold_ > 0) {
+        // Initialize the slices as this is the only function that is assured to
+        // be called in the same order in all the worker machines
+        EncodeDefaultKey(keys[i], values[i].shape().Size(),
```
eric-haibin-lin: This assumes the storage type is kDefaultStorage?

python/mxnet/kvstore/base.py
```diff
@@ -452,4 +457,13 @@ def create(name='local'):
         from .kvstore import KVStore
         kv = KVStore(handle)
         set_kvstore_handle(kv.handle)
-        return kv
+
```
eric-haibin-lin: This means that the server node must have access to the training script, being able to run to line of code that calls `kv.create()`. In comparison, previously in theory users just need to install mxnet on the server nodes, without worrying about other dependencies required for running the training script.  

I wonder if we can still start the server process when importing mxnet. We can add an env var `DMLC_PS_TYPE` which is controlled by `launch.py` and we add an extra flag `--p3` to launch.py. So users need to do `launch.py --p3 -n 2 train.py`. `DMLC_PS_TYPE` is then read when starting the server process. 
We can also check the value of `DMLC_PS_TYPE` at the worker side for consistency when a p3 store is created by the `kv.create()`. What do you think?

python/mxnet/kvstore/kvstore.py
```diff
@@ -104,8 +104,22 @@ def broadcast(self, key, value, out, priority=0):
...
+
+        if use_str_keys:
+            check_call(_LIB.MXKVStoreBroadcastEx(
+                self.handle, mx_uint(len(cvkeys)), cvkeys, mx_uint(len(cokeys)), cokeys,
+                cvals, couts, ctypes.c_int(priority)))
+        else:
+            check_call(_LIB.MXKVStoreBroadcast(
+                self.handle, mx_uint(len(cvkeys)), cvkeys, mx_uint(len(cokeys)), cokeys,
+                cvals, couts, ctypes.c_int(priority)))
+
```
eric-haibin-lin: Shall we have p3store.py that override `is_capable`, `set_optimizer`, etc?

python/mxnet/kvstore/kvstore.py
```diff
@@ -104,8 +104,22 @@ def broadcast(self, key, value, out, priority=0):
...
+        else:
+            cokeys = cvkeys
+            couts = cvals
+
+        if use_str_keys:
+            check_call(_LIB.MXKVStoreBroadcastEx(
+                self.handle, mx_uint(len(cvkeys)), cvkeys, mx_uint(len(cokeys)), cokeys,
+                cvals, couts, ctypes.c_int(priority)))
+        else:
+            check_call(_LIB.MXKVStoreBroadcast(
```
eric-haibin-lin: why do we need `cokeys`?

python/mxnet/kvstore/kvstore.py
eric-haibin-lin: Shall we also add tests to https://github.com/apache/incubator-mxnet/blob/d9291e5ae0b3f757e781bf457bdff8b6a61b8c8d/ci/docker/runtime_functions.sh#L1373

include/mxnet/c_api.h
```diff
@@ -2714,6 +2714,48 @@ MXNET_DLL int MXKVStorePullRowSparseEx(KVStoreHandle handle,
...
+ * \param vals the list of values
+ * \param outs the list of outputs
+ * \param priority the priority of the action
+ * \return 0 when success, -1 when failure happens
+ */
+MXNET_DLL int MXKVStoreBroadcastEx(KVStoreHandle handle,
+                                   mx_uint vnum,
+                                   const char** vkeys,
+                                   mx_uint onum,
+                                   const char** okeys,
```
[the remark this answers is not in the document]
anandj91: I was keeping the C++ Broadcast api similar to the PushPull api. But I guess it is redundant to have separate keys even for pushpull. So I have removed from both APIs.

python/mxnet/kvstore/base.py
anandj91: in that case, ps-lite already require DMLC_PS_VAN_TYPE env variable to be set to use P3Van. Can we use the same env variable for MXNet as well? And we don't need to introduce new --kvstore arguments (dist_p3). In the backend we can check if DMLC_PS_VAN_TYPE is set to p3 and launch P3StoreDist instead of KVStoreDist.
In order to make things easier, as you said, we can add another flag in launch.py --p3. Otherwise users have to set both --p3 and --kvstore dist_p3. Seems redundant.

python/mxnet/kvstore/kvstore.py
anandj91: the register(klass) function uses klass.__name__ as the name of the kvstore. So I guess we need to register multiple classes for dist_p3, dist_device_p3. On the other hand, if we are going with my suggestion in the previous comment, we won't even be having a name for the new kvstore.

I'll get back to this once we finalize what to do with the previous issue.

anandj91: I'll write the test cases, once we finalize the design. otherwise I probably have to rewrite the tests  multiple times. I hope that is okay.

python/mxnet/kvstore/base.py
eric-haibin-lin: Right, i meant `DMLC_PS_VAN_TYPE` as the env var. 

Currently we have a few different communication backends: native, byteps, horovod, p3. When changing the backend, users have to do one of the followings:
- `launch.py`  + `gluon.Trainer(kv='dist')` for native
- `launch.py --horovod`  + `gluon.Trainer(kv='horovod')`
- `launch.py --byteps`  + `gluon.Trainer(kv='byteps')`
- `launch.py --p3`  + `gluon.Trainer(kv='dist_p3')`

For users it should be fairly easy to follow. What do you think?

python/mxnet/kvstore/kvstore.py
eric-haibin-lin: What about adding a name flag to `register`? I had a quick test:
```
import mxnet as mx
import warnings
from mxnet.kvstore import KVStoreBase

@staticmethod
def register(klass, name=None):
    assert(isinstance(klass, type))
    if name is None:
        name = klass.__name__.lower()
    else:
       name = name.lower()
    if name in KVStoreBase.kv_registry:
        warnings.warn('WARNING: New kvstore %s.%s is overriding '
                      'existing kvstore %s.%s' %
                      (klass.__module__, klass.__name__,
                       KVStoreBase.kv_registry[name].__module__,
                       KVStoreBase.kv_registry[name].__name__))
    KVStoreBase.kv_registry[name] = klass
    return klass

mx.kvstore.KVStoreBase.register = register
class MyClass:
    pass

# manually call register multiple times 
mx.kvstore.KVStoreBase.register(MyClass, name='class_dist')
mx.kvstore.KVStoreBase.register(MyClass, name='class_dist_device')

print(KVStoreBase.kv_registry)
```

returns 
```
{'teststore': <class 'mxnet.kvstore.base.TestStore'>, 
'class_dist': <class '__main__.MyClass'>, 
'class_dist_device': <class '__main__.MyClass'>}
```

python/mxnet/kvstore/base.py
anandj91: This sounds good

python/mxnet/kvstore/kvstore.py
anandj91: if a user want to directly use kvstore class, then probably have to use dist_p3 to instantiate kvstore python class and also set DMLC_PS_VAN_TYPE env var. seems a bit redundant to me.

I feel that we don't need separate python class for p3 except for overriding is_capable(). But I think if we are planning to use DMLC_PS_VAN_TYPE in the backend anyway, we probably can raise exception in the backend if someone tries to set optimizer or updater in kvstore. So we don't need any changes in the python frontend at all.

eric-haibin-lin: Sounds good. We don't need to have separate class then. We are already raising exceptions in the backend when setting optimizer for P3. What about also changing `is_capable` in KVStore to return the value based on `DMLC_PS_VAN_TYPE`?

anandj91: yes. sounds good. I'll update the code accordingly

python/mxnet/kvstore/kvstore.py
```diff
@@ -67,6 +68,7 @@ def __init__(self, handle):
         self._updater = None
         self._updater_func = None
         self._str_updater_func = None
+        self._is_p3 = (os.getenv('DMLC_PS_VAN_TYPE', '') == 'p3')
```
eric-haibin-lin: will u also change launch.py?

python/mxnet/kvstore/base.py
anandj91: looking at horovod and byteps, it doesn't seem like it is as easy as passing kv= argument into gluon.Trainer to use horovod and byteps. Using horovod and byteps require some code changes as well. So I'm not sure if it is possible to run launch.py with --horovod/--byteps flag and make them use corresponding communication mechanisms. It would work for kvstore and p3.

python/mxnet/kvstore/kvstore.py
anandj91: sure. I just wanted to confirm if we are planning to only add --p3 to launch.py now and not --horovod or --byteps because of the reasons I mentioned in the previous comment

python/mxnet/kvstore/base.py
eric-haibin-lin: @apeforest is working on the horovod changes, while @chaokunchang is working on bytePS changes. For this PR we can just make the changes required for p3

python/mxnet/model.py
```diff
@@ -69,15 +69,14 @@ def _create_sparse_kvstore(kvstore):
...
     # always update on kvstore
-    update_on_kvstore = True
     if isinstance(kvstore, kvs.KVStore):
         kv = kvstore
     elif isinstance(kvstore, str):
         kv = kvs.create(kvstore)
     else:
         raise TypeError("Cannot create '%s' KVStore with row_sparse parameters. "
                         "The type must be KVStore or str." % kvstore)
-    return (kv, update_on_kvstore)
```
eric-haibin-lin: row_sparse weight requires `update_on_kvstore=True`.  The weights are stored in the server, and the worker only pull a subset of the weight in each iteration for mini-batch training. 
Could you instead add an assertion here `assert kv.is_capable(kvs.KVStoreBase.OPTIMIZER) is True`?

src/kvstore/kvstore_dist.h
```diff
@@ -180,16 +183,22 @@ class KVStoreDist : public KVStoreLocal {
...
+  static int GetNewCustomerId() {
+    return customer_id_++;
+  }
 
   void InitImpl(const std::vector<int>& keys,
                 const std::vector<NDArray>& values) override {
     CheckUnique(keys);
     for (size_t i = 0; i < keys.size(); ++i) {
       comm_->Init(keys[i], values[i].storage_type(), values[i].shape(), values[i].dtype());
+      /* Fix the key encoding */
```
eric-haibin-lin: Does this assume storage_type = default?

python/mxnet/kvstore/kvstore.py
```diff
@@ -349,6 +363,7 @@ def pushpull(self, key, value, out=None, priority=0):
 
         out: NDArray or list of NDArray or list of list of NDArray
             Values corresponding to the keys.
+            `out` should have length as `value`.
```
eric-haibin-lin: Why is there such a restriction?

python/mxnet/model.py
```diff
@@ -69,15 +69,15 @@ def _create_sparse_kvstore(kvstore):
...
+    assert kv.is_capable(kvs.KVStoreBase.OPTIMIZER)
```
eric-haibin-lin: Could you also add an error message: 
"KVStore with sparse weight requires optimizer support. However, type(kv) does not support optimizer. Please consider other kvstore backends (e.g. dist_device) instead."

python/mxnet/kvstore/kvstore.py
anandj91: there is grouping of keys and corresponding ndarrays happening at the backend based on their positions in the list before performing kv operations. Since we are using same key list for both value and out, their layouts should be identical. length check is just for a sanity.

this is the reason why I had both vkeys and okeys separate in the previous implementation.

eric-haibin-lin: You're right. However, if there's one key, having value = a list of 4 arrays, and out = a list of one array is also valid.

anandj91: in that case we need to bring back separate key list for value and out.

python/mxnet/kvstore/kvstore.py
apeforest: Could you clarify a bit more on this? According to this RFC (https://github.com/apache/incubator-mxnet/issues/16795), p3 will be treated as another kvstore just like byteps/horovod. Why would it be different?

eric-haibin-lin: @apeforest because the source code lives inside mxnet, it supports both device and local mode, and can share the same python KVStore class

anandj91: @apeforest We do have a separate class for P3 in the backend though.

eric-haibin-lin:
@anandj91 great work!

Issue #15152, opened 2019-06-04
[CI][nightly] nightly test tutorial failure: test_tutorials.test_python_kvstore


roywei:
FAIL: test_tutorials.test_python_kvstore

http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/NightlyTestsForBinaries/detail/master/335/pipeline/147

roywei:
This tutorial test was passing when running  on 1 GPU machine.
https://github.com/apache/incubator-mxnet/blob/master/docs/tutorials/python/kvstore.md
```
# The numbers used below assume 4 GPUs
gpus = mx.context.num_gpus()
if gpus > 0:
    contexts = [mx.gpu(i) for i in range(gpus)]
else:
    contexts = [mx.cpu(i) for i in range(4)]
```
However, when I changed to P3 instances with 4 gpus in https://github.com/apache/incubator-mxnet/pull/15141. it fails.
```


MXNetError: [01:12:52] src/imperative/./imperative_utils.h:71: Check failed: inputs[i]->ctx().dev_mask() == ctx.dev_mask() (1 vs. 2) : Operator broadcast_add require all inputs live on the same context. But the first argument is on gpu(0) while the 2-th argument is on cpu(0)

Stack trace:

  [bt] (0) /work/mxnet/python/mxnet/../../lib/libmxnet.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x3c) [0x7f08e7052c1c]

  [bt] (1) /work/mxnet/python/mxnet/../../lib/libmxnet.so(mxnet::imperative::GetContext(nnvm::NodeAttrs const&, std::vector<mxnet::NDArray*, std::allocator<mxnet::NDArray*> > const&, std::vector<mxnet::NDArray*, std::allocator<mxnet::NDArray*> > const&, mxnet::Context const&)+0x823) [0x7f08e9fbf343]

  [bt] (2) /work/mxnet/python/mxnet/../../lib/libmxnet.so(mxnet::Imperative::Invoke(mxnet::Context const&, nnvm::NodeAttrs const&, std::vector<mxnet::NDArray*, std::allocator<mxnet::NDArray*> > const&, std::vector<mxnet::NDArray*, std::allocator<mxnet::NDArray*> > const&)+0xdb) [0x7f08e9fcd47b]

[...]

PR #15156, opened 2019-06-05
fix kvstore nightly failure

roywei:
## Description ##
fix https://github.com/apache/incubator-mxnet/issues/15152

**In summary:**
root cause of failure is num of GPUs changed from 2 to 1. NODE_LINUX_GPU is G3.8x wiht 2 GPUs and NODE_LINUX_GPU_P3 is P3.2x with 1 GPU

so

contexts =[mx.gpu(0), mx.gpu(1)]  ->  [mx.gpu(0)] 
b length changed from 2 to 1
b = [mx.nd.ones(shape, ctx) for ctx in contexts]

It seems in kvstore, when pushing a list, only len list lenght >1 , aggregation happens, and everything will be on the same context during update. But when lenght = 1, the sum/aggregation won't happen, causing update with ndarray on different context failed. User should make sure ndarrays are on the same device.

more details and reproduciable code at https://github.com/apache/incubator-mxnet/issues/15152#issuecomment-499193612

### DIFF
--- a/docs/tutorials/python/kvstore.md
+++ b/docs/tutorials/python/kvstore.md
@@ -54,7 +54,7 @@ print(a.asnumpy())
 The data for pushing can be stored on any device. Furthermore, you can push multiple
 values into the same key, where KVStore will first sum all of these
 values and then push the aggregated value. Here we will just demonstrate pushing a list of values on CPU.
-Please note when the sum will only happen if length of the value list is larger than 1.
+Please note summation only happens if the value list is longer than one
 
 ```python
 contexts = [mx.cpu(i) for i in range(4)]
--- a/docs/tutorials/python/kvstore.md
+++ b/docs/tutorials/python/kvstore.md
@@ -75,9 +75,9 @@ For each push, KVStore combines the pushed value with the value stored using an
 control how data is merged:
 
 ```python
-def update(key, input, stored):
+def update(key, value, stored):
     print("update on key: %d" % key)
-    stored += input * 2
+    stored += value * 2
 kv._set_updater(update)
 kv.pull(3, out=a)
 print(a.asnumpy())
@@ -86,7 +86,7 @@ print(a.asnumpy())
 `[[ 4.  4.  4.],[ 4.  4.  4.]]`<!--notebook-skip-line-->
 
 ```python
-kv.push(3, mx.nd.ones(shape))
+kv.push(3, mx.nd.ones(shape, mx.gpu(0)))
 #
 kv.pull(3, out=a)
 print(a.asnumpy())
--- a/docs/tutorials/python/kvstore.md
+++ b/docs/tutorials/python/kvstore.md
@@ -53,15 +53,11 @@ print(a.asnumpy())
 
 The data for pushing can be stored on any device. Furthermore, you can push multiple
 values into the same key, where KVStore will first sum all of these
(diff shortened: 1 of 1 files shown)


roywei:
root cause is GPU changed from 2 to 1. I mistook the instance type used in CI. see updates in https://github.com/apache/incubator-mxnet/issues/15152

piyushghai:
@mxnet-label-bot Add [pr-awaiting-review, KVStore]

docs/tutorials/python/kvstore.md
```diff
@@ -86,7 +86,7 @@ print(a.asnumpy())
 `[[ 4.  4.  4.],[ 4.  4.  4.]]`<!--notebook-skip-line-->
 
 ```python
-kv.push(3, mx.nd.ones(shape))
+kv.push(3, mx.nd.ones(shape, contexts[0]))
```
apeforest: If there are multiple GPUs, will this change instruct user only storing NDArray to the first one?

apeforest: Since this is an user guide page, shouldn't we be more generic to have a simple user experience? I feel it might be better to fix this in nightly pipeline than updating user guide.

eric-haibin-lin: I think we should make the tutorial simpler, just demonstrating pushing a list of cpu arrays to kvstore and pull the result.

roywei: @apeforest had another discussion with @eric-haibin-lin , making it to a list of 4 CPU arrays will provide consistent result. If users has 1/2/3 GPUs, it will produce diffrent results compared to the value on tutorial, and may cause confusion. Also if only 1 GPU, it defeats the purpose of the tutorial.
I have added a note saying sum will only happen if value list is larger than 1.

roywei:
@apeforest any more concerns?
I have tested the following nightly job passed.
```
ci/build.py --docker-registry mxnetci --nvidiadocker --platform ubuntu_nightly_gpu --docker-build-retries 3 --shm-size 1500m /work/runtime_functions.sh nightly_tutorial_test_ubuntu_python2_gpu
```
log
```
Ran 48 tests in 1725.072s

OK
build.py: 2019-06-06 00:18:09,787Z INFO Waiting for status of container 2af8c3bb1fdb for 600 s$
build.py: 2019-06-06 00:18:09,950Z INFO Container exit status: {'Error': None, 'StatusCode': 0$
build.py: 2019-06-06 00:18:09,950Z INFO Container exited with success 👍
build.py: 2019-06-06 00:18:09,950Z INFO Stopping container: 2af8c3bb1fdb
build.py: 2019-06-06 00:18:09,952Z INFO Removing container: 2af8c3bb1fdb
```

docs/tutorials/python/kvstore.md
```diff
@@ -53,15 +53,11 @@ print(a.asnumpy())
 
 The data for pushing can be stored on any device. Furthermore, you can push multiple
 values into the same key, where KVStore will first sum all of these
-values and then push the aggregated value:
+values and then push the aggregated value. Here we will just demonstrate pushing a list of values on CPU.
+Please note when the sum will only happen if length of the value list is larger than 1.
```
apeforest: This sentence is not grammatically correct.

PR #15167, opened 2019-06-06
Pointwise fusion for GPU

ptrendx:
## Description ##
This PR enables fusion of pointwise ops for GPU context, using runtime compilation (NVRTC). I will make and post a design doc in the near future.

Work done by me, @Caenorst, @nvchai and @MoisesHer.

**Important note**: this is **NOT** meant to compete with other long term compilation strategies for MXNet (like integration with TVM etc.). It is intended as a short term solution (and so e.g. does not tackle harder problems like fusion with convolutions/gemms) while those long term solutions are in development.

This is the first stage of this effort, focusing on pointwise ops that do not change the shape of the output (+ slices). Future PRs will tackle broadcast ops.

FYI @zheng-da @junrushao1994 @eric-haibin-lin @szha @KellenSunderland @nvchai 

### Essentials ###

### Changes ###

## Comments ##
- The fusion is controlled by MXNET_USE_FUSION env variable and currently on by default for GPU context (mostly to see how it will behave in the CI environment) - I will probably change it to opt-in later in the review process. 
- It supports both forward and backward pass
- It supports rebinding with different shapes without the need of recompilation of the generated kernel.
- It currently supports Symbol, Module and Gluon (only hybridized without inlining, so mostly static_alloc=True) API.


piyushghai:
@ptrendx Can you look at the CI failures on this PR ? 

ptrendx:
Design doc: https://cwiki.apache.org/confluence/display/MXNET/GPU+Pointwise+fusion

larroy:
Did we document MXNET_USE_FUSION in `docs/faq/env_var.md` and add that is supported only for CUDA?

src/executor/infer_graph_attr_pass.cc
```diff
@@ -209,51 +213,86 @@ nnvm::Graph InferAttr(nnvm::Graph &&ret,
...
+      static auto& is_fusion = Op::GetAttr<exec::TIsFusionHelper>("TIsFusionHelper");
+      if (!is_fusion.get(fwd_ptr->op(), false)) {
+        const IndexedGraph::Node& fnode = idx[inode.control_deps[0]];
+        // use gradient function to find out the correspondence.
+        std::vector<nnvm::NodeEntry> ograd(fwd_ptr->num_outputs());
+        for (size_t i = 0; i < ograd.size(); ++i) {
+          ograd[i].index = static_cast<uint32_t>(i);
+        }
+        // input gradient list
+        auto igrad = fgrad[fwd_ptr->op()](fwd_ptr, ograd);
```
larroy: Can we add a type for readability?

src/operator/fusion/fused_op.cu
```diff
@@ -0,0 +1,396 @@
...
+          const auto& arg1 = variables[{node.inputs[0].node_id, node.inputs[0].index}];
+          code += "const auto " + var_name + " = " + op +
+                  "(" + arg1 + ");\n";
+          CHECK_EQ(outputs[i], 1);
+          variables[{i, 0}] = var_name;
+          continue;
+        }
+
+        if (detail::fused_op_special_ops.find(op_name) != detail::fused_op_special_ops.end()) {
+          const std::vector<std::string>& op_desc = detail::fused_op_special_ops.at(op_name);
```
larroy: why not reuse the iterator that you have when doing find? this is accessing the map two times.

src/operator/fusion/fused_op.cu
```diff
@@ -0,0 +1,396 @@
...
+          std::string fmt = op_desc[0];
+          for (size_t j = 1; j < op_desc.size(); ++j) {
+            const std::string& desc = op_desc[j];
+            std::string sub;
+            if (desc[0] == '_') {
+              // Argument
+              int arg_id = std::stoi(desc.substr(1));
```
larroy: const

src/operator/fusion/fused_op.cu
```diff
@@ -0,0 +1,396 @@
...
+          CHECK_EQ(outputs[i], 1);
+          variables[{i, 0}] = var_name;
+          continue;
+        }
+
+        if (detail::fused_op_special_ops.find(op_name) != detail::fused_op_special_ops.end()) {
+          const std::vector<std::string>& op_desc = detail::fused_op_special_ops.at(op_name);
+          std::string fmt = op_desc[0];
+          for (size_t j = 1; j < op_desc.size(); ++j) {
+            const std::string& desc = op_desc[j];
```
larroy: can't we wrap this detail in a function?

src/operator/fusion/fused_op.cu
```diff
@@ -0,0 +1,396 @@
...
+            fmt.replace(pos, 1, sub);
+          }
+          code += "const auto " + var_name + " = " + fmt + ";\n";
+          CHECK_EQ(outputs[i], 1);
+          variables[{i, 0}] = var_name;
+          continue;
+        }
+
+        if (detail::fused_op_mimo_ops.find(op_name) != detail::fused_op_mimo_ops.end()) {
+          const std::vector<std::vector<std::string>>& op_descs =
```
larroy: same as above.

src/operator/fusion/fused_op.cu
```diff
@@ -0,0 +1,396 @@
...
+          for (const auto& op_desc : op_descs) {
+            var_name = "temp" + std::to_string(temp_name_counter++);
+            std::string fmt = op_desc[0];
+            for (size_t j = 1; j < op_desc.size(); ++j) {
+              const std::string& desc = op_desc[j];
+              std::string sub;
+              if (desc[0] == '_') {
+                // Argument
+                int arg_id = std::stoi(desc.substr(1));
+                sub = variables[{node.inputs[arg_id].node_id, node.inputs[arg_id].index}];
```
larroy: duplicated code with above.

src/operator/fusion/fused_op.cu
```diff
@@ -0,0 +1,396 @@
...
+          continue;
+        }
+        LOG(FATAL) << "Unrecognized op " + op_name;
+      }
+    } else {
+      LOG(FATAL) << "Encountered node with NULL source.";
+    }
+  }
+
+  int counter = 0;
```
larroy: size_t?

src/operator/fusion/fused_op.cu
```diff
@@ -0,0 +1,396 @@
...
+                           const std::vector<OpReqType> &req,
+                           const std::vector<TBlob> &outputs) {
+  using namespace mshadow;
+  std::lock_guard<std::mutex> lock(my_mutex_);
+  CHECK_GE(outputs.size(), 1) << "There needs to be at least 1 output.";
+
+  std::vector<int> in_dtypes;
+  std::vector<int> out_dtypes;
+
+  size_t counter = 0;
```
larroy: does it make sense to check dimensions, use at or loop directly with index? could inputs_ and inputs have different sizes?

src/operator/fusion/fused_op.cu
```diff
@@ -0,0 +1,396 @@
...
+  counter = 0;
+  for (const auto& blob : outputs) {
+    out_dtypes.push_back(blob.type_flag_);
+    initialized_ = initialized_ && (blob.type_flag_ == outputs_[counter].dtype);
+    outputs_[counter].dtype = blob.type_flag_;
+    ++counter;
+  }
+
+  // Get compute capability of the current GPU
+  int dev_id = ctx.run_ctx.ctx.dev_id;
```
larroy: const

larroy: Could we wrap the initialized boolean check in a member function that gets compute cap for the current dev as the code does?

src/operator/fusion/fused_op.cu
```diff
@@ -0,0 +1,396 @@
...
+  initialized_ = initialized_ && cc_major == this->cc_major_;
+  initialized_ = initialized_ && cc_minor == this->cc_minor_;
+  this->cc_major_ = cc_major;
+  this->cc_minor_ = cc_minor;
+
+  initialized_ = initialized_ && (req == saved_reqs_);
+  saved_reqs_ = req;
+
+  if (!initialized_) {
+    this->GenerateCode(req);
```
larroy: Could we wrap codegen into it's own function detail?

src/executor/infer_graph_attr_pass.cc
```diff
@@ -209,51 +213,86 @@ nnvm::Graph InferAttr(nnvm::Graph &&ret,
...
-          if (fis_none(rshape[eid])) {
-            rshape[eid] = rshape[idx.entry_id(fnode.inputs[i])];
-          } else if (!fis_none(rshape[idx.entry_id(fnode.inputs[i])])) {
-            // Need to skip empty forward shape, because it may not be
-            // available now and it is possible to infer the forward
-            // shape in one of the next a few passes
-            CHECK_EQ(rshape[eid], rshape[idx.entry_id(fnode.inputs[i])])
-                << "Backward shape inconsistent with the forward shape";
+
+      static auto& is_fusion = Op::GetAttr<exec::TIsFusionHelper>("TIsFusionHelper");
```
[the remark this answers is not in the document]
ptrendx: Ok, I will do some refactoring there.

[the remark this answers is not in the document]
larroy: Thanks, the work looks awesome, is just that we need to eventually maintain this code.

src/operator/fusion/fused_op.cu
ptrendx: Then I would need to have the iterator in the outer scope, which I think is less readable (as the name would need to be different for each one or another scope would need to be made). And this function is not used in the performance-critical section of the run.

src/executor/infer_graph_attr_pass.cc
ptrendx: ok

src/operator/fusion/fused_op.cu
ptrendx: ok

src/operator/fusion/fused_op.cu
ptrendx: sure.

src/operator/fusion/fused_op.cu
ptrendx: I refactored it out.

src/operator/fusion/fused_op.cu
ptrendx: done.

src/operator/fusion/fused_op.cu
ptrendx: ok

src/operator/fusion/fused_op.cu
ptrendx: ok

src/operator/fusion/fused_op.cu
ptrendx: ok

src/operator/fusion/fused_op.cu
```diff
@@ -0,0 +1,534 @@
...
+            load_index[i] = 1;
+        } else {
+            std::string op_name = source->op()->name;
+            if (detail::fused_op_slice_ops.find(op_name) != detail::fused_op_slice_ops.end()) {
+                load_index[node.inputs[0].node_id] = 0;
+            }
+        }
+    }
+  }
+  for (size_t i = 0; i < g.num_nodes(); ++i) {
```
larroy: indentation is weird.

src/operator/fusion/fused_op-inl.h
```diff
@@ -0,0 +1,906 @@
...
+  return tanf(val);
+}
+
+// Uses output from tan
+template <typename DType>
+inline DType backward_tan(const DType out, const DType grad) {
+  return grad * (out * out + 1);
+}
+
+template <typename DType>
```
larroy: stupid question, but why can't we use mshadow ops for this? (mshadow_op.h)   why reimplement?

src/operator/fusion/fused_op.cu
```diff
@@ -0,0 +1,534 @@
...
+    size_t pos = 0;
+    while ((pos = input->find(old, pos)) != std::string::npos) {
+        input->replace(pos, old.size(), repl);
+        pos += repl.size();
+    }
+}
+
+}  // namespace detail
+
+std::string ParseOpDescription(const std::vector<std::string>& op_desc,
```
larroy: can it be static or in anon ns?

src/operator/fusion/fused_op.cu
larroy: Thanks, looks better.

src/operator/fusion/fused_op-inl.h
ptrendx: Main reason is that it is not really possible to include mshadow_op.h in runtime compilation - we would need to ship A LOT of headers with binary package (including entire mshadow) and make a mechanism for NVRTC to be able to find them. mshadow_op.h also relies on cuda_fp16.h, which would mean that we need to be able to find at runtime CUDA include directory.

src/operator/fusion/fused_op.cu
ptrendx: ok

ptrendx: sure

src/common/exec_utils.h
```diff
@@ -622,6 +622,51 @@ inline nnvm::Graph AssignContext(nnvm::Graph g,
   return g;
 }
 
+inline void CopyGraph(nnvm::Graph *dst, const nnvm::Graph &src, bool copy_variables) {
```
larroy: since this is used in multiple files, could we not inline it and put it in an implementation file (exec_utils.cc) to reduce code bloat.

src/common/exec_utils.h
```diff
@@ -622,6 +622,51 @@ inline nnvm::Graph AssignContext(nnvm::Graph g,
...
+    }
+  }
+  // set the head
+  for (const NodeEntry &e : src.outputs) {
+    (*dst).outputs.emplace_back(NodeEntry{old_new[e.node.get()], e.index, e.version});
+  }
+}
+
+inline bool CheckForInputNameDuplicates(const nnvm::IndexedGraph &idx) {
+  std::set<std::string> names;
```
larroy: can be unordered?

src/executor/node_entry_count.cc
```diff
@@ -0,0 +1,46 @@
...
+ * \brief function that count how many times a node entry is used
+ * \author Clement Fuji Tsang
+ */
+#include "./exec_pass.h"
+
+// TODO(cfujitsang): should this be pushed to nnvm repository ?
+namespace mxnet {
+namespace exec {
+
+NodeEntryMapCounter GetNodeEntryCount(const nnvm::Graph& g) {
```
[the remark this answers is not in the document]
ptrendx: This is actually not needed for this fusion, will remove it

src/executor/exec_pass.h
```diff
@@ -170,6 +190,16 @@ void AttachOpResources(const Graph& g,
  */
 Graph DetectInplaceAddTo(Graph g);
 
+using NodeEntryMapCounter =
+  std::unordered_map<nnvm::NodeEntry, uint32_t, nnvm::NodeEntryHash, nnvm::NodeEntryEqual>;
+/*!\brief
+ * This is to count how many time each output is used by another node (or the output of the graph)
```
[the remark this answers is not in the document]
ptrendx: This is not needed for this fusion, will remove it.

src/operator/fusion/fused_op.cu
ptrendx: I changed it to loop directly with index + a check for the sizes matching.

src/operator/fusion/fused_op.cu
ptrendx: fixed

src/common/exec_utils.h
ptrendx: sure, I will put the functions we introduced in .cc file. However, putting other functions in exec_utils.cc should be done in another PR.

src/common/exec_utils.h
ptrendx: sure

ptrendx:
> Did we document MXNET_USE_FUSION in `docs/faq/env_var.md` and add that is supported only for CUDA?

I added this in the latest commit.

larroy:
Thanks for addressing the comments.

docs/faq/env_var.md
```diff
@@ -309,6 +309,17 @@ If ctypes is used, it must be `mxnet._ctypes.ndarray.NDArrayBase`.
     with float32.
   - Model accuracies do not necessarily improve with this environment variable turned on.
 
+* MXNET_USE_FUSION
```
pengzhao-intel: As I suggested in dev@, could we align the variable to MXNET_SUBGRAPH_BACKEND to make the easy usage for the user? Currently, this env is for CPU operator fusion.

ptrendx: I thought about it but I'm not sure - this fusion does not really use subgraph API because of its limitations (like Gluon support, separate forward and backward fusion), and would not work with the "get_optimized_symbol" API (which using this env variable would imply).

pengzhao-intel: It makes sense. 

@aaronmarkham  do you have any suggestion?
Simple background, @ptrendx please correct me if anything is not correct.

Currently, two environments deliver similar functionality where `MXNET_SUBGRAPH_BACKEND=MKLDNN` is for CPU fusion and `MXNET_USE_FUSION` is for GPU.
I suggest aligning to `MXNET_SUBGRAPH_BACKEN={MKLDNN, CUDNN}` because this is already widely used for operator fusion but the name of `SUBGRAPH` doesn't match with technical details inside GPU implementation.

src/executor/pointwise_fusion_pass.cc
```diff
@@ -0,0 +1,279 @@
...
+                                FCreateNode create_subgraph_node) {
+  for (auto subgraph_set : subgraph_sets) {
+    // Create MXNet subgraph
+    Graph subgraph;
+    const auto sub_outputs_in_main = GetSubgraphOutputs(g, subgraph_set);
+    subgraph.outputs.resize(sub_outputs_in_main.size());
+    for (auto p : sub_outputs_in_main) {
+      subgraph.outputs[p.second] = p.first;
+    }
+    // To generate a subgraph an input have to be replaced by data node (no op)
```
aaronmarkham:
```suggestion
    // To generate a subgraph an input has to be replaced by data node (no op)
```

src/executor/pointwise_fusion_pass.cc
```diff
@@ -0,0 +1,279 @@
...
+    // and it have to be agnostic to the node from which it's an output
```
aaronmarkham:
```suggestion
    // and it has to be agnostic to the node from which it's an output
```

src/executor/pointwise_fusion_pass.cc
```diff
@@ -0,0 +1,279 @@
...
+    // (For exemple even if two inputs are two different outputs from the same node)
```
aaronmarkham:
```suggestion
    // (For example, even if two inputs are two different outputs from the same node)
```

aaronmarkham: This example isn't that clear to me. Seems like there should be a followup sentence. "Even if x... then y"... it's missing the "then y".

src/executor/simple_partition_pass.h
```diff
@@ -0,0 +1,391 @@
...
+                       FCreateNode create_subgraph_node) {
+  for (auto subgraph_set : subgraph_sets) {
+    // Create MXNet subgraph
+    Graph subgraph;
+    const auto sub_outputs_in_main = GetSubgraphOutputs(g, subgraph_set);
+    subgraph.outputs.resize(sub_outputs_in_main.size());
+    for (auto p : sub_outputs_in_main) {
+      subgraph.outputs[p.second] = p.first;
+    }
+    // To generate a subgraph an input have to be replace by data node (no op)
```
aaronmarkham:
```suggestion
    // To generate a subgraph an input has to be replaced by data node (no op)
```

src/executor/simple_partition_pass.h
```diff
@@ -0,0 +1,391 @@
...
+    // and it have to be agnostic to the node from which it's an output
```
aaronmarkham:
```suggestion
    // and it has to be agnostic to the node from which it's an output
```

src/executor/simple_partition_pass.h
```diff
@@ -0,0 +1,391 @@
...
+    // (For exemple even if two inputs are two different outputs from the same node)
```
aaronmarkham:
```suggestion
    // (For example, even if two inputs are two different outputs from the same node)
```
Same concern as before... this needs more info...

src/executor/pointwise_fusion_pass.cc
ptrendx: Sure, will make it clearer.

ptrendx:
@szha @KellenSunderland @eric-haibin-lin Do you have any comments to this PR?

docs/faq/env_var.md
```diff
@@ -309,6 +309,17 @@ If ctypes is used, it must be `mxnet._ctypes.ndarray.NDArrayBase`.
...
+  - Values: 0(false) or 1(true) ```(default=1)```
+  - If this variable is set, MXNet will try fusing some of the operations (pointwise operations only for now).
+  - It works in Symbolic execution as well as in Gluon models hybridized with ```static_alloc=True``` option.
+  - Only applies to MXNet that has been compiled with CUDA (```pip install mxnet-cuXX``` or built from source with ```USE_CUDA=1```)
```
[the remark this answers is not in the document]
ptrendx: It will not do anything then, I will make it clearer.

src/executor/pointwise_fusion_pass.cc
```diff
@@ -0,0 +1,275 @@
...
+      }
+    });
+  }
+  Graph new_graph;
+  new_graph.outputs = g.outputs;
+  return new_graph;
+}
+
+template <typename IsCompatible>
+void AddInputsOnlyCompatible(const Graph &g,
```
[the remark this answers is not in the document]
ptrendx: Sure.

src/operator/fusion/fused_op.cc
```diff
@@ -0,0 +1,269 @@
...
+    os << ")";
+    throw dmlc::ParamError(os.str());
+  }
+  attrs->parsed = FusedOpPtr(new FusedOp(attrs, param));
+}
+
+FusedOp::FusedOp(const nnvm::NodeAttrs* attrs, const FusedOpConfig& config) {
+  this->inputs_ = std::vector<FusedOpEntry>(config.num_inputs);
+  this->outputs_ = std::vector<FusedOpEntry>(config.num_outputs);
+  this->symbol_ = nnvm::Graph();
```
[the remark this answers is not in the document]
ptrendx: Yeah, this is leftover from the very initial implementation that had nnvm::Symbol and not Graph here. I will change it to subgraph_.

tests/python/gpu/test_fusion.py
```diff
@@ -0,0 +1,195 @@
...
+    end = (random.randint(begin[0]+1, shape[0]-1),
+           random.randint(begin[1]+1, shape[1]-1),
+           random.randint(begin[2]+1, shape[2]-1))
+    check_fused_symbol(mx.sym.slice(a, begin=begin, end=end), a=arr1)
+
+@with_seed()
+def test_fusion():
+    check_unary_ops()
+    check_binary_ops()
+    check_other_ops()
```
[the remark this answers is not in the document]
ptrendx: Nosetests does not need main block, it is more of a convenience so that you are able to run the test script via `python test_fusion.py` instead of `nosetests test_fusion.py`. That said, most tests have it, so I will add it here as well.

src/operator/fusion/fused_op.cc
```diff
@@ -0,0 +1,269 @@
...
+  return inferred;
+}
+
+bool FusedOp::InferType(const nnvm::NodeAttrs &attrs,
+                        std::vector<int> *in_attrs,
+                        std::vector<int> *out_attrs) {
+  this->symbol_.attrs.erase("dtype");
+  this->symbol_.attrs.erase("dtype_inputs");
+  std::vector<int> input_types(*in_attrs);
+  this->symbol_ = mxnet::exec::InferType(std::move(this->symbol_),
```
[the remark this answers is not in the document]
ptrendx: I'm not sure if it would ever not be true, but ultimately it does not matter - if the outputs can be inferred that is enough. The only situation where the internal node's attributes are needed is backward, but then the main inferattr would fail due to backward node not knowing its attributes.

karan6181:
@ptrendx gentle ping!

src/executor/graph_executor.cc
```diff
@@ -981,6 +986,41 @@ Graph GraphExecutor::InitGraph(nnvm::Symbol symbol,
   // setup gradient
   nnvm::Graph g = InitFullGraph(symbol, grad_req_types);
 
+#if MXNET_USE_CUDA && !defined(_WIN32)
+  if (dmlc::GetEnv("MXNET_USE_FUSION", true) && default_ctx.dev_mask() == Context::kGPU) {
```
cjolivier01: nit: the second test is faster than the relatively much more slow GetEnv() call, so maybe better to do the GetEnv() second?

src/executor/infer_graph_attr_pass.cc
```diff
@@ -63,6 +63,135 @@ bool ApplyOpInferAttr<int, FInferStorageType>(const nnvm::Graph& g,
...
+}
+
+template<typename FAccessSubgraphType, typename AttrType, typename IsNone>
+void GetAttrFromFusedNode(uint32_t nid,
+                          const nnvm::IndexedGraph& idx,
+                          std::vector<AttrType>* rshape_ptr,
+                          IsNone fis_none,
+                          const std::string& infer_fusion_name) {
+  std::vector<AttrType>& rshape = *rshape_ptr;
+  const auto& inode = idx[nid];
```
cjolivier01: nit: too much use of “auto” where type is not obvious in this review. Makes the code harder to read.

ptrendx:
@cjolivier01 That is simply not true (and "I could post examples but I will not" as well as the single line that you gave as an example, which is actually trivial use case for auto, does not prove your point). In fact the line you pointed at:
```
const auto& inode = idx[nid];
```
without `auto` would look like this
```
const nnvm::IndexedGraph::Node& inode = idx[nid];
```
which makes the code less readable (the fact that `inode` is a node is given by its name, so the typing the type name gives you the same redundant information and actually makes it harder to spot the name of the variable which is more important). 

Here are actual examples of guidelines:
 - [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es11-use-auto-to-avoid-redundant-repetition-of-type-names)
 - [video of a discussion in C++ 2012 and beyond discussing auto usage](http://channel9.msdn.com/Shows/Going+Deep/C-and-Beyond-2012-Scott-Andrei-and-Herb-Ask-Us-Anything#time=25m03s)

The conclusion form those guidelines is that you should be using auto nearly everywhere because it makes the code more future proof (e.g. when implementation of the function change and the type returned is different but still has the same properties, you do not need to change the code that actually uses the function) AND more readable (because it is shorter and you only need to think about the properties of the type that you get from the function call instead of the type itself).

src/executor/graph_executor.cc
ptrendx: Sure, can do that.

cjolivier01:
I looked at those “core guildelines” and it’s basically what I was saying — stuff like ::iterator and repetition or obvious type on same line. where you get the “always use auto” I don’t know. from the video? videos are TLDR-ish, but I certainly wouldn’t trust anyone who makes a video saying to use auto everywhere. that’s just crazy :)

ptrendx:
If you did not want to put the effort to watch the video (portion about auto takes ~4 minutes there) then using "change request" does not seem valid. And if you think that arguments of C++ standard committee members are "just crazy" then maybe you should reconsider your usage of C++?

Also, did you even read what I posted? C++ Core Guidelines (again, maintained by ISO C++ Foundation) text has 1 exception for using auto:
```
Exception
Avoid auto for initializer lists and in cases where you know exactly which type you want and where an initializer might require conversion.
```
There are no other exceptions, so I don't know where you get your `stuff like ::iterator and repetition or obvious type on same line` from.

cjolivier01:
We apparently have very different interpretation of those guidelines. I don’t see anything in that which implies “use it everywhere for everything”. I’m more of the opinion stated here:

While I am not i. the habit of posting stack overflow nonsense as “evidence”, seems to be a consistent understanding about proper auto use:

https://www.acodersjourney.com/c-11-auto/

https://stackoverflow.com/questions/6900459/the-new-keyword-auto-when-should-it-be-used-to-declare-a-variable-type

Using auto everywhere makes C++ code no more readable than something like python.

ptrendx:
Considering Python is quite widely believed to be "easier" language than C++, the `Using auto everywhere makes C++ code no more readable than something like python.` comment is slightly counter intuitive.

Even the blog post you link to actually is saying the same thing - you should be using auto most of the time. And for the example that he starts with (that is supposedly bad use of `auto`):
```
auto a = ConjureMagic();
SetMagic(a);
```
is just wrong, since the alternative (non `auto`) form of let's say
```
std::vector<SuperMagicType> a = ConjureMagic();
SetMagic(a);
```
does not give me any more info on what the code does, but if later the `ConjureMagic` function changes to give back, let's say `std::array<SuperMagicType>` or something, then the second code will either not compile anymore or, even worse, a costly conversion will occur. The actual solution to his problem would be to refactor the `ConjureMagic` function name to actually state what is the result (and not in a `ConjureMagicInteger` way, but in business logic way, like `GetComputationEngine` or something). 

In the StackOverflow thread you posted there was an (not the most upvoted, I will give you that) answer citing this [blog post](https://blogs.msdn.microsoft.com/ericlippert/2011/04/20/uses-and-misuses-of-implicit-typing/) by the principal developer of C# compiler about `var` (which is equivalent to `auto` in C++), which ends with this conclusion:
```
Summing up, my advice is:

Use var when you have to; when you are using anonymous types.
Use var when the type of the declaration is obvious from the initializer, especially if it is an object creation. This eliminates redundancy.
Consider using var if the code emphasizes the semantic "business purpose" of the variable and downplays the "mechanical" details of its storage.
Use explicit types if doing so is necessary for the code to be correctly understood and maintained.
Use descriptive variable names regardless of whether you use "var". Variable names should represent the semantics of the variable, not details of its storage; "decimalRate" is bad; "interestRate" is good.
```

I agree with this, so if you believe that there is necessity to use explicit typing somewhere in this PR in order for it to be correctly understood and maintained, please point me to the specific places in the code where this situation occurs and I will reevaluate those cases - otherwise your feedback is not really actionable.

larroy:
Guys, let's keep the conversation constructive and focused on architecture and approaches rather than nits. In some cases some judicious use of "using" to import namespace can help. for example:

```
const nnvm::IndexedGraph::Node& inode = idx[nid];
```

This can be rewritten to

```
const Node& inode = idx[nid];
```

In this case is better to use the type and reduce long namespace resolution if you are constantly using classes in some namespace.

This would be my preference and advice.
Maybe we need to write a guideline approved by the community? Having these discussions in individual PRs slows things down too much.

❤️

Caenorst:
@larroy I don't think it's recommended as there is nnvm::Node and nnvm::IndexedGraph::Node.

I don't how this is a problem as this line is already in master: https://github.com/apache/incubator-mxnet/blob/01fae12e58d73d5989a23cfc8b7519f4772fdf1f/src/executor/graph_executor.cc#L109 https://github.com/apache/incubator-mxnet/blob/01fae12e58d73d5989a23cfc8b7519f4772fdf1f/src/executor/graph_executor.cc#L1214 for instance

Caenorst:
@larroy I don't think it's recommended as there is nnvm::Node and nnvm::IndexedGraph::Node.

I don't understand how this is a problem as this line is already in master: https://github.com/apache/incubator-mxnet/blob/01fae12e58d73d5989a23cfc8b7519f4772fdf1f/src/executor/graph_executor.cc#L109 https://github.com/apache/incubator-mxnet/blob/01fae12e58d73d5989a23cfc8b7519f4772fdf1f/src/executor/graph_executor.cc#L1214 for instance

cjolivier01:
mxnet doesn’t yet have a c# interface — are you volunteering?  The functionality would be welcomed :)

cjolivier01:
1) i randomly picked that line as representative of auto used for pretty much everything
2) pointing out this or that line is “already in master” at some random place is interesting. If you got cancer, would you have the attitude like “I already have some cancer in me, so clearly more cancer is acceptable”, or would you try to get rid of the cancer if possible, or at the very least slow its growth? If it was a couple of autos here and there, it could pass, but this PR kind of takes that to another level and sets a bad precedent for future code (in case someone wants to point out, well they use “auto” everywhere in this code)...

larroy:
@Caenorst yes in this case you are right. I was trying to convey a general example.

szha:
@ptrendx @cjolivier01 To help move this effort forward, shall we do the following?
1. let's start a discussion on dev@ on the specific topic of `auto` usage, present the arguments, and try to reach consensus there.
2. if @cjolivier01 is willing, we can unblock this PR for now and if necessary address the concern once the discussion concludes.
3. alternatively, if @ptrendx is willing, we could avoid the `auto` usage in this PR to get it unblocked immediately, or wait until the discussion to conclude.

ptrendx:
@eric-haibin-lin Support for fusion without hybridization would be really tricky (you do not know what is the next operation in your graph) and would effectively end up being very similar to hybridization (in order to know what to do you would need to trace the graph and run it only during some sync points). Also if you really need imperative execution (because you have some condition to run different portions of the graph for example) then the overhead of compilation may not be worth it.
Optimizers are interesting use case for fusion - is there any reason to not use Gluon HybridBlocks to implement them?

ptrendx:
@szha As I stated in one of the previous comments - I am open to reevaluating specific cases of `auto` usage in this PR (and they were not given yet). That said, I believe that the main problem with understanding the code that uses `auto` is not the `auto` keyword itself, but the functions that produce values captured by it (that e.g. have a name that does not really say properly what the function is supposed to do).
The discussion on dev@ would be nice.

src/operator/fusion/fused_op.cu
```diff
@@ -0,0 +1,623 @@
...
+                        "--std=c++11",
+                        "-default-device"};
+  const std::string kernel_name_demangled = "FusedKernel_" + kernel_name;
+  NVRTC_CALL(nvrtcAddNameExpression(program, (kernel_name_demangled).c_str()));
+
+  nvrtcResult compileResult = nvrtcCompileProgram(program,  // prog
+                                                  3,        // numOptions
+                                                  opts);    // options
+  // Obtain compilation log from the program.
+  size_t logSize;
```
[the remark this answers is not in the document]
ptrendx: Sure, will change that.

cjolivier01:
Google c++ style guide says basically the same thing as what I’m encouraging:

https://google.github.io/styleguide/cppguide.html#auto

I’d like to review this PR, but it’s unreadable in its current state, imho. I do not intend to have extended arguments on a case-by-case basis, as I don’t think that sort of turnabout would ultimately be constructive.

marcoabreu:
Okay, so lets have a discussion on dev@ and then agree on a style.

cjolivier01:
I am removing my block on this PR and letting in this "auto-abuse" in since I am in a similar industry to NVidia and don't wish to make it look like I am blocking a performance improvement for NVidia GPU's.  I hope that the auto-abuse will be fixed before merging, however.

larroy:
@cjolivier01 you are no longer active in this project and are just blocking PRs and preventing contributions, do the right thing for the community and resign from the PMC.

Jerryzcn:
@ptrendx Does this support AMP? Do you observe any speedup when training mask rcnn?

ptrendx:
@Jerryzcn Yes, it does support AMP. I saw a (pretty small, 1-2%) speedup in the non-FPN version of MaskRCNN from GluonCV.

src/executor/infer_graph_attr_pass.cc
samskalicky: are you planning to use real typenames here instead of auto?

src/executor/simple_partition_pass.h
```diff
@@ -0,0 +1,392 @@
...
+ */
+class BidirectionalGraph {
+ public:
+  struct Node {
+    nnvm::Node* nnvmptr;
+    std::vector<Node*> inputs;
+    std::vector<Node*> outputs;
+  };
+
+  explicit BidirectionalGraph(const Graph &g) {
```
samskalicky: needs more comments. at least explain the algorithm thats written here for each scope. This file specifically but other files too.

src/executor/simple_partition_pass.h
```diff
@@ -0,0 +1,392 @@
...
+    auto& idx = g.indexed_graph();
+    auto num_nodes = idx.num_nodes();
+    nodes.reserve(num_nodes);
+    nnvm2nid.reserve(num_nodes);
+    outputs.reserve(idx.outputs().size());
+    DFSVisit(g.outputs, [this](const nnvm::NodePtr& n) {
```
samskalicky: example comment, fill array of nodes with copy of each node in the graph

src/executor/simple_partition_pass.h
ptrendx: Sure, will do.

src/executor/infer_graph_attr_pass.cc
ptrendx: Done.

rondogency:
Hi @DickJC123 looks like this PR increases the CI centos-gpu testing time by over 100%, which is caused by some gpu tests running time increasing by over 20 times. For example: 

test_operator_gpu.test_sparse_mathematical_core goes from ~13s to ~350s, test_operator_gpu.test_lstm_bidirectional goes from ~15s to ~450s, 
test_operator_gpu.test_rnnrelu_bidirectional goes from ~4s t0 ~250s.

It means this PR may cause some regression, so can you take a look? I did the comparison between this PR and previous PRs.

rondogency:
@ptrendx @Caenorst @nvchai

sxjscience:
Would it be more appropriate to turn off fused_op by default? We can still provide documentation for the user to turn on it manually.

DickJC123:
Thanks for pointing out this large perf change.  I will investigate.

ptrendx:
I would say it is expected - the whole point of the feature is compile the portion of the model (which is much more expensive than just running that part) in other for subsequent iterations to run faster. Since CI runs unit tests, it creates a large number of models and never reuses them, so the compilation overhead can't be amortized. In order to reduce the overall ci time I would say the best way would be to enable fusion only on 1 or 2 configs in CI instead of all of them.

@sxjscience That approach traditionally did not work well - if you have a nonobvious feature that is not enabled by default users have no chance of actually getting benefit from it because that don't know they should be enabling it. Also, the CI example aside, there should never be a usecase where you do not want to enable fusion as it makes the training faster.

sxjscience:
@ptrendx Okay, I just think that we may need more time to test for the cases in GluonNLP and GluonCV.

DickJC123:
After some investigation, I have an explanation and planned fix for the perf regression. To repeat what @ptrendx mentions, the real-time compilation of fused kernels takes additional time up front, with the idea that over many kernel invocations, the compile time will be more than made up for by the increase efficiency of the fused op.  This matches the typical use case (unlike CI), so I believe that fusion should be left enabled by default.

Now one saving thing for each of the 3 tests mentioned by @rondogency is that most of the fused-ops in the test are duplicates of others seen earlier in the test.  In fact <2% of the created fused-ops are unique.  To fix this then I will submit a PR to introduce a 'fused op cache' that will map (source-code, gpu_arch) -> runnable kernel.  This should eliminate most of the runtime compilations and correct in large part the issue flagged here.

PR #15169, opened 2019-06-06
Softmax with length

haojin2:
## Description ##
Softmax with extra length input to specify line length of input on an axis.

### Essentials ###

### Changes ###
- [x] Softmax with length
- [x] Unit test

## Comments ##
Flakiness Check:
```
MXNET_TEST_COUNT=10000 nosetests tests/python/unittest/test_operator.py:test_softmax_with_length
[INFO] Setting module np/mx/python random seeds, use MXNET_MODULE_SEED=507099719 to reproduce.
[23:18:55] src/operator/contrib/../tensor/./../../common/utils.h:450: MXNET_SAFE_ACCUMULATION=1 is recommended for softmax with float16 inputs. See https://mxnet.incubator.apache.org/versions/master/faq/env_var.html for more details.
.
Ran 1 test in 165.490s

OK
MXNET_TEST_COUNT=10000 nosetests tests/python/gpu/test_operator_gpu.py:test_softmax_with_length
[INFO] Setting module np/mx/python random seeds, use MXNET_MODULE_SEED=634837637 to reproduce.
[23:24:32] src/operator/contrib/../tensor/./../../common/utils.h:450: MXNET_SAFE_ACCUMULATION=1 is recommended for softmax with float16 inputs. See https://mxnet.incubator.apache.org/versions/master/faq/env_var.html for more details.
[23:24:32] src/operator/contrib/../tensor/./../../common/utils.h:450: MXNET_SAFE_ACCUMULATION=1 is recommended for softmax with float16 inputs. See https://mxnet.incubator.apache.org/versions/master/faq/env_var.html for more details.
.
Ran 1 test in 137.404s

OK
```

Benchmark results:
CPU: ~1.82x speedup
GPU: ~1.59x speedup

Benchmark script:
```python
import mxnet as mx
import numpy as np
from mxnet.test_utils import check_speed, rand_ndarray

ctx = mx.gpu(0)
ctx = mx.cpu()

shape = (96, 1024, 1024)

len_shape = (96, 1024)

data = rand_ndarray(shape, ctx=ctx)

np_length = np.zeros(len_shape, dtype=np.int32)

all_length = np.arange(1, 1025, 1, dtype=np.int32)

for i in range(len_shape[0]):
    np_length[i, :] = all_length

length = mx.nd.array(np_length, ctx=ctx, dtype=np.int32)

mx_data = mx.sym.Variable("data")
mx_length = mx.sym.Variable("length")

mx_sym = mx.sym.softmax(data=mx_data, length=mx_length, use_length=True, axis=1)

print(check_speed(mx_sym, location={"data": data, "length": length}, ctx=ctx, N=100, typ='whole'))

mx_sym = mx.sym.softmax(data=mx_data, axis=1)

print(check_speed(mx_sym, location={"data": data}, ctx=ctx, N=100, typ='whole'))
```

### DIFF
--- a/src/operator/nn/softmax-inl.h
+++ b/src/operator/nn/softmax-inl.h
@@ -113,6 +113,60 @@ inline void Softmax(Stream<cpu> *s, DType *in, OType *out,
   }
 }
 
+template<typename OP, bool negate, typename AType, typename DType, typename OType,
+         typename IType, int ndim>
+inline void SoftmaxWithLength(Stream<cpu> *s, DType *in, OType *out, IType *length,
+                              Shape<ndim> shape, int axis, const DType temperature) {
+  index_t M = shape[axis];
+  index_t N = shape.Size()/M;
+  Shape<ndim> stride = calc_stride(shape);
+  Shape<ndim> sshape = shape;
+  sshape[axis] = 1;
+  index_t sa = stride[axis];
+
+  #pragma omp parallel for
+  for (int i = 0; i < static_cast<int>(N); ++i) {
+    IType len = length[i];
+    index_t base = unravel_dot(i, sshape, stride);
+
+    DType mmax = negate ? -in[base] : in[base];
+    DType val;
+    for (index_t j = 1; j < len; ++j) {
+      val = negate ? -in[base + j*sa] : in[base + j*sa];
+      if (mmax < val) mmax = val;
+    }
+    for (int j = len; j < M; ++j) {
+      out[base + j*sa] = OType(0.0f);
+    }
+
+    AType sum = AType(0);
+    DType in_val;
+    // By default temperature is 1.0.
+    // Adding a branch here to save the CPU 'divide-by-1' computation at runtime
+    if (temperature == 1.0) {
+      for (index_t j = 0; j < len; ++j) {
+        in_val = negate ? -in[base + j*sa] : in[base + j*sa];
+        sum += std::exp(in_val - mmax);
+      }
+
+      for (index_t j = 0; j < len; ++j) {
+        in_val = negate ? -in[base + j*sa] : in[base + j*sa];
+        out[base + j*sa] = OP::Map(in_val - mmax, sum);
+      }
+    } else {
+      for (index_t j = 0; j < len; ++j) {
+        in_val = negate ? -in[base + j*sa] : in[base + j*sa];
+        sum += std::exp((in_val - mmax)/temperature);
+      }
+
+      for (index_t j = 0; j < len; ++j) {
+        in_val = negate ? -in[base + j*sa] : in[base + j*sa];
+        out[base + j*sa] = OP::Map((in_val - mmax)/temperature, sum);
+      }
+    }
+  }
+}
+
 
 struct softmax_bwd {
   template<typename DType, typename AType>
@@ -235,6 +289,65 @@ inline void Softmax(Stream<gpu> *s, DType *in, OType *out,
   MSHADOW_CUDA_POST_KERNEL_CHECK(softmax_compute_kernel);
 }
 
+template<int x_bits, typename OP, bool negate, typename AType, int ndim,
+  typename DType, typename OType, typename IType>
+__global__ void softmax_with_length_kernel(DType *in, OType *out, IType *length,
+                                           index_t M, int axis, Shape<ndim> sshape,
+                                           Shape<ndim> stride, const double temperature) {
+  const unsigned x_size = 1 << x_bits;
+  __shared__ AType smem[x_size];
+  index_t sa = stride[axis];
+  index_t base = unravel_dot(blockIdx.x, sshape, stride);
+  index_t x = threadIdx.x;
+  IType len = length[blockIdx.x];
+
+  red::maximum::SetInitValue(smem[x]);
+  for (index_t i = x; i < len; i += x_size) {
+    smem[x] = ::max(smem[x], negate ? -in[base + i*sa] : in[base + i*sa]);
+  }
+  __syncthreads();
+  cuda::Reduce1D<red::maximum, x_bits>(smem);
+  __syncthreads();
+  DType smax = smem[0];
+  __syncthreads();
+
+  red::sum::SetInitValue(smem[x]);
+  DType val;
+  for (index_t i = x; i < len; i += x_size) {
+    val = negate ? -in[base + i*sa]:in[base + i*sa];
+    smem[x] += static_cast<AType>(expf((val - smax) / static_cast<AType>(temperature)));
+  }
+  __syncthreads();
+  cuda::Reduce1D<red::sum, x_bits>(smem);
+  __syncthreads();
+  AType ssum = smem[0];
+  __syncthreads();
+
+  for (index_t i = x; i < M; i += x_size) {
+    val = negate ? -in[base + i*sa] : in[base + i*sa];
+    out[base + i*sa] =
+      (i < len) ? OP::Map((val - smax)/static_cast<DType>(temperature), ssum) : OType(0.0f);
+  }
+}
+
+template<typename OP, bool negate, typename AType, typename DType, typename OType,
+         typename IType, int ndim>
+inline void SoftmaxWithLength(Stream<gpu> *s, DType *in, OType *out, IType *length,
+                    Shape<ndim> shape, int axis, const double temperature) {
+  const int x_bits = 7;
+  const int x_size = 1 << x_bits;
+  index_t M = shape[axis];
+  index_t N = shape.Size()/M;
+  Shape<ndim> stride = calc_stride(shape);
+  Shape<ndim> sshape = shape;
+  sshape[axis] = 1;
+
+  softmax_with_length_kernel<x_bits, OP, negate, AType, ndim>
+    <<<N, x_size, 0, mshadow::Stream<gpu>::GetStream(s)>>>(
+      in, out, length, M, axis, sshape, stride, temperature);
+  MSHADOW_CUDA_POST_KERNEL_CHECK(softmax_compute_kernel);
+}
+
 
 template<int x_bits, typename OP1, typename OP2, int Req, bool negate, typename AType, int ndim,
   typename DType, typename OType>
@@ -295,6 +408,7 @@ struct SoftmaxParam : public dmlc::Parameter<SoftmaxParam> {
   int axis;
   dmlc::optional<double> temperature;
   dmlc::optional<int> dtype;
+  dmlc::optional<bool> use_length;
   DMLC_DECLARE_PARAMETER(SoftmaxParam) {
     DMLC_DECLARE_FIELD(axis).set_default(-1)
     .describe("The axis along which to compute softmax.");
@@ -307,6 +421,9 @@ struct SoftmaxParam : public dmlc::Parameter<SoftmaxParam> {
     .set_default(dmlc::optional<int>())
     .describe("DType of the output in case this can't be inferred. "
               "Defaults to the same as input's dtype if not defined (dtype=None).");
+    DMLC_DECLARE_FIELD(use_length)
+    .set_default(dmlc::optional<bool>())
+    .describe("Whether to use the length input as a mask over the data input.");
   }
 };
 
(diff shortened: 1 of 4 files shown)


piyushghai:
@haojin2 Can you look into the CI failures on this ? 

@mxnet-label-bot Add [Operator, pr-awaiting-review]

haojin2:
@szha @eric-haibin-lin Finally the CI passed... Please give a review when you have time.

src/operator/nn/softmax-inl.h
```diff
@@ -113,6 +113,60 @@ inline void Softmax(Stream<cpu> *s, DType *in, OType *out,
   }
 }
 
+template<typename OP, bool negate, typename AType, typename DType, typename OType,
+         typename IType, int ndim>
+inline void SoftmaxWithLength(Stream<cpu> *s, DType *in, OType *out, IType *length,
```
apeforest: Can we merge this function with Softmax (treat length = M in the default case). Creating two functions just to serve one extra argument seems to be an overkill

tests/python/unittest/test_operator.py
```diff
@@ -7838,6 +7871,7 @@ def get_output_names_callback(name, arr):
         except mx.base.MXNetError:
             # skip errors since test is to check all names
             pass
+        print(output_names)
```
apeforest: please remove print statement from unit test.

src/operator/nn/softmax-inl.h
```diff
@@ -464,35 +755,56 @@ void SoftmaxGradCompute(const nnvm::NodeAttrs& attrs,
...
   MXNET_REAL_ACC_TYPE_SWITCH(inputs[0].type_flag_, OType, AType, {
     MSHADOW_REAL_TYPE_SWITCH(outputs[0].type_flag_, DType, {
       MXNET_ASSIGN_REQ_SWITCH(req[0], Req, {
-        if (safe_acc) {
-          if (shape.ndim() == 2) {
-            SoftmaxGrad<OP1, OP2, Req, negate, AType>(
-                ctx.get_stream<xpu>(), inputs[out_idx].dptr<OType>(),
-                inputs[0].dptr<OType>(), outputs[0].dptr<DType>(),
-                shape.get<2>(), axis, static_cast<DType>(temperature));
+        if (!softmax_use_length(attrs)) {
```
apeforest: Can we combine this two conditions with the default length to be M? This condition statement is already complicated enough.

src/operator/nn/softmax-inl.h
```diff
@@ -464,35 +755,56 @@ void SoftmaxGradCompute(const nnvm::NodeAttrs& attrs,
...
+              SoftmaxGrad<OP1, OP2, Req, negate, DType>(
+                  ctx.get_stream<xpu>(), inputs[out_idx].dptr<OType>(),
+                  inputs[0].dptr<OType>(), outputs[0].dptr<DType>(),
+                  shape.get<3>(), axis, static_cast<DType>(temperature));
+            }
           }
         } else {
-          if (shape.ndim() == 2) {
-            SoftmaxGrad<OP1, OP2, Req, negate, DType>(
+          MXNET_INT_TYPE_SWITCH(inputs[2].type_flag_, IType, {
```
apeforest: Do we really need to iterate different int types for length? Can we just cast the type to int64_t.

haojin2:
@apeforest Trust me or not, I've tried what you're suggesting to do on my end way before this PR was raised but found it not easily done without affecting the existing softmax's performance by much. I think I'll stick to this version for now.

src/operator/nn/softmax-inl.h
haojin2: It may not necessarily be int64_t, and what do you mean by "cast the type to int64_t"? Allocating a new buffer on the fly within the operator to hold the casted `length` input? I would consider that a big performance bottleneck.

src/operator/nn/softmax-inl.h
haojin2: Answered below.

tests/python/unittest/test_operator.py
haojin2: Done.

src/operator/nn/softmax-inl.h
haojin2: The `length` is a tensor, not a single value. Please read the test cases carefully.

apeforest:
> @apeforest Trust me or not, I've tried what you're suggesting to do on my end way before this PR was raised but found it not easily done without affecting the existing softmax's performance by much. I think I'll stick to this version for now.

If you have tried earlier, can you post some performance degradation results by trying the other approach? It may help to decide if we want to make some trade off between performance and code simplicity here.

src/operator/nn/softmax-inl.h
apeforest: It does not necessarily be `int64_t`. I think you can use the same data type as M, which is `index_t` I think.

src/operator/nn/softmax-inl.h
apeforest: I understand `length` is a tensor here. This comment is the same for combining this function with SoftMax.

src/operator/nn/softmax-inl.h
apeforest: I understand length is a tensor here. I guess my question is do we really need to care what the dtype of the values from length tensor and use a  `MXNET_INT_TYPE_SWITCH` macro here to iterate. Can we just simply cast them to `index_t` regardless of the dtype of `length` tensor.

haojin2: 1. There's no iteration at all, `TYPE_SWITCH` is always a switch, not a loop. I would suggest that you read the code more carefully.
2. No matter you're doing a cast to whatever type within the kernel or not will not get rid of the `TYPE_SWITCH` at all. One way of doing a constant cast to `index_t` is within the computation kernel, like:
```c++
template<typename IType>
MSHADOW_XINLINE static void Map(int i, IType *buffer) {
  index_t val = static_cast<index_t>(buffer[i]);
}
```
And this would still require a `TYPE_SWITCH` when you're launching the kernel since this kernel is still templated and the input type is not limited.
Or if you really hate the `TYPE_SWITCH` for the compute kernel, you then have to cast your input buffer first:
```c++
Tensor<xpu, 1, index_t> index_t_buffer = ctx.requested[0].get_space_typed<xpu, 1, index_t>(<some shape>);
// Cast your buffer to index_t_buffer
// Launch the compute kernel with index_t_buffer, now you only need one TYPE_SWITCH for your input data type.
```
This I would really hate cause it's not bringing any benefit at all.
Or you only support `length` buffer of type `int64`, then you still need to use `MSHADOW_IDX_TYPE_SWITCH` to provide the error message when `length` is of other data types. Or you could insert an additional check on the data type before the kernel launches, but that would drop support for other integer types.
3. The fact that `length` is a tensor means that `length=M` is not an option, maybe you want to say `length=nullptr`? That would cause additional if branches within the kernel and will especially impact performance on GPU. The way this feature is implemented now puts that if branch on the host rather than within the kernel, which will have minimal performance impact.

src/operator/nn/softmax-inl.h
```diff
@@ -177,6 +231,51 @@ inline void SoftmaxGrad(Stream<cpu> *s, OType *out, OType *ograd,
...
+  index_t N = shape.Size()/M;
+  Shape<ndim> stride = calc_stride(shape);
+  Shape<ndim> sshape = shape;
+  sshape[axis] = 1;
+  index_t sa = stride[axis];
+
+  #pragma omp parallel for
+  for (int i = 0; i < static_cast<int>(N); ++i) {
+    index_t base = unravel_dot(i, sshape, stride);
+    IType len = length[i];
```
apeforest: Can we simply do something like
```suggestion
    index_t len = static_cast<index_t>(length[i]);
```
Shouldn't the data type of elements inside length tensor always be the same as `M`?

src/operator/nn/softmax-inl.h
```diff
@@ -177,6 +231,51 @@ inline void SoftmaxGrad(Stream<cpu> *s, OType *out, OType *ograd,
...
+                                  int axis, const DType temperature) {
+  index_t M = shape[axis];
+  index_t N = shape.Size()/M;
+  Shape<ndim> stride = calc_stride(shape);
+  Shape<ndim> sshape = shape;
+  sshape[axis] = 1;
+  index_t sa = stride[axis];
+
+  #pragma omp parallel for
+  for (int i = 0; i < static_cast<int>(N); ++i) {
```
apeforest: Please use `index_t` instead of `int`

src/operator/nn/softmax-inl.h
```diff
@@ -113,6 +113,60 @@ inline void Softmax(Stream<cpu> *s, DType *in, OType *out,
...
+                              Shape<ndim> shape, int axis, const DType temperature) {
+  index_t M = shape[axis];
+  index_t N = shape.Size()/M;
+  Shape<ndim> stride = calc_stride(shape);
+  Shape<ndim> sshape = shape;
+  sshape[axis] = 1;
+  index_t sa = stride[axis];
+
+  #pragma omp parallel for
+  for (int i = 0; i < static_cast<int>(N); ++i) {
```
apeforest: why cast N to `int`?

src/operator/nn/softmax-inl.h
```diff
@@ -113,6 +113,60 @@ inline void Softmax(Stream<cpu> *s, DType *in, OType *out,
...
+    IType len = length[i];
+    index_t base = unravel_dot(i, sshape, stride);
+
+    DType mmax = negate ? -in[base] : in[base];
+    DType val;
+    for (index_t j = 1; j < len; ++j) {
+      val = negate ? -in[base + j*sa] : in[base + j*sa];
+      if (mmax < val) mmax = val;
+    }
+    for (int j = len; j < M; ++j) {
```
apeforest: use `index_t` here

src/operator/nn/softmax-inl.h
apeforest: Please see my suggestion in the code. Can we just cast length[i] to index_t?

haojin2: Your original comment was "Do we really need to iterate different int types for length? Can we just cast the type to int64_t." and now you're essentially requesting something else in the code.
So regarding your new request, what's the point of casting an integer of a smaller data type to `index_t` if we're using it's original value only?

src/operator/nn/softmax-inl.h
haojin2: Why does it have to be the same data type as `M`? It's a tensor provided by the user so the user could be providing a tensor of any data type.

src/operator/nn/softmax-inl.h
apeforest: My original comment was meant for checking different int types. "iterate" may not be an accurate word here. But nonetheless the suggestion follows the same line. I.e. only use one fixed type instead of various int types.

haojin2: I think your communication was not clear enough in the first place, I've now added cast to `index_t` in some existing code too, are you now comfortable with the changes now?

apeforest: Thanks for making the changes. Sorry, if my earlier comments caused you confusion.

Looking at the current code, I see the `SoftmaxWithLength()` and `Softmax()` are very similar now. (differ by one if block https://github.com/apache/incubator-mxnet/pull/15169/files#diff-be02d7c5660bf5cd623601a501fc7abeR134). Do you think we can combine this two functions into one now? :)

haojin2: No, because your are only looking at the CPU version of the code.

apeforest: Are you referring to these two kernel functions?
https://github.com/apache/incubator-mxnet/blob/50c1c7ca7f2e6a7864e7a0aedc775ae7dd8be091/src/operator/nn/softmax-inl.h#L283
https://github.com/apache/incubator-mxnet/blob/50c1c7ca7f2e6a7864e7a0aedc775ae7dd8be091/src/operator/nn/softmax-inl.h#L339

They also only differ by a few lines. Would you think a better way to consolidate them?

python/mxnet/test_utils.py
```diff
@@ -1242,11 +1242,13 @@ def check_speed(sym, location=None, ctx=None, N=20, grad_req=None, typ="whole",
                     exe.arg_dict.items()}
     else:
         assert isinstance(location, dict), "Expect dict, get \"location\"=%s" %str(location)
-        exe = sym.simple_bind(grad_req=grad_req, ctx=ctx,
-                              **{k: v.shape for k, v in location.items()})
+        # exe = sym.simple_bind(grad_req=grad_req, ctx=ctx,
```
apeforest: Please remove commented code if not needed.

python/mxnet/test_utils.py
```diff
@@ -1242,11 +1242,13 @@ def check_speed(sym, location=None, ctx=None, N=20, grad_req=None, typ="whole",
...
+        #                       **{k: v.shape for k, v in location.items()})
+        exe = sym.bind(ctx=ctx, args=location, grad_req=grad_req)
 
-    for name, iarr in location.items():
-        exe.arg_dict[name][:] = iarr.astype(exe.arg_dict[name].dtype)
+    # for name, iarr in location.items():
```
apeforest: Please remove commented code if not needed.

haojin2: unrelated changes accidentally committed, removed now.

apeforest:
This PR is more like patching an existing function to implement a feature which is not generically needed. From software engineering practice, this PR introduced unnecessary redundancy in the code and the performance impact is not well measured. 

As we discussed offline, if this feature is urgently needed and the code has to be patched in such a way to maintain backward compatibility, it might be okay to ship as it is. However, we should add a TODO in MXNet 2.0 to rewrite softmax from scratch taking into consideration of all the required scenarios and how to make it extensible.

@szha for review and final approve.

src/operator/nn/softmax-inl.h
ptrendx: I have a separate PR  #15545 to optimize softmax GPU implementation, I can look into merging those 2 kernels in that PR.

haojin2:
Thanks all for the approval, merging this now. @ptrendx @szha @apeforest

marcoabreu:
@apeforest did you add the TODO for 2.0 somewhere? https://github.com/apache/incubator-mxnet/pull/15169#issuecomment-512582221

TaoLv:
Sorry for missing the review. Do we need to change MKL-DNN softmax to accommodate the change?

src/operator/nn/softmax.cc
```diff
@@ -92,6 +101,13 @@ Example::
```
eric-haibin-lin: No documentation for example inputs with length?

PR #15171, opened 2019-06-07
Upgrade archive utility and add back FC improvement

anirudh2290:
This reverts commit 65434886f6caa7210ed3ff39cd4e950c023d8328.
Also upgrades archive utility and fixes: https://github.com/apache/incubator-mxnet/issues/15084

## Description ##

- Archive utility version 2.26 doesn't support sizes greater than 4GB. With this FC PR and the compile flags mentioned in the issue, the libmxnet.a archive grows beyond 4 GB. 
- The PR uses a newer archive utility version which handles such archive sizes. The current utility version wont support it, please see : https://sourceware.org/bugzilla/show_bug.cgi?id=14625
- Upgrading this archive utility means that we are forcing the users to use this version of archive utility with make to build mxnet with this set of build flags.
- As discussed with @larroy offline there is a bigger problem related to code bloat which should be addressed and I have opened an issue for this : https://github.com/apache/incubator-mxnet/issues/15182
- Having said that, I am suggesting that we update archive utility and update our docs. The user impact is all developers/end users who build from source and use this combination of flags. They have to upgrade their ar utility. In ubuntu 16.04 it may not be easy (build from source is one way). I am not aware of a better solution in the short term so that we can continue adding/modifying mxnet code. Other suggestions are welcome.

### Essentials ###

### Changes ###

## Comments ##


piyushghai:
@mxnet-label-bot Add [pr-work-in-progress, Backend]

anirudh2290:
@larroy @samskalicky @sandeep-krishnamurthy @Chancebair please help review.

docs/install/build_from_source.md
```diff
@@ -180,6 +180,8 @@ More information on turning these features on or off are found in the following
 There is a configuration file for make,
 [`make/config.mk`](https://github.com/apache/incubator-mxnet/blob/master/make/config.mk), that contains all the compilation options. You can edit it and then run `make` or `cmake`. `cmake` is recommended for building MXNet (and is required to build with MKLDNN), however you may use `make` instead. For building with Java/Scala/Clojure, only `make` is supported.
 
+**NOTE:** When certain set of build flags are set, MXNet archive increases to more than 4 GB. Since MXNet uses archive internally archive runs into a bug ("File Truncated": [bugreport](https://sourceware.org/bugzilla/show_bug.cgi?id=14625)) for archives greater than 4 GB. Please use ar version 2.27 or greater to overcome this bug. Please see https://github.com/apache/incubator-mxnet/issues/15084 for more details.
```
sandeep-krishnamurthy: Can we explicitly give the flags that causes such bloat, so that, not all users building from source need to worry new archive utility?

szha: the most significant part comes from the GPU code, and it's already an explicit switch

anirudh2290: Yes most significant part comes from GPU code. If I provide set of flags that are failing today, no guarantee that tomorrow it would not bloat and not fail for a different set of flags (which includes USE_CUDA=1 and USE_CUDNN=1) and then these instructions may quickly become outdated.

edisongustavo:
I don't think it's a good idea to download a custom `binutils` and use it on CI. It might break in many unexpected ways :/

It also means that MXNet has to stop stating that it supports Ubuntu 16.04. The requirement should be lift to at least 18.04 then.

If the community agrees, then we probably should embark on the quest of upgrading the AMI to the latest Ubuntu 18.04.

@marcoabreu you might want to look at this.

ci/docker/install/ubuntu_ar.sh
```diff
@@ -0,0 +1,36 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# build and install are separated so changes to build don't invalidate
+# the whole docker cache for the image
+
```
marcoabreu: Please add set -ex

docs/install/build_from_source.md
marcoabreu: Is there a way to detect this case so users are not faced with a strange error message but specifically prompted to upgrade ar? I'm afraid that users might not see this specific part of our documentation and just get frustrated. If our build pipeline would automatically notify them (ideally ahead of time or) after the error occured, that would be beneficial to the user experience.

samskalicky:
@edisongustavo We cannot drop support for 16.04 since its LTS period ends April 2021: https://ubuntu.com/about/release-cycle

anirudh2290:
@edisongustavo As I suggested this is a short term solution to unblock the PRs. I imagine that upgrading the AMI would probably take longer and also needs more thought. Note that every day lost here is going to cause more pain for devs pushing code to MXNet.

docs/install/build_from_source.md
anirudh2290: @marcoabreu I have added a warning message for users if the ar is smaller than 2.27

anirudh2290:
@szha @marcoabreu @sandeep-krishnamurthy Is this good to merge ?

anirudh2290:
thanks @marcoabreu . will merge unless someone has objections.

Makefile
```diff
@@ -368,10 +368,32 @@ endif
...
+
+
+endef
+
+$(warning WARNING: Archive utility: ar version being used is less than 2.27.0. $n \
+		   Note that with USE_CUDA=1 flag and USE_CUDNN=1 this is known to cause problems. $n \
+		   For more info see: https://github.com/apache/incubator-mxnet/issues/15084)
+$(shell sleep 5)
+endif
+endif
```
larroy: What about the corresponding modification to CMake?

anirudh2290: we dont yet need it for cmake. we have encountered this issue only with make.

larroy: That's weird. Thanks for your answer. I would expect to have similar amount of object code with both builds.

PR #15210, opened 2019-06-11
Custom Operator Profiling Enhancement

Zha0q1:
## Description ##
fix: https://github.com/apache/incubator-mxnet/issues/15241
I have implemented the new feature. 
Need to add test cases.

### Essentials ###

### DIFF
--- a/src/engine/naive_engine.cc
+++ b/src/engine/naive_engine.cc
@@ -160,9 +160,10 @@ class NaiveEngine final : public Engine {
     this->req_completed_ = false;
     profiler::Profiler *profiler = profiler::Profiler::Get();
     NaiveOpr *opr = nullptr;
+    // Try to get a new name. Only not null for sub-operators of a custom operator
     const bool profiling = opr_name && profiler->IsProfiling(profiler::Profiler::kImperative);
     const char* d_name = profiling && opr_name ?
-                         profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) : NULL;
+                         profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) : nullptr;
     const char* display_name = (d_name ? d_name : opr_name);
     if (profiling) {
       opr = NewOperator(exec_fun, const_vars, mutable_vars,
--- a/src/engine/naive_engine.cc
+++ b/src/engine/naive_engine.cc
@@ -160,7 +160,6 @@ class NaiveEngine final : public Engine {
     this->req_completed_ = false;
     profiler::Profiler *profiler = profiler::Profiler::Get();
     NaiveOpr *opr = nullptr;
-    // Try to get a new name. Only not null for sub-operators of a custom operator
     const bool profiling = opr_name && profiler->IsProfiling(profiler::Profiler::kImperative);
     // GenerateDisplayName() will return a pointer to the correct name of the operator
     const char* display_name = profiling ?
--- a/src/engine/naive_engine.cc
+++ b/src/engine/naive_engine.cc
@@ -29,6 +29,7 @@
 #include "../profiler/profiler.h"
 #include "./openmp.h"
 #include "../common/object_pool.h"
+#include "../profiler/custom_op_profiler.h"
 
 namespace mxnet {
 namespace engine {
@@ -160,6 +161,9 @@ class NaiveEngine final : public Engine {
     profiler::Profiler *profiler = profiler::Profiler::Get();
     NaiveOpr *opr = nullptr;
     const bool profiling = opr_name && profiler->IsProfiling(profiler::Profiler::kImperative);
+    const char* d_name = profiling && opr_name ?
+                         profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) : NULL;
+    const char* display_name = (d_name ? d_name : opr_name);
     if (profiling) {
       opr = NewOperator(exec_fun, const_vars, mutable_vars,
                         prop, opr_name)->Cast<NaiveOpr>();
@@ -168,7 +172,7 @@ class NaiveEngine final : public Engine {
       if (profiler->AggregateEnabled()) {
         attrs.reset(new profiler::ProfileOperator::Attributes());
       }
-      opr->opr_profile.reset(new profiler::ProfileOperator(opr->opr_name, attrs.release()));
+      opr->opr_profile.reset(new profiler::ProfileOperator(display_name, attrs.release()));
       opr->opr_profile->start(exec_ctx.dev_type, exec_ctx.dev_id);
     }
     if (exec_ctx.dev_mask() == gpu::kDevMask) {
--- a/src/engine/naive_engine.cc
+++ b/src/engine/naive_engine.cc
@@ -162,9 +162,10 @@ class NaiveEngine final : public Engine {
     NaiveOpr *opr = nullptr;
     // Try to get a new name. Only not null for sub-operators of a custom operator
     const bool profiling = opr_name && profiler->IsProfiling(profiler::Profiler::kImperative);
-    const char* d_name = profiling && opr_name ?
-                         profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) : nullptr;
-    const char* display_name = (d_name ? d_name : opr_name);
+    // GenerateDisplayName() will return a pointer to the correct name of the operator
+    const char* display_name = profiling ?
+                              profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name)
+                              : opr_name;
     if (profiling) {
       opr = NewOperator(exec_fun, const_vars, mutable_vars,
                         prop, display_name)->Cast<NaiveOpr>();
--- a/src/engine/naive_engine.cc
+++ b/src/engine/naive_engine.cc
@@ -164,8 +164,8 @@ class NaiveEngine final : public Engine {
     const bool profiling = opr_name && profiler->IsProfiling(profiler::Profiler::kImperative);
     // GenerateDisplayName() will return a pointer to the correct name of the operator
     const char* display_name = profiling ?
-                              profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name)
-                              : opr_name;
+                               profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) :
+                               opr_name;
     if (profiling) {
       opr = NewOperator(exec_fun, const_vars, mutable_vars,
                         prop, display_name)->Cast<NaiveOpr>();
--- a/src/engine/naive_engine.cc
+++ b/src/engine/naive_engine.cc
@@ -167,7 +167,7 @@ class NaiveEngine final : public Engine {
     const char* display_name = (d_name ? d_name : opr_name);
     if (profiling) {
       opr = NewOperator(exec_fun, const_vars, mutable_vars,
-                        prop, opr_name)->Cast<NaiveOpr>();
+                        prop, display_name)->Cast<NaiveOpr>();
       opr->profiling = profiling;
       std::unique_ptr<profiler::ProfileOperator::Attributes> attrs;
       if (profiler->AggregateEnabled()) {
--- a/src/engine/threaded_engine.cc
+++ b/src/engine/threaded_engine.cc
@@ -334,17 +334,12 @@ void ThreadedEngine::PushAsync(AsyncFn fn, Context exec_ctx,
   }
 #endif
   const bool profiling = profiler_->IsProfiling(profiler::Profiler::kImperative);
-  // auto start = std::chrono::duration_cast<std::chrono::microseconds>(
-  //     std::chrono::high_resolution_clock::now().time_since_epoch()).count();
-  const char* d_name = profiling && opr_name ? profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) : NULL;
-  // auto end = std::chrono::duration_cast<std::chrono::microseconds>(
-  //     std::chrono::high_resolution_clock::now().time_since_epoch()).count();
-  // LOG(WARNING) << opr_name;
-  // LOG(WARNING) << start;
-  // LOG(WARNING) << end;
-  // LOG(WARNING) << static_cast<double>(end-start) / 1000;
-  const char* display_name = d_name ? d_name : opr_name;
-  ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars, prop, display_name, wait);
+  // Try to get a new name. Only not null for sub-operators of a custom operator
+  const char* d_name = profiling && opr_name ?
+                       profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) : NULL;
+  const char* display_name = (d_name ? d_name : opr_name);
+  ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars,
+                                 prop, display_name, wait);
   opr->temporary = true;
   Push(opr, exec_ctx, priority, profiling);
 }
--- a/src/engine/threaded_engine.cc
+++ b/src/engine/threaded_engine.cc
@@ -334,7 +334,15 @@ void ThreadedEngine::PushAsync(AsyncFn fn, Context exec_ctx,
   }
 #endif
   const bool profiling = profiler_->IsProfiling(profiler::Profiler::kImperative);
+  // auto start = std::chrono::duration_cast<std::chrono::microseconds>(
+  //     std::chrono::high_resolution_clock::now().time_since_epoch()).count();
 	const char* d_name = profiling && opr_name ? profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) : NULL;
+  // auto end = std::chrono::duration_cast<std::chrono::microseconds>(
+  //     std::chrono::high_resolution_clock::now().time_since_epoch()).count();
+  // LOG(WARNING) << opr_name;
+  // LOG(WARNING) << start;
+  // LOG(WARNING) << end;
+  // LOG(WARNING) << static_cast<double>(end-start) / 1000;
 	const char* display_name = d_name ? d_name : opr_name;
 	ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars, prop, display_name, wait);
 	opr->temporary = true;
--- a/src/engine/threaded_engine.cc
+++ b/src/engine/threaded_engine.cc
@@ -336,8 +336,8 @@ void ThreadedEngine::PushAsync(AsyncFn fn, Context exec_ctx,
   const bool profiling = profiler_->IsProfiling(profiler::Profiler::kImperative);
   // GenerateDisplayName() will return a pointer to the correct name of the operator
   const char* display_name = profiling ?
-                            profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name)
-                            : opr_name;
+                             profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) :
+                             opr_name;
   ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars,
                                  prop, display_name, wait);
   opr->temporary = true;
--- a/src/engine/threaded_engine.cc
+++ b/src/engine/threaded_engine.cc
@@ -333,9 +333,11 @@ void ThreadedEngine::PushAsync(AsyncFn fn, Context exec_ctx,
         << device_count_;
   }
 #endif
-  ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars, prop, opr_name, wait);
-  opr->temporary = true;
   const bool profiling = profiler_->IsProfiling(profiler::Profiler::kImperative);
+	const char* d_name = profiling ? profiler::CustomOpProfiler::Get()->GetDisplayName(opr_name) : NULL;
+	const char* display_name = d_name ? d_name : opr_name;
+	ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars, prop, display_name, wait);
+	opr->temporary = true;
   Push(opr, exec_ctx, priority, profiling);
 }
 
--- a/src/engine/threaded_engine.cc
+++ b/src/engine/threaded_engine.cc
@@ -287,8 +287,11 @@ void ThreadedEngine::DeleteOperator(OprHandle op) {
 
 void ThreadedEngine::Push(OprHandle op, Context exec_ctx, int priority, bool profiling) {
   BulkFlush();
-
   ThreadedOpr* threaded_opr = ThreadedOpr::CastFromBase(op);
+  if (profiling) {
+    threaded_opr->opr_name =
+        profiler::CustomOpProfiler::Get()->GenerateDisplayName(threaded_opr->opr_name);
+  }
   OprBlock* opr_block = OprBlock::New();
   opr_block->opr = threaded_opr;
 
@@ -334,12 +337,8 @@ void ThreadedEngine::PushAsync(AsyncFn fn, Context exec_ctx,
   }
 #endif
   const bool profiling = profiler_->IsProfiling(profiler::Profiler::kImperative);
-  // GenerateDisplayName() will return a pointer to the correct name of the operator
-  const char* display_name = profiling ?
-                             profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) :
-                             opr_name;
   ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars,
-                                 prop, display_name, wait);
+                                 prop, opr_name, wait);
   opr->temporary = true;
   Push(opr, exec_ctx, priority, profiling);
 }
--- a/src/engine/threaded_engine.cc
+++ b/src/engine/threaded_engine.cc
@@ -334,10 +334,10 @@ void ThreadedEngine::PushAsync(AsyncFn fn, Context exec_ctx,
   }
 #endif
   const bool profiling = profiler_->IsProfiling(profiler::Profiler::kImperative);
-  // Try to get a new name. Only not null for sub-operators of a custom operator
-  const char* d_name = profiling && opr_name ?
-                       profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) : nullptr;
-  const char* display_name = (d_name ? d_name : opr_name);
+  // GenerateDisplayName() will return a pointer to the correct name of the operator
+  const char* display_name = profiling ?
+                            profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name)
+                            : opr_name;
   ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars,
                                  prop, display_name, wait);
   opr->temporary = true;
--- a/src/engine/threaded_engine.cc
+++ b/src/engine/threaded_engine.cc
@@ -334,7 +334,7 @@ void ThreadedEngine::PushAsync(AsyncFn fn, Context exec_ctx,
   }
 #endif
   const bool profiling = profiler_->IsProfiling(profiler::Profiler::kImperative);
-	const char* d_name = profiling ? profiler::CustomOpProfiler::Get()->GetDisplayName(opr_name) : NULL;
+	const char* d_name = profiling && opr_name ? profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) : NULL;
 	const char* display_name = d_name ? d_name : opr_name;
 	ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars, prop, display_name, wait);
 	opr->temporary = true;
--- a/src/engine/threaded_engine.cc
+++ b/src/engine/threaded_engine.cc
@@ -336,16 +336,16 @@ void ThreadedEngine::PushAsync(AsyncFn fn, Context exec_ctx,
   const bool profiling = profiler_->IsProfiling(profiler::Profiler::kImperative);
   // auto start = std::chrono::duration_cast<std::chrono::microseconds>(
   //     std::chrono::high_resolution_clock::now().time_since_epoch()).count();
-	const char* d_name = profiling && opr_name ? profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) : NULL;
+  const char* d_name = profiling && opr_name ? profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) : NULL;
   // auto end = std::chrono::duration_cast<std::chrono::microseconds>(
   //     std::chrono::high_resolution_clock::now().time_since_epoch()).count();
   // LOG(WARNING) << opr_name;
   // LOG(WARNING) << start;
   // LOG(WARNING) << end;
   // LOG(WARNING) << static_cast<double>(end-start) / 1000;
-	const char* display_name = d_name ? d_name : opr_name;
-	ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars, prop, display_name, wait);
-	opr->temporary = true;
+  const char* display_name = d_name ? d_name : opr_name;
+  ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars, prop, display_name, wait);
+  opr->temporary = true;
   Push(opr, exec_ctx, priority, profiling);
 }
 
--- a/src/engine/threaded_engine.cc
+++ b/src/engine/threaded_engine.cc
@@ -336,7 +336,7 @@ void ThreadedEngine::PushAsync(AsyncFn fn, Context exec_ctx,
   const bool profiling = profiler_->IsProfiling(profiler::Profiler::kImperative);
   // Try to get a new name. Only not null for sub-operators of a custom operator
   const char* d_name = profiling && opr_name ?
-                       profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) : NULL;
+                       profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) : nullptr;
   const char* display_name = (d_name ? d_name : opr_name);
   ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars,
                                  prop, display_name, wait);
--- a/src/operator/custom/custom-inl.h
+++ b/src/operator/custom/custom-inl.h
@@ -148,7 +148,7 @@ class CustomOperator {
             ctx.async_on_complete();
           },
           ctx.run_ctx.ctx, vars, vars2, FnProperty::kNoSkip, 0,
-          "CustomOperator");
+          "Dummy_Wait");
     });
     // increase num_threads if there is not enough threads to execute custom operator
     if (q_.size() > num_free_threads_)
--- a/src/operator/custom/custom-inl.h
+++ b/src/operator/custom/custom-inl.h
@@ -80,9 +80,14 @@ class CustomOperator {
             const std::vector<NDArray>& outputs,
             const std::string op_type = "") {
     if (naive_engine_) {
-      profiler::CustomOpProfiler::Get()->OnCustomBegin(op_type);
-      func();
-      profiler::CustomOpProfiler::Get()->OnCustomEnd();
+      if (profiler::Profiler::Get()->IsProfiling(profiler::Profiler::kImperative)) {
+        profiler::CustomOpProfiler::Get()->OnCustomBegin(op_type);
+        func();
+        profiler::CustomOpProfiler::Get()->OnCustomEnd();
+      }
+      else {
+        func();
+      }
       for (size_t i = 0, out_idx = 0; i < arrs.size(); i++) {
         if (arrs[i].storage_type() == kDefaultStorage ||
             arrs[i].storage_type() == kUndefinedStorage)
@@ -101,9 +106,14 @@ class CustomOperator {
       bool prev_training = Imperative::Get()->set_is_training(training);
 
       try {
-        profiler::CustomOpProfiler::Get()->OnCustomBegin(op_type);
-        func();
-        profiler::CustomOpProfiler::Get()->OnCustomEnd();
+        if (profiler::Profiler::Get()->IsProfiling(profiler::Profiler::kImperative)) {
+          profiler::CustomOpProfiler::Get()->OnCustomBegin(op_type);
+          func();
+          profiler::CustomOpProfiler::Get()->OnCustomEnd();
+        }
+        else {
+          func();
+        }
       } catch (dmlc::Error& e) {
         exception_ =
             std::make_shared<std::exception_ptr>(std::current_exception());
--- a/src/operator/custom/custom-inl.h
+++ b/src/operator/custom/custom-inl.h
@@ -149,8 +149,7 @@ class CustomOperator {
 
             ctx.async_on_complete();
           },
-          ctx.run_ctx.ctx, vars, vars2, FnProperty::kNoSkip, 0,
-          "Dummy_Wait");
+          ctx.run_ctx.ctx, vars, vars2, FnProperty::kNoSkip, 0, "Dummy_Wait");
     });
     // increase num_threads if there is not enough threads to execute custom operator
     if (q_.size() > num_free_threads_)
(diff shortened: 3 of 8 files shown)


src/operator/custom/custom-inl.h
```diff
@@ -76,7 +77,8 @@ class CustomOperator {
             bool training, const std::vector<NDArray>& arrs,
             const std::vector<int>& tags,
             const std::unordered_set<int>& output_tags,
-            const std::vector<NDArray>& outputs) {
+            const std::vector<NDArray>& outputs,
+	          const std::string op_type = "") {
```
access2rohit: nit: indentation

Zha0q1: oh gee, fixing it now

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,93 @@
...
+
+namespace mxnet {
+namespace profiler {
+
+using Tid = std::thread::id;
+using TaskPtr = std::shared_ptr<ProfileTask>;
+
+class CustomOpProfiler {
+public:
+    static CustomOpProfiler* Get() {
```
access2rohit: You can make the constructor of the class as private. Why do you need to use a mutex ?

Zha0q1: Yeah I will make the constructor private. I think the use of mutex is to prevent two threads entering Get() at the same time and creating two instances. This is how class Profiler implements its Get()

src/engine/threaded_engine.cc
```diff
@@ -333,9 +333,11 @@ void ThreadedEngine::PushAsync(AsyncFn fn, Context exec_ctx,
         << device_count_;
   }
 #endif
-  ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars, prop, opr_name, wait);
-  opr->temporary = true;
   const bool profiling = profiler_->IsProfiling(profiler::Profiler::kImperative);
+	const char* d_name = profiling ? profiler::CustomOpProfiler::Get()->GetDisplayName(opr_name) : NULL;
```
access2rohit: nit: indentation

access2rohit: please check indentation everywhere and fix it

Zha0q1: Yeah, it's probably because the tab size is set to different values locally and on here

vandanavk:
@mxnet-label-bot add [Operator, pr-work-in-progress]

src/profiler/profiler.h
```diff
@@ -1225,6 +1250,10 @@ struct ProfileOperator : public ProfileEvent {
   static ProfileDomain domain_;
   /*! \brief Optional operator attributes */
   std::unique_ptr<Attributes> attributes_;
+
+  bool profiling_operator;
+
+  bool profiling_task;
```
anirudh2290: please add doc.

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,93 @@
...
+#include <unordered_set>
+#include <unordered_map>
+#include <thread>
+#include "./profiler.h"
+
+namespace mxnet {
+namespace profiler {
+
+using Tid = std::thread::id;
+using TaskPtr = std::shared_ptr<ProfileTask>;
```
access2rohit: Tid-> thread_id
TaskPtr -> task_ptr
Refer to Google C++ guide for naming conventions: https://google.github.io/styleguide/cppguide.html#Variable_Names

src/profiler/profiler.h
```diff
@@ -1149,8 +1158,20 @@ struct ProfileOperator : public ProfileEvent {
...
+      , profiling_operator(true)
+      , profiling_task(true) {
+    if (strcmp(name, "CustomOperator") == 0) {
+      profiling_operator = false;
+      profiling_task = false;
+    } else if (strcmp(name, "Custom") == 0) {
+      profiling_task = false;
+    } else if (std::string(name).find("::") != std::string::npos) {
+      as_task_.setDomain(&custom_op_domain);
+      SetCategories(custom_op_domain.name());
```
anirudh2290: ideally you won't  need this else if logic. when engine sets the display name the correct display name is automatically. For example if my custom ops MyCustomOp(CustomOp) and MyCustomOp2(CustomOp):
then profiling result should return MyCustomOp::elemwise_add
MyCustomOp2::elemwise_add

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,93 @@
...
+
+class CustomOpProfiler {
```
access2rohit: make constructor as private to make the class a Singleton

src/operator/custom/custom-inl.h
access2rohit: is it possible that op_type can be empty ?(i.e. the user will not specify any name for his/her custom operator ?) Since this method is only invoked for custom operators I want to understand why do we need a default values unless user can specify empty string as name for operator

access2rohit:
Even if the PR is WIP add the description and explain briefly what feature are you adding to MXNet profiler and show how it adds value over existing profiler

src/profiler/profiler.h
Zha0q1: Yes the names are correct. But we still need to set the domain here to "Custom Operator" if the operator turns out to be a "sub-operator", that is, an operator called by a custom operator

src/operator/custom/custom-inl.h
Zha0q1: op_type will not be empty. This function is called in three places: custom.cc Forward(), custom.cc Backward(), and in src/c_api/c_api_function.cc. We need to pass in op_type in the first two places. Having a default value helps avoid creating a new api.

src/engine/threaded_engine.cc
```diff
@@ -333,9 +333,11 @@ void ThreadedEngine::PushAsync(AsyncFn fn, Context exec_ctx,
...
+	const char* display_name = d_name ? d_name : opr_name;
```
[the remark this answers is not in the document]
larroy: Why use a std::string if just const char*?

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,93 @@
...
+public:
+    static CustomOpProfiler* Get() {
+        static std::mutex mtx;
+        static std::shared_ptr<CustomOpProfiler> prof = nullptr;
```
larroy: why shared_ptr and not unique_ptr

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,93 @@
...
+namespace mxnet {
+namespace profiler {
+
+using Tid = std::thread::id;
+using TaskPtr = std::shared_ptr<ProfileTask>;
+
+class CustomOpProfiler {
+public:
+    static CustomOpProfiler* Get() {
+        static std::mutex mtx;
```
larroy: why mutex is static?

src/profiler/custom_op_profiler.h
larroy: protected would be better usually in case you want to extend.

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,93 @@
...
+    tid_to_op_type_[tid] = op_type;
+    tasks_[tid] = std::make_shared<ProfileTask>(task_name.c_str(), &custom_op_domain);
+    tasks_[tid]->start();
+    }
+
+    void OnCustomEnd() {
+    Tid tid = std::this_thread::get_id();
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (tasks_.find(tid) == tasks_.end()) {
+        LOG(WARNING) << "This should never happen";
```
larroy: can you use CHECK so it's an assertion?

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,93 @@
...
+    Tid tid = std::this_thread::get_id();
+    const std::string task_name = op_type + "::pure_python" ;
+    std::lock_guard<std::mutex> lock(mutex_);
+    tid_to_op_type_[tid] = op_type;
+    tasks_[tid] = std::make_shared<ProfileTask>(task_name.c_str(), &custom_op_domain);
+    tasks_[tid]->start();
+    }
+
+    void OnCustomEnd() {
+    Tid tid = std::this_thread::get_id();
```
larroy: can be const

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+private:
+
+    /*! \brief */
+    std::mutex mutex_;
+    /* !\brief task names for sub-operators in custom ops */
+    std::unordered_set<std::string> display_names_;
+    /* */
+    std::unordered_map<Tid, TaskPtr> tasks_;
```
larroy: does it need to be shared_ptr? why not unique_ptr?

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+
+    const char* GetDisplayName(const std::string& op_type) {
+    Tid tid = std::this_thread::get_id();
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (tid_to_op_type_.find(tid) == tid_to_op_type_.end()) {
+        return NULL;
+    }
+    std::string name = tid_to_op_type_[tid] + "::" + op_type;
```
larroy: Why are we mutating the class on a method that says Get? the indentation is all messed up so difficult to follow as it is.

Zha0q1: Thanks! Will fix the issues tomorrow

src/engine/threaded_engine.cc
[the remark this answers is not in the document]
anirudh2290: I would still use std::string to keep consistent with rest of codebase. GetDisplayName should just return std::string, They are also just more maintainable and flexible.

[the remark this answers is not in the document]
Zha0q1: There is a good reason for using chat*. Because display_name is passed to the next function (which takes a char*), we would want to avoid using a local variable here so that we are not pointing to destrcuted data in the future.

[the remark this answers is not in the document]
Zha0q1: The Idea is GetDisplayName() will store the display name in the singleton object, and return a char* to that data. I need to change the function name to something like "AssignDisplayName" though to avoid confusion

src/profiler/custom_op_profiler.h
Zha0q1: I think aliases should be named this way. I referred to the code elsewhere as well

src/profiler/custom_op_profiler.h
Zha0q1: This mutex is not CustomOpProfiler's mutex. It's used when multiple threads are trying to create a singleton object at the same time

src/profiler/profiler.h
```diff
@@ -1098,6 +1102,11 @@ struct ProfileMarker {
   VTUNE_ONLY_CODE(std::unique_ptr<vtune::VTuneInstantMarker> vtune_instant_marker_);
 };
 
+
+class CustomOpProfiler;
```
arcadiaphy: Remove the declaration here.

src/profiler/profiler.h
```diff
@@ -1098,6 +1102,11 @@ struct ProfileMarker {
...
+static ProfileDomain custom_op_domain("Custom Operator");
```
arcadiaphy: Can we move custom_op_domain into class CustomOpProfiler?

src/profiler/profiler.h
arcadiaphy: Please use profiling_operator_ and profiling_task_ instead

src/profiler/profiler.h
Zha0q1: I have removed the unused forward declearation. However, I think we need to keep custom_op_domain here because if we move it into CustomOpProfiler there will be a mutual dependency issue.

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,91 @@
...
+        Tid tid = std::this_thread::get_id(); 
+        std::lock_guard<std::mutex> lock(mutex_);
+        if (tid_to_op_type_.find(tid) == tid_to_op_type_.end()) {
+            return NULL;
+        }
+        std::string op_type = std::string(op_type_ptr);
+        std::string name = tid_to_op_type_[tid] + "::" + op_type;
+        display_names_.insert(name);
+        return display_names_.find(name)->c_str();
+    }
```
arcadiaphy: Is it suitable to use tid to represent custom ops?

The custom worker thread is re-used for all custom ops, so there is no one-to-one mapping between tid and custom op. When the engine executes the sub-ops, perhaps new mappings have already been registered in `tid_to_op_type_`, how can you guarantee that correct op name is returned by `GenerateDisplayName` in thread engine?

Zha0q1: Yes, custom worker threads are reused indeed. However, we are using the mapping while we are pushing sub-operators to the engine, not while executing them, so, the mapping should remain always be to the current custom operator type

Zha0q1: The idea is: worker thread 1 pick up MyOp --> mapping from tid I to MyOp established --> executing callback, pushing sub-operators to engine, using the mapping -> call back finishes --> worker thread move onto next task form the task pool

apeforest:
Please add more description to this PR. What was done, what was tested, any limitationn

src/profiler/custom_op_profiler.h
arcadiaphy: I understand. Another thing is that you should make sure your profiling works in naive engine too.

Zha0q1: Yeah, that I will add and test. Thanks!

Zha0q1:
I have added a new test in test_profiler.py. 
I also manually tested using Naive Engine; the result was the same as using ThreadedEngine. It did not seem like adding "os.environ['MXNET_ENGINE_TYPE'] = 'NaiveEngine'" to my new test case in test_profiler.py would work through. @arcadiaphy

access2rohit:
Can you more test cases like:
customOp1 + RegisteredOp + customOp2

Try to include python processing code in all the above cases

Zha0q1:
> Can you more test cases like:
> customOp1 + RegisteredOp + customOp2
> 
> Try to include python processing code in all the above cases

Added

src/engine/naive_engine.cc
```diff
@@ -159,7 +160,11 @@ class NaiveEngine final : public Engine {
     this->req_completed_ = false;
     profiler::Profiler *profiler = profiler::Profiler::Get();
     NaiveOpr *opr = nullptr;
+    // Try to get a new name. Only not null for sub-operators of a custom operator
     const bool profiling = opr_name && profiler->IsProfiling(profiler::Profiler::kImperative);
+    const char* d_name = profiling && opr_name ?
+                         profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) : nullptr;
+    const char* display_name = (d_name ? d_name : opr_name);
     if (profiling) {
       opr = NewOperator(exec_fun, const_vars, mutable_vars,
```
larroy: Who frees this ptr?

Zha0q1: This printer is retuned by std::string.c_str(). The memory that it points to is managed by the original string and will be deallocated as it goes out of scope

Zha0q1: string's are stored in the singleton CustomOpProfiler object

larroy: I meant opr

Zha0q1: That's a great question. I tried to look it up and I also could not find where the deletion took place

apeforest: Is the d_name here really necessary? Can we simplify the logic here to just the display_name?

src/engine/threaded_engine.cc
```diff
@@ -333,9 +333,14 @@ void ThreadedEngine::PushAsync(AsyncFn fn, Context exec_ctx,
         << device_count_;
   }
 #endif
-  ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars, prop, opr_name, wait);
-  opr->temporary = true;
   const bool profiling = profiler_->IsProfiling(profiler::Profiler::kImperative);
+  // Try to get a new name. Only not null for sub-operators of a custom operator
+  const char* d_name = profiling && opr_name ?
```
apeforest: again, do we really need the variable d_name?

tests/python/unittest/test_profiler.py
```diff
@@ -269,6 +269,129 @@ def check_sorting(debug_str, sort_by, ascending):
...
+
+        def create_operator(self, ctx, shapes, dtypes):
+            return MyAdd()
+
+    file_name = 'test_custom_operator_profiling_multiple_custom_ops.json'
+    enable_profiler(profile_filename = file_name, run=True, continuous_dump=True,\
+                    aggregate_stats=True)
+    inp = mx.nd.zeros(shape=(100, 100))
+    x = inp + 1
+    y = mx.nd.Custom(inp, op_type="MyAdd1")
```
apeforest: Use single quote or double quotes consistently in your test file.

tests/python/unittest/test_profiler.py
```diff
@@ -269,6 +269,129 @@ def check_sorting(debug_str, sort_by, ascending):
...
+        def create_operator(self, ctx, in_shapes, in_dtypes):
+            return Sigmoid()
+
+    file_name = 'test_custom_operator_profiling.json'
+    enable_profiler(profile_filename = file_name, run=True, continuous_dump=True,\
+                    aggregate_stats=True)
+    x = mx.nd.array([0, 1, 2, 3])
+    x.attach_grad()
+    with mx.autograd.record():
+        y = mx.nd.Custom(x, op_type='MySigmoid')
```
apeforest: here is a single quote

src/operator/custom/custom.cc
```diff
@@ -345,7 +345,7 @@ void ForwardEx(const OpStatePtr& state, const OpContext& ctx,
             static_cast<int>(ctx.is_train),
             params.info->contexts[kCustomOpForward]));
       },
-      ctx, false, ctx.is_train, cpys, tags, output_tags, outputs);
+      ctx, false, ctx.is_train, cpys, tags, output_tags, outputs, std::string(params.op_type));
```
apeforest: params.op_type is already a std::string

src/operator/custom/custom.cc
```diff
@@ -415,7 +415,8 @@ void BackwardEx(const OpStatePtr& state, const OpContext& ctx,
         ptrs.size(), const_cast<void**>(ptrs.data()), const_cast<int*>(tags.data()),
         reinterpret_cast<const int*>(req.data()), static_cast<int>(ctx.is_train),
         params.info->contexts[kCustomOpBackward]));
-    }, ctx, false, ctx.is_train, cpys, tags, output_tags, outputs);
+    }, ctx, false, ctx.is_train, cpys, tags, output_tags, outputs,
+       "_backward_" + std::string(params.op_type));
```
apeforest: params.op_type is already a std::string

src/engine/naive_engine.cc
anirudh2290: its done by DeleteOperator

Zha0q1: I think we need to keep it. Because we need to check if the returned char* is nullptr or not

apeforest: How about this?
```suggestion
const char* display_name = profiling ? profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name) : opr_name
```

Then you move the handling of opr_name == nullptr to inside GenerateDisplayName(opr_name). i.e., when opr_name is nullptr, this function also returns a nullptr.

Zha0q1:
@anirudh2290 @arcadiaphy  I have made the requested changes. Can you help me take a look again? Thanks!

src/profiler/custom_op_profiler.h
Zha0q1: Done @arcadiaphy  Could you take a look again?

src/profiler/profiler.h
```diff
@@ -787,6 +787,10 @@ struct ProfileTask : public ProfileDuration {
     NVTX_ONLY_CODE(nvtx_duration_.reset(new nvtx::NVTXDuration(name)));
   }
 
+  void setDomain(ProfileDomain *domain) {
```
apeforest: Add `const` to ProfileDomain?

src/profiler/profiler.h
```diff
@@ -1225,6 +1251,10 @@ struct ProfileOperator : public ProfileEvent {
   static ProfileDomain domain_;
   /*! \brief Optional operator attributes */
   std::unique_ptr<Attributes> attributes_;
+  /*! \brief Whether to profile as operator */
+  bool profiling_operator_;
```
apeforest: Can we make these two bool members const and initlialize them during the class initialization? The reason being that your start() and stop() functions depend on these two variables. If during the execution of the program these two variables are modified between start() and stop(), then stop() will generate undefined behavior.

src/profiler/profiler.h
Zha0q1: We actually would not want that because domain_ is not constant

src/profiler/profiler.h
Zha0q1: That's great advice! Now I am declaring them as const and using the the initializing list

src/engine/naive_engine.cc
```diff
@@ -159,16 +160,21 @@ class NaiveEngine final : public Engine {
     this->req_completed_ = false;
     profiler::Profiler *profiler = profiler::Profiler::Get();
     NaiveOpr *opr = nullptr;
+    // Try to get a new name. Only not null for sub-operators of a custom operator
     const bool profiling = opr_name && profiler->IsProfiling(profiler::Profiler::kImperative);
+    // GenerateDisplayName() will return a pointer to the correct name of the operator
+    const char* display_name = profiling ?
+                              profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name)
```
apeforest: nit: indent

src/engine/threaded_engine.cc
```diff
@@ -333,9 +333,14 @@ void ThreadedEngine::PushAsync(AsyncFn fn, Context exec_ctx,
         << device_count_;
   }
 #endif
-  ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars, prop, opr_name, wait);
-  opr->temporary = true;
   const bool profiling = profiler_->IsProfiling(profiler::Profiler::kImperative);
+  // GenerateDisplayName() will return a pointer to the correct name of the operator
+  const char* display_name = profiling ?
+                            profiler::CustomOpProfiler::Get()->GenerateDisplayName(opr_name)
```
apeforest: nit: indent

src/profiler/profiler.h
```diff
@@ -1149,8 +1158,15 @@ struct ProfileOperator : public ProfileEvent {
     : ProfileEvent(name)
       , as_task_(name, &domain_)
       , name_(name)
-      , attributes_(attributes) {
-    SetCategories(domain_.name());
+      , attributes_(attributes)
+      , profiling_operator_(strcmp(name, "Dummy_Wait"))
+      , profiling_task_(strcmp(name, "Dummy_Wait") && strcmp(name, "Custom")) {
```
apeforest: "Dummy_Wait" is being used in multiple places. It's better to make this a constant string in the header file instead of plain string here.

src/profiler/profiler.h
```diff
@@ -1149,8 +1158,15 @@ struct ProfileOperator : public ProfileEvent {
...
+    if (std::string(name).find("::") != std::string::npos) {
```
apeforest: Maybe just use strstr()?

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  /*!
+   * \brief Called after the callback of custom operators to stop the profile task for python 
+   *        code execution time
+   */
+  void OnCustomEnd() {
+    const Tid tid = std::this_thread::get_id();
+    std::lock_guard<std::mutex> lock(mutex_);
+    CHECK(tasks_.find(tid) != tasks_.end());
```
apeforest: If this CHECK fails, the program will exit and the rest will not be executed. Would it cause any issue?

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,115 @@
...
+    if (!op_type_ptr) {
+      return nullptr;
+    }
+    Tid tid = std::this_thread::get_id();
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (tid_to_op_type_.find(tid) == tid_to_op_type_.end()) {
+      return op_type_ptr;
+    }
+    std::string op_type = std::string(op_type_ptr);
+    std::string name = tid_to_op_type_[tid] + "::" + op_type;
```
apeforest: You don't need this temp variable

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,115 @@
...
+    display_names_.insert(name);
```
apeforest: Consider directly using the return type of the insert function. Then you don't need call the following find() to return.
http://www.cplusplus.com/reference/set/set/insert/

src/profiler/custom_op_profiler.h
Zha0q1: Theoretically this check will never fail. I will add a comment to clarify that.

src/profiler/profiler.h
Zha0q1: Having a constant string in custom-inl.h will introduce a circular dependency issue. I think we should keep it this way

src/profiler/custom_op_profiler.h
apeforest: Why is this resolved, I still did not see a protected/private constructor?

src/profiler/profiler.h
```diff
@@ -1149,8 +1158,15 @@ struct ProfileOperator : public ProfileEvent {
     : ProfileEvent(name)
       , as_task_(name, &domain_)
       , name_(name)
-      , attributes_(attributes) {
-    SetCategories(domain_.name());
+      , attributes_(attributes)
+      , profiling_operator_(strcmp(name, "Dummy_Wait") && strcmp(name, "Custom"))
+      , profiling_task_(strcmp(name, "Dummy_Wait") && strcmp(name, "Custom")) {
+    if (strstr(name, "::")) {
```
apeforest: It would be great if you could turn this simple logic into a function with the explanation you have in this conversation, such as `bool IsSubCustomOperator(const char *)`. Otherwise, I am afraid it is going to be hard to comprehend for future developers.

src/profiler/profiler.h
```diff
@@ -1193,7 +1213,11 @@ struct ProfileOperator : public ProfileEvent {
       if (attributes) {
         name_.append(attributes->to_string().c_str());
       }
-      categories_.set("operator");
+      if (std::string(name).find("::") != std::string::npos) {
```
apeforest: It would be great if you could turn this simple logic into a function with the explanation you have in this conversation, such as bool IsSubCustomOperator(const char *). Otherwise, I am afraid it is going to be hard to comprehend for future developers.

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,114 @@
...
+    return prof.get();
+  }
+  /*!
+   * \brief Called before the callback of custom operators to start a profile task for python 
+   *        code execution time
+   * \param op_type The registed name of the custom operator
+   */
+  void OnCustomBegin(const std::string& op_type) {
+    const Tid tid = std::this_thread::get_id();
+    const std::string task_name = op_type + "::pure_python";
```
apeforest: Add a comment here. What does "pure_python" mean?

src/profiler/custom_op_profiler.h
apeforest: I would rather add this to the output message of CHECK(). If there is a failure, people know what the rootcause is.

src/profiler/custom_op_profiler.h
apeforest: Where is your test for naive engine? Please post the line here. thanks

src/profiler/custom_op_profiler.h
Zha0q1: In an earlier version I used protected constructor, but I revert back because make_shared only allows public constructor.

src/profiler/profiler.h
apeforest: I also feel instead of comparing name to a const string, it might be better to write a function to explicitly call out the logic. `IsCustomOperator(const char *)`. Even for now, I already forgot what the comparison with a "Dummy_Wait" string is for :)

src/profiler/profiler.h
apeforest: Given there are so many confusion about the name, it's better to write a function (with comment) to explain the logic instead of `strstr(name, "::")`

src/profiler/profiler.h
```diff
@@ -1149,8 +1158,15 @@ struct ProfileOperator : public ProfileEvent {
     : ProfileEvent(name)
       , as_task_(name, &domain_)
       , name_(name)
-      , attributes_(attributes) {
-    SetCategories(domain_.name());
+      , attributes_(attributes)
+      , profiling_operator_(strcmp(name, "Dummy_Wait") && strcmp(name, "Custom"))
+      , profiling_task_(strcmp(name, "Dummy_Wait") && strcmp(name, "Custom")) {
+    if (strstr(name, "::")) {
+      as_task_.setDomain(&custom_op_domain);
```
apeforest: Seems this line can be merged with https://github.com/apache/incubator-mxnet/pull/15210/files#diff-b2c0746deeb2e04ea9cdd8139d6b30b2R1159

src/profiler/custom_op_profiler.h
Zha0q1: I actually don't know how to test naive engine. I also left a comment asking that. I tried using os.environ['MXNET_ENGINE_TYPE'] = "NaiveEngine" but it did seem to work

src/profiler/profiler.h
apeforest: This design looks rather strange to me. @arcadiaphy 's suggestion seems more reasonable. Why would it cause mutual dependency and why can't we resolve it?

src/profiler/custom_op_profiler.h
Zha0q1: added

src/profiler/profiler.h
Zha0q1: because if we move it into CustomOpProfiling then we have to include custom_op_profiler.h in profiler.h, but custom_op_profiler.h is already including profiler.h

Zha0q1: It used in a similar fashion here: https://github.com/apache/incubator-mxnet/blob/master/src/c_api/c_api_profile.cc#L41

Zha0q1: We actually only need to use custom_op_domain in profiler.h

src/profiler/custom_op_profiler.h
Zha0q1: added!

src/profiler/custom_op_profiler.h
Zha0q1: I figured it out. Naive test cases added

src/profiler/profiler.h
apeforest: In this case you are calling as_task_ twice with the second call overwrites the previous one. Using a ternary operator during class initialization seems cleaner to me.

src/profiler/profiler.h
apeforest: Please replace this line with the function you added.

Zha0q1: fixed

src/profiler/profiler.h
Zha0q1: I still think that because line 1165 and line 1166 are basically doing the same thing, except for different objects, we should not move 1165 up. Also, I think keeping the initialization list concise is also good

src/profiler/custom_op_profiler.h
apeforest: It seems use a shared_ptr for singleton class itself is not a good design practice:

Please refer to:
https://stackoverflow.com/questions/55555451/how-to-use-stdmake-shared-with-a-singleton-instance
https://codereview.stackexchange.com/questions/13059/thread-safe-singleton-class-using-stdshared-ptr-in-c11

Why do we need make_shared on a singleton object?

src/profiler/profiler.h
apeforest: > because if we move it into CustomOpProfiling then we have to include custom_op_profiler.h in profiler.h, but custom_op_profiler.h is already including profiler.h

Why is it the case? You only move a `ProfileDomain` object into `CustomOpProfiling` class which already contains a data member `static ProfileDomain domain_`. I don't understand why it requires additional header file.

src/profiler/profiler.h
apeforest: I think a good design style is actually keep the body in constructor simple. In fact, if you do
```suggestion
      as_task_(name, IsSubOperatorOfCustom(name) ? &custom_op_domain :  &domain_
```
Then you don't even need a if/else branch for SetCategories. You can simply write:
```suggestion
      SetCategories(domain_.name())
```
because the domain_ here is already customer_domain in your if case.

src/profiler/profiler.h
```diff
@@ -1193,7 +1212,11 @@ struct ProfileOperator : public ProfileEvent {
       if (attributes) {
         name_.append(attributes->to_string().c_str());
       }
-      categories_.set("operator");
+      if (IsSubOperatorOfCustom(const char* name)) {
```
apeforest: You can't write a function definition in if clause, right? :)

src/profiler/custom_op_profiler.h
Zha0q1: Sorry there was a typo in my last comment. I used unique_ptr, but the idea is the same.
I used unique_ptr because CustomOperator used it. Also, because of thread-safety consideration, I used the double-checked locking https://www.baeldung.com/java-singleton-double-checked-locking.

src/profiler/profiler.h
Zha0q1: But that way, to access custom_op_domain, we need to add "include custom_op_profiler.h" to profiler.h

src/profiler/profiler.h
Zha0q1: Copy-pasted the code... I apologize. Will pay more attention

src/profiler/profiler.h
Zha0q1: Now there is a dedicated function

src/profiler/profiler.h
Zha0q1: Sure.

Zha0q1: domain_ is a global variable, not a member variable of ProfileOperator

src/profiler/custom_op_profiler.h
larroy: This is a good idiom, we should apply to other places.

tests/python/unittest/test_profiler.py
```diff
@@ -269,6 +270,134 @@ def check_sorting(debug_str, sort_by, ascending):
...
+        and 'MyAdd2::_plus_scalar' in target_dict['Time']['Custom Operator'] \
+        and '_plus_scalar' not in target_dict['Time']['Custom Operator'] \
+        and 'operator' in target_dict['Time'] \
+        and '_plus_scalar' in target_dict['Time']['operator']
+    profiler.set_state('stop')
+
+def test_custom_operator_profiling_naive():
+    # run the two tests above using Naive Engine
+    run_in_spawned_process(func = test_custom_operator_profiling, env = {'MXNET_ENGINE_TYPE' : "NaiveEngine"})
+    run_in_spawned_process(func = test_custom_operator_profiling_multiple_custom_ops, env = {'MXNET_ENGINE_TYPE' : "NaiveEngine"})
```
larroy: nice

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,126 @@
...
+#define MXNET_PROFILER_CUSTOM_OP_PROFILER_H_
+
+#include <string>
+#include <unordered_set>
+#include <unordered_map>
+#include <thread>
+#include "./profiler.h"
+
+namespace mxnet {
+namespace profiler {
```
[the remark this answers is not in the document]
Zha0q1: I added a check for profiler->kImperative

Zha0q1:
> Lets say an operator got pushed from forward of a custom op but it is an mxnet op. I thought we were going to prepend such ops with custom op name, for example it would say CustomOp1::elemwise_add.
> Is this skipped currently ?
We have this, it happens in PushAsync in NaiveEngine and ThreadedEngine

> Also, I think the Custom and _backward_custom in the profiling output are very confusing. I think we should remove this from profiling output.
Do you mean the custom and back_custom as we had before? I actually have removed them.

> Also, I see big gaps in the profiling output and I am wondering where it is coming from. I hope it is not the overhead of the profiler itself ? Have you verified this ?
I have compared running the same model with & without my modification, I think the overhead is not large. We could talk about the gap you saw in person

src/profiler/custom_op_profiler.h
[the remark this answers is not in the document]
anirudh2290: okay , what happens for symbolic case ? does it fallback to old custom op display for profiling or does it not display anything at all for custom op.

src/operator/custom/custom-inl.h
```diff
@@ -143,8 +149,7 @@ class CustomOperator {
 
             ctx.async_on_complete();
           },
-          ctx.run_ctx.ctx, vars, vars2, FnProperty::kNoSkip, 0,
-          "CustomOperator");
+          ctx.run_ctx.ctx, vars, vars2, FnProperty::kNoSkip, 0, "Dummy_Wait");
```
anirudh2290: Can we change Dummy_Wait -> CustomOperatorWait

src/profiler/custom_op_profiler.h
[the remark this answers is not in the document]
Zha0q1: > okay , what happens for symbolic case ? does it fallback to old custom op display for profiling or does it not display anything at all for custom op.

Currently if you set profile_imperative to False, then you will have no profiling events for pure_python, aka the forward and backward callbacks. This is the case for both symbolic and imperative mode. I think this should be fine though. Considering that within custom operators, 
 users are mostly calling regular NDArray operators in imperative mode, I think if they ever want to peek inside custom operators, they must set profile_imperative to True.

To sum up, if profile_imperative == True, uses would have all the fine grained profiling events including pure_python and sub-operators. If profile_imperative == False, they would nothing.

src/operator/custom/custom-inl.h
Zha0q1: Sure!

src/profiler/custom_op_profiler.h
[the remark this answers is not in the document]
anirudh2290: i think profile_symbolic should also have this info, otherwise users may have a symbolic model with custom op but won't see any info on custom op.

anirudh2290:
@arcadiaphy looks like your comments have been addressed. Can you please take another look ?

src/profiler/profiler.h
```diff
@@ -1240,6 +1277,8 @@ struct ProfileOperator : public ProfileEvent {
   static ProfileDomain domain_;
   /*! \brief Optional operator attributes */
   std::unique_ptr<Attributes> attributes_;
+  /*! \brief Whether to profile or not */
+  const bool profiling;
```
arcadiaphy: please change to profiling_

src/profiler/custom_op_profiler.h
```diff
@@ -0,0 +1,125 @@
...
+    if (!op_type) {
+      return nullptr;
+    }
+    Tid tid = std::this_thread::get_id();
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (tid_to_op_type_.find(tid) == tid_to_op_type_.end()) {
+      return op_type;
+    }
+    std::string name = MakeSubOperatorName(tid, op_type);
+    return display_names_.insert(name).first->c_str();
```
arcadiaphy: why do we need to insert name into unordered_set here?

Zha0q1: Because we are passing a char pointer to profiler. If we just create a local c string we will point to some garbage value when the pointer is dereferenced

src/profiler/profiler.h
Zha0q1: Sure

src/profiler/custom_op_profiler.h
arcadiaphy: Right, the rest part use string literals as opr_name to avoid your mentioned problem. I think the opr_name should use string type instead of const char* to handle dynamically generated names. Nevertheless, it's not related to your PR, I'll approve it now.

Zha0q1: Thank you!

PR #15230, opened 2019-06-13
Updating SymbolBlock.imports to support different dtypes

ThomasDelteil:
## Description ##
Previously it would be impossible to do `SymbolBlock.imports(...)` with non-fp32 parameters.
The problem was that the type inference would fail all the time and parameters types would default to fp32.
Now we use the parameters type as provided in the parameter file, by casting the parameters to them (new `dtype_source` for the `load` `_load_init` and `load_parameters` functions to complement the `cast_dtype` flag)
If the parameters are not provided we attempt an automatic type inference with fp32 input.

### Essentials ###

### DIFF
--- a/tests/python/unittest/test_gluon.py
+++ b/tests/python/unittest/test_gluon.py
@@ -94,7 +94,7 @@ def test_parameter_invalid_access():
     assertRaises(RuntimeError, p1.list_row_sparse_data, row_id)
 
 @with_seed()
-def test_paramdict():
+def test_parameter_dict():
     ctx = mx.cpu(1)
     params0 = gluon.ParameterDict('net_')
     params0.get('w0', shape=(10, 10))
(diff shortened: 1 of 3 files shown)


tests/python/unittest/test_gluon.py
```diff
@@ -2750,6 +2775,17 @@ def test_gluon_param_load():
     net.cast('float16')
     net.load_parameters('test_gluon_param_load.params', cast_dtype=True)
     mx.nd.waitall()
+    
+@with_seed()
+def test_gluon_param_load_dtype_source():
+    net = mx.gluon.nn.Dense(10, in_units=10)
+    net.initialize()
+    net.cast('float16')
```
pengzhao-intel: Would you mind adding a case to load INT8 weight?

xinyu-intel: Also suggest for adding test for loading int8 weight. BTW, I just took a quick try and this works well with my internal patch:)

vandanavk:
@mxnet-label-bot add [Gluon, pr-awaiting-review]

tests/python/unittest/test_gluon.py
ThomasDelteil: @pengzhao-intel @xinyu-intel, thanks for the suggestion, I'm actually unable to find a single layer in Gluon that supports `int8` or `uint8`. 

```python
net = mx.gluon.nn.Conv2D(channels=4, kernel_size=3, in_channels=3)
net.initialize()
net.cast(np.uint8)
net(mx.nd.ones((1,3,224,224), dtype=np.uint8))
```

gives me a 
```text
MXNetError: std::exception
```

Any suggestions without resorting to symbolically fused graphs?

ThomasDelteil: I have added a test for the `ParameterDict` that should be sufficient since it is the same mechanism as the network parameters: see https://github.com/apache/incubator-mxnet/pull/15230/files#diff-962bd5bb7248659d7eb3be37ee8a4c6bR141

ThomasDelteil:
@szha for review

ThomasDelteil:
@pengzhao-intel is that good for merging?

pengzhao-intel:
LGTM 

Sorry for the late reply.

Issue #14159, opened 2019-02-14
[Feature Request] Support fp16 for C Predict API


IvyGongoogle:
hello, there is some materials about how to train a mxnet model with fp16, 
but I can not find how to infer a batch data using fp16 by c++ api, can you give some advises?

looking forward to your reply.

lanking520:
Hi, are you planning to train on C++? Currently, we do have Python supporting float16 if you initialize the dtype to that.

simonmaurer:
interested as well.
@lanking520 so what you're saying is that C++ inference using float16 is not yet possible? is that also true for the C predict API?

lanking520:
Grab someone may know about C++: @leleamol here

IvyGongoogle:
@lanking520  I use python to train a mxnet  model (such as resnet-50 ) with `--dtype=float16`，but I want to using C++ api to infer this trained fp16 model. So there are some documents about how to do it?

eric-haibin-lin:
@leleamol

IvyGongoogle:
I use [example/image-classification/predict-cpp/image-classification-predict.cc](https://github.com/apache/incubator-mxnet/blob/master/example/image-classification/predict-cpp/image-classification-predict.cc) to infer this trained fp16 model, but it can not work and shows error:

```
Segmentation fault: 11

Stack trace returned 8 entries:
[bt] (0) /xxx/libmxnet.so(+0x431b7a) [0x7f19e23f6b7a]
[bt] (1) /xxx/libmxnet.so(+0x3503096) [0x7f19e54c8096]
[bt] (2) /lib64/libc.so.6(+0x35270) [0x7f19e1181270]
[bt] (3) /xxx/libmxnet.so(MXPredSetInput+0x74) [0x7f19e4c08d64]
[bt] (4) ./image-classification-predict() [0x403fb3]
[bt] (5) ./image-classification-predict() [0x402b8a]
[bt] (6) /lib64/libc.so.6(__libc_start_main+0xf5) [0x7f19e116dc05]
[bt] (7) ./image-classification-predict() [0x402e98]
```

kwanking:
have you solved it?

> @lanking520 I use python to train a mxnet model (such as resnet-50 ) with `--dtype=float16`，but I want to using C++ api to infer this trained fp16 model. So there are some documents about how to do it?



[...]

Issue #13335, opened 2018-11-20
MXPredCreate lost dtype


yajiedesign:
## Description
when use MXPredCreate with int8.it will  Type inconsistent error.
 I find https://github.com/apache/incubator-mxnet/blob/master/src/c_api/c_predict_api.cc#L213 cause data dtype to lose.And then cause Type inconsistent.

## Environment info (Required)
----------Python Info----------
Version      : 3.6.0
Compiler     : MSC v.1900 64 bit (AMD64)
Build        : ('default', 'Dec 23 2016 11:57:41')
Arch         : ('64bit', 'WindowsPE')
------------Pip Info-----------
Version      : 18.1
Directory    : D:\Anaconda3\lib\site-packages\pip
----------MXNet Info-----------
Version      : 1.3.1
Directory    : G:\deeplearn\mxnet\python\mxnet
Hashtag not found. Not installed from pre-built package.
----------System Info----------
Platform     : Windows-10-10.0.17134-SP0
system       : Windows
node         : DEVEGA
release      : 10
version      : 10.0.17134

Package used (Python/R/Scala/Julia):C api

## Build info (Required if built from source)

Compiler (gcc/clang/mingw/visual studio):visual studio

MXNet commit hash:
017941e6ca177a15ec65d8bd35f7ccd20ef2b30d

Build config:

## Error Message:
Type inconsistent, Provided = float32, inferred type = int8

## Minimum reproducible example
`  std::string json_file = "int8-symbol.json";
  std::string param_file = "int8-0000.params";
  BufferFile json_data(json_file);
  BufferFile param_data(param_file);

  mx_uint num_input_nodes = 1;
  const char* input_key[1] = { "data" };
  const char** input_keys = input_key;

  const mx_uint input_shape_indptr[2] = { 0,4 };
  const mx_uint input_shape_data[4] = { 1,static_cast<mx_uint>(3) ,static_cast<mx_uint>(1080),static_cast<mx_uint>(1920) };

  PredictorHandle pred_hnd = 0;

  int dev_type = 1;
  int dev_id = 1;

  MXPredCreate(
    static_cast<const char*>(json_data.GetBuffer()),
    static_cast<const char*>(param_data.GetBuffer()),
    static_cast<int>(param_data.GetLength()),
    dev_type,
    dev_id,
    num_input_nodes,
    input_keys,
    input_shape_indptr,
    input_shape_data,
    &pred_hnd);
`

piyushghai:
[...]

Issue #7681, opened 2017-08-31
ValueError: Data provided by data_shapes don't match names specified by data_names ([] vs. ['data'])


jaewoosong:
For bugs or installation issues, please provide the following information.
The more information you provide, the more likely people will be able to help you.

## Environment info
Operating System: Ubuntu 16.04

Compiler: Python 3.6.1 (Anaconda 4.4.0, 64bit), installed by pip

Package used (Python/R/Scala/Julia): Python

MXNet version: 0.11.0

Python version and distribution: Python 3.6.1 (Anaconda 4.4.0, 64bit), installed by pip

## Error Message:
```
Traceback (most recent call last):
  File "/home/jaewoo/temp2.py", line 63, in <module>
    mod.fit(data_iter, num_epoch=5)
  File "/home/jaewoo/Programs/anaconda3/lib/python3.6/site-packages/mxnet/module/base_module.py", line 487, in fit
    self.forward_backward(data_batch)
  File "/home/jaewoo/Programs/anaconda3/lib/python3.6/site-packages/mxnet/module/base_module.py", line 191, in forward_backward
    self.forward(data_batch, is_train=True)
  File "/home/jaewoo/Programs/anaconda3/lib/python3.6/site-packages/mxnet/module/module.py", line 594, in forward
    self.reshape(new_dshape, new_lshape)
  File "/home/jaewoo/Programs/anaconda3/lib/python3.6/site-packages/mxnet/module/module.py", line 457, in reshape
    self.data_names, self.label_names, data_shapes, label_shapes)
  File "/home/jaewoo/Programs/anaconda3/lib/python3.6/site-packages/mxnet/module/base_module.py", line 71, in _parse_data_desc
    _check_names_match(data_names, data_shapes, 'data', True)
  File "/home/jaewoo/Programs/anaconda3/lib/python3.6/site-packages/mxnet/module/base_module.py", line 63, in _check_names_match
    raise ValueError(msg)
ValueError: Data provided by data_shapes don't match names specified by data_names ([] vs. ['data'])
```

## Minimum reproducible example
```
import mxnet as mx
import os
import subprocess
import numpy as np
import matplotlib.pyplot as plt
import tarfile

import warnings
warnings.filterwarnings("ignore", category=DeprecationWarning)

class SimpleIter(mx.io.DataIter):
    def __init__(self, data_names, data_shapes, data_gen,
                 label_names, label_shapes, label_gen, num_batches=10):
[...]

PR #15245, opened 2019-06-14
FP16 Support for C Predict API

anirudh2290:
## Description ##
Adds dtype support for C Predict API.
Fixes: https://github.com/apache/incubator-mxnet/issues/14159 and https://github.com/apache/incubator-mxnet/issues/13335

### Essentials ###

### Changes ###

## Comments ##


vandanavk:

amalgamation/python/mxnet_predict.py
```diff
@@ -133,15 +199,38 @@ def __init__(self, symbol_file,
...
+        provided_arg_type_names = ctypes.POINTER(ctypes.c_char_p)()
+        # provided types
+        provided_arg_type_data = ctypes.POINTER(mx_uint)()
+        if type_dict is not None:
+            provided_arg_type_names = []
+            provided_arg_type_data = []
+            for k, v in type_dict.items():
+                v = np.dtype(v).type
+                if v in _DTYPE_NP_TO_MX:
+                    provided_arg_type_names.append(k)
```
sandeep-krishnamurthy: Here we are depending on index of the element? I remember we had issues due to dependence on position in the past due to different in Python list maintaining elements in Python2 / 3. Should we use a map instead here (name -> type)?

amalgamation/python/mxnet_predict.py
```diff
@@ -160,10 +249,18 @@ def forward(self, **kwargs):
...
+            raise ValueError("number of kwargs should be same as len of type_dict" \
+                             "Please check your forward pass inputs" \
+                             "or type_dict passed to Predictor instantiation")
+
         for k, v in kwargs.items():
             if not isinstance(v, np.ndarray):
                 raise ValueError("Expect numpy ndarray as input")
-            v = np.asarray(v, dtype=np.float32, order='C')
+            if self.type_dict and k in self.type_dict:
+                v = np.asarray(v, dtype=self.type_dict[k], order='C')
```
sandeep-krishnamurthy: Can you help me understand importance of Column major here.

amalgamation/python/mxnet_predict.py
```diff
@@ -160,10 +249,18 @@ def forward(self, **kwargs):
...
+            else:
+                v = np.asarray(v, dtype=np.float32, order='C')
```
sandeep-krishnamurthy: if user provided type is not supported by MXNet, it silently converts to FP32, is this expected?

sandeep-krishnamurthy: nit: Will be better to keep all the dtype, including default ie., np.float32 in the map you are maintaing and remove all explicit np.dtype ?

include/mxnet/c_predict_api.h
```diff
@@ -85,6 +85,44 @@ MXNET_DLL int MXPredCreate(const char* symbol_json_str,
...
+ * \param provided_arg_dtypes
+ *    The provided_arg_dtypes the dtype provided
+ * \param out The created predictor handle.
+ * \return 0 when success, -1 when failure.
+ */
+MXNET_DLL int MXPredCreateEx(const char* symbol_json_str,
+                             const void* param_bytes,
+                             int param_size,
+                             int dev_type, int dev_id,
+                             mx_uint num_input_nodes,
```
sandeep-krishnamurthy: Why this params are non constant?

src/c_api/c_predict_api.cc
```diff
@@ -187,21 +206,41 @@ int _CreatePartialOut(const char* symbol_json_str,
...
     bool infer_complete = (g.GetAttr<size_t>("shape_num_unknown_nodes") == 0);
+    // This is tricky for AMP Use case, for example, with only weights input types
+    // cannot be inferred in AMP. Thus for AMP converted model type_dict will be
+    // required
+    bool infer_type_complete = (g.GetAttr<size_t>("dtype_num_unknown_nodes") == 0);
     CHECK(infer_complete)
       << "The shape information of is not enough to get the shapes";
+    CHECK(infer_type_complete)
+        << "The type information is not enough, please provide input arg_types "
+           "with provided_arg_dtype_names and provided_arg_dtypes";
```
sandeep-krishnamurthy: I think this will not be clear for an MXNet user, he is not setting any provided_arg_dtype_names and provided_arg_dtypes parameters, so if something fails, how to debug?

src/c_api/c_predict_api.cc
```diff
@@ -210,19 +249,31 @@ int _CreatePartialOut(const char* symbol_json_str,
...
     if (arg_params.count(arg_names[i]) != 0) {
       CopyFromTo(arg_params[arg_names[i]], &nd);
     }
     arg_arrays.push_back(nd);
   }
+
   for (size_t i = 0; i < aux_shapes.size(); ++i) {
-    NDArray nd = NDArray(aux_shapes[i], ctx);
+    NDArray nd;
+    if (result_aux_types[i] != -1) {
```
sandeep-krishnamurthy: Since we are doing such check on types throughout, can we otherwise think of a setting a default DType for all params if users are not providing arg types params? So we could get rid of all these checks and always types will be available. Thoughts?

src/c_api/c_predict_api.cc
```diff
@@ -444,6 +538,20 @@ int MXPredGetOutputShape(PredictorHandle handle,
   API_END();
 }
 
+int MXPredGetOutputType(PredictorHandle handle,
+                        mx_uint out_index,
+                        int* out_dtype) {
+  MXAPIPredictor* p = static_cast<MXAPIPredictor*>(handle);
+  API_BEGIN();
+  CHECK_LT(out_index, p->out_arrays.size())
+    << "Index exceed number of outputs";
```
sandeep-krishnamurthy: nit: Can we make the message more easy to comprehend?

amalgamation/python/mxnet_predict.py
anirudh2290: I think this row-major only. Converting all inputs to row major format, which mxnet expects.

amalgamation/python/mxnet_predict.py
anirudh2290: if not supported type is used it will go into if clause and fail. it will silently convert to FP32 to if not provided.

anirudh2290: not sure i understand. can you elaborate.

include/mxnet/c_predict_api.h
anirudh2290: This is just to keep the same interface as MXPredCreate.

src/c_api/c_predict_api.cc
anirudh2290: hmm, i thought about this, but this is not limited to the python user, C Predict API can be used from Python, CPP etc, calling _CreatePartialOut and therefore i used provided_arg_dtype_names and provided_arg_dtypes which are the API param names.

src/c_api/c_predict_api.cc
anirudh2290: This was added as I was foreseeing the AMP change. and didnt even have the 
```
    CHECK(infer_type_complete)
        << "The type information is not enough, please provide input arg_types "
           "with provided_arg_dtype_names and provided_arg_dtypes";
```
Will add a test for AMP for C Predict API and then get back here.

src/c_api/c_predict_api.cc
anirudh2290: okay sure.

amalgamation/python/mxnet_predict.py
sandeep-krishnamurthy: This is one related - https://github.com/apache/incubator-mxnet/issues/7681

It was mostly in data/label shapes and names in Module.

include/mxnet/c_predict_api.h
sandeep-krishnamurthy: Agree. But, I was trying to understand why not const and what is the reason.

src/c_api/c_predict_api.cc
sandeep-krishnamurthy: I don't have a good suggestion, but, this looks very cryptic for a user.

src/c_api/c_predict_api.cc
sandeep-krishnamurthy: Thanks.

amalgamation/python/mxnet_predict.py
anirudh2290: i think the issue is unrelated and limited to module API implementation in python frontend.. i think passing the map this way should be fine.

src/c_api/c_predict_api.cc
anirudh2290: i have added the tests now.

src/c_api/c_predict_api.cc
anirudh2290: I have improved the error message, providing additional ifno for amalgamation python frontend users.

include/mxnet/c_predict_api.h
anirudh2290: const int over int gives little benefits. For example, for callers it doesnt mean anything since the function is going to copy the param. It has the benefit of making it more difficult to use a modified version of the int passed to the function, but its not hard to work around this. Looking at the history, I am guessing in the early days of mxnet the const ints were avoided in the project, and it became used more later on. To make this consistent I have changed this to cons int.

simonmaurer:
thank you!

amalgamation/python/mxnet_predict.py
sandeep-krishnamurthy: I meant, wherever there are defaults we explicitly call 'np.float32'. In other places we use a common lookup with _DTYPE_MX_TO_NP. So I was suggesting if you want to consider moving np.float32 as a default in these (_DTYPE_MX_TO_NP) look up table so there are no hard coded np.float32.

anirudh2290: we use the lookup table to map the dtype int, which comes from mxnet backend to numpy dtype. For cases, we are not determining it from backend dtype I think it is better to use np.float32 rather than _DTYPE_MX_TO_NP[0].

PR #15249, opened 2019-06-14
Sort benchmark output per op name, add some validation to CLI

larroy:
## Description ##

The output now is easier to compare when run multiple times as before the output was in random order (python dict order)

@sandeep-krishnamurthy

### DIFF
--- a/benchmark/opperf/opperf.py
+++ b/benchmark/opperf/opperf.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
+#
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
 # distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


vandanavk:
@mxnet-label-bot add [pr-awaiting-review, Performance]

benchmark/opperf/opperf.py
```diff
@@ -108,26 +113,33 @@ def _parse_mxnet_context(ctx):
...
 
     # 3. PREPARE OUTPUTS
-    save_to_file(final_benchmark_results, user_options.output_file, user_options.output_format)
+    save_to_file(final_benchmark_results, args.output_file, args.output_format)
 
     # 4. Generate list of MXNet operators not covered in benchmarks
     ops_not_covered = get_operators_with_no_benchmark(final_benchmark_results.keys())
     for idx, op in enumerate(ops_not_covered):
         print(f"{idx}. {op}")
+    return 0
```
apeforest: Is return 0 needed for python?

larroy: No, it's because of the wrap in sys.exit below. I always do this in scripts to be more explicit about the return value in error conditions.

PR #15280, opened 2019-06-19
Fixed C++ inference tutorial

NRauschmayr:
## Description ##
Tutorial did not properly preprocess images and as such produce incorrect predictions. The example would set default values for mean but not for std_dev. So if the user does not provide the file that contains those values, the example would crash. The C++ example predicts now the same class as in the Python example. However, the probabilities differ (9.72 versus 10.411) and we should to look into that

### DIFF
--- a/cpp-package/example/inference/inception_inference.cpp
+++ b/cpp-package/example/inference/inception_inference.cpp
@@ -41,6 +41,11 @@ using namespace mxnet::cpp;
 static mx_float DEFAULT_MEAN_R = 123.675;
 static mx_float DEFAULT_MEAN_G = 116.28;
 static mx_float DEFAULT_MEAN_B = 103.53;
+
+static mx_float DEFAULT_STD_R = 0.229;
+static mx_float DEFAULT_STD_G = 0.224;
+static mx_float DEFAULT_STD_B = 0.225;
+
 /*
  * class Predictor
  *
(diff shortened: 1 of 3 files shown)


Roshrini:
@NRauschmayr Thank you for fixing the example. Can you please fix failing CI? 
@leleamol Can you take a look?

NRauschmayr:
I think we have to update the unit-test: it is using an outdated and not maintained model https://github.com/apache/incubator-mxnet/blob/a9458caec285d3c28d16588fef6aef5e219c7472/cpp-package/example/inference/unit_test_inception_inference.sh 
The test should rather use a model from the official model zoo. According to https://mxnet.incubator.apache.org/api/python/gluon/model_zoo.html pre-processing has to be done the following way 
```
image = image/255
normalized = mx.image.color_normalize(image,
                                      mean=mx.nd.array([0.485, 0.456, 0.406]),
                                      std=mx.nd.array([0.229, 0.224, 0.225]))
```

karan6181:
@NRauschmayr Could you please resolve the merge conflicts??

@nswamy @szha Could you please review it? Thanks!

piyushghai:
@NRauschmayr Can you rebase with the latest master please ? 

@leleamol Gentle ping for review.

NRauschmayr:
Sorry for the late reply. I have not done the rebase yet, because I saw that the files I modified were deleted by the following commit https://github.com/apache/incubator-mxnet/pull/15164 I wanted to check first whether or not it makes sense to re-add those files, but I did not have the time yet to compile the latest c++ tutorial of the master branch.

szha:
@mxnet-bot run ci [all]

szha:
OK looks like this part is already refactored into the imagenet_inference.cpp example and it's handling the mean values correct. Still, thanks for proposing the patch.

Issue #15198, opened 2019-06-10
[Feature] Need a debugging utility to dump out intermediate values of Tensor object


apeforest:
Currently, when debugging a MXNet operator, there is no existing utility function to dump out intermediate Tensor and NDArray object. 

Developers have to write their own debugging code which is inefficient and error prone.

We need an easy to use C++ debugging utility to help developers debug MXNet operators in backend.

anirudh2290:
Does set_monitor_callback in executor work for your use case ?

ptrendx:
Monitor callback is good (although not really documented properly - I had to look at the internals of monitor implementation in Module API to know what the Python function should do to access the NDArray), but it only works for Symbol API, not for Gluon, which makes it really hard to debug backward pass/both passes of hybridized models.

larroy:
FYI I'm working on this.

PR #15285, opened 2019-06-20
Graph dumper

larroy:
## Description ##

Utility to dump the computational graph, specially during backward for human consumption.

Dumps the graph to a dot file that can be rendered.

This is intended for developers, although if there's demand it can be exposed easily from python to get the graph from any variable if is_recording is true. It is done from C++ directly, no Python is required.

Added an environment variable which triggers the logic.

It should have no performance impact when it's not enabled as there's only one additional boolean check during backward, it access the environment only once.

@apeforest @anirudh2290 

### Essentials ###


larroy:
https://github.com/apache/incubator-mxnet/issues/15198

src/common/directed_graph.h
```diff
@@ -0,0 +1,201 @@
+  /*
```
anirudh2290: Is this PR trying to address: #15198 . My understanding was that this would be something on the lines of set_monitor_callback for gluon to monitor intermediate nodes in the graph. This seems to be addressing a different problem of IR visualization. For the forward pass, the SaveJSON coupled with visualization already seems to be doing the trick.(which can be extended for backward too if needed) Also , I see TVM has a PrintGraphIR nnvm pass that we can leverage. Do we really need this ?

apeforest:
Hi @larroy could you please paste a snapshot of what the dumped graph looks like? Thanks

src/common/directed_graph.h
larroy: Good question, the intention is only to dump the graph in Backward to dot file or json, so to be precise is only tangential to the original issue. I haven't used set_monitor_callback, but from a quick look seems it's bound to an executor. How would you use it for backward in imperative? I'm not sure where to get an executor in that case.

where is the SaveJSON pass defined? I'm having trouble finding it now due to the lack of types: 

https://github.com/dmlc/tvm/blob/21935dcbf56ad3bd66ebff9891a6bc3865b8106d/nnvm/include/nnvm/pass_functions.h#L59

Could you point to the implementation of SaveJSON pass and the visualization code? Does it convert to dot?

anirudh2290: https://github.com/dmlc/tvm/blob/21935dcbf56ad3bd66ebff9891a6bc3865b8106d/nnvm/src/pass/saveload_json.cc#L314 . You can find visualization code here, It uses digraph to save it in dot format: https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/visualization.py

anirudh2290: > Good question, the intention is only to dump the graph in Backward to dot file or json, so to be precise is only tangential to the original issue. I haven't used set_monitor_callback, but from a quick look seems it's bound to an executor. How would you use it for backward in imperative? I'm not sure where to get an executor in that case.

How to add something set_monitor_callback would require a deepdive but I think it would be somewhere in the cached_op. Another question is in gluon would you want to monitor the tensors at the block level or the symbol level. That also will change things.

larroy: As discussed today, I will only dump the graph and not tensor values. So dumping tensor values it out of scope for this PR. So I won't deal with monitor callback.

anirudh2290: can you summarize why we need this ? if its to visualize the backward pass, why cant we extend the tools that we have ?

larroy: I thought a bit about this and I think you have a valid point, while we could serialize to json using the pass that you mention. Is not useful to call python code to visualize backward, since we want this from inside the engine. So I think for us now makes sense to have a couple of files with the logic to do the custom dump how we want.

I think is desirable to dump to DOT and not only JSON. Also the Symbol2JSONGraph that ends up being called from the links you provided is in the tvm subrepo and I think is best to have the code for the dump in the mxnet repo for modifications.

larroy:
@apeforest I will paste the output next time I run it, right now I'm working on something else. It outputs a dot graph as in the comment here: https://github.com/apache/incubator-mxnet/pull/15285/files#diff-884210dca53fb4c4ac20ca07aa8ecfdfR82

src/common/directed_graph.h
anirudh2290: I guess then, the purpose of the PR is to add visualization tool which would directly dump to dot format instead of using the frontend tool to do a followup pass on the symbol json. I think adding so much more code to avoid one line is a bit too much. I am not blocking this though since we are not exposing this to customer but I think it is not needed.

larroy: What do you suggest exactly? using directly Symbol2JSONGraph?

larroy: I'm just adding a generic graph class with unit tests. I think it might be useful to have as it's generic.

anirudh2290: Create a dummy graph with same outputs as the graph you want to visualize and call SaveJSON on it. Visualize it with python visualization tool.

larroy:
@apeforest 

http://bit.ly/2X6QFYS

src/common/directed_graph.h
larroy: I don't this this is practical as it uses symbol plus needs additional python. I think in this case my code does the job in pure C++ without any needs for python. After considering your suggestion I think it's best to use this C++ approach. The graph is generic, reusable and tested so I don't see a big problem with this.

larroy:
Named the head grads
graphviz (3)

larroy:
First gradient, and second gradient:

```
Forward graph: 
digraph G {
  "var0" -> "log10 node_0"
}
Backward graph: 
digraph G {
  "head grad #0" -> "_backward_log10 node_0_backward"
  "var0" -> "log10 node_0"
  "var0" -> "_backward_log10 node_0_backward"
}
Forward graph: 
digraph G {
  "head grad #0" -> "_backward_log10 node_1"
  "var0" -> "_backward_log10 node_1"
}
Backward graph: 
digraph G {
  "head grad #0" -> "elemwise_mul node_1_backward_grad_grad_inp"
  "head grad #0" -> "_backward_log10 node_1"
  "var0" -> "_backward_log10 node_1"
  "var0" -> "reciprocal node_1_dlogx"
  "_backward_log10 node_1" -> "elemwise_mul node_1_d2ydx2_mid"
  "reciprocal node_1_dlogx" -> "elemwise_mul node_1_d2ydx2_mid"
  "elemwise_mul node_1_d2ydx2_mid" -> "negative node_1_d2ydx2"
  "negative node_1_d2ydx2" -> "elemwise_mul node_1_backward_grad_grad_inp"
}
```

apeforest:
@larroy Thanks for this contribution. The dumped graph does help developers understand more details about computation graph in forward + backward passes. 

However, I also agree with @anirudh2290 there are many overlapping and orthogonal work in this PR compared with existing MXNet/NNVM code base. Can you keep this utility and graph class in a different repo instead of merging it to MXNet. I am afraid having different utilities to do highly similar work may cause more confusion to users. Please let me know if this makes sense.

larroy:
@apeforest How can we keep it in a different repo if we have to call it inside backward?  You mean a private branch?

There's indeed some overlap but it's simpler to have a C++ only component that dumps the graph for debug instead of going through non-specific JSon -> having to introduce python. Also that overlap is with NNVM which is an external repo.

I would vote to have this merged which I though twas the original intention and as @anirudh2290 is not blocking it. 

The PR is ready missing fixing a minor CI issue with amalgamation. If it's not going to be merged I will abandon it.

ptrendx:
I also do not understand why you need a whole new Graph class for the purpose of serializing to dot. nnvm::Graph and nnvm::IndexedGraph should give you everything you need for that, no?

larroy:
@ptrendx I didn't see code that dumps directly to dot, but to Json as discussed above. If you guys have such a problem with introducing a graph class which has unit tests and all, then don't merge this, no problems. I implemented it like this and for me is useful. 🤷🏻‍♂️

ptrendx:
@larroy There is no need to get so defensive about this, I (and I assume others as well) just want to make your contribution better as a result of the review process.
Going back to this graph class - the problem I have with introducing it is that I do not see any value added by it, as (I think) all of the functionality needed is already there in nnvm::Graph and IndexedGraph, while it makes the codebase less readable and maintainable. Do you have an example of functionality that is required but is not covered by the existing classes?

ptrendx:
Also, to clarify - I do not think that reusing JSON dump and other utilities is a good approach to this. My concern is only about the new graph class itself instead of using existing tools. Removing this class would make this PR 400 lines smaller.

larroy:
@ptrendx not getting defensive, that was not my intention, let's not read between the lines. Feedback should be concrete an actionable for efficient use of everybody's time, that's why I indicated that if the feedback is unspecific and not clearly positive with this change I'm fine with closing the PR in interest of everyones time instead of letting it rot here or having a long discussion.

I think there's many ways to implement something. I spent some effort implementing this to help us introspecting backward graphs, probably you could implement it in different ways, but this is the way I did it and I think is clean, concise, generic and has C++ unit tests.

I think having a generic data structure like the one I introduced in this PR is better than using a non-generic structure into which implementation concerns are mixed with the data structure over and over in the implementation itself. This is in a similar but more concerning way to having a linked list by having the next pointer inside the object or as a generic class like std::list.  So for me having generic data structures make the code more maintainable. 

For feedback to be useful it has also to be concrete, in this case I don't see the feedback to my code change proposal is concrete enough for me to take actions. Could you post an example of how would you propose to use indexed graph to do the same effect? Maybe that would help keep the conversation focused. It's also a debugging feature, so I don't feel specially passionate about defending my design decisions here. 

Adding this functionality was requested by @apeforest @kshitij12345 and I while adding higher oder gradient.
I spent some time adding unit tests and making it generic enough that can be re-used for different purposes, so please understand my surprise when there are unspecific questions like why do we need this or why introduce a small single-header data structure.

Thanks for your comments.

ptrendx:
> For feedback to be useful it has also to be concrete, in this case I don't see the feedback to my code change proposal is concrete enough for me to take actions. Could you post an example of how would you propose to use indexed graph to do the same effect? Maybe that would help keep the conversation focused. It's also a debugging feature, so I don't feel specially passionate about defending my design decisions here.

Sure, so I would do this serialization more or less like this pseudo-code:
```
// g - the nnvm::Graph that is being serialized
idx = g.indexed_graph();
for (i = 0; i < idx.num_nodes(); ++i) {
const auto& node = g[i];
for (const auto& input : node.inputs()) {
// input.source is the src node
// node is the dst node
dump "src -> dst" to dot
}
for (const auto& output : idx.outputs) {
// output.source is the src node
// "Output N" is the dst node
dump "src->dst" to dot
}
}
```

larroy:
@ptrendx Thanks

larroy:
It's ironic, the amount of questionable code that has gone in the repo, and having a single header data structure with tests is a big issue. No worries, I will remove it.

larroy:
Just curious, am I the only one who sees the design problems of the indexed graph class?

larroy:
@anirudh2290 @ptrendx @szha please review again, I removed the offending code.

include/mxnet/imperative.h
```diff
@@ -160,8 +160,8 @@ class Imperative {
 
  private:
   friend class NDArray;
-  /*! \brief make constructor protected. */
-  Imperative() {
+  Imperative() :
+    backward_graph_dump_enabled_(dmlc::GetEnv("MXNET_BACKWARD_GRAPH_DUMP", false)) {
```
apeforest: Why do we need a class member for a debugging utility?

include/mxnet/imperative.h
```diff
@@ -191,6 +191,7 @@ class Imperative {
   std::atomic<uint64_t> variable_count_{0};
   /*! \brief default backward bulk size */
   int backward_bulk_size_{0};
+  bool backward_graph_dump_enabled_{};
```
apeforest: not needed.

src/imperative/imperative.cc
```diff
@@ -353,6 +356,11 @@ std::vector<NDArray*> Imperative::Backward(
         << "There are no inputs in computation graph that require gradients.";
   }
 
+  if (backward_graph_dump_enabled_) {
+    std::cout << "Forward graph: " << std::endl;
```
apeforest: Can we leave this to the developer? I think we can just have a function DebugGraph() and not putting all the std::cout logic in a function block.

include/mxnet/imperative.h
larroy: to access the environment only once.

src/imperative/imperative.cc
larroy: ok

larroy:
@apeforest I'm confused by your latest review, have you read the comments above? there's sample output, and samples from the graph, please scroll.

About LogMemoryPlan, I don't think the objective of that function was dumping the graph. I think it served a different purpose.

src/nnvm/graph_dump.h
```diff
@@ -0,0 +1,43 @@
...
+/*!
+ * Copyright (c) 2019 by Contributors
+ * \file graph_dump.h
+ * \brief Utilities to introspect and print the execution graph
+ * \author Pedro Larroy
+ */
+
+#ifndef NNVM_GRAPH_DUMP_H_
+#define NNVM_GRAPH_DUMP_H_
+#pragma once
```
ptrendx: Those are 2 ways to do the same thing (and pragma once is a nonstandard one).

src/nnvm/graph_dump.cc
```diff
@@ -0,0 +1,83 @@
...
+
+using std::vector;
+using std::string;
+using namespace std;
+
+using namespace nnvm;
+
+namespace {
+
+std::string unnamed(const std::string &s) {
```
ptrendx: `unnamed` name does not really fit to what the purpose of this function is. Also, the logic here makes your graph potentially incorrect if a person used the same name (or the same lack of name) for multiple nodes (which is unfortunately allowed by MXNet).
What I would prefer is a function named something like `DeduplicateName` where you would check whether the same name was not used before for different Node and add `#number` to the name (adding `unnamed` or better `_unnamed_` would be fine there too).

src/nnvm/graph_dump.h
larroy: good point, I think I tried pragma once first as is supported by all major compilers but I'm not sure if nvcc supports it.

src/nnvm/graph_dump.cc
larroy: Thanks, makes sense.

include/mxnet/imperative.h
larroy: replied above. I think is more efficient and better debuggability to have a flag than to repeatedly pull the environment.

src/nnvm/graph_dump.cc
larroy: Should I add a static atomic counter for that for example?

larroy: From what I tested most of the nodes have names from operators or variables, and head gradients I added names which makes the graph much better.

ptrendx: I don't think you need an atomic counter for this - just a map that maps the `std::string` to list of `Node*` would be enough to differentiate whether your name is unique or not and what number you need to add to it.
There is nothing (unfortunately) in MXNet preventing you from naming your nodes or variables the same in the same graph. I agree that head grads should have a name, I actually did similar change in my pointwise fusion PR.

larroy: Doesn't it do the same effect with less memory usage to what I proposed just for the purpose of adding a number?

ptrendx: I guess I'm not sure how a counter would help you solve the issue of separate nodes that have the same name.

larroy: Ok I think I got now your intention. I will make the suggested changes.

include/mxnet/imperative.h
apeforest: If this is only used for debugging, I am not sure if adding a field to a production class is a good idea. Can we just pull the environment variable instead? It's not impacting production performance and not going to be used in multithreaded environment.

larroy:
Question, is there interest to get this merged if I address @ptrendx comments? The action points that I have in mind are just deduplicating node names. I have limited bandwidth right now to keep iterating in this pr. Could we summarize which changes are required to get this in or should I close the PR? Thanks.

larroy:
There doesn't seem to be interest to get this in. I will keep it as a private tool. Thanks for the reviews.

PR #15301, opened 2019-06-21
Ignore generated nnvm.cc template. This file is created whenever `make` is run

adis300:
Ignore generated nnvm.cc template in amalgamation dir.


adis300:
Reopening the pull request does not help.

PR #15317, opened 2019-06-21
fix doc for the op sort() and argsort()

suyz526:
## Description ##
Errors in the example of `sort()` and `argsort()`. Those op flatten and sort the array only when `axis=None` is given explicitly.

## Checklist ##
* [x] Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-15317/1/index.html

### DIFF
--- a/src/operator/tensor/ordering_op.cc
+++ b/src/operator/tensor/ordering_op.cc
@@ -114,7 +114,7 @@ Examples::
              [ 1.,  3.]]
 
   // flattens and then sorts
-  sort(x) = [ 1.,  1.,  3.,  4.]
+  sort(x, axis=None) = [ 1.,  1.,  3.,  4.]
 
   // sorts along the first axis
   sort(x, axis=0) = [[ 1.,  1.],
@@ -173,7 +173,7 @@ Examples::
                         [ 0.,  1.,  0.]]
 
   // flatten and then sort
-  argsort(x) = [ 3.,  1.,  5.,  0.,  4.,  2.]
+  argsort(x, axis=None) = [ 3.,  1.,  5.,  0.,  4.,  2.]
 )code" ADD_FILELINE)
 .set_num_inputs(1)
 .set_num_outputs(1)


suyz526:
@mxnet-label-bot add [Doc, pr-awaiting-review]

suyz526:
@aaronmarkham

src/operator/tensor/ordering_op.cc
```diff
@@ -114,7 +114,7 @@ Examples::
              [ 1.,  3.]]
 
   // flattens and then sorts
-  sort(x) = [ 1.,  1.,  3.,  4.]
+  sort(x, axis=None) = [ 1.,  1.,  3.,  4.]
```
aaronmarkham: This sort makes sense. 
But I'm thrown off by the other example. Sorting along the last axis would be 3 then 4 unless it is descending rather than ascending. Does this need to be specified for clarity?

suyz526: Hi, 
Thanks for your reply! 

Do you mean this example?
// sorts along the last axis
`sort(x) = [[ 1.,  4.],
           [ 1.,  3.]]`

Since x = `[[ 1, 4],[ 3, 1]]`, the pairs in the last axis are `[1,4]` and `[3,1]`. Sorting only along the last axis in a descend way would be `[1,4]` and `[1,3]`. So, in my opinion, it is correct? or maybe I'm wrong?

aaronmarkham: I tested it in numpy and it matches fine.

suyz526: Sorry that I didn't get your point. What I did wrong in this PR? 

If the argument `axis=None` is not given, this function will sort the array alone the last axis. However, in the second example, we want the array to be flattened and then sorted.

aaronmarkham: Everything is fine. I might have needed more examples to see the patterns, but I understand that if you're conversant with NumPy then this all works as expected.

roywei:

PR #15331, opened 2019-06-23
[fix] missing input log higher order.

kshitij12345:
@larroy Thank You very much for catching this.

Sorry for the silly mistake.
Can we have a way to test this?

@apeforest @larroy


roywei:
@mxnet-label-bot add [Operator, Backend]

larroy:
I'm still not sure what's the meaning of the backward output for the head gradient input as we discussed before. This week we are at a conference so we might be slow to respond.

I'm not sure how to test this, I think I would need to dump the graph and think about it, as I'm not sure now in which python variable is the gradient of the head gradient stored.

apeforest:
@larroy Those outputs are needed for 3rd order and above gradients.

apeforest:
@kshitij12345 https://github.com/apache/incubator-mxnet/pull/15331/files#diff-0dad60704ce39e602a1907aec6835375R1121 comment should actually be dL/dygrad. Could you please update it as well?

apeforest:
@sxjscience Do we have a use case where the gradient on the gradient of output y is needed? 
i.e. ygrad = dL/dy. How can we test the value of dG/dygrad given G is a function G(x, y) from R^n -> R.

apeforest:
Please update the comment, otherwise LGTM

kshitij12345:
https://github.com/apache/incubator-mxnet/blob/5171e1d92cfc5eefa2c20dfe8ac3fac5351ad19a/src/operator/tensor/elemwise_unary_op_basic.cc#L1120
`dL/dygrad` for this one right?

@larroy @apeforest , I was also wondering if we can check the number of inputs passed at compile time? I have observed the `MakeNode` gets the Op from dynamic registry based on the name. However we actually have information about the number of inputs and outputs for a given Op at compile time.  I tried but couldn't actually figure out. What are your thoughts? How easy or hard would it be to check for valid number of inputs in `MakeNode`? This would help catch these sort of errors at compile time itself.

larroy:
I think ograd[0] is dL/dx_grad

About the number of inputs, you are right that we could check. If it's more than one or two function calls I think is too much overhead and it's going to get caught with the python tests, Also if you don't return enough gradients, there's a check after calls to fgradient, so I think is not a big deal. Up to you if you can come up with something concise.

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -1117,15 +1117,15 @@ MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_log10,
                                                   unary_bwd<mshadow_op::log10_grad>)
 .set_attr<nnvm::FGradient>("FGradient",
   [](const nnvm::NodePtr& n, const std::vector<nnvm::NodeEntry>& ograds) {
-    // ograds[0]: dL/dxgrad
+    // ograds[0]: dL/dygrad
```
larroy: I think this is dL/dx_grad.  The head gradient is the gradient with respect to the previous output right? the previous output is x_grad  or dL/dx     so this thing is dL/(dL/dx)  or dL/dx_grad  in lack of a better notation.

src/operator/tensor/elemwise_unary_op_basic.cc
```diff
@@ -1117,15 +1117,15 @@ MXNET_OPERATOR_REGISTER_BINARY_WITH_SPARSE_CPU_DR(_backward_log10,
...
     // inputs[0]: dL/dy
-    // inputs[1]: x
+    // inputs[1]: x (ElemewiseGradUseIn)
```
larroy: nice comment, helps.

src/operator/tensor/elemwise_unary_op_basic.cc
kshitij12345: I guess it should be, dL/dy_grad as we are computing/returning dL/dx_grad,
Eg.
```
y = f(dx_grad)
L = g(y) # dx_grad formed part of the network and affected loss

During backprop by chain rule,
dL/dx_grad = dL/dy * dy/dx_grad

In comments, we have called dL/dy (mentioned in the above example) as dL/dy_grad
```

That is why we have,
https://github.com/apache/incubator-mxnet/blob/5b95fb3ee3581ba20fe1def336621d68a811e17f/src/operator/tensor/elemwise_unary_op_basic.cc#L1111-L1112

These multiplications performing, 
```
dL/dx_grad = dL/dy * dy/dx_grad
```

larroy: I think the notation is complicating us in excess as it gets pretty hairy.  It's the head gradient of the previous (output) node, which has shape of x, and x_grad. So it has to be related to x, not y.

kshitij12345: Oh. I get it now. If I understand it correctly then, crudely `ograds[0]` is how much does the `x_grad` affect the `L` and then we compute how does `x_grad` change with `x`. Makes sense now. 

Thank you very much. Will reflect it in this and other PRs.

larroy: @kshitij12345 I think what you write makes sense. I'm also unsure about notations, maybe you can come with a better one. If not maybe we leave the comment out, so we can merge the PR, as the code seems to do what's needed.

kshitij12345: Sure. Thanks Again.

karan6181:
@kshitij12345 Could you please resolve the merge conflict and ping the reviewers again to merge it? Thanks!

kshitij12345:
Sure.

kshitij12345:
@larroy @apeforest Gentle ping for review.

kshitij12345:
@apeforest @larroy Gentle Ping.

sxjscience:
I guess we need to add a test case.

kshitij12345:
@sxjscience I am not sure about how to test this. I was expecting that this missing input will cause problem with computing of higher order. Tried for 3rd order, which was computed successfully. For fourth order, 
`Operator _backward_mul is non-differentiable because it didn't register FGradient attribute` (different problem). So I am not very sure as to how to write a test case.

kshitij12345:
@sxjscience @apeforest @larroy Gentle Ping.

kshitij12345:
@apeforest Sure no worries. Thanks.

PR #15369, opened 2019-06-26
Fix build with system's openmp

hubutui:
## Description ##
This pr fix build with openmp by:
1. find openmp using `FindOpenMP` module from cmake
2. if not found, use build openmp from source using 3rd party submodule.

### Essentials ###

This pr is a minor modification of `CMakeLists.txt`, simply find openmp firstly in the system, and then build from source if not found.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -433,7 +433,8 @@ if(USE_OPENMP)
   find_package(OpenMP REQUIRED)
   # This should build on Windows, but there's some problem and I don't have a Windows box, so
   # could a Windows user please fix?
-  if(OPENMP_FOUND)
+  # cmake 3.0 use OPENMP_FOUND, but cmake > 3.7 use OpenMP_FOUND
+  if(OPENMP_FOUND OR OpenMP_FOUND)
       set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
       set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
       set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -433,7 +433,13 @@ if(USE_OPENMP)
   find_package(OpenMP REQUIRED)
   # This should build on Windows, but there's some problem and I don't have a Windows box, so
   # could a Windows user please fix?
-  if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/openmp/CMakeLists.txt
+  if(OPENMP_FOUND)
+      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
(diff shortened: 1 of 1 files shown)


marcoabreu:
Sounds less like a fix but rather a change in the logic if I'm not mistaken.

It would be good to have performance numbers before we move forward with this PR.

@larroy @lebeg @cjolivier01

lebeg:
I'm afraid that this change is rather a complicated one due to long discussions that have not ended yet. You can find more info on the dev mailing list: [[Discussion] Remove bundled llvm OpenMP](https://lists.apache.org/thread.html/4827f0f742b6e7e070da350ea81226d059401527f3072ce8b33c1fdf@%3Cdev.mxnet.apache.org%3E). Initially the proposed change was in this PR [#12160: Remove conflicting llvm OpenMP from cmake builds ](https://github.com/apache/incubator-mxnet/pull/12160).

We have already provided performance data on this, you can find and announcement on dev [here](https://lists.apache.org/thread.html/53030388434c90552d8dbed13488e25c1b153c53fc5970369f715e48@%3Cdev.mxnet.apache.org%3E) and the document [here](https://cwiki.apache.org/confluence/x/2wclBg).

That was not enough, for reasons that are unclear to me, to convince everybody on validity and necessity of this change.

Latest discussions happen in following dev threads: [OMP](https://lists.apache.org/thread.html/a4435d48f4253c85821e19828e57f706582926f70cd9f73cbe2dc080@%3Cdev.mxnet.apache.org%3E) and [[VOTE] Remove conflicting OpenMP from CMake builds](https://lists.apache.org/thread.html/f9897d76cd073f7a83cf6837ccacaf70670a592d727f0c2baa7a503b@%3Cdev.mxnet.apache.org%3E).

@hubutui If you feel motivated enough to proceed with this change you can join the debate.

hubutui:
@lebeg Hmm, actually, do user really care that much about performance affected by different OpenMP? As a packager's opinion, I want a good way to create this package. Maybe we could give user an option to choose the OpenMP, just like BLAS.

lebeg:
Yes, it might be a good idea, but the numbers as stated in the performance doc don't differ that much. Managing this dependency within MXNet isn't a good idea at all in my opinion.

hubutui:
It's been a long time since the discussion. I'll remove `3rdparty/openmp` before building, so that cmake would use openmp from system instead.

anirudhacharya:

larroy:
@hubutui would it be possible to add a CMake option to choose the OpenMP implementation that one wishes for?  There's three options, MKL, 3rdparty OpenMP and the one found in the platform. So I think the pr goes in the good direction but we would need a commandline option.

cjolivier01:
I don't think this is a good idea. we should standardize on one
implementation.  Like I have pointed out, the default make build uses
llvm/intel version.  cmake builds us llvm/intel version.  if libgomp is
still being included, removing it should be investigated.  Trying to
flip-flop between them on a whim is asking for trouble, and currently I
know of no advantage to using libgomp.

hubutui:
@cjolivier01 #9686 suggests transitioning fully to cmake, it would help a lot. Maybe dropping the Makefile someday?

cjolivier01:
Using cmake is another subject altogether.
How is this PR not basically what was already blocked?

cjolivier01:
This is basically a duplicate of turning on llvm openmp.

larroy:
But we are distributing pip wheels with libgomp, at least we should be able to replicate that for development purposes.

szha:
the reason is that there is mismatch between cmake and make builds. pip is currently using make for building libmxnet. for now if you want to replicate that you can use make

larroy:
@szha understood. it would be beneficial to be able to replicate that with CMake by having a selector for the openmp lib until then.

PR #15399, opened 2019-06-28
Add unit tests for TensorRT integration and fix some bugs

Caenorst:
## Description ##
TensorRT integration lacked of unit tests, we instead relied on output comparison of a full network which is not very pertinent, difficult to find the tolerance and not very helpful if it fails.

This PR have two purposes:
 1) Add unit test for all operations:
As we only partition subgraph of at least 2 Ops we always append an identity to each output. we then compare to MXNet, using both TRT FP32 and FP16 computation.
2) Fix a bunch of edge cases bugs that have been exposed by the unit tests:
  - NHWC is currently not compatible with TensorRT
  - Pooling is currently not compatible with count_include_pad and only compatible with pooling_convention Valid
  - FullyConnected without bias cannot be converted to MatMul (as there should be a transpose before) for the moment we decided to remove compatibility, it is quite rare to do FullyConnected without bias anyway
  - Concat is not compatible if the concatenation axis is the batch axis
  - The dropout have to be only enabled on training mode (act as identity)
  - BatchNorm with fix_gamma can have gamma != 1. so we force the value to be 1. when loading it.

### Essentials ###

## Comments ##
- the test of output comparison in a full networks should probably be replaced by an accuracy comparison over a full dataset


anirudhacharya:

Caenorst:
@KellenSunderland @haohuanw

KellenSunderland:
Looks like CI caught a few issues.  For example http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-15399/1/pipeline seems like it should be relevant to this PR.  I'll have a look to see if there's anything else that jumps out at me.

src/operator/subgraph/tensorrt/nnvm_to_onnx.cc
```diff
@@ -157,6 +157,12 @@ std::string ConvertNnvmGraphToOnnx(
   return serialized_onnx_graph;
 }
 
+void ConvertIdentity(NodeProto* node_proto, const NodeAttrs& attrs,
```
KellenSunderland: Any idea if TRT actually optimizes this out?  I've seen this in a few prod services :-/

src/operator/subgraph/tensorrt/tensorrt-inl.h
```diff
@@ -109,13 +111,70 @@ class TensorrtSelector : public SubgraphSelector {
...
+          return false;
+      }
+    }
+
+    if (op_name == "Concat") {
+      const auto& param = nnvm::get<ConcatParam>(n.attrs.parsed);
+      return (param.dim != 0);
+    }
+
+    if (op_name == "Dropout") {
```
KellenSunderland: Again, will TensorRT optimize this out?  We don't want it at inference time right?

src/operator/subgraph/tensorrt/tensorrt-inl.h
```diff
@@ -180,6 +253,17 @@ class TensorrtProperty : public SubgraphProperty {
     n->attrs.name = "TensorRT" + std::to_string(subgraph_id);
     n->attrs.op = Op::Get("_TensorRT");
     CHECK(n->attrs.op);
+    // prevent using Gamma value if using fix_gamma on BatchNorm
+    DFSVisit(new_sym.outputs, [&n](const nnvm::NodePtr& node) {
+      if (node->op() == Op::Get("BatchNorm")) {
```
KellenSunderland: Why not just check for FixGamma = true during nnvm -> Onnx conversion and set gamma to 0 if it's true?

tests/python/gpu/test_tensorrt.py
```diff
@@ -0,0 +1,437 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
KellenSunderland: Super helpful tests, thanks Clement.  It's ok to merge in a subset of these at a time, but I'd try to avoid pushing commented out code / too many todos upstream.

karan6181:
@Caenorst Could you please address the review comments? Thanks!

piyushghai:
@Caenorst Gentle ping...

src/operator/subgraph/tensorrt/tensorrt-inl.h
Caenorst: Dropout have always been seen as identity function in MXNet-TensorRT integration so I don't see any changement on this, regarding to whether or not identity is actually doing a copy or not I'm not quite sure, here is the onnx-tensorrt conversion: https://github.com/onnx/onnx-tensorrt/blob/0ab159579551cabfa05fd66f338357f116e96835/trt_utils.hpp#L169-L180

KellenSunderland: Ok, non-blocking comment for this PR.  I'm just thinking about adding a warning in the future if people are using TRT with operations that don't make sense at inference time (Dropout, Ident, Empty Concats or Copies, etc.)

src/operator/subgraph/tensorrt/tensorrt-inl.h
KellenSunderland: I think this one should be handled upstream.  Have a look at https://github.com/apache/incubator-mxnet/commit/6b22aa455e562db0e2744e94dc6a583191081b05 and let us know if you'd rather use this fix or if ours looks ok.

src/operator/subgraph/tensorrt/nnvm_to_onnx.cc
Caenorst: I believe this should be optimized by ONNX-TRT

src/operator/subgraph/tensorrt/tensorrt-inl.h
Caenorst: I think solution 6b22aa4 looks better, I removed mine

Caenorst:
I don't understand the error on windows-gpu, it doesn't seems related to my modifications...

Caenorst:
@KellenSunderland can we merge it ? (I did a bunch of modifications since the last review that you may wanna review too)

tests/python/gpu/test_tensorrt.py
Caenorst: I removed the commented out unit tests

PR #15405, opened 2019-06-29
Fix memory leak in NaiveEngine

### LINKED ISSUE #15375: Memory leak in Naive engine when profiling
I believe there's a memory leak in Naive engine here:

https://github.com/apache/incubator-mxnet/blob/master/src/engine/naive_engine.cc#L164

I think it would be a great opportunity to refactor how NewOperator is released. As it's also difficult to reason about the release of the operator memory in the threaded engine (looks correct to me, but it takes quite a bit to dig it up and understand).

I would propose returning a unique_ptr directly with a custom deleter using a virtual operator() that is specialized for the engine type which will get the same effect as how operators are being released now. This would make easier to reason about this memory management part.

https://github.com/apache/incubator-mxnet/blob/master/src/engine/threaded_engine.cc#L270

larroy:
Fixes #15375

## Description ##
Fixes a memory leak in NaiveEngine when using operator profiling.

### DIFF
--- a/src/engine/naive_engine.cc
+++ b/src/engine/naive_engine.cc
@@ -159,15 +159,19 @@ class NaiveEngine final : public Engine {
         NaiveEngine::OnComplete, nullptr);
     this->req_completed_ = false;
     profiler::Profiler *profiler = profiler::Profiler::Get();
-    NaiveOpr *opr = nullptr;
+    auto opr_deleter = [this](NaiveOpr* p) {
+      this->DeleteOperator(p);
+    };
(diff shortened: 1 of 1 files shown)


larroy:
Added valgrind run which shows the memory being leaked.

naive_engine_leak

larroy:
@mxnet-label-bot add [Engine, Bug, Memory]

larroy:
@mxnet-label-bot add [Backend]

larroy:

wkcn:
Merged. Thank you!

PR #15117, opened 2019-05-31
nano instructions

aaronmarkham:
## Description ##

This PR adds instructions on how to setup MXNet on a Jetson Nano.

I had a previous PR, #14855, that also added the Nano to CI, but this is stuck. So, I'm decoupling the working instructions from the failing CI part.

Issue #12524, opened 2018-09-11
AddVersion.py breaks the site output in a variety of ways


aaronmarkham:
## Description

[AddVersion.py](https://github.com/aaronmarkham/incubator-mxnet/blob/b56da3cc32fee828c352f0bfb1f05e393ca58040/docs/build_version_doc/AddVersion.py#L77-L78) does some pretty aggressive website modifications just before publishing the full website. It was already known to break on the install page, and now it seems it is responsible for #12168 and #11438. I created a PR to fix this issue by commenting out this one line:
```
outstr = str(content).replace('&lt;', '<').replace('&gt;', '>')
```
But this ended up stopping the versions dropdown from appearing in the final website output. Having the versions dropdown is more important right now than fixing a couple of render issues, so I've submitted a patch that reverts this change along with an update for site consistency: #12426 

Browsing the intermediate outputs from this script, you can see things like `<span class="gp">>>> </span>`.  That's with the line enabled. More research is required to fix this script, so it produces valid html and doesn't break the code block's usage of xml or other tags the require > or <.

Tracking here as well:
https://issues.apache.org/jira/browse/MXNET-987

kalyc:
Thanks for submitting the issue @aaronmarkham 
@mxnet-label-bot[Website, Build, Installation]

kalyc:
@aaronmarkham is this issue still persisting?

aaronmarkham:
Yes

aaronmarkham:
New website launch makes this a moot issue.

PR #15433, opened 2019-07-02
remove comments from nano instructions

aaronmarkham:
## Description ##
Publish process is breaking comments like mentioned here: https://github.com/apache/incubator-mxnet/issues/12524

Instructions are clipped right now: http://mxnet.incubator.apache.org/versions/master/install/install-jetson.html [as reported here](https://github.com/apache/incubator-mxnet/pull/15117#issuecomment-507736472).

So this PR removes the comments to avoid the problem.

### DIFF
--- a/docs/install/install-jetson.md
+++ b/docs/install/install-jetson.md
@@ -61,22 +61,6 @@ sudo pip install \
 
 If you plan to cross-compile you will need to install these dependencies on that computer as well.
 
-<!--
-### Java API
-
-To use the Java inference API you only need the following dependencies:
-
-```
-sudo apt install \
-                        maven \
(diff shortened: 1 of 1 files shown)


ptrendx:
Shouldn't you also remove the Scala&Java instructions from the building from source part if you do not include prerequisites for Java install and remove the `These instructions also cover how to setup MXNet's Java Inference API.` remark in the introduction?

aaronmarkham:
> Shouldn't you also remove the Scala&Java instructions from the building from source part if you do not include prerequisites for Java install and remove the `These instructions also cover how to setup MXNet's Java Inference API.` remark in the introduction?

Yep! Good catch. Thanks.

PR #15487, opened 2019-07-08
Opperf: Support Python<3.6

kshitij12345:
f-Strings are supported only from Python 3.6 onwards.
Have replaced their occurrences with str.format which is supported from Python 2.6.

### DIFF
--- a/benchmark/opperf/opperf.py
+++ b/benchmark/opperf/opperf.py
@@ -141,7 +141,7 @@ def main():
     # 4. Generate list of MXNet operators not covered in benchmarks
     ops_not_covered = get_operators_with_no_benchmark(final_benchmark_results.keys())
     for idx, op in enumerate(ops_not_covered):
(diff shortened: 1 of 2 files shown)


benchmark/opperf/opperf.py
```diff
@@ -140,7 +141,7 @@ def main():
     # 4. Generate list of MXNet operators not covered in benchmarks
     ops_not_covered = get_operators_with_no_benchmark(final_benchmark_results.keys())
     for idx, op in enumerate(ops_not_covered):
-        print(f"{idx}. {op}")
+        print("{idx}. {op}".format(idx, op))
```
apeforest: Should this be 
```suggestion
        print("{idx}. {op}".format(idx=idx, op=op))
```

kshitij12345: Right. Thanks

wkcn:

PR #15489, opened 2019-07-08
Dynamic Library Loading Support

mseth10:
## Description ##

This PR is part of a larger effort of adding new functionality to MXNet that was compiled separately from MXNet. See [here](https://cwiki.apache.org/confluence/display/MXNET/Bring+your+own+Accelerator) and [here](https://cwiki.apache.org/confluence/display/MXNET/Dynamic+CustomOp+Support) for more info.

This PR is the first step that enables libraries to be dynamically loaded into MXNet at runtime .

Features:
- Dynamic library loading functions for both Linux and Windows to load libraries and access its functions
- MXLoadLib that loads and initializes a library
- A new Python API: mx.library.load that calls the new C-API
- Enhancements to both the Makefile and CMake flows to link with the dynamic loader library (dl)

This PR includes the following:
- An example, showing how to write a sample library in C++ (mylib.cc) with its own Makefile, a unittest to verify correctness outside of MXNet (libtest.cc), and a test script showing how to load the library in MXNet (test.py).
- A unit test that loads a precompiled library into MXNet and initializes it. It runs as part of the CI for both Linux and Windows.

### Essentials ###

### Changes ###

## Comments ##


karan6181:
@mxnet-label-bot add [backend, C API, pr-work-in-progress]

mseth10:
@szha I agree with you, I am working on refactoring the PR. Will ping you again for review after I am done.

example/lib_api/mylib.cc
```diff
@@ -0,0 +1,37 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * Copyright (c) 2015 by Contributors
+ * \file my_accel.cc
```
samskalicky: change file name

example/lib_api/test.py
```diff
@@ -0,0 +1,23 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# coding: utf-8
+# pylint: disable=arguments-differ
+
+import mxnet as mx
+
+mx.library.load_lib('mylib.so')
```
samskalicky: should we just call this "load" since its already in the "library" module?

include/mxnet/lib_api.h
```diff
@@ -0,0 +1,44 @@
...
+
+#define INITIALIZE_STR "initialize"
+typedef int (*initialize_t)(int);
+
+extern "C" {
+    /*
+    Function: initialize
+    Parameters:
+    - int : MXNet version passed to library
+    Returns: Success/Failure code
```
samskalicky: need to explain return values: "returns non-zero on error"

include/mxnet/library.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include <dlfcn.h>
+#include <iostream>
+#include "dmlc/io.h"
+
+void* load_lib(const char* path);
+void get_sym(void* handle, void** func, char* name);
+
+template<typename T>
+T get_func(void *lib, char *func_name) {
```
samskalicky: comment on function def?

python/mxnet/library.py
```diff
@@ -0,0 +1,51 @@
...
+    Returns
+    ---------
+    void
+    """
+    #check if path exists
+    if not os.path.exists(path):
+        print('load_lib path "%s" does NOT exist' % path)
+        return None
+    #check if path is to a file
+    if not os.path.isfile(path):
```
samskalicky: should we also check if its a so or dll extension?

src/c_api/c_api.cc
```diff
@@ -90,6 +92,20 @@ inline int MXAPIGetFunctionRegInfo(const FunRegType *e,
 
 // NOTE: return value is added in API_END
 
+int MXLoadLib(const char *path) {
```
samskalicky: comment on function def?

src/c_api/c_api.cc
```diff
@@ -90,6 +92,20 @@ inline int MXAPIGetFunctionRegInfo(const FunRegType *e,
...
+  API_BEGIN();
+  void *lib = load_lib(path);
+  if(!lib)
+    LOG(FATAL) << "Unable to load library";
+
+  initialize_t initialize = get_func<initialize_t>(lib, const_cast<char*>(INITIALIZE_STR));
+  int version = MXNET_VERSION;
+  int flag = initialize(version);
```
samskalicky: initialize((int)MXNET_VERSION)

src/common/library.cc
```diff
@@ -0,0 +1,47 @@
...
+/*!
+ * Copyright (c) 2015 by Contributors
+ * \file library.cc
+ * \brief Dynamically loading accelerator library
+ * and accessing its functions
+ */
+
+#include "../../include/mxnet/library.h"
+
+void* load_lib(const char* path) {
```
samskalicky: comment on function def?

src/common/library.cc
```diff
@@ -0,0 +1,47 @@
...
+  void *handle;
+  handle = dlopen(path, RTLD_LAZY);
+
+  if (!handle) {
+    std::cerr << "Error loading accelerator library: '" << path
```
samskalicky: replace cerr with LOG(FATAL)

src/common/library.cc
```diff
@@ -0,0 +1,47 @@
...
+              << "'\n" << dlerror() << std::endl;
+    return nullptr;
+  }
+  return handle;
+}
+
+void get_sym(void* handle, void** func, char* name) {
+  *func = dlsym(handle, name);
+  if (!(*func)) {
+    std::cerr << "Error getting function '" << name
```
samskalicky: replace cerr with LOG(FATAL)

src/common/library.cc
```diff
@@ -0,0 +1,47 @@
...
+
+  if (!handle) {
+    std::cerr << "Error loading accelerator library: '" << path
+              << "'\n" << dlerror() << std::endl;
+    return nullptr;
+  }
+  return handle;
+}
+
+void get_sym(void* handle, void** func, char* name) {
```
samskalicky: comment on function def?

samskalicky:
Can you clean up the whitespace to fix the sanity build?

```
src/c_api/c_api.cc:98:  Missing space before ( in if(  [whitespace/parens] [5]
```

example/lib_api/mylib.cc
mseth10: Addressed.

example/lib_api/test.py
mseth10: Addressed.

include/mxnet/lib_api.h
mseth10: Addressed.

include/mxnet/library.h
mseth10: Addressed.

python/mxnet/library.py
mseth10: Addressed.

src/c_api/c_api.cc
mseth10: Addressed.

src/c_api/c_api.cc
mseth10: Addressed.

src/common/library.cc
mseth10: Addressed.

src/common/library.cc
mseth10: Addressed.

src/common/library.cc
mseth10: Addressed.

src/common/library.cc
mseth10: Addressed.

src/common/library.cc
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+/*!
+ * Copyright (c) 2015 by Contributors
+ * \file library.cc
+ * \brief Dynamically loading accelerator library
+ * and accessing its functions
+ */
+
+#if defined(_WIN32) || defined(_WIN64) || defined(__WINDOWS__)
```
rondogency: can you put them together like "#define WIN_ENV defined(_WIN32) || defined(_WIN64) || defined(__WINDOWS__)"

example/lib_api/libtest.cc
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+#if defined(_WIN32) || defined(_WIN64) || defined(__WINDOWS__)
+#include <windows.h>
+#else
+#include <dlfcn.h>
+#endif
+
+#include <stdio.h>
+#include "../../include/mxnet/lib_api.h"
```
samskalicky: lets just change this to "lib_api.h" so it matches the mylib.cc file below.

You already have the include path setup in the makefile as: -I ../../include/mxnet so we dont need absolute path in the include statement

example/lib_api/libtest.cc
```diff
@@ -0,0 +1,77 @@
...
+
+#define MXNET_VERSION 10500
+
+int main(void) {
+  // Get a handle to the library.
+#if defined(_WIN32) || defined(_WIN64) || defined(__WINDOWS__)
+  HINSTANCE handle;
+  handle = LoadLibrary(TEXT("lib.dll"));
```
samskalicky: this is called mylib.dll

python/mxnet/library.py
```diff
@@ -0,0 +1,51 @@
...
+    Parameters
+    ---------
+    path : Path to library .so file
+
+    Returns
+    ---------
+    void
+    """
+    #check if path exists
+    if not os.path.exists(path):
```
samskalicky: lets also require users to path an absolute path, so that we dont have to worry about when a user gives a path and the dynamic loader cannot find it. If we force users to provide an absolute path, it should always be found by the dynamic loader when we call dlopen.

we can check by doing:

```
if not os.path.isabs(path):
   print('load path "%s" is not an absolute path' % path)
   return
```

tests/python/unittest/test_library_loading.py
```diff
@@ -0,0 +1,29 @@
...
+import os
+from mxnet.test_utils import download
+
+def test_library_loading():
+    if (os.name=='posix'):
+        lib = 'mylib.so'
+    elif (os.name=='nt'):
+        lib = 'mylib.dll'
+
+    fname = mx.test_utils.download('https://mxnet-demo-models.s3.amazonaws.com/lib_binary/'+lib)
```
samskalicky: lets add this code after line 28:

```
fname = os.path.abspath(fname)
```

src/common/library.cc
```diff
@@ -0,0 +1,89 @@
...
+    char *lpMsgBuf;
+    unsigned long dw = GetLastError();
+    FormatMessage(
+        FORMAT_MESSAGE_ALLOCATE_BUFFER | 
+        FORMAT_MESSAGE_FROM_SYSTEM |
+        FORMAT_MESSAGE_IGNORE_INSERTS,
+        NULL,
+        dw,
+        MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+        lpMsgBuf,
```
samskalicky: we should follow the MS example and pass &lpMsgBuf casting it to (LPSTR) ie. (char*)
https://docs.microsoft.com/en-us/windows/win32/debug/retrieving-the-last-error-code

python/mxnet/library.py
mseth10: Added this check.

tests/python/unittest/test_library_loading.py
mseth10: Added.

src/common/library.cc
mseth10: Done.

example/lib_api/test.py
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# coding: utf-8
+# pylint: disable=arguments-differ
+
```
rondogency: please add more in-file comment on what this file is doing and how it is different from the above libtest file

example/lib_api/libtest.cc
```diff
@@ -0,0 +1,77 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * Copyright (c) 2015 by Contributors
+ * \file libtest.cc
+ * \brief Tests if the library is implemented correctly
```
rondogency: you can add something like 'dynamic library authors can use this test to verify their library implementation'

include/mxnet/lib_api.h
```diff
@@ -0,0 +1,47 @@
...
+#include <cstdlib>
+
+#define INITIALIZE_STR "initialize"
+typedef int (*initialize_t)(int);
+
+extern "C" {
+    /*
+    Function: initialize
+    Parameters:
+    - int : MXNet version passed to library
```
junrushao1994: Do we need to document the format of version number?

samskalicky: Good idea @junrushao1994. Maybe something like this:

```
/*! \brief version number is defined as:
*  MXNET_VERSION = (MXNET_MAJOR*10000 + MXNET_MINOR*100 + MXNET_PATCH)
*/ 
```

src/common/library.cc
mseth10: as discussed offline, this is the way we do it.

include/mxnet/lib_api.h
samskalicky: @zheng-da https://github.com/apache/incubator-mxnet/blob/master/include/mxnet/base.h#L73-L82

example/lib_api/test.py
mseth10: Added

example/lib_api/libtest.cc
mseth10: Added

include/mxnet/lib_api.h
mseth10: Added.

example/lib_api/libtest.cc
```diff
@@ -0,0 +1,79 @@
...
+  handle = dlopen("mylib.so", RTLD_LAZY);
+#endif
+
+  // If the handle is valid, try to get the function address.
+  if (!handle) {
+    printf("Unable to load library\n");
+    return 1;
+  }
+
+  initialize_t func;
```
samskalicky: need a comment here like "get initialize function pointer from the library"

python/mxnet/library.py
```diff
@@ -0,0 +1,55 @@
...
+        print('load path "%s" does NOT exist' % path)
```
samskalicky: need to throw error like:

```
raise MXNetError("cannot use the setter: %s to set attribute" % obj.__name__)
```

python/mxnet/library.py
```diff
@@ -0,0 +1,55 @@
...
+        return
+    #check if path is an absolute path
+    if not os.path.isabs(path):
+        print('load path "%s" is not an absolute path' % path)
+        return
+    #check if path is to a library file
+    _, file_ext = os.path.splitext(path)
+    if not file_ext in ['.so', '.dll']:
+        print('load path "%s" is NOT a library file' % path)
```
samskalicky: and here too

python/mxnet/library.py
```diff
@@ -0,0 +1,55 @@
...
+    ---------
+    void
+    """
+    #check if path exists
+    if not os.path.exists(path):
+        print('load path "%s" does NOT exist' % path)
+        return
+    #check if path is an absolute path
+    if not os.path.isabs(path):
+        print('load path "%s" is not an absolute path' % path)
```
samskalicky: and here

python/mxnet/library.py
```diff
@@ -0,0 +1,55 @@
...
+import os
+from .base import _LIB
+from .base import check_call
+
+def load(path):
+    """Loads library dynamically.
+
+    Parameters
+    ---------
+    path : Path to library .so file
```
samskalicky: change to ".so/dll" file, and mention that the path has to be absolute path

mseth10: Done

python/mxnet/library.py
mseth10: Done

python/mxnet/library.py
mseth10: Done

python/mxnet/library.py
mseth10: Done

example/lib_api/libtest.cc
mseth10: Done

python/mxnet/library.py
```diff
@@ -0,0 +1,54 @@
...
+    path : Path to library .so/.dll file
+
+    Returns
+    ---------
+    void
+    """
+    #check if path exists
+    if not os.path.exists(path):
+        raise MXNetError("load path %s does NOT exist" % path)
+        return
```
samskalicky: remove return

python/mxnet/library.py
```diff
@@ -0,0 +1,54 @@
...
+    #check if path is an absolute path
+    if not os.path.isabs(path):
+        raise MXNetError("load path %s is not an absolute path" % path)
+        return
```
samskalicky: remove return

python/mxnet/library.py
```diff
@@ -0,0 +1,54 @@
...
+    #check if path is to a library file
+    _, file_ext = os.path.splitext(path)
+    if not file_ext in ['.so', '.dll']:
+        raise MXNetError("load path %s is NOT a library file" % path)
+        return
```
samskalicky: remove return

python/mxnet/library.py
mseth10: Done.

python/mxnet/library.py
mseth10: Done.

python/mxnet/library.py
mseth10: Done.

python/mxnet/library.py
```diff
@@ -0,0 +1,51 @@
...
+    #check if path is to a library file
+    _, file_ext = os.path.splitext(path)
+    if not file_ext in ['.so', '.dll']:
+        raise MXNetError("load path %s is NOT a library file" % path)
+
+    byt_obj = path.encode('utf-8')
+    chararr = ctypes.c_char_p(byt_obj)
+    check_call(_LIB.MXLoadLib(chararr))
+
+    return
```
samskalicky: need to remove return

src/common/library.cc
```diff
@@ -0,0 +1,104 @@
...
+#endif
+
+
+/*!
+ * \brief Loads the dynamic shared library file
+ * \param path library file location
+ * \return handle a pointer for the loaded library, nullptr if loading unsuccessful
+ */
+void* load_lib(const char* path) {
+  void *handle;
```
samskalicky: initialize:

```
void *handle=nullptr;
```

src/common/library.cc
```diff
@@ -0,0 +1,104 @@
...
+        FORMAT_MESSAGE_ALLOCATE_BUFFER |
+        FORMAT_MESSAGE_FROM_SYSTEM |
+        FORMAT_MESSAGE_IGNORE_INSERTS,
+        NULL,
+        dw,
+        MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+        reinterpret_cast<char*>(err),
+        0, NULL);
+}
+#endif
```
samskalicky: try this:

```
/*!
 * \brief Retrieve the system error message for the last-error code
 * \param err string that gets the error message
 */
void win_err(char **err) {
#if defined(_WIN32) || defined(_WIN64) || defined(__WINDOWS__)
  uint32_t dw = GetLastError();
  FormatMessage(
        FORMAT_MESSAGE_ALLOCATE_BUFFER |
        FORMAT_MESSAGE_FROM_SYSTEM |
        FORMAT_MESSAGE_IGNORE_INSERTS,
        NULL,
        dw,
        MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
        reinterpret_cast<char*>(err),
        0, NULL);
#endif
}
```

python/mxnet/library.py
mseth10: Removed.

example/lib_api/Makefile
```diff
@@ -0,0 +1,31 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+all:
+	g++ -shared -fPIC mylib.cc -o mylib.so -I ../../include/mxnet
```
larroy: Nice example. Minor suggestion: would you mind parameterizing the compiler (calling $CC and $CXX )so the user doesn't have to edit to change compilers?

example/lib_api/libtest.cc
```diff
@@ -0,0 +1,78 @@
...
+#if defined(_WIN32) || defined(_WIN64) || defined(__WINDOWS__)
+  HINSTANCE handle;
+  handle = LoadLibrary(TEXT("mylib.dll"));
+#else
+  void *handle;
+  handle = dlopen("mylib.so", RTLD_LAZY);
+#endif
+
+  if (!handle) {
+    printf("Unable to load library\n");
```
larroy: suggest cerr  and change to <iostream> instead of stdio.h

example/lib_api/libtest.cc
```diff
@@ -0,0 +1,78 @@
...
+    return 1;
+  }
+
+  // get initialize function address from the library
```
larroy: can we use a better name than "func"?  "initialize something"?

example/lib_api/libtest.cc
```diff
@@ -0,0 +1,78 @@
...
+  initialize_t func;
+#if defined(_WIN32) || defined(_WIN64) || defined(__WINDOWS__)
+  func = (initialize_t) GetProcAddress(handle, INITIALIZE_STR);
+#else
+  func = (initialize_t) dlsym(handle, INITIALIZE_STR);
+#endif
+
+  if (!func) {
+    printf("Unable to get function 'intialize' from library\n");
```
larroy: standard error.

example/lib_api/mylib.cc
```diff
@@ -0,0 +1,37 @@
...
+ * \brief Sample library file
+ */
+
+#include <iostream>
+#include "lib_api.h"
+
+int initialize(int version) {
+  if (version >= 10400) {
+    std::cout << "MXNet version " << version << " supported" << std::endl;
+    return 1;
```
larroy: Does it make sense to return !=0 on error?  This seems opposite of C conventions.

example/lib_api/test.py
```diff
@@ -0,0 +1,29 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
larroy: Can we add shebang with python version using "env"?

`#!/usr/bin/env python3`

include/mxnet/lib_api.h
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright (c) 2015 by Contributors
+ * \file lib_api.h
+ * \brief APIs to interact with libraries
+ */
+#ifndef MXNET_LIB_API_H_
+#define MXNET_LIB_API_H_
+
+#include <cstdlib>
+
+#define INITIALIZE_STR "initialize"
```
larroy: since the macro is global in the preprocessor, shall it be prefixed with MX_ or something? just something to think about.  Also not documented.

include/mxnet/lib_api.h
```diff
@@ -0,0 +1,47 @@
...
+
+/*!
+ * Copyright (c) 2015 by Contributors
+ * \file lib_api.h
+ * \brief APIs to interact with libraries
+ */
+#ifndef MXNET_LIB_API_H_
+#define MXNET_LIB_API_H_
+
+#include <cstdlib>
```
larroy: is it used?

python/mxnet/library.py
```diff
@@ -0,0 +1,49 @@
...
+    path : Path to library .so/.dll file
+
+    Returns
+    ---------
+    void
+    """
+    #check if path exists
+    if not os.path.exists(path):
+        raise MXNetError("load path %s does NOT exist" % path)
+    #check if path is an absolute path
```
larroy: Why does it have to be an absolute path? Can we clarify? What if the library is installed in /usr/lib?

src/common/library.cc
```diff
@@ -0,0 +1,104 @@
...
+ * \brief Loads the dynamic shared library file
+ * \param path library file location
+ * \return handle a pointer for the loaded library, nullptr if loading unsuccessful
+ */
+void* load_lib(const char* path) {
+  void *handle;
+#if defined(_WIN32) || defined(_WIN64) || defined(__WINDOWS__)
+  handle = LoadLibrary(path);
+  if (!handle) {
+    char *lpMsgBuf;
```
larroy: initialize

src/common/library.cc
```diff
@@ -0,0 +1,104 @@
...
+ * \brief Obtains address of given function in the loaded library
+ * \param handle pointer for the loaded library
+ * \param func function pointer that gets output address
+ * \param name function name to be fetched
+ */
+void get_sym(void* handle, void** func, char* name) {
+#if defined(_WIN32) || defined(_WIN64) || defined(__WINDOWS__)
+  *func = GetProcAddress((HMODULE)handle, name);
+  if (!(*func)) {
+    char *lpMsgBuf;
```
larroy: same. Also let's not use hungarian notation, just name the variable semantically, we know the type, the compiler does as well.

tests/python/unittest/test_library_loading.py
```diff
@@ -0,0 +1,32 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# This test checks if dynamic loading of library into MXNet is successful
+
+import mxnet as mx
+import os
+from mxnet.test_utils import download
+
```
larroy: Can you add a decorator to skip the test if the platform is not what you expect? This will fail in other architectures. Please check OS and platform.

Wouldn't it be more resilient to build this stub as part of the build process instead of relying on downloads, this would also solve the above.

example/lib_api/Makefile
mseth10: This Makefile is small enough, we don't need to parameterize the compiler.

example/lib_api/mylib.cc
mseth10: Here, we are using the return value as a boolean, so 1 means True/success.

python/mxnet/library.py
mseth10: Dynamic loader (dlopen) requires the path to be absolute.

src/common/library.cc
mseth10: Done

example/lib_api/libtest.cc
mseth10: Done.

example/lib_api/libtest.cc
mseth10: Done

include/mxnet/lib_api.h
mseth10: Removed.

src/common/library.cc
mseth10: Initialized.

src/common/library.cc
mseth10: Initialized. And changed variable name.

example/lib_api/libtest.cc
mseth10: Changed the name.

example/lib_api/test.py
mseth10: Added.

include/mxnet/lib_api.h
mseth10: Added a prefix and comments.

tests/python/unittest/test_library_loading.py
mseth10: That is a good suggestion Pedro. As discussed offline, the whole point of this PR is to load an external library, so it makes sense to download it here.

mseth10:
@zheng-da I have addressed your comments. Can you please approve?

tests/python/unittest/test_library_loading.py
larroy: This test won't pass in ARM. You need to check the platform and skip this test if the binary is not for the OS / platform. If you want to download an artifact, be it. But then it needs more refinement to skip the test if the platform doesn't match your binary.

larroy: I suggest to use something like 

```
@unittest.skipIf(check_platform(), "blah")
```

If you insist in loading a binary artifact. Otherwise build the library during the build process and check that we are not pre-built as in tools/diagnose.py  for example.

```
     commit_hash = os.path.join(mx_dir, 'COMMIT_HASH')
        if os.path.exists(commit_hash):
            with open(commit_hash, 'r') as f:
                ch = f.read().strip()
                print('Commit Hash   :', ch)
        else:
            print('Commit hash file "{}" not found. Not installed from pre-built package or built from source.'.format(commit_hash))

```
So the test runs only if we are not pre-built.

I think there are two valid options.

mseth10: Added a skip unit test decorator.

tests/python/unittest/test_library_loading.py
```diff
@@ -17,10 +17,13 @@
...
 # This test checks if dynamic loading of library into MXNet is successful
 
-import mxnet as mx
 import os
+import platform
+import unittest
+import mxnet as mx
 from mxnet.test_utils import download
 
+@unittest.skipIf(platform.machine() not in ['x86_64', 'AMD64'], "not all machine types supported")
```
larroy: suggest to check machine and OS, also factor out the check in a function.

mseth10: 'AMD64' is the value given by Windows so checking OS is not required. Refactored the check into a function.

mseth10:

python/mxnet/library.py
ptrendx: So can't we make it absolute programmatically in a Python level as a convenience for the user?

samskalicky: Thanks for the suggestion @ptrendx. We thought about this. We could, but if the user passes in "./libmyacc.so" and Python is running in a different dir, we're going to make the wrong absolute path "/user/home/libmyacc.so". 

So initially we were thinking that it just makes more sense to require the user to provide that as input. It would avoid confusion around "how is this path getting created".

src/common/library.cc
```diff
@@ -0,0 +1,104 @@
...
+  handle = LoadLibrary(path);
+  if (!handle) {
+    char *err_msg = nullptr;
+    win_err(&err_msg);
+    LOG(FATAL) << "Error loading library: '" << path << "'\n" << err_msg;
+    LocalFree(err_msg);
+    return nullptr;
+  }
+#else
+  handle = dlopen(path, RTLD_LAZY);
```
leleamol: Where are we calling the 'dlclose()'? 

How are we checking that we are not calling "dlopen" multiple times for the same library? If we are allowing calling multiple dlopen for the same file, it should match with the same number of dlclose.

src/common/library.cc
```diff
@@ -0,0 +1,104 @@
...
+#if defined(_WIN32) || defined(_WIN64) || defined(__WINDOWS__)
+  *func = GetProcAddress((HMODULE)handle, name);
+  if (!(*func)) {
+    char *err_msg = nullptr;
+    win_err(&err_msg);
+    LOG(FATAL) << "Error getting function '" << name << "' from library\n" << err_msg;
+    LocalFree(err_msg);
+  }
+#else
+  *func = dlsym(handle, name);
```
leleamol: The dlsym can return NULL and it may not necessarily indicate error condition.  The recommended error handling is mentioned in this document.
https://linux.die.net/man/3/dlsym

samskalicky: Thanks Amol, per this: https://stackoverflow.com/a/53590728 I think we can just take the shortcut and assume dlsym returning NULL means the symbol is not found. You have to try pretty hard to get gcc to put a function at the global 0 address.

src/common/library.cc
samskalicky: Thanks @leleamol, per our offline discussion we will take the following actions:
1. Create a map to track file handles and paths to the libraries when opened
2. We'll only allow dlopen to be called once, but if the library is "reloaded" by the user, we'll skip dlopen and go through the same "library registration process" again.
3. Add code to call dlclose for each library opened in the NotifyShutdown functions for each of the engine types in:
https://github.com/apache/incubator-mxnet/blob/eab6da6208fdc7b68437dadc6d7d58ee11c2f528/src/engine/naive_engine.cc#L230
https://github.com/apache/incubator-mxnet/blob/3a4698058840ff79e58709b6135c8fb757a99606/src/engine/threaded_engine.h#L310

samskalicky:
@leleamol Can you please review again to see if the implementation of the dlclose matches what you were thinking?

@anirudh2290 Thanks for the help with the engine destructor info, can you take a look at the changes to the engines:
https://github.com/apache/incubator-mxnet/blob/134cc37d6b306514fde1a405d3b87cc946a56bd1/src/engine/naive_engine.cc#L88-L90
https://github.com/apache/incubator-mxnet/blob/134cc37d6b306514fde1a405d3b87cc946a56bd1/src/engine/threaded_engine.h#L335-L338

src/common/library.cc
leleamol: Ok.. Sounds good.

tests/python/unittest/test_library_loading.py
```diff
@@ -0,0 +1,38 @@
...
+    return platform.machine() not in ['x86_64', 'AMD64']
+
+@unittest.skipIf(check_platform(), "not all machine types supported")
+def test_library_loading():
+    if (os.name=='posix'):
+        lib = 'mylib.so'
+    elif (os.name=='nt'):
+        lib = 'mylib.dll'
+
+    fname = mx.test_utils.download('https://mxnet-demo-models.s3.amazonaws.com/lib_binary/'+lib)
```
marcoabreu: @mseth10 

Please get rid of this asap as this poses a security threat to our system as well as our users since this allows arbitrary code execution. I'd appreciate it if you would provide a replacement within the next 24h. https://github.com/apache/incubator-mxnet/pull/15755 is the corresponding revert PR.

larroy: I suggested before to build this stub during the build process instead of loading from S3.

marcoabreu: Agree

samskalicky:
Thanks @marcoabreu for the catch. As we discussed offline, the problem is not with the feature in this PR but with how test_library_loading is pulling from the S3 bucket in the CI. 

We'll work to build the library as part of the build so that it can be used internally in testing the CI without security risks. We'll move the test pulling from the S3 bucket to the examples directory so that it doesnt run in the CI. 

We may need some help @marcoabreu to get this working in the CI, since we'll need to build this library and then find the path where it lives during the test so we can pass the path to the `load` API

marcoabreu:
> We'll move the test pulling from the S3 bucket to the examples directory so that it doesnt run in the CI.

Could you elaborate?

samskalicky:
@marcoabreu we added an example directory at: example/lib_api. Currently the test.py includes a test that uses the locally built library (build using the included Makefile). But this only provides an example for how to build a library. 

We'll create another example python script that pulls from the S3 bucket and loads that one. But it will just be an example, not a unittest that will run in the CI. 

Do you have other ideas/suggestions on how we can improve testing or provide more examples?

marcoabreu:
No pulling of any prebuilt binaries due to the same reasons, please. I understand you trying to make it more convenient, but I think that we're stepping over a line in that case. Give instructions how to compile the example and the user should be fine.

Just to confirm: we will have a unit test running in ci at the end of the day, correct?

samskalicky:
@marcoabreu I think we're in agreement here. We will be distributing some pre-built libraries in an S3 bucket as examples. We will provide documented examples for how to pull and load these binaries. But these will not be tests that will be automatically run. They will be examples.

Yes, we will have the unit test running in the CI that uses a library that is built in the CI.

wkcn:
I have a suggestion. We can check the `sha1_hash` code of pre-built binary files in the S3 bucket before executing them.

marcoabreu:
While that eliviates the problem from a ci perspective (if the bucket would be owned by the CI owners), it's still a red flag for our users. 

I'm afraid that companies prefer to stay away from the execution of an arbitrary native library that is downloaded from the internet - no matter whether it is backed by a hash or not. It's inevitable in some cases, but we want to reduce the possiblity of compliance scanners flagging us and thus making the usage of the MXNet more complicated than it has to be.

I think the only acceptable way is if the library is getting produced as part of the build process that is entirely retrieved from source. These are also the tenants Apache employs - Apache project just not make any binary releases in any form. Thus let's best stick to that as well :)

PR #15490, opened 2019-07-08
Utility to help developers debug operators: Tensor Inspector

Zha0q1:
Created a separate PR for a new tutorial on usage.
https://github.com/apache/incubator-mxnet/pull/15517

## Background ##

When developing new operators, developers need to deal with Tensors, TBlobs, and NDArrays extensively (for simplicity, we will simply refer them as “tensors”). However, there is no existing tool that helps developers inspect the value of the tensors within an operator, so it’s hard to locate the spot where the computation goes wrong. With that said, we can create a new utility that supports inspecting, dumping and verifying tensor values to alleviate the developer’s need to use hacky prints every time. This new utility will support all the three data types (Tensor, TBlob, NDArray), as well as both CPU and GPU tensors.

Some related GitHub issues: https://github.com/apache/incubator-mxnet/issues/15198

## Methodology ##

First off, to support all three data types, we will need to first convert them to TBlob, so that we can base our new functionalists on one single class and avoid code repetition. To access the value of a tensor/TBlob object, the easiest way is to add some function, such as to_string(), to the TBlob class. Then, we could use shape_ to dereference dptr and extract the actual data from the tensor in a structured format. However, to preserve the integrity of the TBlob class, we would want to avoid making direct changes to tensor_blob.h. Instead, considering that all the member variables of TBlob are all public, we could create a wrapper around TensorBlob, which we would call TensorInspector, and we could then add our new functions to this new wrapper class. We could put this new file tensor_inspector.h in src/common along with a bunch of other utilities. 

```
class TensorInspector {
  private:
    const TBlob tb_;
  public:
    // our new functions
}
```

Besides tensor inspectors which are one-to-one with tensors, I also propose to have a singleton class called InspectorManage which is responsible for data sharing and maintaining locking between the individual TensorInspector objects. This way, we could control some global behaviors within each of the objects.

## The Value of Creating a New Utility ##

It’s important to point out that there is an existing utility that helps print out the value of a tensor, tests/cpp/include/test_util.h, (https://github.com/apache/incubator-mxnet/blob/master/tests/cpp/include/test_util.h#L472). However, this new utility under discussion aims to start from the use cases of the users, who in our case are developers, and has more functionalities than the existing utility. Note: this new utility will use CAccessAsCPU from test_util.h for data copying from GPU memory to CPU memory.

### Create a TensorInspector Object from Tensor, TBlob, and NDarray Objects ###

We can create a TensorInspector object by passing in a Tensor, TBlob, or NDArray object.

```
// Convolution-inl.h, Forward()
Tensor<xpu, 3, DType> weight_3d = in_data[conv::kWeight].get_with_shape<xpu, 3, DType>(
        Shape3(group_, M, K), s);
TensorInspector ti(weight_3d, ctx.run_ctx);

// Similarly, we could do:
TensorInspector ti(/*TBlob obj*/, ctx.run_ctx);
// OR:
TensorInspector ti(/*NDArray obj*/, ctx.run_ctx);
```

### Print Tensor Value (static) ###

We could use `std::string TensorInspector::to_string()` or `void TensorInspector::print_string()` to get a formatted string representation of the entire tensor. We could see the tensor value in a structured way, the data type, as well as the tensor shape. 

```
TensorInspector ti(/*Tensor, Tblob, or Ndarray obj*/, /*RunContext obj*/);
ti.to_string(); // print tensor
```

Screen Shot 2019-07-08 at 4 07 16 PM

### Interactively Print Tensor Value (dynamic) ###

During debugging, situations might occur that ahead of time (at compilation time), we do not know which part of a tensor we need to inspect. Also, sometimes we might want to pause the operator control flow to “zoom into” an erroneous part of a tensor multiple times until we are satisfied to move on. In this regard, we could use void `TensorInspector::interactive_print(std::string tag = "")` like a break point to dynamically specify the part of the tensor to inspect and whether we want to skip the break point. tag is an optional parameter where we can give the break points names so that we can distinguish them. A “Visit” counter would tell us how many times have we stepped into a break point with this tag name.

```
    TensorInspector ti(weight_3d);
    ti.interactive_print("My Convolution");
```

Screen Shot 2019-07-10 at 5 29 07 PM

As we can see from the screenshot, we would be in a loop that asks us for command. We would be prompted with the Tensor info, <f Tensor 20x1x5x5>, and we could enter coordinates such as “0,0” to print only a part of the tensor. We could also type “e” for the entire tensor, “b” to break from this loop, and “s” to break from this loop and skip all the future break points. 

### Check Tensor Value ###

Sometimes, developers might want to check if the tensor contains unexpected values which could be negative values, Nans, or others. We could have a templated api `std::vector<std::vector<int>>TensorInspector::validate_value(const ValueChecker& vc, bool interactive = false, std::string tag = "")` where vc is a bool value checker function that users can define, interactive is whether you want to check the values on the go (similar to interactive print, refer to the first screenshot below), and tag is the name you want to give to the call to identify it. The return is a vector of coordinates (represented with vector) at which checker evaluates to true. 

Screen Shot 2019-07-10 at 5 34 20 PM

To the developer's convenience, we would also pre-define a bunch of value checkers. Refer to the Enum below:

Screen Shot 2019-07-10 at 5 24 41 PM

Developers could skip wiring their own checker by using this API: `std::vector<std::vector<int>>TensorInspector::validate_value(ValueChecker vc, bool interactive = false, std::string tag = "")`

This API can also be used creatively to find coordinates where the values equal to some required value or fall in some range. 

### Dump Tensor Value  ###

Sometimes, we would want to dump the tensor to a file in binary mode, so that we could use, for example, a python script to further analyze the numbers. We could use void TensorInspector::dump_value(std::string tag) to achieve this. We would want to use .npy format for its compatibility with Numpy and its great load efficiency. A benchmark can be found here: https://github.com/mverleg/array_storage_benchmark.

Screen Shot 2019-07-10 at 5 45 44 PM

In this screenshot, we can compare side by side the regular print (left) and the value dump (right). Even without formatting the dumped value, we can see it has a better accuracy.

Screen Shot 2019-07-10 at 5 14 09 PM

Screen Shot 2019-07-10 at 5 14 43 PM

As we can see in the screenshot, the output files are named "{tag}_{visit}.npy", where tag is the name that we give to the call, and visit is the visit count. 

To load the file, we could just do:

```
import numpy as np
a = np.load('abc_1.npy')
print(a)
```

Screen Shot 2019-07-10 at 5 17 29 PM

Notice: in interactive print, we could also do value dump with command "d":

Screen Shot 2019-07-10 at 5 29 07 PM

## Usage ##
To test or use this new utility in your operator, just include the .h file and construct a TensorInspector object, say "ti", as described above. Then, you can use all functionalities listed above by writing, for example, "ti.print_string()" or "ti.interactive_print()". Finally, just run any python script that uses the operator that you modified.

Screen Shot 2019-07-11 at 4 57 41 PM

## TODOs ##
1. Support sparse NDArrays if necessary
2. Check corner cases e.g. if Tblob's shape is unknown

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/src/common/tensor_inspector.h
+++ b/src/common/tensor_inspector.h
@@ -128,7 +128,7 @@ class TensorInspector {
    * \param shape the shape of the tensor
    */
   template<typename DType, typename StreamType>
-  void tensor_info_to_string(StreamType* os, const std::vector<int>& shape) {
+  void tensor_info_to_string(StreamType* os, const std::vector<index_t>& shape) {
     const int dimension = shape.size();
     *os << "<" << typeid(tb_.dptr<DType>()[0]).name() << " Tensor ";
     *os << shape[0];
@@ -156,7 +156,7 @@ class TensorInspector {
     const int dimension = tb_.ndim();
     std::vector<index_t> multiples;
     index_t multiple = 1;
-    for (int i = dimension-1; i >= 0; --i) {
+    for (int i = dimension - 1; i >= 0; --i) {
       multiple *= tb_.shape_[i];
       multiples.push_back(multiple);
     }
@@ -207,7 +207,7 @@ class TensorInspector {
    * \param offset the position of the first value of the desired part of the tensor
    */
   template<typename DType, typename StreamType>
-  void to_string_helper(StreamType* os, const std::vector<int>& sub_shape, index_t offset) {
+  void to_string_helper(StreamType* os, const std::vector<index_t>& sub_shape, index_t offset) {
 #if MXNET_USE_CUDA
     if (tb_.dev_mask() == gpu::kDevMask) {
       TensorInspector(test::CAccessAsCPU(ctx_, tb_, false)(), ctx_)
@@ -223,7 +223,7 @@ class TensorInspector {
     const int dimension = sub_shape.size();
     std::vector<index_t> multiples;
     index_t multiple = 1;
-    for (int i = dimension-1; i >= 0; --i) {
+    for (int i = dimension - 1; i >= 0; --i) {
       multiple *= sub_shape[i];
       multiples.push_back(multiple);
     }
@@ -253,7 +253,8 @@ class TensorInspector {
    * \param sub_shape the sub-shape of the desired part of the tensor; calculated here
    * \param offset the position of the first value of the desired part of the tensor; calculated here
    */
-  void print_locator(const std::vector<int>& pos, std::vector<int>* sub_shape, index_t* offset) {
+  void print_locator(const std::vector<index_t>& pos, std::vector<index_t>* sub_shape,
+    index_t* offset) {
     const int dimension = tb_.ndim();
     int sub_dim = dimension - pos.size();
     sub_shape->resize(sub_dim);
@@ -264,7 +265,7 @@ class TensorInspector {
     }
     index_t sum = 0;
     index_t m = 1;
-    for (int i = pos.size()-1; i >= 0; --i) {
+    for (int i = pos.size() - 1; i >= 0; --i) {
       sum += pos[i] * m;
       m *= tb_.shape_[i];
     }
@@ -277,12 +278,12 @@ class TensorInspector {
    * \param pos the coordinates of the desired part of the tensor, calculated here
    * \param str the string that represents the coordinate
    */
-  bool parse_position(std::vector<int>* pos, const std::string& str) {
+  bool parse_position(std::vector<index_t>* pos, const std::string& str) {
     const int dimension = tb_.ndim();
     std::stringstream ss(str);
-    int i;
-    while (ss >> i) {
-      pos->push_back(i);
+    index_t n;
+    while (ss >> n) {
+      pos->push_back(n);
       if (ss.peek() == ',') {
         ss.ignore();
       }
@@ -291,7 +292,7 @@ class TensorInspector {
       return false;
     }
     for (unsigned i = 0; i < pos->size(); ++i) {
-      if ((*pos)[i] > (tb_.shape_[i]-1)) {
+      if ((*pos)[i] > (tb_.shape_[i] - 1)) {
         return false;
       }
     }
@@ -349,9 +350,9 @@ class TensorInspector {
         }
         continue;
       }
-      std::vector<int> pos;
+      std::vector<index_t> pos;
       if (parse_position(&pos, str)) {
-        std::vector<int> sub_shape;
+        std::vector<index_t> sub_shape;
         index_t offset;
         print_locator(pos, &sub_shape, &offset);
         to_string_helper<DType>(&std::cout, sub_shape, offset);
@@ -472,10 +473,10 @@ class TensorInspector {
    * \brief calculate the coordinate of a value in the tensor, given its index
    * \param idx the index of the value in the tensor
    */
-  std::vector<int> index_to_coordinates(index_t idx) {
+  std::vector<index_t> index_to_coordinates(index_t idx) {
     const int dimension = tb_.ndim();
-    std::vector<int> ret;
-    for (int i = dimension-1; i >= 0; --i) {
+    std::vector<index_t> ret;
+    for (int i = dimension - 1; i >= 0; --i) {
       ret.push_back(idx % tb_.shape_[i]);
       idx /= tb_.shape_[i];
     }
@@ -493,7 +494,7 @@ class TensorInspector {
    * \param tag the name given to this call
    */
   template<typename DType>
-  void check_value_helper(std::vector<std::vector<int>>* ret,
+  void check_value_helper(std::vector<std::vector<index_t>>* ret,
       const std::function<bool(DType)>& checker, bool interactive, std::string tag) {
 #if MXNET_USE_CUDA
     if (tb_.dev_mask() == gpu::kDevMask) {
@@ -512,7 +513,7 @@ class TensorInspector {
           ss  << ", ";
         }
         first_pass = false;
-        std::vector<int> coords = index_to_coordinates(i);
+        std::vector<index_t> coords = index_to_coordinates(i);
         ss << "(" << coords[0];
         for (unsigned int i = 1; i < coords.size(); ++i) {
           ss << ", " << coords[i];
@@ -527,22 +528,40 @@ class TensorInspector {
        InspectorManager::get()->check_value_tag_counter_[tag] += 1;
       while (!InspectorManager::get()->check_value_skip_all_) {
         std::cout << "----------Value Check----------" << std::endl;
+        tensor_info_to_string<DType>(&std::cout);
         if (tag != "") {
           std::cout << "Tag: " << tag << "  Visit: " <<
               InspectorManager::get()->check_value_tag_counter_[tag] <<  std::endl;
         }
         std::cout << count << " value(s) found." << std::endl;
-        std::cout << "\"p\" to print the coordinates, " <<
+        std::cout << "To print a part of the tensor," <<
+            " please specify a position, seperated by \",\"" << std::endl;
+        std::cout << "\"e\" for the entire tensor, " <<
+            "\"p\" to print the coordinates of the values found, " <<
             "\"b\" to break, " <<
             "\"s\" to skip all: ";
         std::string str;
         std::cin >> str;
         if (str == "b") {
           break;
+        } else if (str == "e") {
+          to_string_helper<DType>(&std::cout);
+          continue;
         } else if (str == "p") {
           std::cout << ss.str() << std::endl;
+          continue;
         } else if (str == "s") {
           InspectorManager::get()->check_value_skip_all_ = true;
+          break;
+        }
+        std::vector<index_t> pos;
+        if (parse_position(&pos, str)) {
+          std::vector<index_t> sub_shape;
+          index_t offset;
+          print_locator(pos, &sub_shape, &offset);
+          to_string_helper<DType>(&std::cout, sub_shape, offset);
+        } else {
+          std::cout << "invalid input" << std::endl;
         }
       }
     }
@@ -650,6 +669,18 @@ class TensorInspector {
     write_npy<DType>(header, filename);
   }
 
+   /*!
+   * \brief validate that the shape
+   */
+  inline void validate_shape() {
+    const int dimension = tb_.ndim();
+    CHECK(dimension > 0) << "Tensor Inspector does not support empty tensors " << 
+        "or tensors of unknow shape.";
+    for (int i = 0; i < dimension; ++i) {
+      CHECK(tb_.shape_[i] != 0) << "Invalid tensor shape: shape_[" << i << "] is 0";
+    }
+  }
+
   /* !\brief the tensor blob */
   const TBlob tb_;
   /* !\brief the run context of the tensor */
@@ -666,7 +697,9 @@ class TensorInspector {
    */
   template<typename Device, int dimension, typename DType>
   TensorInspector(const mshadow::Tensor<Device, dimension, DType>& ts, const RunContext& ctx):
-      tb_(ts), ctx_(ctx) {}
+      tb_(ts), ctx_(ctx) {
+    validate_shape();
+  }
 
   /*!
    * \brief construct from TBlob object
@@ -674,7 +707,9 @@ class TensorInspector {
    * \param ctx the run context of the tensor
    */
   TensorInspector(const TBlob& tb, const RunContext& ctx):
-      tb_(tb), ctx_(ctx) {}
+      tb_(tb), ctx_(ctx) {
+    validate_shape();
+  }
 
   /*!
    * \brief construct from NDArray object. Currently this only works with kDefaultStorage
@@ -682,7 +717,9 @@ class TensorInspector {
    * \param ctx the run context of the tensor
    */
   TensorInspector(const NDArray& arr, const RunContext& ctx):
-      tb_(arr.data()), ctx_(ctx) {}
+      tb_(arr.data()), ctx_(ctx) {
+    validate_shape();
+  }
 
   /*!
    * \brief print the tensor to std::cout
@@ -721,9 +758,9 @@ class TensorInspector {
    * \param tag the name given to this call
    */
   template<typename ValueChecker>
-  std::vector<std::vector<int>> check_value(const ValueChecker& checker,
+  std::vector<std::vector<index_t>> check_value(const ValueChecker& checker,
       bool interactive = false, std::string tag = "") {
-    std::vector<std::vector<int>> ret;
+    std::vector<std::vector<index_t>> ret;
     MSHADOW_TYPE_SWITCH(tb_.type_flag_, DType, {
       check_value_helper<DType>(&ret, checker, ret, interactive, tag);
     });
@@ -737,9 +774,9 @@ class TensorInspector {
    * \param interactive wherether to allow the user to interactively check the coordinates
    * \param tag the name given to this call
    */
-  std::vector<std::vector<int>> check_value(CheckerType ct, bool interactive = false,
+  std::vector<std::vector<index_t>> check_value(CheckerType ct, bool interactive = false,
       std::string tag = "") {
-    std::vector<std::vector<int>> ret;
+    std::vector<std::vector<index_t>> ret;
     MSHADOW_TYPE_SWITCH(tb_.type_flag_, DType, {
       check_value_helper<DType>(&ret, get_checker<DType>(ct), interactive, tag);
     });
--- a/src/common/tensor_inspector.h
+++ b/src/common/tensor_inspector.h
@@ -114,7 +114,7 @@ class TensorInspector {
     int dimension = tb_.ndim();
     *os << "<" << typeid(tb_.dptr<DType>()[0]).name() << " Tensor ";
     *os << tb_.shape_[0];
-    for (int i = 1; i < dimension; i++) {
+    for (int i = 1; i < dimension; ++i) {
       *os << 'x' << tb_.shape_[i];
     }
     *os << ">" << std::endl;
@@ -132,7 +132,7 @@ class TensorInspector {
     int dimension = shape.size();
     *os << "<" << typeid(tb_.dptr<DType>()[0]).name() << " Tensor ";
     *os << shape[0];
-    for (int i = 1; i < dimension; i++) {
+    for (int i = 1; i < dimension; ++i) {
       *os << 'x' << shape[i];
     }
     *os << ">" << std::endl;
@@ -154,15 +154,15 @@ class TensorInspector {
     }
 #endif  // MXNET_USE_CUDA
     int dimension = tb_.ndim();
-    std::vector<unsigned int> multiples;
-    int multiple = 1;
-    for (int i = dimension-1; i >= 0; i--) {
+    std::vector<size_t> multiples;
+    size_t multiple = 1;
+    for (int i = dimension-1; i >= 0; --i) {
       multiple *= tb_.shape_[i];
       multiples.push_back(multiple);
     }
     *os << std::string(dimension, '[');
     *os << tb_.dptr<DType>()[0];
-    for (size_t i = 1; i < tb_.shape_.Size(); i++) {
+    for (size_t i = 1; i < tb_.shape_.Size(); ++i) {
       int n = 0;
       for (auto divisor : multiples) {
         n += (i % divisor == 0);
@@ -221,16 +221,16 @@ class TensorInspector {
       return;
     }
     int dimension = sub_shape.size();
-    std::vector<int> multiples;
+    std::vector<size_t> multiples;
     size_t multiple = 1;
-    for (int i = dimension-1; i >= 0; i--) {
+    for (int i = dimension-1; i >= 0; --i) {
       multiple *= sub_shape[i];
       multiples.push_back(multiple);
     }
     std::stringstream ss;
     *os << std::string(dimension, '[');
     *os << dptr[0];
-    for (size_t i = 1; i < multiple; i++) {
+    for (size_t i = 1; i < multiple; ++i) {
       int n = 0;
       for (auto divisor : multiples) {
         n += (i % divisor == 0);
@@ -258,14 +258,14 @@ class TensorInspector {
     int dimension = tb_.ndim();
     int sub_dim = dimension - pos.size();
     sub_shape->resize(sub_dim);
-    int multiple = 1;
-    for (int i = pos.size(), j = 0; i < dimension; i++, j++) {
+    size_t multiple = 1;
+    for (int i = pos.size(), j = 0; i < dimension; ++i, ++j) {
       (*sub_shape)[j] = tb_.shape_[i];
       multiple *= tb_.shape_[i];
     }
-    int sum = 0;
-    int m = 1;
-    for (int i = pos.size()-1; i >= 0; i--) {
+    size_t sum = 0;
+    size_t m = 1;
+    for (int i = pos.size()-1; i >= 0; --i) {
       sum += pos[i] * m;
       m *= tb_.shape_[i];
     }
@@ -291,7 +291,7 @@ class TensorInspector {
     if (pos->size() > dimension) {
       return false;
     }
-    for (unsigned i = 0; i < pos->size(); i++) {
+    for (unsigned i = 0; i < pos->size(); ++i) {
       if ((*pos)[i] > (tb_.shape_[i]-1)) {
         return false;
       }
@@ -474,7 +474,7 @@ class TensorInspector {
   inline std::vector<int> index_to_coordinates(size_t idx) {
     int dimension = tb_.ndim();
     std::vector<int> ret;
-    for (int i = dimension-1; i >= 0; i--) {
+    for (int i = dimension-1; i >= 0; --i) {
       ret.push_back(idx % tb_.shape_[i]);
       idx /= tb_.shape_[i];
     }
@@ -500,11 +500,11 @@ class TensorInspector {
           .check_value_helper<DType>(ret, checker, interactive, tag);
     }
 #endif  // MXNET_USE_CUDA
-    int count = 0;
+    size_t count = 0;
     std::stringstream ss;
     ss << "[";
     bool first_pass = true;
-    for (size_t i = 0; i < tb_.shape_.Size(); i++) {
+    for (size_t i = 0; i < tb_.shape_.Size(); ++i) {
       if (checker(tb_.dptr<DType>()[i])) {
         count += 1;
         if (!first_pass) {
@@ -513,7 +513,7 @@ class TensorInspector {
         first_pass = false;
         std::vector<int> coords = index_to_coordinates(i);
         ss << "(" << coords[0];
-        for (size_t i = 1; i < coords.size(); i++) {
+        for (unsigned int i = 1; i < coords.size(); ++i) {
           ss << ", " << coords[i];
         }
         ss << ")";
@@ -590,7 +590,7 @@ class TensorInspector {
     dict += std::to_string(sizeof(DType));
     dict += "','fortran_order':False,'shape':(";
     dict += std::to_string(tb_.shape_[0]);
-    for (int i = 1; i < tb_.ndim(); i++) {
+    for (int i = 1; i < tb_.ndim(); ++i) {
       dict += ',';
       dict += std::to_string(tb_.shape_[i]);
     }
--- a/src/common/tensor_inspector.h
+++ b/src/common/tensor_inspector.h
@@ -451,7 +451,7 @@ class TensorInspector {
         if (std::is_same<DType, float>::value || std::is_same<DType, double>::value ||
             std::is_same<DType, mshadow::half::half_t>::value) {
           return [] (DType x) {
-                return x += (DType)1.0 / (DType)0.0 || x == -(DType)1.0 / (DType)0.0 &&
+                return x == (DType)1.0 / (DType)0.0 || x == -(DType)1.0 / (DType)0.0 ||
                     x != x;
               };
         } else {
--- a/src/common/tensor_inspector.h
+++ b/src/common/tensor_inspector.h
@@ -81,7 +81,7 @@ enum CheckerType {
                        // will always return false if DType is not a float type
   FiniteChecker,  // check if is finite, will always return false if DType is not a float type
   NormalChecker,  // check if is neither infinity nor NaN
-  AbnormalChecker, // chekc if is infinity or nan
+  AbnormalChecker,  // chekck if is infinity or nan
 };
 
 /**
@@ -484,7 +484,7 @@ class TensorInspector {
 
   /*!
    * \brief check/validate the values within the tensor, find the coordinates
-   * where the lambda evaluates to true
+   * where the value checker evaluates to true
    * \tparam DType the data type
    * \param ret a vector of coordinates which itself is a vector of int; calculated here
    * \param checker the lambda function to check each value of within the tensor
@@ -569,7 +569,8 @@ class TensorInspector {
   }
 
   /*!
-   * \brief dump the value of the tensor to a file with name "tag_[visit count].npy" in npy format
+   * \brief dump the value of the tensor to a file with name "[tag]_[visit count].npy" in npy format
+   * the dump file follows npy 1.0 stantand
    * \tparam DType the data type
    * \param tag the name given to this call
    */
@@ -615,6 +616,8 @@ class TensorInspector {
     file.write(header.c_str(), header.size());
     file.write(reinterpret_cast<char*>(tb_.dptr<DType>()), sizeof(DType) * tb_.shape_.Size());
     file.close();
+    std::cout << "Tensor dumped to file: " <<
+        tag + "_" + std::to_string(visit) + ".npy" << std::endl;
   }
 
   /* !\brief the tensor blob */
@@ -682,7 +685,7 @@ class TensorInspector {
 
   /*!
    * \brief check/validate the values within the tensor, return the coordinates
-   * where the lambda evaluates to true
+   * where the value checker evaluates to true
    * \tparam ValueChecker the type of the lambda
    * \param checker the lambda function to check each value of within the tensor
    * \param interactive wherether to allow the user to interactively check the coordinates
--- a/src/common/tensor_inspector.h
+++ b/src/common/tensor_inspector.h
@@ -63,7 +63,7 @@ struct InspectorManager {
   /* !\brief visit count for check value tags */
   std::unordered_map<std::string, int> check_value_tag_counter_;
   /* !\brief visit count for dump value tags */
-  std::unordered_map<std::string, int> dump_value_tag_counter_;
+  std::unordered_map<std::string, int> dump_to_file_tag_counter_;
 };
 
 /*!
@@ -109,9 +109,9 @@ class TensorInspector {
    * \tparam StreamType the type of the stream object
    * \param os stream object to output to
    */
-  template<typename DType MSHADOW_DEFAULT_DTYPE, typename StreamType>
+  template<typename DType, typename StreamType>
   inline void tensor_info_to_string(StreamType* os) {
-    int dimension = tb_.ndim();
+    const int dimension = tb_.ndim();
     *os << "<" << typeid(tb_.dptr<DType>()[0]).name() << " Tensor ";
     *os << tb_.shape_[0];
     for (int i = 1; i < dimension; ++i) {
@@ -127,9 +127,9 @@ class TensorInspector {
    * \param os stream object to output to
    * \param shape the shape of the tensor
    */
-  template<typename DType MSHADOW_DEFAULT_DTYPE, typename StreamType>
+  template<typename DType, typename StreamType>
   inline void tensor_info_to_string(StreamType* os, const std::vector<int>& shape) {
-    int dimension = shape.size();
+    const int dimension = shape.size();
     *os << "<" << typeid(tb_.dptr<DType>()[0]).name() << " Tensor ";
     *os << shape[0];
     for (int i = 1; i < dimension; ++i) {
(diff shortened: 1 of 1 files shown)


sandeep-krishnamurthy:
Awesome! thank you for the nice utility.  👍 for clear and detailed PR description.

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,554 @@
...
+  template<typename DType MSHADOW_DEFAULT_DTYPE>
+  inline std::function<bool(DType)> build_checker(CheckerType ct) {
+    switch (ct) {
+      case NegativeChecker:
+        return [] (DType x) {
+              return x < 0;
+            };
+      case PositiveChecker:
+        return [] (DType x) {
+              return x < 0;
```
kshitij12345: Probably should be `x > 0`.

Zha0q1: Good catch!

Zha0q1: fixed

Zha0q1:
I could probably rewrite build_checker() (https://github.com/apache/incubator-mxnet/pull/15490/files#diff-9e7d5c2420ecc900c4a85a3c35d91bffR446) with macro to suppress warnings when DType is not floating point. But it's not causing any issues because in such cases the control flow wouldn't go by the undefined operation. Will change if have time.

Zha0q1:
> I could probably rewrite build_checker() (https://github.com/apache/incubator-mxnet/pull/15490/files#diff-9e7d5c2420ecc900c4a85a3c35d91bffR446) with macro to suppress warnings when DType is not floating point. But it's not causing any issues because in such cases the control flow wouldn't go by the undefined operation. Will change if have time.

After experimenting for several hours today, I think there is not a good way to breach at compile time to suppress the warnings. Macro apparently did not work. I also tried simulating C++17 style `constexpr if` with templated structs, which is an idea proposed by https://baptiste-wicht.com/posts/2015/07/simulate-static_if-with-c11c14.html, but that method was compiler specific and did not work either.  

Having no warnings when compiling is nice, but given that this is a developer's tool and that behavior-wise the code has no issue, I think I will leave the warnings.

Zha0q1:
@apeforest @anirudh2290 @access2rohit @larroy @sandeep-krishnamurthy ,
Hi guys, I completed this PR and it's ready for review. The only thing that I do not have yet is support for CSR NDArrays because I want to ask your opinion.  I think I probably do not need to generate the whole matrix. Having something like this in my interactive print should be good enough? (this is the existing test::print)
Screen Shot 2019-07-10 at 6 49 08 PM

access2rohit:
Can you split the PR into 2 separate PRs. One for CPU and one for GPU ?

Zha0q1:
Just realized there were more than one Tensor structs. Added `shadow::` before `Tensor` so compiler would not be confused. https://github.com/apache/incubator-mxnet/pull/15490/commits/a41720d86aa69c3d7147f7ddfea2ea127079bb54#diff-9e7d5c2420ecc900c4a85a3c35d91bffR623

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,717 @@
...
+   */
+  inline void interactive_print(std::string tag = "") {
+    MSHADOW_TYPE_SWITCH(tb_.type_flag_, DType, {
+      interactive_print_helper<DType>(tag);
+    });
+  }
+
+  /*!
+   * \brief check/validate the values within the tensor, return the coordinates
+   * where the lambda evaluates to true
```
rondogency: can you elaborate which lambda you are referring?

Zha0q1: I meant "return the coordinates where the value checker evaluates to true". Will update the comment

Zha0q1:
Created a separate PR for a new tutorial on how to usage.
https://github.com/apache/incubator-mxnet/pull/15517

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,733 @@
...
+   * \tparam DType the data type
+   * \tparam StreamType the type of the stream object
+   * \param os stream object to output to
+   */
+  template<typename DType MSHADOW_DEFAULT_DTYPE, typename StreamType>
+  inline void tensor_info_to_string(StreamType* os) {
+    int dimension = tb_.ndim();
+    *os << "<" << typeid(tb_.dptr<DType>()[0]).name() << " Tensor ";
+    *os << tb_.shape_[0];
+    for (int i = 1; i < dimension; i++) {
```
apeforest: nit: use ++i for performance

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,733 @@
...
+   * \tparam StreamType the type of the stream object
+   * \param os stream object to output to
+   * \param shape the shape of the tensor
+   */
+  template<typename DType MSHADOW_DEFAULT_DTYPE, typename StreamType>
+  inline void tensor_info_to_string(StreamType* os, const std::vector<int>& shape) {
+    int dimension = shape.size();
+    *os << "<" << typeid(tb_.dptr<DType>()[0]).name() << " Tensor ";
+    *os << shape[0];
+    for (int i = 1; i < dimension; i++) {
```
apeforest: same here

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,733 @@
...
+            };
+    }
+    return [] (DType x) {return false;};
+  }
+
+  /*!
+   * \brief calculate the coordinate of a value in the tensor, given its index
+   * \param idx the index of the value in the tensor
+   */
+  inline std::vector<int> index_to_coordinates(size_t idx) {
```
apeforest: use `index_t` instead of `int` for index type.

apeforest:
@eric-haibin-lin This utility looks really useful for debugging mxnet backend. Please take a look if interested.

src/common/tensor_inspector.h
Zha0q1: I remember compiler will optimize this to ++i? Anyways, I will make the change myself

src/common/tensor_inspector.h
Zha0q1: you mean size_t?

src/common/tensor_inspector.h
Zha0q1: fixed

src/common/tensor_inspector.h
Zha0q1: fixed

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,733 @@
...
+ private:
+  /*!
+   * \brief generate the tensor info, including data type and shape 
+   * \tparam DType the data type
+   * \tparam StreamType the type of the stream object
+   * \param os stream object to output to
+   */
+  template<typename DType MSHADOW_DEFAULT_DTYPE, typename StreamType>
+  inline void tensor_info_to_string(StreamType* os) {
+    int dimension = tb_.ndim();
```
larroy: const?

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,733 @@
...
+
+  /*!
+   * \brief output the tensor info, including data type and shape 
+   * \tparam DType the data type
+   * \tparam StreamType the type of the stream object
+   * \param os stream object to output to
+   * \param shape the shape of the tensor
+   */
+  template<typename DType MSHADOW_DEFAULT_DTYPE, typename StreamType>
+  inline void tensor_info_to_string(StreamType* os, const std::vector<int>& shape) {
```
larroy: why inlines?

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,733 @@
...
+  template<typename DType MSHADOW_DEFAULT_DTYPE, typename StreamType>
+  inline void to_string_helper(StreamType* os) {
+#if MXNET_USE_CUDA
+    if (tb_.dev_mask() == gpu::kDevMask) {
+      TensorInspector(test::CAccessAsCPU(ctx_, tb_, false)(), ctx_)
+          .to_string_helper<DType>(os);
+      return;
+    }
+#endif  // MXNET_USE_CUDA
+    int dimension = tb_.ndim();
```
larroy: const?

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,733 @@
...
+    return (reinterpret_cast<char*>(&x)[0]) ? '<' : '>';
+  }
+
+  /*!
+   * \brief dump the value of the tensor to a file with name "[tag]_[visit count].npy" in npy format
+   * the dump file follows npy 1.0 stantand
+   * \tparam DType the data type
+   * \param tag the name given to this call
+   */
+  template<typename DType MSHADOW_DEFAULT_DTYPE>
```
larroy: why the template has two types? I'm confused. Is this a new C++ feature that I'm not aware of?

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,733 @@
...
+    std::string header;
+    header += static_cast<char>(0x93);
+    header += "NUMPY";
+    header += static_cast<char>(0x01);
+    header += static_cast<char>(0x00);
+    header += static_cast<char>((uint16_t)dict.size() & 0x00ff);
+    header += static_cast<char>(((uint16_t)dict.size() >> 8) & 0x00ff);
+    header += dict;
+    InspectorManager::get()->dump_value_tag_counter_[tag] += 1;
+    int visit = InspectorManager::get()->dump_value_tag_counter_[tag];
```
larroy: const?

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,733 @@
...
+    std::ofstream file(tag + "_" + std::to_string(visit) + ".npy",
+        std::ios::out | std::ios::binary);
+    file.write(header.c_str(), header.size());
```
larroy: can you consider using file.exceptions so an exceptions would be raised when IO failure occurs?  In this case I think it can go unnoticed, if my memory doesn't fail.

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,733 @@
...
+      dict += ',';
+      dict += std::to_string(tb_.shape_[i]);
+    }
+    if (tb_.ndim() == 1) {
+       dict += ",";
+    }
+    dict += ")} ";
+    int padding_size = 64 - ((10 + dict.size()) % 64);
+    dict += std::string(padding_size, ' ');
+    dict[dict.size()-1] = '\n';
```
larroy: Consider using .back() for more clarity.

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,733 @@
...
+      check_value_helper<DType>(&ret, build_checker<DType>(ct), interactive, tag);
+    });
+    return ret;
+  }
+
+  /*!
+   * \brief dump the value of the tensor to a file with name "tag_[visit count].npy" in npy format
+   * \param tag the name given to this call
+   */
+  inline void dump_value(std::string tag) {
```
larroy: Could we refine the name to dump to file or similar? is not clear where are we dumping to from the name of the function.

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,733 @@
...
+  }
+
+  /*!
+   * \brief dump the value of the tensor to a file with name "[tag]_[visit count].npy" in npy format
+   * the dump file follows npy 1.0 stantand
+   * \tparam DType the data type
+   * \param tag the name given to this call
+   */
+  template<typename DType MSHADOW_DEFAULT_DTYPE>
+  inline void dump_value_helper(const std::string& tag) {
```
larroy: Can we factor out formatting and IO in separate fns? One function does the formatting and another one does the IO, this is more testable and modular. Also allows to reuse the formatting function....

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,733 @@
...
+    }
+  }
+
+  /*!
+   * \brief build the lambda function, aka the checker, given its type
+   * \tparam DType the data type
+   * \param ct the type of the checker
+   */
+  template<typename DType MSHADOW_DEFAULT_DTYPE>
+  inline std::function<bool(DType)> build_checker(CheckerType ct) {
```
larroy: Please consider refining the name, get_checker or similar should be enough if you are returning just a function. Also why so many inlines? Please only use inline if necessary.

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,733 @@
...
+    InspectorManager::get()->interactive_print_tag_counter_[tag] += 1;
+    while (!InspectorManager::get()->interactive_print_skip_all_) {
+      std::cout << "----------Interactive Print----------" << std::endl;
+      if (tag != "") {
+        std::cout << "Tag: " << tag << "  Visit: " <<
+            InspectorManager::get()->interactive_print_tag_counter_[tag] <<  std::endl;
+      }
+      tensor_info_to_string(&std::cout);
+      std::cout << "Please specify the position, seperated by \",\"" << std::endl;
+      std::cout << "\"e\" for the entire tensor, \"d\" to dump value to file," <<
```
larroy: shall we separate the commands in different lines for readability?

Zha0q1:
> Nice work, just a few comments, looks good.

Thank you for the detailed advices!

src/common/tensor_inspector.h
Zha0q1: I copied the code from `tensor.h`. I think `MSHADOW_DEFAULT_DTYPE` is a macro used for debugging. Probably. I think I don't need it anyways.

src/common/tensor_inspector.h
Zha0q1: changed

src/common/tensor_inspector.h
Zha0q1: fixed

src/common/tensor_inspector.h
Zha0q1: added

src/common/tensor_inspector.h
Zha0q1: added

src/common/tensor_inspector.h
Zha0q1: removed

src/common/tensor_inspector.h
Zha0q1: added

src/common/tensor_inspector.h
Zha0q1: name changed

src/common/tensor_inspector.h
Zha0q1: Changed

src/common/tensor_inspector.h
Zha0q1: Yes!

Zha0q1: separated them

src/common/tensor_inspector.h
Zha0q1: Yeah you are right! I added exception control now

src/common/tensor_inspector.h
Zha0q1: I am a little confused. I know inlines are used to improve performance.

This post https://stackoverflow.com/questions/1932311/when-to-use-inline-function-and-when-not-to-use-it says there are situations where inlines should be avoided. However, I also checked other files in our repo, e.g. https://github.com/apache/incubator-mxnet/blob/master/tests/cpp/include/test_util.h. This file uses inline for every function

larroy: The post that you linked seems good. 

```
use inline instead of #define
very small functions are good candidates for inline: faster code and smaller executables (more chances to stay in the code cache)
the function is small and called very often
```

The compiler knows when to inline functions. Abusing inlines leads to code bloat, and missing stack traces when debugging. The cost overhead of a function call is very minimal, my recommendation is to only explicitly inline if you are doing something very small like returning a value or an arithmetic operation, if you have something more complex like appending to a vector, or a small loop it doesn't make any sense, does more harm than good, and again the compiler will choose when to inline. 

If you have better information on this, feel free to correct me or point to an updated source of information.

We have a code bloat problem in MXNet that surfaces on linking errors, this is partially caused by abuse of inlines like in the file you linked.

Zha0q1: I see! I will make the changes. Thanks for spending the time to look into this!

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,763 @@
...
+    *os << "<" << typeid(tb_.dptr<DType>()[0]).name() << " Tensor ";
+    *os << shape[0];
+    for (int i = 1; i < dimension; ++i) {
+      *os << 'x' << shape[i];
+    }
+    *os << ">" << std::endl;
+  }
+
+  /*!
+   * \brief output the tensor in a structed format
```
ChaiBapchya:
```suggestion
   * \brief output the tensor as a struct
```
Or
```suggestion
   * \brief output the tensor in a struct format
```

Zha0q1: Oh I actually meant "structured format" there. Sorry about the confusion

Zha0q1: I will correct it along with other changes in the next commit

src/common/tensor_inspector.h
apeforest: I meant index_t. This data type is configurable to int64_t if large tensor support is needed: https://github.com/dmlc/mshadow/blob/1d79ecfdb4c9234537e1bf5148f44a1af54501ec/mshadow/base.h#L289

karan6181:
@mxnet-label-bot add [Operator, NDArray, Visualization]

src/common/tensor_inspector.h
Zha0q1: fixed

src/common/tensor_inspector.h
Zha0q1: fixed

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,762 @@
...
+#endif  // MXNET_USE_CUDA
+    DType* dptr = tb_.dptr<DType>() + offset;
+    if (sub_shape.size() == 0) {
+      to_string_helper<DType>(os, dptr);
+      return;
+    }
+    const int dimension = sub_shape.size();
+    std::vector<index_t> multiples;
+    index_t multiple = 1;
+    for (int i = dimension-1; i >= 0; --i) {
```
apeforest: nit: space between '-'

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,762 @@
...
+      multiple *= sub_shape[i];
+      multiples.push_back(multiple);
+    }
+    std::stringstream ss;
+    *os << std::string(dimension, '[');
+    *os << dptr[0];
+    for (index_t i = 1; i < multiple; ++i) {
+      int n = 0;
+      for (auto divisor : multiples) {
+        n += (i % divisor == 0);
```
apeforest: Would it be possible that divisor is zero? What if some sub_shape[i] == 0?

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,762 @@
...
+  /*!
+   * \brief parse the coordinate of the desired part of the tensor, given a string that represents that
+   * coordinate
+   * \param pos the coordinates of the desired part of the tensor, calculated here
+   * \param str the string that represents the coordinate
+   */
+  bool parse_position(std::vector<int>* pos, const std::string& str) {
+    const int dimension = tb_.ndim();
+    std::stringstream ss(str);
+    int i;
```
apeforest: i is not initialized here

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,762 @@
...
+    while (ss >> i) {
+      pos->push_back(i);
+      if (ss.peek() == ',') {
+        ss.ignore();
+      }
+    }
+    if (pos->size() > dimension) {
+      return false;
+    }
+    for (unsigned i = 0; i < pos->size(); ++i) {
```
apeforest: Although it's still correct, it would be better to use a different variable than i here for better readability since you already have a variable i in this block.

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,762 @@
...
+  }
+
+  /*!
+   * \brief helper function to calculate the sub_shape and offset for the desired part of the tensor,
+   * given its coordinates in the original tensor
+   * \param pos the coordinates of the desired part of the tensor
+   * \param sub_shape the sub-shape of the desired part of the tensor; calculated here
+   * \param offset the position of the first value of the desired part of the tensor; calculated here
+   */
+  void print_locator(const std::vector<int>& pos, std::vector<int>* sub_shape, index_t* offset) {
```
apeforest: Use index_t instead of int for pos vector. Essentially index to an element inside a tensor should have data type `index_t`

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,763 @@
...
+      }
+      tensor_info_to_string<DType>(&std::cout);
+      std::cout << "Please specify the position, seperated by \",\"" << std::endl;
+      std::cout << "\"e\" for the entire tensor, " <<
+          "\"d\" to dump value to file, " <<
+          "\"b\" to break, " <<
+          "\"s\" to skip all: ";
+      std::string str;
+      std::cin >> str;
+      if (str == "b") {
```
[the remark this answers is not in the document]
Zha0q1: Yeah we have that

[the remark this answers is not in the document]
Zha0q1: Refer to this screenshot: we are prompted the type and shape of the tensor. Also, if you type in invalid coordinates, they will be rejected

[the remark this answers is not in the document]
Zha0q1: <f Tensor 20x1x5x5>

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,763 @@
...
+        ss << ")";
+        ret->push_back(coords);
+      }
+    }
+    ss << "]" << std::endl;
+    if (interactive) {
+      std::lock_guard<std::mutex> lock(InspectorManager::get()->mutex_);
+       InspectorManager::get()->check_value_tag_counter_[tag] += 1;
+      while (!InspectorManager::get()->check_value_skip_all_) {
+        std::cout << "----------Value Check----------" << std::endl;
```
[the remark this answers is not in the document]
Zha0q1: I got that covered already!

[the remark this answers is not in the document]
Zha0q1: Let's say we have a tensor of shape 20*1*5*5, then if you enter say `0,0` it will print a 5*5 tensor. If you enter `0,0,0,0` the if will print just a single value

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,763 @@
...
+   * \brief build the lambda function, aka the checker, given its type
+   * \tparam DType the data type
+   * \param ct the type of the checker
+   */
+  template<typename DType>
+  std::function<bool(DType)> get_checker(CheckerType ct) {
+    switch (ct) {
+      case NegativeChecker:
+        return [] (DType x) {
+              return x < 0;
```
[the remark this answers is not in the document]
Zha0q1: I think the warnings are unavoidable if we want to have a templated function to return a bunch of value checkers (lambdas). C++ 17 has `constexpr` which can be used to our advantage to do compile type branching. I tried to simulate that using templates tricks (https://baptiste-wicht.com/posts/2015/07/simulate-static_if-with-c11c14.html) but it did not work on our compiler

src/common/tensor_inspector.h
Zha0q1: `I` is used in "while (ss >> i)" where `ss` is a `std::stringstraem`. So I think it's fine?

src/common/tensor_inspector.h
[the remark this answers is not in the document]
Zha0q1: Good point. Will add this

src/common/tensor_inspector.h
Zha0q1: changed

src/common/tensor_inspector.h
Zha0q1: good catch!

src/common/tensor_inspector.h
Zha0q1: fixed

src/common/tensor_inspector.h
Zha0q1: I think sub_shape[i] == 0 will never happen. But I can add an assert to check

src/common/tensor_inspector.h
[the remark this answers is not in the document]
anirudh2290: yes but lets say after i do e twice and then I have to print shape and type, there is no easy way to do so.

src/common/tensor_inspector.h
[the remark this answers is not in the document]
Zha0q1: added!

src/common/tensor_inspector.h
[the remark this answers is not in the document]
Zha0q1: Even without entering any command you know the type and shape. It's on the second line of the screenshot above. Or if I understand you incorrectly can you elaborate more on this use case?

src/common/tensor_inspector.h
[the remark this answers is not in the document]
Zha0q1: Using DEV = 1 will turn the warnings into errors and thus compilation will fail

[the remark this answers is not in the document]
Zha0q1: So I should fix those since developers will build with `DEV = 1`?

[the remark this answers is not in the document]
anirudh2290: since this is just a debugging tool i think it is fine.

src/common/tensor_inspector.h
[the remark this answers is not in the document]
anirudh2290: For example, lets say user types e once. Now he wants to know the shape. But the tensor is so huge that he cannot scroll back to see the shape . he should be able to use just some key to type the shape in this case.

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,799 @@
...
+          .to_string_helper<DType>(os);
+      return;
+    }
+#endif  // MXNET_USE_CUDA
+    const int dimension = tb_.ndim();
+    std::vector<index_t> multiples;
+    index_t multiple = 1;
+    for (int i = dimension - 1; i >= 0; --i) {
+      multiple *= tb_.shape_[i];
+      multiples.push_back(multiple);
```
access2rohit: Make sure you name variables so it is understandable that it is related to a Tensor

multiples -> linear_offset
divisor -> index_size

src/common/tensor_inspector.h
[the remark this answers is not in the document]
anirudh2290: giving it a second look, stages like centos do build wiht make DEV=1 and they should fail in CI. I think its probably CI issue related to caching because of which they are not failing. Please check this otherwise CI may start failing after PR is merged.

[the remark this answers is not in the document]
Zha0q1: Sure

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,799 @@
...
+    }
+    *os << std::string(dimension, '[');
+    *os << tb_.dptr<DType>()[0];
+    for (index_t i = 1; i < tb_.shape_.Size(); ++i) {
+      int n = 0;
+      for (auto divisor : multiples) {
```
access2rohit: same as above

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,799 @@
...
+    for (int i = dimension - 1; i >= 0; --i) {
+      multiple *= sub_shape[i];
+      multiples.push_back(multiple);
+    }
+    std::stringstream ss;
+    *os << std::string(dimension, '[');
+    *os << dptr[0];
+    for (index_t i = 1; i < multiple; ++i) {
+      int n = 0;
+      for (auto divisor : multiples) {
```
access2rohit: same as above

src/common/tensor_inspector.h
apeforest: why cast i?

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,800 @@
...
+          return [] (DType x) {
+                return x == (DType)1.0 / 0.0f || x == -(DType)1.0 / 0.0f;
+              };
+        } else {
+          LOG(WARNING) << "InfChecker only applies to float types. " <<
+              "Lambda will always return false.";
+        }
+        break;
+      case PositiveInfChecker:
+        if (std::is_same<DType, float>::value || std::is_same<DType, double>::value ||
```
apeforest: Can we just use std::isinf?

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,800 @@
...
+      case PositiveChecker:
+        return [] (DType x) {
+              return x > 0;
+            };
+      case ZeroChecker:
+        return [] (DType x) {
+              return x == 0;
+            };
+      case NaNChecker:
+        if (std::is_same<DType, float>::value || std::is_same<DType, double>::value ||
```
apeforest: Can we just use std::isnan? http://www.cplusplus.com/reference/cmath/isnan/

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,800 @@
...
+          return [] (DType x) {
+                return x != (DType)1.0 /  0.0f && x != -(DType)1.0 /  0.0f;
+              };
+        } else {
+          LOG(WARNING) << "FiniteChecker only applies to float types. " <<
+              "Lambda will always return false.";
+        }
+        break;
+      case NormalChecker:
+        if (std::is_same<DType, float>::value || std::is_same<DType, double>::value ||
```
apeforest: http://www.cplusplus.com/reference/cmath/isnormal/

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,800 @@
...
+    const int dimension = tb_.ndim();
+    std::vector<index_t> multiples;
+    index_t multiple = 1;
+    for (int i = dimension - 1; i >= 0; --i) {
+      multiple *= tb_.shape_[i];
+      multiples.push_back(multiple);
+    }
+    *os << std::string(dimension, '[');
+    *os << tb_.dptr<DType>()[0];
+    for (index_t i = 1; static_cast<size_t>(i) < tb_.shape_.Size(); ++i) {
```
apeforest: Why cast i? Either cast tb_.shape_.Size() or simply declare size_t i

src/common/tensor_inspector.h
Zha0q1: The issue is, get_checker is templated, and there is no isinf() for integer types. The code won't compile then

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,800 @@
...
+
+  /*!
+   * \brief parse the coordinate of the desired part of the tensor, given a string that represents that
+   * coordinate
+   * \param pos the coordinates of the desired part of the tensor, calculated here
+   * \param str the string that represents the coordinate
+   */
+  bool parse_position(std::vector<index_t>* pos, const std::string& str) {
+    const int dimension = tb_.ndim();
+    std::stringstream ss(str);
```
apeforest: use istringstream if you only need to read from it.

src/common/tensor_inspector.h
Zha0q1: sadly we are not using c++ 17. otherwise we could do static branching and use isinf()

src/common/tensor_inspector.h
Zha0q1: changed

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,800 @@
...
+    return !pos->empty();
+  }
+
+  /*!
+   * \brief interactive print the tensor value
+   * \tparam DType the data type
+   * \param tag the name given to this call
+   */
+  template<typename DType>
+  void interactive_print_helper(std::string tag) {
```
apeforest: Nice! This must look like a black magic to python developers :)

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,800 @@
...
+    if (tb_.dev_mask() == gpu::kDevMask) {
+      return TensorInspector(test::CAccessAsCPU(ctx_, tb_, false)(), ctx_)
+          .check_value_helper<DType>(ret, checker, interactive, tag);
+    }
+#endif  // MXNET_USE_CUDA
+    index_t count = 0;
+    std::stringstream ss;
+    ss << "[";
+    bool first_pass = true;
+    for (index_t i = 0; static_cast<size_t>(i) < tb_.shape_.Size(); ++i) {
```
apeforest: don't cast i.

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,800 @@
...
+      if (checker(tb_.dptr<DType>()[i])) {
+        count += 1;
```
apeforest: ++count

src/common/tensor_inspector.h
Zha0q1: changed

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,800 @@
...
+        if (!first_pass) {
+          ss  << ", ";
+        }
+        first_pass = false;
+        std::vector<index_t> coords = index_to_coordinates(i);
+        ss << "(" << coords[0];
+        for (unsigned int i = 1; i < coords.size(); ++i) {
```
apeforest: use size_t for i

Zha0q1: changed

src/common/tensor_inspector.h
Zha0q1: changed

src/common/tensor_inspector.h
apeforest: FYI, unsigned is not a good candidate when you use it as data type: https://google.github.io/styleguide/cppguide.html#Integer_Types

src/common/tensor_inspector.h
Zha0q1: changed

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,800 @@
...
+
+  /*!
+   * \brief helper function to calculate the sub_shape and offset for the desired part of the tensor,
+   * given its coordinates in the original tensor
+   * \param pos the coordinates of the desired part of the tensor
+   * \param sub_shape the sub-shape of the desired part of the tensor; calculated here
+   * \param offset the position of the first value of the desired part of the tensor; calculated here
+   */
+  void print_locator(const std::vector<index_t>& pos, std::vector<index_t>* sub_shape,
+    index_t* offset) {
```
access2rohit: nit: indent

Zha0q1: fixed

src/common/tensor_inspector.h
Zha0q1: fixed

src/common/tensor_inspector.h
Zha0q1: fixed

src/common/tensor_inspector.h
Zha0q1: fixed

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,800 @@
...
+    *offset = sum * multiple;
+  }
+
+  /*!
+   * \brief parse the coordinate of the desired part of the tensor, given a string that represents that
+   * coordinate
+   * \param pos the coordinates of the desired part of the tensor, calculated here
+   * \param str the string that represents the coordinate
+   */
+  bool parse_position(std::vector<index_t>* pos, const std::string& str) {
```
access2rohit: How are you handling erroneous input indices ?

Zha0q1: Now I added checks for overly large and negative indices

access2rohit:
LGTM just address my last comment and no need to split this PR into 2 now since you have already addressed most comments

access2rohit:
retrigger the build ... there is some julia error. Probably not because of your code changes. Also add link to the test cases that you covered

Zha0q1:
> retrigger the build ... there is some julia error. Probably not because of your code changes. Also add link to the test cases that you covered

cased covered are described in my tutorial at the end. Link: https://github.com/apache/incubator-mxnet/pull/15517

Zha0q1:
@sandeep-krishnamurthy Looks ready!

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,815 @@
...
+    while (ss >> n) {
+      pos->push_back(n);
+      if (ss.peek() == ',') {
+        ss.ignore();
+      }
+    }
+    if (pos->size() > static_cast<size_t>(dimension)) {
+      return false;
+    }
+    for (unsigned i = 0; i < pos->size(); ++i) {
```
apeforest: replace `unsigned` with `size_t`

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,815 @@
...
+   * \param sub_shape the sub-shape of the desired part of the tensor; calculated here
+   * \param offset the position of the first value of the desired part of the tensor; calculated here
+   */
+  void print_locator(const std::vector<index_t>& pos, std::vector<index_t>* sub_shape,
+      index_t* offset) {
+    const int dimension = tb_.ndim();
+    const int sub_dim = dimension - pos.size();
+    sub_shape->resize(sub_dim);
+    index_t multiple = 1;
+    for (int i = pos.size(), j = 0; i < dimension; ++i, ++j) {
```
apeforest: change int to index_t or size_t

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,815 @@
...
+      (*sub_shape)[j] = tb_.shape_[i];
+      multiple *= tb_.shape_[i];
+    }
+    index_t sum = 0;
+    index_t m = 1;
+    for (int i = pos.size() - 1; i >= 0; --i) {
```
apeforest: change int to index_t

apeforest:
Three more changes. Otherwise good to go. Thanks for your contribution.

Zha0q1:
> @sandeep-krishnamurthy Looks ready!

After a few more changes requested by Lin

src/common/tensor_inspector.h
```diff
@@ -0,0 +1,815 @@
...
+namespace mxnet {
+
+/*!
+ * \brief this singleton struct mediates individual TensorInspector objects
+ * so that we can control the global behavior from each of them
+ */
+struct InspectorManager {
+  static InspectorManager* get() {
+    static std::mutex mtx;
+    static std::unique_ptr<InspectorManager> im = nullptr;
```
larroy: Could you add an utility function encapsulating this idiom? this is being used in several places, and would be good to use this.

Issue #15464, opened 2019-07-04
MKL-DNN gives wrong convolution bias gradient if weights gradient is not requested


matteosal:
## Description
When using MKL-DNN and asking the gradient of a convolution with respect to its biases, the result is wrong unless the gradient with respect to the weights is also requested. 

## Environment info (Required)

```
----------Python Info----------
Version      : 3.7.2
Compiler     : GCC 7.3.0
Build        : ('default', 'Dec 29 2018 06:19:36')
Arch         : ('64bit', '')
------------Pip Info-----------
Version      : 19.0.1
Directory    : /opt/Anaconda/lib/python3.7/site-packages/pip
----------MXNet Info-----------
Version      : 1.5.0
Directory    : /home/matteo/Git/mxnet/python/mxnet
Hashtag not found. Not installed from pre-built package.
----------System Info----------
Platform     : Linux-4.15.0-54-generic-x86_64-with-debian-buster-sid
system       : Linux
node         : mongolius
release      : 4.15.0-54-generic
version      : #58-Ubuntu SMP Mon Jun 24 10:55:24 UTC 2019
----------Hardware Info----------
machine      : x86_64
processor    : x86_64
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              8
On-line CPU(s) list: 0-7
Thread(s) per core:  2
Core(s) per socket:  4
Socket(s):           1
NUMA node(s):        1
Vendor ID:           GenuineIntel
CPU family:          6
Model:               94
Model name:          Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
Stepping:            3
CPU MHz:             2700.094
CPU max MHz:         3500,0000
CPU min MHz:         800,0000
BogoMIPS:            5184.00
Virtualization:      VT-x
L1d cache:           32K
L1i cache:           32K
L2 cache:            256K
L3 cache:            6144K
NUMA node0 CPU(s):   0-7
[...]

PR #15497, opened 2019-07-09
[MKLDNN] Independent gradients requests check with respect to weights and bias of convolution

zixuanweeei:
## Description ##
As it was described in #15464, MXNet with MKL-DNN gives a wrong gradient of a convolution with respect to its biases unless the gradient with respect to its weights is also requested. In the implement of convolution with MKL-DNN, only request for gradients with respect to weights is checked. It should be checked independently for bias.

### Changes ###
- [x] Independently check the requests for gradients with respect to weights and bias of convolution.
- [x] Convolution can give the gradient with respect to any one of weights or bias.

## Comments ##
- No comments.


zixuanweeei:
@pengzhao-intel @ciyongch @TaoLv Please help me review on this PR. Thanks. 😃

pengzhao-intel:
Could you try to add a UT for this case?

zixuanweeei:
> Could you try to add a UT for this case?

Sure. The exists UT passed with this PR on a local test. I will add some UTs for checking the correctness of the results of the gradients' requests combinations.

src/operator/nn/mkldnn/mkldnn_convolution.cc
```diff
@@ -662,21 +662,21 @@ void MKLDNNConvolutionBackward(const nnvm::NodeAttrs& attrs, const OpContext &ct
...
+    if (!param.no_bias && req[conv::kBias]) {
+      auto in_grad_bias = CreateMKLDNNMem(
           in_grad[conv::kBias],
           convBwdWeight.bwdWeights_pd.diff_bias_primitive_desc(), req[conv::kBias]);
       convBwdWeight.SetWeightNewMem(*data_mem, *out_grad_mem,
-                              *in_grad_weight.second, *in_grad_bias.second);
+          *in_grad_weight.second, *in_grad_bias.second);
       MKLDNNStream::Get()->RegisterPrim(convBwdWeight.GetBwdWeights());
       CommitOutput(in_grad[conv::kBias], in_grad_bias);
+    } else {
```
TaoLv: Suggest to check req[conv::kWeight] here.

TaoLv:
@matteosal Could you help to verify this PR with the test case in your project?

src/operator/nn/mkldnn/mkldnn_convolution.cc
zixuanweeei: Sure. I see. There is unnecessary primitive registration without the check enabled. Thanks.

src/operator/nn/mkldnn/mkldnn_convolution.cc
```diff
@@ -662,21 +662,21 @@ void MKLDNNConvolutionBackward(const nnvm::NodeAttrs& attrs, const OpContext &ct
...
+          *in_grad_weight.second, *in_grad_bias.second);
       MKLDNNStream::Get()->RegisterPrim(convBwdWeight.GetBwdWeights());
       CommitOutput(in_grad[conv::kBias], in_grad_bias);
+    } else if (req[conv::kWeight]) {
+      convBwdWeight.SetWeightNewMem(*data_mem, *out_grad_mem,
+          *in_grad_weight.second);
+      MKLDNNStream::Get()->RegisterPrim(convBwdWeight.GetBwdWeights());
     }
-    CommitOutput(in_grad[conv::kWeight], in_grad_weight);
+    if (req[conv::kWeight]) CommitOutput(in_grad[conv::kWeight], in_grad_weight);
```
pengzhao-intel: what's the behavior of req[conv::bias]?

zixuanweeei: It has the same behavior as req[kWeight]. Both of them return the operation request type (OpReqType) to Forward and Backward. We can use it to control the behavior of handling memory of result, like add/copy the result back to the source memory or just do nothing with them.

matteosal:
The example of https://github.com/apache/incubator-mxnet/issues/15464 is fixed here, but I see a failure with this one, where the weights gradient is requested in isolation (so the opposite of https://github.com/apache/incubator-mxnet/issues/15464 ):
```
import mxnet as mx

sym = mx.sym.Convolution(
	mx.sym.Variable('in'), 
	mx.sym.Variable('w'), 
	mx.sym.Variable('b'),
	kernel=(1, 1), 
	num_filter=1
)
args = {
	'in': mx.nd.ones([1, 1, 3, 3]),
	'w': mx.nd.ones([1, 1, 1, 1]),
	'b': mx.nd.ones([1]),
}
grad = {
	'in': mx.nd.zeros([1, 1, 3, 3]),
	'w': mx.nd.zeros([1, 1, 1, 1]),
	'b': mx.nd.zeros([1]),
}
req = {'in': 'null', 'w': 'write', 'b': 'null'}
outgrad = mx.nd.ones([1, 1, 3, 3])

ex = sym.bind(mx.cpu(), args, args_grad=grad, grad_req=req)

ex.forward(True);
ex.backward(out_grads=outgrad);
mx.ndarray.waitall()
```
This is what gets printed to command line:
```
  File "script2.py", line 27, in <module>
    mx.ndarray.waitall()
  File "/home/matteo/Git/mxnet/python/mxnet/ndarray/ndarray.py", line 166, in waitall
    check_call(_LIB.MXNDArrayWaitAll())
  File "/home/matteo/Git/mxnet/python/mxnet/base.py", line 253, in check_call
mxnet.base.MXNetError: std::exception
```

It doesn't fail on master

zixuanweeei:
@matteosal I tested your example with commit 9ca042, and it ran successfully without any exception. Could you test it again with commit 9ca042?

matteosal:
> @matteosal I tested your example with commit [9ca0428](https://github.com/apache/incubator-mxnet/commit/9ca042839b519c91963f77f694673557e9fddfc8), and it ran successfully without any exception. Could you test it again with commit [9ca0428](https://github.com/apache/incubator-mxnet/commit/9ca042839b519c91963f77f694673557e9fddfc8)?

Ops sorry, I've missed that commit. Yes, it works on that

zixuanweeei:
@matteosal Thanks. That's great. I am working on a unit test for this feature. Then we can merge it to master after further review and verification.

zixuanweeei:
@pengzhao-intel @TaoLv Please re-review on this PR. It should be noted that the new unit test function will be unevaluated in context of GPU because of the possible precision degradation resulted from the autotuned cudnn convolution. From a local test, the autotuned convolution has no more than 1.0% mismatches compared to a non-autotuned one, when any of the gradient request is set to be null (atol=1e-3, rtol=1e-3).

pengzhao-intel:
> he possible precision degradation resulted from the autotuned cudnn convolution.

How much degrade from GPU? Could we set a low bar for GPU?

zixuanweeei:
I will take some tests to see whether a low bar works.

pengzhao-intel:
@TaoLv  @ciyongch please take a reveiw too.

karan6181:
@mxnet-label-bot add [MKLDNN, pr-awaiting-review]

tests/python/unittest/test_operator.py
```diff
@@ -1907,6 +1907,88 @@ def test_depthwise_convolution():
...
+                    
+                exe2.forward(is_train=True)
+                np.testing.assert_allclose(exe1.outputs[0].asnumpy(),
+                    exe2.outputs[0].asnumpy(), rtol=rtol, atol=atol)
+                
+                exe2.backward(exe2.outputs[0])
+                for var_name in var_names:
+                    if var_name == "b" and no_bias:
+                        continue
+                    if grad_req2[var_name] == "null":
```
ciyongch: We don't have such case?

zixuanweeei: Yup. It is a very corner use of only requesting the gradient with respect to bias.

zixuanweeei:
@ciyongch The equality assertion error happened with the outputs of convolution forward when any of the gradients requests (`x`, `w`, `b`) was `null`.  Not sure about whether the Backward process drops the numerical precision.

pengzhao-intel:
Thanks for your contribution. Merging now :)

PR #15424, opened 2019-07-01
fixed config.mk and Makefile bugs for installing mkl

nuslq:
## Description ##
UPDATES:
Based on the discussion below with @TaoLv , I have removed the comments for USE_BLAS=mkl in this PR because after testing we find that we can statically link MKL BLAS library by adding USE_BLAS=mkl to the make command line. 

UPDATES:
I tried difference ways to build mxnet from source by hiding each or both of the two changes I made below, and found that only changing lower case "use_blas" to capital format in Makefile would be able to build mxnet with "BLAS_MKL". The other change, adding "USE_BLAS = mkl" before lines 137 - 142, would not make any impact for this purpose (although I am not sure whether or not this change would make impact at other places).

One thing to clarify here is that I used "make" instead of "cmake" to build mxnet from source.

(Brief description on what this PR is about)
I have found two bugs in the Makefile file and make/config.mk file in apache/incubator-mxnet package. These bugs might lead to incorrect installation of Mxnet with MKL library.

Before the changes, I could not find "BLAS_MKL" in mxnet.runtime.Features() outputs. After rebuilt my mxnet with the following changes, I found "BLAS_MKL" in mxnet.runtime.Features() outputs.

In the make/config.mk file, USE_BLAS will be first set to "atlas" for linux or "apple" for osx by default as follows (lines 119 - 124)

```
UNAME_S := $(shell uname -s)
ifeq ($(UNAME_S), Darwin)
USE_BLAS = apple
else
USE_BLAS = atlas
endif
```

Later, the USE_STATIC_MKL would be set "NONE" if users did not set USE_BLAS to "mkl" before the following lines (lines 137 - 142). ~~This would not be corrected even if users set "USE_BLAS=mkl" at the end of the config.mk file or in the command line.~~ (New test shows that USE_BLAS value, thereafter the USE_STATIC_MKL value, in config.mk can be overridden by adding USE_BLAS=mkl to the make command line)
```
# If use MKL only for BLAS, choose static link automatically to allow python wrapper
ifeq ($(USE_BLAS), mkl)
USE_STATIC_MKL = 1
else
USE_STATIC_MKL = NONE
endif
```

~~So, I correct this block to the follows,~~ (removed this change for reason described above)

```
# If use MKL only for BLAS, choose static link automatically to allow python wrapper
# Please Note: You have to set USE_BLAS = mkl here if you want to build mxnet with mkl. Otherwise USE_STATIC_MKL will be set to NONE. 
# USE_BLAS = mkl 
[...]

PR #15498, opened 2019-07-09
Add -DMXNET_USE_OPENMP to Makefiles so runtime info gets updated accordingly

larroy:
## Description ##

Fixes OPENMP info in runtime features.

https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/runtime.py

This would be also needed to fix BLAS flags.
https://github.com/apache/incubator-mxnet/pull/15424

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,6 +182,7 @@ endif
 
 ifeq ($(USE_OPENMP), 1)
 	CFLAGS += -fopenmp
+	CFLAGS += -DMXNET_USE_OPENMP=1
 endif
 
 ifeq ($(USE_NNPACK), 1)


karan6181:
@mxnet-label-bot add [Installation, Makefile, pr-awaiting-review]

larroy:
This is a bit of a departure of how we are testing since we are not testing depending on specific build flags (other than GPU and CPU which we could remove now thanks to libinfo). In that case we should also do all the permutation builds and check the build flags and also for Make and Cmake. This means adding quite a number of recompilations to test this. I'm not sure is worth the increased cost and trouble. Is it justified for you?

larroy:
@mxnet-label-bot add Bug

larroy:
@mxnet-label-bot add [Bug]

PR #15499, opened 2019-07-10
Improve diagnose.py, adding build features info and binary library path.

larroy:
See title.

This is the script used to fill issues, so adding more info will help.

Refined output:

[diagnose.txt](https://github.com/apache/incubator-mxnet/files/3375177/diagnose.txt)

### DIFF
--- a/tools/diagnose.py
+++ b/tools/diagnose.py
@@ -97,6 +97,11 @@ def check_pip():
     except ImportError:
         print('No corresponding pip install for current python.')
 
+def get_build_features_str():
+    import mxnet.runtime
+    features = mxnet.runtime.Features()
+    return '\n'.join(map(str, list(features.values())))
+
 def check_mxnet():
     print('----------MXNet Info-----------')
     try:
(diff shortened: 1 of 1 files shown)


wkcn:
Thank you for the improvement!

I have a suggestion : )
Could we only print the relative path?
Some users don't like the printing of the absolute path with their privacy.

tools/diagnose.py
```diff
@@ -105,13 +110,20 @@ def check_mxnet():
...
+        if os.path.exists(commit_hash):
+            with open(commit_hash, 'r') as f:
+                ch = f.read().strip()
+                print('Commit Hash   :', ch)
+        else:
+            print('Commit hash file "{}" not found. Not installed from pre-built package or built from source.'.format(commit_hash))
+        print('Library      :', mxnet.libinfo.find_lib_path())
+        try:
+            print('Build features:')
+            print(get_build_features_str())
```
apeforest: This will fail if user use this script to diagnose an mxnet script before 1.5.0 release.

apeforest:
One minor comment: the ✖ etc. symbols were not recognized when I import the txt file into MS office or Google Doc. Can we replace these unicode symbols by simple ascii, like Y/N?

larroy:
I think they are valid unicode symbols, what problem are you experiencing with MS Office? I think that is out of scope of this PR as it requires changes to the Python code that prints the feature, feel free to send a different PR for that.

larroy:
@apeforest I don't have any problems with Google docs or Office with those unicode characters.
Screen Shot 2019-07-13 at 7 56 35 PM
Screen Shot 2019-07-13 at 7 57 49 PM

PR #15454, opened 2019-07-04
Julia docs

aaronmarkham:
## Description ##
Since the dmlc.ml domain is gone we lost our Julia docs (and other things). 
This PR adds Julia docs to our regular website deployment pipeline. 💥 Links to the Julia docs will no longer go to a spam/malware website.

Fixes #15415 

**Update**: It also adds a thorough [Ubuntu setup guide for Julia](http://34.201.8.176/versions/julia/install/ubuntu_setup.html#install-the-mxnet-package-for-julia). The existing one doesn't work. However, this needs more eyes and testing. I did get it to work, but after a ton of troubleshooting, so I might have missed some env var or some setting that is required.

**Update 2**: I removed the separate Julia pipeline for docs. I needed to have its installation and config steps and it was duplicating things anyways. So since I had to fix CI, I went ahead and removed the duplication. That being said, sometime in the future it would be nice to revamp the docs build and have stages for each of the doc sets and make it so you don't have to rerun every stage of CI if one particular part of docs generation fails.

## Preview
The Julia docs site ends up in /site in the Julia API area.
http://34.201.8.176/versions/julia/api/julia/site/index.html

PR #15523, opened 2019-07-12
add julia env settings

aaronmarkham:
## Description ##
I missed this env var requirement in the Julia docs PR, #15454. 

## Comment
We should probably add some CI test for the website publish routine so this kind of thing doesn't get overlooked until after a merge.

### DIFF
--- a/docs/build_version_doc/build_all_version.sh
+++ b/docs/build_version_doc/build_all_version.sh
@@ -46,6 +46,17 @@ set -x
 # Set OPTS to any Sphinx build options, like -W for "warnings as errors"
 OPTS=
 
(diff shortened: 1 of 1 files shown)


aaronmarkham:
Nudging CI since so many tests failed for no apparent reason.

marcoabreu:
Well do we do have a website staging pipeline. The question here is why your main pipeline differs from the staging pipeline or if there's work that has to be done to migrate the staging one to match the main pipeline.

aaronmarkham:
@marcoabreu Can you point me to the staging pipeline so I can review the differences?

PR #15524, opened 2019-07-12
refine Nano setup directions

aaronmarkham:
## Description ##
After another run-through on a fresh Nano, plus trying out these new pip wheels, we found some refinements that should make the setup go a little more cleanly.

Thanks for the assist @darlaker

### DIFF
--- a/docs/install/install-jetson.md
+++ b/docs/install/install-jetson.md
@@ -19,13 +19,12 @@
 
 MXNet supports the Ubuntu Arch64 based operating system so you can run MXNet on NVIDIA Jetson Devices, such as the [TX2](http://www.nvidia.com/object/embedded-systems-dev-kits-modules.html) or [Nano](https://developer.nvidia.com/embedded/learn/get-started-jetson-nano-devkit).
 
-These instructions will walk through how to build MXNet and install MXNet's Python language binding. 
+These instructions will walk through how to build MXNet and install MXNet's Python language binding.
 
 For the purposes of this install guide we will assume that CUDA is already installed on your Jetson device. The disk image provided by NVIDIA's getting started guides will have the Jetson toolkit preinstalled, and this also includes CUDA. You should double check what versions are installed and which version you plan to use.
 
-You have several options for installing MXNet:
-1. Use a Jetson MXNet pip wheel for Python development.
-2. Use precompiled Jetson MXNet binaries.
+After installing the prerequisites, you have several options for installing MXNet:
+1. Use a Jetson MXNet pip wheel for Python development and use a precompiled Jetson MXNet binary.
 3. Build MXNet from source
    * On a faster Linux computer using cross-compilation
    * On the Jetson itself (very slow and not recommended)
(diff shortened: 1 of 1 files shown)


aaronmarkham:
Nudging CI again...

docs/install/install-jetson.md
```diff
@@ -60,49 +59,54 @@ sudo pip install \
 ```
 
 If you plan to cross-compile you will need to install these dependencies on that computer as well.
+If you get an error about something being busy, you can restart the nano and this error will go away and you can continue installation of the prerequisites.
 
-### Configure CUDA
+### Download the source & setup some environment variables:
 
-You can check to see what version of CUDA is running with `nvcc`.
+These steps are optional, but some of the following instructions expect MXNet source files and the `MXNET_HOME` environment variable. Also CUDA commands will not work out of the box without updating your path.
```
IvyBazan: "Also, CUDA commands will not work out of the box without updating your path."

docs/install/install-jetson.md
```diff
@@ -128,12 +130,13 @@ To use this with the MXNet Python binding, you must match the source directory's
 ```bash
 cd $MXNET_HOME
 git checkout v1.4.x
-git submodule update --init
+git submodule update --init --recursive
 cd python
 sudo pip install -e .
 ```
```
darlaker: I think we considered adding a note here at the end of Option 1 that notifies the user that at this point, they are done and do not need to continue. I think that still sounds helpful, but that may already be clear with "Option" now in the title.

docs/install/install-jetson.md
```diff
@@ -60,49 +59,54 @@ sudo pip install \
...
 
-Setup your environment variables for MXNet.
+To switch CUDA versions on a device or computer that has more than one version installed, use the following and replace the version as appropriate.
 
 ```bash
-cd ..
-export MXNET_HOME=$(pwd)
-echo "export PYTHONPATH=$MXNET_HOME/python:$PYTHONPATH" >> ~/.rc
-source ~/.rc
+sudo rm /usr/local/cuda
```
darlaker: Maybe add a line here that describes how these commands will remove the directory and then replace it by creating a symlink. But Linux users may not need this called out. :)

docs/install/install-jetson.md
```diff
@@ -60,49 +59,54 @@ sudo pip install \
...
+You can check to see what version of CUDA is running with `nvcc`.
 
 ```bash
-git clone --recursive https://github.com/apache/incubator-mxnet.git mxnet
-cd mxnet
+nvcc --version
 ```
 
-Setup your environment variables for MXNet.
+To switch CUDA versions on a device or computer that has more than one version installed, use the following and replace the version as appropriate.
```
darlaker: Maybe add: "This example installs version 10 and replaces all other versions."

docs/install/install-jetson.md
aaronmarkham: It was always a symlink, not an actual folder. I just briefly freaked out when I thought I deleted the cuda files.

PR #15539, opened 2019-07-15
fix normalize mean error bug

nicklhy:
The normalize operator may produce wrong result when setting the mean arg with a single float val.

### DIFF
--- a/src/operator/image/image_random-inl.h
+++ b/src/operator/image/image_random-inl.h
@@ -339,7 +339,7 @@ void NormalizeOpForward(const nnvm::NodeAttrs &attrs,
   std::vector<float> mean(3);
   std::vector<float> std(3);
   if (param.mean.ndim() == 1) {
-    mean[0] = mean[1] = mean[3] = param.mean[0];
+    mean[0] = mean[1] = mean[2] = param.mean[0];
   } else {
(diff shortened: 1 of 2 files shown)


karan6181:
@mxnet-label-bot add [Operator, pr-awaiting-response]

nicklhy:
> Wow! How did we miss this?
> Good find! Any test you wrote to verify this?

We are upgrading MXNet in our deployment env these days and notice some accuracy drop of our models. This bug seems to exist for at least a few months. Maybe we should add more tests to avoid similar bugs in the future.

wkcn:
Hi, @ChaiBapchya . 

I think it is suitable to merge this PR : ) What do you think?

wkcn:
Hi @nicklhy , could you please add an unittest in [`test_image_normalize`](https://github.com/apache/incubator-mxnet/blob/master/tests/python/unittest/test_operator.py#L8552) ?

Thank you!

nicklhy:
@wkcn Just added some tests for scalar mean/std support in [b952e91](https://github.com/apache/incubator-mxnet/pull/15539/commits/b952e91931140ca4fe7dbbe7460235d3b4bc9276).

wkcn:
@nicklhy Great work! Thank you so much!

PR #15541, opened 2019-07-15
[Opperf] Filter out deprecated ops

ChaiBapchya:
## Description ##
Filters out deprecated operators from `_get_all_mxnet_operators()` function. 
To ensure they are not included while calculating MXNet operators

~~Current random_sampling_operators supports 16. Added 2 ops (normal, uniform)(although deprecated, they are currently supported. Hence)~~

### Essentials ###

### Changes ###
- Removed uniform,normal op

### Test/Verify ###

```
$ export PYTHONPATH=$PYTHONPATH:/path/to/incubator-mxnet/
$ python
>>> from benchmark.opperf.utils.op_registry_utils import _get_all_mxnet_operators
>>> for key in _get_all_mxnet_operators():
>>>     print(key)
>>>
```

~~### How to test/verify ###~~
~~Build~~
~~```~~
~~incubator-mxnet$ make -j8~~
~~incubator-mxnet$ pip install -e python/.~~
~~```~~
~~Test~~
~~```~~
~~>>> import mxnet as mx~~
~~>>> from mxnet import nd ~~
~~>>> from benchmark.opperf.nd_operations.random_sampling_operators import run_mx_random_sampling_operators_benchmarks~~
~~>>> print(run_mx_random_sampling_operators_benchmarks())~~
~~```~~


ChaiBapchya:
@mxnet-label-bot add [Benchmark, Operator]

benchmark/opperf/utils/op_registry_utils.py
```diff
@@ -212,10 +212,13 @@ def get_all_random_sampling_operators():
     # Get all mxnet operators
     mx_operators = _get_all_mxnet_operators()
 
+    # Deprecated random sampling operators
```
sandeep-krishnamurthy: Let us not add deprecated operators? 
May be you can update _get_all_mxnet_operators to not return deprecated ops. What do you think?

ChaiBapchya: Cool. I'll do that. I thought deprecated ops are still supported (until 2.0) Hence thought of including them. But ya makes sense.

benchmark/opperf/utils/op_registry_utils.py
```diff
@@ -55,6 +57,11 @@ def _select_ops(operator_names, filters=("_contrib", "_"), merge_op_forward_back
     mx_operators = {}
     operators_with_backward = []
 
+    # Filter out deprecated operators
```
sandeep-krishnamurthy: Is there a "Deprecated" field in the operator metadata?
Because, listing operator by names is not scalable. Everytime we deprecate a operator, we need to come here to add it which soon because un-maintained.

ChaiBapchya: Did some digging around to find there's no exposed API from backend to retrieve deprecated ops. So at the moment, for the sake of this PR can we go ahead with that. I've created another issue to track the same. Sounds good?

ChaiBapchya: @sandeep-krishnamurthy

piyushghai:
@ChaiBapchya @sandeep-krishnamurthy What's the path forward for this PR ?

ChaiBapchya:
I'm guessing the right way for this PR would be
Create backend C API for getting deprecated ops (instead of my current method of hardcoding it). Just that I am bit stuck on other opperf that I have yet to write the backend part.

PR #15545, opened 2019-07-15
Softmax optimization for GPU

ptrendx:
## Description ##
This PR optimizes Softmax implementation for cases where stride is 1 and the leading dimension is small (up to 20kB of data in that dimension).

There are 2 optimizations in this kernel compared to the previous one:
 - using of the longer datatypes for loading/writing data (so using up to 8B per read/write instead of e.g. 2B in case of fp16 I/O)
 - using persistent storage to reduce the number of memory accesses (previous implementation used 3 reads and 1 write, this implementation uses 1 read and 1 write).

Compared to the previous implementation on fp16 I/O the new kernel is up to 4x faster.

@eric-haibin-lin 

### Essentials ###


karan6181:

src/operator/nn/softmax-inl.h
```diff
@@ -218,6 +219,157 @@ __global__ void softmax_compute_kernel(DType *in, OType *out, index_t M, int axi
...
+  v = redfun(v, __shfl_down_sync(0xffffffff, v, 8));
+  v = redfun(v, __shfl_down_sync(0xffffffff, v, 4));
+  v = redfun(v, __shfl_down_sync(0xffffffff, v, 2));
+  v = redfun(v, __shfl_down_sync(0xffffffff, v, 1));
+  return mshadow::half::half_t(v);
+}
+
+template<typename OP, bool negate, typename AType, typename LType,
+  typename DType, typename OType>
+__global__ void softmax_compute_kernel2(const DType *in, OType *out, const index_t M,
```
eric-haibin-lin: need better name

src/operator/nn/softmax-inl.h
```diff
@@ -218,6 +219,157 @@ __global__ void softmax_compute_kernel(DType *in, OType *out, index_t M, int axi
   }
 }
 
+const int softmax_threads_per_block = 512;
+
+template <typename OP, typename T>
+__device__ inline T warp_reduce(T value, OP redfun) {
```
eric-haibin-lin: this looks like a generic function that can be used elsewhere. Is there a better place to put this function?

src/operator/nn/softmax-inl.h
haojin2: +1 to haibin's comment, maybe something like `softmax_stride_1_compute_kernel` ?

src/operator/nn/softmax-inl.h
```diff
@@ -218,6 +219,157 @@ __global__ void softmax_compute_kernel(DType *in, OType *out, index_t M, int axi
...
+  const int entries_per_load = sizeof(LType)/sizeof(DType);
+  // Due to usage of MSHADOW_TYPE_SWITCH macro we are generating
+  // kernels where sizeof(LType) may be less than sizeof(DType),
+  // resulting in entries_per_load being 0.
+  // This is not a valid combination and is being checked against
+  // in the launcher code. This switch here is just to silence
+  // the division by zero warning generated for such invalid cases.
+  const int row_length = entries_per_load > 0 ? M / entries_per_load : 0;
+
+  const LType * in_aligned = reinterpret_cast<const LType *>(in);
```
haojin2: nit: `LType*` instead of `LType *` as per [Google Style Guide](https://google.github.io/styleguide/cppguide.html#Pointer_and_Reference_Expressions). Same for all other pointer declarations.

haojin2:
@ptrendx Do you have any data on how much performance boost this change is introducing on applicable example workloads?

src/operator/nn/softmax-inl.h
```diff
@@ -218,6 +219,157 @@ __global__ void softmax_compute_kernel(DType *in, OType *out, index_t M, int axi
...
+  return mshadow::half::half_t(v);
+}
+
+template<typename OP, bool negate, typename AType, typename LType,
+  typename DType, typename OType>
+__global__ void softmax_compute_kernel2(const DType *in, OType *out, const index_t M,
+                                       const double temperature, int rows_per_block,
+                                       const index_t total_rows) {
+  __shared__ AType scratch[softmax_threads_per_block];
+  __shared__ LType persistent_storage[20*1024 / sizeof(LType)];
```
apeforest: nit: space between '*' and operands

src/operator/nn/softmax-inl.h
```diff
@@ -218,6 +219,157 @@ __global__ void softmax_compute_kernel(DType *in, OType *out, index_t M, int axi
...
+  LType * out_aligned = reinterpret_cast<LType *>(out);
+
+  for (index_t i = my_id; i < row_length; i += threads_per_row) {
+    out_aligned[base + i] = persistent_storage[my_local_row * row_length + i];
+  }
+}
+
+namespace {
+
+int get_load_type(size_t N) {
```
apeforest: This function also seems generic and maybe we can move it to a common file so all other operators can use?

src/operator/nn/softmax-inl.h
```diff
@@ -218,6 +219,157 @@ __global__ void softmax_compute_kernel(DType *in, OType *out, index_t M, int axi
...
+  } else if (N % 4 == 0) {
+    return kFloat32;
+  } else if (N % 2 == 0) {
+    return kFloat16;
+  } else {
+    return kInt8;
+  }
+}
+
+int get_rows_per_block(size_t N) {
```
apeforest: This function also seems generic and maybe we can move it to a common file so all other operators can use?

src/operator/nn/softmax-inl.h
```diff
@@ -229,10 +381,28 @@ inline void Softmax(Stream<gpu> *s, DType *in, OType *out,
   Shape<ndim> sshape = shape;
   sshape[axis] = 1;
 
-  softmax_compute_kernel<x_bits, OP, negate, AType, ndim>
-    <<<N, x_size, 0, mshadow::Stream<gpu>::GetStream(s)>>>(
-      in, out, M, axis, sshape, stride, temperature);
-  MSHADOW_CUDA_POST_KERNEL_CHECK(softmax_compute_kernel);
+  const int DSize = sizeof(DType);
+  // Using 20 kB of shared memory for persistent storage in the optimized case
+  const int max_opt_M = 20*1024/DSize;
```
apeforest: nit: add space around '*'

src/operator/nn/softmax-inl.h
```diff
@@ -229,10 +381,28 @@ inline void Softmax(Stream<gpu> *s, DType *in, OType *out,
   Shape<ndim> sshape = shape;
   sshape[axis] = 1;
 
-  softmax_compute_kernel<x_bits, OP, negate, AType, ndim>
-    <<<N, x_size, 0, mshadow::Stream<gpu>::GetStream(s)>>>(
-      in, out, M, axis, sshape, stride, temperature);
-  MSHADOW_CUDA_POST_KERNEL_CHECK(softmax_compute_kernel);
+  const int DSize = sizeof(DType);
```
apeforest: use `size_t` for DSize?

src/operator/nn/softmax-inl.h
apeforest: use `size_t` for max_opt_M

sxjscience:
Great! I'm considering to use this kind of warp-level primitives to accelerate our current reduce.

src/operator/nn/softmax-inl.h
sxjscience: Do we need to check the alignment? Since CUDA uses force-alignment, it will potentially raise an error if the address of `in` is not aligned with `LType`.

ptrendx: @sxjscience That is why the code that launches this kernel chooses LType based on the array dimensions - if the leading dimension is odd it will not choose LType larger than DType.

sxjscience: I'm not sure about the answer to this question but my concern is that we need to make sure that the following is true:
```
ASSERT(static_cast<size_t>(in) % sizeof(LType) == 0)
```
mainly due to the force-alignment constraint in CUDA (https://stackoverflow.com/questions/37323053/misaligned-address-in-cuda).

ptrendx: A pointer given by `cudaMalloc` is guaranteed to be aligned to something like 256B or more.

sxjscience: Thanks for correcting me. Another question is: "Would it be possible to handle the cases when the `N` in `get_load_type(size_t N)` can not be divided by 8?" Could we load the majority of the elements using the vectorizing trick and just handle the remainders? (This is just a question and there is no need to revise it in this PR because I think it looks great).

src/operator/nn/softmax-inl.h
ptrendx: Sure

src/operator/nn/softmax-inl.h
ptrendx: Yup, will look into putting it in some better place.

src/operator/nn/softmax-inl.h
ptrendx: :+1:

src/operator/nn/softmax-inl.h
ptrendx: Yup.

src/operator/nn/softmax-inl.h
ptrendx: This is debatable - it is much more coupled to this softmax implementation than the other ones so I would probably leave it where it is for now.

src/operator/nn/softmax-inl.h
ptrendx: :+1:

src/operator/nn/softmax-inl.h
ptrendx: sure

src/operator/nn/softmax-inl.h
ptrendx: :+1:

ptrendx:
@haojin2 For the perf improvement on end to end training - I did not measure yet, measured just the kernel speedup for now. BTW - how important do you think safe accumulation option is (as in, having an option to NOT do it instead of just always doing safe accumulation)? Personally I don't see value in not using it in softmax as it would most probably affect accuracy, and having 3 TYPE_SWITCH makes the compilation time quite big. You did not put the ability to skip it in `softmax_with_length`, should I remove it from regular softmax as well?

src/operator/nn/softmax-inl.h
sxjscience: It looks like common cuda utilities are put here https://github.com/apache/incubator-mxnet/blob/master/src/common/cuda_utils.h

sxjscience: In fact, there are other common functions that spread in other files, e.g., the warp-level wrapper in https://github.com/apache/incubator-mxnet/blob/master/src/operator/nn/layer_norm.cu#L32-L50

sxjscience: We can also put it into https://github.com/apache/incubator-mxnet/blob/master/src/operator/tensor/broadcast_reduce-inl.cuh

ptrendx:
@haojin2 I'm not sure how to progress with this PR - it seems that Windows CI instances do not have enough RAM to process all the templates here. The problem exists even with CPU compilation, where I just fixed your omission of MXNET_SAFE_ACCUMULATION=0 case in softmax with length.

KellenSunderland:
@sxjscience totally agree.  This would provide a lot of benefit across the framework (for example the layernorm op).

@ptrendx
I see what you mean "fatal error C1002: compiler is out of heap space in pass 2".  The CI windows machines should have a fair amount of RAM so this is a little strange.

ptrendx:
@marcoabreu Could you give some advice on that Windows CI problem - are the Windows builder instances much different than the unix ones?

marcoabreu:
Yeah they are. but the heap space error is the same old problem that we just have too many macros I think. Basically we can't add any more operators because the file just grew too large.

Try compiling locally on Unix with optimizations disabled and enable intermediary output. You will see some intermediary files being multiple gigabyte in size.

marcoabreu:
@haojin2 had the same problem btw

marcoabreu:
But the windows instances are c5.18xlarge. they have plenty of ram. We're literally running into the limitations of the compilers

ptrendx:
Would it help if we split the fwd and backward into different files? Those limitations are per file, right?

marcoabreu:
I'd rather split operators into different files. Forward and backwards path kinda belong together, right?

This could also lay the base for dynamic loading of operators where we could ship operators selectively.

KellenSunderland:
Thanks for the info Marco, appreciate you helping with this.

I remember the MSVC team was very slow in moving their compiler process to 64 bit.  Just to be clear they supported compiling 64 bit programs very early on, but the compiler process itself didn't need much ram so it stayed 32-bit.  Apparently now they do have a 64 bit compiler process, but 32-bit is still the default.  Is there any way we could check any of the windows hosts, and see if the compiler process they're using (CL.exe) is running in 32-bit mode in the task manager?  Are any of the compilation processes actually using more than 4GB of ram?

marcoabreu:
I remember that I checked it on Unix and that I saw high ram usage there. I'm currently not able to get data from windows, but the operator compilation has always been a bottleneck.

marcoabreu:
There really is a pattern that these kinds of errors always come when people try to add new operators

src/operator/nn/softmax-inl.h
access2rohit: @ptrendx this function can be made generic by passing `threads_per_block` and `instructions_executed_per_thread` as params. That way we can reuse this functionality. Every time if  we keep adding new functions specific to an operator will keep increasing binary size and no one will revisit once it is left for now.

ptrendx: @access2rohit Yeah, I can see that, thanks for the suggestion. I will put it in the common place as I refactor the softmax files to hopefully get rid of the Windows CI problems and introduce similar optimizations for the backward pass.

src/common/cuda_utils.h
```diff
@@ -615,6 +624,28 @@ __device__ inline DType ldg(const DType* address) {
     return *address;
 #endif
 }
-#endif
+
```
access2rohit: nice ! 
quick question ...... warp primitives don't support float16 yet ?

larroy:
https://developercommunity.visualstudio.com/content/problem/284930/error-c1002-compiler-is-out-of-heap-space-in-pass.html We should update to VS 2017

larroy:
(or Newer)

src/common/cuda_utils.h
ptrendx: No, they don't.

ptrendx:
Ok, it seems that splitting softmax.cc into 3 files, 1 for each operator (softmax, softmin and log_softmax) did the trick fortunately.

src/operator/nn/softmax-inl.h
eric-haibin-lin: @ptrendx could you address this?

ptrendx: Ooops, I forgot about it, will do that tomorrow.

src/operator/nn/softmax-inl.h
```diff
@@ -188,89 +180,77 @@ struct log_softmax_bwd {
...
-    index_t base = unravel_dot(i, sshape, stride);
-    index_t len = static_cast<index_t>(length[i]);
-
-    AType sum = AType(0);
-    for (index_t j = 0; j < len; ++j) {
-      sum += OP1::Map(ograd[base + j*sa], out[base + j*sa]);
+      // By default temperature is 1.0.
+      // Adding a branch here to save the CPU 'divide-by-1' computation at runtime
+      DType final_result;
+      if (temperature == 1.0) {
```
larroy: is this micro-opt really making things better?

src/operator/nn/softmax-inl.h
```diff
@@ -301,7 +282,7 @@ __global__ void softmax_compute_kernel(DType *in, OType *out, index_t M, int axi
 
   red::sum::SetInitValue(smem[x]);
```
larroy: Does have multiple max values affect numerical accuracy? Or are they reduced at some other point to a final max?

src/operator/nn/softmax-inl.h
```diff
@@ -313,71 +294,134 @@ __global__ void softmax_compute_kernel(DType *in, OType *out, index_t M, int axi
...
+  }
+  DType * row = reinterpret_cast<DType *>(persistent_storage + my_local_row * row_length);
+  __syncthreads();
 
-  softmax_compute_kernel<x_bits, OP, negate, AType, ndim>
-    <<<N, x_size, 0, mshadow::Stream<gpu>::GetStream(s)>>>(
-      in, out, M, axis, sshape, stride, temperature);
-  MSHADOW_CUDA_POST_KERNEL_CHECK(softmax_compute_kernel);
-}
+  DType my_max_value;
```
larroy: Can we add a comment or maybe a more descriptive name? is this the max of the stride?

src/operator/nn/softmax-inl.h
```diff
@@ -313,71 +294,134 @@ __global__ void softmax_compute_kernel(DType *in, OType *out, index_t M, int axi
...
+int get_rows_per_block(size_t N) {
+  const int warp_size = 32;
+  // How many read instructions should 1 thread at least do
+  const int read_instructions = 2;
+  const int num_threads = (N + read_instructions - 1) / read_instructions;
+  int num_warps = (num_threads + warp_size - 1) / warp_size;
+  // num_warps needs to be power of 2
+  int used_num_warps = 1;
+  num_warps = std::min(num_warps, softmax_threads_per_block / warp_size);
+  int tmp = num_warps;
```
larroy: Could we factor this out in a separate function with a nice name?

src/operator/nn/softmax-inl.h
apeforest: Yes, I have done performance comparison earlier. This check speed up the operator by 30%

apeforest:
Thanks for refactoring the Softmax functions to make it into one.

src/operator/nn/softmax-inl.h
ptrendx: I did not touch the CPU code besides merging it into a 1 function from 2 to not introduce slowdowns in that path.

src/operator/nn/softmax-inl.h
ptrendx: Not sure I understand.
softmax

That is not very numerically stable, so there is another step introduced that finds the maximum x_i first and compute
softmax
It does not matter which maximum you take.

src/operator/nn/softmax-inl.h
ptrendx: What would you suggest? This is the maximum value that this thread sees.

src/operator/nn/softmax-inl.h
ptrendx: will do together with moving this function to the common place.

src/operator/nn/softmax-inl.h
ptrendx: Done.

src/operator/nn/softmax-inl.h
ptrendx: Done.

src/operator/nn/softmax-inl.h
larroy: let me rephrase. When running Reduce1D, are all the max values reduce into smem[0] / smax? as I understand the xmxax should be the max of {x_i}. It actually matters for numerical issues.

ptrendx: Yes, those intermediate max values are then reduced across threads to the final maximum.

Issue #15465, opened 2019-07-05
[RFC] Integrate TVM into Apache MXNet


yzhliu:
# Problem Statement

Currently in MXNet we implement operator kernels in C++. Developers need to specify the detail logic of each computation, which slows down the development process. Given the fact that we’re moving forward to be numpy-compatible[1], a large amount of operators are to be implemented. Moreover, we also have various of backend to support, including CPUs and GPUs from AMD, ARM, Intel, Nvidia, etc. It requires great effort to implement efficient kernels for each of these hardwares, so as writing test cases for each of the operator+backend combination.

# Proposal

Thus I would like to propose to integrate Apache TVM into Apache MXNet, so that we can leverage its ability to easily implement (in Python) high-performance operator kernels. Here’re some of the advantages,

1. We devised a new approach to implement MXNet kernels in Python, (see PR https://github.com/apache/incubator-mxnet/pull/15345 ). For example, to implement broadcast add, people can write pure Python,

```python
@defop(name="vadd", target="cpu", auto_broadcast=True,
       dtype=AllTypes, ndim=list(range(6)))
def vadd(dtype, ndim):
    A = tvm.placeholder(shape=[tvm.var() for _ in range(ndim)], name='A', dtype=dtype)
    B = tvm.placeholder(shape=[tvm.var() for _ in range(ndim)], name='B', dtype=dtype)
    C = tvm.compute(shape=[tvm.var() for _ in range(ndim)],
                    lambda *index: A[index] + B[index], name='C')
    s = tvm.create_schedule(C.op)
    axes = [axis for axis in C.op.axis]
    fused = s[C].fuse(*axes)
    s[C].parallel(fused)
    return s, [A, B, C]
```

The code above will be compiled to binary and linked into MXNet as a regular function. Note that the same piece of compute definition can be shared across the multiple backends (cpu, gpu, etc.). As it is much more concise than that defined in C++, it can lower the bar of implementing high-performance kernels and improve the development experience. We expect people can develop more efficiently with such approach.

2. Operators in current TVM are already numpy-compatible, we can leverage the efforts there to help our Numpy project.

# Approach

* Build and link TVM dynamic library into MXNet.
* Build infrastructure to write operator kernels in Python, including:
    * The approach for registering TVM operator into MXNet
    * Modify CI to enable TVM operator build, and pack the operator library together with the release binary. We can enable automatic performance tuning[2] later to further improve the performance.

[...]

PR #15550, opened 2019-07-16
Infra to use tvm write op kernels

yzhliu:
## Description ##
This PR implements an infra to let users write op kernels in pure Python. One broadcast operator example can be found in,
* Kernel definition: contrib/tvmop/basic/ufunc.py
* Operator registry: src/operator/contrib/tvmop/ufunc.cc
* Test case: tests/python/unittest/test_tvm_op.py

More background and discussion can be found here: https://github.com/apache/incubator-mxnet/issues/15465

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -734,6 +734,28 @@ if(USE_DIST_KVSTORE)
   list(APPEND mxnet_LINKER_LIBS ${pslite_LINKER_LIBS})
 endif()
 
+if(USE_TVM_OP)
+  add_definitions(-DMXNET_USE_TVM_OP=1)
+  list(APPEND mxnet_LINKER_LIBS ${CMAKE_CURRENT_BINARY_DIR}/3rdparty/tvm/libtvm_runtime.so)
+  include(cmake/BuildTVM.cmake)
+  add_subdirectory("3rdparty/tvm")
+
+  if(NOT Python3_EXECUTABLE)
+    find_package(PythonInterp 3 REQUIRED)
+    set(Python3_EXECUTABLE ${PYTHON_EXECUTABLE} CACHE FILEPATH "Path to the python3 executable")
+    if(NOT Python3_EXECUTABLE)
+      message(FATAL_ERROR "No python3 interpreter found to build TVM operators")
+    endif()
+  endif()
+
+  add_custom_command(TARGET mxnet POST_BUILD
+    COMMAND ${CMAKE_COMMAND} -E env
+          PYTHONPATH="${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/tvm/python:${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/tvm/topi/python:${CMAKE_CURRENT_SOURCE_DIR}/contrib"
+          LD_LIBRARY_PATH="${CMAKE_CURRENT_BINARY_DIR}/3rdparty/tvm/build"
+          ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/contrib/tvmop/compile.py -o${CMAKE_CURRENT_BINARY_DIR}/libtvmop.so
+    )
+endif()
+
 target_link_libraries(mxnet PUBLIC ${mxnet_LINKER_LIBS})
 
 if(USE_PLUGINS_WARPCTC)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,7 @@ mxnet_option(USE_MXNET_LIB_NAMING "Use MXNet library naming conventions." ON)
 mxnet_option(USE_GPROF            "Compile with gprof (profiling) flag" OFF)
 mxnet_option(USE_CXX14_IF_AVAILABLE "Build with C++14 if the compiler supports it" OFF)
 mxnet_option(USE_VTUNE            "Enable use of Intel Amplifier XE (VTune)" OFF) # one could set VTUNE_ROOT for search path
+mxnet_option(USE_TVM_OP           "Enable use of TVM operator build system." OFF)
 mxnet_option(ENABLE_CUDA_RTC      "Build with CUDA runtime compilation support" ON)
 mxnet_option(BUILD_CPP_EXAMPLES   "Build cpp examples" ON)
 mxnet_option(INSTALL_EXAMPLES     "Install the example source files." OFF)
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,14 @@ ifndef AMALGAMATION_PATH
 	AMALGAMATION_PATH = $(ROOTDIR)/amalgamation
 endif
 
+ifndef TVM_PATH
+	TVM_PATH = $(TPARTYDIR)/tvm
+endif
+
+ifndef LLVM_PATH
+	LLVM_PATH = $(TVM_PATH)/build/llvm
+endif
+
 ifneq ($(USE_OPENMP), 1)
 	export NO_OPENMP = 1
 endif
@@ -100,6 +108,12 @@ else
 endif
 CFLAGS += -I$(TPARTYDIR)/mshadow/ -I$(TPARTYDIR)/dmlc-core/include -fPIC -I$(NNVM_PATH)/include -I$(DLPACK_PATH)/include -I$(TPARTYDIR)/tvm/include -Iinclude $(MSHADOW_CFLAGS)
 LDFLAGS = -pthread $(MSHADOW_LDFLAGS) $(DMLC_LDFLAGS)
+
+ifeq ($(USE_TVM_OP), 1)
+	LIB_DEP += lib/libtvm_runtime.so lib/libtvmop.so
+	CFLAGS += -I$(TVM_PATH)/include -DMXNET_USE_TVM_OP=1
+	LDFLAGS += -L$(TVM_PATH)/build -ltvm_runtime
+endif
 
 ifeq ($(ENABLE_TESTCOVERAGE), 1)
         CFLAGS += --coverage
@@ -589,6 +603,30 @@ $(DMLC_CORE)/libdmlc.a: DMLCCORE
 DMLCCORE:
 	+ cd $(DMLC_CORE); $(MAKE) libdmlc.a USE_SSE=$(USE_SSE) config=$(ROOTDIR)/$(config); cd $(ROOTDIR)
 
+TVM_USE_CUDA := OFF
+ifeq ($(USE_CUDA), 1)
+	TVM_USE_CUDA := ON
+	ifneq ($(USE_CUDA_PATH), NONE)
+		TVM_USE_CUDA := $(USE_CUDA_PATH)
+	endif
+endif
+
+lib/libtvm_runtime.so:
+	echo "Compile TVM"
+	[ -e $(LLVM_PATH)/bin/llvm-config ] || sh $(ROOTDIR)/contrib/tvmop/prepare_tvm.sh; \
+	cd $(TVM_PATH)/build; \
+	cmake -DUSE_LLVM="$(LLVM_PATH)/bin/llvm-config" \
+		  -DUSE_SORT=OFF -DUSE_CUDA=$(TVM_USE_CUDA) -DUSE_CUDNN=OFF ..; \
+	$(MAKE) VERBOSE=1; \
+	cp $(TVM_PATH)/build/libtvm_runtime.so $(ROOTDIR)/lib/libtvm_runtime.so; \
+	cd $(ROOTDIR)
+
+lib/libtvmop.so: lib/libtvm_runtime.so $(wildcard contrib/tvmop/*/*.py contrib/tvmop/*.py)
+	echo "Compile TVM operators"
+	PYTHONPATH=$(TVM_PATH)/python:$(TVM_PATH)/topi/python:$(ROOTDIR)/contrib:$PYTHONPATH \
+		LD_LIBRARY_PATH=lib \
+	    python3 $(ROOTDIR)/contrib/tvmop/compile.py -o $(ROOTDIR)/lib/libtvmop.so
+
 NNVM_INC = $(wildcard $(NNVM_PATH)/include/*/*.h)
 NNVM_SRC = $(wildcard $(NNVM_PATH)/src/*/*/*.cc $(NNVM_PATH)/src/*/*.cc $(NNVM_PATH)/src/*.cc)
 $(NNVM_PATH)/lib/libnnvm.a: $(NNVM_INC) $(NNVM_SRC)
@@ -726,6 +764,7 @@ clean: rclean cyclean $(EXTRA_PACKAGES_CLEAN)
 	cd $(DMLC_CORE); $(MAKE) clean; cd -
 	cd $(PS_PATH); $(MAKE) clean; cd -
 	cd $(NNVM_PATH); $(MAKE) clean; cd -
+	cd $(TVM_PATH); $(MAKE) clean; cd -
 	cd $(AMALGAMATION_PATH); $(MAKE) clean; cd -
 	$(RM) -r  $(patsubst %, %/*.d, $(EXTRA_OPERATORS)) $(patsubst %, %/*/*.d, $(EXTRA_OPERATORS))
 	$(RM) -r  $(patsubst %, %/*.o, $(EXTRA_OPERATORS)) $(patsubst %, %/*/*.o, $(EXTRA_OPERATORS))
--- a/Makefile
+++ b/Makefile
@@ -109,12 +109,6 @@ endif
 CFLAGS += -I$(TPARTYDIR)/mshadow/ -I$(TPARTYDIR)/dmlc-core/include -fPIC -I$(NNVM_PATH)/include -I$(DLPACK_PATH)/include -I$(TPARTYDIR)/tvm/include -Iinclude $(MSHADOW_CFLAGS)
 LDFLAGS = -pthread $(MSHADOW_LDFLAGS) $(DMLC_LDFLAGS)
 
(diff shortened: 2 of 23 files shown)


src/operator/contrib/tvmop/ufunc.cc
```diff
@@ -0,0 +1,64 @@
...
+namespace op {
+
+static constexpr char func_vadd_cpu[] = "vadd";
+static constexpr char func_vadd_gpu[] = "cuda_vadd";
+
+template<const char* func>
+void TVMBroadcastCompute(const nnvm::NodeAttrs& attrs,
+                         const mxnet::OpContext& ctx,
+                         const std::vector<TBlob>& inputs,
+                         const std::vector<OpReqType>& req,
```
ZhennanQin: Why req is ignored?

src/operator/tvmop/op_module.cc
```diff
@@ -0,0 +1,117 @@
...
+#include <string>
+#include <vector>
+#include "op_module.h"
+
+using namespace tvm::runtime;
+
+namespace tvm {
+namespace runtime {
+
+void TVMOpModule::Load(const std::string &filepath) {
```
ZhennanQin: Maybe need to load more information like version, backend_support, feature(avx512 or not) to help analysis.

Makefile
```diff
@@ -589,6 +603,30 @@ $(DMLC_CORE)/libdmlc.a: DMLCCORE
...
+	echo "Compile TVM"
+	[ -e $(LLVM_PATH)/bin/llvm-config ] || sh $(ROOTDIR)/contrib/tvmop/prepare_tvm.sh; \
+	cd $(TVM_PATH)/build; \
+	cmake -DUSE_LLVM="$(LLVM_PATH)/bin/llvm-config" \
+		  -DUSE_SORT=OFF -DUSE_CUDA=$(TVM_USE_CUDA) -DUSE_CUDNN=OFF ..; \
+	$(MAKE) VERBOSE=1; \
+	cp $(TVM_PATH)/build/libtvm_runtime.so $(ROOTDIR)/lib/libtvm_runtime.so; \
+	cd $(ROOTDIR)
+
+lib/libtvmop.so: lib/libtvm_runtime.so $(wildcard contrib/tvmop/*/*.py contrib/tvmop/*.py)
```
ZhennanQin: Windows support?

src/operator/contrib/tvmop/ufunc.cc
```diff
@@ -0,0 +1,64 @@
...
+
+NNVM_REGISTER_OP(_contrib_tvm_vadd)
+    .set_num_inputs(2)
+    .set_num_outputs(1)
+    .add_argument("a", "NDArray-or-Symbol", "first input")
+    .add_argument("b", "NDArray-or-Symbol", "second input")
+    .set_attr<mxnet::FInferShape>("FInferShape", BinaryBroadcastShape)
+    .set_attr<nnvm::FInferType>("FInferType", mxnet::op::ElemwiseType<2, 1>)
+    .set_attr<mxnet::FCompute>("FCompute<cpu>", mxnet::op::TVMBroadcastCompute<func_vadd_cpu>)
+    .set_attr<mxnet::FCompute>("FCompute<gpu>", mxnet::op::TVMBroadcastCompute<func_vadd_gpu>);
```
ZhennanQin: If host doesn't have GPU,  will this cause any trouble?

python/mxnet/base.py
```diff
@@ -734,3 +734,8 @@ def write_all_str(module_file, module_all_list):
 
 ctypes.pythonapi.PyCapsule_New.restype = ctypes.py_object
 ctypes.pythonapi.PyCapsule_GetPointer.restype = ctypes.c_void_p
+
+from .runtime import Features
+if Features().is_enabled("TVM_OP"):
```
ZhennanQin: Seems TVM is supported by python only because of this code. Shall we load libtvmop automatically along with loading libmxnet? Then we don't do extra change with front-end language.

Makefile
```diff
@@ -589,6 +603,30 @@ $(DMLC_CORE)/libdmlc.a: DMLCCORE
 DMLCCORE:
 	+ cd $(DMLC_CORE); $(MAKE) libdmlc.a USE_SSE=$(USE_SSE) config=$(ROOTDIR)/$(config); cd $(ROOTDIR)
 
+TVM_USE_CUDA := OFF
```
marcoabreu: Is it really necessary to introduce another variable? Couldn't we just keep USE_CUDA and tie onto that instead?

src/operator/contrib/tvmop/ufunc.cc
yzhliu: Just an example to show how it works

src/operator/tvmop/op_module.cc
yzhliu: agree. maybe add later? haven't been very clear what/how to pass these information and how to use them
Currently as it is loading an AOT compiled module, avx instruction info won't help.

src/operator/contrib/tvmop/ufunc.cc
yzhliu: No. then "func_vadd_gpu" will never be invoked.

Makefile
yzhliu: will add later. need to get a windows...

python/mxnet/base.py
yzhliu: It should be doable. while might need non-trivial surgery to both. cc @tqchen if you have insight.

src/operator/contrib/tvmop/ufunc.cc
ZhennanQin: I mean does tvm have good support for all kinds of OpReqType, like inplace, addto?

src/operator/tvmop/op_module.cc
ZhennanQin: For standard release, there's only one aot compiled module, so the avx512 info seems not useful. But to support multi-version of libtvmop, it's useful to know If the avx512 version is loaded or not.

src/operator/contrib/tvmop/ufunc.cc
ZhennanQin: If a machine doesn't have gpu, will it have trouble to build mxnet because of this code?

Makefile
yzhliu: well, what I am trying to do is to attach `-DUSE_CUDA=$USE_CUDA_PATH` to TVM if `USE_CUDA_PATH` is set, otherwise attach `-DUSE_CUDA=$USE_CUDA`. I think it is fine as it is not exposed to users.

yzhliu:
@marcoabreu cmake support added

Makefile
marcoabreu: Could we have that logic by checking use_cuda then instead? I see your point and it's certainly simpler, but we have a lot of diverging logic branches around the use_cuda flag already, so I'd like to avoid further branching by introducing additional variables that branch off of that.

yzhliu: you mean ignore `USE_CUDA_PATH` ?

contrib/tvmop/opdef.py
```diff
@@ -0,0 +1,106 @@
...
+    """Decorator to define a tvm operator.
+    Parameters
+    ----------
+    name : str
+        function name
+    target : bool
+        {"cpu", "gpu", "cuda"}
+    auto_broadcast : bool
+        auto_broadcast=True allows one to implement broadcast computation
+        without considering whether dimension size equals to one.
```
tqchen: document returns

CMakeLists.txt
```diff
@@ -734,6 +734,28 @@ if(USE_DIST_KVSTORE)
   list(APPEND mxnet_LINKER_LIBS ${pslite_LINKER_LIBS})
 endif()
 
+if(USE_TVM_OP)
```
larroy: Is the definition missing?

src/operator/tvmop/op_module.cc
```diff
@@ -0,0 +1,117 @@
...
+  static const PackedFunc *f_load = Registry::Get("module._LoadFromFile");
+  std::lock_guard<std::mutex> lock(mutex_);
+  Module module = (*f_load)(filepath, "");
+  module_ptr_ = std::make_shared<Module>();
+  *module_ptr_ = module;
+}
+
+inline PackedFunc GetFunction(const std::shared_ptr<Module> &module,
```
larroy: please no inline.

Makefile
marcoabreu: No, what I mean is that the TVM related logic should be in a single place which infers from our base variables USE_TVM, USE_CUDA and USE_CUDA_PATH. Any variables that are now TVM specific should be constructed in a single block that has "if USE_TVM" as condition. These variables don't leave the scope of that block.

Right now, this introduces more and more TVM specific variables on a global level and scattered across the makefile. Instead it would be good if we have a single block to invoke TVM - this block will take care of converting all existing variables to their TVM counterpart. This improves the modularity and acts as a clear separation between our build system and TVM. It acts like a "translator". The issue to me here is that we are mirroring existing variables we already have at the global scope. Combining this fact with conditional operations on top of these mirrored variables, makes it harder to understand the process flow.

I hope that this makes it clearer

marcoabreu: Just look how you did it in cmake. It's a single and self enclosed block. I'd like to see the same here

src/operator/tvmop/op_module.cc
yzhliu: But avx512 flag is used during compiling operators. The binary code in `filepath` here already contains avx-enabled/disabled code.

src/operator/contrib/tvmop/ufunc.cc
yzhliu: I feel it is fine. The code itself won't cause build trouble (I have verified), "FCompute<gpu>" is just a key to NNVM registry. when and where to invoke the registered function is controlled during runtime. As user anyhow cannot construct a gpu-based NDArray on a CPU-only instance, there's no chance to call into such function.

Since there's no cuda code, I don't think we need another `.cu` file and put gpu registry there.

yzhliu:
Please review again @tqchen @marcoabreu @ZhennanQin @larroy

Makefile
yzhliu: addressed, does it look better now?

marcoabreu: Looks great, thank you!

Optional: Indentation (shift them one tab to the right because of the if-block) - but I'll leave it up to you :)

contrib/tvmop/opdef.py
```diff
@@ -0,0 +1,111 @@
...
+        return None
+
+
+def defop(name, target=None, auto_broadcast=False, **kwargs):
+    """Decorator to define a tvm operator.
+    Parameters
+    ----------
+    name : str
+        function name
+    target : bool
```
ZhennanQin: bool ->str

contrib/tvmop/compile.py
```diff
@@ -0,0 +1,58 @@
...
+                func_list = func_list_llvm if operator_def.target == "cpu" else func_list_cuda
+                func_lower = tvm.lower(sch, args,
+                                       name=operator_def.get_op_name(args),
+                                       binds=operator_def.get_binds(args))
+                func_list.append(func_lower)
+
+    lowered_funcs = {get_target("cpu") : func_list_llvm}
+    if len(func_list_cuda) > 0:
+        lowered_funcs[get_target("cuda")] = func_list_cuda
+    func_binary = tvm.build(lowered_funcs, name="tvmop")
```
ZhennanQin: Shall we propagate MXNet feature to libtvmop? E.g. avx and avx2. If one day we support multi-versions for libtvmop, then we'd better save the build option into tvmop and check it at loading time to verify if the expected version is loaded.

src/operator/contrib/tvmop/ufunc.cc
ZhennanQin: OK. But it's better to wrap this code with `#if MXNET_USE_CUDA` to avoid any potential risk.

include/mxnet/libinfo.h
```diff
@@ -185,6 +189,9 @@ enum : unsigned {
   SIGNAL_HANDLER,
   DEBUG,
 
+  // TVM operator
+  TVM_OP,
```
larroy: great!

contrib/tvmop/basic/ufunc.py
```diff
@@ -0,0 +1,50 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
larroy: why is it called ufunc?

contrib/tvmop/prepare_tvm.sh
```diff
@@ -0,0 +1,63 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+#!/bin/sh
+
```
larroy: set -exuo pipefail

contrib/tvmop/basic/ufunc.py
yzhliu: from numpy, universal function

contrib/tvmop/compile.py
yzhliu: That's a good idea. while I'm not sure where to attach the target info at this moment. How about we add a TODO and revisit later?

contrib/tvmop/prepare_tvm.sh
yzhliu: I prefer not, it works for bash, but is illegal in sh.

make/config.mk
```diff
@@ -62,6 +62,9 @@ ADD_LDFLAGS =
 # the additional compile flags you want to add
 ADD_CFLAGS =
 
+# whether to build operators written in TVM
+USE_TVM_OP = 0
```
apeforest: add this flag to make/osx.mk as well.

tests/python/unittest/test_tvm_op.py
```diff
@@ -0,0 +1,34 @@
...
+
+import mxnet as mx
+from mxnet.test_utils import same
+from mxnet.runtime import Features
+
+_features = Features()
+
+def test_tvm_broadcast_add():
+    if _features.is_enabled("TVM_OP"):
+        a = mx.nd.normal(shape=(2, 3, 4))
```
apeforest: use rand_shape_nd from mxnet.test_utils to generate test tensors?

karan6181:
@mxnet-label-bot add [Backend, Operator, pr-awaiting-response]

contrib/tvmop/prepare_tvm.sh
```diff
@@ -0,0 +1,63 @@
...
+    DISTRIB_ID=$(cat /etc/*-release | grep DISTRIB_ID | sed 's/DISTRIB_ID=//g' | tr '[:upper:]' '[:lower:]')
+    DISTRIB_RELEASE=$(cat /etc/*-release | grep DISTRIB_RELEASE | sed 's/DISTRIB_RELEASE=//g' | tr '[:upper:]' '[:lower:]')
+    if [ "$DISTRIB_ID" = "ubuntu" ]; then
+        LLVM_PKG=${LLVM_PKG}-${DISTRIB_ID}-${DISTRIB_RELEASE}
+    else
+        echo "Downloading LLVM only supports Ubuntu. Please download manually."
+        exit 1
+    fi
+else
+    echo "Cannot identify operating system. Try downloading package manually."
```
szha: this script is automatically run and this error message could appear without context to llvm.

Makefile
yzhliu: opinion of aesthetics  :)

src/operator/contrib/tvmop/ufunc.cc
yzhliu: good question. I don't have a good answer to it as for now.

src/operator/tvmop/op_module.h
```diff
@@ -0,0 +1,63 @@
...
+#include <mxnet/base.h>
+#include <mxnet/op_attr_types.h>
+#include <mutex>
+#include <string>
+#include <vector>
+
+namespace tvm {
+namespace runtime {
+
+class Module;
```
szha: is this still needed?

src/operator/contrib/tvmop/ufunc.cc
```diff
@@ -0,0 +1,66 @@
...
+    .set_num_inputs(2)
+    .set_num_outputs(1)
+    .add_argument("a", "NDArray-or-Symbol", "first input")
+    .add_argument("b", "NDArray-or-Symbol", "second input")
+    .set_attr<mxnet::FInferShape>("FInferShape", BinaryBroadcastShape)
+    .set_attr<nnvm::FInferType>("FInferType", mxnet::op::ElemwiseType<2, 1>)
+    .set_attr<mxnet::FCompute>("FCompute<cpu>", mxnet::op::TVMBroadcastCompute<func_vadd_cpu>)
+#if MXNET_USE_CUDA
+    .set_attr<mxnet::FCompute>("FCompute<gpu>", mxnet::op::TVMBroadcastCompute<func_vadd_gpu>);
+#endif  // MXNET_USE_CUDA
```
mikemwx: There will be a missing semicolon if MXNET_USE_CUDA is off.

hgt312: may change to this
```
#if MXNET_USE_CUDA
    .set_attr<mxnet::FCompute>("FCompute<gpu>", mxnet::op::TVMBroadcastCompute<func_vadd_gpu>)
#endif  // MXNET_USE_CUDA
    .set_attr<mxnet::FCompute>("FCompute<cpu>", mxnet::op::TVMBroadcastCompute<func_vadd_cpu>);
```

PR #15560, opened 2019-07-16
Add fp16 support for topk

anirudhacharya:
## Description ##
Add fp16 support for topk operator. Required for certain machine translation tasks( and other NLP related tasks).

### Essentials ###

### Changes ###
 - add fp16 support for topk operator
 - Fixes https://github.com/apache/incubator-mxnet/issues/14125 https://github.com/apache/incubator-mxnet/issues/11156 https://github.com/apache/incubator-mxnet/issues/12705

For review - @anirudh2290 @apeforest @sxjscience
Same as this PR - https://github.com/apache/incubator-mxnet/pull/14912

### DIFF
--- a/src/operator/tensor/ordering_op-inl.h
+++ b/src/operator/tensor/ordering_op-inl.h
@@ -377,7 +377,6 @@ MSHADOW_FORCE_INLINE void TopKSort(const Tensor<gpu, 1, DType>& dat,
    * \param resource temporary resource handler
    * \param src the Source blob
    * \param ret the destination blobs
-   * \param k the K elements to keep
    * \param param the topk parameters
    * \tparam xpu the device type.
    * \tparam DType type of the output value/mask.
--- a/src/operator/tensor/ordering_op-inl.h
+++ b/src/operator/tensor/ordering_op-inl.h
@@ -420,8 +420,6 @@ void TopKImpl(const RunContext &ctx,
     mxnet::op::SortByKeyWorkspaceSize<int, int, xpu>(src.Size()));
   temp_size = std::max(temp_size,
     mxnet::op::SortByKeyWorkspaceSize<int, DType, xpu>(src.Size()));
-  temp_size = std::max(temp_size,
-    mxnet::op::SortByKeyWorkspaceSize<DType, int, xpu>(src.Size()));
   // Additional temp space for gpu full sorts for batch ids.
   temp_size += PadBytes(sizeof(index_t) * src.Size(), alignment);
   // Temp space for cpu sorts.
@@ -563,13 +561,13 @@ void TopK(const nnvm::NodeAttrs& attrs,
           const std::vector<TBlob>& outputs) {
   const TopKParam& param = nnvm::get<TopKParam>(attrs.parsed);
   if (param.ret_typ == topk_enum::kReturnIndices || param.ret_typ == topk_enum::kReturnBoth) {
-    MSHADOW_REAL_TYPE_SWITCH(inputs[0].type_flag_, DType, {
+    MSHADOW_TYPE_SWITCH(inputs[0].type_flag_, DType, {
       MSHADOW_TYPE_SWITCH(param.dtype, IDType, {
         TopKImpl<xpu, DType, IDType>(ctx.run_ctx, ctx.requested[0], req, inputs[0], outputs, param);
       })
     });
   } else {
-    MSHADOW_REAL_TYPE_SWITCH(inputs[0].type_flag_, DType, {
+    MSHADOW_TYPE_SWITCH(inputs[0].type_flag_, DType, {
       TopKImpl<xpu, DType, index_t>(ctx.run_ctx, ctx.requested[0], req, inputs[0], outputs, param);
     });
   }
@@ -695,13 +693,13 @@ void TopKBackward_(const nnvm::NodeAttrs& attrs,
                    const std::vector<TBlob>& outputs) {
   const TopKParam& param = nnvm::get<TopKParam>(attrs.parsed);
   if (param.ret_typ == topk_enum::kReturnBoth) {
-    MSHADOW_REAL_TYPE_SWITCH(inputs[0].type_flag_, DType, {
+    MSHADOW_TYPE_SWITCH(inputs[0].type_flag_, DType, {
       MSHADOW_TYPE_SWITCH(param.dtype, IDType, {
         TopKBackwardImpl<xpu, DType, IDType>(ctx, inputs, req, outputs, param);
       });
     });
   } else if (param.ret_typ == topk_enum::kReturnValue) {
-    MSHADOW_REAL_TYPE_SWITCH(inputs[0].type_flag_, DType, {
+    MSHADOW_TYPE_SWITCH(inputs[0].type_flag_, DType, {
       TopKBackwardImpl<xpu, DType, index_t>(ctx, inputs, req, outputs, param);
     });
   } else {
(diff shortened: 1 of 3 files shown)


karan6181:
@mxnet-label-bot add [FP16, Operator, pr-awaiting-review]

haojin2:
@anirudhacharya Seems like this change is breaking some tests, please fix those.

src/operator/tensor/ordering_op-inl.h
```diff
@@ -420,8 +420,6 @@ void TopKImpl(const RunContext &ctx,
     mxnet::op::SortByKeyWorkspaceSize<int, int, xpu>(src.Size()));
   temp_size = std::max(temp_size,
     mxnet::op::SortByKeyWorkspaceSize<int, DType, xpu>(src.Size()));
-  temp_size = std::max(temp_size,
```
anirudhacharya: there was this - https://github.com/apache/incubator-mxnet/pull/14912#pullrequestreview-237632437 concern here by @sxjscience

sxjscience: I think this line is to make sure that the code will be correct no matter how we change the `SortByKeyWorkspaceSize` function. For example, if we later change our `SortByKeyWorkspaceSize` as something like:
```
mxnet::op::SortByKeyWorkspaceSize<int, DType, xpu>() --> Return 0
mxnet::op::SortByKeyWorkspaceSize<DType, int, xpu>() --> Return 1000000
```
The current revision will raise an error because the `temp_size` is not correctly updated. I suggest keeping this line to avoid issues in the future.

anirudhacharya: Okay. I am not sure why, but without removing these lines, compiling the code with cuda enabled and fp16 enabled fails.

anirudhacharya:
> @anirudhacharya Seems like this change is breaking some tests, please fix those.

the tests that are failing are tests for topk with `int32` dtype in test_ndarray:test_order

src/operator/tensor/ordering_op-inl.h
sxjscience: Please try to locate the source of the error. Removing the line does not solve the problem.

piyushghai:
@anirudhacharya What's the path forward for this PR ?

drivanov:
@anirudhacharya: I see couple bugs in _sort_op-inl.cuh_, which show up when fp16 for topK is implemented. First of all, the lines 52, 63 are identical (see https://github.com/apache/incubator-mxnet/blob/master/src/operator/tensor/sort_op-inl.cuh):
```return static_cast<mshadow::half::half_t>(lhs) < static_cast<mshadow::half::half_t>(rhs);```
therefore:
```
cuda::less_half<half>()
cuda::greater_half<half>()
```
are identical too.
Second, instead of 
```
    thrust::stable_sort_by_key(
      thrust::cuda::par.on(stream),
          key_iter, key_iter + (keys.size(0)), value_iter, cuda::greater_half<half>());
```
(here and on couple of other places) we should use 
```
    thrust::stable_sort_by_key(
      thrust::cuda::par.on(stream),
          key_iter, key_iter + (keys.size(0)), value_iter.get(), cuda::greater_half<half>());
```

**Comment by anirudhacharya** [comment]
@drivanov yes, `cuda::less_half<half>()` and `cuda::greater_half<half>()` are defined similarly and should be looked into. 
But the failure, as far as fp16 for `topk` and this PR goes, is here - https://github.com/apache/incubator-mxnet/blob/master/src/operator/tensor/sort_op-inl.cuh#L75. There is a problem when we try to perform `DeviceRadixSort` with half precision values. The documentation - https://nvlabs.github.io/cub/structcub_1_1_device_radix_sort.html says that `DeviceRadixSort` handles all numeric primitive types, including half precision. Either there is an issue with the way it is being called here or there is an issue with the routine itself.

**Comment by drivanov** [comment]
@anirudhacharya:  `cuda::less_half<half>()` and `cuda::greater_half<half>()` are not only defined **similarly**, but they are identical

**Comment by anirudhacharya** [comment]
> @anirudhacharya: `cuda::less_half<half>()` and `cuda::greater_half<half>()` are not only defined **similarly**, but they are **identica**l and, definitely, it is a bug.

yes, sorry for sounding ambiguous. The intent of my previous comment was to ack that it was a bug.

> As for the problem with `DeviceRadixSort`, yes, it's not comiled for `mshadow::half::half_t`, but it could be compiled for `__half` and I think we could use that. The solution here would be the usage of separate templates `SortPairsWorkspaseSize`...

Thanks for this suggestion, I will try this and see if it works. On the other hand if you have a solution for this already and you would like to make a PR for it, let me know, I can close this and we can merge your code.

**Comment by drivanov** [comment]
@anirudhacharya: Yes, I do have a solution, but it is not quite completed yet. 
1. I added the `np.float16` testing to `test_ndarray.py:test_order` and all these tests are OK. But these tests are using the changes I've made for PR14443 (https://github.com/apache/incubator-mxnet/pull/14443), which is still open.
2. There is another test `test_operator.py:test_order`) which also was design `mx.sym.topk` and other related function. Now I am working on extension of these tests for `np.float16`. I plan to finish this task till tomorrow (08/23/2019)

**Comment by anirudhacharya** [comment]
> Yes, I do have a solution, but it is not quite completed yet.
> 
> 1. I added the `np.float16` testing to `test_ndarray.py:test_order` and all these tests are OK. But these tests are using the changes I've made for PR #14443, which is still open.

then we can probably merge this, since i think this pr is pretty much merge ready.

@sxjscience @haojin2 could you review/merge this.

File: src/operator/tensor/sort_op-inl.cuh [hunk 2]
```diff
@@ -227,18 +251,18 @@ SortByKeyImpl(mshadow::Tensor<gpu, 1, KDType> keys,
...
-    reinterpret_cast<half*>(values.dptr_));
+  thrust::device_ptr<__half> key_iter = thrust::device_pointer_cast(
+    reinterpret_cast<__half*>(keys.dptr_));
+  thrust::device_ptr<__half> value_iter = thrust::device_pointer_cast(
+    reinterpret_cast<__half*>(values.dptr_));
   if (is_ascend) {
     thrust::stable_sort_by_key(
       thrust::cuda::par.on(stream),
-      key_iter, key_iter + (keys.size(0)), value_iter, cuda::less_half<half>());
+      key_iter, key_iter + (keys.size(0)), value_iter.get(), cuda::less_half<__half>());
```
sxjscience: Why adding `.get()` to the `value_iter`?

tests/python/unittest/test_operator.py
```diff
@@ -4372,6 +4347,45 @@ def get_large_matrix():
                            expected=[gt_topk(dat=a_npy, axis=1, ret_typ="indices", k=1,
                                              is_ascend=True)])
 
+    for dtype in [np.float16, np.float32, np.float64]:
```
sxjscience: Would you also test for `int32` and `int64`? Also, the testing for `uint8` and `int8` can be added separately.

src/operator/tensor/sort_op-inl.cuh
drivanov: My experiments shows that without `.get()` it is not working

sxjscience: @drivanov OK, I see. It looks inconsistent at first glance.

tests/python/unittest/test_operator.py
anirudhacharya: it is already test in `test_ndarray.py` here - https://github.com/apache/incubator-mxnet/blob/master/tests/python/unittest/test_ndarray.py#L908

PR #15561, opened 2019-07-16
Julia path patch

aaronmarkham:
## Description ##
Another followup PR to fix the Julia site build and unblock website publishing...
Since the website builds so many versions, it does so in their own folders inside `/mxnet/docs/build_version_doc/apache-mxnet/...` and this makes it difficult for Julia to find when it's expecting the binaries in `/work/mxnet/lib`. After digging through the Julia source and testing a bunch of things I found that `$LD_LIBRARY_PATH` doesn't seem to do anything, but `$MXNET_HOME` is what Julia is looking for. It then manually appends `lib`. 
Unfortunately, according to the [MXNet environment variables](https://mxnet.incubator.apache.org/versions/master/faq/env_var.html) page, this same env var is used elsewhere to store models in `.mxnet`, not a typical `$HOME/incubator-mxnet` or what is often called `$MXNET_ROOT` in tutorials.

Unblocking website publishing is my primary goal, not dealing with conflicting environment variables, so I'm patching this for now, but I think the Julia package should updated to use `$MXNET_LIBRARY_PATH` or `$LD_LIBRARY_PATH` instead. 

My test run used a test CI pipeline that points to my fork:
http://jenkins.mxnet-ci.amazon-ml.com/job/test-website-build/

The "successful" run was here. It fails because there's no place to publish, which is fine. Everything else worked: http://jenkins.mxnet-ci.amazon-ml.com/job/test-website-build/3/console
We might want to keep this test pipeline around... it took me awhile to remember how to create one for testing, and I definitely should have done this thorough test in the first place.

It was able to find Julia, and it used the locally compiled binary based off of the target branch: this one.

I'm hoping this is the last surprise that the CI environment is going to give me for getting Julia to publish.

Previous PRs for the Julia docs publishing:
https://github.com/apache/incubator-mxnet/pull/15554
https://github.com/apache/incubator-mxnet/pull/15523
https://github.com/apache/incubator-mxnet/pull/15454

### DIFF
--- a/docs/mxdoc.py
+++ b/docs/mxdoc.py
@@ -98,7 +98,7 @@ def build_mxnet(app):
 
 def build_julia_docs(app):
     """build Julia docs"""
-    os.environ['MXNET_HOME'] = str(os.path.abspath('{}/..'.format(app.builder.srcdir)))
+    os.environ['MXNET_HOME'] = os.path.abspath(os.path.join(app.builder.srcdir, '..'))
     print("Julia will check for MXNet in {}/lib".format(os.environ.get('MXNET_HOME')))
     dest_path = app.builder.outdir + '/api/julia/site'
     _run_cmd('cd {}/.. && make -C julia/docs'.format(app.builder.srcdir))
--- a/docs/mxdoc.py
+++ b/docs/mxdoc.py
@@ -98,6 +98,8 @@ def build_mxnet(app):
 
 def build_julia_docs(app):
     """build Julia docs"""
+    os.environ['MXNET_HOME'] = str(os.path.abspath('{}/..'.format(app.builder.srcdir)))
+    print("Julia will check for MXNet in {}/lib".format(os.environ.get('MXNET_HOME')))
     dest_path = app.builder.outdir + '/api/julia/site'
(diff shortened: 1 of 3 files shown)


iblis17:
Okay, I can make a patch for rename `MXNET_HOME` to `MXNET_ROOT`, then add a search candidate for `MXNET_LIBRARY_PATH`.

In case of `LD_LIBRARY_PATH`, I think it works already..
This variable will be picked up by C's `dlopen`.

```shell
% LD_LIBRARY_PATH=/home/iblis/git/mxnet/lib julia -e 'using Libdl; println(Libdl.dlopen("libmxnet.so"))' 
Ptr{Nothing} @0x000055b0eca49c50
```

docs/mxdoc.py
```diff
@@ -98,6 +98,8 @@ def build_mxnet(app):
 
 def build_julia_docs(app):
     """build Julia docs"""
+    os.environ['MXNET_HOME'] = str(os.path.abspath('{}/..'.format(app.builder.srcdir)))
```
iblis17:
```suggestion
    os.environ['MXNET_HOME'] = os.path.abspath(os.path.join(app.builder.srcdir, '..'))
```

Issue #8966, opened 2017-12-06
NaiveEngine error in multi-threading 


xinghedyc:
when I use NaiveEngine in openmp multi-threading, I binded 2 executor on gpu(0) and gpu(1). 
and do exe.forward parallelly. However I got some error when running the program:

```
[08:56:51] /data1/yuchendai/lpr_sdk/mxnet/dmlc-core/include/dmlc/logging.h:308: [08:56:51] src/engine/naive_engine.cc:169: Check failed: this->req_completed_ NaiveEngine only support synchronize Push so far

Stack trace returned 10 entries:
[bt] (0) ./bin/test(_ZN4dmlc15LogMessageFatalD1Ev+0x30) [0x45d614]
[bt] (1) /data1/yuchendai/lpr_sdk/mxnet/lib/libmxnet.so(_ZN5mxnet6engine11NaiveEngine9PushAsyncESt8functionIFvNS_10RunContextENS0_18CallbackOnCompleteEEENS_7ContextERKSt6vectorIPNS0_3VarESaISA_EESE_NS_10FnPropertyEiPKc+0x3b4) [0x7f3001f9f8f4]
[bt] (2) /data1/yuchendai/lpr_sdk/mxnet/lib/libmxnet.so(_ZN5mxnet6engine11NaiveEngine4PushEPNS0_3OprENS_7ContextEib+0xad) [0x7f3001f9fd4d]
[bt] (3) /data1/yuchendai/lpr_sdk/mxnet/lib/libmxnet.so(_ZN5mxnet4exec13GraphExecutor6RunOpsEbmm+0x584) [0x7f3002014dc4]
[bt] (4) /data1/yuchendai/lpr_sdk/mxnet/lib/libmxnet.so(MXExecutorForward+0x15) [0x7f3001fd8935]
[bt] (5) ./bin/test(_ZN5mxnet3cpp8Executor7ForwardEb+0x45) [0x47a0f3]
[bt] (6) ./bin/test(_ZN9PLATE_OCR16CustomizationOcr7forwardERKSsRSt6vectorINS_4WordESaIS4_EE+0x30b) [0x4793d9]
[bt] (7) ./bin/test(_ZNK9PLATE_OCR25CustomizationOcrInterface9RecognizeERKSsRNS_12RecognizeResEi+0x53) [0x45cbff]
[bt] (8) ./bin/test() [0x458c77]
[...]

PR #15574, opened 2019-07-17
fix naive engine for multi-threaded inference

arcadiaphy:
## Description ##
In the [cpp example of image classification](https://github.com/apache/incubator-mxnet/tree/master/example/image-classification/predict-cpp), `MXPredCreateMultiThread` is used to perform multi-threaded inference with naive engine. The `req_completed_` variable in naive engine is not thread-safe, which is fixed in this PR.

The problem is reported in #8966.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/src/engine/naive_engine.cc
+++ b/src/engine/naive_engine.cc
@@ -155,9 +155,9 @@ class NaiveEngine final : public Engine {
                  int priority = 0,
                  const char* opr_name = nullptr,
                  bool wait = false) override {
+    bool req_completed = false;
     CallbackOnComplete callback = CreateCallback(
-        NaiveEngine::OnComplete, nullptr);
-    this->req_completed_ = false;
+        NaiveEngine::OnComplete, &req_completed);
     profiler::Profiler *profiler = profiler::Profiler::Get();
     auto opr_deleter = [this](NaiveOpr* p) {
       this->DeleteOperator(p);
@@ -202,7 +202,7 @@ class NaiveEngine final : public Engine {
     for (auto var : mutable_vars) {
       ++var->version_;
     }
-    CHECK(this->req_completed_)
+    CHECK(req_completed)
         << "NaiveEngine only support synchronize Push so far";
     if (profiling) {
       opr->opr_profile->stop();
@@ -235,14 +235,9 @@ class NaiveEngine final : public Engine {
   // callback to oncomplete
   static void OnComplete(Engine *engine, void *param,
                          const dmlc::Error* error) {
-    static_cast<NaiveEngine*>(engine)->req_completed_ = true;
+    bool *req_completed = static_cast<bool*>(param);
+    *req_completed = true;
   }
-  // whether action is completed
-#if DMLC_CXX11_THREAD_LOCAL
-  static thread_local bool req_completed_;
-#else
-  static MX_THREAD_LOCAL bool req_completed_;
-#endif
   /*! \brief whether it is during shutdown phase*/
   std::atomic<bool> shutdown_phase_{false};
   // CPU stream
@@ -265,12 +260,6 @@ class NaiveEngine final : public Engine {
 Engine *CreateNaiveEngine() {
   return new NaiveEngine();
 }
-
-#if DMLC_CXX11_THREAD_LOCAL
-thread_local bool NaiveEngine::req_completed_ = false;
-#else
-MX_THREAD_LOCAL bool NaiveEngine::req_completed_ = false;
-#endif
 
 }  // namespace engine
 }  // namespace mxnet
--- a/src/engine/naive_engine.cc
+++ b/src/engine/naive_engine.cc
@@ -238,7 +238,11 @@ class NaiveEngine final : public Engine {
     static_cast<NaiveEngine*>(engine)->req_completed_ = true;
   }
   // whether action is completed
-  bool req_completed_;
+#if DMLC_CXX11_THREAD_LOCAL
+  static thread_local bool req_completed_;
+#else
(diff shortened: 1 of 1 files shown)


apeforest:
@anirudh2290 Could you please take a look?

marcoabreu:
Thanks for the contribution, but MXNets frontend is not considered threadsafe in general. Thus, I'd prefer if we hold up with this fix.

I feel like we should rather do a proper analysis about the threadsafety of MXNet and then act upon that. Small patches won't really solve the problem but might possibly introduce further issues.

src/engine/naive_engine.cc
```diff
@@ -238,7 +238,11 @@ class NaiveEngine final : public Engine {
     static_cast<NaiveEngine*>(engine)->req_completed_ = true;
   }
   // whether action is completed
-  bool req_completed_;
+#if DMLC_CXX11_THREAD_LOCAL
+  static thread_local bool req_completed_;
```
apeforest: According to C++ standards, thread_local implies static.
https://stackoverflow.com/questions/22794382/are-c11-thread-local-variables-automatically-static

pengzhao-intel:
@ZhennanQin could you help to take a review for the fix?

ZhennanQin:
I just evaluated the thread-safe condition for `NaiveEngine`, and I agree with @marcoabreu that it's not enough with this point fix. There's too much data race when using MXNet in multi-thread usage, even for `NaiveEngine`, we need to take care all class member not only `req_completed_`. Creating private `NaiveEngine` for each thread is another option. So we need a whole picture plan for MXNet multi-thread support.

src/engine/naive_engine.cc
```diff
@@ -238,7 +238,11 @@ class NaiveEngine final : public Engine {
...
+#else
+  static MX_THREAD_LOCAL bool req_completed_;
+#endif
```
anirudh2290: This introduces an additional overhead for thread local, and the number of ops in mxnet model can be in the scale of 100s. I am not sure we should add this without a build flag since we are adding overhead for single threaded use cases too.

anirudh2290:
@marcoabreu @ZhennanQin I agree we need a bigger solution for the problem at hand.Having said that, Looks like the scope of the PR is limited to fix the existing multi threaded inference which is limited to the C Predict API which should be fine.

arcadiaphy:
@apeforest @ZhennanQin @marcoabreu @anirudh2290 
Multi-threaded inference is very common in deploying, so we need to have a solution on this scenario and provide a workable example for users. Let me sum up what I've known now.

Since threaded engine uses multi-thread scheduling for operator computation, so a natural idea is to utilize it for parallel inference which is covered in this [proposal](https://cwiki.apache.org/confluence/display/MXNET/Parallel+Inference+in+MXNet). There are two main problem for this kind of usage in mxnet:

- The threaded engine is not thread-safe, so we need to restrict the computation pushing and result pulling in one thread, otherwise we'll have all kinds of errors in engine.
- For computation pushing, one thread is OK, but when we pull the result using functions like `asnumpy`, it will block the thread and increase the latency.

Due to these problems and the lacking of a workable example, so I focus on the naive engine instead. The only example in mxnet now is cpp predict with `MXPredCreateMultiThread` API, which creates executors for each thread by sharing model parameters, a very seductive option with regard to both speed and memory footprint. In production environment, naive engine looks like a more usable method by restricting one inference to only one thread. Also, I think this way is the basis of other advanced methods, if you can't make it work in naive engine, then very likely you can't make it work in threaded engine too. There are two problems in this method too:

- The naive engine is not thread-safe, which I want to fix in this PR. Without this fix, the cpp example is broken. Maybe we'll have a better fix by planning the bigger solution on parallel inference, but using thread_local is the easiest way now.
- Perhaps there are some issues in MKLDNN calling, I've found a very strange phenomenon which is mentions in #15576. I hope that this can be fixed too since MKLDNN is so good.

arcadiaphy:
@apeforest @anirudh2290 
Do we really need `req_completed_` in naive engine? I'm inclined to remove it now.

anirudh2290:
@arcadiaphy I guess that would be nice to explore. We can remove it only if we can make sure that all exec_funs pushed make a call to callback, since its original purpose is to prevent async calls being pushed.

src/engine/naive_engine.cc
marcoabreu: Please not another build flag - at some point we're gonna spend more money on CI than the NASA :D

Could we make some benchmarks to determine the impact here?

I'm not that deep inside the engine code, so excuse me if that's a stupid question. But why are we declaring the internal variables of the engine as thread local instead of creating thread-local instantiations of the engines. Each thread gets its own engine and we switch all static variables to private ones. The current design of the engine doesn't seem to have the creation of multiple instantiations in mind so we should fix that instead of making all the static variables threadlocal.

marcoabreu:
I understand that you're trying to keep the changes to a minimum, but judging from the static variable definition, it doesn't seem like the engine was designed to be threadsafe. I'd prefer if we make the engine only access internal-variables, get rid of static ones (aka proper object oriented design) and add synchronization to the APIs.

But I don't really have a dog in that hunt, so take my opinion with a grain of salt here.

arcadiaphy:
@anirudh2290 Yes, in order to prevent async call from being pushed, using a single req_completed_ is not enough. We can add the req_completed_ to NaiveOpr and check whether it's been set to true in complete callback.

piyushghai:
@apeforest @anirudh2290 Gentle ping for the review.

PR #15551, opened 2019-07-16
Bypass cuda/cudnn checks if no driver.

DickJC123:
## Description ##
This provides the fix for the issue described in https://github.com/apache/incubator-mxnet/issues/15548.  Thank you @vrakesh for your efforts detecting and posting this issue.  

The problem is that libmxnet.so can be built with MXNET_USE_CUDA and MXNET_USE_CUDNN, but be available on a system with no installed driver (or GPUs).  The absence of a driver can be detected by a 0 driver_version set by cudaDriverGetVersion(&driver_version).  This fix includes code that bypasses the checks introduced by my recent PR https://github.com/apache/incubator-mxnet/pull/15449 for the case when there is no driver.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [X ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ X] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #15591, opened 2019-07-18
Export symbols for checking cuda/cudnn versions

### LINKED ISSUE #15548: MXNet GPU build on CPU machine fails
Post PR #15449 , when trying to build mxnet with GPU support on a CPU instance fails, with the following error

Running: OpWrapperGenerator.py
terminate called after throwing an instance of 'dmlc::Error'
  what():  [00:10:08]include/mxnet/base.h:427: Check failed: e == cudaSuccess (35 vs. 0) :  CUDA: CUDA driver version is insufficient for CUDA runtime version
Stack trace:
  [bt] (0) build/lib/libmxnet.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x25) [0x7fcf63bb2793]
  [bt] (1) build/lib/libmxnet.so(mxnet::Context::GetGPUCount()+0x11b) [0x7fcf66d39325]
  [bt] (2) build/lib/libmxnet.so(+0x5ebcfaa) [0x7fcf66df5faa]
  [bt] (3) /build/lib/libmxnet.so(+0x5ebd180) [0x7fcf66df6180]
  [bt] (4) build/lib/libmxnet.so(+0x5ebd1b7) [0x7fcf66df61b7]
  [bt] (5) /lib64/ld-linux-x86-64.so.2(+0xe9cf) [0x7fcfd1f419cf]
  [bt] (6) /lib64/ld-linux-x86-64.so.2(+0x131f5) [0x7fcfd1f461f5]
  [bt] (7) /lib64/ld-linux-x86-64.so.2(+0xe5e6) [0x7fcfd1f415e6]
  [bt] (8) /lib64/ld-linux-x86-64.so.2(+0x1299a) [0x7fcfd1f4599a]

The cmake setting is as follows

```bash
 cmake \
    -DUSE_OPENCV=${USE_OPENCV:-ON}\
    -DUSE_OPENMP=ON\
    -DUSE_CUDA=${USE_CUDA:-ON}\
    -DUSE_CUDNN=${USE_CUDA:-ON}\
    -DUSE_LAPACK=OFF\
    -DUSE_MKL_IF_AVAILABLE=ON\
    -DUSE_MKLML_MKL=OFF\
    -DUSE_MKLDNN=${USE_MKLDNN:-ON}\
    -DUSE_SSE=ON\
    -DUSE_GPERFTOOLS=OFF\
    -DUSE_JEMALLOC=${USE_JEMALLOC:-ON}\
    -DUSE_PROFILER=${USE_PROFILER:-ON}\
    -DUSE_DIST_KVSTORE=${USE_DIST_KVSTORE:-ON}\
    -DUSE_CPP_PACKAGE=ON\
    -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE \
    -DCUDA_ARCH_NAME=Manual\
    -DCUDA_ARCH_BIN=37,70,75\
    -DCUDA_ARCH_PTX=70,75 \
    -DINSTALL_PYTHON_VERSIONS="${PYTHON_VERSIONS:-2.7;3.4;3.5;3.6}"\
    -DENABLE_CUDA_RTC=${ENABLE_CUDA_RTC:-OFF}\
    -DUSE_SIGNAL_HANDLER=OFF\
    -DUSE_NCCL=${USE_NCCL:-OFF}     

cmake cpp_package_op_h
```

### LINKED ISSUE #15578: MXNet cu100 nightly release breaks Horovod integration tests
## Description
For detailed reproducible steps and environment information, please refer to: https://github.com/horovod/horovod/issues/1217

yuxihu:
Fixes #15578 

Two functions (CudaLibChecks and CuDNNLibChecks) were added in #15551 to fix #15548. These two functions need to be exported in our pip package such that downstream projects consuming mxnet library will not encounter [undefined symbol error](https://github.com/horovod/horovod/issues/1217#issuecomment-512680079) as shown below:

```
OSError: /home/ubuntu/anaconda3/envs/mxnet_p36/lib/python3.6/site-packages/horovod/mxnet/mpi_lib.cpython-36m-x86_64-linux-gnu.so: undefined symbol: _ZN5mxnet7Context13CudaLibChecksEv
```

### DIFF
--- a/make/config/libmxnet.sym
+++ b/make/config/libmxnet.sym
@@ -13,3 +13,4 @@ Java_org_apache_mxnet*
 *on_enter_api*
 *on_exit_api*
 *MXAPISetLastError*
+*LibCheck*
--- a/make/config/libmxnet.ver
+++ b/make/config/libmxnet.ver
@@ -15,5 +15,6 @@
         *on_enter_api*;
         *on_exit_api*;
         *MXAPISetLastError*;
+        *LibCheck*;
     local: *;
 };


yuxihu:
@apeforest @szha @ptrendx @DickJC123 Please help review. Thanks.

yuxihu:

DickJC123:
LGTM.

@yuxihu Thanks for stepping in quickly to provide this fix.  I would appreciate any additional insight you have that might help me from introducing this sort of problem in the future.  For example, I see that the wildcard pattern you introduced to ./make/config/libmxnet.{sym,ver} matches the missing entrypoints:
```
$ nm libmxnet.so | grep LibCheck | grep " T "
0000000003c889d0 T _ZN5mxnet7Context13CudaLibChecksEv
0000000003c89730 T _ZN5mxnet7Context14CuDNNLibChecksEv
```
However, I can also perform:
```
$ nm libmxnet.so | grep ResourceManager | grep Get | grep " T "
0000000003c77c00 T _ZN5mxnet15ResourceManager3GetEv
```
Why is no wildcard pattern similarly needed to cover ResourceManager::Get(), as created in ./include/mxnet/resource.h and ./src/resource.cc?

szha:
Why didn't the API follow existing naming convention?

yuxihu:
> Why didn't the API follow existing naming convention?

@szha What do you suggest to name these two functions CudaLibChecks and CuDNNLibChecks? Adding MX or similar as the prefix? If they can be named to match existing symbol wildcards, we do not need to export them explicitly.

marcoabreu:
+1 to renaming the API to follow the naming scheme.

While we're at it, can we please introduce a check to CI to make sure that all functions match the defined schemes?

yuxihu:
> Why is no wildcard pattern similarly needed to cover ResourceManager::Get(), as created in ./include/mxnet/resource.h and ./src/resource.cc?

@DickJC123 Are you using the mxnet pip package? The symbol is not exported in the pip package.
```
(mxnet_p27) ubuntu@ip-172-31-21-35:~$ pip install mxnet-cu100 --pre
Collecting mxnet-cu100
  Using cached https://files.pythonhosted.org/packages/b7/c4/1e0c4d21ed6dca0890a9db3e23f78c9d8df1651330547c7778aa1a974ea6/mxnet_cu100-1.5.0b20190717-py2.py3-none-manylinux1_x86_64.whl
Requirement already satisfied: graphviz<0.9.0,>=0.8.1 in ./anaconda3/envs/mxnet_p27/lib/python2.7/site-packages (from mxnet-cu100) (0.8.4)
Collecting numpy<2.0.0,>1.16.0 (from mxnet-cu100)
  Downloading https://files.pythonhosted.org/packages/1f/c7/198496417c9c2f6226616cff7dedf2115a4f4d0276613bab842ec8ac1e23/numpy-1.16.4-cp27-cp27mu-manylinux1_x86_64.whl (17.0MB)
    100% |████████████████████████████████| 17.0MB 2.6MB/s
Requirement already satisfied: requests<3,>=2.20.0 in ./anaconda3/envs/mxnet_p27/lib/python2.7/site-packages (from mxnet-cu100) (2.20.0)
Requirement already satisfied: idna<2.8,>=2.5 in ./anaconda3/envs/mxnet_p27/lib/python2.7/site-packages (from requests<3,>=2.20.0->mxnet-cu100) (2.6)
Requirement already satisfied: chardet<3.1.0,>=3.0.2 in ./anaconda3/envs/mxnet_p27/lib/python2.7/site-packages (from requests<3,>=2.20.0->mxnet-cu100) (3.0.4)
Requirement already satisfied: urllib3<1.25,>=1.21.1 in ./anaconda3/envs/mxnet_p27/lib/python2.7/site-packages (from requests<3,>=2.20.0->mxnet-cu100) (1.23)
Requirement already satisfied: certifi>=2017.4.17 in ./anaconda3/envs/mxnet_p27/lib/python2.7/site-packages (from requests<3,>=2.20.0->mxnet-cu100) (2019.3.9)
Installing collected packages: numpy, mxnet-cu100
  Found existing installation: numpy 1.14.5
    Uninstalling numpy-1.14.5:
      Successfully uninstalled numpy-1.14.5
Successfully installed mxnet-cu100-1.5.0b20190717 numpy-1.16.4
You are using pip version 10.0.1, however version 19.1.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```
```
(mxnet_p27) ubuntu@ip-172-31-21-35:~$ nm ./anaconda3/envs/mxnet_p27/lib/python2.7/site-packages/mxnet/libmxnet.so | grep ResourceManager | grep Get | grep " T "
nm: ./anaconda3/envs/mxnet_p27/lib/python2.7/site-packages/mxnet/libmxnet.so: no symbols
```

DickJC123:
Actually no, I was examining the symbols of a locally built libmxnet.so.

I'm still trying to understand the recommended approach for the functionality I've introduced.  Any translation unit that includes include/mxnet/base.h and calls Context::Create(dev_type, dev_id) will need external linkage to the Context::CudaLibChecks() implementation from ./src/base.cc.  On the other hand, CudaLibChecks() is private and Create() is its only user.  What is the recommendation then for preceding the name with MX, implementing in c_api.h, etc.?

yuxihu:
@szha @marcoabreu I just tried to add MX prefix to those two functions.
```
root@f905c57ff7f3:/build/mxnet-build# git diff
diff --git a/include/mxnet/base.h b/include/mxnet/base.h
index c1e2da7..e476734 100644
--- a/include/mxnet/base.h
+++ b/include/mxnet/base.h

-    static void CudaLibChecks();
+    static void MXCudaLibChecks();

-    static void CuDNNLibChecks();
+    static void MXCuDNNLibChecks();

-    CudaLibChecks();
+    MXCudaLibChecks();

-    CuDNNLibChecks();
+    MXCuDNNLibChecks();

diff --git a/src/base.cc b/src/base.cc
index 96d66ad..fc79c6a 100644
--- a/src/base.cc
+++ b/src/base.cc

-void Context::CudaLibChecks() {
+void Context::MXCudaLibChecks() {

-void Context::CuDNNLibChecks() {
+void Context::MXCuDNNLibChecks() {
```
However, the two functions are not exported with the existing wildcard pattern. Am I missing anything here? @szha 
```
root@f905c57ff7f3:/build# nm -D mxnet-build/lib/libmxnet.so | grep MX | grep Lib
000000000252a480 T MXLibInfoFeatures
```

DickJC123:
I suspect the problem is that the name must match MX* and either the mxnet namespace of the new functions or the Context class scope prevents a match.

I can reproduce the error locally.  Is horovod going around the official mxnet api by calling Context::create()?

```
nm -C /usr/local/lib/python3.6/dist-packages/horovod/mxnet/mpi_lib.cpython-36m-x86_64-linux-gnu.so | grep -i context | grep -i mxnet
                 U mxnet::Context::CudaLibChecks()
000000000007e510 t mxnet::Context::Create(mxnet::Context::DeviceType, int)
```

yuxihu:
> I can reproduce the error locally. Is horovod going around the official mxnet api by calling Context::create()?

Yes, in Horovod we call Context::CPU() and Context::GPU() which calls Create() internally. If we have a new C API defined for the Context::Create, we have to change to use it in Horovod which will introduce backward compatibility issue. Adding MX to those two functions does not simply work for us. Exporting symbols is a straightforward working solution, although it might not be the best practice as those two functions are private ones.

szha:
libmxnet.so shouldn't expose symbols that are not part of the API. if it's needed as API, it needs to be added to C-API. I think the former is more appropriate in this context (correction is welcome). so the linked issue indicates problem in horovod build rather than mxnet.

DickJC123:
It seems horovod is going around the official mxnet API by refering directly to NDArray and Context. MXNet provides opaque pointers NDArrayHandle and ContextHandle, but I'm not sure the interface is fully functional, since it appears there is no API call to set a ContexHandle, as in (say) MXContextCreate(...,ContextHandle *out). Anyway, @yuxihu , here's a proposal for getting horovod back on the official API, at least in terms of this problem. Would the following change work for you in terms of compatibility issues?

Horovod now:
```
// Returns pointer to newly created NDArray
// If dev_id equal to CPU_DEVICE_ID, construct Tensor on CPU
// Otherwise construct on GPU
NDArray* TensorUtil::New(int device, int dtype) {
  if (device == CPU_DEVICE_ID) {
    NDArray* my_array = new NDArray(TShape(), Context::CPU(0), false, dtype);
    return my_array;
  } else {
    NDArray* my_array =
        new NDArray(TShape(), Context::GPU(device), false, dtype);
    return my_array;
  }
}
```
Horovod as potentially switched over to official MXNet API:
```
NDArray* TensorUtil::New(int device, int dtype) {
   NDArrayHandle my_array;
  if (device == CPU_DEVICE_ID)
    MXNDArrayCreateEx(nullptr, 0, cpu::kDevMask, 0, false, dtype, &my_array);
  else
    MXNDArrayCreateEx(nullptr, 0, gpu::kDevMask, device, false, dtype, &my_array);
  return static_cast<NDArray*>(my_array);
}
```
The compiles for me but has some runtime issues that may be unrelated to the change.  Maybe you could test?

DickJC123:
Regarding compatibility, old horovod builds that still reference Context::Create() would not link against newer libmxnet.so (where Context::Create() needs the unexported CudaLibChecks()).

DickJC123:
The above change was in the file tensor_util.cc.  Also needed is the following chane in mpi_ops.cc:
Was:
```
  static const auto MX_EXEC_CTX = Context::CPU();
```
Needs to be:
```
static const auto MX_EXEC_CTX = Context();
```
The default constructor of Context creates the equivalent CPU context on device 0, without calling the problematic Context::Create().

I have verified that the sequence of commands first posted in https://github.com/apache/incubator-mxnet/issues/15578.  My version of those commands:
```
pip install mxnet-cu101 --pre
HOROVOD_NCCL_INCLUDE=/usr/include HOROVOD_NCCL_LIB=/usr/lib/x86_64-linux-gnu HOROVOD_NCCL_LINK=SHARED HOROVOD_WITHOUT_PYTORCH=1 HOROVOD_WITHOUT_TENSORFLOW=1 HOROVOD_GPU_ALLREDUCE=NCCL pip install --no-cache-dir <my horovod clone dir>
mpirun -np 2 --allow-run-as-root -H localhost:2 pytest /home/dcarter/horovod2/test/test_mxnet.py
```

yuxihu:
> The default constructor of Context creates the equivalent CPU context on device 0, without calling the problematic Context::Create().

@DickJC123 This saves my day. I actually had the same idea of creating NDArray using C API but struggled on this part. I will create a PR in Horovod and verify. We can close this PR if all looks fine. Will keep you updated. Thanks.

yuxihu:
Fixing the issue in Horovod. Close this PR in favor of https://github.com/horovod/horovod/pull/1244.

PR #15606, opened 2019-07-19
[Opperf] Add array rearrange operators to opperf

ChaiBapchya:
## Description ##
Added Rearray operators to the existing set of opperf benchmarks

### Essentials ###

### Changes ###
- [x] Added following ops
1. transpose
2. swapaxes
3. flip
4. depth_to_space
5. space_to_depth

## Results ##
GPU - https://gist.github.com/ChaiBapchya/51b6c7c2346959f3f72aa1b702eb002d
CPU - https://gist.github.com/ChaiBapchya/37b27d4803e73e0fa5a8bd6e8da11492

## Comments ##
@sandeep-krishnamurthy @apeforest


karan6181:
@mxnet-label-bot add [Operator, Performance, pr-awaiting-review]

benchmark/opperf/rules/default_params.py
```diff
@@ -62,6 +62,15 @@
 # NOTE: Data used is DEFAULT_DATA
 DEFAULT_AXIS = [0]
 
+# For rearrange operators
```
sandeep-krishnamurthy: Thanks for this comments. Helps for easy maintenance.

ChaiBapchya:
Ya right. Moreover, functions like
`stop_gradient`, `identity`, `flatten`, `broadcast_plus`, `broadcast_minus` also don't have avg_forward_time. Need to probe. Any idea?

benchmark/opperf/utils/op_registry_utils.py
```diff
@@ -262,6 +264,26 @@ def get_all_sorting_searching_operators():
...
 
 
+def get_all_rearrange_operators():
+    """Gets all array rearrange operators registered with MXNet.
+
+    Returns
+    -------
+    {"operator_name": {"has_backward", "nd_op_handle", "params"}}
+    """
+    rearrange_ops = ['transpose','swapaxes','flip','depth_to_space','space_to_depth']
```
larroy: This should be a set if repeatedly testing for in.

Could we refactor all these get_ functions and make some kind of yaml file where we add tags to the list of operators, I think a declarative thing would be much easier to maintain than all this logic which can be factored out just to select some ops based on the names.

Let me know if what I'm suggesting is not clear.

larroy: Thanks for clarifying this over chat. Seems there's slightly different logics in some of these functions.

ChaiBapchya:
Also, found that `sum_axis` shows `avg_time_forward_sum` instead of `avg_time_forward_sum_axis`

All these operators are aliases and are untracked or wrongly parsed from the profiler output.
Addressing in upcoming commit.

PR #15611, opened 2019-07-19
[MXNET-978] n-th order gradient test support.

kshitij12345:
## Description ##
n-th order gradient test support function.

### Essentials ###

### Changes ###
- [x] Support test for n-th order.


kshitij12345:
@apeforest @larroy 

While working on a PR (I don't exactly remember which one), test for incorrect implementation of second order gradient was passing due to the very small values. However as sanity check, when I checked for its third order gradient, the assertion failed allowing to catch the issue. 

So I think it would be better to check for the order being implemented and one more (computed by autograd) as a sanity check. 

If you think this is in the right direction, will replace `check_second_order_unary` or make it use `check_nth_order_unary`.

Would like to know your thoughts about the same. 

Thank You.

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -148,30 +165,48 @@ def grad_grad_op(x):
...
-    expected_grad_grad = grad_grad_x.asnumpy() * head_grad_grads.asnumpy() * \
-        y_grad.asnumpy()
-
-    # Validate the gradients.
-    assert_almost_equal(expected_grad_grad, x.grad.asnumpy())
+        for current_order in range(1, order+1):
+            head_grad = nd.random.normal(shape=x.shape)
+            y = autograd.grad(heads=y, variables=x, head_grads=head_grad,
+                              create_graph=True, retain_graph=True)[0]
+            if current_order in orders:
```
larroy: If current_order is not in orders we might have problem zipping?  Is there a case where you wou want 1st and 3rd order but not second?

kshitij12345: I don't think there would be an issue in that case (will confirm it later though).
If `current_order` is not in `orders` case is checked here, 
https://github.com/apache/incubator-mxnet/blob/1f74614391e182e299e2fdcce1036515c4e5fb4f/tests/python/unittest/test_higher_order_grad.py#L41-L42
where first order is not asserted (as per the arguments).

The main thing is that elements in `orders` should be monotonically increasing and they should correspond to elements of the `grad_ops`.

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -148,30 +165,48 @@ def grad_grad_op(x):
...
-
-    # Compute expected values.
-    expected_grad_grad = grad_grad_x.asnumpy() * head_grad_grads.asnumpy() * \
-        y_grad.asnumpy()
-
-    # Validate the gradients.
-    assert_almost_equal(expected_grad_grad, x.grad.asnumpy())
+        for current_order in range(1, order+1):
+            head_grad = nd.random.normal(shape=x.shape)
+            y = autograd.grad(heads=y, variables=x, head_grads=head_grad,
```
larroy: better name instead of mutating y?

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -148,30 +165,48 @@ def grad_grad_op(x):
...
+        # TODO(kshitij12345): Remove
+        check_nth_order_unary(array, sigmoid, [grad_op, grad_grad_op], [1, 2])
+        check_nth_order_unary(array, sigmoid, grad_grad_op, 2)
 
 
 def check_second_order_unary(x, op, grad_grad_op):
+    check_nth_order_unary(x, op, grad_grad_op, 2)
+
+
+def check_nth_order_unary(x, op, grad_ops, orders):
```
larroy: add docstring?

tests/python/unittest/test_higher_order_grad.py
kshitij12345: One option I can think of is, at the start of for loop we'll have `computed_grad = y` (which is deceiving) and replace `y` by `computed_grad` in the for loop.
Up for suggestions.

tests/python/unittest/test_higher_order_grad.py
kshitij12345: Sure thing.

tests/python/unittest/test_higher_order_grad.py
kshitij12345: Have confirmed, that 
```
check_nth_order_unary(array, sin, [grad_op, grad_grad_grad_op], [1, 3])
```
following works.

tests/python/unittest/test_higher_order_grad.py
kshitij12345: I have added the docstring. Could you please take a look to see if it needs more polishing.
Thank You.

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -302,28 +322,72 @@ def grad_grad_op(x):
...
+        grad_ops = [grad_ops]
+
     x = nd.array(x)
-    grad_grad_x = grad_grad_op(x)
     x.attach_grad()
 
-    # Manual head_grads.
-    y_grad = nd.random.normal(shape=x.shape)
-    head_grad_grads = nd.random.normal(shape=x.shape)
+    order = max(orders)
```
apeforest: If orders is monotonically increasing, should this just be orders[-1]?

apeforest:
Ping @sxjscience for review.

tests/python/unittest/test_higher_order_grad.py
kshitij12345: Yes. That will work as well. But I felt `max(orders)` stated the intent more clearly.

kshitij12345:
@sxjscience Gentle ping for review:)

PR #15657, opened 2019-07-25
Eliminate common expressions

ptrendx:
## Description ##
This PR introduces a graph pass that eliminates redundant common expressions in the graph. 

For example, let's look at the graph created from a following Python code:
```
a = mx.sym.Variable()
out = (a + 5) * (a + 5)
```
This graph will compute `a+5` twice, which is wasteful. After the pass introduced in this PR, the actual executed graph will be equivalent to
```
a = mx.sym.Variable()
b = a + 5
out = b * b
```
which computes the `a+5` only once.

@eric-haibin-lin FYI

### Essentials ###

### Changes ###


abhinavs95:
@mxnet-label-bot add [Backend, pr-awaiting-review]

include/mxnet/op_attr_types.h
```diff
@@ -208,6 +208,17 @@ using FCreateOpState = std::function<OpStatePtr (const NodeAttrs& attrs,
...
+
+/*!
+ * \brief Whether the operator always produces the same
+ *        output given the same input.
+ *        This enables certain optimizations
+ *        like common expression elimination.
+ *
+ * \note Register under "FHasDeterministicOutput"
+ */
+using FHasDeterministicOutput = bool;
```
eric-haibin-lin: This is a per-op attribute, right? Are we supposed to mark relevant operators with the right value for this attribute in this PR?

src/imperative/cached_op.cc
```diff
@@ -126,6 +126,8 @@ CachedOp::CachedOp(
         fwd_graph_.outputs.push_back(nodeEntry);
       }
     }
+    fwd_graph_ = exec::EliminateCommonExpr(std::move(fwd_graph_));
+
```
eric-haibin-lin: any unit test?

include/mxnet/op_attr_types.h
ptrendx: This is an option for potential optimization - if the op uses state or requests resources then it is not considered for this optimization. This option enables marking such op as safe for deduplicating (an example of this would be Convolution, which asks for resources, but it is only a workspace). I will add this option to such relevant operators.

src/imperative/cached_op.cc
ptrendx: Frankly I'm not sure how to properly test it - any suggestions?

eric-haibin-lin: Maybe we can test after `FHasDeterministicOutput` property is added to some specific operator?

DickJC123:
Another area to spend some extra thought on involves the primary outputs and the gradients of primary inputs.  If you combined two output NDArrays via CSE, would there be any problem with someone wanting to index through the outputs?  Ditto for input grads.  If a CSE-optimized model generated NDArrays for another model (and that model had mutable inputs), would functionality be preserved?  I think a downstream model gets a copy of the NDArray as the input, so it should work, but do we ever want to optimize away that copy?

CSE may eliminate downstream in-place options for output nodes that have been combined.  Not sure if there's a case where using CSE is a net perf loss, but I doubt it.

And finally, here's a fairly obscure case where the operator state may not be contained entirely in the node attributes (we should probably outlaw this usage, since it assumes no caching of the env var lookup):
```
sym = ...
os.environ['MXNET_MY_OP_BEHAVIOR']=X
sym2 = MyOp(data=sym,...)                        # op behavior affected by captured env var setting
os.environ['MNXET_MY_OP_BEHAVIOR']=Y
sym3 = MyOp(data=sym,...)                        # op behavior affected by captured env var setting
```
Perhaps shifting from the node attr dict to the param struct would catch this case.

ptrendx:
@DickJC123 Ok, answering your questions:
1. This definition of Node equality is quite tied to what I want to do with them, so I believe, at least right now, it belongs to CSE code.
2. While I agree it would be beneficial, it would also be pretty hard to do - the parameter structure is kept without the type information as the `shared_ptr<dmlc::any>`. Comparing memory (which could be dine if `any` gave size of the type held) is also not really feasible since `dict` is actually part of the `Parameter` structure, so that would be different.
3. This is the purpose of this `THasDeterministicOutput` property.
4. I will think about that.

DickJC123:
I believe the case of having one output driving multiple symbol outputs is handled strangely, not that this is an important use case.  The issue occurs with an operator driving 3 or more symbol outputs.  Here's the progression of model compute graphs:
```
foo ----+----> out0
        |
        +----> out1
        |
        +----> out2
```
copy operators inserted:
```
foo ----+----> out0
        |
        +-- copy --> out1
        |
        +-- copy --> out2
```
More CSE applied:
```
foo ----+----> out0
        |
        +-- copy --+----> out1
                   |
                   +----> out2
```
copy operators inserted:
```
foo ----+----> out0
        |
        +-- copy --+----> out1
                   |
                   +-- copy --> out2
```
So N outputs driven by the same node end up being driven by N - 1 serialized copies, not N-1 parallel copies.

What is the use-case that requires these copies to be inserted?

DickJC123:
The EliminateCommonExpr pass is performed first-thing, before the backward nodes have been inserted and so before ctx info (i.e. dev_mask) have been added to the graph.  I think this is the correct place- doing CSE with backward nodes present would be a nightmare.

I'm not fond of the HasDeterministicOutput flag- seems like it's just one more thing the user has to understand and set correctly when making a new operator.  I was hoping that perhaps this property could be inferred by the op's registered resources.  Life would be simple if the resources were determined only by the FResourceRequest(attrs) callback.  The problem is there's also the FResourceRequestEx(attrs, dev_mask, dispatch_mode) callback, which takes a dev_mask and dispatch_mode.  As I point out above, CSE is performed before dev_mask is available from the graph.  The only way I can see doing what I'm suggesting would be to infer the possible settings for dev_mask and dispatch_mode from the registration of other callbacks.  Some possible pseudo-code for this:

```
set<resources> all_possible_resources = {}
if (op_has_registered("FResourceRequestEx") {
  auto f_resource_req_ex = < look up op's FResourceRequestEx >
  if (op_has_registered("FCompute<cpu>")
    all_possible_resources += f_resource_req_ex(attrs, kCpu, DispatchMode::kFCompute)
  if (op_has_registered("FComputeEx<cpu>")
    all_possible_resources += f_resource_req_ex(attrs, kCpu, DispatchMode::kFComputeEx)
  if (op_has_registered("FCompute<gpu>")
    all_possible_resources += f_resource_req_ex(attrs, kGpu, DispatchMode::kFCompute)
  if (op_has_registered("FComputeEx<gpu>")
    all_possible_resources += f_resource_req_ex(attrs, kGpu, DispatchMode::kFComputeEx)
} else if op_has_registered("FResourceRequest") {
  auto f_resource_req = < look up op's FResourceRequest >
  all_possible_resources += f_resource_req(attrs)
}
bool inferred_cse_not_possible = LookForBadResources(all_possible_resources)
```

Regarding the actual name 'THasDeterministicOutput', might this be confused with the cuda notion of deterministic algos?  The Convolution op would be tagged with THasDeterministicOutput=1, but would not truly be deterministic if a convolution algo with atomics is used.

PR #15664, opened 2019-07-26
add int8 bn mkldnn implementation and test

ElaineBao:
## Description ##
Add a new operator - int8 batch norm, mkldnn implementation and test
@pengzhao-intel  @ZhennanQin 
### Usage ###
1. Check the doc in https://github.com/apache/incubator-mxnet/tree/master/example/quantization/README.md to quantize models and do inference.
2. In order to use standalone int8 batch norm instead of fused fp32 batch norm, one should `export MXNET_DISABLE_MKLDNN_FUSE_CONV_BN=1` before using `imagenet_gen_qsym_mkldnn.py` to quantize the model.
3. Suggest to use fuse bn if it can be fused, since it's faster and more accurate. Otherwise use standalone int8 bn.

### Limitation ###
1. Currently int8 bn only support s8 input, since mkldnn batch norm only support s8 input
2. Currently int8 bn cannot support `calib_mode = none`, since when calculating the thresholds on the fly with s8 input, errors are large. One can run with `calib_mode=naïve/entropy`, should have a similar accuracy with fp32 model.

### Performance ###
I tested several models on skylake, which can be used for reference.

Models | FP32 Acc |  INT8 (fuse fp32 bn)  Acc | INT8 (standalone int8 bn) Acc
-|-|-|-
Resnet50 V1 | 0.765 | 0.760 | 0.751
Mobilenet1.0 | 0.722 | 0.720 | 0.677
Inception V3 | 0.782 | 0.782 | 0.772


abhinavs95:
@mxnet-label-bot add [mkldnn, Backend, pr-awaiting-review]

xinyu-intel:
@ElaineBao Can you try this on resnetv2? Theoretically, the performance will be better since lots of bn-relu-conv pattern in this model.

pengzhao-intel:
@ZhennanQin @ciyongch, please help take a review.

pengzhao-intel:
@ElaineBao  could you elaborate the reason for standalone BN leads a bit more accuracy dorp?

ElaineBao:
> @ElaineBao Can you try this on resnetv2? Theoretically, the performance will be better since lots of bn-relu-conv pattern in this model.

that's a good advice, I'll try it and update the performance, thank you.

ElaineBao:
> @ElaineBao could you elaborate the reason for standalone BN leads a bit more accuracy dorp?

Basically the accuracy drop is not because the BN is fused or standalone, it's because the BN is converted from fp32 to int8. 
It's acceptable to have a bit accuracy drop if one op is converted from fp32 to int8, but for int8 bn maybe the drop on accuracy is too much, especially for Mobilenet1.0. I guess it has something to do with the distribution of values of parameters in Mobilenet1.0.
I'll look into it.

xinyu-intel:
@ElaineBao unfuse bn will also introduce standalone quanitzed_activation along with quantized_bn.

ElaineBao:
Hi, all, I've looked into the performance issue, and concluded that as a operator, int8 bn  itself has no performance regression. The accuracy drop happened in some models is due to the combination of int8 bn and other operators, which may cause a poor weight distribution. 
One should analyze and decide which situation is suitable for using int8 bn. And I'll give a doc on how to use it later.

pengzhao-intel:
@ZhennanQin please take a review too.
@ElaineBao please update the doc in here and mark int8 bn to Y
https://mxnet.incubator.apache.org/versions/master/tutorials/mkldnn/operator_list.html

src/operator/nn/batch_norm-inl.h
```diff
@@ -81,19 +93,35 @@ struct BatchNormParam : public dmlc::Parameter<BatchNormParam> {
...
     .describe("Output the mean and inverse std ");
     DMLC_DECLARE_FIELD(axis).set_default(mxnet::op::batchnorm::DEFAULT_AXIS)
-      .describe("Specify which shape axis the channel is specified");
+    .describe("Specify which shape axis the channel is specified");
     DMLC_DECLARE_FIELD(cudnn_off).set_default(false)
-      .describe("Do not select CUDNN operator, if available");
+    .describe("Do not select CUDNN operator, if available");
+    DMLC_DECLARE_FIELD(min_calib_range)
+    .set_default(dmlc::optional<float>())
+    .describe("The minimum scalar value in the form of float32 obtained "
```
ZhennanQin: It's better to point out that the calibration is based on bn output, not its input.

src/operator/nn/batch_norm-inl.h
```diff
@@ -81,19 +93,35 @@ struct BatchNormParam : public dmlc::Parameter<BatchNormParam> {
...
-           this->cudnn_off == other.cudnn_off;
+  bool operator==(const BatchNormParam &other) const {
+    bool flag = this->eps == other.eps && this->momentum == other.momentum &&
+                this->fix_gamma == other.fix_gamma &&
+                this->use_global_stats == other.use_global_stats &&
+                this->output_mean_var == other.output_mean_var && this->axis == other.axis &&
+                this->cudnn_off == other.cudnn_off &&
+                this->min_calib_range.has_value() == other.min_calib_range.has_value() &&
+                this->max_calib_range.has_value() == other.max_calib_range.has_value();
+    if (this->min_calib_range.has_value() && other.min_calib_range.has_value() &&
```
ZhennanQin: Not enough. Please check min and max separately, and consider the condition that only one side has value.

src/operator/nn/batch_norm.cc
```diff
@@ -396,7 +396,7 @@ void BatchNormComputeExCPU(const nnvm::NodeAttrs &attrs,
   CHECK_EQ(inputs.size(), 5U);
   const BatchNormParam &param = nnvm::get<BatchNormParam>(attrs.parsed);
   // MKLDNN batchnorm only works well on the special MKLDNN layout.
-  if (SupportMKLDNNBN(inputs[0], param) && inputs[0].IsMKLDNNData()) {
+  if (SupportMKLDNNBN(inputs[0], param) /*&& inputs[0].IsMKLDNNData() */) {
```
ZhennanQin: Not sure if we can remove this. @TaoLv for double check.

src/operator/quantization/mkldnn/mkldnn_quantized_batch_norm.cc
```diff
@@ -0,0 +1,129 @@
...
+  const float min_data = in_data[quantized_batchnorm::kDataMin].data().dptr<float>()[0];
+  const float max_data = in_data[quantized_batchnorm::kDataMax].data().dptr<float>()[0];
+  const float max_abs_data = std::max(std::abs(min_data), std::abs(max_data));
+
+  float *min_output_ptr = outputs[quantized_batchnorm::kOutMin].data().dptr<float>();
+  float *max_output_ptr = outputs[quantized_batchnorm::kOutMax].data().dptr<float>();
+  if (param.min_calib_range.has_value() && param.max_calib_range.has_value()) {
+    *max_output_ptr =
+        std::max(std::abs(param.min_calib_range.value()), std::abs(param.max_calib_range.value()));
+    *min_output_ptr = -*max_output_ptr;
```
ZhennanQin: why not *min_output_ptr = param.min_calib_range.value()?

src/operator/quantization/mkldnn/mkldnn_quantized_batch_norm.cc
```diff
@@ -0,0 +1,129 @@
...
+                  "don't support calib_mode=None";
+  }
+
+  unsigned flags = mkldnn::use_global_stats | mkldnn::use_scale_shift;
+  auto &fwd = GetBNForward<float>(param, ctx, data, flags);
+  const mkldnn::memory &weight_mem = fwd.GetWeight();
+  CHECK_EQ(weight_mem.get_primitive_desc().get_size(), channel_count * sizeof(float) * 2);
+  float *weight_buf = reinterpret_cast<float *>(weight_mem.get_data_handle());
+
+  NDArray gamma = in_data[quantized_batchnorm::kGamma];
```
ZhennanQin: change to `const NDArray& gamma`

src/operator/quantization/mkldnn/mkldnn_quantized_batch_norm.cc
```diff
@@ -0,0 +1,129 @@
...
+  NDArray beta = in_data[quantized_batchnorm::kBeta];
```
ZhennanQin: As above.

src/operator/quantization/mkldnn/mkldnn_quantized_batch_norm.cc
```diff
@@ -0,0 +1,129 @@
...
+  float *gamma_ptr = gamma.data().dptr<float>();
```
ZhennanQin: Seems gamma only used here. Perhaps you can remove it as 
`float *gamma_ptr = in_data[quantized_batchnorm::kGamma].data().dptr<float>();`

src/operator/quantization/mkldnn/mkldnn_quantized_batch_norm.cc
```diff
@@ -0,0 +1,129 @@
...
+  float *beta_ptr = beta.data().dptr<float>();
+
+  NDArray moving_mean = in_data[quantized_batchnorm::kInMovingMean];
+  NDArray moving_var = in_data[quantized_batchnorm::kInMovingVar];
+  float *moving_mean_ptr = moving_mean.data().dptr<float>();
+  float *moving_var_ptr = moving_var.data().dptr<float>();
+
+  // rescale gamma and beta, to make mean=0 and var=1
+  NDArray rescaled_mean = NDArray(moving_mean.storage_type(), moving_mean.shape(),
```
ZhennanQin: Create temp memory from TmpMemMgr

src/operator/quantization/mkldnn/mkldnn_quantized_batch_norm.cc
```diff
@@ -0,0 +1,129 @@
...
+    weight_buf[channel] = gamma_ptr[channel] * invstd * max_abs_data / (*max_output_ptr);
+    weight_buf[channel_count + channel] =
+        (beta_ptr[channel] - moving_mean_ptr[channel] * gamma_ptr[channel] * invstd) * kInt8Range /
+        (*max_output_ptr);
+    rescaled_mean_ptr[channel] = 0.0f;
+    rescaled_var_ptr[channel] = 1.0f;
+  }
+
+  const NDArray &out = outputs[batchnorm::kOut];
+  auto out_mem = const_cast<NDArray &>(out).CreateMKLDNNData(fwd.GetPd().dst_primitive_desc());
```
ZhennanQin: Use `CreateMKLDNNMem` instead. Avoid using `const_cast` unless you have to.

src/operator/quantization/mkldnn/mkldnn_quantized_batch_norm.cc
```diff
@@ -0,0 +1,129 @@
...
+
+inline static bool QuantizedBatchNormStorageType(const nnvm::NodeAttrs &attrs, const int dev_mask,
+                                                 DispatchMode *dispatch_mode,
+                                                 std::vector<int> *in_attrs,
+                                                 std::vector<int> *out_attrs) {
+  bool dispatched = false;
+  if (!dispatched) {
+    dispatched = MKLDNNStorageType(attrs, dev_mask, true, dispatch_mode, in_attrs, out_attrs);
+  }
+  if (!MKLDNNEnvSet()) {
```
ZhennanQin: Not necessary,  `MKLDNNStorageType` will check this.

src/operator/nn/batch_norm-inl.h
ElaineBao: it's a '==' operator, if min_calib_range has value and max_calib_range doesn't, it's still equal

src/operator/quantization/mkldnn/mkldnn_quantized_batch_norm.cc
ElaineBao: For example, if min_calib_range=-5, max_calib_range=10, then *max_output_ptr=10, *min_output_ptr=-10, it;s symmetric.

src/operator/nn/batch_norm-inl.h
ZhennanQin: How about this->min_calib_range has value and other.min_calib_range doesn't? At this situation, this will use offline calibration, while other will do online calibration, they are not equal.

ElaineBao: it's checked in L117 & L118.

src/operator/nn/mkldnn/mkldnn_batch_norm-inl.h
```diff
@@ -175,6 +175,47 @@ class MKLDNNBNForward {
     }
   }
 
+  void SetDataHandle(const NDArray &data, const mkldnn::memory *mean,
```
ZhennanQin: Don't duplicate code. Make old version on top of this one.

pengzhao-intel:
Thanks for the great works :) Merging now.

Issue #12901, opened 2018-10-22
Flaky test test_operator.test_clip


marcoabreu:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/master/1797/pipeline

```
======================================================================

FAIL: test_operator.test_clip

----------------------------------------------------------------------

Traceback (most recent call last):

  File "C:\Anaconda3\envs\py3\lib\site-packages\nose\case.py", line 197, in runTest

    self.test(*self.arg)

  File "C:\jenkins_slave\workspace\ut-python-gpu\tests\python\unittest\common.py", line 172, in test_new

    orig_test(*args, **kwargs)

  File "C:\jenkins_slave\workspace\ut-python-gpu\tests\python\unittest\test_operator.py", line 3627, in test_clip

    [np.where(data_tmp < 0.6, [1], [0]) * np.where(data_tmp > -0.6, [1], [0])])

  File "C:\jenkins_slave\workspace\ut-python-gpu\windows_package\python\mxnet\test_utils.py", line 1121, in check_symbolic_backward

    equal_nan=equal_nan)

  File "C:\jenkins_slave\workspace\ut-python-gpu\windows_package\python\mxnet\test_utils.py", line 491, in assert_almost_equal

    raise AssertionError(msg)

AssertionError: 

Items are not equal:

Error 100000.002526 exceeds tolerance rtol=0.000010, atol=0.000000.  Location of maximum error:(28, 12), a=0.000000, b=1.000000

 EXPECTED_data: array([[1, 1, 0, ..., 0, 0, 0],

       [1, 0, 1, ..., 1, 1, 1],

       [1, 1, 0, ..., 0, 0, 1],...

 BACKWARD_data: array([[ 1.,  1.,  0., ...,  0.,  0.,  0.],

       [ 1.,  0.,  1., ...,  1.,  1.,  1.],

       [ 1.,  1.,  0., ...,  0.,  0.,  1.],...

-------------------- >> begin captured logging << --------------------

common: INFO: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=708169757 to reproduce.

--------------------- >> end captured logging << ---------------------
```

PR #15688, opened 2019-07-29
Fix backward_clip num inputs and type of clip params

ptrendx:
## Description ##
Fix backward of clip to correctly report 2 inputs instead of 1. Changed the type of a_min and a_max used inside the kernel to avoid numerical inconsistencies in fp16 (even though the final comparison was done in fp32 and the limits were originally in fp32, they were casted to and from fp16 entering the function).

### Essentials ###

### DIFF
--- a/src/operator/tensor/matrix_op.cc
+++ b/src/operator/tensor/matrix_op.cc
@@ -702,7 +702,8 @@ Given an interval, values outside the interval are clipped to the interval edges
 Clipping ``x`` between `a_min` and `a_max` would be::
 
 .. math::
-   clip(x, a_min, a_max) = max(min(x, a_max), a_min))
+
+   clip(x, a_min, a_max) = \max(\min(x, a_max), a_min))
 
 Example::
 
--- a/src/operator/tensor/matrix_op.cc
+++ b/src/operator/tensor/matrix_op.cc
@@ -699,8 +699,9 @@ MXNET_ADD_SPARSE_OP_ALIAS(clip)
 .describe(R"code(Clips (limits) the values in an array.
 
 Given an interval, values outside the interval are clipped to the interval edges.
-Clipping ``x`` between `a_min` and `a_x` would be::
+Clipping ``x`` between `a_min` and `a_max` would be::
 
+.. math::
    clip(x, a_min, a_max) = max(min(x, a_max), a_min))
 
 Example::
--- a/src/operator/tensor/matrix_op.cc
+++ b/src/operator/tensor/matrix_op.cc
@@ -763,7 +763,7 @@ parameter values:
 .add_arguments(ClipParam::__FIELDS__());
 
 NNVM_REGISTER_OP(_backward_clip)
-.set_num_inputs(1)
+.set_num_inputs(2)
 .set_num_outputs(1)
 .set_attr_parser(ParamParser<ClipParam>)
 .set_attr<nnvm::TIsBackward>("TIsBackward", true)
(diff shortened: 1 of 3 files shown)


src/operator/tensor/matrix_op.cc
```diff
@@ -763,7 +763,7 @@ parameter values:
 .add_arguments(ClipParam::__FIELDS__());
 
 NNVM_REGISTER_OP(_backward_clip)
-.set_num_inputs(1)
+.set_num_inputs(2)
```
eric-haibin-lin: I'm surprised to see this was not caught previously. Is there any unit test to unvail this bug?

piyushghai:
@ptrendx Can you also add a unit test for the fix you made ? 

@mxnet-label-bot Add [Operator, Backend, pr-awaiting-review]

src/operator/tensor/matrix_op.cc
ptrendx: I noticed it when trying to attach monitor to the executor which had clip - it failed with assertion about number of inputs mismatch.
I am not aware of any other test that could uncover this - I don't think there are any checks for this.

ptrendx:
@piyushghai Sure, I should be able to do it tomorrow.

ptrendx:
Last commit adds the test that uncovered the initial problem and also fixes #12901

src/operator/tensor/matrix_op.cc
DickJC123: I think the bug in this case is the inconsistency between set_num_inputs(1) for the _backward_clip node and the definition of the forward node:
```
NNVM_REGISTER_OP(clip)
...
 .set_num_inputs(1)
 .set_num_outputs(1)
...
.set_attr<nnvm::FGradient>("FGradient", ElemwiseGradUseIn{ "_backward_clip" })
```
Rather than a unittest, I think the framework should detect this problem in the general case.  The following CHECK_EQ added to ./src/operator/operator_common.h flags this problem in test_clip(), but I haven't determined whether the entire CI is compatible with it:
```
inline std::vector<nnvm::NodeEntry>CreateNodeEntries(
  nnvm::NodePtr pNode,
  const std::vector<nnvm::NodeEntry>* pOgrads = nullptr,
  const std::vector<nnvm::NodeEntry>* pInputs = nullptr) {
  if (pOgrads)
    pNode->inputs.insert(pNode->inputs.end(), pOgrads->begin(), pOgrads->end());

  if (pInputs)
    pNode->inputs.insert(pNode->inputs.end(), pInputs->begin(), pInputs->end());

  CHECK_EQ(pNode->inputs.size(), pNode->num_inputs()) <<
           "Mismatch in number of created vs. expected node inputs.";

  std::vector<nnvm::NodeEntry> ret;
  for (uint32_t i = 0; i < pNode->num_outputs(); ++i)
    ret.emplace_back(nnvm::NodeEntry{pNode, i, 0});

  return ret;
}
```

ptrendx: That is a good idea to have a check like this, but probably in a subsequent PR.

ptrendx:
I'm not sure how to not overload the user with information with all those casts there. I made the other changes to docs.

DickJC123:
My thought with the comment about a_min and a_max casting was to describe precisely the behavior of the op.  While my first suggestion might be too detailed, I still think it important to alert the user to the rounding of the limits, in case they have some home-grown bucketing of the clipped outputs.  How about:
```
The actual limits used by clip are the provided a_min and a_max values rounded to the nearest x.dtype.
```

src/operator/tensor/matrix_op.cc
DickJC123: I have submitted the separate PR https://github.com/apache/incubator-mxnet/pull/15834 that includes this check of actual vs. expected node inputs.  Initial finding is that many nodes are similarly non-compliant, like clip(), so best as a separate PR that doesn't slow down this one.

PR #15713, opened 2019-07-31
make TransposeShape infer shape form both sides

dtracz:
## Description ##
Make the function TransposeShape use both input and output data. Base on this it completes all unknown dimension sizes. If any of them is know on both sides the function checks if they match.

### DIFF
--- a/src/operator/tensor/matrix_op-inl.h
+++ b/src/operator/tensor/matrix_op-inl.h
@@ -344,19 +344,40 @@ inline bool TransposeShape(const nnvm::NodeAttrs& attrs,
   CHECK_EQ(in_attrs->size(), 1U);
   CHECK_EQ(out_attrs->size(), 1U);
   mxnet::TShape& shp = (*in_attrs)[0];
+  mxnet::TShape& out_shp = (*out_attrs)[0];
   CHECK_LE(shp.ndim(), 6) << "Transpose support at most 6 dimensions";
-  mxnet::TShape ret(shp.ndim(), -1);
+  CHECK_NE(shp.ndim(), 0) << "Number of dimensions cannot be 0";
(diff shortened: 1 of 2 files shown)


piyushghai:
@dtracz Can you look into the CI failures ? 
@mxnet-label-bot Add [Operator, Backend, pr-awaiting-review]

src/operator/tensor/matrix_op-inl.h
```diff
@@ -344,19 +344,40 @@ inline bool TransposeShape(const nnvm::NodeAttrs& attrs,
...
+    for (int i = 0; i < out_shp.ndim(); ++i) {
+      ifUnknownAssignElseCheck(ret, i, out_shp[i]);
+      ifUnknownAssignElseCheck(get, shp.ndim()-1-i, out_shp[i]);
+    }
   } else {
-    CHECK_EQ(shp.ndim(), param.axes.ndim());
+    CHECK_EQ(std::max(shp.ndim(), out_shp.ndim()), param.axes.ndim());
     for (int i = 0; i < shp.ndim(); ++i) {
       CHECK(param.axes[i] < static_cast<int64_t>(shp.ndim()));
       ret[i] = shp[param.axes[i]];
```
ptrendx: Shouldn't you also do something with `get` here? Otherwise it will be all `-1`s in the next loop I think.

dtracz: To be honest I just missed it, but fact that it was working despite of it realized me, that all 'second assignments' are redundant. The macro SHAPE_ASSIGN_CHECK does it for me at the end.

ptrendx:
@dtracz As discussed offline, please add a small unit test for this fix, then LGTM.

PR #15715, opened 2019-08-01
Improve diagnose.py to display environment variables

larroy:
## Description ##
see title.

### DIFF
--- a/tools/diagnose.py
+++ b/tools/diagnose.py
@@ -30,23 +30,6 @@ except ImportError:
     from urllib2 import urlopen
 import argparse
 
-def parse_args():
-    """Parse arguments."""
-    parser = argparse.ArgumentParser(
-        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
-        description='Diagnose script for checking the current system.')
-    choices = ['python', 'pip', 'mxnet', 'os', 'hardware', 'network']
-    for choice in choices:
-        parser.add_argument('--' + choice, default=1, type=int,
-                            help='Diagnose {}.'.format(choice))
-    parser.add_argument('--region', default='', type=str,
-                        help="Additional sites in which region(s) to test. \
-                        Specify 'cn' for example to test mirror sites in China.")
-    parser.add_argument('--timeout', default=10, type=int,
-                        help="Connection test timeout threshold, 0 to disable.")
-    args = parser.parse_args()
-    return args
-
 URLS = {
     'MXNet': 'https://github.com/apache/incubator-mxnet',
     'Gluon Tutorial(en)': 'http://gluon.mxnet.io',
@@ -55,6 +38,7 @@ URLS = {
     'PYPI': 'https://pypi.python.org/pypi/pip',
     'Conda': 'https://repo.continuum.io/pkgs/free/',
 }
+
 REGIONAL_URLS = {
     'cn': {
         'PYPI(douban)': 'https://pypi.douban.com/',
@@ -81,6 +65,7 @@ def test_connection(name, url, timeout=10):
     load_elapsed = time.time() - start
     print("Timing for {}: {}, DNS: {:.4f} sec, LOAD: {:.4f} sec.".format(name, url, dns_elapsed, load_elapsed))
 
+
 def check_python():
     print('----------Python Info----------')
     print('Version      :', platform.python_version())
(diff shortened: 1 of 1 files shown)


larroy:

tools/diagnose.py
```diff
@@ -171,6 +161,31 @@ def check_network(args):
...
         test_connection(name, url, args.timeout)
 
+
+def check_environment():
+    print('----------Environment----------')
+    for k,v in os.environ.items():
+        print('{}="{}"'.format(k,v))
+
+
+def parse_args():
```
ChaiBapchya: why is this moved below?

larroy: usually you want it next to main, instead of scrolling all the way up.

tools/diagnose.py
```diff
@@ -171,6 +161,31 @@ def check_network(args):
     for name, url in URLS.items():
         test_connection(name, url, args.timeout)
 
+
+def check_environment():
```
sandeep-krishnamurthy: This prints all the environment variables set by an user. Including their system paths, and everything not related to MXNet. Doesn't this cause a concern for user?

larroy: I thought about filtering MXNET variables, but then there's also other variables which can affect behaviour such as OMP variables or LD_PRELOADS I think the user should remove from their paste any information which they consider sensitive. A priory PATH variables are not sensitive, if anything some keys or tokens might be. Do you suggest some action here?

sandeep-krishnamurthy: I was thinking like AWS credentials or some sensitive details can be part of ENV variables.etc. Will be also environment variables. We cannot list everything and depend on user to remove it IMO.
Do we have a registry for all environment variables?

larroy: I agree. Do you suggest listing only mxnet variables or excluding potential AWS credentials?

sandeep-krishnamurthy: We can have only MXNet variables, but, that list soon may be obsolete, because, new environment variables when introduced may not be added to diagnose.py. So I was asking is there registry or automated way to fetch all environment variables in MXNet.

ChaiBapchya: I couldn't see anything specific to MXNet (that's probably coz I haven't set it.) But this as it stands is more to help user get more information about his own environment, and hence a good to have I guess.

When I ran the above code in Python shell, it returns many irrelevant things. From security standpoint, we should be vary of that.

```
f0189808fb65:Downloads bapac$ python
Python 3.6.5 |Anaconda, Inc.| (default, Apr 26 2018, 08:42:37)
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
>>> import os
>>> for k,v in os.environ.items():
...    print('{}="{}"'.format(k,v))
...
TERM_PROGRAM="iTerm.app"
TERM="xterm-256color"
SHELL="/bin/bash"
TMPDIR="/var/folders/pf/6bc4m4fx0w104sbjkcb0l5rwsrsdhk/T/"
Apple_PubSub_Socket_Render="<>"
TERM_PROGRAM_VERSION="3.2.9"
TERM_SESSION_ID="<>"
USER="bapac"
COMMAND_MODE="unix2003"
SSH_AUTH_SOCK="<>"
__CF_USER_TEXT_ENCODING="0x0:0:0"
PATH="/Users/bapac/anaconda3/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
PWD="/Users/bapac/Downloads"
LANG="en_US.UTF-8"
ITERM_PROFILE="Default"
XPC_FLAGS="0x0"
XPC_SERVICE_NAME="0"
HOME="/Users/bapac"
SHLVL="2"
COLORFGBG="7;0"
ITERM_SESSION_ID="<>"
LOGNAME="bapac"
COLORTERM="truecolor"
_="/Users/bapac/anaconda3/bin/python"
OLDPWD="/Users/bapac/incubator-mxnet"
```

larroy: Question is, should we filter out known variable names to contain sensitive information such as AWS tokens or whitelist MXNET variables. Indeed that list can get out of sync but we could grep them from the https://github.com/apache/incubator-mxnet/blob/master/docs/faq/env_var.md for now.

szha: +1 for whitelisting

marcoabreu: I'd prefer whitelisting over blacklisting

larroy: Ok, thanks for the feedback. Seems that whitelisting will address @sandeep-krishnamurthy concern and is in general agreement with the additional feedback.

larroy:

larroy:

PR #15736, opened 2019-08-02
[OpPerf] Take care of 4d param 

ChaiBapchya:
## Description ##
Previous PR #15606 introduced a bug (if condition resulted in ALWAYS using 4D tensor instead of just using it specific to 2 ops)

## Results ##

Both old and updated results
Old buggy implementation found (1,4,2,4) - 77 times
New updated found it 2 times (`space_to_depth` and `depth_to_space`)

https://gist.github.com/ChaiBapchya/b3c4d14d039daa9e72cafd92fcc3255f

### Essentials ###

### Changes ###
Changed the function signature to take operator as a parameter.
Now I can verify if that operator is an op that needs 4D tensor specifically and only then pass 4D data.

## Comments ##
Missed this because when I run benchmark tests they run without failure. 
Just that input is 4d instead of default 2d tensors (of larger size)


piyushghai:
@mxnet-label-bot Add [Performance, pr-awaiting-review]

ChaiBapchya:
@apeforest @sandeep-krishnamurthy

PR #15744, opened 2019-08-03
[Numpy] Numpy behavior random.uniform()

xidulu:
## Description ##
Current `mx.np.random.uniform` does not support parameters (low and high) as `Ndarray`, as the current version is implemented by wrapping `mxnet.ndarray.random.uniform`, which behaves in a different way **when the output shape is given by the user** compared with native Numpy. 
For example: 
```
low = np.zeros((1,4))
high = np.ones((1,4))
shape = (2,4,4)
```
`mxnet.ndarray.random.uniform(low, high, shape)`  would have output tensor of shape 1x4x2x4x4, however, result from `numpy.random.uniform(low, high, shape)` would have shape 2x4x4.

Another major difference is that,  numpy allows parameters to have different shapes as long as they are broadcastable.
For example: 
```
low = np.zeros((1,4))
high = np.ones((4,4))
shape = (2,4,4)
```
This setting would cause `Operator _sample_uniform inferring shapes failed.` with `mxnet.ndarray.random.uniform`.

This pull request implements a uniform distribution generator with numpy behavior and GPU support.

### Essentials ###

### Changes ###
- [x] New backend for random.uniform (np_uniform_op.*) implemented.
- [x] Current np.random.uniform() in python frontend is renamed to __deprecated_uniform().

## Comments ##
- When both low and high are python scalars and `size` is not defined, a scalar tensor would be returned. This is different from native Numpy, which would return python scalar if both parameters are python scalar.
For example:
```
>>>numpy.random.uniform()
0.1234
>>>mxnet.np.random.uniform()
array(0.1234)
```

### DIFF
--- a/python/mxnet/ndarray/numpy/random.py
+++ b/python/mxnet/ndarray/numpy/random.py
@@ -50,53 +50,6 @@ def _random_helper(random, sampler, params, shape, dtype, ctx, out, kwargs):
                      "but got %s." % type(params[0]))
 
 
-def __deprecated_uniform(low=0.0, high=1.0, size=None, **kwargs):
-    """Draw samples from a uniform distribution.
-
-    Samples are uniformly distributed over the half-open interval
-    ``[low, high)`` (includes low, but excludes high).  In other words,
-    any value within the given interval is equally likely to be drawn
-    by `uniform`.
-
-    Parameters
-    ----------
-    low : float, optional
-        Lower boundary of the output interval.  All values generated will be
-        greater than or equal to low.  The default value is 0.
-    high : float
-        Upper boundary of the output interval.  All values generated will be
-        less than high.  The default value is 1.0.
-    size : int or tuple of ints, optional
-        Output shape.  If the given shape is, e.g., ``(m, n, k)``, then
-        ``m * n * k`` samples are drawn.  If size is ``None`` (default),
-        a scalar tensor containing a single value is returned if
-        ``low`` and ``high`` are both scalars.
-    dtype : {'float16', 'float32', 'float64'}, optional
-        Data type of output samples. Default is 'float32'
-    ctx : Context, optional
-        Device context of output. Default is current context.
-    out : ndarray, optional
-        Store output to an existing ndarray.
-
-    Returns
-    -------
-    out : ndarray
-        Drawn samples from the parameterized uniform distribution.
-
-
-    Notes
-    -----
-    This function currently does not support ``low`` and ``high`` as ndarrays.
-    """
-    dtype = kwargs.pop('dtype', None)
-    if dtype is None:
-        dtype = 'float32'
-    ctx = kwargs.pop('ctx', None)
-    out = kwargs.pop('out', None)
-    return _random_helper(_npi.random_uniform, None,
-                          [low, high], size, dtype, ctx, out, kwargs)
-
-
 def normal(loc=0.0, scale=1.0, size=None, **kwargs):
     """Draw random samples from a normal (Gaussian) distribution.
 
@@ -170,8 +123,10 @@ def uniform(low=0.0, high=1.0, size=None, ctx=None, dtype=None, out=None):
     out : ndarray
         Drawn samples from the parameterized uniform distribution.
     """
-    input_type = (isinstance(low, NDArray), isinstance(high, NDArray))
-    dummy_value = -1.0
+    from ...numpy import ndarray as np_ndarray
+    input_type = (isinstance(low, np_ndarray), isinstance(high, np_ndarray))
+    if dtype is None:
+        dtype = 'float32'
     if ctx is None:
         ctx = current_context()
     if out is not None:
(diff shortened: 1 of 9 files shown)


src/operator/numpy/random/np_uniform_op.h
```diff
@@ -0,0 +1,269 @@
...
+    if (in_attrs->size() == 1U) {
+      SHAPE_ASSIGN_CHECK(*out_attrs, 0, in_attrs->at(0))
+    }
+    // Two scalar case without predefined size.
+    if (in_attrs->size() == 0) {
+      SHAPE_ASSIGN_CHECK(*out_attrs, 0, TShape(0, -1))
+      return true;
+    }
+  }
+  return out_attrs->at(0).ndim() != 0U;;
```
yzhliu:
```suggestion
  return out_attrs->at(0).ndim() != 0U;
```

src/operator/numpy/random/np_uniform_op.h
```diff
@@ -0,0 +1,269 @@
...
+    // Size undeclared.
+    if (in_attrs->size() == 2U) {
+      // Low and high both from ndarray.
+      mxnet::TShape& low = (*in_attrs)[0];
+      mxnet::TShape& high = (*in_attrs)[1];
+      mxnet::TShape out(std::max(low.ndim(), high.ndim()), -1);
+      InferBroadcastShape(low, high, &out);
+      SHAPE_ASSIGN_CHECK(*out_attrs, 0, out);
+    }
+    if (in_attrs->size() == 1U) {
```
yzhliu: better to use "else if" so that it won't be evaluated if previous condition matches.

src/operator/numpy/random/np_uniform_op.h
```diff
@@ -0,0 +1,269 @@
...
+            s, outputs[0].Size(), scalar_pos, stride, oshape,
+            inputs[0].dptr<IType>(), scalar_value,
+            uniform_tensor.dptr_, outputs[0].dptr<OType>());
+        });
+      });
+    });
+  }
+
+  // [tensor tensor] case
+  if (inputs.size() == 2U) {
```
yzhliu:
```suggestion
  else if (inputs.size() == 2U) {
```

src/operator/numpy/random/twoparams_dist_common.h
```diff
@@ -0,0 +1,181 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ *  Copyright (c) 2015 by Contributors
+ * \file etwoparams_dist_common.h
```
yzhliu: how about just name it "dist_common.h"

xidulu: Good suggestion! Working on it.

xidulu: > how about just name it "dist_common.h"

Good suggestion, working on it.

src/operator/numpy/random/np_uniform_op.h
xidulu: fixed.

src/operator/numpy/random/np_uniform_op.h
xidulu: > better to use "else if" so that it won't be evaluated if previous condition matches.

fixed, thx

python/mxnet/ndarray/numpy/random.py
```diff
@@ -50,7 +50,7 @@ def _random_helper(random, sampler, params, shape, dtype, ctx, out, kwargs):
                      "but got %s." % type(params[0]))
 
 
-def uniform(low=0.0, high=1.0, size=None, **kwargs):
+def __deprecated_uniform(low=0.0, high=1.0, size=None, **kwargs):
```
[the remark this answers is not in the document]
reminisce: Yes, I agree we should remove the old version.

python/mxnet/ndarray/numpy/random.py
```diff
@@ -139,6 +139,58 @@ def normal(loc=0.0, scale=1.0, size=None, **kwargs):
...
+        Data type of output samples. Default is 'float32'
+    ctx : Context, optional
+        Device context of output. Default is current context.
+
+    Returns
+    -------
+    out : ndarray
+        Drawn samples from the parameterized uniform distribution.
+    """
+    input_type = (isinstance(low, NDArray), isinstance(high, NDArray))
```
reminisce: We should not allow `NDArray` as input type. Only `mxnet.numpy.ndarray` is allowed.

src/operator/numpy/random/np_uniform_op.h
```diff
@@ -0,0 +1,226 @@
...
+#include "../../mshadow_op.h"
+#include "../../mxnet_op.h"
+#include "../../operator_common.h"
+
+namespace mxnet {
+namespace op {
+
+struct NumpyUniformParam : public dmlc::Parameter<NumpyUniformParam> {
+  int t;
+  float low;
```
reminisce: You can define this as `optional<float> low` and use `low.has_value()` to determine which situation it is.

src/operator/numpy/random/np_uniform_op.h
```diff
@@ -0,0 +1,226 @@
...
+  float high;
```
reminisce: Same as above.

src/operator/numpy/random/np_uniform_op.h
```diff
@@ -0,0 +1,226 @@
...
+                  "1: scalar array "
+                  "2: array scalar "
+                  "3: scalar scalar");
+    DMLC_DECLARE_FIELD(size)
+        .set_default(dmlc::optional<mxnet::Tuple<int>>())
+        .describe("Output shape. If the given shape is, "
+                  "e.g., (m, n, k), then m * n * k samples are drawn. "
+                  "Default is None, in which case a single value is returned.");
+    DMLC_DECLARE_FIELD(ctx)
+    .set_default("")
```
reminisce: Default should be cpu.

src/operator/numpy/random/np_uniform_op.h
```diff
@@ -0,0 +1,226 @@
...
+  int t;
+  float low;
+  float high;
+  std::string ctx;
+  int dtype;
+  dmlc::optional<mxnet::Tuple<int>> size;
+  DMLC_DECLARE_PARAMETER(NumpyUniformParam) {
+    DMLC_DECLARE_FIELD(low);
+    DMLC_DECLARE_FIELD(high);
+    DMLC_DECLARE_FIELD(t)
```
reminisce: If using `optional<float>` for `low` and `high`, I think this parameter is not necessary to be defined.

python/mxnet/ndarray/numpy/random.py
xidulu: > We should not allow `NDArray` as input type. Only `mxnet.numpy.ndarray` is allowed.

Fixed, thx

python/mxnet/numpy/random.py
```diff
@@ -23,7 +23,7 @@
 __all__ = ['uniform', 'normal']
 
 
-def uniform(low=0.0, high=1.0, size=None, **kwargs):
+def __deprecated_uniform(low=0.0, high=1.0, size=None, **kwargs):
```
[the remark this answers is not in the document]
xidulu: > Same as above, just delete this.

Removed, thx

PR #15751, opened 2019-08-05
Fix the bug of `MXEnginePushAsyncND` and `MXEnginePushSyncND`

wkcn:
## Description ##
Sorry that I wrote a bug in the two APIs `MXEnginePushAsyncND` and `MXEnginePushSyncND`, whose argument should be an array pointer of NDArrayHandle.

Issue: #15774 
master branch: #15751 
v1.5.x branch: #15775 

### Essentials ###

### Changes ###
- [x] Modify the type of arguments to 'NDArry**'
- [x] Update the unittest

## Comments ##

### DIFF
--- a/src/c_api/c_api.cc
+++ b/src/c_api/c_api.cc
@@ -1544,11 +1544,11 @@ int MXEnginePushSync(EngineSyncFunc sync_func, void* func_param,
 }
 
 int MXEnginePushAsyncND(EngineAsyncFunc async_func, void* func_param,
-                      EngineFuncParamDeleter deleter, ContextHandle ctx_handle,
-                      NDArrayHandle* const_nds_handle, int num_const_nds,
-                      NDArrayHandle* mutable_nds_handle, int num_mutable_nds,
-                      EngineFnPropertyHandle prop_handle, int priority,
-                      const char* opr_name, bool wait) {
+                        EngineFuncParamDeleter deleter, ContextHandle ctx_handle,
+                        NDArrayHandle* const_nds_handle, int num_const_nds,
+                        NDArrayHandle* mutable_nds_handle, int num_mutable_nds,
+                        EngineFnPropertyHandle prop_handle, int priority,
+                        const char* opr_name, bool wait) {
   API_BEGIN();
   NDArray** const_nds = reinterpret_cast<NDArray**>(const_nds_handle);
   NDArray** mutable_nds = reinterpret_cast<NDArray**>(mutable_nds_handle);
@@ -1564,11 +1564,11 @@ int MXEnginePushAsyncND(EngineAsyncFunc async_func, void* func_param,
 }
 
 int MXEnginePushSyncND(EngineSyncFunc sync_func, void* func_param,
-                     EngineFuncParamDeleter deleter, ContextHandle ctx_handle,
-                     NDArrayHandle* const_nds_handle, int num_const_nds,
-                     NDArrayHandle* mutable_nds_handle, int num_mutable_nds,
-                     EngineFnPropertyHandle prop_handle, int priority,
-                     const char* opr_name) {
+                       EngineFuncParamDeleter deleter, ContextHandle ctx_handle,
+                       NDArrayHandle* const_nds_handle, int num_const_nds,
+                       NDArrayHandle* mutable_nds_handle, int num_mutable_nds,
+                       EngineFnPropertyHandle prop_handle, int priority,
+                       const char* opr_name) {
   API_BEGIN();
   NDArray** const_nds = reinterpret_cast<NDArray**>(const_nds_handle);
   NDArray** mutable_nds = reinterpret_cast<NDArray**>(mutable_nds_handle);
--- a/src/c_api/c_api.cc
+++ b/src/c_api/c_api.cc
@@ -1545,17 +1545,17 @@ int MXEnginePushSync(EngineSyncFunc sync_func, void* func_param,
 
 int MXEnginePushAsyncND(EngineAsyncFunc async_func, void* func_param,
                       EngineFuncParamDeleter deleter, ContextHandle ctx_handle,
-                      NDArrayHandle const_nds_handle, int num_const_nds,
-                      NDArrayHandle mutable_nds_handle, int num_mutable_nds,
+                      NDArrayHandle* const_nds_handle, int num_const_nds,
+                      NDArrayHandle* mutable_nds_handle, int num_mutable_nds,
                       EngineFnPropertyHandle prop_handle, int priority,
                       const char* opr_name, bool wait) {
   API_BEGIN();
-  NDArray* const_nds = static_cast<NDArray*>(const_nds_handle);
-  NDArray* mutable_nds = static_cast<NDArray*>(mutable_nds_handle);
+  NDArray** const_nds = reinterpret_cast<NDArray**>(const_nds_handle);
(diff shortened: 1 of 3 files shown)


src/c_api/c_api.cc
```diff
@@ -1544,18 +1544,18 @@ int MXEnginePushSync(EngineSyncFunc sync_func, void* func_param,
...
-  API_BEGIN();
-  NDArray* const_nds = static_cast<NDArray*>(const_nds_handle);
-  NDArray* mutable_nds = static_cast<NDArray*>(mutable_nds_handle);
+                        EngineFuncParamDeleter deleter, ContextHandle ctx_handle,
+                        NDArrayHandle* const_nds_handle, int num_const_nds,
+                        NDArrayHandle* mutable_nds_handle, int num_mutable_nds,
+                        EngineFnPropertyHandle prop_handle, int priority,
+                        const char* opr_name, bool wait) {
+  API_BEGIN();
+  NDArray** const_nds = reinterpret_cast<NDArray**>(const_nds_handle);
```
wkcn: `reinterpret_cast` is necessary for the cast from `void**` to `NDArray**`.

sxjscience:
Why did the code pass the previous tests?

wkcn:
@sxjscience The previous test is also wrong. In the previous test, I passed the memory address of a NDArray into these two APIs. However, the correct method is to pass the memory address of an array of NDarray pointers.

wkcn:
@sxjscience @anirudh2290 @eric-haibin-lin 
Hi, could you please help take a review? Thank you!

KellenSunderland:
Should we backport this to the v1.5.x branch as well?

wkcn:
@KellenSunderland Yes.

src/c_api/c_api.cc
```diff
@@ -1564,18 +1564,18 @@ int MXEnginePushAsyncND(EngineAsyncFunc async_func, void* func_param,
...
-                     EngineFuncParamDeleter deleter, ContextHandle ctx_handle,
-                     NDArrayHandle const_nds_handle, int num_const_nds,
-                     NDArrayHandle mutable_nds_handle, int num_mutable_nds,
-                     EngineFnPropertyHandle prop_handle, int priority,
-                     const char* opr_name) {
-  API_BEGIN();
-  NDArray* const_nds = static_cast<NDArray*>(const_nds_handle);
-  NDArray* mutable_nds = static_cast<NDArray*>(mutable_nds_handle);
+                       EngineFuncParamDeleter deleter, ContextHandle ctx_handle,
+                       NDArrayHandle* const_nds_handle, int num_const_nds,
```
KellenSunderland: This may count as a breaking change of the public API after a release.  If the intent is to have many NDArrays passed to this function would one option be to concat them together and submit them as one large NDArray?  Would you be able to give some detail about what pieces integrate with this API?  Is it possible we can make a change on that code?

tests/cpp/engine/threaded_engine_test.cc
```diff
@@ -258,47 +258,50 @@ TEST(Engine, PushFunc) {
 TEST(Engine, PushFuncND) {
   auto ctx = mxnet::Context{};
   mxnet::NDArray nd(ctx);
+  std::vector<mxnet::NDArray*> nds;
+  nds.push_back(&nd);
+  void** pnds = reinterpret_cast<void**>(nds.data());
```
KellenSunderland: We might want some tests for cases with more than 1 NDArray present in the argument call.

wkcn: Hi @KellenSunderland, I have added the testcase with more than 1 NDArray : )

KellenSunderland:
LGTM.  After considering the API a little I think I wouldn't' consider this a breaking change.  Consumers of this package would have gotten errors when trying to pass lists of NDArrays to the function.  It's clear that this was the intent of the API.  Instead I'd consider this a patch/bugfix.

sxjscience:
I think it looks good. But where is this API used? Is it used in some other packages?

tests/cpp/engine/threaded_engine_test.cc
```diff
@@ -257,49 +257,80 @@ TEST(Engine, PushFunc) {
...
-  // Test #7
-  LOG(INFO) << "===== Test #7: PushSyncND invalid number of const nds =====";
-  res = MXEnginePushSyncND(FooSyncFunc, nullptr, nullptr, &ctx, &nd, -1, nullptr, 0);
-  EXPECT_EQ(res, -1);
-
-  // Test #8
-  LOG(INFO) << "===== Test #8: PushSyncND invalid number of mutable nds =====";
-  res = MXEnginePushSyncND(FooSyncFunc, nullptr, nullptr, &ctx, nullptr, 0, &nd, -1);
-  EXPECT_EQ(res, -1);
+  std::vector<mxnet::NDArray*> nds;
```
sxjscience: In fact, I'm thinking if it's more appropriate to use
```c++
std::vector<mxnet::NDArray> nds;
```
Do we have to use `std::vector<mxnet::NDArray*>`?

wkcn: We have to use `std::vector<mxnet::NDArray*>`, because the type of the argument of the two APIs is an array pointer of `NDArray*`. Besides, it avoids the potential copy of NDArray.

wkcn: There is an argument Context in the constructor of NDArray. I do not know how to use `std::vector<mxnet::NDArray>`.

sxjscience: Yes, we have to use `std::vector<NDArray*>` if we keep the interface to be `NDArrayHandle *`. However, I'm thinking whether we could directly use the `std::vector<mxnet::NDArray> vec;` and use `nds.emplace_back(mxnet::NDArray(ctx))` or `nds.push_back(std::move(temp_arr))` to fill the vector. In that case, the existing API will not be changed.

wkcn: It sounds good, but I am worry that the type of arguments (`const_vars_handle` and `mutable_vars_handle`) of the two APIs [`MXEnginePushAsync`](https://github.com/apache/incubator-mxnet/blob/master/src/c_api/c_api.cc#L1466) and [`MXEnginePushSync`](https://github.com/apache/incubator-mxnet/blob/master/src/c_api/c_api.cc#L1507) is `EngineVarHandle`, namely `void*`. It casts `void*` to `VarHandle*`, namely `Var**` in https://github.com/apache/incubator-mxnet/blob/master/src/c_api/c_api.cc#L1475. Therefore, I don't know how to decide the type of `const_nds_handle` and `mutable_nds_handle` in `MXEnginePushAsyncND` and `MXEnginePushSyncND`.

wkcn: To keep the consistency, `const_var_handle` and `mutable_var_handle` is the pointer of an array of `VarHandle`, and `const_nds_handle` and `mutable_var_handle` is the pointer of an array of `NDArrayHandle`.

sxjscience: I now understand the logic here. To make the API consistent, I think we should also change the interface of `MXEnginePushAsync` and `MXEnginePushSync`. We should be safe to replace `EngineVarHandle` with `VarHandle*`. Am I right here ? @apeforest @yuxihu

PR #15755, opened 2019-08-05
Revert "Dynamic Library Loading Support"

marcoabreu:
Reverts apache/incubator-mxnet#15489

Reason is the fact that this change poses a security risk: https://github.com/apache/incubator-mxnet/pull/15489/files#diff-e73d371c303c4adefa9914b6c5a6db8cR36 

It allows arbitrary code execution on any computer that executes unit tests. This means it's potentially harmful for our CI system as well as our users. 

I will merge this PR at 7pm GMT on 8/6/2019.

PR #15760, opened 2019-08-06
Fix PR #15489 (Dynamic Library Loading Support)

mseth10:
## Description ##
This PR builds on top of https://github.com/apache/incubator-mxnet/pull/15489, which was reverted by https://github.com/apache/incubator-mxnet/pull/15755.

It includes the following additional changes:
- Changes to Makefile and CMake flows to build sample library as part of MXNet build
- Modifications to CI to stash/un-stash the library binary from build stage to test stage
- Loads the binary as part of unit test

### Essentials ###

### Changes ###

## Comments ##


tests/python/unittest/test_library_loading.py
```diff
@@ -33,6 +33,7 @@ def test_library_loading():
     elif (os.name=='nt'):
         lib = 'mylib.dll'
 
-    fname = mx.test_utils.download('https://mxnet-demo-models.s3.amazonaws.com/lib_binary/'+lib)
-    fname = os.path.abspath(fname)
+    # fname = mx.test_utils.download('https://mxnet-demo-models.s3.amazonaws.com/lib_binary/'+lib)
+    # fname = os.path.abspath(fname)
+    fname = 'Somewhere/In/The/CI'
```
mseth10: @marcoabreu what should we change this to to get the path to the library libsample_lib.so/.dll?

marcoabreu: Check the environment variables. I think there's one called "WORKSPACE" which points to the folder that is allocated for that job

example/lib_api/Makefile
```diff
@@ -0,0 +1,31 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+all:
+	g++ -shared -fPIC mylib.cc -o mylib.so -I ../../include/mxnet
```
larroy: If CI doesn't pass, I suggest to use $(CC) and $(CXX), if not it's not worth another CI round.

samskalicky: @larroy this isnt building in the CI. this is just an example

src/common/library.h
```diff
@@ -0,0 +1,57 @@
...
+/*!
+ * \brief a templated function that fetches from the library
+ * a function pointer of any given datatype and name
+ * \param T a template parameter for data type of function pointer
+ * \param lib library handle
+ * \param func_name function name to search for in the library
+ * \return func a function pointer
+ */
+template<typename T>
+T get_func(void *lib, char *func_name) {
```
larroy: We don't want to get funky inside a mxnet namespace?

src/engine/threaded_engine.h
```diff
@@ -331,6 +332,10 @@ class ThreadedEngine : public Engine {
       kill_.store(true);
     }
     finished_cv_.notify_all();
+    // close opened libraries
+    for (auto const& lib : loaded_libs) {
+      close_lib(lib.second);
```
larroy: why this in the engine? I suggest to do global cleanup in initialize.cc.

related: https://github.com/apache/incubator-mxnet/pull/15762

example/lib_api/Makefile
marcoabreu: Well but it should be - how else it going to be tested?

larroy: I understand, but still this is a PITA if a Makefile doesn't respect the compiler from the environment. That's why I said is not worth another CI round. If you make other changes I would change it. Or is there a reason not to?

samskalicky: How does that work for windows? will it automatically have CXX be "cl" ?

src/engine/threaded_engine.h
mseth10: addressed.

mseth10: addressed

Makefile
```diff
@@ -576,7 +576,7 @@ lib/libmxnet.a: $(ALLX_DEP)
 	@mkdir -p $(@D)
 	ar crv $@ $(filter %.o, $?)
 
-lib/libmxnet.so: $(ALLX_DEP)
+lib/libmxnet.so: $(ALLX_DEP) sample_lib
```
zachgk: If it is not actually a dependency of libmxnet.so, we shouldn't put it here. Can we move it to .PHONY and/or all target instead?

mseth10: Addressed. Moved to all target.

src/common/library.h
```diff
@@ -0,0 +1,57 @@
...
+#ifndef MXNET_COMMON_LIBRARY_H_
+#define MXNET_COMMON_LIBRARY_H_
+
+#include <iostream>
+#include <map>
+#include <string>
+#include "dmlc/io.h"
+
+// map of libraries loaded
+static std::map<std::string, void*> loaded_libs;
```
larroy: This should not be static in a header. I will change it in my PR. This creates a symbol in every compilation unit.

Makefile
```diff
@@ -659,6 +659,9 @@ cpplint:
 pylint:
 	python3 -m pylint --rcfile=$(ROOTDIR)/ci/other/pylintrc --ignore-patterns=".*\.so$$,.*\.dll$$,.*\.dylib$$" python/mxnet tools/caffe_converter/*.py
 
+sample_lib:
+	$(CXX) -shared -fPIC example/lib_api/mylib.cc -o libsample_lib.so -I include/mxnet
```
TaoLv: `libsample_lib.so` cannot be cleaned via `make clean`. Could you please move it to the lib folder?

samskalicky: I dont think we want to move it to the lib folder, since this library is not used for linking or typical LD_LIBRARY_PATH stuff. What if we just add a rm line to the clean target to delete this? We can do it as part of the custom op PR #15921

TaoLv: Thank you for the explanation, @samskalicky . I'm just wondering if we should put a `.so` file in the root folder and I feel difficult to understand the usage of the `.so` from its name.

samskalicky: @TaoLv this has been fixed in the custom Op PR from @rondogency by adding entries to the clean target:
https://github.com/apache/incubator-mxnet/pull/15921/files#diff-b67911656ef5d18c4ae36cb6741b7965R763

TaoLv: Thank you for fixing that.

PR #15794, opened 2019-08-08
Add power, exponent, log ops large tensor support

ChaiBapchya:
## Description ##
Added large tensor support to follow ops
Exponent & Log - exp,expm1, log, log2, log10, log1p
Power - sqrt, rsqrt, cbrt, rcbrt, square, reciprocal

### Essentials ###


tests/nightly/test_large_array.py
```diff
@@ -351,6 +351,69 @@ def test_topk():
     l = nd.topk(b, k=1, axis=-1, dtype=np.int64, ret_typ="value")
     assert l.sum() == np.sum(np.arange(0, SMALL_Y))
 
+def test_exponent_logarithm_operators():
+    a = 2*nd.ones(shape=(LARGE_X, SMALL_Y))
```
apeforest: reuse create_2d_tensor?

tests/nightly/test_large_array.py
```diff
@@ -351,6 +351,69 @@ def test_topk():
...
+    assert result[0][-1] == 1.0986123
+    assert result.shape == a.shape
+
+    # log
+    result = nd.log(a)
+    assert result[0][-1] == 0.6931472
+    assert result.shape == a.shape
+
+def test_power_operators():
+    a = 2*nd.ones(shape=(LARGE_X, SMALL_Y))
```
apeforest: use create_2d_tensor?

tests/nightly/test_large_array.py
ChaiBapchya: a. create_2d_tensor uses 114G (htop reading) for the creating an nd array while same MXNet nd would do it around 40G.
b. np.arange is not really necessary. All we need to do is test if the function works for large arrays.
What do you think?

apeforest: maybe we should change create_2d_array to make it more efficient.

ChaiBapchya: yes. but ill address that in separate PR if that's fine.

access2rohit: Yeah lets remove unnecessary numpy APIs from this test file if they eat up too much memory

PR #15795, opened 2019-08-08
[Numpy] Differentiable svd

hzfan:
## Description ##
Differentiable [svd](https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.svd.html)

### Essentials ###

### Changes ###
- add np.linalg.svd
- add forward and backward tests

## Comments ##
Thank @reminisce and @haojin2 for review and guidance.

### DIFF
--- a/src/operator/c_lapack_api.h
+++ b/src/operator/c_lapack_api.h
@@ -120,6 +120,13 @@ extern "C" {
   MXNET_LAPACK_FSIG_SYEVD(ssyevd, float)
   MXNET_LAPACK_FSIG_SYEVD(dsyevd, double)
 
+  #define MXNET_LAPACK_FSIG_GESVD(func, dtype) \
+    void func##_(char *jobu, char *jobvt, int *m, int *n, dtype *a, int *lda, dtype *s, \
+                 dtype* u, int *ldu, dtype *vt, int *ldvt, dtype *work, int *lwork, int *info);
+
+  MXNET_LAPACK_FSIG_GESVD(sgesvd, float)
+  MXNET_LAPACK_FSIG_GESVD(dgesvd, double)
+
   #ifdef __ANDROID__
     #define MXNET_LAPACK_FSIG_GETRF(func, dtype) \
       int func##_(int *m, int *n, dtype *a, int *lda, int *ipiv, int *info);
@@ -242,6 +249,20 @@ inline void flip(int m, int n, DType *b, int ldb, DType *a, int lda) {
   #define MXNET_LAPACK_sgetrf LAPACKE_sgetrf
   #define MXNET_LAPACK_dgetrf LAPACKE_dgetrf
 
+  #define MXNET_LAPACK_CWRAP_GESVD(prefix, dtype) \
+  inline int MXNET_LAPACK_##prefix##gesvd(int matrix_layout, int m, int n, dtype* ut, \
+                                          int ldut, dtype* s, dtype* v, int ldv, \
+                                          dtype* work, int lwork) { \
+    if (lwork != -1) { \
+      return LAPACKE_##prefix##gesvd(matrix_layout, 'S', 'O', m, n, v, ldv, s, ut, ldut, \
+                                     v, ldv, work); \
+    } \
+    *work = m - 1; \
+    return 0; \
+  }
+  MXNET_LAPACK_CWRAP_GESVD(s, float)
+  MXNET_LAPACK_CWRAP_GESVD(d, double)
+
   #define MXNET_LAPACK_CWRAP_GETRI(prefix, dtype) \
   inline int MXNET_LAPACK_##prefix##getri(int matrix_layout, int n, dtype *a, int lda, \
                                           int *ipiv, dtype *work, int lwork) { \
@@ -361,6 +382,26 @@ inline void flip(int m, int n, DType *b, int ldb, DType *a, int lda) {
   MXNET_LAPACK_CWRAP_SYEVD(ssyevd, float)
   MXNET_LAPACK_CWRAP_SYEVD(dsyevd, double)
 
+  #define MXNET_LAPACK_CWRAP_GESVD(func, dtype) \
+  inline int MXNET_LAPACK_##func(int matrix_layout, int m, int n, dtype* ut, \
+                                 int ldut, dtype* s, dtype* v, int ldv, \
+                                 dtype* work, int lwork) { \
+    if (matrix_layout == MXNET_LAPACK_ROW_MAJOR) { \
+      int info(0); \
+      char jobu('O'); \
+      char jobvt('S'); \
+      func##_(&jobu, &jobvt, &n, &m, v, &ldv, s, v, &ldv, ut, &ldut, work, &lwork, &info); \
+      return info; \
+    } else { \
+      CHECK(false) << "MXNET_LAPACK_" << #func << " implemented for row-major layout only"; \
+      return 1; \
+    } \
+  }
+  MXNET_LAPACK_CWRAP_GESVD(sgesvd, float)
+  MXNET_LAPACK_CWRAP_GESVD(dgesvd, double)
+
+  #define MXNET_LAPACK
+
   // Note: Both MXNET_LAPACK_*getrf, MXNET_LAPACK_*getri can only be called with col-major format
   // (MXNet) for performance.
   #define MXNET_LAPACK_CWRAP_GETRF(prefix, dtype) \
@@ -421,6 +462,12 @@ inline void flip(int m, int n, DType *b, int ldb, DType *a, int lda) {
   int MXNET_LAPACK_##func(int matrix_layout, int n, dtype *a, int lda, \
                           int *ipiv, dtype *work, int lwork);
 
+  #define MXNET_LAPACK_CWRAPPER6(func, dtype) \
+  int MXNET_LAPACK_##func(int matrix_layout, int m, int n, dtype* ut, \
+                          int ldut, dtype* s, dtype* v, int ldv, \
+                          dtype* work, int lwork);
+
+
   #define MXNET_LAPACK_UNAVAILABLE(func) \
   int mxnet_lapack_##func(...);
   MXNET_LAPACK_CWRAPPER1(spotrf, float)
@@ -444,6 +491,10 @@ inline void flip(int m, int n, DType *b, int ldb, DType *a, int lda) {
 
   MXNET_LAPACK_CWRAPPER5(sgetri, float)
   MXNET_LAPACK_CWRAPPER5(dgetri, double)
+
+  MXNET_LAPACK_CWRAPPER6(sgesvd, float)
+  MXNET_LAPACK_CWRAPPER6(dgesvd, double)
+
   #undef MXNET_LAPACK_CWRAPPER1
   #undef MXNET_LAPACK_CWRAPPER2
   #undef MXNET_LAPACK_CWRAPPER3
--- a/src/operator/c_lapack_api.h
+++ b/src/operator/c_lapack_api.h
@@ -249,6 +249,10 @@ inline void flip(int m, int n, DType *b, int ldb, DType *a, int lda) {
   #define MXNET_LAPACK_sgetrf LAPACKE_sgetrf
   #define MXNET_LAPACK_dgetrf LAPACKE_dgetrf
 
+  // Internally A is factorized as U * L * VT, and (according to the tech report)
+  // we want to factorize it as UT * L * V, so we pass ut as u and v as vt.
+  // We also have to allocate at least m - 1 DType elements as workspace as the internal
+  // LAPACKE function needs it to store `superb`. (see MKL documentation)
   #define MXNET_LAPACK_CWRAP_GESVD(prefix, dtype) \
   inline int MXNET_LAPACK_##prefix##gesvd(int matrix_layout, int m, int n, dtype* ut, \
                                           int ldut, dtype* s, dtype* v, int ldv, \
@@ -382,6 +386,8 @@ inline void flip(int m, int n, DType *b, int ldb, DType *a, int lda) {
   MXNET_LAPACK_CWRAP_SYEVD(ssyevd, float)
   MXNET_LAPACK_CWRAP_SYEVD(dsyevd, double)
 
+  // Note: Supports row-major format only. Internally, column-major is used, so all
+  // inputs/outputs are flipped and transposed. m and n are flipped as well.
   #define MXNET_LAPACK_CWRAP_GESVD(func, dtype) \
   inline int MXNET_LAPACK_##func(int matrix_layout, int m, int n, dtype* ut, \
                                  int ldut, dtype* s, dtype* v, int ldv, \
--- a/src/operator/linalg.h
+++ b/src/operator/linalg.h
@@ -195,8 +195,7 @@ int linalg_syevd_workspace_query(const Tensor<xpu, 2, DType>& A,
 
 // CPU/GPU-versions of LAPACK function "gesvd". Please refer to the
 // LAPACK documentation for further details.
-// Note:
-// - V is input and output parameter (overwritten by A)
+// Note: V is input and output parameter (it overwrites A)
 
 template<typename xpu, typename DType>
 void linalg_gesvd(const Tensor<xpu, 2, DType>& UT,
--- a/src/operator/linalg.h
+++ b/src/operator/linalg.h
@@ -191,6 +191,28 @@ int linalg_syevd_workspace_query(const Tensor<xpu, 2, DType>& A,
                                  const Tensor<xpu, 1, DType>& L,
                                  Stream<xpu> *s = 0);
 
+//////////////////////////////// GESVD ////////////////////////////////////////////
+
+// CPU/GPU-versions of LAPACK function "gesvd". Please refer to the
+// LAPACK documentation for further details.
+// Note:
+// - V is input and output parameter (overwritten by A)
+
+template<typename xpu, typename DType>
+void linalg_gesvd(const Tensor<xpu, 2, DType>& UT,
+                  const Tensor<xpu, 1, DType>& L,
+                  const Tensor<xpu, 2, DType>& V,
+                  const Tensor<xpu, 1, DType>& work,
+                  Stream<xpu>* s = 0);
+
+// This function determines the amount of workspace needed for linalg_gesvd
+// which is returned as number of elements of type DType.
+template<typename xpu, typename DType>
+int linalg_gesvd_workspace_query(const Tensor<xpu, 2, DType>& UT,
+                                 const Tensor<xpu, 1, DType>& L,
+                                 const Tensor<xpu, 2, DType>& V,
+                                 Stream<xpu>* s = 0);
+
 //////////////////////////////// GETRF ////////////////////////////////////////////
 
 // CPU/GPU-versions of LAPACK function "getrf". Please refer to the
--- a/src/operator/linalg_impl.h
+++ b/src/operator/linalg_impl.h
@@ -1234,6 +1234,137 @@ LINALG_GPU_SYEVD_WORKSPACE_QUERY(DnDsyevd, double)
 
 #endif  // __CUDACC__
 
+//////////////////////////////// GESVD ////////////////////////////////////////////
+
+// CPU/GPU-versions of LAPACK function "gesvd"
+
+template<typename xpu, typename DType> inline
+void check_gesvd(const Tensor<xpu, 2, DType>& UT,
+                 const Tensor<xpu, 1, DType>& L,
+                 const Tensor<xpu, 2, DType>& V) {
+  // Any checking that helps user debug potential problems.
+  CHECK_LE(V.size(0), V.size(1))
+    << "The second to last dimension of A must be less or equal to the "
+    << "last dimension";
+  CHECK_EQ(UT.size(0), UT.size(1))
+    << "UT must be square matrix";
+  CHECK_EQ(V.size(0), L.size(0))
+    << "V, L have incompatible sizes";
+  CHECK_EQ(V.size(0), UT.size(0))
+    << "V, UT must have compatible sizes";
+}
+
+#define LINALG_CPU_GESVD(fname, DType) \
+template<> inline \
+void linalg_gesvd<cpu, DType>(const Tensor<cpu, 2, DType>& UT, \
+                              const Tensor<cpu, 1, DType>& L, \
+                              const Tensor<cpu, 2, DType>& V, \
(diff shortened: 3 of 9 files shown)


src/operator/c_lapack_api.h
```diff
@@ -242,6 +249,20 @@ inline void flip(int m, int n, DType *b, int ldb, DType *a, int lda) {
   #define MXNET_LAPACK_sgetrf LAPACKE_sgetrf
   #define MXNET_LAPACK_dgetrf LAPACKE_dgetrf
 
+  #define MXNET_LAPACK_CWRAP_GESVD(prefix, dtype) \
+  inline int MXNET_LAPACK_##prefix##gesvd(int matrix_layout, int m, int n, dtype* ut, \
```
mseeger: Not sure I understand this case. Maybe add a comment?

src/operator/c_lapack_api.h
```diff
@@ -361,6 +382,26 @@ inline void flip(int m, int n, DType *b, int ldb, DType *a, int lda) {
   MXNET_LAPACK_CWRAP_SYEVD(ssyevd, float)
   MXNET_LAPACK_CWRAP_SYEVD(dsyevd, double)
 
+  #define MXNET_LAPACK_CWRAP_GESVD(func, dtype) \
```
mseeger: Add comment that due to row-major and internal column-major, the arguments are flipped and transposed, and m and n are flipped as well

src/operator/linalg.h
```diff
@@ -191,6 +191,28 @@ int linalg_syevd_workspace_query(const Tensor<xpu, 2, DType>& A,
                                  const Tensor<xpu, 1, DType>& L,
                                  Stream<xpu> *s = 0);
 
+//////////////////////////////// GESVD ////////////////////////////////////////////
+
+// CPU/GPU-versions of LAPACK function "gesvd". Please refer to the
+// LAPACK documentation for further details.
+// Note:
+// - V is input and output parameter (overwritten by A)
```
mseeger: V is input and output parameter (it overwrites A)

src/operator/linalg_impl.h
```diff
@@ -1234,6 +1234,137 @@ LINALG_GPU_SYEVD_WORKSPACE_QUERY(DnDsyevd, double)
...
+#define LINALG_CPU_GESVD(fname, DType) \
+template<> inline \
+void linalg_gesvd<cpu, DType>(const Tensor<cpu, 2, DType>& UT, \
+                              const Tensor<cpu, 1, DType>& L, \
+                              const Tensor<cpu, 2, DType>& V, \
+                              const Tensor<cpu, 1, DType>& work, \
+                              Stream<cpu> *s) { \
+  check_gesvd(UT, L, V); \
+  DType lwork(0); \
+  MXNET_LAPACK_##fname(MXNET_LAPACK_ROW_MAJOR, V.size(0), V.size(1), \
```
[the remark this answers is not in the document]
mseeger: The reason why I called the workspace query again in the syevd implementation, is that there, work consists of two different workspaces. But here, I think you can just use work.size(0) to pass for lwork, and don't have to do the query again.

src/operator/linalg_impl.h
```diff
@@ -1234,6 +1234,137 @@ LINALG_GPU_SYEVD_WORKSPACE_QUERY(DnDsyevd, double)
...
+  int ret(MXNET_LAPACK_##fname(MXNET_LAPACK_ROW_MAJOR, V.size(0), V.size(1), \
+                               UT.dptr_, UT.stride_, L.dptr_, V.dptr_, V.stride_, \
+                               work.dptr_, static_cast<int>(lwork))); \
+  CHECK_EQ(ret, 0) << #fname << " failed in lapack on cpu."; \
+}
+
+LINALG_CPU_GESVD(sgesvd, float)
+LINALG_CPU_GESVD(dgesvd, double)
+
+// Mangle temp storage requirements for DType and int into a single
```
mseeger: Remove this comment, it only applies to syevd. See my comment above, you have a single workspace here.

src/operator/numpy/linalg/np_gesvd-inl.h
```diff
@@ -0,0 +1,298 @@
...
+          utcol[i * ldut] = -utcol[i * ldut];
+        }
+      }
+    }
+  }
+};
+
+// (UT, L, V) = gesvd(A) [singular value decomposition]
+// - V can overwrite A
+// - Needs workspace (both DType and int), size of which is determined by a
```
mseeger: Only one workspace (DType) needed here

src/operator/numpy/linalg/np_gesvd-inl.h
```diff
@@ -0,0 +1,298 @@
...
+  MSHADOW_SGL_DBL_TYPE_SWITCH(outputs[0].type_flag_, OType, {
+    mshadow::Stream<xpu> *s = ctx.get_stream<xpu>();
+    laop::op(inputs[0].FlatToKD<xpu, 3, OType>(s),
+             outputs[0].FlatToKD<xpu, 3, OType>(s),
+             outputs[1].FlatToKD<xpu, 2, OType>(s),
+             outputs[2].FlatToKD<xpu, 3, OType>(s), ctx, attrs);
+  });
+}
+
+// Helper for gesvd_backward. See technical report for details
```
mseeger: Is the report cited somewhere?

src/operator/numpy/linalg/np_gesvd-inl.h
```diff
@@ -0,0 +1,298 @@
...
+MSHADOW_XINLINE float gesvd_back_helper_eps(float* X) {
+  return 1e-30;
+}
+
+template<>
+MSHADOW_XINLINE double gesvd_back_helper_eps(double* X) {
+  return 1e-100;
+}
+
+struct GesvdBackHelper_dV {
```
mseeger: Comment: dA overwritten by L^-1 dA

src/operator/numpy/linalg/np_gesvd-inl.h
```diff
@@ -0,0 +1,298 @@
...
+      denom = L[offl + i];
+      if (denom < eps) denom = eps;
+      for (int j = 0; j < n; ++j) {
+        dA[offda + i * ldda + j] /= denom;
+      }
+    }
+  }
+};
+
+struct GesvdBackHelper_G1 {
```
mseeger: Comment: X (square) overwritten by L X

src/operator/numpy/linalg/np_gesvd-inl.h
```diff
@@ -0,0 +1,298 @@
...
+    // From here on, we work on dA only
+
+    // Need temporal space, same shape as dUT
+    // invdV:
+    Kernel<GesvdBackHelper_dV, xpu>::Launch
+      (s, dA.size(0), dA.size(1), dA.size(2), L.dptr_, L.stride_, dA.dptr_, dA.stride_);
+
+    // G1:
+    // This copy is just to make sure there are no invalid values (NaN, infinity) in tempM
+    Copy(tempMs, dUT, s);
```
mseeger: Good! I hope our old code does that as well

src/operator/numpy/linalg/np_gesvd-inl.h
```diff
@@ -0,0 +1,298 @@
...
+    Copy(tempMr, dA, s);
```
mseeger: This is wrong. We do not need a temp space of the size of dA

src/operator/numpy/linalg/np_gesvd-inl.h
```diff
@@ -0,0 +1,298 @@
...
+        if (denom1 < eps) denom1 = eps;
+        if (denom2 < eps) denom2 = eps;
+        elem = (X[offx + i * ldx + j] - X[offx + j * ldx + i]) / denom1 / denom2;
+        X[offx + i * ldx + j] = elem * L[offl + j];
+        X[offx + j * ldx + i] = elem * L[offl + i];
+      }
+    }
+    for (int i = 0; i < m; ++i) {
+      elem = DType(0.0);
+      for (int j = 0; j < n; ++j) {
```
mseeger: This is not needed. You have computed this before already, using gemm, just pass in the diagonal.
Also, the function does not need dA and V as input args.

src/operator/numpy/linalg/np_gesvd-inl.h
```diff
@@ -0,0 +1,298 @@
...
+    // Need temporal space, same shape as dUT
+    // invdV:
+    Kernel<GesvdBackHelper_dV, xpu>::Launch
+      (s, dA.size(0), dA.size(1), dA.size(2), L.dptr_, L.stride_, dA.dptr_, dA.stride_);
+
+    // G1:
+    // This copy is just to make sure there are no invalid values (NaN, infinity) in tempM
+    Copy(tempMs, dUT, s);
+    Copy(tempMr, dA, s);
+    gemm::op(dA, V, tempMs, DType(1.0), DType(0.0), false, true, s);
```
mseeger: After this line, you extract the diagonal of this matrix and then pass it to GesvdBackHelper_G2

src/operator/numpy/linalg/np_gesvd-inl.h
```diff
@@ -0,0 +1,298 @@
...
+                 const Tensor<xpu, 3, DType>& dA,
+                 const Tensor<xpu, 3, DType>& tempMs,
+                 const Tensor<xpu, 3, DType>& tempMr,
+                 Stream<xpu>* s, const nnvm::NodeAttrs& attrs) {
+    // Backward of (UT, L, V) = gesvd(A)
+    using namespace mxnet_op;
+    if (dA.dptr_ != dV.dptr_) {
+      Copy(dA, dV, s);
+    }
+    // From here on, we work on dA only
```
mseeger: Could you assign k = dA.size(0), m = dA.size(1), n = dA.size(2) here, and use them below?

src/operator/numpy/linalg/np_gesvd-inl.h
```diff
@@ -0,0 +1,298 @@
...
+    Copy(tempMs, dUT, s);
+    Copy(tempMr, dA, s);
+    gemm::op(dA, V, tempMs, DType(1.0), DType(0.0), false, true, s);
+    Kernel<GesvdBackHelper_G1, xpu>::Launch
+      (s, dA.size(0), dA.size(1), dA.size(2), tempMs.dptr_, tempMs.stride_,
+       L.dptr_, L.stride_);
+    gemm::op(dUT, UT, tempMs, DType(1.0), DType(1.0), true, false, s);
+
+    // G2:
+    Kernel<GesvdBackHelper_G2, xpu>::Launch
```
mseeger: Pass in the diagonal extracted above, and do not pass dA, V.

src/operator/numpy/linalg/np_gesvd-inl.h
```diff
@@ -0,0 +1,298 @@
...
+      (s, dA.size(0), dA.size(1), dA.size(2), tempMs.dptr_, tempMs.stride_,
+       L.dptr_, L.stride_, dL.dptr_, dL.stride_, dA.dptr_, dA.stride_,
+       V.dptr_, V.stride_);
+
+    // G3:
+    gemm::op(tempMs, V, dA, DType(1.0), DType(1.0), false, false, s);
+    gemm::op(UT, dA, tempMr, DType(1.0), DType(0.0), false, false, s);
```
mseeger: It is very annoying we need this large temp space, because we don't have in-place left-multiply with square matrix. The drawback here is that for large matrices (large n), this needs a large temp space. It may be worth avoiding that.

tests/python/unittest/test_numpy_op.py
```diff
@@ -0,0 +1,131 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
mseeger: Very nice test!

src/operator/numpy/linalg/np_gesvd-inl.h
```diff
@@ -0,0 +1,298 @@
...
+             outputs[2].FlatToKD<xpu, 3, OType>(s), ctx, attrs);
+  });
+}
+
+// Helper for gesvd_backward. See technical report for details
+template<typename DType>
+DType gesvd_back_helper_eps(DType* X);
+
+template<>
+MSHADOW_XINLINE float gesvd_back_helper_eps(float* X) {
```
asmushetzel: Would be cleaner to make this constants dependent on values in std::numeric_limits. So something along the lines of 
std::numeric_limits<DType>::epsilon*10 (or whatever number you need to be on the safe side)
That also avoids doing template specialization here.

src/operator/numpy/linalg/np_gesvd-inl.h
asmushetzel: May be cleaner to fill with zero instead of just copying some arbitrary data. I think there is a specific method in mshadow or elsewhere to fill a tensor with such values (if not, copying like here is likely the best way)

src/operator/numpy/linalg/np_gesvd-inl.h
hzfan: Note that tempMs cannot be used to store the extracted diagonal (I have used tempMs to store G1). Do we need extra temp space (of size m) to store the diagonal ?

src/operator/numpy/linalg/np_gesvd-inl.h
```diff
@@ -0,0 +1,298 @@
...
+  template<typename xpu, typename DType>
+  static void op(const Tensor<xpu, 3, DType>& dUT,
+                 const Tensor<xpu, 2, DType>& dL,
+                 const Tensor<xpu, 3, DType>& dV,
+                 const Tensor<xpu, 3, DType>& UT,
+                 const Tensor<xpu, 2, DType>& L,
+                 const Tensor<xpu, 3, DType>& V,
+                 const Tensor<xpu, 3, DType>& dA,
+                 const Tensor<xpu, 3, DType>& tempMs,
+                 const Tensor<xpu, 3, DType>& tempMr,
```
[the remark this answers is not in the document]
hzfan: Could you elaborate how to 
"coding up "in-place" dA <- dot(UT, dA) using a temp space of shape (m, m)"?
The method I have thought about is:
Assume n % m == 0 for simplicity.
We can write dA as blocks: dA = [dA1, dA2, ..., dAx], where dAi is of shape (m, m) and x = n / m
so dot(UT, dA) = [dot(UT, dA1), ..., dot(UT, dAx], and each dot can be achieved with temp space of shape (m, m)
I don't know whether I have understood correctly.

src/operator/numpy/linalg/np_gesvd-inl.h
hzfan: Our old code does that. I learnt from our old code (syevd) about this.

src/operator/numpy/linalg/np_gesvd-inl.h
hzfan: Not yet. The public technical report (https://arxiv.org/pdf/1710.08717.pdf) does not include details about svd.

src/operator/numpy/linalg/np_gesvd-inl.h
hzfan: The only way I know to fill tensor with zero is to:
```
    tempMs.FlatTo1D() = 0;
    tempMr.FlatTo1D() = 0;
```
But our old code (syevd) sticks to copying. Which do you think is better?

hzfan: The copy in syevd is [here](https://github.com/apache/incubator-mxnet/blob/67daae7eff1ef06254693bc1bc30bbfc23b2cf99/src/operator/tensor/la_op-inl.h#L861)

src/operator/numpy/linalg/np_gesvd-inl.h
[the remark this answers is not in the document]
mseeger: Yes exactly, something like you are saying. Essentially you slice the large matrix up into (m, m) blocks, and then you can even just call gemm in a loop. This may be the easiest, in fact. What I mean, you mask out (m, m) blocks of dA and iterate over them, always replacing one block B with dot(UT, B). For that, you need an (m, m) temp space, but that you have.

src/operator/numpy/linalg/np_gesvd-inl.h
mseeger: Yes true, but that is really small extra space. But yes, have to allocate that as well. Instead, we can get rid of tempMr (below)

src/operator/numpy/linalg/np_gesvd-inl.h
mseeger: Ah, you are right. I will have the new report version being uploaded.

src/operator/numpy/linalg/np_gesvd-inl.h
mseeger: Aha, so it seems I am to blame for that. But it is fine

src/operator/numpy/linalg/np_gesvd-inl.h
hzfan: imho, X (square) overwritten by X L

mseeger: Yes, it is all transposed, because of this row/col major issue

src/operator/numpy/linalg/np_gesvd-inl.h
mseeger: Could you cite the arxiv paper in the code? The new version with SVD will be uploaded in the next few days, way before this CR will get merged. Thanks.

src/operator/c_lapack_api.h
hzfan: The LAPACK_gesvd function interface differs in signature from the MXNET_LAPACK-signature and have to be wrapped (as is stated [here](https://github.com/apache/incubator-mxnet/blob/67191c4df9ba363605c59b332c1a8096573309e0/src/operator/c_lapack_api.h#L201)). So this is basically a wrapper of LAPACK_gesvd.

I added some comments about how to use the LAPACK_gesvd. Its official document can be found [here](https://software.intel.com/en-us/mkl-developer-reference-c-gesvd).

src/operator/c_lapack_api.h
hzfan: Added

src/operator/linalg.h
hzfan: Fixed

src/operator/linalg_impl.h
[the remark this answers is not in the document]
hzfan: Yes, you are right. I have removed the query.

src/operator/linalg_impl.h
hzfan: Removed

src/operator/numpy/linalg/np_gesvd-inl.h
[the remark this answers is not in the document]
mseeger: Please ask if this is still unclear (but your comment is what I have in mind). You can mask out blocks of dA simply by moving the pointer, everything else (stride, etc) remains the same, because the n-axis is the continuous one. The final block will be (m, m2), where m2 <= m, but that should also not be a problem.

src/operator/numpy/linalg/np_gesvd-inl.h
hzfan: Fixed

src/operator/numpy/linalg/np_gesvd-inl.h
hzfan: Comment added

src/operator/numpy/linalg/np_gesvd-inl.h
hzfan: Comment added

src/operator/numpy/linalg/np_gesvd-inl.h
hzfan: Yes, fixed.

src/operator/numpy/linalg/np_gesvd-inl.h
hzfan: Fixed

src/operator/numpy/linalg/np_gesvd-inl.h
hzfan: Fixed

src/operator/numpy/linalg/np_gesvd-inl.h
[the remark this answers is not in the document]
hzfan: Now my implementation loops over the m/n blocks (each of shape (m, m) and reduce the temp space needed to (m, m).

src/operator/numpy/linalg/np_gesvd-inl.h
hzfan: I tried it out. But I found std::numeric_limits::epsilon() cannot be accessed in Cuda. So I stick with the original implementation for now.

src/operator/numpy/linalg/np_gesvd-inl.h
hzfan: I changed the copying to filling with zeros.

hzfan:
Just updated the code.
The two main points updated:
- reuse the diagonal results of (L^-1 dV VT) with an additional temp space of shape (m,)
- avoid tempMr (which is of shape (m, n)) by joining blocks of dots.

Now the total temp space used is (k, m, m) + (k, m)

src/operator/numpy/linalg/np_gesvd-inl.h
hzfan: Yes. Cited.

src/operator/numpy/linalg/np_gesvd-inl.h
```diff
@@ -0,0 +1,302 @@
...
+
+    // G2:
+    Kernel<GesvdBackHelper_G2, xpu>::Launch
+      (s, k, m, n, tempM.dptr_, tempM.stride_,
+       L.dptr_, L.stride_, dL.dptr_, dL.stride_,
+       tempMd.dptr_, tempMd.stride_);
+
+    // G3:
+    gemm::op(tempM, V, dA, DType(1.0), DType(1.0), false, false, s);
+    for (int i = 0; i < n; i += m) {
```
hzfan: Here I go through the blocks (0:m, i: i + m). Other attrs of tensors like stride_, stream_ are not changed, as you said. The ncols is the m2 you just mentioned. It is used for the last block.

mseeger: Add comment:
dA <- dot(UT, dA). Loop over (k, m, m) blocks to avoid large temporary memory

hzfan: Comment added.

hzfan:
Thank @mseeger  and @asmushetzel for guidance and review.

Issue #15790, opened 2019-08-08
Can't train CIFAR10 with debug flag using CMake / GPU / MKL


larroy:
## Description

train cifar_10 doesn't seem to make any progress when MXNet is built with CMake and debug flags.

## Environment info (Required)

Console  is flooded with OMP warnings:

```
OMP: Error #13: Assertion failure at kmp_runtime.cpp(6481).
OMP: Hint: Please submit a bug report with this message, compile and run commands used, and machine configuration info including native compiler and operating system versions. Faster response will be obtained by including all program sources. For information on submitting this issue, please see https://bugs.llvm.org/.
Assertion failure at kmp_runtime.cpp(6481): __kmp_team_pool == __null.
OMP: Error #13: Assertion failure at kmp_runtime.cpp(6481).
OMP: Hint: Please submit a bug report with this message, compile and run commands used, and machine configuration info including native compiler and operating system versions. Faster response will be obtained by including all program sources. For information on submitting this issue, please see https://bugs.llvm.org/.
Assertion failure at kmp_runtime.cpp(6481): __kmp_team_pool == __null.
OMP: Error #13: Assertion failure at kmp_runtime.cpp(6481).
OMP: Hint: Please submit a bug report with this message, compile and run commands used, and machine configuration info including native compiler and operating system versions. Faster response will be obtained by including all program sources. For information on submitting this issue, please see https://bugs.llvm.org/.
Assertion failure at kmp_runtime.cpp(6481): __kmp_team_pool == __null.
OMP: Error #13: Assertion failure at kmp_runtime.cpp(6481).
OMP: Hint: Please submit a bug report with this message, compile and run commands used, and machine configuration info including native compiler and operating system versions. Faster response will be obtained by including all program sources. For information on submitting this issue, please see https://bugs.llvm.org/.
Version      : 1.6.0
Directory    : /home/piotr/mxnet_master_cmake_debug/python/mxnet
Commit hash file "/home/piotr/mxnet_master_cmake_debug/python/mxnet/COMMIT_HASH" not found. Not installed from pre-built package or built from source.
Library      : ['/home/piotr/mxnet_master_cmake_debug/python/mxnet/../../build/libmxnet.so']
Build features:
✔ CUDA
✔ CUDNN
✖ NCCL
✔ CUDA_RTC
✖ TENSORRT
✔ CPU_SSE
✔ CPU_SSE2
✔ CPU_SSE3
✔ CPU_SSE4_1
[...]

PR #15808, opened 2019-08-08
Add option to choose between OMP implementations

larroy:
## Description ##
Adds a build switch to choose OpenMP implementations:

Workaround for #15790 

@karan6181

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,7 @@ mxnet_option(USE_CUDA             "Build with CUDA support"   ON)
 mxnet_option(USE_OLDCMAKECUDA     "Build with old cmake cuda" OFF)
 mxnet_option(USE_NCCL             "Use NVidia NCCL with CUDA" OFF)
 mxnet_option(USE_OPENCV           "Build with OpenCV support" ON)
-mxnet_option(USE_OPENMP           "Build with Openmp support" ON)
+mxnet_option(USE_OPENMP           "Build with Openmp support" ON) # OFF | ON | PLATFORM | BUNDLED
 mxnet_option(USE_CUDNN            "Build with cudnn support"  ON) # one could set CUDNN_ROOT for search path
 mxnet_option(USE_SSE              "Build with x86 SSE instruction support" ON IF NOT ARM)
 mxnet_option(USE_F16C             "Build with x86 F16C instruction support" ON) # autodetects support if ON
@@ -432,14 +432,13 @@ endif()
 
 # ---[ OpenMP
 if(USE_OPENMP)
-  find_package(OpenMP REQUIRED)
   # This should build on Windows, but there's some problem and I don't have a Windows box, so
   # could a Windows user please fix?
-  if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/openmp/CMakeLists.txt
-     AND SYSTEM_ARCHITECTURE STREQUAL "x86_64"
-     AND NOT MSVC
-     AND NOT CMAKE_CROSSCOMPILING)
-
+  if(USE_OPENMP STREQUAL "BUNDLED" AND EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/openmp/CMakeLists.txt
+      AND SYSTEM_ARCHITECTURE STREQUAL "x86_64"
+      AND NOT MSVC
+      AND NOT CMAKE_CROSSCOMPILING)
+    message("Using bundlded LLVM OpenMP from 3rdparty")
     # Intel/llvm OpenMP: https://github.com/llvm-mirror/openmp
     set(OPENMP_STANDALONE_BUILD TRUE)
     set(LIBOMP_ENABLE_SHARED TRUE)
@@ -452,7 +451,9 @@ if(USE_OPENMP)
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
     add_definitions(-DMXNET_USE_OPENMP=1)
-  else()
+  elseif(USE_OPENMP STREQUAL "PLATFORM" OR USE_OPENMP STREQUAL "ON")
+    find_package(OpenMP REQUIRED)
+    message("Using platform provided OpenMP")
     if(OPENMP_FOUND)
       set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
       set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
@@ -460,6 +461,8 @@ if(USE_OPENMP)
       set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
       add_definitions(-DMXNET_USE_OPENMP=1)
     endif()
+  else()
+    message(FATAL_ERROR "USE_OPENMP takes values [PLATFORM, BUNDLED, OFF]")
   endif()
 elseif(UNIX AND NOT ANDROID)
   list(APPEND mxnet_LINKER_LIBS pthread)
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "3rdparty/dmlc-core"]
 	path = 3rdparty/dmlc-core
-	url = https://github.com/dmlc/dmlc-core.git
+	url = https://github.com/larroy/dmlc-core.git
 [submodule "3rdparty/ps-lite"]
 	path = 3rdparty/ps-lite
 	url = https://github.com/dmlc/ps-lite
--- a/cmake/cmake_options.yml
+++ b/cmake/cmake_options.yml
@@ -20,7 +20,7 @@ USE_CUDA: "OFF" # Build with CUDA support
 USE_OLDCMAKECUDA: "OFF" # Build with old cmake cuda
 USE_NCCL: "OFF" # Use NVidia NCCL with CUDA
 USE_OPENCV: "ON" # Build with OpenCV support
-USE_OPENMP: "ON" # Build with Openmp support
+USE_OPENMP: "ON" # Build with Openmp support 
 USE_CUDNN: "ON" # Build with cudnn support) # one could set CUDNN_ROOT for search path
 USE_SSE: "ON" # Build with x86 SSE instruction support IF NOT ARM
 USE_F16C: "ON" # Build with x86 F16C instruction support) # autodetects support if "ON"
--- a/cmake/cmake_options.yml
+++ b/cmake/cmake_options.yml
@@ -20,7 +20,7 @@ USE_CUDA: "OFF" # Build with CUDA support
 USE_OLDCMAKECUDA: "OFF" # Build with old cmake cuda
 USE_NCCL: "OFF" # Use NVidia NCCL with CUDA
 USE_OPENCV: "ON" # Build with OpenCV support
-USE_OPENMP: "ON" # Build with Openmp support 
+USE_OPENMP: "ON" # Build with Openmp support
 USE_CUDNN: "ON" # Build with cudnn support) # one could set CUDNN_ROOT for search path
 USE_SSE: "ON" # Build with x86 SSE instruction support IF NOT ARM
 USE_F16C: "ON" # Build with x86 F16C instruction support) # autodetects support if "ON"


larroy:
@mxnet-label-bot add [Build, CMake]

larroy:
@mxnet-label-bot add [Breaking]

CMakeLists.txt
```diff
@@ -433,11 +434,11 @@ if(USE_OPENMP)
...
-  if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/openmp/CMakeLists.txt
-     AND SYSTEM_ARCHITECTURE STREQUAL "x86_64"
-     AND NOT MSVC
-     AND NOT CMAKE_CROSSCOMPILING)
-
+  if(USE_OPENMP_BUNDLED_LLVM AND EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/openmp/CMakeLists.txt
+      AND SYSTEM_ARCHITECTURE STREQUAL "x86_64"
+      AND NOT MSVC
+      AND NOT CMAKE_CROSSCOMPILING)
+    message("Using bundlded LLVM OpenMP")
```
ChaiBapchya: nitpick: bundled

CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ mxnet_option(USE_OLDCMAKECUDA     "Build with old cmake cuda" OFF)
 mxnet_option(USE_NCCL             "Use NVidia NCCL with CUDA" OFF)
 mxnet_option(USE_OPENCV           "Build with OpenCV support" ON)
 mxnet_option(USE_OPENMP           "Build with Openmp support" ON)
+mxnet_option(USE_OPENMP_BUNDLED_LLVM "Build with bundled llvm openmp from 3rdparty" OFF)
```
apeforest: What happens when USE_OPENMP is OFF and USE_OPENMP_BUNDLED_LLVM ON?

larroy: USE_OPENMP takes precedence, so no OMP is used.

apeforest: Is this documented somewhere? Can we set mxnet_option(USE_OPENMP   3RD_PARTY | BUILT_IN)?

I am not an expert in CMake so I can not really decide whether this is the most professional way. LGTM for now. But it's better to have someone from CI team to merge this PR.

lebeg:
> LGTM. But I am not an expert in CMake so I can not really decide whether this is the most proper way. @marcoabreu @lebeg for review.

I think what you were [proposing](https://github.com/apache/incubator-mxnet/pull/15808#discussion_r313211314) would be a better way to handle this. Instead of `ON/OFF` I would define an order in which the Omp versions would be preferred. Like a list `OMP_SYSTEM, OMP_BUNDLED, OMP_OFF`.

For a case like `OMP_SYSTEM` if it's not found it would be an error.

larroy:
@lebeg you might be right, but getting this working is urgent as we can't debug. I would propose doing it incrementally or getting your contribution in this direction.

CMakeLists.txt
larroy: I used this for consistency. What you propose is also valid. I would do this incrementally as right now debugging is blocked.

apeforest:
@larroy What's the effort required to implement the multi-selection as you earlier proposed? If this only blocks debugging using CMake and not regular users or developers using Make, I am more leaning towards @lebeg suggestion.

larroy:
CI is blocked now, and it's going to be blocked for a few days. There's problems with Windows and other things. If I make changes I'm afraid this won't be merged for a while. Up to you guys to decide. I think this is urgent enough to have it merged and then refine it. I would say this is a critical fix for being able to debug, as other people need to debug issues related to other tasks. (Make in debug requires updating the linker which is something I would prefer to avoid in my workstation).

The effort to implement your suggestion should be around 30 minutes.

marcoabreu:
I'd also prefer the multi selection.

larroy:
Can you guys review again?

CMakeLists.txt
```diff
@@ -431,10 +433,14 @@ endif()
 
 # ---[ OpenMP
 if(USE_OPENMP)
+  if (NOT OPENMP_FLAVOR)
```
apeforest: Where is OPENMP_FLAVOR from?

CMakeLists.txt
```diff
@@ -16,15 +16,15 @@ if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/build/private/local_config.cmake)
   include(${CMAKE_CURRENT_SOURCE_DIR}/build/private/local_config.cmake)
 endif()
 
+
```
apeforest: As we already agreed, please do a diff before sending the PR. Remove unnecessary blank lines in PR please.

apeforest:
@larroy I thought the suggested change is to use multi-selection for USE_OPENMP not adding additional variable:

It would look like:
```
USE_OPENMP:           "Select with Openmp support"    OMP_SYSTEM |  | OMP_BUNDLED | OMP_OFF)
```

@lebeg Please let me know if my understanding is correct. Thanks

CMakeLists.txt
larroy: From cmake command line. Is the selector as requested.

larroy:
@apeforest I think what you propose is not homogeneous with what we have. I think is better to leave an intitial boolean switch and a fine grained selector. It also keeps backwards compatibility.

lebeg:
> @larroy I thought the suggested change is to use multi-selection for USE_OPENMP not adding additional variable:
> 
> It would look like:
> 
> ```
> USE_OPENMP:  "Configure Openmp support"    OMP_SYSTEM |  | OMP_BUNDLED | OMP_OFF)
> ```
> 
> @lebeg Please let me know if my understanding is correct. Thanks

Yes, this is more or less what I had in mind. For the sake of backwards compatibility (and maybe simplicity) keeping the option `ON` would default to `OMP_SYSTEM`.

lebeg:
There seem to be some unintended changes in this PR, otherwise it looks good

larroy:
They are for CI, a change in dmlc is needed first:

https://github.com/dmlc/dmlc-core/pull/558

larroy:
@lebeg please approve then

CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ mxnet_option(USE_CUDA             "Build with CUDA support"   ON)
 mxnet_option(USE_OLDCMAKECUDA     "Build with old cmake cuda" OFF)
 mxnet_option(USE_NCCL             "Use NVidia NCCL with CUDA" OFF)
 mxnet_option(USE_OPENCV           "Build with OpenCV support" ON)
-mxnet_option(USE_OPENMP           "Build with Openmp support" ON)
+mxnet_option(USE_OPENMP           "Build with Openmp support" ON) # OFF | ON | PLATFORM | BUNDLED
```
apeforest: What's the difference between ON and the rest? Can you add some comments here?

CMakeLists.txt
```diff
@@ -452,14 +451,18 @@ if(USE_OPENMP)
...
+    message("Using platform provided OpenMP")
     if(OPENMP_FOUND)
       set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
       set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
       set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
       set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
       add_definitions(-DMXNET_USE_OPENMP=1)
     endif()
+  else()
+    message(FATAL_ERROR "USE_OPENMP takes values [PLATFORM, BUNDLED, OFF]")
```
apeforest: "USE_OPENMP takes values [PLATFORM, BUNDLED, OFF, ON]")?

apeforest:
> They are for CI, a change in dmlc is needed first:
> 
> [dmlc/dmlc-core#558](https://github.com/dmlc/dmlc-core/pull/558)

Why is it needed for this PR?

CMakeLists.txt
```diff
@@ -432,14 +432,13 @@ endif()
...
 if(USE_OPENMP)
-  find_package(OpenMP REQUIRED)
   # This should build on Windows, but there's some problem and I don't have a Windows box, so
   # could a Windows user please fix?
-  if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/openmp/CMakeLists.txt
-     AND SYSTEM_ARCHITECTURE STREQUAL "x86_64"
-     AND NOT MSVC
-     AND NOT CMAKE_CROSSCOMPILING)
-
+  if(USE_OPENMP STREQUAL "BUNDLED" AND EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/openmp/CMakeLists.txt
```
marcoabreu: The first check and the others should be separate to throw appropriate error messages

CMakeLists.txt
```diff
@@ -452,14 +451,18 @@ if(USE_OPENMP)
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
     add_definitions(-DMXNET_USE_OPENMP=1)
-  else()
+  elseif(USE_OPENMP STREQUAL "PLATFORM" OR USE_OPENMP STREQUAL "ON")
```
marcoabreu: What's the difference between on and platform

szha:
@cjolivier01 if you'd like your change request to be considered a veto, please provide a technical reason. You may refer to the [Apache Voting Process](https://www.apache.org/foundation/voting.html#Veto) for details.

Reopening given that @larroy didn't indicate that he'd like to abandon this change.

CMakeLists.txt
larroy: They are the same, for backwards compatibility.

CMakeLists.txt
larroy: Can you explain what you mean exactly?

larroy:
Submodule update is necessary as build was failing, there's a fix in the submodule that should be merged soon.

cjolivier01:
this is a repeat of a previously vetoed proposal by Pedro and another by someone he tried to sneak a similar change in, just under a different guise. I really don’t have the energy to fight openly discuss this anymore after two years.

cjolivier01:
in addition, on a fundamental level, flippantly switching between omp libraries while taking nothing else into considerations is extraordinarily hazardous (ie what’s mkl bringing in?).

larroy:
This has repeatedly caused crashes reported by several users which didn't happen when we used the platform OMP which is also the one shipped with binary releases. You are doing a disservice to the community with this attitude, the users are free to choose the OMP implementation that they want.

szha:
@cjolivier01 I didn't realize that. If this is indeed a repetition of a change proposal already vetoed, would you mind linking to it?

I only recall a case where @lebeg attempted to remove the bundled omp, which is different from what @larroy is proposing here. If this is what you're referring to, given the difference, would you mind providing a technical reason for rejecting this change?

cjolivier01:
There is plenty of info out there already on this issue including discussions on dev. I will not be forever to refuge this thing every time Pedro wants to sneak basically the exact same change in every few mo this for two years+ while never actually root-causing the issue (he’s been told what the issue was — not the OMP library’s fault — it’s a big in mxnet — see the issue he is locked out of), but ignores repeatedly.

larroy:
@cjolivier01 stop holding this project hostage. This will be reported to the relevant decision makers in Apache. I hope your powers are revoked due to your track record of destructive actions.

CMakeLists.txt
marcoabreu: If someone uses the option "bundled" but the "exists" check fails, they'd receive an error message that they selected an invalid mode.

marcoabreu:
If the submodule update fixes the build, please put it in a separate PR as already stated.

marcoabreu:
While I try to not engage in the discussion above whether this is a match to the previously vetoed PR, I agree that I see some similarities. 

With regards to #15790: This PR doesn't actually solve the problem (which appears to be the same as previously) but gives users a different OMP library choice. That effectively means that the default method we use is still broken and users would have to know that they have to use a different one. I might be no expert in that field, but I agree that to this day we only read something about some assertion being hit and nothing further. From what I can see, there has been no root causing (to the actual root) but a lot based on assumptions. This change seems quite massive (introducing a new OMP Backend choice) while not being based on real data. So i understand the frustration from both sides here: Chris' request for root causing is constantly being denied, while the root cause for pedro seems apparent. But since we are a community and we would like to make a well educated decision as a group, I'd agree with the general request for more root causing in order to create transparency. 

Thinking about it, I agree that this is the same change in a different costume that already had been vetoed with proper technical reasons - but I understand that Chris is getting frustrated by constantly having to repeat his request.

So yeah, I agree with his veto that I'd like to see a deep dive. That should be collected somewhere and if we are at a crossroads, we should bring the topic up for a broader vote on dev@.

marcoabreu:
Another proposal to avoid the difficult personal situation is to hand this issues over to somebody else. Maybe another engineer at Amazon or somebody at Intel or Nvidia would like to be of assistance here. I'm certain that they are also quite capable of debugging concurrency issues. Wdyt?

PR #15815, opened 2019-08-09
Numpy add numpy op hanning, hamming, blackman

gyshi:
## Description ##

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/python/mxnet/ndarray/numpy/_op.py
+++ b/python/mxnet/ndarray/numpy/_op.py
@@ -34,7 +34,7 @@ __all__ = ['zeros', 'ones', 'full', 'add', 'subtract', 'multiply', 'divide', 'mo
            'trunc', 'logical_not', 'arcsinh', 'arccosh', 'arctanh', 'tensordot',
            'linspace', 'expand_dims', 'tile', 'arange', 'split', 'concatenate', 'stack', 'vstack', 'mean',
            'maximum', 'minimum', 'swapaxes', 'clip', 'argmax', 'std', 'var', 'indices', 'copysign',
-           'ravel']
+           'ravel', 'hanning', 'hamming', 'blackman']
 
 
 @set_module('mxnet.ndarray.numpy')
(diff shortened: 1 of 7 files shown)


src/operator/numpy/np_window_op.h
```diff
@@ -0,0 +1,139 @@
...
+    .describe("Context of output, in format [cpu|gpu|cpu_pinned](n)."
+    "Only used for imperative calls.");
+    DMLC_DECLARE_FIELD(dtype)
+    .set_default(mshadow::kFloat64)
+    MXNET_ADD_ALL_TYPES
+    .describe("Data-type of the returned array.");
+  }
+};
+
+inline bool NumpyWindowsShape(const nnvm::NodeAttrs& attrs,
```
haojin2: put this in `.cc` file

python/mxnet/ndarray/numpy/_op.py
```diff
@@ -2432,3 +2433,249 @@ def indices(dimensions, dtype=_np.int32, ctx=None):
     else:
         raise ValueError("The dimensions must be sequence of ints")
 # pylint: enable=redefined-outer-name
+
+
+@set_module('mxnet.ndarray.numpy')
+def hanning(M, dtype=None, ctx=None):
```
comaniac: I still prefer to use float64 as the default type in the function argument instead of having a checker at the beginning of the function. The reason is that the type of `dtype` would be `Union[None, str, numpy.dtype]` in the current implementation, which is inconsistent with the docstring saying that  `dtype : str or numpy.dtype`. Also, it makes the semantic ambiguous and should be avoided especially in Python 3.7. If the default value of `dtype` in the argument is `np.float64` or 'float64' then we can enforce its type to be `Union[str, numpy.dtype]` as documented.

gyshi: it's good, i think i should use float32 as the defaulet type,

PR #15823, opened 2019-08-09
[Numpy] operator arctan2

tingying2020:
numpy operator arctan2


python/mxnet/ndarray/numpy/_op.py
```diff
@@ -293,3 +293,89 @@ def power(x1, x2, out=None):
...
+    Note the order of the parameters. `arctan2` is defined also when `x2` = 0
+    and at several other special points, obtaining values in
+    the range ``[-pi, pi]``:
+
+    >>> x = np.array([1, -1])
+    >>> y = np.array([0, 0])
+    >>> np.arctan2(x, y)
+    array([ 1.5707964, -1.5707964])
+    """
+    return _ufunc_helper(x1, x2, _npi.arctan2, _np.arctan2,\
```
haojin2: no need for backslash `\` here.

tingying2020: fixed, thanks!

PR #15839, opened 2019-08-10
Add register_op_hook for gluon

anirudh2290:
## Description ##

This PR adds a register_op_hook for Block. One can register_op_hook to monitor all the ops that are called by the HybridBlock after hybridization. For example:

```
def mon_callback(node_name, opr_name, arr):
      print(py_str(node_name))
      print(py_str(opr_name))
model = mx.gluon.nn.HybridSequential(prefix="dense_")
with model.name_scope():
     model.add(mx.gluon.nn.Dense(2))
model.initialize()
model.hybridize()
model.register_op_hook(mon_callback, monitor_all=True)
model(mx.nd.ones((2, 3, 4)))
```

This will print the following : 
```
dense_dense0_fwd_data
FullyConnected
dense_dense0_fwd_weight
FullyConnected
dense_dense0_fwd_bias
FullyConnected
dense_dense0_fwd_output
FullyConnected
```

Setting monitor_all=False will print only the output : 
```
dense_dense0_fwd_output
FullyConnected
```

### What is not supported ###

1. register_op_hook on non hybridized blocks is not supported. If a block is not a hybrid block and register_op_hook is called on it, it will be registered only on HybridBlocks which are children of the blocks. The callback will only be invoked after hybridization.
2. Calback is not supported when static_shape=True for call to hybridize.

cc @Vikas-kum 

### Essentials ###

### Changes ###

## Comments ##


src/imperative/imperative_utils.cc
```diff
@@ -120,6 +121,18 @@ void InvokeOperator(const nnvm::IndexedGraph& idx,
       *arrays[eid] = NDArray();
     }
   }
+  if (callback) {
+    for (size_t i = 0; i < ndinputs.size(); ++i) {
+      NDArray* cpy = new NDArray(*ndinputs[i]);
+      callback(node.source->attrs.name.c_str(), node.source->op()->name.c_str(),
```
Vikas-kum: above you mention that cb takes one string and a array handle. 
here i see 2 strings, am i missing anything here?

src/imperative/imperative_utils.cc
```diff
@@ -120,6 +121,18 @@ void InvokeOperator(const nnvm::IndexedGraph& idx,
...
+               reinterpret_cast<void*>(cpy));
+    }
+    for (size_t i = 0; i < ndoutputs.size(); ++i) {
+      NDArray* cpy = new NDArray(*ndoutputs[i]);
+      callback(node.source->attrs.name.c_str(), node.source->op()->name.c_str(),
```
Vikas-kum: i guess first is tensor name, 2nd is operator name, 3rd is value. Can you confirm this?
Is there a way to distinguish if the tensor is input or output ?

Also, i don't see monitor_all being used, is it used somewhere implicitly not in this PR?

src/imperative/imperative_utils.cc
anirudh2290: this is still wip, the comment is wrong and needs to be cleaned up , it takes 2 strings.

src/imperative/imperative_utils.cc
anirudh2290: yes you are right. there is a way to distinguish by having a postfix after node name for outputs. Alternatively we can have different callback registration apis for inputs and outputs. Currently it is common for both.

monitor_all is not being used yet and the intention was to use it for inputs vs inputs and outputs, similar to the way it is done with graph executor.

anirudh2290:
It looks like my changed files are not being picked up and probably cached files are being picked up for CI. @marcoabreu would you know how I can force clear the cache ?

src/common/utils.cc
```diff
@@ -51,5 +51,62 @@ void CastStorageDispatch<cpu>(const OpContext& ctx,
...
+    for (size_t i = 0; i < node->num_outputs(); ++i) {
+      output_names.emplace_back(std::to_string(i));
+    }
+  }
+
+  for (size_t i = 0; i < node->num_outputs(); ++i) {
+    if (state_arrays[idx.entry_id(nid, i)]->is_none()) {
+      continue;
+    }
+    NDArray *cpy = new NDArray(*state_arrays[idx.entry_id(nid, i)]);
```
rahul003: Can we avoid the copy and just pass a reference to the current array?

python/mxnet/gluon/block.py
```diff
@@ -833,6 +848,12 @@ def _deferred_infer_shape(self, *args):
     def _call_cached_op(self, *args):
         if self._cached_op is None:
             self._build_cache(*args)
+        assert self._cached_op, "cached op is not None"
+        if self._callback:
+            self._cached_op._register_op_hook(self._callback, self._monitor_all)
+            if len(self._flags) >= 2 and self._flags[1]:
+                warnings.warn("Callback is not supported when static_shape=True "
```
rahul003: Why's this? If it's not supported, do not register the callback?

rahul003: else say something like experimental for static_shape=True if you are not confident, but it might work

src/common/utils.cc
```diff
@@ -51,5 +51,62 @@ void CastStorageDispatch<cpu>(const OpContext& ctx,
...
+    }
+  }
+
+  for (size_t i = 0; i < node->num_inputs(); ++i) {
+    const nnvm::NodeEntry &input = node->inputs[i];
+    if (state_arrays[idx.entry_id(input)]->is_none()) {
+      continue;
+    }
+    NDArray *cpy = new NDArray(*state_arrays[idx.entry_id(input)]);
+    std::string name = inode.source->attrs.name + "_" + input_names[i];
```
rahul003: Is this naming convention standard across MXNet? For example exporting of graph in mxboard? https://github.com/awslabs/mxboard/blob/686933cfe466af324b53fcfe043093d37a0e38d1/python/mxboard/summary.py#L364

python/mxnet/gluon/block.py
anirudh2290: Yes I will say experimental, good point thanks. The reason this is experimental is because for certain code path the cached op directly calls engine push instead of calling InvokeOp. adding the callbacks in such cases was much more difficult since we just had the opr and the engine vars not the ndarray references. Please see line: https://github.com/apache/incubator-mxnet/blob/master/src/imperative/cached_op.cc#L689

For this case, it is not supported.

src/common/utils.cc
anirudh2290: this wont create a copy of data, though to make changes to ndarray metadata for copy without changing the original we still need this copy.

src/common/utils.cc
anirudh2290: I am not sure about mxboard, but in general in mxnet, the naming convention for input data doesnt prepend the node name. but since for register_op_hook we need to monitor intermediate nodes we are prepending node name here. Following the convention similar to register_monitor_callback.

apeforest:
Have we measured how much performance overhead this might introduce?

anirudh2290:
@apeforest no we haven't measured perf overhead since This is only used for monitoring. This should not be used in production or for perf intensive use cases.

src/common/utils.cc
anirudh2290: I have changed it to experimental and also added check both for static_alloc and static_shape.

anirudh2290:
Thank you @Vikas-kum, @rahul003, @apeforest for the review!!

joapolarbear:
@anirudh2290 It seems this pr may affect the trace results, below shows the visualized results using chrome://tracing when training a bert model.
image
In contrast, before I use this pr, I get the following dense traces. Many forward and backpropagation operators miss in the above traces. 
image
FYI, I encounter the following warning when using this pr. I am not sure whether this warning has any effect.
image

anirudh2290:
@joapolarbear do you just see much spacing out or is it that ops are missing in one but not in other. registering an op hook is expensive and called at the beginning and end of each op execution and may kill performance. It is meant to be used for debugging and should be commented out during profiling.

eric-haibin-lin:
> registering an op hook is expensive and called at the beginning and end of each op execution and may kill performance. 


@anirudh2290 did we measure how much overhead it is to invoke a dummy hook?

anirudh2290:
@eric-haibin-lin i didnt measure that, since this was intended only for debugging not to be used in production use cases. If I had to make a guesstimate i would say it would be very similar to invoking a dummy hook with set_monitor_callback in module.

python/mxnet/gluon/block.py
```diff
@@ -833,6 +848,12 @@ def _deferred_infer_shape(self, *args):
     def _call_cached_op(self, *args):
         if self._cached_op is None:
             self._build_cache(*args)
+        assert self._cached_op, "cached op is not None"
+        if self._callback:
+            self._cached_op._register_op_hook(self._callback, self._monitor_all)
+            if len(self._flags) >= 2 and (self._flags[1] or self._flags[0]):
```
kovtun77, line 854: `self._flags = list(kwargs.items())`
The condition is true even when it is not supposed to be. For example if
`self._flags = [('static_alloc', False), ('static_shape', False), ('inline_limit', 2)]`

PR #15855, opened 2019-08-12
[Numpy] cumprod

hzfan:
## Description ##
Use tvm to implement numpy compatible [cumprod](https://docs.scipy.org/doc/numpy/reference/generated/numpy.cumprod.html)

### Changes ###
- Add cumprod forward/backward with tvm
- Add tests and docs

## Comments ##
- do not work until https://github.com/dmlc/tvm/pull/3723 is merged.
- Thank @yzhliu for guidance and review.


python/mxnet/_numpy_op_doc.py
```diff
@@ -52,3 +52,49 @@ def _np_zeros_like(a):
...
+    >>> np.cumprod(a, axis=0)
+    array([[ 1.,  2.,  3.],
+           [ 4., 10., 18.]])
+    
+    The cumulative product for each row (i.e. over the columns) of `a`:
+    
+    >>> np.cumprod(a, axis=1)
+    array([[  1.,   2.,   6.],
+           [  4.,  20., 120.]])
+    """
```
hgt312: add `pass` and one more blank line

hzfan: fixed

contrib/tvmop/basic/ufunc.py
```diff
@@ -48,3 +51,186 @@ def vadd_gpu(dtype, ndim):
...
+    else:
+        ret, c_list, s_list = cumprod(X, ishape, dtype, ndim, axis)
+    s = tvm.create_schedule(ret.op)
+    return s, X, ret, c_list, s_list
+
+
+@defop(name="vcumprod", target="cpu", auto_broadcast=False,
+       dtype=AllTypesButHalf, ndim=list(range(0, 6)), axis=list(range(0, 5)) + [None],
+       attrs=["axis"], attrs_valid=lambda **kwargs: kwargs['axis'] is None or kwargs['axis'] < kwargs['ndim'])
+def vcumprod(dtype, ndim, axis):
```
reminisce: What does `v` mean here? In `vadd`, I think it used to refer to "add two **v**ectors" in the beginning, but that function was later extended to support auto broadcasting. So the prefix `v` may be not appropriate anymore. @yzhliu Could you confirm? Thanks.

yzhliu: should be just `cumprod`

hzfan: renamed as cumprod

hzfan:
I created 2 new files: cumprod.py and cumprod.cc. If there are better namespaces, plz let me know.

reminisce:
Regarding the namespace choice in the frontend (python), how about we just follow what NumPy does? For example, `cumprod` is in `numpy/core/fromnumeric.py`. We can create `tvmop/core/fromnumeric.py` and put `cumprod` in it. This can serve as an easy and clear guideline for others to follow in the future.

hzfan:
> Regarding the namespace choice in the frontend (python), how about we just follow what NumPy does? For example, `cumprod` is in `numpy/core/fromnumeric.py`. We can create `tvmop/core/fromnumeric.py` and put `cumprod` in it. This can serve as an easy and clear guideline for others to follow in the future.

Great! I created two files:
  - contrib/tvmop/core/fromnumeric.py
  - src/operator/contrib/tvmop/core/fromnumeric.cc

contrib/tvmop/core/fromnumeric.py
```diff
@@ -0,0 +1,203 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# coding: utf-8
+import tvm
+import topi
+from .. import defop, AllTypes, AllTypesButHalf
+
+def kernel_cumprod(X, ishape, dtype, ndim, axis):
```
haojin2: Suggestion: Maybe we can make this a common scan kernel, so that in the future similar scan operators could re-use this.

hzfan: Good. I will do it. What about moving the generalized kernel to `contrib/tvmop/utils.py`?

hzfan:
Have Rebased. And a minor change: move the dispatch of req to func `set_req` in `src/operator/tvmop/utils.h` so it can be reused.

hzfan:
Another minor modification: Rename `src/operator/contrib/tvmop/core/fromnumeric.cc` as `src/operator/numpy/np_cumprod.cc` (following the naming convention of `cumsum.cc`)

ci/docker/runtime_functions.sh
```diff
@@ -367,6 +367,7 @@ build_ubuntu_cpu_openblas() {
     set -ex
     export CC="gcc"
     export CXX="g++"
+    pip3 install --user psutil
```
marcoabreu: Please do package installation in the dockerfiles

marcoabreu: Also, why is this required?

hzfan: I tried to install it in `ci/docker/install/ubuntu_python.sh`. It worked fine for unix-cpu, but edge fails. 

Edge runs `ci/build.py --docker-registry mxnetci -p test.arm_qemu ./runtime_functions.py` and one step of it is 
```
Step 5/17 : RUN /work/ubuntu_python.sh
 ---> Using cache
 ---> 7535d4f37ed9
```
But it seems that the package I want to install (psutil) cannot be installed on edge device. 

Could you plz kindly point out where to install the `psutil` package?

hzfan: A part of `tvm`, depends on `psutil`. As we are using `tvm` to generate ops, `psutil` is needed.

hzfan: To be more specific, `tvm/python/tvm/rpc/server.py` depends on `psutil`

yzhliu:
@hzfan Could you disable fp16 on cuda for now? We will fix it later.

ci/docker/runtime_functions.sh
marcoabreu: How about doing it in https://github.com/apache/incubator-mxnet/blob/master/ci/docker/install/ubuntu_tvm.sh ?

hzfan: Thank you for the suggestion. I tried it and found that `Dockerfile.build.ubuntu_gpu_cu*` runs `ubuntu_tvm.sh`, but `Dockerfile.build.ubuntu_cpu` does not run it. So I think doing it in `ubuntu_tvm.sh` may not fix the error in unix-cpu.

Can I separate `ubuntu_python.sh` into two files, one for arm device and the other for common x86-64 device? This worked for me.

marcoabreu: How about you add Ubuntu tvm to the CPU dockerfile? I mean if we want to run tvm cpu tests, we should also install it, right?

No, please don't separate it.

hzfan: > How about you add Ubuntu tvm to the CPU dockerfile?

Yes, but I will also need a new file `ubuntu_tvm_cpu.sh` because `ubuntu_tvm.sh` installs tvm with cuda and cannot be run on cpu.

> I mean if we want to run tvm cpu tests, we should also install it, right?

No, it is not installed in ci. TVM is a 3rdparty in MXNet, so it is installed in cmake.

Will a new file like `ubuntu_tvm_cpu.sh` be acceptable?

hzfan: The dependency on psutil has been sidestepped without modifying ci. Thank you for your help.

szha:
@hzfan @yzhliu what's the status for this?

PR #15870, opened 2019-08-13
update support MKLDNN BN conditions

ElaineBao:
## Description ##
update BN conditions of when to run into MKLDNN.

## Changes ##
- [ ] support BN with arbitrary channel size
- [ ] support BN with different MKLDNN layout

## Performance ##
1. resnet50 v2 has one BN with shape (3 x 224 x 224), which cannot run into MKLDNN before changing the condition. Performance comparison before and after the change is as below.
2. When a model has fp32 bn along with other int8 ops, it triggers the IsMKLDNNData() check and cannot run int MKLDNN before changing the condition. Performance comparison is as below.

OP | Time (ms) |  Calls | Avg (ms/call)  
-|-|-|-
5000 MKLDNN BN (various shapes) + 100 native BN with shape (3x224x224) | 8127.24 | 5100 | 1.59
5000 MKLDNN BN (various shapes) + 100 MKLDNN BN with shape (3x224x224) | 6947.11 | 5100 | 1.36
fp32 native BN | 21909.83 | 2700 | 8.12
fp32 MKLDNN BN | 3794.58 | 2700 |  1.41

### DIFF
--- a/src/operator/nn/batch_norm.cc
+++ b/src/operator/nn/batch_norm.cc
@@ -384,7 +384,6 @@ static inline bool SupportMKLDNNBN(const NDArray &input, const BatchNormParam &p
   mxnet::TShape shape = input.shape();
   return SupportMKLDNN(input) && shape.ndim() == 4
       && param.axis == mxnet::op::batchnorm::DEFAULT_AXIS
-      && shape[param.axis] % 8 == 0
       && !mxnet::op::batchnorm::disable_mkl;
 }
 
@@ -396,7 +395,7 @@ void BatchNormComputeExCPU(const nnvm::NodeAttrs &attrs,
   CHECK_EQ(inputs.size(), 5U);
   const BatchNormParam &param = nnvm::get<BatchNormParam>(attrs.parsed);
   // MKLDNN batchnorm only works well on the special MKLDNN layout.
-  if (SupportMKLDNNBN(inputs[0], param) && inputs[0].IsMKLDNNData()) {
+  if (SupportMKLDNNBN(inputs[0], param)) {
     std::vector<NDArray> in_data(inputs.begin(), inputs.begin() + batchnorm::kInMovingMean);
     std::vector<NDArray> aux_states(inputs.begin() + batchnorm::kInMovingMean, inputs.end());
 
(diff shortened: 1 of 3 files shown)


pengzhao-intel:
Thanks for your contribution. Could you add a test case to cover the specific size of 8?

src/operator/nn/batch_norm.cc
```diff
@@ -420,8 +419,7 @@ void BatchNormGradComputeExCPU(const nnvm::NodeAttrs &attrs,
 
   mxnet::TShape shape = inputs[0].shape();
   // MKLDNN batchnorm only works well on the special MKLDNN layout.
-  if (SupportMKLDNNBN(inputs[0], param)
-      && (inputs[3].IsMKLDNNData() || inputs[0].IsMKLDNNData())) {
+  if (SupportMKLDNNBN(inputs[0], param)) {
```
pengzhao-intel: why we don't need to check these two conditions?

ElaineBao: It's the same reason as line 399, as described in Performance part, when a model has fp32 bn along with other int8 ops, it triggers the IsMKLDNNData() check and cannot run int MKLDNN, which cause a slower speed.

src/operator/nn/batch_norm.cc
```diff
@@ -396,7 +395,7 @@ void BatchNormComputeExCPU(const nnvm::NodeAttrs &attrs,
   CHECK_EQ(inputs.size(), 5U);
   const BatchNormParam &param = nnvm::get<BatchNormParam>(attrs.parsed);
   // MKLDNN batchnorm only works well on the special MKLDNN layout.
-  if (SupportMKLDNNBN(inputs[0], param) && inputs[0].IsMKLDNNData()) {
+  if (SupportMKLDNNBN(inputs[0], param)) {
     std::vector<NDArray> in_data(inputs.begin(), inputs.begin() + batchnorm::kInMovingMean);
```
pengzhao-intel: Please remove the comments in L397

pengzhao-intel:
Seems something wrong on CI or your code. Please rebase the code and try again.
@ZhennanQin could you help to take a look with @ElaineBao ?

ElaineBao:
Seems something wrong on CI, retrigger it again. If it's still failed, will look into it.

pengzhao-intel:
It's not easy to pass CI. Merging now and thanks for your contribution.

Issue #13958, opened 2019-01-22
Windows builds running out of heap space


KellenSunderland:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fwindows-cpu/detail/PR-13917/9/pipeline

c:\program files (x86)\microsoft visual studio 14.0\vc\include\xlocnum(1144) : fatal error C1002: compiler is out of heap space in pass 2

jom: C:\jenkins_slave\workspace\build-cpu\build\CMakeFiles\mxnet.dir\build.make [CMakeFiles\mxnet.dir\src\operator\tensor\elemwise_unary_op_basic.cc.obj] Error 1

jom: C:\jenkins_slave\workspace\build-cpu\build\CMakeFiles\Makefile2 [CMakeFiles\mxnet.dir\all] Error 2

jom: C:\jenkins_slave\workspace\build-cpu\build\Makefile [all] Error 2

frankfliu:
@mxnet-label-bot add [Windows, CI, build, CMake]

marcoabreu:
Interesting, haven't seen that one before. Considering we got ~300GB of Swap enabled on Windows instances, it surprises me that we are allegedly running out of memory. 

Do you have an idea what this could be caused by or if Swap is not applicable to that usecase?

Also, I found something:

> "Link Time Code Generation" should be set to "Profile Guided Optimization - Optimization (/LTCG:PGOptimize)" instead of being blank.

Source: https://social.msdn.microsoft.com/Forums/sqlserver/en-US/d2c4bb60-e558-4dc6-a0ba-47611d45bc86/c1002-compiler-is-out-of-heap-space-in-pass-2?forum=vcgeneral

stereomatchingkiss:
> Also, I found something:

I found that too, but can't solve the issue, you can refer to #14203

haojin2:
I'm running into this issue for #14359 too. Any updates?

stereomatchingkiss:
> 
> 
> I'm running into this issue for #14359 too. Any updates?

I would like to know too, maybe you could try 1.3.1, unless it works on windows. 
You could check [this blog](https://qtandopencv.blogspot.com/2019/03/build-mxnet-131-on-windows.html) if you want to compile 1.3.1 on windows, I have to warn you the journey of compile mxnet 1.3.1 on windows from source is not a pleasant one

haojin2:
[...]

PR #15770, opened 2019-08-06
[WIP] Reduce numerical error on numerical gradient calculations

larroy:
## Description ##

During numerical grad, the output symbol is multiplied by a random matrix. This increases the numerical errors dramatically. With this change, the gradient is still checked at the right location, but without the reduced precission. The benefit by adding a constant instead of multiplying is also that the gradient output in symbolic is always the same and the numerical one changes very little.

Fixes #11720
Overall will reduce flakiness of tests using numerical gradients

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

PR #15882, opened 2019-08-13
Move Windows CI build to a 64-bit toolchain to fix 'out of heap space'.

DickJC123:
## Description ##
'Out of heap space' errors having been plaguing our Windows CI builds for some time (https://github.com/apache/incubator-mxnet/issues/13958).  The errors appear to be a result of the limited virtual memory of the 32-bit version of cl.exe being used.  After some trial and error (due to not having direct access to the build machines) this PR upgrades the Windows compilation toolchain to 64-bit programs.  The PR leaves the compiler version at VS2015.  A VS2017 compiler appears to be installed onto the CI machines but cannot be used yet because compatible OpenCV libraries are not present.

I recommend a quick review due to the inconvenience of having these sporadic build failures.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/tests/python/unittest/test_random.py
+++ b/tests/python/unittest/test_random.py
@@ -893,14 +893,15 @@ def test_zipfian_generator():
 def test_shuffle():
     def check_first_axis_shuffle(arr):
         stride = int(arr.size / arr.shape[0])
-        column0 = arr.reshape((arr.size,))[::stride].sort()
+        column0 = arr.reshape((arr.size,))[::stride]
         seq = mx.nd.arange(0, arr.size - stride + 1, stride, ctx=arr.context)
-        assert (column0 == seq).prod() == 1
-        for i in range(arr.shape[0]):
-            subarr = arr[i].reshape((arr[i].size,))
-            start = subarr[0].asscalar()
-            seq = mx.nd.arange(start, start + stride, ctx=arr.context)
-            assert (subarr == seq).prod() == 1
+        assert (column0.sort() == seq).prod() == 1
+        # Check for ascending flattened-row sequences for 2D or greater inputs.
+        if stride > 1:
+            ascending_seq = mx.nd.arange(0, stride, ctx=arr.context)
+            equalized_columns = arr.reshape((arr.shape[0], stride)) - ascending_seq
+            column0_2d = column0.reshape((arr.shape[0],1))
+            assert (column0_2d == equalized_columns).prod() == 1
 
     # This tests that the shuffling is along the first axis with `repeat1` number of shufflings
     # and the outcomes are uniformly distributed with `repeat2` number of shufflings.
--- a/ci/build_windows.py
+++ b/ci/build_windows.py
@@ -148,7 +148,7 @@ def windows_build(args):
 
     with remember_cwd():
         os.chdir(path)
-        cmd = "\"{}\" && cmake -G \"NMake Makefiles JOM\" {} {}".format(args.vcvars,
+        cmd = "\"{}\" amd64 && cmake -G \"NMake Makefiles JOM\" {} {}".format(args.vcvars,
                                                                         CMAKE_FLAGS[args.flavour],
                                                                         mxnet_root)
         logging.info("Generating project with CMake:\n{}".format(cmd))
--- a/ci/build_windows.py
+++ b/ci/build_windows.py
@@ -36,7 +36,7 @@ from subprocess import check_call
 from util import *
 
 KNOWN_VCVARS = {
-    'VS 2015': r'C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\amd64\vcvarsamd64.bat',
+    'VS 2015': r'C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat',
     'VS 2017': r'C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars64.bat'
 }
 
@@ -154,7 +154,7 @@ def windows_build(args):
         logging.info("Generating project with CMake:\n{}".format(cmd))
         check_call(cmd, shell=True)
 
-        cmd = "\"{}\" && jom".format(args.vcvars)
+        cmd = "\"{}\" amd64 && jom".format(args.vcvars)
         logging.info("Building with jom:\n{}".format(cmd))
 
         t0 = int(time.time())
--- a/ci/build_windows.py
+++ b/ci/build_windows.py
@@ -220,7 +220,7 @@ def main():
 
     parser.add_argument("--vcvars",
         help="vcvars batch file location, typically inside vs studio install dir",
-        default=KNOWN_VCVARS['VS 2017'],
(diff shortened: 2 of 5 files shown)


DickJC123:
Normally I would not combine different fixes into one PR, but when fighting simultaneous CI flakiness, I feel it's justified.  With my latest commit, I've sped up test_random.test_shuffle to under 2 minutes to keep the CI from timing out at the 3 hour point.  Previously, I'd seen test_shuffle take from 10 - 50 minutes.  Not sure why the disparity- perhaps multiple CI runners are running on the same machine?   I did this purely in the Python test-result-checking part, and did not alter the loop-counts or logic of the test.  @asitstands was the last significant contributor to this test.

tests/python/unittest/test_random.py
```diff
@@ -893,14 +893,15 @@ def compute_expected_prob():
...
-        column0 = arr.reshape((arr.size,))[::stride].sort()
+        column0 = arr.reshape((arr.size,))[::stride]
         seq = mx.nd.arange(0, arr.size - stride + 1, stride, ctx=arr.context)
-        assert (column0 == seq).prod() == 1
-        for i in range(arr.shape[0]):
-            subarr = arr[i].reshape((arr[i].size,))
-            start = subarr[0].asscalar()
-            seq = mx.nd.arange(start, start + stride, ctx=arr.context)
-            assert (subarr == seq).prod() == 1
+        assert (column0.sort() == seq).prod() == 1
```
ChaiBapchya: curious to know what difference it makes by moving sort?

ChaiBapchya:
@apeforest @larroy @anirudh2290 Would be great to get this reviewed quickly and merged if it helps speed up the CI!

tests/python/unittest/test_random.py
DickJC123: There are no random shapes in test_shuffle so its runtime, even give the sort, should be fairly consistent.  My only explanation for the runtime variation is that maybe there's multiple cpu runners on the same machine.  Looking for confirmation from @marcoabreu.

Moving the sort was just a style preference, not performance driven.  The biggest runtime savings came from introducing the `if stride > 1:` clause and avoiding needless work for the last big 1D test input.  The rest was just some further perf polishing.

I keep stumbling on flakey tests with this PR, so my last commit is fixing laop_6.  A bit of bad luck I would say.

marcoabreu: we run up to 4 jobs per windows-cpu machine. 1 job per windows-gpu machine.

ptrendx:
@marcoabreu That is not 100% true though - for elementwise ops that is the case and I fully agree we should split them, but for most operators there is 1 file per operator. The issue in those cases comes from the fact that we need to support many combinations input/output types (and things like accumulation types in some cases) which ends up in a long list of templates to instantiate.

marcoabreu:
If we reduce/split elementwise ops, is the problem solved then?

ptrendx:
No, it is not. As I said, the number of templates that we need to generate in some other cases is a problem - that was the problem with softmax for example, which prompted e.g. this: https://github.com/apache/incubator-mxnet/blob/master/src/operator/tensor/broadcast_reduce-inl.cuh#L620 which is bad, since it actually makes Windows build different in terms of functionality than other builds.

larroy:
@marcoabreu let's unblock CI and discuss what you propose later. Blocking a critical change to unblock CI and development takes priority.

ChaiBapchya:
Yes please. I've been retriggering my 5 PRs for last 2 days (atleast 2-3 times per PR)
Would be great if we address this time-out issue!

marcoabreu:
Hi, there might be a misunderstanding. The issue here is not the cause for the timeouts.

DickJC123:
@larroy I looked again at the CI failure of laop_6: http://jenkins.mxnet-ci.amazon-ml.com/blue/rest/organizations/jenkins/pipelines/mxnet-validation/pipelines/windows-gpu/branches/PR-15882/runs/10/nodes/116/steps/186/log/?start=0.  It appears to be unrelated to tolerances, so I reverted my change to that test.

My change to test_shuffle was unrelated to tolerances- just sped up the checking code without altering any functionality.

larroy:
@DickJC123 I looked at laop tests, one root cause is the multiplication by a "random projection" done in the test, this initial multiplication creates an introduction of fp error. I had a look at this and tried to address, but I then had failures in other tests and run ouf of time allocated for this.

https://github.com/apache/incubator-mxnet/pull/15770

larroy:
Thanks for the clarification. I already approved.

larroy:
Marco. I understand this is a bug in the vs compiler version we use. I don't think this deprecates 32 bit support as the target arch doesn't change. We are compiling for 32 bit in arm7 right?

marcoabreu:
We had the same issue with clang if I recall correctly. So I'd say that this is not a bug but rather a symptom.

marcoabreu:
Can we get rid of the compiler upgrade for now and observe if everything is fine now? For the other improvements I'd be happy to get them in. You can also just put them in a separate PR.

larroy:
Why is this PR stuck?

DickJC123:
This PR was created at a time when there were multiple CI issues (OOM on Windows toolchain, CI timeouts due to test_shuffle, and laop_6 flakeyness).  This PR tried to solve all 3 problems in order to get to a passing/mergeable state.  Since all of the issues have been solved in a fashion, I'm closing this PR.

The approach this PR took to move the Windows CI to the 64-bit toolchain may be helpful as a reference in the future.  Instead of adopting this PR though, we broke up some of the larger compiles and stayed on the 32-bit windows toolchain.  The way I look at it, by staying on the 32-bit toolchain, we are somewhat artificially forcing developers to keep their compilation units small.  This is not a bad thing, since it helps a multi-threaded compile to finish more quickly in the presence of errors.  @ptrendx points out that with some valid templating approaches, reducing compile size may not be always possible.

If we run up against a Windows compile that we don't want to work around by breaking up the file, ping me to dust off and resubmit this PR.

marcoabreu:
Happy to move forward with the upgrade to 64bit

DickJC123:
@marcoabreu Sounds like if I resubmitted the core of this PR, you'd support it.  Anything specific behind your shifted stance w.r.t. windows 64-bit toolchain?  I know we went ahead and broke up some major compilation units, which you were pushing for.

marcoabreu:
It seems like I was under the impression that we are dropping support of some visual studio version, but that seemed like a fluke. It's fine now, sorry for the confusion. Happy to move forward with the upgrade to 64bit

larroy:
If we want to monitor binary sizes, a good way to do it would be to put a safeguard on CI that checks that separately.

ChaiBapchya:
> It seems like I was under the impression that we are dropping support of some visual studio version, but that seemed like a fluke. It's fine now, sorry for the confusion. Happy to move forward with the upgrade to 64bit

Are you referring to this PR - https://github.com/apache/incubator-mxnet/pull/16712
In order to upgrade VC2015 -> 2017, need opencv to be updated

This PR for opencv update is failing due to "authorization issue" (Hence I am stuck)
[Restricted access] https://github.com/anirudh2290/mxnet-ci/pull/2

PR #15886, opened 2019-08-14
Graph Partition API

mseth10:
## Description ##
This PR introduces enhanced Python and C APIs to trigger graph partitioning for a given backend. These APIs provide an option to the user to infer shapes, dtypes and stypes before partitioning. They add the ability to pass optional arguments to subgraph property.

This PR makes the following modifications to subgraph property:
- Added the ability for the subgraph property to preprocess the graph and initialize any custom state that is required during partitioning
- Added the ability for the subgraph property to post-process the graph and do things like collect statistics on the quality of partitioning.
- Enabled subgraph property to reject creating a sub-optimal subgraph. For example, the subgraph property could reject subgraphs with only one operator in it.

We reuse the existing graph partitioning algorithm and control using subgraph property classes. Since we’re modifying a symbol object, we add this API to the Symbol class. Users will call `optimize_for` on an existing symbol object to create a new partitioned symbol.

Here is an example showing partitioning without args, which means infer shape/type will NOT be called before partitioning.  It is similar to `GetBackendSymbol`.
```
import mxnet as mx
sym, args, aux = mx.model.load_checkpoint('resnet-50', 0)
sym = sym.optimize_for('default')
exe = sym.bind(ctx=mx.cpu(), args=args, aux_states=aux, grad_req='null')
```

Here is an example showing partitioning with args, which means infer shape/type is called before partitioning.
```
import mxnet as mx
sym, args, aux = mx.model.load_checkpoint('resnet-50', 0)
sym = sym.optimize_for('default', args)
exe = sym.bind(ctx=mx.cpu(), args=args, aux_states=aux, grad_req='null')
```

Here is an example showing partitioning with kwargs to set a list of ops to be excluded from subgraphs.
```
import mxnet as mx
sym, args, aux = mx.model.load_checkpoint('resnet-50', 0)
sym = sym.optimize_for('default', args, excluded_ops=['BatchNorm'])
exe = sym.bind(ctx=mx.cpu(), args=args, aux_states=aux, grad_req='null')
```

Next steps:
- Add example(s) and test(s) for `excluded_ops` passed as arguments to `optimize_for` as part of the PR on support for custom subgraph properties
- Add functionality for users to pass shape/type/stype dicts as kwargs to `optimize_for` just like we have it in `simple_bind`

### Essentials ###

### DIFF
--- a/python/mxnet/symbol/symbol.py
+++ b/python/mxnet/symbol/symbol.py
@@ -1437,9 +1437,39 @@ class Symbol(SymbolBase):
         return Symbol(handle)
 
 
-    def optimize_for(self, backend, ctx=None, args=None, **kwargs):
-        """Partition symbol and optimize it for a given backend"""
+    def optimize_for(self, backend, args=None, ctx=None, **kwargs):
+        """Partition symbol and optimize it for a given backend
+
+        Parameters
+        ----------
+        backend : str
+            The name of backend, as registered in `SubgraphBackendRegistry`
+
+        args : list of NDArray or dict of str to NDArray, optional
+            Input arguments to the symbol.
+            - If type is a list of `NDArray`, the order is the same as that of `list_arguments()`.
+            - If type is a dict of str to `NDArray`, then it maps the name of arguments
+              to the corresponding `NDArray`.
+
+        ctx : Context, optional
+            Device context. Used to infer stypes.
+
+        kwargs : Optional arguments
+            Passed on to `PrePartition` and `PostPartition` functions of `SubgraphProperty`
+
+        Returns
+        -------
+        out : SymbolHandle
+            The created symbol for target backend.
+        """
         out = SymbolHandle()
+        assert isinstance(backend, str)
+        if args is None:
+            args = []
+            args_handle = []
+        else:
+            listed_arguments = self.list_arguments()
+            args_handle, args = self._get_ndarray_inputs('args', args, listed_arguments, False)
         if ctx is None:
             dev_type = -1
             dev_id = 0
@@ -1448,12 +1478,6 @@ class Symbol(SymbolBase):
                 raise TypeError("Context type error")
             dev_type = ctx.device_typeid
             dev_id = ctx.device_id
-        if args is None:
-            args = []
-            args_handle = []
-        else:
-            listed_arguments = self.list_arguments()
-            args_handle, args = self._get_ndarray_inputs('args', args, listed_arguments, False)
         key_list = []
         val_list = []
         if kwargs is not None:
--- a/python/mxnet/symbol/symbol.py
+++ b/python/mxnet/symbol/symbol.py
@@ -1437,7 +1437,7 @@ class Symbol(SymbolBase):
         return Symbol(handle)
 
 
-    def optimizeFor(self, backend, ctx=None, args=None, **kwargs):
+    def optimize_for(self, backend, ctx=None, args=None, **kwargs):
         """Partition symbol and optimize it for a given backend"""
         if ctx is not None and args is not None):
             infer_shape_type = False
@@ -1447,13 +1447,6 @@ class Symbol(SymbolBase):
                 raise TypeError("Context type error")
             listed_arguments = self.list_arguments()
             args_handle, args = self._get_ndarray_inputs('args', args, listed_arguments, False)
-            aux_states = []
-            aux_args_handle, aux_states = self._get_ndarray_inputs(
-                'aux_states', aux_states, self.list_auxiliary_states(), False)
-            ctx_map_keys = []
-            ctx_map_dev_types = []
-            ctx_map_dev_ids = []
-
         out = SymbolHandle()
         check_call(_LIB.MXOptimizeForBackend(self.handle,
                                              c_str(backend),
@@ -1461,14 +1454,8 @@ class Symbol(SymbolBase):
                                              ctypes.c_bool(infer_shape_type)
                                              ctypes.c_int(ctx.device_typeid),
                                              ctypes.c_int(ctx.device_id),
-                                             mx_uint(len(ctx_map_keys)),
-                                             c_str_array(ctx_map_keys),
-                                             c_array_buf(ctypes.c_int, array('i', ctx_map_dev_types)),
-                                             c_array_buf(ctypes.c_int, array('i', ctx_map_dev_ids)),
                                              mx_uint(len(args)),
-                                             args_handle,
-                                             mx_uint(len(aux_states)),
-                                             aux_args_handle))
+                                             args_handle)
         return out
 
 
--- a/python/mxnet/symbol/symbol.py
+++ b/python/mxnet/symbol/symbol.py
@@ -1437,6 +1437,12 @@ class Symbol(SymbolBase):
         return Symbol(handle)
 
 
+    def optimizeFor(self, backend):
+        """Partition symbol and optimize it for a given backend"""
+        out = SymbolHandle()
+        check_call(_LIB.MXOptimizeForBackend(self.handle, c_str(backend), ctypes.byref(out)))
+
+
     # pylint: disable=too-many-locals
     def simple_bind(self, ctx, grad_req='write', type_dict=None, stype_dict=None,
                     group2ctx=None, shared_arg_names=None, shared_exec=None,
--- a/python/mxnet/symbol/symbol.py
+++ b/python/mxnet/symbol/symbol.py
@@ -1440,26 +1440,31 @@ class Symbol(SymbolBase):
     def optimize_for(self, backend, ctx=None, args=None, **kwargs):
         """Partition symbol and optimize it for a given backend"""
         out = SymbolHandle()
-        if ctx is None or args is None:
-            infer_shape_type = False
+        if ctx is None:
+            dev_type = -1
+            dev_id = 0
         else:
-            infer_shape_type = True
             if not isinstance(ctx, Context):
                 raise TypeError("Context type error")
+            dev_type = ctx.device_typeid
+            dev_id = ctx.device_id
+        if args is None:
+            args = []
+            args_handle = []
+        else:
             listed_arguments = self.list_arguments()
             args_handle, args = self._get_ndarray_inputs('args', args, listed_arguments, False)
         key_list = []
         val_list = []
         if kwargs is not None:
-            for key, value in kwargs.iteritems():
+            for key, val in kwargs.iteritems():
                 key_list.append(key)
                 val_list.append(val)
         check_call(_LIB.MXOptimizeForBackend(self.handle,
                                              c_str(backend),
                                              ctypes.byref(out),
-                                             ctypes.c_bool(infer_shape_type),
-                                             ctypes.c_int(ctx.device_typeid),
-                                             ctypes.c_int(ctx.device_id),
+                                             ctypes.c_int(dev_type),
+                                             ctypes.c_int(dev_id),
                                              mx_uint(len(args)),
                                              args_handle,
                                              mx_uint(len(key_list)),
--- a/python/mxnet/symbol/symbol.py
+++ b/python/mxnet/symbol/symbol.py
@@ -1437,7 +1437,7 @@ class Symbol(SymbolBase):
         return Symbol(handle)
 
 
-    def optimize_for(self, backend, args=None, **kwargs):
+    def optimize_for(self, backend, ctx=None, args=None, **kwargs):
         """Partitions current symbol and optimizes it for a given backend,
         returns new partitioned symbol.
 
@@ -1463,6 +1463,11 @@ class Symbol(SymbolBase):
         """
         out = SymbolHandle()
         assert isinstance(backend, str)
+
+        if ctx is None:
+            ctx = current_context()
+        assert isinstance(ctx,Context)
+        
         if args is None:
             args = []
             args_handle = c_array(NDArrayHandle, [])
@@ -1476,6 +1481,7 @@ class Symbol(SymbolBase):
             val_list.append(str(val))
         check_call(_LIB.MXOptimizeForBackend(self.handle,
                                              c_str(backend),
+                                             ctypes.c_int(ctx.device_typeid),
                                              ctypes.byref(out),
                                              mx_uint(len(args)),
                                              args_handle,
--- a/python/mxnet/symbol/symbol.py
+++ b/python/mxnet/symbol/symbol.py
@@ -1463,25 +1463,24 @@ class Symbol(SymbolBase):
         assert isinstance(backend, str)
         if args is None:
             args = []
-            args_handle = []
+            args_handle = c_array(NDArrayHandle, [])
         else:
             listed_arguments = self.list_arguments()
             args_handle, args = self._get_ndarray_inputs('args', args, listed_arguments, False)
         key_list = []
         val_list = []
-        if kwargs is not None:
-            for key, val in kwargs.iteritems():
-                key_list.append(key)
-                val_list.append(val)
+        for key, val in kwargs.items():
+            key_list.append(key)
+            val_list.append(str(val))
         check_call(_LIB.MXOptimizeForBackend(self.handle,
                                              c_str(backend),
                                              ctypes.byref(out),
                                              mx_uint(len(args)),
                                              args_handle,
                                              mx_uint(len(key_list)),
-                                             c_str_array([key for key in key_list]),
-                                             c_str_array([str(val) for val in val_list])))
-        return out
+                                             c_str_array(key_list),
+                                             c_str_array(val_list)))
+        return Symbol(out)
 
 
     # pylint: disable=too-many-locals
--- a/python/mxnet/symbol/symbol.py
+++ b/python/mxnet/symbol/symbol.py
@@ -1438,7 +1438,8 @@ class Symbol(SymbolBase):
 
 
     def optimize_for(self, backend, args=None, **kwargs):
-        """Partition symbol and optimize it for a given backend
+        """Partitions current symbol and optimizes it for a given backend,
+        returns new partitioned symbol.
 
         Parameters
         ----------
@@ -1446,12 +1447,13 @@ class Symbol(SymbolBase):
             The name of backend, as registered in `SubgraphBackendRegistry`
 
         args : list of NDArray or dict of str to NDArray, optional
-            Input arguments to the symbol.
+            Input arguments to the symbol, required to infer shapes/types before partitioning
+
             - If type is a list of `NDArray`, the order is the same as that of `list_arguments()`.
             - If type is a dict of str to `NDArray`, then it maps the name of arguments
               to the corresponding `NDArray`.
 
-        kwargs : Optional arguments
+        kwargs : optional arguments
             Passed on to `PrePartition` and `PostPartition` functions of `SubgraphProperty`
 
         Returns
--- a/python/mxnet/symbol/symbol.py
+++ b/python/mxnet/symbol/symbol.py
@@ -1446,15 +1446,15 @@ class Symbol(SymbolBase):
         backend : str
             The name of backend, as registered in `SubgraphBackendRegistry`
 
-        ctx : Context, optional
-            Device context, used to infer stypes
-
         args : list of NDArray or dict of str to NDArray, optional
             Input arguments to the symbol, required to infer shapes/types before partitioning
 
             - If type is a list of `NDArray`, the order is the same as that of `list_arguments()`.
             - If type is a dict of str to `NDArray`, then it maps the name of arguments
               to the corresponding `NDArray`.
+
+        ctx : Context, optional
+            Device context, used to infer stypes
 
         kwargs : optional arguments
             Passed on to `PrePartition` and `PostPartition` functions of `SubgraphProperty`
@@ -1467,16 +1467,17 @@ class Symbol(SymbolBase):
         out = SymbolHandle()
         assert isinstance(backend, str)
 
-        if ctx is None:
-            ctx = current_context()
-
-        assert isinstance(ctx, Context)
         if args is None:
             args = []
             args_handle = c_array(NDArrayHandle, [])
         else:
             listed_arguments = self.list_arguments()
             args_handle, args = self._get_ndarray_inputs('args', args, listed_arguments, False)
+
+        if ctx is None:
+            ctx = current_context()
+        assert isinstance(ctx, Context)
+
         key_list = []
         val_list = []
         for key, val in kwargs.items():
--- a/python/mxnet/symbol/symbol.py
+++ b/python/mxnet/symbol/symbol.py
@@ -1440,7 +1440,7 @@ class Symbol(SymbolBase):
     def optimize_for(self, backend, ctx=None, args=None, **kwargs):
         """Partition symbol and optimize it for a given backend"""
         out = SymbolHandle()
-        if ctx is not None and args is not None):
+        if ctx is None or args is None:
             infer_shape_type = False
         else:
             infer_shape_type = True
@@ -1457,14 +1457,14 @@ class Symbol(SymbolBase):
         check_call(_LIB.MXOptimizeForBackend(self.handle,
                                              c_str(backend),
                                              ctypes.byref(out),
-                                             ctypes.c_bool(infer_shape_type)
+                                             ctypes.c_bool(infer_shape_type),
                                              ctypes.c_int(ctx.device_typeid),
                                              ctypes.c_int(ctx.device_id),
                                              mx_uint(len(args)),
                                              args_handle,
                                              mx_uint(len(key_list)),
                                              c_str_array([key for key in key_list]),
-                                             c_str_array([str(val) for val in val_list]))
+                                             c_str_array([str(val) for val in val_list])))
         return out
 
 
--- a/python/mxnet/symbol/symbol.py
+++ b/python/mxnet/symbol/symbol.py
@@ -1437,7 +1437,7 @@ class Symbol(SymbolBase):
         return Symbol(handle)
 
 
-    def optimize_for(self, backend, ctx=None, args=None, **kwargs):
+    def optimize_for(self, backend, args=None, ctx=None, **kwargs):
         """Partitions current symbol and optimizes it for a given backend,
         returns new partitioned symbol.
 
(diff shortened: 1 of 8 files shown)


pengzhao-intel:
@ZhennanQin for review the compatible  :)

ZhennanQin:
What's the difference between `sym.partition('default')` and `sym.get_backend_symbol('default')`?

python/mxnet/symbol/symbol.py
```diff
@@ -1437,6 +1437,13 @@ def _gen_atomic_symbol(self):
         return Symbol(handle)
 
 
+    def partition(self, name):
```
samskalicky: - rename to "optimizeFor"
- return a copy of the modified input symbol

mseth10: Done!

samskalicky:
@ZhennanQin Here in this PR we're trying to do the following things:

1. Clarify the API/function-names. Initially we wanted the API to be called "sym.partition" since we were going to partition the symbol. But given the latest PR that enables multiple subgraph properties to be specified we dropped this idea. Instead, to users the outcome is that they are "optimizing the symbol, for a specific backend". So now we propose the API to be called "optimizeFor" and it will take the name of the backend. So this way it will be much more understandable to users that they are "optimizing the symbol for a specific backend". For example: sym.optimizeFor("TensorRT") or sym.optimizeFor("MKLDNN") or sym.optimizeFor("EIA"). We expect that users could call this more than once for the same model. For example, users could first optimize for EIA (that would group supported nodes into subgraphs) and then optimize for MKLDNN for the remaining nodes that were not partitioned into subgraphs.

2. We also want to enable users to pass more configuration options directly to the subgraph property to tailor the partitioning. One example is to provide a better experience for users to provide a blacklist (or list of operators to exclude from subgraphs). Another could be setting the minimum number of nodes allowed in a subgraph (ie. 3). 

3. Enable the API to optionally perform shape/type propagation before partitioning. For example, some backends may only support operators if their inputs are a particular type (ie. FP16 or Int8) or shape (2D, 3D, or 4D). 

4. Currently the subgraph properties are state-less. We want to enable the subgraph properties to have a prepartition and postpartition APIs to do setup or cleanup around the actual partitioning call. This could be used to setup a whitelist of supported operators, or pre-analyze the graph before partitioning begins and Select is called on the Selector. 

In this PR we want to enable this functionality for users to more easily use partitioning. In a later PR we want to enable subgraph properties to be dynamically loaded from libraries.

python/mxnet/symbol/symbol.py
```diff
@@ -1437,6 +1437,12 @@ def _gen_atomic_symbol(self):
         return Symbol(handle)
 
 
+    def optimizeFor(self, backend):
```
samskalicky: optional arguments:
- ctx, args

use bind as an example: https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/symbol/symbol.py#L1705
https://github.com/apache/incubator-mxnet/blob/05f3ae13a39835c4690eb5d9567f457be84189e3/src/c_api/c_api_executor.cc#L135

src/operator/subgraph/subgraph_property.h
```diff
@@ -221,6 +221,10 @@ class SubgraphProperty {
     return nullptr;
   }
 
+  virtual void PrePartition() {}
```
samskalicky: should we pass the graph to Pre/PostPartition functions?

we also need to pass the map of "options" to the subgraph prop somehow

ZhennanQin:
@samskalicky Thanks for the explanation, it helps a lot to understand this PR. Basically, I think `sym.optimizeFor('MKLDNN')` is the same as current `sym.get_backend_symbol('MKLDNN')`, so I don't see the reason to duplicate it. Maybe we can directly rename `get_backend_symbol` to `optimizeFor`?

samskalicky:
Basically, I think sym.optimizeFor('MKLDNN') is the same as current sym.get_backend_symbol('MKLDNN'), so I don't see the reason to duplicate it. Maybe we can directly rename get_backend_symbol to optimizeFor?

@ZhennanQin Thanks for your comments, we'll try and clarify our thought process. Apologies for the lack of clarity in this PR. You are correct in the current status of the code that @mseth10 has committed so far today. The PR is still WIP and more changes are coming, lets not make decisions on what is there today. 

Instead lets focus on the items i mentioned before (they will get into the PR description before we attempt to merge, I promise! :-D). We will add more arguments to the optimizeFor API in the coming commits. We need to add arguments to enable us to to do shape/type propagation, and we want to accept arbitrary options that we pass to the subgraph property. 

These new API changes will not be compatible with the current get_backend_symbol API.

samskalicky:
@mseth10 we also need to add the ability to reject creating a subgraph in the buildsubgraph.cc file:

https://github.com/apache/incubator-mxnet/blob/e98fea3165670157090f2a2f644890452443803c/src/operator/subgraph/build_subgraph.cc#L574

We want to enable the subgraph property to return a null node to say reject

python/mxnet/symbol/symbol.py
```diff
@@ -1437,6 +1437,41 @@ def _gen_atomic_symbol(self):
         return Symbol(handle)
 
 
+    def optimizeFor(self, backend, ctx=None, args=None, **kwargs):
```
rondogency: should be optimize_for, since we don't do camelCase in api functions

python/mxnet/symbol/symbol.py
mseth10: added optional arguments

src/operator/subgraph/subgraph_property.h
mseth10: added graph and options_map as parameters

python/mxnet/symbol/symbol.py
mseth10: changed function name

src/c_api/c_api_symbolic.cc
```diff
@@ -1181,3 +1182,59 @@ int MXShallowCopySymbol(SymbolHandle src, SymbolHandle* out) {
...
+  }
+  std::vector<std::pair<std::string, std::string>> options_map;
+  for (mx_uint i = 0; i < num_options; ++i) {
+    options_map.emplace_back(keys[i], vals[i]);
+  }
+  const auto backend = mxnet::op::SubgraphBackendRegistry::Get()->GetSubgraphBackend(backend_name);
+  const auto& subgraph_prop_list = backend->GetSubgraphProperties();
+  for (auto property : subgraph_prop_list) {
+    nnvm::Graph g = orig_g;
+    property->SetAttr("graph", g);
```
samskalicky: dont think we need this if we're going to pass it into the prePartition function

src/c_api/c_api_symbolic.cc
```diff
@@ -1181,3 +1182,59 @@ int MXShallowCopySymbol(SymbolHandle src, SymbolHandle* out) {
...
+    property->PrePartition(g, options_map);
+    g.attrs["subgraph_property"] = std::make_shared<nnvm::any>(property);
+    g = ApplyPass(std::move(g), "BuildSubgraph");
+    g.attrs.erase("subgraph_property");
+    property->PostPartition(g, options_map);
+    property->RemoveAttr("graph");
```
samskalicky: if we get rid of set, remove this one too

python/mxnet/symbol/symbol.py
```diff
@@ -1437,6 +1437,37 @@ def _gen_atomic_symbol(self):
...
 
+    def optimize_for(self, backend, ctx=None, args=None, **kwargs):
+        """Partition symbol and optimize it for a given backend"""
+        out = SymbolHandle()
+        if ctx is not None and args is not None):
+            infer_shape_type = False
+        else:
+            infer_shape_type = True
+            if not isinstance(ctx, Context):
+                raise TypeError("Context type error")
```
samskalicky: if you raise an exception here, that means ctx is required argument

python/mxnet/symbol/symbol.py
```diff
@@ -1437,6 +1437,37 @@ def _gen_atomic_symbol(self):
         return Symbol(handle)
 
 
+    def optimize_for(self, backend, ctx=None, args=None, **kwargs):
+        """Partition symbol and optimize it for a given backend"""
+        out = SymbolHandle()
+        if ctx is not None and args is not None):
```
samskalicky: should this be OR?

src/common/exec_utils.h
```diff
@@ -621,6 +621,42 @@ inline nnvm::Graph AssignContext(nnvm::Graph g,
...
+                                     StorageTypeVector arg_stypes,
+                                     const Context& default_ctx,
+                                     const std::map<std::string, Context>& ctx_map,
+                                     const std::vector<Context>& in_arg_ctxes,
+                                     const std::vector<Context>& aux_state_ctxes,
+                                     bool partial_shape = false) {
+  const auto& indexed_graph = g.indexed_graph();
+  const auto num_forward_inputs = indexed_graph.input_nodes().size();
+  g = AssignContext(g, default_ctx, ctx_map, in_arg_ctxes, {},
+                   aux_state_ctxes, {}, num_forward_inputs, g.outputs.size());
```
samskalicky: this is the only consumer of default_ctx, is it really necessary to just do the shape/type prop that we're looking for pre-partitioning?

mseth10: AssignContext sets 'context' attribute to graph
https://github.com/apache/incubator-mxnet/blob/master/src/common/exec_utils.h#L512

which is used to set 'dev_mask' graph attribute later in InferStorageType
https://github.com/apache/incubator-mxnet/blob/master/src/executor/infer_graph_attr_pass.cc#L713-L719

ZhennanQin:
@samskalicky @mseth10 As this PR is still WIP, please ping me when this PR is ready to review. I agree with you that we need a more powerful API to put dtype, shape and ctx into consideration to make partition more accurate.

src/common/exec_utils.h
samskalicky: @mseth10 do we need to run InferStorageType? or can we skip it for the purpose of partitioning?

python/mxnet/symbol/symbol.py
mseth10: Modified the condition

src/c_api/c_api_symbolic.cc
mseth10: Graph attribute not used anywhere, removing it.

src/c_api/c_api_symbolic.cc
mseth10: Removed.

python/mxnet/symbol/symbol.py
```diff
@@ -1437,6 +1437,37 @@ def _gen_atomic_symbol(self):
...
+        """Partition symbol and optimize it for a given backend"""
+        out = SymbolHandle()
+        if ctx is None or args is None:
+            infer_shape_type = False
+        else:
+            infer_shape_type = True
+            if not isinstance(ctx, Context):
+                raise TypeError("Context type error")
+            listed_arguments = self.list_arguments()
+            args_handle, args = self._get_ndarray_inputs('args', args, listed_arguments, False)
```
samskalicky: are we overwriting "args" here thats the input to the optimize_for function?

mseth10: It's suppose to be that way. Similar to: https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/symbol/symbol.py#L1799

src/common/exec_utils.h
mseth10: As discussed offline, we need InferStorageType, which needs context.

tests/python/unittest/test_subgraph_op.py
```diff
@@ -146,11 +146,135 @@ def get_executor(sym, subgraph_backend=None, op_names=None, original_exec=None):
...
+                assert name in exe1.aux_dict
+                exe1.aux_dict[name][:] = mx.nd.random.uniform(shape=exe1.aux_dict[name].shape)
+        exe1.forward()
+
+        # partition before simple_bind
+        check_call(_LIB.MXSetSubgraphPropertyOpNames(c_str(subgraph_backend), mx_uint(len(op_names)),
+                                                     c_str_array(op_names)))
+        part_sym = sym.optimize_for(subgraph_backend)
+        check_call(_LIB.MXRemoveSubgraphPropertyOpNames(c_str(subgraph_backend)))
+        exe2 = part_sym.simple_bind(ctx=mx.current_context(), grad_req='null')
```
samskalicky: add a new line between removeSubgraphProp and bind

tests/python/unittest/test_subgraph_op.py
```diff
@@ -146,11 +146,135 @@ def get_executor(sym, subgraph_backend=None, op_names=None, original_exec=None):
...
+        for name in input_names:
```
samskalicky: What is this loop name checking looking for? Can we add a newline and a comment?

tests/python/unittest/test_subgraph_op.py
mseth10: Done

src/operator/subgraph/subgraph_property.h
```diff
@@ -221,6 +222,12 @@ class SubgraphProperty {
     return nullptr;
   }
 
+  virtual void PrePartition(const nnvm::Graph& g,
+    const std::vector<std::pair<std::string, std::string>>& options_map) {}
+
+  virtual void PostPartition(const nnvm::Graph& g,
+    const std::vector<std::pair<std::string, std::string>>& options_map) {}
```
samskalicky: Dont think we need options_map in the postPartition args list. Any options that need to be saved can be saved in the SubgraphProperty Class.

mseth10:
@ZhennanQin ping for review

src/operator/subgraph/subgraph_property.h
mseth10: Removed options_map from PostPartition

tests/python/unittest/test_subgraph_op.py
mseth10: The loop assigns random inputs to executor generated using simple_bind. Both executors share the same set of inputs. Refactored the code and added comments.

tests/python/unittest/test_subgraph_op.py
```diff
@@ -146,11 +146,137 @@ def get_executor(sym, subgraph_backend=None, op_names=None, original_exec=None):
...
+
     def check_subgraph_exe(sym, subgraph_backend, op_names):
         _check_subgraph_exe1(sym, subgraph_backend, op_names)
         _check_subgraph_exe2(sym, subgraph_backend, op_names)
         _check_subgraph_exe3(sym, subgraph_backend, op_names)
         _check_subgraph_exe4(sym, subgraph_backend, op_names)
+        _check_subgraph_exe5(sym, subgraph_backend, op_names)
+        _check_subgraph_exe6(sym, subgraph_backend, op_names)
+        _check_subgraph_exe7(sym, subgraph_backend, op_names)
+        _check_subgraph_exe8(sym, subgraph_backend, op_names)
```
rondogency: should we separate those tests into a separate test file? since the previous tests are testing the 'subgraph building' but the new tests are testing partition api

samskalicky: Its the same test. subgraph building is partitioning.

src/c_api/c_api_symbolic.cc
```diff
@@ -1199,3 +1200,73 @@ int MXShallowCopySymbol(SymbolHandle src, SymbolHandle* out) {
...
+  }
+  const auto backend = mxnet::op::SubgraphBackendRegistry::Get()->GetSubgraphBackend(backend_name);
+  const auto& subgraph_prop_list = backend->GetSubgraphProperties();
+  for (auto property : subgraph_prop_list) {
+    property->PrePartition(g, options_map);
+    g.attrs["subgraph_property"] = std::make_shared<nnvm::any>(property);
+    g = ApplyPass(std::move(g), "BuildSubgraph");
+    g.attrs.erase("subgraph_property");
+    property->PostPartition(g);
+    s->outputs = g.outputs;
```
rondogency: can we move this outside the loop?

samskalicky: we're looping over subgraph_prop_list, and for each property we call `g.attrs["subgraph_property"] = std::make_shared<nnvm::any>(property);`

samskalicky: Sorry @rondogency, i retract that statement above. Lets move the output assignment outside of the loop

mseth10: Agreed. I will make this change.

mseth10: Done.

src/c_api/c_api_symbolic.cc
```diff
@@ -1199,3 +1200,73 @@ int MXShallowCopySymbol(SymbolHandle src, SymbolHandle* out) {
...
+          << "args[" << i << "] is on context: " << in_arg.ctx()
+          << ", whereas args[0] is on context: " << default_ctx
+          << ". All args must be on the same context.";
+    }
+    const auto& indexed_graph = g.indexed_graph();
+    const auto num_forward_inputs = indexed_graph.input_nodes().size();
+    g.attrs["context"] = std::make_shared<nnvm::any>(
+        exec::ContextVector(indexed_graph.num_nodes(), default_ctx));
+    // infer shapes
+    g = exec::InferShape(std::move(g), std::move(arg_shapes), "__shape__");
```
ZhennanQin: Shall we call `InferForwardAttrs` instead? Then we can maintain single copy of this functionality.
https://github.com/apache/incubator-mxnet/blob/d8b6e47b91a737d29c0589bda8640aee69d8599b/src/executor/graph_executor.cc#L1606

src/c_api/c_api_symbolic.cc
```diff
@@ -1199,3 +1200,73 @@ int MXShallowCopySymbol(SymbolHandle src, SymbolHandle* out) {
...
+                         const char** keys,
+                         const char** vals) {
+  nnvm::Symbol *s = new nnvm::Symbol();
+  API_BEGIN();
+  nnvm::Symbol *sym = static_cast<nnvm::Symbol *>(sym_handle);
+  *s = sym->Copy();
+  nnvm::Graph g = Symbol2Graph(*s);
+  if (len) {
+    NDArray **in_args_ptr = reinterpret_cast<NDArray**>(in_args_handle);
+    Context default_ctx = in_args_ptr[0]->ctx();
```
ZhennanQin: Why we use this ctx? User may bind new ctx which is different with this, like the code you provided:
```
import mxnet as mx
sym, args, aux = mx.model.load_checkpoint('resnet-50', 0)
sym = sym.optimize_for('default', args)
exe = sym.bind(ctx=mx.cpu(), args=args, aux_states=aux, grad_req='null')
```
How to deal with default_ctx != mx.cpu()?

src/c_api/c_api_symbolic.cc
```diff
@@ -1199,3 +1200,73 @@ int MXShallowCopySymbol(SymbolHandle src, SymbolHandle* out) {
...
+          << ", whereas args[0] is on context: " << default_ctx
+          << ". All args must be on the same context.";
+    }
+    const auto& indexed_graph = g.indexed_graph();
+    const auto num_forward_inputs = indexed_graph.input_nodes().size();
+    g.attrs["context"] = std::make_shared<nnvm::any>(
+        exec::ContextVector(indexed_graph.num_nodes(), default_ctx));
+    // infer shapes
+    g = exec::InferShape(std::move(g), std::move(arg_shapes), "__shape__");
+    if (g.GetAttr<size_t>("shape_num_unknown_nodes") != 0U) {
```
ZhennanQin: Seems this code doesn't support partial inter shape, while `InferForwardAttrs` supports that.

src/operator/subgraph/build_subgraph.cc
```diff
@@ -572,27 +572,30 @@ void CreateSubgraphNode(nnvm::Graph* g,
...
+        BiDirectedNode* sn = simple_nodes[nid].get();
+        for (BiDirectedNode* dest_node : subgraph_nodes) {
+          sn->outputs.erase(dest_node->node);
+        }
+        sn->outputs[n.get()].push_back(i);
       }
-      sn->outputs[n.get()].push_back(i);
     }
   }
 #if DEBUG_SUBGRAPH
```
ZhennanQin: Consider the case that `#if DEBUG_SUBGRAPH` is true.

src/operator/subgraph/subgraph_property.h
```diff
@@ -221,6 +222,11 @@ class SubgraphProperty {
     return nullptr;
   }
 
+  virtual void PrePartition(const nnvm::Graph& g,
```
ZhennanQin: Can you provide an example how to use this API? Please take `excluded_ops=['BatchNorm']` as an example.

tests/python/unittest/test_subgraph_op.py
ZhennanQin: Consider to add a test for 
```
sym = sym.optimize_for('default', args, excluded_ops=['BatchNorm'])
```

src/c_api/c_api_symbolic.cc
ZhennanQin: InferStorageTypes is not only used by MKLDNN, but important for all backends, because it defines whether it's a sparse computation or not. Not all backends can support sparse tensor. I think we need to gather storage type along with dtype and shape for subgraph partition.

src/c_api/c_api_symbolic.cc
ZhennanQin: Partial infer shape is useful even for the backend that needs explicit shapes/types.  MXNet newly introduced some operators that generate dynamic out shape, e.g. boolean_mask. If a network uses that op, then there will be some shape can't be inferred. If we support partial infer shape, then that backend can still make partition based on the known shapes/types, otherwise, that backend can't be used for the this graph.

src/operator/subgraph/subgraph_property.h
ZhennanQin: Got your point. Because in this PR, Prepartition and PostPartition are just 2 empty virtual function, so I can't image the usage of them. Your explanation makes sense to me. But I still suggest to implement  `excluded_ops` to `default` backend as an example(considering `default` backend is an example, we can use it to example the features newly added), then other developer can know how to use this 2 APIs.

src/c_api/c_api_symbolic.cc
ZhennanQin: I think the differences are minor, and they won't cause problems. Overall, I think we need to unify the build subgraph process to make it behaves all the same either from bind API or optimize_for API.

ZhennanQin: That's fine with me. Alternatively, it's better to add some comments to mention this, if other developer changed `InferForwardAttrs` for the subgraph process in `bind` API, probably that change need to apply to `optimize_for` as well.

ZhennanQin:
I guess we have agreements for all my concerns. So LGTM.

src/c_api/c_api_symbolic.cc
ZhennanQin: For mkldnn backend, the correct ctx is important for getting expected result from `InferStorageTypes`, so please consider to accepts correct ctx from user. I expect the API would be like
```
sym = sym.optimize_for('default', ctx= mx.cpu(), args) #same context with bind
exe = sym.bind(ctx=mx.cpu(), args=args, aux_states=aux, grad_req='null')
```

ZhennanQin: I don't think that check is necessary, as user can use different `args` between `optimize_for` and `bind`. 
Another case is, to support multiple backends mixed running, user may run optimize_for many times. Like,
```
sym = sym.optimize_for('MKLDNN', ctx= mx.cpu(), args) 
sym = sym.optimize_for('backendA', ctx= mx.gpu(), args) # backend A to handle the rest graph
sym = sym.optimize_for('backendB', ctx= mx.cpu(), args) # backend B to handle the rest graph
exe = sym.bind(ctx=mx.cpu(), args=args, aux_states=aux, grad_req='null')
```
For this case, context is mainly decided by graph partition, not bind.

ZhennanQin: > Instead we'll pull the context from where the args reside.

args are normally loaded by
```
sym, args, aux = mx.model.load_checkpoint('resnet-50', 0)
```
No contexts are specified. That's why we need to pass the expected contexts during bind. The contexts from args doesn't mean anything.

>We expect that users could call this more than once for the same model. For example, users could first optimize for EIA (that would group supported nodes into subgraphs) and then optimize for MKLDNN for the remaining nodes that were not partitioned into subgraphs.

That's the comments you provided earlier. So I thought you're working to support multi-context execution. If not, then you mean we only support running multiple `optimize_for` for same contexts? Even for that, we need to pass correct contexts for `optimize_for`, not the gathered one from args.

Now again, the correct contexts are important for `InferStorageTypes` for MKLDNN backend, so we need to pass correct context(same as bind) to `optimize_for`. If you don't want to introduce `ctx` for `optimize_for` API, then I suggest to query the expected context from backend property, and use it to pass `InferStorageTypes`. Or, you can figure out other solution to solve this, I just want to ensure that subgraph partition can get same `InferStorageTypes` results as bind.

ZhennanQin: Finally you got my point. Thanks for your patients. Either solution is OK for me, you can make the decision.

python/mxnet/symbol/symbol.py
```diff
@@ -1437,6 +1437,54 @@ def _gen_atomic_symbol(self):
         return Symbol(handle)
 
 
+    def optimize_for(self, backend, args=None, **kwargs):
```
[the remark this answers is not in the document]
mseth10: I agree with you on having support for passing shape/type/stype dicts just like simple_bind. I will add that support in a later PR. Since we accept kwargs now, it would not be a breaking API change.

I understand your concern about having a more explicit api, but I prefer keeping it this way for the following reason. Users are expected to know what arguments are required by the backend. The backend errors out if the required arguments are not provided whether or not we have an explicit flag. At this point, it's not a choice for users to not infer shape/type/stype when the backend requires it, or vice versa.

samskalicky:
@ZhennanQin can you please approve? @mseth10 has added the "Next steps" in the PR description to save the outcome of our discussions from the PR comments.

samskalicky:

samskalicky:

PR #15906, opened 2019-08-15
[Numpy] Numpy operator diff

xiezhq-hermann:
## Description ##
Numpy compatible operator [diff](https://docs.scipy.org/doc/numpy/reference/generated/numpy.diff.html)

### Essentials ###

### Changes ###
- Numpy compatible operator diff added, including tests and docs

## Comments ##
Thank you for your effort to review! @haojin2 @reminisce

### DIFF
--- a/src/operator/numpy/np_diff-inl.h
+++ b/src/operator/numpy/np_diff-inl.h
@@ -1,0 +1,220 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ * \file np_diff-inl.h
+ * \brief Function definition of numpy-compatible diff operator
+ */
+
+#ifndef MXNET_OPERATOR_NUMPY_NP_DIFF_INL_H_
+#define MXNET_OPERATOR_NUMPY_NP_DIFF_INL_H_
+
+#include <mxnet/base.h>
+#include <mxnet/operator_util.h>
+#include <vector>
+#include "../mxnet_op.h"
+#include "../operator_common.h"
+#include "../tensor/broadcast_reduce_op.h"
+
+namespace mxnet {
+namespace op {
+
+struct DiffParam : public dmlc::Parameter<DiffParam> {
+  int n, axis;
+  dmlc::optional<mxnet::TShape> prepend;
+  dmlc::optional<mxnet::TShape> append;
+  DMLC_DECLARE_PARAMETER(DiffParam) {
+    DMLC_DECLARE_FIELD(n).set_default(1).describe(
+        "The number of times values are differenced."
+        " If zero, the input is returned as-is.");
+    DMLC_DECLARE_FIELD(axis).set_default(-1).describe(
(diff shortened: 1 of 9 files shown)


tests/python/unittest/test_numpy_op.py
```diff
@@ -880,6 +880,69 @@ def get_new_shape(shape, axis):
...
+@with_seed()
+@use_np
+def test_np_diff():
+    def np_diff_backward(ograd, n, axis):
+        res = ograd
+        for i in range(n):
+            res = _np.negative(_np.diff(res, n=1, axis=axis, prepend=0, append=0))
+        return res
+
+    @use_np
```
haojin2: get rid of this line.

src/operator/numpy/np_diff-inl.h
```diff
@@ -0,0 +1,215 @@
...
+    DMLC_DECLARE_FIELD(n).set_default(1).describe(
+        "The number of times values are differenced."
+        " If zero, the input is returned as-is.");
+    DMLC_DECLARE_FIELD(axis).set_default(-1).describe(
+        "Axis along which the cumulative sum is computed."
+        " The default (None) is to compute the diff over the flattened array.");
+  }
+};
+
+inline void YanghuiTri(std::vector<int>* buffer, int n) {
```
haojin2: Get rid of `inline` here.

src/operator/numpy/np_diff-inl.h
```diff
@@ -0,0 +1,215 @@
...
+template <typename xpu>
+void DiffBackwardImpl(const OpContext& ctx, const TBlob& ograd,
+                      const TBlob& igrad, const int n, const int axis) {
+  using namespace mshadow;
+  using namespace mxnet_op;
+
+  CHECK_GE(n, 0);
+  int axis_checked = CheckAxis(axis, igrad.ndim());
+  if (n >= igrad.shape_[axis_checked])
+    return;  // nothing in the ograd and igrad
```
haojin2: for all one-line if-statements, do either:
```c++
if (<condition>) {
  // your code
}
```
or
```c++
if (<condition>) // your code
```

src/operator/numpy/np_diff-inl.h
xiezhq-hermann: The `inline` is finally kept since the function here is really concise and no need to be put in .cc

tests/python/unittest/test_numpy_op.py
```diff
@@ -1519,6 +1519,58 @@ def hybrid_forward(self, F, a):
...
+    class TestDiff(HybridBlock):
+        def __init__(self, n=1, axis=-1):
+            super(TestDiff, self).__init__()
+            self._n = n
+            self._axis = axis
+
+        def hybrid_forward(self, F, a):
+            return F.np.diff(a, n=self._n, axis=self._axis)
+
+    shapes = [tuple(random.randrange(10) for i in range(random.randrange(6))) for j in range(5)]
```
comaniac: Please design static shapes for testing so that we could guarantee all corner cases are covered.

xiezhq-hermann: Hi, the random shape generator should have covered the corner cases, including zero dim, 1d to 5d input shape. Actually, the codes have been tested by static shapes and the random generator here is for shorter testing time.
Do you think it's necessary to change back still? If yes, I'll do it.

comaniac: I prefer to have a static shape list tho. Here are my concerns, and you are welcome to let me know if those look not an issue in your opinion:

- This code may be maintained or improved by others in the future, but anyone new to this code will get confused about what tests it has covered.
- If the code is modified in the future that fails to cover certain cases, it is hard to capture because re-run the test may not give you the same shape. We should keep the unit test result deterministic.

xiezhq-hermann: I agreed your opinions and that's what I did before, but the present random cases generator should be concise and easy to understand. 
The tests are switched to random cases mainly for the balance of time cost and test coverage, which is advised by @haojin2 , could you reach an agreement?

comaniac: I don't quite get the point about the balance of time cost and test coverage tho. It seems to me that declaring a static list (i.e., shape = [[...], [...], ...]) won't cost more than calling random generator on the fly. I would even prefer the way that you generate a list of shapes randomly and hardcoded the generated list in this unit test. On the other hand, I'm also fine if you and @haojin2 also agreed with the random generator way in case you guys have other concerns I missed.

PR #15921, opened 2019-08-15
dynamic custom operator support

samskalicky:
## Description ##
Enhancements to dynamic library loading to support custom operators in libraries.
- added MXTensor/MXDtype structure, and versioning to lib_api.h
- added similar NNVM register op-like capability for custom ops
- operators are found in the library, and re-registered in MXNet during library loading
- operators are re-registered from mx.nd.op to mx.nd and mx.sym.op to mx.sym shortcuts
- Created a new example library in "example/lib_ops" with a GEMM operator

Initially, this project was [proposed on the CWiki](https://cwiki.apache.org/confluence/display/MXNET/Dynamic+CustomOp+Support), however the design has evolved since the initial proposal. The current design is described below.

## Design ##
The goal of this PR to to enable operators to be implemented in separate libraries and loaded at runtime

The main constraint is to maintain a low-level C-types only boundary between MXNet and the library to simplify the building and compiling of external libraries. 

Working backwards from the user, users register operators with easy-to-use function prototypes like:
```
int myForward(std::map<std::string,std::string> attrs,
               std::vector<MXTensor> inputs, 
               std::vector<MXTensor> outputs);
```

Users' `Forward` (ie. `FCompute`) functions are called by a helper function `_opCallFCompute` that converts the C-types passed across the library boundary to the familiar STL types. This function is implemented in the `lib_api.h` header file that users compile with their library. 
```
int _opCallFCompute(fcomp_t fcomp, const char* const* keys, const char* const* vals, int num,
                    const int64_t** inshapes, int* indims, void** indata, int* intypes, int num_in,
                    const int64_t** outshapes, int* outdims, void** outdata, int* outtypes, int num_out);
```

In MXNet's C API, the `_opCallFCompute` function is found in the library. A lambda function `fcomp_conv` is created for each operator loaded from the library to convert from MXNet-types to C-types. Then these C-types are passed to `_opCallFCompute`. 

```
auto fcomp_conv = [=](const nnvm::NodeAttrs& attrs,
                      const OpContext& ctx,
                      const std::vector<TBlob>& inputs,
                      const std::vector<OpReqType>& req,
                      const std::vector<TBlob>& outputs);
```

The same design is used for: parseAttrs, inferShape, inferType, etc.

Finally, an operator is re-registered in MXNet with the lambda function like:
```
nnvm::Op &regOp = dmlc::Registry<nnvm::Op>::Get()->__REGISTER_OR_GET__(name);
regOp.set_attr<FCompute>("FCompute<cpu>",fcomp_conv);
```

Once the C API returns back to Python in the `load` function in `library.py`, we regenerate the Python bindings and re-register the operator shortcuts to mx.nd and mx.sym.

After the `load` function returns back to the user's Python code, they can then use their operators just like any other operate:
```
mx.nd.myCustomOp(A,B,C)
```

### Current Features ###
- custom CPU operators
- stateless & stateful operators
- custom subgraph operators
- Memory resource request
- 

### Future/Next-steps ###
(to be done in a separate PR)
- custom GPU operators
- Random number generator resource request
- sparse data types
- DLTensor
- migrate lambda functions in MXLoadLib in src/c_api/c_api.cc to classes defined elsewhere
- Documentation, add the "library" python package to the namespace to the doc: https://mxnet.apache.org/api/python/docs/api/ ?

### Essentials ###

### Changes ###
- updated path to be absolute in example/lib_api/test.py
- update the Makefile/CMakeLists.txt to build the new example in lib_ops instead of lib_api
- moved library import in python/mxnet/__init__.py to after ndarray/symbol
- added operator reregistration to python/mxnet/library.py
- added operator discovery/registration in MXLoadLib in src/c_api/c_api.cc


samskalicky:
@wkcn while this PR is not quite done yet, it would be great to get some early feedback since the design/implementation has changed since our initial discussion. Let me know what you think, thanks!

wkcn:
Hi @samskalicky , is it possible to use `mshadow` in dynamic custom operator?
It will be convenient to leverage some math helpers.

samskalicky:
> Hi @samskalicky , is it possible to use `mshadow` in dynamic custom operator?
> It will be convenient to leverage some math helpers.

Sure, users can compile whatever they want into their libraries. But they will compile mshadow separately from MXNet. 

The main goal in architecting the custom op support is to lower the barrier for entry to compiling custom operators. Currently users have to recompile all of mxnet to integrate their custom ops. So we want to do better with this project. If users have a custom op that does not uses mshadow they shouldnt have to compile with it.

wkcn:
@samskalicky Nice! We can provide some higher-level language (e.g. C++) examples with mshadow.

samskalicky:
@wkcn any thoughts on how we can modify this to re-register an operator multiple times?

https://github.com/apache/incubator-mxnet/pull/15921/files#diff-3288207540b433242fe8da1390c68d3fR423-R427

Im thinking of the scenario when a user loads multiple libraries that each have the same operator. So its possible that an operator can be re-registers over and over (obviously only the last registered one will be used). But it would suck if a user couldnt use an operator in another library if it also defined an operator they already loaded from a different library.

wkcn:
@samskalicky It may be helpful to define a `namespace` to avoid re-registering an operator multiple times. The namespace can be defined in the source code of custom operators, or when loading the library, like `mx.library.load('myop.lib', namespace='mynamespace')`, then  users call `mx.nd.mynamespace.myop`.

If the name conflict still exists, we will raise a warning or an exception.

samskalicky:
@wkcn the problem is in NNVM, not python. 

Here, we register attributes on an operator:

https://github.com/apache/incubator-mxnet/blob/711f9a364d463018395f84059edb8e2b3eabb052/src/c_api/c_api.cc#L423-L427

`set_attr` takes a pvalue that defaults to 10:

https://github.com/dmlc/nnvm/blob/master/include/nnvm/op.h#L237-L239

And here in the set_attr function theres a check that makes sure attributes are not re-registered with the same pvalue:

https://github.com/dmlc/nnvm/blob/master/include/nnvm/op.h#L461-L464

So my though was to just set it to 11, so that lets us re-register it once. But the 2nd time it will fail since there already is an 11. 

But my question was if you had any suggestions on how to extend this so that we can re-register over and over.

The use case is the following:
- Library1 has {opA, opC, opD}
- Library2 has {opB, opC, opE}

Lets say the user wants to use: opA, opB, and opD. So they load both Library1 & Library2. So loading Library1 succeeds registering opC. But then loading Library2 fails since opC is already registered. In this case the user doesnt care about opC, so we are ok with opC being overwritten by the implementation in Library2.

wkcn:
@samskalicky I see. Could we add a prefix name to avoid name conflict?
For example,

Library1 has {opA, opC, opD}
Library2 has {opB, opC, opE}

Firstly, we load lib1. They are registered as 1_opA, 1_opC, 1_opD.
Then we load lib2, and lib2 registers 1_opB, *2_opC*, 1_opE.
If we load lib2 again, they are called as 2_opB, 3_opC, 2_opE.

There is not any name conflict in nnvm.
In the front end, we can map them to some new names.

e.g.

lib1 = mx.library.load(“lib1.so”, namespace=“lib1”)
op = mx.nd.lib1.opA(...)
op2 = lib1.nd.opA(...)

wkcn:
Hi @rondogency , do you have any idea?

samskalicky:
I think that hurts the ease of use to force namespacing. Plus if we internally set lib1.opA then users wont know how to call mx.nd.lib1.opA or mx.nd.lib1_opA (whatever the naming scheme is). If we give them a way to set the namespace name, its still not as easy to use as mx.nd.opA. Plus we want users to be able to override internal MXNet ops.

For example mx.nd.Convolution does not support FP16 on CPU. So if a user had a need for that they could write their own. So we need ops to be overwritten, and be in the mx.nd namespace. 

Im thinking the best solution might be to track the number of times an operator is registered and just increment the pvalue for now.

wkcn:
Agree that users is allowed to define the namespace manually and override internal operators.
In the case you gave, 
```
Library1 has {opA, opC, opD}
Library2 has {opB, opC, opE}
```
Users may want opA, opB, opE, opD, and opC in library1. They likewise choose opA, opB, opE, opD, and opC in library2.
I don't think pvalue can address the problem.

samskalicky:
@wkcn should we consider removing the nnvm pvalue restriction that attrs can be set multiple time here:

https://github.com/dmlc/nnvm/blob/dab5ce8ab6adbf4edd8bd2fa89f1a99f343b6e38/include/nnvm/op.h#L461-L464

Or add an argument to optionally allow override, but default to current behavior?

Like this:

‘’’
  inline Op& set_attr(const std::string& attr_name,  // NOLINT(*)
                      const ValueType& value,
                      int plevel = 10, bool override=false);

template<typename ValueType>
inline Op& Op::set_attr(  // NOLINT(*)
    const std::string& attr_name,
    const ValueType& value,
    int plevel,
    bool override) {
  CHECK_GT(plevel, 0)
      << "plevel in set_attr must be greater than 0";
  // update the attribute map of the key by creating new empty if needed.
  UpdateAttrMap(attr_name,
                [this, attr_name, value, plevel](any* pmap) {
      // the callback is in lockscope so is threadsafe.
      if (pmap->empty()) {
        OpMap<ValueType> pm;
        pm.attr_name_ = attr_name;
        *pmap = std::move(pm);
      }
      CHECK(pmap->type() == typeid(OpMap<ValueType>))
          << "Attribute " << attr_name
          << " of operator " << this->name
          << " is registered as inconsistent types"
          << " previously " << pmap->type().name()
          << " current " << typeid(OpMap<ValueType>).name();
      std::vector<std::pair<ValueType, int> >& vec =
          nnvm::get<OpMap<ValueType> >(*pmap).data_;
      // resize the value type.
      if (vec.size() <= index_) {
        vec.resize(index_ + 1,
                   std::make_pair(ValueType(), 0));
      }
      std::pair<ValueType, int>& p = vec[index_];
      CHECK(p.second != plevel && !override)
          << "Attribute " << attr_name
          << " of operator " << this->name
          << " is already registered with same plevel=" << plevel;
      if (p.second < plevel) {
        vec[index_] = std::make_pair(value, plevel);
      }
    });
  return *this;
}
‘’’

wkcn:
We don't need to change the NNVM code. For multiple registration, we can use different names to register them in the back-end. Users only focus on the front-end, which uses `namespace` to distinguish different implementations of CustomOp.

For example, users want to register a ConvOp to override the original Convolutional Op:
```python
mx.library.load('MyConvOp.so', override=True)
# mx.nd.Convolution will be the custom operator
```
If users doesn't want to override the original ConvOp,
```python
mx.library.load('MyConvOp.so', namespace='myop')
# mx.nd.myop.Convolution is the custom operator, and mx.nd.Convolution is not overrided.
```

The flag `override` is `False` by default. If users don't set it `True` and register an operator multiple times, it will raises the exception to prompt that `The operator xxx has been registered. Please rename it`.

samskalicky:
That’s a good idea @wkcn but we have users that will be using c or cpp APIs with custom ops. So it will need to go into the backend. 

@junrushao1994 any thoughts?

example/lib_ops/mylib.cc
```diff
@@ -0,0 +1,139 @@
...
+  return 1; //no error
+}
+
+REGISTER_OP(sam)
+.setFCompute(myFCompute)
+.setParseAttrs(parseAttrs)
+.setInferType(inferType)
+.setInferShape(inferShape);
+
+int initialize(int version) {
```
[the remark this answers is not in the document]
wkcn: Agree major10000 + minor100 + patch.

samskalicky:
@wkcn just talked to @junrushao1994 offline and its not ideal and will be a pain-in-the-@$$ to get that functionality merged into nnvm and then update the submodule dependency commit in MXNet for that. Especially since nnvm is not high priority now that we're moving to relay. 

So I think for now i'll implement a map that track the operator re-registrations and just uses incremented plevel values to allow calling set_attr multiple times. its definitely a hack, but i think its the best option for now.

src/c_api/c_api.cc
```diff
@@ -92,16 +93,384 @@ inline int MXAPIGetFunctionRegInfo(const FunRegType *e,
...
+    // validate custom operator functions from the dynamic library
+    CHECK(fcomp_fp != nullptr) << "Error loading '" << name
+                            << "' custom op, FCompute function was not set.";
+    CHECK(parse_fp != nullptr) << "Error loading '" << name
+                            << "' custom op, ParseAttrs function was not set.";
+    CHECK(type_fp  != nullptr) << "Error loading '" << name
+                            << "' custom op, InferType function was not set.";
+    CHECK(shape_fp != nullptr) << "Error loading '" << name
+                            << "' custom op, InferShape function was not set.";
+
```
samskalicky: Lets add a CHECK for mutate_fp too

src/c_api/c_api.cc
```diff
@@ -92,16 +93,384 @@ inline int MXAPIGetFunctionRegInfo(const FunRegType *e,
...
+      }
+
+      return mutate_indices_list;
+    };
+
+    auto infer_storage_type = [=](const nnvm::NodeAttrs& attrs,
+                                  const int dev_mask,
+                                  DispatchMode* dispatch_mode,
+                                  std::vector<int>* in_stypes,
+                                  std::vector<int>* out_stypes) {
```
samskalicky: until we decide to actively support sparse, lets error out if in_stypes contain sparse types

include/mxnet/lib_api.h
```diff
@@ -18,33 +18,627 @@
...
+                                    int**, int*);
+
+#define MXLIB_OPCALLCREATEOPSTATE_STR "_opCallCreateOpState"
+typedef int (*opCallCreateOpState_t)(createOpState_t, const char* const*, const char* const*, int,
+                                     void**);
+
+#define MXLIB_OPCALLFSTATEFUL_STR "_opCallFStateful"
+typedef int (*opCallFStateful_t)(fstateful_t, void*,
+                                 const int64_t**, int*, void**, int*, int,
+                                 const int64_t**, int*, void**, int*, int);
```
samskalicky: We need to enable OpResource for Stateful ops too like we do with FCompute

example/lib_ops/subgraph_lib.cc
```diff
@@ -0,0 +1,125 @@
...
+MXReturnValue forwardStateful(CustomStatefulOp* op_inst,
+                              std::vector<MXTensor> inputs,
+                              std::vector<MXTensor> outputs) {
+  MyStatefulOp* my_op_inst = static_cast<MyStatefulOp*>(op_inst);
+  if (my_op_inst == nullptr) {
+    std::cout << "stateful op loading failed" << std::endl;
+    return MX_FAIL;
+  }
+
+  my_op_inst->Forward();
```
samskalicky: Need to pass inputs/outputs to forward function

include/mxnet/lib_api.h
```diff
@@ -18,33 +18,627 @@
...
+    return *this;
+  }
+  CustomOp& setCreateOpState(createOpState_t func) {
+    create_op_state = func;
+    return *this;
+  }
+  CustomOp& setForwardStateful(fstateful_t func) {
+    fstateful = func;
+    return *this;
+  }
```
samskalicky: We need to add some error checking in lib_api.h to prevent users from registering the wrong combo of functions. Currently we have 2 scenarios:

Basic Op Registration requires:
- Parse Attrs
- Infer Type
- Infer Shape
- Forward

__optional__
- Gradient --> Lets rename to Backward (to match with Forward)
- Mutate inputs

Stateful Op Registration requires:
- Parse Attrs
- Infer Type
- Infer Shape
- Create Op State
- Forward Stateful

__optional__
- Gradient --> Lets rename to Backward (to match with Forward)
- Mutate inputs

We need to enforce that the Basic Op Forward is not registered with Create Op State or Forward Stateful and vice-versa. This will give users an error when compiling their operator libraries before loading in MXNet

include/mxnet/lib_api.h
```diff
@@ -18,33 +18,627 @@
...
+ private:
+  void* instance;
+};
+
+/*!
+ * \brief An prototype interface class for library author creating stateful op
+ */
+class CustomStatefulOp {
+ public:
+  virtual void Forward() = 0;
```
samskalicky: Shouldnt we make the constructor virtual and = 0 too? Or is there somewhere we need to be able to construct instances of the super class?

example/lib_ops/subgraph_lib.cc
```diff
@@ -0,0 +1,125 @@
...
+  ~MyStatefulOp() {}
+
+ private:
+  std::string subgraph_sym;
+  int count;
+};
+
+MXReturnValue createOpState(std::map<std::string,std::string> attrs,
+                            CustomStatefulOp** op_inst) {
+  *op_inst = new MyStatefulOp("json", 0);
```
samskalicky: Lets add a check for "subgraph_sym_json" key in the attrs map, and error out if its missing.
Lets get the "json" string from the attrs map using the "subgraph_sym_json" key rather than setting the placeholder "json" string. 

We'll assume the subgraph property sets this attr, later. 

```
CHECK(attrs.find("subgraph_sym_json") != attrs.end()) << "Error! Expected key '" << "subgraph_sym_json" << "' in Node Attributes";
*op_inst = new MyStatefulOp(attrs["subgraph_sym_json"], 0);
```

include/mxnet/lib_api.h
```diff
@@ -18,33 +18,601 @@
...
+  virtual MXReturnValue Forward(std::vector<MXTensor> inputs,
+                                std::vector<MXTensor> outputs,
+                                OpResource op_res) = 0;
+  virtual MXReturnValue Backward(std::vector<MXTensor> inputs,
+                                 std::vector<MXTensor> outputs,
+                                 OpResource op_res) {
+    std::cout << "Error! Operator does not support backward" << std::endl;
+    return MX_FAIL;
+  }
+  virtual ~CustomStatefulOp() = 0;
```
samskalicky: lets not force the user to implement this. it also solves the weirdness around pure virtual destructor issue. I suggest we remove this line.

Makefile
```diff
@@ -664,6 +664,9 @@ cpplint:
 pylint:
 	python3 -m pylint --rcfile=$(ROOTDIR)/ci/other/pylintrc --ignore-patterns=".*\.so$$,.*\.dll$$,.*\.dylib$$" python/mxnet tools/caffe_converter/*.py
 
+sample_lib:
+	$(CXX) -shared -fPIC -std=c++11 example/lib_ops/gemm_lib.cc -o libsample_lib.so -I include/mxnet
+
```
samskalicky: Can we please add an rm libsample_lib.so to the clean target so this is removed and the project can be cleaned up properly?

rondogency: that's a good point

example/lib_ops/gemm_lib.cc
```diff
@@ -123,13 +114,17 @@ MXReturnValue backward(std::map<std::string, std::string> attrs,
...
+  float *At = new float[n*k];
+  float *Bt = new float[k*m];
+
   transpose(A, At, n, k);
   transpose(B, Bt, k, m);
 
   gemm(dC, Bt, dA, n, k, m);
   gemm(At, dC, dB, n, k, m);
 
+  free(At);
```
wkcn: delete []At;
delete []Bt;

new and delete are matching, and malloc and free are matching.

rondogency: thanks for pointing it out!

rondogency:
@samskalicky @wkcn @szha @junrushao1994 @eric-haibin-lin This PR is ready for review.
c_api.cc and lib_api.h files are the most important
c_api.cc loads custom op and register in nnvm; lib_api.h has all prototypes used in custom library
example/extensions/*.cc(py) are simple custom operator examples
other files include unittest, build change and python wrapper change

samskalicky:
Failing centos-gpu job is failing with Numpy issue #16358 unrelated to this PR. We've sync'ed with the authors and they confirmed. Assume all CI jobs are passing at this point until they can address the issue

example/extensions/lib_custom_op/Makefile
```diff
@@ -0,0 +1,27 @@
...
+all: subgraph_lib gemm_lib
+
+gemm_lib:
+	g++ -shared -fPIC -std=gnu++0x gemm_lib.cc -o libgemm_lib.so -I ../../../include/mxnet
+
+subgraph_lib:
+	g++ -shared -fPIC -std=gnu++0x subgraph_lib.cc -o libsubgraph_lib.so -I ../../../include/mxnet
+
+clean:
+	rm -rf *.so
```
wkcn: rm libgemm_lib.so libsubgraph_lib.so may be better, because users may create other *.so files in this directory.

example/extensions/lib_custom_op/gemm_lib.cc
```diff
@@ -0,0 +1,246 @@
...
+ * Copyright (c) 2019 by Contributors
+ * \file gemm_lib.cc
+ * \brief Sample 2D gemm custom operator implementation library file
+ */
+
+#include <iostream>
+#include "lib_api.h"
+
+// main matrix multiplication routine
+void gemm(float* A, float* B, float* C, unsigned n, unsigned k, unsigned m) {
```
wkcn: I suggest that we set the inputs as constant values.

example/extensions/lib_custom_op/gemm_lib.cc
```diff
@@ -0,0 +1,246 @@
...
+    for (j = 0; j < m; j++) {
+      C[i*m+j] = 0;
+      for (kk = 0; kk < k; kk++) {
+        C[i*m+j] += A[i*k+kk] * B[kk*m+j];
+      }
+    }
+  }
+}
+
+void transpose(float* A, float* At, unsigned n, unsigned m) {
```
wkcn: the same as above. const float* A may be better : )

example/extensions/lib_custom_op/gemm_lib.cc
```diff
@@ -0,0 +1,246 @@
...
+MXReturnValue parseAttrs(std::map<std::string, std::string> attrs, int* num_in, int* num_out) {
+  *num_in = 2;
+  *num_out = 1;
+  return MX_SUCCESS;
+}
+
+MXReturnValue inferType(std::map<std::string, std::string> attrs,
+                        std::vector<int> &intypes,
+                        std::vector<int> &outtypes) {
+  // validate inputs
```
wkcn: we should check whether the input type is float32 in inferType rather than the forward function.

example/extensions/lib_custom_op/gemm_lib.cc
```diff
@@ -0,0 +1,246 @@
...
+  unsigned n = inshapes[0][0];
+  unsigned k = inshapes[0][1];
+  unsigned kk = inshapes[1][0];
+  unsigned m = inshapes[1][1];
+  if (k != kk) {
+    std::cout << "Exected first input axis 1 equals to second input axis 0" << std::endl;
+    return MX_FAIL;
+  }
+
+  outshapes[0].push_back(n);
```
wkcn: Is it available that outshapes[0] = {n, m}; ?

include/mxnet/lib_api.h
```diff
@@ -18,33 +18,756 @@
...
+#include <string>
+#include <iostream>
+#include <utility>
+
+#define MX_LIBRARY_VERSION 1
+
+/*!
+ * \brief External Tensor data types
+ */
+enum MXDType {
```
wkcn: Is it consistent with DLPack or mshadow?

include/mxnet/lib_api.h
```diff
@@ -18,33 +18,756 @@
...
+ */
+struct MXTensor {
+  MXTensor() : data(NULL) {}
+
+  MXTensor(void *data, const std::vector<int64_t> &shape, MXDType dtype)
+  : data(data), shape(shape), dtype(dtype) {}
+
+  /*! \brief helper function to cast data pointer */
+  template<typename data_type>
+  inline data_type* getData() {
```
wkcn: I suggest to check data type in the getData function.
For the name of functions, getData -> data, getDataSize -> size, getDataShape -> shape may be better.

include/mxnet/lib_api.h
rondogency: this is consistent with mshadow, not dlpack

include/mxnet/lib_api.h
rondogency: I think it should be user's job to figure out what data type to cast given tensor dtype info. The mshadow macro or similar mechanism might be too complex for the custom op, especially it can be done by a few lines of hard-coded types. What do you think?

wkcn: Agree : )

include/mxnet/lib_api.h
```diff
@@ -18,33 +18,754 @@
...
+    return size;
+  }
+
+  // data is flatten 1D repr of tensor, elements are in continuous memory
+  // user can access each element using the shape of tensor
+  // it may also point to data allocated on gpu
+  void *data_ptr;
+
+  // shape is in [2,3,4] format to represent high-dim tensor
+  std::vector<int64_t> shape;
```
junrushao1994: Do we assume that we share a same c++ runtime so that the allocators behind std::vector are consistent?

include/mxnet/lib_api.h
```diff
@@ -18,33 +18,754 @@
...
+
+  // type can only be MXDType enum types
+  MXDType dtype;
+
+  // gpu flag to specify the data tensor storage location
+  bool is_gpu;
```
junrushao1994: There might be some issues with multiple gpus, in which we might not have idea about on which gpu the tensor resides

include/mxnet/lib_api.h
```diff
@@ -18,33 +18,754 @@
...
+};
+
+/*!
+ * \brief resource malloc function to allocate memory inside Forward/Backward functions
+ */
+typedef void* (*xpu_malloc_t)(void*, int);
```
junrushao1994: Do you guys prefer int64_t?

junrushao1994: What about alignment?

include/mxnet/lib_api.h
```diff
@@ -18,33 +18,754 @@
...
+  OpResource(xpu_malloc_t xm, void* _xm) : xpu_malloc(xm), _xpu_malloc(_xm) {}
+
+  /*! \brief allocate memory controlled by MXNet */
+  void* alloc(int size) {
+    return xpu_malloc(_xpu_malloc, size);
+  }
+
+ private:
+  xpu_malloc_t xpu_malloc;
+  void* _xpu_malloc;
```
junrushao1994: what's this :-)

include/mxnet/lib_api.h
```diff
@@ -18,33 +18,754 @@
...
+  }
+  json_type type;
+  int num;
+  std::string str;
+  std::vector<json_val> list;
+  std::map<json_val, json_val> map;
+};
+
+/*! \brief functions used for parsing JSON */
+struct Json_Parser {
```
junrushao1994: nits: naming style

include/mxnet/lib_api.h
```diff
@@ -18,33 +18,754 @@
...
+  /*! \brief map of entries in registry */
+  std::vector<T*> entries;
+};
+
+/*!
+ * \brief Macros to help with string concat
+ * Annoyingly, the concat_ and concat macros are necessary to
+ * be able to use __COUNTER__ in an identifier name 
+ */
+#define _STR_CONCAT_(__a, __b) __a ## __b
```
junrushao1994: Add some prefix to those macros so that there is no conflict

src/c_api/c_api.cc
```diff
@@ -92,16 +95,589 @@ inline int MXAPIGetFunctionRegInfo(const FunRegType *e,
...
+      CustomStatefulOp* state_op_inst = op.get_instance();
+      CHECK(state_op_inst != nullptr)
+      << "Error MXNet cannot load custom stateful operator'" << name_str << "'";
+
+      if (forward) {
+        CHECK(state_op_inst->Forward(c_inputs, c_outputs, op_res))
+        << "Error calling ForwardStateful for custom operator '" << name_str << "'";
+      } else {
+        CHECK(state_op_inst->Backward(c_inputs, c_outputs, op_res))
+        << "Error calling BackwardStateful for custom operator '" << name_str << "'";
```
samskalicky: We cant call Forward/Backward directly, we need a _opCallStatefulFCompute function like _opCallFCompute that rebuilds the MXTensors in lib_api.h, not directly here in MXNet

Because we cannot pass vector objects between MXNet and the libraries. Vectors are a class, and we can only pass C-types (pointers, ints, etc.) for binary compatibility. This lets the library creator use a different version of gcc and/or cstdlib or STL, etc.

include/mxnet/lib_api.h
samskalicky: Good catch @junrushao1994, we dont assume MXNet and the custom operator library use the same C++ runtime

include/mxnet/lib_api.h
samskalicky: Another good catch, this not currently relevant since we dont support GPU operators yet. Currently we only register FCompute<cpu> for now. This should be addressed in the followup PR where we should also take a device_id or something to identify which device the array was allocated on.

include/mxnet/lib_api.h
samskalicky: This just passes the value to:
```
mshadow::Tensor<mxnet::cpu, 1, char> data =
        resource.get_space_typed<mxnet::cpu, 1, char>(mshadow::Shape1(size), cpu_stream);
```

So it doesnt need to be aligned, but the user can choose to align as they want. Again, currently this is only for allocating CPU memory via the MXNet Resource manager. 

Currently we're only support 32-bit tensor dimensions in this PR, we'll look to support large tensor operators in the next PR (after large tensor support is formally added in their forthcoming PRs).

include/mxnet/lib_api.h
samskalicky: This is a pointer to the cpu_malloc lambda function that eventually just passes the value to:
```
mshadow::Tensor<mxnet::cpu, 1, char> data =
        resource.get_space_typed<mxnet::cpu, 1, char>(mshadow::Shape1(size), cpu_stream);
```

include/mxnet/lib_api.h
samskalicky: change to JsonParser

include/mxnet/lib_api.h
samskalicky: change to MX_STR_CONCAT. Also change:
STRINGIFY --> MX_STRINGIFY
_REGISTER_NAME_ --> MX_REGISTER_NAME_
_REGISTER_DEF_ --> MX_REGISTER_DEF_

Leave REGISTER_OP unchanged, otherwise we need to go and change all the examples. This does not conflict with the NNVM_REGISTER_OP macro so it can stay for now.

include/mxnet/lib_api.h
junrushao1994: I see. If it is only for MXNet resource manager, I think it is totally good :-)

junrushao1994:
Hey, thank you guys for the nice work! BTW, would you mind if you guys clearly state why DLTensor is not adopted, which I believe would be useful for other community members for refenrece

samskalicky:
> Hey, thank you guys for the nice work! BTW, would you mind if you guys clearly state why DLTensor is not adopted, which I believe would be useful for other community members for refenrece

@wkcn (who implemented DLTensor support in MXNet) and I had a long discussion about this. In fact, we did investigate supporting DLTensor:
https://github.com/samskalicky/incubator-mxnet/blob/custom_op/example/custom_op/test.cc#L14

One takeaway we had was that it would not be easy or convenient to modify the structure of DLPack. The reason is that DLPack is used commonly in multiple deep learning frameworks, and we should keep the consistence of DLPack. So building MXNet custom operators on top of DLTensor would limit our future extensibility of MXNet and custom operator support. 

One example of this would be adding a "layout" field to the tensor structure (ie. NCHW). This is something that I have heard as a request, but is [not currently something the DLPack community is willing to accept](https://github.com/dmlc/dlpack/pull/42). The MXTensor structure in this work is compatible with DLPack/DLTensor, so any user that wants to convert from MXTensor to DLTensor can do so by simply setting the fields in a DLTensor without copying data, and with a very small overhead. 

Just because we're not using DLTensor now, does not mean that we cannot support it directly in a future PR. If enough users want this feature, the work in this PR can easily be extended to simply pass DLTensors from MXNet to the custom operators in the external library.

junrushao1994:
@samskalicky Thank you Sam for the explanation!

src/c_api/c_api.cc
```diff
@@ -406,6 +409,7 @@ int MXLoadLib(const char *path) {
       mshadow::Stream<mxnet::cpu> *cpu_stream = ctx.get_stream<mxnet::cpu>();
 
       // create lambda that captures stream & resource objects
+      // the memory pointer returned will eventually return to user
```
samskalicky: nit: consider rephrasing

src/c_api/c_api.cc
```diff
@@ -547,37 +552,39 @@ int MXLoadLib(const char *path) {
...
+        out_dims.push_back(outputs[i].shape().ndim());
+        out_types.push_back(outputs[i].dtype());
       }
 
       // get memory resource
       const Resource &resource = ctx.requested[0];
       mshadow::Stream<mxnet::cpu> *cpu_stream = ctx.get_stream<mxnet::cpu>();
 
       // create lambda that captures stream & resource objects
+      // the memory pointer returned will eventually return to user
```
samskalicky: nit: consider rephrasing

samskalicky:
Note: the two failing CI jobs are related to numpy:
sanity:
```
************* Module mxnet.numpy_op_signature

python/mxnet/numpy_op_signature.py:21:0: E0402: Attempted relative import beyond top-level package (relative-beyond-top-level)
```
unix-gpu:
```
Setting test np/mx/python random seeds, use MXNET_TEST_SEED=1746065449 to reproduce.

ERROR
```

junrushao1994:
@samskalicky the pylint issue in ci is really weird. very often i failed it, did nothing but retrigger, it disappears

example/extensions/lib_custom_op/gemm_lib.cc
```diff
@@ -0,0 +1,240 @@
...
+  float* A = inputs[1].data<float>();
+  float* B = inputs[2].data<float>();
+  float* dA = outputs[0].data<float>();
+  float* dB = outputs[1].data<float>();
+  // set tensor shapes
+  unsigned n = inputs[1].shape[0];
+  unsigned k = inputs[1].shape[1];
+  unsigned m = inputs[2].shape[1];
+
+  float *At = new float[k*n];
```
eric-haibin-lin: Are we suggesting users manage memory themselves? Why not using mxnet storage API to request for memory?

python/mxnet/__init__.py
```diff
@@ -87,6 +86,9 @@
 
 from . import gluon
 
+# Dynamic library module should be done after ndarray and symbol are initialized
+from . import library
```
eric-haibin-lin: Can we also add this namespace to the doc: https://mxnet.apache.org/api/python/docs/api/ ?

example/extensions/lib_custom_op/gemm_lib.cc
samskalicky: @rondogency can we use OpResource to allocate the memory instead of using "new" like @eric-haibin-lin is suggesting?
We need to make sure we only make one call to allocate memory, so make sure to request the amount for both At and Bt and then split up the memory manually. For example:
```
float *mem = alloc();
float *At = mem;
float *Bt = mem + (k*n);
```

python/mxnet/__init__.py
samskalicky: @eric-haibin-lin We are not working on docs/tutorials in this PR. Ive added an entry to the PR description in the "Future/Next-steps" section that we'll address in the next PR.

example/extensions/lib_custom_op/gemm_lib.cc
rondogency: yes I will make the change

wkcn: Maybe we should distinguish CPU and GPU allocation.
For example, an GPU operator need to copy the GPU data into CPU, then compute it in CPU.

samskalicky: @wkcn remember in this PR there is only CPU operator support. The next PR will have GPU support and we’ll make sure to distinguish cpu/gpu allocations.

junrushao1994:
Just talked with @szha and he found out a better way to make `MXTensor` not diverge from the de facto standard tensor format `DLTensor` -- we can have a `MXTensor` class which contains a `DLTensor` as its member. Just like [what TVM did](https://github.com/dmlc/tvm/blob/6b0359b440135b19116ded681be9bee0d7d4c985/include/tvm/runtime/ndarray.h#L242-L294) for its own tensor format.

samskalicky:
> Just talked with @szha and he found out a better way to make `MXTensor` not diverge from the de facto standard tensor format `DLTensor` -- we can have a `MXTensor` class which contains a `DLTensor` as its member. Just like [what TVM did](https://github.com/dmlc/tvm/blob/6b0359b440135b19116ded681be9bee0d7d4c985/include/tvm/runtime/ndarray.h#L242-L294) for its own tensor format.

Thanks @junrushao1994  & @szha thats a good idea. I added the DLTensor to the list of features for the next custom Op PR.

junrushao1994:
Sounds nice!

rondogency:
talked with @szha offline and we agree on putting DLTensor support to the next PR

wkcn:
Hi @szha, this PR provides an approach to create ABI-compatibility custom operator.

There is some limitation in the existing interface ‘MXCustomOpRegister’, which creates custom operator as a task of mx.nd.Custom (https://github.com/apache/incubator-mxnet/blob/master/src/operator/custom/custom-inl.h#L216). It drops the performance, because of the thread lock and restricted workers. Additionally, ‘ MXCustomOpRegister’ does not provide the CUDA stream.

This PR will register custom Op, which is a NNVM op, rather than a task of a NNVM op(mx.nd.Custom).  It does not drop the performance, and it is available to handle all resources, like CUDA stream, memory allocation.

example/extensions/lib_custom_op/gemm_lib.cc
```diff
@@ -0,0 +1,233 @@
...
+/*
+ * Executes C = A * B
+ * inputs[0] = A; inputs[1] = B; outputs[0] = C
+ */
+MXReturnValue forward(std::map<std::string, std::string> attrs,
+                      std::vector<MXTensor> inputs,
+                      std::vector<MXTensor> outputs,
+                      OpResource res) {
+  // simple example of using runtime data type
+  if (inputs[0].dtype == kFloat32) {
```
rondogency: @eric-haibin-lin in the current implementation, we make user hard-code the data type they wanna cast to, which mitigates the burden of using the lengthy MSHADOW_TYPE_SWITCH macro. I am just wondering as a user if you would prefer this or the implementation mxnet is current using?

This question also applies for the upcoming gpu support: does users wanna use Kernel interface, or they wanna write the cuda launch code themselves?

eric-haibin-lin: We should not force users to use MSHADOW_TYPE_SWITCH and Kernel interface. Sometimes as an user I only care about float32. I should have the option to use any existing cuda kernels and easily port it as a custom op

rondogency:
@szha Because we want to make lib_api.h self-contained without any dependencies. The end goal is that the user will only copy lib_api.h to their workspace, install MXNet pip wheel, and write their operator based on the tutorial doc and examples we provided. Using existing MXNet interfaces both introduces extra dependencies, and complicate user experience for users who are not familiar with MXNet source code

rondogency:
@szha we can also copy the implementation of dmlc::registry & dlpack to lib_api.h, but it doesn't solve the diverge problem you described. Plus those changes are not necessarily go to this PR as it focuses on backend registration, and if user really wants the complete MXNet solution we can refactor it in the upcoming PRs.

include/mxnet/lib_api.h
```diff
@@ -63,7 +211,46 @@ struct MXTensor {
...
+      dltensor.dtype.code = kDLInt;
+      dltensor.dtype.bits = 32;
+      break;
+    case kInt8:
+      dltensor.dtype.code = kDLInt;
+      dltensor.dtype.bits = 8;
+      break;
+    case kInt64:
+      dltensor.dtype.code = kDLInt;
+      dltensor.dtype.bits = 64;
```
wkcn: We need an exception when dtype is invalid.

include/mxnet/lib_api.h
```diff
@@ -38,6 +38,154 @@
 
 #define MX_LIBRARY_VERSION 1
 
+/*
+ * Import from DLPack https://github.com/dmlc/dlpack/blob/master/include/dlpack/dlpack.h
```
wkcn: dlpack is 3rdparty of MXNet. Could we include <dlpack/dlpack.h> rather than copy its code?

samskalicky: We want to keep everything in a single header file

wkcn: I see.

include/mxnet/lib_api.h
samskalicky: What if we just do this:
```
default:
      dltensor.dtype.code = 0;
      dltensor.dtype.bits = 0;
      std::cout << "Error! Invalid dtype flag: " << dtype << std::endl;
```
Since we're not including any glog we cant do "LOG(FATAL)".

wkcn: We can throw a C++ exception.

samskalicky: Done, changed to:
https://github.com/apache/incubator-mxnet/pull/15921/commits/5761891d039c49e34ab1da86ed802239d83e545d#diff-5c218b5f3e585d7cc7e4ca8b88470d40R256-R258

include/mxnet/lib_api.h
```diff
@@ -252,7 +253,9 @@ struct MXTensor {
     default:
       dltensor.dtype.code = 0;
       dltensor.dtype.bits = 0;
-      std::cout << "Error! Invalid dtype flag: " << dtype " when constructing MXTensor" << std::endl;  
+      throw std::runtime_error("Error! Invalid dtype flag: "
```
wkcn:
```c++
throw std::runtime_error(std::string("Error! Invalid dtype flag: ")
```

include/mxnet/lib_api.h
```diff
@@ -35,6 +35,7 @@
 #include <string>
 #include <iostream>
 #include <utility>
+#include <exception>
```
wkcn: `#include <stdexcept>`

`std::runtime_error` is in `stdexcept`
https://zh.cppreference.com/w/cpp/error/runtime_error

samskalicky: Done

samskalicky:
@yzhliu any idea why this is failing?
```
  File "/work/mxnet/contrib/tvmop/compile.py", line 20, in <module>
    import tvm
  File "/work/mxnet/3rdparty/tvm/python/tvm/__init__.py", line 23, in <module>
    from . import tensor
  File "/work/mxnet/3rdparty/tvm/python/tvm/tensor.py", line 20, in <module>
    from ._ffi.node import NodeBase, NodeGeneric, register_node, convert_to_node
  File "/work/mxnet/3rdparty/tvm/python/tvm/_ffi/node.py", line 24, in <module>
    from .node_generic import NodeGeneric, convert_to_node, const
  File "/work/mxnet/3rdparty/tvm/python/tvm/_ffi/node_generic.py", line 23, in <module>
    from .base import string_types
  File "/work/mxnet/3rdparty/tvm/python/tvm/_ffi/base.py", line 60, in <module>
    _LIB, _LIB_NAME = _load_lib()
  File "/work/mxnet/3rdparty/tvm/python/tvm/_ffi/base.py", line 52, in _load_lib
    lib = ctypes.CDLL(lib_path[0], ctypes.RTLD_GLOBAL)
  File "/usr/lib/python3.5/ctypes/__init__.py", line 347, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: /work/build/3rdparty/tvm/libtvm.so: file too short
ninja: build stopped: subcommand failed.
```

@wkcn can you please restart the unix-cpu job and see if it was transient?

wkcn:
@samskalicky , I do not have the permission to retrigger the one of CI, but I push a empty commit to retrigger the all.

samskalicky:
Thanks @wkcn, i ran into some flaky test failures and had to push another empty commit:
```
=====================================================================
FAIL: test_quantization_mkldnn.test_requantize_int32_to_int8

  File "/usr/local/lib/python3.5/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/usr/local/lib/python3.5/dist-packages/nose/util.py", line 620, in newfunc
    return func(*arg, **kw)
  File "/work/mxnet/tests/python/mkl/../unittest/common.py", line 177, in test_new
    orig_test(*args, **kwargs)
  File "/work/mxnet/tests/python/mkl/../quantization/test_quantization.py", line 186, in test_requantize_int32_to_int8
    check_requantize_with_symbol((3, 4, 10, 10))
  File "/work/mxnet/tests/python/mkl/../quantization/test_quantization.py", line 181, in check_requantize_with_symbol
    assert_almost_equal(qdata_int8.asnumpy(), qdata_int8_np)
  File "/work/mxnet/python/mxnet/test_utils.py", line 624, in assert_almost_equal
    raise AssertionError(msg)

AssertionError: 
Items are not equal:
Error 1562.500000 exceeds tolerance rtol=1.000000e-05, atol=1.000000e-20 (mismatch 0.083333%).
Location of maximum error: (2, 0, 9, 8), a=63.00000000, b=64.00000000
 ACTUAL: array([[[[ 106,   17,  -56, ...,   88,   -9,  -38],
         [ 107, -120,  -49, ...,  -78,   81,   93],
         [-100,  -90,  -17, ...,   84,   49, -118],...
 DESIRED: array([[[[ 106,   17,  -56, ...,   88,   -9,  -38],
         [ 107, -120,  -49, ...,  -78,   81,   93],
         [-100,  -90,  -17, ...,   84,   49, -118],...

*** Maximum errors for vector of size 1200:  rtol=1e-05, atol=1e-20

  1: Error 1562.500000  Location of error: (2, 0, 9, 8), a=63.00000000, b=64.00000000


FAIL: test_operator_gpu.test_fast_lars

  File "C:\Python37\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\jenkins_slave\workspace\ut-python-gpu\tests\python\gpu\../unittest\common.py", line 177, in test_new
    orig_test(*args, **kwargs)
  File "C:\jenkins_slave\workspace\ut-python-gpu\tests\python\gpu\test_operator_gpu.py", line 328, in test_fast_lars
    check_fast_lars(w_dtype, g_dtype, shapes, ctx, tol1, tol2)
  File "C:\jenkins_slave\workspace\ut-python-gpu\tests\python\gpu\test_operator_gpu.py", line 310, in check_fast_lars
    assert_almost_equal(ref_new_lrs.asnumpy(), mx_new_lrs.asnumpy(), atol=tol2, rtol=tol2)
  File "C:\jenkins_slave\workspace\ut-python-gpu\windows_package\python\mxnet\test_utils.py", line 624, in assert_almost_equal
    raise AssertionError(msg)

AssertionError: 
Items are not equal:
Error 2.844314 exceeds tolerance rtol=1.000000e-06, atol=1.000000e-06 (mismatch 1.724138%).
Location of maximum error: (33,), a=0.01896909, b=0.01897199
 ACTUAL: array([0.00013492, 0.00045911, 0.00021266, ..., 0.00068631, 0.0002449 ,
       0.00085557], dtype=float32)
 DESIRED: array([0.00013493, 0.00045911, 0.00021266, ..., 0.00068631, 0.0002449 ,
       0.00085556], dtype=float32)

*** Maximum errors for vector of size 58:  rtol=1e-06, atol=1e-06

  1: Error 2.844314  Location of error: (33,), a=0.01896909, b=0.01897199
```

samskalicky:
More flaky test failures:
```
FAIL: test_operator_gpu.test_bulking_operator_gpu
    self.test(*self.arg)
    orig_test(*args, **kwargs)
  File "/work/mxnet/tests/python/gpu/test_operator_gpu.py", line 2398, in test_bulking_operator_gpu
    _test_bulking(_test_bulking_in_process)
  File "/work/mxnet/tests/python/gpu/test_gluon_gpu.py", line 553, in _test_bulking
    time_per_iteration):
  File "/work/mxnet/tests/python/gpu/../unittest/common.py", line 334, in run_in_spawned_process
    assert p.exitcode == 0, "Non-zero exit code %d from %s()." % (p.exitcode, func.__name__)
AssertionError: Non-zero exit code -6 from _test_bulking_in_process().
common: INFO: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=298254377 to reproduce.
```

samskalicky:
pipeline timed out, for centos-cpu, had to push again.

rondogency:
@wkcn 1.6 code freeze is tomorrow, so are you ok with this one not going into the 1.6 release? It is because none of us have time to maintain it until late November. After code freeze then we can merge it on Friday, and user can use nightly build to access this feature.

wkcn:
@rondogency No problem : )

wkcn:
Hi @samskalicky and @rondogency , is it ready to merge this PR?

samskalicky:
> Hi @samskalicky and @rondogency , is it ready to merge this PR after CI passes?

Yes! We're soooooo ready to merge :) 

Thanks @zachgk for reruning the unix_cpu job!

wkcn:
I will merge this PR after the CI passes.
Thank all contributors!

rondogency:
@wkcn Big thank to Jackie for the merging work!

tests/python/unittest/test_extensions.py
```diff
@@ -21,15 +21,16 @@
...
-from mxnet.test_utils import download, is_cd_run
+from mxnet.test_utils import download, is_cd_run, assert_almost_equal
 
 def check_platform():
     return platform.machine() not in ['x86_64', 'AMD64']
 
 @unittest.skipIf(check_platform(), "not all machine types supported")
 @unittest.skipIf(is_cd_run(), "continuous delivery run - ignoring test")
-def test_library_loading():
+def test_custom_op():
```
TaoLv: It has a strong assumption that the case will be called from mxnet root folder. Otherwise, the libsample_lib.so will not be found.

```
$ cd tests/python/unittest/
$ nosetests -v test_extensions:test_custom_op
test_extensions.test_custom_op ... ERROR

ERROR: test_extensions.test_custom_op
  File "/home/lvtao/miniconda3/envs/mxnet/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/lvtao/Workspace/mxnet-official/tests/python/unittest/test_extensions.py", line 41, in test_custom_op
    raise MXNetError("library %s not found " % lib)
mxnet.base.MXNetError: library libsample_lib.so not found

Ran 1 test in 0.005s

FAILED (errors=1)

```

wkcn: Is it possible to use mx.libinfo.find_lib_path to find the library?
https://github.com/apache/incubator-mxnet/blob/93228649340bcacb8056d47d8f6f8a78a9805ae4/python/mxnet/libinfo.py#L26

rondogency: Gotcha I will fix it in the next PR

PR #15938, opened 2019-08-18
Tvm broadcast backward

hzfan:
## Description ##
Use tvm to implement vadd backward with broadcast.

### Changes ###
- add vadd backward

## Comments ##
- As vadd is not a Numpy op, 0-dim and 0-size is not supported.
- For now, I implemented infra-level things on op-level, like
  - dispatch of different req
  - dispatch of backward

I think code for these may be further reused in the future. It'll be great if we can have a consistent interface for tvm op and hide the dispatch of things like req and backward.

Thank @yzhliu and @junrushao1994 for the brilliant "compressed bit string" idea.

### DIFF
--- a/contrib/tvmop/basic/ufunc.py
+++ b/contrib/tvmop/basic/ufunc.py
@@ -52,6 +52,15 @@ def vadd_gpu(dtype, ndim):
     return s, [A, B, C]
 
 
+def assign_by_req(a, req):
+    b = tvm.placeholder(a.shape, name='assign_by_req_b', dtype=a.dtype)
+    if (req == "kAddTo"):
+        c = tvm.compute(a.shape, lambda *idx: a[idx] + b[idx])
+    else:
+        c = tvm.compute(a.shape, lambda *idx: a[idx])
+    return b, c
+
+
 def reduce_axes(X, axes, reducer):
     def get_index(idx, ridx):
         j = 0
@@ -71,31 +80,34 @@ def reduce_axes(X, axes, reducer):
     return ret
 
 
-def compute_backward_vadd(dtype, ndim, reduce1st):
+def compute_backward_vadd(dtype, ndim, reduce1st, req):
     axes = ([reduce1st, 1 - reduce1st] * ndim)[:ndim]
     X = tvm.placeholder([tvm.var() for _ in range(ndim)], name='X', dtype=dtype)
     reducer = tvm.comm_reducer(lambda x, y: x + y,
         lambda t: tvm.const(0, dtype=t), name="sum")
     ret = reduce_axes(X, axes, reducer)
-    s = tvm.create_schedule(ret.op)
-    return s, X, ret, [ret]
+    in_grad_a, in_grad = assign_by_req(ret, req)
+    s = tvm.create_schedule(in_grad.op)
+    return s, X, in_grad_a, in_grad, [ret, in_grad]
 
 
-@defop(name="backward_vadd", target="cpu", dtype=AllTypes,
-       ndim=list(range(1, 6)), reduce1st=[0, 1], attrs=["reduce1st"])
-def backward_vadd(dtype, ndim, reduce1st):
-    s, X, ret, c_list = compute_backward_vadd(dtype, ndim, reduce1st)
+@defop(name="backward_vadd", target="cpu", dtype=AllTypes, 
+       ndim=list(range(1, 6)), reduce1st=[0, 1],
+       req=["kWriteTo", "kAddTo"], attrs=["reduce1st", "req"])
+def backward_vadd(dtype, ndim, reduce1st, req):
+    s, X, in_grad_a, in_grad, c_list = compute_backward_vadd(dtype, ndim, reduce1st, req)
     for t in c_list:
         axes = [axis for axis in t.op.axis]
         fused = s[t].fuse(*axes)
         s[t].parallel(fused)
-    return s, [X, ret]
+    return s, [X, in_grad_a, in_grad]
 
 
 @defop(name="cuda_backward_vadd", target="gpu", dtype=["float32", "float64"],
-       ndim=list(range(1, 6)), reduce1st=[0, 1], attrs=["reduce1st"])
-def backward_vadd_gpu(dtype, ndim, reduce1st):
-    s, X, ret, c_list = compute_backward_vadd(dtype, ndim, reduce1st)
+       ndim=list(range(1, 6)), reduce1st=[0, 1],
+       req=["kWriteTo", "kAddTo"], attrs=["reduce1st", "req"])
+def backward_vadd_gpu(dtype, ndim, reduce1st, req):
+    s, X, in_grad_a, in_grad, c_list = compute_backward_vadd(dtype, ndim, reduce1st, req)
     num_thread = 64
     for t in c_list:
         block_x = tvm.thread_axis("blockIdx.x")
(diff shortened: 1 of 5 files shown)


junrushao1994:
CC @tqchen if you have bandwidth

contrib/tvmop/basic/ufunc.py
```diff
@@ -48,3 +50,71 @@ def vadd_gpu(dtype, ndim):
     s[C].bind(bx, tvm.thread_axis("blockIdx.x"))
     s[C].bind(tx, tvm.thread_axis("threadIdx.x"))
     return s, [A, B, C]
+
+
+def assign_by_req(a, req):
```
yzhliu: move to sth like common.py?

src/operator/contrib/tvmop/ufunc.cc
```diff
@@ -37,29 +38,88 @@ namespace op {
...
+    }
+    TShape oshape(ov.begin(), ov.end()), ishape(iv.begin(), iv.end());
+    TBlob ograd_tvm(ograd.reshape(oshape).dltensor());
+    TBlob igrad_tvm(igrad.reshape(ishape).dltensor());
+    std::string funcname = std::string(func) + "reduce1st_" + std::to_string(flag);
+    // dispatch by req
+    funcname += "req_";
+    MXNET_ASSIGN_REQ_SWITCH(req[k], req_type, {
+      if (req_type == kWriteTo) {
+                funcname += "kWriteTo";
```
yzhliu: alignment

src/operator/contrib/tvmop/ufunc.cc
```diff
@@ -37,29 +38,88 @@ namespace op {
...
+                                     const std::vector<OpReqType>& req,
+                                     const std::vector<TBlob>& outputs) {
+  CHECK_EQ(inputs.size(), 1U);
+  CHECK_EQ(outputs.size(), 2U);
+  int ndim = inputs[0].shape_.ndim();
+  for (int k = 0; k < 2; ++k) {
+    // dispatch by backward
+    std::vector<int> ov, iv;
+    const TBlob& ograd = inputs[0], igrad = outputs[k];
+    bool flag = ograd.size(0) != igrad.size(0);
```
yzhliu: better to use int and explicitly assign the value.

src/operator/contrib/tvmop/ufunc.cc
```diff
@@ -37,29 +38,88 @@ namespace op {
...
+      } else {
+        ov.back() *= ograd.size(i);
+      }
+    }
+    for (int i = flag; i < ov.size(); i += 2) {
+      iv.push_back(ov[i]);
+    }
+    TShape oshape(ov.begin(), ov.end()), ishape(iv.begin(), iv.end());
+    TBlob ograd_tvm(ograd.reshape(oshape).dltensor());
+    TBlob igrad_tvm(igrad.reshape(ishape).dltensor());
```
yzhliu: please add some comments to elaborate the ideas.

src/operator/contrib/tvmop/ufunc.cc
```diff
@@ -37,29 +38,88 @@ namespace op {
...
+                                     const std::vector<TBlob>& outputs) {
+  CHECK_EQ(inputs.size(), 1U);
+  CHECK_EQ(outputs.size(), 2U);
+  int ndim = inputs[0].shape_.ndim();
+  for (int k = 0; k < 2; ++k) {
+    // dispatch by backward
+    std::vector<int> ov, iv;
+    const TBlob& ograd = inputs[0], igrad = outputs[k];
+    bool flag = ograd.size(0) != igrad.size(0);
+    for (int i = 0; i < ndim; ++i) {
```
reminisce: If my understanding is correct, there seems to be an assumption that `ograd.ndim = igrad.ndim`, which is not necessarily true. I think you need to prepend axes before `igrad` if `igrad.ndim < ograd.ndim` and then use the logic here.

hzfan: Yes, `igrad.ndim = ograd.ndim` is assumed.

@yzhliu suggests padding the input to 5-dim, which is the largest possible dim supported by this op. The padding will 1) reduce the number of kernels 2) handle the `igrad.ndim < ograd.ndim` issue. But there may be loss in performance.

I think prepending axes before `igrad` to make it `ograd.dim` requires more kernels, but the performance is better. It is a tradeoff.

contrib/tvmop/basic/ufunc.py
hzfan: Shall we use the existing contrib/tvmop/utils.py or create a contrib/tvmop/basic/common.py?

src/operator/contrib/tvmop/ufunc.cc
hzfan: What about expand it into a if-else?

src/operator/contrib/tvmop/ufunc.cc
reminisce: Please correct me if my understanding is wrong, but don't you still need kernels generated for `ndims < 5` since you will collapse consecutive dimensions where reduction is performed? For example, given a 5d shape `(2, 3, 4, 5, 6)`, and perform reduction on `axis=(1, 2)`, the `tblob` will be first reshaped into `(2, 12, 30)`, and then reduce on `axis=1`.  In this case, do you need a kernel generated for 3D shapes?

hzfan: I think we can pad the shape after dimension collapse. In this case, the `tblob` will be reshaped into `(2, 12, 30, 1, 1)` and then reduce on `axis=[1, 3]`.

reminisce: I see. I am in favor of the approach with less kernels generated. We can revisit the performance concern if that turns out to be an issue.

contrib/tvmop/basic/ufunc.py
yzhliu: utils.py is fine

src/operator/contrib/tvmop/ufunc.cc
yzhliu: sounds good

src/operator/contrib/tvmop/ufunc.cc
hzfan: Added in ufunc.py

contrib/tvmop/basic/ufunc.py
hzfan: Moved.

src/operator/contrib/tvmop/ufunc.cc
hzfan: I pushed a new version, where the inputs and outputs are padded to 5 dim.

src/operator/contrib/tvmop/ufunc.cc
hzfan: Aligned.

contrib/tvmop/basic/ufunc.py
```diff
@@ -48,3 +50,71 @@ def vadd_gpu(dtype, ndim):
...
+    b = tvm.placeholder(a.shape, name='assign_by_req_b', dtype=a.dtype)
+    if (req == "kAddTo"):
+        c = tvm.compute(a.shape, lambda *idx: a[idx] + b[idx])
+    else:
+        c = tvm.compute(a.shape, lambda *idx: a[idx])
+    return b, c
+
+
+def reduce_axes(X, axes, reducer):
```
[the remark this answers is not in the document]
hzfan: Yes. Added in ufunc.py

PR #15940, opened 2019-08-19
Add boolean ndarray

reminisce:
## Description ##
- Added boolean ndarray infra in mshadow.
- Implemented comparison operators: equal, not_equal, greater, greater_equal, less, and less_equal to use `np.bool_` as their output tensors' `dtype`.

### Essentials ###

## Comments ##
Follow-up work includes:
- More operators that can consume boolean ndarrays. To list a few that must be finished before the next release:
  - sum
  - cast
  - boolean_mask
- `ndarray` [boolean indexing](https://www.numpy.org/devdocs/reference/arrays.indexing.html#boolean-array-indexing)

Thank @yzhliu @hzfan for the help on debugging.


src/operator/numpy/np_elemwise_broadcast_op.cc
```diff
@@ -182,5 +114,211 @@ MXNET_OPERATOR_REGISTER_NP_BINARY_SCALAR(_npi_rpower_scalar)
...
+  const size_t num_args = 3;  // one input tensor, one scalar param, and one output
+  type_codes.resize(num_args);
+  values.resize(num_args);
+
+  // input tensor setup
+  type_codes[0] = kArrayHandle;
+  values[0].v_handle = const_cast<DLTensor*>(&(tblobs[0].dltensor()));
+
+  // scalar param
+  type_codes[1] = kDLFloat;
```
reminisce: @yzhliu Since I need to pass a `double` param to the op func generated by TVM, I cannot use the `Call` function defined in the `TVMOpModule`. I moved the logic of preparing `TVMArgs` up here from the `Call` function to MXNet op's `FCompute` function and added an independent `CallEx` in `TVMOpModule` to just invoke the kernel. We can discuss the change of the API to cater for more use cases.

Makefile
```diff
@@ -473,11 +473,13 @@ CFLAGS += -I$(TVM_PATH)/include -DMXNET_USE_TVM_OP=1
 LDFLAGS += -L$(ROOTDIR)/lib -ltvm_runtime -Wl,-rpath,'$${ORIGIN}'
 
 TVM_USE_CUDA := OFF
+TVM_OP_CUDA_ARCH := NONE
```
marcoabreu: Any particular reason you are introducing a second set instead of using the arch set variable we already have? In which use case would these two differ?

src/operator/operator_tune.cc
```diff
@@ -84,6 +85,16 @@ struct static_init_var {
...
+  __macro$(__VA_ARGS__, float); \
+  __macro$(__VA_ARGS__, double); \
+  __macro$(__VA_ARGS__, mshadow::half::half_t); \
+  __macro$(__VA_ARGS__, uint8_t); \
+  __macro$(__VA_ARGS__, int8_t); \
+  __macro$(__VA_ARGS__, int32_t); \
+  __macro$(__VA_ARGS__, int64_t); \
+  __macro$(__VA_ARGS__, bool)
+
 #define IMPLEMENT_WORKLOAD_VALUE_FOR_TYPE(__op$, __typ$) \
```
larroy: could you please add a comment to this macro to clarify?

tests/python/unittest/test_numpy_op.py
```diff
@@ -240,27 +236,38 @@ def is_int(dtype):
     in_data_dim = random.choice([2, 3, 4])
     shape = rand_shape_nd(in_data_dim, dim=3)
     acc_type = {'float16': 'float32', 'float32': 'float64', 'float64': 'float64',
-                'int8': 'int32', 'int32': 'int64', 'int64': 'int64'}
+                'int8': 'int32', 'int32': 'int64', 'int64': 'int64', 'bool': 'int64'}
     for hybridize in [False, True]:
```
larroy: would using https://docs.python.org/3.7/library/itertools.html#itertools.product help readability of the code and make it less nested?

Makefile
reminisce: Reverted. Thanks for the suggestion.

tests/python/unittest/test_numpy_op.py
reminisce: Thanks for the suggestion. Will consider refactoring it in the following PRs.

PR #15953, opened 2019-08-20
Add Median,p50,p99 to python profiler

ChaiBapchya:
## Description ##
Profile more metrics (than just average; also fix incorrect avg calculation)

### Essentials ###


benchmark/opperf/utils/profiler_utils.py
```diff
@@ -228,10 +229,16 @@ def python_profile(func):
 
     @functools.wraps(func)
     def python_profile_it(*args, **kwargs):
-        start_time = time.perf_counter()    # 1
-        res = func(*args, **kwargs)
-        end_time = time.perf_counter()      # 2
-        run_time = end_time - start_time    # 3
+        runs = args[1]
+        modified_args = (args[0], 1)
```
apeforest: what is this for?

apeforest:
Thanks for the contribution. How to enable the percentile output for users? Or it comes as default?

ChaiBapchya:
Comes as default if chosen profiler as python

benchmark/opperf/utils/profiler_utils.py
ChaiBapchya: so Im trying to reuse the existing code.
Other function uses runs and runs a for loop here
https://github.com/ChaiBapchya/incubator-mxnet/blob/bb3e9fc65b756716b4f010c98ed2cfe733f52386/benchmark/opperf/utils/ndarray_utils.py#L45

I reuse the above function. MXNet profiler is able to log the data on a per run basis without explicitly storing that info after every run/iteration

But we need to do that. So that's what I am doing it here

https://github.com/ChaiBapchya/incubator-mxnet/blob/bb3e9fc65b756716b4f010c98ed2cfe733f52386/benchmark/opperf/utils/profiler_utils.py#L236

Not sure if this is the best way. But I wanted to reuse the code and ensure it is backward compatible. Hence this workaround.

Let me know if I should clarify further.

benchmark/opperf/utils/common_utils.py
```diff
@@ -108,12 +112,19 @@ def _prepare_op_benchmark_result(op, op_bench_result, profiler):
             inputs = value
         elif "avg_time" in key:
             avg_time = value
+        elif "p50_" in key:
```
apeforest: why not p50_time in key?

ChaiBapchya: coz it's enough to check if p50 exists in key
there won't be p50_a and p50_b

ChaiBapchya: Added it nonetheless

benchmark/opperf/utils/profiler_utils.py
```diff
@@ -228,10 +232,16 @@ def python_profile(func):
 
     @functools.wraps(func)
     def python_profile_it(*args, **kwargs):
-        start_time = time.perf_counter()    # 1
-        res = func(*args, **kwargs)
-        end_time = time.perf_counter()      # 2
-        run_time = end_time - start_time    # 3
+        runs = args[1]
+        modified_args = (args[0], 1, args[2])
```
access2rohit: Can you give an example case what are now sent in modified_args ?

ChaiBapchya: Currently when you run opperf
args are - `[op, runs, args]`
runs  is the # of times we want to run this operator

For python profiler to work, we need to call the operator explicitly those many times
Hence runs = args[1] stores the number of times user wanted to run the operator

Now, the alias takes the argument to run MXNet profiler on it
We're overwriting on that, hence we pass run=1 to MXNet profiler
modified args = `[op, 1, args]`
Does that answer your question?

PR #15969, opened 2019-08-22
Partitioning Gluon HybridBlocks

samskalicky:
## Description ##
Adds partitioning support for Gluon HybridBlocks. This is a continuation of the partitioning support for Symbol #15886

## Design ##
In Gluon, a HybridBlock contains a Symbol after hybridizing and executing a forward pass. The Symbol is contained and managed within the block. The partitioning logic will be integrated into the hybridize flow. 

There are many ways to create a Gluon Hybrid block and after this process, users call the `hybridize()` function to start the flow. We add two new arguments to support partitioning: `backend` which is a string corresponding to the `subgraph_backend` name, and `opt_args` which is a map of arguments that should be passed to the `subgraph_property` during partitioning. These values are stored until used during the first inference call. Heres an example specifying these new arguments:
```
net = create()
net.hybridize(backend='default', opt_args={excluded_ops=['BatchNorm']}) 
```

Notice that in the above example, the new arguments have the same value as the example in #15886. These arguments will ultimately be passed to a call to the `optimize_for` API. 

In the Gluon, the hybridize flow starts before the first inference. The Symbol object is created in the `_build_cache` function:
https://github.com/apache/incubator-mxnet/blob/bd67723da96e6d36e72c9a42535a4fe68f234a71/python/mxnet/gluon/block.py#L933-L934
We'll add a new line of code to partition it and pass the new arguments from the hybridize call:
```
def _build_cache(self, *args):
        data, out = self._get_graph(*args)
        if self.backend:
                out = out.optimize_for(self.backend, **self.opt_args)
```

This supports the partitioning flow without shape/type propagation. Some backends do not need shapes and types so there is no reason to require it for all backends. Other backends will require shapes and types in order to partition the model correctly (examples being backends that only support float16 and not float32, or only support small shapes and not large ones). 

For the partitioning with with shape/type propagation we can get the args to the model from the parameters in the Gluon block. By default, the initialization of Gluon parameters may be delayed. If the parameters are not initialized yet, we'll continue with the flow shown in the code snippet above that does not infer shapes/types. 

In Gluon users can force initialization ([see this guide](https://d2l.ai/chapter_deep-learning-computation/deferred-init.html#forced-initialization)) and if all parameters are initialized after calling `hybridize` and setting the `backend` name, we will pass the arguments from the Gluon parameters into the `optimize_for` API to infer shapes/types before partitioning. This gives the user the control over partitioning in the same way that they do for Symbol API. Heres a code snippet to produce the arg array and pass it to `optimize_for`:
```
arg_array = []
try:
    for name in out.list_arguments():
        if name in data_names.keys():
            arg_array.append(args[data_names[name]])
        else:
            arg_array.append(params.get(name))
except DeferredInitializationError:
    arg_array = None
except RuntimeError:
    arg_array = None
out = out.optimize_for(self._backend, arg_array, ctx, **self._backend_args)
```

The context will be gathered from the inputs to the model like this:
```
ctx = args[0].context
```
Context is required to infer storage types.

### Note 
Partitioning is done as part of the hybridize flow, when building the cachedOp. So if shapes change between infer calls the graph is not re-partitioned. 

### Essentials ###

### Changes ###
- Refactor subgraph tests, returns input names and shapes (needed for gluon). improves test architecture to be more clear about what we're testing (subgraph API, optimize for, gluon).


leezu:
Should this be run prior to training or prior to exporting the HybridBlock? Could/Should it be run automatically?

samskalicky:
Waiting on #15886 to be merged to re-use optimize_for API call on symbol

python/mxnet/gluon/block.py
```diff
@@ -932,6 +932,9 @@ def _get_graph(self, *args):
 
     def _build_cache(self, *args):
         data, out = self._get_graph(*args)
+        if self._backend:
+            # To do: pass in all arguments
+            out = out.optimize_for(self._backend)
```
samskalicky: also pass `self.optargs` to `optimize_for` like:
```
out = out.optimize_for(self._backend, **self.optargs)
```

python/mxnet/gluon/block.py
```diff
@@ -1178,6 +1183,28 @@ def hybrid_forward(self, F, x, *args, **kwargs):
         # pylint: disable= invalid-name
         raise NotImplementedError
 
+    def optimize_for(self, backend, **kwargs):
```
samskalicky: we're getting rid of this function right?

guanxinq: Yes. Removed.

python/mxnet/gluon/block.py
guanxinq: Updated. Thanks!

python/mxnet/gluon/block.py
```diff
@@ -1040,7 +1043,9 @@ def register_child(self, block, name=None):
         super(HybridBlock, self).register_child(block, name)
         self._clear_cached_op()
 
-    def hybridize(self, active=True, **kwargs):
+    def hybridize(self, active=True, backend='', optargs={}, **kwargs):
```
samskalicky: maybe `optargs` should be `backend_args` for consistency? we will continue to pass kwargs to the C++ CachedOp instance. So want to disambiguate against options that will be used for partitioning.

python/mxnet/gluon/block.py
```diff
@@ -932,6 +932,9 @@ def _get_graph(self, *args):
 
     def _build_cache(self, *args):
         data, out = self._get_graph(*args)
+        if self._backend:
+            # To do: pass in all arguments
+            out = out.optimize_for(self._backend, **self.optargs)
```
samskalicky: can we add a comment that we're partitioning the graph here?

guanxinq: Added.

python/mxnet/gluon/block.py
eric-haibin-lin: please set default to `None` instead of `''`

python/mxnet/gluon/block.py
eric-haibin-lin: Why not do it inside `self._get_graph`?

samskalicky: agreed, lets move it here:
https://github.com/apache/incubator-mxnet/blob/557a29b7024e86979b5927ae81209d8c32d8511d/python/mxnet/gluon/block.py#L927-L928

samskalicky:
Thanks @guanxinq for the latest update! I think we need to call optimize for again here too when we create a SymbolBlock, otherwise the partitioning wont happen:
https://github.com/apache/incubator-mxnet/blob/3d18974fdc990b7def2401fae8e46fb0b030442f/python/mxnet/gluon/block.py#L1340
Because once `self._cached_graph` is set the previous code wont be executed in the `_get_graph` function

samskalicky:
@eric-haibin-lin we shouldnt partition inside `_get_graph` since its also called here:
https://github.com/apache/incubator-mxnet/blob/3d18974fdc990b7def2401fae8e46fb0b030442f/python/mxnet/gluon/block.py#L1068

python/mxnet/gluon/block.py
```diff
@@ -926,6 +928,11 @@ def _get_graph(self, *args):
                 out = self.hybrid_forward(symbol, *grouped_inputs, **params)  # pylint: disable=no-value-for-parameter
             out, self._out_format = _flatten(out, "output")
 
+            if self._backend:
+                # To do: pass in all arguments
```
leezu: Will this be addressed in this PR or later?

samskalicky: Yes, it will be addressed in this PR. we're working on it now

python/mxnet/gluon/block.py
guanxinq: Just as Sam mentioned "Because once self._cached_graph is set the previous code wont be executed in the _get_graph function". It is better to keep this change.

python/mxnet/gluon/block.py
```diff
@@ -954,6 +955,18 @@ def _build_cache(self, *args):
             unused = ', '.join(list(param_names - set(used_param_names)))
             warnings.warn("Parameter %s is not used by any computation. "
                           "Is this intended?"%unused, stacklevel=4)
+        if self._backend:
+            arg_array = []
+            ctx = None
+            for name in out.list_arguments():
+                if name in data_names.keys():
+                    arg_array.append(args[data_names[name]])
```
samskalicky: should we assign ctx here too?

python/mxnet/gluon/block.py
```diff
@@ -954,6 +955,18 @@ def _build_cache(self, *args):
             unused = ', '.join(list(param_names - set(used_param_names)))
             warnings.warn("Parameter %s is not used by any computation. "
                           "Is this intended?"%unused, stacklevel=4)
+        if self._backend:
+            arg_array = []
+            ctx = None
```
samskalicky: I think we only need to assign `ctx` to `args[0].context` since in gluon there must be at least one input to the model and all inputs/params have to be on the same context. Then we dont need to do it in the loop.

python/mxnet/gluon/block.py
```diff
@@ -954,6 +955,18 @@ def _build_cache(self, *args):
...
+            for name in out.list_arguments():
+                if name in data_names.keys():
+                    arg_array.append(args[data_names[name]])
+                else:
+                    if ctx is None:
+                        ctx = params.get(name)._ctx_list[0]
+                    arg_array.append(ndarray.random.uniform(shape=params.get(name)._shape))
```
samskalicky: can we just make this:
```
arg_array.append(params.get(name))
```

python/mxnet/gluon/block.py
guanxinq: Fixed

python/mxnet/gluon/block.py
guanxinq: Done

tests/python/unittest/test_subgraph_op.py
```diff
@@ -18,351 +18,448 @@
...
+
+    # compare outputs
+    outputs1 = exe1.outputs
+    outputs2 = exe2.outputs
+    assert len(outputs1) == len(outputs2)
+    for i in range(len(outputs1)):
+        assert_almost_equal((outputs1[i] - outputs2[i]).abs().sum().asnumpy(), np.zeros(shape=(1,)))
+
+def check_subgraph_exe6(sym, subgraph_backend, op_names):
+    """Call optimize_for to trigger graph partitioning without infer shapes/types before,
```
samskalicky: lets change this comment, in this test we are inferring/shapes types

tests/python/unittest/test_subgraph_op.py
```diff
@@ -18,351 +18,448 @@
...
+    exe2.forward()
+    
+    # compare outputs
+    outputs1 = exe1.outputs
+    outputs2 = exe2.outputs
+    assert len(outputs1) == len(outputs2)
+    for i in range(len(outputs1)):
+        assert_almost_equal((outputs1[i] - outputs2[i]).abs().sum().asnumpy(), np.zeros(shape=(1,)))
+    
+def check_subgraph_exe9(sym, subgraph_backend, op_names):
```
samskalicky: Need a comment here explaining the unique-ness of this test

tests/python/unittest/test_subgraph_op.py
```diff
@@ -18,351 +18,448 @@
...
+    inputs = [mx.sym.var(i, dtype=mx_real_t) for i in sym[1]]
+    sym_block = nn.SymbolBlock(sym[0], inputs)
+    sym_block.initialize()
+    shapes = sym[2]
+    x = [mx.nd.random.uniform(shape=s) for s in shapes]
+    outputs1 = sym_block(*x)
```
samskalicky: In this test you're doing an inference before hybridizing with the subgraph_backend. In this case that means we'll pass shapes/types. Can we add another function without this to test not inferring shapes/types?

tests/python/unittest/test_subgraph_op.py
```diff
@@ -18,351 +18,448 @@
...
+def test_subgraph_backend_sym_v2():
+    check_subgraph_backend_sym('default_v2')
+
+def test_subgraph_backend_gluon():
+    check_subgraph_backend_gluon('default')
+
+def test_subgraph_backend_gluon_v2():
+    check_subgraph_backend_gluon('default_v2')
+
+def test_subgraph_backend_gluon_ext1():
```
samskalicky: Need description of function/test

tests/python/unittest/test_subgraph_op.py
```diff
@@ -18,351 +18,448 @@
...
+    def get_net():
```
samskalicky: if we dont call get_net more than once lets move the function body out

tests/python/unittest/test_subgraph_op.py
```diff
@@ -18,351 +18,448 @@
...
+    net.hybridize(backend = subgraph_backend)
+    outputs2 = net(x)
+    check_call(_LIB.MXRemoveSubgraphPropertyOpNamesV2(c_str(subgraph_backend)))
+
+    # compare outputs
+    assert len(outputs1) == len(outputs2)
+    for i in range(len(outputs1)):
+        assert_almost_equal((outputs1[i] - outputs2[i]).abs().sum().asnumpy(), np.zeros(shape=(1,)))
+
+def test_subgraph_backend_gluon_ext2():
```
samskalicky: need description

tests/python/unittest/test_subgraph_op.py
```diff
@@ -18,351 +18,448 @@
...
+
+        def hybrid_forward(self, F, x):
+            x = F.relu(self.fc1(x))
+            x = F.relu(self.fc2(x))
+            return self.fc3(x)
+
+    x = nd.random.normal(shape=(1, 512))
+    net = Net()
+    net.collect_params().initialize()
+    outputs1 = net(x)
```
samskalicky: same deal here, you're running an inference before partitioning

tests/python/unittest/test_subgraph_op.py
```diff
@@ -18,351 +18,448 @@
...
+    ret1 = mx.sym.cos(ret)
+    ret2 = mx.sym.sin(ret)
+    ret = ret1 + ret2
+    ret = mx.sym.BatchNorm(ret)
+    ret = mx.sym.BatchNorm(ret)
+    # Return the same and shape of 'data' and auxiliary states
+    return  (ret, ['data', *ret.list_auxiliary_states()], [(2, 3, 10, 10), (3,), (3,), (3,), (3,)])
+
+def network_structure_4():
+    # the last op has multiple duplicate outputs
```
larroy: I like the comments on these tests

tests/python/unittest/test_subgraph_op.py
```diff
@@ -18,351 +18,448 @@
...
+    for i in range(len(exe.outputs)):
+        assert_almost_equal((exe.outputs[i] - partitioned_exe.outputs[i]).abs().sum().asnumpy(),
+                            np.zeros(shape=(1,)))
+
+def check_subgraph_exe2(sym, subgraph_backend, op_names):
+    """Use env var MXNET_SUBGRAPH_BACKEND=default to trigger graph partitioning in simple_bind
+    and compare results of the partitioned sym and the original sym."""
+    def get_executor(sym, subgraph_backend=None, op_names=None, original_exec=None):
+        if subgraph_backend is not None:
+            os.environ['MXNET_SUBGRAPH_BACKEND'] = subgraph_backend
```
larroy: nit: there's a decorator somewhere to restore the environment using with, this is not really safe if there's an exception and the environment is left in a bad state.

tests/python/unittest/test_subgraph_op.py
```diff
@@ -18,351 +18,448 @@
...
+                nn.Dense(2))
+        return net
+
+    x = nd.random.normal(shape=(1, 512))
+    net = get_net()
+    net.collect_params().initialize()
+    outputs1 = net(x)
+
+    subgraph_backend = 'default'
+    op_names = []
```
samskalicky: ???

python/mxnet/gluon/block.py
```diff
@@ -954,6 +955,22 @@ def _build_cache(self, *args):
...
+            arg_array = []
+            ctx = args[0].context
+            try:
+                for name in out.list_arguments():
+                    if name in data_names.keys():
+                        arg_array.append(args[data_names[name]])
+                    else:
+                        arg_array.append(params.get(name).data())
+
+            except DeferredInitializationError:
```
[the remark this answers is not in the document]
samskalicky: Per offline discussion with @leezu, Gluon should always give shapes as long as we do this:
https://github.com/apache/incubator-mxnet/blob/93b72c03d9871b71fa274b3c34aa6a4473be2155/python/mxnet/gluon/block.py#L1032-L1040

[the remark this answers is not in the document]
leezu: Ie. instead of handling except DeferredInitializationError as `arg_array = None`, you need to infer the parameter shapes earlier on based on the symbol, before creating the CachedOp

python/mxnet/gluon/block.py
```diff
@@ -954,6 +955,22 @@ def _build_cache(self, *args):
             unused = ', '.join(list(param_names - set(used_param_names)))
             warnings.warn("Parameter %s is not used by any computation. "
                           "Is this intended?"%unused, stacklevel=4)
+        if self._backend:
+            arg_array = []
+            ctx = args[0].context
```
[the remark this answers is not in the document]
guanxinq: Sure.

tests/python/unittest/test_subgraph_op.py
guanxinq: Updated.

tests/python/unittest/test_subgraph_op.py
guanxinq: Added.

tests/python/unittest/test_subgraph_op.py
guanxinq: Added.

tests/python/unittest/test_subgraph_op.py
guanxinq: Done.

tests/python/unittest/test_subgraph_op.py
guanxinq: Added.

tests/python/unittest/test_subgraph_op.py
guanxinq: Updated.

tests/python/unittest/test_subgraph_op.py
guanxinq: Wiil add tests without the inference before partition and refactor tests.

python/mxnet/gluon/block.py
```diff
@@ -1040,7 +1061,12 @@ def register_child(self, block, name=None):
         super(HybridBlock, self).register_child(block, name)
         self._clear_cached_op()
 
-    def hybridize(self, active=True, **kwargs):
+    def hybridize(self, active=True, backend=None, backend_args=None, **kwargs):
+        self._backend = backend
+        if backend_args is None:
+            self._backend_args = {}
```
mseth10: do we really need this? isn't it default?

samskalicky: Yes, we "splat" the backend_args later in the code, so has to be a dictionary or the "splat" operator (ie. the *) will error out

tests/python/unittest/test_subgraph_op.py
samskalicky: This is old code, we're not adding it in this PR just refactoring it. So we'll ignore the nit for this PR

python/mxnet/gluon/block.py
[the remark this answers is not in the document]
samskalicky: @leezu This is not always guaranteed to work. The input shapes may not be enough to infer shapes through the model. And the shapes inferred may not the be the same as the actual shapes used at inference time. 

Some backends may not be ok with this. They may error out saying the shapes presented at partition time do not match those at runtime. So the safe approach is either to provide the exact shapes at partition time, or not at all.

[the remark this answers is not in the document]
samskalicky: Plus, during training we dont have the params yet (shapes/types) so we need to partition without them. For inference we do have the shapes/types, so it makes sense to have a flow that works for both use-cases.

The question is how we enable this flow with Gluon-style handling (ie. just make it work internally by catching the exception, or have the user pass in a flag to explicitly skip the attempt to get the params)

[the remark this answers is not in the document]
leezu: DeferredInitializationError is with respect to parameter shapes.

> The input shapes may not be enough to infer shapes through the model.

If the input shapes are not sufficient to infer parameter shapes, forward will not be possible.
It's correct that dynamic shape operators prevent shape inference of the complete model.

My concern here is only to handle correctly the case where parameter shape can be inferred correctly and the PR currently does not. Let's discuss more about the complete model shape inference.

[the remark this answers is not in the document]
samskalicky: Thanks again @leezu! made the code changes based on our offline discussion!

samskalicky:
@guanxinq I made some code refactors based on suggestions from @leezu most of it was just refactoring where params are initialized. But we did find that there is a difference between `params.get(name)` and `params[name]` that was also causing some confusion. 

But the new code should work now without catching any exceptions.

tests/python/unittest/test_subgraph_op.py
samskalicky: i set fullyconnected

tests/python/unittest/test_subgraph_op.py
samskalicky: i changed it so the block is recreated before we partition and weights are copied over, so results can be compared

python/mxnet/gluon/block.py
```diff
@@ -967,6 +968,30 @@ def _build_cache(self, *args):
                 self._cached_op_args.append((False, params[name]))
         flags = [('data_indices', data_indices), ('param_indices', param_indices)] + \
                 self._flags
+
+        args, _ = _flatten(args, "input")
+        args_without_none = [ele for ele in args if ele is not None]
+        try:
+            cargs = [args_without_none[i] if is_arg else i.data()
```
guanxinq: cargs is a local variable and is not used in this function. Does it still make sense to keep cargs here?

python/mxnet/gluon/block.py
```diff
@@ -967,6 +968,27 @@ def _build_cache(self, *args):
                 self._cached_op_args.append((False, params[name]))
         flags = [('data_indices', data_indices), ('param_indices', param_indices)] + \
                 self._flags
+
+        args, _ = _flatten(args, "input")
+        args_without_none = [ele for ele in args if ele is not None]
```
mseth10: unused, can be removed

python/mxnet/gluon/block.py
```diff
@@ -967,6 +968,27 @@ def _build_cache(self, *args):
...
+        try:
+            for is_arg, i in self._cached_op_args:
+                if not is_arg:
+                    tmp = i.data()
```
mseth10: can we remove assignment and just call `i.data()`

python/mxnet/gluon/block.py
guanxinq: Removed.

python/mxnet/gluon/block.py
guanxinq: Done.

mseth10:
@leezu your comments have been addressed. can you please review again?

python/mxnet/gluon/block.py
```diff
@@ -1040,7 +1052,12 @@ def register_child(self, block, name=None):
         super(HybridBlock, self).register_child(block, name)
         self._clear_cached_op()
 
-    def hybridize(self, active=True, **kwargs):
+    def hybridize(self, active=True, backend=None, backend_args=None, **kwargs):
```
eric-haibin-lin: hmmm. This is specific for hybridblock? Can we add documentation?

leezu: Based on prior discussion with @samskalicky, documentation should describe what happens if input shapes change in subsequent forward calls. (Ie. currently no repartitioning is triggered).

python/mxnet/gluon/block.py
rondogency: I think default None is the safest option, anyway we don't need this line

we need to enforce users to pass a dictionary (since user may pass a string), so we need to add a check below before assign it to _backend_args

python/mxnet/gluon/block.py
```diff
@@ -1040,7 +1052,12 @@ def register_child(self, block, name=None):
...
+        self._backend = backend
+        if backend_args is None:
+            self._backend_args = {}
+        else:
+            self._backend_args = backend_args
```
rondogency: we need to enforce users to pass a dictionary (since user may pass a string), so we need to add a check below before assign it to _backend_args

samskalicky: lets add something like
```
if isinstance(backend_args, dict)
```

python/mxnet/gluon/block.py
guanxinq: Added the description for hybridblock hybridize().

eric-haibin-lin: Didn't see it - did you push?

python/mxnet/gluon/block.py
guanxinq: Fixed. Thanks.

python/mxnet/gluon/block.py
guanxinq: Pushed now.

python/mxnet/gluon/block.py
```diff
@@ -1053,11 +1041,31 @@ def register_child(self, block, name=None):
...
+        """Activates or deactivates :py:class:`HybridBlock` s recursively. Has no effect on
+        non-hybrid children.
+
+        Parameters
+        ----------
+        active : bool, default True
+            Whether to turn hybrid on or off.
+        backend : str
+            The name of backend, as registered in `SubgraphBackendRegistry`, default None
+        backend_args : dict of optional arguments, optional
```
mseth10: nit: optional twice

guanxinq: Fixed.

python/mxnet/gluon/block.py
mseth10: we still need an `else` block.
```
else:
    self._backend_args = {}
```

mseth10: Okay, we don't need `else` as it is initialized to {}.

python/mxnet/gluon/block.py
```diff
@@ -656,20 +656,8 @@ def initialize(self, init=initializer.Uniform(), ctx=None, verbose=False,
...
-            Whether to turn hybrid on or off.
-        static_alloc : bool, default False
-            Statically allocate memory to improve speed. Memory usage may increase.
-        static_shape : bool, default False
-            Optimize for invariant input shapes between iterations. Must also
-            set static_alloc to True. Change of input shapes is still allowed
-            but slower.
-        """
+        """ Please refer description of HybridBlock hybridize().
+        """
```
mseth10: can you get rid of trailing whitespaces

python/mxnet/gluon/block.py
```diff
@@ -967,6 +968,26 @@ def _build_cache(self, *args):
...
+                    i._finish_deferred_init()
+
+        if self._backend:
+            ctx = args[0].context
+            # get list of params in the order of out.list_arguments
+            arg_array = [args[data_names[name]] if name in data_names.keys() else params[name].data()
+                         for name in out.list_arguments()]
+            # Partition the graph.
+            out = out.optimize_for(self._backend, arg_array, ctx, **self._backend_args)
+
```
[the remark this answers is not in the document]
mseth10: As discussed offline, it's not required to add a comment here considering currently accepted norms in MXNet. We can discuss this in a wider community to come up with a set of guidelines to be followed while adding comments.

python/mxnet/gluon/block.py
eric-haibin-lin: the concept of SubgraphBackendRegistry, PostPartition, etc are new and not very straightforward to users. Is it possible to also add a link to any tutorial that teaches user how to register a subgraph backend?

mseth10: As discussed offline, we plan to add a tutorial as part of our next PR and link it to the example. I have put together the TODO list for the next PR in this github issue #17532 .

Issue #15975, opened 2019-08-22
Crashing test test_laop6


marcoabreu:
```
test_operator_gpu.test_laop_6 ... 

Error running tests, python exited with status code C0000005
```

http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fwindows-gpu/detail/PR-15972/1/pipeline/

ChaiBapchya:
Faced this issue atleast 5-6 times and had mentioned about it in this issue https://github.com/apache/incubator-mxnet/issues/11720

marcoabreu:
Awesome, thanks! But lets keep the tests separate.

ChaiBapchya:
```
======================================================================
ERROR: test_operator_gpu.test_laop_6
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Python37\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Python37\lib\site-packages\nose\util.py", line 620, in newfunc
    return func(*arg, **kw)
  File "C:\jenkins_slave\workspace\ut-python-gpu\tests\python\gpu\../unittest\common.py", line 177, in test_new
    orig_test(*args, **kwargs)
  File "C:\jenkins_slave\workspace\ut-python-gpu\tests\python\gpu\../unittest\test_operator.py", line 6899, in test_laop_6
    check_grad(test_logabsdet, [a])
  File "C:\jenkins_slave\workspace\ut-python-gpu\tests\python\gpu\../unittest\test_operator.py", line 6869, in <lambda>
    atol=atol_bw, dtype=dtype)
  File "C:\jenkins_slave\workspace\ut-python-gpu\windows_package\python\mxnet\test_utils.py", line 992, in check_numeric_gradient
    symbolic_grads = {k:executor.grad_dict[k].asnumpy() for k in grad_nodes}
  File "C:\jenkins_slave\workspace\ut-python-gpu\windows_package\python\mxnet\test_utils.py", line 992, in <dictcomp>
    symbolic_grads = {k:executor.grad_dict[k].asnumpy() for k in grad_nodes}
  File "C:\jenkins_slave\workspace\ut-python-gpu\windows_package\python\mxnet\ndarray\ndarray.py", line 2406, in asnumpy
    ctypes.c_size_t(data.size)))
[...]

PR #15976, opened 2019-08-22
Disable laop_6

marcoabreu:
Crashes due to #15975


larroy:
What's the root cause of the crash? this test was not crashing not long ago.

marcoabreu:
Next step is to investigate the crash, fix it and re-enable the test.

gyshi:
my ci has an error  with laop_6 , 
https://github.com/apache/incubator-mxnet/pull/15976,
you will see detail messages in my ci tests

Issue #6087, opened 2017-05-03
mx.sym.constant


jmacglashan:
For symbol constructions, there is var, zeros, and ones. Can we also get a constant one that receives a list/numpy/mx.ndarray and makes a constant symbol out of it? Working it in via variables is a bit cumbersome, particularly because the Constant initialization doesn't work for anything not ending with the string 'weight'.

piiswrong:
yeah that's something I wanted to do but didn't have time for.
Basically you serialize an ndarray into string and pass it as argument to an op. The op then decode it and copy it to output everytime.

yuruofeifei:
I'm interested in this. @piiswrong 
Can we just create a const op?

Atomu2014:
my friend has implemented a simple version of constant initializer @panzheyi

panzheyi:
I implemented a possible solution.
A constant can be regarded as a variable which doesn't get gradient.
So you can firstly create a variable, and then block its gradient.
Here's my code:
import mxnet as mx
import numpy as np

@mx.init.register
class MyConstant(mx.init.Initializer):
    """Initializes the weights to a scalar value.

    Parameters
    ----------
    value : float
        Fill value.
    """
    def __init__(self, value):
        super(MyConstant, self).__init__(value=value)
        self.value = value

    def _init_weight(self, _, arr):
        arr[:] = mx.nd.array(self.value)

batch_size = 10

const_arr = np.ones((5, 5)).tolist()
a = mx.sym.Variable('a', shape = (5, 5), init = MyConstant(value = const_arr))
a = mx.sym.BlockGrad(a)
data = mx.sym.Variable('data')
loss = mx.sym.MakeLoss(mx.sym.broadcast_add(a, data))


mod = mx.mod.Module(loss, data_names = ['data'], label_names = [])
mod.bind(data_shapes = [('data', (batch_size, 5, 5)),])
mod.init_params(initializer = mx.init.Uniform())
mod.init_optimizer(optimizer = 'sgd', optimizer_params = (('learning_rate', 0.005),))


a = mx.nd.ones((5, 5))
data = np.ones((1000, 5, 5))

dataiter = mx.io.NDArrayIter(data = {'data': data}, batch_size = batch_size)
dataiter.reset()
for batch_id, databatch in enumerate(dataiter):
	mod.forward_backward(databatch)
	mod.update()
	print mod.get_outputs()[0].asnumpy()

panzheyi:
I implemented a possible solution.
[...]

PR #15993, opened 2019-08-24
Add a contrib operator for Constant

vandanavk:
## Description ##
MXNet doesn't have a constant op for sym API (similar to gluon.constant). The closest is to map it to BlockGrad as discussed in https://github.com/apache/incubator-mxnet/issues/6087. It would be useful to have a constant operator in general, and it is also required to convert Pytorch models to MXNet through ONNX.

Adding support for this operator here.

### Essentials ###

### Changes ###
- Add a constant operator in contrib, add relevant tests
- Add ONNX conversion for contrib.constant, and relevant tests


vandanavk:
@mxnet-label-bot add [Operator, pr-awaiting-review, ONNX]

src/operator/contrib/constant-inl.h
```diff
@@ -0,0 +1,91 @@
...
+#include "../tensor/matrix_op-inl.h"
+
+namespace mxnet {
+namespace op {
+
+struct ConstantParam : public dmlc::Parameter<ConstantParam> {
+  mxnet::Tuple<float> value;
+  int dtype;
+  DMLC_DECLARE_PARAMETER(ConstantParam) {
+    DMLC_DECLARE_FIELD(value)
```
ZhennanQin: Consider to rename `value` to `shape`.

src/operator/contrib/constant.cc
```diff
@@ -0,0 +1,65 @@
...
+NNVM_REGISTER_OP(_contrib_constant)
+.describe(R"code(Creates a constant tensor for a value.
+Example::
+
+  v1 = (1, 2)
+  constant_op = symbol.contrib.constant(value=v1)
+  executor = constant_op.simple_bind(ctx=cpu())
+  executor.forward(is_train=True)
+  executor.outputs
+  [ -1.  2.]
```
ZhennanQin: Here's the most confusing part. I assume v1 is the shape of constant output, but where to specify the output value? On the other words, why the output is [-1, 2]?

ZhennanQin:
I mean, If we did some constant folding optimization in the graph, and generate a new tensor which is not filled with same values. For example,
```
scaled_weight = weight * scale, 
out = convolution(weight = scaled_weight)
```
For this code, that scale multiply will execute many times during each iteration of inference. Because weight and scale are all constant, so we can fold them before inference, and use a constant node to replace the folded result in the graph. The value holding by constant node is vary, which can't be generated with simple pattern.

src/operator/contrib/constant-inl.h
ZhennanQin: The describe is 
```
.describe("The target shape");
```
Let's change this alternately?

anirudhacharya:
can you reference this issue in the PR description - https://github.com/apache/incubator-mxnet/issues/6087

PR #16003, opened 2019-08-26
add uint8 bn mkldnn implementation

ElaineBao:
## Description ##
add uint8 batchnorm, mkldnn implementation and test
@PatricZhao @ZhennanQin 

### Usage ###
Check the doc in https://github.com/apache/incubator-mxnet/tree/master/example/quantization/README.md to quantize models and do inference.
Quantized bn will be used automatically when a bn operator cannot be fused.

### Performance ###
In most cases, bn can be fused, so quantized bn is not introduced. In reset50 v2, some of the bn operators are standalone, quantizing these bn give a performance as follows:

| Model | FP32 (Top-1 / Top-5) | Fusion + fp32 bn | Fusion + int8 bn |
| - | - | - | - |
Resnet50 v2 | 0.764 / 0.935 | 0.722 / 0.901 | 0.712 / 0.897

### DIFF
--- a/example/quantization/imagenet_gen_qsym_mkldnn.py
+++ b/example/quantization/imagenet_gen_qsym_mkldnn.py
@@ -216,7 +216,7 @@ if __name__ == '__main__':
             if exclude_first_conv:
                 excluded_sym_names += ['resnetv10_conv0_fwd']
         elif args.model.find('resnet') != -1 and args.model.find('v2') != -1:
-            excluded_sym_names += ['resnetv20_flatten0_flatten0']
+            excluded_sym_names += ['resnetv20_flatten0_flatten0', 'resnetv20_stage1_batchnorm0_fwd']
             if exclude_first_conv:
                 excluded_sym_names += ['resnetv20_conv0_fwd']
         elif args.model.find('vgg') != -1:
--- a/example/quantization/imagenet_gen_qsym_mkldnn.py
+++ b/example/quantization/imagenet_gen_qsym_mkldnn.py
@@ -216,6 +216,7 @@ if __name__ == '__main__':
             if exclude_first_conv:
                 excluded_sym_names += ['resnetv10_conv0_fwd']
         elif args.model.find('resnet') != -1 and args.model.find('v2') != -1:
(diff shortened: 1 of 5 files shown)


example/quantization/imagenet_gen_qsym_mkldnn.py
```diff
@@ -216,7 +216,7 @@ def save_params(fname, arg_params, aux_params, logger=None):
             if exclude_first_conv:
                 excluded_sym_names += ['resnetv10_conv0_fwd']
         elif args.model.find('resnet') != -1 and args.model.find('v2') != -1:
-            excluded_sym_names += ['resnetv20_flatten0_flatten0']
+            excluded_sym_names += ['resnetv20_flatten0_flatten0', 'resnetv20_stage1_batchnorm0_fwd']
```
xinyu-intel: why exclude the first one?

ElaineBao: This is for the sake of accuracy, if do not exclude this layer, top-1 accuracy will drop to 52.3. Reason of this accuracy drop is under investigation.

tests/python/quantization/test_quantization.py
```diff
@@ -918,15 +916,9 @@ def check_qsym_forward(qsym, qarg_params, qaux_params, data_shape, label_shape=N
...
-            if qdtype == 'int8' and is_test_for_mkldnn() and name in ['sym1', 'sym2', 'sym3']:
-              print('skipped testing test_quantize_model_with_forward for mkldnn cpu int8 since it is not supported yet')
-              continue
-            elif qdtype == 'uint8' and is_test_for_mkldnn() and name in ['sym1']:
-              print('skipping test_quantize_model_with_forward for mkldnn cpu uint8 since it is not supported yet')
-              continue
-            elif qdtype == 'int8' and is_test_for_gpu() and name in ['sym1']:
-              print('skipped testing test_quantize_model_with_forward for gpu int8 since it is not supported yet')
-              continue
+            if is_test_for_gpu() and name in ['sym1']:
```
xinyu-intel: gpu bn will be exlcuded automatically now, so no need to add this.

PR #16006, opened 2019-08-26
[tvm][numpy] operator deg2rad

tingying2020:
Numpy operator deg2rad implemented with tvm.

@haojin2


python/mxnet/ndarray/numpy/_op.py
```diff
@@ -1905,3 +1905,40 @@ def get_list(arrays):
 
     arrays = get_list(arrays)
     return _npi.stack(*arrays, axis=axis, out=out)
+
+
+
```
haojin2: one less blank line

tingying2020: Done, thanks.

PR #16013, opened 2019-08-27
Windows cmake flags cleanup

larroy:
## Description ##

See tittle. Doesn't change any behaviour.
Flags were set to 0 and 1 instead of ON OFF
Removed USE_PROFILER
#15595


yuxihu:
The changes look straightforward. Two questisons:
1. ON/OFF is the preferred/recommended way of setting cmake flags? 0/1 also worked, right?
2. Why removing USE_PROFILER? Is it ON in other platforms? I think we have profiler related tests. Would it be affected?

larroy:
USE_PROFILER was deprecated by @ChaiBapchya ON OFF is preferred as it was requested to have multi value flag for the openmp flag in a different PR.

PR #16061, opened 2019-09-01
Allow loading from model files with empty weights.

JiadingGai:
This patch extends mxnet.model.load_checkpoint to support loading from saved graphs with empty weights. Weightless graphs can arise from several scenarios; e.g., graphs that only contain mxnet.ndarray ops such as repeat, reshape etc. The feature is quite useful in testing each individual component of a large neural network. Many thanks to Joshua Z. Zhang for suggesting the fix.

### DIFF
--- a/python/mxnet/model.py
+++ b/python/mxnet/model.py
@@ -431,6 +431,7 @@ def load_params(prefix, epoch):
     aux_params = {}
     if not save_dict:
         logging.warning("Params file '%s' is empty", '%s-%04d.params' % (prefix, epoch))
+        return (arg_params, aux_params)
     for k, v in save_dict.items():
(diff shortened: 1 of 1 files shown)


python/mxnet/model.py
```diff
@@ -431,6 +431,7 @@ def load_params(prefix, epoch):
     aux_params = {}
     if not save_dict:
         logging.warning("Params file '%s' is empty", '%s-%04d.params' % (prefix, epoch))
+        return (arg_params, aux_params)
```
wkcn: Thanks for your contribution!
In the original code, the function will return empty weights in the line 440 when loading an empty file.

wkcn:
Merged. Thank you for the fix : )

PR #16066, opened 2019-09-02
Getting rid of excessive repetitions in test_pick

gyshi:
before remove for loop, 
test time is too long  cpu:130.647s . gpu:188.873s
after remove for loop, cpu:0.229s .  gpu:14.004s
because i think  for loop is redundancy.


gyshi:
@haojin2 , thx for reviewing

PR #16067, opened 2019-09-02
Reduce time cost of test_operator.test_norm, test_operator. test_laop_5 and test_sparse_operator.test_elemwise_binary_ops

gyshi:
because test_norm and  test_elemwise_binary_ops use  rand_shape_2d()。 by default, rand_shape_2d generate shape(0,0) to (10, 10), so it will generate too large shape. for some operator, it will consume much time. so i  limit it .

for test_laop_5, i modify for loop from (1, 10) to (1, 5), beacuse i it's ok!  so the test time from 71.902s to 16.569s.

### DIFF
--- a/tests/python/unittest/test_operator.py
+++ b/tests/python/unittest/test_operator.py
@@ -3704,7 +3704,7 @@ def test_norm():
 
     ctx = default_context()
     data = mx.symbol.Variable('data')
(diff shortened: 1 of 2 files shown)


gyshi:
@haojin2 , thx for reviewing

PR #16078, opened 2019-09-02
[Dataset] Add take, filter API to dataset

eric-haibin-lin:
## Description ##
Add dataset.filter, dataset.take APIs. @zhreshold 

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/python/mxnet/gluon/data/dataset.py
+++ b/python/mxnet/gluon/data/dataset.py
@@ -101,7 +101,7 @@ class Dataset(object):
         from . import Sampler
         if not isinstance(sampler, Sampler):
             raise TypeError('Invalid sampler type: %s. Expected gluon.data.Sampler instead.'%
-                             type(sampler))
+                            type(sampler))
         return _SampledDataset(self, sampler)
 
     def transform(self, fn, lazy=True):
--- a/python/mxnet/gluon/data/dataset.py
+++ b/python/mxnet/gluon/data/dataset.py
@@ -175,8 +175,8 @@ class _FilteredDataset(Dataset):
     def __init__(self, dataset, fn):
         self._dataset = dataset
         self._indices = []
-        for i in range(len(dataset)):
-            if fn(dataset[i]):
+        for i, sample in enumerate(dataset):
+            if fn(sample):
                 self._indices.append(i)
 
     def __len__(self):
--- a/python/mxnet/gluon/data/dataset.py
+++ b/python/mxnet/gluon/data/dataset.py
@@ -61,7 +61,8 @@ class Dataset(object):
         Dataset
             The filtered dataset.
         """
-        return _FilteredDataset(self, fn)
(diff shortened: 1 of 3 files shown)


eric-haibin-lin:
Then the Sampler needs to take a dataset as the input for its constructor, which is quite different from existing samplers. Do we really want to do that?
We also need more functionalities like concat, reduce and window (ref: https://www.tensorflow.org/api_docs/python/tf/data/Dataset#concatenate) which will be separate APIs and cannot be expressed as sampling.

zhreshold:
this one looks good to me, just wanna confirm if it's compatible with LazyTransform dataset?

eric-haibin-lin:
yes. I added more tests with lazy transform dataset.

python/mxnet/gluon/data/dataset.py
```diff
@@ -135,6 +170,38 @@ def __call__(self, x, *args):
             return (self._fn(x),) + args
         return self._fn(x)
 
+class _FilteredDataset(Dataset):
+    """Dataset with a filter applied"""
+    def __init__(self, dataset, fn):
+        self._dataset = dataset
+        self._indices = []
```
leezu: `self._indices = [i for for i, sample in enumerate(dataset) if fn(sample)]`
should be faster for large datasets.

python/mxnet/gluon/data/dataset.py
```diff
@@ -40,6 +40,41 @@ def __getitem__(self, idx):
     def __len__(self):
         raise NotImplementedError
 
+    def filter(self, fn):
```
leezu: This will eagerly retrieve all values from a `lazy=True` transformed dataset. That should be documented?
(The transformation function will also be applied again in `__getitem__`)

eric-haibin-lin: Yes I'll document that

python/mxnet/gluon/data/dataset.py
eric-haibin-lin: Good point I'll update it

python/mxnet/gluon/data/dataset.py
zhreshold: Is it better in some cases that lazy evaluation should be applied after `filter`, i.e., 

data.filter(xxx).transform is way better than data.transform.filter in some situations.

eric-haibin-lin: Both are valid operations and I don't think we can stop users from doing transform() before filter(), since sometimes we need to actually filter the data based on the transformed data. For now I'd resort to better documentation

python/mxnet/gluon/data/dataset.py
leezu: Sorry there was a typo in above code sample. You need to remove a `for`

PR #16097, opened 2019-09-05
[numpy] array ufunc and array function protocols

reminisce:
## Description ##
This PR implemented NumPy's [array ufunc](https://numpy.org/neps/nep-0013-ufunc-overrides.html) and [array function](https://numpy.org/neps/nep-0018-array-function-protocol.html) protocols so that `mxnet.numpy.ndarray` can be fed into a subset of official NumPy operators and get dispatched to the corresponding MXNet operators for execution. See the following for examples.

```python
import os

import numpy as onp  # the official NumPy's version must be >= 1.17
from mxnet import np as mx_np

a = mx_np.random.uniform(size=(1, 3))
b = mx_np.random.uniform(size=(2, 1))
assert type(onp.concatenate([a, a])) == mx_np.ndarray
assert type(onp.sum(a)) == mx_np.ndarray
```

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/python/mxnet/numpy_dispatch_protocol.py
+++ b/python/mxnet/numpy_dispatch_protocol.py
@@ -1,0 +1,214 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""Utils for registering NumPy array function protocol for mxnet.numpy ops."""
+
+from __future__ import absolute_import
+import functools
+import numpy as _np
+from . import numpy as mx_np  # pylint: disable=reimported
+from .numpy.multiarray import _NUMPY_ARRAY_FUNCTION_DICT, _NUMPY_ARRAY_UFUNC_DICT
+
+
+def _find_duplicate(strs):
+    str_set = set()
+    for s in strs:
+        if s in str_set:
+            return s
+        else:
+            str_set.add(s)
+    return None
+
+
+def _implements(numpy_function):
+    """Register an __array_function__ implementation for MyArray objects."""
+    def decorator(func):
+        _NUMPY_ARRAY_FUNCTION_DICT[numpy_function] = func
+        return func
+    return decorator
+
+
+def with_array_function_protocol(func):
+    """A decorator for functions that expect array function protocol.
+    The decorated function only runs when NumPy version >= 1.17."""
+    from distutils.version import LooseVersion
+    cur_np_ver = LooseVersion(_np.__version__)
+    np_1_17_ver = LooseVersion('1.17')
+
+    @functools.wraps(func)
+    def _run_with_array_func_proto(*args, **kwargs):
+        if cur_np_ver >= np_1_17_ver:
+            try:
+                func(*args, **kwargs)
+            except:
+                raise RuntimeError('Running function {} with NumPy array function protocol failed'
+                                   .format(func.__name__))
+
+    return _run_with_array_func_proto
+
+
+def with_array_ufunc_protocol(func):
+    """A decorator for functions that expect array ufunc protocol.
+    The decorated function only runs when NumPy version >= 1.15."""
+    from distutils.version import LooseVersion
+    cur_np_ver = LooseVersion(_np.__version__)
+    np_1_15_ver = LooseVersion('1.15')
+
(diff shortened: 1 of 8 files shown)


python/mxnet/array_function_protocol.py
```diff
@@ -0,0 +1,98 @@
...
+
+def _implements(numpy_function):
+    """Register an __array_function__ implementation for MyArray objects."""
+    def decorator(func):
+        _NUMPY_ARRAY_FUNCTION_DICT[numpy_function] = func
+        return func
+    return decorator
+
+
+_NUMPY_ARRAY_FUNCTION_LIST = [
```
szha: is this the list of operators where our operator definition complies with the official numpy behavior?

reminisce: Yes, it is the operators in our codebase that is dispatchable through the array function protocol. Others will be dispatched through either fluent methods or array ufunc protocol, which will be implemented later.

szha: Awesome. We can use this list to track the full-compatibility with numpy operators once we start measuring it.

reminisce:
@szha This PR is ready for merge if there are no more comments. Thanks.

szha:
LGTM. @leezu would you mind taking a look too?

python/mxnet/numpy_dispatch_protocol.py
```diff
@@ -0,0 +1,214 @@
...
+    from distutils.version import LooseVersion
+    cur_np_ver = LooseVersion(_np.__version__)
+    np_1_17_ver = LooseVersion('1.17')
+
+    @functools.wraps(func)
+    def _run_with_array_func_proto(*args, **kwargs):
+        if cur_np_ver >= np_1_17_ver:
+            try:
+                func(*args, **kwargs)
+            except:
```
leezu: Why is it necessary to catch all possible exceptions here and to overwrite their error message with a generic 'Running ... failed'?

reminisce:
@leezu Thanks for the review. I tried the commands you showed, but it works for me. Could you confirm that you are using NumPy with version >= 1.17?
```python
>>> from mxnet import npx, np
>>> a = np.arange(10)
>>> npx.set_np()
>>> a = np.arange(10)
>>> import numpy as onp
>>> print(onp.__version__)
1.17.1
>>> onp.dot(a, a)
array(285.)
>>> type(onp.dot(a, a))
<class 'mxnet.numpy.ndarray'>
```

python/mxnet/numpy_dispatch_protocol.py
reminisce: Good point. I will append the generic message to the real exception.

reminisce:
@leezu Re: Can test be improved? Yes, absolutely. This PR is for building the skeleton of testing suite. It enables `mxnet.numpy` operator developers to easily add test cases from the official NumPy package to MXNet (`OpArgMngr` in `test_numpy_interoperatibility.py` is for storing those test cases. We will send in those test cases in the follow-up PRs.

leezu:
> @leezu Could you confirm that you are using NumPy with version >= 1.17?

You're right, I was using numpy 1.16 and it ran into the well-known issue (I guess there is some issue tracking it, but not sure) that numpy doesn't correctly convert ndarrays to numpy arrays.

Would it make sense to support Numpy 1.16 when `NUMPY_EXPERIMENTAL_ARRAY_FUNCTION=1` environment variables is defined? If not, how about bumping the required minimum numpy version in `setup.py` to 1.17?

https://github.com/apache/incubator-mxnet/blob/e87995db6f8ac7e8d4424001e1db5296398562f6/python/setup.py#L33

reminisce:
@leezu I prefer to bump the required version to 1.17 to use this feature for two reasons. However, I'm not sure if we should bump the numpy version requirement in this or a separate PR. Maybe @szha can provide some insights.

1. This functionality was claimed as experimental in 1.16. There might be bugs that we are not aware of, and consequently would have to deal with unexpected behaviors. I tried using numpy 1.16 during the development. However, there is some issue with NumPy correctly dispatching calls to MXNet even when the environment variable is set (I printed the environ value and verified it's 1) using nosetests. In 1.17, there is no such problem.

2. Asking users to set the environment variable manually to enable this functionality may open the door to unknown issues, especially under multi-threaded circumstances (the one I saw may be related to this). To close the possibilities of using it incorrectly and make the user experience straightforward without surprises, we can stick with numpy versions >= 1.17.

szha:
The downside of upgrading is the potential conflict of the numpy dependency version between mxnet’s and other packages’. Shall we just produce proper error message for older numpy versions instead?

reminisce:
I will add proper error message when it is used with numpy 1.16 versions. Thanks.

szha:
@leezu I'm referring to the situation where another package on an end user's environment declares a dependency on numpy 1.16, in which case the control of upgrading numpy isn't in the hands of the end user.

leezu:
This seems quite hypothetical / uncommon to me. I'm not convinced if we need to take consideration of this, given the user experience impact of not declaring the correct numpy version dependency and our current strategy of throwing runtime errors or having numpy throw errors such as `ValueError: dot: too many dimensions in result` (which is not very helpful for users).
I don't want to block this PR though.

reminisce:
@leezu @szha How about we merge this PR first and later consider if it's necessary to bump the numpy version to 1.1?

szha:
I'm OK with bumping the version and merging the PR. The error message during installation is sufficiently clear.

PR #16115, opened 2019-09-07
[Numpy] . implement numpy op sinc with tvm

gyshi:
https://docs.scipy.org/doc/numpy/reference/generated/numpy.sinc.html

this op is different with other op, if your data type is "int" type, your output  data type is float64, but  it has no backward.


gyshi:
@mxnet-label-bot  numpy

gyshi:
@yzhliu @reminisce

PR #16120, opened 2019-09-08
Mkldnn v1.0

dchoi77:
## Description ##
An error in Doc.

### Essentials ###

### Changes ###
- The Doc of the function nn.Conv2DTranspose does not explain the input parameter 'padding' correctly. It is simply a copy of the part used in nn.Conv2D. You need to correct the meaning of padding. The same thing is applied to nn.Conv1DTranspose and nn.Conv3DTranspose.

## Comments ##


TaoLv:
Sorry the feature branch is not ready for upstreaming yet. Closing to save the CI resources. Thanks.

PR #16122, opened 2019-09-08
Add fast implementation of LARS

Caenorst:
## Description ##
Add a Fast implementation of LARS.

### Essentials ###

### Changes ###
- Add multi_sum_sq Op: computing sum of squares of multiple arrays (1 sum per array).
- Add multi_lars Op: computing LARS with the sum of squares of weights and gradients.
- Add preloaded_multi_sgd which a version of SGD where learning rate are MXNet array instead of list of parameters.
- Add SGDwFastLARS optimizer
- Add test for the new Ops and optimizer.

## Comments ##
This code have been used for MLPerf v0.6 benchmarks. It's especially pertinent if you are training with small local batch size as optimizers usually don't scale with batch size.

## Credits ##
- The CPU part is from @drivanov
- As said in the comments of multi_sum_sq files, the main approach is coming from https://github.com/NVIDIA/apex


python/mxnet/optimizer/optimizer.py
```diff
@@ -781,6 +784,240 @@ def update(self, index, weight, grad, state):
         ftml_update(weight, grad, prev_d, prev_v, prev_z, out=weight,
                     lr=lr, wd=wd, **kwargs)
 
+@register
+class SGDwFastLARS(Optimizer):
```
anirudhacharya: is this different from LBSGD defined here - https://github.com/apache/incubator-mxnet/pull/16122/files#diff-0c893416e9e93fbd94dfaa9fa6c13d67R1022

Currently LBSGD implementation is buggy, maybe this should replace that rather than create SGDwFastLARS?

Caenorst: Thank you for quick review.

LB stand for "Large batch" indicating that it will use diverse techniques. As warmup could be implemented directly with lr_scheduler (LARS contrarily to what LBSGD indicate is not a warmup strategy but an optimizer). I think the only reason to use such name would be to have the choice between LARS and LARC optimizers (which will be coming in another PR), or even other techniques for large batch. 

I think it could make sense as both implementation are very similar although I'm not sure it's ergonomic to have the selection of optimizer as an argument of `Optimizer`.

anirudhacharya: yes, i am not suggesting we have a parameter that will switch the optimizer from LARS to LARC. Maybe we could just deprecate( or mark it for deprecation) LBSGD, since it does not work.

python/mxnet/optimizer/optimizer.py
```diff
@@ -781,6 +784,240 @@ def update(self, index, weight, grad, state):
...
+    def __init__(self, momentum=0.0, lazy_update=True, lars_eta=0.001, lars_eps=0,
```
anirudhacharya: could you add documentation like here - https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/optimizer/optimizer.py#L727

python/mxnet/optimizer/optimizer.py
Caenorst: I think deprecation is the best option, please reviewers thumb up / down this comment for opinion

python/mxnet/optimizer/optimizer.py
Caenorst: Definitely

python/mxnet/optimizer/optimizer.py
ptrendx: I am in favorof deprecation of `LBSGD`, but still I would prefer a better name - "Fast" in "SGDwFastLARS" does not really make sense (fast compared to what?). Maybe just `LARSSGD` or something like that?

Caenorst: We could also just call them LARC or LARS.

As a point of reference, TF calls it LARSOptimizer:
https://www.tensorflow.org/api_docs/python/tf/contrib/opt/LARSOptimizer

Caenorst:
`SyntaxError: only named arguments may follow *expression`

I'm not seeing this error on my local machine, what version of Python is on CI ?

python/mxnet/optimizer/optimizer.py
Caenorst: Is this good enough ?

python/mxnet/optimizer/optimizer.py
Caenorst: I renamed it LARS, and also remove the "lars" redundant prefix from `eta` and `eps` let me know if it's fine like that

tests/python/gpu/test_optimizer.py
```diff
@@ -0,0 +1,93 @@
...
+    tanh3 = mx.sym.Activation(data=fc1, act_type="tanh")
+    # second fullc
+    fc2 = mx.sym.FullyConnected(data=tanh3, num_hidden=10)
+    # loss
+    lenet = mx.sym.SoftmaxOutput(data=fc2, name='softmax')
+    #pylint: enable=no-member
+    return lenet
+
+@with_seed()
+def test_lars():
```
anirudhacharya: how long does this test take? should it be put in nightly tests instead of unit tests? 

should we a test in `python/unittest/test_optimizer.py`?

Caenorst: It's taking 10s with my V100

Caenorst: So, I'm already testing the MXNet Ops in https://github.com/apache/incubator-mxnet/pull/16122/files#diff-4758fb9329d438de2836db2634a8f5f7R270-R422. which are the only non-python part of the optimizer, what my test is adding is to show that it's properly behaving: i.e allow you to train a network with a bigger batch. Making a fully python of it wouldn't help me testing that (the python could be wrong as well, let's say if I misunderstood the publication)

Caenorst:
I don't understand the errors in the CI, can somebody help me ?

tests/python/gpu/test_optimizer.py
Caenorst: @anirudhacharya Does swapping my training test to nightly test and just unit testing my MXNet Ops in `/tests/python/gpu/test_operator_gpu.py` enough ? If yes I believe I'm all done for this PR :)

anirudhacharya: sorry, missed seeing this. i will take another look. And, yes it would be good to move time consuming tests to nightly tests. 

I understand that the python mock optimizer could be wrong if our understanding of the paper is incorrect. Let us try to catch them in the code review process.

anirudhacharya: apart from the above mentioned test issue, everything else lgtm.

python/mxnet/optimizer/optimizer.py
```diff
@@ -781,6 +784,266 @@ def update(self, index, weight, grad, state):
...
+            stype = weight.stype if self.lazy_update else 'default'
+            momentum = zeros(weight.shape, weight.context, dtype=weight.dtype, stype=stype)
+        return momentum
+
+    def _l2norm(self, v, rescale=False):
+        """L2 Norm implementation"""
+        v = v.astype('float32')
+        if rescale:
+            v *= self.rescale_grad
+        norm = NDnorm(v).asnumpy()[0]
```
eric-haibin-lin: Is this intended? I thought having a blockingc call is bad

PR #16131, opened 2019-09-09
Fix for duplicate subgraph inputs/outputs

### LINKED ISSUE #16108: Subgraph API creates duplicate inputs and outputs
## Description
Subgraph API can create duplicated tensors for both inputs to a subgraph and outputs from a subgraph. This happens when there are multiple nodes in a subgraph that consume the same input, or when there are multiple nodes that consume a single subgraph output. This results in tensor duplication, and causes OOM due to excessive memory usage. 

Consider the duplicate input problem first. Lets say we have the following graph:
```
A --> B
  \
    --> C
```
In this graph node A has a single output, and nodes B and C consume A's output. If nodes B and C are added to a subgraph, the graph is changed to the following:
```
A --> S

S [ A0 --> B
     A1 --> C]
```
Heres, nodes B and C are moved into the subgraph, and their dependency on node A (which is not in the subgraph) is broken, and an input node is created for each: A0, A1. The subgraph will have 2 inputs (A0, A1) and will both have dependencies to node A. 

Now lets consider the the duplicate output problem. Consider the same graph as above:
```
A --> B
  \
    --> C
```
But this time lets only node A is added to the subgraph, and nodes B and C are left out. Then  the graph is changed to the following:
```
S --> B
  \
    --> C

S [ A]
```
Here, B and C have dependencies on the output from A inside the subgraph. In this case, the subgraph will have 2 outputs, one for each consumer: B and C. But both outputs will be defined as coming from the same output of node A inside the subgraph.

samskalicky:
## Description ##
Fixes #16108

Subgraph API can create duplicated tensors for both inputs to a subgraph and outputs from a subgraph. This happens when there are multiple nodes in a subgraph that consume the same input, or when there are multiple nodes that consume a single subgraph output. This results in tensor duplication, and causes OOM due to excessive memory usage.

### Fix for duplicate inputs ###
Currently, inputs to a subgraph are duplicated when there are multiple nodes in a subgraph consuming the same input (see diagram below). In this PR we change the behavior of CutGraphInputs function to store a NodeEntry for each input to the subgraph in a map, and re-use that node for each node in the subgraph that consumes that input (see diagram below). This prevents input nodes from being duplicated, and results in less copies of tensors at runtime and a much smaller memory footprint.

duplicateInputs

### Fix for duplicate outputs ###
Similar to the inputs, when an output from a subgraph is consumed by multiple nodes outside the subgraph, the output is duplicated for each consumer (see diagram below). In this PR we change the behavior of CreateSubgraphNode to collapse duplicate output dependencies to a single output by checking if neighboring output entries are the same. If so we do not add another output, preventing duplicates. 

Then in the ConnectSubgraphOutputs function we create dependencies from each node outside the subgraph that consumes a subgraph output. If neighboring nodes share the same subgraph output, they point to the same subgraph output.

duplicateOutputs

### Example ###
Added an example graph with duplicate inputs/outputs to the `test_subgraph_op.py`. Heres the original graph partitioned before this PR. Notice that it has two input nodes: `data0` and `data1` that both come from the same top level `data`. Similarly, notice that it has two outputs coming from the same node `_plus0`.
```
{
  "nodes": [
    {
      "op": "null", 
      "name": "data", 
      "attrs": {"__shape__": "(1,)"}, 
      "inputs": []
    }, 
    {
      "op": "_CachedOp", 
      "name": "_CachedOp0", 
      "inputs": [[0, 0, 0], [0, 0, 0]], 
      "subgraphs": [
        {
          "nodes": [
            {
              "op": "null", 
              "name": "data0", 
              "inputs": []
            }, 
            {
              "op": "sin", 
              "name": "sin0", 
              "inputs": [[0, 0, 0]]
            }, 
            {
              "op": "null", 
              "name": "data1", 
              "inputs": []
            }, 
            {
              "op": "sin", 
              "name": "sin1", 
              "inputs": [[2, 0, 0]]
            }, 
            {
              "op": "elemwise_add", 
              "name": "_plus0", 
              "inputs": [[1, 0, 0], [3, 0, 0]]
            }
          ], 
          "arg_nodes": [0, 2], 
          "node_row_ptr": [0, 1, 2, 3, 4, 5], 
          "heads": [[4, 0, 0], [4, 0, 0]]
        }
      ]
    }, 
    {
      "op": "cos", 
      "name": "cos0", 
      "inputs": [[1, 0, 0]]
    }, 
    {
      "op": "cos", 
      "name": "cos1", 
      "inputs": [[1, 1, 0]]
    }, 
    {
      "op": "elemwise_sub", 
      "name": "_minus0", 
      "inputs": [[2, 0, 0], [3, 0, 0]]
    }
  ], 
  "arg_nodes": [0], 
  "node_row_ptr": [0, 1, 3, 4, 5, 6], 
  "heads": [[4, 0, 0]], 
  "attrs": {"mxnet_version": ["int", 10700]}
}
```
Heres the same graph partitioned after this PR. Notice that now there is only 1 input node `data0` and both `sin` ops use it. Also notice that now there is only 1 output from the subgraph, and both `cos` ops use it. 
```
{
  "nodes": [
    {
      "op": "null", 
      "name": "data", 
      "attrs": {
        "__profiler_scope__": "<unk>:", 
        "__shape__": "(1,)"
      }, 
      "inputs": []
    }, 
    {
      "op": "_CachedOp", 
      "name": "_CachedOp0", 
      "inputs": [[0, 0, 0]], 
      "subgraphs": [
        {
          "nodes": [
            {
              "op": "null", 
              "name": "data0", 
              "inputs": []
            }, 
            {
              "op": "sin", 
              "name": "sin0", 
              "attrs": {"__profiler_scope__": "<unk>:"}, 
              "inputs": [[0, 0, 0]]
            }, 
            {
              "op": "sin", 
              "name": "sin1", 
              "attrs": {"__profiler_scope__": "<unk>:"}, 
              "inputs": [[0, 0, 0]]
            }, 
            {
              "op": "elemwise_add", 
              "name": "_plus0", 
              "attrs": {"__profiler_scope__": "<unk>:"}, 
              "inputs": [[1, 0, 0], [2, 0, 0]]
            }
          ], 
          "arg_nodes": [0], 
          "node_row_ptr": [0, 1, 2, 3, 4], 
          "heads": [[3, 0, 0]]
        }
      ]
    }, 
    {
      "op": "cos", 
      "name": "cos0", 
      "attrs": {"__profiler_scope__": "<unk>:"}, 
      "inputs": [[1, 0, 0]]
    }, 
    {
      "op": "cos", 
      "name": "cos1", 
      "attrs": {"__profiler_scope__": "<unk>:"}, 
      "inputs": [[1, 0, 0]]
    }, 
    {
      "op": "elemwise_sub", 
      "name": "_minus0", 
      "attrs": {"__profiler_scope__": "<unk>:"}, 
      "inputs": [[2, 0, 0], [3, 0, 0]]
    }
  ], 
  "arg_nodes": [0], 
  "node_row_ptr": [0, 1, 2, 3, 4, 5], 
  "heads": [[4, 0, 0]], 
  "attrs": {"mxnet_version": ["int", 20000]}
}
```


samskalicky:
@zheng-da @reminisce could you please review?

samskalicky:

samskalicky:
@ZhennanQin for review. Any suggestions on how to better fix the problem?

src/operator/subgraph/subgraph_property.h
```diff
@@ -296,8 +296,20 @@ class SubgraphProperty {
    */
   virtual void ConnectSubgraphOutputs(const nnvm::NodePtr subgraph_node,
                                       std::vector<nnvm::NodeEntry*>* output_entries) const {
+    // Collapse output_entries pointing to same NodeEntry
```
ZhennanQin: Can we implement this outside of ConnectSubgraphOutputs? In other words, can we ensure that `output_entries` in `ConnectSubgraphOutputs` doesn't have duplicated entries, and we can create mapping in build_subgraph to achieve that.

samskalicky: Any suggestion on how to do this without creating a map or changing the interface/API of createSubgraphOutputs?

HahTK: It is important to be clear about 2 things :
1. "output_entries" are misnamed. They are really input_entries_consuming_subgraph_output.
2. Therefore, output_entries do not have duplicate entries. The number of output_entries is correct ! (or rather the the number of input_entries_consuming_subgraph_output is correct)

Consider the following 3 node json example -
(json simplified to eliminate irrelevant fields)

```
nodes : [ 
  { name : "NodeA" }, #nodes[0]
  { name : "NodeB",   #nodes[1]
    inputs : [[0,0,0]] }, #pointing at NodeA 
  { name : "NodeC",  #nodes[2]
    inputs : [[0,0,0]] } #pointing at NodeA 
]
```

Suppose NodeA gets turned into a subgraph with 1 node. If we do that then, 
The so called "output_entries" are :
1. nodes[1][inputs][0] #0th input of NodeB
2. nodes[2][inputs[0] #0th input of NodeC

Therefore the variable names "output_entries" is actually refering to the the "input_entries_consuming_subgraph_output". 
The correct number of "output_entries" is therefore 2. 
We can not collapse this down to 1 because both NodeB and NodeC need to point back to NodeA

The real outputs are "sym.outputs" in the build_graph.cc code in the PR. 
We do de-duplicate those.

ZhennanQin: Of course each entries have their unique address, which means, they are presenting different data consuming for this node's output. But, if some entries are the same(same node, index and version), then they should still be the 'same' after subgraph partition(same means the entry value is the same among those entries, but of course the entry value is different from before partition).
Take your example as example,
After partition, the result would be
```
nodes : [ 
  { name : "SubgraphA" }, #subgraph_nodes[0], have 2 outputs, the output[1] presents the origin output of  NodeA
  { name : "NodeB",   #nodes[1]
    inputs : [[0,1,0]] }, #pointing at SubgraphA, repsents the output data require of NodeA 
  { name : "NodeC",  #nodes[2]
    inputs : [[0,1,0]] } #pointing at SubgraphA, repsents the output data require of NodeA 
]
```
So, here's what we can do to simplify the work in `ConnectSubgraphOutputs` 

1. Create a map about unique_entry -> origin entry address(Not necessary, just for better performance).
2. Pass unique_entry to `ConnectSubgraphOutputs`, and return a map of  unique_entry -> unique_entry_after_partition.
3. Then we know, how each original entry should be replaced for. Then do the replacement on all entry address based on above 2 maps.

What I what to do here is, we should avoid backend developer to handle the duplicated entries, instead, backend should only tells the new connection after partition. Take above case as example, backend developer should only tells that the output data requirement on NodeA now should pointing to index 1 on new subgraph node through `ConnectSubgraphOutputs`. And build_subgraph pass should propagate this change on all original entries which points at NodeA.

samskalicky: @zhennanqin are you suggesting we change the API for connect subgraph outputs? That would be a breaking change with lots of implications. 

Are you ok with doing this in 1.6 timeframe? Maybe we should take @hahtk suggestion for now and change the API for the 2.0 release?

ZhennanQin: Don't need to change API. The pseudo code is,
```
std::vector<nnvm::NodeEntry*> unique_output_entires; // Collect the unique entry, let's collect the first entry
std::unordered_map<nnvm::NodeEntry*, std::vector<nnvm::NodeEntry*>> output_entires_map; // Collect all duplicated
// Collect above infomation
for (auto entry_ptr : output_entries) {
  if (*entry_ptr is unique) { // replace this with nnvm::NodeEntryEqual logic
    unique_output_entires.push_back(entry_ptr);
  } else {
    output_entires_map[first_entry_of_it].push_back(entry_ptr);
  }
}
// Pass unique_output_entires to ConnectSubgraphOutputs and collect the change
subg_prop->ConnectSubgraphOutputs(n, &unique_output_entires);
// Now we know how to make the change to all duplicated
for (auto entry_ptr : unique_output_entires) {
  auto duplicated_entries = output_entires_map[entry_ptr];
  for (auto& i : duplicated_entries) {
    *i = *entry_ptr;
  }
}
```

ZhennanQin: You're right. But changing inside ConnectSubgraphOutputs() won't help for the backends that override this function, and because of duplicated output entries, most of backends had already overrided this function. Your proposal doesn't work for them. Even for newly added backends, they will most likely continue overriding ConnectSubgraphOutputs() because the outputs connection may have a different sequence. 

To make things more clear as you requested, I suggest to introduce an new api, maybe called `ConnectSubgraphUniqueOutputs`, then the code would be like,
```
virtual void ConnectSubgraphUniqueOutputs(const nnvm::NodePtr subgraph_node,
                                          std::vector<nnvm::NodeEntry*>* output_entries) const {
  size_t idx = 0;
  for (size_t i = 0; i < output_entries->size(); ++i) {
    *output_entries->at(i) = nnvm::NodeEntry{subgraph_node, idx++, 0};
}

virtual void ConnectSubgraphOutputs(const nnvm::NodePtr subgraph_node,
                                      std::vector<nnvm::NodeEntry*>* output_entries) const {
std::vector<nnvm::NodeEntry*> unique_output_entires; // Collect the unique entry, let's collect the first entry
std::unordered_map<nnvm::NodeEntry*, std::vector<nnvm::NodeEntry*>> output_entires_map; // Collect all duplicated
// Collect above infomation
for (auto entry_ptr : output_entries) {
  if (*entry_ptr is unique) { // replace this with nnvm::NodeEntryEqual logic
    unique_output_entires.push_back(entry_ptr);
  } else {
    output_entires_map[first_entry_ptr_of_it].push_back(entry_ptr);
  }
}
// Pass unique_output_entires to ConnectSubgraphUniqueOutputs and collect the change
ConnectSubgraphUniqueOutputs(n, &unique_output_entires);
// Now we know how to make the change to all duplicated
for (auto entry_ptr : unique_output_entires) {
  auto duplicated_entries = output_entires_map[entry_ptr];
  for (auto& i : duplicated_entries) {
    *i = *entry_ptr;
  }
}
```
Then backend developer can only override ConnectSubgraphUniqueOutputs() to avoid handle duplicated outputs.

samskalicky: Hey Guys,

I think we're getting offtrack here. Lets just clarify what we're working toward. We want the subgraph to only have unique outputs, so this code is necessary in creating the sym.outputs correctly:

https://github.com/apache/incubator-mxnet/pull/16131/files#diff-962e4e83b570943c087e477058485e25R578-R587

Then, the problem is we need to wire up the those unique outputs to each entry in output_entries. Since we pass the new symbol "sym" into connectSubgraphOutputs indirectly via the node "n", technically the set of unique outputs from the subgraph is already available in the connectSubgraphOutputs function. 

To clarify, the ideal solution would be to pass in a map from output node to output entry and then let connectSubgraphOutputs just wire them up as it sees fit. 

Do we agree that we do not want to change the API of connectSubgraphOutputs to enable this? Is there any reason not to just do this now? it shouldnt be too much work to go change, there is only a few subgraph properties that would need to change:

https://github.com/apache/incubator-mxnet/search?q=connectSubgraphOutputs&unscoped_q=connectSubgraphOutputs

And I think the creators of those wouldnt mind just fixing this once and for all now. This would go in the 1.6 release.

ZhennanQin: @samskalicky Before this PR, backend still has the chance to remove the duplicated sym.outputs in CreateSubgraphNode(). The existing APIs are enough to solve the problem you're trying to fix, so I think this PR is to make all of this happen in a automatic way. But the change in connectSubgraphOutputs() doesn't achieve that goal, but making the solution semi-automatic. 

From a backend developer position, I rather prefer to have full control of my code(current APIs works in this way), or making most of the code automatically(the way I suggest to change). For the semi-automatic solution, developer have to spend extra effort to figure out which parts are handled automatically(sym.outputs), and which parts are not(connectSubgraphOutputs). And also increased the change that backend developer to make a bug when overriding connectSubgraphOutputs(I can't find any reason for connecting entries with same value to different sym.outputs, but by mistake).

Anyway, I don't against to make the change in your way. At least maybe we should add some comments in connectSubgraphOutputs to highlight the difference between sym.outputs and output_entries, and let developer notice that they need to handle the duplicated entires carefully?

samskalicky: @ZhennanQin Would you approve changing the connectSubgraphOutputs API to pass in a map instead for output_entries? Then it will simplify the implementation of connectSubgraphOutputs and help backend developers avoid making bugs by having to dedupe in this weird way. 

@HahTK and I are willing to make this change. 

Heres what we propose:

Change FindOutputEntries to populate a map<int, vector<nodeentry*> > instead of the vector<nodeentry*>. The integer refers to the i'th output of the subgraph, and nodeentry pointer is the same as it is in output_entries. The job of connectSubgraphOutputs is to connect all nodeentries in the vector to that indexed output of the subgraph.

Then in CreateSubgraphNode (in build_subgraph.cc) the code will look like:

```
nnvm::Symbol sym;
  size_t idx = 0;
  sym.outputs.resize(output_map.size());
  for (size_t i = 0; i < output_map.size(); ++i) {
    sym.outputs[i] = *output_map[i][0];
  }
```

And ConnectSubgraphOutputs will take the output_map and do:

```
for (size_t i = 0; i < output_map->size(); ++i) {
   for(auto e : output_map[i]) {
     *e = nnvm::NodeEntry{subgraph_node, static_cast<uint32_t>(i), 0};
    }
}
```

This vastly simplifies/clarifies the code backend devs need to write. And helps avoid confusion and prevents bugs.

ZhennanQin: This solution is more clear! I like this change. Please go ahead to implement it. Thanks.

src/operator/subgraph/build_subgraph.cc
```diff
@@ -450,17 +450,18 @@ void FindInputEntries(const nnvm::Graph& g,
...
- * \param output_entries output entries of the subgraph
+ * \param output_map output entries of the subgraph
  */
 void FindOutputEntries(nnvm::Graph* g,
                        const std::vector<BiDirectedNodePtr>& simple_nodes,
                        const std::vector<BiDirectedNode*>& subgraph_nodes,
                        const std::unordered_map<const nnvm::NodeEntry*, size_t>&
                          entry_top_order_map,
-                       std::vector<nnvm::NodeEntry*>* output_entries) {
+                       std::map<int, std::vector<nnvm::NodeEntry*> > &output_map) {
```
ZhennanQin: Use `std::vector<std::vector<nnvm::NodeEntry*>>` instead. This is fast(O(1) for search and push_back), and can avoid element missing(for std::map, we need additional check to ensure that for each i < output_map.size(), output_map[i] exists).

samskalicky:
@ZhennanQin ive made the changes, do they still look good? If so we can start going through and updating all other subgraph properties.

Regarding the update, what do you think we should do with properties that implement the standard logic for connectSubgraphInputs like:

https://github.com/apache/incubator-mxnet/blob/35e6d7a0edac416532bb754ba4e570a9df67463e/src/operator/subgraph/mkldnn/mkldnn_conv_property.h#L236-L243

Should we just eliminate this definition and continue to use the parent/standard implementation?

ZhennanQin:
LGTM. Please coordinate all backends  with new ConnectSubgraphOutputs().

DickJC123:
I've noticed that MXNet does not preserve what I'm calling 'output independence' during it's memory planning (issue forthcoming).  While I'm talking here about the model as a whole, I've noticed many graph passes are applied to subgraphs as well, so perhaps the issue is relevent to this PR.  Any update on status here?

samskalicky:
Hi @DickJC123 sorry I havent been able to prioritize finishing off this PR. IMO the problem is similar and endemic in mxnet but not related to this PR. Please do file an issue at the very least for posterity.

DickJC123:
I have not begun to work on this, and my plate is fairly full, so someone else can jump in if they want.  The issue can be fixed narrowly to match the failing case I posted, but I was hoping for someone to also understand/fix the 'endemic' issue @samskalicky mentions in https://github.com/apache/incubator-mxnet/pull/16131.

samskalicky:
@mseth10 looks like we made an aux an arg somehow, thats why the MKLDNN tests are failing:
| in_args_map | arg_names | arg_names |
|---|---|---|
data|
conv1_weight|
bn1_gamma|
bn1_beta|
conv2_weight|
bn2_gamma|
bn2_beta|

src/operator/subgraph/subgraph_property.h
```diff
@@ -342,8 +342,27 @@ class SubgraphProperty {
...
+      const std::string& var_name = output_names[0];
+
+      auto it = name_count_map.find(var_name);
+      // if it is not in the map yet, add it and increment the output index
+      if (name_count_map.end() == it) {
+        name_count_map.emplace(var_name, nnvm::NodeEntry{subgraph_node, idx, 0});
+        idx++;
+      }
+      // set output to index from map
+      *output_entries->at(i) = name_count_map[var_name];
```
samskalicky, line 365: @Kh4L should we do something like this for the tensorrt-inl.h `ConnectSubgraphOutputs` too?

src/operator/subgraph/build_subgraph.cc
```diff
@@ -605,20 +609,31 @@ void CreateSubgraphNode(nnvm::Graph* g,
...
-    sym.outputs[i] = *output_entries[i];
+    if (i == 0) {  // add first entry
+      sym.outputs[idx] = *output_entries[i];
+    } else if (!node_equal(sym.outputs[idx], *output_entries[i])) {  // compare to see if unique
+      // add non-dupe entries
+      idx++;
+      sym.outputs[idx] = *output_entries[i];
+    }  // else skip over dupe entries
   }
+  sym.outputs.resize(idx+1);
```
samskalicky, line 626: @Kh4L heres the code that deduplicates the outputs from the subgraph itself. With this, the subgraph only contains 1 output entry for each output instead of an output entry for each consuming node (where there may be duplicate entries for multiple consumers of a single output)

src/operator/subgraph/build_subgraph.cc
```diff
@@ -537,10 +537,11 @@ void FindOutputEntries(nnvm::Graph* g,
  */
 void CutGraphInputs(const std::vector<nnvm::NodeEntry*> &input_entries,
                     std::vector<nnvm::NodeEntry> *orig_entries,
+                    std::vector<nnvm::NodeEntry> *unique_inputs,
```
HahTK, line 540: There is an alternative implementation that does not need t track the actual unique inputs separately.
A counter would allow us to correctly modify orig_entries into unique_entries.
This keeps the function signature unchanged and minimizes changes elsewhere.

It is already working in a private build. I add it here for consideration if desired

samskalicky: Thanks for the suggestion @hahtk, but we cant modify `orig_entries` since we need it unmodified for ReattachGraphInputs:
https://github.com/apache/incubator-mxnet/blob/e249d71ad4621afcba7f0f3af77095a3f9a4bc83/src/operator/subgraph/build_subgraph.cc#L578
this is used to reject a subgraph from the reviewSubgraph API:
https://github.com/apache/incubator-mxnet/blob/e249d71ad4621afcba7f0f3af77095a3f9a4bc83/src/operator/subgraph/build_subgraph.cc#L657
the example you provided doesnt work in the latest version of MXNet where you can reject a subgraph by returning nullptr from CreateSubgraphNode here: https://github.com/apache/incubator-mxnet/blob/e249d71ad4621afcba7f0f3af77095a3f9a4bc83/src/operator/subgraph/build_subgraph.cc#L630-L633

HahTK: The proposed change was tested on 1.5.1 and it seems like newer versions of MXNet have new features that needs orig_entries to be preserved. So acknowledged !

HahTK (approved):

waytrue17 (approved):
Looks good to me, thanks!

mseth10 (approved):
LGTM. Thanks for the fix.

samskalicky:
> @mseth10 looks like we made an aux an arg somehow, thats why the MKLDNN tests are failing:
> 
> in_args_map	arg_names	arg_names
> data	data	data
> conv1_weight	conv1_weight	conv1_weight
> bn1_gamma	bn1_gamma	bn1_gamma
> bn1_beta	bn1_beta	bn1_beta
> conv2_weight	conv2_weight	conv2_weight
> bn2_gamma	bn2_gamma	bn2_gamma
> bn2_beta	bn2_beta	bn2_beta
> **bn2_moving_mean**
> So this is why the error says
> 
> ```
> MXNetError: Check failed: arg_names.size() == in_args_map.size() (8 vs. 7) : 
> ```
> 
> Theres one extra name in `arg_names` for `bn2_moving_mean`

For the record @mseth10 i finally found the issue that caused this "making an aux an arg" problem: https://github.com/apache/incubator-mxnet/pull/19112#issuecomment-690810445

PR #15768, opened 2019-08-06
Fix gradient tensor mutate in `{adam/ftrl/rmprop/rmspropalex}_update`.

kshitij12345:
## Description ##
Rescaling the gradient used to update the data of `grad` passed as input.
Detailed in #15759 

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] Fix the bug.
- [x] Relevant test.

PR #16198, opened 2019-09-18
[fix] Update `test_update_ops_mutation` tolerance

kshitij12345:
Fixes:
https://github.com/apache/incubator-mxnet/pull/15768#issuecomment-532046408

@apeforest  @sxjscience @Vikas-kum

### DIFF
--- a/tests/python/unittest/test_ndarray.py
+++ b/tests/python/unittest/test_ndarray.py
@@ -21,6 +21,7 @@ from distutils.version import LooseVersion
 from itertools import permutations, combinations_with_replacement
 import os
 import pickle as pkl
+import functools
 from nose.tools import assert_raises, raises
 from common import with_seed, assertRaises, TemporaryDirectory
 from mxnet.test_utils import almost_equal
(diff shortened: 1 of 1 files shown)


tests/python/unittest/test_ndarray.py
```diff
@@ -1887,14 +1888,16 @@ def check_save_load(save_is_np_shape, load_is_np_shape, shapes, save_throw_excep
     check_save_load(True, True, [(2, 0, 1), (0,), (), (), (0, 4), (), (3, 0, 0, 0), (2, 1), (0, 5, 0)], False, False)
 
 
-@with_seed()
-def test_update_ops_mutation():
-    def assert_mutate(x, y, op):
+def _test_update_ops_mutation_impl():
+    assert_allclose = functools.partial(
+                np.testing.assert_allclose, rtol=1e-10)
```
Vikas-kum: are you making it more sensitive. Will that cause more failures going forward? 
I think you need different comparison function which says check for equality upto this precision - 
Something like - assert_array_almost_equal

Vikas-kum: @sxjscience @kshitij12345 sorry there was race between I writing this and Xingjian merging the change. Can you please look at this comment.

sxjscience: I think we need to lower the rtol because sometimes it fails the assert_raise check

Vikas-kum: IIUC , with the current check is if difference is less than 1e-10 , test will fail. Is this correct ?
I think it should be I don't care about 7th(or 8th precision or anything above) precision, it can go arbitrary lower than what I care about.

sxjscience: We are `assert_raise`, which means that if the difference is larger than rtol, `assert_allclose` will fail and it will raise the AssertionError.

kshitij12345: @Vikas-kum @sxjscience Thanks for taking a look.

The new tolerance of `assert_allclose` ([numpy doc](https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_almost_equal.html) mentions to use `assert_allclose` over `assert_almost_equal`) is now 1e-10, which means that it is more stricter now when checking for NDArray not being changed. 

A consequence of this as it is used in `assert_mutated` is, it also allows smaller updates, i.e. -5.9604645e-08 (failure case update) to be valid change . 

Thus now we have stricter check on whether any element of NDArray has changed or not (with stricter tolerance) while allowing smaller valid updates (i.e. anything above `rtol` of 1e-10 will be a valid update).

> I think it should be I don't care about 7th(or 8th precision or anything above) precision, it can go arbitrary lower than what I care about.

(**I don't know much about this but**) With the instability of floating point number (rounding errors), I guess it is better to have some tolerance like 1e-10. I don't know if it is good idea to go to Machine Epsilon (lowest possible) as `rtol` for this.

sxjscience: @kshitij12345 I think it should be safe to remove the test_mutate check.

PR #16207, opened 2019-09-19
Bump numpy version >=1.17.0

reminisce:
## Description ##
To adopt numpy `__array_function__` protocol for better `ndarray` interoperability, we need numpy >= 1.17.0. See discussion in https://github.com/apache/incubator-mxnet/pull/16097.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -30,7 +30,11 @@ if "--inplace" in sys.argv:
 else:
     from setuptools import setup
     from setuptools.extension import Extension
-    kwargs = {'install_requires': ['numpy>=1.17.0,<2.0.0', 'requests>=2.20.0,<3', 'graphviz<0.9.0,>=0.8.1'], 'zip_safe': False}
+    kwargs = {
+        'install_requires': ['requests>=2.20.0,<3', 'graphviz<0.9.0,>=0.8.1']
+            .append('numpy>=1.17.0,<2.0.0' if sys.version_info[0] > 2 else 'numpy>1.16.0,<2.0.0'),
+        'zip_safe': False
+    }
 
 with_cython = False
 if '--with-cython' in sys.argv:
--- a/python/setup.py
+++ b/python/setup.py
@@ -30,7 +30,7 @@ if "--inplace" in sys.argv:
 else:
     from setuptools import setup
     from setuptools.extension import Extension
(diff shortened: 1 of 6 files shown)


szha:
@leezu FYI

python/setup.py
```diff
@@ -30,7 +30,11 @@
 else:
     from setuptools import setup
     from setuptools.extension import Extension
-    kwargs = {'install_requires': ['numpy>=1.17.0,<2.0.0', 'requests>=2.20.0,<3', 'graphviz<0.9.0,>=0.8.1'], 'zip_safe': False}
+    kwargs = {
+        'install_requires': ['requests>=2.20.0,<3', 'graphviz<0.9.0,>=0.8.1']
+            .append('numpy>=1.17.0,<2.0.0' if sys.version_info[0] > 2 else 'numpy>1.16.0,<2.0.0'),
```
leezu: Why not use https://www.python.org/dev/peps/pep-0508/#environment-markers ?

leezu:
`numba<0.45.1` does not work with numpy 1.17. So users of numba must upgrade when using numpy 1.17. I believe we should nevertheless bump the MXNet numpy dependency version. But FYI, to make you aware of the caveat.

python/setup.py
reminisce: Thanks for pointing this out. I'm not familiar with the tool, but will adopt it in this PR after CI passes.

python/setup.py
```diff
@@ -30,7 +30,12 @@
 else:
     from setuptools import setup
     from setuptools.extension import Extension
-    kwargs = {'install_requires': ['numpy>1.16.0,<2.0.0', 'requests>=2.20.0,<3', 'graphviz<0.9.0,>=0.8.1'], 'zip_safe': False}
+    kwargs = {
+        'install_requires': ['requests>=2.20.0,<3', 'graphviz<0.9.0,>=0.8.1']
+            .append('numpy>=1.17.0,<2.0.0' if sys.version_info.major >= 3 and sys.version_info.minor >= 5
+                    else 'numpy>1.16.0,<2.0.0'),
```
szha: what's the concern for just switching to `>= 1.17`? was it CI?
setup.py may be evaluated on a platform that builds binary distributions that is different from the platform on which the distribution is installed. in this scenario, the condition won't be evaluated on the target platform

leezu: `1.17` is Python 3 only.
If the condition is not evaluated on the target platform, we can't bump the version without dropping Python 2 support..

szha: agreed. let's put this change on hold

PR #16215, opened 2019-09-19
New ops for RCNN + old ops improvements for RCNN

Jerryzcn:
## Description ##
1. Box Encoder for RCNN
2. Box Decoder for RCNN
3. amp_multicast can cast to narrowest type now
4.  roi_align ignore rois that have batchid less than 0

Have trained mask-rcnn with the change, and the convergence is the same as before the change


Jerryzcn:
> can we have unit tests for `box_encode` and `box_decode`?

sure

src/operator/contrib/bounding_box-inl.h
```diff
@@ -787,6 +787,284 @@ void BipartiteMatchingBackward(const nnvm::NodeAttrs& attrs,
...
+  // assign output shape
+  mxnet::TShape oshape = ashape;
+  SHAPE_ASSIGN_CHECK(*out_attrs, 0, oshape);
+  SHAPE_ASSIGN_CHECK(*out_attrs, 1, oshape);
+  return shape_is_known(oshape);
+}
+
+struct box_encode {
+  template<typename DType>
+  MSHADOW_XINLINE static void Map(int i, DType *out_targets, DType *out_masks,
```
apeforest: Can we use `index_t` here for i, m, n in order to support large tensor?

src/operator/contrib/bounding_box-inl.h
```diff
@@ -787,6 +787,284 @@ void BipartiteMatchingBackward(const nnvm::NodeAttrs& attrs,
...
+  // assign output shape
+  mxnet::TShape oshape = dshape;
+  SHAPE_ASSIGN_CHECK(*out_attrs, 0, oshape);
+  return shape_is_known(oshape);
+}
+
+template<int anchor_encode, bool has_clip>
+struct box_decode {
+  template<typename DType>
+  MSHADOW_XINLINE static void Map(int i, DType *out, const DType *x,
```
apeforest: Same here. Please use `index_t` for tensor index.

src/operator/contrib/bounding_box-inl.h
Jerryzcn: understood.

src/operator/contrib/bounding_box-inl.h
Jerryzcn: understood

PR #16178, opened 2019-09-16
[WIP]improving argmax perf

access2rohit:
## Description ##
Removes mshadow implementation of argmax and add fresh implementation for it

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Performance Numbers ###
Machine: 
Intel(R) Xeon(R) CPU E5-2686 v4 @ 2.30GHz
32 Core
L1d cache:             32K
L1i cache:             32K
L2 cache:              256K
L3 cache:              46080K

|          | input             | axis | int32_p50 | int64_p50 | int64/int32 |
|----------|-------------------|------|-----------|-----------|-------------|
| Master   | 1024, 1024        | 1    | 0.71      | 0.99      | 139.51      |
| new_impl | 1024, 1024        | 1    | 0.78      | 0.66      | 84.11       |
| master   | 1024, 1024        | 0    | 8.30      | 17.77     | 213.96      |
| new_impl | 1024, 1024        | 0    | 0.76      | 0.75      | 98.70       |
| master   | 1024, 1024        | NA   | 5.41      | 12.10     | 223.52      |
| new_impl | 1024, 1024        | NA   | 0.27      | 0.31      | 113.24      |
| master   | 10000, 10000      | 1    | 39.28     | 79.76     | 203.06      |
| new_impl | 10000, 10000      | 1    | 52.15     | 35.66     | 68.38       |
[...]

PR #16218, opened 2019-09-19
Improving performance of argmax operator

drivanov:
## Description ##
In average, this implementation of `argmax` operator runs 4.2x faster on CPU and 7.8x faster on GPU, than the previous one.

### Essentials ###
- [x ] Code is well-documented: 

### Changes ###
- [x] Two new kernels are implemented. 

## Comments ##
- The performance of new argmax was tested on 69 different pairs `(shape, axis)` of 2- and 3-dimensional shapes. Following table contains some performance results for run time improvement:
```
                CPU:       GPU:        
Average:       4.163x     7.753x
Maximal:      15.287x   100.157x 
```
Significant improvement achieved mostly by using multiple threads working on the same vector.


src/operator/tensor/broadcast_reduce_op_index.cu
```diff
@@ -43,5 +43,18 @@ NNVM_REGISTER_OP(pick)
...
+template<>
+int DefineNumbWorkers<gpu>(const TShape &shape, int axis) {
+  const auto nSteps = static_cast<uint32_t>(shape[axis]);
+  const auto nThreads = shape.Size()/nSteps;
+  if (nThreads > nSteps)
+    return 1;
+
+  const auto a = static_cast<float>(nSteps)/nThreads;
+  const auto b = log2f(a);
+  const auto numbWorkers = pow(2, (b * 5 + 28)/11);
```
wkcn: Why is the number of workers pow(2, (b * 5 + 28)/11)?

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -556,6 +556,162 @@ inline bool ReduceAxesOpForwardStorage(const nnvm::NodeAttrs& attrs,
   return dispatched;
 }
 
+struct argmax {
+  template<typename DType>
+  MSHADOW_XINLINE static void Map(int i, const int nWorkers, const DType *in_data, DType *out_data,
```
apeforest: Please use `index_t` for Map index so that we can support large tensor.

apeforest:
It seems this PR is working on the same operator as https://github.com/apache/incubator-mxnet/pull/16178. Can you run a profiling using https://github.com/apache/incubator-mxnet/tree/master/benchmark/opperf and show the performance gain of your change?

src/operator/tensor/broadcast_reduce_op_index.cu
drivanov: This is just a heuristic. Experimenting with different values `(shape, axis, numbWorkers)`, I accumulated a lot of data, it turned out that they almost perfectly corresponded to this formula.

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -556,6 +556,162 @@ inline bool ReduceAxesOpForwardStorage(const nnvm::NodeAttrs& attrs,
...
+  }
+};
+
+struct argmax_reduce {
+  template<typename DType>
+  MSHADOW_XINLINE static void Map(int i, const int nWorkers, const DType *in_data, DType *out_data,
+               const size_t step, const size_t shift, void *pIdx, const bool use_uin16) {
+    const DType *pCurr = in_data + i % step + shift * (i / step);
+    int maxIdx;
+    if (use_uin16) {
```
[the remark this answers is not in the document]
drivanov: Yes, I also thought about it, and I even wrote one. Unfortunately, I could not compile this template. Perhaps, I will try one more time.

src/operator/tensor/broadcast_reduce_op.h
drivanov: OK, I will do that.

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -556,6 +556,162 @@ inline bool ReduceAxesOpForwardStorage(const nnvm::NodeAttrs& attrs,
...
+    const size_t num_items = nWorkers * shape.Size() / shape[axis];
+    pIdxMemory = AllocateDTypeMemory<xpu, uint32_t>(ctx, num_items);
+    if (pIdxMemory)
+      break;
+
+    // Check if indexes can be stored in uint16 format
+    if (shape[axis] <= UINT16_MAX) {
+      pIdxMemory = AllocateDTypeMemory<xpu, uint16_t>(ctx, num_items);
+      if (pIdxMemory) {
+        use_uint16 = false;
```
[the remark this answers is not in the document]
drivanov: Good catch! It's is definitely a bug and I will fix it

drivanov:
> It seems this PR is working on the same operator as #16178. Can you run a profiling using https://github.com/apache/incubator-mxnet/tree/master/benchmark/opperf and show the performance gain of your change?

I will do it and will collect performance data for all three version: existing one, #16178 and #16218

src/operator/tensor/broadcast_reduce_op.h
drivanov: OK, I will do that.

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -556,6 +556,179 @@ inline bool ReduceAxesOpForwardStorage(const nnvm::NodeAttrs& attrs,
...
+      const auto in = i / nWorkers;
+      iw = i % nWorkers;
+      pCurr += in % step + shift * (in / step) + iw * step;
+      nSteps = (nSteps + nWorkers - 1 - iw) / nWorkers;
+      step *= nWorkers;
+    } else {
+      pCurr += i % step + shift * (i / step);
+    }
+
+    uint32_t maxIdx = 0;
```
apeforest: Can we use `index_t` here because we need to support large tensor?

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -556,6 +556,179 @@ inline bool ReduceAxesOpForwardStorage(const nnvm::NodeAttrs& attrs,
...
+      out_data[i] = maxIdx;    // output of argmax
+    }
+  }
+};
+
+struct argmax_reduce {
+  template<typename DType, typename IType>
+  MSHADOW_XINLINE static uint32_t BestIdx(const int nWorkers, const DType *pCurr,
+                                          const uint32_t step, const IType *pIdxStorage) {
+    uint32_t maxIdx = *pIdxStorage;
```
apeforest: Can we use `index_t` here because we need to support large tensor?

apeforest:
@drivanov Thanks a lot for your contribution and profiling the performance. This looks really great. We'd love to merge your PR ASAP. I just had a few comments regarding large tensor support. Since we plan to support large tensor (index over MAX_INT32), could you please verify if you change can run through a large array by compiling MXNet with compiler flag USE_INT64_TENSOR_SIZE on?

We've already had a nightly test for this operator: https://github.com/apache/incubator-mxnet/blob/master/tests/nightly/test_large_array.py#L1031

Thanks!

access2rohit:
@drivanov can you share the build procedure of your code base so your results can be reproduced ?
How did you benchmark your code ?
I tried to use opperf utility present in MXNet and i get errors 
```
home/ubuntu/anaconda3/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
INFO:root:Begin Benchmark - argmax
INFO:root:Complete Benchmark - argmax
[{'argmax': [{'avg_time_argmax': 0.29419124126434326, 'p50_time_argmax': 0.28957799077033997, 'p90_time_argmax': 0.3093678504228592, 'p99_time_argmax': 0.3325981646776199, 'inputs': {'data': (1024, 1024), 'axis': 1}}]}]
INFO:root:Begin Benchmark - argmax
INFO:root:Complete Benchmark - argmax
[{'argmax': [{'avg_time_argmax': 0.38427121937274933, 'p50_time_argmax': 0.38124993443489075, 'p90_time_argmax': 0.39454028010368347, 'p99_time_argmax': 0.40961168706417084, 'inputs': {'data': (1024, 1024), 'axis': 0}}]}]
  File "perf_argmax.py", line 7, in <module>
    print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(1024,1024)}],warmup=10, runs=25, profiler='python'))
  File "/home/ubuntu/PERF/drivanov/incubator-mxnet/benchmark/opperf/utils/benchmark_utils.py", line 137, in run_performance_test
    benchmark_result = _run_nd_operator_performance_test(op, inputs, run_backward, warmup, runs, args_list, kwargs_list, profiler)
  File "/home/ubuntu/PERF/drivanov/incubator-mxnet/benchmark/opperf/utils/benchmark_utils.py", line 69, in _run_nd_operator_performance_test
    _, _ = benchmark_helper_func(op, warmup, [], **kwargs_list[0])
  File "/home/ubuntu/PERF/drivanov/incubator-mxnet/benchmark/opperf/utils/profiler_utils.py", line 241, in python_profile_it
    res = func(*modified_args, **kwargs)
  File "/home/ubuntu/PERF/drivanov/incubator-mxnet/benchmark/opperf/utils/ndarray_utils.py", line 52, in nd_forward_backward_and_profile
    nd.waitall()
  File "/home/ubuntu/PERF/drivanov/incubator-mxnet/python/mxnet/ndarray/ndarray.py", line 188, in waitall
    check_call(_LIB.MXNDArrayWaitAll())
  File "/home/ubuntu/PERF/drivanov/incubator-mxnet/python/mxnet/base.py", line 254, in check_call
mxnet.base.MXNetError: [22:48:58] src/operator/tensor/./broadcast_reduce_op.h:689: Global reduction not supported yet
Stack trace:
  [bt] (0) /home/ubuntu/PERF/drivanov/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x32) [0x7fb239db4cb2]
  [bt] (1) /home/ubuntu/PERF/drivanov/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(void mxnet::op::ArgMax<mshadow::cpu>(nnvm::NodeAttrs const&, mxnet::OpContext const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&, std::vector<mxnet::OpReqType, std::allocator<mxnet::OpReqType> > const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&)+0x92) [0x7fb23b3973d2]
  [bt] (2) /home/ubuntu/PERF/drivanov/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(mxnet::imperative::PushFCompute(std::function<void (nnvm::NodeAttrs const&, mxnet::OpContext const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&, std::vector<mxnet::OpReqType, std::allocator<mxnet::OpReqType> > const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&)> const&, nnvm::Op const*, nnvm::NodeAttrs const&, mxnet::Context const&, std::vector<mxnet::engine::Var*, std::allocator<mxnet::engine::Var*> > const&, std::vector<mxnet::engine::Var*, std::allocator<mxnet::engine::Var*> > const&, std::vector<mxnet::Resource, std::allocator<mxnet::Resource> > const&, std::vector<mxnet::NDArray*, std::allocator<mxnet::NDArray*> > const&, std::vector<mxnet::NDArray*, std::allocator<mxnet::NDArray*> > const&, std::vector<unsigned int, std::allocator<unsigned int> > const&, std::vector<mxnet::OpReqType, std::allocator<mxnet::OpReqType> > const&)::{lambda(mxnet::RunContext)#1}::operator()(mxnet::RunContext) const+0x2b2) [0x7fb23c8b8ae2]
  [bt] (3) /home/ubuntu/PERF/drivanov/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(std::_Function_handler<void (mxnet::RunContext), mxnet::imperative::PushFCompute(std::function<void (nnvm::NodeAttrs const&, mxnet::OpContext const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&, std::vector<mxnet::OpReqType, std::allocator<mxnet::OpReqType> > const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&)> const&, nnvm::Op const*, nnvm::NodeAttrs const&, mxnet::Context const&, std::vector<mxnet::engine::Var*, std::allocator<mxnet::engine::Var*> > const&, std::vector<mxnet::engine::Var*, std::allocator<mxnet::engine::Var*> > const&, std::vector<mxnet::Resource, std::allocator<mxnet::Resource> > const&, std::vector<mxnet::NDArray*, std::allocator<mxnet::NDArray*> > const&, std::vector<mxnet::NDArray*, std::allocator<mxnet::NDArray*> > const&, std::vector<unsigned int, std::allocator<unsigned int> > const&, std::vector<mxnet::OpReqType, std::allocator<mxnet::OpReqType> > const&)::{lambda(mxnet::RunContext)#1}>::_M_invoke(std::_Any_data const&, mxnet::RunContext&&)+0x17) [0x7fb23c8b8cc7]
  [bt] (4) /home/ubuntu/PERF/drivanov/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(+0x3772bfe) [0x7fb23ce12bfe]
  [bt] (5) /home/ubuntu/PERF/drivanov/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(mxnet::engine::ThreadedEngine::ExecuteOprBlock(mxnet::RunContext, mxnet::engine::OprBlock*)+0x995) [0x7fb23ce0d115]
  [bt] (6) /home/ubuntu/PERF/drivanov/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(std::_Function_handler<void (std::shared_ptr<dmlc::ManualEvent>), mxnet::engine::ThreadedEnginePerDevice::PushToExecute(mxnet::engine::OprBlock*, bool)::{lambda()#1}::operator()() const::{lambda(std::shared_ptr<dmlc::ManualEvent>)#1}>::_M_invoke(std::_Any_data const&, std::shared_ptr<dmlc::ManualEvent>&&)+0xfe) [0x7fb23ce208de]
  [bt] (7) /home/ubuntu/PERF/drivanov/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(std::thread::_Impl<std::_Bind_simple<std::function<void (std::shared_ptr<dmlc::ManualEvent>)> (std::shared_ptr<dmlc::ManualEvent>)> >::_M_run()+0x4a) [0x7fb23ce0bb7a]
  [bt] (8) /home/ubuntu/anaconda3/bin/../lib/libstdc++.so.6(+0xafc5c) [0x7fb249e03c5c]
```

drivanov:
@access2rohit: 
- When building I do not use any special tricks. flags, etc. I just launch  `make`.
- For banchmark I am using 
```
def test_argmax()
```
from `/opt/mxnet/tests/python/unittest/test_operator.py` and I launch it by using following command:
```
nosetests --verbose -s --logging-level=DEBUG /opt/mxnet/tests/python/unittest/test_operator.py:test_argmax
```
(BTW, `--logging-level=DEBUG` option needed to see the seed which is used by the test for randomly generated tensors. Otherwise the seed will be printed only if test should fail, which is not a case for this benchmark)
- To make the results 100% reproducible or/and benchmark different `argmax`'s with the same set of randomly tensors, you need to use the same seed. For instance, for the benchmark I did for your #16178, I was using `MXNET_TEST_SEED=1643676822`. If you want to reproduce these results, just call:
```
> export MXNET_TEST_SEED=1643676822
>  nosetests --verbose -s --logging-level=DEBUG /opt/mxnet/tests/python/unittest/test_operator.py:test_argmax
```
BTW, 
- if you will need to use another randomly used seed, just call 
```
> unset MXNET_TEST_SEED
```
before launching `nosetests --verbose -s --logging-level=DEBUG ...`.

drivanov:
@access2rohit:
Regarding the crash you see in your benchmark... Following line from your output:
```
mxnet.base.MXNetError: [22:48:58] src/operator/tensor/./broadcast_reduce_op.h:689: Global reduction not supported yet
``` 
shows that there is some problem with the definition of `axis` parameter. I am not sure, perhaps, it's is not defined in your test at all or it's default value is not yet supported. This is an error message from following to lines of C++ code:
```
  const ReduceAxisParam& param = nnvm::get<ReduceAxisParam>(attrs.parsed);
  if (!param.axis) LOG(FATAL) << "Global reduction not supported yet";
```
When everything fine, `param.axis` should NOT be `null_ptr`, otherwise `param.axis.values()` would fail.

access2rohit:
This code doesn't fail on master. But if your code is failing then you are changing the behaviour of a customer facing API.

the following is my code that I use to benchmark

```
import mxnet as mx
from mxnet import nd

print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(1024,1024), "axis":1}],warmup=10, runs=25, profiler='python'))
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(1024,1024), "axis":0}],warmup=10, runs=25, profiler='python'))
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(1024,1024)}],warmup=10, runs=25, profiler='python'))
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(10000,10000), "axis":1}],warmup=10, runs=25, profiler='python'))
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(10000,10000), "axis":0}],warmup=10, runs=25, profiler='python'))
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(10000,10000)}],warmup=10, runs=25, profiler='python'))
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(1000, 1000, 100), "axis":2}],warmup=10, runs=25, profiler='python'))
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(1000, 1000, 100), "axis":1}],warmup=10, runs=25, profiler='python'))
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(1000, 1000, 100), "axis":0}],warmup=10, runs=25, profiler='python'))
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(1000, 1000, 100)}],warmup=10, runs=25, profiler='python'))
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(100, 100, 100, 100), "axis":3}],warmup=10, runs=25, profiler='python'))
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(100, 100, 100, 100), "axis":2}],warmup=10, runs=25, profiler='python'))
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(100, 100, 100, 100), "axis":1}],warmup=10, runs=25, profiler='python'))
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(100, 100, 100, 100), "axis":0}],warmup=10, runs=25, profiler='python'))
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(100, 100, 100, 100)}],warmup=10, runs=25, profiler='python'))
```

drivanov:
Later today I will try your approach:
```
import mxnet as mx
from mxnet import nd

print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(1024,1024), "axis":1}],warmup=10, runs=25, profiler='python'))
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(1024,1024), "axis":0}],warmup=10, runs=25, profiler='python'))
```

src/operator/tensor/broadcast_reduce_op.h
drivanov: @apeforest: I could do that, but are you sure that you really need this? According to following comments from `config.mk`
```
# Use int64_t type to represent the total number of elements in a tensor
# This will cause performance degradation reported in issue #14496
# Set to 1 for large tensor with tensor size greater than INT32_MAX i.e. 2147483647
# Note: the size of each dimension is still bounded by INT32_MAX
USE_INT64_TENSOR_SIZE = 1
```
**the size of each dimension is still bounded by INT32_MAX**. I think, even `uint32_t` currently used here should be more than enough.

drivanov:
@access2rohit: OK, I know what is wrong with your test launching my `argmax`. Everything is fine, when axis is specified by it's value (0<=axis < dimension): 
```
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(1024,1024), "axis":0}],warmup=10, runs=25, profiler='python'))
```
OR as "-1" (which is equivalent to `axis = lastDimension = dimension - 1` 
```
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(1024,1024), "axis":-1}],warmup=10, runs=25, profiler='python'))
```
But when `axis` is **not directly specified**: 
```
print(run_performance_test(nd.argmax, run_backward=True, dtype='float32', ctx=mx.cpu(),inputs=[{"data":(1024,1024)}],warmup=10, runs=25, profiler='python')
```
we are getting crash with the error message:
```
mxnet.base.MXNetError: [21:07:34] src/operator/tensor/./broadcast_reduce_op.h:683: Global reduction not supported yet
```
When I was implementing new version of `argmax`, I left everything which was in previously used interface, including following two lines: 
```
  const ReduceAxisParam& param = nnvm::get<ReduceAxisParam>(attrs.parsed);
  if (!param.axis) LOG(FATAL) << "Global reduction not supported yet";
```
If you don't see this problem in your implementation of `argmax`, it means that you in some way implemented  **"Global reduction"**. I will investigate your code and I will try to do something similar.

drivanov:
@access2rohit: I understood what you did for `param.axis.has_value() == False` and tomorrow I will push corresponding changes into my #16218. Till then, if you want, you can run my code in all your benchmarks where `axis` is explicitly defined.

src/operator/tensor/broadcast_reduce_op.h
apeforest: We are working to support the size of each dimension beyond INT32_MAX in MXNet 1.6. That's why I suggest you to use `index_t`. In 1.5.1, `index_t` is int32_t by default; if the compiler flag USE_INT64_TENSOR_SIZE = 1, index_t is int64_t. In 1.6 release, this compiler flag will be ON by default.

drivanov:
@access2rohit: I just pushed changes fixing the `argmax` with the global reduction. All your benchmarks running #16218 should be OK.

src/operator/tensor/broadcast_reduce_op.h
drivanov: DONE.

src/operator/tensor/broadcast_reduce_op.h
drivanov: DONE.

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -556,6 +556,227 @@ inline bool ReduceAxesOpForwardStorage(const nnvm::NodeAttrs& attrs,
...
+    axis = 1;
+  }
+
+  void *pIdxMemory = nullptr;
+  auto nWorkers = DefineNumbWorkers<xpu>(shape, axis);
+  idxStorageType storageType = undefined_st;
+  while (nWorkers > 1) {
+    const size_t num_items = nWorkers * shape.Size() / shape[axis];
+#if MXNET_USE_INT64_TENSOR_SIZE
+    pIdxMemory = AllocateDTypeMemory<xpu, index_t>(ctx, num_items);
```
apeforest: `index_t` is not always `int64_t` as is defined here: https://github.com/apache/incubator-mxnet/blob/master/3rdparty/mshadow/mshadow/base.h#L291

Actually why do we need this check of storageType? Can we just pass in `index_t` and let compiler do the job?

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -556,6 +556,227 @@ inline bool ReduceAxesOpForwardStorage(const nnvm::NodeAttrs& attrs,
...
+    if (pIdxMemory) {
+      storageType = int64_st;
+      break;
+    }
+#endif
+    if (shape[axis] <= UINT32_MAX) {
```
apeforest: I think `uint32_t` is not being used for shape index any more. It's either `int64_t` or `int32_t` depending on the compiler flag. As I commented earlier, it might be cleaner to just pass in `index_t` directly.

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -556,6 +556,227 @@ inline bool ReduceAxesOpForwardStorage(const nnvm::NodeAttrs& attrs,
...
+      pIdxMemory = AllocateDTypeMemory<xpu, uint32_t>(ctx, num_items);
+      if (pIdxMemory) {
+        storageType = uint32_st;
+        break;
+      }
+    }
+    // Check if indexes can be stored in uint16 format
+    if (shape[axis] <= UINT16_MAX) {
+      pIdxMemory = AllocateDTypeMemory<xpu, uint16_t>(ctx, num_items);
```
apeforest: Is this working? I thought the index inside tensor object always use int32 or int64. Would this cause overflow if uint16_t is passed in. Also based on Google C++ style guide, we should avoid using unsigned integer to represent data type: https://google.github.io/styleguide/cppguide.html#Integer_Types

src/operator/tensor/broadcast_reduce_op_index.cu
apeforest: Could you add a line of comment about this heuristic?

apeforest: Is this only applicable to argmax? or it can be generalized?

tests/python/unittest/test_operator.py
```diff
@@ -7535,6 +7535,15 @@ def test_softmax():
     check_smoothed_softmax_grad(default_context())
 
 
+def create_data(batch_size, num_labels, ctx=None, dtype=None, H=None, W=None):
```
apeforest: Maybe you could reuse this test util https://github.com/apache/incubator-mxnet/blob/df34e768ed5d4dd334c0dce7ce42384d7b62ee8f/python/mxnet/test_utils.py#L92?

tests/python/unittest/test_operator.py
```diff
@@ -7604,6 +7612,161 @@ def _softmaxoutput_normalization(multi_output, use_ignore, normalization):
                     multi_output, use_ignore, normalization)
 
 
+@with_seed()
+def test_argmax():
+    ci_test = True       # Use False if you need to collect aggregated information about argmax performance
```
apeforest: Thanks for adding the script for performance test. But I feel this is probably not the best place to add performance test in unit test. Maybe can you separate the performance test logic and add it to https://github.com/apache/incubator-mxnet/blob/master/tests/nightly/test_large_array.py?

tests/python/unittest/test_operator.py
```diff
@@ -7604,6 +7612,161 @@ def _softmaxoutput_normalization(multi_output, use_ignore, normalization):
...
+                        buff += "   %6.3f (%3d)" % (end - begin, getNumWorkers(shape, axis, ctx=ctx))
+                    else:
+                        buff += "   %6.3f (%3d)" % (end - begin, nWorkers)
+                else:
+                    buff += "   %6.3f" % (end - begin)
+
+            if j == jMax - 1:
+                if nWorkers > 0:
+                    buff += "   ** %3d" % getNumWorkers(shape, axis, ctx=ctx)
+                print(buff)
```
apeforest: Please do not add print in unit test

drivanov: Operator `printt(buff)` you mentioned will never being called from unit test, because of `ci_test` value:
```
def test_argmax():
    ci_test = True       # Use False if you need to collect aggregated information about argmax performance
```
But when running benchmarks, profiling or debugging it's convenient to have this output on the screen.

tests/python/unittest/test_operator.py
drivanov: Thanks for the idea! In fact, I myself did not like a lot of `if`'s in this code. I have separated performance and unit test, but I think it would be better to keep
```
def profiling_argmax():
```
in `tests/python/unittest/test_operator.py`

The truth is: I tried to add it to `tests/nightly/test_large_array.py`, but it did not work for me. For some reason Python has some problem with this modules first lines:
```
root@627d15c17420:/opt/mxnet# python
Python 3.5.2 (default, Nov 12 2018, 13:43:14) 
[GCC 5.4.0 20160609] on linux
>>> import numpy as np
>>> import mxnet as mx
>>> from mxnet.test_utils import rand_ndarray, assert_almost_equal, rand_coord_2d
>>> from mxnet import gluon, nd
>>> from tests.python.unittest.common import with_seed
ImportError: No module named 'tests.python'
>>> 
```

tests/python/unittest/test_operator.py
drivanov: This `print(buff)` operator was not called for `ci_test = True` and after I splitted `test_argmax` in two function, it is not in unit test anymore.

src/operator/tensor/broadcast_reduce_op.h
drivanov: Actually, in my initial version (when the use of `int64_t` was not implemented):
```
 auto nWorkers = DefineNumbWorkers<xpu>(shape, axis);
 bool use_uin16 = false;
  while (nWorkers > 1) {
    const size_t num_items = nWorkers * shape.Size() / shape[axis];
      pIdxMemory = AllocateDTypeMemory<xpu, uint32_t>(ctx, num_items);
      if (pIdxMemory)
        break;

    // Check if indexes can be stored in uint16 format
    if (shape[axis] <= UINT16_MAX) {
      pIdxMemory = AllocateDTypeMemory<xpu, uint16_t>(ctx, num_items);
      if (pIdxMemory) {
        use_uin16 = true;
        break;
      }
    }

    // Decreasing the number of workers - less memory will be needed
    nWorkers >>= 1;
  }
```
I was getting "optimal" value of `nWorkers` first, and I tried to allocate memory for indexes in "unsigned" formats.  Perhaps, it is not always true (and, please, correct me, if I am wrong), but I noticed that with `uint32_t` pointers the program runs 5-10% faster than with `uint16_t`. Because of that, I was trying to allocate memory in that order. 
When 
- (a) for some value of `nWorkers` the program fails to allocate `uint32_t *` memory AND
- (b) `shape[axis] <= UINT16_MAX`, therefore it is still possible to use `uint16_t *` memory, 

I was trying to allocate and use `uint16_t *` memory for the same value of `nWorkers`. But, if
- (a) is not possible to use "shorter" memory (`shape[axis] > UINT16_MAX)` OR 
- (b) the attempt to allocate it failed, 

I reduced nWorkers >> = 1 and started a new round of memory allocation with smaller `nWorkers`. (Yes, now `nWorkers` is not optimal, but it's better to as much as possible workers!).

After I started to use `index_t`, I tried to follow the same pattern.

**NOTE,** when `index_t` is defined as `int64_t` I also see 5-10% run time degradation, when 
`int64_t *` memory is used for indexes storage. Therefore, with regard to runtime, this is not the best solution, but at least there is some logic here. Best solution would be use `int64_t` pointers **if and only if** it is impossible to store indexes in `uint32_t` format, otherwise use 
`uint32_t` or `uint_16_t` formats.

Definitely, I could forget about optimization of run time and use **only** the pointers of type `index_t *`. In that case the loop will be quite simple:
```
  while (nWorkers > 1) {
    const size_t num_items = nWorkers * shape.Size() / shape[axis];
    pIdxMemory = AllocateDTypeMemory<xpu, index_t>(ctx, num_items);
    if (pIdxMemory)
       break;
    nWorkers >>= 1;
  }
```

So it looks like we have following three options:
(1) keep current implementation (which looks logical, but note the best one when `index_t` is defined as `int64_t`)
(2) store indexes only `index_t` format (which is the simplest, but not the best one with respect to run time)
(3) implement the **"best"** solution: use `int64_t ` format only when `index_t` is defined as 'int64_t' AND `shape[axis] > UINT32_MAX`  

Which one do you prefer?

src/operator/tensor/broadcast_reduce_op.h
drivanov: Yes, I understand that `index_t` is not always `int64_t ` and it's why I used
```
#if MXNET_USE_INT64_TENSOR_SIZE
 ...
#endif
```
Please see below my detailed explanation regarding the usage of pointers of `int64_t/uin32_t/uint16_t` types.

src/operator/tensor/broadcast_reduce_op.h
drivanov: No, this will never cause a problem because before allocation of pointer of `int64_t/uint32_t/uint16_t` types, we always check if it is possible to store indexes in corresponding formats. 
For instance
```
if (shape[axis] <= UINT16_MAX)
 pIdxMemory = AllocateDTypeMemory<xpu, uint16_t>(ctx, num_items);
```
The index will never be bigger than the size of corresponding dimension `shape[axis]`. 
Therefore, we should no overflow for any type of `index_t`.

Regarding the Google C ++ style guide, about not-using an unsigned integers, I'm not sure if this is a universal tip, especially when we are talking about a significant difference in runtime (and 5-10% I would consider a significant difference).

apeforest: > No, this will never cause a problem because before allocation of pointer of `int64_t/uint32_t/uint16_t` types, we always check if it is possible to store indexes in corresponding formats.
> For instance
> 
> ```
> if (shape[axis] <= UINT16_MAX)
>  pIdxMemory = AllocateDTypeMemory<xpu, uint16_t>(ctx, num_items);
> ```
> 
> The index will never be bigger than the size of corresponding dimension `shape[axis]`.
> Therefore, we should no overflow for any type of `index_t`.
> 
> Regarding the Google C ++ style guide, about not-using an unsigned integers, I'm not sure if this is a universal tip, especially when we are talking about a significant difference in run time (and 5-10% degradation which I observed, I would consider a significant difference).

Interesting. So by changing from uint32_t to int32_t, there is a 5~10% runtime degradation?

tests/python/unittest/test_operator.py
drivanov: Firts of all,
```
def random_arrays(*shapes):
    """Generate some random numpy arrays."""
    arrays = [np.random.randn(*s).astype(default_dtype())
              for s in shapes]
```
generates only `np.arrays`'s and we need `mx.nd.array`'s here.
Secondly, all elements of 'np.array' can only be of type `np.foat32`, and `dtype` is the input parameter of
```
def create_data( ... dtype=None, ...):
```
Third, the same approach is used to create data for 
```
def test_softmax_output_normalization():
 . . .
        if multi_output:
            data = create_data(batch_size, num_labels, ctx=None, H=H, W=W)
            label_shape = (batch_size, H, W)
        else:
            data = create_data(batch_size, num_labels)
            label_shape = (batch_size, )
```
I reused this one which, in my opinion is more versatile.

src/operator/tensor/broadcast_reduce_op.h
drivanov: @apeforest: I got email with your comments
```
Interesting. So by changing from uint32_t to int32_t, there is a 5~10% runtime degradation?
```
For some reasons I don't see it here... Anyway, I will respond on this one.
It's what you could found in https://github.com/apache/incubator-mxnet/blob/master/make/osx.mk : 
```
# Use int64_t type to represent the total number of elements in a tensor
# This will cause performance degradation reported in issue #14496
# Set to 1 for large tensor with tensor size greater than INT32_MAX i.e. 2147483647
# Note: the size of each dimension is still bounded by INT32_MAX
USE_INT64_TENSOR_SIZE = 0
```

Here is the link to that issue: https://github.com/apache/incubator-mxnet/issues/14496
The degradation here is also approx 7%

apeforest: I added this compiler flag in the make file. The performance degradation is from int32_t to int64_t.

My earlier comment to your PR is choosing int32_t over uint32_t and I was curious to know if there is any performance difference between int32_t and uint32_t

drivanov: I saw approx. 5% degradation when I was using `uint32_t`.I'm not sure, maybe it's just some kind of noise, and for such a conclusion, additional testing is necessary.
To make the code more robust, I decided not to use `unsigned` indexes (see https://github.com/apache/incubator-mxnet/pull/16218/commits/bf843d55de98d04fe19ee20703cd33b2912f7ef1 and https://github.com/apache/incubator-mxnet/pull/16218/commits/c7bf38d4e4ca481aa5004c446407cab38f1b70f1)

ptrendx:
@access2rohit Were you able to benchmark this PR compared to yours?

PR #16234, opened 2019-09-22
Fix the wrong result of sum(inf, inf) and mean(inf, inf)

wkcn:
## Description ##
Hi, there.
I fix the wrong result of sum(inf, inf) and mean(inf, inf).

Test Case:
```python
import mxnet as mx
import numpy as np

x = mx.nd.array([np.inf, np.inf])
print(mx.nd.mean(x))
print(mx.nd.sum(x))

x = mx.nd.array([-np.inf, -np.inf])
print(mx.nd.mean(x))
print(mx.nd.sum(x))

x = mx.nd.array([np.inf, -np.inf])
print(mx.nd.mean(x))
print(mx.nd.sum(x))
```
The result should be `inf` rather than `nan`.

### Essentials ###

### Changes ###
- [x] Add `isnan_typed` and `isinf_typed` in mshadow
- [x] Remove `isnan_typed` in `src/operator/mshadow_op.h`
- [x] Add relative testcases

## Comments ##
The results of `argmin` and `argmax` are inconsistent with that of NumPy. For example,
```
np.array([0, np.nan]).argmin() = 1
mx.nd.array([0, np.nan]).argmin() = 0
```
It need to be fixed later.


3rdparty/mshadow/mshadow/base.h
```diff
@@ -791,12 +863,9 @@ struct maximum {
...
   template<typename DType>
   MSHADOW_XINLINE static void Reduce(volatile DType& dst,  volatile DType src) { // NOLINT(*)
-    using namespace std;
-#ifdef __CUDACC__
-    dst = ::max(dst, src);
-#else
-    dst = max(dst, src);
-#endif  // __CUDACC__
+    if (!isnan_typed::IsNan(dst)) {
+      if (!(dst > src)) dst = src;
```
wkcn: I'm not sure whether the compiler will optimize it into `dst <= src`.

3rdparty/mshadow/mshadow/base.h
```diff
@@ -791,12 +863,9 @@ struct maximum {
...
   template<typename DType>
   MSHADOW_XINLINE static void Reduce(volatile DType& dst,  volatile DType src) { // NOLINT(*)
-    using namespace std;
-#ifdef __CUDACC__
-    dst = ::max(dst, src);
-#else
-    dst = max(dst, src);
-#endif  // __CUDACC__
+    if (!isnan_typed::IsNan(dst)) {
+      if (!(dst >= src)) dst = src;
```
wkcn: I'm not sure whether the compiler will optimize it into `dst < src`.

julia/test/unittest/ndarray.jl
```diff
@@ -1525,8 +1525,8 @@ function test_argmax()
          NaN 2 6]
     x = NDArray(A)
 
-    @test copy(argmax(x, dims = 1)) == [1 1 2]
-    @test copy(argmax(x, dims = 2)) == reshape([2, 3], :, 1)
+    @test copy(argmax(x, dims = 1)) == [2 1 2]
```
wkcn: The result is consistent with Julia built-in function `argmin` and `argmax`

iblis17: Please consider update the docstring of `argmax`/`argmin` in `julia/src/ndarray/reduction.jl` as well.

wkcn: I have updated it : )

3rdparty/mshadow/mshadow/extension/reduce_with_axis.h
```diff
@@ -112,7 +112,7 @@ struct Plan<ReduceWithAxisExp<Reducer, SrcExp, DType, dimsrc, mask, dimdst>, DTy
         index_t z = (x*size_+k)*trailing_+y;
         DType tmp = res;
         Reducer::Reduce(res, src_.Eval(z/last_, z%last_));
-        if (tmp != res) {
+        if (tmp != res && !isnan_typed::IsNan(tmp)) {
```
wkcn: `NaN` is treated as greater/less than all other values in argmax/ argmin.
NaN is not equal to any number including NaN.

wkcn:
Hi @marcoabreu @access2rohit , could you please help take a review?
The PR fixes the wrong result of `sum`, `mean`, `argmin` and `argmax` when inputs contain `inf` or `NaN`.
Thank you!

julia/test/unittest/ndarray.jl
```diff
@@ -1547,8 +1547,8 @@ function test_argmin()
          NaN 2 6]
     x = NDArray(A)
 
-    @test copy(argmin(x, dims = 1)) == [1 2 1]
-    @test copy(argmin(x, dims = 2)) == reshape([1, 2], :, 1)
+    @test copy(argmin(x, dims = 1)) == [2 2 1]
```
access2rohit: @wkcn as per numpy it should be [1, 2, 1] its argmin. right ?

julia/test/unittest/ndarray.jl
```diff
@@ -1547,8 +1547,8 @@ function test_argmin()
...
+    @test copy(argmin(x, dims = 2)) == reshape([1, 1], :, 1)
```
access2rohit: shouldn't this be [1,2]
previous argmin tests are fine w.r.t numpy

julia/test/unittest/ndarray.jl
```diff
@@ -1525,8 +1525,8 @@ function test_argmax()
          NaN 2 6]
     x = NDArray(A)
 
-    @test copy(argmax(x, dims = 1)) == [1 1 2]
-    @test copy(argmax(x, dims = 2)) == reshape([2, 3], :, 1)
+    @test copy(argmax(x, dims = 1)) == [2 1 2]
+    @test copy(argmax(x, dims = 2)) == reshape([2, 1], :, 1)
```
access2rohit: both correct ! :)

tests/python/unittest/test_operator.py
```diff
@@ -9286,21 +9286,26 @@ def test_sample_normal_default_shape():
     assert s.shape == (1, 1)
 
 
-def test_min_max_inf():
-    dtypes = [np.float32, np.double]
-    elem_list = [-1, 1, 0, np.inf, -np.inf]
-
+def test_inf_and_nan():
```
access2rohit: Why is test operator ? You can actually add it in test_ndarray.py. Actually you can add the following code 
```
allow_nan = np.random.randint(0, 2)
if allow_nan:
    total_nans = np.random.randint(0, dat.size//10+1)
    dat.ravel()[np.random.choice(dat.size, total_nans, replace=False)] = np.nan
```
after this line: https://github.com/apache/incubator-mxnet/blob/72b4d9b8261e0783989447ad78a09e8573aee853/tests/python/unittest/test_ndarray.py#L598
This will randomly decide whether to add Nans to input or not and put nans (roughly 10% of the size of input) inside the input randomly

3rdparty/mshadow/mshadow/extension/reduce_with_axis.h
access2rohit: In C++ i found using gdb that Nan are treated as largest. You probably don't need to check that. But as per C++ documentation the behavior is not defined so its good to keep that check !

3rdparty/mshadow/mshadow/base.h
```diff
@@ -600,6 +600,64 @@ struct divto {
...
+  template<>
+  MSHADOW_XINLINE bool IsInf(volatile long double val) {
+    return isinf(val);
+  }
+  template<>
+  MSHADOW_XINLINE bool IsInf(volatile mshadow::half::half_t val) {
+    return (val.half_ & 0x7fff) == 0x7c00;
+  }
+}  // namespace isinf_typed
+
```
access2rohit: Good !

3rdparty/mshadow/mshadow/extension/reduce_with_axis.h
wkcn: It is necessary to check NaN. For example, argmin/argmax should return the first index in the array [NaN, NaN], however (NaN != NaN) is true.

julia/test/unittest/ndarray.jl
wkcn: In NumPy, if the tensor contains NaN, it always return the index of NaN.

wkcn: import numpy as np
a = np.array([[1,5,3], [np.nan, 2, 6]])
print(a.argmin(0)) # 1, 1, 0
print(a.argmin(1)) # 0, 0

It always return the index of NaN if NaN exists.

tests/python/unittest/test_operator.py
wkcn: Thank you! I will move the testcase into test_ndarray.py

julia/test/unittest/ndarray.jl
iblis17:
```julia
julia> A
2×3 Array{Float64,2}:
   1.0  5.0  3.0
 NaN    2.0  6.0

julia> argmin(A, dims = 1)
1×3 Array{CartesianIndex{2},2}:
 CartesianIndex(2, 1)  CartesianIndex(2, 2)  CartesianIndex(1, 3)
```
note that Julia is 1-based index

wkcn: Yes. argmin/argmax in Julia are consistent with that in NumPy.

julia/test/unittest/ndarray.jl
wkcn: I have updated the testcase.

tests/python/unittest/test_operator.py
wkcn: Hi @access2rohit , I have moved the testcase into test_ndarray.py.

3rdparty/mshadow/mshadow/base.h
wkcn: Thx : )

julia/test/unittest/ndarray.jl
```diff
@@ -1515,8 +1515,8 @@ function test_argmax()
          4 2 6]
     x = NDArray(A)
 
-    @test copy(argmax(x, dims = 1)) == [2 1 2]
-    @test copy(argmax(x, dims = 2)) == reshape([2, 3], :, 1)
+    @test copy(argmax(x, dims = 1)) == [x[1] for x in argmax(A, dims = 1)]
+    @test copy(argmax(x, dims = 2)) == [x[2] for x in argmax(A, dims = 2)]
```
iblis17: This would be great.
```suggestion
    @test copy(argmax(x, dims = 2)) == [x[2] for x ∈ argmax(A, dims = 2)]
```

Just type `\in<tab>` in Julia REPL

wkcn: I have updated it. Thank you : )

wkcn:
Hi @reminisce and @haojin2 , could you please help take a review?

This PR makes the following functions consistent with NumPy.
```
sum, mean, argmin, argmax
```

Thank you!

wkcn:
Hi @eric-haibin-lin , could you please help take a review?
It is a bug, which outputs a wrong result or an inconsistent result with that of NumPy.

eric-haibin-lin:
Would you mind also add what is the result before this fix?

wkcn:
Hi @eric-haibin-lin , I have updated the test result : )

szha:
cc @reminisce

wkcn:
Ping : )

3rdparty/mshadow/mshadow/base.h
```diff
@@ -606,6 +606,64 @@ struct divto {
...
+  MSHADOW_XINLINE bool IsNan(volatile double val) {
+    return isnan(val);
+  }
+  template<>
+  MSHADOW_XINLINE bool IsNan(volatile long double val) {
+    return isnan(val);
+  }
+  template<>
+  MSHADOW_XINLINE bool IsNan(volatile mshadow::half::half_t val) {
+    return (val.half_ & 0x7fff) > 0x7c00;
```
marcoabreu: Can you turn these magic values into constants with documentation? While I get 0x7ffff, 0x7c00 for example, looks quite arbitrary.

tests/python/unittest/test_ndarray.py
```diff
@@ -601,16 +629,22 @@ def test_reduce_inner(numpy_reduce_func, nd_reduce_func, multi_axes):
...
+            mx_arr = mx.nd.array(dat, dtype=dtype)
+            ndarray_ret = nd_reduce_func(mx_arr, axis=axes, keepdims=keepdims)
             if type(ndarray_ret) is mx.ndarray.NDArray:
                 ndarray_ret = ndarray_ret.asnumpy()
             assert (ndarray_ret.shape == numpy_ret.shape) or \
                    (ndarray_ret.shape == (1,) and numpy_ret.shape == ()), "nd:%s, numpy:%s" \
                                                          %(ndarray_ret.shape, numpy_ret.shape)
-            err = np.square(ndarray_ret - numpy_ret).mean()
-            assert err < 1E-4
+            if check_dtype:
```
marcoabreu: Could you elaborate why you're introducing so much branching into a test? If the results are inconsistent, we should rather improve the test instead of skipping the checks. I'd love to have more detail

marcoabreu:
I'll merge after the feedback has been addressed :)

3rdparty/mshadow/mshadow/base.h
wkcn: Hi @marcoabreu , I add two constants `MSHADOW_HALF_SIGN_BIT` and `MSHADOW_HALF_EXPONENT_BITS` in [3rdparty/mshadow/mshadow/half.h](https://github.com/apache/incubator-mxnet/pull/16234/commits/9183cb3a2284a7d920bb720a19c38dc49d66fdea#diff-0d54cfcb67beaf0744f1efe71b45bfdfR352), and replace these two magic values.

tests/python/unittest/test_ndarray.py
wkcn: Hi @marcoabreu , here is the explanation.
1. So much branching
We need to test all reduce operators, like `min, max, argmin, argmax, sum, mean` when the inputs contain `-inf, +inf, nan`.

2. Skipping the checks
I replace the old check with [a new one](https://github.com/apache/incubator-mxnet/pull/16234/files#diff-69757562d07268150de8b369ff5b6b61R642).

PR #16264, opened 2019-09-25
Numpy operators: `lcm`, `tril`, `identity` and `take`

hgt312:
Each commit is an opeator.


reminisce:
The ops have been reviewed separately in each PR. Merging now.

Issue #13247, opened 2018-11-13
Support None as input in forward and hybrid_forward?


yifeim:
## Description
(Hybrid) forward supports `None` as defaults, but not as inputs. I was wondering if there are any fundamental reasons why `None` cannot be taken as input - if the pattern is consistent and I do not change the computation graph? Similarly, `None` in the output should also be allowed.

## Environment info (Required)

```
----------Python Info----------
Version      : 3.6.5
Compiler     : GCC 7.2.0
Build        : ('default', 'Apr 29 2018 16:14:56')
Arch         : ('64bit', '')
------------Pip Info-----------
Version      : 10.0.1
Directory    : /home/ec2-user/anaconda3/envs/mxnet_p36/lib/python3.6/site-packages/pip
----------MXNet Info-----------
Version      : 1.3.0
Directory    : /home/ec2-user/anaconda3/envs/mxnet_p36/lib/python3.6/site-packages/mxnet
Commit Hash   : b3be92f4a48bce62a5a8424271871c2f81c8f7f1
----------System Info----------
Platform     : Linux-4.14.77-69.57.amzn1.x86_64-x86_64-with-glibc2.9
system       : Linux
node         : ip-172-16-72-155
release      : 4.14.77-69.57.amzn1.x86_64
version      : #1 SMP Tue Nov 6 21:32:55 UTC 2018
----------Hardware Info----------
machine      : x86_64
processor    : x86_64
----------Network Test----------
Setting timeout: 10
Timing for MXNet: https://github.com/apache/incubator-mxnet, DNS: 0.0016 sec, LOAD: 0.4079 sec.
Timing for Gluon Tutorial(en): http://gluon.mxnet.io, DNS: 0.1175 sec, LOAD: 0.1057 sec.
Timing for Gluon Tutorial(cn): https://zh.gluon.ai, DNS: 0.2767 sec, LOAD: 0.1359 sec.
Timing for FashionMNIST: https://apache-mxnet.s3-accelerate.dualstack.amazonaws.com/gluon/dataset/fashion-mnist/train-labels-idx1-ubyte.gz, DNS: 0.0274 sec, LOAD: 0.1023 sec.
Timing for PYPI: https://pypi.python.org/pypi/pip, DNS: 0.0023 sec, LOAD: 0.3404 sec.
Timing for Conda: https://repo.continuum.io/pkgs/free/, DNS: 0.0081 sec, LOAD: 0.3184 sec.
```

[...]

PR #16280, opened 2019-09-25
[Gluon] Support None argument in HybridBlock

sxjscience:
## Description ##
Solves https://github.com/apache/incubator-mxnet/issues/13247

We will later consider to support dict, OrderedDict, namedtuple and customized python classes in Gluon 2.0 @leezu @junrushao1994 

Support the following way of writing the hybrid_forward:

```python
class Foo(HybridBlock):
    def hybrid_forward(self, F, a, b):
        if a is None and b is not None:
            return b
        elif b is None and a is not None:
            return a
        elif a is not None and b is not None:
            return a + b
        else:
            raise NotImplementedError
```

### Essentials ###

### Changes ###
- [x] HybridBlock argument as None, tests

### DIFF
--- a/tests/python/unittest/test_gluon.py
+++ b/tests/python/unittest/test_gluon.py
@@ -454,6 +454,18 @@ def test_hybrid_block_none_args():
             else:
                 raise NotImplementedError
 
+    class FooDefault(gluon.HybridBlock):
+        def hybrid_forward(self, F, a, b=None):
+            if a is None and b is not None:
+                return b
+            elif b is None and a is not None:
+                return a
+            elif a is not None and b is not None:
+                return a + b
+            else:
+                raise NotImplementedError
+
+
     class FooNested(gluon.HybridBlock):
         def __init__(self, prefix=None, params=None):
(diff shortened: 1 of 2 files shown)


tests/python/unittest/test_gluon.py
```diff
@@ -441,6 +441,63 @@ def test_sparse_hybrid_block():
...
+            data = self.f3(data, b)
+            return data
+
+    for arg_inputs in [(None, mx.nd.ones((10,))),
+                       (mx.nd.ones((10,)), mx.nd.ones((10,))),
+                       (mx.nd.ones((10,)), None)]:
+        foo1 = FooNested(prefix='foo_nested_hybridized')
+        foo1.hybridize()
+        foo2 = FooNested(prefix='foo_nested_nohybrid')
+        for _ in range(2):
```
wkcn: It will be better to annotate that the reason of the two loops is to test hybridized cache : )

sxjscience: Yes, it triggers the part of the code when the operators are cached.

szha:
@eric-haibin-lin FYI on the merge.

PR #16335, opened 2019-09-30
[Numpy] Global numpy shape flag

stu1130:
## Description ##

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/include/mxnet/c_api.h
+++ b/include/mxnet/c_api.h
@@ -1173,7 +1173,8 @@ MXNET_DLL int MXAutogradIsTraining(bool* curr);
 MXNET_DLL int MXIsNumpyShape(bool* curr);
 /*!
  * \brief set numpy compatibility switch
- * \param is_np_shape 1 when numpy shape semantics is on, 0 when off
+ * \param is_np_shape 1 when numpy shape semantics is thread local on,
+ *        2 when numpy shape semantics is global on and 0 when off
  * \param prev returns the previous status before this set
  * \return 0 when success, -1 when failure happens
  */
--- a/include/mxnet/imperative.h
+++ b/include/mxnet/imperative.h
@@ -35,6 +35,17 @@
 #include "./ndarray.h"
 
 namespace mxnet {
+  /*! \brief there are three numpy shape flags based on priority.
+   * GlobalOn
+   *   turn on numpy shape flag globally, it includes thread local.
+   *   The flag can be seen in any thread.
+   * ThreadLocalOn
+   *   only turn on thread local numpy shape flag, it cannot be seen
+   *   in other threads.
+   * Off
+   *   turn off numpy shape flag globally.
+   * */
+  enum NumpyShape{Off, ThreadLocalOn, GlobalOn};
 /*! \brief runtime functions for NDArray */
 class Imperative {
  public:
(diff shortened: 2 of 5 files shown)


stu1130:
will reopen once test is added

stu1130:
@reminisce @eric-haibin-lin I added the test

ci/docker/runtime_functions.sh
```diff
@@ -1532,6 +1532,13 @@ nightly_estimator() {
     nosetests test_sentiment_rnn.py
 }
 
+nightly_global_numpy_flag() {
```
marcoabreu: Is it necessary to define a new run? Can you make it part of the existing nightly test suite instead?

stu1130: I put it into estimator nightly test

marcoabreu: Ok, so then there's no need to add this here, right? Or maybe you want to elaborate your intention :)

stu1130: Although putting in estimator together is not ideal, as long as the test is run alone, it's fine

stu1130:
@marcoabreu could you review again Thanks

stu1130:
Hey @marcoabreu we really need this feature. Please review it. Thanks

ci/docker/runtime_functions.sh
```diff
@@ -1524,12 +1524,15 @@ nightly_scala_demo_test_cpu() {
     bash bin/run_im.sh
 }
 
-nightly_estimator() {
+nightly_estimator_numpy_global_flag() {
```
marcoabreu: Sorry I'm not following. Why is this even a nightly test?

Also, I don't think that this will work. When you change the name of the function but not the point where it gets referenced (IF it gets referenced), then it would break, right? But lets forget about that.

Why don't you make this a normal unit test? The test doesn't seem to fit into a constraint that warrants nightly execution.

stu1130: The feature is only going to be used in MXNet subproject (even numpy willl not use), so I think it make sense to put in nightly test.

I changed the tests/nightly/JenkinsfileForBinaries to reference this function. Do I still miss something?

The test should only be run alone, since it affect the behavior of numpy shape in other tests that is run parallelly. i.e. It will crash our operator unit test if I put this in noraml unit test

marcoabreu: Could you elaborate which subproject?

Oh right, you change the Jenkinsfile. Sorry about that!

We don't parallelize our tests, so it should be fine to just run it together with the others. The only thing you have to watch is to clean up properly.

marcoabreu: To me this seems like a normal feature - independent of where it's used, so I'd be in favour if we add a regular unit test.

stu1130: The project is not open source yet.

AFAIK, nosetest run tests in parallel. Currently numpy shape is claimed as thread_local variable, this feature enable global functionaility where we can see the flag from any thread. It means we might break non-numpy-compatible operator test that is run parallelly because MXNet engine treat them as numpy-compatible.

Although the payload is small, I don't want MXNet developers bothered by this test and it's actually more like an integration test.

marcoabreu: We run nosetests without parallelization :)

stu1130: ok I'll try to add this in regular unit test

PR #16342, opened 2019-09-30
adding redirects so that old website API links surfaced from searches

ThomasDelteil:
Reopened from @sad- PR


docs/static_site/src/.htaccess
```diff
@@ -1,4 +1,3 @@
-Options -Indexes
```
marcoabreu: Could you elaborate? I think we prefer running a 404 page rather than displaying the directory index

ThomasDelteil: @aaronmarkham did you test that 404 are still respected and we don't display the index when you removed that line?
@marcoabreu my understanding is that this no indexing would have been set at the virtual host level in the apache config already.

aaronmarkham: I removed that because infra said it wasn't supported and it was breaking on staging.
https://issues.apache.org/jira/browse/INFRA-19187

aaronmarkham: Once I removed that line this code would work on staging... with it, the code would result in a 500 error. I didn't test variations, but if you review the link and the docs link that they refer to for supported features, maybe you can turn on something else that will give you the desired effect.

marcoabreu: Ah I see, they restricted the usage to a certain set of commands. Fine to go ahead.

aaronmarkham:
This was already approved on soji's PR, so I'm merging now.

docs/static_site/src/.htaccess
ThomasDelteil: http://mxnet.incubator.apache.org/api/perl/docs/tutorials/ the option is available on prod, we need to ask Apache infra to align beta / prod configs

PR #16351, opened 2019-10-02
[IR-Bridge] Support attrs for operators: convolution, batch norm, relu

junrushao1994:
@tqchen @szha @sxjscience @eric-haibin-lin @leezu 

Ideally, we should stop using `Attrs`. But it may not be a good time now as the infra hasn't been done.

## Directory structure

PREFIX=mxnet/src/v3/

- Folder `$PREFIX/include/bridge`: headers for converters, like mxnet's nnvm (legacy_nnvm.h).
- Folder `$PREFIX/include/op`: operator definition, which is supposed to be independent from frontend.

### DIFF
--- a/src/v3/include/ir.h
+++ b/src/v3/include/ir.h
@@ -1,0 +1,188 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*!
+ * Copyright (c) 2019 by Contributors
+ * \file ir.h
+ * \author Junru Shao
+ */
+#pragma once
+#if MXNET_USE_TVM_OP && !defined MXNET_AMALGAMATION
+// This is a compatibility layer between MXNet v3 and Relay
+// We will borrow basically everything from TVM/Relay to here.
+
+#include <tvm/attrs.h>
+#include <tvm/ir.h>
+#include <tvm/runtime/c_runtime_api.h>
+#include <tvm/runtime/packed_func.h>
+#include <tvm/node/container.h>
+#include <tvm/node/memory.h>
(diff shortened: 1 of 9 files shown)


src/v3/src/bridge/legacy_nnvm/attrs.cc
```diff
@@ -0,0 +1,120 @@
...
+    };
+    const auto &nnvm_attrs =
+        nnvm::get<mxnet::op::ActivationParam>(attrs.parsed);
+    if (op_map.count(nnvm_attrs.act_type)) {
+      return CallNode::make(op_map[nnvm_attrs.act_type], args, {});
+    }
+  }
+  LOG(INFO) << "Warning: cannot recognize NNVM operator " << op->name
+            << ", fallback to add";
+  return CallNode::make(Op::Get("add"), args, {}, {});
```
szha: shall we log fatal here? 😆

src/v3/src/bridge/legacy_nnvm/ir.cc
```diff
@@ -58,57 +65,58 @@ static void PrintIndexedGraph(const nnvm::Graph &g) {
     std::string op_name = op ? op->name : "None";
     if (input_nodes.count(i)) {
       input_cnt += 1;
-      op_name = (op ? op->name + " [input " : "[input ") + std::to_string(input_cnt) + "]";
+      op_name = (op ? op->name + " [input " : "[input ") +
+                std::to_string(input_cnt) + "]";
     } else {
       op_name = op ? op->name : "None";
     }
     std::cout << "  i = " << i << ", op = " << op_name
```
szha: LOG?

src/v3/include/ir.h
```diff
@@ -0,0 +1,188 @@
...
+using tvm::relay::TypeReporter;
+
+// Relay Functors
+using tvm::relay::ExprFunctor;
+
+}  // namespace ir
+}  // namespace v3
+}  // namespace mxnet
+
+#define MNM_DEF_NODE_TYPE_INFO(TypeName, Parent) TVM_DECLARE_NODE_TYPE_INFO(TypeName, Parent)
```
szha: let's remember to revisit the name for MNM

junrushao1994: Hmmm good point, I will do another check

src/v3/src/bridge/legacy_nnvm/ir.cc
junrushao1994: This is piece of program that basically did visualization, and will be removed once the entire pipeline is done. So probably `cout` looks more casual and unlikely to forget :-)

src/v3/src/bridge/legacy_nnvm/attrs.cc
junrushao1994: Then many unittests would crash because the operator is unsupported ><

szha: We can disable the tests that are failing and re-enable them when we achieve better coverage.

junrushao1994: Per offline discussion with @szha, I am going to write a separate Jenkins file

PR #16355, opened 2019-10-02
Embedding gradient performance optimization on GPU

MoisesHer:
## Description ##
This PR includes a specific Embedding-backward operator for GPU.
Two new CUDA kernels have been implemented for improving the performance of the operator when using GPU.
According to our measurements on Volta GPUs, the previous version was taken 2.2ms,
whereas the new implementation takes 0.3ms, i.e. more than 7x speedup.
### Essentials ###
- Unit tests already existed (tests/python/gpu/test_operator_gpu:test_embedding_with_type) and changes do not affect correctness.
### Changes ###
- [x ] Embedding-backward operator for GPU, test: tests/python/gpu/test_operator_gpu:test_embedding_with_type

### DIFF
--- a/src/operator/tensor/indexing_op.cu
+++ b/src/operator/tensor/indexing_op.cu
@@ -569,6 +569,7 @@ __global__ void EmbeddingFindBounds(const IType *sorted_data,
                                     const index_t data_dim,
                                     const index_t vocab_dim) {
   const index_t id = blockIdx.x * blockDim.x + threadIdx.x;
+  if (id>=vocab_dim) return;
 
   // Binary search to find lower bound: stored at bounds[0..vocab_dim-1]
   IType lower_bound = 0;
@@ -582,9 +583,12 @@ __global__ void EmbeddingFindBounds(const IType *sorted_data,
       lower_bound = mean + 1;
   }
   bool found_row = (sorted_data[lower_bound] == id);
-
-  if (id < vocab_dim) {
-    bounds[id] = (found_row) ? lower_bound : -1;
+  if (!found_row) {
+    bounds[id] = -1;
+    bounds[vocab_dim + id] = -2;
+    return;
+  } else {
+    bounds[id] = lower_bound;
   }
 
   // Binary search to find upper bound: stored at bounds[vocab_dim..2*vocab_dim-1]
@@ -597,11 +601,7 @@ __global__ void EmbeddingFindBounds(const IType *sorted_data,
     else
       upper_bound = mean - 1;
   }
-  found_row = (sorted_data[upper_bound] == id);
-
-  if (id < vocab_dim) {
-    bounds[vocab_dim + id] = (found_row) ? upper_bound : -1;
-  }
+  bounds[vocab_dim + id] = upper_bound;
 }
 
 /*
@@ -612,7 +612,6 @@ __global__ void EmbeddingFindBounds(const IType *sorted_data,
  * \param grad_out output gradient data
  * \param embbedding_dim dimension of the dense embedding
  * \param vocab_dim maximum number of unique indices in the data array: tokens vocabulary size
- * \param rows_per_block number of grad_in rows to be computed by each block
  * \param req write/add/null
  */
 template <typename LType, typename DType, typename IType>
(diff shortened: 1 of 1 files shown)


src/operator/tensor/indexing_op.cu
```diff
@@ -545,6 +545,245 @@ void TakeOpForward<gpu>(const nnvm::NodeAttrs& attrs,
...
 }
 
+/*
+ * \brief returns integer log2(a) rounded up
+ */
+inline int ilog2(unsigned int a) {
+  int k = 1;
+  while (a >>= 1) k++;
+  return k;
+}
```
ptrendx: Please move this function to unnamed namespace.

ptrendx:
@sxjscience FYI

src/operator/tensor/indexing_op.cu
```diff
@@ -545,6 +545,247 @@ void TakeOpForward<gpu>(const nnvm::NodeAttrs& attrs,
...
+    mean = (lower_bound + upper_bound) / 2;
+    if (id <= sorted_data[mean])
+      upper_bound = mean;
+    else
+      lower_bound = mean + 1;
+  }
+  bool found_row = (sorted_data[lower_bound] == id);
+
+  if (id < vocab_dim) {
+    bounds[id] = (found_row) ? lower_bound : -1;
```
ptrendx: Maybe early exit? How long is this kernel?

src/operator/tensor/indexing_op.cu
```diff
@@ -545,6 +545,247 @@ void TakeOpForward<gpu>(const nnvm::NodeAttrs& attrs,
...
+  LType Lvalue[1];
+  DType* Dvalues = reinterpret_cast<DType*>(Lvalue);
+
+  for (index_t row=0; row < rows_per_block; ++row) {
+    IType my_row = blockIdx.x * rows_per_block + row;
+    if (my_row < vocab_dim) {
+      // Read lower and upper bounds for current row
+      IType lower_bound = index_bounds[my_row];
+      IType upper_bound = index_bounds[vocab_dim + my_row];
+      int nOccurrences = (lower_bound != -1) ? (upper_bound - lower_bound + 1) : 0;
```
ptrendx: Super small hack and probably not worth it, but if you set `upper_bound` to -2 if you did not find the lower bound, then you would not need this conditional here.

src/operator/tensor/indexing_op.cu
```diff
@@ -545,6 +545,247 @@ void TakeOpForward<gpu>(const nnvm::NodeAttrs& attrs,
...
+
+      for (index_t emb_id=threadIdx.x; emb_id < aligned_emb_dim; emb_id += blockDim.x) {
+        // Initialize grad_in
+        if (req == kAddTo) {
+          grad_in_row[threadIdx.x] = aligned_grad_in[my_row * aligned_emb_dim + emb_id];
+        } else {
+          grad_in_row[threadIdx.x] = 0.0;
+        }
+        // Add all rows from grad_out according to indices in data
+        if (nOccurrences) {
```
ptrendx: Does this help? It should be already handled the same way by the for loop.

src/operator/tensor/indexing_op.cu
```diff
@@ -545,6 +545,247 @@ void TakeOpForward<gpu>(const nnvm::NodeAttrs& attrs,
...
+  // Find lower & upper bounds of each possible index
+  const int threads_block_bounds = 128;
+  const int nblocks_bounds = (vocab_dim + threads_block_bounds - 1) / threads_block_bounds;
+  EmbeddingFindBounds<<<nblocks_bounds, threads_block_bounds, 0, Stream<gpu>::GetStream(s)>>>(
+                  sorted_data.dptr_, bounds_index.dptr_, data_dim, vocab_dim);
+
+  // Compute Gradient
+  int ltype = mxnet::common::cuda::get_load_type(embbedding_dim * sizeof(DType));
+  MXNET_LOAD_TYPE_SWITCH(ltype, LType, {
+    const int rows_per_block = 1;
```
ptrendx: Do you support `rows_per_block != 1` in the kernel? If not, then it should be removed.

sxjscience:
Nice! LGTM. So the BinarySearch version of FindBounds has complexity O(|V| log |N|) where |V| is the vocabulary size and |N| is the number of indices. I guess our initial version has complexity O(|N|) for finding the boundaries. Thus, in some workloads (in which |N| is small), the O(N) version might be faster.

ptrendx:
Moises did a performance comparison between the new version and both the old one and the old buggy one. The new kernel is faster than the old working version in all cases and ~same in speed as the buggy one. The biggest performance change is seen actually when changing how many different elements are in the input data (as small number of distinct elements limits parallelism in the backward pass).

PR #16384, opened 2019-10-06
[DOC] Change mailing list url in footer to point to instructions about how to subscribe

marcoabreu:
If users just send an email to the email list without being subscribed, it will just be swallowed.

### DIFF
--- a/docs/static_site/src/_includes/footer.html
+++ b/docs/static_site/src/_includes/footer.html
@@ -4,8 +4,7 @@
             <div class="col-4">
                 <h4 class="footer-category-title">Resources</h4>
                 <ul class="contact-list">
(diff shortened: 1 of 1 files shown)


marcoabreu:
I think these links expire after some time. As far I can recall there's an Apache infra managed link we can use.

szha:
@marcoabreu the register link can be set to expire. The link to channels usually don't expire. For example, GluonNLP's slack link has been alive for about a year now. http://gluon-nlp.mxnet.io/community/index.html

marcoabreu:
Erm the link to me shows that is has expired 😄 . But whatever, we can use http://s.apache.org/slack-invite (from https://cwiki.apache.org/confluence/display/INFRA/Slack+Guest+Invites)

PR #16391, opened 2019-10-08
cuDNN non-persistant bidirectional RNN dgrad sync fix

DickJC123:
## Description ##
Background: A non-deterministic failure of test_operator_gpu.py:test_lstm_bidirectional that was observed just once in our own CI was found to be due to cuDNN's RNN dgrad implementation.  cuDNN launches many kernels as part of the RNN dgrad into various auxiliary streams that are different than the primary user-settable stream of the cuDNN handle.  A final kernel launched by cuDNN into one of these aux streams was not being synchronized (via events) back to the handle's stream.  When MXNet's RNN::Backward() returns, a gradient summation kernel can be launched by the GPU worker into its main stream and this kernel's execution could potentially overlap or even precede that of the final cuDNN RNN dgrad kernel.  MXNet's calling of cuDNN's wgrad immediately after dgrad makes this data-race failure exceedingly rare, and in fact we discovered the problem by code inspection, not by being able to reproduce the original CI failure.

The first commit of this PR will demonstrate the failure.  The approach is to have MXNet now skip the RNN wgrad operation if grad_req = {'data':'add', 'parameters':'none'}, plus expand the test_lstm_bidirectional test to invoke this case.  Skipping wgrad aggravates the data race by allowing MXNet to launch the gradient summation kernel sooner after the RNN dgrad kernels are launched with no intervening wgrad gpu activity.

Once the failure is solidly demonstrated, a follow-up commit will supply the fix.  The fix will use cuda events directed at the legacy default stream to ensure all dgrad GPU activity is complete before the wgrad or other kernels begin.  Unlike a cudaDeviceSynchronize(), the fix will not block the CPU.  nvprof-based timing analysis of the fix shows no measurable difference in timing for the single-RNN case analyzed.

The fix is needed for all versions of cuDNN supported by MXNet (so up to the current v7.6.4) and is needed only for non-persistent bidirectional RNNs.

### Essentials ###

### Changes ###

## Comments ##


src/operator/rnn-inl.h
```diff
@@ -1484,6 +1493,20 @@ class RNNOp {
...
+
+#if MXNET_USE_CUDNN == 1 && defined(__CUDACC__)
+  // cuDNN versions up to and including v7.6.4 did not sync a last dgrad kernel back to the main
+  // cudnn handle's stream (non-persistant algo, bidirectional only).  This could result in silent
+  // non-determinstic failures with very low probability, seen more often when wgrad is bypassed.
+  inline void SyncDgrad() {
+    if (CUDNN_VERSION <= 7604 && dgrad_sync_needed_) {
+      // Without blocking the CPU, create a synchronization point of all current GPU activity.
+      CUDA_CALL(cudaEventRecord(dgrad_sync_event_, cudaStreamLegacy));
+      CUDA_CALL(cudaStreamWaitEvent(cudaStreamLegacy, dgrad_sync_event_, 0));
```
ptrendx: You don't need to wait on this event, just recording is fine.

tests/python/unittest/test_operator.py
```diff
@@ -149,6 +149,9 @@ def test_lstm_bidirectional():
     check_rnn_consistency(fused, stack, T, N, I, H, 'write')
     check_rnn_consistency(fused, stack, T, N, I, H, 'add')
     check_rnn_consistency(fused, stack, T, N, I, H, 'null')
+    # Repeated to expose low-probability failure in cuDNN v7.6.4 and earlier
```
ptrendx: You can probably take it out now, right?

src/operator/rnn-inl.h
DickJC123: I've simplified this as you directed, but put in a comment to say that the wait is not needed in this case.

DickJC123:
I've rebased the commits of the PR to latest master in an attempt to avoid pylint failures, as suggested by @reminisce .  If this succeeds, does this suggest a bug in the way the CI creates the repo-under-test?

src/operator/rnn-inl.h
```diff
@@ -1484,6 +1493,20 @@ class RNNOp {
...
+
+#if MXNET_USE_CUDNN == 1 && defined(__CUDACC__)
+  // cuDNN versions up to and including v7.6.4 did not sync a last dgrad kernel back to the main
+  // cudnn handle's stream (non-persistant algo, bidirectional only).  This could result in silent
+  // non-determinstic failures with very low probability, seen more often when wgrad is bypassed.
+  inline void SyncDgrad() {
+    if (CUDNN_VERSION <= 7604 && dgrad_sync_needed_) {
+      // Without blocking the CPU, create a synchronization point of all current GPU activity.  No
+      // need to call cudaStreamWaitEvent- cudaEventRecord on the legacy default stream suffices.
+      CUDA_CALL(cudaEventRecord(dgrad_sync_event_, cudaStreamLegacy));
```
haojin2: Hi @DickJC123, I'm encountering `cudaErrorInvalidResourceHandle` error here when I'm trying to run this (notebook)[https://github.com/d2l-ai/d2l-en/blob/numpy2/chapter_natural-language-processing/sentiment-analysis-rnn.md] in dive into deep learning textbook. Could you help with a fix to that?

PR #16377, opened 2019-10-05
[WIP] Fixing the API organization

ThomasDelteil:
@sojiadeshina @aaronmarkham 

@mseeger @asmushetzel @eric-haibin-lin 

This PR simplify the organization of the documentation. We now have the best of both world, automatically generated package level documentation with full coverage, with the ability to have semantic organization at the top. We added back a JS generated API reference on the left for each module.

Screen Shot 2019-10-04 at 8 14 36 PM

PR #16392, opened 2019-10-08
[WIP] Improving Python Docs API

sojiadeshina:
## Description ##

Reorganizes python API docs in the new website. We now have coverage on all the mxnet python submodules, building on #16377

@ThomasDelteil @aaronmarkham @eric-haibin-lin 

Changed the API front page to provide quick links for frequently used packages, this is also reflected on the top level of the toc-tree in the left hand side.

Screen Shot 2019-10-07 at 5 55 23 PM

There's also a new page that allows you to easily navigate to all the submodules.

Screen Shot 2019-10-07 at 5 44 46 PM


docs/python_docs/_static/autodoc.js
```diff
@@ -0,0 +1,49 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/* Customizations to the Sphinx auto module plugin output */
+function auto_index() {
+    var targets = $("dl.class>dt,dl.function>dt");
```
aaronmarkham: Can this function be described?
We ran into issues with the last site because the Sphinx customizations and js were undocumented.

docs/python_docs/environment.yml
```diff
@@ -36,3 +36,4 @@ dependencies:
   - breathe
   - mock
   - awscli
+  - autodocsumm
```
aaronmarkham: Is that from here? https://github.com/Chilipp/autodocsumm

docs/python_docs/python/Makefile
```diff
@@ -47,7 +46,7 @@ build/%: %
 html: $(OBJ)
 	mkdir -p build
 	cp Makefile_sphinx build/Makefile
-	sphinx-autogen build/api/*.rst build/api/*/*.rst   -t build/_templates/
+	sphinx-autogen build/api/*.rst build/api/**/*.rst   -t build/_templates/
```
aaronmarkham: Is that `**` some recursive notation for file paths?

docs/python_docs/python/api/gluon/model_zoo/index.rst
```diff
@@ -15,21 +15,20 @@
    specific language governing permissions and limitations
    under the License.
 
-model_zoo.vision
-=================
+gluon.model_zoo.vision
+======================
 
 .. automodule:: mxnet.gluon.model_zoo.vision
+   :noindex:
```
aaronmarkham: What is the reason on when you use this approach:
```
.. automodule:: mxnet.gluon.model_zoo.vision
   :noindex:
``` 
...versus this:
```
.. automodule:: mxnet.gluon.loss
    :members:
    :imported-members:
    :autosummary:
```

docs/python_docs/python/api/gluon/contrib/index.rst
```diff
@@ -146,27 +147,21 @@ API Reference
 
 .. automodule:: mxnet.gluon.contrib
     :members:
-    :imported-members:
```
aaronmarkham: When do you decide to not include `:imported-members:`?

docs/python_docs/python/api/gluon/nn/index.rst
```diff
@@ -36,46 +36,24 @@ We group all layers in these two modules according to their categories.
...
-   nn.SymbolBlock
 
 
 Sequential containers
 ---------------------
 
 
 .. autosummary::
-    :toctree: _autogen
     :nosignatures:
```
aaronmarkham: This was questioned in another issue. What are some set with `:nosignatures:` and others are not?

docs/python_docs/python/api/initializer/index.rst
```diff
@@ -0,0 +1,23 @@
...
+   "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+   KIND, either express or implied.  See the License for the
+   specific language governing permissions and limitations
+   under the License.
+
+mxnet.initializer
+=================
+
+.. automodule:: mxnet.initializer
+    :members:
```
aaronmarkham: Why no inherited-members?

docs/python_docs/environment.yml
sojiadeshina: yeah that's where it's from.

docs/python_docs/python/api/gluon/contrib/index.rst
sojiadeshina: I take out imported members when the package level module also incidentally imports stuff from elsewhere that's already documented. For example here I think this also imports mxnet.ndarray which leads to a ton of warnings during the sphinx build.

docs/python_docs/python/api/gluon/model_zoo/index.rst
sojiadeshina: The difference is this.

For model_zoo.vision, I wanted to keep the documentation organized by the different model architectures: resnet, vgg etcs. it would have been super cluttered if it was just done all automatically. There are a ton of pretrained models in the model zoo so it's worth it to have them organized by sections.

The second approach in mxnet.gluon.loss automatically generates the documentation for all classes in the package and :autosummary: allows you to create a summary table at top automatically. Since there are not that many loss functions the autogenerated table summary is sufficient for this class.

docs/python_docs/python/api/gluon/nn/index.rst
sojiadeshina: This is an aesthetic/readability choice and the idea here is similar to the comment in the other issue. Here the doc is listing different classes under the different groupings so there's no need to have the method signatures.

At the bottom when autosummary generates an alphabetical listing of the classes, the method signatures are included.

docs/python_docs/python/api/initializer/index.rst
sojiadeshina: do you mean :imported-members:? if so then it's the same as : https://github.com/apache/incubator-mxnet/pull/16392#discussion_r332820265

if you meant :inherited-members: I think that only works for .. autoclass::

docs/python_docs/python/Makefile
sojiadeshina: That's the idea but actually I think the effect is the same with a single `*` .

docs/python_docs/_static/autodoc.js
```diff
@@ -0,0 +1,49 @@
...
+            }
+        }
+        html += "</ul>";
+        li_node.append(html);
+        li_node.prepend("<a><span id='autodoc_toggle' onclick='$(\"#autodoc\").toggle()'>[toggle]</span></a>")
+    } else {
+        setTimeout(auto_index, 500);
+    }
+}
+$(document).ready(auto_index);
```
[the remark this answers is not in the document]
sojiadeshina: Linter didn't yell at me for this so I guess it's fine

docs/python_docs/_static/autodoc.js
sojiadeshina: Sure.
Essentially what's going on is it's going through all the classes/methods on the page and truncating the name to just the last two parts i.e from something like `mxnet.gluon.nn.contrib` to just `nn.contrib` and then adding that to a list that can be toggled so that the sidebar has information and anchors to specific methods on the page even when the sphinx toc-tree doesn't.

 @ThomasDelteil wrote this part so he can correct me if i missed anything or give a more detailed description of what's going on.

docs/python_docs/python/api/gluon/contrib/index.rst
aaronmarkham: > I take out imported members when the package level module also incidentally imports stuff from elsewhere that's already documented. For example here I think this also imports mxnet.ndarray which leads to a ton of warnings during the sphinx build.

Ok that's good then, because we need to get warnings back down to zero.

docs/python_docs/python/api/initializer/index.rst
aaronmarkham: > do you mean :imported-members:? if so then it's the same as : [#16392 (comment)](https://github.com/apache/incubator-mxnet/pull/16392#discussion_r332820265)
> 
> if you meant :inherited-members: I think that only works for .. autoclass::

I noticed that sometimes it's `imported-members` and sometimes it's `inherited-members`. I would like to better understand the distinction and when one should be used, both should be used, none should be used.

aaronmarkham:
@sojiadeshina Once this PR goes in you can publish to staging. Plus, I'm having the same sanity test issue even after rebasing, so you might want track that. https://github.com/apache/incubator-mxnet/pull/16411

ThomasDelteil:
@ChaiBapchya please restart CI

ChaiBapchya:
Retriggered the windows-gpu, windows-cpu

ChaiBapchya:
@ThomasDelteil 
@aaronmarkham @eric-haibin-lin Ready to review/merge

PR #16398, opened 2019-10-08
Aggregated adamw update

drivanov:
## Description ##
MxNet operator for aggregated Adam update

### Essentials ###

### Changes ###
- [x] New operator allows to make Adam update for multiple gradients in one kernel.
tests, (and when applicable, API doc)
- [x] Test, for previously used "single" Adam update modified and now it also includes the use of `clip_gradient` parameter and random variations for `lr`, `eta`, `wd` and `shape`.


ptrendx:
@eric-haibin-lin FYI

python/mxnet/ndarray/contrib.py
```diff
@@ -542,14 +545,17 @@ def isnan(data):
     [1. 0.]
     <NDArray 2 @cpu(0)>
     """
-    return data != data # pylint: disable=comparison-with-itself
+    return data != data  # pylint: disable=comparison-with-itself
 
-def adamw_update(weight, grad, mean, var, rescale_grad, lr, eta, beta1=0.9, beta2=0.999,
-                 epsilon=1e-8, wd=0, clip_gradient=-1, out=None, name=None, **kwargs):
+def getRescaleGrad(rescale_grad, ctx):
```
ptrendx: Maybe check whether rescale_gradis already on the proper context and skip the copy if it is?

src/operator/contrib/adamw-inl.h
```diff
@@ -151,7 +135,7 @@ struct MPAdamWUpdate {
                const std::vector<TBlob> &outputs,
                const float rescale_grad) {
     using namespace mxnet_op;
-    AdamWParam param = nnvm::get<AdamWParam>(attrs.parsed);
+    const auto &param = nnvm::get<AdamWParam>(attrs.parsed);
```
ptrendx: Put the `&` next to `auto` (so `const auto& param` instead of `const auto &param`) to be consistent with the rest of the code.

src/operator/contrib/adamw-inl.h
```diff
@@ -211,6 +192,314 @@ struct AdamWUpdate {
...
+  mxnet_op::Stream<xpu>* s = ctx.get_stream<xpu>();
+  pParam->clip_gradient = p.clip_gradient;
+  pParam->beta1 = p.beta1;
+  pParam->beta2 = p.beta2;
+
+  pParam->epsilon = p.epsilon;
+
+  pParam->count = p.num_weights;
+  pParam->max_size = 0;
+  const bool isSame = std::is_same<DType, MPDType>::value;
```
ptrendx: `constexpr` since it is stronger than `const`.

tests/python/gpu/test_operator_gpu.py
```diff
@@ -49,6 +49,7 @@
 from test_contrib_operator import test_multibox_target_op
```
ptrendx: Add it here instead of adding a new line?

python/mxnet/ndarray/contrib.py
drivanov: Copy is always skipped when `rescale_grad` is defined in a proper context:
```
    def as_in_context(self, context):
        if self.context == context:
            return self
        return self.copyto(context)
```

ptrendx: Oh, that's good to know. Thanks :-)

src/operator/contrib/adamw-inl.h
drivanov: I could do that, but it does not look like a standard for the whole MxNet, perhaps, just for `adamw-inl.h`. I searched for `auto &` and `auto&` in all `*h, *.cc, *.cu` files. As I could see, the probability of this or that option is approximately 50%.

ptrendx: I understand, but inside a single file we should try to make it consistent :-).

src/operator/contrib/adamw-inl.h
drivanov: I will make this change.

src/operator/contrib/adamw-inl.h
drivanov: OK, I will change that.

python/mxnet/ndarray/contrib.py
```diff
@@ -542,14 +546,17 @@ def isnan(data):
     [1. 0.]
     <NDArray 2 @cpu(0)>
     """
-    return data != data # pylint: disable=comparison-with-itself
+    return data != data  # pylint: disable=comparison-with-itself
 
-def adamw_update(weight, grad, mean, var, rescale_grad, lr, eta, beta1=0.9, beta2=0.999,
-                 epsilon=1e-8, wd=0, clip_gradient=-1, out=None, name=None, **kwargs):
+def getRescaleGrad(rescale_grad, ctx=mx.cpu()):
```
eric-haibin-lin: nit: _get_rescale_grad

python/mxnet/ndarray/contrib.py
```diff
@@ -548,15 +548,15 @@ def isnan(data):
     """
     return data != data  # pylint: disable=comparison-with-itself
 
-def getRescaleGrad(rescale_grad, ctx=mx.cpu()):
+def _getRescaleGrad(rescale_grad, ctx=mx.cpu()):
```
eric-haibin-lin: nit pick on the style: we use snake_case for python methods, and underscore prefix for non-public methods.

drivanov: DONE

ptrendx:
Generally I would opt for cleaning the optimizers so that only the `multi_` versions of the optimizers code exist (so use the same code for both multi and single tensor versions). The next step would be to refactor common code for handling multiple tensors out.

eric-haibin-lin:
Thank you @drivanov

PR #16408, opened 2019-10-09
Add MXNet Ops for fast multihead attention

Caenorst:
## Description ##
Add new optimized Ops for Multihead attention

### Essentials ###

### Changes ###
- Add 4 Ops: (matmul(K,Q) and matmul(attention_weights, V). For both self-attention and encoder-decoder
- Add unit test for those Ops

## Comments ##
- https://github.com/Caenorst/gluon-nlp/tree/fast_mha shows a example of integration in BERT, it will not be PRed to gluon-nlp as it's breaking it (out of BERT usage).
- Those Ops require to have a different layout (sequence, batch, encoding) except for the masked softmax / dropout
- Those Ops change the ordering of projection weights, which means a pretrained BERT without those Ops need to have the weights processed as in: https://github.com/Caenorst/incubator-mxnet/commit/e98761456ba0343664ba550e056f00db31516ac7#diff-4758fb9329d438de2836db2634a8f5f7R2505-R2519 in order to use those Ops.
- The argument `bwd_ignore_zero_init` allow to further speedup and reduce the memory consumption but is only giving good results with `MXNET_EXEC_ENABLE_ADDTO` set to 1, it's also a dirty trick as it's actually not "adding to" but initializing (it rely on the fact that the two ops inputs are actually complete (not overlapping and using the whole tensor), despite using the same tensor.


src/operator/contrib/transformer.cu
```diff
@@ -22,12 +22,898 @@
...
+    const int32_t attn_batches      = inputs[2].shape_[0];
+    const int32_t q_seq_len         = inputs[2].shape_[1];
+    const int32_t embed_dim         = output_lin_kv_dim / 2;
+    int32_t head_dim                = embed_dim / params.heads;
+    const int32_t lead_dim_kv       = attn_batches * head_dim * 2;
+    const int32_t batch_stride_kv   = 2 * head_dim;
+    const float alpha               = 1.f;
+
+    if (req[0] != kNullOp) {
+      if (req[0] == kWriteTo && !params.bwd_ignore_zero_init) {
```
sxjscience: Why should we add this flag?

src/operator/contrib/transformer.cu
```diff
@@ -22,12 +22,898 @@
...
+#include <cuda.h>
+#include <cuda_runtime.h>
+#include <cuda_fp16.h>
+#include <cuda_profiler_api.h>
+
 #include <mxnet/base.h>
 #include "./transformer-inl.h"
+#include "../../common/cuda_utils.h"
+
+#include "cutlass/cutlass.h"
```
sxjscience: I guess we need to add cutlass as a third-party submodule like cub.

src/operator/contrib/transformer.cu
sxjscience: I cannot  well understand the rationale behind this flag. I think we can determine whether we need to inplace_accumulate the gradient when  `req[0] == kAddTo`

Caenorst: As I put in the description, this is a hack if you use the Ops within an multihead attention the input tensor is an interlacing of projected QKV,  one Op do backward on QK part the other do backward on V (and attention weights). because of this you actually never need to "accumulate" the gradients between both Ops. It's hack that you only enable if you use this flag `bwd_ignore_zero_init`. If you intend to use this Op out of the official multihead attention don't enable this flag and the behavior will be as expected.

src/operator/contrib/transformer.cu
Caenorst: Nice catch, gonna fix that

.gitmodules
```diff
@@ -26,3 +26,6 @@
 [submodule "3rdparty/nvidia_cub"]
 	path = 3rdparty/nvidia_cub
 	url = https://github.com/NVlabs/cub.git
+[submodule "3rdparty/cutlass"]
```
marcoabreu: May I propose to have a dev@ discussion before adding a new dependency?

Caenorst: Sure you can start one. On my side if there is no Cutlass then those will only work under certain dimensions.

sxjscience: @marcoabreu Sure, let's start a dev discussion. Also, it would be really nice if we could later move the BatchedGEMM accelerations to the linalg_impl.h so other ops can be benefited : https://github.com/apache/incubator-mxnet/blob/15ea40d9bf7da3c4618ca45a6d023d9b0fb1c295/src/operator/linalg_impl.h#L325-L368
@ptrendx @eric-haibin-lin @szha

src/operator/contrib/transformer.cu
```diff
@@ -22,12 +22,898 @@
...
+      CutlassGemm_FP32Accum<kRowMajor, kColumnMajor, 2, 2, 8>(stream, m, n, k, alpha, a, lda,
+        strideA, b, ldb, strideB, beta, c, ldc, strideC, batchCount);
+    } else if (!(lda & 0x1) && !(ldb & 0x1) && !(ldc & 0x3)) {
+      CutlassGemm_FP32Accum<kRowMajor, kColumnMajor, 2, 2, 4>(stream, m, n, k, alpha, a, lda,
+        strideA, b, ldb, strideB, beta, c, ldc, strideC, batchCount);
+    } else if (!(lda & 0x1) && !(ldb & 0x1) && !(ldc & 0x1)) {
+      CutlassGemm_FP32Accum<kRowMajor, kColumnMajor, 2, 2, 2>(stream, m, n, k, alpha, a, lda,
+        strideA, b, ldb, strideB, beta, c, ldc, strideC, batchCount);
+    } else {
+      CublasStridedBatchedGemm(s, transA, transB, m, n, k, alpha, a, lda, strideA, b, ldb,
```
sxjscience: One issue of the optimizations here is that it will increase the library size.

.gitmodules
marcoabreu: Please feel free to start one yourself, thanks!

Caenorst: Update: We decided to drop Cutlass from now, there are some cases on which Cublas is actually working better and it make the PR more simple

src/operator/contrib/transformer.cc
```diff
@@ -29,6 +29,163 @@
...
+                                                mxnet::ShapeVector* in_shape,
+                                                mxnet::ShapeVector* out_shape) {
+  CHECK_EQ(in_shape->size(), 2);
+  auto qkv_shape = in_shape->at(0);
+  auto att_shape = in_shape->at(1);
+  CHECK_EQ(qkv_shape.ndim(), 3);
+  CHECK_EQ(att_shape.ndim(), 3);
+  CHECK_EQ(qkv_shape[0], att_shape[1]);
+  CHECK_EQ(qkv_shape[0], att_shape[2]);
+  CHECK_EQ(qkv_shape[2] % 3, 0);
```
ptrendx: Could you make some meaningful error messages when the shape does not match your expectation?

src/operator/contrib/transformer.cc
```diff
@@ -29,6 +29,163 @@
...
+                                           mxnet::ShapeVector* in_shape,
+                                           mxnet::ShapeVector* out_shape) {
+  const auto& params = nnvm::get<InterleavedMatMulParam>(attrs.parsed);
+  CHECK_EQ(in_shape->size(), 2);
+  auto q_shape = in_shape->at(0);
+  auto kv_shape = in_shape->at(1);
+  CHECK_EQ(q_shape.ndim(), 3);
+  CHECK_EQ(kv_shape.ndim(), 3);
+  CHECK_EQ(q_shape[2] * 2, kv_shape[2]);
+  CHECK_EQ(q_shape[1], kv_shape[1]);
```
ptrendx: Same here

src/operator/contrib/transformer.cc
```diff
@@ -29,6 +29,163 @@
...
+
+static bool InterleavedMatMulEncDecValAttShape(const NodeAttrs& attrs,
+                                               mxnet::ShapeVector* in_shape,
+                                               mxnet::ShapeVector* out_shape) {
+  const auto& params = nnvm::get<InterleavedMatMulParam>(attrs.parsed);
+  CHECK_EQ(in_shape->size(), 2);
+  auto kv_shape = in_shape->at(0);
+  auto att_shape = in_shape->at(1);
+  CHECK_EQ(kv_shape[0], att_shape[2]);
+  CHECK_EQ(kv_shape[1] * params.heads, att_shape[0]);
```
ptrendx: And here

src/operator/contrib/transformer.cu
```diff
@@ -22,12 +22,568 @@
...
+
 #include <mxnet/base.h>
 #include "./transformer-inl.h"
+#include "../../common/cuda_utils.h"
 
 namespace mxnet {
 namespace op {
 
+// gemm_switch_fp32accum and the functions called are almost fully copied from:
+// MLPerf v0.6 submission repository from NVIDIA by https://github.com/kevinstephano
```
ptrendx: This is no longer the case, right?

src/operator/contrib/transformer.cc
```diff
@@ -29,6 +29,231 @@
...
+  return true;
+}
+
+NNVM_REGISTER_OP(interleaved_matmul_selfatt_qk)
+.describe(R"code(Compute the matrix multiplication between the projections of
+queries and keys in multihead attention use as self attention.
+
+the input must be a single tensor of interleaved projections
+of queries, keys and values following the layout:
+(seq_length, batch_size, num_heads * head_dim * 3)
```
eric-haibin-lin: Does this require the weight of k, q, v concatenated together? For LAMB it calculates the weight and grad norm per parameter, so for gradient update, we cannot use the weight stored in concatenated form, I assume?

src/operator/contrib/transformer.cc
```diff
@@ -29,6 +29,231 @@
...
+NNVM_REGISTER_OP(interleaved_matmul_encdec_valatt)
+.describe(R"code(Compute the matrix multiplication between the projections of
+values and the attention weights in multihead attention use as encoder-decoder.
+
+the inputs must be a tensor of interleaved projections of
+keys and values following the layout:
+(seq_length, batch_size, num_heads * head_dim * 2)
+
+and the attention weights following the layout:
+(batch_size, seq_length, seq_length)
```
eric-haibin-lin: Can we add math formula, or at least the unfused version, for others to understand what it computes?

src/operator/contrib/transformer.cc
Caenorst: Adding the concatenation reduce by about 20% the speedup due to multihead attention. I think we can think about an improvement but meanwhile that is still a speedup. I would encourage to make an analysis of LAMB coefficients difference within multihead attention blocks, maybe directly applying it on the concatenation of weights would be fine :man_shrugging: .

Caenorst:
I had to remove some tests on my Ops, the behaviour for bwd_ignore_zero_init cannot be tested as we can't enable the kAddTo yet.

src/operator/contrib/transformer.cc
Caenorst: As said in the description it's a matrix multiplication between the projections of queries and keys, it's equivalent to a batched_dot excepted that the inputs are within the same tensor and in a different layout.

I think if I try to include a math formula it's will be more misleading for the reader, while the example in tests/python/gpu/test_operator_gpu.py are simple to understand.

eric-haibin-lin:
@aaronmarkham is the website preview functionality still working after the website upgrade? I cannot see the preview of this PR: https://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-16408/12/index.html

tests/python/gpu/test_operator_gpu.py
```diff
@@ -2493,13 +2493,334 @@ def test_arange_like_dtype():
...
+        assert(grads_orig[k].dtype == grads_opti[k].dtype)
+        assert(grads_orig[k].shape == grads_opti[k].shape)
+        assert_allclose(grads_orig[k], grads_opti[k], rtol=1e-2, atol=1e-3)
+
+def test_multihead_attention_selfatt():
+    #os.environ['MXNET_EXEC_ENABLE_ADDTO'] = '0'
+    check_multihead_attention_selfatt(bwd_ignore_zero_init=False)
+    #os.environ['MXNET_EXEC_ENABLE_ADDTO'] = '1'
+    #check_multihead_attention_selfatt(bwd_ignore_zero_init=False)
+    #check_multihead_attention_selfatt(bwd_ignore_zero_init=True)
```
sxjscience: Why comment these lines?

Caenorst: My comment : https://github.com/apache/incubator-mxnet/pull/16408#issuecomment-543947555 refer to this modification

Apparently the usage of kAddTo is not working yet on this repository. and the behavior with `bwd_ignore_zero_init=True` is only correct if you enable it. I can add a little TODO comment like "to uncomment once kAddTo feature is working"

TaoLv:
Actually I personally don't like this kind of super fused operator which doesn't have explicit semantics and also lacks the flexibility as a general operator. I'm not sure if we should encourage our users to call these strange and half-baked operators in their models and how to do that if they already implemented their models? Besides, it also increase the burden for the community to maintain the complicated logic in the code.
I totally agree that fusions can bring performance improvement and make our framework outperforms other competing products. But I still think it would be better if fusions can be done in following ways:
1. Graph fusion pass: we have subgraph in NNVM and more features in the coming Relay IR;
2. Accelerating library support: integrate APIs and primitives from library and make them an operator is more acceptable. Just like what we did for fused RNN layers?
3. Maintain your customized fused operators in a separated repository and teach users how to integrate them into MXNet if they really want. I think the dynamic custom operator feature may help in this situation?
Just my two cents.

src/operator/contrib/transformer-inl.h
```diff
@@ -34,6 +34,18 @@
...
 namespace op {
 
+struct InterleavedMatMulParam : public dmlc::Parameter<InterleavedMatMulParam> {
+  int heads;
+  bool bwd_ignore_zero_init;
+  DMLC_DECLARE_PARAMETER(InterleavedMatMulParam) {
+    DMLC_DECLARE_FIELD(heads)
+    .describe("Set number of heads");
+    DMLC_DECLARE_FIELD(bwd_ignore_zero_init)
+    .describe("Make backward pass ignore AddTo and not init to 0.")
```
eric-haibin-lin: I don't think this one sentence explains this flag very well. When are users supposed to turn this on? Would it affect users who use gradient accumulation for self-attention?

Caenorst: Sure, it's kind of tricky to explain. Until we don't have the gradient accumulation feature working the user should not use this flag, it's cheating on purpose relying on the fact that the two ops of self attention are using complementary parts of the input tensor (not overlapping and the sum of all is the full tensor). If you want to use gradient accumulation with another Op you should not enable this flag. Any indication on wording would be appreciated.

Caenorst:
@TaoLv I agree that the dynamic custom operators feature would be a great fit here, but since it is not yet ready contrib is the next best place for such operator, also I did not modified any core component of MXNet or any existing model. Of course having an automatic fusion is better, but you first need the Op before having the fusion.

@eric-haibin-lin We intend to provide as a follow up a wrapper op that could be used for this and cudnn multihead attention that would unify this, but we don't want to wait for the cudnn multihead attention to cover all the usecases we needed right now.

eric-haibin-lin:
I can no longer find `MXNET_EXEC_ENABLE_ADDTO` in mxnet. Is this flag required?

Caenorst:
@eric-haibin-lin Only with `bwd_ignore_zero_init=True`, otherwise it works find regardless. I let this flag hoping that `MXNET_EXEC_ENABLE_ADDTO` get back soon

src/operator/contrib/transformer-inl.h
samskalicky: can you add more of what you said to the description? this is good info

tests/python/gpu/test_operator_gpu.py
```diff
@@ -2493,13 +2493,334 @@ def test_arange_like_dtype():
...
+
+    def convert_bias(F, q_bias, k_bias, v_bias, num_heads):
+        q_bias = F.reshape(q_bias, shape=(num_heads, -1))
+        k_bias = F.reshape(k_bias, shape=(num_heads, -1))
+        v_bias = F.reshape(v_bias, shape=(num_heads, -1))
+        all_bias = F.stack(q_bias, k_bias, v_bias, axis=1)
+        all_bias = F.reshape(all_bias, shape=(-1,))
+        return all_bias
+
+    dtype='float16'
```
samskalicky: should we test with other types? What types do these operators support?

Caenorst: I just added test for fp32 (the ops support fp16 and fp32), and a warning on bwd_ignore_zero_init to avoid usage without kAddTo functionality

src/operator/contrib/transformer-inl.h
```diff
@@ -34,6 +34,19 @@
...
 
+struct InterleavedMatMulParam : public dmlc::Parameter<InterleavedMatMulParam> {
+  int heads;
+  bool bwd_ignore_zero_init;
+  DMLC_DECLARE_PARAMETER(InterleavedMatMulParam) {
+    DMLC_DECLARE_FIELD(heads)
+    .describe("Set number of heads");
+    DMLC_DECLARE_FIELD(bwd_ignore_zero_init)
+    .describe("Make backward pass ignore AddTo and not init to 0. "
+              " /!\\ Only enable with MXNET_ENABLE_EXEC_ADDTO fonctionality")
```
marcoabreu: Is that constraint enforced and throws and error if the flag is not set?

Caenorst: The problem is that MXNET_ENABLE_EXEC_ADDTO is used only during binding. So the user could potentially bind one symbol with MXNET_ENABLE_EXEC_ADDTO=0 then swap it to MXNET_ENABLE_EXEC_ADDTO=1 (for instance to bind another symbol). If I check for the flag both network will see it at 1, while only the 2nd network will actually use the functionality.

marcoabreu: Do you envision that to happen in regular use cases? In case you don't want to raise an exception, we could still print a warning in case the flag is not set

Caenorst: It can happen, but probably not frequently. I can add a warning if you are fine with it.

ptrendx:
@marcoabreu, @samskalicky Are your comments resolved?

src/operator/contrib/transformer-inl.h
Caenorst: I found a way to raise an exception while cover my edge case. By checking with a custom ParamParser, as it's called during bindind. Hope the solution suit you :)

marcoabreu: Very neat solution, good job and thank you!

eric-haibin-lin:
I am still not comfortable with the `bwd_ignore_zero_init ` flag. The `MXNET_ENABLE_EXEC_ADDTO ` env var does not exist nor take any effect in the current version of mxnet. I'd suggest we remove this flag and related code for this PR. After `MXNET_ENABLE_EXEC_ADDTO ` is properly introduced by other PRs, we can add relevant optimization around it. 

> I do see there are some useful discussions about the design and flags used in this PR. Also seems you already have considerations about future improvement with MHA from cuDNN, dynamic customized operator or with subgraph and fusion pass. So I would hope you can prepare a design proposal to cleanly describe the design, the limitations, user suggestions, and your further plans. Please also cross post the proposal to dev@ so it can get fully discussed and I also think the community can learn much from that. If possible, some guys can help to fill the missing part on CPU side.
> BTW, what does that mean to MXNet Gluon if the CPU implementation is missing for an operator? @eric-haibin-lin

I agree a design proposal is helpful, but it is not required for a contrib op. The purpose for contrib ops is that ppl who are interested in this op can try and use the op in gluonnlp Block as @Caenorst suggested, and developers can gather timely feedbacks. 

The work on official multi-head attention block API / operator API, would definitely require a proposal and discussion on dev@ since it involves developers from different hardwares including CPU and GPU. 

As far as gluon-nlp is concerned, we will not integrate this op (without CPU support) to the official API for now. However, we will redirect users who are interested to the example usage provided by @Caenorst if they only care about using GPUs.

Caenorst:
Following @eric-haibin-lin comment I removed the flag `bwd_ignore_zero_init` and will put it back once AddTo functionality is supported

Caenorst:
Please @TaoLv as @eric-haibin-lin indicated there is no impact on Gluon, I removed the flag, so there is no reason to block this PR.

TaoLv:
Thank you @eric-haibin-lin for clarifying the impact on MXNet Gluon. My change request is not only for that but also for an elaborative proposal about these new operators. Although they will live under `contrib op` category, I didn't see a clear scope definition about `contrib op` in MXNet.
To make it clear, I hope below items can be described in the proposal:
1. What's the current design? APIs, features, formulas, limitations, performance, ...
2. Why cuDNN MHT, dynamic operator, or subgraph doesn't help here?
3. Do we need call for contribution for CPU implementation?
4. How long will these operators exist? If I understand correctly, once they're in, less likely they will be removed before next major release. So `contrib op` is not a trivial thing.
5. Will they be replaced with other implementations? with the same operator interfaces? If not, how about backward compatibility?
6. Do we encourage users and downstream projects to use them?
7. Known issues or further plan to improve?
I hope the proposal and more discussions can help other developers as well as users to understand these operators better and also make MXNet development more open.
I don't want to be a hostile here. So it's not necessary for you to finish the proposal before merging this PR.

aaronmarkham:
> @aaronmarkham is the website preview functionality still working after the website upgrade? I cannot see the preview of this PR: https://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-16408/11/index.html

No, a variety of features of the new site didn't work on S3. 
To preview, you need to follow the directions on the wiki, or use the devmenu features I added in this PR: https://github.com/apache/incubator-mxnet/pull/16514

Caenorst:
@TaoLV:
1) The current design are two separated Ops which represent each matrix multiplication part of the multihead attention on both case of self-attention and encoder-decoder the performance depend on the model implementation / hardware. As an example on our internal contains we have about 15% improvement on BERT training. The Ops force to modify the layout to (sequence_length, batch), and interleave the weights for input linear projection.

2) cuDNN MHA Ops are not as fast as those Ops yet. Subgraph API can be doable to automatically fuse to those ops, but the implementation is not obvious as the two Ops are inter-dependent. Those ops require to modify the layout which we can't do on the whole model without user involvement yet.

3) I don't think a CPU implementation would be important for now. Those ops have been designed to have specific performance improvements on GPUs, as there are no plans to have them on Gluon-NLP those Ops are here to give the option to the users to have a non-negligible speedup on GPU if they need to. We do provide an example of implementation on https://github.com/Caenorst/gluon-nlp/tree/fast_mha, but don't intend to present it as a "standard" implementation of multihead attention.
We do plan to send a design proposal to dev@ for the CuDNN fused multihead attention integration we're prototyping. For that efforts, we are looking for contributions from the CPU side to make the feature better.

4) We don't know, it depend on future performance / ergonomic improvement and users feedbacks.

5) We don't know, it depend on future performance / ergonomic improvement and users feedbacks.

6) NVIDIA would encourage and help users that are looking for a very efficient implementation of multihead attention on GPU. We wouldn't encourage the usage for somebody who's looking for "A BERT implementation" for both CPU and GPU.

7) Unfortunately because of lack of kAddTo full functionality there is a cudaMemset that could be avoided, we are looking forward completing this functionality. We have several options for improvement but we want to release some tests internally before making any plans. We would like to have a single op for the whole MHA but it constraint to force the _masked_softmax + dropout inside the Op, and that would remove some of the flexbility of the current set of Ops, hence our current decision to scatter it in two ops.

TaoLv:
Thanks for your response @Caenorst. Looking forward to your general proposal for cuDNN MHA integration. Now I'm withdrawing the change request.

DickJC123:
Thanks Tao.  Looking foward to working with you and others on MXNet's MHA definition.

tests/python/gpu/test_operator_gpu.py
```diff
@@ -2493,13 +2493,327 @@ def test_arange_like_dtype():
...
+    grads_orig = {k : v.asnumpy() for k, v in executor.grad_dict.items()}
+    assert_allclose(att_score_orig, att_score_opti, rtol=1e-2, atol=1e-3)
+    assert_allclose(output_orig, output_opti, rtol=1e-2, atol=1e-3)
+
+    for k in grads_opti.keys():
+        assert(grads_orig[k].dtype == grads_opti[k].dtype)
+        assert(grads_orig[k].shape == grads_opti[k].shape)
+        assert_allclose(grads_orig[k], grads_opti[k], rtol=1e-2, atol=1e-3)
+
+def test_multihead_attention_selfatt():
```
szha: this test should have been marked with minimum cudnn/cuda version requirement, since the feature is not available in earlier versions

TaoLv: These tests have been moved to unittests/test_operator.py. Please see the discussion here: https://github.com/apache/incubator-mxnet/pull/17138#issuecomment-568329046.

leezu:
BTW, `cublasGemmStridedBatchedEx` used in this PR is broken in Cuda 10.1 which will cause crashes on p2 instances. Seems fixed in Cuda 10.2 (as per release notes).

larroy:
So it's related to cuda and not the arch?

leezu:
No. My comment was wrong

PR #16416, opened 2019-10-09
Dgl ops 2

ChaiBapchya:
## Description ##
Added tests for following ops (required by DGL)
Input - Large tensor array
Hyperbolic ops
- arccosh
- arcsinh
- arctanh
- cosh
- sinh
- tanh

Misc
- sign
- batch_dot
- divide

Logical Ops
- logical_and
- logical_not
- logical_or
- logical_xor

Regression
- Linear Reg
- Logistic Reg
### Essentials ###

### Changes ###
- [x] Add tests to test_large_array.py


tests/nightly/test_large_array.py
```diff
@@ -841,7 +841,9 @@ def test_div():
     b = 3*nd.ones(shape=(LARGE_X, SMALL_Y))
     c = b
```
larroy: c=b.__div__(a)  is more clear. If CI fails might be a nit to refine.

tests/nightly/test_large_array.py
```diff
@@ -1212,6 +1214,80 @@ def test_full():
     assert a[-1][-1] == 3
 
 
+def test_hyperbolic():
+    def test_arccosh(a):
```
larroy: nit if CI fails: could you add a comment about testing only the last element?

ChaiBapchya: We are doing the last element check to save on time. But this is being done for each test (>50) so doesn't make sense to add for this specific test. In that case, should i add it to the top of the function?

sxjscience:
I think we've tested for batch_dot before

ChaiBapchya:
Batch_dot operator hasn't been tested for large array before. Atleast not in the test_large_array.py file

tests/nightly/test_large_array.py
```diff
@@ -1199,6 +1185,118 @@ def test_full():
...
+
+def test_sign():
+    a = mx.nd.random.normal(-1,1, shape=(LARGE_X, SMALL_Y))
+    mx_res = mx.nd.sign(a)
+    assert_almost_equal(mx_res[-1][-1].asnumpy(), np.sign(a[-1][-1].asnumpy()))
+
+
+def test_logical():
+    def test_logical_and(a, b):
+        mx_res = mx.nd.logical_and(a, b)
```
anirudh2290: all tests here test logical_and

tests/nightly/test_large_array.py
```diff
@@ -1199,6 +1185,118 @@ def test_full():
...
+
+    def test_arcsinh(a):
+        mx_res = mx.nd.arcsinh(a)
+        assert_almost_equal(mx_res[-1][-1].asnumpy(), np.arcsinh(a[-1][-1].asnumpy()))
+
+    def test_arctanh(a):
+        a[-1][-1] = 0   # arctanh of 1 is inf, assert_almost_equal gives "divide by 0" for comparing 2 inf values
+        mx_res = mx.nd.arctanh(a)
+        assert_almost_equal(mx_res[-1][-1].asnumpy(), np.arctanh(a[-1][-1].asnumpy()))
+
```
anirudh2290: can this be merged inside the trignometric functions added by : https://github.com/apache/incubator-mxnet/pull/16477

access2rohit: initial work division was done in this way. Can we do it later in a separate PR ?

anirudh2290: please do it in one of the four PRs that you two have opened. Shouldnt be very difficult. Let me know if this is the case.

tests/nightly/test_large_array.py
ChaiBapchya: good catch! fixed

tests/nightly/test_large_array.py
ChaiBapchya: merged

ChaiBapchya:
@anirudh2290 It looks same coz we are testing same functions. Agreed that we can bring both (large array input dim >=2) and large vector under same file. But for now going with this. Adding "merging array & vector tests" to the todo.

access2rohit:
@ChaiBapchya can you restart sanity test on your PR

tests/nightly/test_large_array.py
```diff
@@ -1201,6 +1187,76 @@ def test_full():
...
+        mx_res = mx.nd.logical_or(a, b)
+        assert_almost_equal(mx_res[-1][-1].asnumpy(), np.logical_and(a[-1][-1].asnumpy(), b[-1][-1].asnumpy()))
+
+    def test_logical_not(a, b):
+        mx_res = mx.nd.logical_not(a, b)
+        assert_almost_equal(mx_res[-1][-1].asnumpy(), np.logical_and(a[-1][-1].asnumpy(), b[-1][-1].asnumpy()))
+
+    def test_logical_xor(a, b):
+        mx_res = mx.nd.logical_xor(a, b)
+        assert_almost_equal(mx_res[-1][-1].asnumpy(), np.logical_and(a[-1][-1].asnumpy(), b[-1][-1].asnumpy()))
```
anirudh2290: comparing with logical_and everywhere ? is this tested ?

tests/nightly/test_large_array.py
```diff
@@ -1081,35 +1071,31 @@ def test_log_softmax():
...
     assert c.shape == a.shape
     assert c[0][-1] == 2
 
 
 def test_isub():
     a = nd.array(np.array(np.full((SMALL_Y, LARGE_X), 3)))
     b = nd.array(np.ones((SMALL_Y, LARGE_X)))
-    c = a
-    c -= b
+    c = a-b
```
anirudh2290: nit space between operators. 
```
a - b
```

tests/nightly/test_large_array.py
```diff
@@ -1351,6 +1407,48 @@ def check_tan():
...
+        # expected ouput for indices=(0, 1, -3, -2, -1) after applying cosh()
+        expected_output = [1, np.cosh(1), np.cosh(np.pi/2), np.cosh(3*np.pi/4), np.cosh(np.pi)]
+        assert_correctness_of_trigonometric_ops(y, expected_output)
+
+    def check_tanh():
+        x = create_input_for_trigonometric_ops([-1/4, -1/2, 0, 1/4, 1/2])
+        y = nd.tanh(x)
+        # expected ouput for indices=(0, 1, -3, -2, -1) after applying tanh()
+        expected_output = [np.tanh(-1/4), np.tanh(-1/2), 0, np.tanh(1/4), np.tanh(1/2)]
+        assert_correctness_of_trigonometric_ops(y, expected_output)
```
anirudhacharya: why are trigonometric ops being compared with the `assert_correctness_of_trigonometric_ops` function. why not use `assert_almost_equal` from the test `utils.py` file?

access2rohit: @anirudhacharya assert almost equals require ndarray inputs. In Large Tensor tests we are only inspecting certain specific indices of the output. Therefore instead of using `assert_almost_equals` I have written a function that does the follwoing:
`np.abs(output-expected)<=atol`

anirudhacharya: > assert almost equals require ndarray inputs.

`assert_almost_equal` accepts both `np.ndarray` and `mx.nd.array` - [see here](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/test_utils.py#L534-L541)

> In Large Tensor tests we are only inspecting certain specific indices of the output

why certain specific indices only? even if we do not want to compare the whole array, shouldn't the indices at least be randomized?

> that does the follwoing:
> `np.abs(output-expected)<=atol`

it is useful to have both absolute and relative tolerance when comparing arrays with floating point values, because tolerance should be seen as a percentage of the original value. 
```
abs(Expected_Value - Computed_Value) < RTOL * abs(Expected_Value) + ATOL
```
eg. a change of 0.1 does not seem much when comparing 100 and 100.1. But when comparing 0.1 and 0.2 a difference of 0.1 makes a lot of difference and should ideally fail the test.

access2rohit: So basically this function is specific to testing only Large Tensors. I agree that `assert_almost_equals` is the ideal way to but check each individual value will consume a lot of memory and will take a lot of time resulting in nightly timeout. This functions checks for certain predetermined indices of the output so the correctness is ensured. Since the initial input is small and then it is made repetitive so this check works. These values have been manually checked by comparing output from ndarray and numpy trigonometric ops and the atol was found to be 1e-3. Thats why only atol is used. Currently `assert_correctness_of_trigonometric_ops` is only specific to use case of large tensors

anirudhacharya: I understand you checking individual values instead of whole arrays for large tensors. but a few concerns in this file - 
 - there seems to be erroneous comments like here - https://github.com/apache/incubator-mxnet/pull/16416/files#diff-9ee911616af04047075035c95cf542fbL1298 in the `create_input_for_trigonometric_ops` function in line 1298. Please clean those up.
 - instead of using hard coded values like 1e-3 in `assert_correctness_of_trigonometric_ops` can they be passed as `atol` arguments to these utility functions. It can have a default value of `1e-3`.

I will merge this now, since these changes are not specific to this PR. But make sure you address these errors and clean up the `test_large_array` file in a subsequent PR.

ChaiBapchya: Addressed it here #16585

Issue #16399, opened 2019-10-08
ndarray.load crashes MXNet GPU builds on CPU machines


leezu:
## Description
For some parameter files below crash happens. It does not happen for all parameter files.

## Environment info (Required)

```
----------Python Info----------
Version      : 3.7.3
Compiler     : GCC 5.4.0 20160609
Build        : ('default', 'Jun 13 2019 13:24:27')
Arch         : ('64bit', 'ELF')
------------Pip Info-----------
Version      : 19.0.3
Directory    : /home/ubuntu/.pyenv/versions/3.7.3/lib/python3.7/site-packages/pip
----------MXNet Info-----------
Version      : 1.5.1
Directory    : /home/ubuntu/.local/lib/python3.7/site-packages/mxnet
Commit Hash   : c9818480680f84daa6e281a974ab263691302ba8
Library      : ['/home/ubuntu/.local/lib/python3.7/site-packages/mxnet/libmxnet.so']
Build features:
✔ CUDA
✔ CUDNN
✔ NCCL
✔ CUDA_RTC
✖ TENSORRT
✔ CPU_SSE
✔ CPU_SSE2
✔ CPU_SSE3
✔ CPU_SSE4_1
✔ CPU_SSE4_2
✖ CPU_SSE4A
✔ CPU_AVX
✖ CPU_AVX2
✖ OPENMP
✖ SSE
✔ F16C
✖ JEMALLOC
✖ BLAS_OPEN
✖ BLAS_ATLAS
✖ BLAS_MKL
✖ BLAS_APPLE
✔ LAPACK
✖ MKLDNN
✔ OPENCV
✖ CAFFE
✖ PROFILER
✔ DIST_KVSTORE
✖ CXX14
✖ INT64_TENSOR_SIZE
✔ SIGNAL_HANDLER
✖ DEBUG
----------System Info----------
Platform     : Linux-4.4.0-1095-aws-x86_64-with-debian-stretch-sid
system       : Linux
node         : ip-172-31-84-3
release      : 4.4.0-1095-aws
version      : #106-Ubuntu SMP Wed Sep 18 13:33:48 UTC 2019
----------Hardware Info----------
machine      : x86_64
processor    : x86_64
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                4
On-line CPU(s) list:   0-3
Thread(s) per core:    2
Core(s) per socket:    2
Socket(s):             1
[...]

PR #16432, opened 2019-10-11
Load NDArray only to GPU if GPU is present

leezu:
## Description ##
Fix https://github.com/apache/incubator-mxnet/issues/16399

### Essentials ###

### Changes ###
- [X] Fix crash when attempting to load a NDArray that was saved with GPU context on a MXNet runtime built with Cuda but when no GPUs are present

## Comments ##
- I'm not sure if the CI actually tests the case of "GPU build of MXNET but no GPU present". So the test added here may always pass on CI.

### DIFF
--- a/src/ndarray/ndarray.cc
+++ b/src/ndarray/ndarray.cc
@@ -1819,7 +1819,13 @@ bool NDArray::Load(dmlc::Stream *strm) {
     *this = std::move(temp); return true;
   } else {
 #if MXNET_USE_CUDA
-    *this = temp.Copy(ctx); return true;
+    int device_count = -1;
+    cudaGetDeviceCount(&device_count);
+    if (device_count > 0) {
+      *this = temp.Copy(ctx); return true;
+    } else {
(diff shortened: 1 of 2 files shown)


ZhennanQin:
For a machine with GPU, if user want to run model on CPU only, will this change works for him?

leezu:
This change only affects the case where a MXNet cuda-enabled build is used on a CPU-only machine. It avoids a crash that currently happens. So the use-case you describe is not affected.

ZhennanQin:
So if user want to use MXNet cuda-enabled build on a GPU equipped machine, but want to run model on its CPU, is this case supported?

leezu:
Yes. It is already supported currently. Nothing is changed with respect to that.

The problem addressed here, is that a serialized ndarray encodes the context it lived on before serialization. When loading it, and no GPU is present, MXNet crashes as it attempts to load the array to GPU. With this PR, we fallback to loading to CPU. This is also done for CPU-only builds of MXNet.

PR #16446, opened 2019-10-11
Aggregated zero grad

drivanov:
## Description ##
New MxNet operator for zeroing of multiple arrays.

### Essentials ###
- [x] Changes are complete

### Changes ###
- [x] Previously existed `test_zero_grad()` is extended for different `dtypes`. Besides, now it uses a  newly implemented `mx.nd.reset_arrays` operator and also contains test for zeroing of multiple arrays of different shapes.


src/operator/contrib/reset_arrays-inl.h
```diff
@@ -0,0 +1,99 @@
...
+                           std::vector<int>* in_type,
+                           std::vector<int>* out_type) {
+  const auto& param_ = dmlc::get<ResetArraysParam>(attrs.parsed);
+  CHECK_EQ(in_type->size(), param_.num_arrays);
+  const int dtype = (*in_type)[0];
+  CHECK_NE(dtype, -1) << "First input must have specified type";
+  for (size_t i = 0; i < in_type->size(); ++i) {
+    const auto currType = (*in_type)[i];
+    if (currType == -1)
+      (*in_type)[i] = dtype;
```
ptrendx: I think I prefer the approach from `ResetArraysShape` function where you do just return `false` from the function if you don't know the type of all inputs instead of assigning them to some value. Also I don't like the fact that you can fail here if the first input does not have type specified. Please make this function closer to what the `ResetArraysShape` function is doing.

drivanov: I completely agree with you on both issues. But, it seems that Moises (who was the author of this code) was trying to follow common pattern which is used in 25 other MxNet operators (to find all of them, I searched for"First input must have specified type"). Do you think that for this operator we should break we should break this pattern?

ptrendx: That pattern is actually an error and potentially breaks some networks that would otherwise work - MXNet's shape and type inference is a multi pass algorithm, where it is possible to not get the full graph inferred during the first pass, but still eventually succeed.

drivanov: OK, I will make this change for that operator.

PR #16448, opened 2019-10-11
Minor type issues in Squeeze

junrushao1994:
@zheng-da 

This issue caused an incorrect type signature in our auto-generated C++ wrapper in CPP package, [link](https://github.com/apache/incubator-mxnet/blob/master/cpp-package/scripts/OpWrapperGenerator.py).

With this bug exists, the cpp-package would generate the first argument of signature `const std::vector<Symbol> &`, but the correct one should be `const Symbol &`.


junrushao1994:
This issue caused an incorrect type signature in our auto-generated C++ wrapper in CPP package, [link](https://github.com/apache/incubator-mxnet/blob/master/cpp-package/scripts/OpWrapperGenerator.py).

junrushao1994:
I see...I didn't realize that it fixes an issue that @zheng-da and @eric-haibin-lin mentioned before...I will add a testcase...

eric-haibin-lin:
@junrushao1994 you're the man!

PR #16465, opened 2019-10-13
[Gluon] [Fix] Fix HybridBlock when hybridize is not called

sxjscience:
## Description ##
When `.hybridize()` is not called, the forward function of HybridBlock supports a broader range of inputs. For example, the combination of `ndarray` and scalar values. The following code will run smoothly without error:

```python
from mxnet import gluon
class FooHybrid(gluon.HybridBlock):
    def hybrid_forward(self, F, a, b):
        if isinstance(a, (list, tuple)):
            a = sum(a)
        if isinstance(b, (list, tuple)):
            b = sum(b)
        return a + b
foo_hybrid = FooHybrid()
a = foo_hybrid(mx.nd.ones((10,)), 1)
b = foo_hybrid(mx.nd.ones((10,)), mx.nd.ones((10,)))
```

However, when hybridized, the second line will trigger an error:

```python
from mxnet import gluon
class FooHybrid(gluon.HybridBlock):
    def hybrid_forward(self, F, a, b):
        if isinstance(a, (list, tuple)):
            a = sum(a)
        if isinstance(b, (list, tuple)):
            b = sum(b)
        return a + b
foo_hybrid = FooHybrid()
foo_hybrid.hybridize()
a = foo_hybrid(mx.nd.ones((10,)), 1)
# line below triggers the error
b = foo_hybrid(mx.nd.ones((10,)), mx.nd.ones((10,)))
```

https://github.com/apache/incubator-mxnet/pull/16280 tries to make sure that in both cases, HybridBlock will raise an error. However, it will make the code backward-incompatible. This PR fixes this backward-incompatible problem and adds new tests for such behavior.

### Essentials ###

### Changes ###
- [x] HybridBlock same as Block when not hybridized, test

### DIFF
--- a/python/mxnet/gluon/block.py
+++ b/python/mxnet/gluon/block.py
@@ -93,6 +93,53 @@ class _BlockScope(object):
         _BlockScope._current.value = self._old_scope
 
 
+def _gather_type_ctx_info(args):
+    """Analyze the elements inside the nested args object and find:
+        - If there exists ndarray
+        - If there exists symbol
+        - All contexts appearing in args
+
+    Parameters
+    ----------
+    args : list or NDArray or Symbol
+        Could be a nested architecture.
+
+    Returns
+    -------
+    has_symbol : bool
+        Whether the elements in args contains symbols
+    has_ndarray : bool
+        Whether the elements in args contains ndarrays
+    ctx_set : set of mxnet.context.Context
+        Contains all possible contexts of the inner ndarrays in args. Can be empty if there is no
+        ndarray inside args.
+    first_ctx : mxnet.context.Context or None
+        Context of the first appeared NDArray (for backward-compatibility)
+    """
+    if isinstance(args, NDArray):
+        return False, True, {args.context}, args.context
+    elif isinstance(args, Symbol):
+        return True, False, set(), None
+    elif isinstance(args, (list, tuple)):
+        has_symbol = False
+        has_ndarray = False
+        ctx_set = set()
+        first_ctx = None
+        for ele in args:
+            ele_has_sym, ele_has_nd, ele_ctx_set, ele_first_ctx =\
+                _gather_type_ctx_info(ele)
+            has_symbol = has_symbol or ele_has_sym
+            has_ndarray = has_ndarray or ele_has_nd
+            if first_ctx is None and ele_first_ctx is not None:
+                first_ctx = ele_first_ctx
+            ctx_set = ctx_set | ele_ctx_set
+            if has_symbol and has_ndarray:
+                break
+        return has_symbol, has_ndarray, ctx_set, first_ctx
+    else:
+        return False, False, set(), None
+
+
 def _flatten(args, inout_str):
     """Parse the arguments into a flattened list + an additional format array.
     The format array stores the structure of the original arguments to help reconstruct the inputs.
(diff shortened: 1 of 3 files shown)


sxjscience:
@leezu

python/mxnet/gluon/block.py
```diff
@@ -1054,34 +1098,16 @@ def register_op_hook(self, callback, monitor_all=False):
...
-        for ele in flatten_args:
-            if isinstance(ele, NDArray):
-                if is_ndarray is False:
-                    raise ValueError('In HybridBlock, we do not support mixed NDArrays and Symbols'
-                                     ' types for the input.\n'
-                                     'Received types are: {}.'
-                                     .format([type(ele) for ele in flatten_args]))
-                is_ndarray = True
-                exist_sym_nd = True
-                ctx = ele.context
```
leezu: This line introduced in #16280 is not compatible with the previous context handling. Previously, always `x.context` is used as default context. https://github.com/apache/incubator-mxnet/pull/16280/files#diff-29da832c2145752f3906a2f71c7b63baL982
Does it matter?

python/mxnet/gluon/block.py
```diff
@@ -1054,34 +1098,16 @@ def register_op_hook(self, callback, monitor_all=False):
...
-        if is_ndarray:
+        has_symbol, has_ndarray, ctx_set = _gather_type_ctx_info([x] + list(args))
+        if has_symbol and has_ndarray:
+            raise ValueError('In HybridBlock, we do not support mixed NDArrays and Symbols'
+                             ' types for the input. Please check the type of the args.\n')
+        if not has_symbol and not has_ndarray:
+            raise ValueError('In HybridBlock, there must be one NDArray or one Symbol in the input.'
+                             ' Please check the type of the args.\n')
+        if has_ndarray:
+            ctx = next(iter(ctx_set))
```
leezu: `set` is unordered, thus `next(iter(ctx_set))` may differ?

sxjscience: Nice catch, should sort it beforehand.

python/mxnet/gluon/block.py
sxjscience: It is chosen like this because x can be None now.

leezu: Then it should be set to the first non-`None` argument, not the last?

sxjscience: Actually, all ctxs are supposed to be the same. For example, we should not allow the mixing of cpu and gpu contexts. However, we currently allow to do so because we will need to mix  `cpu`, `cpu_pinned`, and `cpu_shared`.

leezu: Thus we should use the first non-`None` argument not to break backwards compatibility? ` cpu, cpu_pinned, cpu_shared` are different contexts after all

sxjscience: @leezu I think the difference between using the first or last non-None argument is not very important here. We need to just make sure that we will finally pick a meaningful one. In fact, the previous implementation has not checked the ctx of all the args.

leezu: As the previous implementation hasn't enforced all contexts being equal, we shouldn't start picking a different array to determine the context. As you stated above, it's valid to use a mix of  `cpu, cpu_pinned, cpu_shared` contexts.
For example, after your change, `cpu_pinned` or `cpu_shared` may be picked as default context instead of `cpu` if the user passed a  `cpu_pinned` or `cpu_shared`  as last argument. The extra overhead could cause a performance regression (all parameters will be made available under default context).
No need to risk this given there is no advantage?

sxjscience: @leezu It's also possible that, previously cpu_pinned is picked as the default argument and after the change, the correct cpu context is picked as the default. My point is we need to probably give special treatment of the `cpu, cpu_pinned, cpu_shared`. What's your opinion?

sxjscience: @leezu I agree that the backward-compatible issue is valid. Let me first make it to be backward-compatible. However, this does not fix the issue of the cpu, cpu_pinned, cpu_shared combination.

leezu: We should get rid of choosing one array and using it's context as default context. For parameters, users should get the array via `self.weight.data(ctx)`. For the time being I suggest not to break the behaviour, to avoid unintended consequences

Issue #16495, opened 2019-10-15
docs for gluon.data.* are missing


aaronmarkham:
Probably related to git filters when adding the files.
https://stackoverflow.com/questions/8006393/force-add-despite-the-gitignore-file

PR #16496, opened 2019-10-15
fix missing docs due to git add issues

sojiadeshina:
## Description ##
Some files were missing from the #16392 pr due to gitignore issue. They are now included here. 

This should allow the pages for gluon.data to be rendered properly.

Fixes https://github.com/apache/incubator-mxnet/issues/16495


sojiadeshina:
@aaronmarkham can we restart unix-gpu?

aaronmarkham:
> @aaronmarkham can we restart unix-gpu?

I restarted it for you, but now that the kvstore test that was timing out has been removed, if this fails again, I'd rebase, so you can get past this test that keeps failing.

PR #16512, opened 2019-10-16
detect number of procs during sphinx build

aaronmarkham:
## Description ##
I ran into a build issue when making the python docs on an instance with 16 processors. The makefile for sphinx was hardcoded to 32 which I guess was fine for CI.

The error was like this:

```
/work/mxnet/docs/python_docs/python/build/tutorials/packages/gluon/training/fit_api_tutorial.ipynb:395: WARNING: File not found: 'tutorials/packages/gluon/blocks/save_load_params.html#saving-model-parameters-to-file'

Sphinx parallel build error:
IndexError: list index out of range
```

I turned off parallel builds with Sphinx and it worked, then switched to this new code that detects the number of processors in the Makefile and that worked.

## Testing
1. Clone the repo.
2. [Install Docker and Docker for python and make it so docker doesn't need sudo to run](https://github.com/apache/incubator-mxnet/blob/master/ci/README.md).
3. Run the following two commands to build the mxnet binary, then build the python docs.

```
ci/build.py --docker-registry mxnetci --platform ubuntu_cpu_lite /work/runtime_functions.sh build_ubuntu_cpu_docs
ci/build.py --docker-registry mxnetci --platform ubuntu_cpu_python /work/runtime_functions.sh build_python_docs
```

### DIFF
--- a/docs/python_docs/python/Makefile_sphinx
+++ b/docs/python_docs/python/Makefile_sphinx
@@ -18,8 +18,25 @@
 # Makefile for Sphinx documentation
 #
 
+
+NPROCS := 1
+OS := $(shell uname)
+export NPROCS
+
+ifeq ($J,)
+
+ifeq ($(OS),Linux)
+  NPROCS := $(shell grep -c ^processor /proc/cpuinfo)
+else ifeq ($(OS),Darwin)
+  NPROCS := $(shell system_profiler | awk '/Number of CPUs/ {print $$4}{next;}')
+endif # $(OS)
+
+else
+  NPROCS := $J
+endif # $J
+
 # You can set these variables from the command line.
-SPHINXOPTS    = -j 32 -c ../scripts
+SPHINXOPTS    = -j$(NPROCS) -c ../scripts
 SPHINXBUILD   = sphinx-build
 PAPER         =
 BUILDDIR      = _build
--- a/docs/python_docs/python/Makefile_sphinx
+++ b/docs/python_docs/python/Makefile_sphinx
@@ -18,12 +18,12 @@
 # Makefile for Sphinx documentation
 #
 
-
+# Begin number of processors detection
 NPROCS := 1
 OS := $(shell uname)
 export NPROCS
 
-ifeq ($J,)
+ifeq ($(NUMJOBS),)
 
 ifeq ($(OS),Linux)
   NPROCS := $(shell grep -c ^processor /proc/cpuinfo)
@@ -32,8 +32,9 @@ else ifeq ($(OS),Darwin)
 endif # $(OS)
 
 else
-  NPROCS := $J
-endif # $J
+  NPROCS := $(NUMJOBS)
+endif # $(NUMJOBS)
+# End number of processors detection
 
 # You can set these variables from the command line.
 SPHINXOPTS    = -j$(NPROCS) -c ../scripts


docs/python_docs/python/Makefile_sphinx
```diff
@@ -18,8 +18,25 @@
...
 
+
+NPROCS := 1
+OS := $(shell uname)
+export NPROCS
+
+ifeq ($J,)
+
+ifeq ($(OS),Linux)
+  NPROCS := $(shell grep -c ^processor /proc/cpuinfo)
```
larroy: can we use nproc here?

docs/python_docs/python/Makefile_sphinx
```diff
@@ -18,8 +18,25 @@
 # Makefile for Sphinx documentation
 #
 
+
+NPROCS := 1
+OS := $(shell uname)
+export NPROCS
+
+ifeq ($J,)
```
larroy: can we change J to numjobs or something different than 1 character variable?

docs/python_docs/python/Makefile_sphinx
aaronmarkham: I tried that first, but I guess you can't do that in a Makefile... so I looked it up and found this referenced a couple of times. It did the trick...

docs/python_docs/python/Makefile_sphinx
aaronmarkham: not sure... is it really important? I'd rather not suffer through another CI pass...

szha: yes, it's important. code is usually written once and read by many people many times. having a good variable name will help make their life easier.

aaronmarkham: Ok, please re-review; I updated the variable name.

aaronmarkham: Well, [that change broke it](http://jenkins.mxnet-ci.amazon-ml.com/job/mxnet-validation/job/website/job/PR-16512/2/display/redirect). I think "J" was probably required. IDK enough about it though...

szha: environment variable names in makefile require parentheses around them

docs/python_docs/python/Makefile_sphinx
larroy: ok, makes sense.

docs/python_docs/python/Makefile_sphinx
larroy: or {}

larroy: https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_6.html

aaronmarkham:
> Also if we have troubles with make we can use a python thing to build the sphinx docs, there are python alternatives. Make is just a dumb depgraph.

This file is what's included in Sphinx so people expect it. But sure, we could use something different to invoke sphinx.

docs/python_docs/python/Makefile_sphinx
aaronmarkham: ok this works now!

aaronmarkham:
Kind of ridiculous how many times I've had to restart the tests on this PR.

Restarting centos-gpu now due to failing here: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fcentos-gpu/detail/PR-16512/4/pipeline

larroy:
Can we collect the failures to understand what were the root causes?

larroy:
In the one you linked S3 failed, so we could add a retry.

Issue #14288, opened 2019-02-28
Flaky test: test_operator_gpu.test_dropout


junrushao1994:
http://jenkins.mxnet-ci.amazon-ml.com/blue/rest/organizations/jenkins/pipelines/mxnet-validation/pipelines/unix-gpu/branches/PR-14270/runs/5/nodes/271/steps/624/log/?start=0

```
======================================================================
FAIL: test_operator_gpu.test_dropout
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/usr/local/lib/python2.7/dist-packages/nose/util.py", line 620, in newfunc
    return func(*arg, **kw)
  File "/work/mxnet/tests/python/gpu/../unittest/common.py", line 173, in test_new
    orig_test(*args, **kwargs)
  File "/work/mxnet/tests/python/gpu/../unittest/test_operator.py", line 6107, in test_dropout
    check_dropout_ratio(1.0, shape, cudnn_off=False)
  File "/work/mxnet/tests/python/gpu/../unittest/test_operator.py", line 6040, in check_dropout_ratio
    check_correctness(exe, exe.arg_arrays[0].asnumpy(), ratio)
  File "/work/mxnet/tests/python/gpu/../unittest/test_operator.py", line 6005, in check_correctness
    assert output_zeroes == len(input)
AssertionError: 
-------------------- >> begin captured logging << --------------------
common: INFO: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=623337751 to reproduce.
--------------------- >> end captured logging << ---------------------
```

junrushao1994:
See also

http://jenkins.mxnet-ci.amazon-ml.com/blue/rest/organizations/jenkins/pipelines/mxnet-validation/pipelines/unix-gpu/branches/PR-14270/runs/5/nodes/271/steps/624/log/?start=0

junrushao1994:
@mxnet-label-bot add [Test, Flaky]

perdasilva:
Seeing it again: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/master/650/pipeline

Making a PR to disable test until it is fixed.

ChaiBapchya:
Was able to reproduce this issue.
With `cudnn_off=True`, 

```
[...]

PR #13896, opened 2019-01-16
Cudnn dropout

szha:
## Description ##
Use dropout in CuDNN

Tested on p3.2x (V100). Test case:
```python
import mxnet as mx
a = mx.nd.ones((10, 200, 300, 500), ctx=mx.gpu(0))
a.attach_grad()
mx.autograd.set_recording(True)
%timeit mx.nd.Dropout(a, 0.5, mode='always').wait_to_read()
```

|          | fwd   | fwd+bwd |
|----------|-------|---------|
| cudnn GPU  | ~46ms~ 4.3ms  | ~48ms~ 15ms   |
| prev GPU | 305ms | 321ms   |

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] CuDNN Dropout.
- [x] skip resource request for parallel random when using cudnn forward.
- [x] extend cpptest to support stateful operators, resourcerequestex, ResourceRequest::kCuDNNDropoutDesc.
- [x] fix bug where the backward pass on an inference-mode dropout forward didn't act as identity.
- [x] add kCuDNNDropoutDesc resource for sharing dropout state across operators.
- [x] replace dropout with identity when it's not dropping anything.

## Comments ##
- `cudnnSetDropoutDescriptor` is an expensive call due to initialization on each of the stream multiprocessor on a GPU. Since cudnn 7, `cudnnRestoreDropoutDescriptor` becomes available so that the initialized space can be cached. This descriptor is currently used in both Dropout op and RNN op. We need a mechanism for caching this so that initialization on each stream happens only once, as the same desc can be shared among operators.

PR #16532, opened 2019-10-18
fix dropout gpu seed

### LINKED ISSUE #15662: cudnn Dropout reproducibility
https://github.com/apache/incubator-mxnet/blob/master/src/operator/nn/dropout-inl.h#L491-L493

The seed of CUDNN dropout does not respect the random seed in mxnet

roywei:
## Description ##
fix #15662


tests/python/gpu/test_operator_gpu.py
```diff
@@ -2507,6 +2507,18 @@ def test_arange_like_dtype():
         for v in out:
             assert v.dtype == t
 
+# do not remove seed
+@with_seed(1234)
+def test_dropout_with_seed():
+    a = mx.nd.ones((10, 10), ctx=mx.gpu())
+    dropout = mx.gluon.nn.Dropout(0.5)
+    with mx.autograd.record():
```
wkcn: I have a suggestion.
Firstly, replace `@with_seed(1234)` with `@with_seed()`.
Then add the following code,
```python
info = np.iinfo(np.int64)
seed = np.random.randint(info.min, info.max)
```
Finally, we add `mx.random.seed(seed)` before each `dropout(a)`.

wkcn: We can move the unittest into `tests/python/unittest/test_operator.py`, since `test_dropout_with_seed` is also necessary for CPU.

src/operator/nn/dropout-inl.h
```diff
@@ -253,7 +253,18 @@ class DropoutOp {
...
+      Random<cpu, unsigned> *prnd = ctx.requested[1].get_random<cpu, unsigned>(stream);
+      Tensor<cpu, 1, char> workspace =
+        ctx.requested[2].get_space_typed<cpu, 1, char>(Shape1(1), stream);
+      // slice workspace
+      char *workspace_ptr = workspace.dptr_;
+      Tensor<cpu, 1, unsigned> random_number =
+        Tensor<cpu, 1, unsigned>(reinterpret_cast<unsigned *>(workspace_ptr),
+                                 Shape1(1), stream);
+      prnd->GetRandInt(random_number);
+      uint64_t seed_ = 17 + reinterpret_cast<uint64_t>(&random_number) % 4096;  // NOLINT(runtime/threadsafe_fn)
```
wkcn: Why is it `&random_number` ?

src/operator/nn/dropout-inl.h
```diff
@@ -262,7 +262,7 @@ class DropoutOp {
         Tensor<xpu, 1, unsigned>(reinterpret_cast<unsigned *>(workspace_ptr),
                                  Shape1(1), s);
       prnd->GetRandInt(random_number);
-      uint64_t seed_ = 17 + reinterpret_cast<uint64_t>(&random_number) % 4096;
+      uint64_t seed_ = 17 + reinterpret_cast<uint64_t>(random_number.dptr_) % 4096;
```
wkcn: random_number[0]

roywei: i get `dropout-inl.h(265): error: invalid type conversion`

roywei:
Still investigate why local unit test passed, but CI constantly failed, seems seed is not fixed in CI.

On local GPU, running the following passed:
1. nosetests single test passed 10000 times
2. nosetests all test_operator_gpu passed
3. runing https://github.com/apache/incubator-mxnet/issues/15662#issuecomment-540911324 directly from python

However, this test failed on CI with both `mx.seed(fixed_seed)` and `@with_seed(fixed_seed)` decorator.

I wil try to reproduce CI failure locally first. Or try to add this to nightly test where less nosetests are executed at the same time. Suspect other nosetest running parallel on CI workers will affect the result.

roywei:
cc @eric-haibin-lin @sxjscience

src/operator/nn/dropout-inl.h
szha: dptr_ is a pointer and I don't think you mean to use its address here.

roywei:
I am able to reproduce CI failure locally now by running the following on P3.8x with DLAMI
`ci/build.py --docker-registry mxnetci --nvidiadocker --platform ubuntu_gpu_cu101 --docker-build-retries 3 --shm-size 500m /work/runtime_functions.sh unittest_ubuntu_python2_gpu`

result:
```
FAIL: test_operator_gpu.test_dropout_with_seed
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/usr/local/lib/python2.7/dist-packages/nose/util.py", line 620, in newfunc
    return func(*arg, **kw)
    orig_test(*args, **kwargs)
  File "/work/mxnet/tests/python/gpu/../unittest/test_operator.py", line 6946, in test_dropout_with_seed
    assert_almost_equal(b.asnumpy(), c.asnumpy())
  File "/work/mxnet/python/mxnet/test_utils.py", line 624, in assert_almost_equal
    raise AssertionError(msg)
AssertionError:
Items are not equal:
Error 100000002004087734272.000000 exceeds tolerance rtol=1.000000e-05, atol=1.000000e-20 (mismatch at least 0.110000%).
Location of maximum error: (0, 1), a=2.00000000, b=0.00000000
 ACTUAL: array([[0., 2., 2., ..., 2., 0., 0.],
       [0., 2., 2., ..., 0., 0., 2.],
       [2., 2., 2., ..., 0., 0., 2.],...
 DESIRED: array([[2., 0., 2., ..., 2., 0., 2.],
       [2., 2., 2., ..., 2., 2., 2.],
       [2., 0., 0., ..., 2., 2., 2.],...

*** Maximum errors for vector of size 10000:  rtol=1e-05, atol=1e-20
--------------------- >> end captured stdout << ----------------------
common: INFO: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=179619306 to reproduce.

```

**However, running the test standalone with the same seed failed with CI enviroment passed:**
```
MXNET_TEST_SEED=179619306 nosetests --logging-level=DEBUG --verbose -s  tests/python/gpu/test_operator_gpu.py:test_dropout_with_seed
[INFO] Setting module np/mx/python random seeds, use MXNET_MODULE_SEED=980748466 to reproduce.
[WARNING] *** test-level seed set: all "@with_seed()" tests run deterministically ***
test_operator_gpu.test_dropout_with_seed ... [INFO] Setting test np/mx/python random seeds, use MXNET_TEST_SEED=179619306 to reproduce.
[07:36:44] ../src/base.cc:84: Upgrade advisory: this mxnet has been built against cuDNN lib version 7401, which is older than the oldest version tested by CI (7600).  Set MXNET_CUDNN_LIB_CHECKING=0 to quiet this warning.
ok

Ran 1 test in 13.896s

OK
```

src/operator/nn/dropout-inl.h
```diff
@@ -253,7 +253,16 @@ class DropoutOp {
...
+      Random<xpu, unsigned> *prnd = ctx.requested[1].get_random<xpu, unsigned>(s);
+      Tensor<xpu, 1, char> workspace =
+        ctx.requested[2].get_space_typed<xpu, 1, char>(Shape1(1), s);
+      // slice workspace
+      char *workspace_ptr = workspace.dptr_;
+      Tensor<xpu, 1, unsigned> random_number =
+        Tensor<xpu, 1, unsigned>(reinterpret_cast<unsigned *>(workspace_ptr),
+                                 Shape1(1), s);
+      prnd->GetRandInt(random_number);
+      uint64_t seed_ = 17 + reinterpret_cast<uint64_t>(&random_number[0]) % 4096;
```
wkcn: Why does it need the modulus operator `%` ? the modulus operator makes the `seed_` between 0+17~4096+17

roywei: I'm just keeping the original logic here:
https://github.com/roywei/incubator-mxnet/blob/master/src/operator/nn/dropout-inl.h#L95

and here:
https://github.com/roywei/incubator-mxnet/blob/master/src/operator/nn/dropout-inl.h#L491

while fixing seed to respect mxnet seed.

roywei:
> Hi @roywei , could you please add a unit-test for multi-GPU? The result of Dropout on multi-GPU should be different.
> Thanks a lot : )

I believe GPU unit tests are running on instances with 1GPU. I will try to move the entire test to nightly tests where it's using P3 instances with 4 gpus. I can add multi-gpu test there. Hopefully the seed can be properly fixed with less parallel jobs on CI workers.

src/operator/nn/dropout-inl.h
wkcn: I think it should be `uint64_t seed_ = 17 + static_cast<uint64_t>(random_number[0]) % 4096;`, because the type of `random_number[0]` is `unsigned`.

https://github.com/apache/incubator-mxnet/blob/master/3rdparty/mshadow/mshadow/tensor.h#L591

src/operator/nn/dropout-inl.h
wkcn: I think it should be `uint64_t seed_ = 17 + static_cast<uint64_t>(random_number[0]) % 4096;`, because the type of `random_number[0]` is `unsigned`.

https://github.com/apache/incubator-mxnet/blob/master/3rdparty/mshadow/mshadow/tensor.h#L591

tests/nightly/test_dropout_with_seed.py
```diff
@@ -0,0 +1,43 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import mxnet as mx
+import numpy as np
+from mxnet.test_utils import assert_almost_equal
+
+
+def test_dropout_with_seed():
```
marcoabreu: Add the with seed annotation

marcoabreu:
Ci runs on g3.8xlarge with 2 GPUs

tests/nightly/test_dropout_with_seed.py
roywei: I'm manually choosing a random seed and set it before each forward, so with_seed decorator will not take effect. See comment: https://github.com/apache/incubator-mxnet/pull/16532#discussion_r336370708

marcoabreu: In line 29 you're generating a random number to feed the seed though, so that generator needs to be fed with a seed as well

src/operator/nn/dropout-inl.h
roywei: this will give a segfault during dropout. Also why would dropout on mult-gpu return different result? I thought the seed in fixed at global? so dropout on different GPU will use the same seed, thus return the same result?

wkcn: Sorry that I didn't express it clearly. If different GPUs use the same seed, the result of drouout of different GPU should be the same. When training a model, do GPUs use different random seed?

src/operator/nn/dropout-inl.h
```diff
@@ -253,7 +253,16 @@ class DropoutOp {
...
+      Random<xpu, unsigned> *prnd = ctx.requested[1].get_random<xpu, unsigned>(s);
+      Tensor<xpu, 1, char> workspace =
+        ctx.requested[2].get_space_typed<xpu, 1, char>(Shape1(1), s);
+      // slice workspace
+      char *workspace_ptr = workspace.dptr_;
+      Tensor<xpu, 1, unsigned> random_number =
+        Tensor<xpu, 1, unsigned>(reinterpret_cast<unsigned *>(workspace_ptr),
+                                 Shape1(1), s);
+      prnd->GetRandInt(random_number);
+      uint64_t seed_ = 17 + static_cast<uint64_t>(random_number[0]) % 4096;
```
eric-haibin-lin: The tensor is on GPU, we need to explicitly copy it back to CPU using cudaMemCopy. You might get garbage value if you just access data on GPU mem address directly

tests/nightly/test_dropout.py
```diff
@@ -0,0 +1,80 @@
...
+
+    mx.random.seed(seed, ctx=mx.gpu(0))
+    with mx.autograd.record():
+        result1 = dropout(data1)
+
+    mx.random.seed(seed, ctx=mx.gpu(0))
+    with mx.autograd.record():
+        result2 = dropout(data1)
+
+    mx.random.seed(seed, ctx=mx.gpu(0))
```
wkcn: `mx.random.seed(seed, ctx=mx.gpu(1))` ?

roywei: I'm trying to fix seed on gpu 0 only, so gpu1 still have random seed, so result3 and result2 will be different. Otherwise, I will need to create a different `seed2` and fix it on gpu1, it would take the same effect. (different seed on gpu0 and gpu1 leading to result2 and result3 to be different)

roywei: updated

src/operator/nn/dropout-inl.h
```diff
@@ -253,7 +253,19 @@ class DropoutOp {
...
+      // slice workspace
+      char *workspace_ptr = workspace.dptr_;
+      Tensor<xpu, 1, unsigned> random_number =
+        Tensor<xpu, 1, unsigned>(reinterpret_cast<unsigned *>(workspace_ptr),
+                                 Shape1(1), s);
+      prnd->GetRandInt(random_number);
+      // copy generated random int to cpu
+      unsigned data = 0;
+      CUDA_CALL(cudaMemcpy(&data, &random_number[0], sizeof(unsigned), cudaMemcpyDeviceToHost));
+      uint64_t seed_ = 17 + static_cast<uint64_t>(data) % 4096;
```
eric-haibin-lin: @ptrendx @DickJC123 any concern for the fix?

roywei:
Passed nightly test locally:
command:
```
ci/build.py --docker-registry mxnetci --nvidiadocker --platform ubuntu_nightly_gpu --docker-build-retries 3 --shm-size 500m /work/runtime_functions.sh nightly_python
```

result:
```
Ran 2 tests in 351.086s

OK
build.py: 2019-10-22 17:05:18,213Z INFO Waiting for status of container 5d41b56c614c for 600 s.
build.py: 2019-10-22 17:05:18,443Z INFO Container exit status: {'StatusCode': 0, 'Error': None}
build.py: 2019-10-22 17:05:18,443Z INFO Container exited with success 👍
build.py: 2019-10-22 17:05:18,443Z INFO Stopping container: 5d41b56c614c
build.py: 2019-10-22 17:05:18,445Z INFO Removing container: 5d41b56c614c
```

src/operator/nn/dropout-inl.h
```diff
@@ -253,7 +253,19 @@ class DropoutOp {
                            const TBlob &mask,
                            const TBlob &out) {
       Stream<xpu> *s = ctx.get_stream<xpu>();
```
szha: not directly related to the changes in this PR: there's no chance for this to run in CPU so the xpu template here isn't necessary.

roywei: yes, and if we can access CPU here it will be much easier, there is a getRandInt() for CPU that directly returns an unsigned number.

src/operator/nn/dropout-inl.h
ptrendx: So there are multiple problems with this:
 - why are you trying to seed the dropout every time you do forward? It will not actually do anything I believe as `get_cudnn_dropout_desc` after the first call calls [cudnnRestoreDropoutDescriptor](https://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html#cudnnRestoreDropoutDescriptor) which seems to ignore the seed value. If you actually want to seed the dropout every time (via `cudnnSetDropoutDescriptor`) that would be super costly as seeding random number generator is way more expensive than actually using it.
 - you should never use `cudaMemcpy` in the operator code. If you really need to copy values from GPU to CPU you should use `cudaMemcpyAsync` followed by `cudaStreamSynchronize`. The difference is `cudaMemcpy` synchronizes on all streams (so it waits on all GPU activity and prevents all subsequent work to be done on all worker threads) vs `cudaStreamSynchronize` which waits only on the stream that you pass as argument (so other GPU workers are not affected).

roywei:
Just noticed our `test_dropout` unittest is disabled due to flakiness. (https://github.com/apache/incubator-mxnet/issues/14288)
 
I have verified locally this test passed with my PR, the test is still flaky though (failed with 10000 runs, use MXNET_TEST_SEED=107821594)

```
python3 -m "nose" tests/python/gpu/test_operator_gpu.py:test_dropout
[INFO] Setting module np/mx/python random seeds, use MXNET_MODULE_SEED=1842556802 to reproduce.
[22:22:55] ../src/base.cc:84: Upgrade advisory: this mxnet has been built against cuDNN lib version 7401, which is older than the oldest version tested by CI (7600).  Set MXNET_CUDNN_LIB_CHECKING=0 to quiet this warning.
.
Ran 1 test in 16.091s

OK
```

src/operator/nn/dropout-inl.h
eric-haibin-lin: What about having the following logic inside `get_cudnn_dropout_desc`:

```
  if (!state_space->handle.size) {
      request = ResourceManager::Get()->Request(cpu_ctx, ResourceRequest::kRandom)
      seed = request.GetRandInt()
      CUDNN_CALL(cudnnSetDropoutDescriptor(..., seed))
} else {
    // use a dummy seed (e.g. 0)  for cudnnRestoreDropoutDescriptor
}

```
and we remove the seed argument for `get_cudnn_dropout_desc`?

eric-haibin-lin: On a separate note, I see 34 occurrence of `cudaMemcpy` in various places (mainly operators) in the codebase, we probably need to do some cleanup:

```
src/kvstore/kvstore_utils.cu:  CUDA_CALL(cudaMemcpy(sort_output_ptr, dptr, sort_output_bytes,
src/kvstore/kvstore_utils.cu:  CUDA_CALL(cudaMemcpy(&num_selected_out, num_selected_ptr, num_selected_bytes,
src/ndarray/ndarray_function.cu:      CUDA_CALL(cudaMemcpy(&nnr_out, &row_flg[num_rows-1], sizeof(dim_t),
src/operator/contrib/adamw.cu:    CUDA_CALL(cudaMemcpy(&scale, scale_blob.dptr<DType>(), sizeof(DType),
src/operator/contrib/boolean_mask.cu:  CUDA_CALL(cudaMemcpy(&valid_num, &prefix_sum[idx_size - 1], sizeof(int32_t),
src/operator/contrib/index_array.cu:    CUDA_CALL(cudaMemcpy(workspace.dptr_, cpu_workspace.data(), sizeof(int64_t) * (2 * naxes),
src/operator/contrib/index_array.cu:    CUDA_CALL(cudaMemcpy(workspace.dptr_, inshape.data(), sizeof(dim_t) * ndim,
src/operator/contrib/multi_proposal.cu:  FRCNN_CUDA_CHECK(cudaMemcpy(&mask_host[0],
src/operator/contrib/multi_proposal.cu:    FRCNN_CUDA_CHECK(cudaMemcpy(workspace_proposals.dptr_, &anchors[0],
src/operator/contrib/multi_proposal.cu:        FRCNN_CUDA_CHECK(cudaMemcpy(keep, &_keep[0], sizeof(int) * _keep.size(),
src/operator/contrib/proposal.cu:  FRCNN_CUDA_CHECK(cudaMemcpy(&mask_host[0],
src/operator/contrib/proposal.cu:    FRCNN_CUDA_CHECK(cudaMemcpy(workspace_proposals.dptr_,
src/operator/contrib/proposal.cu:    FRCNN_CUDA_CHECK(cudaMemcpy(&cpu_im_info[0], im_info.dptr_,
src/operator/contrib/proposal.cu:    FRCNN_CUDA_CHECK(cudaMemcpy(keep, &_keep[0], sizeof(int) * _keep.size(),
src/operator/numpy/np_boolean_mask_assign.cu:    CUDA_CALL(cudaMemcpy(&valid_num, &prefix_sum[mask_size], sizeof(size_t),
src/operator/numpy/np_nonzero_op.cu:  CUDA_CALL(cudaMemcpy(&valid_num, &prefix_sum[in_size - 1], sizeof(int32_t),
src/operator/numpy/np_nonzero_op.cu:      CUDA_CALL(cudaMemcpy(out.data().dptr<int64_t>(), &temp, sizeof(int64_t),
src/operator/numpy/np_unique_op.cu:    CUDA_CALL(cudaMemcpy(&valid_num, thrust::raw_pointer_cast(&prefix_sum[input_size - 1]),
src/operator/numpy/np_unique_op.cu:    CUDA_CALL(cudaMemcpy(&valid_num, thrust::raw_pointer_cast(&prefix_sum[temp_shape[0] - 1]),
src/operator/numpy/np_unique_op.cu:      CUDA_CALL(cudaMemcpy(outputs[0].data().dptr<DType>(), inputs[0].data().dptr<DType>(),
src/operator/numpy/random/dist_common.cu:CUDA_CALL(cudaMemcpy(dst, src, sizeof(float), cudaMemcpyDeviceToHost));
src/operator/numpy/random/dist_common.cu:CUDA_CALL(cudaMemcpy(dst, src, sizeof(double), cudaMemcpyDeviceToHost));
src/operator/numpy/random/np_multinomial_op.cu:  CUDA_CALL(cudaMemcpy(&pvals_[0], input, sizeof(DType) * prob_length,
src/operator/rnn-inl.h:      CUDA_CALL(cudaMemcpy(sequence_length_cpu_itype,  sequence_length_ptr_gpu,
src/operator/tensor/cast_storage-inl.cuh:  CUDA_CALL(cudaMemcpy(&nnr, &row_flg[num_rows - 1], sizeof(dim_t), cudaMemcpyDeviceToHost));
src/operator/tensor/cast_storage-inl.cuh:        CUDA_CALL(cudaMemcpy(&nnz, &(indptr[num_rows]), sizeof(IType), cudaMemcpyDeviceToHost));
src/operator/tensor/dot-inl.cuh:          CUDA_CALL(cudaMemcpy(&nnr, nnr_ptr, nnr_bytes, cudaMemcpyDeviceToHost));
src/operator/tensor/dot-inl.cuh:            CUDA_CALL(cudaMemcpy(&nnr_out, &row_flg_out[num_cols_l-1], sizeof(dim_t),
src/operator/tensor/elemwise_binary_op_basic.cu:        CUDA_CALL(cudaMemcpy(&nnr_out, &common_row_table[num_rows-1], sizeof(nnvm::dim_t),
src/operator/tensor/indexing_op.cu:  CUDA_CALL(cudaMemcpy(&is_valid, is_valid_ptr, sizeof(char),
src/operator/tensor/indexing_op.cu:  CUDA_CALL(cudaMemcpy(&nnr, grad_row_idx + data_size, sizeof(RType),
src/operator/tensor/indexing_op.cu:        CUDA_CALL(cudaMemcpy(&nnr, &prefix_sum[num_rows-1], sizeof(dim_t),
src/operator/tensor/matrix_op.cu:        CUDA_CALL(cudaMemcpy(&nnr, &out_indptr[indptr_len-1], sizeof(RType),
src/operator/tensor/square_sum.cu:    CUDA_CALL(cudaMemcpy(&is_diff, is_diff_ptr, sizeof(int32_t), cudaMemcpyDeviceToHost));
```

eric-haibin-lin: @reminisce @haojin2 can we fix the `cudaMemcpy` calls in the numpy op? They impact GPU performance

roywei:
tested again nightly and unit test passed. ran performance test and get similar speed as in https://github.com/apache/incubator-mxnet/pull/13896

```
In [1]: import mxnet as mx
   ...: a = mx.nd.ones((10, 200, 300, 500), ctx=mx.gpu(0))
   ...: a.attach_grad()
   ...: mx.autograd.set_recording(True)
   ...: %timeit mx.nd.Dropout(a, 0.5, mode='always').wait_to_read()

[10:00:10] ../src/base.cc:84: Upgrade advisory: this mxnet has been built against cuDNN lib version 7401, which is older than the oldest version tested by CI (7600).  Set MXNET_CUDNN_LIB_CHECKING=0 to quiet this warning.
4.51 ms ± 6.66 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)```

File: src/operator/nn/dropout-inl.h (hunk 8 above)
Comment (szha) [reply src/operator/nn/dropout-inl.h]: @eric-haibin-lin let's create an issue for tracking

Comment (eric-haibin-lin) [reply src/operator/nn/dropout-inl.h]: I already created one https://github.com/apache/incubator-mxnet/issues/16583

File: tests/nightly/test_dropout.py [hunk 10]
```diff
@@ -0,0 +1,49 @@
...
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..', 'python', 'unittest'))
+from common import with_seed
+import unittest
+
+import mxnet as mx
+import numpy as np
+from mxnet.test_utils import assert_almost_equal
+from nose.tools import assert_raises
+
+@with_seed()
```
eric-haibin-lin: would you mind adding a unit test for the RNN op, too? Thx!

roywei: Added, but cpu does not work, gpu works after my PR. I think it's unrelated to my change, created a separate issue here: https://github.com/apache/incubator-mxnet/issues/16604

Workaround is manual append a dropout layer after rnn on cpu.

eric-haibin-lin:
@ptrendx any concerns on the new fix?

src/operator/rnn-inl.h
```diff
@@ -1360,15 +1360,16 @@ class RNNOp {
...
       }
       // Only update the probability by passing in a null dropout_states ptr
       DType* dropout_states = NULL;
       size_t dropout_bytes = 0;
+      // use dummy seed as state is null
       CUDNN_CALL(cudnnSetDropoutDescriptor(dropout_desc_, s->dnn_handle_,
                                            param_.p,  // discard probability
                                            dropout_states, dropout_bytes,
-                                           seed_));
+                                           0));
```
DickJC123: In the minimum, the comment should be updated.  The way these calls work is:
cudnnSetDropoutDescriptor(..., dropout_states==NULL,...)    // Set dropout probability **and seed**, leave states alone.
cudnnSetDropoutDescriptor(..., dropout_states!=NULL,...)    // Set dropout probability and seed, init states based on these values.

cudnnRestoreDropoutDescriptor()    // Set dropout probability, seed and states ptr from provided args.

roywei: Ok, and I found another issue, the seed will be fixed during resource initialization. So once if dropout layer is created, event if we don't fix  seed, each dropout result will be the same.

If we want each forward to have a different result, and the random result must respect mxnet random seed. The only solution is to get mxnet seed during each forward. Is that right?

roywei: new test case
```
@with_seed()
def test_dropout_with_seed():
    info = np.iinfo(np.int32)
    seed = np.random.randint(info.min, info.max)
    _test_dropout(mx.cpu(), seed)
    _test_dropout(mx.gpu(), seed)
    _test_dropout(mx.cpu())
    _test_dropout(mx.gpu())

def _test_dropout(ctx, seed=None):
    data = mx.nd.ones((100, 100), ctx=ctx)
    dropout = mx.gluon.nn.Dropout(0.5)

    if seed:
        mx.random.seed(seed)
    with mx.autograd.record():
        result1 = dropout(data)

    if seed:
        mx.random.seed(seed)
    with mx.autograd.record():
        result2 = dropout(data)
    if seed:
        # dropout on gpu should return same result with fixed seed
        assert_almost_equal(result1.asnumpy(), result2.asnumpy())
    else:
        # seed not fixed, result should be different
        with assert_raises(AssertionError):
            assert_almost_equal(result1.asnumpy(), result2.asnumpy())
```

ptrendx: Dick is looking into it, but I don't think sentence `event if we don't fix seed, each dropout result will be the same` is true. What is true though, is the fact that the dropout as is implemented now would not respond to the change of seed on the MXNet side after the initial creation of the op. Seeding each forward would destroy performance, so how about a solution like this - if you know that you will cache the value of seed (like in the dropout descriptor resource case), every time when you get the descriptor it should internally ask the random resource "did a user set a new seed?" (which could be implemented as a set in the random resource that would keep track of who asked about this already and resetting that set when user calls to `mxnet.random.seed`). If the answer is "no", then no reseeding is required, but if the answer is yes, the dropout descriptor should be reseeded with a new value.

eric-haibin-lin: echoing @ptrendx 's comment, we should implement SetSeed() for the cudnn resource in  https://github.com/apache/incubator-mxnet/blob/master/src/resource.cc#L174-L201

DickJC123:
@roywei I've been playing with a variant of your proposed test in which I set the seed to two different values for the two models and expect the results to be different.  This fails, because the results are identical even with the differing seeds.  The two models each get their own gpu random resource, but the two are seeded by cpu random number generators that are identical.

The problem here is that the cpu rng's are not responding to mx.random.seed(), and instead have their seed set to 0.  The reason is that cpu rngs are requested from the ResourceManager, and ResourceManager is a thread-local variable.  The main python thread (performing the mx.random.seed()) only affects the **global_seed_** data member for its ResourceManager instance, which does not affect the seeds of the cpu rngs that are requested of the worker thread's ResourceManager.

DickJC123:
Yeah, this is going to be tricky to fix solidly, particularly considering models with multiple dropouts and rnns.

If all the gpu rngs of a worker indeed share the same rng state, then only the seed from the first rng in a model effects the initialization (the other seeds are ignored).  Also, I believe it is non-deterministic which GPU worker handles an operator during execution, so this represents a problem (could be the cause of some low-probability test failures).

I believe moving the gpu rng state to be a resource was motivated by the high initialization overhead, particularly painful for imperative models.  Would it be possible to add a seed argument at the python level to dropout and rnn operators, with the understanding that by setting the seed, the operator will get its own rng state (at some memory and initialization expense)?  Not setting a seed would grab the global resource- not sure how the determinism would work, but it would be fast.

roywei:
closing in favor of https://github.com/apache/incubator-mxnet/pull/17547

PR #16540, opened 2019-10-19
fix pylint in CI

roywei:
## Description ##

https://github.com/apache/incubator-mxnet/issues/16541

temporary disable pylint check on tools/caffe_converter/*.py
as no change was made but it kept failing for PRs.


roywei:
cc @reminisce

PR #16553, opened 2019-10-20
Fix for wrong reqs set after switching from training to inference

ptrendx:
## Description ##
`StaticAllocMemory` in `CachedOp` used `storage_inplace_index` attribute (generated by `MXPlanMemory` graph pass) to assign reqs for edges in the graph. However, the `MXPlanMemory` pass is called only once per the type of memory plan (full, forward or backward). While the memory plan itself is stored in separate attributes in the graph (`forward_mem_plan`, `full_mem_plan` etc.), the `storage_inplace_index` (and so the reqs) were overwritten by the last called `MXPlanMemory`.

The following code:

```
with mx.autograd.record():
   result = net(x)
result.backward()

result2 = net(x)

with mx.autograd.record():
   result3 = net(x)
result3.backward()
```

calls first the plan memory for the full graph and then for just the forward graph. The third invocation to `net` does not invoke a plan memory pass. Let us assume that inside `net` is an op that produces the output needed for the backward pass only when needed (req not set to `kNullOp`). Since the reqs are overwritten by the second `net` invocation, that output's req is set to `kNullOp` (because there is no backward pass there). Then the 3rd invocation to `net` does not change the req value and so the op does not produces the required output - `result3` gradient is therefore produced using the stale values and so is wrong.

This PR fixes it by changing the `StaticAllocMemory` to use per mem plan values to assign reqs (`storage_inplace_index_forward` etc.) to keep the benefits of caching (`MXPlanMemory` called once per type) while ensuring correctness.

@eric-haibin-lin 

### Essentials ###

### DIFF
--- a/src/imperative/cached_op.cc
+++ b/src/imperative/cached_op.cc
@@ -32,6 +32,22 @@ DMLC_REGISTER_PARAMETER(CachedOpConfig);
 
 constexpr uint32_t kEidNotExist = std::numeric_limits<uint32_t>::max();
 
+const std::string CachedOp::FULL = "full";
+const std::string CachedOp::FORWARD = "forward";
+const std::string CachedOp::BACKWARD = "backward";
+const std::string CachedOp::REF_COUNT = "ref_count";
+const std::string CachedOp::MEM_PLAN = "mem_plan";
+const std::string CachedOp::STORAGE_PLAN = "storage_plan";
+
+namespace {
+
+std::string AddPrefix(const std::string& prefix,
+                      const std::string& s) {
+  return prefix + "_" + s;
+}
+
+}  // namespace
+
 struct CachedOp::GraphInfo {
   nnvm::Graph fwd_graph;
   nnvm::Graph full_graph;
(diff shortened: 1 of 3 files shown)


szha:
As follow-up:
1. ideally we should be able to test for this explicitly. given that we lack the utility for such testing, I want to avoid blocking this fix on this basis
2. the keys/magic strings should ideally be centralized, maybe in an enum.

ptrendx:
For 1: this error was found when working on pointwise fusion, which uses reqs to compile the right code. I don't know if there is any other operator which may not write to an output if there is no backward pass for it (I know `mx.sym.contrib.box_nms` would have that behavior, but the current implementation writes to the output only used for backward every time, without looking at the req value). I can add the test in pointwise fusion PR, show that it fails and then merge with master once this fix is in to show that it does not fail anymore.

For 2: I agree, probably as `const static` members of `CachedOp`?

szha:
sounds good to both

ptrendx:
@eric-haibin-lin Yes, it had this bug since the beginning I believe, but I don't think there were any ops that would expose it until now.

ptrendx:
Ok, I added the test to #15167 that should fail because of this issue. I will move the string constants to `const static` members of cached op, merge this PR and then merge master to #15167 to show it fixes the problem.

ptrendx:
As expected, test failed in the new test here: http://jenkins.mxnet-ci.amazon-ml.com/job/mxnet-validation/job/centos-gpu/job/PR-15167/50/display/redirect

PR #16570, opened 2019-10-21
Showing proper error message when an attempt is made to create large tensor but MXNet is not built with it

access2rohit:
## Description ##
Throwing error message to customers that they need to build MXNet with flag USE_INT64_TENSOR_SIZE=1 when user attempts to create large NDarrays for e.g.
- nd.array()
- nd.ones()
- nd.zeros()
- nd.full()
- nd.random() etc.

### Essentials ###

### Testing ###
```
ubuntu@ip-172-31-82-110 ~/incubator-mxnet (err_msg_int32) $ MXNET_TEST_COUNT=1 nosetests --logging-level=DEBUG --verbose -s tests/python/unittest/test_operator.py:test_large_tensor_disabled_err_msg
/home/ubuntu/anaconda3/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from 'float' to 'np.floating' is deprecated. In future, it will be treated as 'np.float64 == np.dtype(float).type'.
[INFO] Setting module np/mx/python random seeds, use MXNET_MODULE_SEED=721807427 to reproduce.
test_operator.test_large_tensor_disabled_err_msg ... ok

Ran 1 test in 0.042s

OK
```

### DIFF
--- a/python/mxnet/ndarray/ndarray.py
+++ b/python/mxnet/ndarray/ndarray.py
@@ -155,6 +155,11 @@ def _new_alloc_handle(shape, ctx, delay_alloc, dtype=mx_real_t):
             ctypes.c_int(int(_DTYPE_NP_TO_MX[np.dtype(dtype).type])),
             ctypes.byref(hdl)))
     else:
+        size = 1
+        for idx in shape:
+            size = size * idx
+        if size > 2**32:
+            print("Size of tensor you are trying to allocate is larger than 2^32 elements. Please build with flag USE_INT64_TENSOR_SIZE=1")
         check_call(_LIB.MXNDArrayCreateEx(
             c_array_buf(mx_uint, native_array('I', shape)),
             mx_uint(len(shape)),
--- a/python/mxnet/ndarray/ndarray.py
+++ b/python/mxnet/ndarray/ndarray.py
@@ -158,9 +158,10 @@ def _new_alloc_handle(shape, ctx, delay_alloc, dtype=mx_real_t):
         size = 1
         for idx in shape:
             size = size * idx
-        if size > 2**32:
-            raise Exception("[Python] Size of tensor you are trying to allocate is larger than 2^32 elements. " +
-                  "Please build with flag USE_INT64_TENSOR_SIZE=1")
+        if size > 2**31:
+            raise Exception("[_new_alloc_handle] Size of tensor you are trying to allocate is " +
+                            "larger than 2^31 elements. Please build with flag " +
+                            "USE_INT64_TENSOR_SIZE=1")
         check_call(_LIB.MXNDArrayCreateEx(
             c_array_buf(mx_uint, native_array('I', shape)),
             mx_uint(len(shape)),
--- a/python/mxnet/ndarray/ndarray.py
+++ b/python/mxnet/ndarray/ndarray.py
@@ -159,7 +159,8 @@ def _new_alloc_handle(shape, ctx, delay_alloc, dtype=mx_real_t):
         for idx in shape:
             size = size * idx
         if size > 2**32:
-            print("Size of tensor you are trying to allocate is larger than 2^32 elements. Please build with flag USE_INT64_TENSOR_SIZE=1")
+            raise Exception("[Python] Size of tensor you are trying to allocate is larger than 2^32 elements. " +
+                  "Please build with flag USE_INT64_TENSOR_SIZE=1")
         check_call(_LIB.MXNDArrayCreateEx(
             c_array_buf(mx_uint, native_array('I', shape)),
             mx_uint(len(shape)),
--- a/python/mxnet/ndarray/ndarray.py
+++ b/python/mxnet/ndarray/ndarray.py
@@ -58,6 +58,7 @@ _STORAGE_TYPE_UNDEFINED = -1
 _STORAGE_TYPE_DEFAULT = 0
 _STORAGE_TYPE_ROW_SPARSE = 1
 _STORAGE_TYPE_CSR = 2
+_SIGNED_INT32_UPPER_LIMIT = (2**31 - 1)
 
 # pylint: disable= no-member
 _DTYPE_NP_TO_MX = {
@@ -155,10 +156,12 @@ def _new_alloc_handle(shape, ctx, delay_alloc, dtype=mx_real_t):
             ctypes.c_int(int(_DTYPE_NP_TO_MX[np.dtype(dtype).type])),
             ctypes.byref(hdl)))
     else:
+        # When shape is larger than unit32 then there is an overflow error at python end itself.
+        # It needs to be caught here since the call doesn't even reach backend.
         size = 1
         for idx in shape:
             size = size * idx
-        if size > 2**31:
+        if size > _SIGNED_INT32_UPPER_LIMIT:
             raise Exception("[_new_alloc_handle] Size of tensor you are trying to allocate is " +
                             "larger than 2^31 elements. Please build with flag " +
                             "USE_INT64_TENSOR_SIZE=1")
--- a/src/operator/tensor/init_op.h
+++ b/src/operator/tensor/init_op.h
@@ -272,6 +272,11 @@ inline bool InitShape(const nnvm::NodeAttrs& attrs,
   CHECK_EQ(in_attrs->size(), 0U);
   CHECK_EQ(out_attrs->size(), 1U);
   mxnet::TShape param_shape = param.shape;
+  if (!features::is_enabled(features::INT64_TENSOR_SIZE)) {
+    CHECK_LT(param_shape.Size(), (int64_t{1} << 32) - 1) <<
+              "Size of tensor you are trying to allocate is larger than "
+              "2^32 elements. Please build with flag USE_INT64_TENSOR_SIZE=1";
+  }
   if (!Imperative::Get()->is_np_shape()) {
     common::ConvertToNumpyShape(&param_shape);
   }
@@ -336,8 +341,11 @@ inline bool InitStorageType(const nnvm::NodeAttrs& attrs,
 template <bool is_integer = false, typename ValueType, typename xpu>
 void Fill(mshadow::Stream<xpu> *s, const TBlob& b, const OpReqType req, ValueType val) {
   // If b is a zero-size tensor, do nothing.
-  CHECK_LT(b.Size(), (uint32_t{1} << 31) - 1) << "Size of tensor you are trying to allocate is larger than "
-                                 "2^32 elements. Please build with flag USE_INT64_TENSOR_SIZE=1";
+  if (!features::is_enabled(features::INT64_TENSOR_SIZE)) {
+    CHECK_LT(b.Size(), (int64_t{1} << 32) - 1) <<
+              "Size of tensor you are trying to allocate is larger than "
+              "2^32 elements. Please build with flag USE_INT64_TENSOR_SIZE=1";
+  }
   if (b.Size() == 0) return;
   if (req != kNullOp) {
     const size_t size = b.Size();
@@ -582,7 +590,13 @@ inline bool RangeShape(const nnvm::NodeAttrs& attrs,
   }
   const double out_size = std::ceil((param.stop.value() - param.start) / param.step)
                           * param.repeat;
-  SHAPE_ASSIGN_CHECK(*out_attrs, 0, mxnet::TShape({static_cast<nnvm::dim_t>(out_size)}));
+  mxnet::TShape output_shape = mxnet::TShape({static_cast<nnvm::dim_t>(out_size)});
+  if (!features::is_enabled(features::INT64_TENSOR_SIZE)) {
+    CHECK_LT(output_shape.Size(), (int64_t{1} << 32) - 1) <<
+              "Size of tensor you are trying to allocate is larger than "
+              "2^32 elements. Please build with flag USE_INT64_TENSOR_SIZE=1";
+  }
+  SHAPE_ASSIGN_CHECK(*out_attrs, 0, output_shape);
   return true;
 }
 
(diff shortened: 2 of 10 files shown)


access2rohit:

python/mxnet/ndarray/ndarray.py
```diff
@@ -155,6 +155,11 @@ def _new_alloc_handle(shape, ctx, delay_alloc, dtype=mx_real_t):
             ctypes.c_int(int(_DTYPE_NP_TO_MX[np.dtype(dtype).type])),
             ctypes.byref(hdl)))
     else:
+        size = 1
+        for idx in shape:
+            size = size * idx
+        if size > 2**32:
+            print("Size of tensor you are trying to allocate is larger than 2^32 elements. Please build with flag USE_INT64_TENSOR_SIZE=1")
```
marcoabreu: Error?!

access2rohit: @marcoabreu how to you handle errors in python end? I was looking for existing implementations. Can you point out one ?

marcoabreu: Generally I'd say ArgumentException, but you can also let the backend handle throwing an error

python/mxnet/ndarray/ndarray.py
```diff
@@ -155,6 +155,11 @@ def _new_alloc_handle(shape, ctx, delay_alloc, dtype=mx_real_t):
             ctypes.c_int(int(_DTYPE_NP_TO_MX[np.dtype(dtype).type])),
             ctypes.byref(hdl)))
     else:
+        size = 1
+        for idx in shape:
+            size = size * idx
+        if size > 2**32:
```
leezu: This also happens for `size < 2**32`. For example with the matrix of shape `(18545452, 200)` in https://github.com/dmlc/gluon-nlp/issues/981.

Please make the check more accurate

python/mxnet/ndarray/ndarray.py
```diff
@@ -155,6 +155,13 @@ def _new_alloc_handle(shape, ctx, delay_alloc, dtype=mx_real_t):
             ctypes.c_int(int(_DTYPE_NP_TO_MX[np.dtype(dtype).type])),
             ctypes.byref(hdl)))
     else:
+        size = 1
+        for idx in shape:
+            size = size * idx
+        if size > 2**31:
+            raise Exception("[_new_alloc_handle] Size of tensor you are trying to allocate is " +
```
marcoabreu: Shall we not just let the backend handle that logic and bubble up the exception? The frontend shouldn't care about these kinds of things I'd say

marcoabreu: Feel free to let me know if you'd like me to elaborate

access2rohit: actually if you look at the C API call below it is passing shape as `c_array_buf(mx_uint, native_array('I', shape)),` so the code i a lot of cases break here it self saying that integer value passed overflows unit32 range. Therefore, a check here is unavoidable. Let me know if this makes sense. If you have a suggestion to make this check better, I am happy to hear it :)

marcoabreu: Makes sense, thanks for elaborating. Would you mind adding a comment?

access2rohit:
> Would it be possible to also add some tests?

@marcoabreu I definitely will. But I need to make sure the functionality works fine first.

marcoabreu:
Sure, sorry for asking prematurely

src/operator/tensor/init_op.h
```diff
@@ -272,10 +272,22 @@ inline bool InitShape(const nnvm::NodeAttrs& attrs,
   CHECK_EQ(in_attrs->size(), 0U);
   CHECK_EQ(out_attrs->size(), 1U);
   mxnet::TShape param_shape = param.shape;
+  if (shape_is_known(param_shape) && !features::is_enabled(features::INT64_TENSOR_SIZE)) {
```
apeforest: Why not directly check the compiler flag here?

marcoabreu: I personally prefer that way - looks cleaner :)

python/mxnet/ndarray/ndarray.py
apeforest: As proposed by Frank in the design review, it would be better to check int64 related condition inside C API, right? I agree with Marco, we should minimize the backend logic checking in frontend.

access2rohit: @apeforest the call will fail here itself. It won't even go to backend as stated in my comment above: https://github.com/apache/incubator-mxnet/pull/16570/files#r338846832

src/operator/tensor/init_op.h
access2rohit: in certain cases shape is not know at the beginning, then param_shape.Size() throws a different error, which is not a good user experience. That's why this check is required.

access2rohit: @apeforest @marcoabreu do you think this makes sense now ?

python/mxnet/ndarray/ndarray.py
access2rohit: > Makes sense, thanks for elaborating. Would you mind adding a comment?

Sure !

marcoabreu: Didn't we have 32 and 64 bit C api functions?

It sounds counter intuitive initially, but if we have overloaded c api functions we could always expose them. If we're compiled without 64bit Support, the c api would then return an error instead of passing on the request. IMO this design makes the typing a bit more explicit and transparent (while I have to admit that I generally dislike the verbose explicity in c++) instead of hiding all the magic in a simple typedef override and some preprocessor statements.

The logic in the frontend would then be to call the 32bit c api functions with small tensors and 64bit one with large tensors. Of course it needs some tweaking because that's not a fail proof design against overflow of the intermediary results.

What it seems like to me is that a C Api is supposed to be a static contract that is always present. But with the 64bit Support we're now introducing versions of the c api that depend on the compile options. This then breaks the exchangability and decoupling. In the end it's a matter of taste, but I think we should uphold ourselves to a standard where the c api never changes but the backend might behave differently depending on the chosen options. At the same time, we might want to make it a rule that c api function definitions must not be conditional. 

It's a bit derailing, but what do you think about that idea?

python/mxnet/ndarray/ndarray.py
access2rohit: @leezu I have fixed that. Please see the output below and verify.
`
ubuntu@ip-172-31-82-110 ~/fast_text $ python code.py
/home/ubuntu/anaconda3/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  File "code.py", line 4, in <module>
    model = nlp.model.train.FasttextEmbeddingModel.load_fasttext_format(wordvectors)
  File "/home/ubuntu/anaconda3/lib/python3.6/site-packages/gluonnlp/model/train/embedding.py", line 277, in load_fasttext_format
    self.initialize(ctx=ctx)
  File "/home/ubuntu/incubator-mxnet/python/mxnet/gluon/block.py", line 632, in initialize
    self.collect_params().initialize(init, ctx, verbose, force_reinit)
  File "/home/ubuntu/incubator-mxnet/python/mxnet/gluon/parameter.py", line 887, in initialize
    v.initialize(None, ctx, init, force_reinit=force_reinit)
  File "/home/ubuntu/incubator-mxnet/python/mxnet/gluon/parameter.py", line 464, in initialize
    self._finish_deferred_init()
  File "/home/ubuntu/incubator-mxnet/python/mxnet/gluon/parameter.py", line 344, in _finish_deferred_init
    data = zeros_fn(**kwargs)
  File "/home/ubuntu/incubator-mxnet/python/mxnet/ndarray/utils.py", line 67, in zeros
    return _zeros_ndarray(shape, ctx, dtype, **kwargs)
  File "/home/ubuntu/incubator-mxnet/python/mxnet/ndarray/ndarray.py", line 4641, in zeros
    return _internal._zeros(shape=shape, ctx=ctx, dtype=dtype, **kwargs)
  File "<string>", line 36, in _zeros
  File "/home/ubuntu/incubator-mxnet/python/mxnet/_ctypes/ndarray.py", line 107, in _imperative_invoke
    ctypes.byref(out_stypes)))
  File "/home/ubuntu/incubator-mxnet/python/mxnet/base.py", line 254, in check_call
mxnet.base.MXNetError: [00:25:17] src/operator/numpy/../tensor/init_op.h:276: Check failed: param_shape.Size() < (int64_t{1} << 31) - 1 (3709090400 vs. 2147483647) : [InitShape-input] Size of tensor you are trying to allocate is larger than 2^31 elements. Please build with flag USE_INT64_TENSOR_SIZE=1
Stack trace:
  [bt] (0) /home/ubuntu/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x34) [0x7efc0bd7df44]
  [bt] (1) /home/ubuntu/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(bool mxnet::op::InitShape<mxnet::op::InitOpParam>(nnvm::NodeAttrs const&, std::vector<mxnet::TShape, std::allocator<mxnet::TShape> >*, std::vector<mxnet::TShape, std::allocator<mxnet::TShape> >*)+0x339) [0x7efc0ca83045]
  [bt] (2) /home/ubuntu/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(std::_Function_handler<bool (nnvm::NodeAttrs const&, std::vector<mxnet::TShape, std::allocator<mxnet::TShape> >*, std::vector<mxnet::TShape, std::allocator<mxnet::TShape> >*), bool (*)(nnvm::NodeAttrs const&, std::vector<mxnet::TShape, std::allocator<mxnet::TShape> >*, std::vector<mxnet::TShape, std::allocator<mxnet::TShape> >*)>::_M_invoke(std::_Any_data const&, nnvm::NodeAttrs const&, std::vector<mxnet::TShape, std::allocator<mxnet::TShape> >*&&, std::vector<mxnet::TShape, std::allocator<mxnet::TShape> >*&&)+0x61) [0x7efc0bd96b9b]
  [bt] (3) /home/ubuntu/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(std::function<bool (nnvm::NodeAttrs const&, std::vector<mxnet::TShape, std::allocator<mxnet::TShape> >*, std::vector<mxnet::TShape, std::allocator<mxnet::TShape> >*)>::operator()(nnvm::NodeAttrs const&, std::vector<mxnet::TShape, std::allocator<mxnet::TShape> >*, std::vector<mxnet::TShape, std::allocator<mxnet::TShape> >*) const+0x76) [0x7efc1053c68e]
  [bt] (4) /home/ubuntu/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(mxnet::imperative::SetShapeType(mxnet::Context const&, nnvm::NodeAttrs const&, std::vector<mxnet::NDArray*, std::allocator<mxnet::NDArray*> > const&, std::vector<mxnet::NDArray*, std::allocator<mxnet::NDArray*> > const&, mxnet::DispatchMode*)+0x4c5) [0x7efc106a42b2]
  [bt] (5) /home/ubuntu/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(mxnet::Imperative::Invoke(mxnet::Context const&, nnvm::NodeAttrs const&, std::vector<mxnet::NDArray*, std::allocator<mxnet::NDArray*> > const&, std::vector<mxnet::NDArray*, std::allocator<mxnet::NDArray*> > const&)+0x285) [0x7efc106aadf9]
  [bt] (6) /home/ubuntu/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(MXImperativeInvokeImpl(void*, int, void**, int*, void***, int, char const**, char const**)+0x195) [0x7efc11007089]
  [bt] (7) /home/ubuntu/incubator-mxnet/python/mxnet/../../lib/libmxnet.so(MXImperativeInvokeEx+0x87) [0x7efc110073fa]
  [bt] (8) /home/ubuntu/anaconda3/lib/python3.6/lib-dynload/../../libffi.so.6(ffi_call_unix64+0x4c) [0x7efc1a86eec0]
`

access2rohit:

access2rohit:
@apeforest @ChaiBapchya @zheng-da This PR is ready for review

python/mxnet/ndarray/ndarray.py
access2rohit: `The logic in the frontend would then be to call the 32bit c api functions with small tensors and 64bit one with large tensors. Of course it needs some tweaking because that's not a fail proof design against overflow of the intermediary results.`

A: Actually its not that straight, the operator code still needs to use 32 bit and 64 bit versions. Since the performance of both versions of operators in not the same using 64bit version of both versions of C API is not the ideal choice.

`But with the 64bit Support we're now introducing versions of the c api that depend on the compile options. This then breaks the exchangability and decoupling. In the end it's a matter of taste, but I think we should uphold ourselves to a standard where the c api never changes but the backend might behave differently depending on the chosen options`

A: This is possible. We can have both 32 bit and 64 bit version of all the operators in the the code. This would require duplicate versions of same code which involves a lot of inline and templatized code. This can be used to dynamically switch between 32bit and 64bit versions of the operators keeping all the C APIs as 64bit. This was considered earlier but as you yourself already know have pointed out in many PRs it would lead to blowing up of the generated binaries `libmxnet.so` so I decided not to move ahead with this approach.

Hope this addresses your concerns

python/mxnet/ndarray/ndarray.py
```diff
@@ -155,6 +155,15 @@ def _new_alloc_handle(shape, ctx, delay_alloc, dtype=mx_real_t):
             ctypes.c_int(int(_DTYPE_NP_TO_MX[np.dtype(dtype).type])),
             ctypes.byref(hdl)))
     else:
+        # When shape is larger than unit32 then there is an overflow error at python end itself.
+        # It needs to be caught here since the call doesn't even reach backend.
+        size = 1
+        for idx in shape:
+            size = size * idx
+        if size > 2**31:
```
apeforest: Can we use 2**31 as a constant instead of expression here to save some runtime?

PR #16579, opened 2019-10-22
add retry to S3 uploads for test results

larroy:
See title.

https://github.com/apache/incubator-mxnet/pull/16512

### DIFF
--- a/ci/Jenkinsfile_utils.groovy
+++ b/ci/Jenkinsfile_utils.groovy
@@ -134,11 +134,13 @@ def collect_test_results_unix(original_file_name, new_file_name) {
         // Thus, we have to pick a name manually and rename the files so that they can be stored separately.
         sh 'cp ' + original_file_name + ' ' + new_file_name
         archiveArtifacts artifacts: new_file_name
-        try {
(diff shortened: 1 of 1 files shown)


marcoabreu:
Good idea. Will this rethrow the original exception if the retry limit is reached?

ChaiBapchya:
Nice idea. But do we know if it will actually upload the xml within 3 attempts? Coz I have seen S3 upload fail many times when some test fails. S3upload is independent of test failure. But just to be sure.

larroy:
@marcoabreu I don't know. Looks like people are using it assuming it handle exceptions: https://gist.github.com/daipresents/c7f982ffce9c1bda73b4387d6b1ae3da
Do you know better?

@ChaiBapchya ok, what would you suggest instead?

marcoabreu:
Seems like it's handling in exactly that fashion.

marcoabreu:
Could you also add some diagnostics to debug the error? The execution in the mentioned PR said success but then had the return code 1. Seems odd

ChaiBapchya:
> 
> @ChaiBapchya ok, what would you suggest instead?

No my point was have you by any chance tested that if PR fails on a particular build, does it still upload on S3 or does it fail everytime build fails? But nevermind, it's tough to check that. (I tried doing it myself but its like searching for needle in haystack.

larroy:
I wouldn't think it uploads to s3 if the PR fails.

PR #16582, opened 2019-10-22
[Gluon] Don't serialize shared parameters twice

leezu:
## Description ##
`Block.save_parameters` currently stores a separate copy of shared parameters for sub-block which uses the shared parameter. This is wasteful. Instead only store a single copy.
This change is backwards compatible; all existing valid serialized files continue to work. However, newly serialized files will be smaller in case they contain shared parameters.

### Essentials ###

### Changes ###
- [X] Don't serialize shared parameters twice in Gluon `Block.save_parameters`

## Comments ##
@szha @eric-haibin-lin

### DIFF
--- a/python/mxnet/gluon/block.py
+++ b/python/mxnet/gluon/block.py
@@ -527,7 +527,7 @@ class Block(object):
 
         if not allow_missing:
             # Shared parameters are stored only a single time as of MXNet 1.6.
-            # We thus retrieve all prefixes (à la _collect_params_with_prefix)
+            # We thus retrieve all prefixes (through _collect_params_with_prefix)
             # that a shared parameter is used with. Check that there are no
             # missing parameters that were not yet already loaded from the
             # shared version.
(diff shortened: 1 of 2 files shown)


python/mxnet/gluon/block.py
```diff
@@ -510,15 +524,26 @@ def load_parameters(self, filename, ctx=None, allow_missing=False,
 
         if not any('.' in i for i in loaded.keys()):
             # legacy loading
-            del loaded
+            loaded = None
+            # We cannot `del loaded` as it is used in a generator expression
+            # below. `del loaded` would be a SyntaxError in Python 2.
```
szha: this comment is confusing as it comments on code that doesn't exist.

in general, comments should be minimal, and concise, and only out of necessity. code should be self-documenting.

leezu:
Addressed comments

python/mxnet/gluon/block.py
leezu: It's non-obvious that `del loaded` is a SyntaxError in Python 2. But I explained the explanation as it is confusing. Rather let's just instruct future readers to replace the instruction with `del loaded` once Python 2 support is gone.

PR #16497, opened 2019-10-15
Large Vector tests for DGL Ops Part 2

ChaiBapchya:
## Description ##
add hyperbolic, logical, sign and regression tests for large vector

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- [x] Code is well-documented: 
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] Add tests to test_large_array.py
- [x] Fix issue with test_concat introduced in PR #15960

PR #16585, opened 2019-10-23
C Api for simplebind, fix comment for trigoops, add atol to assert

ChaiBapchya:
## Description ##
Simple_bind (used in Symbolic mode) used to fail for Large tensor (esp when individual dimension was of size > `2**32`

PR https://github.com/apache/incubator-mxnet/pull/16497 introduces large vector tests 
test_regression for Logistic and Linear regression has been written for symbolic mode
And it fails otherwise. This PR will fix it.

### Essentials ###

### Changes ###
- [x] M       include/mxnet/c_api.h
- [x] M       python/mxnet/symbol/symbol.py
- [x] M       src/c_api/c_api_executor.cc
- [x] M       tests/nightly/test_large_array.py


src/c_api/c_api_executor.cc
```diff
@@ -849,6 +816,152 @@ int MXExecutorSimpleBindEx(SymbolHandle symbol_handle,
...
+                           const char** provided_arg_stype_names,
+                           const int* provided_arg_stypes,
+                           const uint32_t num_shared_arg_names,
+                           const char** shared_arg_name_list,
+                           int* shared_buffer_len,
+                           const char** shared_buffer_name_list,
+                           NDArrayHandle* shared_buffer_handle_list,
+                           const char*** updated_shared_buffer_name_list,
+                           NDArrayHandle** updated_shared_buffer_handle_list,
+                           uint32_t* num_in_args,
```
anirudh2290: please add api doc

src/c_api/c_api_executor.cc
```diff
@@ -515,44 +515,11 @@ int MXExecutorSimpleBind(SymbolHandle symbol_handle,
...
- * \param aux_states list_auxiliary_states associated with the current executor
- * \param shared_exec_handle shared excutor handle passed from _bind_ith_exec
- * \param out the handle of the executor to be created
- */
-int MXExecutorSimpleBindEx(SymbolHandle symbol_handle,
+
+namespace mxnet {
+
+template<typename DType>
+int SimpleBindExMaster(SymbolHandle symbol_handle,
```
anirudh2290: internal functions in C API generally begin with _

ChaiBapchya: also @access2rohit suggested renaming it with SimpleBindExImpl
So combining both suggestions - function name should be `_SimpleBindExImpl` ?

anirudh2290: _SimpleBindImpl

src/c_api/c_api_executor.cc
access2rohit: @anirudh2290 shouldn't they be same for both 64 and w/o 64 just like in c_api.h ?

anirudh2290: yes its still good to add it, in case some params change in future. Also one line about what the difference is would be good.

access2rohit: It makes sense to add a line in general about APIs with 64 and w/o 64 appended. But if we do it for all its just duplication of documentation, which will lead to increased no. of lines in file. IMO we should simply add a comment to the top of the file saying APIs with 64 appended to them are for large tensor support. @anirudh2290 do you agree? Or do you still want duplicate documentation to be added to each API signature ?

ChaiBapchya: i didn't add APIdoc as if would have been same as MXExecutorSimpleBindEx

ChaiBapchya:
```
nosetests tests/nightly/test_large_vector.py:test_regression
.
Ran 1 test in 745.939s
```

src/c_api/c_api_executor.cc
anirudh2290: As mentioned offline, I would prefer documentation: one reason is for frontend developers to know what these APIs do, by looking at them in isolation. Also, for docs generated here: https://mxnet.apache.org/api/cpp/docs/api/include_2mxnet_2c__api_8h.html#acecaa93405fc612183fc8780e6d5ae74

ChaiBapchya: I have added TODO for myself for adding api docs whereever int64 specific is used. I will shortly send out PR for the same. (Missed ur comment before triggering the CI hence not committing here again)

ChaiBapchya:
Also addressed the suggestions by @anirudhacharya in #16416

tests/nightly/test_large_array.py
```diff
@@ -1295,17 +1295,17 @@ def check_trunc():
 
 
 def create_input_for_trigonometric_ops(vals):
-    # Creates large vector input of size(LARGE_X*10, SMALL_Y/10) from vals using tile operator
+    # Creates large vector input of size(LARGE_X*10, SMALL_Y/10) from vals using broadcast_to operator
```
access2rohit: @anirudhacharya

tests/nightly/test_large_vector.py
```diff
@@ -710,6 +710,39 @@ def test_full():
...
 
+def test_regression():
+    shape = (LARGE_X, )
+
+    def check_regression(symbol, forward, shape):
+        # init executor
+        data_s = mx.symbol.Variable('data')
+        label_s = mx.symbol.Variable('label')
+        out_s = symbol(data=data_s, label=label_s)
+        exe = out_s.simple_bind(ctx=mx.cpu(0), data=shape, label=shape)
```
anirudh2290: what about infer shape. is it tested with bigger sizes ?

ChaiBapchya: haven't checked infer shape specifically. but we are checking output of regressions with numpy outputs for entire vector (so i am guessing shape check indirectly happens)
i can add an explicit check for shape if it is really needed.

anirudh2290: no i mean calling sym.infer_shape with large sizes. do we have a test for that ? if not we should add it. many users use this api

ChaiBapchya: oh we don't have test for it. adding now.

ChaiBapchya: Added the test.

anirudh2290: cool, does it work ?

ChaiBapchya:
```
nosetests tests/nightly/test_large_vector.py:test_infer_shape
.
Ran 1 test in 0.001s

OK
```

Issue #16576, opened 2019-10-21
[Numpy][Bug] einsum bug


sxjscience:
Example:
```python
import mxnet as mx
import numpy as np
import numpy.testing as npt
from mxnet.gluon import HybridBlock



class MultiNDimBatchDot1(HybridBlock):
    def hybrid_forward(self, F, lhs, rhs):
        """

        Parameters
        ----------
        F
        lhs :
            Shape (N0, N1, T0, C)
        rhs :
            Shape (N0, N1, T1, C)

        Returns
        -------
        ret :
            Shape (N0, N1, T0, T1)
        """
        return F.batch_dot(F.reshape(lhs, (-3, 0, 0)),
                           F.reshape(rhs, (-3, 0, 0)), transpose_b=True).reshape((-4, -1, 8, 0, 0))


class MultiNDimBatchDot2(HybridBlock):
    def hybrid_forward(self, F, lhs, rhs):
        """

        Parameters
        ----------
        F
        lhs :
            Shape (N0, N1, T0, C)
        rhs :
            Shape (N0, N1, T1, C)

        Returns
        -------
        ret :
            Shape (N0, N1, T0, T1)
        """
        return F.np.einsum('abiz,abjz->abij', lhs, rhs)


batch_dot1 = MultiNDimBatchDot1()


batch_dot1.hybridize()

lhs = mx.np.array(np.random.normal(0, 1, (64, 8, 128, 512)), dtype=np.float32, ctx=mx.cpu())
rhs = mx.np.array(np.random.normal(0, 1, (64, 8, 128, 512)), dtype=np.float32, ctx=mx.cpu())
mx.npx.waitall()

gt = np.einsum('abiz,abjz->abij', lhs.asnumpy(), rhs.asnumpy())
out = batch_dot1(lhs.as_nd_ndarray(), rhs.as_nd_ndarray())

npt.assert_allclose(gt, out.asnumpy(), rtol=1E-3, atol=1E-3)


mx.npx.set_np()
batch_dot2 = MultiNDimBatchDot2()
batch_dot2.hybridize()

out2 = batch_dot2(lhs, rhs)
npt.assert_allclose(gt, out2.asnumpy(), rtol=1E-3, atol=1E-3)
```

The CPU results are not correct.

Also, if we use the gpu, it will trigger a segfault:

```python
import mxnet as mx
import numpy as np
import numpy.testing as npt
from mxnet.gluon import HybridBlock



[...]

PR #14830, opened 2019-04-29
Use env var to enforce safe accumulation in ReduceAxesCompute

haojin2:
## Description ##
As title.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] New environment variable to control accumulation behavior.
- [x] Corresponding test coverage
- [x] Corresponding docs.

## Comments ##
@eric-haibin-lin 
Flakiness checked on GPU:
```
MXNET_TEST_COUNT=10000 nosetests tests/python/gpu/test_operator_gpu.py:test_norm
[INFO] Setting module np/mx/python random seeds, use MXNET_MODULE_SEED=292245351 to reproduce.
.
----------------------------------------------------------------------
Ran 1 test in 256.972s

OK
```

PR #16589, opened 2019-10-23
Fix index overflow bug in einsum

hzfan:
## Description ##
Fix index overflow bug in einsum as mentioned in #16576

## Comments ##
Thank @reminisce for guidance and review.

### DIFF
--- a/tests/python/unittest/test_numpy_op.py
+++ b/tests/python/unittest/test_numpy_op.py
@@ -3378,10 +3378,11 @@ def test_np_einsum():
                                                                     _np.dot(_np.dot(args[0], args[1]).T, _np.ones((2, 2))))),
         # broadcast bug
         ('ij, ij -> i', [(1, 4), (2, 4)], lambda *args: (_np.sum(args[1], axis=0)[None, :],
-                                                           _np.tile(args[0], [2, 1]))),
+                                                         _np.tile(args[0], [2, 1]))),
         # issue #16576
-        ('abiz,abjz->abij', [(64, 8, 128, 512), (64, 8, 128, 512)], lambda *args: (_np.matmul(_np.ones((64, 8, 128, 128)), args[1]),
-                                                                                   _np.matmul(_np.ones((64, 8, 128, 128)), args[0]))),
+        # commented due to long running time
+        # ('abiz,abjz->abij', [(64, 8, 128, 512), (64, 8, 128, 512)], lambda *args: (_np.matmul(_np.ones((64, 8, 128, 128)), args[1]),
+        #                                                                            _np.matmul(_np.ones((64, 8, 128, 128)), args[0]))),
     ]
     dtypes = ['int32', 'float32', 'float64']
     for hybridize in [False, True]:
--- a/tests/python/unittest/test_numpy_op.py
+++ b/tests/python/unittest/test_numpy_op.py
@@ -3384,13 +3384,15 @@ def test_np_einsum():
         # ('abiz,abjz->abij', [(64, 8, 128, 512), (64, 8, 128, 512)], lambda *args: (_np.matmul(_np.ones((64, 8, 128, 128)), args[1]),
         #                                                                            _np.matmul(_np.ones((64, 8, 128, 128)), args[0]))),
     ]
-    dtypes = ['int32', 'float32', 'float64']
+    dtypes = ['float16', 'float32', 'float64', 'int32']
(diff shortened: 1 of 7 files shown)


hzfan:
@mxnet-label-bot  add [Numpy]

tests/python/unittest/test_numpy_op.py
```diff
@@ -3377,16 +3377,20 @@ def dbg(name, data):
...
-    dtypes = ['int32', 'float16', 'float32', 'float64']
+    dtypes = ['int32', 'float32', 'float64']
     for hybridize in [False, True]:
         for dtype in dtypes:
             for config in configs:
                 for optimize in [False, True]:
-                    rtol = 1e-0 if dtype == 'float16' else 1e-3
-                    atol = 1e-1 if dtype == 'float16' else 1e-5
+                    rtol = 1e-0 if dtype == 'float16' else 1e-1
+                    atol = 1e-1 if dtype == 'float16' else 1e-1
```
sxjscience: Why 1e0 and 1e-1?

hzfan: When shape is large, the behavior somehow diverges from official NumPy, so after the `np.einsum('abiz,abjz->abij', lhs, rhs)` test case is added (which sums up 512 numbers), I have to relax the tolerance.

Different implementation details can result in the minor difference, like intermediate result type. I require the intermediate var to be of the same type as input vars. But official numpy may has some type promotion here (of which I am not sure). This difference grows as the input size grows.

Also, note that exactly the same result can be produced with int inputs.

sxjscience: I see. Could we compare with some ground-truth values other than the numpy-produced ones? Also, we may set the accumulation type to be float32, which may also improve the performance.

hzfan: Yes, I think it’s a good idea to set the accumulation type. A promotion rule (from input type to accumulation type) may do this, like int32 to int64, float16 to float32, float32 to float64, float64 to float64 etc. I think it can be done in another pr.

As far as I know, MXNet numpy operators have been using official numpy as ground truth. If there are better options, plz let me know.

sxjscience: @hzfan For testing, we always compare with the implementation that is as close to the ground truth as possible. One solution is to benchmark against a higher precision output by numpy, e.g., float64. Setting the `rtol` and `atol` to some large values is not the correct approach. Also, you may set the accumulation type to a higher precision dtype like what we are doing in safe accumulate: https://github.com/apache/incubator-mxnet/pull/14830

sxjscience: @hzfan  For `np.einsum('abiz,abjz->abij', lhs, rhs)`, is it possible to fall back to Batched GEMM? I've communicated with @reminisce offline and einsum is quite important in implementing neural attention. This is critical as users will use it to implement their own model.

sxjscience: @hzfan You may also refer to this tutorial in TF: https://www.tensorflow.org/guide/keras/rnn

hzfan: @sxjscience Falling back to Batched GEMM is possible, but quite a lot work needs to be done. Now both official numpy and our implementation uses `tensordot` as a fallback, but `tensordot` cannot handle batch dot.

hzfan: @sxjscience safe accumulation has been added. Thank you for the suggestion.

haojin2:
@hzfan When you have shrunk some tolerance I would suggest trying multiple trials (like 1000) locally on your side to verify the flakiness a bit before merging, other than that this PR LGTM so far. @sxjscience Any more feedbacks?

sxjscience:
LGTM

hzfan:
@haojin2 I will do it. Now each test takes around 12 minutes on cpu (simply 1 forward takes around 5s, but we have different hybridize, dtype, optimize, forward, backward...), so 1000 tests take around 8.5 days, which is too long.

I will enforce hybridize=True, optimize=False, and use dtypes=`['float16', 'float32', 'float64']`, run 1000 tests. It should finish within a day.

haojin2:
@hzfan Please do that, I think hybridized or not will generally not have effect on the numerical precison. 
Also you could probably use only ~500 trials, a test that succeeds with 99% of chance could only pass 500 trials with less than 1% chance.

Issue #16604, opened 2019-10-24
RNN op with dropout cannot use fixed seed on CPU


roywei:
As follow up on https://github.com/apache/incubator-mxnet/pull/16532

I found out rnn operator with dropout result cannot be fixed with fixed seed on CPU.
GPU (cudnn) is fine after PR https://github.com/apache/incubator-mxnet/pull/16532

```
@with_seed()
def test_rnn_with_seed():
    info = np.iinfo(np.int32)
    seed = np.random.randint(info.min, info.max)
    
    _test_rnn(seed, mx.cpu())
    _test_rnn(seed, mx.gpu())

def _test_rnn(seed, ctx):
    data = mx.nd.ones((5, 3, 10), ctx=ctx)
    rnn = mx.gluon.rnn.RNN(100, 3, dropout=0.5)
    rnn.initialize(ctx=ctx)
    mx.random.seed(seed)
    with mx.autograd.record():
        result1 = rnn(data)

    mx.random.seed(seed)
    with mx.autograd.record():
        result2 = rnn(data)
    # dropout on gpu should return same result with fixed seed
    assert_almost_equal(result1.asnumpy(), result2.asnumpy())

```

Current workaround is DO NOT use dropout in rnn and manually append a dropout layer after rnn, the following works:
```
def _test_rnn(seed, ctx):
    data = mx.nd.ones((5, 3, 10), ctx=ctx)
    rnn = mx.gluon.rnn.RNN(100, 3, dropout=0.)
    rnn.initialize(ctx=ctx)
    dropout = mx.gluon.nn.Dropout(0.5)
    with mx.autograd.record():
        result1 = rnn(data)
        mx.random.seed(seed)
        o1 = dropout(result1)

    with mx.autograd.record():
        result2 = rnn(data)
        mx.random.seed(seed)
        o2 = dropout(result2)
    # dropout on gpu should return same result with fixed seed
    assert_almost_equal(o1.asnumpy(), o2.asnumpy())
```

roywei:
This implementation on lstm, gru and rnn should respect mxnet seed:

https://github.com/apache/incubator-mxnet/blob/master/src/operator/rnn_impl.h#L159

pengzhao-intel:
@zixuanweeei please take a look for this

zixuanweeei:
> @zixuanweeei please take a look the fix from @roywei :)

Sure. And thanks for @roywei 's great work.

samskalicky:
[...]

PR #16605, opened 2019-10-24
Fix rnn dropout on CPU

roywei:
fix https://github.com/apache/incubator-mxnet/issues/16604

depend on https://github.com/apache/incubator-mxnet/pull/16532, created a separate PR in case my fix on rnn cpu has problem. I don't want to impact the merge of https://github.com/apache/incubator-mxnet/pull/16532


tests/nightly/test_fixed_seed.py
```diff
@@ -0,0 +1,69 @@
...
+
+    mx.random.seed(seed)
+    with mx.autograd.record():
+        result1 = dropout(data)
+
+    mx.random.seed(seed)
+    with mx.autograd.record():
+        result2 = dropout(data)
+    # dropout on gpu should return same result with fixed seed
+    assert_almost_equal(result1.asnumpy(), result2.asnumpy())
```
marcoabreu: Didn't you also have tests that verified that different seeds return different results?

roywei: According to the discussion [here](https://github.com/apache/incubator-mxnet/pull/16532#discussion_r337794095), we choose to generate a random seed from MXNet CPU Random module  to avoid seeding every time during forward. The benefit is we only seed once during cuddn dropout initialization, and use CPU random module will avoid calling cuda mem copy. Drawback is we can not fix seed on different device. Note that the original code also cannot seed on different device. Currently we didn't see any usage of such functionality, so we chose this solution for performance.

marcoabreu: Understand, thanks for elaborating

PR #16644, opened 2019-10-27
[mkldnn-v1.0] Minor fix of mkldnn-v1.0 transition

zixuanweeei:
## Description ##
Minor fix.

### Changes ###
- [x] Remove two repetitive mkdir and rm directory operations from mkldnn.mk.
- [x] Delete redundant whitespace from ndarray.cc.
- [x] Rename variables of backward primitive descriptors and primitives in mkldnn_act*.
- [x] Use const_iterator instead of iterator in mkldnn_rnn.cc.

## Comments ##
- The two complex macro condition checks of `MXNET_USE_MKLDNN == 1` need another PR to become more readable and cleaner.
@pengzhao-intel @TaoLv @rongzha1


src/operator/nn/mkldnn/mkldnn_rnn.cc
```diff
@@ -676,7 +676,8 @@ void MKLDNNRnnOp::Init(const OpContext &ctx,
       "Layer vector's size has a different value than the number of fusion.";
   if (dst_.size() < num_fusion - 1) {
     int data_dtype = outputs[rnn_enum::kOut].dtype();
-    for (auto fwd = fwd_inf_vec_.begin(); fwd < fwd_inf_vec_.end() - 1; ++fwd)
+    for (std::vector<MKLDNNRnnForward>::const_iterator fwd = fwd_inf_vec_.begin();
+        fwd < fwd_inf_vec_.end() - 1; ++fwd)
```
ZhennanQin: strange code style. Use `fwd != fwd_inf_vec_.end() - 1` instead. For curious, why fwd can't be fwd_inf_vec_.end() - 1? Shouldn't we remove the last element from fwd_inf_vec_?

zixuanweeei: Here we need `fwd_inf_vec_.size() - 1` spaces for the intermediate results of the multiple fused layers. And for the result of the last fused layer, `outputs[rnn_enum::kOut]` could provide the space. Hence, `forward_inf_vec_.back()` is excluded when allocates the spaces for intermediate results.

ZhennanQin: Please make your reply as a code comment here. Then other developer will know the reason behind by reading it.

zixuanweeei: Thanks for reminding me. I will add the comment.

pengzhao-intel:
Merging this now to run test in the master together.

Issue #16431, opened 2019-10-10
[RFC] MXNet Multithreaded Inference Interface


anirudh2290:
Thanks to @nswamy for his inputs and design discussions related to this project and @frankfliu for explaining the requirements and the use case from customer perspective.

# Problem Statement

One of the big un-catered for use cases in MXNet is loading a model and being able to run parallel inference on the model from multiple threads while sharing the parameters. There are multiple user requests for the same [[1]](https://github.com/apache/incubator-mxnet/issues/3946). There also has been a lot of confusion around the current state of MXNet with respect to thread safety.

This doc attempts to address three things : 

1. Tries to clarify the current state of MXNet with respect to thread safety.
2. Tries to give an idea of the benefits to expect from adding this feature.
3. Attempts to solve the problem of parallel inference by providing a multi-threaded inference API ( C APIs and frontend APIs in CPP and Python), 

# Current State of MXNet Thread Safety

## MXNet Dependency Engine Thread Safety

Examining MXNet dependency engine code, it looks like it was designed  to be thread safe. Tried to push Convolution op from multiple threads into MXNet Engine, to see if there are any issues with thread safety. Used CPP Package for the same. The script is provided here : https://github.com/anirudh2290/mxnet/tree/multithreaded_inference_poc/cpp-package/example/multithreading_engine_push_mxnet_op.cpp

```
./build/cpp-package/example/multithreading_engine_push_mxnet_op 2
```

The script pushes Convolution op to the engine from multiple threads. You can verify the correctness of the op with this script : 
https://github.com/anirudh2290/mxnet/tree/multithreaded_inference_poc/test_cached_op_ts_check.py

```
python3 test_cached_op_ts_check.py
```

## MXNet Graph Executor Thread Safety

Removed NaiveEngine only restriction for C Predict API and tried to run multi threaded inference with C Predict API using ThreadedEngine by commenting the check : https://github.com/anirudh2290/mxnet/tree/multithreaded_inference_poc/src/c_api/c_predict_api.cc

When running this example the program core dumps with memory leaks in Graph Executor Bind. This shows that graph executor is not thread safe. 

## Cached Op (Gluon Backend) Thread Safety

Try to create cached op in the main thread and spawn multiple threads to invoke the same cached op inside each of the threads. Here is the script which does the same : https://github.com/anirudh2290/mxnet/tree/multithreaded_inference_poc/cpp-package/example/multithreading_engine_push_cached_op.cpp

```
# Usage
[...]

PR #9055, opened 2017-12-13
Re-enable all op segments when in batch mode

KellenSunderland:
## Description ##
This patch fixes a performance regression that is causing a roughly 20% slowdown in some inference scenarios.  I haven't investigated in depth how wide spread this issue is, but at a minimum it's affecting inference with a resnet model and a batch size of 1.

I'm not totally familiar with the commit that makes the change, but to me it looks like the regression occurred because of a refactor in the graph executor that improves readability.  The performance problems occur because we introduce a number of cudaStreamSyncronize calls which lower our overall gpu utilization (likely due to a restricted execution plan that leads to a lower overall SM occupancy).  It's not clear to me if this would cause any issues when running inference in a non-GPU environment.

It is possible that the change in operator segmentation behaviour was intended and has other useful implications.  If so we should be able to modify the code so that we get both the intended benefits, and the faster performance.

## Investigation ##
After verifying the regression on a demonstration model we did some high level measurements to see if any trends emerged.  We saw two stats that helped diagnose the issue.   First the gpu usage during inference dropping from ~95% to ~85%.  Second we had a large increase in the number of cudaStreamSynchronize calls.

0.9.5  cuda calls:
```
Time(%)      Time     Calls       Avg       Min       Max  Name
 30.59%  16.6780s      2401  6.9463ms  5.7920us  31.420ms  cudaStreamSynchronize
 22.59%  12.3190s    105146  117.16us  22.368us  12.287ms  cudaLaunch
 17.10%  9.32475s        16  582.80ms  195.55us  4.66099s  cudaStreamCreateWithFlags
```
0.12 cuda calls:
```
==2726== API calls:
Time(%)      Time     Calls       Avg       Min       Max  Name
 46.13%  37.5542s     48601  772.70us  5.1200us  5.5922ms  cudaStreamSynchronize
 17.72%  14.4214s    102946  140.09us  22.432us  9.2845ms  cudaLaunch
 11.32%  9.21538s        16  575.96ms  222.94us  4.60608s  cudaStreamCreateWithFlags
...
```

Note the number of cudaStreamSynchronize calls increases from 2401 to 48601.  Once this PR has been applied our models return to exactly 2401 calls.

[...]

PR #16654, opened 2019-10-28
Multithreaded Inference Support

anirudh2290:
## Description ##

To get some background on the PR, please go through the related github issue : https://github.com/apache/incubator-mxnet/issues/16431 . Below, I have added a brief overview for the PR.

## Brief Overview ##

1. This PR adds a multithreaded inference interface (EXPERIMENTAL) for MXNet. The approach it takes for this is to add a thread safe version of cached op with minimal set of features required for inference.
2. Adds C API to create cached op thread safe version: The API: CreateCachedOpEX takes an additional parameter `thread_safe`. When `thread_safe` is set to true, invokes on the cached op handle will invoke the thread safe version of cached op.
```
  /*!
   * \brief create cached operator, allows to choose thread_safe version
   * of cachedop
   */
  MXNET_DLL int MXCreateCachedOpEX(SymbolHandle handle,
                                   int num_flags,
                                   const char** keys,
                                   const char** vals,
                                   CachedOpHandle *out,
                                   bool thread_safe DEFAULT(false));
```

3. This PR adds tests and verification for thread safety of engine, for running parallel inference with the thread safe version of CachedOp with different models and different number of threads and different number of inferences per thread.
4. Model Testing has been limited to : resnet18, resnet50_v1, resnet152_v1
5. Adds CI stages and CI tests for running the thread safety tests. Please see : `capi-cpp-package GPU` in `unix-gpu` stage.

## Feedback/Acknowledgements ##

Below is the feedback/help received from offline code/design reviews:

1. @nswamy - Early inputs and offline design discussions related to this project, specifically with respect to dmlc::ThreadLocalStore.
2. @frankfliu - Explaining the requirements and use case for MXNet backend from a customer perspective, code review suggestions to have thread safe API for Create only and to determine whether to invoke cached op/cached op threadsafe based on handle.
3. @andreaolgiati  - Suggestion to document current threading model of MXNet (still a TODO), requesting adding support for dynamic shapes for NLP case (not supported yet), suggested documenting the long term plan with respect to thread safe inference in mxnet. (Documented in limitations and Future work section of the doc, will add required github issues).
4. @lanking520 , @larroy , @zachgk - Suggestion to add a new API for Threadsafe cached op version, since modifying existing one will break backward compatibility.
5. @frankfliu and @oorqueda - Suggestion for continuous maintenance (addition of support) of the models supported table. Will be maintained on the doc and RFC issue.
6. @samskalicky - offline code review suggestions with respect to code comments for locking strategy and comments for changes made for thread safe cached op.
7. @vikas-kum - better document the performance benefits. [still a TODO]

Thanks to all above for the suggestions, review and feedback.

Thanks to @marcoabreu and @arcadiaphy for the valuable comments on the RFC. 

Additionally, thanks to @roywei, @access2rohit , @stu1130 , @ChaiBapchya , @rondogency , @yuxihu, @vrakesh, @Keerthanvasist, @josephevans for code/design review.

### Essentials ###


include/mxnet/c_api.h
```diff
@@ -1274,10 +1274,28 @@ MXNET_DLL int MXCreateCachedOpEx(SymbolHandle handle,
...
+/*!
+ * \brief create cached operator, allows to choose thread_safe version
+ * of cachedop
+ */
+MXNET_DLL int MXCreateCachedOpEX(SymbolHandle handle,
+                                 int num_flags,
+                                 const char** keys,
+                                 const char** vals,
+                                 CachedOpHandle *out,
+                                 bool thread_safe = false);
```
zachgk: Aren't default arguments like this only supported in C++ and this file is just C?

anirudh2290: Yes i need to DEFAULT(false) like https://github.com/apache/incubator-mxnet/blob/master/include/mxnet/c_api.h#L3157

ci/docker/runtime_functions.sh
```diff
@@ -1353,6 +1373,19 @@ integrationtest_ubuntu_gpu_cpp_package() {
...
 }
 
+integrationtest_ubuntu_gpu_capi_cpp_package() {
+    set -ex
+    export PYTHONPATH=./python/
+    export LD_LIBRARY_PATH=/work/mxnet/lib:$LD_LIBRARY_PATH
+    python3 -c "import mxnet as mx; mx.test_utils.download_model(\"imagenet1k-resnet-18\"); mx.test_utils.download_model(\"imagenet1k-resnet-152\"); mx.test_utils.download_model(\"imagenet1k-resnet-50\");"
+    build/tests/cpp/mxnet_unit_tests --gtest_filter="ThreadSafety.*"
+    build/tests/cpp/mxnet_unit_tests --gtest_filter="ThreadSafety.*" --thread-safety-with-cpu
+    export MXNET_ENGINE_TYPE=NaiveEngine
```
larroy: Why NaiveEngine here? Could you add comments?

anirudh2290: This is to run thread safety tests in NaiveEngine mode. I have added a comment.

anirudh2290:
@zachgk @larroy Thanks for the review. I have addressed it.

ci/docker/runtime_functions.sh
```diff
@@ -1212,6 +1230,8 @@ unittest_ubuntu_cpugpu_perl() {
 
 unittest_cpp() {
     set -ex
+    export PYTHONPATH=./python/
+    python3 -c "import mxnet as mx; mx.test_utils.download_model(\"imagenet1k-resnet-18\"); mx.test_utils.download_model(\"imagenet1k-resnet-152\"); mx.test_utils.download_model(\"imagenet1k-resnet-50\");"
```
larroy: I don't like this inlined python very much, it's also duplicated code below, it could be factored out into a utility script or a function.

docs/static_site/src/pages/api/cpp/docs/tutorials/multi_threaded_inference.md
```diff
@@ -0,0 +1,324 @@
...
+$ make
+```
+
+If you have built mxnet from source with cmake, please uncomment the specific lines for cmake build or set the following environment variables: `MKLDNN_BUILD_DIR (default is $(MXNET_ROOT)/3rdparty/mkldnn/build)`, `MKLDNN_INCLUDE_DIR (default is $(MXNET_ROOT)/3rdparty/mkldnn/include)`, `MXNET_LIB_DIR (default is $(MXNET_ROOT)/lib)`.
+
+### Download the model and run multi threaded inference example
+To download a model use the `get_model.py` script. This downloads a model to run inference.
+
+```python
+python3 get_model.py --model <model_name>
```
larroy: why not use this get model above?

docs/static_site/src/pages/api/cpp/docs/tutorials/multi_threaded_inference.md
```diff
@@ -0,0 +1,324 @@
...
+              << "Usage: ./multi_threaded_inference [model_name] [num_threads] [is_gpu] apple.jpg"
+              << std::endl
+              << "Example: ./.multi_threaded_inference imagenet1k-inception-bn 1 0 apple.jpg"
+              << std::endl
+              << "NOTE: Thread number ordering will be based on the ordering of file inputs" << std::endl
+              << "NOTE: Epoch is assumed to be 0" << std::endl;
+    return EXIT_FAILURE;
+  }
+  std::string model_name = std::string(argv[1]);
+  int num_threads = std::atoi(argv[2]);
```
larroy: let's use std::to_string instead of unsafe atoi in general. This should be a linter rule.

docs/static_site/src/pages/api/cpp/docs/tutorials/multi_threaded_inference.md
```diff
@@ -0,0 +1,324 @@
...
+  int num_inputs = out.ListInputs().size();
+
+  for (size_t i = 0; i < data_arr.size(); ++i) {
+    data_arr[i] = input_arrs[i].Copy(ctx);
+  }
+  prepare_input_data(softmax_shape, ctx, num_threads, &softmax_arr);
+  std::map<std::string, mxnet::cpp::NDArray> parameters;
+  mxnet::cpp::NDArray::Load(param_file, 0, &parameters);
+
+  for (std::string name : out.ListInputs()) {
```
larroy: reference?  does this code compile? Can be included in the doc? how do we keep it current.

example/multi_threaded_inference/multi_threaded_inference.cc
```diff
@@ -0,0 +1,351 @@
...
+  int num_inputs = out.ListInputs().size();
+
+  for (size_t i = 0; i < data_arr.size(); ++i) {
+    data_arr[i] = input_arrs[i].Copy(ctx);
+  }
+  prepare_input_data(softmax_shape, ctx, num_threads, &softmax_arr);
+  std::map<std::string, mxnet::cpp::NDArray> parameters;
+  mxnet::cpp::NDArray::Load(param_file, 0, &parameters);
+
+  for (std::string name : out.ListInputs()) {
```
larroy: const reference?

src/imperative/cached_op.h
```diff
@@ -26,8 +26,180 @@
 #include <utility>
 #include <string>
 #include <unordered_map>
+#include <map>
+#include "../operator/operator_common.h"
+#include "../operator/subgraph/common.h"
+#include "./imperative_utils.h"
 
 namespace mxnet {
+namespace {
```
larroy: does this anon ns need to be inside mxnet?

src/imperative/cached_op.h
```diff
@@ -26,8 +26,180 @@
...
+  static const char BACKWARD[] = "backward";
+  static const char REF_COUNT[] = "ref_count";
+  static const char MEM_PLAN[] = "mem_plan";
+  static const char STORAGE_PLAN[] = "storage_plan";
+
+std::string AddPrefix(const std::string& prefix,
+                      const std::string& s) {
+  return prefix + "_" + s;
+}
+void CreateFullGraph(const nnvm::Symbol& sym,
```
larroy: missing function documentation

src/imperative/cached_op.h
```diff
@@ -26,8 +26,180 @@
...
+        }
+        fwd_graph->outputs.emplace_back(std::move(copy_node));
+      } else {
+        dedup_out.emplace(nodeEntry, 0);
+        fwd_graph->outputs.push_back(nodeEntry);
+      }
+    }
+  }
+
+  bool do_elim_common_expr = dmlc::GetEnv("MXNET_ELIMINATE_COMMON_EXPR", true);
```
larroy: can this be factored out in smaller functions?

src/imperative/cached_op.h
```diff
@@ -26,8 +26,180 @@
...
+  }
+
+  // construct full graph
+  {
+    full_graph->outputs = fwd_graph->outputs;
+    for (const auto& i : grad_graph->outputs) full_graph->outputs.emplace_back(i);
+  }
+}
+
+void SetRefCounts(nnvm::Graph* fwd_graph, const nnvm::Graph& full_graph) {
```
larroy: doc?

src/imperative/cached_op.h
```diff
@@ -26,8 +26,180 @@
...
+                                                                             REF_COUNT));
+  for (size_t i = 0; i < num_forward_entries; ++i) full_ref_count.at(i) += temp_ref_count[i];
+  fwd_graph->attrs[AddPrefix(FULL, REF_COUNT)] =
+      std::make_shared<dmlc::any>(std::move(full_ref_count));
+}
+
+void OptimizeGraph(nnvm::Graph * full_graph, nnvm::Graph * fwd_graph, nnvm::Graph * grad_graph,
+                   const Context& context, size_t num_forward_outputs, const bool inlining) {
+#if MXNET_USE_CUDA && !defined(_WIN32)
+  if (context.dev_mask() == kGPU &&
```
larroy: can this be factored out?

src/imperative/cached_op_threadsafe.cc
```diff
@@ -0,0 +1,373 @@
...
+  static const std::vector<const Op *> zero_ops{Op::Get("zeros_like"),
+                                                Op::Get("_zeros")};
+  static const auto _copy_op = Op::Get("_copy");
+  config_.Init(flags);
+
+  if (config_.static_shape) {
+      CHECK(config_.static_alloc) << "static_alloc must be True when static_shape is True";
+  }
+
+  // construct forward graph
```
larroy: why not factor out instead of putting a comment and a block of code?

src/imperative/cached_op_threadsafe.cc
```diff
@@ -0,0 +1,373 @@
...
+      tmp.reserve(indexed_graph.input_nodes().size());
+      for (size_t i = 0; i < indexed_graph.input_nodes().size(); ++i) {
+        tmp.emplace_back(i);
+      }
+      config_.data_indices.assign(tmp.begin(), tmp.end());
+    }
+  }
+}
+
+OpStatePtr CachedOpThreadSafe::DynamicForward(const Context& default_ctx,
```
larroy: doc?

src/imperative/cached_op_threadsafe.cc
```diff
@@ -0,0 +1,373 @@
...
+  const auto &idx = g.indexed_graph();
+  size_t num_inputs = idx.input_nodes().size();
+  size_t max_nodes = runtime.info.fwd_graph.indexed_graph().num_nodes();
+  runtime.op_states.resize(max_nodes);
+  auto &states = runtime.op_states;
+
+  // Allocate entries
+  // This buff is thread local and used to store intermediate
+  // nodes in the graph
+  buff.resize(idx.num_node_entries());
```
larroy: This coding style is similar as in the backward pass, and I'm not a fan of this code style because I find it not very readable due to back and forth the indexed graph, indices, buffers and arrays. I think it can be abstracted more and make it easier to follow by factoring out in functions where the indexes don't appear. In the end it's the way a big chunk of mxnet is so this is like another drop in the ocean, but anyway.

ci/docker/runtime_functions.sh
anirudh2290: I think writing a function for this is an overkill.

docs/static_site/src/pages/api/cpp/docs/tutorials/multi_threaded_inference.md
anirudh2290: If you are hinting at the undefined behavior that will happen only for outside int ranges. As this is for num_threads for an example script this is unlikely. Otherwise, I think atoi seems fine for the use case here.

docs/static_site/src/pages/api/cpp/docs/tutorials/multi_threaded_inference.md
anirudh2290: no, this is a doc and I had added this to refer to the code that I was refering to in the cc file. I changed it to the code blocks in the cc file, to make it easier to maintain.

example/multi_threaded_inference/multi_threaded_inference.cc
anirudh2290: done.

src/imperative/cached_op.h
anirudh2290: this was from before. it was to keep these functions local to the translation unit.

src/imperative/cached_op.h
anirudh2290: this function was from before. added doc.

src/imperative/cached_op.h
anirudh2290: this function was from before, its outside the scope of this pr. Feel free to add a new PR for refactoring this.

src/imperative/cached_op.h
anirudh2290: added doc.

src/imperative/cached_op.h
anirudh2290: this function was from before, refactoring this is outside the scope of this pr.

src/imperative/cached_op_threadsafe.cc
anirudh2290: refactored.

src/imperative/cached_op_threadsafe.cc
anirudh2290: done.

src/imperative/cached_op_threadsafe.cc
anirudh2290: feel free to open a new pr for refactor.

docs/static_site/src/pages/api/cpp/docs/tutorials/multi_threaded_inference.md
```diff
@@ -0,0 +1,199 @@
...
+2. Explain how one can use C API and thread safe version of cached op, along with CPP package to achieve iultithreaded inference. This will be useful for end users as well as frontend developers of different language bindings
+3. Discuss the limitations of the above approach
+4. Future Work
+
+## Current state of Thread Safety in MXNet
+
+Examining the current state of thread safety in MXNet we can arrive to the following conclusion:
+
+1. MXNet Dependency Engine is thread safe (except for WaitToRead invoked inside a spawned thread. Please see Limitations section)
+2. Graph Executor which is Module/Symbolic/C Predict API backend is not thread safe
```
eric-haibin-lin: Awesome documetnation and knowledge sharing! Could you share what class/component used in graph executor/CachedOp are not thread safe?

anirudh2290: Thanks for the kind words @eric-haibin-lin . For graph executor,  I saw core dumps with memory leaks for graph executor bind which indicated that Graph Executor Bind was not thread safe. I didnt dig more into what exactly in Graph Executor bind, it could be one of the NNVM passes. For cached op, the issues where 1. with respect to not using thread local variables for intermediate states buffer (buff) in CachedOp::DynamicForward 2. Issues related to hang in accept4 call in cuda lib when  op push is not serialized in CachedOp::Forward.

docs/static_site/src/pages/api/cpp/docs/tutorials/multi_threaded_inference.md
```diff
@@ -0,0 +1,199 @@
...
+To download a model use the `get_model.py` script. This downloads a model to run inference.
+
+```python
+python3 get_model.py --model <model_name>
+```
+e.g.
+```python
+python3 get_model.py --model imagenet1k-inception-bn
+```
+Only the supported models with `get_model.py` work with multi threaded inference.
```
apeforest: Where can the user find the supported model names? Can you provide a link here?

docs/static_site/src/pages/api/cpp/docs/tutorials/multi_threaded_inference.md
```diff
@@ -0,0 +1,199 @@
...
+## Multithreaded inference in MXNet with C API and CPP Package
+
+### Prerequisites
+To complete this tutorial you need to:
+- Learn the basics about [MXNet C++ API](/api/cpp)
+- Build MXNet from source with make/cmake
+- Build the multi-threaded inference example
+
+### Setup the MXNet C++ API
+To use the C++ API in MXNet, you need to build MXNet from source with C++ package. Please follow the [built from source guide](/get_started/ubuntu_setup.html), and [C++ Package documentation](/api/cpp)
```
apeforest: Can I just use C API without building the cpp package?

example/multi_threaded_inference/multi_threaded_inference.cc
```diff
@@ -0,0 +1,351 @@
...
+#include <fstream>
+#include <thread>
+#include <iomanip>
+#include <chrono>
+#include <mxnet/ndarray.h>
+#include <opencv2/opencv.hpp>
+#include <mxnet/c_predict_api.h>
+#include "mxnet-cpp/MxNetCpp.h"
+
+const mx_float DEFAULT_MEAN = 117.0;
```
apeforest: Let's just use float. I don't think we will need this typedef any more.

example/multi_threaded_inference/multi_threaded_inference.cc
```diff
@@ -0,0 +1,351 @@
...
+    if (random_uniform) {
+      mxnet::cpp::Operator("_random_uniform")(begin, end)
+          .Invoke((*data_arr)[i]);
+    }
+    mxnet::cpp::NDArray::WaitAll();
+  }
+}
+
+// Run inference on a model
+void run_inference(const std::string& model_name, const std::vector<mxnet::cpp::NDArray>& input_arrs,
```
apeforest: just mxnet::NDArray?

example/multi_threaded_inference/multi_threaded_inference.cc
```diff
@@ -0,0 +1,351 @@
...
+                 " static_shape: " + std::to_string(static_shape);
+  std::string json_file = model_name + "-symbol.json";
+  std::string param_file = model_name + "-0000.params";
+  auto out = mxnet::cpp::Symbol::Load(json_file);
+  std::string static_alloc_str = static_alloc ? "true" : "false";
+  std::string static_shape_str = static_shape ? "true" : "false";
+
+  // Prepare context
+# if MXNET_USE_CUDA == 1
+  mxnet::Context backend_ctx;
```
apeforest: Same question for mxnet::Context vs mxnet::cpp::Context

example/multi_threaded_inference/multi_threaded_inference.cc
```diff
@@ -0,0 +1,351 @@
...
+              << "NOTE: Epoch is assumed to be 0" << std::endl;
+    return EXIT_FAILURE;
+  }
+  std::string model_name = std::string(argv[1]);
+  int num_threads = std::atoi(argv[2]);
+  bool is_gpu = std::atoi(argv[3]);
+  CHECK(num_threads == argc - 4) << "Number of files provided, should be same as num_threads";
+  std::vector<std::string> test_files;
+  for (size_t i = 0; i < argc - 4; ++i) {
+  test_files.emplace_back(argv[4 + i]);
```
apeforest: indentation

src/imperative/cached_op.h
```diff
@@ -26,8 +26,195 @@
...
+std::string AddPrefix(const std::string& prefix,
+                      const std::string& s) {
+  return prefix + "_" + s;
+}
+
+/* \brief create a forward graph from they Symbol */
+void CreateForwardGraph(const nnvm::Symbol &sym, nnvm::Graph *fwd_graph) {
+  using namespace nnvm;
+  static const auto _copy_op = Op::Get("_copy");
+  {
```
apeforest: Can we write a function instead of using braces to mark the scope.

src/imperative/cached_op.h
```diff
@@ -26,8 +26,195 @@
...
+  using namespace nnvm;
+  static const std::vector<const Op*> zero_ops{Op::Get("zeros_like"), Op::Get("_zeros")};
+  CreateForwardGraph(sym, fwd_graph);
+
+  bool do_elim_common_expr = dmlc::GetEnv("MXNET_ELIMINATE_COMMON_EXPR", true);
+  if (do_elim_common_expr)
+    *fwd_graph = exec::EliminateCommonExpr(std::move(*fwd_graph));
+
+  // construct backward graph
+  {
```
apeforest: Same here. Using a function instead of a pair of braces can improve readability.

src/imperative/cached_op.h
```diff
@@ -26,8 +26,195 @@
...
+        << "There are no inputs in computation graph that require gradients.";
+
+    *grad_graph = pass::MXGradient(
+        *fwd_graph, fwd_graph->outputs, xs, *ograd_entries,
+        exec::AggregateGradient, nullptr, nullptr,
+        zero_ops, "_copy");
+  }
+
+  // construct full graph
+  {
```
apeforest: Same here.

src/imperative/cached_op.h
```diff
@@ -26,8 +26,195 @@
...
+  fwd_graph->outputs = std::vector<nnvm::NodeEntry>(full_graph->outputs.begin(),
+                                                    full_graph->outputs.begin() +
+                                                    num_forward_outputs);
+  *grad_graph = nnvm::Graph();
+  grad_graph->outputs = std::vector<nnvm::NodeEntry>(full_graph->outputs.begin() +
+                                                     num_forward_outputs,
+                                                     full_graph->outputs.end());
+  SetRefCounts(fwd_graph, *full_graph);
+}
+
```
apeforest: remove redundant blank lines.

src/imperative/cached_op.cc
```diff
@@ -32,247 +32,12 @@ DMLC_REGISTER_PARAMETER(CachedOpConfig);
...
-  std::vector<uint32_t> bwd_input_eid;
-};
-
 struct CachedOp::DynamicRuntime {
   GraphInfo info;
   std::vector<NDArray> buff;
   std::vector<OpStatePtr> op_states;
 };
 
-void CreateFullGraph(const nnvm::Symbol& sym,
```
apeforest: Any particular reason you moved these function implementation from .cc to .h?

src/imperative/cached_op_threadsafe.h
```diff
@@ -0,0 +1,129 @@
...
+// Used in dynamic_forward
+#if DMLC_CXX11_THREAD_LOCAL
+    static thread_local std::vector<NDArray> buff;
+#else
+    static MX_THREAD_LOCAL std::vector<NDArray> buff;
+#endif
+
+
+
+class CachedOpThreadSafe : public CachedOp {
```
apeforest: Maybe I missed something, but why can't we modify the CachedOp to make it thread safe rather than creating a new inherited class? In which scenarios users want to choose a thread unsafe CachedOp and is there any guideline somewhere?

tests/cpp/operator/mkldnn_operator_test.cc
```diff
@@ -719,7 +707,10 @@ void TestOpEx(const OpAttrs &forward_attrs, const OpAttrs &backwards_attrs) {
             Context(), forward_attrs.attrs, inputs, ex_outputs, req,
             DispatchMode::kFComputeEx, mxnet::OpStatePtr());
         Engine::Get()->WaitForAll();
-        AssertEqual(outputs, ex_outputs);
+        // TODO(unassigned): Need to fix op, should work for the whole vector
```
apeforest: Can you assign TODO here? otherwise I am afraid it might remain TODO forever :)

apeforest:
Thanks for this big contribution! In the future, it might be easier for reviewers if the PR can be broken into smaller pieces :)

src/imperative/cached_op_threadsafe.cc
```diff
@@ -0,0 +1,342 @@
...
+  if (config_.static_shape) {
+      CHECK(config_.static_alloc) << "static_alloc must be True when static_shape is True";
+  }
+
+  // construct forward graph
+  CreateForwardGraph(sym.Copy(), &fwd_graph_);
+  SetForwardRefCounts(&fwd_graph_);
+
+  // Set param indices
+  {
```
ChaiBapchya: Same as pointed by Lin and Pedro before. Factor out to a function.

tests/cpp/engine/thread_local_test.cc
```diff
@@ -56,7 +56,7 @@ static int ThreadSafetyTest(int num, std::vector<int>* tmp_inputs, std::vector<i
     return 0;
 }
 
-TEST(ThreadLocal, verify_thread_safety) {
```
ChaiBapchya: Why is this changed?

tests/cpp/thread_safety/thread_safety_test.cc
```diff
@@ -0,0 +1,667 @@
...
+        }
+        arr_handles[i][j].push_back(softmax_arr[i][j].GetHandle());
+      }
+    }
+    CachedOpHandle hdl = CachedOpHandle();
+    get_expected_results_multiple(out, flag_keys, flag_vals, &arr_handles,
+                                  num_threads, &result_expected, &hdl);
+
+
+    // Create thread safe cahced op
```
ChaiBapchya:
```suggestion
    // Create thread safe cached op
```

src/imperative/cached_op.cc
```diff
@@ -32,247 +32,12 @@ DMLC_REGISTER_PARAMETER(CachedOpConfig);
 
 constexpr uint32_t kEidNotExist = std::numeric_limits<uint32_t>::max();
 
-const char CachedOp::FULL[] = "full";
```
eric-haibin-lin: Could you describe the main changes made for cached_op.cc/h (code moved/removed/added)?

docs/static_site/src/pages/api/cpp/docs/tutorials/multi_threaded_inference.md
anirudh2290: I used CPP package because using it makes writing these tests and example a little easier. It can also be written solely with C API.

example/multi_threaded_inference/multi_threaded_inference.cc
anirudh2290: changed to float

example/multi_threaded_inference/multi_threaded_inference.cc
anirudh2290: the cpp ndarray is a frontend ndarray which is implemented like any other frontend ndarray i.e. save the ndarray handle (pointer to the ndarray) and perform C API operations on it.

example/multi_threaded_inference/multi_threaded_inference.cc
anirudh2290: frontend version of the context data structure. provides getters and setters for dev_type and dev_id.

example/multi_threaded_inference/multi_threaded_inference.cc
anirudh2290: fixed.

src/imperative/cached_op.h
anirudh2290: this was part of a refactor in this pr. earlier this whole chunk of code was called directly from forward of cached op. The braces are not needed here. I have removed it.

src/imperative/cached_op.h
anirudh2290: this is the only place this function is called from. I think we can wait till this is needed elsewhere and move it to its own function.

src/imperative/cached_op.h
anirudh2290: similar to above, i think we can wait till this is needed elsewhere and move it to its own function.

src/imperative/cached_op.h
anirudh2290: removed.

src/imperative/cached_op.cc
anirudh2290: to allow to be used in both CachedOp and CachedOpThreadSafe.

src/imperative/cached_op_threadsafe.h
anirudh2290: the thread safe cached op has not achieved feature parity with the current cached op. Lot of things not supported including training, bulking, multi context use case etc.  Guideline is to use the thread safe cached op for multi threaded inference use case and default one for everything else.

tests/cpp/operator/mkldnn_operator_test.cc
anirudh2290: i didn't know who to assign it to. I have assigned to @pengzhao-intel for now, please feel free to suggest other names.

CMakeLists.txt
```diff
@@ -825,7 +829,6 @@ if(MSVC AND USE_MXNET_LIB_NAMING)
   set_target_properties(mxnet PROPERTIES OUTPUT_NAME "libmxnet")
 endif()
 
-add_subdirectory(tests)
```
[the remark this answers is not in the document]
anirudh2290: this is for the tests to be built after the cpp package, because of the additional dependency on cpp package for some tests.

cpp-package/include/mxnet-cpp/symbol.hpp
```diff
@@ -151,6 +151,18 @@ inline std::vector<std::string> Symbol::ListArguments() const {
   }
   return ret;
 }
+
+inline std::vector<std::string> Symbol::ListInputs() const {
+  std::vector<std::string> ret;
+  mx_uint size;
```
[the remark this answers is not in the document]
anirudh2290: can you elaborate ?

example/multi_threaded_inference/Makefile
```diff
@@ -0,0 +1,67 @@
...
+    # Uncomment below line for CMake build
+    #export MKLDNN_INCLUDE_DIR = $(MXNET_ROOT)/3rdparty/mkldnn/include
+endif
+
+CFLAGS += -I$(MXNET_ROOT)/include -I$(CPP_PACKAGE)/include -I$(USE_CUDA_PATH)/include -I$(MKLDNN_INCLUDE_DIR) -I$(MKLDNN_BUILD_DIR)/include
+
+# If MXNET_LIB_DIR env variable set use that, otherwise defaults to MXNET_ROOT/build
+ifndef MXNET_LIB_DIR
+    MXNET_LIB_DIR=$(MXNET_ROOT)/lib
+    # cmake default by default
```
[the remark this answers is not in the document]
anirudh2290: hehe, not sure either. will remove it!

docs/static_site/src/pages/api/cpp/docs/tutorials/multi_threaded_inference.md
```diff
@@ -0,0 +1,199 @@
...
+```
+e.g.
+
+```bash
+./multi_threaded_inference imagenet1k-inception-bn 2 1 grace_hopper.jpg dog.jpg
+```
+
+The above script spawns 2 threads, shares the same cachedop and params among two threads, and runs inference on GPU. It returns the inference results in the order in which files are provided.
+
+NOTE: This example is to demonstrate the multi-threaded-inference with cached op. The inference results work well only with specific models (e.g. imagenet1k-inception-bn). The results may not necessarily be very accurate because of different preprocessing step required etc.
```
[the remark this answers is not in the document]
anirudh2290: What I mean is that if you use a random image to test here. the inference results may not be very accurate. the width and height of image is hardcoded and in the example and it may require specific preprocessing steps depending on image to get accurate result.

docs/static_site/src/pages/api/cpp/docs/tutorials/multi_threaded_inference.md
```diff
@@ -0,0 +1,199 @@
...
+op since the `thread_safe` (which is the last parameter to `MXCreateCachedOpEX`) is set to
+true. When this is set to false, it will invoke CachedOp instead of CachedOpThreadSafe.
+
+
+### Step 4: Prepare lambda function which will run in spawned threads
+
+[https://github.com/apache/incubator-mxnet/example/multi_threaded_inference/multi_threaded_inference.cc#L248-L262](multi_threaded_inference.cc#L248-262)
+
+The above creates the lambda function taking the thread number as the argument.
+If `random_sleep` is set it will sleep for a random number (secs) generated between 0 to 5 seconds.
```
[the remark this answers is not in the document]
anirudh2290: it can be 0-10 also, i just wanted to simulate a multi-threaded scenario where different threads call forward at different times.

example/multi_threaded_inference/multi_threaded_inference.cc
```diff
@@ -0,0 +1,351 @@
...
+              << "Example: ./.multi_threaded_inference imagenet1k-inception-bn 1 0 apple.jpg"
+              << std::endl
+              << "NOTE: Thread number ordering will be based on the ordering of file inputs" << std::endl
+              << "NOTE: Epoch is assumed to be 0" << std::endl;
+    return EXIT_FAILURE;
+  }
+  std::string model_name = std::string(argv[1]);
+  int num_threads = std::atoi(argv[2]);
+  bool is_gpu = std::atoi(argv[3]);
+  CHECK(num_threads == argc - 4) << "Number of files provided, should be same as num_threads";
```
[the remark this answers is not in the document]
anirudh2290: good point, i can change this.

tests/cpp/engine/thread_local_test.cc
anirudh2290: just making the test naming consistent.

src/imperative/cached_op.cc
anirudh2290: added comments to cached_op.cc, moved the constant strings and the helper functions AddPrefix, CreateFullGraph, SetRefCounts, OptimizeRefCounts, CachedOpState to cached_op.h file. Refactored CreateForwardGraph and SetForwardRefCounts into its own function.

src/imperative/cached_op.h
larroy: Anon namespace is fine, the question is why is inside mxnet when is dealing with nnvm.

src/imperative/cached_op.h
```diff
@@ -26,8 +26,191 @@
...
+std::string AddPrefix(const std::string& prefix,
+                      const std::string& s) {
+  return prefix + "_" + s;
+}
+
+/* \brief create a forward graph from the Symbol */
+void CreateForwardGraph(const nnvm::Symbol &sym, nnvm::Graph *fwd_graph) {
+  using namespace nnvm;
+  static const auto _copy_op = Op::Get("_copy");
+  NodeEntryMap<size_t> dedup_out;
```
larroy: Why this needs to map to size_t when you are treating just as a set?

src/imperative/cached_op.h
larroy: Is not about being used elsewhere, is about readability. If you don't need to dive deep is better to read a function call and skip the details.

src/imperative/cached_op_threadsafe.cc
larroy: You are changing this code now, there's should be no need to immediately refactor in the future.

anirudh2290: i am just following existing style. Refactoring cached op is not in the scope of this PR.

src/imperative/cached_op.h
anirudh2290: This is specific to cached op and mxnet. The location is correct.

src/imperative/cached_op.h
anirudh2290: its not a set, NodeEntryMap is typedef'd to following in NNVM:   

template<typename ValueType>
  using NodeEntryMap = std::unordered_map<NodeEntry, ValueType, NodeEntryHash, NodeEntryEqual>;

src/imperative/cached_op.h
anirudh2290: Added a new function.

src/imperative/cached_op.cc
apeforest: Could you add these to PR description?

src/imperative/cached_op_threadsafe.cc
```diff
@@ -0,0 +1,329 @@
...
+ * used to store intermediate nodes in the graph
+ */
+OpStatePtr CachedOpThreadSafe::DynamicForward(const Context& default_ctx,
+                                              const std::vector<NDArray*>& inputs,
+                                              const std::vector<NDArray*>& outputs) {
+  using namespace nnvm;
+  using namespace imperative;
+
+  {
+  auto state_ptr = GetCachedOpState(default_ctx);
```
apeforest: indent? Also, better to refactor into a function.

src/imperative/cached_op_threadsafe.cc
```diff
@@ -0,0 +1,329 @@
...
+  auto op_state = OpStatePtr::Create<DynamicRuntime>();
+  auto &runtime = op_state.get_state<DynamicRuntime>();
+  {
```
apeforest: Refactor into a function would be much better for readability

src/imperative/cached_op_threadsafe.cc
```diff
@@ -0,0 +1,329 @@
...
+#include "./imperative_utils.h"
+#include "../executor/exec_pass.h"
+#include "./cached_op_threadsafe.h"
+#include "../profiler/profiler.h"
+#include "../operator/operator_common.h"
+#include "../operator/subgraph/common.h"
+
+namespace mxnet {
+
+DMLC_REGISTER_PARAMETER(CachedOpThreadSafeConfig);
```
eric-haibin-lin: Is it necessary to have another implementation for cached op? What does it take to make the original cached op thread safe? This makes the code hard to maintain. Is there a plan of future works of thread_safe cached op? Do we plan to merge it back to original cached op?

rondogency: anirudh answered here: https://github.com/apache/incubator-mxnet/pull/16654#discussion_r366074941 but I think if we can complete all the todo items it's better to merge back

rondogency:
@anirudh2290 Thanks for this big contribution! one suggestion is that lots of developers complaining about cached op lacking docblocks and comments. since you're familiar with cached op, can you add a bit more docs on either your cached op or the original one? It is one time effort and can benefit the whole community

src/imperative/cached_op_threadsafe.cc
anirudh2290: > Is it necessary to have another implementation for cached op? What does it take to make the original cached op thread safe? This makes the code hard to maintain.

Hi @eric-haibin-lin , the scope of this project was to allow for multi threaded inference with the cached op. This is used in the DJL project (https://github.com/awslabs/djl). There are two options for implementation : 1. either include all these changes with a different code path inside cached op code or 2. create a new implementation for thread safe cached op. Between the two, I find the latter to be easier to maintain. To make the existing cached op thread safe (as @rondogency pointed out), it would require training support, bulking support, multi context support, subgraph api support, dynamic shape support, ensuring that there are no perf regressions. At the very least, there is a lot of additional testing required. 

> Is there a plan of future works of thread_safe cached op?

For the future works of the thread_safe cached op, at this point we want to continue maintaining and addressing customer feedback for the current thread safe cached op version and address performance issues. 

> Do we plan to merge it back to original cached op?

I am not aware of a plan for this currently.

anirudh2290:
@rondogency I am happy to add more comments for the thread safe cached op or the code that I have touched in this PR.

anirudh2290:
> On the other hand, I am not sure if we can accept this cached op without bulking and subgraph support. Subgraph is used for multiple accelerators for MXNet inference. Also, bulking is essential to inference speed on GPUs. For instance, we saw 20% perf regression if bulking scope is smaller than the optimal scope #9055 and it would be even more significant if no bulking is enabled at all. I am not sure if this can be shipped with these unknown limitations. Thinking about how to better support thread safe bulking and subgraph may affect the implementation and design.

1. I have added bulking support. 

2. For subgraphing, If I am correct, MXNet still doesnt support subgraph API with gluon. It is still supported only with symbol and MXNET_SUBGRAPH_BACKEND env variable works only with graph executor. What I meant to say is that the param subgraph in CachedOp is not supported. Now you can use a symbol which has already been converted and replaced with subgraph ops with the cached_op_threadsafe version. I have added a test to demonstrate that. I think this should address your concern with respect to subgraphing.

> Do we have evidence that the newly introduced cached op is performant? I'm concerned about the mutex for the whole forward function - that means no thread can concurrently push operations to the engine. What I observed is - that could take a long time depending on the model used.

Although the mutex in forward is just pushing the ops, which may have delays between threads, the execution of these ops will be scheduled in parallel, especially when there is no dependency from the ops different threads and should improve performance.

https://github.com/awslabs/djl did some benchmarks and were able to see improved throughput upto 2x with naive engine. I was also able to see performance improvements upto 1.7x when you increase the number of worker threads on the threaded engine for cpu.

anirudh2290:
I have addressed all the comments here. Please let me know if you have additional comments.

pengzhao-intel:
We're on the long vocation and will back to office in 10th Feb.

anirudh2290:
Thanks a lot for your kind words Chris! Appreciate your dedication towards the mxnet community!

cjolivier01:
hahaha...

larroy:
Did you run performance benchmarks to see that there's no regression? as I understand this is a change that could have an impact on performance in the engine. Would be useful to run the usual models we have in the performance dashboards.

leezu:
@anirudh2290 this seems to have broken C++ examples. See https://github.com/apache/incubator-mxnet/issues/17514

eric-haibin-lin:
@zhreshold FYI

PR #16555, opened 2019-10-20
Upgrade MKL-DNN dependency to v1.0

TaoLv:
## Description ##
This is an effort from the whole Intel MXNet team. Credits belong to everyone in the team.

This PR upgrades the 3rdparty/mkldnn dependency to it's v1.0.x release which has many API breaking changes and is not backward compatible to the previous v0.x versions. Hence this PR also changes the integration code in operators while keep the integration methodology and software architecture the same. Development has been done on the feature branch `mkldnn-v1.0` and tracked via github project: https://github.com/apache/incubator-mxnet/projects/16. 

Please see the discussion here: https://lists.apache.org/thread.html/f46ab920f18795496eafe713e6e9e561c684e06189085cec17b401dc@%3Cdev.mxnet.apache.org%3E
Please see MKL-DNN v1.0 RFC here: https://github.com/intel/mkl-dnn/tree/rfc-api-changes-v1.0/doc/rfc/api-v1.0

As mentioned in the dev@ thread, this PR also removes the mklml and iomp5 library which previously are distributed along with MXNet pip package. So it also fixes the license issue in https://github.com/apache/incubator-mxnet/issues/15544. It's a requirement for 1.6.0 release.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at https://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
[...]

PR #16663, opened 2019-10-29
[mkldnn-v1.0] Use memcpy instead of set_handle with unidirectional 1-layer RNN

zixuanweeei:
## Description ##
Fix the issue filed by https://github.com/apache/incubator-mxnet/pull/16555#issuecomment-547000155. Previously, we don't need `Concat` to concatenate pieces of weights to a continuous space with unidirectional 1-layer RNN. Thus, `mkldnn::memory::set_data_handle(ptr)` is directly used to fetch the weights, and `ptr` is the pointer of weights shifted from `NDArray`. When it runs into `gru` mode, a swap mechanism of gates is used to swap weights of the reset gate and the update gate. If we use `set_data_handle` here, it will crash the naive memory from `NDArray`.  As a result, the gru operator swaps the naive memory every iteration when `is_train=True`. It will then give inconsistent results. We should use `memcpy` here.

### Changes ###
- [x] Move `set_data_handle` to `memcpy`.

### DIFF
--- a/src/operator/nn/mkldnn/mkldnn_rnn.cc
+++ b/src/operator/nn/mkldnn/mkldnn_rnn.cc
@@ -483,8 +483,10 @@ void MKLDNNRnnForward::SetWeightsMem(MKLDNNRnnMemMgr* mgr, void *w_ptr, void *b_
   char *weights_ptr = static_cast<char *>(w_ptr);
   size_t wx_bytes = GetRnnGatesNum(param_.mode) * param_.state_size *
         param_.input_size * dtype_bytes;  //* DIMS: ngates x state_size x input_size
+  size_t wh_bytes = GetRnnGatesNum(param_.mode) * param_.state_size *
+        param_.state_size * dtype_bytes;  //* DIMS: ngates x state_size x state_size
   char *l2r_wx = weights_ptr;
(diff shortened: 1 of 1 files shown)


ZhennanQin:
This change will impact performance. Why v0.20 doesn't have this issue? Is there any solutions other than memcpy?

zixuanweeei:
> This change will impact performance. Why v0.20 doesn't have this issue? Is there any solutions other than memcpy?

v0.20 swaps the GRU memory on native NDArray space and then swaps back to the original layout. It may have more overhead over the first iteration. Thus, v1.0 uses a different mechanism that directly adjusts `mkldnn::memory`. Previously, we use OpenMP for-loop to apply the swap and copy. I have had an experiment measuring the performance of the operator from stl and the one wrapped by OpenMP. It seems that `std::memcpy` needs another ~4 milliseconds (clx-8280) to finish the operation with 1e7 long float array. It may not severely impact the performance in the inference procedure, as we copy and swap the memory only once at the first iteration. Honestly, it is not true in forward_training and backward. There are lots of `memcpy` to deliver the gradients of weights. I will try to simplify and accelerate them in this PR.

pengzhao-intel:
@zixuanweeei take a look for the CI failure

zixuanweeei:
`IMPERATIVE.FullyConnectedOp` failed again. @pengzhao-intel

Issue #16658, opened 2019-10-28
[Numpy] Boolean array bug


sxjscience:
```python
import mxnet as mx
import numpy as np
from mxnet.gluon import HybridBlock
mx.npx.set_np()

class CastBlock(HybridBlock):
	def hybrid_forward(F, x):
		return x.astype(np.float32)

f_cast = CastBlock()
f_cast.hybridize()
out = f_cast(mx.np.ones((10,)) > 0)
print(out)
```

```
MXNetError: [09:43:24] src/operator/tensor/elemwise_unary_op.h:461: Unknown type enum 7
```

sxjscience:
The problem only occurs after I called hybridize.

sxjscience:
After offline discussion with @haojin2, it could be solved by fixing this line https://github.com/apache/incubator-mxnet/blob/86ed5f5c0a72d1e34b29b0396432d04888274ce2/src/operator/tensor/elemwise_unary_op.h#L454 to add WITH_BOOL

reminisce:
boolean ndarrays are only supported in a few ops right now. We need to enable them gradually in more ops. Thanks for reporting this.

sxjscience:
@reminisce I agree. At least we should fix cast, otherwise we cannot use `>`, `<`.

reminisce:
@sxjscience  `astype` should be working with boolean ndarrays. I think using `cast` is not as pythonic as using `astype`.

sxjscience:
@reminisce In this case, `.astype` will not trigger the error if the block is not hybridized. However, once it's hybridized, it will report the error message.

reminisce:
@sxjscience I will investigate to see why there is a such difference.

reminisce:
In imperative mode, `astype` used `copy` operator which has fully supported boolean ndarrays, while in symbolic mode, `astype` calls `cast`, which has not been fully enabled to support boolean ndarrays.

Issue #16650, opened 2019-10-28
[Bug][Numpy] Cannot expand_dims of bool array


sxjscience:
```python
import mxnet as mx
mx.npx.set_np()
a = mx.np.ones((10,10))
b = mx.np.expand_dims(a < 1, axis=1)
print(b)
```

Error:
```
MXNetError: [03:33:12] src/operator/nn/mkldnn/../../tensor/./../mxnet_op.h:670: Unknown type enum 7
Stack trace:
  [bt] (0) /home/ubuntu/mxnet/python/mxnet/../../lib/libmxnet.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x32) [0x7fd31c83eac2]
  [bt] (1) /home/ubuntu/mxnet/python/mxnet/../../lib/libmxnet.so(void mxnet::op::UnaryOp::IdentityCompute<mshadow::cpu>(nnvm::NodeAttrs const&, mxnet::OpContext const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&, std::vector<mxnet::OpReqType, std::allocator<mxnet::OpReqType> > const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&)+0x154b) [0x7fd31c939bcb]
[...]

PR #16664, opened 2019-10-29
Miscellaneous fix for several numpy issues

haojin2:
## Description ##
#16658 #16650 and inconsistent type when an array is created from an official numpy array:
```Python
from mxnet import np
import numpy as _np
a = np.array(_np.ones((3, ), dtype=_np.float16))
# a's dtype should be float16
# while the previous implementation it was float32
# after this fix it should become float16
```

### Essentials ###

### Changes ###
- [x] Fix for `expand_dims`'s boolean support
- [x] Fix for #16658 
- [x] Make resulting ndarray's dtype consistent with a source official numpy ndarray
- [x] corresponding tests

### DIFF
--- a/python/mxnet/numpy/multiarray.py
+++ b/python/mxnet/numpy/multiarray.py
@@ -577,7 +577,7 @@ class ndarray(NDArray):
             if not isinstance(key, tuple) or len(key) != 0:
                 raise IndexError('scalar tensor can only accept `()` as index')
             if isinstance(value, numeric_types):
-                self.full(value)
+                self._full(value)
             elif isinstance(value, ndarray) and value.size == 1:
                 if value.shape != self.shape:
                     value = value.reshape(self.shape)
@@ -1993,10 +1993,12 @@ def array(object, dtype=None, ctx=None):
     """
     if ctx is None:
         ctx = current_context()
-    if isinstance(object, ndarray):
+    if isinstance(object, (ndarray, _np.ndarray)):
         dtype = object.dtype if dtype is None else dtype
     else:
         dtype = _np.float32 if dtype is None else dtype
+        if hasattr(object, "dtype"):
+            dtype = object.dtype
         if not isinstance(object, (ndarray, _np.ndarray)):
             try:
                 object = _np.array(object, dtype=dtype)
--- a/python/mxnet/numpy/multiarray.py
+++ b/python/mxnet/numpy/multiarray.py
@@ -1996,13 +1996,16 @@ def array(object, dtype=None, ctx=None):
     if isinstance(object, (ndarray, _np.ndarray)):
         dtype = object.dtype if dtype is None else dtype
     elif isinstance(object, NDArray):
-        raise ValueError("")
+        raise ValueError("If you're trying to create a mxnet.numpy.ndarray "
+                         "from mx.nd.NDArray, please use the zero-copy as_np_ndarray function.")
     else:
         if dtype is None:
             dtype = object.dtype if hasattr(object, "dtype") else _np.float32
         try:
             object = _np.array(object, dtype=dtype)
         except Exception as e:
+            # printing out the error raised by official NumPy's array function
+            # for transparency on users' side
             raise TypeError('{}'.format(str(e)))
     ret = empty(object.shape, dtype=dtype, ctx=ctx)
     if len(object.shape) == 0:
--- a/python/mxnet/numpy/multiarray.py
+++ b/python/mxnet/numpy/multiarray.py
@@ -1995,15 +1995,15 @@ def array(object, dtype=None, ctx=None):
         ctx = current_context()
     if isinstance(object, (ndarray, _np.ndarray)):
         dtype = object.dtype if dtype is None else dtype
+    elif isinstance(object, NDArray):
+        raise ValueError("")
     else:
-        dtype = _np.float32 if dtype is None else dtype
-        if hasattr(object, "dtype"):
-            dtype = object.dtype
-        if not isinstance(object, (ndarray, _np.ndarray)):
-            try:
-                object = _np.array(object, dtype=dtype)
-            except Exception as e:
-                raise TypeError('{}'.format(str(e)))
+        if dtype is None:
+            dtype = object.dtype if hasattr(object, "dtype") else _np.float32
+        try:
+            object = _np.array(object, dtype=dtype)
+        except Exception as e:
+            raise TypeError('{}'.format(str(e)))
     ret = empty(object.shape, dtype=dtype, ctx=ctx)
     if len(object.shape) == 0:
         ret[()] = object
--- a/src/operator/tensor/elemwise_unary_op.h
+++ b/src/operator/tensor/elemwise_unary_op.h
@@ -451,7 +451,7 @@ void CastCompute(const nnvm::NodeAttrs& attrs,
   Stream<xpu> *s = ctx.get_stream<xpu>();
   MSHADOW_TYPE_SWITCH_WITH_BOOL(outputs[0].type_flag_, DstDType, {
     Tensor<xpu, 1, DstDType> out = outputs[0].FlatTo1D<xpu, DstDType>(s);
-    MSHADOW_TYPE_SWITCH(inputs[0].type_flag_, SrcDType, {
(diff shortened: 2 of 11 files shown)


python/mxnet/numpy/multiarray.py
```diff
@@ -1993,10 +1993,12 @@ def array(object, dtype=None, ctx=None):
...
     if ctx is None:
         ctx = current_context()
-    if isinstance(object, ndarray):
+    if isinstance(object, (ndarray, _np.ndarray)):
         dtype = object.dtype if dtype is None else dtype
     else:
         dtype = _np.float32 if dtype is None else dtype
+        if hasattr(object, "dtype"):
+            dtype = object.dtype
         if not isinstance(object, (ndarray, _np.ndarray)):
```
sxjscience: I feel that this if-statement is not necessary because we've tested it in Line 1996.

python/mxnet/numpy/multiarray.py
```diff
@@ -577,7 +577,7 @@ def __setitem__(self, key, value):
             if not isinstance(key, tuple) or len(key) != 0:
                 raise IndexError('scalar tensor can only accept `()` as index')
             if isinstance(value, numeric_types):
-                self.full(value)
+                self._full(value)
```
reminisce: I think we should delete the class method `_full` from `ndarray` since NumPy ndarray class does not have it. Wherever it's used, we should replace it with the direct `full` operator call.

src/operator/tensor/elemwise_unary_op.h
```diff
@@ -451,7 +451,7 @@ void CastCompute(const nnvm::NodeAttrs& attrs,
   Stream<xpu> *s = ctx.get_stream<xpu>();
   MSHADOW_TYPE_SWITCH_WITH_BOOL(outputs[0].type_flag_, DstDType, {
     Tensor<xpu, 1, DstDType> out = outputs[0].FlatTo1D<xpu, DstDType>(s);
-    MSHADOW_TYPE_SWITCH(inputs[0].type_flag_, SrcDType, {
+    MSHADOW_TYPE_SWITCH_WITH_BOOL(inputs[0].type_flag_, SrcDType, {
```
reminisce: Please reorganize `test_np_ndarray_astype` so that this case would be covered. Ideally, we should do
```python
dtypes = [np.int8, np.uint8, np.int32, np.float16, np.float32, np.float64,  np.bool, np.bool_,
          'int8', 'uint8', 'int32', 'float16', 'float32', 'float64', 'bool']
for itype, otype in itertools.product(dtypes, dtypes):
    verify_astype(itype, otype)
```

python/mxnet/numpy/multiarray.py
```diff
@@ -1993,10 +1993,12 @@ def array(object, dtype=None, ctx=None):
     """
     if ctx is None:
         ctx = current_context()
-    if isinstance(object, ndarray):
+    if isinstance(object, (ndarray, _np.ndarray)):
         dtype = object.dtype if dtype is None else dtype
     else:
```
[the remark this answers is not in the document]
haojin2: Okay, will add

python/mxnet/numpy/multiarray.py
haojin2: Sure

src/operator/tensor/elemwise_unary_op.h
haojin2: Done.

python/mxnet/numpy/multiarray.py
```diff
@@ -1993,10 +1993,12 @@ def array(object, dtype=None, ctx=None):
...
         dtype = _np.float32 if dtype is None else dtype
+        if hasattr(object, "dtype"):
+            dtype = object.dtype
```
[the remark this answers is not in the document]
haojin2: Done.

python/mxnet/numpy/multiarray.py
haojin2: removed

python/mxnet/numpy/multiarray.py
haojin2: turns out `_full` is also used in `python/mxnet/ndarray/ndarray.py` for indexing, so keeping it for now.

tests/python/unittest/test_numpy_ndarray.py
```diff
@@ -483,24 +484,29 @@ def __init__(self, dtype, copy):
...
-    for dtype in [np.int8, np.uint8, np.int32, np.float16, np.float32, np.float64, np.bool, np.bool_,
-                  'int8', 'uint8', 'int32', 'float16', 'float32', 'float64', 'bool']:
+    dtypes = [np.int8, np.uint8, np.int32, np.float16, np.float32, np.float64, np.bool, np.bool_,
+              'int8', 'uint8', 'int32', 'float16', 'float32', 'float64', 'bool']
+
+    for itype, otype in itertools.product(dtypes, dtypes):
         for copy in [True, False]:
             for hybridize in [True, False]:
-                check_astype_equal(dtype, copy, copy is False and mx_data.dtype == dtype, hybridize)
+                check_astype_equal(itype, otype, copy, hybridize)
```
haojin2: @reminisce

python/mxnet/numpy/multiarray.py
```diff
@@ -1993,15 +1993,17 @@ def array(object, dtype=None, ctx=None):
     """
     if ctx is None:
         ctx = current_context()
-    if isinstance(object, ndarray):
+    if isinstance(object, (ndarray, _np.ndarray)):
         dtype = object.dtype if dtype is None else dtype
+    elif isinstance(object, NDArray):
+        raise ValueError("")
```
marcoabreu: Add message?

python/mxnet/numpy/multiarray.py
```diff
@@ -1993,15 +1993,17 @@ def array(object, dtype=None, ctx=None):
...
-            try:
-                object = _np.array(object, dtype=dtype)
-            except Exception as e:
-                raise TypeError('{}'.format(str(e)))
+        if dtype is None:
+            dtype = object.dtype if hasattr(object, "dtype") else _np.float32
+        try:
+            object = _np.array(object, dtype=dtype)
+        except Exception as e:
+            raise TypeError('{}'.format(str(e)))
```
marcoabreu: Is such a broad exception handling the right thing? Also, you're stripping all the exception information and just printing the message - nested exception or rethrowing would make more sense here I'd say

marcoabreu: Also, it's missing the message.

haojin2: It's not missing the message here, the exception is being converted into a string and formatted in the `{}`'s position.

python/mxnet/numpy/multiarray.py
haojin2: Will add

python/mxnet/numpy/multiarray.py
marcoabreu: You're stripping away the exception stack (in a broad catch all clause) and further information that would give more information. In how far does removing that info add value?

Could you elaborate what you're trying to achieve here? Catching all exceptions and rethrowing them with stripped information is usually not something I see around

haojin2: So here it's simply calling the official numpy's `array` function, but Numpy does not show any "call stack" at all:
```
>>> np.array(1.0, order="MXNET")
TypeError: order not understood
```
Catching it here and simply print it out is a more transparent way of showing the users the original error I guess...

marcoabreu: Okay, thanks for elaborating! Can you add a short comment to document? Fine with me to move forward

haojin2: Let's see if this round of CI check passes, if it does I'll probably prefer to save 1 CI cycle for this... I'll combine the doc into a future PR maybe, lemme know if you're fine with that. BTW error message has been added with respect to your other comment.

marcoabreu: Fine with me, thanks!

haojin2:
@eric-haibin-lin Good for merge?

Issue #16049, opened 2019-08-30
Initializers should implement `__eq__`


canerturkmen:
## Description
HybridBlocks do not initialize with params when parameter initializers are explicitly given, and two initializers don't match (by reference).

## Environment info (Required)
```
----------Python Info----------
Version      : 3.6.5
Compiler     : GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.10.44.4)
Build        : ('default', 'Aug 19 2019 21:45:20')
Arch         : ('64bit', '')
------------Pip Info-----------
Version      : 9.0.3
Directory    : /Users/caner/VENVS/mxnet/lib/python3.6/site-packages/pip
----------MXNet Info-----------
Version      : 1.5.0
Directory    : /Users/caner/VENVS/mxnet/lib/python3.6/site-packages/mxnet
Commit Hash   : 75a9e187d00a8b7ebc71412a02ed0e3ae489d91f
Library      : ['/Users/caner/VENVS/mxnet/lib/python3.6/site-packages/mxnet/libmxnet.so']
Build features:
✖ CUDA
✖ CUDNN
✖ NCCL
✖ CUDA_RTC
✖ TENSORRT
✔ CPU_SSE
✔ CPU_SSE2
✔ CPU_SSE3
✔ CPU_SSE4_1
✔ CPU_SSE4_2
✖ CPU_SSE4A
✔ CPU_AVX
✖ CPU_AVX2
✖ OPENMP
✖ SSE
✖ F16C
✖ JEMALLOC
✖ BLAS_OPEN
✖ BLAS_ATLAS
✖ BLAS_MKL
✖ BLAS_APPLE
✔ LAPACK
✖ MKLDNN
✔ OPENCV
✖ CAFFE
✖ PROFILER
✔ DIST_KVSTORE
✖ CXX14
✖ INT64_TENSOR_SIZE
✔ SIGNAL_HANDLER
✖ DEBUG
----------System Info----------
Platform     : Darwin-17.7.0-x86_64-i386-64bit
system       : Darwin
node         : caner.local
release      : 17.7.0
version      : Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64
----------Hardware Info----------
machine      : x86_64
processor    : i386
b'machdep.cpu.brand_string: Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz'
[...]

PR #16680, opened 2019-10-31
Implement Initializer.__eq__

leezu:
## Description ##
When sharing parameters between Blocks, gluon checks that all attributes of to-be-shared parameters match. However, as mxnet initializers don't implement `__eq__`, the check will always fail. Thus it is currently impossible to share parameters between Blocks with custom initializers.

See https://github.com/apache/incubator-mxnet/issues/16049

### Essentials ###

### Changes ###
- [X] Support sharing parameters between blocks with custom initializers

## Comments ##
@szha @sxjscience

### DIFF
--- a/python/mxnet/initializer.py
+++ b/python/mxnet/initializer.py
@@ -268,6 +268,11 @@ class Initializer(object):
             '"weight", "bias", "gamma" (1.0), and "beta" (0.0).' \
             'Please use mx.sym.Variable(init=mx.init.*) to set initialization pattern' % name)
 
+    def __eq__(self, other):
+        if not isinstance(other, Initializer):
+            return NotImplemented
+        # pylint: disable=unidiomatic-typecheck
+        return type(self) is type(other) and self._kwargs == other._kwargs
 
 # pylint: disable=invalid-name
 _register = registry.get_register_func(Initializer, 'initializer')
--- a/tests/python/unittest/test_gluon.py
+++ b/tests/python/unittest/test_gluon.py
@@ -3119,6 +3119,21 @@ def test_squeeze_consistency():
         shape = (np.random.randint(1, 10), np.random.randint(1, 10), 1)
         block(mx.nd.ones(shape))
 
+def test_shared_parameters_with_non_default_initializer():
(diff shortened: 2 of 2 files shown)


python/mxnet/initializer.py
```diff
@@ -268,6 +268,10 @@ def _init_default(self, name, _):
             '"weight", "bias", "gamma" (1.0), and "beta" (0.0).' \
             'Please use mx.sym.Variable(init=mx.init.*) to set initialization pattern' % name)
 
+    def __eq__(self, other):
+        if not isinstance(other, Initializer):
+            return NotImplemented
+        return type(self) == type(other) and self._kwargs == other._kwargs
```
szha: does the second statement behave correctly if it involves constant initializer (w/ numpy array, NDArray, and mxnet numpy array)?

leezu: We can say it behaves correctly, if we expect `Constant(10.0) != Constant(np.array([10.0])`.

leezu: With this PR at least `Constant(10.0) == Constant(10.0)`, which was not true before...

tests/python/unittest/test_gluon.py
```diff
@@ -3119,6 +3119,19 @@ def forward(self, x):
...
+    class MyBlock(gluon.HybridBlock):
+        def __init__(self, **kwargs):
+            super(MyBlock, self).__init__(**kwargs)
+
+            with self.name_scope():
+                self.param = self.params.get("param", shape=(1, ), init=mx.init.Constant(-10.0))
+
+    bl = MyBlock()
+    bl2= MyBlock(params=bl.collect_params())
+    assert bl.param is bl2.param
```
wkcn: Thanks for your contribution!
Could you add the unittest 'b1.param is not bl3.param’ and the test for mx.init.Initializer?

leezu: What do you mean? Actually `bl2 = MyBlock(params=bl.collect_params())` performs internally a check that the two initializes are equal, which will fail without this PR.

PR #16700, opened 2019-11-01
[Profiler] Make the trace name unique

joapolarbear:
## Description ##
Current trace names are not unique, for example, if a model consists of two `Conv2d` layers, the corresponding trace names are the same, something like `Convolution`. In order to make each layer in trace results hold a unique name, the following changes have been made:

### Changes ###
- [x]  For Module API, update `src/executor/graph_executor.cc`, substitute `inode.source->op()->name` with `opr_names += inode.source->attrs.name `.
- [x] For Gluon API
** Update `src/engine/threaded_engine.cc`, allow profiler mode to be `kSymbolic`.
** Update `src/imperative/imperative_utils.h`, use `attrs.name` to substitute `op->name`. To maintain the lifecycle of `attrs.name`, use a global variable (type unordered_set string) to store it. To force profile the input and output op, assign priority 1 to the first and last node when calling PushSync, if `attrs.name` is not empty.

An example of the test results is shown below:
```
{
    "name": "hybridsequential0_conv0_fwd",
    "cat": "operator",
    "ph": "B",
    "ts": 1572512095582317,
    "pid": 64,
    "tid": 1623894732768404781
 }
...
{
    "name": "hybridsequential0_conv1_fwd",
    "cat": "operator",
     "ph": "E",
    "ts": 1572512095582642,
    "pid": 64,
    "tid": 1623894732768404781
}
```
Now, `conv` layers is identified with an index, which make them unique in a single step.

### Essentials ###
- [x] To my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### DIFF
--- a/src/engine/naive_engine.cc
+++ b/src/engine/naive_engine.cc
@@ -119,6 +119,15 @@ class NaiveEngine final : public Engine {
   void Push(OprHandle op, Context exec_ctx, int priority = 0, bool profiling = false) override {
     profiler::Profiler *profiler = profiler::Profiler::Get();
     NaiveOpr *opr = op->Cast<NaiveOpr>();
+
+    //huhanpeng
+    if (opr->opr_name) {
+      std::cout << "src/engine/naive_engine.cc:Push" 
+            << " opr_name:" << opr->opr_name
+            << " profiling:" << profiling
+            << std::endl << std::flush;
+    }
+
     opr->profiling = profiling && profiler->IsProfiling(profiler::Profiler::kSymbolic);
     this->PushAsync([&](RunContext ctx, CallbackOnComplete on_complete) {
         if (opr->profiling) {
@@ -160,6 +169,7 @@ class NaiveEngine final : public Engine {
     profiler::Profiler *profiler = profiler::Profiler::Get();
     NaiveOpr *opr = nullptr;
     const bool profiling = opr_name && profiler->IsProfiling(profiler::Profiler::kImperative);
+
     if (profiling) {
       opr = NewOperator(exec_fun, const_vars, mutable_vars,
                         prop, opr_name)->Cast<NaiveOpr>();
--- a/src/engine/naive_engine.cc
+++ b/src/engine/naive_engine.cc
@@ -119,15 +119,6 @@ class NaiveEngine final : public Engine {
   void Push(OprHandle op, Context exec_ctx, int priority = 0, bool profiling = false) override {
     profiler::Profiler *profiler = profiler::Profiler::Get();
     NaiveOpr *opr = op->Cast<NaiveOpr>();
-
-    //huhanpeng
-    if (opr->opr_name) {
-      std::cout << "src/engine/naive_engine.cc:Push" 
-            << " opr_name:" << opr->opr_name
-            << " profiling:" << profiling
-            << std::endl << std::flush;
-    }
-
     opr->profiling = profiling && profiler->IsProfiling(profiler::Profiler::kSymbolic);
     this->PushAsync([&](RunContext ctx, CallbackOnComplete on_complete) {
         if (opr->profiling) {
@@ -169,7 +160,6 @@ class NaiveEngine final : public Engine {
     profiler::Profiler *profiler = profiler::Profiler::Get();
     NaiveOpr *opr = nullptr;
     const bool profiling = opr_name && profiler->IsProfiling(profiler::Profiler::kImperative);
-
     if (profiling) {
       opr = NewOperator(exec_fun, const_vars, mutable_vars,
                         prop, opr_name)->Cast<NaiveOpr>();
--- a/src/engine/threaded_engine.cc
+++ b/src/engine/threaded_engine.cc
@@ -292,6 +292,14 @@ void ThreadedEngine::Push(OprHandle op, Context exec_ctx, int priority, bool pro
   OprBlock* opr_block = OprBlock::New();
   opr_block->opr = threaded_opr;
 
+  //huhanpeng
+  // if (threaded_opr->opr_name) {
+  //   std::cout << "src/engine/threaded_engine.cc:Push" 
+  //         << " opr_name:" << threaded_opr->opr_name
+  //         << " profiling:" << profiling
+  //         << std::endl << std::flush;
+  // }
+
   opr_block->wait.store(static_cast<int>(
       threaded_opr->const_vars.size() +
       threaded_opr->mutable_vars.size() + 1));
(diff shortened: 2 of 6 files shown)


src/engine/threaded_engine.cc
```diff
@@ -335,7 +335,9 @@ void ThreadedEngine::PushAsync(AsyncFn fn, Context exec_ctx,
 #endif
   ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars, prop, opr_name, wait);
   opr->temporary = true;
-  const bool profiling = profiler_->IsProfiling(profiler::Profiler::kImperative);
+  // const bool profiling = profiler_->IsProfiling(profiler::Profiler::kImperative);
```
marcoabreu: Can you remove the commented out code

src/executor/graph_executor.cc
```diff
@@ -1175,7 +1175,8 @@ void GraphExecutor::InitCachedOps() {
   for (uint32_t nid = 0; nid < idx.num_nodes(); ++nid) {
     const auto& inode = idx[nid];
     if (inode.source->is_variable()) continue;
-    op_nodes_[nid].opr_name = inode.source->op()->name.c_str();
+    // op_nodes_[nid].opr_name = inode.source->op()->name.c_str();
```
marcoabreu: Same here

src/executor/graph_executor.cc
```diff
@@ -1557,7 +1558,8 @@ GraphExecutor::CachedSegOpr GraphExecutor::CreateCachedSegOpr(size_t topo_start,
     std::copy(op_node.use_vars.begin(), op_node.use_vars.end(),
               std::inserter(use_vars, use_vars.end()));
     ret.exec_list.push_back(exec);
-    opr_names += inode.source->op()->name + ",";
+    // opr_names += inode.source->op()->name + ",";
```
marcoabreu: And here

src/imperative/imperative_utils.h
```diff
@@ -60,6 +60,9 @@ struct EngineOprSeg {
 
 using MemoryPlanVector = std::vector<MemoryPlanInfo>;
 
+// cached segment operator name (needs a longer lifecycle than cached_seg_opr_)
+static std::unordered_set<std::string> cached_seg_opr_names_;
```
marcoabreu: Could this be a potential memory hug in case a lot of models are instantiated and disposed within the same application lifecycle? Or when does the clean up happen?

src/imperative/imperative_utils.h
```diff
@@ -575,12 +581,12 @@ inline void PushOperator(const OpStatePtr& state,
...
             run(rctx, engine::CallbackOnComplete());
           }, ctx, read_vars, write_vars, FnProperty::kNormal,
-          0, op->name.c_str());
+          0, *iter != "" ? iter->c_str() : op->name.c_str()); // huhanpeng: modified for profiling
     } else {
       CHECK(exec_type == ExecType::kAsync);
       Engine::Get()->PushAsync(
           run, ctx, read_vars, write_vars, FnProperty::kAsync,
-          0, op->name.c_str());
+          0, *iter != "" ? iter->c_str() : op->name.c_str()); // huhanpeng: modified for profiling
```
marcoabreu: You could replace the comment with explaining what it does, e.g. "append attribute name to make profiling unique" or whatever :)

Also generally, we try to avoid people's names within the code unless it's a bug/Todo. Feel free to add yourself to the contributors.md!

src/imperative/imperative_utils.h
```diff
@@ -982,10 +994,11 @@ inline void CreateEngineOpSeg(
 
     // Create opr segment for previous nodes.
     if (stop && nid > seg_start) {
+      opr_name = idx[seg_start].source->attrs.name;
```
marcoabreu: I think you can move this and the other statements below a bit more into the condition to avoid the operation in case the condition is not met.

src/imperative/imperative_utils.h
ChaiBapchya: Yes definitely add your name here - https://github.com/apache/incubator-mxnet/blob/master/CONTRIBUTORS.md

src/engine/threaded_engine.cc
```diff
@@ -335,7 +335,9 @@ void ThreadedEngine::PushAsync(AsyncFn fn, Context exec_ctx,
 #endif
   ThreadedOpr *opr = NewOperator(std::move(fn), const_vars, mutable_vars, prop, opr_name, wait);
   opr->temporary = true;
-  const bool profiling = profiler_->IsProfiling(profiler::Profiler::kImperative);
```
ChaiBapchya: Previously this would check if it's imperative. Now you've extended the scope to Symbolic as well.
Could you explain why? Would be great to either have this commented in code or PR description

src/executor/graph_executor.cc
```diff
@@ -1581,7 +1583,7 @@ GraphExecutor::CachedSegOpr GraphExecutor::CreateCachedSegOpr(size_t topo_start,
     }
     on_complete();
   };
-  opr_names.pop_back();
+  opr_names.pop_back(); // delete the last ','
```
ChaiBapchya: Nitpick: Lint issue - At least two spaces is best between code and comments
As a general suggestion - `make lint` and `make lint` catches Lint issues (recommended to run these before pushing PR / commits

src/imperative/imperative_utils.h
ChaiBapchya: Valid question. I don't see any explicit memory cleanup so I'm guessing it will be cleaned up when it goes out of scope? @joapolarbear you could either 
a. justify why we don't need to explicitly clean up this set 
or better
b. clean it once it is not needed

src/engine/threaded_engine.cc
joapolarbear: OK👌

src/engine/naive_engine.cc
```diff
@@ -223,6 +223,9 @@ class NaiveEngine final : public Engine {
     shutdown_phase_.store(true);
   }
 
+  const char *GetCachedName(std::string opr_names) override {
```
eric-haibin-lin: This leads to build failure: 
```
src/engine/naive_engine.cc: In member function 'virtual const char* mxnet::engine::NaiveEngine::GetCachedName(std::__cxx11::string)':

src/engine/naive_engine.cc:227:3: error: no return statement in function returning non-void [-Werror=return-type]

   }


```

PR #16729, opened 2019-11-05
[NumPy][TVM] NumPy Unary Operator Using TVM Infra

hgt312:
For unary operators, **forward** and **backward-use-none** use the same computation function.
Similarly, **backward-use-in** and **backward-use-out** use the same computation function.

Examples (`deg2rad`, `rad2deg`, and `abs`):
Python part:

- `contrib/tvmop/core/umath/operator.py`
- `contrib/tvmop/core/umath/register.py`

C++ part:

- `src/operator/numpy/np_elemwise_unary_op_basic.cc`
- `src/operator/numpy/np_elemwise_unary_op_basic.cu`


contrib/tvmop/utils.py
```diff
@@ -20,7 +20,16 @@
 
 AllTypes = ["float32", "float64", "float16", "uint8", "int8", "int32", "int64"]
 RealTypes = ["float32", "float64", "float16"]
-
+acc_type_resolver = {
```
reminisce: `acc` usually means accumulation for reduce-like ops. Seems here it's only for unary element-wise ops. Shall we rename to something like `unary_result_type` to echo `np.result_type`? Also, I think it would be better to implement this as a function, instead of using a dict.

hgt312: Yeah. A function is better, but this dict is useful, it should be kept.

reminisce:
Waiting for the fix of large code gen from TVM...

Issue #15968, opened 2019-08-21
1 bit mask for Dropout 


eric-haibin-lin:
Currently the mask generated by dropout op uses the same dtype and shape as the input. 1 bit per coordinate should be sufficient. Since dropout is heavily used after activations, this will save a lot of memory for training. 

@ptrendx @DickJC123 @PatricZhao @TaoLv

frankfliu:
@mxnet-label-bot add [feature request, operator]

eric-haibin-lin:
@apeforest

apeforest:
Working on it now.

PR #16735, opened 2019-11-06
Use single-bit for mask in dropout operator

apeforest:
## Description ##
Use single bit in mask for dropout to reduce memory.
This PR fixes https://github.com/apache/incubator-mxnet/issues/15968

The current implementation does not yet support param axes. It therefore will not pass all unit tests. It is used for benchmarking BERT model for now. I will add support for axes param before merging it to master.


src/operator/nn/dropout.cc
```diff
@@ -119,18 +119,12 @@ Example::
...
     for (int i = 0; i < param.axes.ndim(); ++i) {
       dshape[param.axes[i]] = 1;
     }
-    out_shape->push_back(dshape); */
-  } else {
-    mxnet::TShape mshape(1, static_cast<dim_t>(ceil(static_cast<double>(dshape.Size()) / 8)));
-    out_shape->push_back(mshape);
   }
-
+  mxnet::TShape mshape(1, static_cast<dim_t>(ceil(static_cast<double>(dshape.Size()) / 8)));
```
wkcn: mxnet::TShape mshape(1, static_cast<dim_t>((dshape.Size() + 7) / 8));

apeforest: because it's faster?

src/operator/nn/dropout-inl.h
```diff
@@ -187,28 +188,101 @@ class DropoutOp {
...
                                     const DType *input_data,
                                     const real_t pkeep) {
       RNG_KERNEL_LOOP(xpu, DType, id, gen, N, step, {
         const real_t rand_num = static_cast<real_t>(genImpl.uniform());
-        mask_out[i] = mshadow_op::threshold_eq::Map<real_t>(rand_num, pkeep) * (1.0f / pkeep);
-        dropout_out[i] = input_data[i] * mask_out[i];
-      });
+        // mask_out is set per bit position
+        // therefore bitwise shift need to be performed here
+        auto maskIdx = i / 8;
```
eric-haibin-lin: maskIdx -> mask_idx

eric-haibin-lin: Same comment for offset, val

src/operator/nn/dropout-inl.h
```diff
@@ -187,28 +188,101 @@ class DropoutOp {
...
+        auto maskOffset = i % 8;
+        bool maskVal = mshadow_op::threshold_eq::Map<real_t>(rand_num, pkeep);
+        if (maskVal) {
+          // set bit
+          mask_out[maskIdx] |= 1U << maskOffset;
```
eric-haibin-lin: will this lead to race condition if the same maskIdx is being set by multiple threads? Shall each thread handle at least 8 bits?

src/operator/nn/dropout-inl.h
```diff
@@ -187,28 +188,101 @@ class DropoutOp {
...
+  };
+
+  struct DropoutBackwardKernel {
+    MSHADOW_XINLINE static void Map(index_t i,
+                                    OpReqType req,
+                                    DType *igrad,
+                                    DType *ograd,
+                                    const uint8_t *mask,
+                                    const real_t pkeep) {
+      auto maskIdx = i / 8;
```
eric-haibin-lin: maskIdx-> mask_idx. same comments for offset, val

src/operator/nn/dropout-inl.h
```diff
@@ -187,28 +188,101 @@ class DropoutOp {
...
+                                    DType *dropout_out,
+                                    uint8_t *mask_out,
                                     const real_t pkeep) {
       RNG_KERNEL_LOOP(xpu, DType, id, gen, N, step, {
         const real_t rand_num = static_cast<real_t>(genImpl.uniform());
-        mask_out[i] = mshadow_op::threshold::Map<real_t>(rand_num, pkeep) * (1.0f / pkeep);
-      });
+        // mask_out is set per bit position
+        // therefore bitwise shift need to be performed here
+        auto maskIdx = i / 8;
```
eric-haibin-lin: will this lead to race condition?

src/operator/nn/dropout.cc
wkcn: Yes : )

src/operator/nn/dropout-inl.h
apeforest: will do

src/operator/nn/dropout.cc
```diff
@@ -118,14 +118,18 @@ Example::
...
+    }
   }
-  out_shape->push_back(dshape);
+  mxnet::TShape mshape(1, static_cast<dim_t>(ceil(static_cast<double>(dshape.Size()) / 8)));
+  out_shape->push_back(mshape);
   return true;
 })
 .set_attr<nnvm::FInferType>("FInferType", [](const nnvm::NodeAttrs& attrs,
       std::vector<int> *in_type, std::vector<int> *out_type) {
+  using namespace mshadow;
```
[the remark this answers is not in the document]
apeforest: will remove

src/operator/nn/dropout-inl.h
apeforest: Good catch. I was thinking of setting the step to 8 but forgot to update it in the macro.

src/operator/nn/dropout-inl.h
apeforest: will change

src/operator/nn/dropout-inl.h
apeforest: After more careful checking, actually the race condition will not happen in this implementation. The reason is the `step` variable passed to the `Kernel<OP, xpu>::Launch()` is always a multiple of 8. And therefore, one thread will sequentially process a multiple of 8 items in the `RNG_KERNEL_LOOP`. 

This logic is not clearly seen from this implementation and has dependency on RandGenerator<xpu>::kMinNumRandomPerThread (currently is 64). I will try to make it more explicit in the code for better readability as well as bug-prone.

src/operator/nn/dropout-inl.h
apeforest: After checking into it more, I found ideally this should not happen because `RandGenerator<xpu>::kMinNumRandomPerThread` is 64 and therefore by design the step size inside `LaunchRNG` should be a multiple of 8. But then I looked into that piece of code again and found it looks like a bug in calculating the `step`. Please review my latest change in  src/operator/random/sampler.h and let me know if it makes sense. Thanks.

src/operator/nn/dropout-inl.h
apeforest: See comment above.

src/operator/nn/dropout-inl.h
```diff
@@ -440,31 +519,38 @@ class DropoutOp {
         }
 #endif  // MXNET_USE_CUDNN_DROPOUT && defined(__CUDACC__)
         // standard case for dropout
-        CHECK_EQ(grad.Size(), mask.Size());
+        CHECK_LE(grad.Size(), mask.Size() * 8);
```
TaoLv: `CHECK_EQ((grad.Size() + 7) / 8, mask.Size());` ?

tests/python/unittest/test_operator.py
```diff
@@ -6955,7 +6955,7 @@ def test_stack():
 
 
 @with_seed()
-@unittest.skip("test fails intermittently. temporarily disabled till it gets fixed. tracked at https://github.com/apache/incubator-mxnet/issues/14288")
+#@unittest.skip("test fails intermittently. temporarily disabled till it gets fixed. tracked at https://github.com/apache/incubator-mxnet/issues/14288")
```
TaoLv: remove?

src/operator/nn/dropout-inl.h
TaoLv: Is this for loop parallelized?

tests/python/unittest/test_operator.py
apeforest: This test is still flaky. I commented it out only to test my PR. I suppose we should enable it after the flakiness issue is resolved?

src/operator/nn/dropout-inl.h
```diff
@@ -187,28 +188,98 @@ class DropoutOp {
...
+    MSHADOW_XINLINE static void Map(index_t i,
+                                    OpReqType req,
+                                    DType *igrad,
+                                    DType *ograd,
+                                    const uint8_t *mask,
+                                    const real_t pkeep) {
+      auto mask_idx = i / 8;
+      uint8_t mask_offset = i % 8;
+      bool mask_val = (mask[mask_idx] >> mask_offset) & 1U;
+      KERNEL_ASSIGN(igrad[i], req, mask_val * ograd[i] * (1 / pkeep));
```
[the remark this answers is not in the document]
apeforest: line 204 and 207 are set/clear bit. Line 223 is get bit. Could you please explain why the expression 
```
bool mask_val = mask[mask_idx] & (1U << mask_offset);
```
is better?

src/operator/nn/dropout-inl.h
```diff
@@ -396,10 +475,9 @@ class DropoutOp {
               mshadow::Shape<NDim> rstride = mxnet_op::calc_stride(new_rshape.get<NDim>());
               mxnet_op::Kernel<mxnet_op::binary_broadcast_kernel<NDim, mshadow_op::mul>, xpu>::
               template LaunchEx(s, new_oshape.Size(), req[dropout::kOut],
-              lstride, rstride, oshape,
-              in.dptr<DType>(),
-              mask.dptr<DType>(), out.dptr<DType>());
-            });
+                                lstride, rstride, oshape, in.dptr<DType>(),
+                                temp.dptr_, out.dptr<DType>());
+            })
```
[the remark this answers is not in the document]
apeforest: Added

src/operator/nn/dropout-inl.h
apeforest: Done

src/operator/nn/dropout-inl.h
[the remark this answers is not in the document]
TaoLv: Hah, no strong opinion. Just see doing shift on immediate number more often. Feel free to pick either one.

tests/python/unittest/test_operator.py
eric-haibin-lin: Can we enable the non-cudnn tests, and mark the cudnn dropout test as flaky? Otherwise this PR is not checked with unit test

src/operator/nn/dropout-inl.h
eric-haibin-lin: In general I do not recommend writing code this way. There is not documenntation nor  guarantee that `kMinNumRandomPerThread` will always be greater than 8 in the future. Nor does the dropout operator document any assumption about the value of `kMinNumRandomPerThread`. The code is delicate and will be broken if some contributor changes `kMinNumRandomPerThread` to values like 4. If there's any assumption, we should add an explicit check so that it won't be broken in the future

apeforest: Fair point. I will refactor this piece of code.

src/operator/nn/dropout-inl.h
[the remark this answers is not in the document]
apeforest: Thanks for explanation. Updated.

TaoLv:
@apeforest Thank you for the nice work! Do you have any numbers to share?
- memory usage of a model in which dropout workspace used to be a problem?
- operator performance benchmark?

eric-haibin-lin:
For GPT-2, the memory usage goes from 30GB to 26GB. For BERT, it goes from 26GB to 23GB. I didn't notice much difference in training throughput.

apeforest:
@TaoLv Thanks for your review. I ran operator profining using benchmark.opperf.utils.benchmark_utils.run_performance_test. The result shows speed up in forward but some degradation in backward pass.

w/ this change:
```
[{'Dropout': [{'avg_time_forward_Dropout': 1.3266, 'max_storage_mem_alloc_cpu/0': 4259.8398, 'avg_time_backward_Dropout': 0.2682, 'inputs': {'data': (1024, 1024), 'p': 0.5}}]}]
```

w/o this change:
```
[{'Dropout': [{'avg_time_forward_Dropout': 1.7864, 'max_storage_mem_alloc_cpu/0': 6291.4561, 'avg_time_backward_Dropout': 0.1836, 'inputs': {'data': (1024, 1024), 'p': 0.5}}]}]
```

TaoLv:
@apeforest Thank you for testing it out. Given memory is not always a concern, can we make bit mask an option for dropout?

eric-haibin-lin:
@TaoLv I don't think adding an option is necessary. can we improve the backward kernel?

TaoLv:
@apeforest Could you please also test the operator performance with USE_BLAS=mkl?

apeforest:
@TaoLv I had problem building/running mxnet with USE_BLAS=mkl. I will appreciate if you could give a hand: https://github.com/apache/incubator-mxnet/issues/17366

eric-haibin-lin:
@TaoLv why is BLAS related to this PR???

TaoLv:
@eric-haibin-lin some VSL functions [1] are used to generate random numbers for dropout. VSL is part of MKL library [2].

[1] https://github.com/apache/incubator-mxnet/blob/master/src/operator/nn/dropout-inl.h#L108
[2] https://software.intel.com/en-us/mkl-developer-reference-c-statistical-functions

apeforest:
Update: Merged https://github.com/apache/incubator-mxnet/pull/17403  to enable unit tests of dropout on CPU and GPU w/o cuDNN. I will build MXNet with MKL and run the perf test again.

apeforest:
@TaoLv With MKL Dropout:
[✖ CUDA, ✖ CUDNN, ✖ NCCL, ✖ CUDA_RTC, ✖ TENSORRT, ✔ CPU_SSE, ✔ CPU_SSE2, ✔ CPU_SSE3, ✔ CPU_SSE4_1, ✔ CPU_SSE4_2, ✖ CPU_SSE4A, ✔ CPU_AVX, ✖ CPU_AVX2, ✔ OPENMP, ✖ SSE, ✔ F16C, ✖ JEMALLOC, ✖ BLAS_OPEN, ✖ BLAS_ATLAS, ✔ BLAS_MKL, ✖ BLAS_APPLE, ✔ LAPACK, ✔ MKLDNN, ✔ OPENCV, ✖ CAFFE, ✖ PROFILER, ✖ DIST_KVSTORE, ✖ CXX14, ✖ INT64_TENSOR_SIZE, ✖ SIGNAL_HANDLER, ✔ DEBUG, ✖ TVM_OP]

This PR:
```
[{'Dropout': [{'avg_time_Dropout': 28.989554492291063, 'p50_time_Dropout': 30.037737917155027, 'p90_time_Dropout': 30.27475003618747, 'p99_time_Dropout': 30.33297127345577, 'inputs': {'data': (1024, 1024)}}]}]
[{'Dropout': [{'avg_time_forward_Dropout': 25.1095, 'avg_time_backward_Dropout': 5.6493, 'max_storage_mem_alloc_cpu/0': 4259.8398, 'inputs': {'data': (1024, 1024)}}]}]
```

Master:
```
[{'Dropout': [{'avg_time_Dropout': 2.1497764391824603, 'p50_time_Dropout': 1.9523266237229109, 'p90_time_Dropout': 2.6631804881617427, 'p99_time_Dropout': 2.8045210894197234, 'inputs': {'data': (1024, 1024)}}]}]
[{'Dropout': [{'avg_time_forward_Dropout': 0.7749, 'avg_time_backward_Dropout': 0.3057, 'max_storage_mem_alloc_cpu/0': 6291.4561, 'inputs': {'data': (1024, 1024)}}]}]
```

eric-haibin-lin:
so we get 15x slow down?

apeforest:
> so we get 15x slow down?

Yes, it seems so. Not sure if it is because the extra bit-operation logic we introduced or the modification to omp parallel clause. Need more investigation. @TaoLv Can MKL VSL support bit-wise mask so we don't have to do the extra conversion which seems redundant in the MKL case.

apeforest:
@eric-haibin-lin The big slow down was due to the use of dynamic loop scheduling in omp when assigning a chunk of 8 to each thread. After I changed it to static scheduling, the performance is improved, but still slower than master due to the extra bit operations needed in the kernel and the fact that we need to assign a batch of 8 when parallelizing the for loop. @TaoLv Any other suggestion to speed up the omp section? Thanks.

This PR:
```
[{'Dropout': [{'avg_time_Dropout': 3.4434730419889092, 'p50_time_Dropout': 3.2913365866988897, 'p90_time_Dropout': 4.017965029925109, 'p99_time_Dropout': 4.7498174966312945, 'inputs': {'data': (1024, 1024)}}]}]
```

Master:
```
[{'Dropout': [{'avg_time_Dropout': 2.1684831893071532, 'p50_time_Dropout': 1.962667447514832, 'p90_time_Dropout': 2.8079885290935636, 'p99_time_Dropout': 2.967591730412097, 'inputs': {'data': (1024, 1024)}}]}]
```

apeforest:
@TaoLv Could you please help to review the change to MKL Dropout part? Thanks

TaoLv:
> Can MKL VSL support bit-wise mask so we don't have to do the extra conversion which seems redundant in the MKL case.

Sorry, I'm not aware of that. I will take a look at the code changes. Will you re-consider the suggestion of making bit-mask an option?

TaoLv:
@apeforest  Could you please share your benchmarking scripts?

src/operator/nn/dropout-inl.h
```diff
@@ -123,25 +124,33 @@ class DropoutOp {
...
+      outptr[i] = dataptr[i] * mkl_mask[i] * pk_1;
+      auto mask_idx = i >> 3;  // div 8
+      uint8_t mask_offset = i & 7;  // mod 8
+      if (mkl_mask[i]) {
+        // set bit
+        mask.dptr_[mask_idx] |= 1U << mask_offset;
+      } else {
+        // clear bit
+        mask.dptr_[mask_idx] &= ~(1U << mask_offset);
+      }
```
TaoLv: Can you try the below code snippet? Blocking might help on cache contention.
```cpp
    const int nblk = count / 64;

#pragma omp parallel for num_threads(nthr) schedule(static, 8)
    for (index_t nb = 0; nb < nblk; ++nb) {
      for (index_t k = 0; k < 64; ++k) {
        const index_t i = nb * 64 + k;
        outptr[i] = dataptr[i] * mkl_mask[i] * pk_1;
        auto mask_idx = i >> 3;  // div 8
        uint8_t mask_offset = i & 7;  // mod 8
        if (mkl_mask[i]) {
          // set bit
          mask.dptr_[mask_idx] |= 1U << mask_offset;
        } else {
          // clear bit
          mask.dptr_[mask_idx] &= ~(1U << mask_offset);
        }
      }
    }

    // tail
    if (nblk * 64 < count) {
      for (index_t i = nblk * 64; i < count; ++i) {
        outptr[i] = dataptr[i] * mkl_mask[i] * pk_1;
        auto mask_idx = i >> 3;  // div 8
        uint8_t mask_offset = i & 7;  // mod 8
        if (mkl_mask[i]) {
          // set bit
          mask.dptr_[mask_idx] |= 1U << mask_offset;
        } else {
          // clear bit
          mask.dptr_[mask_idx] &= ~(1U << mask_offset);
        }
      }
    }
```

apeforest:
@TaoLv Thanks for your review and suggestion. 

Using blocking does help:

```
[{'Dropout': [{'avg_time_Dropout': 2.9556092573329806, 'p50_time_Dropout': 2.9536145739257336, 'p90_time_Dropout': 2.9735330026596785, 'p99_time_Dropout': 3.0410749791190033, 'inputs': {'data': (1024, 1024)}}]}]
```

Here is my performance script:

```
#!/usr/bin/python
import mxnet as mx
from mxnet import nd

mx.random.seed(17)
res = run_performance_test(nd.Dropout, run_backward=True, dtype='float32', ctx=mx.cpu(),
                           inputs=[
                               {"data" : (1024, 1024)}
                           ],
                           warmup=20, runs=100, profiler='python')
print(res)
```

I am not in favor of adding an option. It exposes internal implementation (e.g. using bit mask or not) to users and is also counterintuitive (why not use bit-mask). If sacrificing performance (to some extent) can help improve usability, I think we need to consider the trade off.

apeforest:
@eric-haibin-lin I posted the latest performance results in the PR description. Please review this PR again and let me know if it's good to ship.

eric-haibin-lin:
I'm ok with the result. @TaoLv any concern?

TaoLv:
> I'm ok with the result. @TaoLv any concern?

It's still 1.36x slower. I will take another look today.

> If sacrificing performance (to some extent) can help improve usability, I think we need to consider the trade off.

As I mentioned above, I'm not taking this as a general usability issue. I don't think we want to sacrifice the performance on CPU while memory size is not a concern there. @pengzhao-intel

pengzhao-intel:
> > I'm ok with the result. @TaoLv any concern?
> 
> It's still 1.36x slower. I will take another look today.
> 
> > If sacrificing performance (to some extent) can help improve usability, I think we need to consider the trade off.
> 
> As I mentioned above, I'm not taking this as a general usability issue. I don't think we want to sacrifice the performance on CPU while memory size is not a concern there. @pengzhao-intel

It will be a concern for the performance drop because we are working on model training recently.
I didn't follow up on all discussions in the thread. One quick question:
Does the slow down come from more computation in the new algorithm or the sub-optimal implementation?

apeforest:
@roywei Using the test script in https://github.com/apache/incubator-mxnet/pull/13896
Build | runtime (before) | runtime (after) 
---|---|---
CPU w/ MKL | 262 ms ± 1.2 ms | 337 ms ± 12.5 ms


Using python timer to measure CPU performance with MKL:

This PR:

```
[{'Dropout': [{'avg_time_Dropout': 1.1714265774935484, 'p50_time_Dropout': 1.1715246364474297, 'p90_time_Dropout': 1.190436165779829, 'p99_time_Dropout': 1.2154309218749404, 'inputs': {'data': (1024, 1024)}}]}]
```

Master:
```
[{'Dropout': [{'avg_time_Dropout': 0.6394564639776945, 'p50_time_Dropout': 0.6996351294219494, 'p90_time_Dropout': 1.045508868992329, 'p99_time_Dropout': 1.59036863129586, 'inputs': {'data': (1024, 1024)}}]}]
```

TaoLv:
Does the `avg_time_Dropout` include backward time? @apeforest

src/operator/nn/dropout-inl.h
```diff
@@ -152,15 +181,20 @@ class DropoutOp {
...
+    const uint8_t *maskptr = mask.dptr_;
+    const index_t count = grad.shape_[0] * grad.shape_[1];
+    const float pk_1 = 1.0f / this->pkeep_;
+    const int nthr = engine::OpenMP::Get()->GetRecommendedOMPThreadCount();
+#pragma omp parallel for num_threads(nthr) schedule(static, 8)
+    for (index_t i = 0; i < count; ++i) {
+      auto mask_idx = i >> 3;  // div 8;
+      uint8_t mask_offset = i & 7;  // mod 8
+      bool mask_val = maskptr[mask_idx] & (1U << mask_offset);
+      ingradptr[i] = outgradptr[i] * mask_val * pk_1;
```
TaoLv: Let's also use blocking in the backward path:

```cpp
    const int blk_size = 64;
    const int nblk = count / blk_size;

#pragma omp parallel for num_threads(nthr) schedule(static, 8)
    for (index_t b = 0; b < nblk; ++b) {
      for (index_t k = 0; k < blk_size; ++k) {
        index_t i = b * blk_size + k;
        auto mask_idx = i >> 3;  // div 8;
        uint8_t mask_offset = i & 7;  // mod 8
        bool mask_val = maskptr[mask_idx] & (1U << mask_offset);
        ingradptr[i] = outgradptr[i] * mask_val * pk_1;
      }
    }

    // tail
    if (nblk * blk_size < count) {
      for (index_t i = nblk * blk_size; i < count; ++i) {
        auto mask_idx = i >> 3;  // div 8;
        uint8_t mask_offset = i & 7;  // mod 8
        bool mask_val = maskptr[mask_idx] & (1U << mask_offset);
        ingradptr[i] = outgradptr[i] * mask_val * pk_1;
      }
    }
  }
```

apeforest:
> Does the `avg_time_Dropout` include backward time? @apeforest

Yes, it includes backward time as my `run_backward` is set to True

src/operator/nn/dropout-inl.h
apeforest: Sure

apeforest: After more thoughts, I think we actually don't need to do blocking in the backward pass as there is no write to maskptr and hence no race condition either.

apeforest:
> Does the slow down come from more computation in the new algorithm or the sub-optimal implementation?
@PatricZhao The slowdown comes from extra computation in the new algorithm when dropout uses MKL. MKL already does the masking but stores each mask as integer. The new algorithm simply repackage this int32 based mask into bit-based mask and therefore introduced extra runtime. In the ideal case, it would be to enhance MKL dropout to store mask using bits. But it requires modification of the VSL APIs.

src/operator/nn/dropout-inl.h
TaoLv: We're writing to `ingradptr`. We also hope the elements in one cache line will be handled by one openmp thread. With the original parallelization, one cache line is loaded and only one element in it is handled by the current thread. For the next thread, it need load the same cache line, and handle the next element.

apeforest: However there is no read from `ingradptr`, therefore this is not a case of the false sharing, right? I tried this block and didn't noticed any noticeable performance gain.

TaoLv:
> Does the slow down come from more computation in the new algorithm or the sub-optimal implementation?

The new implementation increases both memory load and additional bit-wise operations. So performance slow down is expected.

pengzhao-intel:
> > Does the slow down come from more computation in the new algorithm or the sub-optimal implementation?
> 
> The new implementation increases both memory load and additional bit-wise operations. So performance slow down is expected.

What algorithm is used in TF and pytorch?

TaoLv:
> What algorithm is used in TF and pytorch?

@pengzhao-intel I don't think TF has a fused dropout operator. It's implement with several small operators. See https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/nn_ops.py#L4456. So the backward path should go through the backward of these small operators. Hence no bit-mask there.

For PyTorch, I see there is a fused one: https://github.com/pytorch/pytorch/blob/master/tools/autograd/templates/Functions.cpp#L634. The mask tensor should either be Boolean or has compatible type as grad. So no bit-mask either. 

> In the ideal case, it would be to enhance MKL dropout to store mask using bits. But it requires modification of the VSL APIs.

@apeforest , so far there is no dropout functionality in MKL or MKL-DNN. Here we just use VSL to generate random values. So even we can generate bit-mask, it will increase additional computation for `mask_val * 1.0 / pkeep` which can be reused from forward path.

apeforest:
> > Does the slow down come from more computation in the new algorithm or the sub-optimal implementation?
> 
> The new implementation increases both memory load and additional bit-wise operations. So performance slow down is expected.

The memory load is actually reduced even in the case of MKL, right?

TaoLv:
> Why does it increase memory load?

If there are N elements, per the Bernoulli distribution generation in VSL, we still need to allocation memory and write `N*4` bytes to it. To generate bit mask, we need load the `N*4` bytes back and write `N/8` bytes with bits.

apeforest:
> > Why does it increase memory load?
> 
> If there are N elements, per the Bernoulli distribution generation in VSL, we still need to allocate memory and write `N*4` bytes to it. To generate bit mask, we need load the `N*4` bytes back and write `N/8` bytes with bits.

The memory for bit-mask is not extra memory. `N*sizeof(DType)` was used in the master branch: https://github.com/apache/incubator-mxnet/blob/master/src/operator/nn/dropout.cc#L124

So for the MKL dropout case, 
master branch uses memory `N*4` + `N*sizeof(DType)` vs. this PR `N*4` + `N/8`. This memory reduction is verified through the MXNet profiler results reported in the PR description section.

eric-haibin-lin:
@PatricZhao @TaoLv what do you suggest as the resolution? If CPU performance is a concern, shall we add env_var to control the behavior? Do you agree in the long term we want to push for dropout API in MKLDNN with 1-bit mask?

apeforest:
Given your concern about the performance degradation in the case of MKL dropout, I have disabled this feature when MKL dropout is used. Please review the PR again and let me know if you think this is good to go. Thanks!

apeforest:
Hi @TaoLv I reverted my last commit "
It makes the code too bristle and also involves very complicate logic to check memory allocation at runtime. Here are the main reasons:

(1) MKL dropout support is currently not complete. It does not work if the input data type is smaller than int32 and it does not support broadcast option (when the option axes is specified). This limitation enforces a check at runtime which is not possible in the InferShape function

e.g. In this function, I will need to check if the dtype is greater than int32 in order to use a different shape for MKL Dropout. 
https://github.com/apache/incubator-mxnet/pull/16735/files#diff-74c4dc433970c5df31a5e2c4b57c8d71R127

(2) Having different Dropout engine at runtime (based on data type and ) may cause inconsistency in the mixed precision case. Introducing another difference in mask memory allocation complicates this even further.

I think we should focus on enhancing MKL Dropout so that it (1) supports all the different cases as non MKL dropout (2) supports bit-mask.

Please let me know what you think. Thanks!

Lin

sxjscience:
Is there anyone that can take a look at this PR ?

PR #16746, opened 2019-11-07
[docs] Fix runtime feature detection documentation

larroy:
## Description ##

as title. 

Documentation was not appearing in the index nor sidebar.
Adds usage example to mxnet.runtime

@ThomasDelteil @aaronmarkham


larroy:
@aaronmarkham @zachgk @lanking520  can we merge this?

larroy:
@ChaiBapchya does CPP affect the library build?

aaronmarkham:
> @ChaiBapchya does CPP affect the library build?

You have to build from source with that flag in order to use CPP. You can't just take a lib you built without the flag and setup the binding (like you do with the other lang bindings)... if that's what you're asking... At least, this is my understanding.

larroy:
That's a good call for contributions in that case.

PR #16750, opened 2019-11-07
[Numpy] add custom op sort

Alicia1529:
## Description ##
add custom NumPy operator sort


python/mxnet/ndarray/numpy/_op.py
```diff
@@ -344,6 +344,133 @@ def take(a, indices, axis=None, mode='raise', out=None):
...
+        [1., 3.]])
+    >>> np.sort(a, axis=None)     # sort the flattened array
+    array([[1., 1., 3., 4.]])
+    >>> np.sort(a, axis=0)        # sort along the first axis
+    array([[1., 1.],
+        [3., 4.]])
+    """
+    if order is not None:
+        raise NotImplementedError("Not support at the moment")
+    if axis is not None and (axis - 1) > len(a.shape):
```
reminisce: You may also need to put this check in `infer_shape` function so that symbol api can catch it.

python/mxnet/numpy_op_fallback.py
```diff
@@ -49,6 +49,48 @@ def _register_helper(prop_cls):
...
+        raise NotImplementedError('Operator sort does not support gradient computation')
+
+
+@register('sort_fallback')
+class SortProp(operator.CustomOpProp):
+    """Fallback sort operator properties."""
+    def __init__(self, axis, kind):
+        super(SortProp, self).__init__(need_top_grad=True)
+        self._axis = ast.literal_eval(axis)
+        self._kind = kind if isinstance(kind, str) else ast.literal_eval(kind)
```
reminisce: `kind` is always a string here, so `isinstance(kind, str)` always returns true. According to documentation, which says that kind is one of `{None, ‘quicksort’, ‘mergesort’, ‘heapsort’, ‘stable’}`, I think you can do
```python
self._kind = None if kind == 'None' else kind
```

python/mxnet/numpy/multiarray.py
```diff
@@ -2276,6 +2276,127 @@ def take(a, indices, axis=None, mode='raise', out=None):
...
+    `CPython listsort.txt <https://github.com/python/cpython/blob/3.7/Objects/listsort.txt>`_.
+    'mergesort' and 'stable' are mapped to radix sort for integer data types. Radix sort is an
+    O(n) sort instead of O(n log n).
+
+    Examples
+    --------
+    >>> a = np.array([[1,4],[3,1]])
+    >>> np.sort(a)                # sort along the last axis
+    array([[1., 4.],
+        [1., 3.]])
```
reminisce: Align this.

python/mxnet/ndarray/numpy/_op.py
```diff
@@ -344,6 +344,133 @@ def take(a, indices, axis=None, mode='raise', out=None):
...
+          [1., 3.]])
+    >>> np.sort(a, axis=None)     # sort the flattened array
+    array([[1., 1., 3., 4.]])
+    >>> np.sort(a, axis=0)        # sort along the first axis
+    array([[1., 1.],
+          [3., 4.]])
+    """
+    if order is not None:
+        raise NotImplementedError("Not support at the moment")
+    if axis is not None and ((axis + 1) > len(a.shape) or abs(axis) > len(a.shape)):
```
hgt312:
```suggestion
    if axis is not None and (axis > a.ndim - 1 or axis < -a.ndim) :
```

python/mxnet/numpy_op_fallback.py
```diff
@@ -49,6 +49,48 @@ def _register_helper(prop_cls):
...
+
+    def list_arguments(self):
+        return ['a']
+
+    def infer_shape(self, in_shape):
+        if self._axis is None and len(in_shape[0]) == 0:
+            out_shape = (1,)
+        elif self._axis is None:
+            out_shape = (np.prod(in_shape[0]),)
+        elif (self._axis + 1) > len(in_shape[0]) or abs(self._axis) > len(in_shape[0]):
```
hgt312: same here

sxjscience:
@reminisce @Alicia1529 @hgt312 We have sort in ndarray: https://github.com/apache/incubator-mxnet/blob/master/src/operator/tensor/ordering_op.cc#L109-L160. This operator is quite important and sorting with GPU will improve the performance. I think we should not use the CustomOp to do that.

python/mxnet/numpy_op_fallback.py
```diff
@@ -49,6 +49,49 @@ def _register_helper(prop_cls):
...
+        super(Sort, self).__init__()
+        self._axis = axis
+        self._kind = kind
+
+    def forward(self, is_train, req, in_data, out_data, aux):
+        out = np.sort(in_data[0].asnumpy(), self._axis, self._kind)
+        self.assign(out_data[0], req[0], _mx_np.array(out, dtype=out.dtype, ctx=out_data[0].ctx))
+
+    def backward(self, req, out_grad, in_data, out_data, in_grad, aux):
+        raise NotImplementedError('Operator sort does not support gradient computation')
```
sxjscience: Sort can be made to be differentiable. The gradient is to  pass back the gradient to the stored indices.

wkcn: I think it is suitable to use numpy.argsort to get the index in forward.

PR #16760, opened 2019-11-08
[Estimator] Improve usability and fix logging inconsistencies

leezu:
## Description ##
gluon.contrib.estimator uses a Logger shared between all event handlers that is
obtained via logging.getLogger(gluon.contrib.estimator.event_handlers). This
logger used to be configured when creating a
gluon.contrib.estimator.LoggingHandler. This configuration was inconsistent,
because each time a LoggingHandler was created, the global Logger was modified.

Instead, this commit clearly separates the LoggingHandler (responsible for
logging to the Logger during estimator.fit) from the configuration of the
Logger (which users can do via the introduced getLogger API).

Further, there was previously no way for user-defined EventHandlers to log to
the correct Logger instance.

We also disable log handler propagation by default and change the default log
stream for gluon estimator to sys.stdout.

### Essentials ###

### Changes ###
- [X] Introduce `gluon.contrib.estimator.getLogger` API for users to configure and obtain the logger class used by gluon estimator. This supersedes configuring the logger by adding constructor arguments to `gluon.contrib.estimator.LoggingHandler`, as the logger is a global object and it does not make sense to configure it as part of a class constructor (of which we can expect some users to create multiple objects during a sessione.)

## Comments ##
- This change is backward incompatible, but is important to make. It only changes experimental APIs. The reason is that the previous API design is inconsistent and not user friendly. It is inconsistent, because a global object is modified whenever a `LoggingHandler` object is created; this leads to buggy behaviour if the user creates multiple `LoggingHandler`s, as all of them will share the same underlying Python Logger. Instead, we must make it explictit that the same Logger is used.
- This change improves usability, by switching the default log stream from `sys.stderr` to `sys.stdout`. The following screenshot highlights the need for this. Users can change this default by modifying the logger returned by the `getLogger` API, if needed.

image

Compare this to 

image


leezu:
Superseded by https://github.com/apache/incubator-mxnet/pull/16810

Issue #16646, opened 2019-10-27
[Bug][Numpy] zeros_like, ones_like does not respect dtype flag


sxjscience:
```python
import mxnet as mx
mx.npx.set_np()

a = mx.np.ones((10,10), dtype=mx.np.float32)
b = mx.np.zeros_like(a, dtype=mx.np.int32)
print(b.dtype)
```
Returns
```
float32
```

xidulu:
I take a brief look at the `infertype` function for `zeros_like`:
https://github.com/apache/incubator-mxnet/blob/master/src/operator/numpy/np_init_op.cc#L103
The parameter type seems to be ignored.
The following experiment verified my conjecture:
```
In [1]: from mxnet import np

In [2]: a = np.ones(10)

In [3]: b = np.zeros_like(a, dtype="Life is short, use Pytorch")

In [4]: b
Out[4]: array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])

In [5]: b.dtype
Out[5]: dtype('float32')
```

Alicia1529:
This bug has been fixed in pr #16804

PR #16778, opened 2019-11-11
custom op full_like

Alicia1529:
## Description ##
Add numpy operator full_like

### DIFF
--- a/python/mxnet/numpy_op_fallback.py
+++ b/python/mxnet/numpy_op_fallback.py
@@ -50,6 +50,61 @@ def register(op_name, imperative=True, symbolic=True):
 
 
 @use_np  # enforce np shape and array semantics for all the methods in this class
+class FullLike(operator.CustomOp):
+    """Fallback to NumPy full_like operator."""
+    def __init__(self, fill_value, dtype):
+        super(FullLike, self).__init__()
+        self._fill_value = fill_value
+        self._dtype = dtype
+
+    def forward(self, is_train, req, in_data, out_data, aux):
+        out = np.full_like(in_data[0].asnumpy(), self._fill_value, dtype=self._dtype)
+        self.assign(out_data[0], req[0], _mx_np.array(out, dtype=out.dtype, ctx=out_data[0].ctx))
+
+    def backward(self, req, out_grad, in_data, out_data, in_grad, aux):
+        raise NotImplementedError('Operator full_like does not support gradient computation')
+
+
+@register('full_like_fallback')
+class FullLikeProp(operator.CustomOpProp):
+    """Fallback full_like operator properties."""
+    def __init__(self, fill_value, dtype):
+        super(FullLikeProp, self).__init__(need_top_grad=True)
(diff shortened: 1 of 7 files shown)


python/mxnet/numpy_op_fallback.py
```diff
@@ -49,6 +49,73 @@ def _register_helper(prop_cls):
...
+        self._fill_value = fill_value
+        self._dtype = dtype
+        self._order = order 
+        self._subok = subok
+        self._shape = shape
+
+    def forward(self, is_train, req, in_data, out_data, aux):
+        np_version = np.version.version
+        if np_version == '1.17.0':
+            out = np.full_like(in_data[0].asnumpy(), self._fill_value, dtype=self._dtype, order=self._order,
```
wkcn: It will be faster to use `mx.nd.full(shape, fill_value)`

hgt312:
@mxnet-label-bot add [Numpy]

tests/python/unittest/test_numpy_op.py
```diff
@@ -4463,6 +4464,52 @@ def hybrid_forward(self, F, a):
...
+    class TestFullLike(HybridBlock):
+        def __init__(self, fill_value, dtype):
+            super(TestFullLike, self).__init__()
+            self._fill_value = fill_value
+            self._dtype = dtype
+
+        def hybrid_forward(self, F, x, *args, **kwargs):
+            return F.np.full_like(x, self._fill_value, self._dtype)
+
+    if StrictVersion(platform.python_version()) < StrictVersion('3.0.0'):
```
reminisce: You can decorate the test function with this condition as the following.
```python
@unittest.skipIf(StrictVersion(platform.python_version()) < StrictVersion('3.0.0'), "reason to skip")
def test_np_full_like():
    ...
```

Alicia1529: thanks!

sxjscience:
I suggest not use the customop for full_like, which is very similar to ones_like

sxjscience:
@Alicia1529, You can refer to the implementation of ones_like and zeros_like: https://github.com/apache/incubator-mxnet/blob/61c8bafdcfee129e4f7a491438a2402e6762ddd9/src/operator/numpy/np_init_op.cc#L100-L134

Alicia1529:
> @Alicia1529, You can refer to the implementation of ones_like and zeros_like:
> 
> https://github.com/apache/incubator-mxnet/blob/61c8bafdcfee129e4f7a491438a2402e6762ddd9/src/operator/numpy/np_init_op.cc#L100-L134

Thanks for the suggestions!

In pr #16804 a c++ version of full_like is implemented. Zeros_like and ones_like are also rewritten by calling full_like with different fill_value in the front-end to fix the type inference bug in issue #16646 

But I am still fixing some problems in python2 test, so it has not been merged. This pr is older and should be closed.

sxjscience:
@Alicia1529 Is it safe to close this issue?

Alicia1529:
@sxjscience  Yeah, thanks

PR #16810, opened 2019-11-14
[Gluon] Improve estimator usability and fix logging logic

leezu:
## Description ##
gluon.contrib.estimator used a global Logger obtained via
`logging.getLogger('gluon.contrib.estimator.event_handlers')`. This logger used
to be configured every time a gluon.contrib.estimator.LoggingHandler was
created, which is a bug. We can't modify a global Logger instance whenever the
user creates an Estimator and a LoggingHandler.

Instead, this commit separates the LoggingHandler (responsible for logging
metadata during estimator.fit) from the configuration of the Logger.

We expose the Logger as attribute of the Estimator, and configure it to output
to stdout by default. Instructions are given how users can configure the
Estimator.logger to log to a file instead.

### Essentials ###

## Comments ##
- This change is backward incompatible, but is important to make. It only changes experimental APIs. The reason is that the previous API design is inconsistent and not user friendly. It is inconsistent, because a global object is modified whenever a `LoggingHandler` object is created; this leads to buggy behaviour if the user creates multiple `LoggingHandler`s, as all of them will share the same underlying Python Logger. Instead, we should use a separate logger for each estimator.
- This change improves usability, by switching the default log stream from `sys.stderr` to `sys.stdout`. The following screenshot highlights the need for this.

image

Compare this to 

image


leezu:
@ptrendx can we add this change to the 1.6 branch? Should I open a PR or will you merge it similar to https://github.com/apache/incubator-mxnet/pull/16832 ?

ptrendx:
@leezu yes, go ahead and make a PR to 1.6.x branch

PR #16826, opened 2019-11-15
Add missing imports in python/mxnet/contrib/__init__.py 

leezu:
## Description ##
Add missing imports in python/mxnet/contrib/__init__.py 

## Comments ##
Without this change, users need to `import mxnet; import mxnet.contrib.amp; print(mxnet.contrib.amp)` before they can use amp.
With this change, users can `import mxnet; print(mxnet.contrib.amp)`

### DIFF
--- a/python/mxnet/contrib/__init__.py
+++ b/python/mxnet/contrib/__init__.py
@@ -30,6 +30,7 @@ from . import tensorboard
 from . import amp
 from . import text
 from . import onnx
+from . import svrg_optimization
 from . import io
 from . import quantization
 from . import quantization as quant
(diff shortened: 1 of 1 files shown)


leezu:
This is blocked by some dependencies between contrib module and ndarray module, as well as hacks in the ndarray module that require contrib to be imported before ndarray.

Should be revisited after dropping such hacks.

```
~/src/mxnet-master/python/mxnet/base.py in _init_op_module(root_namespace, module_name, make_op_func)
    611     # use mx.nd.contrib or mx.sym.contrib from now on
    612     contrib_module_name_old = "%s.contrib.%s" % (root_namespace, module_name)
--> 613     contrib_module_old = sys.modules[contrib_module_name_old]
    614     submodule_dict = {}
    615     for op_name_prefix in _OP_NAME_PREFIX_LIST:
```

PR #16829, opened 2019-11-15
[Numpy][Operator] 'where' Implementation in MXNet

hgt312:
https://docs.scipy.org/doc/numpy/reference/generated/numpy.where.html?highlight=where#numpy.where


hgt312:
@mxnet-label-bot add [Numpy]

xidulu:
Two question:
1. Numpy supports the following usage:
```
>>> import numpy as np
>>> a = np.random.uniform(size=(4,1))
>>> np.where(a > 0.5)
(array([0, 1, 2]), array([0, 0, 0]))
```
in short: x and y are optional ( either *both* or *neither* of x and y should be given)

2. How did you handle the scalar case mentioned in the docstring?:
https://github.com/apache/incubator-mxnet/pull/16829/files#diff-2867f3bd5e893c32c128044e32d821d6R7353-R7359

hgt312:
@xidulu 
1. According to NumPy's doc, use `np.nonzero`, we have it.
> When only condition is provided, this function is a shorthand for np.asarray(condition).nonzero(). Using nonzero directly should be preferred, as it behaves correctly for subclasses. The rest of this documentation covers only the case where all three arguments are provided.
2. Does not support scalar now, just use `np.array` to wrap it.

src/operator/numpy/np_where_op-inl.h
```diff
@@ -0,0 +1,266 @@
...
+}
+
+inline bool NumpyWhereOpType(const nnvm::NodeAttrs& attrs,
+                             std::vector<int>* in_attrs,
+                             std::vector<int>* out_attrs) {
+  CHECK_EQ(in_attrs->size(), 3U)
+    << "where operator takes 3 arguments (" << in_attrs->size() << " given)";
+  CHECK_EQ(out_attrs->size(), 1U);
+  CHECK_EQ(in_attrs->at(1), in_attrs->at(2));
+  TYPE_ASSIGN_CHECK(*out_attrs, 0, in_attrs->at(1));
```
anirudh2290: you should also add backward inference. inferring input types from output. One way to write this easily, would be to use ElemwiseType or ElemwiseAttr.

python/mxnet/ndarray/numpy/_op.py
```diff
@@ -5308,3 +5308,67 @@ def nan_to_num(x, copy=True, nan=0.0, posinf=None, neginf=None, **kwargs):
         return _npi.nan_to_num(x, copy=copy, nan=nan, posinf=posinf, neginf=neginf, out=None)
     else:
         raise TypeError('type {} not supported'.format(str(type(x))))
+
+
+@set_module('mxnet.ndarray.numpy')
+def where(condition, x, y):
```
reminisce: We need to support the case where there is only one input. You can do this for the imperative mode only by falling back to `condition.nonzero()`.

src/operator/numpy/np_where_op-inl.h
```diff
@@ -0,0 +1,266 @@
...
+    auto cidx = static_cast<index_t>(mxnet_op::dot(coord, cstride));
+    if (is_left) {
+      KERNEL_ASSIGN(grad[base], req, datac[cidx] != CType(0) ? datao[base] : DType(0));
+    } else {
+      KERNEL_ASSIGN(grad[base], req, datac[cidx] == CType(0) ? datao[base] : DType(0));
+    }
+  }
+};
+
+inline bool NumpyWhereOpShape(const nnvm::NodeAttrs& attrs,
```
reminisce: Put shape/type inference functions in .cc if they are not called from other files.

src/operator/numpy/np_where_op-inl.h
```diff
@@ -0,0 +1,266 @@
...
+#include "../../common/utils.h"
+#include "../mxnet_op.h"
+#include "../mshadow_op.h"
+#include "../operator_common.h"
+#include "np_broadcast_reduce_op.h"
+
+namespace mxnet {
+namespace op {
+
+#define NUMPY_WHERE_MAX_DIM 5
```
reminisce: Use `constexpr` as `#define` is not scope-controlled.

reminisce: Change the name to `np_where_max_ndim`.

src/operator/numpy/np_where_op-inl.h
hgt312: solved

src/operator/numpy/np_where_op-inl.h
```diff
@@ -0,0 +1,218 @@
...
+    expanded_lshape[j + ndim_delta] = (outputs[0].shape_)[j];
+  }
+  // get expanded rshape
+  TShape expanded_rshape(broadcast::MAX_DIM, 1);
+  ndim_delta = expanded_rshape.ndim() - outputs[1].shape_.ndim();
+  for (int j = 0; j < outputs[1].shape_.ndim(); ++j) {
+    expanded_rshape[j + ndim_delta] = (outputs[1].shape_)[j];
+  }
+
+  MSHADOW_TYPE_SWITCH_WITH_BOOL(inputs[0].type_flag_, DType, {
```
haojin2: Consider do:
```
const TBlob& cond = inputs[0];
const TBlob& x = inputs[1];
const TBlob& y = inputs[2];
const TBlob& out = outputs[0];
```
and substitute all the `inputs[i]` `outputs[j]` to improve readability.

src/operator/numpy/np_where_op-inl.h
```diff
@@ -0,0 +1,218 @@
...
+    }
+    in_strides[i] = mxnet_op::calc_stride(expanded_ishape.get<broadcast::MAX_DIM>());
+  }
+  TShape expanded_oshape(broadcast::MAX_DIM, 1);
+  const int ndim_delta = expanded_oshape.ndim() - outputs[0].shape_.ndim();
+  for (int j = 0; j < outputs[0].shape_.ndim(); ++j) {
+    expanded_oshape[j + ndim_delta] = (outputs[0].shape_)[j];
+  }
+  Shape<broadcast::MAX_DIM> oshape = expanded_oshape.get<broadcast::MAX_DIM>();
+  MSHADOW_TYPE_SWITCH_WITH_BOOL(outputs[0].type_flag_, DType, {
```
haojin2: Consider do:
```
const TBlob& cond = inputs[0];
const TBlob& x = inputs[1];
const TBlob& y = inputs[2];
const TBlob& out = outputs[0];
```
and substitute all the `inputs[i]` `outputs[j]` to improve readability.

PR #16835, opened 2019-11-16
Fix test_gluon.py:test_sync_batchnorm when number of GPUS > 4

larroy:
Port of https://github.com/apache/incubator-mxnet/pull/16834


larroy:
@ptrendx

Issue #11346, opened 2018-06-20
Logic operator returns NDArrays of float, not int / boolean


junrushao1994:
## Description
Logical operators for NDArrays returns a float NDArray, which is counterintuitive and inconsistent of the description of the documentation. For example, [doc here](https://mxnet.incubator.apache.org/api/python/ndarray/ndarray.html#mxnet.ndarray.not_equal).

## Steps to reproduce
It can be produced by running the code in the doc. It seems that the dtype of result NDArray is the same as the dtype of inputs.

For example, when x and y are of dtype int32, (x == y) returns int32.

```python
In [1]: import mxnet as mx
In [2]: x = mx.nd.ones((2,3)).astype("int32")
In [3]: y = mx.nd.arange(2).reshape((2,1)).astype("int32")
In [4]: (x == y).dtype
Out[4]: numpy.int32
```

However, when x and y are of dtype float32, (x == y) returns float32

```python
In [1]: import mxnet as mx
In [2]: x = mx.nd.ones((2,3)).astype("float32")
In [3]: y = mx.nd.arange(2).reshape((2,1)).astype("float32")
In [4]: (x == y).dtype
Out[4]: numpy.float32
```

haojin2:
MXNet does not have bool type support right now.

junrushao1994:
@haojin2 I mean it could be more reasonable if the returned type is `numpy.int`

haojin2:
We may not want to make `numpy.int8` as the default here, if you really need the dtype to be `numpy.int8` you can add a `cast` in your model or script.

junrushao1994:
If it is by design, I guess it will be great to mention in the documentation. Closing this issue for now.

haojin2:
@junrushao1994 I think that's a fair point, we can definitely add that soon.

PR #16836, opened 2019-11-16
Fix InferAttr/InferShapeAttr not calling inference for all nodes in a graph

ptrendx:
## Description ##
Currently type/shape inference of the operator is only called if the values of input and output attributes are not currently known. This is a bug as it may lead to some operators never seeing the values of those inputs/outputs attributes during InferAttr pass and so never validating them.

As an example, this code snippet runs without throwing any exception in current MXNet:
```
v = mx.sym.Variable('V')
s = mx.sym.transpose(v)
x = mx.sym.Variable('x')
s2 = x + v
s3 = s + s2
e = s3.simple_bind(ctx=mx.cpu(), x=(2,3), grad_req='null')
```
and it errors out only when calling `e.forward` (in 1.5.1, in 1.6.0 new 2D transpose implementation removed the check for shape inside the forward function of the operator so it does not error out even then). This is because the `transpose` node is early in the topologically sorted graph and ran its shape inference earlier (on unknown input and output shapes) than the `_Plus` operators, which then set all of the shapes by themselves, which prevented `transpose` from running its shape inference again to spot that they are not suitable for this operator. 

In this PR, I track which operators are actually finished inferring attributes (including validating the attributes they got if all of the inputs/outputs are already known at the point of running their attribute inference).

### Essentials ###

## Comments ##
- Currently I check that the operator finished inferring attributes if all input and output attributes are set after the call to operator's InferShape/InferType. This should not be necessary as those functions return the value that should be equivalent. However, since this value was never actually checked, there are a lot of operators that lie and return `true` even if not everything was inferred correctly. This should be tackled in a separate PR.

### DIFF
--- a/src/executor/infer_graph_attr_pass.cc
+++ b/src/executor/infer_graph_attr_pass.cc
@@ -401,7 +401,8 @@ nnvm::Graph InferAttr(nnvm::Graph &&ret,
       if (!is_fusion_helper.get(fwd_ptr->op(), false)) {
         GetAttrFromForwardNode(nid, idx, &rshape, &inference_finished, fis_none);
       } else {
-        GetAttrFromFusedNode<FAccessSubgraphType>(nid, idx, &rshape, &inference_finished, fis_none, infer_fusion_name);
+        GetAttrFromFusedNode<FAccessSubgraphType>(nid, idx, &rshape, &inference_finished,
+                                                  fis_none, infer_fusion_name);
       }
     } else {
       DispatchMode* dispatch_mode = nullptr;
--- a/src/executor/infer_graph_attr_pass.cc
+++ b/src/executor/infer_graph_attr_pass.cc
@@ -67,6 +67,7 @@ template<typename AttrType, typename IsNone>
 inline void GetAttrFromForwardNode(const uint32_t nid,
                                    const nnvm::IndexedGraph &idx,
                                    std::vector<AttrType>* rshape_ptr,
+                                   std::vector<bool>* inference_finished,
                                    IsNone fis_none) {
   std::vector<AttrType>& rshape = *rshape_ptr;
   const nnvm::IndexedGraph::Node& inode = idx[nid];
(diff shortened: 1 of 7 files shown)


ptrendx:
@junrushao1994 The failure in `test_contrib_control_flow.test_while_loop_simple_forward` test in this PR is a legitimate bug in `WhileLoopType` function (https://github.com/apache/incubator-mxnet/blob/master/src/operator/control_flow.cc#L764), where you set the output type of condition to `float`. I do not really understand how that function is supposed to work (e.g. you set the `cond_out_type` but then do not use it?), could you take a look at it?

ptrendx:
@reminisce The failure in `test_numpy_op.test_np_transpose` in this PR is a legitimate bug, where `Transpose` op's `TransposeShape` function used to infer shape does not support scalar inputs (https://github.com/apache/incubator-mxnet/blob/master/src/operator/tensor/matrix_op-inl.h#L423). Please help in fixing this (or disabling testing of scalar inputs if `Transpose` not supporting scalars is expected).

ptrendx:
@samskalicky @apeforest FYI, see last 2 comments.

ptrendx:
@reminisce Actually, the required change to `Transpose` seems simple, I will do it in this PR.

junrushao1994:
At the time I wrote the control flow code, it is still infeasible to have int64 output for many operators, like `==` or `<` - this is why this line is hard-coded as float32...

ptrendx:
Not sure I understand @junrushao1994 - I believe the problem here is that you are forcing the float32 output type here even if the operator could output something else (and in the case of the test that fails in this PR, previously the operator did not have a chance to verify that it is correct). Why not leave the output type of the condition part as `-1` so that it can be inferred?

junrushao1994:
@ptrendx At the time that I was writing this piece of code, operators do return float32, see [this](https://github.com/apache/incubator-mxnet/issues/11346).

junrushao1994:
@ptrendx I think leaving it -1 is a nice proposal. Could you fix this? Thanks!

Issue #16765, opened 2019-11-08
FusedOp Failing Static Linked Build


zachgk:
## Description
The build is currently failing for the statically linked build that is used for Scala Maven Publishing. This is blocking the currently nightly snapshot and must also be fixed before building the release jars as well.

```
build/src/operator/fusion/fused_op_gpu.o: In function `void mxnet::FusedOp::Forward<mshadow::gpu>(nnvm::NodeAttrs const&, mxnet::OpContext const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&, std::vector<mxnet::OpReqType, std::allocator<mxnet::OpReqType> > const&, std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob> > const&)':

tmpxft_00008748_00000000-5_fused_op.compute_70.cudafe1.cpp:(.text._ZN5mxnet7FusedOp7ForwardIN7mshadow3gpuEEEvRKN4nnvm9NodeAttrsERKNS_9OpContextERKSt6vectorINS_5TBlobESaISC_EERKSB_INS_9OpReqTypeESaISH_EESG_+0x1287): undefined reference to `cuLaunchKernel'

tmpxft_00008748_00000000-5_fused_op.compute_70.cudafe1.cpp:(.text._ZN5mxnet7FusedOp7ForwardIN7mshadow3gpuEEEvRKN4nnvm9NodeAttrsERKNS_9OpContextERKSt6vectorINS_5TBlobESaISC_EERKSB_INS_9OpReqTypeESaISH_EESG_+0x12a6): undefined reference to `cuGetErrorString'

collect2: error: ld returned 1 exit status

make: *** [bin/im2rec] Error 1

make: *** Waiting for unfinished jobs....

2019-11-01 20:27:09,794 - root - INFO - Waiting for status of container 00fc4568b4c9 for 600 s.

2019-11-01 20:27:10,060 - root - INFO - Container exit status: {'StatusCode': 2, 'Error': None}

2019-11-01 20:27:10,061 - root - ERROR - Container exited with an error 😞

2019-11-01 20:27:10,061 - root - INFO - Executed command for reproduction:
```

See full log at http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/restricted-publish-artifacts/detail/master/287/pipeline/
Main Scala nightly pipeline at http://jenkins.mxnet-ci.amazon-ml.com/job/restricted-publish-artifacts/job/master/

[...]

PR #16838, opened 2019-11-16
USE_NVRTC -> ENABLE_CUDA_RTC to fix maven build.  Add compile-guard to fusion.

DickJC123:
This PR fixes the maven build issue https://github.com/apache/incubator-mxnet/issues/16765 reported by @zachgk and first seen with PR https://github.com/apache/incubator-mxnet/pull/15167.  The fix involved-

1. Renaming the build flag used by maven (namely `USE_NVRTC`) to the name used by the pip builds (`ENABLE_CUDA_RTC`).  The Makefile currently looks at ENABLE_CUDA_RTC (not USE_NVRTC) and in response sets the compile flag MXNET_ENABLE_CUDA_RTC=1 and adds `-lcuda -lnvrtc` to the LDFLAGS.  It was the absence of these ld flags when creating im2rec that caused the issue filed.

2. Adding compile guards to the newly-introduced files  ./src/operator/fusion/* in a form similar to those that protect ./src/common/rtc.cc.  Now both cuda runtime-compilation facilities are included into the build (or not) based on MXNET_ENABLE_CUDA_RTC.

### Essentials ###

### Changes ###

## Comments ##


ptrendx:
> LGTM, but I'm thinking about backwards compatibility - should the change of the build flag be of concern?

I would say it is just fixing the bug and bringing the Scala build to parity with other builds (especially since the value of the flag is not changed, so the intent to enable this was there, it's just that it was never actually doing anything).

lanking520:
Honestly speaking, there is no backend build difference between pip build and scala for now. we can also use the pip mk on Scala/Java build without problems. @zachgk let's track on the deprecation on the maven mkl files.

ptrendx:
`[03:39:03] /work/mxnet/src/executor/attach_op_execs_pass.cc:355: Neither FCompute nor FComputeEx registered _FusedOp` - @DickJC123 I believe it is still not guarded properly (seems that the graph pass works but the actual .cu file that contains `FCompute` is not included).

PR #7698, opened 2017-09-01
Second order gradient and Subgraph execution

piiswrong:

PR #16854, opened 2019-11-19
Change Reshape's shape size check to EQ

Kh4L:
## Description ##
Change non-recorded `NDArray::Reshape` to accept only shapes with the same volume, in order to match NumPy behavior.

NumPy behavior:
```
>>> a = np.ones(shape=(2,4))
>>> a
array([[1., 1., 1., 1.],
       [1., 1., 1., 1.]])
>>> a.reshape(1)
ValueError: cannot reshape array of size 8 into shape (1,)
```

MXNet behavior prior to the change:
```
>>> a = mx.nd.ones(shape=(2,4))
>>> a.reshape(1)

[1.]
<NDArray 1 @cpu(0)>
```

MXNet behavior after the change:
```
>>> a = mx.nd.ones(shape=(2,4))
>>> a.reshape(1)
  File "/opt/mxnet/python/mxnet/ndarray/ndarray.py", line 1067, in reshape
    ctypes.byref(handle)))
  File "/opt/mxnet/python/mxnet/base.py", line 252, in check_call
mxnet.base.MXNetError: [17:03:38] src/ndarray/ndarray.cc:239: Check failed: shape_.Size() == shape.Size() (8 vs. 1) : NDArray.Reshape: target shape must have the same size as current shape.
```

### Essentials ###

### Changes ###
- [X] Change Reshape check

### DIFF
--- a/src/ndarray/ndarray.cc
+++ b/src/ndarray/ndarray.cc
@@ -243,8 +243,9 @@ NDArray NDArray::MKLDNNDataReshape(const mxnet::TShape &shape) const {
 
 NDArray NDArray::Reshape(const mxnet::TShape &shape) const {
   CHECK(!is_none()) << "NDArray is not initialized";
-  CHECK_GE(shape_.Size(), shape.Size())
-    << "NDArray.Reshape: target shape size is larger current shape";
+  CHECK_EQ(shape_.Size(), shape.Size())
+    << "NDArray.Reshape: target shape must have the same size as "
+    << "current shape.";
   NDArray ret = this->Detach();
   // If the shape doesn't change, we can just return it now.
   if (ret.shape_ == shape)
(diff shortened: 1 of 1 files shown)


Kh4L:
Looking at the failing tests, it was looks that it was intentional.

> Would be great if previous committers who pushed this code can chime in

Yep I guess it would be the best :)

In any case, I don’t think that it is a good idea to leave this silent "slicing".
I actually had a silent problem when writing some code today, and had to go through the code to realize that was happening.

numpy (and other numpy-like FW) enforces the sizes to match, so most of users would expect it to be the case with MXNet.

ChaiBapchya:
Usability wise makes total sense to keep it compatible with Numpy and hence this fix +1

Git blame on the line shows #7698 last changed it 2yrs back
@piiswrong @eric-haibin-lin  any idea?

Issue #16850, opened 2019-11-19
[Numpy] expand_dims throws delay_alloc error


stu1130:
MXNet version: pip install mxnet-mkl --pre as of today
```
>>> import mxnet
>>> from mxnet import np, npx
>>> npx.set_np()
>>> a = np.array([]).reshape(2, 1, 0)
>>> np.expand_dims(a, 2).wait_to_read()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/leecheng/anaconda3/lib/python3.7/site-packages/mxnet/ndarray/ndarray.py", line 2350, in wait_to_read
    check_call(_LIB.MXNDArrayWaitToRead(self.handle))
  File "/Users/leecheng/anaconda3/lib/python3.7/site-packages/mxnet/base.py", line 255, in check_call
    raise MXNetError(py_str(_LIB.MXGetLastError()))
mxnet.base.MXNetError: [23:11:49] src/ndarray/ndarray.cc:502: Check failed: delay_alloc:
Stack trace:
  [bt] (0) 1   libmxnet.so                         0x000000011c7cc029 mxnet::op::MKLDNNLeakyReluBackward(nnvm::NodeAttrs const&, mxnet::OpContext const&, std::__1::vector<mxnet::NDArray, std::__1::allocator<mxnet::NDArray> > const&, mxnet::OpReqType const&, mxnet::NDArray const&) + 4265
  [bt] (1) 2   libmxnet.so                         0x000000011ea457c4 mxnet::NDArray::Chunk::SetMKLMem(mxnet::TShape const&, int) + 1668
  [bt] (2) 3   libmxnet.so                         0x000000011ea45cc0 mxnet::NDArray::GetMKLDNNData() const + 560
  [bt] (3) 4   libmxnet.so                         0x000000011c80e941 mxnet::op::MKLDNNReshapeFwd::MKLDNNReshapeFwd(mxnet::OpReqType const&, mxnet::NDArray const&, mxnet::NDArray const&) + 225
  [bt] (4) 5   libmxnet.so                         0x000000011c80fe65 mxnet::op::GetReshapeForward(mxnet::OpReqType const&, mxnet::NDArray const&, mxnet::NDArray const&) + 245
[...]

PR #16856, opened 2019-11-19
Fix zero-size problem in expand_dims MKLDNN

reminisce:
Fixed https://github.com/apache/incubator-mxnet/issues/16850. Added test.


reminisce:
@stu1130

Issue #16830, opened 2019-11-15
CI error in unix gpu test_quantization_gpu.test_quantized_conv


leezu:
more details

http://jenkins.mxnet-ci.amazon-ml.com/blue/rest/organizations/jenkins/pipelines/mxnet-validation/pipelines/unix-cpu/branches/PR-16810/runs/6/nodes/300/steps/700/log/?start=0
```
FAIL: test_quantization_mkldnn.test_quantized_conv
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/usr/local/lib/python3.5/dist-packages/nose/util.py", line 620, in newfunc
    return func(*arg, **kw)
  File "/work/mxnet/tests/python/mkl/../unittest/common.py", line 177, in test_new
    orig_test(*args, **kwargs)
  File "/work/mxnet/tests/python/mkl/../quantization/test_quantization.py", line 277, in test_quantized_conv
    check_quantized_conv((3, 4, 28, 28), (3, 3), 128, (1, 1), (1, 1), False, qdtype)
  File "/work/mxnet/tests/python/mkl/../quantization/test_quantization.py", line 273, in check_quantized_conv
    assert cond == 0
AssertionError: 
-------------------- >> begin captured stdout << ---------------------
skipped testing quantized_conv for mkldnn cpu int8 since it is not supported yet
skipped testing quantized_conv for mkldnn cpu int8 since it is not supported yet

--------------------- >> end captured stdout << ----------------------
-------------------- >> begin captured logging << --------------------
common: INFO: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=752964439 to reproduce.
--------------------- >> end captured logging << ---------------------
```

pengzhao-intel:
@xinyu-intel could you help take a look ?

haojin2:
Happening again: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-cpu/detail/PR-16862/7/pipeline/299
```
======================================================================

FAIL: test_quantization_mkldnn.test_quantized_conv

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/usr/local/lib/python3.5/dist-packages/nose/case.py", line 198, in runTest

    self.test(*self.arg)

[...]

PR #16866, opened 2019-11-20
skip quantized conv flaky case

xinyu-intel:
## Description ##

temporary skip this case.
@pengzhao-intel 

#16830

### Essentials ###

### Changes ###

## Comments ##


ChaiBapchya:
Are you trying to use this PR for testing the issue? If that's the case it doesn't make sense coz its a costly experiment in that case.

xinyu-intel:
Some flaky tests always fail in CI but cannot reproduce easily from local even with docker env on EC2. So debug with the CI environment can be more efficiency to find the root cause.

access2rohit:
@xinyu-intel this is not a proper use of public resources. If you have issues with reproducing your issues on local then you should ask on dev@ channel thats a good place to start. Using CI for debugging causes a fresh run of all 11 test suits on your PR each time you push new changes. A single run of all 11 tests costs approximately 200$. Please be mindful of that.

larroy:
@xinyu-intel you can use the ci/build.py tool to easily reproduce what runs on CI. it's dockerized so it's going to produce the same result with very high probability:  https://cwiki.apache.org/confluence/display/MXNET/Reproducing+test+results#Reproducingtestresults-2.1.Build

xinyu-intel:
okay, I'll ask in dev list next time.

xinyu-intel:
@larroy I have already used this script and it is very easy for use. However, in contrast with "very high probability", this case also cannot be reproduced with`ci\docker.py` scripts on c5.18xlarge:)
I'll skip this case to save the cost. But you should fix the `https error` first:(http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-cpu/detail/PR-16866/11/pipeline

pengzhao-intel:
Let's skip this test first :)

TaoLv:
@ChaiBapchya @access2rohit @larroy Really appreciate your suggestions and frugal spirit for MXNet development. I think this PR is not the first time for contributors to use CI for debugging purpose, and most likely it will not be the last one. I just took two simple filters on the PR list:
https://github.com/apache/incubator-mxnet/pulls?utf8=%E2%9C%93&q=is%3Apr+in%3Atitle+%22do+not+merge%22
https://github.com/apache/incubator-mxnet/pulls?utf8=%E2%9C%93&q=is%3Apr+in%3Atitle+%22debug%22
@access2rohit Can you explain your PRs at the top of the filtered list? If you really worry about the usage of CI resources, could you please bring this topic to @dev to forbid this behavior and note it somewhere if you get a consensus there?

I believe @xinyu-intel has tried his best to reproduce this issue on both local machine and also on EC2 instances to better emulate the CI environment. But unfortunately, he cannot reproduce the error and I think it's not the first time that we cannot reproduce an error out of CI.

Another choice is maybe someone can contribute to the CI system so developers can only trigger just one test suits rather than all of the 11 tests suits. It will be very helpful especially to document contributors.

access2rohit:
@TaoLv I was already pointed out by @szha about the consequences of opening such PRs already and I closed those PRs. Also, if I had done that before doesn't mean I cannot suggest other developers not to do the same mistake as I did. Yes, if such a behaviour becomes widespread I will start a thread on @dev list.

TaoLv:
@access2rohit Good to learn that.
@xinyu-intel Please try it again to see if you can reproduce the problem out of CI. If not, seems that we have no other choice but to disable this UT.

wkcn:
Hi @access2rohit and @larroy, I have several suggestions about MXNet CI.
1. Allow trusty contributors or committers to get access to the machines of MXNet CI by ssh
  Contributors may could not reproduce issues on their local machines. They don't have the server, as the same as that on MXNet CI. When some unit-tests fail on CI, they can only read the log, but they couldn't look for the reason in detail. For example, we met the problem about CI driver, like PR #16985. Since of no permission, other contributors couldn't address the issue. Contributos may need to access to the CI machine, in order to solve the problem.

It is better to provide some machines to debug : )

2. Allow users to do single run or stop the unit-test.
Because of network error or precision problem, unit-tests may fail. Contributors need to restart all tests recently. If CI supports single run, the cost of CI will be reduced. If contributors submit a new commit, the old CI test should be stopped. When contributors realize that there is a mistake in this PR, they should have the permission to stop the CI test.
For document update (such as fixing typo), it will be better to build the documents only.

Thanks.

marcoabreu:
#1 is certainly not an option. The CI is a productive system and granting ssh access is a security concern. Amazon security was very explicit when we went through the security audit.

tests/python/quantization/test_quantization.py
```diff
@@ -200,8 +200,9 @@ def check_quantized_conv(data_shape, kernel, num_filter, pad, stride, no_bias, q
         if is_test_for_native_cpu():
             print('skipped testing quantized_conv for native cpu since it is not supported yet')
             return
-        elif qdtype == 'int8' and is_test_for_mkldnn():
-            print('skipped testing quantized_conv for mkldnn cpu int8 since it is not supported yet')
+        elif is_test_for_mkldnn():
+            # (TODO)Xinyu: https://github.com/apache/incubator-mxnet/issues/16830
+            print('skipped testing quantized_conv for mkldnn cpu since it is a flaky case')
```
ChaiBapchya, line 205: @xinyu-intel Any chance we switch this back on?

PR #16871, opened 2019-11-20
Enforces NDArray type in get_symbol

Kh4L:
## Description ##
Giving a Python object different than NDArray to `get_symbol` causes a Segfault, we need to enforce the type.
Add an assert for type checking.

### Essentials ###

### Changes ###
- [x] Add type check

### DIFF
--- a/python/mxnet/autograd.py
+++ b/python/mxnet/autograd.py
@@ -360,6 +360,8 @@ def get_symbol(x):
     Symbol
         The retrieved Symbol.
     """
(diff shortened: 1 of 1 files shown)


ptrendx:
Could you make the error message a little bit more clear (like adding "Incorrect type of the argument" or something like that?)

Kh4L:
> Could you make the error message a little bit more clear (like adding "Incorrect type of the argument" or something like that?)

Done

PR #16878, opened 2019-11-21
add micro to pearsonr

zburning:
## Description ##
add micro to pearson correlation coefficient.

### Essentials ###

### Changes ###

## Comments ##


python/mxnet/metric.py
```diff
@@ -1438,13 +1449,46 @@ class PearsonCorrelation(EvalMetric):
...
+
+    def update_variance(self, new_values, *aggregate):
+        #Welford's online algorithm for variance update
+        (count, mean, m_2) = aggregate
+        count += len(new_values)
+        delta = new_values - mean
+        mean += numpy.sum(delta / count)
+        delta_2 = new_values - mean
+        m_2 += numpy.sum(delta * delta_2)
+        return (count, mean, m_2)
```
sxjscience: There is no need to wrap it inside a tuple, we may just return `count, mean, m_2`.

sxjscience:
Nice add! Could you also add the test here? https://github.com/apache/incubator-mxnet/blob/a8b31a239f5d5ed0ebff0f3be44b5e5534e0b3f5/tests/python/unittest/test_metric.py#L265-L272

Also, I suggest test against the scipy version: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.pearsonr.html

zburning:
Actually, I also test the run time performance locally. But the current test_metric_perf.py doesn't test micro performance. Do you think it's necessary to add it? @sxjscience

python/mxnet/metric.py
```diff
@@ -1438,13 +1449,46 @@ class PearsonCorrelation(EvalMetric):
...
-                 output_names=None, label_names=None):
+                 output_names=None, label_names=None, average='macro'):
+        self.average = average
         super(PearsonCorrelation, self).__init__(
             name, output_names=output_names, label_names=label_names,
             has_global_stats=True)
+        if self.average == 'micro':
+            self.reset_micro()
+
+    def reset_micro(self):
```
leezu: Let's make these attributes below private.
For example `self._sse_p` instead of `self.sse_p` etc.

python/mxnet/metric.py
```diff
@@ -1457,16 +1501,37 @@ def update(self, labels, preds):
...
+    def get(self):
+        if self.num_inst == 0:
+            return (self.name, float('nan'))
+        if self.average == 'macro':
+            return (self.name, self.sum_metric / self.num_inst)
+        else:
+            n = self.label_nums
+            pearsonr = self.conv / ((n-1) * numpy.sqrt(self.sse_p / (n - 1)) * numpy.sqrt(self.sse_l / (n - 1)))
+            return (self.name, pearsonr)
+
```
leezu: Seems there are too many newlines here

leezu:
@ptrendx this change may be good to include to 1.6. Do you have any concerns?

sxjscience:
@leezu For this one, we may not included it in R1.6.0 since it adds new functionality.

leezu:
Current metrics is broken as it is inconsistent. Thus this functionality is a bugfix and should be included. There's no reason to leave it out.

sxjscience:
My concern is that it adds the average flag and is not a mere bug fix. @ptrendx will decide whether it is suitable to be included or not.

leezu:
The current pearsonr metric is broken, because we cannot actually calculate the pearsonr but only an average of batch-wise pearsonrs. We should not ship broken features without need. That would be masochistic, deriving pleasure from our user's pain inflicted by the broken feature.
As this PR has very limited scope there should be no concern to include it from my perspective. But of course @ptrendx will take the final decision.

python/mxnet/metric.py
```diff
@@ -590,7 +592,8 @@ def update(self, labels, preds):
 
 
 class _BinaryClassificationMetrics(object):
-    """Private container class for classification metric statistics. True/false positive and
+    """
+    Private container class for classification metric statistics. True/false positive and
```
sxjscience: Fix the docstyle, the original version is correct.

python/mxnet/metric.py
```diff
@@ -686,7 +689,8 @@ def global_fscore(self):
             return 0.
 
     def matthewscc(self, use_global=False):
-        """Calculate the Matthew's Correlation Coefficent"""
+        """Calculate the Matthew's Correlation Coefficent
+        """
```
sxjscience: Move to the previous line.

sxjscience: I'll merge once this problem is addressed.

python/mxnet/metric.py
```diff
@@ -610,8 +609,7 @@ def __init__(self):
         self.global_true_negatives = 0
 
     def update_binary_stats(self, label, pred):
-        """
-        Update various binary classification counts for a single (label, pred)
+        """Update various binary classification counts for a single (label, pred)
```
leezu: Should all be on a single line https://www.python.org/dev/peps/pep-0257/

python/mxnet/metric.py
```diff
@@ -592,8 +592,7 @@ def update(self, labels, preds):
 
 
 class _BinaryClassificationMetrics(object):
-    """
-    Private container class for classification metric statistics. True/false positive and
+    """Private container class for classification metric statistics. True/false positive and
```
leezu: Add two newlines prior to "True/false positive and" https://www.python.org/dev/peps/pep-0257/

python/mxnet/metric.py
```diff
@@ -153,7 +153,8 @@ def reset(self):
         self.global_sum_metric = 0.0
 
     def reset_local(self):
-        """Resets the local portion of the internal evaluation results to initial state."""
+        """Resets the local portion of the internal evaluation results
```
leezu: Single line https://www.python.org/dev/peps/pep-0257/

python/mxnet/metric.py
```diff
@@ -371,7 +372,8 @@ def reset(self):
             pass
 
     def reset_local(self):
-        """Resets the local portion of the internal evaluation results to initial state."""
```
leezu: Single line https://www.python.org/dev/peps/pep-0257/

python/mxnet/metric.py
```diff
@@ -1597,7 +1654,8 @@ def reset(self):
         self.reset_local()
 
     def reset_local(self):
-        """Resets the local portion of the internal evaluation results to initial state."""
```
leezu: Single line https://www.python.org/dev/peps/pep-0257/

PR #15600, opened 2019-07-19
import mshadow source

szha:
## Description ##
import mshadow source. https://github.com/dmlc/mshadow/issues/373. https://lists.apache.org/thread.html/c1ba34330b0eb52ef3a3a30da6d34964a35a01c320e93067e94ed306@%3Cdev.mxnet.apache.org%3E

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] import mshadow source
- [ ] update directory structure and build system

PR #16881, opened 2019-11-21
Add TypeFlag=>string macro

Kh4L:
## Description ##
Add a macro mapping mshadow type_flag to strings, to improve debuggability. 

### Essentials ###

### DIFF
--- a/include/mxnet/base.h
+++ b/include/mxnet/base.h
@@ -85,21 +85,6 @@
  */
 #define PROFILER_MESSAGE_FUNCNAME (__FUNCTION__)
 
-/*!
- * \brief get string from a type_flag (int)
- */
-#define MSHADOW_TYPE_TO_STRING(type_flag)                                                \
-((type_flag >= 0 && static_cast<size_t>(type_flag) < ::mshadow::TypeFlagStrings.size())  \
-  ? ::mshadow::TypeFlagStrings[type_flag]                                                \
-  : "Unknown Type")
-
-namespace mshadow {
-/*
- * \brief array mapping mshadow types defined in dmlc/base.h to strings
-*/
-extern const std::array<std::string, 8> TypeFlagStrings;
-}  // namespace mshadow
-
 /*! \brief namespace of mxnet */
 namespace mxnet {
 /*! \brief mxnet cpu */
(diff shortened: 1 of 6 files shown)


include/mxnet/base.h
```diff
@@ -85,6 +85,18 @@
  */
 #define PROFILER_MESSAGE_FUNCNAME (__FUNCTION__)
 
+/*!
+ * \brief get string from a type_flag (int)
+ */
+#define MSHADOW_TYPE_TO_STRING(type_flag) ::mshadow::TypeFlagStrings[type_flag]
```
wkcn: Thanks for your contribution!
Could you please add a check ‘0 <= type_flag < len(TypeFlagStrings)’?
If the check fails, it will return UnknownType: type_flag.

wkcn:
I prefer to add the type name in DataType class.
https://github.com/apache/incubator-mxnet/blob/master/3rdparty/mshadow/mshadow/base.h#L321

Kh4L:
> I prefer to add the type name in DataType class, and get the type name from ‘mshadow::DataType<DType>::kName’.
> https://github.com/apache/incubator-mxnet/blob/master/3rdparty/mshadow/mshadow/base.h#L321

AFAIK, we can't change `/mshadow/base.h`, as it is part of mshadow, that is archived.

include/mxnet/base.h
Kh4L: Done

wkcn:
Hi @Kh4L,
mshadow is not archived. It has been a part of MXNet, and we can modify it.
https://github.com/apache/incubator-mxnet/pull/15600

Kh4L:
@wkcn @haojin2 understood, I will change it to use this

wkcn:
Hi @haojin2 , the request has been addressed. Could you please help take a review? Thank you!

wkcn:
Hi @haojin2

haojin2:
@Kh4L @wkcn  What is the reason for moving the exisiting function to mshadow?

wkcn:
@haojin2 The reason is that the dtypes belong to mshadow, and it will be more convenient to get the name of mshadow::dtype.

Kh4L:
@haojin2  that's right, as @wkcn , I moved it to mshadow for consistecy, where most of dtype routines are defined.

Issue #16879, opened 2019-11-21
loss for training and evaluation in estimator could be different


liuzh91:
## Description
In current estimator implementation, fit_batch and evaluate_batch use the same loss function.
Code snippet in the `fit_batch` is shown below:
```
     with autograd.record():
            pred = [self.net(x) for x in data]
            loss = [self.loss(y_hat, y) for y_hat, y in zip(pred, label)]
```

The code snippet for evaluate_batch is shown below:
```
        data, label = self._get_data_and_label(val_batch, self.context, batch_axis)
        pred = [self.net(x) for x in data]
        loss = [self.loss(y_hat, y) for y_hat, y in zip(pred, label)]
```

both training and evaluation are using the same loss function `self.loss` to compute the batch loss. In many use cases, it does not hold true. For example, when training LSTM, user may use joint 
regularization loss during training whereas standard cross entropy during evaluation. 

When writing customized estimator, it is cumbersome to define a new loss when evaluation does not share the same loss with training. So it would be good if estimator api could include two losses: `self.train_loss` and `self.evaluate_loss` to tackle different cases separately.

leezu:
How about introducing a new `evaluation_loss` or `evaluate_loss` argument to the constructor. If it is None, we use the train loss during evaluation. For backwards compatibility, let's keep `self.loss` and just introduce a new `self.evaluation_loss`. What do you think? 
Will you open a PR to fix this issue?

liuzh91:
> How about introducing a new `evaluation_loss` or `evaluate_loss` argument to the constructor. If it is None, we use the train loss during evaluation. For backwards compatibility, let's keep `self.loss` and just introduce a new `self.evaluation_loss`. What do you think?
> Will you open a PR to fix this issue?

It is viable to do that. I'll fix this issue.

sxjscience:
Fixed in #16888, Thanks @liuzh91

PR #16888, opened 2019-11-22
Add evaluation_loss to the estimator base class.

liuzh91:
## Description ##
[Bug_fix] Add evaluation loss member in estimator class. The purpose of add the evaluation loss is to decouple the training loss with the evaluation loss in the fit_batch() and evaluate_batch() methods. Please refer to the issue for further information: https://github.com/apache/incubator-mxnet/issues/16879


python/mxnet/gluon/contrib/estimator/estimator.py
```diff
@@ -64,13 +64,17 @@ def __init__(self, net,
...
-                 context=None):
+                 context=None,
+                 evaluation_loss=None):
 
         self.net = net
         self.loss = self._check_loss(loss)
         self._train_metrics = _check_metrics(metrics)
         self._add_default_training_metrics()
         self._add_validation_metrics()
+        self.evaluation_loss = None
```
leezu: How about changing this line to `self.evaluation_loss = self.loss`. Then you can remove the if else clause in the `def evaluate_batch`

liuzh91: If users need customized evaluation loss in the base estimator, they cannot do that any more.

leezu: Why not? The semantics is simpler: `self.evaluation_loss` always refers to the loss used during evaluation. The proposed change is to simply initialize `self.evaluation_loss` with `self.loss` if the user does not specify `evaluation_loss` explicitly.

liuzh91: Yes, you are right.

leezu:
Could you also add a test case to https://github.com/apache/incubator-mxnet/blob/master/tests/python/unittest/test_gluon_estimator.py to prevent future regressions of this feature

python/mxnet/gluon/contrib/estimator/estimator.py
```diff
@@ -59,6 +59,8 @@ class Estimator(object):
         Trainer to apply optimizer on network parameters.
     context : Context or list of Context
         Device(s) to run the training on.
+    evaluation_loss: gluon.loss.loss
+        Loss (objective) function to calculate during evaluation.
```
leezu: It's not clear that `evaluation_loss` will use `loss` if it is None. Could you improve the description here?

PR #16715, opened 2019-11-04
Lamb optimizer update

access2rohit:
## Description ##
adding to new operators:
- lamb_update_phase1
- lamb_update_phase2
Link to paper: https://arxiv.org/pdf/1904.00962.pdf

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- [x] Code is well-documented: 
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] lamb_update, tests, (and when applicable, API doc)

### Testing ###
```
[DEBUG] 1000 of 1000: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=2052238159 to reproduce.
ok

----------------------------------------------------------------------
Ran 1 test in 5085.147s

OK
```

PR #16893, opened 2019-11-23
Multi-tensor LAMB

MoisesHer:
Multi-tensor LAMB Optimizer  **(in development / debugging)**

### Essentials ###
- [X] Changes are complete
- A test of the multi-LAMB optimizer is added in tests/python/unittest/test_optimizer:test_multilamb

### Changes ###
- [X] Added Mxnet operator (CPU & GPU): _multi_lamb_update (and mix precision version _multi_mp_lamb_update). Given weights and gradients of multiple tensors, it updates all of them in parallel
- [X] Python LAMB optimizer was updated to launch the multi-tensor version if MXNET_OPTIMIZER_AGGREGATION_SIZE > 1


leezu:
What's the relation with https://github.com/apache/incubator-mxnet/pull/16715/? #16715 recently added 'LAMB' optimizer to `python/mxnet/optimizer/optimizer.py`. Your code is currently in conflict.

MoisesHer:
> What's the relation with #16715? #16715 recently added 'LAMB' optimizer to `python/mxnet/optimizer/optimizer.py`. Your code is currently in conflict. Please resolve the conflict by merging or rebasing on master.
I merged master, so conflict is resolved. This operator is similar to #16715 but instead of updating a single Tensor, it updates multiple-tensors simultaneously. Thus, it expose more parallelism.

leezu:
Thanks for the clarification. Is it necessary to expose it as a separate `multiLamb` optimizer? Can it be integrated with #16715? Why do we need the less parallel implementation? Or are there other differences? Sorry, I didn't read through the code yet.

python/mxnet/ndarray/contrib.py
```diff
@@ -605,3 +605,27 @@ def multi_mp_adamw_update(weights, grads, mean, var, weights32, rescale_grad, lr
                                                     etas=etas,
                                                     name=name,
                                                     **kwargs)
+
+def multi_lamb_update(weights, grads, mean, var, temp_g, step_count,
+                      out=None, name=None, num_tensors=0, **kwargs):
```
eric-haibin-lin: Could you add some documentation on the input types for these two functions? Otherwise it shows nothing on the API website

python/mxnet/optimizer/optimizer.py
```diff
@@ -1051,6 +1052,95 @@ def update_multi_precision(self, index, weight, grad, state):
         self._update_impl(index, weight, grad, state,
                           multi_precision=use_multi_precision)
 
+@register
+class MultiLAMB(Optimizer):
+    """multiLAMB optimizer.
+    """
+    def __init__(self, learning_rate=0.001, beta1=0.9, beta2=0.999, epsilon=1e-6,
```
eric-haibin-lin: Can we set the default value of lower_bound = None, upperbound = None, and bias_correction=True? Both TF and Pytorch are using the version that does not bound the value of r1 and perform bias correction by default. We should use the same interface as the existing lamb optimizer: https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/optimizer/optimizer.py#L1249-L1254

I noticed that the GPU kernel always performs upper/lower bound clipping on r1. Can we add a branch to skip clipping if upper/lower bound = -1?

src/operator/contrib/multi_lamb-inl.h
```diff
@@ -0,0 +1,332 @@
...
+      Shape1(kernel_params.ntensors), s);
+    pos_wspace += kernel_params.ntensors * sizeof(float);
+    Tensor<xpu, 1, int> block_to_tensor(reinterpret_cast<int*>(&workspace[pos_wspace]),
+      Shape1(kernel_params.nchunks), s);
+    pos_wspace += kernel_params.nchunks * sizeof(int);
+    Tensor<xpu, 1, int> block_to_chunk(reinterpret_cast<int*>(&workspace[pos_wspace]),
+      Shape1(kernel_params.nchunks), s);
+
+    MultiSumSqRun<xpu>(weights, kernel_params.ntensors, r1.dptr_, s);
+    call_kernel1<MPDType, DType>(s, kernel_params, param, block_to_tensor.dptr_,
```
eric-haibin-lin: It looks like MultiSumSq uses atomic add and introduces non-determinism.. We should probably provide a way for users to have deterministic updates. I don't think we need a MultiLAMB optimizer, but rather we reuse the LAMB optimizer class and dispatch to the multi-lamb ops if aggregate size is larger than 1. 
This way, if a user wants deterministic updates, he/she can at least achieve that via setting the aggregation size env var.

python/mxnet/optimizer/optimizer.py
```diff
@@ -1051,6 +1052,95 @@ def update_multi_precision(self, index, weight, grad, state):
...
+        self.upper_bound = upper_bound
+        self.bias_correction = bias_correction
+        self.aggregate_num = max(1, min(50, int(os.getenv('MXNET_OPTIMIZER_AGGREGATION_SIZE', "50"))))
+
+    def create_state(self, index, weight):
+        stype = weight.stype
+        dtype = weight.dtype
+        return (zeros(weight.shape, weight.context, dtype=dtype, stype=stype), # mean
+                zeros(weight.shape, weight.context, dtype=dtype, stype=stype), # variance
+                zeros(weight.shape, weight.context, dtype=dtype, stype=stype)) # temp_g
```
eric-haibin-lin: I am a bit confused for the use of temp_g here. Is this mainly used as a temporarly workspace? Is the value of temp_g at step `t-1` used by the update at step `t`? If so would it be possible to allocate g in the op via resource request? Otherwise the state size is increased by 50% when saving the training state. On the other hand, it is inconsistent with the existing LAMB optimizer which has (mean, var) as the state.

src/operator/contrib/multi_lamb-inl.h
```diff
@@ -0,0 +1,332 @@
...
+    DMLC_DECLARE_FIELD(wd)
+    .set_default(0.0f)
+    .describe("Weight decay augments the objective function with a "
+              "regularization term that penalizes large weights. "
+               "The penalty scales with the square of the magnitude of each weight.");
+    DMLC_DECLARE_FIELD(rescale_grad)
+    .set_default(1.0f)
+    .describe("Gradient rescaling factor");
+    DMLC_DECLARE_FIELD(lower_bound)
+    .set_default(1e-3f)
```
eric-haibin-lin: We can set the default value of lower_bound and upper_bound to -1. And in the GPU kernel, we skip the clipping if the value is -1.

src/operator/contrib/multi_lamb.cu
```diff
@@ -0,0 +1,254 @@
...
+  MPDType r_mean[ILP_LAMB];
+  MPDType r_var[ILP_LAMB];
+  MPDType r_g[ILP_LAMB];
+
+  for (size_t i=startPos; i < stopPos && i < kernel_params.sizes[tensorID];
+                                                  i+= blockDim.x*ILP_LAMB) {
+#pragma unroll
+      for (int ii = 0; ii < ILP_LAMB; ii++) {
+          int load_pos = i + ii*blockDim.x;
+          if (load_pos < stopPos && load_pos < kernel_params.sizes[tensorID]) {
```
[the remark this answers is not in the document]
MoisesHer: done, thanks

python/mxnet/optimizer/optimizer.py
MoisesHer: done, thanks

python/mxnet/optimizer/optimizer.py
MoisesHer: done, thanks

src/operator/contrib/multi_lamb-inl.h
MoisesHer: done, thanks

MoisesHer:
> Ping @MoisesHer Is the MultiLAMB optimizer a generalization of LAMB optimizer? If so, why do we keep the LAMB optimizer? If not, please add documentation or a reference to the docstring. Thank you!

Sorry, since we were having different numbers of states I was not sure how to reuse the previous optimizer. After Haibin suggestion, now we have same numbers of states.
I have merged both optimizers. If aggregation number is <=1 or gradients/weights is a unique NDArray, the single-tensor implementation is used, otherwise, the multi-tensor implementation is used.

eric-haibin-lin:
could you update the code based on https://github.com/apache/incubator-mxnet/pull/17002/ ?

python/mxnet/optimizer/optimizer.py
```diff
@@ -1259,53 +1260,99 @@ def __init__(self, learning_rate=0.001, beta1=0.9, beta2=0.999, epsilon=1e-6,
...
-
-        if multi_precision:
-            mean, var = state[1]
-            weight32 = state[0]
-            g = mp_lamb_update_phase1(weight, grad, mean, var, weight32, wd=wd, **kwargs)
+        if self.aggregate_num <= 1 or not isinstance(index, (tuple, list)):
+            assert(isinstance(weight, NDArray))
+            assert(isinstance(grad, NDArray))
+            self._update_count(index)
+            lr = self._get_lr(index)
```
eric-haibin-lin: let's pass a list of lr, wd to the ops since they can be different.

src/operator/contrib/multi_lamb-inl.h
```diff
@@ -0,0 +1,357 @@
...
+};
+
+template<typename DType, typename MPDType>
+struct MultiLAMBKernelParam {
+  static const int N = 45;
+  size_t ntensors;
+  size_t max_size;
+  size_t total_size;
+  size_t sizes[N];
+  size_t tensor2temp_g[N];
```
eric-haibin-lin: Can this op handle more than 45 tensors? Is it safe to sage MXNET_OPTIMIZER_AGGREGATION_SIZE > 45?

python/mxnet/optimizer/optimizer.py
```diff
@@ -1259,53 +1260,101 @@ def __init__(self, learning_rate=0.001, beta1=0.9, beta2=0.999, epsilon=1e-6,
...
                   'rescale_grad': self.rescale_grad}
 
-        if self.clip_gradient:
-            kwargs['clip_gradient'] = self.clip_gradient
-
-        if multi_precision:
-            mean, var = state[1]
-            weight32 = state[0]
-            g = mp_lamb_update_phase1(weight, grad, mean, var, weight32, wd=wd, **kwargs)
+        if self.aggregate_num <= 1 or not isinstance(index, (tuple, list)):
```
eric-haibin-lin: MXNET_OPTIMIZER_AGGREGATION_SIZE=1 does not seem to work because weight will be a list instead of NDArray

src/operator/contrib/multi_lamb-inl.h
```diff
@@ -0,0 +1,357 @@
...
+    .set_default(1)
+    .describe("Number of tensors");
+  }
+};
+
+template<typename ParamType, int input_stride>
+inline bool MultiLAMB_InferShape(const nnvm::NodeAttrs& attrs,
+                                 mxnet::ShapeVector *in_attrs,
+                                 mxnet::ShapeVector *out_attrs) {
+  const ParamType& param = dmlc::get<ParamType>(attrs.parsed);
```
eric-haibin-lin: Should we also assert the number of tensors to be fewer than 45?

python/mxnet/optimizer/optimizer.py
```diff
@@ -1259,53 +1260,121 @@ def __init__(self, learning_rate=0.001, beta1=0.9, beta2=0.999, epsilon=1e-6,
...
+              t = self._index_update_count[index]
+              weight_ptr = weight
+              grad_ptr = grad
+              if multi_precision:
+                mean, var = state[1]
+                weight32 = state[0]
+              else:
+                mean, var = state
+            else:
+              assert(len(index)==self.aggregate_num)
```
eric-haibin-lin: maybe it's cleaner to have one line for this case:
```
if isinstance(index, (tuple, list)):
    index, weight, grad, state = index[0], weight[0], grad[0], state[0]

# then followed by line 1278 - 1290 shared by both cases
```

src/operator/contrib/multi_lamb.cc
```diff
@@ -0,0 +1,246 @@
...
+                               kernel_params,
+                               r1, r2,
+                               temp_g,
+                               param.lower_bound, param.upper_bound,
+                               req);
+}
+
+DMLC_REGISTER_PARAMETER(MultiLAMBParam);
+
+std::vector<std::string> LAMBParamToVector(uint32_t num_args, const char *pName[], size_t nParams) {
```
eric-haibin-lin: nit: pName -> p_name

eric-haibin-lin:
Thank you @MoisesHer for addressing all the review comments

PR #16898, opened 2019-11-25
Sparse int64 Large tensor support

ChaiBapchya:
## Description ##
Adds support to Sparse tensor creation

### Essentials ###

### Changes ###
- [x] src/c_api/c_api.cc
- [x]  python/mxnet/ndarray/sparse.py
- [x] tests/nightly/test_large_array.py

### Test ###
```
python -m nose tests/nightly/test_large_array.py:test_sparse_dot
.
Ran 1 test in 3.688s

OK
```

### DIFF
--- a/include/mxnet/c_api.h
+++ b/include/mxnet/c_api.h
@@ -643,14 +643,14 @@ MXNET_DLL int MXNDArrayCreateSparseEx(int storage_type,
 
 MXNET_DLL int MXNDArrayCreateSparseEx64(int storage_type,
                                         const int64_t *shape,
-                                        int ndim,
+                                        uint32_t ndim,
                                         int dev_type,
                                         int dev_id,
                                         int delay_alloc,
                                         int dtype,
                                         uint32_t num_aux,
                                         int *aux_type,
-                                        int *aux_ndims,
+                                        uint32_t *aux_ndims,
                                         const int64_t *aux_shape,
                                         NDArrayHandle *out);
 
--- a/include/mxnet/c_api.h
+++ b/include/mxnet/c_api.h
@@ -643,14 +643,14 @@ MXNET_DLL int MXNDArrayCreateSparseEx(int storage_type,
 
 MXNET_DLL int MXNDArrayCreateSparseEx64(int storage_type,
                                         const int64_t *shape,
-                                        uint32_t ndim,
+                                        int ndim,
                                         int dev_type,
                                         int dev_id,
                                         int delay_alloc,
                                         int dtype,
                                         uint32_t num_aux,
                                         int *aux_type,
-                                        uint32_t *aux_ndims,
+                                        int *aux_ndims,
                                         const int64_t *aux_shape,
                                         NDArrayHandle *out);
 
--- a/python/mxnet/ndarray/sparse.py
+++ b/python/mxnet/ndarray/sparse.py
@@ -98,8 +98,8 @@ def _new_alloc_handle(stype, shape, ctx, delay_alloc, dtype, aux_types, aux_shap
         ctypes.c_int(int(_DTYPE_NP_TO_MX[np.dtype(dtype).type])),
         num_aux,
         c_array_buf(ctypes.c_int, native_array('i', aux_type_ids)),
-        c_array_buf(mx_uint, native_array('I', aux_shape_lens)),
-        c_array_buf(mx_uint, native_array('I', aux_shapes)),
+        c_array_buf(mx_int64, native_array('q', aux_shape_lens)),
+        c_array_buf(mx_int64, native_array('q', aux_shapes)),
         ctypes.byref(hdl)))
     return hdl
 
--- a/python/mxnet/ndarray/sparse.py
+++ b/python/mxnet/ndarray/sparse.py
@@ -41,7 +41,7 @@ import numpy as np
 from ..base import NotSupportedForSparseNDArray
 from ..base import _LIB, numeric_types
 from ..base import c_array_buf, mx_real_t, integer_types
-from ..base import mx_uint, NDArrayHandle, check_call
+from ..base import NDArrayHandle, check_call
 from ..context import Context, current_context
 from . import _internal
 from . import op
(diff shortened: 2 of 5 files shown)


include/mxnet/c_api.h
```diff
@@ -643,14 +643,14 @@ MXNET_DLL int MXNDArrayCreateSparseEx(int storage_type,
 
 MXNET_DLL int MXNDArrayCreateSparseEx64(int storage_type,
                                         const int64_t *shape,
-                                        int ndim,
+                                        uint32_t ndim,
```
apeforest: Isn't ndim type always int? why change it here?

include/mxnet/c_api.h
```diff
@@ -643,14 +643,14 @@ MXNET_DLL int MXNDArrayCreateSparseEx(int storage_type,
...
                                         int dev_type,
                                         int dev_id,
                                         int delay_alloc,
                                         int dtype,
                                         uint32_t num_aux,
                                         int *aux_type,
-                                        int *aux_ndims,
+                                        uint32_t *aux_ndims,
```
apeforest: Same here.

tests/nightly/test_large_array.py
```diff
@@ -1678,6 +1678,16 @@ def test_gather():
     assert np.sum(arr[idx[0]] == 2) == SMALL_Y
 
 
+def test_sparse_dot():
+    shape = (LARGE_X, 2)
```
apeforest: Is this large enough?

eric-haibin-lin:
Is there a check list of what to change for int64 support in general?

include/mxnet/c_api.h
ChaiBapchya: Number of dimensions in Large tensor would still be <2**32 right?
Value of single dimension could be > 2**32.

ChaiBapchya:
@eric-haibin-lin while I don't think there is uniform guideline for int64 support across all ops. But in most cases, switching the datatype of all the variables holding "indices" from 32bit dtype to 64bit dtype solves the issue. One still has to test that to ensure it doesn't break the functionality and supports int64.

@apeforest @access2rohit any inputs?

tests/nightly/test_large_array.py
ChaiBapchya: Good catch! Updated.

include/mxnet/c_api.h
access2rohit: > Isn't ndim type always int? why change it here?

keeping it consistent with 32bit version: https://github.com/apache/incubator-mxnet/pull/16898/files#diff-0ce80a2c739b88d1b42356fc2aff2630L633

include/mxnet/c_api.h
access2rohit: https://github.com/apache/incubator-mxnet/pull/16898/files#diff-0ce80a2c739b88d1b42356fc2aff2630R640

access2rohit:
> @eric-haibin-lin while I don't think there is uniform guideline for int64 support across all ops. But in most cases, switching the datatype of all the variables holding "indices" from 32bit dtype to 64bit dtype solves the issue. One still has to test that to ensure it doesn't break the functionality and supports int64.
> 
> @apeforest @access2rohit any inputs?

Plus ensure that the C API being used supports int64_t for indices as well. On language binding side int64_t values need to be passed to C++ backend. For e.g. ctype.int64 should be used for passing large values else they will get truncated at the language binding interface itself.

include/mxnet/c_api.h
apeforest: Why not make everything to int to be consistent? I think internally ndim is always int.

include/mxnet/c_api.h
apeforest: Same as above.

python/mxnet/ndarray/sparse.py
```diff
@@ -41,7 +41,7 @@
 from ..base import NotSupportedForSparseNDArray
 from ..base import _LIB, numeric_types
 from ..base import c_array_buf, mx_real_t, integer_types
-from ..base import mx_uint, NDArrayHandle, check_call
+from ..base import mx_uint, NDArrayHandle, check_call, mx_int64
```
apeforest: I thought we had an effort removing all user-defined types such as mx_uint. Why do we introduce them again? Can we use a standard data type such as ctypes.int64.

include/mxnet/c_api.h
access2rohit: @apeforest just to clarify you want to change 32bit C API signature as well ?
If we do that then it might be a behaviour change for this particualt C API
If we only change it for 64 bit. Then https://github.com/apache/incubator-mxnet/pull/16898/files#diff-3288207540b433242fe8da1390c68d3fR852 
method would require additional template parameter to differentiate ndim type to be used for 32bit version and 64 bit version.

python/mxnet/ndarray/sparse.py
ChaiBapchya: Wasn't aware about it. Addressed it.

include/mxnet/c_api.h
apeforest: i c. In that case, I guess we can use uint32_t for both 32-bit and 64-bit API. But I would also check with Numpy project owner to see if -1 is not used for numpy operator. Thanks!

ChaiBapchya: @reminisce Confirmed with him. -1 is used to indicate "unknown". Internally mxnet does use int. Hence it makes sense to add 64bit API with ndim as int. 
Since 32 bit API existed before with uint32_t, not sure if it makes sense to have a API breaking change.

apeforest:
Adding @reminisce to review the C API change for numpy operator.

python/mxnet/ndarray/sparse.py
```diff
@@ -88,19 +88,34 @@ def _new_alloc_handle(stype, shape, ctx, delay_alloc, dtype, aux_types, aux_shap
...
+            ctypes.c_int(int(_DTYPE_NP_TO_MX[np.dtype(dtype).type])),
+            num_aux,
+            c_array_buf(ctypes.c_int, native_array('i', aux_type_ids)),
+            c_array_buf(ctypes.c_int, native_array('i', aux_shape_lens)),
+            c_array_buf(ctypes.c_int64, native_array('q', aux_shapes)),
+            ctypes.byref(hdl)))
+    else:
+        check_call(_LIB.MXNDArrayCreateSparseEx(
+            ctypes.c_int(int(_STORAGE_TYPE_STR_TO_ID[stype])),
+            c_array_buf(mx_uint, native_array('I', shape)),
```
apeforest: Shouldn't this be ctypes.c_int as well?

python/mxnet/ndarray/sparse.py
```diff
@@ -88,19 +88,34 @@ def _new_alloc_handle(stype, shape, ctx, delay_alloc, dtype, aux_types, aux_shap
...
+            ctypes.c_int(len(shape)),
+            ctypes.c_int(ctx.device_typeid),
+            ctypes.c_int(ctx.device_id),
+            ctypes.c_int(int(delay_alloc)),
+            ctypes.c_int(int(_DTYPE_NP_TO_MX[np.dtype(dtype).type])),
+            num_aux,
+            c_array_buf(ctypes.c_int, native_array('i', aux_type_ids)),
+            c_array_buf(ctypes.c_int, native_array('i', aux_shape_lens)),
+            c_array_buf(mx_uint, native_array('I', aux_shapes)),
```
apeforest: Same here

python/mxnet/ndarray/sparse.py
ChaiBapchya: Missed it!

include/mxnet/c_api.h
reminisce: Thanks for the discussion. Using `int` for `ndim` is consistent with MXNet convention.

src/operator/tensor/cast_storage-inl.h
```diff
@@ -283,14 +283,14 @@ struct CopyCsrDataToDns {
...
-  MSHADOW_XINLINE static void Map(int i,
+  MSHADOW_XINLINE static void Map(index_t i,
                                   DType* dns_data,
                                   const CType* col_idx,
                                   const IType* indptr,
                                   const DType* csr_data,
                                   const nnvm::dim_t num_cols) {
     const nnvm::dim_t offset = i * num_cols;
-    for (IType j = indptr[i]; j < indptr[i+1]; ++j) {
+    for (index_t j = static_cast<index_t>(indptr[i]); j < static_cast<index_t>(indptr[i+1]); ++j) {
```
apeforest: Why change the IType to index_t here?

apeforest:
One last comment. Otherwise looks good to me. Thanks!

eric-haibin-lin:
Does https://cwiki.apache.org/confluence/display/MXNET/Large+Tensor+Support correctly document the current practice/changes?

src/operator/tensor/cast_storage-inl.h
ChaiBapchya: I was under assumption indices for LTS should be int64 (`index_t`). Hence changed the j variable dtype as it iterates over the csr_data to create dns_data

ChaiBapchya:
> Does https://cwiki.apache.org/confluence/display/MXNET/Large+Tensor+Support correctly document the current practice/changes?

Yes. It enlists the 3 steps needed to extend LTS for operators (includes changes on Front-end language binding side and Back-end C++ api side). I updated the wiki. Please verify. @eric-haibin-lin

src/operator/tensor/cast_storage-inl.h
ChaiBapchya: Found that IType and CType will be int64 anyway. No need for hard-code to index_t

src/c_api/c_api.cc
```diff
@@ -972,6 +972,44 @@ int MXNDArrayCreateSparseEx(int storage_type,
       Context::Create(static_cast<Context::DeviceType>(dev_type), dev_id),
       delay_alloc != 0,
       dtype, aux_types, aux_shapes);
+}
+
+int MXNDArrayCreateSparseEx(int storage_type,
+                            const uint32_t *shape,
+                            uint32_t ndim,
```
apeforest: Can we also change the ndim type here from uint32_t to int? It looks strange to me why 32-bit version uses uint32_t while 64-bit version uses int.

src/c_api/c_api.cc
```diff
@@ -972,6 +972,44 @@ int MXNDArrayCreateSparseEx(int storage_type,
...
+                            int dev_id,
+                            int delay_alloc,
+                            int dtype,
+                            uint32_t num_aux,
+                            int *aux_type,
+                            uint32_t *aux_ndims,
+                            const uint32_t *aux_shape,
+                            NDArrayHandle *out) {
+  API_BEGIN();
+  CreateSparseNDArray<uint32_t, uint32_t>(storage_type, shape, ndim, dev_type, dev_id,
```
apeforest: Can we also change the ndim type here from uint32_t to int? It looks strange to me why 32-bit version uses uint32_t while 64-bit version uses int.

ChaiBapchya:

Issue #16869, opened 2019-11-20
Cannot manipulate gradients during training in estimator


liuzh91:
**Description**
In current estimator framework, api fit_batch() forward the training batch through the network, compute the loss and gradients and update network parameters with gradients. The default 
fit_batch() function is shown as below:

```
def fit_batch(self, train_batch, batch_axis=0):
        data, label = self._get_data_and_label(train_batch, self.context, batch_axis)
        batch_size = train_batch[0].shape[batch_axis]

        with autograd.record():
            pred = [self.net(x) for x in data]
            loss = [self.loss(y_hat, y) for y_hat, y in zip(pred, label)]

        for l in loss:
            l.backward()

        self.trainer.step(batch_size)

        return data, label, pred, loss`
```
The problem is in some use cases like gradient accumulation, users may not update gradient at each batch. They apply gradients on network parameters periodically. There is no way to implement the gradient accumulation feature in customized estimators.

One possible option is to create a handler to update/apply gradients at the end of each batch, but it violates the definition of handlers since handlers are not supposed to apply gradients on network parameters.  Gradient updates should be processed in the fit_batch() api.

So a viable way to resolve the problem is to decompose the fit_batch() function into two methods:
-  computations of loss/gradient  
-  applying gradients on network parameters 

between these two methods we could add some customized api to process gradients, e.g., gradient clipping or gradient accumulation.

leezu:
Thanks @liuzh91 for raising the issue. I agree the current `fit_batch` is not sufficient to suppose the use-case.

What do you mean with "handlers are not supposed to apply gradients on network parameters"?

If we allow handlers to apply gradient to network parameter, an alternative solution to your problem could be to remove the `trainer.step` from the current `def fit_batch` implementation and introduce a `GradientUpdateHandler`? Similar to how `LoggingHandler` and `MetricHandler` are constructed by default in `estimator.fit`, if not provided by the user, the `GradientUpdateHandler` would be default implement the `estimator.step` behavior.

Users would be free to provide their own subclass of `GradientUpdateHandler` that could extend the handler to do gradient clipping or gradient accumulation.

liuzh91:
> Thanks @liuzh91 for raising the issue. I agree the current `fit_batch` is buggy as it cannot support this standard use-case.
> 
> What do you mean with "handlers are not supposed to apply gradients on network parameters"?
> 
[...]

PR #16900, opened 2019-11-25
introduce  gradient update handler to the  base estimator

liuzh91:
## Description ##
This change add default gradient update handler to the base estimator class. The purpose of this change is to decouple the forward/backward computation with the gradient apply process. In some use cases such as gradient accumulation, gradient updates are not executed during each training batch.  See issue description #16869.

### DIFF
--- a/python/mxnet/gluon/contrib/estimator/estimator.py
+++ b/python/mxnet/gluon/contrib/estimator/estimator.py
@@ -24,7 +24,7 @@ import logging
 import sys
 import warnings
 
-from .event_handler import MetricHandler, ValidationHandler, LoggingHandler, StoppingHandler
+from .event_handler import MetricHandler, ValidationHandler, LoggingHandler, StoppingHandler, GradientUpdateHandler
 from .event_handler import TrainBegin, EpochBegin, BatchBegin, BatchEnd, EpochEnd, TrainEnd
 from .event_handler import _check_event_handlers
 from .utils import _check_metrics, _suggest_metric_for_loss, _check_handler_metric_ref
@@ -302,12 +302,10 @@ class Estimator(object):
 
         with autograd.record():
             pred = [self.net(x) for x in data]
-            loss = [self.loss(y_hat, y) for y_hat, y in zip(pred, label)]
+            loss = [self.loss(y_hat, y) / batch_size for y_hat, y in zip(pred, label)]
 
         for l in loss:
             l.backward()
-
-        self.trainer.step(batch_size)
 
         return data, label, pred, loss
 
@@ -413,6 +411,9 @@ class Estimator(object):
 
         # no need to add to default handler check as StoppingHandler does not use metrics
         added_default_handlers.append(StoppingHandler(self.max_epoch, self.max_batch))
+
+        if not any(isinstance(handler, GradientUpdateHandler) for handler in event_handlers):
+            added_default_handlers.append(GradientUpdateHandler())
 
         if not any(isinstance(handler, MetricHandler) for handler in event_handlers):
             added_default_handlers.append(MetricHandler(train_metrics=self.train_metrics))
--- a/python/mxnet/gluon/contrib/estimator/estimator.py
+++ b/python/mxnet/gluon/contrib/estimator/estimator.py
@@ -302,7 +302,7 @@ class Estimator(object):
 
         with autograd.record():
             pred = [self.net(x) for x in data]
-            loss = [self.loss(y_hat, y) / batch_size for y_hat, y in zip(pred, label)]
+            loss = [self.loss(y_hat, y) for y_hat, y in zip(pred, label)]
 
         for l in loss:
             l.backward()
@@ -358,6 +358,7 @@ class Estimator(object):
 
         self.max_epoch = epochs
         self.max_batch = batches
+        self.batch_axis = batch_axis
 
         # provide default handlers
         event_handlers = self._prepare_default_handlers(val_data, event_handlers)
--- a/python/mxnet/gluon/contrib/estimator/event_handler.py
+++ b/python/mxnet/gluon/contrib/estimator/event_handler.py
@@ -28,7 +28,6 @@ import numpy as np
 from ....metric import CompositeEvalMetric, EvalMetric
 from ....metric import Loss as metric_loss
 from .utils import _check_metrics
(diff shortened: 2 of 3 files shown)


ptrendx:
@leezu What is the rationale behind marking this PR as 1.6? This looks like adding a new functionality.

tests/python/unittest/test_gluon_estimator.py
```diff
@@ -332,39 +332,41 @@ def test_default_handlers():
...
     est = Estimator(net=net,
                     loss=loss,
                     metrics=train_acc,
                     trainer=trainer,
                     context=ctx)
-    # no handler(all default handlers), no warning
+    # no handler except gradient update handler (all default handlers), no warning
     with warnings.catch_warnings(record=True) as w:
-        est.fit(train_data=train_data, epochs=num_epochs)
+        est.fit(train_data=train_data, epochs=num_epochs, event_handlers=[gradient_update])
```
roywei: you don't need to add `gradient_update` here right? as it's added as a default handler?

liuzh91: My mistake. I'll remove this handler here.

leezu:
@ptrendx the estimator API is currently experimental and is not yet up to more complex real-world use-cases. @liuzh91 is currently converting GluonNLP training scripts to make use of the Estimator API and ran into several remaining shortcomings of the API.
We shouldn't release 1.6 with a known broken version of Estimator *if fixes are available*. Thus if this PR is approved and merged to master in time for the 1.6 release, it may be good to backport. Given the scope of the estimator API and the current shortcomings we can consider these PRs as bugfixes. If we don't fix the API with the 1.6 release, we may be bound to the broken API due to backwards compatibility commitments.

Do you think backporting the fix would be reasonable or do you have concerns? When do you plan to tag a release candidate?

These extends the fixes commited started November 2019 https://github.com/apache/incubator-mxnet/commits/master/python/mxnet/gluon/contrib/estimator.

python/mxnet/gluon/contrib/estimator/estimator.py
```diff
@@ -295,13 +295,11 @@ def fit_batch(self, train_batch, batch_axis=0):
 
         with autograd.record():
             pred = [self.net(x) for x in data]
-            loss = [self.loss(y_hat, y) for y_hat, y in zip(pred, label)]
+            loss = [self.loss(y_hat, y) / batch_size for y_hat, y in zip(pred, label)]
```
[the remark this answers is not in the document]
liuzh91: I'll look into this problem.

liuzh91:
> Thank you for the improvement! 2 concerns.
> Also could you point an example which require custom gradient handler? (gradient clipping or aggregation)

Thank u for the review. 

For the gradient update example, one use case of using gradient accumulation appears when training a transformer. (https://github.com/dmlc/gluon-nlp/blob/master/scripts/machine_translation/train_transformer.py#L320) Because the size of parameters in the transformer network is too large, we can compute gradient for a small batch of data examples. In this case, the gradient is updated periodically on the weight parameters.

python/mxnet/gluon/contrib/estimator/event_handler.py
```diff
@@ -130,13 +130,15 @@ class MetricHandler(EpochBegin, BatchEnd):
     ----------
     train_metrics : List of EvalMetrics
         Training metrics to be updated at batch end.
+    priority : scalar
+        Priority level of the MetricHandler
```
leezu: Can you explain that the priority is used for ascending sorting of the handlers?

python/mxnet/gluon/contrib/estimator/event_handler.py
```diff
@@ -176,14 +178,17 @@ class ValidationHandler(TrainBegin, BatchEnd, EpochEnd):
     batch_period : int, default None
         How often to run validation at batch end, by default
         :py:class:`ValidationHandler` does not validate at batch end.
+    priority: scalar, default -1000
+        Priority level of the ValidataionHandler
```
leezu: Typo

python/mxnet/gluon/contrib/estimator/event_handler.py
```diff
@@ -235,14 +240,17 @@ class LoggingHandler(TrainBegin, TrainEnd, EpochBegin, EpochEnd, BatchBegin, Bat
         Training metrics to be logged, logged at batch end, epoch end, train end.
     val_metrics : list of EvalMetrics
         Validation metrics to be logged, logged at epoch end, train end.
+    priority : scalar, default np.Inf
+        Priority level of the LoggingHandler
```
leezu: Explanation of the priority level would be helpful for users.

python/mxnet/gluon/contrib/estimator/event_handler.py
```diff
@@ -706,3 +714,29 @@ def train_end(self, estimator, *args, **kwargs):
...
+
+class GradientUpdateHandler(BatchEnd):
+    """Gradient Update Handler that apply gradients on network weights
+
+    :py:class:`GradientUpdateHandler` takes the priority level. It updates weight parameters
+    at the end of each batch
+
+    Parameters
+    ----------
+    priority : scalar, default -np.Inf
```
leezu: Should a integer be used instead of `-np.Inf`? Maybe there are some use-cases where a handler should be run prior to the gradient handler?

python/mxnet/gluon/contrib/estimator/event_handler.py
liuzh91: I have fixed these documentation issues.

liuzh91:
>     1. Does this require the user to specify grad_req = 'add' for gradient accumulation? When is the gradient reset to zeros?
> 
>     2. Now GradientUpdateHandler is added to the list of default handlers. If users explicitly create estimator with `event_handlers=[metric, logging]`, will `GradientUpdateHandler` still be included in the estimator and performing updates correctly?

1. The gradient update handler is for general gradient update. When we implement gradient accumulation handler in the future, we will inherit this class.

2. Yes, it is added as a default handler. Please refer to (https://github.com/apache/incubator-mxnet/blob/5878451c4b476d8771d93000ff5d51b81071997d/tests/python/unittest/test_gluon_estimator.py#L369).

Issue #16904, opened 2019-11-26
Upgrade MXNet version to 1.6.0


samskalicky:
Need to uprev to 1.6 here:
https://github.com/apache/incubator-mxnet/blob/v1.6.x/include/mxnet/base.h#L76

samskalicky:
@mxnet-label-bot add [R1.6.0]

ptrendx:
Removing 1.6 label, since the relevant PR was merged to 1.6.

PR #16905, opened 2019-11-26
updating MXNet version to 1.6.0 in base.h for C APIs

access2rohit:
## Description ##
updating MXNet version to 1.6.0 in base.h for C APIs
https://github.com/apache/incubator-mxnet/issues/16904

### Essentials ###

### DIFF
--- a/include/mxnet/base.h
+++ b/include/mxnet/base.h
@@ -73,7 +73,7 @@
 /*! \brief major version */
 #define MXNET_MAJOR 1
 /*! \brief minor version */
-#define MXNET_MINOR 5
+#define MXNET_MINOR 6
 /*! \brief patch version */
 #define MXNET_PATCH 0
 /*! \brief mxnet version */


access2rohit:
@lanking520

access2rohit:
> This is a key problem with the publish, we should bump it once we make the release

yes

access2rohit:
@ptrendx

samskalicky:
@access2rohit shouldnt we be changing master to 1.7?

access2rohit:
> ?

@samskalicky Sure! I think we should first bump master to 1.6 and then again to 1.7. Does that makes sense ?

ptrendx:
@access2rohit Since the 1.6.x branch is already diverged from master, there is no need to do this 1.6 step I think - we should just go directly to 1.7 (which I believe requires changes in more places).

access2rohit:
> @access2rohit Since the 1.6.x branch is already diverged from master, there is no need to do this 1.6 step I think - we should just go directly to 1.7 (which I believe requires changes in more places).

Yes it requires changes in more places. IMO we should follow proper progression by moving from 1.5->1.6->1.7 to keep the change history in master clean and consistent. 
@ptrendx do you agree ?

PR #16909, opened 2019-11-26
Build with cv 4.0

ewail:
## Description ##
Build with cv 4.0

### Changes ###
Change CV_IMWRITE_PNG_COMPRESSION to cv::IMWRITE_PNG_COMPRESSION
Change CV_IMWRITE_JPEG_QUALITY to cv::IMWRITE_JPEG_QUALITY

## Comments ##
- Adapt to the new opencv.

### DIFF
--- a/R-package/src/im2rec.cc
+++ b/R-package/src/im2rec.cc
@@ -169,11 +169,11 @@ void IM2REC::im2rec(const std::string & image_lst, const std::string & root,
   std::vector<unsigned char> encode_buf;
   std::vector<int> encode_params;
   if (encoding == std::string(".png")) {
-    encode_params.push_back(CV_IMWRITE_PNG_COMPRESSION);
+    encode_params.push_back(cv::IMWRITE_PNG_COMPRESSION);
     encode_params.push_back(quality);
     LOG(INFO) << "PNG encoding compression: " << quality;
   } else {
-    encode_params.push_back(CV_IMWRITE_JPEG_QUALITY);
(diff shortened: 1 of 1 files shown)


wkcn:
Thank you for the contribution!
We need the compatibility with the lower version of OpenCV.
Could you please define some macros in [`src/io/opencv_compatibility.h`](https://github.com/apache/incubator-mxnet/blob/master/src/io/opencv_compatibility.h), rather than replacing them?

ewail:
> Thank you for the contribution!
> 
> We need the compatibility with the lower version of OpenCV. The lower version doesn't define `cv::IMWRITE_PNG_COMPRESSION` and `cv::IMWRITE_JPEG_QUALITY`.
> 
> I recommend to re-define the macro when using OpenCV4, like [src/io/opencv_compatibility.h](https://github.com/apache/incubator-mxnet/blob/master/src/io/opencv_compatibility.h#L52-L53).
> 
> ```c++
> #if CV_VERSION_MAJOR >= 4
> #include <opencv2/opencv.hpp>
> #define CV_IMWRITE_PNG_COMPRESSION cv::IMWRITE_PNG_COMPRESSION
> #define CV_IMWRITE_JPEG_QUALITY cv::IMWRITE_JPEG_QUALITY
> #endif  // CV_VERSION_MAJOR >= 4
> ```
> 
> Thanks again : )

Many Thanks, and let me test that.

wkcn:
@ewail Welcome to submit the PR : )

PR #16922, opened 2019-11-27
Update LoggingHandler to support logging per interval

### LINKED ISSUE #16918: Configurable log interval for estimator
Currently the gluon Estimator has a LoggingHandler, which supports logging per batch and per epoch. It would be good to have a configurable logging interval (e.g. logging every 50 batches). Logging per epoch does not provide enough information, while logging per batch is tooooo verbose

liuzh91:
## Description ##
Add support of logging per interval to the default LoggingHandler. It acts as a middle ground between logging per batch and logging per epoch. Please refer to the description in Issue

Fixes #16918

### DIFF
--- a/python/mxnet/gluon/contrib/estimator/event_handler.py
+++ b/python/mxnet/gluon/contrib/estimator/event_handler.py
@@ -241,17 +241,16 @@ class LoggingHandler(TrainBegin, TrainEnd, EpochBegin, EpochEnd, BatchBegin, Bat
     """
 
     LOG_PER_EPOCH = 1
-    LOG_PER_BATCH = 2
-    LOG_PER_INTERVAL = 3
+    LOG_PER_INTERVAL = 2
 
     def __init__(self, verbose=LOG_PER_EPOCH,
                  train_metrics=None,
                  val_metrics=None,
                  log_interval=1):
         super(LoggingHandler, self).__init__()
-        if verbose not in [self.LOG_PER_EPOCH, self.LOG_PER_BATCH, self.LOG_PER_INTERVAL]:
+        if verbose not in [self.LOG_PER_EPOCH, self.LOG_PER_INTERVAL]:
             raise ValueError("verbose level must be either LOG_PER_EPOCH or "
-                             "LOG_PER_BATCH, received %s. "
+                             "LOG_PER_INTERVAL, received %s. "
                              "E.g: LoggingHandler(verbose=LoggingHandler.LOG_PER_EPOCH)"
                              % verbose)
         self.verbose = verbose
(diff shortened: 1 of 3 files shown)


python/mxnet/gluon/contrib/estimator/event_handler.py
```diff
@@ -286,21 +293,31 @@ def train_end(self, estimator, *args, **kwargs):
         estimator.logger.info(msg.rstrip(', '))
 
     def batch_begin(self, estimator, *args, **kwargs):
-        if self.verbose == self.LOG_PER_BATCH:
+        if self.verbose >= self.LOG_PER_BATCH:
```
eric-haibin-lin: this assumes the order of these constant variables. Better to use `if self.verbose in [var1, var2]` instead.

liuzh91: I'll fix this.

liuzh91:
> Shall we add a unit test for this?

What assertion shall we make in this case?

leezu:
You could assert the number of lines logged. For example, if you specify training for 10 batches and log every 5 batches, you may test for two lines being logged.

liuzh91:
> You could assert the number of lines logged. For example, if you specify training for 10 batches and log every 5 batches, you may test for two lines being logged.

Ok, I'll take a try.

liuzh91:
> Thanks for the fix. Shall we remove `LOG_PER_BATCH` as it can be replaced by `LOG_PER_INTERVAL`?
> Since this is in contrib, I think it's fine and less options will be easier for users.

I have thought about the option before. Most parts of logging batch and logging interval are overlapped. But during batch logging, we only logged the training metrics:
```
            for metric in self.train_metrics:
                # only log current training loss & metric after each batch
                name, value = metric.get()
                msg += '%s: %.4f, ' % (name, value)
``` 
But during interval logging, I think it is better to log both training and validation metrics. So `LOG_PER_BATCH` is still necessary?
```
          for monitor in self.train_metrics + self.val_metrics:
                name, value = monitor.get()
                msg += '%s: %.4f, ' % (name, value)
```
Or we can merge `LOG_PER_BATCH` and `LOG_PER_INTERVAL`, then we use `self.log_interval` to check whether using validation metrics or not.

leezu:
I'm actually not sure if logging both `self.train_metrics + self.val_metrics` during `LOG_PER_INTERVAL` is good. 
Currently the notion of `val_metrics` and `train_metrics` is not clearly decoupled. Thinking about how to tackle https://github.com/apache/incubator-mxnet/issues/16959 may also help to clarify the relation of `eval_metrics` and `train_metrics` and how their values should be updated and logged.

liuzh91:
> I'm actually not sure if logging both `self.train_metrics + self.val_metrics` during `LOG_PER_INTERVAL` is good.
> Currently the notion of `val_metrics` and `train_metrics` is not clearly decoupled. Thinking about how to tackle #16959 may also help to clarify the relation of `eval_metrics` and `train_metrics` and how their values should be updated and logged.

It makes sense to me. I'll merge `LOG_PER_BATCH` and `LOG_PER_INTERVAL` and leave out the `val_metrics`.

python/mxnet/gluon/contrib/estimator/event_handler.py
```diff
@@ -241,17 +241,16 @@ class LoggingHandler(TrainBegin, TrainEnd, EpochBegin, EpochEnd, BatchBegin, Bat
...
 
     LOG_PER_EPOCH = 1
-    LOG_PER_BATCH = 2
-    LOG_PER_INTERVAL = 3
+    LOG_PER_INTERVAL = 2
 
     def __init__(self, verbose=LOG_PER_EPOCH,
                  train_metrics=None,
                  val_metrics=None,
                  log_interval=1):
```
leezu: Do you think it would make sense to drop the `verbose` argument completely? For example, if `log_interval=None`, we only log per epoch. If log_interval is set to any positive integer, we log according to that interval?
For example, the following constructor:
```
def __init__(self, log_interval=None,
                 train_metrics=None,
                 val_metrics=None):
```

liuzh91: It can be done in this way, but I think verbose is more user friendly. If we use `log_interval=None` in the logging handler, user may need to understand this hack from the documentation. I think we should leave verbose here.

leezu: `log_interval='epoch' `?

liuzh91: Then the type of the variable will be confusing. In the comments of `LoggingHandler`, we have `log_interval: int, default 1`. If we assign a value other than integer type, user may complain about it.

Even though there is no type declaration in python, I believe it is still convention that a variable should only be of one type. It is also beneficial for maintenance.

leezu: My concern is that having constants
``` python 
LOG_PER_EPOCH = 1
LOG_PER_INTERVAL = 2
```
is not very pythonic. For example users will need to specify

``` python 
est = Estimator(Estimator.LOG_PER_INTERVAL, log_interval=20, train_metrics=X)
```

users may easily by mistake specify

``` python 
est = Estimator(log_interval=20, train_metrics=X)
```

and wonder about why things don't work.

Instead, it's much simpler to 

``` python 
est = Estimator(log_interval='epoch', train_metrics=X)
est = Estimator(log_interval=20, train_metrics=X)
```

or

 ``` python 
est = Estimator('epoch', train_metrics=X)
est = Estimator(20, train_metrics=X)
```

Having variable type arguments is fine in Python / in this case IMO.

Comment (liuzh91) [reply python/mxnet/gluon/contrib/estimator/event_handler.py]: I see. I'll remove the verbose part and add some documentation to address this issue.

File: python/mxnet/gluon/contrib/estimator/event_handler.py [hunk 3]
```diff
@@ -227,33 +227,22 @@ class LoggingHandler(TrainBegin, TrainEnd, EpochBegin, EpochEnd, BatchBegin, Bat
...
-        verbose=LOG_PER_EPOCH: display metrics every epoch
-        verbose=LOG_PER_BATCH: display metrics every batch
-        verbose=LOG_PER_INTERVAL: display metrics every interval of batches
     train_metrics : list of EvalMetrics
         Training metrics to be logged, logged at batch end, epoch end, train end.
     val_metrics : list of EvalMetrics
         Validation metrics to be logged, logged at epoch end, train end.
-    log_interval: int, default 1
-        Logging interval during training. 1 is equivalent to LOG_PER_BATCH.
+    log_interval: int or str, default 'epoch'
```
leezu: Make this the first argument?

liuzh91: Ok

PR #16923, opened 2019-11-27
[WIP] Fallback mechanism for mx.np operators

reminisce:
## Description ##
Fallback mechanism for `mx.np` operators.

@haojin2

### DIFF
--- a/python/mxnet/numpy/multiarray.py
+++ b/python/mxnet/numpy/multiarray.py
@@ -46,7 +46,7 @@ from ..ndarray import numpy as _mx_nd_np
 from ..ndarray.numpy import _internal as _npi
 from ..ndarray.ndarray import _storage_type
 from .utils import _get_np_op
-from .fallback import *
+from .fallback import *  # pylint: disable=wildcard-import,unused-wildcard-import
 from . import fallback
 
 __all__ = ['ndarray', 'empty', 'empty_like', 'array', 'shape',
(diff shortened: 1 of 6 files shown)


python/mxnet/numpy/multiarray.py
```diff
@@ -202,13 +220,13 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):  # pylint: disable=
...
-                                 'the _NUMPY_ARRAY_UFUNC_LIST. Please make sure you are '
-                                 'using NumPy >= 1.15.0 and the operator implementation '
-                                 'is compatible with NumPy. Then add the operator name '
-                                 'to the list.'
-                                 .format(name))
-            return mx_ufunc(*inputs, **kwargs)
+                # try to fallback to official NumPy op
+                onp_op = _get_np_op(name)
+                new_inputs = [arg.asnumpy() if isinstance(arg, ndarray) else arg for arg in inputs]
+                out = onp_op(*new_inputs, **kwargs)
```
wkcn: It will break the computational graph, and could not compute the gradient.

reminisce: We are aware of this. More sophisticated fallback mechanism is illustrated in https://github.com/apache/incubator-mxnet/pull/16698 by leveraging `CustomOp`. To reach 100% NumPy op coverage within a month, this the simplest and fastest pathway though. In the future, we will gradually replace those fallback ops with native implementation in backend.

wkcn: It may be better to use `mx.autograd.Function` to wrap these numpy operators.

xidulu: @wkcn 
 `mx.autograd.Function`  currently does not support DeepNumpy, some extra infrastructure is required.
Also, I am not sure if `mx.autograd.Function` could be integrated into HybridBlock, I cannot find corresponding cases covered in the unit tests.

PR #16928, opened 2019-11-27
[CI] Disable builds based on ubuntu 14.04 as Ubuntu 14 is EOL and the…

larroy:
… container can't be built right now.

### DIFF
--- a/src/ndarray/ndarray.cc
+++ b/src/ndarray/ndarray.cc
@@ -684,7 +684,7 @@ void NDArray::CopyFrom(const mkldnn::memory &mem) {
     ptr_->Reorder2Default();
 
   const mkldnn::memory *this_mem = GetMKLDNNData();
-  MKLDNNCopy(mem, this_mem);
+  MKLDNNMemoryCopy(mem, this_mem);
 }
 
 mkldnn::memory *NDArray::CreateMKLDNNData(const mkldnn::memory::desc &desc) {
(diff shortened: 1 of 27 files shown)


larroy:
@mseth10 @samskalicky @lanking520

larroy:
See errors here: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-cpu/detail/master/1339/pipeline

leezu:
Do you plan to re-enable this test based on Ubuntu 16.04 shortly after?

leezu:
If we want to continue using 14.04, we may need to sign up for https://ubuntu.com/esm ?

leezu:
@larroy actually it is sufficient to disable the requirement for "ci/jenkins/mxnet-validation/unix-cpu" to pass on the Github side. Then it doesn't matter if the test fails or passes.
It's impossible to merge this PR anyways before disabling "ci/jenkins/mxnet-validation/unix-cpu" requirement

marcoabreu:
Instead of turning them off, we can also just adjust the sources file which tells Apt where to grab the packages (replace live with archive repo). It's normal that the main distribution servers get turned off once something goes into EOL. But since we still rely on 1404 for binary distribution, we need to keep checking or update.

See https://askubuntu.com/questions/91815/how-to-install-software-or-upgrade-from-an-old-unsupported-release for instructions.

leezu:
@larroy how to move forward with this? All PR's are currently blocked by this issue.

larroy:
@leezu we are working on this.

larroy:
@marcoabreu feel free to open a PR.  

@marcoabreu cache refresh is failing again, and you blocked a previous PR that helped me diagnose (the one about output) again it's me having to fix this. 

There's 3 jobs failing now on master which prevent PRs from being merged, unix_gpu centos and cache refresh which is also impacting builds.

larroy:
Trying to repro this in EC2 and seems to run, so I might have to get the CI environment AMI to reproduce. Weird.

PR #16941, opened 2019-11-28
Fix crashing on Windows in ObjectPool ~ctor

dmitry-markeshov:
## Description ##
On Windows in ObjectPool ~ctor _aligned_free should be used instead free
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc?view=vs-2019


wkcn:
Merged. Thank you for the fix : )

PR #16955, opened 2019-12-01
[Dataset] add flatten API to dataset

zburning:
## Description ##
add flatten API to dataset. It can be useful when one transforms a single sample into multiple samples.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/python/mxnet/gluon/data/dataset.py
+++ b/python/mxnet/gluon/data/dataset.py
@@ -184,6 +184,19 @@ class Dataset(object):
             The transformed dataset.
         """
         return self.transform(_TransformFirstClosure(fn), lazy)
+    
+    def flatten(self):
+    """Returns a new dataset with samples flattened.
+
+    It is usefull to call after transform() when your transformer function 'fn' transform
+    each sample into multipul samples.
+
+    Returns
+    -------
+    Dataset
+        The result dataset.
+    """
+    return SimpleDataset(list(itertools.chain.from_iterable(self)))
 
 
 class SimpleDataset(Dataset):
--- a/python/mxnet/gluon/data/dataset.py
+++ b/python/mxnet/gluon/data/dataset.py
@@ -190,6 +190,7 @@ class Dataset(object):
 
     It is usefull to call after transform() when your transformer function 'fn' transform
     each sample into multipul samples.
+    Note that the items in dataset should be iterable, e.g., list.
 
     Returns
     -------
--- a/tests/python/unittest/test_gluon_data.py
+++ b/tests/python/unittest/test_gluon_data.py
@@ -360,6 +360,14 @@ def test_dataset_take():
         total += sample
     assert total == expected_total
 
+def test_dataset_flatten():
+    length = 100
+    a = mx.gluon.data.SimpleDataset([i for i in range(length)])
+    b = mx.gluon.data.SimpleDataset([[4 * i + j for j in range(4)] for i in range(length // 4)])
+    b = b.flatten()
+    assert len(a) == length
+    assert a == b
+    
 def test_dataloader_scope():
     """
     Bug: Gluon DataLoader terminates the process pool early while
--- a/tests/python/unittest/test_gluon_data.py
+++ b/tests/python/unittest/test_gluon_data.py
@@ -364,9 +364,17 @@ def test_dataset_flatten():
     length = 100
     a = mx.gluon.data.SimpleDataset([i for i in range(length)])
     b = mx.gluon.data.SimpleDataset([[4 * i + j for j in range(4)] for i in range(length // 4)])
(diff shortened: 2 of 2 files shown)


python/mxnet/gluon/data/dataset.py
```diff
@@ -185,6 +185,20 @@ def transform_first(self, fn, lazy=True):
         """
         return self.transform(_TransformFirstClosure(fn), lazy)
 
+    def flatten(self):
+        """Returns a new dataset with samples flattened.
+
+        It is usefull to call after transform() when your transformer function 'fn' transform
+        each sample into multipul samples.
```
leezu: typo: multipul

python/mxnet/gluon/data/dataset.py
```diff
@@ -185,6 +185,20 @@ def transform_first(self, fn, lazy=True):
...
+        Note that the items in dataset should be iterable, e.g., list.
```
leezu: The items in the dataset must be iterable.

python/mxnet/gluon/data/dataset.py
```diff
@@ -185,6 +185,20 @@ def transform_first(self, fn, lazy=True):
...
+
+        Returns
+        -------
+        Dataset
+            The result dataset.
```
leezu: The flattened dataset.

sxjscience:
Is it similar to the `flat_map` in TF? https://www.tensorflow.org/api_docs/python/tf/data/Dataset#flat_map

szha:
@sxjscience it's only the flatten part. Since our dataset API already supports lazy transformation, it doesn't seem necessary to entangle the transform function with flatten.

python/mxnet/gluon/data/dataset.py
```diff
@@ -185,6 +185,20 @@ def transform_first(self, fn, lazy=True):
...
+        It is usefull to call after transform() when your transformer function 'fn' transform
+        each sample into multiple samples.
+        Note that the items in dataset must be iterable, e.g., list.
+
+        Returns
+        -------
+        Dataset
+            The flattened dataset.
+        """
+        return SimpleDataset(list(itertools.chain.from_iterable(self)))
```
szha: would it make sense to include a lazy flatten option? Just a thought.

sxjscience:
@szha I mean that the added feature is similar to the flat_map in TF. What's our plan for enhancing the Dataset API? Making it lazy is not possible I guess.

szha:
@sxjscience yes, it's similar to flat_map, minus the map part. Using itertools should make it possible for lazy flattening. I'm not sure what use cases there would be yet as it merely came from analogy. Plan for dataset API would be a larger topic.

sxjscience:
@szha I mean if there is a plan for matching the behavior of TF Dataset, we'd better provide similar functionalities. I don't think we can `getitem` via `itertool`. Making this call lazy requires that. The major assumption of `Dataset` is that we can `__getitem__`.

szha:
@sxjscience I don't think there's such plan. By lazy I meant that `__getitem__` and `__len__` can be the trigger to the actual flattening.

sxjscience:
Explicitly constructing the dataset may be a better choice than adding a `flatten` method.
```python
new_dataset = preprocess_function(dataset)
```

From my perspective, the major design choice of `gluon.dataset` is to support `__getitem__` + lazy evaluation in `transform()`. With the help of lazy evaluation, we can generate the data on-the-fly and the overall data processing pipeline uses less memory. However, the `flatten` method  is equivalent to this python one-liner `SimpleDataset(list(itertools.chain.from_iterable(self)))` and there is no speed/memory benefit.

Moreover, think about the case where each sample is a (data, label) pair. Calling flatten() will make the dataset look like `[data0, label0, data1, label1, ...]`, which is not very meaningful.

I suggest we should just use `SimpleDataset(list(itertools.chain.from_iterable(self)))` to implement this functionality.

sxjscience:
Sometimes the user may want to write their own data preprocessing pipeline, like

https://github.com/huggingface/transformers/blob/b3d834ae11381ca493da97f717d77d185ca7d780/examples/utils_squad.py#L190-L418

This is doable with the current Gluon Dataset and we need to teach the user about this usage.

python/mxnet/gluon/data/dataset.py
zburning: I guess you mean having a new class for example, flattenedDataset which implements __getitem__ and __len__ ?

leezu: @szha The overhead of triggering `__getitem__` and `__len__` instead of just flattening directly may be too large.

szha: OK

Issue #16942, opened 2019-11-29
Validation model can be different from fit model in estimator class


liuzh91:
## Description
In current estimator class implementation, validation model is the same network as the training model `self.net`. Code of `evaluate_batch()` is shown below:

```
        data, label = self._get_data_and_label(val_batch, self.context, batch_axis)
        pred = [self.net(x) for x in data]
        loss = [self.evaluation_loss(y_hat, y) for y_hat, y in zip(pred, label)]
```

This assumption does not hold true in the general case. It is common to have a different validation model than the training model on many tasks, e.g., the language model.

```
   model_eval = nlp.model.AWDRNN(args.model, len(vocab), args.emsize, args.nhid, args.nlayers,
                                  args.tied, args.dropout, args.weight_dropout,
                                  args.dropout_h, args.dropout_i, args.dropout_e)
   model = nlp.model.train.AWDRNN(args.model, len(vocab), args.emsize, args.nhid, args.nlayers,
                                   args.tied, args.dropout, args.weight_dropout,
                                   args.dropout_h, args.dropout_i, args.dropout_e)
```
Please refer to (https://github.com/dmlc/gluon-nlp/blob/master/scripts/language_model/word_language_model.py#L154) for a detailed reference.

PR #16957, opened 2019-12-02
Include eval_net the validation model in the estimator api

liuzh91:
## Description ##
We add validation model eval_net to the base estimator API. Users may use different models during training and validation. In this PR, we make the following contribution:

-  We add the implementation of eval_net in the estimator class
-  We add unit test for the introduced feature to ensure that parameter sharing works correctly for self.eval_net and self.net

Fixes issue #16942.


leezu:
CC @ptrendx part of the #16922, #16900 series. If feasible, let's also include it.

PR #16971, opened 2019-12-04
Extend estimator.evaluate() to support event handlers

### LINKED ISSUE #16959: Estimator.evaluate does not support logging
Gluon estimator does not log evaluation progress. It does not have the event_handler option. Shall we add it?

liuzh91:
## Description ##
Motivated by the issue #16959, we find that we could use existing `LoggingHandler` and `MetricHandler` for validation purpose. We extend `estimator.evaluate()` to support default and customized handlers. Our contributions can be summarized below:

- We extend estimator.evaluate() to support event_handlers
- We replace the metrics in estimator constructor with train_metrics and val_metrics
- We decouple the dependence of default event handlers on training or validation metrics 
- We add test script to verify the correctness of evaluate function 

Fix #16959.

### DIFF
--- a/python/mxnet/gluon/contrib/estimator/estimator.py
+++ b/python/mxnet/gluon/contrib/estimator/estimator.py
@@ -230,6 +230,8 @@ class Estimator(object):
             self._train_metrics.append(metric_loss(loss_name))
 
         for metric in self._train_metrics:
+            # add training prefix to the metric name
+            # it is useful for event handlers to distinguish them from validation metrics
             metric.name = 'training ' + metric.name
 
     def _add_validation_metrics(self):
@@ -237,6 +239,8 @@ class Estimator(object):
             self._val_metrics = [copy.deepcopy(metric) for metric in self._train_metrics]
 
         for metric in self._val_metrics:
+            # add validation prefix to the metric name
+            # it is useful for event handlers to distinguish them from training metrics
             if 'training' in metric.name:
                 metric.name = metric.name.replace('training', 'validation')
             else:
--- a/python/mxnet/gluon/contrib/estimator/estimator.py
+++ b/python/mxnet/gluon/contrib/estimator/estimator.py
@@ -256,7 +256,6 @@ class Estimator(object):
 
     def evaluate_batch(self,
                        val_batch,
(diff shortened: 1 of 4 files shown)


python/mxnet/gluon/contrib/estimator/event_handler.py
```diff
@@ -305,15 +293,22 @@ def batch_end(self, estimator, *args, **kwargs):
 
     def epoch_begin(self, estimator, *args, **kwargs):
         if self.verbose >= self.LOG_PER_EPOCH:
+            is_training = False
+            for metric in self.metrics:
+                if 'training' in metric.name:
```
leezu: Please add a comment and describe that this test always works thanks to the hack in `Estimator.__init__`

python/mxnet/gluon/contrib/estimator/estimator.py
```diff
@@ -202,13 +206,17 @@ def _add_default_training_metrics(self):
             self._train_metrics.append(metric_loss(loss_name))
 
         for metric in self._train_metrics:
-            metric.name = "training " + metric.name
+            metric.name = 'training ' + metric.name
```
leezu: Please add a comment describing that this is needed for `LoggingHandler` to improve the logging output and to detect if it is in training or evaluation mode.

liuzh91: I will add the comment here.

python/mxnet/gluon/contrib/estimator/event_handler.py
liuzh91: I'll comment this.

python/mxnet/gluon/contrib/estimator/estimator.py
liuzh91: The naming is not exactly hack. It already exists in the code for displaying name during logging. Anyway I have added comments to resolve the issue.

ptrendx:
@leezu - do you also want this in 1.6? I'm cherry-picking the other stuff, might as well take this since it is merged.

leezu:
@ptrendx, yes, that would be great

PR #16973, opened 2019-12-04
Fix ci cuda problems

larroy:
## Description ##

update CUDNN from 7.5.1.10 to 7.6.0.64 in ubuntu_gpu_cu101

### DIFF
--- a/src/operator/cross_device_copy.cc
+++ b/src/operator/cross_device_copy.cc
@@ -60,6 +60,16 @@ class CrossDeviceCopyProp : public OperatorProperty {
     return true;
   }
 
+  bool InferType(std::vector<int> *in_type,
+                 std::vector<int> *out_type,
+                 std::vector<int> *aux_type) const {
+    CHECK_EQ(in_type->size(), 1) << "Input:[data]";
+    if (in_type->at(0) == -1) return false;
+    out_type->clear();
+    out_type->push_back(in_type->at(0));
+    return true;
+  }
+
   OperatorProperty* Copy() const override {
     return new CrossDeviceCopyProp();
   }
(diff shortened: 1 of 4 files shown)


marcoabreu:
Guys I feel like we have half a dozen PRs for the same thing now. Who owns this? Can we please consolidate these efforts? Everyone states that they fixed something, but nothing is merged or seems to be tested. What's going on here?

larroy:
We are trying to fix it. Opening PRs to see results and compare. Not sure what's your point. We are also using the dev CI for that matter, but it's a complex interaction of the environment and the software inside the nvidia docker container.

tests/python/unittest/test_model_parallel.py
```diff
@@ -17,6 +17,7 @@
 
 import numpy as np
 import mxnet as mx
+from mxnet.test_utils import *
```
apeforest: What is this change for?

src/operator/cross_device_copy.cc
```diff
@@ -60,6 +60,16 @@ class CrossDeviceCopyProp : public OperatorProperty {
     return true;
   }
 
+  bool InferType(std::vector<int> *in_type,
```
apeforest: How would this fix CI?

tests/python/unittest/test_model_parallel.py
larroy: This is not needed, thanks.

src/operator/cross_device_copy.cc
larroy: Looks like I pulled an additional commit while rebasing, seems obvious.

Issue #16789, opened 2019-11-12
[Numpy] Transpose does not support kAddTo and kNullOp


sxjscience:
https://github.com/apache/incubator-mxnet/blob/02f4f0503de60cce41989f1bb779ee8c7b943e4b/src/operator/numpy/np_matrix_op-inl.h#L122

Also, it does not support kNullOp, which has been supported in https://github.com/apache/incubator-mxnet/commit/56999399a44467cbd1d0fbdf9cbba8b46ac53956#diff-d5c9c7d539907caef01a08a74e16592b

sxjscience:
@leezu

leezu:
Let's consider backporting the fix to `mx.nd.Transpose`.

FYI @zburning

samskalicky:
@lanking520 assign @ChaiBapchya

PR #16979, opened 2019-12-05
[Bugfix] [Numpy] Add `kAddTo` and kNullOp to Transpose

sxjscience:
## Description ##

- Add kAddTo and kNullOp to np.transpose and fix https://github.com/apache/incubator-mxnet/issues/16789
- Fix the problem that the previous np.tranpose support duplicated axis:

Previously, we can run the following code without any problem:
```python
import mxnet as mx
mx.npx.set_np()
a = mx.np.ones((3, 4, 5))
b = mx.np.transpose(a, axes=(0, 0, 1))
print(b.asnumpy())
```

Now, we will raise an MXNetError

### Essentials ###

### Changes ###
- [x] kAddTo and kNullOp to transpose, test


sxjscience:
@leezu @zburning This should fix the XLNet problem.

src/operator/tensor/matrix_op.cc
```diff
@@ -283,11 +283,12 @@ static void TransposeComputeExCPU(const nnvm::NodeAttrs& attrs,
...
   }
   const TransposeParam& param = nnvm::get<TransposeParam>(attrs.parsed);
-  CHECK_EQ(req[0], kWriteTo) << "Transpose does not support kWriteInplace and kAddTo";
+  CHECK(req[0] == kWriteTo || req[0] == kAddTo) <<
+      "Transpose only supports kNullOp, kWriteTo and kAddTo";
   CHECK_EQ(inputs.size(), 1U);
   CHECK_EQ(outputs.size(), 1U);
 
-  if (SupportMKLDNNTranspose(param, inputs[0])) {
+  if (SupportMKLDNNTranspose(param, inputs[0]) && req[0] == kWriteTo) {
```
TaoLv: Thanks!

src/operator/tensor/pseudo2DTranspose_op-inl.cuh
```diff
@@ -39,22 +39,29 @@ namespace mxnet {
 namespace op {
 namespace cuda {
 
-
-template <typename DType, typename CType>
+/*!
+ * \brief The `transpose_pseudo2D` based on chosen vectorized types. It transpose an array of
```
sxjscience: @ptrendx I've also added the doc here for the `transpose_pseudo2D`. Correct me if you find any problem.

src/operator/tensor/pseudo2DTranspose_op-inl.cuh
```diff
@@ -39,22 +39,29 @@ namespace mxnet {
...
+ * \tparam DType Data type
+ * \tparam CType The type to load the data.
+ * \tparam TSR the vectorized ratio.
+ * \tparam is_addto Whether to perform out += transpose(data) or out = transpose(data)
+ */
+template <typename DType, typename CType, int TSR, bool is_addto>
 __global__ void transpose_pseudo2D(DType* out, DType* inp,
                                    const index_t m, const index_t n,
                                    const index_t nIterY, const index_t nIterZ) {
-  const index_t TSR = sizeof(CType)/sizeof(DType);  // TypeSizeRatio
```
ptrendx: Why did you move this to template argument? It should be fine here (you can even make it `constexpr`).

src/operator/tensor/pseudo2DTranspose_op-inl.cuh
```diff
@@ -78,23 +85,34 @@ __global__ void transpose_pseudo2D(DType* out, DType* inp,
         }
         __syncthreads();
 
-        // read from shared to registers
-        transp_t tmp[TSR];
+        // read from shared to local registers
+        CType tmp[TSR];
         #pragma unroll
         for (index_t i = 0; i < TSR; i++) {
+          DType* tmp_dptr = reinterpret_cast<DType*>(&tmp[i]);
```
ptrendx: Hmmm, I'm not sure if this will still be handled properly by the compiler, let me test that.

src/operator/tensor/pseudo2DTranspose_op-inl.cuh
sxjscience: I moved it to template to avoid the divide by zero error. Because in the other loop we are using switch to switch over the size of DType.

sxjscience: Reports a `use too much shared memory error` after I change this. May be it's feasible to first put it into the template

sxjscience: Okay, I think I know where the problem is. Let me try another time.

src/operator/tensor/pseudo2DTranspose_op-inl.cuh
sxjscience: Have you find any problem? I can compile that and it also passed CI.

src/operator/tensor/matrix_op-inl.h
```diff
@@ -323,62 +336,77 @@ void TransposeImpl(RunContext ctx,
...
+      Tensor<xpu, 1, DType> in = src.get_with_shape<xpu, 1, DType>(mshadow::Shape1(src.Size()), s);
+      Tensor<xpu, 1, DType> out = ret.get_with_shape<xpu, 1, DType>(mshadow::Shape1(ret.Size()), s);
+      if (!is_addto) {
+        // Use memcpy to accelerate the speed
+        Copy(out, in, s);
+      } else {
+        mxnet_op::Kernel<mxnet_op::op_with_req<mshadow_op::identity, kAddTo>, xpu>::Launch(
+            s, ret.Size(), out.dptr_, in.dptr_);
+      }
+      return;
```
reminisce: `return` can be moved out of the macro to simplify expanded code a little bit.

src/operator/tensor/matrix_op-inl.h
```diff
@@ -323,62 +336,77 @@ void TransposeImpl(RunContext ctx,
...
-        } else {
-          out = in.T();
-        }
+      Tensor<xpu, 2, DType> in = src.get<xpu, 2, DType>(s);
+      Tensor<xpu, 2, DType> out = ret.get<xpu, 2, DType>(s);
+      if (ctx.get_ctx().dev_mask() == cpu::kDevMask) {
+        Transpose2D<DType, is_addto>(in.dptr_, out.dptr_, in.shape_[0], in.shape_[1]);
       } else {
-        Copy(out, in, s);
+        LOG(FATAL) << "Not Implemented. We should never reach here. Report an issue in Github.";
```
reminisce: Can we improve the error message by pointing out the case should have been handled by `transpose_pseudo2D`?

src/operator/tensor/matrix_op-inl.h
```diff
@@ -323,62 +336,77 @@ void TransposeImpl(RunContext ctx,
...
   if (isPseudo2DTranspose(axes)) {
     MSHADOW_TYPE_SWITCH(ret.type_flag_, DType, {
-      transpose_pseudo2D<DType>(ret, src, axes, s);
+      transpose_pseudo2D<DType, is_addto>(ret, src, axes, s);
     });
     return;
   }
 #endif
+  // Special handle the identity case
+  if (isIdentityTranspose(axes)) {
```
reminisce: `isIdentityTranspose` -> `IsIdentityTranspose`.

tests/python/unittest/test_numpy_op.py
```diff
@@ -1313,45 +1315,56 @@ def __init__(self, axes=None):
 
         def hybrid_forward(self, F, a):
             return F.np.transpose(a, self.axes)
+    test_workloads = [[(), [(), None]],
```
reminisce: Seems `ndim=1` is not tested?

src/operator/tensor/matrix_op-inl.h
sxjscience: moved

src/operator/tensor/pseudo2DTranspose_op-inl.cuh
ptrendx: Hi, I only now had a chance to look into it. No, there is not any problem - I was worried that the compiler could get confused by this and put the `tmp` array in local memory instead of registers, but I tested and it does not do it.

sxjscience: Great! Thanks for clarifying 👍

src/operator/tensor/matrix_op-inl.h
```diff
@@ -323,62 +336,79 @@ void TransposeImpl(RunContext ctx,
...
-     }
      case 2: {
-      mshadow::Tensor<xpu, 2, DType> in = src.FlatTo2D<xpu, DType>(s);
-      mshadow::Tensor<xpu, 2, DType> out = ret.FlatTo2D<xpu, DType>(s);
-
-      if (axes[0] == 1 && axes[1] == 0) {
-        if (ctx.get_ctx().dev_mask() == cpu::kDevMask) {
-          Transpose2D<DType>(in.dptr_, out.dptr_, in.shape_[0], in.shape_[1]);
-        } else {
-          out = in.T();
```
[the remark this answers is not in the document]
sxjscience: @ChaiBapchya We have already had `transpose_pseudo2D`, which covers the case of 2D Transpose in GPU.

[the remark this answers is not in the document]
ChaiBapchya: Ah right..

TaoLv:
@sxjscience Is it possible to create a case that Transpose has req kAddTo? I'm considering how to support that in MKL-DNN backend. Thanks!

ptrendx:
@TaoLv just transpose the weight before using it (e.g. take nchw weight, transpose and use in nhwc convolution), and set grad_req as 'add' in simple_bind (gluon example would be pretty similar).

PR #16980, opened 2019-12-05
change windows build system

yajiedesign:
## Description ##
now, the huge image size causes many troubles.etc large image can't laod. some compile error.
now, the PIP build use compressed fatbin by hack nvcc,but it will cause a lot of memory,and general users have difficulty compiling on their own.
this pr change build system,it spilt different cuda arch to different dll.etc mxnet_30.mxnet_50.Then use a warpdll to automatically identify the gpu sm version. Load the corresponding dll. And forward all calls.
this pr also contains tools to automatically generate warp dll and will be executed automatically during compilation.


tools/gen_warp.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include <fstream>
+
+#define  IMAGE_SIZEOF_SIGNATURE 4
+
+
+DWORD rva_to_foa(IN DWORD RVA, IN PIMAGE_SECTION_HEADER section_header) 
+{
+
+	size_t count = 0;
+	//���ж����ĸ�����
```
lanking520: Need a translation

CMakeLists.txt
```diff
@@ -18,61 +19,84 @@ endif()
...
-mxnet_option(USE_CUDA             "Build with CUDA support"   ON)
-mxnet_option(USE_OLDCMAKECUDA     "Build with old cmake cuda" OFF)
-mxnet_option(USE_NCCL             "Use NVidia NCCL with CUDA" OFF)
-mxnet_option(USE_OPENCV           "Build with OpenCV support" ON)
-mxnet_option(USE_OPENMP           "Build with Openmp support" ON)
-mxnet_option(USE_CUDNN            "Build with cudnn support"  ON) # one could set CUDNN_ROOT for search path
-mxnet_option(USE_SSE              "Build with x86 SSE instruction support" ON IF NOT ARM)
-mxnet_option(USE_F16C             "Build with x86 F16C instruction support" ON) # autodetects support if ON
-mxnet_option(USE_LAPACK           "Build with lapack support" ON)
+# Some things have order. This must be put in front alone
```
apeforest: Can you resume the earlier alignment? This is harder to read.

CMakeLists.txt
```diff
@@ -18,61 +19,84 @@ endif()
...
+  USE_MKL_IF_AVAILABLE
+  AND
+  (NOT APPLE)
+  AND
+  (NOT MSVC)
+  AND
+  (CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "x86_64")
+  AND
+  (NOT CMAKE_CROSSCOMPILING))
+mxnet_option(USE_OPERATOR_TUNING "Enable auto-tuning of operators" ON IF NOT
```
apeforest: Can you resume the earlier alignment.

CMakeLists.txt
```diff
@@ -18,61 +19,84 @@ endif()
...
-mxnet_option(USE_TVM_OP           "Enable use of TVM operator build system." OFF)
-mxnet_option(ENABLE_CUDA_RTC      "Build with CUDA runtime compilation support" ON)
-mxnet_option(BUILD_CPP_EXAMPLES   "Build cpp examples" ON)
-mxnet_option(INSTALL_EXAMPLES     "Install the example source files." OFF)
-mxnet_option(USE_SIGNAL_HANDLER   "Print stack traces on segfaults." ON)
-mxnet_option(USE_TENSORRT         "Enable inference optimization with TensorRT." OFF)
-mxnet_option(USE_ASAN             "Enable Clang/GCC ASAN sanitizers." OFF)
-mxnet_option(ENABLE_TESTCOVERAGE  "Enable compilation with test coverage metric output" OFF)
-mxnet_option(USE_INT64_TENSOR_SIZE "Use int64_t to represent the total number of elements in a tensor" OFF)
+mxnet_option(USE_GPROF "Compile with gprof (profiling) flag" OFF)
```
apeforest: Same here

yajiedesign:
every body,ci test is ok .

CMakeLists.txt
```diff
@@ -619,17 +624,17 @@ endif()
 
 if(USE_CUDA)
   if(FIRST_CUDA)
-    mshadow_select_nvcc_arch_flags(NVCC_FLAGS_ARCH)
+    mshadow_select_nvcc_arch_flags(NVCC_FLAGS_ARCH cuda_arch)
     string(REPLACE ";" " " NVCC_FLAGS_ARCH "${NVCC_FLAGS_ARCH}")
-    set(CMAKE_CUDA_FLAGS "${NVCC_FLAGS_ARCH}")
+    # set(CMAKE_CUDA_FLAGS "${NVCC_FLAGS_ARCH}")
```
leezu: ?

ci/build_windows.py
```diff
@@ -115,6 +117,7 @@ class BuildFlavour(Enum):
         '-DCUDA_ARCH_NAME=Manual '
         '-DCUDA_ARCH_BIN=52 '
         '-DCUDA_ARCH_PTX=52 '
+        '-DCUDA_ARCH_LIST=5.2 '
```
leezu: Why add `CUDA_ARCH_LIST`?

yajiedesign: the first class cuda use CUDA_ARCH_LIST choose or set arch

leezu: I don't think it works currently. I got this warning when testing it:

```
Manually-specified variables were not used by the project:

    CUDA_ARCH_LIST
```

It is fixed as part of https://github.com/apache/incubator-mxnet/pull/17031

yajiedesign: currently works  fine on ci.The warning may be caused by other modifications.
image

leezu: I just tried it again. It only works with `CUDA_ARCH_LIST=5.2` but not with `CUDA_ARCH_LIST=7.0`. It seems somewhere an outdated arch detection script is used.

yajiedesign:
@leezu now imageSize is too large. Windows cannot load too large DLL.So we must try to reduce the imagesize size.
previously I reduced the size by modifying nvcc.But this method takes up too much memory when starting multiple processes.So I designed a new scheme

yajiedesign:
@leezu The pip size will be solved by other methods.

leezu:
@yajiedesign do you mean on Windows 32 bit? Do we have any need to support 32 bit build?

leezu:
Regarding the memory size, in my understanding if you use multiple processes, that all load the mxnet dll, the memory among them will be shared.
Or is there some problem that prevents the memory reuse?

yajiedesign:
@leezu the previously reduced size method broken dll shared.

yajiedesign:
@leezu not 32 bit. 64bit dll are also image size limit.This is determined by PE file structure.Reference https://stackoverflow.com/questions/6976693/what-is-the-maximum-size-of-a-pe-file-on-64-bit-windows

yajiedesign:
@leezu Pip is mentioned here because the old method is use to PIP build,

ci/build_windows.py
yajiedesign: yes.only in https://github.com/apache/incubator-mxnet/pull/16980/files#diff-101d74caab6e808c727c01a427bd927eR137

leezu:
> @leezu not 32 bit. 64bit dll are also image size limit.This is determined by PE file structure.Reference https://stackoverflow.com/questions/6976693/what-is-the-maximum-size-of-a-pe-file-on-64-bit-windows

Thanks @yajiedesign. It seems one of the root causes is that Windows does not support position independent code and PE format must always use 32bit relative addressing.

The reason I ask you so many questions is that a) I'm concerned that not many people are familiar with Windows, so we need to document the reasoning that went into this patch b) fundamentally the problem you attempt to solve here also affects unix systems and it would be good to have a cross-platform solution. (Thanks to ELF superiority over Windows PE, I understand that the problem is more pressing on Windows and we may need to have a Windows specific intermediate solution at first)

What do you think?

Maybe @samskalicky's dynamic loading of operators effort can solve this in a cross-platform way?

yajiedesign:
Yes, it is urgent under windows. The current scheme is private and has many problems
This solution may be portable to Linux. But I don't know it is need under Linux.I'm not familiar with linux.
I think merge it first to make windows available, and then consider cross platform issues.

leezu:
@yajiedesign can you provide a link for the current scheme? Or is the code not available at all? Where does it run?

Which scheme do you mean? Do you mean the code used to build the windows binary at https://pypi.org/project/mxnet-cu101/1.6.0b20190926/#files

How large is the binary without the current private scheme? What happens without the private scheme?

yajiedesign:
@leezu not current scheme link.it only on my build machine. It's complicated and dirty.
yes current scheme used to build pip pkg.
without current scheme .image size about 1.8gb.it can't load.

leezu:
Thanks for providing more details.

I think it would be good to move the files introduced here from `tools` to `tools/windows` or `tools/windowsbuild` and add a `README.md` there which documents the motivation (ie. the information you provided during the discussion). Then it should be fine to merge.

You can also consider making the warping feature optional in the Windows build, as it adds overhead and is not needed apart from the pip build that needs to include all cuda arches.

CMakeLists.txt
```diff
@@ -60,6 +60,8 @@ option(USE_ASAN "Enable Clang/GCC ASAN sanitizers." OFF)
 option(ENABLE_TESTCOVERAGE "Enable compilation with test coverage metric output" OFF)
 option(USE_INT64_TENSOR_SIZE "Use int64_t to represent the total number of elements in a tensor" OFF)
 option(BUILD_CYTHON_MODULES "Build cython modules." OFF)
+option(USE_SPLIT_ARCH_DLL   "Build mxnet with spilt different dll every cuda arch in windows." ON)
```
leezu: It's only valid on Windows, so how about hiding the option via `cmake_dependent_option` on other systems.

`cmake_dependent_option(USE_SPLIT_ARCH_DLL "Build a separate DLL for each Cuda arch (Windows only)." ON "NOT MSVC" OFF)`

yajiedesign:
ok.

yajiedesign:
ci test ok

tools/windowsbuild/warp_dll.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include <iostream>
+#include <cuda_runtime.h>
+#include <algorithm>
+#include <Windows.h>
+
+
+
+int find_version()
+{
+	int known_sm[] = { 30,35,37,50,52,60,61,70,75 };
```
marcoabreu: Can you make that dynamic so we don't have to update it in the future?

yajiedesign:
@marcoabreu  done

yajiedesign:
can we merge?

yajiedesign:
ci test ok.

CMakeLists.txt
```diff
@@ -192,9 +195,11 @@ else()
     add_definitions(-DDMLC_USE_CXX11=1)
     add_definitions(-DDMLC_USE_CXX14=1)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
+    set(CMAKE_CUDA_STANDARD 14)
```
leezu: It's already set in line 93 https://github.com/apache/incubator-mxnet/blob/ee31ae54bf7099c6ceda4a4bd1d7ccd8c431dea9/CMakeLists.txt#L93

CMakeLists.txt
```diff
@@ -192,9 +195,11 @@ else()
...
   elseif(SUPPORT_CXX11)
     add_definitions(-DDMLC_USE_CXX11=1)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+    set(CMAKE_CUDA_STANDARD 11)
```
leezu: It's already set in line 88
https://github.com/apache/incubator-mxnet/blob/ee31ae54bf7099c6ceda4a4bd1d7ccd8c431dea9/CMakeLists.txt#L88

yajiedesign:
ping

ptrendx:
@yajiedesign Could you make a PR with backported changes needed for Windows build to 1.6 branch as you mentioned in the mail to dev@?

yajiedesign:
@ptrendx ok

lordofgod:
@yajiedesign 

这个issue的修改可能会导致cpp_package的OpWrapperGenerator无法正常生成op.h文件, 错误码127。原因是warp_dll.cpp文件的119行HMODULE hm = LoadLibraryW(dll_name);使用了相对路径。调用OpWrapperGenerator时的dll搜索路径不是libmxnet.dll的所在目录，导致这段代码找不到依赖的dll库。这里是不是使用绝对路径更好一些。

This pr will cause the "OpwrapperGenerator.py" failed to generate "op.h" with error code 127, while building with cpp_package. The reason is use of relative dll path at "HMODULE hm = LoadLibraryW(dll_name);"  in the new added file warp_dll.cpp, Line 119. "OpWrapperGenerator.py" 's path is different from libmxnet.dll, thus may cause the 127 error while calling the generator script, which infers that sub-dll's load failure in libmxnet.dll. If it's better to use absolute path here, or other strategy for solving this.

dll search path reference: https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order

Issue #16930, opened 2019-11-28
Plug and Play fit_batch() for the estimator class


liuzh91:
## Description
In the current estimator implementation, fit_batch() is a class method of the estimator class. A common workflow of fit_batch() is that the model `self.net` forwards the training batch to generate outputs and compute loss functions. The problem is that such design is not flexible enough with different model forward interfaces on the same task. For example, fit_batch() of the base estimator trains the current batch on the label prediction task:
```
        with autograd.record():
            pred = [self.net(x) for x in data]
            loss = [self.loss(y_hat, y) for y_hat, y in zip(pred, label)]
```
In the above example, the model forward interface of `self.net` is `def forward(self, inputs)` with the return value of predict labels.  The estimator is compatible with any model using this forward interface. However, if we have another model for the label prediction task with a different forward interface `def forward(self, inputs, input_length)`, the base estimator is not compatible with this model even though both models share the same loss functions, training and evaluation metrics. A real world example can be found at LM models (https://github.com/dmlc/gluon-nlp/blob/c03665bafb1e0fe0fa5c2a59bbb4845393fbf9ba/src/gluonnlp/model/train/language_model.py). `StandardRNN` and `AWDRNN` shares the same forward interface, whereas `BigRNN` has a different one.

Forward interface of `StandardRNN` and `AWDRNN`:
```
def __call__(self, inputs, begin_state=None):
```
Forward interface of `BigRNN`:
```
def forward(self, inputs, label, begin_state, sampled_values):
```

A straightforward workaround is to create a new customized estimator for each model interface. It will bring the issue that we need to create a standalone estimator each time we see a new model interface even on the same task. In machine learning community, it is common to see different model forward logic on the same task. This approach will leads to prohibitively many estimators for some simple task. In the above LM example, we need to create a `vanillaRNNEstimator` and a `BigRNNEstimator` even most of the training logic between these two estimators are the same.

To prevent the above estimator explosion issue, we suggest adding support of a plug and play customized `fit_batch()` which is similar to the `event_handlers` for the estimator class. Given an existing estimator `est`,  we modify the `fit()` method to take an extra argument of `fit_batch_handler`. So we can call `est.fit(train_data=data_loader, epochs=epochs, fit_batch_handler=fit_StandardRNN_batch)` or `est.fit(train_data=data_loader, epochs=epochs, fit_batch_handler=fit_BigRNN_batch)` to use models with different interface.
[...]

PR #16678, opened 2019-10-31
[Estimator] refactor estimator to allow overriding evaluate/fit of a batch

szha:
## Description ##
refactor estimator to allow overriding evaluate/fit of a batch

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] refactor estimator with `evaluate_batch` and `fit_batch`.

PR #16982, opened 2019-12-05
Add support of plug and play fit_batch and evaluate_batch

liuzh91:
## Description ##
Due to the potential estimator explosion issue introduced in issue #16930, we add support of plug and play `fit_batch()` and `evaluate_batch()` methods for the base estimator class. Concretely,
we define a new class `BatchProcessor` and use it to wrap the `fit_batch()` and `evaluate_batch()` methods. Our contribution can be summarized as below

- Add a new class `BatchProcessor` to support plug and play `fit_batch` and `evaluate_batch` for existing estimators.
- Add test case to test the plug and play `fit_batch()` and `evaluate_batch()` methods

Fix Issue #16930

### DIFF
--- a/python/mxnet/gluon/contrib/estimator/batch_processor.py
+++ b/python/mxnet/gluon/contrib/estimator/batch_processor.py
@@ -20,7 +20,6 @@
 """Gluon Batch Processor for Estimators"""
 
 from ...utils import split_and_load
-from ....metric import Loss as metric_loss
 from .... import autograd
 
 __all__ = ['BatchProcessor']
@@ -96,7 +95,6 @@ class BatchProcessor(object):
         """
         data, label = self._get_data_and_label(train_batch, estimator.context, batch_axis)
 
-        batch_size = train_batch[0].shape[batch_axis]
         with autograd.record():
             pred = [estimator.net(x) for x in data]
             loss = [estimator.loss(y_hat, y) for y_hat, y in zip(pred, label)]
--- a/python/mxnet/gluon/contrib/estimator/batch_processor.py
+++ b/python/mxnet/gluon/contrib/estimator/batch_processor.py
@@ -61,7 +61,7 @@ class BatchProcessor(object):
             Batch axis to split the validation data into devices.
         """
         data, label = self._get_data_and_label(val_batch, estimator.context, batch_axis)
-        pred = [estimator.net(x) for x in data]
+        pred = [estimator.eval_net(x) for x in data]
         loss = [estimator.evaluation_loss(y_hat, y) for y_hat, y in zip(pred, label)]
 
         return data, label, pred, loss
--- a/python/mxnet/gluon/contrib/estimator/batch_processor.py
+++ b/python/mxnet/gluon/contrib/estimator/batch_processor.py
@@ -17,20 +17,18 @@
 
 # coding: utf-8
 # pylint: disable=wildcard-import, unused-argument, too-many-ancestors
-"""Gluon EstimatorModels for Estimators"""
-
-import warnings
+"""Gluon Batch Processor for Estimators"""
 
 from ...utils import split_and_load
 from ....metric import Loss as metric_loss
 from .... import autograd
 
-__all__ = ['EstimatorModel']
+__all__ = ['BatchProcessor']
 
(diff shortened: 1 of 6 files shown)


python/mxnet/gluon/contrib/estimator/estimator.py
```diff
@@ -269,9 +271,14 @@ def evaluate(self,
 
         for metric in val_metrics:
             metric.reset()
+        estimator_ref = self
 
         for _, batch in enumerate(val_data):
-            self.evaluate_batch(batch, val_metrics, batch_axis)
+            if self.estimator_model is None:
+                self.evaluate_batch(batch, val_metrics, batch_axis)
```
leezu: This line provides backwards compatibility to the unreleased API introduced in https://github.com/apache/incubator-mxnet/pull/16678/ . As that API has not been released yet, we don't need to provide compatibility for it.
Instead it may better to keep the code simple by removing `evaluate_batch` and `fit_batch` from the `Estimator` and only using the plug-and-play version. What do you think?

python/mxnet/gluon/contrib/estimator/estimator.py
```diff
@@ -383,8 +390,11 @@ def fit(self, train_data,
                 for handler in batch_begin:
                     handler.batch_begin(estimator_ref, batch=batch)
 
-                _, label, pred, loss = self.fit_batch(batch, batch_axis)
-
+                if self.estimator_model is None:
+                    _, label, pred, loss = self.fit_batch(batch, batch_axis)
```
leezu: Same as above

python/mxnet/gluon/contrib/estimator/estimator_model.py
```diff
@@ -0,0 +1,112 @@
...
+from ...utils import split_and_load
+from ....metric import Loss as metric_loss
+from .... import autograd
+
+__all__ = ['EstimatorModel']
+
+class EstimatorModel(object):
+    """EstimatorModel Class for plug and play fit_batch & evaluate_batch
+
+    :py:class:`EstimatorModel` can be used to replace fit_batch() and evaluate_batch()
```
leezu: You may want to provide more documentation here?

python/mxnet/gluon/contrib/estimator/estimator.py
liuzh91: I think wrapping the `fit_batch` and `evaluate_batch` in one class is better. But I will brainstorm a better class name for the two methods.

python/mxnet/gluon/contrib/estimator/estimator.py
liuzh91: I will brainstorm a better name.

python/mxnet/gluon/contrib/estimator/estimator_model.py
liuzh91: OK, I'll add more explanation later.

python/mxnet/gluon/contrib/estimator/estimator.py
leezu: Agree, but the point here is about the `Estimator.fit_batch` and `Estimator.evaluate_batch`. Currently there's a code duplication with `EstimatorModel.fit_batch` etc

liuzh91: I have removed the duplicate code for `fit_batch` and `evaluate_batch`.

python/mxnet/gluon/contrib/estimator/batch_processor.py
```diff
@@ -27,6 +27,10 @@
 
 class BatchProcessor(object):
     """BatchProcessor Class for plug and play fit_batch & evaluate_batch
+    During training or validation, data are divided into minibatches for processing. This
```
leezu: Missing newline https://www.python.org/dev/peps/pep-0257/

liuzh91: Problem solved.

leezu:
@ptrendx, let's also consider this one if it's merged to master in time (expected to be merged in 12 hours after fixing the conflict)

PR #16997, opened 2019-12-06
Fix pip setup.py for windows.

frankfliu:
## Description ##
1. Add build windows pip wheel feature. 
2. Add cu101
3. Fix setup bugs.

### Essentials ###


tools/pip/setup.py
```diff
@@ -28,6 +28,9 @@
 if platform.system() == 'Linux':
     sys.argv.append('--universal')
     sys.argv.append('--plat-name=manylinux1_x86_64')
+if platform.system() == 'Windows':
+    sys.argv.append('--universal')
```
leezu: Do you have some more insights into why this works?
It's not supported by wheel https://packaging.python.org/guides/distributing-packages-using-setuptools/#universal-wheels

frankfliu: @yajiedesign may provide better answer.

@leezu --universal mean can run on both p2 and p3. I don't think we have C extensions. We only use ctype to bind to a share library. Linux pip setup also use "--universal".

yajiedesign: @leezu because we don't compile C extensions through python.We compile is self and call via ctypes use it.for python this is a pure python package.

leezu: We do have Cython for C extensions. If we don't use Cython for the pypi build, I agree with you.

yajiedesign: @leezu windows pip not use Cython.I don't know about Linux,

frankfliu:

tools/pip/setup.py
```diff
@@ -148,20 +153,46 @@ def has_ext_modules(self):
...
     shutil.copy(os.path.join(os.path.dirname(LIB_PATH[0]), 'libquadmath.so.0'), os.path.join(CURRENT_DIR, 'mxnet'))
     package_data['mxnet'].append('mxnet/libquadmath.so.0')
+elif platform.system() == 'Windows':
+    package_data['mxnet'].append(os.path.abspath(os.path.join('mxnet', os.path.basename(LIB_PATH[0]))))
+    package_data['mxnet'].append(os.path.abspath(os.path.join('mxnet', 'libmxnet.lib')))
+    for root, dirs, files in os.walk('win_third/openblas'):
+        for file in files:
+            shutil.copy(os.path.join(root, file), os.path.join(CURRENT_DIR, 'mxnet'))
+            package_data['mxnet'].append(os.path.abspath(os.path.join('mxnet', file)))
+    if variant != 'CPU':
```
larroy: could we wrap this in a function for readability and maint.?

frankfliu: I actually don't think wrap with a function will improve readability in this case.

The code added is party of if...elif.. condition, move it into function make it harder to compare difference between Linux and Windows.

PR #15762, opened 2019-08-06
Refactor LibraryInitializer so it's thread safe. Fixes random sporadical concurrency crashes.

larroy:
## Description ##

This PR improves initialization of the Library which uses unsafe passing of environment variables to child processes through pthread_atfork handlers. Calling setenv concurrently with getenv leads to random crashes which have been reported in several ocassions but so far haven't been fixed.

This patch avoids using getenv and setenv and it should address those random crashes while improving handling of the library intialization in the LibraryInitializer object.

Fixes #13438
Fixes #14979

Fixes minor issues with #15755 

Aggregated dynamic library code in LibraryInitializer.
Verified Dtor is called on exit.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- Instead of changing OMP_NUM_THREADS it directly set_thread_max to 1, even though OMP is disabled.
- LibraryInitializer is run on (library) construction instead of static initialization so it's run before any other static initializer.
- Improve documentation of concurrency environment variables.
- Moved current_process_id to common

PR #16999, opened 2019-12-07
Prevent after-fork number of OMP threads being bigger than 1.

larroy:
## Description ##

This could happen if it was set in the environment. As we are setting engine::OpenMP::Get()->set_enabled(false) in initialize.cc in the child after forking, the behaviour goes back to what it was before #15762 was introduced.

Regions using omp get the threads count from GetRecommendedOMPThreadCount, so if omp is disabled they will get 1 thread and run serially

@anirudh2290  @apeforest

### DIFF
--- a/src/engine/openmp.cc
+++ b/src/engine/openmp.cc
@@ -84,6 +84,7 @@ void OpenMP::set_reserve_cores(int cores) {
 int OpenMP::GetRecommendedOMPThreadCount(bool exclude_reserved) const {
 #ifdef _OPENMP
   if (enabled_) {
+    // OMP_NUM_THREADS was set in the environment at the time of static initialization
     if (omp_num_threads_set_in_environment_) {
       return omp_get_max_threads();
     }
--- a/src/engine/openmp.cc
+++ b/src/engine/openmp.cc
@@ -83,10 +83,10 @@ void OpenMP::set_reserve_cores(int cores) {
 
 int OpenMP::GetRecommendedOMPThreadCount(bool exclude_reserved) const {
 #ifdef _OPENMP
-  if (omp_num_threads_set_in_environment_) {
-    return omp_get_max_threads();
-  }
   if (enabled_) {
+    if (omp_num_threads_set_in_environment_) {
+      return omp_get_max_threads();
+    }
     int thread_count = omp_get_max_threads();
     if (exclude_reserved) {
       if (reserve_cores_ >= thread_count) {
@@ -100,8 +100,9 @@ int OpenMP::GetRecommendedOMPThreadCount(bool exclude_reserved) const {
       return thread_count;
     }
     return omp_thread_max_;
+  } else {
+    return 1;
   }
-  return 1;
 #else
   return 1;
 #endif
--- a/tests/python/unittest/test_engine.py
+++ b/tests/python/unittest/test_engine.py
@@ -17,6 +17,9 @@
 
 import nose
 import mxnet as mx
+import os
+import unittest
+from mxnet.test_utils import EnvManager
 
 def test_bulk():
     with mx.engine.bulk(10):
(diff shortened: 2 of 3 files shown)


src/engine/openmp.cc
```diff
@@ -83,10 +83,10 @@ void OpenMP::set_reserve_cores(int cores) {
 
 int OpenMP::GetRecommendedOMPThreadCount(bool exclude_reserved) const {
 #ifdef _OPENMP
-  if (omp_num_threads_set_in_environment_) {
-    return omp_get_max_threads();
-  }
   if (enabled_) {
+    if (omp_num_threads_set_in_environment_) {
+      return omp_get_max_threads();
+    }
```
anirudh2290: this is the correct behavior, but i wonder if this will cause performance issues, for certain scenarios where clients of this class depend on GetRecommendedOMPThreadCount() to return the right number even if set_enabled is false. I only see set_enabled used in initialize so i think it should be fine. Other (probably less riskier) option is to set omp_num_threads_set_in_environment_ to false in the child process, but since it is private it will require additional setter and getter.

tests/python/unittest/test_engine.py
```diff
@@ -30,6 +33,44 @@ def test_bulk():
...
+                assert cpid == pid
+                exit_status = status >> 8
+                assert exit_status == 0
+            except:
+                pass
+        else:
+            os.close(w)
+            rfd = os.fdopen(r, 'r')
+            rfd.read(1)
+            omp_max_threads = mx.base._LIB.omp_get_max_threads()
```
anirudh2290: for our case we need to test GetRecommendedThreadCount. in the parent it should be same as env variable, in child it should be 1. CPP test would be easier to write.

src/engine/openmp.cc
larroy: I don't see it as a big risk, I went through the usage of the method. Maybe I missed something but the only non omp usage seems this one according to my IDE.

https://github.com/apache/incubator-mxnet/blob/master/src/kvstore/kvstore_utils.cc#L38

tests/python/unittest/test_engine.py
larroy: C++ test is platform dependent though. There's no fork in windows. That's why I didn't bother with C++.  Do you see it as necessary for approval?

larroy: C++ test is platform dependent though. There's no fork in windows. That's why I didn't bother with C++. Do you see it as necessary for approval?

src/engine/openmp.cc
anirudh2290: okay i guess its fine. only other thing that worries me is the pragma omp parallel for used without any num_threads. I was thinking most should use GetRecommendedThreadCount, but there are a lot of places in code which use omp parallel for without a thread count. You mentioned that you called set_omp_num_threads in the child handler of pthread_atfork. didnt that help ?

tests/python/unittest/test_engine.py
anirudh2290: windows doesnt support pthread_atfork so it should be fine. I think some kind of test is necessary.

tests/python/unittest/test_engine.py
```diff
@@ -30,6 +33,44 @@ def test_bulk():
...
+    """
+    with EnvManager('OMP_NUM_THREADS', '42'):
+        r, w = os.pipe()
+        pid = os.fork()
+        if pid:
+            os.close(r)
+            wfd = os.fdopen(w, 'w')
+            wfd.write('a')
+            omp_max_threads = mx.base._LIB.omp_get_max_threads()
+            print("Parent omp max threads: {}".format(omp_max_threads))
```
apeforest: Please remove print statement from test.

tests/python/unittest/test_engine.py
```diff
@@ -30,6 +33,44 @@ def test_bulk():
...
+                exit_status = status >> 8
+                assert exit_status == 0
+            except:
+                pass
+        else:
+            os.close(w)
+            rfd = os.fdopen(r, 'r')
+            rfd.read(1)
+            omp_max_threads = mx.base._LIB.omp_get_max_threads()
+            print("Child omp max threads: {}".format(omp_max_threads))
```
apeforest: Please remove print statement from test.

src/engine/openmp.cc
```diff
@@ -83,10 +83,10 @@ void OpenMP::set_reserve_cores(int cores) {
 
 int OpenMP::GetRecommendedOMPThreadCount(bool exclude_reserved) const {
 #ifdef _OPENMP
-  if (omp_num_threads_set_in_environment_) {
-    return omp_get_max_threads();
-  }
   if (enabled_) {
+    if (omp_num_threads_set_in_environment_) {
```
apeforest: Can we add some comments on under which scenario this would happen? The logic in this class has become too complicated to follow.

tests/python/unittest/test_engine.py
larroy: They are not a problem, and not printed to the console by nosetests. This is a non-issue.

src/engine/openmp.cc
larroy: sure, makes sense.

larroy:
@szha no. What behaviour do you suggest would be reasonable under a notebook? Leave omp threads set to 1 or to the generous value. Is there a way to detect the runtime is Jupyter?

leezu:
@szha do you have any reference on this? If you start python in your shell, the process is also forked. But as `exec` is executed after the fork, there are no problems.

src/engine/openmp.cc
larroy: looks that's implementation dependent. We should test more, but I think let's get this done first.

tests/python/unittest/test_engine.py
larroy: Added cpp test.

tests/python/unittest/test_engine.py
larroy: let's leave it for now, as nosetest adsorbs stdout output anyway. So basically unit test is the only place where prints are acceptable.

src/engine/openmp.cc
larroy: Is it better now?

PR #17002, opened 2019-12-07
Multi_sum_sq review, AtomicAdd removal

MoisesHer:
## Description ##
Modified multi_sum_sq operator to avoid nondeterministic behavior, which was potentially caused by AtomicAdd operation on GPU kernel.
  
### Essentials ###

- [x] All changes have test coverage: an specific test for multi_sum_sq operator has been included

### Changes ###
- [x] Modified GPU implementation of multi_sum_sq operator. The operator is launching several CUDA blocks per tensor. In the previous version, an atomic operation was used to reduce the sumSQ coming from different blocks. In this new version, we instead use a temporal storage space to write the reduction from each block, and we launch a second kernel to reduce those.


tests/python/gpu/test_operator_gpu.py
```diff
@@ -310,6 +310,35 @@ def check_fast_lars(w_dtype, g_dtype, shapes, ctx, tol1, tol2):
...
+
+    dtypes = ['float16','float32', 'float64']
+    for ctx in [mx.gpu(0)]:
+        for dtype in dtypes:
+            nparam = np.random.randint(min_nparam + 1, max_nparam + 1)
+            shapes = [np.random.randint(mindim, maxdim + 1, size=maxndim) for i in range(nparam)]
+            lowTol = ctx == mx.cpu(0) and ('float16'in [dtype])
+            tol1 = 1e-3 if lowTol else 1e-5
+            tol2 = 1e-6 if lowTol else 1e-7
+            check_multi_sum_sq(dtype, shapes, ctx, tol1, tol2)
```
eric-haibin-lin: since this PR fixes the non-determinism, is it possible to add a test that calculates squaresum twice and assert the output is deterministic?

MoisesHer: yes, that makes sense... just added it

tests/python/gpu/test_operator_gpu.py
```diff
@@ -310,6 +310,38 @@ def check_fast_lars(w_dtype, g_dtype, shapes, ctx, tol1, tol2):
...
     assert_almost_equal(ref_new_lrs.asnumpy(), mx_new_lrs.asnumpy(), atol=tol2, rtol=tol2)
 
+def check_multi_sum_sq(dtype, shapes, ctx, tol1, tol2):
+    values_arr = [np.random.rand(*shape).astype(dtype) * 10. for shape in shapes]
+
+    mx_vals = _make_ndarrays(values_arr, ctx=ctx)
+    sum_sq = mx.nd.multi_sum_sq(*mx_vals, num_arrays=len(shapes))
+    sum_sq2 = mx.nd.multi_sum_sq(*mx_vals, num_arrays=len(shapes))
+    # checks that operator is deterministic
+    assert_almost_equal(sum_sq.asnumpy(), sum_sq2.asnumpy(), atol=1e-9, rtol=1e-9)
```
eric-haibin-lin: instead of using assert_almost_equal, shall we use `np.array_equal` for `sum_sq.asnumpy(), sum_sq2.asnumpy()`? It checks if two arrays are exactly the same

MoisesHer: yes that is better, thanks

tests/python/gpu/test_operator_gpu.py
```diff
@@ -310,6 +310,36 @@ def check_fast_lars(w_dtype, g_dtype, shapes, ctx, tol1, tol2):
...
+    mindim = 50000
+    maxdim = 3200000
+    maxndim = 1
+
+    dtypes = ['float16','float32', 'float64']
+    for ctx in [mx.gpu(0)]:
+        for dtype in dtypes:
+            nparam = np.random.randint(min_nparam + 1, max_nparam + 1)
+            shapes = [np.random.randint(mindim, maxdim + 1, size=maxndim) for i in range(nparam)]
+            lowTol = ctx == mx.cpu(0) and ('float16'in [dtype])
```
eric-haibin-lin: low_tol

src/operator/contrib/multi_sum_sq.cu
```diff
@@ -106,32 +97,67 @@ __global__ void MultiSumSqKernel(int chunk_size,
...
+      if (chunks_this_tensor > param.max_chunks_per_tensor)
+        param.max_chunks_per_tensor = chunks_this_tensor;
+    }
+     // temporary storage for the reduction of each block
+    size_t workspace_size = nInputs * param.max_chunks_per_tensor * sizeof(float);
+    Tensor<gpu, 1, char> workspace =
+    ctx.requested[multi_sum_sq::kTempSpace].get_space_typed<gpu, 1, char>(
+      Shape1(workspace_size), s);
+    Tensor<gpu, 1, float> block_reductions(reinterpret_cast<float*>(&workspace[0]),
+      Shape1(nInputs * param.max_chunks_per_tensor), s);
```
eric-haibin-lin: n_inputs

src/operator/contrib/multi_sum_sq.cu
```diff
@@ -43,15 +43,14 @@ struct MultiSumSqKernelParam {
...
 template<typename DType>
 __device__ __forceinline__ DType reduce_block_into_lanes(DType* x,
-                                                         DType val,
-                                                         int lanes = 1,
-                                                         bool share_result = false) {
-  int tid = threadIdx.x + threadIdx.y * blockDim.x;
-  int blockSize = blockDim.x * blockDim.y;  // blockSize is intended to be a multiple of 32.
+                                                         DType val) {
+  int tid = threadIdx.x;
+  int blockSize = blockDim.x;
```
eric-haibin-lin: block_size

MoisesHer:
> MXNet uses CamelCase for functions and snake_case for variables. Would you mind updating the var names below? Thanks. Otherwise looks good to me

Thanks !   I think I corrected all of them

src/operator/contrib/multi_sum_sq.cu
```diff
@@ -43,96 +43,121 @@ struct MultiSumSqKernelParam {
...
+    for (int ii = 0; ii < ILP && i < i_max; ++ii, i += blockDim.x) {
       const auto incoming_val = static_cast<float>(x[i]);
       val += incoming_val * incoming_val;
     }
   }
+  const float final = ReduceBlockIntoLanes(vals, val);
+
+  if (threadIdx.x == 0) {
+    block_reductions[(start_tensor_id + tensor_loc) * param.max_chunks_per_tensor +
+                    param.block_to_chunk[blockIdx.x]] = final;
```
shuo-ouyang, line 109: Maybe we should change the variable name here? `= final` specifies that a virtual function cannot be overridden in a derived class.

PR #17007, opened 2019-12-07
[BUGFIX] Fix race condition in kvstore.pushpull

eric-haibin-lin:
## Description ##
There was a race condition in the previous implementation for the `pushpull` API. We have to add `stored.WaitToRead()` before sending it back to workers. 
@anandj91 

### Essentials ###
- [] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ x To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###

## Comments ##

### DIFF
--- a/src/kvstore/kvstore_dist_server.h
+++ b/src/kvstore/kvstore_dist_server.h
@@ -369,14 +369,22 @@ class KVStoreDistServer {
        * If pull flag is set, respond immediately with the updated values
        * Otherwise, only send the notification
        */
-      if (update_buf->request[0].pull) {
+      bool has_pull = false;
+      for (const auto& req : update_buf->request) {
+        has_pull = has_pull || update_buf->request[0].pull;
+      }
+      if (has_pull) {
+        // if there is a pull request, perform WaitToRead() once before DefaultStorageResponse
         if (has_multi_precision_copy(type)) CopyFromTo(stored, store_[key]);
         stored.WaitToRead();
         for (const auto& req : update_buf->request) {
-          DefaultStorageResponse(type, key, req, req_data, server);
+          if (req.pull) {
+            DefaultStorageResponse(type, key, req, req_data, server);
+          }
         }
         update_buf->request.clear();
       } else {
+        // otherwise, send response directly
         for (const auto& req : update_buf->request) {
           server->Response(req);
         }
--- a/src/kvstore/kvstore_dist_server.h
+++ b/src/kvstore/kvstore_dist_server.h
@@ -364,21 +364,26 @@ class KVStoreDistServer {
       if (log_verbose_)  {
         LOG(INFO) << "sent response to " << update_buf->request.size() << " workers";
       }
-      for (const auto& req : update_buf->request) {
-        /**
-         * Request can be for either push, pull or pushpull
-         * If pull flag is set, respond immediately with the updated values
-         * Otherwise, only send the notification
-         */
-        if (req.pull) {
+      /**
+       * Request can be for either push, pull or pushpull
+       * If pull flag is set, respond immediately with the updated values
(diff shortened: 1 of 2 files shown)


src/kvstore/kvstore_dist_server.h
```diff
@@ -364,21 +364,34 @@ class KVStoreDistServer {
...
+          }
+        }
+        update_buf->request.clear();
+      } else {
+        // otherwise, send response directly
+        for (const auto& req : update_buf->request) {
           server->Response(req);
         }
+        update_buf->request.clear();
+        if (has_multi_precision_copy(type)) CopyFromTo(stored, store_[key]);
```
apeforest: Can we extract these two lines out of the if and else block since it is called in both conditions.

src/kvstore/kvstore_dist_server.h
```diff
@@ -364,21 +364,34 @@ class KVStoreDistServer {
...
-         * Otherwise, only send the notification
-         */
-        if (req.pull) {
-          DefaultStorageResponse(type, key, req, req_data, server);
-        } else {
+        has_pull = has_pull || req.pull;
+      }
+      if (has_pull) {
+        // if there is a pull request, perform WaitToRead() once before DefaultStorageResponse
+        if (has_multi_precision_copy(type)) CopyFromTo(stored, store_[key]);
```
apeforest: nit: it's better to move the CopyFromTo() into a different line and add braces around the if block.

tests/nightly/dist_device_sync_kvstore.py
```diff
@@ -55,33 +58,30 @@ def init_kv():
...
             arr = [mx.nd.ones(shape, ctx=mx.gpu(j)) * scale for j in range(num_gpus)]
             val = mx.nd.zeros(shape)
-            if use_pushpull:
-                kv.pushpull('3', arr, out=val)
-            else:
-                kv.push('3', arr)
-                kv.pull('3', out=val)
+            kv.push('9', arr)
+            kv.pull('9', out=val)
+            check_diff_to_scalar(val, num)
```
apeforest: Should we pass my_rank instead of num to this function?

tests/nightly/dist_device_sync_kvstore.py
```diff
@@ -55,33 +58,30 @@ def init_kv():
...
             big_arr = [mx.nd.ones(big_shape, ctx=mx.gpu(j)) * scale for j in range(num_gpus)]
             big_val = mx.nd.zeros(big_shape)
-            if use_pushpull:
-                kv.pushpull('99', big_arr, out=big_val)
-            else:
-                kv.push('99', big_arr)
-                kv.pull('99', out=big_val)
+            kv.push('99', big_arr)
+            kv.pull('99', out=big_val)
+            check_diff_to_scalar(big_val, num)
```
apeforest: Same here.

src/kvstore/kvstore_dist_server.h
eric-haibin-lin: The order is different. In this branch, it is done after `Response` with ACK so that we send back message as early as possible. We then push to engine to perform copy. 
For the previous case, we need to response with real data instead of ACK, we actually need to perform copy first.

src/kvstore/kvstore_dist_server.h
eric-haibin-lin: In general we follow google's cpp style guide. For this case we inline the short statement to improve readability. https://google.github.io/styleguide/cppguide.html#Conditionals

Issue #16745, opened 2019-11-06
[Numpy] Cannot print the numpy scalar with format string


sxjscience:
```python
import mxnet as mx
mx.npx.set_np()
a = mx.np.array(1.0)
print('{:2f}'.format(a))
```
Error message:
```
TypeError: non-empty format string passed to object.__format__
```

In numpy:
```python
import numpy as np
a = np.array(1.0)
print('{:2f}'.format(a))
```
Returns
```
1.000000
```

PR #17009, opened 2019-12-08
[Numpy] support boolean indexing

Alicia1529:
## Description ##
1. Support simple boolean indexing in __getitem__ and __setitem__

2. Use the returned value after indexing one array to index another array
```
# after 
>>> array = np.array([1,2,3,4,5], dtype = np.int32)
>>> array
array([1, 2, 3, 4, 5], dtype=int32)
>>> idxed = array[0]
>>> idxed
array(1, dtype=int32)
>>> array[idxed]
array(2, dtype=int32)
```

3. ~Enable to index 0-d array with 0~ (cancel this change)
```
>>> idxed
array(1, dtype=int32)
>>> idxed[0]
array(1, dtype=int32)
>>> idxed[()]
array(1, dtype=int32)
```

4. Enable index with empty tuple 
```
# before
>>> b.shape # (3,3,3)
>>> b[1,(),:] # get item
# after
>>> b = np.arange(81).reshape(3,3,3,3)
>>> b[:,()]
array([], shape=(3, 0, 3, 3))
>>> b[:,(),0]
array([], shape=(3, 0, 3))
>>> b[0,(),0]
array([], shape=(0, 3))
>>> b[0,(),0] = 1000 # no change
```

5.  fix: slice does not support boolean type
```
# before
>>> b
array([[False, False, False, False, False, False, False],
       [False, False, False, False, False, False, False],
       [False, False, False, False, False, False, False],
       [ True,  True,  True,  True,  True,  True,  True],
       [ True,  True,  True,  True,  True,  True,  True]])
>>> b[:,1]
  File "/Users/luoting/Desktop/mxnet/python/mxnet/numpy/multiarray.py", line 1056, in __repr__
    array_str = self.asnumpy().__repr__()
  File "/Users/luoting/Desktop/mxnet/python/mxnet/ndarray/ndarray.py", line 2552, in asnumpy
    ctypes.c_size_t(data.size)))
  File "/Users/luoting/Desktop/mxnet/python/mxnet/base.py", line 278, in check_call
# after
>>> b[:,1]
array([False, False,  False, True,  True])
```

6. fix: cannot print the mx.numpy scalar with format string [issue](https://github.com/apache/incubator-mxnet/issues/16745)

### DIFF
--- a/python/mxnet/numpy/multiarray.py
+++ b/python/mxnet/numpy/multiarray.py
@@ -65,6 +65,10 @@ __all__ = ['ndarray', 'empty', 'array', 'shape', 'zeros', 'zeros_like', 'ones', 
 _NDARRAY_UNSUPPORTED_INDEXING = -1
 _NDARRAY_BASIC_INDEXING = 0
 _NDARRAY_ADVANCED_INDEXING = 1
+_NDARRAY_EMPTY_TUPLE_INDEXING = 2
+_NDARRAY_BOOLEAN_INDEXING = 3
+_NDARRAY_INT_BOOLEAN_INDEXING = 4
+_NDARRAY_SLICE_BOOLEAN_INDEXING = 5
 
 
 # This function is copied from ndarray.py since pylint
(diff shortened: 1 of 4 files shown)


python/mxnet/numpy/multiarray.py
```diff
@@ -324,6 +327,134 @@ def _set_np_advanced_indexing(self, key, value):
...
+            if pos != len(key) - 1:
+                raise NotImplementedError('only support boolean array at the end of the idces '
+                                          'when it is mixed with integers')
+            for idx in range(pos):
+                data = data[key[idx]]
+                pos -= 1
+
+        if isinstance(value, numeric_types):
+            _npi.boolean_mask_assign_scalar(data=data, mask=mask, value=value, start_axis=pos, out=data)
+        elif isinstance(value, NDArray):
```
haojin2: should check
```python
elif isinstance(value, ndarray):
```

reminisce:
I think it should error out in the following case since semantically, there is no axis for 0-d arrays and NumPy raises the error as well.
```python
>>> import numpy as np
>>> a = np.array(3432)
>>> a
array(3432)
>>> a[0]
IndexError: too many indices for array
```

Alicia1529:
> I think it should error out in the following case since semantically, there is no axis for 0-d arrays and NumPy raises the error as well.
> 
> ```python
> >>> import numpy as np
> >>> a = np.array(3432)
> >>> a
> array(3432)
> >>> a[0]
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> IndexError: too many indices for array
> ```
Sorry about it. I removed this change just now.

Issue #16795, opened 2019-11-12
[RFC] Unified API for Distributed Data Parallel Training


eric-haibin-lin:
## Background  
Data parallel training is the most common distributed training technique when it comes to multiple GPUs or multiple hosts. Currently, several communication backends provide functionalities for communicating tensors across devices/hosts for data parallel training. For MXNet users, there are a few options: 
1. native kvstore
2. [p3 kvstore](https://www.sysml.cc/doc/2019/75.pdf)
3. [horovod](https://github.com/horovod/horovod/) 
4. [bytePS](https://github.com/bytedance/byteps/)

These different implementations provide different APIs:
- native kvstore
  - high level APIs: `mx.gluon.Trainer`
  - low level APIs: `kv.push`, `kv.pull`, `kv.init`
- horovod
  - high level APIs: `hvd.init()`, `hvd.DistributedTrainer`
  - low level APIs: `hvd.broadcast`, `hvd.allreduce`
- bytePS
  - high level APIs: `bps.init()`, `bps.DistributedTrainer`
  - low level APIs: `byteps_declare_tensor`, `byteps_push_pull`

Here, high level APIs refers to the API a typical novice user uses for a distributed training job. To communicate tensors not managed by `Trainer` or `DistributedTrainer`s, users may refer to the low level APIs to send/receive a custom tensor. 

## Problem Statement
Sometimes we want to easily switch between these different distributed communication backends and compare which one performs the best for a particular distributed training environment. Due to different APIs of these implementations, it requires lots of user code changes to try each one of them. It typically involves custom logics to:
1. launch python processes for distributed training job ([BytePS launch](https://github.com/bytedance/byteps/blob/master/docs/step-by-step-tutorial.md#mxnet) v.s. [horovod](https://github.com/horovod/horovod#running-horovod)) 
2. initialize communication backends ([example code](https://github.com/eric-haibin-lin/gluon-nlp/blob/benchmark/scripts/bert/run_pretraining.py#L187-L228))
3. create (Distributed)Trainers ([example code](https://github.com/eric-haibin-lin/gluon-nlp/blob/benchmark/scripts/bert/run_pretraining.py#L297-L303))
4. send custom tensors ([example code](https://github.com/eric-haibin-lin/gluon-nlp/blob/benchmark/scripts/bert/run_pretraining.py#L582-L586))

## Proposal 

[...]

PR #17010, opened 2019-12-08
[API] unified API for custom kvstores

eric-haibin-lin:
## Description ##
depends on https://github.com/apache/incubator-mxnet/pull/17007 
https://github.com/apache/incubator-mxnet/issues/16795 
See the `KVStoreBase` for all interfaces defined in `kvstore/base.py`. See `TestStore` class in the same file as an example of custom single node kvstore which implements minimal required interfaces. 

The interfaces include:
- broadcast
- pushpull
- num_workers
- rank
- type
- is_capable

We can reuse these interfaces for other communication backends such as p3 and horovod. @anandj91 @apeforest 

### Essentials ###

### Changes ###
Added test for:
- pushpull with teststore in `test_kvstore_custom.py`
- broadcast with teststore in `test_kvstore_custom.py`
- gluon trainer with teststore in `test_gluon_trainer.py`

## Comments ##


eric-haibin-lin:
@ChaokunChang FYI

python/mxnet/__init__.py
```diff
@@ -84,22 +86,20 @@
...
 from . import gluon
 
-# Dynamic library module should be done after ndarray and symbol are initialized
-from . import library
-from . import tvmop
-
-__version__ = base.__version__
-
+# use mx.kv as short for kvstore
 # Dist kvstore module which launches a separate process when role is set to "server".
```
apeforest: Maybe you also want to update the comment here. In the case of HVD, there will be no "server" process.

python/mxnet/kvstore/base.py
```diff
@@ -0,0 +1,452 @@
...
+
+def _ctype_dict(param_dict):
+    """Returns ctype arrays for keys and values(converted to strings) in a dictionary"""
+    assert(isinstance(param_dict, dict)), \
+        "unexpected type for param_dict: " + str(type(param_dict))
+    c_keys = c_array(ctypes.c_char_p, [c_str(k) for k in param_dict.keys()])
+    c_vals = c_array(ctypes.c_char_p, [c_str(str(v)) for v in param_dict.values()])
+    return (c_keys, c_vals)
+
+class KVStoreBase(object):
```
apeforest: Do you see a case of using allgather API?

python/mxnet/kvstore/base.py
```diff
@@ -0,0 +1,452 @@
...
+    """An abstract key-value store interface for data parallel training."""
+
+    def broadcast(self, key, value, out, priority=0):
+        """ Broadcast the `value` NDArray at rank 0 to all ranks,
```
apeforest: This API seems to have no direct control of rank. Maybe we don't need to mention it here?

python/mxnet/kvstore/base.py
```diff
@@ -0,0 +1,452 @@
...
+        out : NDArray, or list of NDArray
+            Values corresponding to the key to store the result
+
+        priority : int, optional
+            The priority of the operation.
+            Higher priority operations are likely to be executed before other actions.
+        """
+        raise NotImplementedError()
+
+    def pushpull(self, key, value, out=None, priority=0):
```
apeforest: If horovod-based KVStore is going to be derived from this base, maybe we should consider removing this API from this base class.

python/mxnet/kvstore/base.py
```diff
@@ -0,0 +1,452 @@
...
+
+        Parameters
+        ----------
+        fname : str
+            Path to input states file.
+        """
+        raise NotImplementedError()
+
+    @property
+    def type(self):
```
apeforest: What exactly does the type mean here? Different types of backend? If so, maybe write it clearly in comment, or use a more expressive API name.

python/mxnet/kvstore/base.py
```diff
@@ -0,0 +1,452 @@
...
+
+        Returns
+        -------
+        type : str
+            the string type
+        """
+        raise NotImplementedError()
+
+    @property
+    def rank(self):
```
apeforest: Does ps-based KVStore use the rank information? If not, we should remove it from base class.

python/mxnet/kvstore/base.py
```diff
@@ -0,0 +1,452 @@
...
+    then ``dist_sync`` behaves like ``local`` with batch size ``n * b``.
+
+    ``dist_device_sync``: Identical to ``dist_sync`` with the difference similar
+    to ``device`` vs ``local``.
+
+    ``dist_async``: Performs asynchronous updates.
+    The weights are updated whenever gradients are received from any machine.
+    No two updates happen on the same weight at the same time. However, the order is not
+    guaranteed.
+
```
apeforest: Please also add 'nccl' and 'horovod' description here. And I suppose p3 and byteps can be covered any one of the types above?

python/mxnet/kvstore/kvstore.py
```diff
@@ -327,39 +344,37 @@ def pushpull(self, key, value, out=None, priority=0):
...
             Values corresponding to the keys.
 
         out: NDArray or list of NDArray or list of list of NDArray
             Values corresponding to the keys.
 
         priority : int, optional
-            The priority of the pull operation.
-            Higher priority pull operations are likely to be executed before
-            other pull actions.
+            The priority of the operation.
```
apeforest: Does priority 0 mean the highest or the lowest? Please specify.

python/mxnet/model.py
```diff
@@ -122,10 +130,10 @@ def _initialize_kvstore(kvstore, param_arrays, arg_params, param_names, update_o
     """Initialize kvstore"""
     for idx, param_on_devs in enumerate(param_arrays):
         name = param_names[idx]
-        kvstore.init(name, arg_params[name])
-
-        if update_on_kvstore:
-            kvstore.pull(name, param_on_devs, priority=-idx)
+        if not update_on_kvstore or arg_params[name].stype != 'default':
```
apeforest: Where is the type 'default' defined? Also, it's better for readability to avoid using double negative logic.

python/mxnet/model.py
```diff
@@ -155,9 +163,12 @@ def _update_params_on_kvstore(param_arrays, grad_arrays, kvstore, param_names):
             continue
         name = param_names[index]
         # push gradient, priority is negative index
-        kvstore.push(name, grad_list, priority=-index)
         # pull back the weights
-        kvstore.pull(name, arg_list, priority=-index)
+        if grad_list[0].stype == 'default' and arg_list[0].stype == 'default':
```
apeforest: Where is 'default' defined?

python/mxnet/model.py
```diff
@@ -171,9 +182,11 @@ def _update_params(param_arrays, grad_arrays, updater, num_device,
         if kvstore:
             name = param_names[index]
             # push gradient, priority is negative index
-            kvstore.push(name, grad_list, priority=-index)
-            # pull back the sum gradients, to the same locations.
-            kvstore.pull(name, grad_list, priority=-index)
+            if grad_list[0].stype == 'default' and arg_list[0].stype == 'default':
```
apeforest: Where is 'default' defined?

tests/nightly/dist_device_sync_kvstore_custom.py
```diff
@@ -0,0 +1,96 @@
...
+init_test_keys_device_big = [str(i) for i in range(500,600)]
+
+shape = (2, 3)
+big_shape = (1200, 1200)        # bigger than MXNET_KVSTORE_BIGARRAY_BOUND
+
+kv = mx.kv.create(args.name)
+my_rank = kv.rank
+my_num_workers = kv.num_workers
+
+def test_push_pull():
```
apeforest: nit: please use the name test_pushpull() to be consistent with the called API name.

tests/python/unittest/test_kvstore.py
```diff
@@ -283,6 +283,26 @@ def test_get_type():
     kvtype = 'local_allreduce_cpu'
     kv = mx.kv.create(kvtype)
     assert kv.type == kvtype
+    kvtype = 'teststore'
+    kv = mx.kv.create(kvtype)
+    assert kv.type == kvtype
+
+@with_seed()
+def test_custom_store():
```
apeforest: Should this method go to test_kvstore_custom.py ?

tests/python/unittest/test_kvstore_custom.py
```diff
@@ -0,0 +1,165 @@
...
+    check_aggregator(init_kv('device'), 'a', str_keys)
+    check_aggregator(init_kv('teststore'), 3)
+    check_aggregator(init_kv('teststore'), 'a')
+
+@with_seed()
+def test_get_type_device():
+    kvtype = 'device'
+    kv = mx.kv.create(kvtype)
+    assert kv.type == kvtype
+
```
apeforest: Do we have to test set_optimizer_states and load_optimizer_states as well?

python/mxnet/gluon/trainer.py
```diff
@@ -354,14 +366,30 @@ def allreduce_grads(self):
...
+            return
+        for i, param in enumerate(self._params):
+            if param.grad_req != 'null':
 
-                    self._kvstore.push(i, param.list_grad(), priority=-i)
-                    if not self._update_on_kvstore:
-                        self._kvstore.pull(i, param.list_grad(), priority=-i,
+                grad_list = param.list_grad()
+                # sparse gradients, call push and pull separately
+                if grad_list[0].stype != 'default':
```
apeforest: Please avoid double negative condition to improve readability.

eric-haibin-lin: where is the double negative condition? There is only one negative condition, right?

python/mxnet/kvstore/base.py
eric-haibin-lin: Horovod can override pushpull to perform allreduce

python/mxnet/kvstore/base.py
eric-haibin-lin: We do. Otherwise users don't know which rank's tensor are being broadcast.

python/mxnet/kvstore/base.py
eric-haibin-lin: pushpull without optimizer is basically the same as doing allreduce with gradients

python/mxnet/kvstore/base.py
eric-haibin-lin: what about updating the doc to be "Returns the type of this kvstore backend" ?

python/mxnet/kvstore/base.py
eric-haibin-lin: it does

python/mxnet/kvstore/base.py
eric-haibin-lin: We should only update the description once p3/horovod integration is merged. I expect the horovod integration PR to update this document. Otherwise, it will be misleading to talk about a non-existing backend

python/mxnet/kvstore/kvstore.py
eric-haibin-lin: the type is int. int can be positive or negative. So 0 is neither highest or lowest.

python/mxnet/model.py
eric-haibin-lin: ok, i can update the branch condition. The stype is specified in ndarray documentation e.g. https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/ndarray/utils.py#L51-L52 

default = dense
row_sparse = row_sparse
csr = compressed sparse row

tests/nightly/dist_device_sync_kvstore_custom.py
eric-haibin-lin: ok

tests/python/unittest/test_kvstore.py
eric-haibin-lin: i can do that

tests/python/unittest/test_kvstore_custom.py
eric-haibin-lin: i can add that

python/mxnet/__init__.py
eric-haibin-lin: Sure. I think there will be a separate PR for unifying the launching method of different backends, but yes I can add some clarification here in the doc first

python/mxnet/kvstore/base.py
eric-haibin-lin: For APIs like allgather, they are not involved in the usual data parallel training practice using either module or gluon trainer. The native kvstore does not support it either. However, users are free to call hvd.allgather when they use hvd backend to gather any tensor they want.

python/mxnet/kvstore/base.py
eric-haibin-lin: I will add more comments on this function for clarification with allreduce backends.

eric-haibin-lin:
Thanks for the detailed review comments. Just updated the PR.

python/mxnet/gluon/trainer.py
apeforest: Sorry, I guess I read it wrong.

python/mxnet/kvstore/base.py
apeforest: By the naming itself, pushpull indicate a parameter-server structure. It might also put some restrictions if we want to implement some allreduce-based operations in the future.

What do you think of stripping this method out of the base class. And let PS-based KVstore class use pushpull and Allreduce-based KVStores use allreduce. I feel it's easier to understand from user's point of view.

python/mxnet/kvstore/kvstore.py
apeforest: So maybe we want to mention the priority is in ascending order? e.g. the larger the value the higher the priority?

python/mxnet/model.py
apeforest: Okay, I guess I overlooked. The stype here is the storage type, not the type of kvstore.

python/mxnet/kvstore/base.py
apeforest: Discussed @eric-haibin-lin offline. Using one API interface simplifies the coding in Training object. It might be a good idea to add an alias allreduce for non-PS based strategies for better readability.

PR #14743, opened 2019-04-19
change mxnet_option behavior

yajiedesign:
## Description ##
change mxnet_option behavior
 
In the past, when the condition in mxnet_option was false, the option would be deleted completely instead of set to `OFF`. This is not the expected behavior. Now, change it to set to `OFF` when the condition is false.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [X] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

PR #17018, opened 2019-12-09
Replace mxnet_option macro with standard CMAKE_DEPENDENT_OPTION

leezu:
## Description ##
Replace mxnet_option macro with standard CMAKE_DEPENDENT_OPTION.
Using standard language constructs improves maintainability and eases reading the code.

### Essentials ###

### Changes ###
- [X] Replace mxnet_option macro with CMAKE_DEPENDENT_OPTION part of CMake "standard library"

## Comments ##
@cjolivier01 please confirm the change in `FindMKL.cmake` is correct. Previously, based on #14743, if the condition was false, the option was set to `OFF`.


szha:
cc @junrushao1994 who may want to offer suggestions

junrushao1994:
I think I am a very beginner in using cmake. Looks like `CMAKE_DEPENDENT_OPTION ` provides a native way to express cmake options, which is great :-)

CMakeLists.txt
```diff
@@ -18,46 +18,49 @@ endif()
...
-mxnet_option(USE_ASAN             "Enable Clang/GCC ASAN sanitizers." OFF)
-mxnet_option(ENABLE_TESTCOVERAGE  "Enable compilation with test coverage metric output" OFF)
-mxnet_option(USE_INT64_TENSOR_SIZE "Use int64_t to represent the total number of elements in a tensor" OFF)
-mxnet_option(BUILD_CYTHON_MODULES "Build cython modules." OFF)
+option(USE_CUDA "Build with CUDA support"   ON)
+option(USE_OLDCMAKECUDA "Build with old cmake cuda" OFF)
+option(USE_NCCL "Use NVidia NCCL with CUDA" OFF)
+option(USE_OPENCV "Build with OpenCV support" ON)
+option(USE_OPENMP "Build with Openmp support" ON)
+cmake_dependent_option(USE_CUDNN "Build with cudnn support" ON "USE_CUDA" OFF) # one could set CUDNN_ROOT for search path
```
yajiedesign: The operation of cmake_dependent_option is not the same as before.

> This macro presents an option to the user only if a set of other conditions are true

leezu: I think we should follow the best practice of CMake, which is to expose the option only when it's meaningful. Specifically for CUDNN, it's common practice to only expose it if Cuda is enabled

https://github.com/pytorch/pytorch/blob/04b93244762e0126887c5fcc79fd8cffc2a776c6/CMakeLists.txt#L143-L145

What problem do you see if we present the option to the user only when CUDA is set to true?

leezu: Also there is some logic for `set(USE_CUDNN OFF)` If Cuda is set to 0. So this change just simplifies the logic

yajiedesign: @leezu There is nothing wrong here, the problem is the  33 lines.

CMakeLists.txt
```diff
@@ -18,39 +18,41 @@ endif()
...
+cmake_dependent_option(USE_SSE "Build with x86 SSE instruction support" ON "NOT ARM" OFF)
+option(USE_F16C "Build with x86 F16C instruction support" ON) # autodetects support if ON
+option(USE_LAPACK "Build with lapack support" ON)
+option(USE_MKL_IF_AVAILABLE "Use MKL if found" ON)
+cmake_dependent_option(USE_MKLDNN "Build with MKL-DNN support" ON
+  "USE_MKL_IF_AVAILABLE;NOT APPLE;NOT MSVC;CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL \"x86_64\";NOT CMAKE_CROSSCOMPILING" OFF)
```
yajiedesign: this line.cmake_dependent_option hidden USE_MKLDNN  on MSVC.but we needed is off by default.

CMakeLists.txt
```diff
@@ -18,39 +18,41 @@ endif()
...
+cmake_dependent_option(USE_OPERATOR_TUNING  "Enable auto-tuning of operators" ON "NOT MSVC" OFF)
```
yajiedesign: this line same.

CMakeLists.txt
leezu: What do you mean? It is off by default, only if all of the following conditions are true, we set it to ON. That's exactly the same logic as we had before:

`"USE_MKL_IF_AVAILABLE;NOT APPLE;NOT MSVC;CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL \"x86_64\";NOT CMAKE_CROSSCOMPILING"`

vs before

`USE_MKL_IF_AVAILABLE AND (NOT APPLE) AND (NOT MSVC) AND (CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "x86_64") AND (NOT CMAKE_CROSSCOMPILING))`

The `;` stands for `AND`

CMakeLists.txt
leezu: See above, I think its correct

CMakeLists.txt
yajiedesign: From the documentation of cmake_dependent_option, if the condition is false. The option will be hidden, which will cause MKLDNN to not be opened in the MSVC environment.

yajiedesign: Maybe I understand the document wrong. I'll do a test

yajiedesign: image
not wrong.cmake_dependent_option hidden option with MSVC.

leezu: Which document do you refer to? You can check the official documentation https://cmake.org/cmake/help/v3.16/module/CMakeDependentOption.html

> This macro presents an option to the user only if a set of other conditions are true. When the option is not presented a default value is used, but any value set by the user is preserved for when the option is presented again.

Thus the default value is used if the condition is false. Thus we have the same logic.

leezu: I'm not familiar with the Windows UI. What does your screenshot show?

leezu: Do you mean we want to allow users on MSVC to use MKLDNN? Then indeed the logic is wrong.

yajiedesign: The operation logic of mxnet_optiond has been modified. The new version of mxnet_option does not hide the options. It just adjusts the default values ​​based on the conditions.
https://github.com/apache/incubator-mxnet/pull/14743

leezu: If it is supported, why does it have a different default value? Let's use the same default value (`ON`) then on all supported platforms? If it is not supported, then hiding is correct.

CMakeLists.txt
leezu: Agree

CMakeLists.txt
leezu: Let's discuss the assimilation of default values separately. I fixed the hiding problem for now with `if else` condition. So now it's at least very easy to understand what is going on (whereas before users need to study the Macro implementation to understand the behaviour..)

PR #17027, opened 2019-12-09
Merge make/pip and make/maven configurations

zachgk:
## Description ##
Merge the make/maven and make/pip configuration folders used in the static linked build. Currently there should not be any differences between them, and this will help ensure that differences are not created accidentally.

@lanking520 

### Essentials ###


tools/staticbuild/build_lib.sh
```diff
@@ -18,7 +18,7 @@
 # under the License.
 
 # This script builds the libraries of mxnet.
-make_config=make/${STATIC_BUILD_TARGET}/${STATIC_BUILD_TARGET}_${PLATFORM}_${VARIANT}.mk
+make_config=make/${STATIC_BUILD_TARGET}/${PLATFORM}_${VARIANT}.mk
```
szha: if there's no difference then the `${STATIC_BUILD_TARGET}` should be removed from the path too.

zachgk: I think this might be better because it would be easier if there is a difference in the future or if another language binding uses this build but needs changes to the configuration

szha: if you anticipate divergence then there's no need to merge them

Issue #14979, opened 2019-05-17
[BUG] Using a package with MKL and GPU versions, using python to open a new process will cause an error


fierceX:
Hardware and version information:

----------Python Info----------
Version      : 3.6.8
Compiler     : GCC 7.3.0
Build        : ('default', 'Dec 30 2018 01:22:34')
Arch         : ('64bit', '')
------------Pip Info-----------
Version      : 19.1.1
Directory    : /home/bird/miniconda3/envs/test/lib/python3.6/site-packages/pip
----------MXNet Info-----------
Version      : 1.4.1
Directory    : /home/bird/miniconda3/envs/test/lib/python3.6/site-packages/mxnet
Hashtag not found. Not installed from pre-built package.
----------System Info----------
Platform     : Linux-4.15.0-50-generic-x86_64-with-debian-buster-sid
system       : Linux
node         : ctmp
release      : 4.15.0-50-generic
version      : #54-Ubuntu SMP Mon May 6 18:46:08 UTC 2019
----------Hardware Info----------
machine      : x86_64
processor    : x86_64
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              8
On-line CPU(s) list: 0-7
Thread(s) per core:  2
Core(s) per socket:  4
Socket(s):           1
NUMA node(s):        1
Vendor ID:           GenuineIntel
CPU family:          6
Model:               94
Model name:          Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz
Stepping:            3
CPU MHz:             800.218
CPU max MHz:         4000.0000
CPU min MHz:         800.0000
BogoMIPS:            6816.00
Virtualization:      VT-x
L1d cache:           32K
L1i cache:           32K
L2 cache:            256K
L3 cache:            8192K
NUMA node0 CPU(s):   0-7
[...]

PR #17039, opened 2019-12-11
Fix omp fork race conditions and assert issue

cjolivier01:
## Description ##
Fix for: https://github.com/apache/incubator-mxnet/issues/14979

### Essentials ###
- [ X] Code is well-documented: 
- [X ] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

## Comments ##

* When forking, sometimes mxnet can hook its fork handler before openmp's, depending upon static startup order as well as variables such as whether OMP_NUM_THREADS is set.  This is mitigated by forcing an OMP call before mxnet hooks a fork handler as well as first-thing after forking as the child.  This avoids two scenarios:
  1) An OMP API call by mxnet before openmp's child fork handler has run, reinitializing its state to the new process
  2) Starting a thread which may call an OpenMP API call (or open a parallel region) before or **during** OMP library's child fork handler.

* A considerable amount of multithreading is done by mxnext both during static init as ell as in atfork() handlers, and we may wish to change this over time.  Due to the non-deterministic startup order of static init (and thus sometimes fork handling), it can cause unpredictable results.

* Build omp in release mode to stop harmless assert.  Comment notes that assert is harmless and represents a variable that fails to get set to null in their atfork_child handler (they set other similar variables to to null). The variable is set to NULL in the next line of code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -249,7 +249,7 @@ if(USE_TENSORRT)
 endif()
 
 # please note that when you enable this, you might run into an linker not being able to work properly due to large code injection.
-# you can find more information here https://github.com/apache/incubator-mxnet/issues/15971 
+# you can find more information here https://github.com/apache/incubator-mxnet/issues/15971
 if(ENABLE_TESTCOVERAGE)
   message(STATUS "Compiling with test coverage support enabled. This will result in additional files being written to your source directory!")
   find_program( GCOV_PATH gcov )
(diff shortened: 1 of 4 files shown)


cjolivier01:
I actually also did change (removed from this PR in order to keep it simple) that stops forcing the creation of the engine during static init and the fork handlers, if it is desired.  This has a couple of effects including reducing the overhead of resources and threads created simply by the library being loaded, which then, in turn, causes a lot of thread churn (Stop(), Start()) when forking.  Even if hey never use mxnet, we spin up a lot of stuff unnecessarily. It actually gets  lot faster on startp when this is done.

szha:
> I actually also did change (removed from this PR in order to keep it simple) that stops forcing the creation of the engine during static init and the fork handlers, if it is desired. This has a couple of effects including reducing the overhead of resources and threads created simply by the library being loaded, which then, in turn, causes a lot of thread churn (Stop(), Start()) when forking. Even if hey never use mxnet, we spin up a lot of stuff unnecessarily. It actually gets lot faster on startp when this is done.

The change sounds useful. Do you plan to commit the change?

cjolivier01:
> > I actually also did change (removed from this PR in order to keep it simple) that stops forcing the creation of the engine during static init and the fork handlers, if it is desired. This has a couple of effects including reducing the overhead of resources and threads created simply by the library being loaded, which then, in turn, causes a lot of thread churn (Stop(), Start()) when forking. Even if hey never use mxnet, we spin up a lot of stuff unnecessarily. It actually gets lot faster on startp when this is done.
> 
> The change sounds useful. Do you plan to commit the change?

If I have time soon

PR #17047, opened 2019-12-11
Add build flag for x86 arch

TaoLv:
## Description ##
Add a new build flag `USE_X86_ARCH` for CPU build, so users can leverage advanced x86 instructions when build mxnet from source code. Previously, only sse3 is supported.

The new flag is OFF by default.

### Essentials ###

### Changes ###

## Comments ##


ZhennanQin:
Shall we add this to the MXNet build features list? For now we've had below CPU features:
```
✔ CPU_SSE
✔ CPU_SSE2
✔ CPU_SSE3
✔ CPU_SSE4_1
✔ CPU_SSE4_2
✖ CPU_SSE4A
✔ CPU_AVX
✖ CPU_AVX2
```

It's better we can update this list from the march specified. This will help to locate the bug from certain instruction sets.

chinakook:
Can we link this dynamically, like MKL?

TaoLv:
@larroy Do you have any suggestion about the name? I choose this one because:
1. `-march` is under the section `x86 options` of [gcc document](https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html). So I should be more friendly is the user is familiar with gcc options.
2. It helps to differentiate from CUDA arch or Arm arch.

TaoLv:
> Does it also need to interact with USE_SSE? if you set march I don't think we need to set sse level explicitly.

Yes, there is no need to set -msse3 explicitly if -march has a more advanced arch. But I keep -msse3 here as a gate keeper since we won't want to support lower end machines. @larroy

CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ option(USE_OPENMP "Build with Openmp support" ON)
 cmake_dependent_option(USE_CUDNN "Build with cudnn support" ON "USE_CUDA" OFF) # one could set CUDNN_ROOT for search path
 cmake_dependent_option(USE_SSE "Build with x86 SSE instruction support" ON "NOT ARM" OFF)
 option(USE_F16C "Build with x86 F16C instruction support" ON) # autodetects support if ON
+option(USE_X86_ARCH         "Build with x86 options for -march" OFF)
```
leezu: `option` can only be `ON` or `OFF`. I suspect users want to specify string values such as `native` here.

leezu:
> Yes, there is no need to set -msse3 explicitly if -march has a more advanced arch. But I keep -msse3 here as a gate keeper since we won't want to support lower end machines. @larroy

Why do we need such gate keeper? How about defaulting to `march=native -mtune=native`?

CMakeLists.txt
TaoLv: Thanks! How about changing to:
`set(USE_X86_ARCH "" CACHE STRING "Build with x86 options for -march")`

leezu: Yes. But please see my question about default value below.

TaoLv:
> Why do we need such gate keeper? How about defaulting to march=native -mtune=native?

I also considered to have native as the default value. But I'm not sure how would this affect the process of nightly release or the usage of other users. If they build MXNet on a high end machine and want to deploy it to a lower end one, it might have problems.

leezu:
We need to set `-march` and `-mtune` reasonably when building releases for other users (such as nightly build). But there's no problem to do that on our side.

> If they build MXNet on a high end machine and want to deploy it to a lower end one, it might have problems.

It's already true for the SSE3 instructions, though arguably SSE3 is widely available, so the case is unlikely.

One solution would be to to point out in the setup instructions that `native` needs to be overwritten when planning to use the build on a different machine?

szha:
Binary release is one simpler use case of cross compilation. We should probably build it as a preset in cmake which includes fine-grained control at instruction set level. Currently we assume SSE2/3 and AVX2 in binary releases and, yes it would be disastrous if we were to release a binary that require AVX512...

CMakeLists.txt
larroy: @leezu it actually accepts other values that ON OFF, it's a bit misleading.

leezu:
@larroy besides `march` there is `mtune`. But on ARM it's `mcpu` instead of `mtune`. We also need to handle such discrepancy

TaoLv:
@larroy We already have CUDA_ARCH, why we cannot have X86_ARCH? I don't think it's a good idea to mix up x86 arch with ARM arch into one build flag given that their have quite different options. The -march option also need be passed to MKL-DNN build, supporting both x86 and ARM in one build flag will complicate the build logic here.

@leezu I read "Specifying -march=cpu-type implies -mtune=cpu-type." from https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html. Do we still need put it into the command line?

@szha May I know your suggestion for the default value of this flag?

leezu:
@TaoLv From the wording alone I'd agree with you. But it seems that the wording is not precise:

Quoting https://lemire.me/blog/2018/07/25/it-is-more-complicated-than-i-thought-mtune-march-in-gcc/

> This has lead me to believe that “-march” trumps “-mtune” meaning that if you set “-march=native”, then the “-mtune” is effectively irrelevant.
> 
> I was wrong.

In particular,  `march=native` doesn't imply `mtune=native`. But for example, `march=broadwell` implies `mtune=broadwell`. Thus if we use `native`, we must specify both `march` and `mtune`.

TaoLv:
I see. Not sure if it's a problem of the documentation or the stale version of compiler. Running `gcc -Q --help=target -march=native` with gcc 7.3, my cascade lake machine is identified as knl and -mtune is set to generic. :(

Is the below logic good to you?
```
ifneq ($(USE_X86_ARCH), NONE)
	CFLAGS += -march=$(USE_X86_ARCH) -mtune=$(USE_X86_ARCH)
endif
```

larroy:
I'm in favor of the intention of this change, but the details need to be ironed out.

TaoLv:
> I'm in favor of the intention of this change, but the details need to be ironed out.

Thank you @larroy. Sorry I'm not going to add the ARM part. It's out of my intention of this PR and I don't have any environment to test it out. If you're interested, you're very welcomed to contribute for that.
Regarding the runtime feature list, as you're the author of that, do you think it's enough to add new AVX512 instruction sets to the list here: https://github.com/apache/incubator-mxnet/blob/master/src/libinfo.cc#L67?  -march have many options, I'm not sure how to reflect the user choice to the feature list.

larroy:
can we just name the flag "USE_MARCH_MTUNE" ?

leezu:
On ARM there's also `mcpu`

TaoLv:
@eric-haibin-lin The cmake part should be covered by #17468. I'm not sure if we still need the changes for makefile as the project is moving towards cmake build system. @leezu

szha:
this change doesn't seem needed anymore for master branch. feel free to reopen and send towards 1.x branch if still deemed necessary.

PR #15834, opened 2019-08-09
Check expected vs. actual Node inputs.  Many operators non-compliant.

DickJC123:
## Description ##
PR https://github.com/apache/incubator-mxnet/pull/15688 fixed, among other things, a discrepancy between the number of inputs specified in the _backward_clip operator definition via .set_num_inputs() and the number of inputs actually created for the backward node (1 vs 2, respectively).  Use of monitors (for debugging) requires these input counts to be equal.

This PR supplies the check of expected vs. actual node input counts and exposes the fact that many operators have a discrepancy.  As suggested by @ptrendx, this check is best supplied in this separate PR, as it might take some time to fix all the non-compliant operators.  The first commit only supplies the check, not operator fixes, and so is expected to fail.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [X ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #17049, opened 2019-12-11
Fix operators lying about their number of inputs

ptrendx:
## Description ##
This PR fixes a number of backward operators that advertise the wrong number of inputs (either by having a wrong number passed to `set_num_inputs` or leaving it empty which defaults to 1 input). It also introduces a check in `MakeNode` to catch this error.

Having consistent number of inputs and outputs is important because tools like `Monitor` rely on them (and currently some of the networks are impossible to inspect during backward because of this) and they are needed for more advanced manipulation of the computational graph.

### Essentials ###


szha:
how should we test for this systematically?

ptrendx:
I introduced a test for it in the `MakeNode` function which should cover vast majority of the cases. There is still however a possibility of an escape if you make a node without inputs and then add them manually in your `FGradient`. For that case only an actual graph validation pass would catch that (NNVM does not really give any enforcement mechanisms for this).

src/operator/nn/concat.cc
```diff
@@ -394,6 +394,14 @@ CONCAT_FORWARD_ATTRS
 .add_arguments(ConcatParam::__FIELDS__());
 
 NNVM_REGISTER_OP(_backward_Concat)
+.set_num_inputs([](const NodeAttrs& attrs) {
+#if MXNET_USE_MKLDNN
```
TaoLv: `#if MXNET_USE_MKLDNN == 1`

ptrendx: :+1:

DickJC123:
Thanks for carrying through on this issue.  I remember running into the problem and getting only as far as describing the issue with this incomplete PR https://github.com/apache/incubator-mxnet/pull/15834.  I reviewed my implementation of the input check there and I like that you have pushed the check down into MakeNode().  However, I notice in my implementation, I was protecting the routine MakeNonLossGradNode, where you don't have a similar check because MakeNode() is called with inputs=null.  Could you review my commit above and add a check for MakeNonLossGradNode if you feel it enhances the protection?

PR #17031, opened 2019-12-10
Use CMake standard library to handle CUDA

leezu:
## Description ##
Use CMake standard library to handle CUDA.
Makes our code more maintainable and may fix bugs, because we rely on code tested by all CMake users, instead of the small subset of MXNet CMake users.

Include FindCUDAToolkit that replaces the deprecated FindCUDA module as of CMake 1.17.
https://gitlab.kitware.com/cmake/cmake/merge_requests/4093

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at https://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [X] Refactor cuda handling in cmake build 

## Comments ##
- Bump CMake minimum version requirement due to
  https://gitlab.kitware.com/cmake/cmake/issues/19199
- Lazy consensus for the minimum version bump: https://lists.apache.org/thread.html/82ead692d172918de3afdeb27dbd533d37e1c084aa5738c5e353bbfd%40%3Cdev.mxnet.apache.org%3E

PR #17063, opened 2019-12-13
Fix armv7 build by updating the base container and pin it

### LINKED ISSUE #16753: fail to build using docker
## Description
Tried building library using docker through the command: _python3 ci/build.py -p armv7_ but compilation failed because it seems docker is pulling some binaries from and old repository (jessie). Tried running the same code with --force-yes with the same result.

Also tried to compile directly on Raspberry pi, but it fails because of insufficient VM, even after I increased it to 4096M. I'll move to cross-compilation next.

### Error Message
WARNING: The following packages cannot be authenticated!
  autoconf
E: There are problems and -y was used without --force-yes
The command '/bin/sh -c /work/deb_ubuntu_ccache.sh' returned a non-zero code: 100
  File "ci/build.py", line 454, in <module>
    sys.exit(main())
  File "ci/build.py", line 364, in main
    num_retries=args.docker_build_retries, no_cache=args.no_cache)
  File "ci/build.py", line 116, in build_docker
    run_cmd()
  File "/Volumes/External/mxnet/ci/util.py", line 81, in f_retry
    return f(*args, **kwargs)
  File "ci/build.py", line 114, in run_cmd
    check_call(cmd)
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/subprocess.py", line 363, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['docker', 'build', '-f', 'docker/Dockerfile.build.armv7', '--build-arg', 'USER_ID=501', '--build-arg', 'GROUP_ID=20', '--cache-from', 'mxnetci/build.armv7', '-t', 'mxnetci/build.armv7', 'docker']' returned non-zero exit status 100.

## To Reproduce
Just run python3 ci/build.py -p armv7

### Steps to reproduce

1. Just run python3 ci/build.py -p armv7

1. Added --force-yes when installing autoconf
2. Added https package just in case authentication was due to https site
3. Realized that autoconf may not be available because jessie distribution may be too old?

## Environment

We recommend using our script for collecting the diagnositc information. Run the following command and paste the outputs below:
```
curl --retry 10 -s https://raw.githubusercontent.com/dmlc/gluon-nlp/master/tools/diagnose.py | python

# paste outputs here
```
Version      : 3.7.5
Compiler     : Clang 11.0.0 (clang-1100.0.33.12)
Build        : ('default', 'Nov  2 2019 10:02:20')
Arch         : ('64bit', '')
Version      : 19.3.1
Directory    : /Volumes/External/pyenv/tvmenv/lib/python3.7/site-packages/pip
Version      : 1.4.1
Directory    : /Volumes/External/pyenv/tvmenv/lib/python3.7/site-packages/mxnet
An error occured trying to import mxnet.
This is very likely due to missing missing or incompatible library files.
  File "<stdin>", line 122, in check_mxnet
AttributeError: module 'mxnet.util' has no attribute 'get_gpu_count'

Platform     : Darwin-19.0.0-x86_64-i386-64bit
system       : Darwin
node         : Catalyst-RD.local
release      : 19.0.0
version      : Darwin Kernel Version 19.0.0: Thu Oct 17 16:17:15 PDT 2019; root:xnu-6153.41.3~29/RELEASE_X86_64
machine      : x86_64
processor    : i386
b'machdep.cpu.brand_string: Intel(R) Core(TM) i7-4850HQ CPU @ 2.30GHz'
b'machdep.cpu.features: FPU VME DE PSE TSC MSR PAE MCE CX8 APIC SEP MTRR PGE MCA CMOV PAT PSE36 CLFSH DS ACPI MMX FXSR SSE SSE2 SS HTT TM PBE SSE3 PCLMULQDQ DTES64 MON DSCPL VMX SMX EST TM2 SSSE3 FMA CX16 TPR PDCM SSE4.1 SSE4.2 x2APIC MOVBE POPCNT AES PCID XSAVE OSXSAVE SEGLIM64 TSCTMR AVX1.0 RDRAND F16C'
b'machdep.cpu.leaf7_features: RDWRFSGS TSC_THREAD_OFFSET BMI1 AVX2 SMEP BMI2 ERMS INVPCID FPU_CSDS MDCLEAR IBRS STIBP L1DF SSBD'
b'machdep.cpu.extfeatures: SYSCALL XD 1GBPAGE EM64T LAHF LZCNT RDTSCP TSCI'
Setting timeout: 10
Timing for MXNet: https://github.com/apache/incubator-mxnet, DNS: 0.0310 sec, LOAD: 0.4920 sec.
Timing for GluonNLP GitHub: https://github.com/dmlc/gluon-nlp, DNS: 0.0005 sec, LOAD: 0.5698 sec.
Timing for GluonNLP: http://gluon-nlp.mxnet.io, DNS: 0.0007 sec, LOAD: 0.0289 sec.
Timing for D2L: http://d2l.ai, DNS: 0.0008 sec, LOAD: 0.3076 sec.
Timing for D2L (zh-cn): http://zh.d2l.ai, DNS: 0.0006 sec, LOAD: 0.1726 sec.
Timing for FashionMNIST: https://repo.mxnet.io/gluon/dataset/fashion-mnist/train-labels-idx1-ubyte.gz, DNS: 0.0007 sec, LOAD: 0.3055 sec.
Timing for PYPI: https://pypi.python.org/pypi/pip, DNS: 0.0089 sec, LOAD: 1.1580 sec.
Timing for Conda: https://repo.continuum.io/pkgs/free/, DNS: 0.0005 sec, LOAD: 0.1895 sec.

larroy:
Fixes #16753

### DIFF
--- a/ci/docker/Dockerfile.build.armv7
+++ b/ci/docker/Dockerfile.build.armv7
@@ -18,7 +18,10 @@
 #
 # Dockerfile to build MXNet for Android ARMv7
 
-FROM mxnetcipinned/dockcross-linux-armv7:11262018
+# The container is pinned for preventing CI failures on updates, swap below to use 
+# the upstream container
+#FROM dockcross/linux-armv7
+FROM mxnetci/dockcross-linux-armv7-pinned
 
 ENV ARCH armv7l
 ENV HOSTCC gcc


larroy:
@leezu

leezu:
Thanks @larroy. The same update needs to be done for the other arches as the failure mode affects all arches. Either in this PR or separate PRs.

ci/docker/Dockerfile.build.armv7
```diff
@@ -18,7 +18,10 @@
 #
 # Dockerfile to build MXNet for Android ARMv7
 
-FROM mxnetcipinned/dockcross-linux-armv7:11262018
+# The container is pinned for preventing CI failures on updates, swap below to use 
+# the upstream container
+#FROM dockcross/linux-armv7
+FROM mxnetci/dockcross-linux-armv7-pinned
```
leezu: Should this container be given another tag besides `latest` https://hub.docker.com/r/mxnetci/dockcross-linux-armv7-pinned/tags

larroy: What do you mean?

leezu: If you only specify the name `mxnetci/dockcross-linux-armv7-pinned`, instead of `mxnetci/dockcross-linux-armv7-pinned:TAG`, docker will always pull the version tagged as `latest`: `mxnetci/dockcross-linux-armv7-pinned:latest`.
But that version changes whenever someone pushes a new version of `dockcross-linux-armv7-pinned` to `mxnetci`. So it becomes impossible to test if an update of `dockcross-linux-armv7-pinned` breaks the CI. Instead all CI runs will be changed immediately.

If instead you add a tag `TAG` for the particular version of the `mxnetci/dockcross-linux-armv7-pinned` container at https://hub.docker.com/r/mxnetci/dockcross-linux-armv7-pinned/tags, our Dockerfiles here on Github can mention the tag `mxnetci/dockcross-linux-armv7-pinned:TAG`. Then, when pushing a new update of `dockcross-linux-armv7-pinned` to `mxnetci` and tagged as `TAG2`, the CI will continue to use the old version until a PR is opened to update to use `TAG2` on github.

larroy: I think that's the idea, that we push the container if it can be built, so we don't have to constantly update this file.

leezu: But then there is no need to use `mxnetci`. Then we can just directly use the `dockcross` repository.
You mentioned that in the past `dockcross` sometimes broke our build. I thought we used pinning to fix that?

marcoabreu: Mxnetci is a fully automatically handled repository while MXNetcipinned is manually curated. In case there's a security breach, the pinned repo wouldn't be impacted. Since the pinned image usually is no longer available at dockcross since they don't have tagging, the image would then be lost if the image gets deleted by the automatic process. Thus, I'd recommend to stay on the separate pinned repo

larroy: The pinned image needs to be maintained automatically. It's a good point with the credentials. We might consider having it in a private pipeline with a different repo to preserve those credentials.

larroy:
@mxnet-label-bot add pr-awaiting-review

larroy:

larroy:
@leezu how do you suggest to proceed to reach consensus?

leezu:
Maybe consensus is the wrong word. Essentially I'm not clear what's the advantage of using `mxnetci/dockcross-X` without version tag and automating the updates to mxnetci compared to directly using the upstream `dockcross`.

leezu:
Regarding Hotfix, `armv7` is not sufficient, we need to update all arches. It's straightforward to update, ie. https://github.com/apache/incubator-mxnet/pull/17031/commits/d3571d1052a07fea38af932a9e0ea7e1b2ad0f5c passes CI

larroy:
Well, why do you think it was pinned in the first place? It happened a few times that dockcross updating their containers broke the CI as we couldn't compile anymore. Otherwise it wouldn't be pinned. Cross compilation is very tricky, so if it starts failing at some point and people don't know how to fix it then the most likely outcome is that the job would be disabled.

In which branch is that commit that you linked? I'm too stupid to find which branch is in with the github UI.

leezu:
@larroy fair enough, but if you don't pin it inside the code-base but rather by pushing the relevant container to the `mxnetcipinned` docker repository automatically, the same breakage will occur.
Or how would you suggest to avoid it?
My understanding is that the only way to avoid it would be to pin it inside the mxnet codebase as done currently?

larroy:
One possible idea is to push the pinned container only if the build works. Or leave it pinned until we have the need to update and is done manually. What do you think is better? or you have a better idea?

leezu:
Yes, I agree the update should be done automatically when it doesn't break the CI build.
Instead of replicating the checking logic in a separate system, we can use the source code to fix the container (via tag) and have a bot to open pull requests whenever the base container is updated.

Then the CI will test the pull request as it normal.

leezu:
As part of #16753, the docker containers have been updated and switched to the upstream version. 
https://github.com/apache/incubator-mxnet/issues/17151 tracks reintroducing pinning.

PR #17069, opened 2019-12-13
fix for number of inputs/outputs for backward custom ops

samskalicky:
## Description ##
The initial custom operator support provides all possible inputs to backward operators: inputs, outputs, and gradients. The user is responsible for compute the correct number of inputs and outputs, and the abstraction logic tells MXNet that the backward operator will have the correct number of inputs.

This fixes the problem where we dont account for the number of gradients in computing the number of inputs for a backward operator. Previously we computed: "num_inputs + num_outputs", but the correct computation should be: "num_inputs + 2*num_outputs" since there is an input gradient for each output argument of the forward operator. 

### Essentials ###

### DIFF
--- a/src/c_api/c_api.cc
+++ b/src/c_api/c_api.cc
@@ -265,7 +265,7 @@ int MXLoadLib(const char *path) {
                            &num_in, &num_out))
       << "Error calling ParseAttrs::num_outputs for custom operator '" << name_str << "'";
 
-      return num_in + num_out;
+      return num_in + 2*num_out;
     };
 
     // lambda function to call infer shape


wkcn:
https://github.com/apache/incubator-mxnet/blob/5f9686af87ecedef12bce6f1025f1deed44aef0a/src/c_api/c_api.cc#L671-L672

What is `num_inputs` and `num_outpus`? Do they have the same meaning in forward and backward?

samskalicky:
> Hi. What is `inputs` and `outpus`? Do they have the same meaning in forward and backward?

In the forward pass, an operator takes its inputs and produces its outputs. In the backward pass, an operator receives an input gradient for each of its outputs and produces an output gradient for each of its inputs. The picture looks something like this:

MXNetOpArgs

In the forward pass the inputs are: A,B and the outputs are: C
In the backward pass the inputs are: dC, A, B, C and the outputs are: dA, dB

In MXNet you can specify which inputs/outputs from the forward function you want to keep for the backward pass. To simplify the first commit of custom op we just made it so that every possible input/output was available during the backward pass. Heres the code where we tell MXNet that we want to have all inputs/outputs available for the backward function:

https://github.com/apache/incubator-mxnet/blob/bbdc1c3627ad5254c049c2bb871ecb4527d7dc14/src/c_api/c_api.cc#L498-L512

grad_reg is an input when registering the FGradient attribute on the forward function:

https://github.com/apache/incubator-mxnet/blob/bbdc1c3627ad5254c049c2bb871ecb4527d7dc14/src/c_api/c_api.cc#L666

So the fix here is to correctly tell MXNet how many inputs and outputs we have for the backward pass. Previously MXNet wasnt checking or erroring out when the numbers didnt match. But @ptrendx is implementing a new feature to check #17049 so this fix will help him continue.

wkcn:
I see. Thanks @samskalicky for the detailed explanation!

PR #17114, opened 2019-12-19
Improve the speed of the pointwise fusion graph pass

### LINKED ISSUE #17105: Bug with fusion
## Description
I'm running the script at https://github.com/dmlc/gluon-nlp/tree/master/scripts/language_model/run_glue.py. The command line to reproduce:
python run_glue.py --task MRPC --gpu 8 --batch_size 32
 
I have the following observation:
1. With model.hybridize()
With mxnet-cu100 >= 1.6.0b20191102, the time cost of the first batch is more than 283s with a single GPU, more than 1000s with 4 GPUs. 
With mxnet-cu100==1.6.0b20191101, the time cost of the first batch is about 11s  with a single GPU.

2. With model.hybridize() and  os.environ['MXNET_USE_FUSION'] = '0'
for both mxnet-cu100 == 1.6.0b201911102 and 1.6.0b20191215, the time cost of the first batch is around 11s.

2. Without hybridize():
The performance are nearly the same.

### Error Message
(Paste the complete error message. Please also include stack trace by setting environment variable `DMLC_LOG_STACK_TRACE_DEPTH=10` before running your script.)

## To Reproduce
(If you developed your own code, please provide a short script that reproduces the error. For existing examples, please provide link.)

### Steps to reproduce

1.
2.

1.
2.

## Environment

We recommend using our script for collecting the diagnositc information. Run the following command and paste the outputs below:
```
curl --retry 10 -s https://raw.githubusercontent.com/dmlc/gluon-nlp/master/tools/diagnose.py | python

# paste outputs here
```

ptrendx:
## Description ##
Fixes #17105 

This PR fixes 2 problems (that together lead to the huge time spent doing fusion reported in #17105).

The first problem was the `get_subsets` function - a part of the pointwise fusion graph pass. It involves a step that ensures there are no cycles created in a graph after the fusion step. It does so in 2 steps:
 - first it looks at each incompatible node and separates its inputs and outputs into disjoint sets
 - then for each node `n` it looks at those separations and creates a set of nodes that are incompatible to be in the same fusion with `n`

Both of those steps were improved:
 - the first step was optimized via observation that if all inputs to an incompatible node `n` are also incompatible, then the separation set produced by that node is strictly smaller than the ones produced by those inputs, so there is no need to create it
 - the second step was greatly optimized by the observation, that when considering which nodes are incompatible for fusion with a given node `n`, if the incompatible node producing the separation set was already part of the set of nodes `n` is incompatible with, then also every node in that separation set was already included there as well. In the repro script from the issue, this change cut the number of insertion trials in the forward fusion pass from ~1M to ~2000.

The other problem fixed in this PR is that due to variable scope of `state_ptr` in `CachedOp::Forward` method, the state of `CachedOp` was taken twice for any context, resulting in 2 calls into graph optimization methods instead of 1.

In my local environment, the time spent on fusion pass in the repro script came down from over 200s to 4.5s on a single GPU.

@leezu @zburning @Caenorst @samskalicky FYI 

### Essentials ###

### Changes ###
- [x] Improved the speed of finding the subsets compatible with fusion
- [x] Fixed a bug in `CachedOp::Forward` method where the shared_ptr to `CachedOpState` was taken twice, resulting in launching the optimization step twice.

### DIFF
--- a/src/executor/pointwise_fusion_pass.cc
+++ b/src/executor/pointwise_fusion_pass.cc
@@ -35,6 +35,7 @@
 #include "../operator/fusion/fused_op-inl.h"
 #include "../operator/fusion/fused_op.h"
 #include "../operator/operator_common.h"
+#include <chrono>
 
 namespace mxnet {
 namespace exec {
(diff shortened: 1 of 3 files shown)


ptrendx:
@zburning Were you able to validate that this fixes the issue you reported?

zburning:
@ptrendx Yes!  I test it on a p3 instance, the first batch costs 11.4s for 4 gpus. Thank you for your quick fix.

PR #17129, opened 2019-12-20
Add support of iterable objects in estimator api

liuzh91:
## Description ##
`fit()` and `evaluate()` methods of the base `Estimator` class support inputs from `iterable Object`.

### Essentials ###


python/mxnet/gluon/contrib/estimator/estimator.py
```diff
@@ -281,7 +281,7 @@ def evaluate(self,
 
         Parameters
         ----------
-        val_data : DataLoader
+        val_data : iterable Object
```
leezu: Should be `iterable object`?

python/mxnet/gluon/contrib/estimator/estimator.py
```diff
@@ -281,7 +281,7 @@ def evaluate(self,
...
             Validation data loader with data and labels.
```
leezu: It's unclear that `data loader` refers to the `gluon.DataLoader`. It would be good to clarify that any iterable object yielding a sequence of batches is acceptable and that the `gluon.DataLoader` is one instance of such iterables.
Same applies to the fit_batch below.

python/mxnet/gluon/contrib/estimator/estimator.py
```diff
@@ -358,10 +359,11 @@ def fit(self, train_data,
...
         """
-        if not isinstance(train_data, DataLoader):
-            raise ValueError("Estimator only support input as Gluon DataLoader. Alternatively, you "
-                             "can transform your DataIter or any NDArray into Gluon DataLoader. "
-                             "Refer to gluon.data.dataloader")
+        try:
+            is_iterable = iter(train_data)
+        except TypeError as te:
+            raise ValueError("Estimator only support input as iterable Object. Alternatively, you "
+                             "can transform your DataIter or any NDArray into iterable Object. ")
```
leezu: It's not very clear. How about still referencing DataLoader as one example?

python/mxnet/gluon/contrib/estimator/estimator.py
```diff
@@ -358,10 +359,11 @@ def fit(self, train_data,
         batch_axis : int, default 0
             Batch axis to split the training data into devices.
         """
-        if not isinstance(train_data, DataLoader):
-            raise ValueError("Estimator only support input as Gluon DataLoader. Alternatively, you "
-                             "can transform your DataIter or any NDArray into Gluon DataLoader. "
-                             "Refer to gluon.data.dataloader")
+        try:
+            is_iterable = iter(train_data)
```
leezu: Calling `iter(train_data)` can have side-effects, such as starting prefetching. See https://github.com/dmlc/gluon-nlp/blob/297e14732849f2b8025de6e2bf80ac05cab82b44/src/gluonnlp/data/stream.py#L341

Thus `iter` must only be invoked once. The enumerate call below invokes `iter` a second time implicitly. If you want to invoke `iter` manually, you need to use a while loop and `next` to iterate over the elements until `StopIteration` is raised.

liuzh91: what about using `hasattr(train_data, '__iter__')` to check if the `train_data` is iterable or not?

leezu: You need to check both. Thus `hasattr(train_data, '__iter__') or hasattr(train_data, '__getitem__')` should be fine.

See 

> An object capable of returning its members one at a time. Examples of iterables include all sequence types (such as list, str, and tuple) and some non-sequence types like dict, file objects, and objects of any classes you define with an `__iter__()` method or with a `__getitem__()` method that implements Sequence semantics.

https://docs.python.org/3/glossary.html

liuzh91: ok, it makes sense to me.

python/mxnet/gluon/contrib/estimator/estimator.py
```diff
@@ -281,20 +281,19 @@ def evaluate(self,
...
             will be added if not specified explicitly.
         """
-        try:
-            is_iterable = iter(val_data)
-        except TypeError as te:
-            raise ValueError("Estimator only support input as Iterable. Alternatively, you "
-                             "can transform your DataIter or any NDArray into an iterable object.")
+        if not hasattr(val_data, '__iter__') and not hasattr(val_data, '__getitem__'):
+            raise ValueError("Estimator only support iterable object as input. For example, you "
+                             "can provide a customized data loader os input")
```
leezu: "os" typo

liuzh91: nice catch

PR #17138, opened 2019-12-21
Interleaved MHA for CPU path

TaoLv:
## Description ##
This PR fills the CPU counterpart of PR https://github.com/apache/incubator-mxnet/pull/16408.
Only FP32 is supported yet.

### Essentials ###

### Changes ###

## Comments ##


TaoLv:
@eric-haibin-lin @pengzhao-intel @Caenorst Please help to review. I copied the UTs from test_operator_gpu.py to test_operator.py but didn't remove the original ones since it seems they have minimal cuda version requirement. Let me know if it's not appropriate. Thanks!

tests/python/unittest/test_operator.py
```diff
@@ -9373,6 +9373,330 @@ def check_random_uniform():
...
+    assert_allclose(att_score_orig, att_score_opti, rtol=1e-2, atol=1e-3)
+    assert_allclose(output_orig, output_opti, rtol=1e-2, atol=1e-3)
+
+    for k in grads_opti.keys():
+        assert(grads_orig[k].dtype == grads_opti[k].dtype)
+        assert(grads_orig[k].shape == grads_opti[k].shape)
+        assert_allclose(grads_orig[k], grads_opti[k], rtol=1e-2, atol=1e-3)
+
+@with_seed(12345)
+def test_multihead_attention_encdec():
```
eric-haibin-lin: do we still need `check_multihead_attention_selfatt ` in test_operator_gpu.py?

src/operator/contrib/transformer.cc
```diff
@@ -122,6 +122,531 @@ static bool InterleavedMatMulEncDecValAttShape(const NodeAttrs& attrs,
   return true;
 }
 
+void strided_batch_sgemm(bool transA, bool transB,
```
eric-haibin-lin: Is it better to move this function to the file with linalg prefix?

src/operator/contrib/transformer.cc
```diff
@@ -122,6 +122,531 @@ static bool InterleavedMatMulEncDecValAttShape(const NodeAttrs& attrs,
...
+                        kv_seq_len,
+                        kv_seq_len * q_seq_len,
+                        beta,
+                        queries_grads,
+                        lead_dim_q,
+                        batch_stride_q,
+                        attn_batches);
+  }
+  if (req[1] != kNullOp) {
+    if (req[1] == kWriteTo) {
```
eric-haibin-lin: Shall we also check req[1] == kWriteInplace?

src/operator/contrib/transformer.cc
```diff
@@ -122,6 +122,531 @@ static bool InterleavedMatMulEncDecValAttShape(const NodeAttrs& attrs,
...
+  const index_t attn_batches      = inputs[2].shape_[0];
+  const index_t q_seq_len         = inputs[2].shape_[1];
+  const index_t embed_dim         = output_lin_kv_dim / 2;
+  const index_t head_dim          = embed_dim / params.heads;
+  const index_t lead_dim_kv       = attn_batches * head_dim * 2;
+  const index_t batch_stride_kv   = 2 * head_dim;
+  const float alpha               = 1.f;
+
+  if (req[0] != kNullOp) {
+    if (req[0] == kWriteTo) {
```
eric-haibin-lin: Shall we also check req[1] == kWriteInplace?

src/operator/contrib/transformer.cc
eric-haibin-lin: Shall we also document why memset is required here

TaoLv:
> could you elaborate "it seems they have minimal cuda version requirement" and why we duplicate the test?

Because of this line: https://github.com/apache/incubator-mxnet/blob/master/tests/python/gpu/test_operator_gpu.py#L2863
But since the oldest cuda version in CI is 9.2 so I think it's safe to remove this check and copy tests to test_operator.py. I will remove the duplicated part in test_operator_gpu.py.

eric-haibin-lin:
Thanks for the reply. Look forward to the update

src/operator/contrib/transformer.cc
TaoLv: I don't expect this function will be widely used except these mha operators. Also linalg* functions need high level abstraction but I'm not sure how to wrap the pointers a/b/c here into Tensor while still keeping the m/n/k and strides arguments, and how to implement the gpu part.

src/operator/contrib/transformer.cc
TaoLv: > Shall we also check req[1] == kWriteInplace?

FInplaceOption is not registered in this backward operator. Do we still need check kWriteInplace here?

TaoLv: > Shall we also document why memset is required here

Reference to the implementation on gpu path: https://github.com/apache/incubator-mxnet/blob/master/src/operator/contrib/transformer.cu#L160.
I thought it relates to the discussion about the bwd_ignore_zero_init flag and gradient accumulation in the original PR. @Caenorst any comments here?

eric-haibin-lin: @Caenorst shall we change it to `req[1] == kWriteTo || req[1] == kWriteInplace ` for GPU code path, too?

pengzhao-intel:
Is this PR good to merge? @TaoLv @eric-haibin-lin

TaoLv:
@eric-haibin-lin Code is re-based. `memset` is needed to pass the unit tests. We can revisit the kWriteInplace check in a follow up PR if we notice any performance problem there. As the FInplaceOption is not enabled for these operators, I think it's safe for not checking it. Let me know what do you think. Thnanks!

szha:
>  But since the oldest cuda version in CI is 9.2 so I think it's safe to remove this check and copy tests to test_operator.py. I will remove the duplicated part in test_operator_gpu.py.

@TaoLv  This is not true for the CD pipeline for cu90. See [this failure](http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/restricted-mxnet-cd%2Fmxnet-cd-release-job/detail/mxnet-cd-release-job/494/pipeline/647#step-1029-log-1633)

TaoLv:
@szha Any suggestion for this case? It's shared for both CPU and GPU test, can we add the cuda requirement decorator to it? Also if we're still releasing cu90 flavor, why it was removed from CI?

PR #17169, opened 2019-12-25
[Numpy] op empty_like, add nan_to_num to dispatch

Alicia1529:
As title

### DIFF
--- a/python/mxnet/ndarray/numpy/_op.py
+++ b/python/mxnet/ndarray/numpy/_op.py
@@ -28,7 +28,7 @@ from ...context import current_context
 from . import _internal as _npi
 from ..ndarray import NDArray
 
-__all__ = ['shape', 'zeros', 'zeros_like', 'ones', 'ones_like', 'full', 'full_like', 'invert', 'delete',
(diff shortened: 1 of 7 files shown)


python/mxnet/ndarray/numpy/_op.py
```diff
@@ -372,6 +372,75 @@ def full_like(a, fill_value, dtype=None, order='C', ctx=None, out=None): # pylin
...
+           [2.0e-323, 2.5e-323, 3.0e-323]])
+    """
+    dtype_list = {None:'None', _np.int8:'int8', _np.uint8:'uint8', _np.int32:'int32',
+                  _np.int64:'int64', _np.float16:'float16', _np.float32:'float32',
+                  _np.float64:'float64', _np.bool_:'bool'}
+    try:
+        dtype = dtype if isinstance(dtype, str) else dtype_list[dtype]
+    except:
+        raise NotImplementedError("Do not support this dtype at this moment")
+    return _npi.empty_like_fallback(prototype, dtype=dtype, order=order, subok=subok, shape=shape)
```
reminisce: Should check `order` and `subok` as well since the finally returned ndarray is an mxnet ndarray which does not feature those parameters.

PR #16089, opened 2019-09-04
[Numpy] Random.choice implemented

xidulu:
## Description ##
As title
https://docs.scipy.org/doc/numpy-1.16.1/reference/generated/numpy.random.choice.html

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] Uniform/Weighted sampling with/without replacement implemented.
- [x] Allow input to be array/symbol. 

## Comments ##
- Weighted sampling has some performance issues currently, as TopK used in weighted sample without replacement is implemented using `sort`.

- Heavily referenced from https://github.com/pytorch/pytorch/pull/18624

PR #17174, opened 2019-12-25
Fix a build bug on aarch64.

TsingWei:
## Description ##
Fix a build bug on aarch64. #16089

### Essentials ###
I finished coding on this PR, just a simple fix.


CMakeLists.txt
```diff
@@ -143,6 +143,7 @@ else(MSVC)
     check_cxx_compiler_flag("-msse2"     SUPPORT_MSSE2)
   else()
     set(SUPPORT_MSSE2 FALSE)
+    add_definitions(-DMSHADOW_USE_SSE=0)
```
TaoLv: Should be moved to L178-L182?
```cmake
if(SUPPORT_MSSE3)
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -msse3")
elseif(SUPPORT_MSSE2)
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -msse2")
else()
    add_definitions(-DMSHADOW_USE_SSE=0)
endif()
```

TsingWei: Right.

Issue #17140, opened 2019-12-21
mx.nd._internal._mul_scalar and other scalar operators give incorrect results when PySide2 is used


nickguletskii:
## Description

Creating an instance of `PySide2.QtCore.QCoreApplication` makes `mx.nd._internal._mul_scalar` and other scalar operators function incorrectly.

## To Reproduce
```python
import argparse

import mxnet as mx
import numpy as np

if __name__ == '__main__':
    parser = argparse.ArgumentParser(description="Demonstrate an MXNet incompatibility with PySide2.")
    parser.add_argument("--import-pyside", dest="import_pyside", action="store_true", default=False)
    parser.add_argument("--create-app", dest="create_app", action="store_true", default=False)
    parser.add_argument("--scalar", type=float, default=0.6)
    args = parser.parse_args()
    if args.import_pyside:
        from PySide2.QtCore import QCoreApplication

        if args.create_app:
            app = QCoreApplication()

    tensor = mx.nd.full(shape=(375, 600, 3), val=255).astype(np.int32)
    tensor = tensor.astype(np.float32)
    scalar = args.scalar
    print("All of the following operations should yield the same result:")
    print("Multiply numpy ndarray by scalar: ", (tensor.asnumpy() * scalar).max())
    print("Multiply MXNet NDArray by MXNet NDArray: ", (tensor * mx.nd.array([[[scalar]]])).max().asscalar())
    print("Multiply MXNet NDArray by float: ", (tensor * scalar).max().asscalar())
```
### Steps to reproduce
(Paste the commands you ran that produced the error.)

1. Install a Python 3.7.2 virtual environment with PySide2 and MXNet.
2. Run the provided script without any arguments. You should see the following results:
```
All of the following operations should yield the same result:
Multiply numpy ndarray by scalar:  153.0
Multiply MXNet NDArray by MXNet NDArray:  153.0
Multiply MXNet NDArray by float:  153.0
```
3. Run the provided script with `--import-pyside`. You should observe the same results.
4. Run the provided script with `--import-pyside --create-app`. The following output will be printed into the console:
```
All of the following operations should yield the same result:
Multiply numpy ndarray by scalar:  153.0
Multiply MXNet NDArray by MXNet NDArray:  153.0
Multiply MXNet NDArray by float:  0.0
```
5.  Run the provided script with `--import-pyside --create-app --scalar 1.6`.The following output will be printed into the console:
```
All of the following operations should yield the same result:
[...]

Issue #16134, opened 2019-09-10
Incorrect subtraction


RomanSteinberg:
## Description
The result of subtraction is influenced by third party code.

## Environment info (Required)

```
----------Python Info----------
Version      : 3.6.8
Compiler     : GCC 8.0.1 20180414 (experimental) [trunk revision 259383
Build        : ('default', 'Jan 14 2019 11:02:34')
Arch         : ('64bit', 'ELF')
------------Pip Info-----------
Version      : 9.0.1
Directory    : /home/deploy/insightface/venv/lib/python3.6/site-packages/pip
----------MXNet Info-----------
Version      : 1.5.0
Directory    : /home/deploy/insightface/venv/lib/python3.6/site-packages/mxnet
Commit Hash   : 75a9e187d00a8b7ebc71412a02ed0e3ae489d91f
Library      : ['/home/deploy/insightface/venv/lib/python3.6/site-packages/mxnet/libmxnet.so']
Build features:
✔ CUDA
✔ CUDNN
✔ NCCL
✔ CUDA_RTC
✖ TENSORRT
✔ CPU_SSE
✔ CPU_SSE2
✔ CPU_SSE3
✔ CPU_SSE4_1
✔ CPU_SSE4_2
✖ CPU_SSE4A
✔ CPU_AVX
✖ CPU_AVX2
✖ OPENMP
✖ SSE
✔ F16C
✖ JEMALLOC
✖ BLAS_OPEN
✖ BLAS_ATLAS
✖ BLAS_MKL
✖ BLAS_APPLE
✔ LAPACK
✖ MKLDNN
✔ OPENCV
✖ CAFFE
✖ PROFILER
✔ DIST_KVSTORE
✖ CXX14
✖ INT64_TENSOR_SIZE
✔ SIGNAL_HANDLER
✖ DEBUG
----------System Info----------
Platform     : Linux-4.18.0-15-generic-x86_64-with-Ubuntu-18.04-bionic
system       : Linux
node         : finapolis
release      : 4.18.0-15-generic
version      : #16~18.04.1-Ubuntu SMP Thu Feb 7 14:06:04 UTC 2019
----------Hardware Info----------
machine      : x86_64
processor    : x86_64
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              8
On-line CPU(s) list: 0-7
Thread(s) per core:  2
Core(s) per socket:  4
Socket(s):           1
[...]

PR #17177, opened 2019-12-26
[1.x] Fix incorrect calculation results when the C locale is set to a locale that uses commas as the decimal separator

nickguletskii:
## Description ##
Currently, many operators utilize `std::stod` to convert strings into floating point numbers. This causes incorrect calculations (#17140, #16134) when the C locale is set to a locale which uses commas (`,`) as decimal separators.

This pull request replaces calls to `std::stod` and `std::stof` to the locale-invariant `dmlc::stod` and `dmlc::stof` respectively.

### The scope of this patch ###

This patch should fix a large portion of interactions through Python and JVM frontends, since they use locale-invariant serialization in order to pass parameters into MXNet's C API.

However, frontends which utilize C locale-aware serialization (i.e. call `sprintf` or similar) may break when using locales which don't use `.` as the decimal separator. They will have to be fixed in a separate patch. I also suspect that they are already broken, because operators utilising dmlc-core parameter parsing were already using locale-invariant serialization.

### Further steps ###
STL streams are heavily used within the codebase, both for serialization and for forming user-friendly messages. Fortunately, they don't seem to be affected by the C standard library locale settings. However, if someone sets the STL locale by calling [std::locale::global](https://en.cppreference.com/w/cpp/locale/locale/global), MXNet's API will be broken. In order to ensure that this doesn't happen, all streams which are used for serialization will have to be imbued with the "C" locale (not the locale set in the C standard library).

It would be nice to see a more principled approach to serialization in MXNet 2.0, e.g. using a binary format for communication between the frontend and the backend. In addition to solving locale-related issues, this would probably result in a smaller invocation overhead.

### Locale-invariant serialization vs locale-aware serialization ###

As a side-note, using locale-aware serialization is not an option, simply because using `,` as the decimal separator adds ambiguities to tuple serialization, e.g. `(4,4,3)` can be a tuple of 3 integers, or a tuple of 2 floats.

### Essentials ###

### Changes ###
- [x] Replace `std::stod` with `dmlc::stod` and `std::stof with `dmlc::stof`.
- [x] Add a test that tests locale invariance for scalar ops.

## Comments ##
- May break Julia and R code when using a locale that uses `,` as the decimal separator. However, since there was no consistency between the various operators before, it is not unlikely that the code was already broken.

### DIFF
--- a/cpp-package/include/mxnet-cpp/optimizer.h
+++ b/cpp-package/include/mxnet-cpp/optimizer.h
@@ -32,6 +32,7 @@
 #include <string>
 #include <memory>
 #include <functional>
+#include <dmlc/strtonum.h>
 #include "mxnet-cpp/base.h"
 #include "dmlc/logging.h"
 #include "mxnet-cpp/ndarray.h"
@@ -84,7 +85,7 @@ class Optimizer {
   Optimizer *SetLRScheduler(std::unique_ptr<LRScheduler> lrScheduler) {
     CHECK(lrScheduler);
     lrScheduler_ = std::move(lrScheduler);
-    lrScheduler_->SetLR(std::stof(params_["lr"]));
+    lrScheduler_->SetLR(dmlc::stof(params_["lr"]));
     return this;
   }
   /*!
--- a/cpp-package/include/mxnet-cpp/optimizer.h
+++ b/cpp-package/include/mxnet-cpp/optimizer.h
@@ -27,12 +27,12 @@
 #ifndef MXNET_CPP_OPTIMIZER_H_
 #define MXNET_CPP_OPTIMIZER_H_
 
+#include <dmlc/strtonum.h>
 #include <map>
 #include <vector>
 #include <string>
 #include <memory>
 #include <functional>
-#include <dmlc/strtonum.h>
 #include "mxnet-cpp/base.h"
 #include "dmlc/logging.h"
 #include "mxnet-cpp/ndarray.h"
--- a/cpp-package/include/mxnet-cpp/optimizer.hpp
+++ b/cpp-package/include/mxnet-cpp/optimizer.hpp
@@ -26,6 +26,7 @@
 #ifndef MXNET_CPP_OPTIMIZER_HPP_
 #define MXNET_CPP_OPTIMIZER_HPP_
 
+#include <dmlc/strtonum.h>
 #include <algorithm>
 #include <utility>
 #include <numeric>
@@ -33,7 +34,6 @@
 #include <cmath>
 #include <string>
 #include <vector>
-#include <dmlc/strtonum.h>
 #include "mxnet-cpp/optimizer.h"
 #include "mxnet-cpp/op.h"
 #include "mxnet-cpp/op_map.h"
--- a/cpp-package/include/mxnet-cpp/optimizer.hpp
+++ b/cpp-package/include/mxnet-cpp/optimizer.hpp
@@ -33,6 +33,7 @@
 #include <cmath>
 #include <string>
 #include <vector>
+#include <dmlc/strtonum.h>
 #include "mxnet-cpp/optimizer.h"
 #include "mxnet-cpp/op.h"
 #include "mxnet-cpp/op_map.h"
@@ -116,11 +117,11 @@ inline float Optimizer::GetLR_(int index) {
   if (nullptr != lrScheduler_) {
     return lrScheduler_->GetLR(num_update_);
   }
-  return std::stof(params_["lr"]);
+  return dmlc::stof(params_["lr"]);
 }
 
 inline float Optimizer::GetWD_(int index) {
-  float wd = std::stof(params_["wd"]);
+  float wd = dmlc::stof(params_["wd"]);
   return wd;
 }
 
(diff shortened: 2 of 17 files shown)


leezu:
@nickguletskii can resolve the conflicts so this PR may be merged?

Are you interested in working on the "more principled approach to serialization in MXNet 2.0, e.g. using a binary format for communication between the frontend and the backend. In addition to solving locale-related issues, this would probably result in a smaller invocation overhead."?

stu1130:
+1 need this fix on MXNet 1.7

ciyongch:
@nickguletskii, as we're doing 1.7 release recently, can you also help to backport this PR to v1.x branch as @stu1130 mentioned, which will be included in MXNet 1.7 as well? Thanks!

nickguletskii:
Sorry for accidentally adding so many reviewers: this was a side effect from rebasing onto the v1.x branch, which didn't exist when I first created this PR.

@leezu Judging by the documentation for the new FFI, it does seem that using it would solve the problem. However, that would require all operators, optimizers and other parts of the codebase to be ported to the new FFI. Therefore, I don't think this patch is necessary on the master branch.

marcoabreu:
Shall we add a check which throws an error in sanity if people use the STD method?

nickguletskii:
@marcoabreu I think it would be better to add a (nightly) CI job that will run the whole test suite with a locale that is vastly different from `en_US.UTF-8`. While this PR solves the issues related to the usage of ``std::stod`` and ``std::stof``, there's no guarantee that there aren't any more subtle locale-related issues that are quietly truncating/changing values inside MXNet.

nickguletskii:
@mxnet-bot run ci windows-cpu

nickguletskii:

marcoabreu:
Sure, but a text based scan is a lot quicker and less costly than running the fully suite. They are not exclusive, but enforcing the use of the dmlc version as part of the sanity check should bring a lot of value for low investment.

leezu:
We already run the testsuite on 2 Linux platforms: Ubuntu and CentOS7. I suggest to switch the locale for the CentOS7 platform instead of introducing another nighlty build. @nickguletskii would you like to help open a PR to the master branch? You'd need to change this Dockerfile https://github.com/apache/incubator-mxnet/blob/master/ci/docker/Dockerfile.build.centos7_cpu https://github.com/apache/incubator-mxnet/blob/master/ci/docker/Dockerfile.build.centos7_gpu

ciyongch:
Adding to 1.7.0 roadmap https://github.com/apache/incubator-mxnet/issues/16864

leezu:
@mxnet-bot run ci [windows-gpu, unix-gpu]

nickguletskii:
> We already run the testsuite on 2 Linux platforms: Ubuntu and CentOS7. I suggest to switch the locale for the CentOS7 platform instead of introducing another nighlty build. @nickguletskii would you like to help open a PR to the master branch? You'd need to change this Dockerfile https://github.com/apache/incubator-mxnet/blob/master/ci/docker/Dockerfile.build.centos7_cpu https://github.com/apache/incubator-mxnet/blob/master/ci/docker/Dockerfile.build.centos7_gpu

Good idea! I've made a pull request that updates the CentOS CI jobs and changes the tests to respect the locale environment variables: #18097

leezu:
@mxnet-bot run ci [windows-gpu, unix-gpu]

ciyongch:
@nickguletskii , can you help to take a look at/solve the failed CI? As the code freeze date is postponed to April 25 PST, please make sure this PR is included both in v1.x and v1.7.x branch as well before the freeze date. Thanks!

stu1130:
@ciyongch seems like the PR is blocked by 
https://github.com/apache/incubator-mxnet/pull/18097 and #18025 @nickguletskii right?

leezu:
@stu1130 these shouldn't be blockers

@mxnet-bot run ci [windows-gpu, unix-gpu]

ciyongch:
@stu1130 thanks for pointing out the blocked items for this PR. I saw the first one (#18097) is inactive for two days, the latter one is keeping active updated. 
Gentle ping to @nickguletskii if we can catch up the freeze date for this feature? Thanks!

nickguletskii:
@ciyongch From what I've seen, the CI failures are from CI timing out. I don't think this PR is causing the CI failures, since the CI seems to hang up on different tests each time...

leezu:
@mxnet-bot run ci [all]

ciyongch:
@nickguletskii thanks for your information, then re-trigger the failed CI would be helpful for merging. Can you help to backport this PR to v1.7.x branch as well?

nickguletskii:
@mxnet-bot run ci [unix-gpu]

leezu:
@mxnet-bot run ci [unix-gpu]

@nickguletskii the deadlock is not deterministic, so until the root cause is fixed / the respective CI tests are disabled (https://github.com/apache/incubator-mxnet/pull/18151) to at least have a stable CI, retriggering will still help.

leezu:
@mxnet-bot run ci [unix-gpu]

ciyongch:
The back ported PR to v1.7.x branch https://github.com/apache/incubator-mxnet/pull/18147 passed CI :)

leezu:
Merge based on the CI results on v1.7.x.

Thanks @nickguletskii

PR #17187, opened 2019-12-27
[MKLDNN] Support channel wise quantization for FullyConnected

ciyongch:
## Description ##
Add channel-wise quantization support for FullyConnected, which will bring accuracy benefit for some models such as BERT SQuAD, etc. 
The default quantization keeps the same as now which is using tensor-wise quantization, user can switch to use channel-wise quantization by setting arguments `quantize_granularity="channel-wise"` when calling `quantize_model()`.

### Essentials ###

### Changes ###

## Comments ##

@pengzhao-intel @TaoLv

### DIFF
--- a/include/mxnet/c_api.h
+++ b/include/mxnet/c_api.h
@@ -1932,6 +1932,7 @@ MXNET_DLL int MXSymbolInferTypePartial(SymbolHandle sym,
  * \param quantized_dtype the quantized destination type for input data
  * \param calib_quantize **Deprecated**. quantize op will always be calibrated if could
  * \param quantize_mode quantize mode to be used in quantize pass
+ * \param quantize_granularity quantize granularity, tensor-wise or channel-wise
  * \param out_num_calib_names return the number of nodes to be calibrated
  * \param out_calib_names return the node names to be calibrated
  */
@@ -1944,8 +1945,8 @@ MXNET_DLL int MXQuantizeSymbol(SymbolHandle sym_handle,
                                const char **excluded_op_names,
                                const uint32_t num_offline, const char **offline_params,
                                const char *quantized_dtype, const bool calib_quantize,
-                               const char *quantize_mode, uint32_t* out_num_calib_names,
-                               const char ***out_calib_names);
+                               const char *quantize_mode, const char *quantize_granularity,
+                               uint32_t* out_num_calib_names, const char ***out_calib_names);
 
 /*!
  * \brief Convert a symbol into a mixed precision symbol with cast operators for target dtype casting
--- a/python/mxnet/contrib/quantization.py
+++ b/python/mxnet/contrib/quantization.py
@@ -112,7 +112,7 @@ def _quantize_symbol(sym, ctx, excluded_symbols=None, excluded_operators=None,
         The mode that quantization pass to apply.
     quantize_granularity: str
         The granularity of quantization, currently supports 'tensor-wise' and 'channel-wise'
-        quantization.
+        quantization. The default value is 'tensor-wise'.
 
     """
     num_excluded_symbols = 0
@@ -465,7 +465,7 @@ def quantize_model(sym, arg_params, aux_params,
                    ctx=cpu(), excluded_sym_names=None, excluded_op_names=None, calib_mode='entropy',
                    calib_data=None, num_calib_examples=None,
                    quantized_dtype='int8', quantize_mode='smart',
-                   quantize_granularity='channel-wise', logger=None):
+                   quantize_granularity='tensor-wise', logger=None):
     """User-level API for generating a quantized model from a FP32 model w/ or w/o calibration.
     The backend quantized operators are only enabled for Linux systems. Please do not run
     inference using the quantized models on Windows for now.
@@ -523,7 +523,7 @@ def quantize_model(sym, arg_params, aux_params,
         'smart' means quantization pass will smartly choice which operator should be quantized.
     quantize_granularity: str
         The granularity of quantization, currently supports 'tensor-wise' and 'channel-wise'
-        quantization.
+        quantization. The default value is 'tensor-wise'.
     logger : Object
         A logging object for printing information during the process of quantization.
 
@@ -553,6 +553,9 @@ def quantize_model(sym, arg_params, aux_params,
     if quantized_dtype not in ('int8', 'uint8', 'auto'):
         raise ValueError('unknown quantized_dtype %s received,'
                          ' expected `int8`, `uint8` or `auto`' % quantized_dtype)
+    if quantize_granularity not in ('tensor-wise', 'channel-wise'):
+        raise ValueError('unkonwn quantize_granularity %s received,',
+                         ' expected `tensor-wise` or `channel-wise`.' % quantize_granularity)
     qsym, calib_layer = _quantize_symbol(sym, ctx, excluded_symbols=excluded_sym_names,
                                          excluded_operators=excluded_op_names,
                                          offline_params=list(arg_params.keys()),
@@ -607,7 +610,8 @@ def quantize_model_mkldnn(sym, arg_params, aux_params,
                           data_names=('data',), label_names=('softmax_label',),
                           ctx=cpu(), excluded_sym_names=None, excluded_op_names=None,
                           calib_mode='entropy', calib_data=None, num_calib_examples=None,
-                          quantized_dtype='int8', quantize_mode='smart', logger=None):
+                          quantized_dtype='int8', quantize_mode='smart',
+                          quantize_granularity='tensor-wise', logger=None):
     """User-level API for generating a fusion + quantized model from a FP32 model
     w/ or w/o calibration with Intel MKL-DNN.
     The backend quantized operators are only enabled for Linux systems. Please do not run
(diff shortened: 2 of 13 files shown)


python/mxnet/contrib/quantization.py
```diff
@@ -109,6 +110,9 @@ def _quantize_symbol(sym, ctx, excluded_symbols=None, excluded_operators=None,
         The quantized destination type for input data.
     quantize_mode: str
         The mode that quantization pass to apply.
+    quantize_granularity: str
+        The granularity of quantization, currently supports 'tensor-wise' and 'channel-wise'
```
pengzhao-intel: how about simplifying to `tensor` and `channel`?

python/mxnet/contrib/quantization.py
```diff
@@ -459,7 +464,8 @@ def quantize_model(sym, arg_params, aux_params,
                    data_names=('data',), label_names=('softmax_label',),
                    ctx=cpu(), excluded_sym_names=None, excluded_op_names=None, calib_mode='entropy',
                    calib_data=None, num_calib_examples=None,
-                   quantized_dtype='int8', quantize_mode='smart', logger=None):
+                   quantized_dtype='int8', quantize_mode='smart',
+                   quantize_granularity='channel-wise', logger=None):
```
pengzhao-intel: Is the default value "channel-wise"?

ciyongch: Good catch, forget to change back to `tensor-wise` as default value after validation :) Will change default value to `tensor-wise`.

python/mxnet/contrib/quantization.py
ciyongch: I thought `tensor` and `channel` is not that descriptive for the argument. There's also `per-axis`/`per-channel` and `per-tensor` naming, which is similar to `channel-wise` and `tensor-wise`. Better to keep this?

pengzhao-intel: OK

ciyongch:
> also add `quantize_granularity` in `quantize_model_mkldnn` api and `quantize_net_v2` api and make original type in `quantize_net` api. GluonNLP will use `quantize_net_v2` api.
> By the way, for the previous quantized models, will they still work with this patch?

Thanks @xinyu-intel,  the new patch was updated to include those changes to support broader user-level quantization API. This patch is backward compatible, so no worry for the existing quantized models.

include/mxnet/c_api.h
```diff
@@ -1944,8 +1945,8 @@ MXNET_DLL int MXQuantizeSymbol(SymbolHandle sym_handle,
                                const char **excluded_op_names,
                                const uint32_t num_offline, const char **offline_params,
                                const char *quantized_dtype, const bool calib_quantize,
-                               const char *quantize_mode, uint32_t* out_num_calib_names,
-                               const char ***out_calib_names);
+                               const char *quantize_mode, const char *quantize_granularity,
+                               uint32_t* out_num_calib_names, const char ***out_calib_names);
```
TaoLv: Break backward compatibility?

ciyongch: This API was called by `_quantize_symbol()` only, and  the new argument `quantize_granularity` carries the default vale of `tensor-wise` from Python to C++ perspective. 
For users doing quantization with `quantization.py`, setting `quantize_granularity` to `channel-wise` explicitly to do channel-wise quantization for FullyConnected, otherwise it will do tensor-wise quantization which is same as before (the default behavior).

src/operator/quantization/quantized_batch_norm.cc
```diff
@@ -135,11 +135,9 @@ NNVM_REGISTER_OP(BatchNorm)
...
     return node;
   })
-.set_attr<FAvoidQuantizeInput>("FAvoidQuantizeInput", [](const NodeAttrs &attrs, size_t index) {
-  if (index == 0)
-    return false;
-  else
-    return true;
+.set_attr<FAvoidQuantizeInput>("FAvoidQuantizeInput", [](
+  const NodeAttrs &attrs, const size_t index, const std::string quantize_granularity) {
+  return (index == 0) ? false : true;
```
TaoLv: `return (index != 0);` ?

src/operator/quantization/quantized_indexing_op.cc
```diff
@@ -181,11 +181,9 @@ NNVM_REGISTER_OP(Embedding)
...
     return node;
   })
-.set_attr<FAvoidQuantizeInput>("FAvoidQuantizeInput", [](const NodeAttrs &attrs, size_t index) {
-  if (index == 0)
-    return true;
-  else
-    return false;
+.set_attr<FAvoidQuantizeInput>("FAvoidQuantizeInput", [](
+  const NodeAttrs &attrs, const size_t index, const std::string quantize_granularity) {
+  return (index == 0) ? true : false;
```
TaoLv: `return (index == 0);` ?

src/operator/subgraph/mkldnn/mkldnn_fc.cc
```diff
@@ -115,84 +151,153 @@ void SgMKLDNNFCOp::Forward(const OpContext &ctx,
...
+      if (mkldnn_param.enable_float_output) {
+        support_channelwise_scale = true;
+      }
+      // channel_wise  support_channelwise_scale  result
+      // True          True                       True
+      // True          False                      Error
+      // False         True/False                 False
+      if (channel_wise && !support_channelwise_scale) {
+        LOG(FATAL)
+          << "Currently, channel-wise quantization requires fuse requantize or dequantize.";
```
TaoLv: Is it something that users may encounter? If so, what kind of action is suggested?

tests/python/mkl/test_subgraph.py
```diff
@@ -767,7 +773,7 @@ def test_pos_conv_bn_sum_act():
               "softrelu": True,
               "relu6": False,
               "leakyrelu": True,
-              "gelu": True}
+              "gelu": False}
```
TaoLv: why change it false?

python/mxnet/contrib/quantization.py
```diff
@@ -978,7 +1004,7 @@ def __exit__(self, exc_type, exc_value, traceback):
         net.collect_params().reset_ctx(ctx)
     return net
 
-def quantize_net(network, quantized_dtype='auto', quantize_mode='full',
+def quantize_net(network, quantized_dtype='auto', quantize_mode='full', quantize_granularity='tensor-wise',
```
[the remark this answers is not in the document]
xinyu-intel: it's better to keep.

src/operator/quantization/quantized_batch_norm.cc
ciyongch: No, the attribute of this op is different than others. Check [here](https://github.com/apache/incubator-mxnet/blob/master/src/operator/quantization/quantized_batch_norm.cc#L139).

python/mxnet/contrib/quantization.py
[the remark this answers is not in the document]
ciyongch: Looks like the only difference between `quantize_net_v2` and `quantize_net` is the new param `LayerOutputCollector` so far, `quantize_granularity` will be another new param introduced by this PR.
As there's not any other difference, how about jsut combine these two API into one?
Otherwise, there could be `v3`, `v4` version in the future if there's new param needed.

src/operator/quantization/quantized_batch_norm.cc
ciyongch: good.

src/operator/quantization/quantized_indexing_op.cc
ciyongch: good.

src/operator/subgraph/mkldnn/mkldnn_fc.cc
ciyongch: This is not the common case, but this would happen in the case of setting `MXNET_DISABLE_MKLDNN_QFC_FLOAT_OUTPUT` or `MXNET_DISABLE_MKLDNN_QFC_FUSE_ALL` to true, which means quantized Fullyconnected will not fused with either `requantize` or `dequantize`.  
So this msg will give a hint to user enable fusion to enable this feature.

tests/python/mkl/test_subgraph.py
ciyongch: Convolution with post sum and activation only supports "relu" currently, previously UT didn't catch such failure.

python/mxnet/contrib/quantization.py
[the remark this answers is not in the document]
TaoLv: I don't know if there is a better way to handle this kind of change. Since `quantize_net` is an user interface and has been released, adding new parameters to it might break user workloads. One convention is to add `v2` implementation, deprecate the original one, and replace it in the next major release. The `v2` implementation is not released yet, so I think it's safe to change it to add `quantize_granularity`. If it's released, yes, we might need `v3` implementation.

src/operator/subgraph/mkldnn/mkldnn_fc.cc
TaoLv: Thanks for the explanation. How about suggesting in the error message to check these two env variables and set correct values?

python/mxnet/contrib/quantization.py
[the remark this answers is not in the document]
ciyongch: That make sense, so let's keep the `quantize_net` api as is, only adding the new params in `v2` version.

src/operator/subgraph/mkldnn/mkldnn_fc.cc
ciyongch: No problem :)

ciyongch:
@TaoLv @ZhennanQin @xinyu-intel please help to review the latest changes :)

TaoLv:
Thank you for the contribution. Merging now~

PR #17098, opened 2019-12-17
Disable OpenMP offloading support for 3rdparty/openmp

leezu:
## Description ##
OpenMP offloading was introduced some time during the past two years and is
enabled by default. With upgrading 3rdparty/openmp in
#17012 it was unintentionally made part of the MXNet CMake build.

But we don't use OpenMP offloading and the Cuda target in the llvm OpenMP
Offloading build is broken in our setting. Thus disable it.

Due to bug in CMake < 3.13, this requires updating to CMake 3.13. See https://cmake.org/cmake/help/latest/policy/CMP0077.html

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [X] Changes are complete (i.e. I finished coding on this PR)
- [X] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [X] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at https://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [X] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [X] Disable OpenMP offloading support for 3rdparty/openmp in CMake build, which was enabled by mistake

PR #17012, opened 2019-12-08
Upgrade 3rdparty/openmp to release_90 version

leezu:
## Description ##
Fixes https://github.com/apache/incubator-mxnet/issues/10856

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [X] Changes are complete (i.e. I finished coding on this PR)
- [X] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [X] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at https://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [X] Upgrade 3rdparty/openmp to release_90 version

## Comments ##
See https://github.com/apache/incubator-mxnet/issues/10856 for a discussion on the symptom. It seems it's due to a bug in llvm openmp fixed in the meantime.

The previously failing assertion is still part of the `release_90` version. It is at https://github.com/llvm-mirror/openmp/blob/release_90/runtime/src/kmp_runtime.cpp#L6616
But the assertion does not fail anymore (based on my test with `cmake -DUSE_CUDA=0 -DUSE_MKLDNN=0 -DCMAKE_EXPORT_COMPILE_COMMANDS=1 -DBUILD_CYTHON_MODULES=1 -DUSE_OPENCV=1 -DCMAKE_BUILD_TYPE=Debug  ..`, which leads to assertion failure without this PR: https://github.com/apache/incubator-mxnet/issues/10856#issuecomment-562485521) 

You can compare the lines in blame mode and conclude they are the same

https://github.com/llvm-mirror/openmp/blame/37c72127e90360a020f351f18d9cccfc30e5145a/runtime/src/kmp_runtime.cpp#L6481
https://github.com/llvm-mirror/openmp/blame/release_90/runtime/src/kmp_runtime.cpp#L6616

PR #17193, opened 2019-12-30
Backport 3rdparty/openmp fixes

leezu:
## Description ##
Backport #17098 #17039 #17012 #17155

The commits contain fixes for the cmake build with openmp enabled.
The cmake build is not used for binary releases on pypi, but the fixes may be helpful for people or organizations building with cmake from source.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0.2)
+cmake_minimum_required(VERSION 3.13)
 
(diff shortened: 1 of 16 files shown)


ptrendx:
@leezu is there a fix for those numpy errors? We need to backport that as well.

leezu:
@ptrendx no real fix yet as far as I'm aware. For now numpy 1.18 is marked as unsupported. I'll include the commit that marks 1.18 as unsupported.

PR #17208, opened 2020-01-03
gather_nd: check bound and wrap negative indices

Alicia1529:
1. check if indices are out of bound 

2. wrap negative indices


src/operator/tensor/indexing_op.cc
```diff
@@ -435,6 +435,63 @@ inline void SparseEmbeddingOpBackwardRspImpl<cpu>(const bool deterministic,
...
+      offset += strides[j] * static_cast<index_t>(indices[j*N + i]);
+    }
+    for (index_t j = 0; j < K; ++j) {
+      KERNEL_ASSIGN(out[i*K + j], req, data[offset+j]);
+    }
+  }
+};
+
+template<typename cpu>
+void GatherNDForward(const nnvm::NodeAttrs& attrs,
```
reminisce: No need to use template function. You can name this `GatherNDForwardCPU`.

src/operator/tensor/indexing_op.cu
```diff
@@ -437,6 +436,98 @@ inline void SparseEmbeddingOpBackwardRspImpl<gpu>(const bool deterministic,
...
+  Kernel<set_zero, gpu>::Launch(s, 1, is_valid_ptr);
+  Kernel<is_valid_check_gather_nd, gpu>::Launch(s, M, is_valid_ptr, data_ptr, N, mshape);
+  CUDA_CALL(cudaMemcpyAsync(&is_valid, is_valid_ptr, sizeof(char),
+                            cudaMemcpyDeviceToHost, mshadow::Stream<gpu>::GetStream(s)));
+  CUDA_CALL(cudaStreamSynchronize(mshadow::Stream<gpu>::GetStream(s)));
+  return is_valid == 0;
+}
+
+template<typename gpu>
+void GatherNDForward(const nnvm::NodeAttrs& attrs,
```
reminisce: Same here. `GatherNDForwardGPU`.

src/operator/tensor/indexing_op.cu
```diff
@@ -437,6 +436,98 @@ inline void SparseEmbeddingOpBackwardRspImpl<gpu>(const bool deterministic,
...
+
+/*! \brief If there are out-of-bound indices, out will be assigned to 1.
+ */
+struct is_valid_check_gather_nd {
+  template<typename DType>
+  MSHADOW_XINLINE static void Map(int i, char* out, const DType* data,
+                                  const index_t N, const mshadow::Shape<10> mshape) {
+    index_t n = N - 1;
+    while (n >= 0) {
+      if (data[i*N + n] < -mshape[i] || data[i*N + n] > mshape[i] - 1) *out = 1;
```
haojin2: You can already break when this if branch evaluates to `True`.

haojin2:
Maybe also add some tests to verify this new checking.

tests/python/unittest/test_operator.py
```diff
@@ -7166,6 +7166,35 @@ def check(data, idx):
...
+            # IndexError: index -5 is out of bounds for axis 1 with size 3
+            pass
+
+    # check if the negative indices are wrapped correctly
+    indices1 = mx.nd.array([[0, 1, -1], [0, 1, -2]])
+    indices2 = mx.nd.array([[0, 1, 1], [0, 1, 1]])
+    data1 = mx.nd.gather_nd(data, indices1)
+    data2 = mx.nd.gather_nd(data, indices2)
+    assert_almost_equal(data1, data2, rtol=1e-5, atol=1e-5)
+
```
haojin2: one more blank line below.

eric-haibin-lin:
@szha we should not mix commits in a PR. It makes it hard to search for the commit that fixes " NormalizeError" APIs

PR #17214, opened 2020-01-05
[tvmop] support cuda multi-arch compilation

yzhliu:
accept the same (multi) arch setting provided by Makefile

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -759,11 +759,15 @@ if(USE_DIST_KVSTORE)
   list(APPEND mxnet_LINKER_LIBS ${pslite_LINKER_LIBS})
 endif()
 
-if(USE_TVM_OP)
-  list(APPEND mxnet_LINKER_LIBS ${CMAKE_CURRENT_BINARY_DIR}/3rdparty/tvm/libtvm_runtime.so)
+function(BuildTVMOP)
+  # scope the variables in BuildTVM.cmake to avoid conflict
   include(cmake/BuildTVM.cmake)
   add_subdirectory("3rdparty/tvm")
+endfunction()
 
+if(USE_TVM_OP)
+  list(APPEND mxnet_LINKER_LIBS ${CMAKE_CURRENT_BINARY_DIR}/3rdparty/tvm/libtvm_runtime.so)
+  BuildTVMOP()
   if(NOT Python3_EXECUTABLE)
     find_package(PythonInterp 3 REQUIRED)
     set(Python3_EXECUTABLE ${PYTHON_EXECUTABLE} CACHE FILEPATH "Path to the python3 executable")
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -840,7 +841,7 @@ install(TARGETS ${MXNET_INSTALL_TARGETS}
 
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/dlpack/include/ DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/dmlc-core/include/ DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
-if(USE_MKLDNN)
+if(INSTALL_MKLDNN)
```
TaoLv: What's the problem with `USE_MKLDNN`?

yzhliu: in cmake/BuildTVM.cmake I set `USE_MKLDNN` to OFF, does it affect this variable used later in root cmake? what is the best approach to scope an cmake option?

TaoLv: I see. So both MXNet and TVM have this `USE_MKLDNN` flag but you don't want to enable it in TVM. Does the `<PARENT_SCOPE>` help here? https://cmake.org/cmake/help/latest/command/set.html#set-normal-variable

yzhliu: actually I need something opposite of `PARENT_SCOPE`

```cmake
option(USE_MKLDNN ON)
if(USE_MKLDNN)
...
endif()
if(USE_TVM_OP)
  include(cmake/BuildTVM.cmake)   <- in BuildTVM.cmake I set(USE_MKLDNN OFF)
  ...
endif()
if(USE_MKLDNN) <- here it is OFF, while should be ON
...
endif()
```

leezu: @yzhliu Yes, parent scope is affected in the current case. To fix this problem, you can make use of CMakes function scopes. Ie define a new function insede the BuildTVM.cmake, and do all variable changes inside that function.

You can take 
https://github.com/apache/incubator-mxnet/blob/3e638b4fc4de959c7afc614facc336a796347c9d/CMakeLists.txt#L407-L414
as a reference

CMakeLists.txt
```diff
@@ -758,11 +759,15 @@ if(USE_DIST_KVSTORE)
   list(APPEND mxnet_LINKER_LIBS ${pslite_LINKER_LIBS})
 endif()
 
-if(USE_TVM_OP)
-  list(APPEND mxnet_LINKER_LIBS ${CMAKE_CURRENT_BINARY_DIR}/3rdparty/tvm/libtvm_runtime.so)
+function(BuildTVMOP)
+  # scope the variables in BuildTVM.cmake to avoid conflict
```
yzhliu: @TaoLv @leezu How about this?

contrib/tvmop/compile.py
```diff
@@ -105,18 +105,13 @@ def get_cuda_arch(arch):
         return arch.split(';')
 
     # the arch string contains '-arch=sm_xx'
+    archs = []
     flags = arch.split()
     for flag in flags:
-        if flag.startswith('-arch='):
-            return flag[len('-arch='):]
+        if flag.startswith('-gencode') or flag.startswith('arch='):
```
leezu: `if flag.startswith('-gencode')` should never be True, as then line 104 would have been True already?

yzhliu: an example input arch string would be
`-gencode arch=compute_30,code=sm_30 -gencode arch=compute_35,code=sm_35 -gencode arch=compute_50,code=sm_50 -gencode arch=compute_52,code=sm_52 -gencode arch=compute_60,code=sm_60 -gencode arch=compute_61,code=sm_61 -gencode arch=compute_70,code=sm_70 -gencode arch=compute_75,code=[sm_75,compute_75] --fatbin-options -compress-all`

L104 has semicolon ; after "gencode", I believe it's for single arch compilation @reminisce if I understand correctly.

leezu: L104 is added for the use-case that the `arch` parameter in `get_cuda_arch(arch)` is based on cmake's `CUDA_SELECT_NVCC_ARCH_FLAGS`. But that should always have a semicolon. Is the example input arch string based on the Makefile setup?

reminisce: I believe the example comes from the Makefile setup:

https://github.com/apache/incubator-mxnet/blob/83a23b0988b55c1483f5ecec941207357492bcff/Makefile#L429-L442

yzhliu: sorry I'm a bit confused, should I remove L102-L105?

leezu: It seems CMake and make based builds pass inputs of two different formats to the `get_cuda_arch(arch)`. Before removing one condition in `get_cuda_arch(arch)`, the Makefile based build would need to be adapted to use `;` as a separator. `;` is used internally by cmake to represent lists.

For now I think it's fine to keep both, because later we anyways plan to remove the Makefile based build

yzhliu:
Thanks @reminisce @leezu @TaoLv for help.

contrib/tvmop/compile.py
```diff
@@ -99,24 +99,16 @@ def get_cuda_arch(arch):
...
-    if arch.startswith('-gencode;'):
-        return arch.split(';')
-
-    # the arch string contains '-arch=sm_xx'
-    flags = arch.split()
+    # an example of arch string,
+    # -gencode arch=compute_30,code=sm_30 -gencode arch=compute_35,code=sm_35
+    # -gencode;arch=compute_75,code=[sm_75,compute_75] --fatbin-options -compress-all
+    archs = []
+    flags = arch.replace("-gencode;", "-gencode ").split()
```
leezu: Should this be `arch.replace(";", " ").split()`? Consider your example of `gencode arch=compute_30,code=sm_30 -gencode arch=compute_35,code=sm_35`. The equivalent case for cmake is `gencode;arch=compute_30,code=sm_30;-gencode;arch=compute_35,code=sm_35`.

Alternatively you can change the CMakeList `    set(TVM_OP_COMPILE_OPTIONS "${TVM_OP_COMPILE_OPTIONS}" "--cuda-arch" "\"${CUDA_ARCH_FLAGS}\"")` to `    set(TVM_OP_COMPILE_OPTIONS "${TVM_OP_COMPILE_OPTIONS}" "--cuda-arch" "\"${CUDA_ARCH_FLAGS_SPACES}\"")`

leezu: @yzhliu

Issue #17219, opened 2020-01-06
Flaky test: test_numpy_op.test_npi_boolean_assign


TaoLv:
## Description
test_numpy_op.test_npi_boolean_assign

## Occurrences
```
MXNET_TEST_SEED=952277814 nosetests -s -v test_numpy_op:test_npi_boolean_assign
test_numpy_op.test_npi_boolean_assign ... [INFO] Setting test np/mx/python random seeds, use MXNET_TEST_SEED=952277814 to reproduce.
((3, 4), (3, 4), 0)
((3, 0), (3, 0), 0)
((), (), 0)
((2, 3, 4, 5), (2, 3), 0)
ERROR

======================================================================
ERROR: test_numpy_op.test_npi_boolean_assign
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/lvtao/miniconda3/envs/mxnet/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/lvtao/Workspace/mxnet-official/tests/python/unittest/common.py", line 221, in test_new
    orig_test(*args, **kwargs)
  File "/home/lvtao/miniconda3/envs/mxnet/lib/python3.6/site-packages/mxnet/util.py", line 321, in _with_np_shape
    return func(*args, **kwargs)
  File "/home/lvtao/miniconda3/envs/mxnet/lib/python3.6/site-packages/mxnet/util.py", line 505, in _with_np_array
    return func(*args, **kwargs)
  File "/home/lvtao/Workspace/mxnet-official/tests/python/unittest/test_numpy_op.py", line 1327, in test_npi_boolean_assign
    np_data[np_mask, :, :] = val
ValueError: shape mismatch: value array of shape (0,) could not be broadcast to indexing result of shape (0,4,5)
-------------------- >> begin captured logging << --------------------
common: INFO: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=952277814 to reproduce.
root: INFO: NumPy-shape semantics has been activated in your code. This is required for creating and manipulating scalar and zero-size tensors, which were not supported in MXNet before, as in the official NumPy library. Please DO NOT manually deactivate this semantics while using `mxnet.numpy` and `mxnet.numpy_extension` modules.
--------------------- >> end captured logging << ---------------------

----------------------------------------------------------------------
Ran 1 test in 0.052s

FAILED (errors=1)
```

## What have you tried to solve it?
None

[...]

PR #17222, opened 2020-01-06
fix flaky test: boolean index and fix bugs

Alicia1529:
1. fix the problem in issue #17219: 
As the variable val is a mxnet.numpy.array, the operation `np_data[np_mask] = val` actually fall back to mxnet operation
2. handle boolean indexing on 0-d array and index as primitive boolean type. For example:
```
>>> from mxnet import np, npx
>>> npx.set_np()
>>> import numpy as _np
>>> mx_a = np.array(1)
>>> np_a = _np.array(1)
>>> mx_a[True]
array([1.])
>>> mx_a[False]
array([])
>>> np_a[True]
array([1])
>>> np_a[False]
array([], dtype=int64)
```

### DIFF
--- a/tests/python/unittest/test_numpy_op.py
+++ b/tests/python/unittest/test_numpy_op.py
@@ -1296,12 +1296,12 @@ def test_npi_boolean_assign():
 
     for hybridize in [False]:
         for config in configs:
-            print(config)
             dshape, mshape, start_axis = config
             test_data = np.random.uniform(size=dshape)
-            mx_mask = np.around(np.random.uniform(size=mshape))
-            valid_num = int(mx_mask.sum())
-            np_mask = mx_mask.asnumpy().astype(_np.bool)
+            np_mask = _np.random.choice(a=[False, True], size=mshape)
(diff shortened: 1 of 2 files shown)


tests/python/unittest/test_numpy_op.py
```diff
@@ -1296,10 +1296,9 @@ def hybrid_forward(self, F, a, mask, value):
 
     for hybridize in [False]:
         for config in configs:
-            print(config)
             dshape, mshape, start_axis = config
             test_data = np.random.uniform(size=dshape)
-            mx_mask = np.around(np.random.uniform(size=mshape))
+            mx_mask = np.around(np.random.uniform(size=mshape)).astype(np.bool)
```
Kh4L: Nitpicking:

```suggestion
            mx_mask = np.random.choice(a=[False, True], size=mshape)
```

Alicia1529: Thanks for your suggestions. I added this change just now : )

Issue #17226, opened 2020-01-06
Compile from Source instructions re-use CI install scripts


leezu:
## Description
The "compile from source" instructions re-use CI install scripts. But the CI scripts contain hacky modifications that shouldn't be applied to user systems silently. For example

https://github.com/apache/incubator-mxnet/blob/83a23b0988b55c1483f5ecec941207357492bcff/ci/docker/install/ubuntu_core.sh#L55-L66

We don't have any logic to undo such system changes when the user uninstall mxnets, neither is it a good practice to do such system changes in a install script whose audience are novice users..

Issue #17180, opened 2019-12-26
Need a better build script for CMake


apeforest:
## Description
The current instruction for users to build from source using Cmake is not very user-friendly. Users have to remember each flag whenever they build mxnet.

It will be nice to have a simple setup.py and users only need to run the following as Pytorch:
```
python setup.py install
```

leezu:
We also need to handle other language bindings. But yes, the user experience can be improved.

apeforest:
Closing the issue as the current cmake build step has been greatly simplified using a config.cmake file thanks to @leezu

PR #17229, opened 2020-01-06
Fix and clean up Ubuntu build from source instructions

leezu:
## Description ##
Fix and clean up Ubuntu build from source instructions.

Fixes https://github.com/apache/incubator-mxnet/issues/17226

### Essentials ###

### Changes ###
- [X] Simplify and fix Ubuntu build from source instructions

## Comments ##
Justification for respective removals
- Binary Python install instructions: 1) The instructions were outdated and incomplete 2) This is a "Build From Source Guide". Why include binary package install instructions? 3) There are better instructions for the binary python install at the Get Started Page. Thus link to that page instead.
- `ci/docker/install/ubuntu_{core,python}.sh` "quick" install: 1) https://github.com/apache/incubator-mxnet/issues/17226 2) Only supports Ubuntu 16.04
- `./install_mxnet_ubuntu_python.sh`: Does not exist anymore
- `dev_menu.py build`: The `dev_menu.py build` does things differently compared to the simple build instructions in this tutorial. For example, it uses virtualenv and contains other unrelated features such as building docker containers. I think it adds confusion for new users if we include it in the setup guide. (CC @larroy)

### DIFF
--- a/docs/static_site/src/pages/get_started/ubuntu_setup.md
+++ b/docs/static_site/src/pages/get_started/ubuntu_setup.md
@@ -22,19 +22,26 @@ permalink: /get_started/ubuntu_setup
 <!--- specific language governing permissions and limitations -->
 <!--- under the License. -->
 
-# Installing MXNet on Ubuntu
+# Installing MXNet from source on Ubuntu
 
-The following installation instructions are for installing MXNet on computers running **Ubuntu 16.04**. Support for later versions of Ubuntu is [not yet available](#contributions).
-<hr>
+The following installation instructions are for building MXNet from source on
+computers running **Ubuntu 16.04** or higher. For instructions to build MXNet
+from source on other platforms, see the general [Build From Source
+guide](build_from_source).
+
+Instead of building from source, you can install a binary version of MXNet. For
+that, please follow the information at [Get Started](get_started).
+
+Building MXNet from source is a two-step process:
+
+1. Build the shared library from the MXNet C++ source code.
+2. (optional) Install the supported language-specific packages for MXNet.
 
 ## Contents
 
 * [CUDA Dependencies](#cuda-dependencies)
-* [Quick Installation](#quick-installation)
-    * [Python](#install-mxnet-for-python)
-    * [pip Packages](#pip-package-availability)
-* [Build from Source](#build-mxnet-from-source)
-* [Installing Language Packages](#installing-language-packages-for-mxnet)
+* [Build the MXNet shared library from source](#build-mxnet-from-source)
+* [Install Language Packages](#installing-language-packages-for-mxnet)
     * [R](#install-the-mxnet-package-for-r)
     * [Julia](#install-the-mxnet-package-for-julia)
     * [Scala](#install-the-mxnet-package-for-scala)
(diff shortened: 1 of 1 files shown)


szha:
Is the verification of this instruction not automated? I thought we had automatic verification for it which wouldn't allow it to be broken

leezu:
There were multiple separate instructions. One of them are the CI scripts and those were verified. Other's were not verified and one of them was broken (`./install_mxnet_ubuntu_python.sh` doesn't exist anymore).

However, the CI scripts contain CI specific code and shouldn't be used on user systems.

larroy:
@szha I'm not aware of such automation, I think there's something for notebooks / tutorials, not for installation instructions. I could be wrong.

docs/static_site/src/pages/get_started/ubuntu_setup.md
```diff
@@ -234,13 +140,13 @@ If building on CPU and using MKL and MKL-DNN (make sure MKL is installed accordi
...
-If building on GPU (make sure you have installed the [CUDA dependencies first](#cuda-dependencies)):
-Cuda 10.1 in Ubuntu 18.04 builds fine but is not currently tested in CI.
+For a GPU-enabled build make sure you have installed the
+[CUDA dependencies first](#cuda-dependencies)) and run:
 
 ```bash
     rm -rf build
     mkdir -p build && cd build
-    cmake -GNinja \
+    ~/.local/bin/cmake -GNinja \
```
apeforest: Is this path to cmake required?

apeforest:
Thanks a lot for cleaning up this instruction. The current one is so confusing and not working as well. 

One more thing is that I found current CMake still not working perfectly.
e.g. if I add flag -DUSE_NCCL, it does not work. I have to use `make` if I want to use NCCL with CUDA. 
Also, it's not very user-friendly to type all different flags of cmake. It would be nice to have a setup.py like pytorch.

I have a feature request here: https://github.com/apache/incubator-mxnet/issues/17180

docs/static_site/src/pages/get_started/ubuntu_setup.md
leezu: Yes, because we use pip install --user to install an up-to-date cmake. Often ~/.local/bin will be part of user's PATH. But it's not guaranteed and not true on default Ubuntu installation. So it's safer to specify the complete path.

leezu:
@ptrendx This PR may be nice to backport to 1.6. What do you think?

ptrendx:
Website lives kind of on its own anyway, so there is no need to include this change in 1.6 package for it to apply to 1.6 release. I would be against including things that are not essential as it is very easy to just get into this loop of adding new shiny things one after another and never actually releasing the thing.

leezu:
@ptrendx isn't the website built from 1.6 branch? I agree it's not necessary to include the change in the release, but could be good to include for the website.

docs/static_site/src/pages/get_started/ubuntu_setup.md
```diff
@@ -147,67 +81,35 @@ On Ubuntu versions 16.04 or later, you need the following dependencies:
...
-Build with CMake and ninja, without GPU and without MKL.
+**Step 3:** Build MXNet core shared library.
+
+For a CPU-only build with OpenBLAS math library run:
 
 ```bash
     rm -rf build
     mkdir -p build && cd build
-    cmake -GNinja \
+    ~/.local/bin/cmake -GNinja \
```
apeforest: Somehow I found using this absolute path causes inconvenience as my 'cmake' is not installed on this path, nor is the default one on AWS DLAMI. Do you have a better idea to not using a fixed path?

larroy: it would be better to just use CMake and specify what's the minimum required version

leezu: @larroy some users may not know how to understand cmake.
@apeforest DLAMI encourages users to use conda, which causes problems for above installation steps.

I'll open a PR to improve the experience shortly

PR #17230, opened 2020-01-06
Additional fix for vector access. 

aws-taylor:
See https://github.com/apache/incubator-mxnet/commit/9634786f96388004f68c223d72e120ad425c2f12 for the original.

## Description ##
Fixes an assertion thrown by operator[] of std::vector when MxNet is compiled with certain STL hardening flags, especially since GCC 8. This pull request is to fix a small block that was missed in the original PR.

### DIFF
--- a/3rdparty/mshadow/mshadow/dot_engine-inl.h
+++ b/3rdparty/mshadow/mshadow/dot_engine-inl.h
@@ -421,12 +421,9 @@ struct BLASEngine<cpu, double> {
   CBLAS_TRANSPOSE p_transa[GROUP_SIZE] = {cblas_a_trans};
   CBLAS_TRANSPOSE p_transb[GROUP_SIZE] = {cblas_b_trans};
 
-  std::vector<const double*> pp_A;
-  std::vector<const double*> pp_B;
-  std::vector<double*> pp_C;
-  pp_A.reserve(batch_count);
-  pp_B.reserve(batch_count);
-  pp_C.reserve(batch_count);
+  std::vector<const double*> pp_A(batch_count, nullptr);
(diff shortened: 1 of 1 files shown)


3rdparty/mshadow/mshadow/dot_engine-inl.h
```diff
@@ -421,12 +421,9 @@ struct BLASEngine<cpu, double> {
   CBLAS_TRANSPOSE p_transa[GROUP_SIZE] = {cblas_a_trans};
   CBLAS_TRANSPOSE p_transb[GROUP_SIZE] = {cblas_b_trans};
 
-  std::vector<const double*> pp_A;
-  std::vector<const double*> pp_B;
-  std::vector<double*> pp_C;
-  pp_A.reserve(batch_count);
-  pp_B.reserve(batch_count);
-  pp_C.reserve(batch_count);
+  std::vector<const double*> pp_A(batch_count, nullptr);
```
wkcn: Thank you for the fix!
It will be better to use `std::array`, since `std::vector` will allocate the memory, which is greater than `batch_count`.

aws-taylor: I did this because it's precisely the same logic used by the earlier commit.

That being said, I believe the STL guarantees that non-class types will be zero initialized when using the constructor form you propose, so I would expect the performance of the two forms to be identical.

wkcn:
Hi @aws-taylor , I could not retrigger the CI.
Could you please retrigger it (push an empty commit)? Thank you!

wkcn:
The PR has been merged. Thank you!

PR #17255, opened 2020-01-09
[DO NOT REVIEW]Set np default dtype( float32 <-> float64)

JiangZhaoh:
## Description ##
This is only a draft and need to change a lot.
Please don't review it until I totally finish it.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/include/mxnet/imperative.h
+++ b/include/mxnet/imperative.h
@@ -46,6 +46,7 @@ namespace mxnet {
    *   turn off numpy shape flag globally.
    * */
   enum NumpyShape{Off, ThreadLocalOn, GlobalOn};
+  typedef NumpyShape NumpyDefaultDtype;
 /*! \brief runtime functions for NDArray */
 class Imperative {
  public:
@@ -136,6 +137,34 @@ class Imperative {
     }
     return old;
   }
+  /*! \brief return current ndarray default dtype,
+   *  float32(false), float64(true).
+   * */
+  int is_np_default_dtype() const {
+    if (is_np_default_dtype_global_) {
+      return 2;
+    }
+    return is_np_default_dtype_thread_local_ ? 1 : 0;
+  }
+  /*! \brief specify ndarray default dtype off, thread local on or global on. */
+  bool set_is_np_default_dtype(int is_np_default_dtype) {
+    NumpyDefaultDtype flag = static_cast<NumpyDefaultDtype>(is_np_default_dtype);
+    bool old = this->is_np_default_dtype();
+    switch (flag) {
+      case GlobalOn:
+        is_np_default_dtype_global_ = true;
+        is_np_default_dtype_thread_local_ = true;
+        break;
+      case ThreadLocalOn:
+        is_np_default_dtype_thread_local_ = true;
+        break;
+      case Off:
+        is_np_default_dtype_global_ = false;
+        is_np_default_dtype_thread_local_ = false;
+        break;
+    }
+    return old;
+  }
   /*! \brief to record operator, return corresponding node. */
   void RecordOp(nnvm::NodeAttrs&& attrs,
                 const std::vector<NDArray*>& inputs,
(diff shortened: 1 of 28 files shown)


include/mxnet/imperative.h
```diff
@@ -136,6 +137,34 @@ class Imperative {
     }
     return old;
   }
+  /*! \brief return current ndarray default dtype,
+   *  float32(false), float64(true).
+   * */
+  int is_np_default_dtype() const {
+    if (is_np_default_dtype_global_) {
+      return 2;
```
marcoabreu: Could you use an enum instead?

include/mxnet/imperative.h
```diff
@@ -136,6 +137,34 @@ class Imperative {
...
+    }
+    return is_np_default_dtype_thread_local_ ? 1 : 0;
+  }
+  /*! \brief specify ndarray default dtype off, thread local on or global on. */
+  bool set_is_np_default_dtype(int is_np_default_dtype) {
+    NumpyDefaultDtype flag = static_cast<NumpyDefaultDtype>(is_np_default_dtype);
+    bool old = this->is_np_default_dtype();
```
marcoabreu: Type mismatch

include/mxnet/imperative.h
```diff
@@ -136,6 +137,34 @@ class Imperative {
...
+        break;
+      case ThreadLocalOn:
+        is_np_default_dtype_thread_local_ = true;
+        break;
+      case Off:
+        is_np_default_dtype_global_ = false;
+        is_np_default_dtype_thread_local_ = false;
+        break;
+    }
+    return old;
```
marcoabreu: ?

include/mxnet/imperative.h
JiangZhaoh: Sure.

JiangZhaoh:
> This PR changes so many things and it feels odd. I'm afraid it will cause quite some confusion when people explicitly set the type, change the default and then wonder when it's not being used - the UX wouldn't be great
Could you please tell me how can I finish this function if I don't change these files? I know some files maybe don't need to change, but I think most of them definitely need to be modified to assure consistency. If you have better idea, please explain it and I'm willing to change my method.

JiangZhaoh:
> I don't know... Somehow this change feels wrong to me - a default is a default and people can always specify the dtype if they would like to have a particular one. This PR changes so many things and it feels odd. I'm afraid it will cause quite some confusion when people explicitly set the type, change the default and then wonder when it's not being used - the UX wouldn't be great

Could you please tell me how can I finish this function if I don't change these files? I know some files maybe don't need to change, but I think most of them definitely need to be modified to assure consistency. If you have better idea, please explain it and I'm willing to change my method.

marcoabreu:
My question is rather towards whether the functionality itself will provide a good user experience - the implementation is of secondary nature at this point.

haojin2:
@marcoabreu 
If you explicitly specified some dtype in an op with a `dtype` option, it means that you're not expecting to see whatever the default dtype is, isn't that a very natural thing? I'm not seeing any UX problem following this kind of logic.
This new default type option is part of the new NumPy-compatible user experience for a closer compatibility with existing code written in NumPy, a lot of which pose the assumptions that array creation functions return arrays with float64 data type by default, and failure to meet such assumption could lead to crashes and unexpected numerical bugs. For a more smooth transition for those code to NP on MXNet, providing such an option instead of having to change all array creation functions in those code to explicitly say `dtype=np.float64` would, I assume, be providing a better UX.

marcoabreu:
Okay, so there's a functionality in numpy to set the default dtype or is this approach chosen as a compromise to not alter the array creation functions while still offering the possibility to have  parity with numpy?

haojin2:
@marcoabreu
Let's probably take a look at this example to help you understand:
```python
import numpy as np
a = np.ones((4096, 1024)) / 1000000
# original code, a is of type float64
```
```python
from mxnet import np, npx
npx.set_np()
# transitioning to NP on MXNet, swapping the library
a = np.ones((4096, 1024)) / 1000000
# a is of type float32, which might not be able to accurately represent very small numbers, could possibly cause some precision errors in later computations
a = np.ones((4096, 1024), dtype=np.float64) / 1000000
# to recover the same precision, need to touch the original code to explicitly specify the dtype,
# number of lines changed is the same as number of array creation functions used with in the whole file.
```
```python
from mxnet import np, npx
npx.set_np()
# transitioning to NP on MXNet, swapping the library
npx.set_np_default_type()
# switch the default type, this function call could be further merged to be part of set_np function
a = np.ones((4096, 1024)) / 1000000
# now a is of type float64, not interfering with the original assumption of precisions without any changes to the original code.
# only one more line is added, not the same as the total # of array creations in the script.
```

marcoabreu:
Thanks for elaborating, makes sense.

Merging it into set_np sounds like a good addition, indeed.

I know that the code is not ready for review yet, I still have a gut feeling that we are approaching this from a strange angle. Sometimes, the dtype parameter is removed, sometimes it's set to float, sometimes to none and there are probably more cases. 

Do we have a concept how exactly we'd like to have this introduced? Right now it feels a bit cluttered and too many things are changed at the same time in different ways.

Not trying to criticize here, just trying to give a bit of early input from an external view.

haojin2:
@marcoabreu It's really case by case for each op, it should only affect array creation ops, and this is not the final version yet. I think we need to clean up the code that assign the default types in the python frontend to go with this newly-defined behavior. Thanks for you input

PR #17265, opened 2020-01-10
Add bfloat16 floating-point format support based on AMP 

rongzha1:
## Description ##
Bfloat16 is wildly used in Deep Learning especially on training to get a better performance.

This PR is to add bf16 support based on mxnet AMP(Automatic Mixed Precision) module .

### Essentials ###
- [done ] Changes are complete (i.e. I finished coding on this PR)
- [done ] All changes have test coverage:
- [ done] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

## Comments ##

This PR has passed unitest and preci test in local machine.
Unit tests are added for this PR.

@ZhennanQin @ElaineBao @xinyu-intel @TaoLv @PatricZhao


.gitmodules
```diff
@@ -6,7 +6,7 @@
 	url = https://github.com/dmlc/ps-lite
 [submodule "3rdparty/dlpack"]
 	path = 3rdparty/dlpack
-	url = https://github.com/dmlc/dlpack
+	url = https://github.com/ElaineBao/dlpack.git
```
ptrendx: Will need to be changed once the changes are merged to upstream dlpack. Leaving this comment as a reminder ;-).

3rdparty/mshadow/mshadow/base.h
```diff
@@ -312,6 +338,11 @@ enum TypeFlag {
   kInt8  = 5,
   kInt64 = 6,
   kBool = 7,
+  kInt16 = 8,
+  kUint16 = 9,
+  kUint32 = 10,
+  kUint64 = 11,
```
ptrendx: Why adding those additional types here? No operator supports them anyway, right?

3rdparty/mshadow/mshadow/bfloat.h
```diff
@@ -0,0 +1,167 @@
...
+    return RTYPE(static_cast<float>(a) OP float(b));  /* NOLINT(*) */     \
+  }                                                                       \
+  MSHADOW_BF16_OPERATOR_TYPE(float, float, OP)                            \
+  MSHADOW_BF16_OPERATOR_TYPE(double, double, OP)                          \
+  MSHADOW_BF16_OPERATOR_TYPE(float, int8_t, OP)                           \
+  MSHADOW_BF16_OPERATOR_TYPE(float, uint8_t, OP)                          \
+  MSHADOW_BF16_OPERATOR_TYPE(float, int32_t, OP)                          \
+  MSHADOW_BF16_OPERATOR_TYPE(float, uint32_t, OP)                         \
+  MSHADOW_BF16_OPERATOR_TYPE(float, int64_t, OP)                          \
+  MSHADOW_BF16_OPERATOR_TYPE(float, uint64_t, OP)
```
ptrendx: Returning float or double, while understandable, is a different behavior to the one currently done for half_t type. Could we discuss this and make them consistent?

example/quantization/imagenet_inference.py
```diff
@@ -99,7 +100,34 @@ def score(sym, arg_params, aux_params, data, devs, label_name, max_num_examples,
...
+        elif model_name.find('vgg') != -1:
+            excluded_sym_names += ['vgg0_conv0_fwd']
+        elif model_name.find('squeezenet1') != -1:
+            excluded_sym_names += ['squeezenet0_conv0_fwd']
+        elif model_name.find('mobilenet') != -1 and model_name.find('v2') == -1:
+            excluded_sym_names += ['mobilenet0_conv0_fwd']
+        elif model_name.find('mobilenet') != -1 and model_name.find('v2') != -1:
+            excluded_sym_names += ['mobilenetv20_conv0_fwd']
+        elif model_name.find('inceptionv3') != -1:
+            excluded_sym_names += ['inception30_conv0_fwd']
```
ptrendx: Why? Is there an accuracy issue without those exclusions?

python/mxnet/contrib/amp/amp.py
```diff
@@ -43,14 +44,17 @@
 from ... import optimizer as opt
 from .loss_scaler import LossScaler
 
+bfloat16 = np.dtype([('bfloat16', np.uint16)])
```
ptrendx: Can we have this dtype accessible (as `mx.bfloat16` or something similar)?

python/mxnet/gluon/parameter.py
```diff
@@ -289,11 +289,18 @@ def _load_init(self, data, ctx, cast_dtype=False, dtype_source='current'):
...
+                    "Failed loading Parameter '%s' from saved params: " \
+                    "dtype incompatible expected %s vs saved %s. " \
+                    "Set cast_dtype=True to cast the dtype of saved params."%(
+                        self.name, str(self.dtype), str(data.dtype))
+                else:
+                    assert np.dtype(self.dtype).type == data.dtype, \
+                    "Failed loading Parameter '%s' from saved params: " \
+                    "dtype incompatible expected %s vs saved %s. " \
+                    "Set cast_dtype=True to cast the dtype of saved params."%(
+                        self.name, str(self.dtype), str(data.dtype))
```
ptrendx: Aren't those 2 codepaths the same?

3rdparty/mshadow/mshadow/base.h
ZhennanQin: This is to align the definition with DLPack. Otherwise we have to preserve those numbers. Even through we don't use them currently, it's no harm to add them.

3rdparty/mshadow/mshadow/bfloat.h
ZhennanQin: Sure. Any suggestion here?

example/quantization/imagenet_inference.py
ZhennanQin: Not for accuracy, but for performance purpose. This could be removed once more bfloat16 hardware added.

python/mxnet/contrib/amp/amp.py
ZhennanQin: This is a good topic, and I want to have a discussion for this. 
Currently, MXNet doesn't have its own type system. It's simply using Numpy.dtype. Numpy doesn't natively support bfloat16, so we define bfloat16 as a numpy customized type.
Pros: compatible with current design, isinstance(bfloat16, np.dtype) could return True.
cons: bfloat16.name doesn't work, have to use bfloat16.names[0] instead.
Another solution is, creating MXNet's own data type system, just like pytorch and tf. This is a big API change, so we wish this can be done when upgrading to MXNet 2.0.

Currently, we prefer this approach to enable bfloat16 in MXNet 1.x, and refactor it in MXNet 2.0.

include/mxnet/ndarray.h
```diff
@@ -770,6 +770,12 @@ class NDArray {
    */
   NDArray Reorder2Default() const;
 
+    /*
+   * This creates a new NDArray using f32 with the reordered data.
+   * It doesn't affect the data of the original NDArray.
+   */
+  NDArray Reorder2DefaultFp32() const;
```
eric-haibin-lin: Adding dtype specific interface looks very ad-hoc

python/mxnet/ndarray/ndarray.py
```diff
@@ -83,6 +84,7 @@
     5: np.int8,
     6: np.int64,
     7: np.bool_,
+    12: np.dtype([('bfloat16', np.uint16)]),
```
eric-haibin-lin: why 12?

src/engine/naive_engine.cc
```diff
@@ -55,7 +55,7 @@ class NaiveEngine final : public Engine {
     std::vector<VarHandle> const_vars;
     std::vector<VarHandle> mutable_vars;
     FnProperty prop;
-    const char* opr_name;
+    std::string opr_name;
```
eric-haibin-lin: Was this an issue blocking this feature? just curious

python/mxnet/ndarray/ndarray.py
ElaineBao: This is to align the `TypeFlag` defined in mshadow

include/mxnet/ndarray.h
rongzha1: OK, will change it to Reorder2DefaultFloatFormat()

.gitmodules
pengzhao-intel: Definitely :) We're working on PR the related code in dlpack https://github.com/dmlc/dlpack/issues/45

src/engine/naive_engine.cc
ZhennanQin: This is a bugfix for naive engine. If my remember is correct, MXNet already had this fixed with different approach. So we can drop this change from this PR.

eric-haibin-lin: I don't think it's merged yet. I'm ok with the change here.

ElaineBao:
> For bfloat16 training, which loss scalar is recommended? Do we also need to perform NaN checks?

Bfloat16 has the same dynamic range as float32, since they have the same exponent bits. So it can represent gradients directly, it doesn't require loss scaling like fp16.

example/quantization/imagenet_inference.py
pengzhao-intel: Please add a comment for this temp performance solution and will convert all conv layers later.

3rdparty/mshadow/mshadow/base.h
```diff
@@ -988,6 +1034,7 @@ struct minimum {
 };
 }  // namespace red
 
+#ifndef __NVCC__
```
ptrendx: I don't like it - can we make a similar thing as in the case of fp16 for CPUs that did not support F16C instructions (i.e. code that runs but may be slower than the code for hardware natively supporting bfloat16)?

ptrendx: You can implement `atomicAdd` (which seems to be the problem you are facing)with `atomicCAS` like this: https://github.com/apache/incubator-mxnet/blob/master/src/common/cuda_utils.h#L702-L721

eric-haibin-lin:
@ElaineBao thanks for the explanation

3rdparty/mshadow/mshadow/base.h
ZhennanQin: We don't have enough background / knowledge to enable Bfloat16 on GPU side. So probably we can't make the change you proposed. Alternately, any code refactoring on GPU side is welcome. you may change this in following PR.

python/mxnet/gluon/parameter.py
rongzha1: np.dtype(self.dtype) is different from np.dtype(self.dtype).type
https://docs.scipy.org/doc/numpy/reference/generated/numpy.dtype.type.html
https://docs.scipy.org/doc/numpy/reference/generated/numpy.dtype.html#numpy.dtype

.gitmodules
```diff
@@ -6,7 +6,7 @@
 	url = https://github.com/dmlc/ps-lite
 [submodule "3rdparty/dlpack"]
 	path = 3rdparty/dlpack
-	url = https://github.com/dmlc/dlpack
+	url = https://github.com/dmlc/dlpack.git
```
xinyu-intel: keep same.

src/operator/nn/fully_connected.cc
```diff
@@ -237,7 +238,7 @@ static bool BackwardFCStorageType(const nnvm::NodeAttrs& attrs,
   bool dispatched = false;
   if (!dispatched && common::ContainsOnlyStorage(*in_attrs, mxnet::kDefaultStorage)) {
     dispatched = storage_type_assign(out_attrs, mxnet::kDefaultStorage,
-                                     dispatch_mode, DispatchMode::kFCompute);
+                                     dispatch_mode, DispatchMode::kFComputeEx);
```
xinyu-intel: we may need to enable dnnl fc bwd in another PR since there is an known issue.

.gitmodules
rongzha1: OK

src/operator/nn/fully_connected.cc
rongzha1: Thanks for reminder

pengzhao-intel:
@ptrendx thanks for your review.  Feel free to let me know if you have other concerns, we are going to merge PR recently.

TaoLv:
Hi @leezu, @larroy, this PR can pass the builds for ARM but always hit the time out of the test. 
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fedge/detail/PR-17265/29/pipeline

We don't have any environment to reproduce. Could you please take a look or have any suggestion for further debugging?

3rdparty/mshadow/mshadow/bfloat.h
```diff
@@ -0,0 +1,167 @@
+/*!
```
TaoLv: @szha Do we need Apache license header for this new file?

szha: yes

zhreshold:
@larroy Do you have idea how to display more logs for the edge tests? It consistently fail at this stage.

3rdparty/mshadow/mshadow/bfloat.h
rongzha1: done  thanks

pengzhao-intel:
I am merging now. If there're any other comments, we can resolve by new PR.

Issue #17006, opened 2019-12-07
[RFC] Custom Operator Part 2


samskalicky:
## Description
Request for comments on the next PR for enhancing custom operator support

Heres some suggestions from the initial PR (Part 1):
- custom GPU operators
- Random number generator resource request
- sparse data types
- migrate lambda functions in MXLoadLib in src/c_api/c_api.cc to classes defined elsewhere
- Documentation, add the "library" python package to the namespace to the doc: https://mxnet.apache.org/api/python/docs/api/ ?

## References
- initial PR (Part 1): #15921

wkcn:
Hi @samskalicky , thank you for the contribution!
I have several suggestions.

- custom GPU operators
  1. Provide CUDA stream in `OpResource`.
  2. Share the same function on CPU and GPU.
      Users can discriminate the context by `MXTensor::dltensor::ctx`
- Call framework specific math helper
  It is important for a custom operator. Users may call gemm, even convolution op in custom op.

Thanks.

rondogency:
Need to include a fix for the test error https://github.com/apache/incubator-mxnet/pull/15921#pullrequestreview-328686634

larroy:
@wkcn could you explain your suggestion? calling gemm back into the framework which gets dispatched to GPU or CPU?

samskalicky:
We should create a namespace for the stuff in the lib_api.h file as suggested by @larroy:
https://github.com/apache/incubator-mxnet/pull/15760/files#r311756416

wkcn:
@larroy Users may need matrix operators and DNN Op(e.g. ReLU, Conv) when writing a custom Op. Although they can implement it by third-party libraries, it is more convenient to use the built-in functions in MXNet.

ptrendx:
Custom ops should be able to set the inplace property.

kpuatamazon:
Speed.  All those `std::string` and `std::unordered_map` objects don't come cheaply.  

I compared an integrated fork with a custom operator.  

https://github.com/kpuatamazon/incubator-mxnet/tree/intgemm integrated version end-to-end Sockeye performance (based on 1.6.0):
```
real	2m57.962s
user	7m3.986s
sys	0m6.724s
```
Custom operator version (based on 1.7.x. because it had to be for custom operators):
```
real	3m16.879s
user	7m43.727s
sys	0m8.273s
```
Conditions:
[...]

PR #17270, opened 2020-01-11
Dynamic custom operator GPU support

rondogency:
## Description ##
Add GPU support for custom operators. 
This is a continuation of custom operators project, initial CPU support is implemented here: #15921

## Design ##
Working backward from the user. Here is an example of custom operator forward function for GPU. 

Notice the function interface is the same as CPU operator. The input and output tensors are already in GPU, so you don't need to memcpy them to GPU. 

You need the CUDA stream from OpResource in order to launch your CUDA kernel on the correct GPU, and mx_stream_t is defined as cudaStream_t when you compile the code with NVCC.
```
MXReturnValue forwardGPU(std::map<std::string, std::string> attrs,
                         std::vector<MXTensor> inputs,
                         std::vector<MXTensor> outputs,
                         OpResource res) {
    ... 
    mx_stream_t cuda_stream = res.get_cuda_stream();
    gpu_forward<<<grid, block, 0, cuda_stream>>>(out_data, in_data, N);
    return MX_SUCCESS;
}
```

Then user will register a single operator with both CPU and GPU computation logic, by specifying the device type in register function. This registration works for any context as user only passes in a string.
```
REGISTER_OP(my_op)
.setForward(forwardCPU, "cpu")
.setForward(forwardGPU, "gpu")
.setBackward(backwardCPU, "cpu")
.setBackward(backwardGPU, "gpu");
```

Then user compiles the library similar to compiling the CPU operator library. Python usage is exactly the same as CPU custom operators
```
import mxnet as mx
mx.library.load(os.path.abspath('libmy_op_lib.so'))
a = mx.nd.array([[1,2,3],[4,5,6]], ctx=mx.gpu())
b = mx.nd.array([[7],[8],[9]], ctx=mx.gpu())
mx.nd.my_op(a,b)
```

## This PR is not ##
- Supporting CPU function is one library and GPU function in another library. Both CPU and GPU functions have to be in the same library.
- Making inferShape and inferType context aware
- Supporting dynamic loading of custom context

### Essentials ###

### Changes ###
- Add Fcompute<gpu> registration, and pass NDArray context to custom library in c_api.cc
- Add context info to MXTensor class in lib_api.h
- Add lib_custom_op/relu.cu example file containing full registration of custom operator "my_relu", and add both CPU and GPU kernel functions in that file
- Modify lib_custom_op/Makefile to compile .cu file using nvcc to custom library

## Comments ##


src/c_api/c_api.cc
```diff
@@ -720,8 +751,11 @@ int MXLoadLib(const char *path) {
         gradOp.set_attr<bool>("TIsLayerOpBackward", true, plevel);
         gradOp.set_attr<FStatefulComputeEx>("FStatefulComputeEx<cpu>",
                                             fstateful_backward, plevel);
+        gradOp.set_attr<FStatefulComputeEx>("FStatefulComputeEx<gpu>",
```
eric-haibin-lin: What is the target supported contexts for this feature? Do we target just cpu and gpu, or we want to support other hardware backends, too?

Currently the dispatch logic is inside FCompute, which is a bit different from existing mxnet users' experience. Usually the FCompute only declares the computation, and leave the dispatch logic to MXNet executor. And it's unclear how it supports the case where the same op is extended by a library for Intel CPUs and NVIDIA GPUs - they may hard-code the dispatch logic to only care about their own hardware. How do we handle such conflicts? 

Furthermore, currently the infer_shape/infer_dtype is not context-aware, i.e. CPU and GPU infers the same dtype. However, it may not be true (e.g. cpu supports fp32 and bfloat16, and gpu supports fp32 and fp16). How do we handle these attribute conflict? 

I had a short discussion with @yzhliu and we saw two potential fixes:
1. make infer_shape/infer_dtype context aware. This way we can have different infer_dtype function for cpu & gpu. MXNet needs to dispatch to the function based on the current context. For example, `op.set_attr<FInferType>("FInferType<cpu>", my_infer_type_function)` for cpu specific type inference, and `op.set_attr<FInferType>("FInferType<gpu>", my_infer_type_function_gpu)`for gpu. 
2. Another way is to register ops with different names (e.g. 'cpu_gemm' and 'gpu_gemm'). This way they can have different infer_attr functions. But we don't want users to modify their model definition in the training script to these names. To mitigate that we can have an API to allow user to provide a mapping (e.g. {'gemm' -> 'cpu_gemm'}) for mxnet to map an op to another op registered in the backend. 

Finally, is there a plan to support dynamic custom context? :P @samskalicky

samskalicky: > What is the target supported contexts for this feature? Do we target just cpu and gpu, or we want to support other hardware backends, too?

Yes, lets just target the currently supported contexts. We can swing back around later when we add support dynamic loading of contexts. We'll make sure to implement something generic in the PR (ie. setting context with string rather than enum) so it will just work. 

> Currently the dispatch logic is inside FCompute, which is a bit different from existing mxnet users' experience. Usually the FCompute only declares the computation, and leave the dispatch logic to MXNet executor. And it's unclear how it supports the case where the same op is extended by a library for Intel CPUs and NVIDIA GPUs - they may hard-code the dispatch logic to only care about their own hardware. How do we handle such conflicts?
 
Good catch. We'll change it so that users can specify context and Forward/Backward function in the registration. But, for custom operators we can only support what the current implementation in MXNet allows. Which is that the top level scope is an operator, and it has implementations for different contexts.

What you're describing is top level being context and inside of that having a distinct operator registration. Like you describe next, this organization is not supported in MXNet. We should discuss this as part of a separate feature enhancement than this PR (the scope of this PR is to add GPU support to custom operators -- only). 

> Furthermore, currently the infer_shape/infer_dtype is not context-aware, i.e. CPU and GPU infers the same dtype. However, it may not be true (e.g. cpu supports fp32 and bfloat16, and gpu supports fp32 and fp16). How do we handle these attribute conflict?
> 
> I had a short discussion with @yzhliu and we saw two potential fixes:
> 
> 1. make infer_shape/infer_dtype context aware. This way we can have different infer_dtype function for cpu & gpu. MXNet needs to dispatch to the function based on the current context. For example, `op.set_attr<FInferType>("FInferType<cpu>", my_infer_type_function)` for cpu specific type inference, and `op.set_attr<FInferType>("FInferType<gpu>", my_infer_type_function_gpu)`for gpu.
> 2. Another way is to register ops with different names (e.g. 'cpu_gemm' and 'gpu_gemm'). This way they can have different infer_attr functions. But we don't want users to modify their model definition in the training script to these names. To mitigate that we can have an API to allow user to provide a mapping (e.g. {'gemm' -> 'cpu_gemm'}) for mxnet to map an op to another op registered in the backend.

This is a good idea, we should have a separate PR to add this new feature to MXNet backend, and then extend it out to custom operators. But this is out of scope for this PR. 

Another point is that current custom operator support does not allow for registering CPU implementations in one library, and GPU implementations for the same operator in another. This merging of functionality from different libraries is a good idea for a future feature.

> Finally, is there a plan to support dynamic custom context? :P @samskalicky

I'll add it to the list behind: custom data loaders, graph passes, etc... :D

eric-haibin-lin: Are you in favor of approach 1 or approach 2? They have different indication for the recommended nnvm_registration API for custom ops

samskalicky: I think that I support approach 1.

samskalicky: I prefer approach 1, lets not do name mangling :D

include/mxnet/lib_api.h
```diff
@@ -215,25 +225,43 @@ struct MXTensor {
...
+    else if (ctx.dev_type == "vulcan")
+      dltype = kDLVulkan;
+    else if (ctx.dev_type == "metal")
+      dltype = kDLMetal;
+    else if (ctx.dev_type == "vpi")
+      dltype = kDLVPI;
+    else if (ctx.dev_type == "rocm")
+      dltype = kDLROCM;
+    else
+      dltype = kDLExtDev;
```
samskalicky: should this be else or "ext"?

include/mxnet/lib_api.h
```diff
@@ -566,16 +609,20 @@ typedef MXReturnValue (*createOpState_t)(std::map<std::string, std::string>,
...
-    isSGop(false) {}
-  ~CustomOp() {}
-  CustomOp& setForward(fcomp_t fcomp) {
-    forward = fcomp;
+    parse_attrs(NULL), infer_type(NULL), infer_shape(NULL), mutate_inputs(NULL),
+    create_opstate(NULL), isSGop(false) {}
+  CustomOp& setForward(fcomp_t fcomp, std::string ctx) {
+    char* cstr = new char[ctx.length()+1];
+    strncpy(cstr, ctx.c_str(), ctx.length()+1);
+    forward_ctx_cstr.push_back(cstr);
```
samskalicky: i think we should use const char* here like we do for the op_name

include/mxnet/lib_api.h
```diff
@@ -602,12 +649,18 @@ class CustomOp {
...
   }
+  ~CustomOp() {}
 
   /*! \brief operator name */
   const char* name;
+
+  /*! \brief each fcomp function is associated with a context speficied by a string*/
+  std::vector<const char*> forward_ctx_cstr;
+  std::vector<fcomp_t> forward_fp;
+  std::vector<const char*> backward_ctx_cstr;
```
samskalicky: should this be a map?

include/mxnet/lib_api.h
```diff
@@ -982,17 +1045,22 @@ extern "C" {
     // create a vector of tensors for inputs
     std::vector<MXTensor> inputs(num_in);
     for (int i = 0; i < num_in; i++) {
-      inputs[i].setTensor(indata[i], (MXDType)intypes[i], inshapes[i], indims[i], inIDs[i]);
+      std::string ctx_str(indev_type[i]);
+      MXContext inctx = {ctx_str, indev_id[i]};
+      inputs[i].setTensor(indata[i], (MXDType)intypes[i], inshapes[i], indims[i],
```
samskalicky: string can be initialized with char* http://www.cplusplus.com/reference/string/string/string/ (see constructor 4)

can we just inline everything and not create intermediary variables?
```
inputs[i].setTensor(indata[i], (MXDType)intypes[i], inshapes[i], indims[i], inIDs[i], {ctx_str, indev_id[i]});
```

src/c_api/c_api.cc
```diff
@@ -97,9 +97,31 @@ inline int MXAPIGetFunctionRegInfo(const FunRegType *e,
 }
 
 // NOTE: return value is added in API_END
+/*!
+ * \brief Convert NDArray to C type arrays for passing to custom library
+ */
+void NDArrayToCTypeArray(const std::vector<NDArray>& inputs,
```
samskalicky: here we're converting a vector of NDArrays to vectors of C types, so shouldnt the name be "NDArrayToCTypes"?

src/c_api/c_api.cc
```diff
@@ -450,32 +480,59 @@ int MXLoadLib(const char *path) {
...
-                              const std::vector<OpReqType>& req,
-                              const std::vector<NDArray>& outputs) {
-      return fcomp_lambda(fcomp_fp, attrs, ctx, inputs, req, outputs);
+    auto forward_cpu_lambda = [=](const nnvm::NodeAttrs& attrs,
+                                  const OpContext& ctx,
+                                  const std::vector<NDArray>& inputs,
+                                  const std::vector<OpReqType>& req,
+                                  const std::vector<NDArray>& outputs) {
+      CHECK(forward_ctx_map.count("cpu") > 0) << "CPU Forward function is not implemented";
+      return fcomp_lambda(forward_ctx_map.at("cpu"), attrs, ctx, inputs, req, outputs);
```
samskalicky: so this implictly copies the entire forward_ctx_map object into the lambda. do we want to do this?

src/c_api/c_api.cc
```diff
@@ -635,11 +698,20 @@ int MXLoadLib(const char *path) {
       CHECK(state_op_inst != nullptr)
       << "Error MXNet cannot load custom stateful operator'" << name_str << "'";
 
+      // pass the gpu stream associated with the context to custom library
+      void* gpu_stream = nullptr;
+      if (inputs[i].ctx().dev_mask() == Context::kGPU) {
```
samskalicky: what is `i` here for `inputs[i]`? shouldnt this be `inputs[0]`?

src/c_api/c_api.cc
```diff
@@ -450,32 +480,59 @@ int MXLoadLib(const char *path) {
         return ptr;
       };
 
+      // pass the gpu stream associated with the context to custom library
+      void* gpu_stream = nullptr;
+      if (inputs[i].ctx().dev_mask() == Context::kGPU) {
```
samskalicky: same as before for fstateful

src/c_api/c_api.cc
```diff
@@ -685,28 +757,26 @@ int MXLoadLib(const char *path) {
...
+      regOp.set_attr<FResourceRequest>("FResourceRequest", DefaultSubgraphOpResourceRequest, plevel);
+      regOp.set_attr<nnvm::FMutateInputs>("FMutateInputs", DefaultSubgraphOpMutableInputs, plevel);
     }
 
     // optionally add stateful forward
     if (create_opstate_fp != nullptr) {
       regOp.set_attr<FCreateOpState>("FCreateOpState", create_opstate, plevel);
       regOp.set_attr<FStatefulComputeEx>("FStatefulComputeEx<cpu>",
                                         fstateful_forward, plevel);
+      regOp.set_attr<FStatefulComputeEx>("FStatefulComputeEx<gpu>",
```
samskalicky: shouldnt we only register this conditionally?

src/c_api/c_api.cc
```diff
@@ -685,28 +757,26 @@ int MXLoadLib(const char *path) {
...
+                                        fstateful_forward, plevel);
     } else {
-      regOp.set_attr<FComputeEx>("FComputeEx<cpu>", forward_lambda, plevel);
+      regOp.set_attr<FComputeEx>("FComputeEx<cpu>", forward_cpu_lambda, plevel);
+      regOp.set_attr<FComputeEx>("FComputeEx<gpu>", forward_gpu_lambda, plevel);
```
samskalicky: shouldnt we only register this conditionally?

include/mxnet/lib_api.h
```diff
@@ -763,9 +818,9 @@ typedef int (*opCallInferType_t)(inferType_t, const char* const*, const char* co
 
 #define MXLIB_OPCALLFCOMP_STR "_opCallFCompute"
 typedef int (*opCallFComp_t)(fcomp_t, const char* const*, const char* const*, int,
-                             const int64_t**, int*, void**, int*, size_t*, int,
-                             const int64_t**, int*, void**, int*, size_t*, int,
-                             xpu_malloc_t, void*);
+                             const int64_t**, int*, void**, int*, size_t*, char**, int*, int,
+                             const int64_t**, int*, void**, int*, size_t*, char**, int*, int,
+                             xpu_malloc_t, void*, void*);
```
samskalicky: lets refactor and use the same function signature as the actual function later on

include/mxnet/lib_api.h
```diff
@@ -833,20 +891,26 @@ extern "C" {
...
-    *name = op.name;
-    *fcomp = op.forward;
-    *fgrad = op.backward;
-    *parse = op.parse_attrs;
-    *type = op.infer_type;
-    *shape = op.infer_shape;
-    *mutate = op.mutate_inputs;
-    *create_op = op.create_opstate;
-    *isSGop = op.isSGop;
+    CustomOp *op = &(Registry<CustomOp>::get()->get(idx));
```
samskalicky: maybe try:
```
    CustomOp& op = Registry<CustomOp>::get()->get(idx);
```

samskalicky: or change 
```
 T* get(int idx) {
    return entries[idx];
  }
```

src/c_api/c_api.cc
samskalicky:
```
if(forward_ctx_map.count("cpu") > 0) {
    fcomp_t fcomp = forward_ctx_map.at("cpu");
    auto forward_cpu_lambda = [=](const nnvm::NodeAttrs& attrs,
                                  const OpContext& ctx,
                                  const std::vector<NDArray>& inputs,
                                  const std::vector<OpReqType>& req,
                                  const std::vector<NDArray>& outputs) {
      return fcomp_lambda(fcomp, attrs, ctx, inputs, req, outputs);
 } ;
}

File: src/c_api/c_api.cc (hunk 6 above)
Comment (rondogency) [reply src/c_api/c_api.cc]: agree

File: include/mxnet/lib_api.h (hunk 3 above)
Comment (rondogency) [reply include/mxnet/lib_api.h]: changed string in parameter to be const char *, as user gonna pass in a static string, in this way we can avoid a redundant copy

File: include/mxnet/lib_api.h [hunk 14]
```diff
@@ -740,68 +787,81 @@ class Registry {
...
-                                        const char* const*, int,
-                                        char***, char***, int*);
+typedef int (*partCallAcceptSubgraph_t)(acceptSubgraph_t acceptSubgraph, const char *json,
+                                        int subgraph_id, int *accept, const char* const* opt_keys,
+                                        const char* const* opt_vals, int num_opts,
+                                        char*** attr_keys, char*** attr_vals, int *num_attrs);
 
 #define MXLIB_INITIALIZE_STR "initialize"
-typedef int (*initialize_t)(int);
+typedef int (*initialize_t)(int version);
```
samskalicky: thanks for putting all the names, this makes the code much more readable/maintainable

src/c_api/c_api.cc
```diff
@@ -164,39 +186,60 @@ int MXLoadLib(const char *path) {
...
+             &createop_ctx, &createop_fp, &createop_count,
+             &parse_fp, &type_fp, &shape_fp, &mutate_fp);
+
+    // construct maps of context to forward/backward custom library function
+    std::unordered_map<std::string, fcomp_t> forward_ctx_map;
+    std::unordered_map<std::string, fcomp_t> backward_ctx_map;
+    std::unordered_map<std::string, createOpState_t> createop_map;
+    for (int i=0; i<forward_count; i++) {
+      std::string ctx_str(forward_ctx[i]);
+      forward_ctx_map[ctx_str] = forward_fcomp[i];
```
samskalicky: should we check that `ctx_str` was already in `forward_ctx_map` and error out if we find it (since that means the user registered the same context twice)?

src/c_api/c_api.cc
```diff
@@ -408,24 +452,13 @@ int MXLoadLib(const char *path) {
...
-        out_data.push_back(outputs[i].data().dptr_);
-        out_shapes.push_back(outputs[i].shape().data());
-        out_dims.push_back(outputs[i].shape().ndim());
-        out_types.push_back(outputs[i].dtype());
-        out_verIDs.push_back(outputs[i].version());
-      }
+      NDArrayToCTypes(inputs, in_data, in_shapes, in_dims, in_types,
+                      in_verIDs, in_dev_type, in_dev_id);
+      NDArrayToCTypes(outputs, out_data, out_shapes, out_dims, out_types,
+                      out_verIDs, out_dev_type, out_dev_id);
```
samskalicky: this is a great refactor! saves space and adds clarity!!

src/c_api/c_api.cc
```diff
@@ -563,11 +587,21 @@ int MXLoadLib(const char *path) {
...
       // create a pointer to hold custom op state object
+      // only create one stateful op depending on passing context
+      // user can add new supported context and call to custom library
       void* state_op_inst = nullptr;
-      CHECK(callCreateOpState(create_opstate_fp, attr_keys.data(), attr_vals.data(),
-                              attr_keys.size(), &state_op_inst))
-      << "Error calling CreateOpState for custom operator '" << name_str << "'";
-
+      if (ctx.dev_mask() == Context::kCPU) {
+        CHECK(createop_map.count("cpu") > 0) << "CPU CreateOpState not implemented";
```
samskalicky: this is good, but can we made the error message more clear by adding the operator name at the end so the message is more like: CPU CreateOpState is not implemented for operator zzzzz

src/c_api/c_api.cc
rondogency: I did the context dedup check in lib_api.h to catch the error early

src/c_api/c_api.cc
rondogency: actually it doesn't save space, so I did a bigger change refactor lol

src/c_api/c_api.cc
rondogency: agree

rondogency:
@samskalicky @eric-haibin-lin appreciate another review. now custom library can allocate both gpu and cpu memory, and all dispatch logics are in mxnet

example/extensions/lib_custom_op/relu_lib.cu
```diff
@@ -0,0 +1,195 @@
...
+MXReturnValue forwardGPU(std::map<std::string, std::string> attrs,
+                         std::vector<MXTensor> inputs,
+                         std::vector<MXTensor> outputs,
+                         OpResource res) {
+    float* in_data = inputs[0].data<float>();
+    float* out_data = outputs[0].data<float>();
+
+    // test on memory resource allocation
+    void *workspace_cpu = res.alloc_cpu(8 * sizeof(float));
+    void *workspace_gpu = res.alloc_gpu(8 * sizeof(float));
```
samskalicky: its great that you're testing calling these functions, but you dont do anything with the allocated space. lets either use it for something, or not alloc at all.

example/extensions/lib_custom_op/relu_lib.cu
```diff
@@ -0,0 +1,195 @@
...
+}
+
+MXReturnValue backwardGPU(std::map<std::string, std::string> attrs,
+                          std::vector<MXTensor> inputs,
+                          std::vector<MXTensor> outputs,
+                          OpResource res) {
+    float* in_data = inputs[0].data<float>();
+    float* out_data = outputs[0].data<float>();
+
+    cudaStream_t cuda_stream = reinterpret_cast<cudaStream_t>(res.get_cuda_stream());
```
samskalicky: should we just have `get_cuda_stream` return a cudaStream_t type? We can use `#ifdef __NVCC__` around the code to protect when the user doesnt compile with nvcc (see [reference here](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#nvcc-identification-macro)) and just use `void*` type in the `#else` block.

It will simplify the user's code from:
```
cudaStream_t cuda_stream = reinterpret_cast<cudaStream_t>(res.get_cuda_stream());
```
to
```
cudaStream_t cuda_stream = res.get_cuda_stream();
```

example/extensions/lib_custom_op/relu_lib.cu
```diff
@@ -0,0 +1,195 @@
...
+.setParseAttrs(parseAttrs)
+.setInferType(inferType)
+.setInferShape(inferShape)
+.setForward(forwardCPU, "cpu")
+.setForward(forwardGPU, "gpu")
+.setBackward(backwardCPU, "cpu")
+.setBackward(backwardGPU, "gpu");
+
+
+
```
samskalicky: can we just reduce this down to 1 extra empty line?

example/extensions/lib_custom_op/relu_lib.cu
```diff
@@ -0,0 +1,195 @@
...
+REGISTER_OP(my_state_relu)
+.setParseAttrs(parseAttrs)
+.setInferType(inferType)
+.setInferShape(inferShape)
+.setCreateOpState(createOpStateCPU, "cpu")
+.setCreateOpState(createOpStateGPU, "gpu");
+
+MXReturnValue initialize(int version) {
+    if (version >= 10400) {
```
samskalicky: can we change the version number in the examples to 10600 since thats the current version in the master branch?

include/mxnet/lib_api.h
```diff
@@ -328,16 +358,31 @@ typedef void* (*xpu_malloc_t)(void*, int);
  */
 class OpResource {
  public:
-  OpResource(xpu_malloc_t cm, void* ca) : cpu_malloc(cm), cpu_alloc(ca) {}
+  OpResource(xpu_malloc_t cm, void* ca, xpu_malloc_t gm, void* ga, void* st)
```
samskalicky: I know you're just following the existing precedent, but can we change the names `cm` to cpu_mem_alloc so its more readable?

include/mxnet/lib_api.h
```diff
@@ -328,16 +358,31 @@ typedef void* (*xpu_malloc_t)(void*, int);
...
+    : cpu_malloc(cm), gpu_malloc(gm), cpu_alloc(ca), gpu_alloc(ga), cuda_stream(st) {}
 
   /*! \brief allocate memory controlled by MXNet */
```
samskalicky: can we change the comment to "allocate cpu memory"?

include/mxnet/lib_api.h
```diff
@@ -328,16 +358,31 @@ typedef void* (*xpu_malloc_t)(void*, int);
...
-  void* alloc(int size) {
+  void* alloc_cpu(int size) {
     return cpu_malloc(cpu_alloc, size);
   }
 
+  /*! \brief allocate memory controlled by MXNet */
```
samskalicky: can we change the comment to "allocate gpu memory"?

include/mxnet/lib_api.h
```diff
@@ -328,16 +358,31 @@ typedef void* (*xpu_malloc_t)(void*, int);
...
+  void* alloc_gpu(int size) {
+    return gpu_malloc(gpu_alloc, size);
+  }
+
+  /*! \brief return the gpu stream object */
+  void* get_cuda_stream() {
```
samskalicky: As mentioned above, consider doing this instead:
```
#if defined(__NVCC__)
 cudaStream_t
#else
 void*
#endif
```

include/mxnet/lib_api.h
```diff
@@ -328,16 +358,31 @@ typedef void* (*xpu_malloc_t)(void*, int);
...
+
  private:
-  xpu_malloc_t cpu_malloc;
-  void* cpu_alloc;
+  /*! \brief wrapper to allocation lambda function */
+  xpu_malloc_t cpu_malloc, gpu_malloc;
+  /*! \brief lambda function to return allocated memory handle */
+  void *cpu_alloc, *gpu_alloc;
+  /*! \brief cuda stream passed from MXNet */
+  void *cuda_stream;
```
samskalicky: and again here, or do this once up above and use everywhere:
```
#if defined(__NVCC__)
 typedef cudaStream_t mx_stream_t;
#else
 typedef void* mx_stream_t;
#endif
```

include/mxnet/lib_api.h
```diff
@@ -594,26 +640,58 @@ class CustomOp {
...
+  /*! \brief vector repr of ctx map to be easily loaded from c_api */
+  std::vector<const char*> forward_ctx_cstr, backward_ctx_cstr, create_op_ctx_cstr;
+  std::vector<fcomp_t> forward_fp, backward_fp;
+  std::vector<createOpState_t> create_op_fp;
+
+ private:
+  void raiseDuplicateContextError() {
+    std::string op_name_str(name);
+    throw std::runtime_error(
+      "Error! Register multiple functions under same context for operator '"
```
samskalicky: how about "Error! Cannot register multiple functions under same context for operator '" ...

include/mxnet/lib_api.h
```diff
@@ -1056,26 +1171,30 @@ extern "C" {
...
-                           outIDs[i]);
+                           outIDs[i], {outdev_type[i], outdev_id[i]});
     }
-    OpResource res(cpu_malloc, cpu_alloc);
+
+    OpResource res(cpu_malloc, cpu_alloc, gpu_malloc, gpu_alloc, stream);
+
     CustomStatefulOp* op_ptr = reinterpret_cast<CustomStatefulOp*>(state_op);
-    if (is_forward) {
+    if (is_forward == 1) {
```
samskalicky: the equal comparison is not necessary since a you're passing either a 1 or a 0 in c_api.cc and `if(0)` evaluates to false

src/c_api/c_api.cc
```diff
@@ -99,7 +99,135 @@ inline int MXAPIGetFunctionRegInfo(const FunRegType *e,
...
+  };
+
+  // create lambda without captures so that we can cast it to function pointer
+  // this needs to be a lambda function so that we can do the decltype cast
+  typedef decltype(cpu_alloc) alloc_type_cpu;
+  auto cpu_malloc = [](void* _cpu_alloc, int size) {
+    // cast the void* argument to the type for the cpu_alloc lambda function
+    alloc_type_cpu* cpualloc = static_cast<alloc_type_cpu*>(_cpu_alloc);
+    // call cpu_alloc to actually allocate memory and get the pointer
+    void* ptr = (*cpualloc)(size);
```
samskalicky: is there some reason we're doing this in two lines instead of just `return (*cpualloc)(size)`?

src/c_api/c_api.cc
```diff
@@ -99,7 +99,135 @@ inline int MXAPIGetFunctionRegInfo(const FunRegType *e,
...
+      attr_vals.push_back(kv.second.c_str());
+    }
+    // call fcompute function
+    CHECK(callFComp(fcomp_fp, attr_keys.data(), attr_vals.data(), attr_keys.size(),
+                    in_shapes.data(), in_dims.data(), in_data.data(), in_types.data(),
+                    in_verIDs.data(), in_dev_type.data(), in_dev_id.data(), in_data.size(),
+                    out_shapes.data(), out_dims.data(), out_data.data(), out_types.data(),
+                    out_verIDs.data(), out_dev_type.data(), out_dev_id.data(), out_data.size(),
+                    cpu_malloc, &cpu_alloc, gpu_malloc, &gpu_alloc, cuda_stream))
+    << "Error calling FCompute for custom operator '" << op_name << "'";
```
samskalicky: should we make this unique for the stateful operator? like: "Error calling FCompute for custom stateful operator '" << op_name << "'"

src/c_api/c_api.cc
```diff
@@ -99,7 +99,135 @@ inline int MXAPIGetFunctionRegInfo(const FunRegType *e,
...
+  }
+
+  if (state_ptr != nullptr) {
+    // retrieve op state object created from CreateOpState
+    CustomStatefulOpWrapper& op = state_ptr->get_state<CustomStatefulOpWrapper>();
+    CustomStatefulOp* state_op_inst = op.get_instance();
+    CHECK(state_op_inst != nullptr)
+    << "Error MXNet cannot load custom stateful operator'" << op_name << "'";
```
samskalicky: this error message makes it seem like there was a problem with loading the library/operator, when actually its a problem with MXNet unable to retrieve the op_state that should have been created already. Maybe the message should say "Error! op_state is null for operator '" << op_name

src/c_api/c_api.cc
```diff
@@ -563,101 +623,30 @@ int MXLoadLib(const char *path) {
...
-    };
-
-    auto fstateful_backward = [=](const OpStatePtr& state_ptr,
-                                  const OpContext& ctx,
-                                  const std::vector<NDArray>& inputs,
-                                  const std::vector<OpReqType>& req,
-                                  const std::vector<NDArray>& outputs) {
-      fstateful_lambda(false, state_ptr, ctx, inputs, req, outputs);
-    };
+    /* -------------- BELOW ARE CUSTOM OPERATOR REGISTRATION --------------- */
```
samskalicky: I think you mean "BELOW IS THE REGISTRATION FOR CUSTOM OPERATORS"

src/c_api/c_api.cc
```diff
@@ -685,28 +674,54 @@ int MXLoadLib(const char *path) {
...
-      regOp.set_attr<FComputeEx>("FComputeEx<cpu>", forward_lambda, plevel);
+      if (forward_ctx_map.count("cpu") > 0) {
+        auto forward_cpu_lambda = [=](const nnvm::NodeAttrs& attrs,
+                                      const OpContext& ctx,
+                                      const std::vector<NDArray>& inputs,
+                                      const std::vector<OpReqType>& req,
+                                      const std::vector<NDArray>& outputs) {
+          CustomFComputeDispatcher(name_str, callFComp, forward_ctx_map.at("cpu"), &attrs,
+                                   nullptr, 0, nullptr, ctx, inputs, req, outputs);
+        };
```
samskalicky: I think we should call `forward_ctx_map.at("cpu")` outside of the lambda and just use the `fcomp_t` otherwise it will capture the whole `forward_ctx_map` map

samskalicky:

samskalicky:
@rondogency can you try this:
```
template <class T>
class Registry {
 public:
  /*!
   * \brief get singleton pointer to class
   * \returns pointer to class
   */
  static Registry* get() {
    static Registry inst __attribute__ ((visibility ("hidden")));
    return &inst;
  }
```

example/extensions/lib_custom_op/gemm_lib.cc
```diff
@@ -203,9 +210,9 @@ class MyStatefulGemm : public CustomStatefulOp {
 
 MXReturnValue createOpState(std::map<std::string, std::string> attrs,
                             CustomStatefulOp** op_inst) {
-  int count = 0;
-  if (attrs.count("test_kw") > 0)
-    count = std::stoi(attrs["test_kw"]);
+  // testing passing of keyward arguments
```
ptrendx: Not: keyword

example/extensions/lib_custom_op/relu_lib.cu
```diff
@@ -0,0 +1,193 @@
...
+__global__ void relu_gpu_forward(float *out, float *in, int64_t N) {
+    int tid = blockIdx.x * blockDim.x + threadIdx.x;
+    if (tid < N)
+        out[tid] = in[tid] > 0 ? in[tid] : 0;
+}
+
+__global__ void relu_gpu_backward(float *out, float *in, int64_t N) {
+    int tid = blockIdx.x * blockDim.x + threadIdx.x;
+    if (tid < N)
+        out[tid] = in[tid] > 0 ? 1 : 0;
```
ptrendx: Random thing, but this is actually wrong, right? It should not be 1, but 1*incoming gradient, right?

example/extensions/lib_custom_op/test_relu.py
```diff
@@ -0,0 +1,69 @@
...
+print("--------start stress test---------")
+a = mx.nd.uniform(shape=(1000,1000,100), ctx=mx.cpu())
+b = mx.nd.uniform(shape=(1000,1000,100), ctx=mx.gpu())
+t1 = time.time()
+r1 = mx.nd.my_relu(a)
+t2 = time.time()
+r2 = mx.nd.my_relu(b)
+t3 = time.time()
+print("CPU running time:")
+print(t2 - t1)
```
ptrendx: You should have `mx.nd.waitall()` somewhere, otherwise it is a wrong measurement.

samskalicky:
> One question I have is how would you use other resources than temp workspace in the custom op, like a random resource?

We're tracking all the other operator features to add for custom ops here: #17006 and RNGs are on the list. Please add more items there and we'll get to it in the next PR.

example/extensions/lib_custom_op/relu_lib.cu
rondogency: thanks for pointing out the error. I have made the fix

example/extensions/lib_custom_op/test_relu.py
rondogency: thanks for the pointing out the error. I didn't pay attention as we would do more tests including benchmark tests in the coming PRs. I have made the fix now.

samskalicky:
@rondogency if you're going to add a new library for the GPU tests you need to modify: CMakeLists.txt, Jenkins_steps.groovy. In the CI cmake is used, and you need to store the library between build/test stages of the CI by modifying the groovy file. 

See the subgraph property PR modifications for these files that add `libsubgraph_lib.so`:
https://github.com/apache/incubator-mxnet/pull/17034/files

rondogency:
@ptrendx thanks for your comments! I have resolved those comments, and I will appreciate if you could take another quick look and approve this.

CMakeLists.txt
```diff
@@ -752,6 +741,31 @@ elseif(MSVC)
...
+  if (USE_CUDA)
+    target_compile_options(customop_gpu_lib PUBLIC -shared)
+  endif()
+elseif(MSVC)
+  target_compile_options(customop_lib PUBLIC /LD)
+  target_compile_options(subgraph_lib PUBLIC /LD)
+  set_target_properties(customop_lib PROPERTIES PREFIX "lib")
+  set_target_properties(subgraph_lib PROPERTIES PREFIX "lib")
+  if (USE_CUDA)
+    target_compile_options(customop_gpu_lib PUBLIC /LD)
```
samskalicky: can you try adding:
```
target_compile_options(target_compile_options  PUBLIC "$<COMPILE_LANGUAGE:CUDA>>:-Xcompiler=-fPIC>")
```

samskalicky: also try this so we can debug the cmake generated makefile:
```
  set(CMAKE_VERBOSE_MAKEFILE ON)
```

samskalicky:
@rondogency looks like the windows build/test is working now with those cmake changes:
```
test_operator_gpu.test_custom_op_gpu ... 
MXNet version 10600 supported
[
10:16:08] C:\jenkins_slave\workspace\build-gpu\src\c_api\c_api.cc:286: 
Found 2 operators in library
[
10:16:08] C:\jenkins_slave\workspace\build-gpu\src\c_api\c_api.cc:350: 	Op[0] my_relu
[
10:16:08] C:\jenkins_slave\workspace\build-gpu\src\c_api\c_api.cc:350: 	Op[1] my_state_relu
[
10:16:08] C:\jenkins_slave\workspace\build-gpu\src\c_api\c_api.cc:785: Found 0 partitioners in library
ok (0.6834s)
```

Now we just need to work through the flaky tests

CMakeLists.txt
```diff
@@ -752,6 +741,32 @@ elseif(MSVC)
...
+  if (USE_CUDA)
+    target_compile_options(customop_gpu_lib PUBLIC -shared)
+  endif()
+elseif(MSVC)
+  target_compile_options(customop_lib PUBLIC /LD)
+  target_compile_options(subgraph_lib PUBLIC /LD)
+  set_target_properties(customop_lib PROPERTIES PREFIX "lib")
+  set_target_properties(subgraph_lib PROPERTIES PREFIX "lib")
+  if (USE_CUDA)
+    set(CMAKE_VERBOSE_MAKEFILE ON)
```
samskalicky: can we remove this now that windows builds are working?

Makefile
```diff
@@ -664,11 +664,19 @@ cpplint:
...
+EXT_LIBS = custom_op_lib subgraph_lib
+ifeq ($(USE_CUDA), 1)
+	EXT_LIBS += custom_op_gpu_lib
+endif
+extension_libs: $(EXT_LIBS)
+
+custom_op_lib:
+	$(CXX) -shared -fPIC -std=c++11 example/extensions/lib_custom_op/gemm_lib.cc -o build/libcustomop_lib.so -I include/mxnet
+custom_op_gpu_lib:
+	$(NVCC) -shared -std=c++11 -Xcompiler -fPIC example/extensions/lib_custom_op/relu_lib.cu -o build/libcustomop_gpu_lib.so -I include/mxnet
```
ptrendx: Why don't you use the NVCCFLAGS and CUDA_ARCH flags here?

tests/python/unittest/test_extensions.py
```diff
@@ -148,3 +156,47 @@ def test_subgraph():
     out3 = exe3.forward()
     # check that result matches one executed by MXNet
     assert_almost_equal(out[0].asnumpy(), out3[0].asnumpy(), rtol=1e-3, atol=1e-3)
+
+@unittest.skipIf(check_platform(), "not all machine types supported")
+@unittest.skipIf(is_cd_run(), "continuous delivery run - ignoring test")
```
ptrendx: Why skipping this test on CD run?

Makefile
samskalicky: we dont need to use the same MXNet flags to build the library, its totally separate

example/extensions/lib_custom_op/test_relu.py
```diff
@@ -0,0 +1,83 @@
...
+print("--------start backward compute--------")
+out_grad = mx.nd.ones((2,2), ctx=mx.gpu())
+exe.backward([out_grad])
+exe_base.backward([out_grad])
+print(in_grad)
+print(in_grad_base)
+
+print("--------start testing larger ndarray---------")
+a = mx.nd.uniform(shape=(100,100,100), ctx=mx.cpu())
+b = mx.nd.uniform(shape=(100,100,100), ctx=mx.gpu())
```
ptrendx: There should be `mx.nd.waitall` after this line for the timings to be any good.

Makefile
rondogency: it is a very small library simply for illustration purpose, so it doesn't necessarily use the existing NVCC and CUDA_ARCH flags used for compile MXNet

ptrendx: sure, makes sense.

tests/python/unittest/test_extensions.py
samskalicky: Looks like this was added in this PR #16127 and @rondogency just copied the same from teh other test in teh test_extensions.py file. We're just following the existing standard. If @perdasilva and @szha want to enable it for the CD run they can do that in a future PR

example/extensions/lib_custom_op/test_relu.py
samskalicky: to not include the memory allocation in the performance timing?

ptrendx: yeah, and to not time those uniform ops.

tests/python/unittest/test_extensions.py
ptrendx: ok

include/mxnet/lib_api.h
```diff
@@ -203,6 +214,16 @@ enum MXDType {
   kUNSET = 100,
 };
 
+/*!
+ * \brief Context info passing from MXNet OpContext
+ * dev_type is string repr of supported context, currently only "cpu" and "gpu"
+ * dev_id is the device index where the tensor locates
+ */
+typedef struct {
+  std::string dev_type;
```
wkcn: I worry about the ABI compatibility of std::string, since MXNet and custom lib may be built in different version of compilers or different compilers, e.g. gcc5 and gcc 9, gcc and clang. The implementation of std::string may be different. A solution is to use a C string char*.

rondogency: no, this string only serves as an easy way for custom library to check the device type, the actual context names passed from custom library to MXNet are still in const char *. Line 971 in lib_api.h is what we passed through the boundary

wkcn: I see. Thank you!

rondogency: I will appreciate if you can make another review!

example/extensions/lib_custom_op/test_relu.py
rondogency: thanks for approving the PR. I am thinking if we can merge the PR now, since we will do more benchmark tests in the coming PRs and it is a small example, so it's not worth the CI run.

samskalicky: @wkcn if you agree would you please merge this PR?

rondogency: I have made a PR on CustomOp GPU support doc #17486 and I will made this fix on the example in that PR as soon as this PR gets merged.

include/mxnet/lib_api.h
```diff
@@ -594,26 +657,58 @@ class CustomOp {
...
+ private:
+  void raiseDuplicateContextError() {
+    std::string op_name_str(name);
+    throw std::runtime_error(
+      "Error! Error! Cannot register multiple functions under same context for operator '"
+      + op_name_str + "'");
+  }
+
+  /*! \brief dedup context maps - static string ctx to custom function */
+  std::unordered_map<const char*, fcomp_t> forward_ctx_map, backward_ctx_map;
```
wkcn: The key of `forward_ctx_map`, `backward_ctx_map` and `create_op_ctx_map` are the pointer value of a string, rather than the content. Although the pointers of the same content are the same in a library, the pointers are different in different libraries. Using `std::string` as the key may be better.

rondogency: This map is used for dedup inside each library, and if loading multiple libraries it would be different operators, since for each custom library there will be one registry object created and custom operators are registered individually in each library.

wkcn: Thank you! I see.

rondogency:
@wkcn Thanks! Can you help me to merge it?

wkcn:
Merged. : )

example/extensions/lib_custom_op/test_relu.py
rondogency: The fix is in #17516 along with one more fix

PR #17284, opened 2020-01-13
Fix language selection in get_started options.js

leezu:
## Description ##
Previously, selecting 'R' will highlight 'Clojure' and 'Perl', as both of them contain 'r'. Further the documentation for Clojure and Perl will be included, which is confusing.

image


aaronmarkham:
Can I preview this change somewhere?

leezu:
I also tested the change by modifying the file using Chrome's local overrides feature. I don't think PRs can be currently previewed as part of the CI built.

Issue #17239, opened 2020-01-07
Cmake with NCCL flag does not work.


apeforest:
## Description
If I build mxnet with NCCL using cmake, it failed with "Could not find NCCL libraries" even though my NCCL is installed at /usr/local/cuda/include
## Reproduce
```
cmake -GNinja -DUSE_CUDA=ON -DCMAKE_CUDA_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Release -DUSE_CUDNN=ON -DUSE_NCCL=ON ..

CMake Warning at CMakeLists.txt:299 (message):
  Could not find NCCL libraries
```
## Environment

We recommend using our script for collecting the diagnositc information. Run the following command and paste the outputs below:
```
curl --retry 10 -s https://raw.githubusercontent.com/dmlc/gluon-nlp/master/tools/diagnose.py | python

----------Python Info----------
Version      : 3.6.6
Compiler     : GCC 7.2.0
Build        : ('default', 'Jun 28 2018 17:14:51')
Arch         : ('64bit', '')
------------Pip Info-----------
Version      : 19.3.1
Directory    : /home/ubuntu/anaconda3/lib/python3.6/site-packages/pip
----------MXNet Info-----------
No MXNet installed.
----------System Info----------
Platform     : Linux-4.4.0-1096-aws-x86_64-with-debian-stretch-sid
system       : Linux
node         : ip-172-31-20-50
release      : 4.4.0-1096-aws
version      : #107-Ubuntu SMP Thu Oct 3 01:51:58 UTC 2019
----------Hardware Info----------
machine      : x86_64
processor    : x86_64
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                64
On-line CPU(s) list:   0-63
Thread(s) per core:    2
Core(s) per socket:    16
Socket(s):             2
NUMA node(s):          2
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 79
Model name:            Intel(R) Xeon(R) CPU E5-2686 v4 @ 2.30GHz
Stepping:              1
CPU MHz:               2699.984
CPU max MHz:           3000.0000
CPU min MHz:           1200.0000
BogoMIPS:              4600.11
Hypervisor vendor:     Xen
Virtualization type:   full
[...]

PR #17297, opened 2020-01-14
Fix NCCL Cmake autodetect issue

ChaiBapchya:
## Description ##
Fixes https://github.com/apache/incubator-mxnet/issues/17239

### Essentials ###

### Changes ###
modified:   cmake/Modules/FindNCCL.cmake

## Test ##
Tested with following command
```
cmake -GNinja -DUSE_CUDA=ON -DCMAKE_CUDA_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Release -DUSE_CUDNN=ON -DUSE_NCCL=ON ..
```

Without this branch, it failed with NCCL not found
```
Could NOT find NCCL (missing: NCCL_INCLUDE_DIRS NCCL_LIBRARIES)
CMake Warning at CMakeLists.txt:636 (message):
  Could not find NCCL libraries
```

With this PR, it passes.

### DIFF
--- a/cmake/Modules/FindNCCL.cmake
+++ b/cmake/Modules/FindNCCL.cmake
@@ -43,6 +43,18 @@ find_path(NCCL_INCLUDE_DIRS
   $ENV{NCCL_DIR}/include
   )
 
+# if CUDA_TOOLKIT_ROOT_DIR is not specified, try symlink
+if (NOT CUDA_TOOLKIT_ROOT_DIR)
+  if (UNIX)
+    set (search_paths "/usr/local/cuda")
+  endif()
+
+  find_path(NCCL_INCLUDE_DIRS
+  NAMES nccl.h
+  PATHS ${search_paths}
+  )
+endif()
+
 find_library(NCCL_LIBRARIES
   NAMES nccl
   HINTS
--- a/cmake/Modules/FindNCCL.cmake
+++ b/cmake/Modules/FindNCCL.cmake
@@ -42,8 +42,6 @@ endif()
 find_path(NCCL_INCLUDE_DIRS
   NAMES nccl.h
   HINTS
-  ${NCCL_ROOT}
-  ${NCCL_ROOT}/include
   ${NCCL_INCLUDE_DIR}
   ${NCCL_ROOT_DIR}
   ${NCCL_ROOT_DIR}/include
@@ -55,10 +53,6 @@ find_library(NCCL_LIBRARIES
   NAMES nccl
   HINTS
   ${NCCL_LIB_DIR}
-  ${NCCL_ROOT}
-  ${NCCL_ROOT}/lib
-  ${NCCL_ROOT}/lib/x86_64-linux-gnu
-  ${NCCL_ROOT}/lib64
   ${NCCL_ROOT_DIR}
   ${NCCL_ROOT_DIR}/lib
   ${NCCL_ROOT_DIR}/lib/x86_64-linux-gnu
@@ -72,15 +66,15 @@ if (UNIX)
   set (search_paths "/usr/local/cuda")
 
   find_path(NCCL_INCLUDE_DIRS
-   NAMES nccl.h
-   PATHS ${search_paths}
-   PATH_SUFFIXES include
+    NAMES nccl.h
+    PATHS ${search_paths}
+    PATH_SUFFIXES include
   )
 
   find_library(NCCL_LIBRARIES
-   NAMES nccl
-   PATHS ${search_paths}
-   PATH_SUFFIXES lib
+    NAMES nccl
+    PATHS ${search_paths}
+    PATH_SUFFIXES lib
   )
 endif()
 
--- a/cmake/Modules/FindNCCL.cmake
+++ b/cmake/Modules/FindNCCL.cmake
@@ -31,28 +31,19 @@
 # install NCCL in the same location as the CUDA toolkit.
 # See https://github.com/caffe2/caffe2/issues/1601
 
-set(NCCL_ROOT_DIR "" CACHE PATH "Folder contains NVIDIA NCCL")
+set(NCCL_ROOT $ENV{NCCL_ROOT} CACHE PATH "Folder contains NVIDIA NCCL")
+set(NCCL_INCLUDE_DIR $ENV{NCCL_INCLUDE_DIR} CACHE PATH "Folder contains NVIDIA NCCL headers")
+set(NCCL_LIB_DIR $ENV{NCCL_LIB_DIR} CACHE PATH "Folder contains NVIDIA NCCL libraries")
 
-# first check in the /usr/local/cuda before other paths
-if (UNIX)
-  set (search_paths "/usr/local/cuda")
-
-  find_path(NCCL_INCLUDE_DIRS
-  NAMES nccl.h
-  PATHS ${search_paths}
-  PATH_SUFFIXES include
-  )
-
-  find_library(NCCL_LIBRARIES
-  NAMES nccl
-  PATHS ${search_paths}
-  PATH_SUFFIXES lib
-  )
+if ($ENV{NCCL_ROOT_DIR})
+  message(WARNING "NCCL_ROOT_DIR is deprecated. Please set NCCL_ROOT instead.")
 endif()
 
 find_path(NCCL_INCLUDE_DIRS
   NAMES nccl.h
   HINTS
+  ${NCCL_ROOT}
+  ${NCCL_ROOT}/include
   ${NCCL_INCLUDE_DIR}
   ${NCCL_ROOT_DIR}
   ${NCCL_ROOT_DIR}/include
@@ -64,6 +55,10 @@ find_library(NCCL_LIBRARIES
   NAMES nccl
   HINTS
   ${NCCL_LIB_DIR}
+  ${NCCL_ROOT}
+  ${NCCL_ROOT}/lib
+  ${NCCL_ROOT}/lib/x86_64-linux-gnu
+  ${NCCL_ROOT}/lib64
   ${NCCL_ROOT_DIR}
   ${NCCL_ROOT_DIR}/lib
   ${NCCL_ROOT_DIR}/lib/x86_64-linux-gnu
@@ -71,6 +66,23 @@ find_library(NCCL_LIBRARIES
   ${CUDA_TOOLKIT_ROOT_DIR}/lib64
   $ENV{NCCL_DIR}/lib
   )
+
+# if not found in any of the above paths, finally, check in the /usr/local/cuda for UNIX systems
+if (UNIX)
+  set (search_paths "/usr/local/cuda")
+
+  find_path(NCCL_INCLUDE_DIRS
+   NAMES nccl.h
+   PATHS ${search_paths}
+   PATH_SUFFIXES include
+  )
+
+  find_library(NCCL_LIBRARIES
+   NAMES nccl
+   PATHS ${search_paths}
+   PATH_SUFFIXES lib
+  )
+endif()
 
 include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(NCCL DEFAULT_MSG NCCL_INCLUDE_DIRS NCCL_LIBRARIES)
--- a/cmake/Modules/FindNCCL.cmake
+++ b/cmake/Modules/FindNCCL.cmake
@@ -31,10 +31,6 @@
 # install NCCL in the same location as the CUDA toolkit.
 # See https://github.com/caffe2/caffe2/issues/1601
 
-set(NCCL_ROOT $ENV{NCCL_ROOT} CACHE PATH "Folder contains NVIDIA NCCL")
-set(NCCL_INCLUDE_DIR $ENV{NCCL_INCLUDE_DIR} CACHE PATH "Folder contains NVIDIA NCCL headers")
-set(NCCL_LIB_DIR $ENV{NCCL_LIB_DIR} CACHE PATH "Folder contains NVIDIA NCCL libraries")
-
 if ($ENV{NCCL_ROOT_DIR})
   message(WARNING "NCCL_ROOT_DIR is deprecated. Please set NCCL_ROOT instead.")
 endif()
--- a/cmake/Modules/FindNCCL.cmake
+++ b/cmake/Modules/FindNCCL.cmake
@@ -33,17 +33,7 @@
 
 set(NCCL_ROOT_DIR "" CACHE PATH "Folder contains NVIDIA NCCL")
 
-find_path(NCCL_INCLUDE_DIRS
-  NAMES nccl.h
-  HINTS
-  ${NCCL_INCLUDE_DIR}
-  ${NCCL_ROOT_DIR}
-  ${NCCL_ROOT_DIR}/include
-  ${CUDA_TOOLKIT_ROOT_DIR}/include
-  $ENV{NCCL_DIR}/include
-  )
-
-# if CUDAToolkit_FOUND is not found, try default location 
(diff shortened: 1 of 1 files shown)


cmake/Modules/FindNCCL.cmake
```diff
@@ -33,6 +33,26 @@
 
 set(NCCL_ROOT_DIR "" CACHE PATH "Folder contains NVIDIA NCCL")
 
+# if CUDAToolkit_FOUND is not found, try default location
+#if (NOT CUDAToolkit_FOUND)
```
apeforest: Space?

cmake/Modules/FindNCCL.cmake
```diff
@@ -33,6 +33,26 @@
...
+  if (UNIX)
+    set (search_paths "/usr/local/cuda")
+
+    find_path(NCCL_INCLUDE_DIRS
+    NAMES nccl.h
```
apeforest: Indent

cmake/Modules/FindNCCL.cmake
```diff
@@ -33,6 +33,26 @@
...
+    )
+
+    find_library(NCCL_LIBRARIES
+    NAMES nccl
+    PATHS ${search_paths}
+    PATH_SUFFIXES lib
+    )
+
+  endif()
+  #endif()
```
apeforest: Remove

cmake/Modules/FindNCCL.cmake
```diff
@@ -33,6 +33,26 @@
...
 set(NCCL_ROOT_DIR "" CACHE PATH "Folder contains NVIDIA NCCL")
 
+# if CUDAToolkit_FOUND is not found, try default location
+#if (NOT CUDAToolkit_FOUND)
+  if (UNIX)
+    set (search_paths "/usr/local/cuda")
+
+    find_path(NCCL_INCLUDE_DIRS
+    NAMES nccl.h
+    PATHS ${search_paths}
```
leezu: Should this be HINTS?

ChaiBapchya: https://cmake.org/cmake/help/v3.4/command/find_path.html
Looks like either Hints or paths is permitted. Any specific reason to use hints?

leezu: When writing the comment I thought PATHS may skip default locations whereas HINTS would be defaults + specified locations. Thanks for pointing out they are equivalent.

ChaiBapchya:
Verified this works by running following
Fresh cmake build
```
rm -rf build
mkdir -p build && cd build
cmake -GNinja \
    -DUSE_CUDA=ON \
    -DUSE_MKL_IF_AVAILABLE=OFF \
    -DCMAKE_CUDA_COMPILER_LAUNCHER=ccache \
    -DCMAKE_C_COMPILER_LAUNCHER=ccache \
    -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
    -DCMAKE_BUILD_TYPE=Release \
    -DUSE_NCCL=ON \
..
ninja
```
Python binding
```
cd ..
pip install -e python/.
python3 -c "from mxnet.runtime import feature_list; print(feature_list())"
[✔ CUDA, ✔ CUDNN, ✔ NCCL, ✔ CUDA_RTC, ✖ TENSORRT, ✔ CPU_SSE, ✔ CPU_SSE2, ✔ CPU_SSE3, ✔ CPU_SSE4_1, ✔ CPU_SSE4_2, ✖ CPU_SSE4A, ✔ CPU_AVX, ✖ CPU_AVX2, ✔ OPENMP, ✖ SSE, ✔ F16C, ✔ JEMALLOC, ✔ BLAS_OPEN, ✖ BLAS_ATLAS, ✖ BLAS_MKL, ✖ BLAS_APPLE, ✔ LAPACK, ✖ MKLDNN, ✔ OPENCV, ✖ CAFFE, ✖ PROFILER, ✖ DIST_KVSTORE, ✖ CXX14, ✖ INT64_TENSOR_SIZE, ✔ SIGNAL_HANDLER, ✖ DEBUG, ✖ TVM_OP]
```

ChaiBapchya:

cmake/Modules/FindNCCL.cmake
```diff
@@ -33,6 +33,23 @@
 
 set(NCCL_ROOT_DIR "" CACHE PATH "Folder contains NVIDIA NCCL")
 
+# first check in the /usr/local/cuda before other paths
```
leezu: "first check" may break the assumption that `NCCL_ROOT_DIR` can be used to specify nccl root directory?

Would adding `/usr/local/cuda` as last item in `find_path`, `find_library` below be sufficient? Or does it throw an error if `/usr/local/cuda` doesn't exist?

ChaiBapchya: I'm not sure why it breaks the assumption. Can you help me rephrase the comment? I can remove "first" if that's what confuses the users.

leezu: Quoting from https://cmake.org/cmake/help/latest/command/find_path.html

> This command is used to find a directory containing the named file. A cache entry named by `<VAR>` is created to store the result of this command. If the file in a directory is found the result is stored in the variable and the search will not be repeated unless the variable is cleared. If nothing is found, the result will be `<VAR>-NOTFOUND`, and the search will be attempted again the next time find_path is invoked with the same variable.

Thus if there is some `nccl.h` at `/usr/local/cuda` but the user specifies `NCCL_ROOT_DIR=/home/ubuntu/myversionofnccl`, the `/usr/local/cuda` is wrongly used? Or do I misunderstand the doc? 

Btw, as per https://cmake.org/cmake/help/latest/policy/CMP0074.html `NCCL_ROOT_DIR` should probably be called `NCCL_ROOT` (or `NCCL_ROOT` should be added in addition to `NCCL_ROOT_DIR`).

cmake/Modules/FindNCCL.cmake
```diff
@@ -33,6 +33,23 @@
...
+if (UNIX)
+  set (search_paths "/usr/local/cuda")
+
+  find_path(NCCL_INCLUDE_DIRS
+  NAMES nccl.h
```
apeforest: Not familiar with cmake, but do you need add indentation here?

cmake/Modules/FindNCCL.cmake
```diff
@@ -33,6 +33,23 @@
...
+  PATHS ${search_paths}
+  PATH_SUFFIXES include
+  )
+
+  find_library(NCCL_LIBRARIES
+  NAMES nccl
```
apeforest: Not familiar with cmake, but do you need add indentation here?

cmake/Modules/FindNCCL.cmake
ChaiBapchya: Ya. Even I saw this NCCL_ROOT and found that NCCL_ROOT_DIR is deprecated hence no longer to be used in PyTorch repo
https://github.com/pytorch/pytorch/blob/f94aab45fd7b1e7e32d8ea4324da6d3632a0d412/cmake/Modules/FindNCCL.cmake#L21

However, I couldn't find official cmake doc saying so. Nonetheless, I think I should add that too.

ChaiBapchya: > If there is some nccl.h at /usr/local/cuda but the user specifies NCCL_ROOT_DIR=/home/ubuntu/myversionofnccl, the /usr/local/cuda is wrongly used? Or do I misunderstand the doc?

I agree, since "search" will not be repeated until var is set, it makes sense to assume whatever location is found first is final. So in that case first check has to be env var and second for default.

ChaiBapchya: Surprisingly, I couldn't find a check for `/usr/local/cuda` in case of pytorch's FindNCCL
I have created a discuss question on their forum anyway - https://discuss.pytorch.org/t/findnccl-cmake-search-for-usr-local-cuda/66891

leezu: > I saw this NCCL_ROOT and found that NCCL_ROOT_DIR is deprecated hence no longer to be used in PyTorch repo
> https://github.com/pytorch/pytorch/blob/f94aab45fd7b1e7e32d8ea4324da6d3632a0d412/cmake/Modules/FindNCCL.cmake#L21
> 
> However, I couldn't find official cmake doc saying so. Nonetheless, I think I should add that too.

Thanks. The official doc should be https://cmake.org/cmake/help/latest/policy/CMP0074.html

cmake/Modules/FindNCCL.cmake
ChaiBapchya: yes good catch! indentation needed.

cmake/Modules/FindNCCL.cmake
ChaiBapchya: Moved /usr/local/cuda check to be after other checks
Also added NCCL_ROOT alongside NCCL_ROOT_DIR
Also added deprecation warning for ROOT_DIR

leezu: @ChaiBapchya Thank you!

After reading CMP0074 again,
> [...] now searches prefixes specified by the `<PackageName>_ROOT` CMake variable and the `<PackageName>_ROOT` environment variable. Package roots are maintained as a stack so nested calls to all find_* commands inside find modules and config packages also search the roots as prefixes.

seems to imply that it's not necessary to list `NCCL_ROOT` as part of `find_path` and `find_library`, but rather it's done automatically. Do you understand it in the same way?
Also https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure (paragraph "The set of installation prefixes is constructed using the following steps [...]") suggests this.

If so,  `set(NCCL_ROOT $ENV{NCCL_ROOT} CACHE PATH "Folder contains NVIDIA NCCL")` would be superfluous. Also it seems none of the other Find modules check the `X_LIB_DIR` and `X_LIB_DIR` env variables.

ChaiBapchya: 1. NCCL_ROOT auto check - I guess it is done automatically. Though it's not mentioned explicitly anywhere. Hence asked here - https://stackoverflow.com/questions/59783425/does-cmake-find-path-find-library-search-for-packagename-root-by-default

2. 
> Also it seems none of the other Find modules check the X_LIB_DIR and X_LIB_DIR env variables.
Not sure what you implying? You think that is superfluous too?

ChaiBapchya:
Also @leezu any idea why it's failing for Windows CPU Cmake. Unable to discern from the Error log.

leezu:
Not sure. I restarted the job to make sure it's not an intermittent issue.

ChaiBapchya:
Wow! It passed! Weird! @leezu thanks for retrigger

cmake/Modules/FindNCCL.cmake
```diff
@@ -55,6 +67,23 @@ find_library(NCCL_LIBRARIES
   $ENV{NCCL_DIR}/lib
   )
 
+# if not found in any of the above paths, finally, check in the /usr/local/cuda for UNIX systems
+if (UNIX)
+  set (search_paths "/usr/local/cuda")
+
+  find_path(NCCL_INCLUDE_DIRS
+   NAMES nccl.h
```
apeforest: Sorry for being picky, but isn't the indent supposed to be 2 space for all?

cmake/Modules/FindNCCL.cmake
```diff
@@ -55,6 +67,23 @@ find_library(NCCL_LIBRARIES
   $ENV{NCCL_DIR}/lib
   )
 
+# if not found in any of the above paths, finally, check in the /usr/local/cuda for UNIX systems
+if (UNIX)
```
apeforest: What happens if not UNIX? is there any error/warning message issued?

cmake/Modules/FindNCCL.cmake
ChaiBapchya: 1. Autocheck will be done provided `NO_DEFAULT_PATH` or `NO_PACKAGE_ROOT_PATH` are not set. So I'm assuming that's the case.
Refer : https://stackoverflow.com/questions/59783425/does-cmake-find-path-find-library-search-for-packagename-root-by-default/59788313#59788313

cmake/Modules/FindNCCL.cmake
leezu: A variable will be set to not found. The calling cmake module needs to check that variable and error out if nccl is required.

apeforest:
@ChaiBapchya Please fix the indent problem as I pointed out. Otherwise, this PR is good to go for me.

cmake/Modules/FindNCCL.cmake
leezu: 1. Yes, you should remove the manual checks for `NCCL_ROOT`.

2. I don't think it's superfluous (ie it does something), but it seems not necessary because it seems non-standard (ie other Find modules don't check the environment variables, so it's not expected that you check them here)

cmake/Modules/FindNCCL.cmake
```diff
@@ -31,7 +31,13 @@
 # install NCCL in the same location as the CUDA toolkit.
 # See https://github.com/caffe2/caffe2/issues/1601
 
-set(NCCL_ROOT_DIR "" CACHE PATH "Folder contains NVIDIA NCCL")
+set(NCCL_ROOT $ENV{NCCL_ROOT} CACHE PATH "Folder contains NVIDIA NCCL")
```
leezu: > You can also set(ENV{variable_name} value) and get $ENV{variable_name} environment variables, though it is generally a very good idea to avoid them.

https://cliutils.gitlab.io/modern-cmake/chapters/basics/variables.html

Should the env variable checks here be removed? You could just delete these 3 lines.

ChaiBapchya: So I am guessing I should remove it here too?

https://github.com/ChaiBapchya/incubator-mxnet/blob/a9abba5a934755813481e6583435298a0c317957/cmake/Modules/FindNVTX.cmake#L18

ChaiBapchya: or is it specifically to do with ENV vars. In that case we can set "" to these 3 variables - NCCL_ROOT, NCCL_INCLUDE_DIR and NCCL_LIB_DIR
Since that seems to be the norm

apeforest: @ChaiBapchya let's focus on one thing in a PR. I would change the NVTX_ROOT_DIR in separate PR if needed.

leezu: @ChaiBapchya Yes, for this PR my concern is only that you are introducing environment variable depencies, which is not standard practice and not needed.
So for this PR, let's just remove the 3 lines.
For a later PR, NVTX can be refactored.

ChaiBapchya: @leezu Setting empty string to a cache variable does serve a purpose (declaring a variable so as to check if it is set or not in future)
So if NCCL_ROOT_DIR is not already set then it assigns a default value to it (right now since we have deleted this line, if variable is not set via command line, it won't exist -> hence can't use it in this file)

What do you think?

https://stackoverflow.com/questions/59867283/whats-the-point-of-using-empty-string-to-set-cache-variable-in-cmake/59868538#59868538

leezu: Yes, I agree with the Stackoverflow answer. In your case, it would "assigns description" to the variable. But as these are deprecated variables I don't think we need to add description.
The standard / supported variable is `NCCL_ROOT` as per https://cmake.org/cmake/help/latest/policy/CMP0074.html

Issue #17292, opened 2020-01-13
Can't run horovod with latest nightly wheel


apeforest:
## Description
Cannot run horovod with latest nightly wheel. It could mean the 1.6 release have same problem too. The last working nightly wheel was 12/30/2019

### Error Message
```
OSError: /home/ubuntu/.virtualenvs/build/lib/python3.6/site-packages/horovod/mxnet/mpi_lib.cpython-36m-x86_64-linux-gnu.so: undefined symbol: _Z20MXAPIHandleExceptionRKSt9exception
```

### Steps to reproduce
```
pip install https://apache-mxnet.s3-us-west-2.amazonaws.com/dist/2019-12-30/dist/mxnet_cu92-1.6.0b20191230-py2.py3-none-manylinux1_x86_64.whl
pip install horovod
python -c "import horovod.mxnet as hvd"
```
## What have you tried to solve it?

1. Tried mxnet-cu100 nightly wheel on 12/30/2019. It worked
2. Tried mxnet-cu92 nightly wheel on 12/30/2019. It worked

## Environment

```
----------Python Info----------
Version      : 3.6.6
Compiler     : GCC 7.2.0
Build        : ('default', 'Jun 28 2018 17:14:51')
Arch         : ('64bit', '')
------------Pip Info-----------
Version      : 19.3.1
Directory    : /home/ubuntu/.virtualenvs/build/lib/python3.6/site-packages/pip
----------MXNet Info-----------
Version      : 1.6.0
Directory    : /home/ubuntu/.virtualenvs/build/lib/python3.6/site-packages/mxnet
Num GPUs     : 8
Hashtag not found. Not installed from pre-built package.
----------System Info----------
Platform     : Linux-4.4.0-1096-aws-x86_64-with-debian-stretch-sid
system       : Linux
node         : ip-172-31-20-50
release      : 4.4.0-1096-aws
version      : #107-Ubuntu SMP Thu Oct 3 01:51:58 UTC 2019
----------Hardware Info----------
machine      : x86_64
processor    : x86_64
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                64
On-line CPU(s) list:   0-63
Thread(s) per core:    2
Core(s) per socket:    16
Socket(s):             2
NUMA node(s):          2
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 79
[...]

PR #17298, opened 2020-01-14
[MXNET-1438] Adding SDML loss function

anjishnu:
## Description ##

### Essentials ###
- [x] Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- [x] For user-facing API changes, API doc string has been updated. 

### Changes ###
- [x] Added new loss for information retrieval - Smoothed Deep Metric Learning Loss with relevant API doc.

## Comments ##
- Added loss function outlined in https://arxiv.org/abs/1905.12786 - example to follow.


haojin2:
@anjishnu Please address the sanity errors: http://jenkins.mxnet-ci.amazon-ml.com/job/mxnet-validation/job/sanity/job/PR-17298/1/display/redirect . Also can you randomize your unit test to ensure that we're covering more numerically different cases?

anjishnu:
Sure will address the sanity cases - 

Can you give an example of a unit test that is appropriately randomized so I can base it on that?

haojin2:
@anjishnu Like this: https://github.com/apache/incubator-mxnet/pull/17298/files#diff-b129d416cc6980021cb35420a66e4d3aR334

anjishnu:
It looks a little tricky to port this into the 'fit' and 'score' paradigm since this is a retrieval specific loss function which uses the other elements in a batch as implicit negative samples - and I'm not sure how cleanly it fits into the Module API for this kind of test. 

The loss only guarantees that associated pairs will be closer in distance space after learning then non-associated pairs. Maybe I can write something equivalent using the gluon API, to train a small network and ensure it learns the right associations. I'll come up with a proposal shortly.

haojin2:
@anjishnu I was actually only asking for getting the input data randomized with the original version of test code untouched.

anjishnu:
@haojin2 if I randomized the input data in the original test code the losses would would have different values during each run (SDML loss imposes a distribution over the relative distances of data points in a minibatch) - so you cannot compare the output against precomputed loss values any more - thus the original unit test procedure cannot be reused.

That's why I added a test that fits a toy model to some toy data instead. The current test was running in ~50 ms on my machine on CPU. Would love to hear your thoughts on how to improve on this.

anjishnu:
Verifying that mention of this PR in #17292 is a typo - author meant to refer to #17208 - we are not breaking horovod

anjishnu:
@haojin2 Are there any serious concerns with the new unit test?

haojin2:
@anjishnu I'm only concerned about if there's any flakiness in such a test, to verify, please try
`MXNET_TEST_COUNT=500 nosetests tests/python/unittest/test_loss.py:test_sdml_loss` on your end and lemme know if it passes.

anjishnu:
@haojin2 
With the original confirguration I was getting around 1 in 100 or 1 in 200 failures, which is not acceptable for a unit test.

Changing the hyperparameters (e.g. increasing dimensionality, lowering N) does make it more robust (tried up to MXNET_TEST_COUNT=2000 runs) but there's always a chance of failure

Other options:
- Go with the original precomputed values approach, but maybe over a larger range of precomputed values.

- Set up a similar test ... but use correlated data instead of using purely random data:
   - Generate the positive samples that are actually correlated to the paired data rather than purely random.
   - Always use some set of samples that are known to converge successfully
- Add a test to see loss reduction (e.g. to <0.05 as in many of the other tests)

Which of these would be your preferred approach - I have tested them all to MXNET_TEST_COUNT=2000

anjishnu:
@haojin2 could you take a look at the latest version - the new unit test (which looks at loss reduction) is passing MXNET_TEST_COUNT=2000

haojin2:
@anjishnu it seems ok. I've re-triggered your failed website build, I think it's good for merge after that passes.

anjishnu:
@haojin2 ok, thanks! sounds good!

haojin2:
@anjishnu Merged, thx for the contribution.

python/mxnet/gluon/loss.py
```diff
@@ -930,3 +930,118 @@ def _cosine_similarity(self, F, x, y, axis=-1):
...
+        Given the question president of France in batch 1 the model will
+        learn to predict french president comparing it with all the other
+        vectors in batch 2
+        """
+        batch_size = x1.shape[0]
+        labels = self._compute_labels(F, batch_size)
+        distances = self._compute_distances(x1, x2)
+        log_probabilities = F.log_softmax(-distances, axis=1)
+        # multiply for the number of labels to obtain the correct loss (gluon kl_loss averages instead of sum)
+        return self.kl_loss(log_probabilities, labels.as_in_context(distances.context)) * batch_size
```
szha, line 1043: this doesn't work in sym. 1.x will need to be fixed while 2.0 will be switched to deferred compute mode

PR #16890, opened 2019-11-22
Mkldnn fullyConnect bwd bug fix

rongzha1:
## Description ##
fix mkldnn fc bwd bug due to data inplace
when data inplace, if do bwd_data will change src data which will lead bwd_weight calc error.

@PatricZhao @TaoLv @xinyu-intel 
### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [done ] Changes are complete (i.e. I finished coding on this PR)
- [done ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [done ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #17318, opened 2020-01-15
Enable MKL-DNN FullyConnected backward

TaoLv:
## Description ##
Discussed with @rongzha1 offline. This PR incorporates the code changes in #16890 which was reverted previously.

### Essentials ###

### Changes ###

## Comments ##


TaoLv:
This is ready for review now. I also changed the cpp test to address the FullyConnectedOp failure reported before. The main reason is the tensor shape is relative large and the max value in the tensor is 50. It will lead to float number accumulation error. @rongzha1 @ciyongch @pengzhao-intel

tests/cpp/include/test_mkldnn.h
```diff
@@ -330,7 +330,7 @@ inline void PrintVerifyMsg(const NDArrayAttrs &arr1, const NDArrayAttrs &arr2) {
  */
 inline std::vector<NDArrayAttrs> GetTestInputArrays(
     int types = ArrayTypes::All, bool rand = false,
-    std::vector<float> scale = {1}, bool spatial_data_format = false) {
+    std::vector<float> scale = {1}, bool spatial_data_format = false, int max = 50) {
```
pengzhao-intel: Is the new parameter for better usability?

tests/cpp/include/test_mkldnn.h
```diff
@@ -63,24 +63,24 @@ struct TestArrayShapes {
...
-inline static void InitDefaultArray(NDArray *arr, bool is_rand = false) {
+inline static void InitDefaultArray(NDArray *arr, bool is_rand = false, int max = 50) {
   const TBlob &blob = arr->data();
   mshadow::default_real_t *data = blob.dptr<mshadow::default_real_t>();
   int size = blob.Size();
 
   for (int i = 0; i < size; i++)
     if (is_rand) {
-      data[i] = (std::rand() % 100) - 50;
+      data[i] = (std::rand() % (max * 2)) - max;
```
ciyongch: How about change to `data[i] = std::rand() * 1.0 / RAND_MAX - 0.5;`? As `max = 1` will only generate two values: -1.0and 0.0 .

tests/cpp/include/test_mkldnn.h
```diff
@@ -63,24 +63,24 @@ struct TestArrayShapes {
...
     } else {
-      data[i] = i % 100 - 50;
+      data[i] = i % (max * 2) - max;
```
ciyongch: Same as above, how about change to something like `data[i] = i * 2.0 / size - 1.0` to generate [-1.0, 1.0)?

tests/cpp/include/test_mkldnn.h
TaoLv: Because I don't want to affect other test case which still use the default max=50 to generate integers in [50, 50). But For the FullyConnectedOp, I want to generate relative small numbers. With the given code, the elements will be -1 and 0. Any suggestion?

tests/cpp/include/test_mkldnn.h
TaoLv: See https://github.com/apache/incubator-mxnet/pull/17318#discussion_r379465354.

tests/cpp/include/test_mkldnn.h
ciyongch: I've no idea about why the range was set to [-50, 50) previously, and I can't figure out any specific reasons to use this range for the tests (any upper bound test?). It'll be great if you have any background for it.
But anyway, the tensors with only two values (-1 and 0, 50% are 0) might not be a good candidate for the tests.

TaoLv: Okay, i will change to generate float numbers in [-max, max) rather than integer numbers. Previously I thought sparse (say 50% zeros) is also a way to avoid float number accumulation error.

TaoLv: @ciyongch Failed to do so. There are cases doing bit correction check. It seems we cannot pass the tests with these floating numbers.
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-17318/17/pipeline/
https://github.com/apache/incubator-mxnet/blob/master/tests/cpp/include/test_mkldnn.h#L605

ciyongch: With `memcmp` to check the results, then the only choice is integer numbers. Is it reasonable to check the results by `AssertEqual` within a small enough threshold like 1e-6, then we can keep the floating number with better distribution? 
Or we can just increase `max` to filling more different numbers other than only -1 and 0.
What do you think?

TaoLv: I have reverted the changes for floating numbers. Changing ``memcmp` to `AssertEqual` is out of the scope of this PR, so I will keep it as is.

> Or we can just increase max to filling more different numbers other than only -1 and 0.

I was thinking about including number 2 into the generated tensor but found that with the given shapes, there still has chance to get error. That means for the worst case, the intermediate accumulation value will be > 2^24, so the 1 will be ignored when accumulating another 1 to  it.

ciyongch: Ok, let's keep it as is for now.

PR #17121, opened 2019-12-19
Include 3rdparty/jemalloc (Fix for Ubuntu 18.10+, Debian Stable, etc.)

leezu:
## Description ##
As of jemalloc 5, jemalloc default build can not be used in libraries that `dlopen`ed. However, `libmxnet.so` is `dlopen`ed by Python (ctypes). We can thus not link to system `libjemalloc.so` but must rather link to a `libjemalloc` compiled as part of our build process.

jemalloc 5 is distributed as part of Ubuntu 18.10 and higher, as well as Debian Stable.
Users on these systems will be unable to compile MXNet with `USE_JEMALLOC=ON` (which is the default).

This PR contains the necessary changes.

Fixes https://github.com/apache/incubator-mxnet/issues/17145

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [X] Changes are complete (i.e. I finished coding on this PR)
- [X] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [X] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at https://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [X] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [X] Include 3rdparty/jemalloc

## Comments ##
- We include a patched version of `jemalloc` that adds CMake support. Once the patch is mainlined, we can switch to the official repo. The patch is currently hosted at https://github.com/leezu/jemalloc/tree/add/cmake-2019. All credits go to @mattsta (see https://github.com/jemalloc/jemalloc/issues/303)

PR #17324, opened 2020-01-15
USE_JEMALLOC=OFF by default

leezu:
## Description ##
As of jemalloc 5, jemalloc default build can not be used in libraries that are dlopened. However, libmxnet.so is dlopened by Python (ctypes). To use MXNet with jemalloc 5, users must not to system libjemalloc.so but must rather link to a libjemalloc compiled with special parameters to allow dlopen to work.

jemalloc 5 is distributed as part of Ubuntu 18.10 and higher, as well as Debian Stable. Users on these systems will be unable to compile MXNet with USE_JEMALLOC=ON.

An alternative is to build jemalloc as part of our build. See https://github.com/apache/incubator-mxnet/pull/17121 for an implementation. Given the lack of review of https://github.com/apache/incubator-mxnet/pull/17121 and the reliance of a patched jemalloc version to allow integration into our build, let's first proceed with the current PR.

### Changes ###
- [X] Change USE_JEMALLOC default to OFF


ChaiBapchya:
Should this be brought to the notice of dev community? I remember when certain build flag (MKLDNN/Large tensor) is to be toggled, it's discussed first. 

On other side, since this one toggles it off so I guess we can go ahead..

leezu:
Thanks, I will send a mail to dev.

PR #17321, opened 2020-01-15
Fixes #17304 Flaky Test -> test_higher_order_grad.test_tanh

kshitij12345:
@apeforest @reminisce 

Going through the code for `_backward_tanh`, it is implemented as `1 - (tanh^2(x))`, which is equivalent to `sech^2(x)` or `(1/cosh(x))^2`. 

However for the failed seed, I have verified that `1 - (tanh^2(x))` is not coming equivalent to `(1/ cosh(x))^2` for the randomly generated array of dim 4 (will try to investigate further for the cause).

For now replacing `grad_op` with its equivalent `1- tanh^2(x)` looks to work without the need to relax tolerance.

PR #17325, opened 2020-01-15
Fix Flaky Test Higher Order Grad

kshitij12345:
While recently fixing #17321 , I stumbled upon cases where `test_arctanh`, `test_arcsin` failing due to getting value which were out of the functions domain.

The way we generated the test array from normal distribution meant that even if very low probability, the chance of an out of domain value showing up was still there.

To this end, have defined a simple wrapper around `np.random.uniform` (similar to `random_arrays`), which lets us specify the lower and upper bound of the generated array.

A pro of using uniform distribution is that we generate samples equal from the whole range of the distribution unlike normal which is centred around mean thus allowing in better testing.

### DIFF
--- a/python/mxnet/test_utils.py
+++ b/python/mxnet/test_utils.py
@@ -102,8 +102,13 @@ def random_arrays(*shapes):
     return arrays
 
 
-def random_uniform_arrays(*shapes, low=0.0, high=1.0, dtype=default_dtype()):
+def random_uniform_arrays(*shapes, **kwargs):
     """Generate some random numpy arrays."""
+    low = kwargs.pop('low', 0.0)
+    high = kwargs.pop('high', 1.0)
+    dtype = kwargs.pop('dtype', default_dtype())
+    if len(kwargs) > 0:
+        raise TypeError('Got unexpected argument/s : ' + str(kwargs.keys()))
     arrays = [np.random.uniform(low, high, size=s).astype(dtype)
               for s in shapes]
     return arrays
--- a/python/mxnet/test_utils.py
+++ b/python/mxnet/test_utils.py
@@ -104,7 +104,7 @@ def random_arrays(*shapes):
 
 def random_uniform_arrays(*shapes, low=0.0, high=1.0):
     """Generate some random numpy arrays."""
-    arrays = [np.array(np.random.uniform(), dtype=default_dtype())
+    arrays = [np.array(np.random.uniform(low, high), dtype=default_dtype())
               if len(s) == 0 else np.random.uniform(low, high, size=s).astype(default_dtype())
               for s in shapes]
     if len(arrays) == 1:
--- a/python/mxnet/test_utils.py
+++ b/python/mxnet/test_utils.py
@@ -102,6 +102,16 @@ def random_arrays(*shapes):
     return arrays
 
 
+def random_uniform_arrays(*shapes, low=0.0, high=1.0):
+    """Generate some random numpy arrays."""
+    arrays = [np.array(np.random.uniform(), dtype=default_dtype())
+              if len(s) == 0 else np.random.uniform(low, high, size=s).astype(default_dtype())
+              for s in shapes]
+    if len(arrays) == 1:
+        return arrays[0]
+    return arrays
+
+
 def random_sample(population, k):
     """Return a k length list of the elements chosen from the population sequence."""
     assert 0 <= k <= len(population)
--- a/python/mxnet/test_utils.py
+++ b/python/mxnet/test_utils.py
@@ -102,13 +102,10 @@ def random_arrays(*shapes):
     return arrays
 
 
-def random_uniform_arrays(*shapes, low=0.0, high=1.0):
+def random_uniform_arrays(*shapes, low=0.0, high=1.0, dtype=default_dtype()):
     """Generate some random numpy arrays."""
-    arrays = [np.array(np.random.uniform(low, high), dtype=default_dtype())
-              if len(s) == 0 else np.random.uniform(low, high, size=s).astype(default_dtype())
+    arrays = [np.random.uniform(low, high, size=s).astype(dtype)
               for s in shapes]
-    if len(arrays) == 1:
-        return arrays[0]
     return arrays
 
 
--- a/src/nnvm/node_op_util.h
+++ b/src/nnvm/node_op_util.h
@@ -80,6 +80,12 @@ class NodeOpGen {
                                                    dependent_node->attrs.name + "_negative",
                                                    {x}, nullptr, &dependent_node)};
     }
+
+    nnvm::NodeEntry ones_like(const nnvm::NodeEntry &x) {
+        return nnvm::NodeEntry{mxnet::op::MakeNode("ones_like",
+                                                   dependent_node->attrs.name + "_oneslike",
+                                                   {x}, nullptr, &dependent_node)};
+    }
 };
 
 }  // namespace util
--- a/tests/python/unittest/test_higher_order_grad.py
+++ b/tests/python/unittest/test_higher_order_grad.py
@@ -159,12 +159,8 @@ def test_arccos():
 
     for dim in range(1, 5):
         shape = rand_shape_nd(dim)
-        array = random_arrays(shape)
-        # Hack: Decrease std_dev to make
-        # sure all elements
-        # are in range -1 to 1
-        # i.e. Domain of arccos
-        array *= 0.2
+        # Domain of arccos is [-1, 1]
+        array = random_uniform_arrays(shape, low=-1.0, high=1.0)
         check_second_order_unary(array, arccos, grad_grad_op)
 
 
--- a/tests/python/unittest/test_higher_order_grad.py
+++ b/tests/python/unittest/test_higher_order_grad.py
@@ -145,7 +145,7 @@ def test_arcsin():
     for dim in range(1, 5):
         shape = rand_shape_nd(dim)
         # Domain of arcsin is [-1, 1]
-        array = random_uniform_arrays(shape, low=-1.0, high=1.0)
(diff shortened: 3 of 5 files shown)


kshitij12345:
@apeforest @larroy @sxjscience Please review.

kshitij12345:
It is failing the test with multiple trials. Still needs to be fixed.

python/mxnet/test_utils.py
```diff
@@ -102,6 +102,16 @@ def random_arrays(*shapes):
     return arrays
 
 
+def random_uniform_arrays(*shapes, low=0.0, high=1.0):
+    """Generate some random numpy arrays."""
+    arrays = [np.array(np.random.uniform(low, high), dtype=default_dtype())
+              if len(s) == 0 else np.random.uniform(low, high, size=s).astype(default_dtype())
+              for s in shapes]
```
sxjscience: The if-condition is not necessary. We can directly write
```python
arrays = [np.random.uniform(low, high, size=s) for s in shapes]
```

sxjscience: Also, we'd better add a dtype flag to the function, which is None by default and set it to default_dtype() if it's None.

larroy: I don't like to return different types based on inputs, the function should just return a list of arrays. other than that PR looks fine and approach is good. thanks for the PR.

kshitij12345: @sxjscience Will do that.

@larroy Agreed even I don't like that. However note that `random_arrays` in `test_utils.py` has that. I wanted `random_uniform_arrays` to be a drop-in replacement for `random_arrays` if and where necessary. But I guess it makes sense to not worry about it.
https://github.com/apache/incubator-mxnet/blob/04c3eec3e8ea793614244e18a89c6b3a13f12d6a/python/mxnet/test_utils.py#L95-L102

tests/python/unittest/test_higher_order_grad.py
```diff
@@ -27,7 +27,8 @@
 from common import with_seed
 import mxnet
 from mxnet import nd, autograd, gluon
-from mxnet.test_utils import assert_almost_equal, random_arrays, rand_shape_nd, same
+from mxnet.test_utils import (
+    assert_almost_equal, random_arrays, random_uniform_arrays, rand_shape_nd, same)
```
apeforest: Do you still need import random_arrays in this case? Also can this new method replace random_arrays?

apeforest:
Thanks a lot for your contribution! Looks good to me except one question

tests/python/unittest/test_higher_order_grad.py
kshitij12345: Other tests are still using `random_arrays`, probably they can be replaced with `random_uniform_arrays` but will have to verify once.

kshitij12345:
@sxjscience @apeforest @larroy 

While trying to track the causes for failures, I found that there is an issue with the way we are computing  the second order gradients.

Sample : 
https://github.com/apache/incubator-mxnet/blob/49df604ac3bb156e446671403c86a3535af9615d/src/operator/tensor/elemwise_unary_op_trig.cc#L201-L207

While computing `x_grad`,  elements of `dydx` can be `0`, which leads to `Nan` at that position in our computation, instead of it being `0`.

Most of the failures that I have checked in CI seem to be of this case.
Example : #17362 where we expect 0 but end up with `Nan`
```
Items are not equal:

Error nan exceeds tolerance rtol=1.000000e-05, atol=1.000000e-20 (mismatch 0.154321%).

Location of maximum error: (1, 6, 2, 1), a=0.00000000, b=nan

 ACTUAL: array([[[[-0.03453423,  0.1168007 , -0.00583075],

         [ 0.02170432,  0.37017354,  0.09864384],

         [-0.02544087, -0.06709251, -0.34824234]],...

 DESIRED: array([[[[-0.03453423,  0.1168007 , -0.00583075],

         [ 0.02170432,  0.37017348,  0.09864385],

         [-0.02544087, -0.0670925 , -0.34824237]],...

*** Maximum errors for vector of size 648:  rtol=1e-05, atol=1e-20



  1: Error nan  Location of error: (1, 6, 2, 1), a=0.00000000, b=nan



--------------------- >> end captured stdout << ----------------------

common: INFO: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=402611671 to reproduce.

```

---------------------------------------------------------

**Mitigation** : 

The reason we divide by `dydx` is to get first gradient from previous computation instead of computing it again. 

One solution might be to replace 
```
auto x_grad = op.div(dydx_mul_grad_x, dydx); 
```
with actually computing the first degree gradient
```
auth x_grad = nnvm::NodeEntry{mxnet::op::MakeNode("_backward_arcsin",
                                                   dependent_node->attrs.name + "_mul_scalar",
                                                   {x}, &nullptr, &dependent_node)};
```
Will try this and update how this goes.

Let me know if you have any other idea.

kshitij12345:
Verification of Trig Operators with 
```
MXNET_TEST_COUNT=10000 nosetests ./tests/python/unittest/test_higher_order_grad.py:test_opname
```

* [ ] sin
* [ ] cos
* [ ] tan
* [ ] sinh
* [ ] cosh
* [ ] tanh
* [ ] arcsin
* [ ] arccos
* [ ] arctan
* [ ] arcsinh
* [ ] arccosh
* [ ] arctanh
* [ ] radians

kshitij12345:
Verification of remaining Unary Operators with 
```
MXNET_TEST_COUNT=10000 nosetests ./tests/python/unittest/test_higher_order_grad.py:test_opname
```

* [ ] relu
* [ ] log
* [ ] log2
* [ ] log10
* [ ] square
* [ ] expm1
* [ ] log1p
* [ ] reciprocal
* [ ] abs
* [ ] clip
* [ ] dropout
* [ ] sigmoid
* [ ] sqrt
* [ ] cbrt
* [ ] rsqrt
* [ ] rcbrt

kshitij12345:
@larroy @apeforest Is CI unstable at the moment? Tests are failing but the CI shows pending status.

apeforest:
Yes, we encountered (again) CI random timeout issue recently. Please bear with us and re-trigger CI again. Sorry for the inconvenience.

kshitij12345:
@apeforest Sure. No worries. Just wanted to confirm that it is not related to this PR.

larroy:
> @sxjscience @apeforest @larroy
> 
> While trying to track the causes for failures, I found that there is an issue with the way we are computing the second order gradients.
> 
> Sample :
> https://github.com/apache/incubator-mxnet/blob/49df604ac3bb156e446671403c86a3535af9615d/src/operator/tensor/elemwise_unary_op_trig.cc#L201-L207
> 
> While computing `x_grad`, elements of `dydx` can be `0`, which leads to `Nan` at that position in our computation, instead of it being `0`.
> 
> Most of the failures that I have checked in CI seem to be of this case.
> Example : #17362 where we expect 0 but end up with `Nan`
> 
> ```
> Items are not equal:
> 
> Error nan exceeds tolerance rtol=1.000000e-05, atol=1.000000e-20 (mismatch 0.154321%).
> 
> Location of maximum error: (1, 6, 2, 1), a=0.00000000, b=nan
> 
>  ACTUAL: array([[[[-0.03453423,  0.1168007 , -0.00583075],
> 
>          [ 0.02170432,  0.37017354,  0.09864384],
> 
>          [-0.02544087, -0.06709251, -0.34824234]],...
> 
>  DESIRED: array([[[[-0.03453423,  0.1168007 , -0.00583075],
> 
>          [ 0.02170432,  0.37017348,  0.09864385],
> 
>          [-0.02544087, -0.0670925 , -0.34824237]],...
> 
> -------------------- >> begin captured stdout << ---------------------
> 
> 
> 
> *** Maximum errors for vector of size 648:  rtol=1e-05, atol=1e-20
> 
> 
> 
>   1: Error nan  Location of error: (1, 6, 2, 1), a=0.00000000, b=nan
> 
> 
> 
> --------------------- >> end captured stdout << ----------------------
> 
> -------------------- >> begin captured logging << --------------------
> 
> common: INFO: Setting test np/mx/python random seeds, use MXNET_TEST_SEED=402611671 to reproduce.
> 
> --------------------- >> end captured logging << ---------------------
> ```
> 
> **Mitigation** :
> 
> The reason we divide by `dydx` is to get first gradient from previous computation instead of computing it again.
> 
> One solution might be to replace
> 
> ```
> auto x_grad = op.div(dydx_mul_grad_x, dydx); 
> ```
> 
> with actually computing the first degree gradient
> 
> ```
> auth x_grad = nnvm::NodeEntry{mxnet::op::MakeNode("_backward_arcsin",
>                                                    dependent_node->attrs.name + "_mul_scalar",
>                                                    {x}, &nullptr, &dependent_node)};
> ```
> 
> Will try this and update how this goes.
> 
> Let me know if you have any other idea.

looks good, I guess it's easier than have another operator like div_zero which I'm not sure we have.

kshitij12345:
@apeforest @sxjscience @larroy Can you please take a final look. Tests have finally passed.

src/nnvm/node_op_util.h
```diff
@@ -80,6 +80,12 @@ class NodeOpGen {
                                                    dependent_node->attrs.name + "_negative",
                                                    {x}, nullptr, &dependent_node)};
     }
+
+    nnvm::NodeEntry ones_like(const nnvm::NodeEntry &x) {
```
apeforest: Don't we already have an ones_like operator already for use? https://github.com/apache/incubator-mxnet/blob/971ea4f0adb5bd52cdcdee3e15f390ac05a3d379/src/operator/tensor/init_op.cc#L185

apeforest:
Thanks for your patience re-triggering CI for so many times. We are working on a new CI system and hope it can alleviate the current burden from developers. Stay tuned please :)

PR #17128, opened 2019-12-19
[DEBUG] enable custom error type

szha:
## Description ##
enable throwing custom error type from backend

With this patch, by prepending error messages with the error type, the frontend will automatically throw exception of that type if the type has been registered.

<details>
<summary>An example in a check in transpose and its effect</summary>

### Logging in C++

```cpp
    CHECK_EQ(axes_set.size(), axes.ndim()) << "ValueError: Repeated axis in transpose."
                                           << " param.axes = "
                                           << param.axes;
```

### Effect

```python
In [1]: from mxnet import np

In [2]: dat = np.random.normal(0, 1, (3, 4, 5))

In [3]: dat.transpose((0, 0, 1))
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-3ad259b4e371> in <module>
----> 1 dat.transpose((0, 0, 1))

~/mxnet-distro/mxnet-build/python/mxnet/numpy/multiarray.py in transpose(self, *axes)
   1460             elif axes[0] is None:
   1461                 axes = None
-> 1462         return _mx_np_op.transpose(self, axes=axes)
   1463
   1464     def flip(self, *args, **kwargs):

~/mxnet-distro/mxnet-build/python/mxnet/ndarray/register.py in transpose(a, axes, out, name, **kwargs)

~/mxnet-distro/mxnet-build/python/mxnet/_ctypes/ndarray.py in _imperative_invoke(handle, ndargs, keys, vals, out, is_np_op, output_is_list)
    105         c_str_array(keys),
    106         c_str_array([str(s) for s in vals]),
--> 107         ctypes.byref(out_stypes)))
    108
    109     create_ndarray_fn = _np_ndarray_cls if is_np_op else _ndarray_cls

~/mxnet-distro/mxnet-build/python/mxnet/base.py in check_call(ret)
    271     """
    272     if ret != 0:
--> 273         raise get_last_ffi_error()
    274
    275

ValueError: Traceback (most recent call last):
  File "src/operator/numpy/np_matrix_op.cc", line 77
ValueError: Check failed: axes_set.size() == axes.ndim() (2 vs. 3) : Repeated axis in transpose. param.axes = [0,0,1]

```

</details>

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
[...]

PR #17348, opened 2020-01-16
Fix Horovod build error due to missing exported symbols

apeforest:
## Description ##
Export MXAPIHandle symbol and fixes https://github.com/apache/incubator-mxnet/issues/17292

Here are the causes of the problem:

1.  Horovod uses MX_API_BEGIN() and MX_API_END() from mxnet/c_api_error.h to catch and throw errors in horovod APIs: https://github.com/horovod/horovod/blob/master/horovod/mxnet/mpi_ops.cc#L224
MX_API_BEGIN() is a macro that calls MXAPIHandleException https://github.com/apache/incubator-mxnet/blob/master/include/mxnet/c_api_error.h#L36. Before #17128, MXAPIHandleException is an inline function. And therefore when #17128 introduced a new function call NormalizeError() inside MXAPIHandleException it broke Horovod integration because the symbol of NormalizeError is not whitelist by MXNet distribution.
2. #17298 removed NormalizeError() from MXAPIHandleException and made it not inline. https://github.com/apache/incubator-mxnet/pull/17208/files#diff-875aa4c013dbd73b044531e439e8afddR67. This time the error becomes undefined symbol of MXAPIHandleException.

So to summarize, the problem is not that Horovod requires MXAPIHandleException function to be inline. The rootcause is that MXNet did not whitelist the symbol *MXAPIHandleException* but only whitelisted the symbols that are being used inside MXAPIHandleException function. It was okay when the function MXAPIHandleException is inline, but became a problem when it's not. A good practice is to whitelist symbol *MXAPIHandleException* instead of its internal functions.
## Comments ##
@szha Please help to review.

### DIFF
--- a/make/config/libmxnet.sym
+++ b/make/config/libmxnet.sym
@@ -1,6 +1,5 @@
-MX*
+*MX*
 NN*
-_MX*
 _NN*
 mx*
 nn*
@@ -12,4 +11,3 @@ Java_org_apache_mxnet*
 *Storage*Get*
 *on_enter_api*
 *on_exit_api*
-*MXAPIHandleException*
--- a/make/config/libmxnet.sym
+++ b/make/config/libmxnet.sym
@@ -12,4 +12,4 @@ Java_org_apache_mxnet*
 *Storage*Get*
 *on_enter_api*
 *on_exit_api*
-*MXAPISetLastError*
+*MXAPIHandleException*
--- a/make/config/libmxnet.sym
+++ b/make/config/libmxnet.sym
@@ -1,10 +1,7 @@
 *MX*
-NN*
-_NN*
-mx*
-nn*
-_mx*
-_nn*
+*NN*
+*mx*
+*nn*
 Java_org_apache_mxnet*
 *NDArray*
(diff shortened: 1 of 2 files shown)


make/config/libmxnet.sym
```diff
@@ -12,4 +12,4 @@ Java_org_apache_mxnet*
 *Storage*Get*
 *on_enter_api*
 *on_exit_api*
-*MXAPISetLastError*
+*MXAPIHandleException*
```
szha: Isn't this covered by the first line? If not, should we expand the pattern for the first line instead?

apeforest: I believe the first line misses the prefix before MX. Not sure if we want to whitelist all *MX* APIs.

szha: Let's fix the first line then

apeforest: Updated. Please review again. Thanks!

apeforest:
@szha Thanks for the review and update. Please let me know if this is good to ship.

PR #17358, opened 2020-01-17
np.broadcast_to extension

haojin2:
## Description ##
Extending `broadcast_to` to infer shape from input tensor, useful under symbolic mode.

### Essentials ###

### Changes ###

## Comments ##
FYI @szhengac @sxjscience @xidulu

### DIFF
--- a/src/operator/numpy/np_broadcast_reduce_op_value.cc
+++ b/src/operator/numpy/np_broadcast_reduce_op_value.cc
@@ -467,17 +467,21 @@ bool NumpyBroadcastToShape(const nnvm::NodeAttrs& attrs,
   mxnet::TShape& ishape = (*in_attrs)[0];
   if (!mxnet::shape_is_known(ishape)) return false;
   const BroadcastToParam& param = nnvm::get<BroadcastToParam>(attrs.parsed);
-  CHECK(mxnet::shape_is_known(param.shape))
-      << "the objective shape for broadcasting array must be known";
   CHECK_LE(ishape.ndim(), param.shape.ndim())
       << "shape " << ishape << " is not broadcastable to " << param.shape;
+  TShape pshape = param.shape;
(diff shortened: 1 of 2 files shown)


szhengac:
Great!

src/operator/numpy/np_broadcast_reduce_op_value.cc
```diff
@@ -467,17 +467,21 @@ bool NumpyBroadcastToShape(const nnvm::NodeAttrs& attrs,
...
-    CHECK(ishape[j] == param.shape[i] || ishape[j] == 1)
-        << "shape " << ishape << " is not broadcastable to " << param.shape;
+    if (pshape[i] == -2) {
+      pshape[i] = ishape[j];
+    }
+    CHECK(ishape[j] == pshape[i] || ishape[j] == 1)
+        << "shape " << ishape << " is not broadcastable to " << pshape;
   }
-  SHAPE_ASSIGN_CHECK(*out_attrs, 0, param.shape);
+  CHECK(mxnet::shape_is_known(pshape))
```
reminisce: I think `-1` is internally used as unknown dim size. It should be fine to use `-1` in  a destination shape set by users with similar meaning as in `npx.reshape`.

haojin2: @szhengac Are you okay with changing the api to use -1 for copying a dim?

szhengac: `-1` is for shape inference, which is different from copy. So I do not think we should use `-1`.

reminisce: `-1` does not mean copying dim size in NumPy; it actually means deducing shape with the available information. The point is we should not `CHECK(mxnet::shape_is_known(pshape))` since it implies that the destination shape is unknown which is different from NumPy semantics.

Issue #17352, opened 2020-01-17
Error in mkldnn unit test


apeforest:
## Description
When I ran nosetests -v tests/python/mkl/test_mkldnn.py, all the tests passed. 
If I ran nosetests -v tests/python/mkl/test_mkldnn.py:test_mkldnn_engine_threading, it failed with 
```
test_mkldnn.test_mkldnn_engine_threading ... ERROR

======================================================================
ERROR: test_mkldnn.test_mkldnn_engine_threading
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/.virtualenvs/mxnet/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/ubuntu/src/mxnet/tests/python/mkl/test_mkldnn.py", line 98, in test_mkldnn_engine_threading
    assert_almost_equal(y[0, 0, 0, 0], np.array(0.016711406))
  File "/home/ubuntu/src/mxnet/python/mxnet/test_utils.py", line 610, in assert_almost_equal
    aTmp[index] = bTmp[index] = 0
TypeError: 'numpy.float32' object does not support item assignment
-------------------- >> begin captured stdout << ---------------------

*** Maximum errors for vector of size 1:  rtol=1e-05, atol=1e-20

  1: Error 1920385.369517  Location of error: (), a=0.33763480, b=0.01671141

--------------------- >> end captured stdout << ----------------------
```
## To Reproduce
nosetests -v tests/python/mkl/test_mkldnn.py:test_mkldnn_engine_threading

apeforest:
@PatricZhao @TaoLv Did you encounter this before? Thanks

TaoLv:
Yes, we're aware of this problem as in our local validation system, python tests are split into single run for each test. We didn't dive into the problem as we thought it's a problem of nosetests or how the nosetests file is implemented. I remember these was an similar issue but cannot find it anymore.

TaoLv:
@juliusshufan

PR #17386, opened 2020-01-20
fixed seed for mkldnn test

heaseny:
## Description ##
This PR is intended to fix the issue in #17352 with fixed seed.
Fix https://github.com/apache/incubator-mxnet/issues/17352

### Essentials ###
- [ ] Changes are complete 

### Changes ###
- [ ] Use fixed seed 1234 for the unit test: test_mkldnn_ndarray_slice
- [ ] Use fixed seed 1234 for the unit test: test_mkldnn_engine_threading

### DIFF
--- a/tests/python/mkl/test_mkldnn.py
+++ b/tests/python/mkl/test_mkldnn.py
@@ -61,6 +61,7 @@ def test_mkldnn_model():
     except:  # pylint: disable=bare-except
         assert 0, "test_mkldnn_model exception in bind and execution"
 
+@with_seed(1234)
 def test_mkldnn_ndarray_slice():
     ctx = mx.cpu()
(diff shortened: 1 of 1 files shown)


tests/python/mkl/test_mkldnn.py
```diff
@@ -95,7 +97,7 @@ def __getitem__(self, key):
     for _ in loader:
         y = net(mx.nd.array(np.ones(X))).asnumpy()
         # output should be 016711406 (non-mkldnn mode output)
```
TaoLv: you should also change the comment line here.

TaoLv:
@apeforest This change should fix the problem in #17352. Could you please review and check?

PR #17387, opened 2020-01-20
[TVM][Windows] TVM Ops windows build

vexilligera:
## Description ##
Linux link and Windows works differently, so I had to paste contrib/tvmop/space.py to python/mxnet/space.py, otherwise python front-end installation fails since Windows recognizes the link file as .py file. If someone knows any better solution please inform me.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -134,7 +134,7 @@ if(MSVC)
   if(USE_F16C)
     message("F16C instruction set is not yet supported for MSVC")
   endif()
-  set(CMAKE_C_FLAGS "/MP")
+  # set(CMAKE_C_FLAGS "/MP")
   set(CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} /bigobj")
 else()
   include(CheckCXXCompilerFlag)
@@ -419,7 +419,7 @@ if(USE_OPENMP)
   # could a Windows user please fix?
   if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/openmp/CMakeLists.txt
      AND SYSTEM_ARCHITECTURE STREQUAL "x86_64"
-     AND NOT MSVC
+     # AND NOT MSVC
      AND NOT CMAKE_CROSSCOMPILING)
     load_omp()
     list(REMOVE_ITEM mxnet_LINKER_LIBS iomp5)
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -134,7 +134,7 @@ if(MSVC)
   if(USE_F16C)
     message("F16C instruction set is not yet supported for MSVC")
   endif()
-  set(CMAKE_C_FLAGS "/MP")
+  # set(CMAKE_C_FLAGS "/MP")
```
leezu: Why this change?

python/mxnet/space.py
```diff
@@ -1 +1,214 @@
-../../contrib/tvmop/space.py
```
leezu: ?

vexilligera: Linux link and Windows shortcut work differently, so I had to paste contrib/tvmop/space.py to python/mxnet/space.py, otherwise python front-end installation fails since Windows recognizes the link file as .py file.

CMakeLists.txt
```diff
@@ -419,7 +419,7 @@ if(USE_OPENMP)
   # could a Windows user please fix?
   if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/openmp/CMakeLists.txt
      AND SYSTEM_ARCHITECTURE STREQUAL "x86_64"
-     AND NOT MSVC
+     # AND NOT MSVC
```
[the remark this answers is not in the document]
vexilligera: This can be built on Windows correctly. I will remove the comment later.

CMakeLists.txt
vexilligera: /MP caused some errors earlier, I will take a look later and add this back if it works properly.

CMakeLists.txt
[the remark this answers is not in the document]
leezu: Ok. Please then also remove `could a Windows user please fix?` comment

CMakeLists.txt
vexilligera: /MP causes error if the compiler is Clang-cl. I will enable the option if the compiler is MSVC.

python/mxnet/space.py
leezu: Would it be possible to fix the root cause instead? Duplicating the file is not a good strategy as such duplicates tend to get out of sync

vexilligera: @hzfan is there any alternative? A workaround I'm using now is to copy the file when running prepare_tvm.ps1 script.

hzfan: I think copying the file at compile time is good. Why not remove this `python/mxnet/space.py` since it will be generated by your workaround? This way we can avoid duplication.

Also, just FYI, if we go with copying, we may have to make some extra modification to ci. Now ci separates the compiling phase and the testing phase. CI will copy the output of compilation (like libmxnet.so) to the testing phase. In our case, `python/mxnet/space.py` becomes part of the compilation output and must be copied.

vexilligera: @hzfan If I remove space.py, ONNX build failed probably because it was built before TVM. I think we need a better way of doing this.

hzfan: @vexilligera Could you try adding `python/mxnet/space.py` to `Jenkins_steps.groovy`, so that it will be copied from compilation phase to testing phase

Issue #17389, opened 2020-01-21
Indexing ndarray should give empty array output


reminisce:
## Description
When indexing an ndarray with an index greater than the dim size, it should produce an empty ndarray as in NumPy, instead of throwing an error.

```python
>>> import mxnet as mx
>>> mx.npx.set_np()
>>> a = mx.np.arange(1)
>>> a
array([0.])
>>> a[1:]  # index exceeds dim size, should result in an empty array instead of throwing an error
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ubuntu/unison/mxnet4/cpu/python/mxnet/numpy/multiarray.py", line 663, in __getitem__
    return self._slice(key.start, key.stop)
  File "/home/ubuntu/unison/mxnet4/cpu/python/mxnet/ndarray/ndarray.py", line 1343, in _slice
    start, stop, _ = _get_index_range(start, stop, self.shape[0])
  File "/home/ubuntu/unison/mxnet4/cpu/python/mxnet/ndarray/ndarray.py", line 3100, in _get_index_range
    raise IndexError('Slicing start %d exceeds limit of %d' % (start, length))
IndexError: Slicing start 1 exceeds limit of 1
>>> a.asnumpy()[1:]  # Okay in NumPy
array([], dtype=float32)
>>> a
array([0.])
>>> 
```
@Alicia1529

reminisce:
Fixed in https://github.com/apache/incubator-mxnet/pull/17391

PR #17391, opened 2020-01-21
fix issue #17389: Indexing ndarray should give empty array output

Alicia1529:
This pr is to fix the indexing problem in issue #17389: Indexing ndarray should give empty array output

### DIFF
--- a/python/mxnet/ndarray/ndarray.py
+++ b/python/mxnet/ndarray/ndarray.py
@@ -3095,9 +3095,9 @@ def _get_index_range(start, stop, length, step=1):
     elif start < 0:
         start += length
         if start < 0:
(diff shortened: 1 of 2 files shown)


python/mxnet/ndarray/ndarray.py
```diff
@@ -3110,9 +3110,9 @@ def _get_index_range(start, stop, length, step=1):
     elif stop < 0:
         stop += length
         if stop < 0:
-            raise IndexError('Slicing stop %d exceeds limit of %d' % (stop-length, length))
+            stop = 0
```
reminisce: Should `stop` be -1? Otherwise, a[0] will not be included in the sliced array.

PR #17413, opened 2020-01-23
cmake: remove outdated comments

leezu:


CMakeLists.txt
```diff
@@ -674,13 +677,7 @@ if(USE_PLUGIN_CAFFE)
...
 endif()
 
-if(NOT MSVC)
-  # Only add c++11 flags and definitions after cuda compiling
-  add_definitions(-DDMLC_USE_CXX11)
-  add_definitions(-DMSHADOW_IN_CXX11)
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
-else()
+if(MSVC)
```
ChaiBapchya: previously it was added after cuda compiling
1. why was it done before?
2. why is it no longer needed?
Thanks

leezu: I'm not sure why it was done before. @jermainewang added the comment and may be able to clarify.

leezu: It's no longer needed as the build works fine without it.

PR #17430, opened 2020-01-24
cmake: add config files similar to make/config.mk

leezu:
## Description ##
Add config files similar to make/config.mk and update the `ubuntu_setup` guide.

For cmake build, the following features are currently unsupported (in comparison to `make/config.mk` Makefile based compillation). Unlike makefile based compilation, the cmake build is cross platform.

- LIBJPEG_TURBO
- NNPACK
- HDFS (hadoop)
- S3

CC @eric-haibin-lin @apeforest @yajiedesign


config/config.cmake
```diff
@@ -0,0 +1,143 @@
...
+set(USE_CUDA ON CACHE BOOL "Build with CUDA support")
+
+# Target NVIDIA GPU achitecture.
+# Format: Auto | Common | All | LIST(ARCH_AND_PTX ...)
+# - "Auto" detects local machine GPU compute arch at runtime.
+# - "Common" and "All" cover common and entire subsets of architectures
+# - ARCH_AND_PTX : NAME | NUM.NUM | NUM.NUM(NUM.NUM) | NUM.NUM+PTX
+# - NAME: Fermi Kepler Maxwell Kepler+Tegra Kepler+Tesla Maxwell+Tegra Pascal Volta Turing
+# - NUM: Any number. Only those pairs are currently accepted by NVCC though:
+#        2.0 2.1 3.0 3.2 3.5 3.7 5.0 5.2 5.3 6.0 6.2 7.0 7.2 7.5
```
eric-haibin-lin: Why is there `NUM.NUM(NUM.NUM)` ?

config/config.cmake
```diff
@@ -0,0 +1,143 @@
...
+
+# Building with lapack is only effective when compiled with any of
+# openblas/apple/atlas/mkl blas versions
+set(USE_LAPACK ON CACHE BOOL "Build with lapack support")
+
+
+#---------------------------------------------
+# CPU instruction sets: The support is autodetected if turned ON
+#---------------------------------------------
+set(USE_SSE ON CACHE BOOL "Build with x86 SSE instruction support")
```
eric-haibin-lin: Are these two frequently used configs?

config/config.cmake
```diff
@@ -0,0 +1,143 @@
...
+set(USE_F16C ON CACHE BOOL "Build with x86 F16C instruction support")
+
+
+#----------------------------
+# distributed computing
+#----------------------------
+set(USE_DIST_KVSTORE OFF CACHE BOOL "Build with DIST_KVSTORE support")
+# set(USE_HDFS OFF CACHE BOOL "Allow read and write to HDFS directly; requires hadoop")  # TODO Not yet supported in CMake build
+# # path to libjvm.so. required if USE_HDFS=1
+# LIBJVM=$(JAVA_HOME)/jre/lib/amd64/server
```
eric-haibin-lin: what are these commented lines?

config/config.cmake
```diff
@@ -0,0 +1,143 @@
...
+# When compiling on a machine without GPU, autodetection will fail and you
+# should instead specify the target architecture manually. If no architecture is
+# detected or specified, compilation will target all available architectures,
+# which significantly slows down the build .
+set(MXNET_CUDA_ARCH "Auto" CACHE STRING "Target NVIDIA GPU achitecture")
+
+set(ENABLE_CUDA_RTC OFF CACHE BOOL "Build with CUDA runtime compilation support")
+set(USE_CUDNN ON CACHE BOOL "Build with cudnn support, if found")
+# set(USE_NVTX ON CACHE BOOL "Build with NVTX support")  # TODO Currently always autodetected in CMake
+set(USE_NCCL "Use NVidia NCCL with CUDA" OFF)
```
eric-haibin-lin: We can probably remove it since nccl kvstore is seldom used

config/config.cmake
```diff
@@ -0,0 +1,143 @@
...
+
+# set(CMAKE_C_COMPILER "" CACHE BOOL "C compiler")
+# set(CMAKE_CXX_COMPILER "" CACHE BOOL "C++ compiler")
+# set(CMAKE_CUDA_COMPILER "" CACHE BOOL "Cuda compiler (nvcc)")
+
+# Uncomment the following line to compile with debug information
+# set(CMAKE_BUILD_TYPE Debug CACHE STRING "CMake build type")
+
+# Whether to build operators written in TVM
+set(USE_TVM_OP OFF CACHE BOOL "Enable use of TVM operator build system.")
```
eric-haibin-lin: Is it possible to reorder the configurations in this cmake, so that the frequently used ones appear first? For instance: USE_CUDA, USE_CUDNN, USE_MKLDNN

config/config.cmake
leezu: `# TODO Not yet supported in CMake build`

config/config.cmake
leezu: Order follows `config.mk`. We can change it

config/config.cmake
leezu: It's a feature of upstream https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindCUDA/select_compute_arch.cmake

Let's document it better (the documentation you mention is taken from upstream, but leaves room for improvement). For example, `"7.0(6.0)"` translates to `-gencode;arch=compute_60,code=sm_70`.

config/config.cmake
leezu: Let's move it to the bottom of the file into a category of seldomly used features.

config/config.cmake
leezu: I expect there are some power users that need it. In particular, they are important when you build on a different machine compared to where you run MXNet. In that case autodetection may find F16C is available, but on your target machine it's not. 
Currently the selection of options is based on what is included in `config.mk`.

config/config.cmake
```diff
@@ -0,0 +1,129 @@
...
+#---------------------------------------------
+# Common libraries
+#---------------------------------------------
+set(USE_OPENCV ON CACHE BOOL "Build with OpenCV support")
+set(OPENCV_ROOT "" CACHE BOOL "OpenCV install path. Supports autodetection.")
+
+set(USE_OPENMP ON CACHE BOOL "Build with Openmp support")
+
+set(USE_MKL_IF_AVAILABLE ON CACHE BOOL "Use Intel MKL if found")
+set(USE_MKLDNN ON CACHE BOOL "Build with MKL-DNN support")
```
TaoLv: How does this work with the if condition here: https://github.com/apache/incubator-mxnet/blob/master/CMakeLists.txt#L40? Will the check be skipped?

leezu: The check is not skipped, but the `option` call becomes a no-op.

> In CMake 3.13 and above the option() command prefers to do nothing when a normal variable of the given name already exists. It does not create or update a cache entry or remove the normal variable. The new behavior is consistent between the first and later runs in a build tree. This policy provides compatibility with projects that have not been updated to expect the new behavior.

https://cmake.org/cmake/help/latest/policy/CMP0077.html

config/config.cmake
```diff
@@ -0,0 +1,129 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+#-------------------------------------------------------------------------------
+#  Template configuration for compiling MXNet
+#
+#  If you want to change the configuration, please use the following
+#  steps. Assume you are on the root directory of mxnet. First copy the this
```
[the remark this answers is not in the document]
leezu: Thanks. The typo was copy-pasted from https://github.com/apache/incubator-mxnet/blob/47349ce50004d52af7844ea7c3c509de96678928/make/config.mk#L22

config/config.cmake
leezu: Could you advise if `if(USE_MKL_IF_AVAILABLE AND (NOT APPLE) AND (NOT MSVC) AND (CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "x86_64") AND (NOT CMAKE_CROSSCOMPILING))` condition is correct?

Is MKL-DNN unsupported on Apple and MSVC?
Ideally we can use MKL-DNN for builds for all x86_64 platforms and thus the `config.cmake` file introduced in this PR is valid for all of these.

TaoLv: MKL-DNN should also support Apple and MSVC. See https://github.com/intel/mkl-dnn/#runtime-dependencies.
Previously we decided to enable MKL-DNN by default only on linux machines, so Apple and MSVC are excluded. One note is that, if MSVC, `CMAKE_HOST_SYSTEM_PROCESSOR` may not be `x86_64`. Also if we want to default MKL-DNN on all platforms, similar change should be applied to makefile too.

PR #17452, opened 2020-01-27
Add test cases for indexing ndarrays with zero-dim array

reminisce:
## Description ##
Fix the problem of indexing ndarrays with zero-dim arrays as indices not treated as integers. After the fix, the following should be equivalent.
```python
>>> from mxnet import np, npx
>>> npx.set_np()
>>> a = np.array([1, 2, 3], dtype='int32')
>>> b = np.arange(3, dtype='int32')
>>> b[1] == b[a[0]]
```

@haojin2 @sxjscience @astonzhang


python/mxnet/ndarray/ndarray.py
```diff
@@ -3055,6 +3060,8 @@ def get_indexing_dispatch_code(key):
                 return _NDARRAY_EMPTY_TUPLE_INDEXING
             if getattr(idx, 'dtype', None) == np.bool_:
                 num_bools += 1
+            if getattr(idx, 'ndim', None) == 0:  # zero-dim array as index treated as integer
```
sxjscience: Should we support indexing with float32 0-dim array?

```python
import numpy as np
a = np.ones((2, 2))
b = np.array(1.0)
a[b]
```
Currently, numpy will raise an error.

haojin2: Currently we don's support either:
```
>>> from mxnet import np, npx
>>> npx.set_np()
>>> a = np.ones((2, 2))
>>> b = np.array(1.0)
>>> a[b]
  File "/home/ubuntu/4-mxnet/python/mxnet/numpy/multiarray.py", line 675, in __getitem__
    indexing_dispatch_code = get_indexing_dispatch_code(key)
  File "/home/ubuntu/4-mxnet/python/mxnet/ndarray/ndarray.py", line 3064, in get_indexing_dispatch_code
    ''.format(idx, type(idx))
ValueError: NDArray does not support slicing with key 1.0 of type <class 'float'>.
```
Do you have some existing use case where the indices could be float types? If so we could try to add support for that.

sxjscience: I think we should raise an error if it's float32. So it should be good if the new implementation also raises an error.

haojin2: ok

sxjscience: @haojin2 I realized that we should have already handled the ndim=0 case. In the `indexing_key_expand_implicit_axes` function, we will convert the 0-dim array to python types by call `.item()`.

https://github.com/apache/incubator-mxnet/blob/910c608f682a47fc2c43375b5f5a426b563e5821/python/mxnet/ndarray/ndarray.py#L2975

We will always call `indexing_key_expand_implicit_axes` before `get_indexing_dispatch_code`.

reminisce: @sxjscience I thought it's not supported. The error message reported by @mli is actually about misuse of float numbers as integers which is consistent with NumPy's behavior. I will remove the code in ndarray while keeping the test cases in the PR.

PR #17466, opened 2020-01-29
Bert gemms true fp16

MoisesHer:
## Description ##
This PR allows true FP16 in CUBLAS gemms if environment variable MXNET_FC_TRUE_FP16 is set to true

### Essentials ###
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator):
tests/python/gpu/test_gluon_gpu:test_gemms_true_fp16

### Changes ###
- [x] General half_precision Gemm: allows true fp16 if MXNET_FC_TRUE_FP16 is set
- [x] Transformer: allows true fp16 if MXNET_FC_TRUE_FP16 is set


src/operator/linalg_impl.h
```diff
@@ -261,25 +262,43 @@ void linalg_gemm<gpu, mshadow::half::half_t>(const Tensor<gpu, 2, mshadow::half:
...
+  using TrueFP16Type = mshadow::half::half_t;
+  using PseudoFP16Type = typename CublasType<mshadow::half::half_t>::ScaleType;
+  TrueFP16Type trueFP16_alpha = static_cast<TrueFP16Type>(alpha);
+  TrueFP16Type trueFP16_beta = static_cast<TrueFP16Type>(beta);
+  PseudoFP16Type pseudoFP16_alpha = static_cast<PseudoFP16Type>(alpha);
+  PseudoFP16Type pseudoFP16_beta = static_cast<PseudoFP16Type>(beta);
+  const void *alpha_ptr;
+  const void *beta_ptr;
+  cudaDataType_t computeType;
+  bool use_true_fp16 = dmlc::GetEnv("MXNET_FC_TRUE_FP16", false);
```
eric-haibin-lin: For new env vars, please document in https://github.com/apache/incubator-mxnet/blob/master/docs/static_site/src/pages/api/faq/env_var.md

docs/static_site/src/pages/api/faq/env_var.md
```diff
@@ -353,6 +353,10 @@ If ctypes is used, it must be `mxnet._ctypes.ndarray.NDArrayBase`.
   - Values: 0(false) or 1(true) ```(default=1)```
   - This variable controls whether to use the MKL-DNN backend in fused RNN operator for CPU context. There are two fusion implementations of RNN operator in MXNet. The MKL-DNN implementation has a better performance than the naive one, but the latter is more stable in the backward operation currently.
 
+* MXNET_FC_TRUE_FP16
+  - Values: 0(false) or 1(true) ```(default=0)```
+  - If this variable is set to true, MXNet will performs true FP16 computation in CUBLAS gemms when input datatype is float16.
```
ptrendx: Could you describe it slightly more? I don't think average user will understand what "true fp16" means. Maybe something like this: 
If this variable is set to true, MXNet will perform fp16 accumulation when using cuBLAS and input datatype is set to float16. This could increase the speed of the computation, but might result in loss of accuracy. This makes this setting useful mainly for inference usecases.

@eric-haibin-lin What do you think?

ptrendx:
What is the GPU arch of CI? I believe it could be Kepler still, where this would not work. We need to check for the gpu architecture before enabling fp16 compute.

PR #17448, opened 2020-01-27
cmake: staticbuild for binary distribution

leezu:
## Description ##
Add staticbuild scripts for binary distribution based on CMake.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [X] Changes are complete (i.e. I finished coding on this PR)
- [X] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [X] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at https://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [X] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [X] cmake: staticbuild for binary distribution

Issue #14664, opened 2019-04-10
mxnet-cu100mkl "illegal instruction" on CPU


mjpost:
On my desktop (with *no* GPU), with CUDA 10 libraries loaded, when I attempt to `import mxnet` in Python, I get the following error:

```
$ pip list | grep mxnet
mxnet-cu100mkl    1.3.1     
$ python
Python 3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 01:22:34) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import mxnet
Illegal instruction
$
```

But everything works fine in another conda environment with `mxnet-92mkl` and CUDA 9.2 libraries loaded:

```
$ pip list | grep mxnet
mxnet-cu92mkl    1.3.1     
$ python
Python 3.6.6 |Anaconda, Inc.| (default, Jun 28 2018, 17:14:51) 
[GCC 7.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import mxnet
>>>
```

Is there any advice how to fix this? I cannot find a similar issue filed.

mjpost:
I also tried upgrading to MXNet 1.4.0, but got the same "Illegal instruction" problem.

wkcn:
Why do you install the GPU version in a machine without GPU? It seems that the program crashes since CUDA library files not found.
You can install the CPU version of MXNet: mxnet-mkl

mjpost:
The GPU libraries are available and in my path. If I unload them, I get a different message from the linker.

The reason is that I am submitting jobs to a cluster. I have one job that uses MXNet but just for data preparation so it doesn't need a GPU. It's a pain to have to maintain separate environments. As long as the libraries are found, it seems MXNet should be able to work smoothly with a CPU device, instead of a GPU one?

Plus, this used to work fine.

wkcn:
I will test it. I have the environment with CUDA 10 and without GPU.

mjpost:
Thank you!

wkcn:
Sorry that the version of CUDA in my machine is 10.1.105. It seems that there is no any related pre-build 
 MXNet package.

lanking520:
@szha

szha:
Since cu92mkl and cu100mkl only differs in cuda/cudnn versions, I'm guessing this is related to cudnn, since the library is statically linked. @mjpost, does it work if you try to build from source? You can find the configuration in #8671

[...]

PR #17468, opened 2020-01-29
Add -march=native -mtune=native to config/config.cmake

leezu:
## Description ##
Add settings to `config/config.cmake` to optimize for local CPU architecture.

### Changes ###
- [X] Add -march=native -mtune=native to config/config.cmake

## Comments ##
Related https://github.com/apache/incubator-mxnet/pull/17047

@TaoLv would providing different default `config/config.cmake` files for different architectures with the respective compiler flags be an easier way forward than adding flags like `USE_X86_ARCH` to the `CMakeLists.txt`. What do you think?

For example, for ARM, the config file would instead contain 
```
# CPU Architecture to optimize for. With default settings, your build may not
# work on earlier CPUs than the one used by the system you are compiling on.
# To work with older CPU, set the -march, -mtune variables below to the
# respective architecture name of the target CPU (see GCC documentation).
set(CMAKE_C_FLAGS "-mcpu=native" CACHE STRING "C compiler flags")
set(CMAKE_CXX_FLAGS "-mcpu=native" CACHE STRING "C++ compiler flags")
```
as per https://github.com/apache/incubator-mxnet/pull/17047#issuecomment-565274798, https://community.arm.com/developer/tools-software/tools/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu


TaoLv:
Thank you for adding this! This needs to modify the config.cmake carefully if we want to build binary for another platform, ie. using cmake for the convenient releases.

leezu:
For the binary distribution I suggest to use another set of config files. Please take a look at https://github.com/apache/incubator-mxnet/pull/17448

TaoLv:
It would also change the default behavior of cmake build - previously it can be built on a machine and run on a lower end machine, but now it probably will lead to illegal instruction error?

leezu:
`cmake -C ../config.cmake` didn't exist previously, in that sense we don't break the default behavior. The config file is a new feature in MXNet 1.7 and we're free to set the defaults (IMO).

You are right that there may be people that copy the `config.cmake` file, don't read it and run into this issue when copying `libmxnet.so` to a different computer.

We can comment out the lines introduced by this PR by default to decrease the likelihood of running into the issue you describe. Even thene https://github.com/apache/incubator-mxnet/issues/14664 would mean that `libmxnet.so` would stop working on CPU without AVX.

Would you recommend to comment out the `-march=native -mtune=native` by default?

TaoLv:
@leezu Could you please rebase? Then we can merge if CI passes. Because I see the last validation is 7 days ago.

leezu:
I plan to merge this after the static build instructions are improved.

leezu:
As the staticbuild is now easy to do via `python ci/build.py --platform centos7_cpu /work/runtime_functions.sh build_static_libmxnet cpu`, this PR is ready

leezu:

PR #17491, opened 2020-01-30
Remove dilation restriction for conv3d

reminisce:
## Description ##
Dilation > 1 was restricted due to legacy cudnn versions (older than 6.0.20). The restriction is removed for modern cudnn versions.

### DIFF
--- a/tests/python/unittest/test_operator.py
+++ b/tests/python/unittest/test_operator.py
@@ -2600,6 +2600,10 @@ def test_convolution_dilated_impulse_response():
     for dil in [ (1,1), (2,2), (3,3) ]:
         for ks in [ (3,3), (4,4), (2,3), (3,2), (1,1) ]:
             test_run_convolution_dilated_impulse_response(dil=dil, kernel_shape=ks)
+    # 3D
+    for dil in [ (1,1,1), (2,2,2), (3,3,3) ]:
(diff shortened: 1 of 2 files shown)


tests/python/unittest/test_operator.py
```diff
@@ -2600,6 +2600,10 @@ def test_convolution_dilated_impulse_response():
     for dil in [ (1,1), (2,2), (3,3) ]:
         for ks in [ (3,3), (4,4), (2,3), (3,2), (1,1) ]:
             test_run_convolution_dilated_impulse_response(dil=dil, kernel_shape=ks)
+    # 3D
+    for dil in [ (1,1,1), (2,2,2), (3,3,3) ]:
```
wkcn: Is it supported for the dilation `(0, 1, 1)` and `(1, 0, 0)`?

reminisce: Dilation is required to be an integer greater than 0 in mxnet.
https://github.com/apache/incubator-mxnet/blob/a726c406964b9cd17efa826738a662e09d973972/src/operator/nn/convolution.cc#L240-L241

If it's zero, the kernel size becomes 1. Is there a use case where dilation needs to be zero? I don't see any technical reasons of not being able to do so though.

wkcn: Sorry that I treated dilation as padding by mistake.

Issue #16438, opened 2019-10-11
Prebuilt scala package for 2.12


clouds56:
## Description
I found there's prebuilt jar for scala 2.11, now 2.12 is popular and 2.13 has come out. I think it is a good time to provide prebuilt package for scala 2.12 (or even make it the default one).
I tried to built it myself, but I have to keep multiple java version, since the tool required for build `javah` is removed in jdk 10 (I'm using jdk 12).

## Environment info (Required)

For Scala user, please provide:
1. Java version: (`java -version`) 12.0.2+10
2. Maven version: (`mvn -version`) 3.6.1
3. Scala runtime if applicable: (`scala -version`) 2.12.8

## Build info (Required if built from source)

Compiler (gcc/clang/mingw/visual studio): clang 8.0.1

MXNet commit hash: 15ea40d9bf7da3c4618ca45a6d023d9b0fb1c295

ddavydenko:
@mxnet-label-bot , add [Scala, FeatureRequest]

atiqsayyed:
@yzhliu  @nswamy  
I've  locally build 2.12 version, I want to help with this issue but I need some guidance about the current build package and how to go ahead with it.
Could you guys please guide me as to how can i start on this??

zachgk:
@atiqsayyed If you want to start on this, we mostly need to update the scala maven build. Right now, we have a profile "scala-2.11". We would want to create a similar profile for scala-2.12 to switch between the versions. We should move the scala.version and scala.binary.version properties into those profiles as well so it can be changed. Then, see if it can actually build both versions.

hmf:
Newbie to MxNet,

Can anyone tell me if it is possible to build for 2.13? 

TIA

PR #17503, opened 2020-02-01
Add Scala 2.12 and 2.13 cross-compilation (#16438)

fomkin:
## Description ##

Currently MXNet support only ancient Scala 2.11. I found mentions of try to add support of Scala 2.12 but looks like the try was failed despite the fact that PR was merged. This PR is another try.

## Comments ##
- API compatibility was completely saved
- Add `scala-2.11`,  `scala-2.12` and `scala-2.13` Maven profiles. Now one of the profiles should be active. This condition is checked by `maven-enforcer-plugin`
- Add Support of Scala 2.13 collections to C++ JNI code
- Module "Examples" builds only for 2.11. Dependencies were obsolete and there is no way to port examples without rewriting of logic (this is out of scope of the PR)
- Module "Spark ML" builds only for 2.11. It depends on Spark 1.6.3 which supports only 2.11
- *Possible performance degradation in 2.13 version*. Scala 2.13 has new collections library. Default `Seq` is immutable now, so I need to ivoke `toSeq` for each `ArrayBuffer` to copy it to immutable sequence

### Essentials ###

### To solve ###

1. [ ] I did something and `BertQA` starts to compile (looks like it was ignored before). It breaks compilation of examples. I don't know what to do. I'm nood in neural networks and not know how no modify the code to make it compile.
2. [ ] Makefile and CI integration.


leezu:
Just FYI, there are currently some issues with the CI. So not every failure of the tests you observe is due to your changes.

I suggest you also take a look at https://github.com/apache/incubator-mxnet/issues/16167#issue-493529102 . For example, the Module "Examples" you mention above would likely be rendered obsolete as part of MXNet 2. 

> In terms of frontend, this roadmap focuses mostly on Python-frontend since MXNet has been taking a Python-first approach. The expectation with respect to other language bindings is that they would evolve along with the backend evolution and make use of the improvements. Given that breaking changes can occur, maintainers of different language bindings are expected to participate in related interface definition discussions.

fomkin:
Examples do not compile because I've updated `scala-maven-plugin` to the latest version. It fixes some cases where *.java files are ignoring during compilation. May I remove examples from the build? It will fix CI.

leezu:
Please CC the persons who contributed the respective example (see git history of the file) in this issue.

scala-package/assembly/pom.xml
```diff
@@ -26,7 +26,7 @@
     <relativePath>../pom.xml</relativePath>
   </parent>
 
-  <artifactId>mxnet-full_2.11</artifactId>
+  <artifactId>mxnet-full</artifactId>
```
zachgk: Is there a reason you are removing the _{scala.binary.version} from the internal artifactId? This could end up breaking other users

scala-package/core/src/test/scala/org/apache/mxnet/NDArraySuite.scala
```diff
@@ -357,24 +357,24 @@ class NDArraySuite extends FunSuite with BeforeAndAfterAll with Matchers {
...
       assert(CheckUtils.reldiff(result1.toArray, range1.toArray) <= 1e-4f)
 
-      val result2 = (0.0 until stop.toDouble by step.toDouble)
+      val result2 = (BigDecimal(0.0) until BigDecimal(stop) by BigDecimal(step))
         .flatMap(x => Array.fill[Float](repeat)(x.toFloat))
       val range2 = NDArray.arange(stop, step = step, repeat = repeat)
       assert(CheckUtils.reldiff(result2.toArray, range2.toArray) <= 1e-4f)
 
-      val result3 = 0f to stop by 1f
+      val result3 = (BigDecimal(0) until BigDecimal(stop) by BigDecimal(1)).map(_.toFloat)
```
zachgk: BigDecimal(0) **to** BigDecimal(stop)?

scala-package/pom.xml
```diff
@@ -423,23 +590,11 @@
       <version>1.7.7</version>
       <scope>provided</scope>
     </dependency>
-    <dependency>
-      <groupId>org.scalatest</groupId>
-      <artifactId>scalatest_2.11</artifactId>
```
zachgk: Can you try to keep the dependencies in the dependency section so it is easier to keep track of them. For scalatest, you could just use `<artifactId>scalatest_${scala.binary.version}</artifactId>`. If you need to vary the version, you could use a property.

scala-package/assembly/pom.xml
fomkin: This is an artifactId in the Maven _reactor_ scope. Concrete Scala version suffix adds on deploy/install phase. It means that artifact published to the Central will have Scala version suffix. See `deploy/pom.xml`

scala-package/pom.xml
fomkin: OK. But `scala-parallel-collections` in `core/pom.xml` will remain in profile dependencies because in 2.11 and 2.12 it is part of standard library.

fomkin: I tried to do this. But it doesn't working. Looks like `maven-remote-resources-plugin` doesn't understand properties in dependencies. Here is error log https://gist.github.com/fomkin/a26aefc59eaca1148dfed3719a8144f0

fomkin:
@zachgk 
> In BertQA.java lines 77 and 79, try to use null as the second argument to the softmaxParam constructor.

Thanks. Now it compiles.

scala-package/core/src/test/scala/org/apache/mxnet/NDArraySuite.scala
fomkin: Thanks!

scala-package/pom.xml
fomkin: Well. It solves by removing _default_ properties: `maven-remote-resources-plugin` ignores the property from a profile if it has default values.

fomkin:
But throws exception :face_with_head_bandage:

scala-package/assembly/pom.xml
fomkin: When install scala-package locally, `~/.m2/repository` will contain artifact without suffix. I sure that it advanced usage option, and user know what he or she doing. At least he or she explicitly sets profile with Scala version.

gigasquid:
> @fomkin Removing the examples may fix the CI now, but it will be bad for the project down the line as those examples would no longer be maintained. It is odd that those examples were not working. I looked at the current problem. In BertQA.java lines 77 and 79, try to use `null` as the second argument to the softmaxParam constructor. If this doesn't work or other examples are broken, I can help fix them or you can ask whoever created the example (see git blame).
> 
> @gigasquid Would this affect clojure at all?

No - That shouldn't affect Clojure. Thanks for thinking about us though :)

fomkin:
@gigasquid May you look at this http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-cpu/detail/PR-17503/7/pipeline#step-600-log-1655 ? What I had do wrong?

fomkin:
@lanking520 may you look at `BertQA.java`? The code use obsolete API. It never compiled in CI before because of bug in `scala-maven-plugin`. Now it compiles and breaks the build. I tried to fix this giving same NDArray as length to softmaxParam. It became compilable but tests fails with the error: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-17503/7/pipeline#step-804-log-1708

gigasquid:
> @gigasquid May you look at this http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-cpu/detail/PR-17503/7/pipeline#step-600-log-1655 ? What I had do wrong?

I stand corrected. It does affect the Clojure package as well. A library we use depends on the old version of scala and would need to be forked/upgraded to work https://github.com/t6/from-scala/blob/master/project.clj

TaoLv:
You may also want to set `use_length=True`? See the API definition: https://mxnet.incubator.apache.org/api/python/docs/api/symbol/symbol.html#mxnet.symbol.softmax

fomkin:
@gigasquid Looks like it simple to downgrade Scala 2.11 back to 2.11.8. Upgrading Scala version is not the purpose of this PR.

zachgk:
> You may also want to set `use_length=True`? See the API definition: https://mxnet.incubator.apache.org/api/python/docs/api/symbol/symbol.html#mxnet.symbol.softmax

@TaoLv Is something wrong with the softmax operator? It says that the length argument is required, but I think it is supposed to be optional (and used to be optional). The examples in the documentation you linked to don't include a length argument. The current problem is that one of our java examples originally called `softmax(data)`, but now the signature is `softmax(data, length)` so we need to figure out what to pass for the length argument.

PR #17438, opened 2020-01-25
[submodule] Update tvm to the latest

yzhliu:
In order to bring the latest features,
* Reduce cuda compile time: https://github.com/apache/incubator-tvm/pull/4684
* Enable multi-arch cuda compile

@reminisce @haojin2

PR #17519, opened 2020-02-04
[tvmop] use size_var for placeholder & compute to reduce size of codegen

yzhliu:
It's a follow up of https://github.com/apache/incubator-mxnet/pull/17438


yzhliu:
Thanks @reminisce

Issue #16852, opened 2019-11-19
Could not build with CUDA_ARCH_LIST=Common


hubutui:
## Description
I could not build with `CUDA_ARCH_LIST=Common` on ArchLinux.
### Error Message
[mxnet-git.log](https://github.com/apache/incubator-mxnet/files/3863306/mxnet-git.log)

## To Reproduce
```shell
  cmake_opts=(
    -DBUILD_CPP_EXAMPLES:BOOL=OFF
    -DBUILD_TESTING:BOOL=OFF
    -DCMAKE_BUILD_TYPE:String=Release
    -DCMAKE_EXE_LINKER_FLAGS="$(pkg-config --libs cblas)"
    -DCMAKE_INSTALL_LIBDIR:PATH=lib
    -DCMAKE_INSTALL_PREFIX:PATH=/usr
    -DCMAKE_SHARED_LINKER_FLAGS="$(pkg-config --libs cblas)"
    -DCMAKE_SKIP_INSTALL_RPATH:BOOL=ON
    -DINSTALL_DOCUMENTATION:BOOL=OFF
    -DUSE_BLAS=open
    -DUSE_CPP_PACKAGE:BOOL=ON
    -DUSE_DIST_KVSTORE:BOOL=ON
    -DUSE_GPERFTOOLS:BOOL=OFF
    -DUSE_JEMALLOC:BOOL=OFF
    -DUSE_LAPACK:BOOL=ON
    -DUSE_MKL_IF_AVAILABLE:BOOL=OFF
    -DUSE_OPENMP:BOOL=ON
    -DUSE_S3:BOOL=ON
)
  cmake \
    ${cmake_opts[@]} \
    -DCUDA_ARCH_LIST=${cudaarch} \
    -DUSE_CUDA:BOOL=ON \
    -DUSE_CUDNN:BOOL=ON \
    -DUSE_MKLDNN:BOOL=ON \
    -DUSE_NCCL:BOOL=ON \
    -DUSE_OPENCV:BOOL=OFF \
    ..
  make
  cd ../python
  python setup.py build --with-cython
```

## Environment info
Same as https://github.com/apache/incubator-mxnet/issues/15263#issue-457222298

## What have you tried to solve it?

1. It's OK to build with `CUDA_ARCH_LIST=Pascal`
[...]

PR #17521, opened 2020-02-05
cmake: don't build PTX and 3.5 arch if cuda arch detection fails

leezu:
## Description ##
Currently, if `USE_CUDA=ON` but architecture detection fails due to building on a machine without GPU, we will build for `3.0;3.5;5.0;5.2;6.0;6.1;7.0;7.0+PTX;7.5;7.5+PTX`. Thereby we create too large object files, triggering https://github.com/apache/incubator-mxnet/issues/16852.

Thus, this PR introduces patched `select_compute_arch.cmake`, such that the fallback arches are `3.0;5.0;5.2;6.0;6.1;7.0;7.5`, ie. dropping PTX and 3.5 so that #16852 is not triggered.
We don't need 3.5, because code compiled for 3.0 also runs on 3.5 hardware and 3.5 hardware is 7 years old. If someone really cares about performance on 3.5, they can still manually set `MXNET_CUDA_ARCH=3.5`.

Upstream bug https://gitlab.kitware.com/cmake/cmake/issues/20275

### Changes ###
- [X] cmake: don't build PTX and 3.5 arch if cuda arch detection fails


ptrendx:
`We don't need 3.5, because code compiled for 3.0 also runs on 3.5 hardware` - I do not believe that is the case, those archs are not binary compatible. I would say that the better choice would be to remove SM3.0 (as SM3.5 is still accessible via p2 AWS instance). We do not need PTX for 7.0 if we include PTX for 7.5.

ptrendx:
Not including PTX at all would make it impossible to run MXNet on future GPUs without recompilation.

leezu:
> `We don't need 3.5, because code compiled for 3.0 also runs on 3.5 hardware` - I do not believe that is the case, those archs are not binary compatible. I would say that the better choice would be to remove SM3.0 (as SM3.5 is still accessible via p2 AWS instance). We do not need PTX for 7.0 if we include PTX for 7.5.

Why is this wrong? Binary forward compatibility is explicitly described in Nvidia docs https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#binary-compatibility

leezu:
> Not including PTX at all would make it impossible to run MXNet on future GPUs without recompilation.

That's true, but given the amount of cuda code in MXNet, PTX JIT compilation does not provide a good user experience (program will hang very long?) and is thus not enabled in the Makefile builds as well AFAIK.

leezu:
ping @ptrendx

leezu:
Ping @ptrendx

ptrendx:
Ok, I think you are right that 3.0 will in fact work for 3.5. That said, again, do we actually care about SM 3.0 (and 5.0 for that matter, as it was used only in a few GPUs)?

leezu:
> Ok, I think you are right that 3.0 will in fact work for 3.5. That said, again, do we actually care about SM 3.0 (and 5.0 for that matter, as it was used only in a few GPUs)?

I'm not sure if we have any data about the distribution of SMs of our users. Thus my main motivation here is to preserve backwards compatibility. 

Also, this change does not affect our binary distribution (python wheels), but just the default for manual build-from-source builds. So I expect if performance is important to whomever is doing the build, they will figure out to set the right architecture. Thus choosing a conservative default would be fine?

Is there any data about how large the performance difference is between running 3.0 binary compared to 3.5 binary on a 3.5 architecture?

ptrendx:
I see. There should not be much difference - the only difference in the generated instructions could be usage of `ldg` in SM_35 case, but I checked and it is used only in 2 places in MXNet.

leezu:
Thanks for checking. Given the small difference between SM 3 and 3.5, let's keep the default to 3.0?

Do you have any further suggestions for the PR? I still need to figure out why the license check is failing, but once that is done can I go ahead to merge?

leezu:
@ptrendx let's merge this PR as is? Without it users can run into #16852 easily.

PR #17525, opened 2020-02-05
MXNET-1444: replace NULL by nullptr literal in cpp-package

olk:
It is better to use the nullptr literal for null-pointers because NULL can be an integer literal with value zero (since C++11).


olk:
I did a forced push to my forked repo because of typos in the commit message - should I close and create a new pull request?

leezu:
No need for a new PR. My concern is that you added a file named `0001-replace-NULL-by-nullptr-literal-in-backend.patch `

olk:
Yes, I encountered my fault - I'll fix it ...

Issue #16376, opened 2019-10-05
 [RFC] Deferred compute in imperative interface to unify imperative and symbolic interface


leezu:
A new **deferred computation** (DC) argument to the imperative MXNet APIs is
proposed. If enabled, memory allocation and computation is deferred as long as
possible. Users can export the computational graph recorded during deferred
computation, which enables hybridization support.

Arrays for which DC is enabled are called **lazy**. Other arrays are called
**normal**. Inplace operations on lazy arrays are unsupported.

Storage allocation and computation for lazy arrays is deferred until their
results are required by conversion to numpy or use as input to an operator
creating a normal array. Accessing attributes such as `shape` can also trigger
computation if the attribute can't be inferred.

**Update: The proposed implementation in https://github.com/apache/incubator-mxnet/pull/17530/ differs slightly from the API previously described in this RFC. Thus I deleted the API docs in this RFC. Please refer to the PR.** For example, a global state is used to enable / disable deferred compute, instead of introducing a new invocation API `MXImperativeDeferredInvokeEx`.

### FAQ

**How about Autograd, `NDArray.autograd_entry_` and `AGInfo`?**
Autograd inside deferred computation (DC) mode can be supported.

Relation of Autograd and DC: While autograd’s `RecordOp` provides a similar
recording functionality to the deferred computation, the autograd graph is not
the same as a computational graph: `NDArray::Detach()` serves to detach a node
from the autograd graph by deleting `NDArray.entry_`, though the `NodeEntry` is
still required for reconstructing the computational history of how this NDArray
came to be.

**Are reqs like `kInPlace` supported?**
No. For now only `kWriteTo` is supported in DC mode.

The plan is to replace inplace operations with `kWriteTo` operations, writing to
a new (lazy) array. The framework should be smart enough to decide when to reuse
memory and when not. It shouldn’t be required for users to specify that they
want an inplace operation.

**How is context attribute handled, specifically context changes?**

Cross-device copy must be represented as operator (`CrossDeviceCopyOp`) which
requires special handling in the graph executor.

**How is incomplete shape information handled?**
`shape` property triggers computation if shape is accessed and can't be inferred completely.
Users can access `static_shape` if they wan't to avoid triggering computation.

## Python (Gluon)

Based on DC, hybridization in Gluon is simplified:

Instead of implementing `def hybrid_forward(self, F, x, ...)` in `HybridBlock`,
users can opt to implement `def forward(self, x, ...)` in `HybridBlock`.

[...]

PR #17530, opened 2020-02-05
Add deferred compute support

leezu:
## Description ##
Implements https://github.com/apache/incubator-mxnet/issues/16376

### Essentials ###

### Changes ###
- [X] Add imperative deferred compute for Gluon 2

## Comments ##

### DIFF
--- a/include/mxnet/imperative.h
+++ b/include/mxnet/imperative.h
@@ -224,9 +224,8 @@ class Imperative {
                              const std::vector<NDArray*>& inputs,
                              const std::vector<NDArray*>& outputs);
   /*! \brief obtain symbol representation of deferred compute session. */
-  nnvm::Symbol GetDeferredComputeSymbol(
-    const std::vector<std::pair<NDArray *, std::string>> &inputs,
-    const std::vector<NDArray *> &outputs);
+  nnvm::Symbol GetDeferredComputeSymbol(const std::vector<NDArray *> &outputs);
+  /*! \brief associate arrays with variables for deferred compute */
   void SetDeferredComputeVariable(NDArrayHandle *arrays, SymbolHandle *variables, const int num);
   /*! \brief */
   OpStatePtr Invoke(const Context& default_ctx,
--- a/include/mxnet/imperative.h
+++ b/include/mxnet/imperative.h
@@ -93,10 +93,6 @@ class Imperative {
   /*! \brief DCInfo datastructure to enable deferred computation */
   class DCInfo {
    public:
-    DCInfo() {
-      // Default constructor provided for the sake of any.h. Should not be used.
-      throw std::invalid_argument("Unsupported default constructor");
-    }
     explicit DCInfo(const std::vector<NDArray *> &inputs,
                     const std::vector<NDArray *> &outputs);
 
@@ -228,9 +224,9 @@ class Imperative {
                              const std::vector<NDArray*>& inputs,
                              const std::vector<NDArray*>& outputs);
   /*! \brief obtain symbol representation of deferred compute session. */
-  nnvm::Symbol *GetDeferredComputeSymbol(
-      const std::vector<std::pair<NDArray *, std::string>> &inputs,
-      const std::vector<NDArray *> &outputs);
+  nnvm::Symbol GetDeferredComputeSymbol(
+    const std::vector<std::pair<NDArray *, std::string>> &inputs,
+    const std::vector<NDArray *> &outputs);
   /*! \brief */
   OpStatePtr Invoke(const Context& default_ctx,
                     const nnvm::NodeAttrs& attrs,
--- a/include/mxnet/imperative.h
+++ b/include/mxnet/imperative.h
@@ -227,6 +227,7 @@ class Imperative {
   nnvm::Symbol GetDeferredComputeSymbol(
     const std::vector<std::pair<NDArray *, std::string>> &inputs,
     const std::vector<NDArray *> &outputs);
+  void SetDeferredComputeVariable(NDArrayHandle *arrays, SymbolHandle *variables, const int num);
   /*! \brief */
   OpStatePtr Invoke(const Context& default_ctx,
                     const nnvm::NodeAttrs& attrs,
--- a/include/mxnet/imperative.h
+++ b/include/mxnet/imperative.h
@@ -92,11 +92,11 @@ class Imperative {
 
   /*! \brief DCInfo datastructure to enable deferred computation */
   class DCInfo {
-  public:
-    explicit DCInfo() {
+   public:
+    DCInfo() {
       // Default constructor provided for the sake of any.h. Should not be used.
       throw std::invalid_argument("Unsupported default constructor");
-    };
+    }
     explicit DCInfo(const std::vector<NDArray *> &inputs,
                     const std::vector<NDArray *> &outputs);
 
@@ -121,7 +121,7 @@ class Imperative {
                           const std::vector<NDArray *> &inputs,
                           const std::vector<NDArray *> &outputs);
 
-  private:
+   private:
     friend class Imperative;
 
     /*! \brief Copies of input NDArrays
@@ -225,8 +225,8 @@ class Imperative {
                 std::vector<bool>* p_save_outputs = nullptr);
   /*! \brief to record operator, return corresponding node. */
   void RecordDeferredCompute(nnvm::NodeAttrs&& attrs,
-                             std::vector<NDArray*>& inputs,
-                             std::vector<NDArray*>& outputs);
+                             const std::vector<NDArray*>& inputs,
+                             const std::vector<NDArray*>& outputs);
   /*! \brief obtain symbol representation of deferred compute session. */
   nnvm::Symbol *GetDeferredComputeSymbol(
       const std::vector<std::pair<NDArray *, std::string>> &inputs,
--- a/include/mxnet/imperative.h
+++ b/include/mxnet/imperative.h
@@ -56,7 +56,8 @@ class Imperative {
     OpReqType grad_req;
     OpStatePtr state;
     std::vector<NDArray> outputs;
-    std::vector<NDArray> out_grads;
+    std::vector<NDArray> out_grads;  // used to hold gradient arrays the user is
+                                     // interested in (marked variables)
     bool fresh_out_grad;
 
     AGInfo() :
@@ -79,7 +80,7 @@ class Imperative {
     }
 
     static bool IsNone(const NDArray& arr) {
-      return arr.entry_.node == nullptr || arr.entry_.node->info.empty();
+      return arr.autograd_entry_.node == nullptr || arr.autograd_entry_.node->info.empty();
     }
 
     static bool IsVariable(const nnvm::ObjectPtr& node) {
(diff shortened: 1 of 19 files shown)


python/mxnet/_deferred_compute.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    prev = ctypes.c_int()
+    check_call(_LIB.MXNDArraySetDeferredComputeEnabled(
+        ctypes.c_int(is_deferred_compute), ctypes.byref(prev)))
+    return bool(prev.value)
+
+
+@contextlib.contextmanager
+def context():
+    # Like other MXNet context manager, this bleeds state across concurrent
```
eric-haibin-lin: Can you create a github issue to trace it, or mention potential workarounds?

src/imperative/imperative.cc
```diff
@@ -544,4 +681,60 @@ std::vector<NDArray*> Imperative::Backward(
...
+
+void Imperative::DCInfo::Compute(const NDArray &arr) {
+  if (Imperative::DCInfo::IsComputed(arr))
+    return;
+
+  DCInfo &info = Imperative::DCInfo::Get(arr.deferredcompute_entry_.node);
+  info.is_computed_ = true;  // We will Invoke at the end of this function.
+
+  // Recursively compute input arrays
+  for (NDArray &input : info.inputs_) {
```
reminisce: change to use `const NDArray&` ?

python/mxnet/_deferred_compute.py
```diff
@@ -0,0 +1,95 @@
...
+    # code: "Context managers that have state should use Context Variables
+    # instead of threading.local() to prevent their state from bleeding to
+    # other code unexpectedly, when used in concurrent code."
+    val = set_deferred_compute(True)
+    yield
+    set_deferred_compute(val)
+
+
+def get_symbol(input_arrays, output_arrays, input_names=None):
```
reminisce: Can we have a parameter taking in symbol class since we will need to support the symbols compatible with numpy ops.
```python
def get_symbol(input_arrays, output_arrays, input_names=None, sym_cls=Symbol):
```

src/imperative/imperative.cc
```diff
@@ -273,10 +288,132 @@ void Imperative::RecordOp(
...
+      << inputs.size() << " were specified.";
+
+  CHECK_EQ(missing_inputs.size(), 0)
+      << "ValueError: Invalid input to GetDeferredComputeSymbol. "
+      << missing_inputs.size() << " required inputs unspecified.";
+
+  // Deep copy of symbol as subsequent calls to this function may change the
+  // name of input variables.
+  s = s.Copy();
+  Symbol *out = new Symbol();
```
reminisce: Related to the above comment, it's better to move this to the caller and simply do `return s.Copy()` here to prevent potential memory leak when an exception is thrown in the following line.

src/c_api/c_api_ndarray.cc
```diff
@@ -428,3 +436,42 @@ int MXCachedOpRegisterOpHook(NDArrayHandle handle,
...
+  }
+  for (int i = 0; i < num_outputs; ++i) {
+    NDArray *array = reinterpret_cast<NDArray *>(output_handles[i]);
+    outputs.emplace_back(array);
+  }
+
+  // Obtain Symbeol
+  *out = Imperative::Get()->GetDeferredComputeSymbol(inputs, outputs);
+
+  API_END();
```
reminisce: It's better to use `API_END_HANDLE_ERROR` to handle potential memory leak caused by throwing exceptions. For example,
https://github.com/apache/incubator-mxnet/blob/93c123daee0241845c3e7f6452f91ab2d3d144c8/src/c_api/c_api.cc#L1068-L1080

include/mxnet/ndarray.h
```diff
@@ -83,7 +83,7 @@ class NDArray {
  public:
   /*! \brief default constructor */
   NDArray()
-    : entry_(nullptr) {
+    : autograd_entry_(nullptr) {
```
samskalicky: why dont we initialize `deferredcompute_entry_` in these constructors?

python/mxnet/_deferred_compute.py
samskalicky: does this have any issues with the work in #16654 ?

include/mxnet/imperative.h
```diff
@@ -88,6 +89,77 @@ class Imperative {
...
     }
   };
+
+  /*! \brief DCInfo datastructure to enable deferred computation */
+  class DCInfo {
+   public:
+    DCInfo() {
+      // Default constructor provided for the sake of any.h. Should not be used.
+      throw std::invalid_argument("Unsupported default constructor");
+    }
```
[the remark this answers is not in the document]
leezu: Good catch. This was only needed in an earlier, unpublished version of this PR.

include/mxnet/ndarray.h
leezu: It's not required? We just use the `NodeEntry` default constructor.

python/mxnet/_deferred_compute.py
leezu: This refers to problems if users use https://docs.python.org/3/library/asyncio.html

src/c_api/c_api_ndarray.cc
```diff
@@ -428,3 +436,41 @@ int MXCachedOpRegisterOpHook(NDArrayHandle handle,
   op->RegisterOpHook(clbk, monitor_all);
   API_END();
 }
+
+int MXNDArrayIsDeferredComputeEnabled(int *curr) {
```
samskalicky: why dont we call this just `IsDeferredCompute`?

src/c_api/c_api_ndarray.cc
```diff
@@ -428,3 +436,41 @@ int MXCachedOpRegisterOpHook(NDArrayHandle handle,
...
+  API_BEGIN();
+  *curr = Imperative::Get()->is_deferred_compute();
+  API_END();
+}
+
+int MXNDArraySetDeferredComputeEnabled(int deferred_compute, int *prev) {
```
samskalicky: why dont we call this `SetDeferredCompute` like the python API above `set_deferred_compute`

src/imperative/cached_op.h
```diff
@@ -294,7 +294,7 @@ void SetInputIndices(const nnvm::Graph& fwd_graph,
   const auto& indexed_graph = fwd_graph.indexed_graph();
   if (data_indices->ndim() || param_indices.ndim()) {
     CHECK_EQ(data_indices->ndim() + param_indices.ndim(),
-             indexed_graph.input_nodes().size());
+             static_cast<const int>(indexed_graph.input_nodes().size()));
```
samskalicky: Is this what fixed that weird error
```
[2020-02-05T20:57:05.668Z] /work/mxnet/3rdparty/mshadow/mshadow/././logging.h:75:34: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
[2020-02-05T20:57:05.668Z]  #define CHECK_EQ(x, y) CHECK((x) == (y))
[2020-02-05T20:57:05.668Z]                                   ^
[2020-02-05T20:57:05.668Z] /work/mxnet/3rdparty/mshadow/mshadow/././logging.h:68:9: note: in definition of macro 'CHECK'
[2020-02-05T20:57:05.668Z]    if (!(x))                                                \
[2020-02-05T20:57:05.668Z]          ^
[2020-02-05T20:57:05.668Z] src/imperative/./cached_op.h:290:5: note: in expansion of macro 'CHECK_EQ'
[2020-02-05T20:57:05.668Z]      CHECK_EQ(data_indices->ndim() + param_indices.ndim(),
[2020-02-05T20:57:05.668Z]      ^
[2020-02-05T20:57:05.668Z] cc1plus: all warnings being treated as errors
```

src/imperative/imperative.cc
```diff
@@ -273,8 +288,126 @@ void Imperative::RecordOp(
...
+        << "recording in deferred compute mode.";
+    // However, an inplace operation on a non-deferred compute array inside
+    // deferred compute scope will work. For example:
+    // a = mx.nd.arange(10)
+    // with dc.context():
+    //     a[:5] = 0
+  }
+  DispatchMode dispatch_mode = DispatchMode::kUndefined;
+  Context default_ctx = Context::CPU();
+  Context ctx = imperative::GetContext(attrs, inputs, outputs, default_ctx);
```
samskalicky: nit: do we need to declare default_ctx rather than just pass `Context::CPU()` into the `GetContext` function?

src/imperative/imperative.cc
```diff
@@ -273,8 +288,126 @@ void Imperative::RecordOp(
...
+        // to be updated.
+        Imperative::DCInfo &dcinfo = Imperative::DCInfo::Get(node);
+        const NDArray *array = dcinfo.input_handles_.at(i);
+
+        // Make sure this array is part of GetDeferredComputeSymbol inputs
+        auto is_equal = [array](const std::pair<NDArray *, std::string> &input) {
+          return array == std::get<0>(input);
+        };
+
+        // std::vector<std::pair<NDArray *, std::string>>::iterator input_search =
```
samskalicky: nice! thanks for clarifying the type!

src/ndarray/ndarray.cc
```diff
@@ -2088,6 +2117,25 @@ void NDArray::SyncCheckFormat(const bool full_check) const {
...
+  Engine::Get()->WaitForVar(ptr_->var);
+}
+
+void NDArray::WaitToWrite() const {
+  if (is_none()) return;
+  Imperative::DCInfo::Compute(*this);
+  /*!
+   * Push an empty mutable function to flush all preceding reads to the
+   * variable.
+   */
```
samskalicky: nit: does this need to be a multi-line comment?

src/c_api/c_api_ndarray.cc
leezu: For consistency with autograd APIs, lets call it `MXNDArraySetIsDeferredCompute`

src/c_api/c_api_ndarray.cc
leezu: `MX` prefix is required for symbol visibility. `NDArray` as this is part of the `NDArray` related functions.

src/imperative/imperative.cc
leezu: That one shouldn't have been a comment. I didn't figure out the correct type at the time of writing and used auto as intermediate solution.
Correct type is `std::vector<std::pair<NDArray *, std::string>>::const_iterator` (note the `const`) and is used now.

src/imperative/cached_op.h
leezu: Yes. Unfortunately there seems to be further undefined behavior and the CI needs more fixing (see failure status below.. Can't reproduce any of these failures locally..)

leezu: Was able to reproduce with Makefile based build and older gcc. With that setup, only 

``` c++
#include "./imperative_utils.h"
#include "./cached_op.h"
```

works in `imperative.cc`, but not

``` c++
#include "./cached_op.h"
#include "./imperative_utils.h"
```

which was used in the PR initially.

leezu:
@eric-haibin-lin enabling deferred compute mode enables recording and deferring the computation. 

It does not automatically export the recording to a CachedOp and use that for accelerated computation. This logic is implemented in `HybridBlock.__call__`. `HybridBlock.__call__`, which automatically enables deferred compute mode for the computation of the Block. Users should not enable deferred compute outside manually (in fact it's an internal API as of this PR).

We can add a `@hyridize` decorator in the future, which helps to hybridize functions based on deferred compute.

Specifically to your code example, there are a few more considerations:

**Cross device execution**
`copyto` is not supported by CachedOp yet. Thus while we can use `copyto` inside deferred compute mode, exporting to Symbol and running the Symbol will fail.
**Autograd**
Autograd recording and deferred compute recording are currently mutually exclusive. You can't enable both at the same time. But you don't need to: Deferred compute provides the symbolic representation of the forward computation. We don't need autograd while obtaining this symbolic representation, ie. we can just disable autograd inside the deferred compute scope. `CachedOp` takes the symbol and facilitates computing gradients.

It will be useful to support autograd inside deferred compute in the future, for higher order grads.

> what happens if a user enables DC for the program below? 

`MXNetError("Autograd recording is not supported during deferred compute mode.")`

https://github.com/apache/incubator-mxnet/blob/e64e03cb27b43f369c27c6629f1238103fa875e0/src/imperative/imperative.cc#L204-L214

https://github.com/apache/incubator-mxnet/blob/e64e03cb27b43f369c27c6629f1238103fa875e0/src/imperative/imperative.cc#L295-L299

src/c_api/c_api_ndarray.cc
```diff
@@ -107,9 +107,17 @@ void MXImperativeInvokeImpl(AtomicSymbolCreator creator,
...
-  if (Imperative::Get()->is_recording()) {
-    Imperative::Get()->RecordOp(std::move(attrs), ndinputs, ndoutputs, state);
+  if (Imperative::Get()->is_deferred_compute()) {
+    Imperative::Get()->RecordDeferredCompute(std::move(attrs), ndinputs, ndoutputs);
+  } else {
+    for (NDArray* input : ndinputs) {
+      Imperative::DCInfo::Compute(*input);
+    }
+    auto state = Imperative::Get()->Invoke(
+      Context::CPU(), attrs, ndinputs, ndoutputs);
```
zhreshold: is it assumes the context is always CPU?

src/imperative/imperative.cc
```diff
@@ -544,4 +705,60 @@ std::vector<NDArray*> Imperative::Backward(
...
+  for (const NDArray &input : info.inputs_) {
+    Compute(input);
+  }
+
+  // Prepare pointers
+  std::vector<NDArray *> ndinputs, ndoutputs;
+  ndinputs.reserve(info.inputs_.size());
+  ndoutputs.reserve(info.outputs_.size());
+  for (NDArray &input : info.inputs_)
+    ndinputs.push_back(&input);
```
zhreshold: can you point me to the machanism for maintaining pointers to ndarray valid through the lifetime of DCInfo?

src/c_api/c_api_ndarray.cc
leezu: This context is only used as `default_ctx`. (Also, this line is not changed in this PR.)

src/imperative/imperative.cc
leezu: `&input` here points to an array in `info.inputs_`, which is `std::vector<NDArray> inputs_`. Ie. it contains proper NDArrays and not only pointers to NDArrays that could become invalid.
We control the lifetime of the NDArrays in `std::vector<NDArray> inputs_`. They are copies, created based on the ndarray handle passed when invoking an operator and created in in `DCInfo::Create`. The copies are destroyed here in `DCInfo::Compute` after computation finished.

Note that the copies share the same underlying storage chunk. The storage chunk is deallocated once the frontend doesn't hold any handle to an NDArray owning that chunk anymore and once all copies located in DCInfo.inputs_ are destroyed.

zhreshold: thx

src/imperative/imperative.cc
```diff
@@ -273,7 +288,153 @@ void Imperative::RecordOp(
...
+                                       const std::vector<NDArray *> &inputs,
+                                       const std::vector<NDArray *> &outputs) {
+  CHECK(!is_recording())
+      << "Autograd recording is not supported during deferred compute mode.";
+
+  for (const NDArray *output : outputs) {
+    CHECK(DCInfo::IsNone(*output))
+        << "Inplace operations (+=, -=, x[:]=, etc) are not supported when "
+        << "recording in deferred compute mode.";
+    // However, an inplace operation on a non-deferred compute array inside
```
sxjscience: Does it mean that these two code snippets will be different?
- Without error:
```python
a = mx.np.arange(10)
with dc.context():
    a[:5] = 0
```
- With error:
```python
with dc.context():
    a = mx.np.arange(10)
    a[:5] = 0
```

leezu: Correct. In the first case we just record `_slice_assign_scalar` Op. The second case is unsupported because the `output` of `_slice_assign_scalar` is a ndarray that already has a `deferredcompute_entry_` attribute.

eric-haibin-lin:
LGTM

PR #17546, opened 2020-02-07
enabling build stage gpu_int64 to enable large tensor nightly runs

access2rohit:
## Description ##
Fixes nightly build failure due to absence of large tensor build artifact required for testing large tensor support on a nightly basis.

### Essentials ###

### DIFF
--- a/tests/nightly/JenkinsfileForBinaries
+++ b/tests/nightly/JenkinsfileForBinaries
@@ -40,25 +40,16 @@ core_logic: {
           utils.pack_lib('gpu', mx_lib_cpp_example_mkl)
         }
       }
-    }/*,
+    },
     'CPU: USE_INT64_TENSOR_SIZE': {
-      node(NODE_LINUX_CPU) {
+      node(NODE_LINUX_GPU) {
         ws('workspace/build-cpu-int64') {
           utils.init_git()
           utils.docker_run('ubuntu_nightly_cpu', 'build_ubuntu_cpu_large_tensor', false)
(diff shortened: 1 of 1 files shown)


access2rohit:

access2rohit:
@apeforest can you review this

tests/nightly/JenkinsfileForBinaries
```diff
@@ -49,7 +49,7 @@ core_logic: {
           utils.pack_lib('cpu_int64', mx_cmake_lib)
         }
       }
-    },
+    }*/,
```
apeforest: I think this is a little bit confusing. We are actually testing cpu context on GPU platforms here. The reason we don't use CPU node is simply the CPU node type does not have big enough memory as GPU node. 

We should either modify the CPU node type with a memory optimized one such as R5 (this is the ideal solution)

Or we rename the pipeline stage so that it's clear to people these are tests running on CPUs.

access2rohit: I can change “GPU: USE_INT64_TENSOR_SIZE” -> “USE_INT64_TENSOR_SIZE” to enable first.

Changing instance type is a temp fix. Theoritically everything should be able to run on CPU. So, fixing MXNet memory management would be the ideal solution but that would take lot of time. Switching to R5 will bump up our costs a bit too.

Let me know if "USE_INT64_TENSOR_SIZE" makes sense as interim solution to enable LT on nightly first.

access2rohit:
@apeforest updated the PR after incorporating your suggestions as discussed offline.

access2rohit:

Issue #7410, opened 2017-08-10
mxnet random seed should not be fixed by default


leezu:
mxnet seems to use a fixed seed for its random number generator. This is not good. To make sure research results are valid, experiments need to be repeated with different random initializations. To the best of my knowledge it is common practice that the seed for the random number generator is chosen by default randomly at startup. At least that is the behaviour of numpy.

As mxnet does not follow the expectated behaviour researchers may wrongly assume that their results are statistically significant, given that they'd expect a random seed.

Compare what happens when running the following two files repeatedly:
```
➜  ~ cat nprandom.py 
import numpy as np
print(np.random.normal(size=(2,2)))
➜  ~ cat mxrandom.py 
import mxnet as mx
print(mx.nd.random_normal(shape=(2,2)))
```
```
➜  ~ python3 nprandom.py
[[ 0.07005017 -0.10707515]
 [ 0.99219416  0.16670986]]
➜  ~ python3 nprandom.py
[[ 1.2185212  -1.56751472]
 [ 1.14549379 -1.90969064]]
```

```
➜  ~ python3 mxrandom.py

[[ 2.21220636  1.16307867]
 [ 0.7740038   0.48380461]]
<NDArray 2x2 @cpu(0)>
➜  ~ python3 mxrandom.py

[[ 2.21220636  1.16307867]
 [ 0.7740038   0.48380461]]
<NDArray 2x2 @cpu(0)>
```

I.e. mxnet random number generator uses the same seed..

leezu:
@mli @piiswrong maybe worth fixing before the new release. Or at least explicitly warning about this.

szha:
This issue is closed due to lack of activity in the last 90 days. Feel free to ping me to reopen if this is still an active issue. Thanks!
Also, do please check out our [forum](https://discuss.mxnet.io/) (and [Chinese version](https://discuss.gluon.ai/)) for general "how-to" questions.

larroy:
I think this is the appropriate behaviour. If you want different seeds, you should seed yourself using time or the systems random pool.

imgemp:
[...]

PR #17547, opened 2020-02-07
Fix cudnn Dropout reproducibility

roywei:
Fix https://github.com/apache/incubator-mxnet/issues/15662
This will replace #16532 with an alternative solution.
Please refer to the discussion in  https://github.com/apache/incubator-mxnet/pull/16532#discussion_r339268855

1. Added `GetSeed()` in GPU random resource, it's similar to CPU version.
2. During cudnn dropout, check whether random seed has been changed and reset cudnn dropout descriptor's seed if random seed changed.

Benefit:
1. This avoided generating a random int (tensor on gpu) on GPU, copy to CPU and and pass it to `cudnnDropoutGetStatesSize`.
2. get the seed from gpu random resource is fast, we can afford to do it every forward.

Drawback:
Will be affected by https://github.com/apache/incubator-mxnet/issues/7410, by default mxnet's random seed is fixed.

### DIFF
--- a/3rdparty/mshadow/mshadow/random.h
+++ b/3rdparty/mshadow/mshadow/random.h
@@ -55,7 +55,7 @@ class Random<cpu, DType> {
 #if MSHADOW_IN_CXX11
     rnd_engine_.seed(seed);
 #endif
-    this->rseed_ = static_cast<unsigned>(seed);
+    this->rseed_ = static_cast<uint64_t>(seed);
   }
   /*!
    * \brief get random seed used in random generator
@@ -274,7 +274,7 @@ class Random<cpu, DType> {
   /*! \brief use c++11 random engine. */
   std::mt19937 rnd_engine_;
   /*! \brief random number seed used in random engine */
-  unsigned rseed_;
+  uint64t rseed_;
 
 #else
 
@@ -404,7 +404,7 @@ class Random<gpu, DType> {
     // Now set the seed.
     curandStatus_t status;
     status = curandSetPseudoRandomGeneratorSeed(gen_, static_cast<uint64_t>(seed));
-    this->rseed_ = static_cast<unsigned>(seed);
+    this->rseed_ = static_cast<uint64_t>(seed);
     CHECK_EQ(status, CURAND_STATUS_SUCCESS) << "Set CURAND seed failed.";
   }
   /*!
--- a/3rdparty/mshadow/mshadow/random.h
+++ b/3rdparty/mshadow/mshadow/random.h
@@ -61,7 +61,7 @@ class Random<cpu, DType> {
    * \brief get random seed used in random generator
    * \return seed in unsigned
    */
-  inline unsigned GetSeed() const {
+  inline uint64_t GetSeed() const {
     return rseed_;
   }
   /*!
@@ -411,7 +411,7 @@ class Random<gpu, DType> {
     * \brief get random seed used in random generator
     * \return seed in unsigned
     */
-  inline unsigned GetSeed() const {
+  inline uint64_t GetSeed() const {
     return rseed_;
   }
   /*!
@@ -474,7 +474,7 @@ class Random<gpu, DType> {
   uniform(Shape<dim> shape);
 
  private:
-  unsigned rseed_;
+  uint64_t rseed_;
   inline void GenGaussian(float *dptr, size_t size, float mu, float sigma) {
     curandStatus_t status;
     status = curandGenerateNormal(gen_, dptr, size, mu, sigma);
--- a/3rdparty/mshadow/mshadow/random.h
+++ b/3rdparty/mshadow/mshadow/random.h
@@ -404,7 +404,15 @@ class Random<gpu, DType> {
     // Now set the seed.
     curandStatus_t status;
     status = curandSetPseudoRandomGeneratorSeed(gen_, static_cast<uint64_t>(seed));
+    this->rseed_ = static_cast<unsigned>(seed);
     CHECK_EQ(status, CURAND_STATUS_SUCCESS) << "Set CURAND seed failed.";
+  }
+  /*!
+    * \brief get random seed used in random generator
+    * \return seed in unsigned
+    */
+  inline unsigned GetSeed() const {
+    return rseed_;
   }
   /*!
    * \brief get a set of random integers
@@ -466,6 +474,7 @@ class Random<gpu, DType> {
   uniform(Shape<dim> shape);
 
  private:
+  unsigned rseed_;
   inline void GenGaussian(float *dptr, size_t size, float mu, float sigma) {
     curandStatus_t status;
     status = curandGenerateNormal(gen_, dptr, size, mu, sigma);
--- a/3rdparty/mshadow/mshadow/random.h
+++ b/3rdparty/mshadow/mshadow/random.h
@@ -274,7 +274,7 @@ class Random<cpu, DType> {
   /*! \brief use c++11 random engine. */
   std::mt19937 rnd_engine_;
   /*! \brief random number seed used in random engine */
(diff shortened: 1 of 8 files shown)


roywei:
@DickJC123 @ptrendx could you help take a look again? Thanks

3rdparty/mshadow/mshadow/random.h
```diff
@@ -404,8 +404,16 @@ class Random<gpu, DType> {
...
     curandStatus_t status;
     status = curandSetPseudoRandomGeneratorSeed(gen_, static_cast<uint64_t>(seed));
+    this->rseed_ = static_cast<unsigned>(seed);
     CHECK_EQ(status, CURAND_STATUS_SUCCESS) << "Set CURAND seed failed.";
   }
+  /*!
+    * \brief get random seed used in random generator
+    * \return seed in unsigned
+    */
+  inline unsigned GetSeed() const {
```
apeforest: Why can't we just use uint64_t as the data type for rseed_?

3rdparty/mshadow/mshadow/random.h
```diff
@@ -466,6 +474,7 @@ class Random<gpu, DType> {
   uniform(Shape<dim> shape);
 
  private:
+  unsigned rseed_;
```
apeforest: use uint64_t?

src/operator/nn/dropout-inl.h
```diff
@@ -255,8 +255,13 @@ class DropoutOp {
       Stream<xpu> *s = ctx.get_stream<xpu>();
 
       // set dropout state.
-      ctx.requested[0].get_cudnn_dropout_desc(&dropout_desc_, s, 1.0f - this->pkeep_, seed_);
-
+      Random<xpu, unsigned> *prnd = ctx.requested[1].get_random<xpu, unsigned>(s);
+      uint64_t rng_seed = (uint64_t) prnd->GetSeed();
```
apeforest: If GetSeed() returns uint64_t you don't need the cast here

src/operator/rnn.cc
```diff
@@ -180,6 +180,7 @@ static std::vector<ResourceRequest> RNNResourceEx(const NodeAttrs& attrs, const
     const RNNParam& param = nnvm::get<RNNParam>(attrs.parsed);
     if (param.p != 0 && 1.0f - param.p > 0) {
       request.emplace_back(ResourceRequest::kCuDNNDropoutDesc);
+        request.emplace_back(ResourceRequest::kRandom);
```
apeforest: nit: indent here not consistent.

karan6181:
@DickJC123 @ptrendx Can you please review this PR once again? Thanks!

apeforest:
@roywei Could you please rebase and update this PR? Thanks!

ChaiBapchya:
@roywei with CI issue fixed. let's rebase to include the fix. Thanks!

access2rohit:
restarted CI jobs. @roywei can you get your PR merged once they pass CI

ChaiBapchya:
@access2rohit without rebasing, the fix #17962 won't be included [for windows-gpu]

roywei:
@ChaiBapchya @access2rohit  Rebased! Thanks!

src/operator/rnn-inl.h
```diff
@@ -1495,7 +1500,7 @@ class RNNOp {
   cudnnRNNInputMode_t input_mode_;
   cudnnDropoutDescriptor_t dropout_desc_;
   Storage::Handle reserve_space_;
-  uint64_t seed_ = 17 + rand() % 4096;  // NOLINT(runtime/threadsafe_fn)
```
leezu, line 1498: This `seed_` is used for `rand_r` in the forward implementations. That's very bad practice, not portable and should be fixed.
As this PR already passes CI, it's fine to fix it in a follow-up PR. In fac, https://github.com/apache/incubator-mxnet/pull/17984 is blocked by the usage of `rand_r` so I may fix it there.

More info: https://channel9.msdn.com/Events/GoingNative/2013/rand-Considered-Harmful

src/operator/nn/dropout-inl.h
```diff
@@ -255,8 +255,13 @@ class DropoutOp {
...
       // set dropout state.
-      ctx.requested[0].get_cudnn_dropout_desc(&dropout_desc_, s, 1.0f - this->pkeep_, seed_);
-
+      Random<xpu, unsigned> *prnd = ctx.requested[1].get_random<xpu, unsigned>(s);
+      uint64_t rng_seed = prnd->GetSeed();
+      // reset dropout descriptor if rng seed changed.
+      bool reset = seed_ != rng_seed;
+      seed_ = rng_seed;
+      ctx.requested[0].get_cudnn_dropout_desc(&dropout_desc_, s, 1.0f - this->pkeep_,
+          seed_, reset);
```
szha, line 264: @roywei there are more than 1 random number generators in the resources (4 by default) and they could have different seeds. The result of rotating random number generator seeds here is that cudnn dropout state is reinitialized very often. @sxjscience observed that there's significant performance impact because of this.

rondogency: I believe @roywei is already working on this issue, since this regression has been caught by our benchmark last week

roywei: I spent some time looking into it and actually the problem is dropout's forward is entering the [re-init cudnn dropout desc logic](https://github.com/apache/incubator-mxnet/pull/17547/files#diff-cc7bb408eba92cdd6fad0e590b76fabeL438) every time during forward even without my change. This is true for nd/np and gluon, false for symbol dropout. So originally it was already reinitializing in every forward. The reason it does not cause any performance regression is because it's always using the `seed_` defined in `uint64_t seed_ = 17 + rand() % 4096; ` and never changed, it won't listen to MXNet's PRNG (which is the problem this PR is trying to fix). So if the seed didn't change, event if you re-init cudnn dropout descriptor, it won't take any time. My PR changed the seed, so it was actually re-init every forward, thus the regression.

However, it works fine under symbol case. I guess is somehow for symbolic, every forward is using the same Dropout node and the state check took effect the didn't go into re-init logic. For imperative case the state check is always empty and went into the re-init logic.

Compare the following two code, one is gluon and one is symbol, if I print some log during reinitialization in the original code without my change [here](https://github.com/apache/incubator-mxnet/pull/17547/files#diff-cc7bb408eba92cdd6fad0e590b76fabeL438). It will print out every forward in ND/NP/Gluon, not the case in Symbol.
```
import mxnet as mx
data = mx.nd.ones((10, 200, 300, 500), ctx=mx.gpu(0))
dropout = mx.gluon.nn.Dropout(0.5)
# with or without hybridize is the same result
dropout.hybridize()
with mx.autograd.record():
        result1 = dropout(data)
        result2 = dropout(result1)
```
print 2 times
```
re-init dropout desc
re-init dropout desc
```

Symbol:
```
import mxnet as mx
data = mx.nd.ones((10, 200, 300, 500), ctx=mx.gpu(0))
net = mx.sym.Variable("data")
net = mx.sym.Dropout(data=net, p=0.5, cudnn_off=False)
exe = net.simple_bind(mx.gpu(0), data=data.shape)
result1 = exe.forward(is_train=True, data=data)
result2 = exe.forward(is_train=True, data=result1[0])
```

print 1 time
```
re-init dropout desc
```

Given this situation, I don't have a good solution as checking the state handle size does not work in imperative mode, checking PRNG seed will also not work. I will revert this PR for now as it's causing regressions for models using dropout.

cc @szha @sxjscience @apeforest

szha: @roywei the links in your comment don't seem to point to what they refer to so I'm a bit confused. Where did you insert the 're-init dropout desc' comment?

PR #17570, opened 2020-02-11
Ignore config.cmake in git and remove old files

apeforest:
## Description ##
Remove old files to make the cmake instruction consistent with our man page.
Add config.cmake to gitignore for better developer experience.

@leezu Please review.


larroy:
What's the rationale for removing dev menu? The main goal is to reproduce CI results, build documentation and reproduce failures. I understand that the installation instructions were updated and that's fine. But I think the two things are unrelated.

apeforest:
@larroy I don't see it being used in our current CI any more. Where do we use it? 
This caused confusion to new users. We have seen in the slack channel some users were trying to modify cmake_options.yml and build mxnet.

apeforest:
Update PR per offline discussion with @larroy

larroy:
Thanks, could you update dev_menu.py to use the new way to configure CMake ?  right now it will be broken.

leezu:
Essentially if think for that you need to delete

``` python
    def read_config(self):
        assert os.path.isfile(self.cmake_options_yaml)
        with open(self.cmake_options_yaml, 'r') as f:
            self.cmake_options = yaml.safe_load(f)

    def _cmdlineflags(self):
        res = []
        for opt,v in self.cmake_options.items():
            res.append('-D{}={}'.format(opt,v))
        return res

def cmake_command(self) -> str:
        """
        :return: Cmake command to run given the options
        """

```

and replace `cmake_command` call with something like `['cmake', '-C', 'config.cmake']`?

apeforest:
Updated

larroy:
Did you try to run the tool? Looks like it's broken now.

larroy:
Where is the config.cmake? or where is documented how to use it?, I looked at the new build from source instructions are we are still calling plain cmake.

leezu:
Calling plain cmake is intended and correct. You just need to pass the `-C path/to/config.cmake` option.
You can find the default files in https://github.com/apache/incubator-mxnet/tree/master/config
#17602 renamed the config.cmake to linux.cmake etc

PR #17592, opened 2020-02-14
Add cmake build support for macOS static build.

frankfliu:
## Description ##
Add cmake build support for macOS static build.
Added --without-libidn2 flag for libcurl build.

### Essentials ###

## Comments ##
- CMake build still fail with libtiff 4.0.10 version. macOS build only work with 4.0.9, will create separate PR.

### DIFF
--- a/tools/dependencies/curl.sh
+++ b/tools/dependencies/curl.sh
@@ -35,6 +35,7 @@ if [[ ! -f $DEPS_PATH/lib/libcurl.a ]]; then
         CONFIG_FLAG="--with-darwinssl"
     fi
     ./configure $CONFIG_FLAG \
+                --without-libidn2 \
                 --with-zlib \
                 --with-nghttps2 \
                 --without-zsh-functions-dir \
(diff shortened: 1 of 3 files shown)


tools/dependencies/curl.sh
```diff
@@ -35,6 +35,7 @@ if [[ ! -f $DEPS_PATH/lib/libcurl.a ]]; then
         CONFIG_FLAG="--with-darwinssl"
     fi
     ./configure $CONFIG_FLAG \
+                --without-libidn2 \
```
leezu: > No, there's no known security reason to avoid enabling libidn2 in curl builds.
For generic curl builds I would recommend building with it so that users can
use international domain names in URLs

https://curl.haxx.se/mail/lib-2017-10/0158.html

Why disable this?

leezu:
As per https://github.com/leezu/mxnet/commit/0bbd44c45f34c37bf70c90635e29a5fcf2e6f0cc/checks/447232316/logs libtiff 4.1 does not work either.

leezu:
Thank you @frankfliu ! Should now be fixed by https://github.com/apache/incubator-mxnet/pull/17602 ?

PR #17602, opened 2020-02-15
Fix OS X staticbuild, update docs and add tests

leezu:
## Description ##
Fix OS X staticbuild and add tests based on Github Actions.

### DIFF
--- a/docs/static_site/src/pages/get_started/osx_setup.md
+++ b/docs/static_site/src/pages/get_started/osx_setup.md
@@ -24,176 +24,151 @@ permalink: /get_started/osx_setup
 
 # Installing MXNet from source on OS X (Mac)
 
-**NOTE:** For pre-built MXNet with Python, please refer to the [new install guide]({{'/get_started'|relative_url}}).
+The following installation instructions are for building MXNet from source. For
+instructions to build MXNet from source on other platforms, see the general
+[Build From Source guide](build_from_source).
 
-Installing MXNet is a two-step process:
+Instead of building from source, you can install a binary version of MXNet. For
+that, please follow the information at [Get Started](get_started).
+
+Building MXNet from source is a two-step process:
 
 1. Build the shared library from the MXNet C++ source code.
-2. Install the supported language-specific packages for MXNet.
+2. (optional) Install the supported language-specific packages for MXNet.
 
-**Note:** To change the compilation options for your build, edit the ```make/config.mk``` file and submit a build request with the ```make``` command.
+If you plan to build with GPU, you need to set up the environment for CUDA and
+cuDNN. Please follow the [NVIDIA CUDA Installation Guide for Mac OS
+X](https://docs.nvidia.com/cuda/cuda-installation-guide-mac-os-x/index.html) and
+[cuDNN Installation
+Guide](https://docs.nvidia.com/deeplearning/sdk/cudnn-install/index.html#install-mac).
+Note that CUDA stopped supporting macOS in 2019 and future versions of CUDA may
+not support macOS.
 
-## Prepare Environment for GPU Installation
+## Contents
 
-This section is optional. Skip to next section if you don't plan to use GPUs. If you plan to build with GPU, you need to set up the environment for CUDA and cuDNN.
+* [Build the MXNet shared library from source](#build-mxnet-from-source)
+* [Install Language Packages](#installing-language-packages-for-mxnet)
+    * [R](#install-the-mxnet-package-for-r)
+    * [Julia](#install-the-mxnet-package-for-julia)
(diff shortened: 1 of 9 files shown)


ChaiBapchya:
Verified this works by doing the following
Mac OS X 10.14.6
1. Get the source code
```
git clone --recursive https://github.com/leezu/mxnet.git
cd mxnet/
git fetch origin
git checkout --track origin/osxci
```

2. Install dependencies
- Homebrew
```
gcc --version
ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
brew update
```
- Rest
```
brew install nasm automake ninja libtool
```

3. Gave me error for cmake command not found
```
brew install cmake
```

4. Static Build with CPU
```
CMAKE_STATICBUILD=1 ./tools/staticbuild/build.sh cpu
```

It works perfectly. Takes a little over an hour to build.
Result - Static linked MXNet library available as a dylib file : `mxnet/lib/libmxnet.dylib`

Minor modification : Update Readme to reflect libmxnet.dylib (if its is Mac) here : https://github.com/apache/incubator-mxnet/blob/master/tools/staticbuild/README.md

ChaiBapchya:
Verified normal build for this PR on Mac OSX 10.14.6 : followed the instructions for Source build (using CMake) as mentioned in this PR

leezu:
Thanks @ChaiBapchya. Let's update the staticbuild/Readme to reflect `libmxnet.dylib` when fixing `staticbuild/build_wheel.sh` (given CI has passed for this PR and that the difference between `.so` and `.dylib` was already present before this PR).

docs/static_site/src/pages/get_started/osx_setup.md
```diff
@@ -24,176 +24,151 @@ permalink: /get_started/osx_setup
...
-	pip install jupyter
+git clone --recursive https://github.com/apache/incubator-mxnet.git mxnet
+cd mxnet
+cp config/darwin.cmake config.cmake
 ```
 
-### Build MXNet Shared Library
+Please edit the config.cmake file based on your needs. The file contains a
+series of `set(name value CACHE TYPE "Description")` entries. For example, to
+build without Cuda, change `set(USE_CUDA ON CACHE TYPE "Build with CUDA
```
apeforest: nit: Cuda --> CUDA

docs/static_site/src/pages/get_started/osx_setup.md
```diff
@@ -24,176 +24,151 @@ permalink: /get_started/osx_setup
...
+without GPU, MXNet build can't autodetect your GPU architecture and will target
+all available GPU architectures. Please set the `MXNET_CUDA_ARCH` variable in
+`config.cmake` to your desired cuda architecture to speed up the build.
 
-After you have installed the dependencies, pull the MXNet source code from Git and build MXNet to produce an MXNet library called ```libmxnet.so```. You can clone the repository as described in the following code block, or you may try the [download links](download) for your desired MXNet version.
+To (optionally) build with MKL math library, please install MKL first based on
+the guide in [Math Library Selection](build_from_source#math-library-selection).
 
-The file called ```osx.mk``` has the configuration required for building MXNet on OS X. First copy ```make/osx.mk``` into ```config.mk```, which is used by the ```make``` command:
+**Step 3:** Build MXNet core shared library.
```
apeforest: These instructions and everything below seem to be identical to linux setup instruction. Could we consolidate them to reduce duplication?

docs/static_site/src/pages/get_started/osx_setup.md
```diff
@@ -24,176 +24,151 @@ permalink: /get_started/osx_setup
...
+# Install Homebrew
+/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
 
-Now we can start to build MXNet.
+# Install dependencies
+brew install cmake ninja ccache opencv
+```
 
-## Build the Shared Library
+`opencv` is an optional dependency. You can delete it from above `brew install`
```
apeforest: It might be clearer to explain what's the consequence if `opencv` is not installed so users can make an informed decision.

docs/static_site/src/pages/get_started/osx_setup.md
leezu: Agree. This will require more work though. Essentially the consolidated page will require client-side logic that allows users to select the operating system of their interest and based on the selection hides the unrelated instructions. Another option is file transclusion, but markdown may not support this.

Thus, the majority of the displayed consolidated page would be shared, and only the dependency installation would be OS specific.

Given this is not the focus of this PR, let's deduplicate in a separate PR once Windows instructions have been simplified as well.

PR #17617, opened 2020-02-17
[Large Tensor] Fixed Spatial Transformer op

connorgoggins:
## Description ##
The Spatial Transformer op was previously breaking on large tensor (dimension >= 2^32) data. With the following input:
```
run_performance_test(nd.SpatialTransformer, run_backward=True, inputs=[{'data': (2, 2**29,1,6), 'loc': nd.random_normal(shape=(2,6)), 'transform_type': 'affine', 'sampler_type': 'bilinear', 'target_shape': (2,6)}], warmup=1, runs=1)
```
the following error was thrown:
```
*** Error in `python3': double free or corruption (out): 0x00007f36bbffe010 ***
```

To root cause this issue, I ran the previous command in a Python script with GDB, and found that the underlying problem was in the iteration portion of the forward and backward methods of `spatial_transformer.cc`. Several of the variables used in the iteration used the `int` dtype when they should have been using `index_t` to properly handle long int indices. I switched these variables to `index_t` in the forward and backward methods, and after rebuilding, the previous input command displayed the correct output:
```
INFO:root:Begin Benchmark - SpatialTransformer
INFO:root:Complete Benchmark - SpatialTransformer
[{'SpatialTransformer': [{'inputs': {'data': (2, 536870912, 1, 6), 'loc': '<NDArray 2x6 @cpu(0)>', 'transform_type': 'affine', 'sampler_type': 'bilinear', 't
arget_shape': (2, 6)}, 'max_storage_mem_alloc_cpu/0': 102005472.0, 'avg_time_forward_SpatialTransformer': 551614.125, 'avg_time_backward_SpatialTransformer':
 511034.0938}]}]
```

### Essentials ###
- [x] Code is well-documented

### Changes ###
- M src/operator/tensor/spatial_transformer.cc

## Comments ##
Tested on r5dn.24xl-ubuntu 16.04 and p2.16xl-ubuntu 16.04 with
1. Individual op run
2. Full OpPerf run

## Results ##
The key difference between CPU and GPU tests was the instance type (r5dn.24xl for CPU, p2.16xl for GPU). All relevant build flags remain the same, and both were tested using CPU context.

[Single operator test - SpatialTransformer op (GPU)](https://gist.github.com/connorgoggins/7b3765e4e6f54f7841fb14e0930221ca)
[Single operator test - SpatialTransformer op (CPU)](https://gist.github.com/connorgoggins/d97d64e73702a342bdeb170d30aef04f)

[Full OpPerf test (GPU)](https://gist.github.com/connorgoggins/8b0563eaf98119980a5d36b7c61d796e)
[Full OpPerf test (CPU)](https://gist.github.com/connorgoggins/4bebae70a85f7d9f754124187f52f383)

### DIFF
--- a/src/operator/spatial_transformer.cc
+++ b/src/operator/spatial_transformer.cc
@@ -58,10 +58,10 @@ inline void BilinearSamplingForward(const Tensor<cpu, 4, DType> &output,
           const DType top_left_x_w = 1.0 - (x_real - top_left_x);
           const index_t data_index = n * i_c * i_h * i_w + c * i_h * i_w +
                                  top_left_y * i_w + top_left_x;
-          index_t top_left_v = 0;
-          index_t top_right_v = 0;
-          index_t bottom_left_v = 0;
-          index_t bottom_right_v = 0;
+          DType top_left_v = 0;
+          DType top_right_v = 0;
+          DType bottom_left_v = 0;
+          DType bottom_right_v = 0;
           index_t lower_bound = 0;
           if (between(top_left_x, lower_bound, i_w-1) &&
               between(top_left_y, lower_bound, i_h-1))
@@ -114,10 +114,10 @@ inline void BilinearSamplingBackward(const Tensor<cpu, 4, DType> &input_grad,
             const index_t data_index = n * i_c * i_h * i_w + c * i_h * i_w +
                                    top_left_y * i_w + top_left_x;
             // calc 4 vertex value in input data
-            index_t top_left_v = 0;
-            index_t top_right_v = 0;
-            index_t bottom_left_v = 0;
-            index_t bottom_right_v = 0;
+            DType top_left_v = 0;
+            DType top_right_v = 0;
+            DType bottom_left_v = 0;
+            DType bottom_right_v = 0;
             index_t lower_bound = 0;
             if (between(top_left_x, lower_bound, i_w-1) &&
                 between(top_left_y, lower_bound, i_h-1)) {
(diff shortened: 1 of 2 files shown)


connorgoggins:

access2rohit:
@connorgoggins can you tell me what are the possible types for DType here:
https://github.com/apache/incubator-mxnet/pull/17617/files#diff-64979618e95b2211ff9317b0830fbbd2R37

connorgoggins:
@access2rohit DType here represents the type of the interior elements of the input tensors, and since SpatialTransformer only supports floating point types then DType could be `float32` or `float64` (also potentially `float16`, but this DType doesn't seem to have been implemented for GEMM on CPU yet).

src/operator/spatial_transformer.cc
```diff
@@ -58,10 +58,10 @@ inline void BilinearSamplingForward(const Tensor<cpu, 4, DType> &output,
           const DType top_left_x_w = 1.0 - (x_real - top_left_x);
           const index_t data_index = n * i_c * i_h * i_w + c * i_h * i_w +
                                  top_left_y * i_w + top_left_x;
-          index_t top_left_v = 0;
-          index_t top_right_v = 0;
-          index_t bottom_left_v = 0;
-          index_t bottom_right_v = 0;
+          DType top_left_v = 0;
```
ChaiBapchya: why is this changed back to Dtype?
It's the index position right?

connorgoggins: @ChaiBapchya changing DType to index_t for these specific variables makes the op give incorrect on standard inputs (e.g. the inputs in the CI run) - the values generated in the output NDArray are all integers instead of floats. This is due to the fact that these variables do not represent the index positions (as I also originally believed), but the underlying values at the vertices.

ChaiBapchya: Cool. good catch.
_v indicates the value at that particular index. Thanks!

PR #17585, opened 2020-02-13
Dynamic subgraph property doc

samskalicky:
## Description ##
Add a brief tutorial doc on Custom partitioning tutorial and example for the feature implemented in #15969. This PR addresses two of the items in #17532
- Adds an example using Gluon to partition a model
- Adds fixes to custom operator support for the backward pass of subgraphOps
- changes int* and size to vector<int> for `ids` in `acceptSubgraph` API

## This PR is not
- Updating docs on the MXNet website on extending MXNet
- Consolidating custom extension related docs on the website

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at https://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- Renamed `backend_args` to `backend_opts` for the `hybridize` API to disambiguate from model args (ie. params) since it contains options that are passed to the backend.
- Moved GPU customOp tests and split tests into diff files for CPU and GPU
  - CPU tests are in **unittests/test_extensions.py** for test_custom_op (GeMM) and test_subgraph (exp, log)
  - GPU tests are in **gpu/test_extensions_gpu.py** for test_custom_op_gpu (ReLU)

Issue #17236, opened 2020-01-07
[RFC] Custom subgraph property part2


samskalicky:
## Description
Request for comments on the next PR for enhancing custom subgraph property support

Heres some suggestions from the initial PR (Part 1):

Heres some suggestions from the initial PR (Part 1):
- docs, readme, tutorial
- dynamic graph passes

## References
- initial PR (Part 1): #17034

samskalicky:
@mxnet-label-bot add RFC

samskalicky:
Current implementation (as of 3/18/2020) only supports combining ops into subgraphs generically. does not allow for fine grained control to say which ops should go into which subgraphs. Need to modify `supportedOps` API to change the `ids` vector from bool (supported or not) to integer (to say which subgraph this node should go in to). This will allow for coloring and fine grained control of ops in a subgraph. 
https://github.com/apache/incubator-mxnet/blob/07d0b738384cba011ce519bcc41aeeaaafcd22c1/example/extensions/lib_subgraph/subgraph_lib.cc#L179-L181
Will also need to modify custom_subgraph_property.h selector to use these int values to define the subgraph instead of just looking for existence of the operator name.
https://github.com/apache/incubator-mxnet/blob/41d534be3ffd7b45846f555366230f4e811d8751/src/operator/subgraph/partitioner/custom_subgraph_property.h#L47-L54

Open question:
How do we handle the cases where we want to remove args/aux from the top level of the model, migrating those into the subgraph? If we compile a subgraph with Args/Aux then those tensor values are effectively frozen in the compiled binary so we need to be clear with users that they can no longer be changed in the top level model

samskalicky:
Need to support removing args/aux from custom graph pass, re-using the optimized symbol when calling optimize_for in Gluon repeatedly

PR #17623, opened 2020-02-19
Dynamic subgraph compile support

samskalicky:
## Description ##
This PR adds support for passing the NDArrays from the existing `optimize_for` API down to the `reviewSubgraph` function in an external library. It also adds a new API for HybridBlock called `optimize_for` that can partition the model without running a forward pass. 
### Feature changes
- Adds new API to HybridBlock `optimize_for` that partitions the model but does not call the cachedOp
- Modifies the subgraph library example to optionally require args to be provided
- Adds annotation on subgraph inputs for the name of the original param so that inputs can be mapped and passes annotations to input nodes of subgraphs
- Adds support for tensors in MKLDNN format, calls `Reorder2Default`
### New tests
- Adds a new test to partition operators that directly consume params
- add a new model to test where ops to be partitioned have args/params
### Bug Fixes
- fixes bug in passing `ids` vector by value instead of by reference
- fixes bug in passing copies of attributes instead of by reference
- fixes bug where `_cached_graph` was not updated after partitioning
- fixes memory leak where user-specified attributes on subgraph ops were not freed if subgraph was rejected
- fixes problem incorrectly indexing into shape/dtype maps when annotating the graph
### Docs
- Updates the README doc with the latest changes described above

## Design - Passing NDArrays ##
In #15886 the `optimize_for` API was added  to the Symbol class to give users an easy API to use to partition their models. The `args` argument took the params to the model to use for shape/type inference. But the NDArray values were never used. In this PR, we pass the NDArray data values to the backend library and also add the ability to pass auxiliary params too. Now, the `optimize_for` API looks like:
```
sym = sym.optimize_for('default', args, aux, ctx=mx.cpu())
```

On the backend library side, the `reviewSubgraph` API has an additional arguments for `args` and `aux` that are maps of named MXTensors. 
```
MXReturnValue reviewSubgraph(std::string json, int subraph_id, bool* accept,
                             std::unordered_map<std::string, std::string>& options,
                             std::unordered_map<std::string, std::string>& attrs) {
                             std::unordered_map<std::string, std::string>& attrs,
                             std::map<std::string, MXTensor>& args,
                             std::map<std::string, MXTensor>& aux);
```

These additional maps of args and aux will allow backends to compile subgraphs and use param/weight values during compilation. For example, this will enable TensorRT to be implemented as a backend library and eliminate the `init_tensorrt_params` API that is needed to provide the params to the TensorRT backend. It will also enable compiling subgraphs with TVM and other compile-centric backends. 

This PR also tags subgraph inputs with a new attribute "argName" if the input to the subgraph comes directly from a model param/weight. This will be used when backend libraries lookup params for subgraph inputs since the names of subgraph inputs are modified from the original model param name. The "argName" can be looked up in the new `args` or `aux` arguments passed to the `reviewSubgraph` API to get the actual tensor values. Heres an example partitioned graph, notice that the original input param is named `a` and the input to the subgraph is named `a0`. You can see the "argName" attribute is set on the input refer to original name `a`.

```
{
  "nodes": [
    {
      "op": "null", 
      "name": "a", 
      "attrs": {
        "dtype": "0", 
        "shape": "[3,2]"
      }, 
      "inputs": []
    }, 
    {
      "op": "_custom_subgraph_op", 
      "name": "_op0", 
      "attrs": {"myKey": "myVal"}, 
      "inputs": [[0, 0, 0]], 
      "subgraphs": [
        {
          "nodes": [
            {
              "op": "null", 
              "name": "a0", 
              "attrs": {
                "argName": "a", 
                "isArg": "True", 
                "isAux": "False"
              }, 
              "inputs": []
            }, 

            ...

        }
      ]
    }
  ], 
}
```

## Design - Partitioning HybridBlock without infer ##
In #15969 the `hybridize` API was modified to accept the backend name and partition the model during the hybridize flow. This flow is for users who intend to run inference immediately after partitioning. But for users that want to partition but not run inference, they are out of luck. When compiling the model, a user will compile on a machine suitable for compilation. And that compilation may take tens of minutes depending on the model and optimization strategy. Then they will export their model, and copy it to the machine they intent to run inference on. 

In this flow, the machine for running compilation may not be suitable for running a complete forward pass. In this PR we add an `optimize_for` API to the HybridBlock class that runs most of what is part of the hybridize flow including the forward pass. But it does not actually call the cachedOp. 

The new HybridBlock `optimize_for` API combines the argument lists of `forward` and `hybridize` like:

```
def optimize_for(self, x, *args, backend=None, backend_opts=None, **kwargs)
```

Notice that the first two args are the same as in the `forward` API and the last 3 args are the same as in the `hybridize` API. The `active` argument is dropped since it will default to true in this function in order to partition the model. 

The expectation is that users will do the following:

```
block.optimize_for(x, backend='myProp')
block.export('partitioned')
```

This is equivalent to:

```
block.hybridize(backend='myProp')
block(x)
block.export('partitioned')
```

But does not actually execute the forward pass. Users can still use the partitioned block to run forward passes just like with hybridize, for example:

```
block.optimize_for(x, backend='myProp')
block(x)
```

Calling `optimize_for` is equivalent to `hybridize` in this usage, and since the cachedOp is already created before `block(x)`, it is not recreated during the forward pass. 

### Essentials ###


samskalicky:
If #17585 gets merged first, then ill update the doc with the new compile info. If this one gets merged first ill got modify #17585 to add the description.

samskalicky:

src/c_api/c_api.cc
```diff
@@ -823,9 +823,9 @@ int MXLoadLib(const char *path) {
 
       // MXNET_REGISTER_SUBGRAPH_PROPERTY(customBackend, CustomSubgraphProperty);
       mxnet::op::SubgraphBackendRegistry::Get()->__REGISTER_CUSTOM_PROPERTY__(name_str,
-                            std::make_shared<mxnet::op::CustomSubgraphProperty>(
-                           strategy_str, callSupportedOps, supportedOps_fp,
-                           callAcceptSubgraph, acceptSubgraph_fp, callFree, op_name_str));
+                               std::make_shared<mxnet::op::CustomSubgraphProperty>(
```
mseth10: looks like some unintentional spacing changes were made. can you please fix them if you end up pushing another commit?

src/operator/subgraph/build_subgraph.cc
```diff
@@ -561,10 +561,12 @@ void CutGraphInputs(const std::vector<nnvm::NodeEntry*> &input_entries,
     nnvm::ObjectPtr n = nnvm::CreateVariableNode(
         var_name + std::to_string(name_count_map[var_name]));
     // set attribute for subgraph input to indicate if it is from an arg/param to model
-    if (e->node->is_variable())
+    if (e->node->is_variable()) {
       n->attrs.dict["isArg"] = "True";
-    else
+      n->attrs.dict["argName"] = var_name;
```
mseth10: where is "argName" getting used?

src/c_api/c_api.cc
samskalicky: sure, will do

src/operator/subgraph/build_subgraph.cc
samskalicky: added an example to the PR description

samskalicky:
What do you think @HahTK?

HahTK:
It sounds like the goal for passing data is to allow data to be compiled into the bin (example tensort bin) for that subgraph to avoid an init step.

If the weights are in the bin, then we structuring this such that weights can not be changed with calling optimize_for again using new weights ?

Is there any reason why we only do this for args and not auxs ?

samskalicky:
> It sounds like the goal for passing data is to allow data to be compiled into the bin (example tensort bin) for that subgraph to avoid an init step.
> 
> If the weights are in the bin, then we structuring this such that weights can not be changed with calling optimize_for again using new weights ?

Using the weights in `optimize_for` is a one-way flow. You cannot call `optimize_for` again with new weights. You would need to call it on the original graph with new weights. The assumption is that weights would only be used for inference. So presumably the model is already trained, and the weights are frozen. 

> Is there any reason why we only do this for args and not auxs ?

Thanks for pointing this out. I'll add them too

samskalicky:
@HahTK ive added the ability to pass params down for aux args, and your suggestion to enable removing the params to the enhancements issue: #17236

HahTK:
@samskalicky :
I'll let the reviewers go through the code in detail but from perusal and with the aux added the feature looks good.

Things to think about in the future :
1. For backends with very long compile time, a weight_update function may be useful.
It also serves as a path to dynamically swap models (if so desired).

2. Some backends may absorb the weights in to the bin itself.
For those we should provide an option for the partitioner to remove the redundant weight nodes that have been fully absorbed into bins. 
The corner case would of course be weights that are only partly absorbed.
This would reduce memory footprint and improve speed by avoid the need to shove around unused weights

I think both of these items can be done in the future but just enumerating them here

samskalicky:
@sergeykolychev I had to update the optimize_for API call in the perl bindings. Can you please review/approve for the change here:
https://github.com/apache/incubator-mxnet/pull/17623/files#diff-cc128e9c950abe76c19162ce01fe2402

example/extensions/lib_subgraph/subgraph_lib.cc
```diff
@@ -218,10 +218,27 @@ MXReturnValue mySupportedOps(std::string json,
...
 MXReturnValue myReviewSubgraph(std::string json, int subraph_id, bool* accept,
                                std::unordered_map<std::string, std::string>& options,
-                               std::unordered_map<std::string, std::string>& attrs) {
+                               std::unordered_map<std::string, std::string>& attrs,
+                               std::map<std::string, MXTensor>& args,
+                               std::map<std::string, MXTensor>& aux) {
   for (auto kv : options) {
     std::cout << "option: " << kv.first << " ==> " << kv.second << std::endl;
   }
+  for (auto kv : args) {
```
HahTK: are we missing auxs here ?

example/extensions/lib_subgraph/subgraph_lib.cc
```diff
@@ -218,10 +218,27 @@ MXReturnValue mySupportedOps(std::string json,
...
+    std::cout << "arg: " << kv.first << " ==> (";
+    for (auto s : kv.second.shape)
+      std::cout << s << ",";
+    std::cout << ") [";
+    for (int i=0; i<kv.second.size(); i++)
+      std::cout << kv.second.data<float>()[i] << ", ";
+    std::cout << "]" << std::endl;
+  }
+  if(options.count("reqArgs") > 0 && args.size() == 0) {
```
HahTK: ditto missing auxs again ?
we should probably have a reqAuxs option too

example/extensions/lib_subgraph/test_subgraph.py
```diff
@@ -88,3 +91,54 @@
...
+sym_block2.initialize()
+sym_block2.optimize_for(mx.nd.ones((3,2)), mx.nd.ones((3,2)), backend='myProp')
+sym_block2.export('partitioned')
+
+
+###############################################
+# Test with subgraph directly consuming params
+###############################################
+# example model, ops to be partitioned have args
+d2 = mx.sym.exp(a)
```
HahTK: I don't understand the goal of this test.
The point of optimize_for() is to export the partitioned graph without running inference.
Therefore, in the real world use case, the user will only run inference on the imported partition.

The test should re-import the partitioned graph and test that right ?
Testing the sym directly tells us nothing about the success of the export/import.

NOTE : There may still be value to testing the sym directly but that is secondary to the intended use model

python/mxnet/gluon/block.py
```diff
@@ -1026,6 +1030,53 @@ def _call_cached_op(self, *args):
             out = [out]
         return _regroup(out, self._out_format)
 
+    def optimize_for(self, x, *args, backend=None, backend_opts=None, **kwargs):
```
HahTK: is this also missing auxs ?

src/operator/subgraph/build_subgraph.cc
HahTK: sanity check .... are we attaching "auxName" to auxes ?

example/extensions/lib_subgraph/README.md
```diff
@@ -120,11 +122,8 @@ There are several essential building blocks for making a custom partitioner:
...
 
             REGISTER_PARTITIONER(my_part_name)
-            .addStrategy("strategy1", 
-                          supportedOps, 
-                          "_custom_subgraph_op")
-            .setReviewSubgraph("strategy1", 
-                                reviewSubgraph);
+            .addStrategy("strategy1", supportedOps, "_custom_subgraph_op")
+            .setReviewSubgraph("strategy1", reviewSubgraph);
```
rondogency: don't forget the add gluon optimize_for you just wrote to this doc

example/extensions/lib_subgraph/subgraph_lib.cc
```diff
@@ -205,7 +205,7 @@ MXReturnValue mySupportedOps(std::string json,
     }
 
     //check if op dtype is float
-    if(dtype == kFloat32) {
+    if((dtype == kFloat32 && options.count("reqFloat") > 0) || options.count("reqFloat") == 0) {
```
rondogency: what is "reqFloat"?

perl-package/AI-MXNetCAPI/mxnet.i
```diff
@@ -1633,6 +1633,8 @@ int MXOptimizeForBackend(SymbolHandle sym_handle,
                                    const mx_uint in,
                                    NDArrayHandle* in,
                                    const mx_uint in,
+                                   NDArrayHandle* in,
```
rondogency: should we have the same names for this function params?

example/extensions/lib_subgraph/subgraph_lib.cc
```diff
@@ -177,7 +177,7 @@ REGISTER_OP(_custom_subgraph_op)
 const std::vector<std::string> op_names({"exp","log"});
 
 MXReturnValue mySupportedOps(std::string json,
-                             std::vector<bool> ids,
+                             std::vector<bool>& ids,
```
rondogency: can you elaborate this ids in either the code or the doc? like "this vector flags all the ops in the given symbol graph to be partitioned into subgraphs"

example/extensions/lib_subgraph/subgraph_lib.cc
samskalicky: no, this is just an example implementation of the subgraph library. it can do whatever the user wants it to do.

example/extensions/lib_subgraph/subgraph_lib.cc
samskalicky: this is just an example implementation of the subgraph library. it can do whatever the user wants it to do. 

Theres no reason to require aux by default, depending on the model it may or may not have any aux params. The smarter thing a user could do is check and see if the subgraph has aux inputs, and then in the aux map is empty error out.

example/extensions/lib_subgraph/test_subgraph.py
samskalicky: this is just giving us a baseline to compare again built-in MXNet computation flow. and again its just an example.

python/mxnet/gluon/block.py
samskalicky: nope. in Gluon args here means inputs to the model. not the same as args/aux in symbol/module flow

src/operator/subgraph/build_subgraph.cc
samskalicky: no. argName will be set for all "variables" which includes both args and aux. for an aux param, isArg = True && isAux = true && argName = 'bob'

example/extensions/lib_subgraph/subgraph_lib.cc
samskalicky: changed the comment to clarify:
```
//check if op dtype is float, and if option was specified to require float types 
```

perl-package/AI-MXNetCAPI/mxnet.i
samskalicky: I was just following the style that was already there. @sergeykolychev to verify

example/extensions/lib_subgraph/subgraph_lib.cc
samskalicky: In the doc it says
> The 2nd argument is an array of booleans, one for each operator in the model. When traversing the graph, operators to be partitioned into subgraphs are identified and an entry is set to `true` for the index in the `ids` array corresponding to the node ID.

Is that sufficient?

example/extensions/lib_subgraph/README.md
samskalicky: thanks for the reminder, done

samskalicky:
Thanks for the feedback @HahTK and @rondogency! I made some changes and replied to your comments/suggestions

example/extensions/lib_subgraph/subgraph_lib.cc
HahTK: Agreed. I did not notice that this was an example

example/extensions/lib_subgraph/subgraph_lib.cc
HahTK: Agreed

example/extensions/lib_subgraph/test_subgraph.py
HahTK: agreed

src/operator/subgraph/build_subgraph.cc
samskalicky: Heres an example from resnet's batchNorm operator:
```
{
      "op": "null",
      "name": "resnetv10_batchnorm0_beta0",
      "attrs": {
        "argName": "resnetv10_batchnorm0_beta",
        "isArg": "True",
        "isAux": "False"
      },
      "inputs": []
    },
    {
      "op": "null",
      "name": "resnetv10_batchnorm0_running_mean0",
      "attrs": {
        "argName": "resnetv10_batchnorm0_running_mean",
        "isArg": "True",
        "isAux": "True"
      },
      "inputs": []
    },
```

samskalicky: Heres an example of an input to a subgraph thats an output from another operator (ie. not a param):
```
{
      "op": "null", 
      "name": "_plus0_output0", 
      "attrs": {
        "isArg": "False", 
        "isAux": "False"
      }, 
      "inputs": []
    }, 
```

python/mxnet/gluon/block.py
```diff
@@ -1026,6 +1030,53 @@ def _call_cached_op(self, *args):
...
+        """Activates or deactivates :py:class:`HybridBlock` s recursively. Has no effect on
+        non-hybrid children.
```
leezu: Should this docstring be extended to clarify the difference with `hybridize`?

python/mxnet/gluon/block.py
```diff
@@ -1026,6 +1030,53 @@ def _call_cached_op(self, *args):
...
+        static_alloc : bool, default False
+            Statically allocate memory to improve speed. Memory usage may increase.
+        static_shape : bool, default False
+            Optimize for invariant input shapes between iterations. Must also
+            set static_alloc to True. Change of input shapes is still allowed
+            but slower.
+        """
+
+        # do hybrize API call
+        self.hybridize(True, backend, backend_opts, **kwargs)
```
leezu: Could you elaborate on your choice of `optimize_for` calling `hybridize`, instead of requiring users to call `hybridize` and then `optimize_for`?

samskalicky: The meaning of `optimize_for` API in Symbol class is to partition the model. In Gluon to partition we have to hybridize to build the symbol, and then partition it. So it made sense to integrate that inside as a single step for the user. 

GluonCV does a similar thing with its `export_block` API where they call hybridize under the hood instead to provide a better user experience with a single API call. 
https://github.com/dmlc/gluon-cv/blob/9921323263d3698ce3e03a9eb4ada208b7b8bce3/gluoncv/utils/export_helper.py#L129-L133

python/mxnet/gluon/block.py
samskalicky: Ya i think i just copied from hybridize's doc string. I'll update this so its more clear what the function does and refer to hybridize. i'll add something like we put in the README.md:
> Immediately partitions a HybridBlock using the specified backend. Combines the work done in the `hybridize` API with part of the work done in the forward pass without calling the CachedOp. Can be used in place of `hybridize`, afterwards `export` can be called or inference can be run. 

What do you think?

leezu: I think that's reasonable as a separate paragraph in the docstring. But it'd be too confusing as a sentence summary. Maybe for the first sentence, you could describe what optimize_for is used for. Ie. enable users to optimize for a backend without doing an actual forward pass. The concept of CachedOp may not be clear to users.

samskalicky: just pushed the change, lmk if that addresses your concern

python/mxnet/gluon/block.py
rondogency: what if user calls hybridize and then call optimize_for? nothing will happen or it throws on unrelated error message?

python/mxnet/symbol/symbol.py
```diff
@@ -1475,13 +1482,19 @@ def optimize_for(self, backend, args=None, ctx=None, **kwargs):
...
             args = []
             args_handle = c_array(NDArrayHandle, [])
         else:
-            listed_arguments = self.list_arguments()
-            args_handle, args = self._get_ndarray_inputs('args', args, listed_arguments, False)
+            args_handle, args = self._get_ndarray_inputs('args', args,
+                                                         self.list_arguments(), False)
 
+        if aux is None or len(aux) == 0:
+            aux = []
```
rondogency: do we need this line?

python/mxnet/gluon/block.py
samskalicky: calling `hybridize` and then `optimize_for` is fine. Since `optimize_for` will just call `hybridize` again internally anyway.

Calling `optimize_for` and then `hybridize` though will reset the cachedOp since `hybridize` calls `_clear_cached_op`. So not a problem per-se. User just needs to realize what they are doing.

include/mxnet/lib_api.h
```diff
@@ -234,10 +234,15 @@ enum MXReturnValue {
  */
 struct MXTensor {
   MXTensor() : data_ptr(nullptr), dtype(kUNSET), verID(0) {}
-
+  MXTensor(const MXTensor& oth) : data_ptr(oth.data_ptr), shape(oth.shape),
+    dtype(oth.dtype), verID(oth.verID), ctx(oth.ctx) {
+    setDLTensor();
```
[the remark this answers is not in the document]
samskalicky: In this case, its the copy constructor. So we're constructing an MXTensor from another one, and we need to setup the DLTensor for the new one that we're constructing. This is used in std::map

python/mxnet/gluon/block.py
```diff
@@ -1026,6 +1030,69 @@ def _call_cached_op(self, *args):
...
+            raise ValueError('In HybridBlock, there must be one NDArray as input.'
+                             ' Please check the type of the args.\n')
+        if len(ctx_set) > 1:
+            raise ValueError('Find multiple contexts in the input, '
+                             'After hybridized, the HybridBlock only supports one input '
+                             'context. You can print the ele.ctx in the '
+                             'input arguments to inspect their contexts. '
+                             'Find all contexts = {}'.format(ctx_set))
+
+        self._build_cache(x, *args)
```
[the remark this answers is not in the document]
samskalicky: `_build_cache` combines (x, args) back into a single tuple:
https://github.com/apache/incubator-mxnet/blob/ade7d48fc7b977e28213ec24b628c61dc1c0c6f0/python/mxnet/gluon/block.py#L923

python/mxnet/symbol/symbol.py
samskalicky: Yes, because later on line 1513 we do `mx_uint(len(aux))`

include/mxnet/lib_api.h
```diff
@@ -407,6 +412,8 @@ class OpResource {
  */
 /*! \brief Macro to help passing serialized subgraph through attribute dict */
 #define SUBGRAPH_SYM_JSON "subgraph_sym_json"
+#define MX_DTYPE "__dtype__"
+#define MX_SHAPE "__shape__"
```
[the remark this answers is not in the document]
samskalicky: changed to:
```
#define MX_STR_SUBGRAPH_SYM_JSON "subgraph_sym_json"
#define MX_STR_DTYPE "__dtype__"
#define MX_STR_SHAPE "__shape__"
```

tests/python/unittest/test_extensions.py
```diff
@@ -167,3 +167,17 @@ def test_subgraph():
     out4 = sym_block(mx.nd.ones((3,2)),mx.nd.ones((3,2)))
     # check that result matches one executed by MXNet
     assert_almost_equal(out[0].asnumpy(), out4[0].asnumpy(), rtol=1e-3, atol=1e-3)
+
+    # Gluon Hybridize partitioning with shapes/types
+    sym_block2 = nn.SymbolBlock(sym, [a,b])
+    sym_block2.initialize()
+    a_data = mx.nd.ones((3,2))
+    b_data = mx.nd.ones((3,2))
```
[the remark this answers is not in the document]
samskalicky: I dont think it matters here since we're not testing any mathematical code. Just partitioning, we're not adding any operators.

src/c_api/c_api_symbolic.cc
```diff
@@ -1383,7 +1403,26 @@ int MXOptimizeForBackend(SymbolHandle sym_handle,
...
+    g.attrs["in_args"] = std::make_shared<nnvm::any>(in_args_ptr);
+    g.attrs["in_arg_names"] = std::make_shared<nnvm::any>(arg_names);
+
+    NDArray **in_aux_ptr = static_cast<NDArray**>(nullptr);
+    std::vector<std::string> aux_names;
+    g.attrs["in_aux"] = std::make_shared<nnvm::any>(in_aux_ptr);
+    g.attrs["in_aux_names"] = std::make_shared<nnvm::any>(aux_names);
   }
+
+
```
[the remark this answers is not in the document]
samskalicky: it needed two lines because the code above and below were fighting and couldnt stand eachother, so it needs some extra room ;-)

Removed it in the latest commit

include/mxnet/lib_api.h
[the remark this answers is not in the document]
rondogency: then if it's copying we don't need the setTensor call below. should we just make setTensor call here?

samskalicky:

samskalicky:
@leezu i think this PR is ready to go. But lets leave it open until next week, I was hoping to get a pass from the Nvidia folks for applicability for TensorRT usage. @ptrendx and @Caenorst

mseth10:
@ptrendx @Caenorst Can you guys please review this PR for TensorRT usage?

src/c_api/c_api.cc
```diff
@@ -122,14 +122,28 @@ void CustomFComputeDispatcher(const std::string op_name,
...
-    in_dims.push_back(inputs[i].shape().ndim());
-    in_types.push_back(inputs[i].dtype());
-    in_verIDs.push_back(inputs[i].version());
-    const char* ctx_str = inputs[i].ctx().dev_mask() == Context::kCPU ? "cpu" : "gpu";
-    in_dev_type.push_back(ctx_str);
-    in_dev_id.push_back(inputs[i].ctx().real_dev_id());
+    const NDArray& in_nd = inputs[i];
+    // reorder data if in MKLDNN format
+    if (in_nd.IsMKLDNNData()) {
+      const NDArray& tmp_nd = in_nd.Reorder2Default();
```
mseth10: Can we just keep `in_nd.Reorder2Default()` inside if block, the rest looks same for if and else block.

samskalicky: `Reorder2Default` returns a copy:
https://github.com/apache/incubator-mxnet/blob/4dddb087293965af7e7af8d0eaead60a9051f0fc/src/ndarray/ndarray.cc#L597

samskalicky: i pushed a change to use pointers instead so that the code isnt duplicated, let me know how that looks to you

src/operator/subgraph/partitioner/custom_subgraph_property.h
```diff
@@ -308,6 +350,63 @@ class  CustomSubgraphProperty: public SubgraphProperty {
       return nullptr;
     }
   }
+
+  virtual void InitSubgraphInputs(std::vector<nnvm::NodeEntry*>* input_entries,
+                                  std::vector<nnvm::NodeEntry>* orig_input_entries) const {
+    std::cout << "in InitSubgraphInputs" << std::endl;
```
mseth10: remove this?

samskalicky: durrr....thanks. just pushed the change

src/operator/subgraph/partitioner/custom_subgraph_property.h
```diff
@@ -188,31 +278,133 @@ class  CustomSubgraphProperty: public SubgraphProperty {
...
+        e->node->attrs.dict["__dtype__"] = ss.str();
+      }
+
+      if (orig.node->attrs.dict.count("__shape__") > 0) {
+        // get shape string from other node
+        std::string& shape = orig.node->attrs.dict["__shape__"];
+        int idx = 0;
+        // find the beginning of the output shape for the particular output index
+        for (unsigned x=0; x < orig.index; x++)
+          idx = shape.find("[", idx+1);
```
mseth10: check if `idx` points to the right index

src/operator/subgraph/partitioner/custom_subgraph_property.h
```diff
@@ -188,31 +278,133 @@ class  CustomSubgraphProperty: public SubgraphProperty {
...
+      }
+
+      // pass down other attributes if available
+      if (orig.node->attrs.dict.count("__dtype__") > 0) {
+        // get dtype string from other node
+        std::string& dtype = orig.node->attrs.dict["__dtype__"];
+        int idx = 0;
+        // find the beginning of the output dtype for the particular output index
+        for (unsigned x=0; x < orig.index; x++)
+          idx = dtype.find("[", idx+1);
```
mseth10: modify this for list of integers

src/operator/subgraph/partitioner/custom_subgraph_property.h
```diff
@@ -188,31 +278,133 @@ class  CustomSubgraphProperty: public SubgraphProperty {
...
-        call_free_(attr_keys[i]);
+
+      // set shapes
+      {
+        std::stringstream ss;
+        ss << "[";
+        for (unsigned i=0; i < sym.outputs.size(); i++) {
+          nnvm::Node* n = sym.outputs[i].node.get();
+          if (n->attrs.dict.count("__shape__") > 0) {
+            std::string& shape = n->attrs.dict["__shape__"];
```
mseth10: modify logic for the case when `n` is a subgraph node and `shape` a list of lists

samskalicky: fixed

src/operator/subgraph/partitioner/custom_subgraph_property.h
samskalicky: fixed

src/operator/subgraph/partitioner/custom_subgraph_property.h
samskalicky: fixed

samskalicky:
@mseth10 thanks! lmk what you think of the latest commit

mseth10:
> @mseth10 thanks! lmk what you think of the latest commit

All latest changes look good to me. Thanks @samskalicky for addressing multiple subgraph issues in this PR.

leezu:
@ptrendx any concerns on TensorRT or is this good to merge?

src/operator/subgraph/partitioner/custom_subgraph_property.h
```diff
@@ -188,31 +278,121 @@ class  CustomSubgraphProperty: public SubgraphProperty {
...
+          if (e.node->attrs.dict.count("__shape__") > 0) {
+            std::string& shape = e.node->attrs.dict["__shape__"];
+            // add this shape to the list
+            ss << getShapeAt(shape, e.index);
+          }
+          if (i < sym.outputs.size()-1)
+            ss << ",";
+        }
+        ss << "]";
+        n->attrs.dict["__shape__"] = ss.str();
```
samskalicky: nit: use MX_STR_SHAPE instead of "__shape__"

src/operator/subgraph/partitioner/custom_subgraph_property.h
```diff
@@ -188,31 +278,121 @@ class  CustomSubgraphProperty: public SubgraphProperty {
...
+          if (e.node->attrs.dict.count("__dtype__") > 0) {
+            std::string& dtype = e.node->attrs.dict["__dtype__"];
+            // add this dtype to the list
+            ss << getDtypeAt(dtype, e.index);
+          }
+          if (i < sym.outputs.size()-1)
+            ss << ",";
+        }
+        ss << "]";
+        n->attrs.dict["__dtype__"] = ss.str();
```
samskalicky: nit: use MX_STR_DTYPE instead of "__dtype__"

samskalicky:
@leezu, Reviewed offline with @ptrendx and @Caenorst. Decided that there are a few missing features needed for TensorRT to use this compile API. But decided to list them here https://github.com/apache/incubator-mxnet/issues/17236 and implement them in a subsequent PR. We can go ahead and merge this one as it is still useful for compiling for TVM. 

After this PR is merged i'll open another PR and start working on the missing features next.

PR #17629, opened 2020-02-19
[Urgent] Reduce load on CI due to excessive log flood

larroy:
## Description ##
as title


szhengac:
After merging this PR, I found that my PR #17400 fails to build on Windows, though I have retriggered CI several times.

PR #17632, opened 2020-02-20
[Large Tensor] Fixed RNN op

connorgoggins:
## Description ##
The RNN op was previously breaking on large tensor (dimension >= 2^32) data. With the following input:
```
run_performance_test(nd.RNN, run_backward=True, inputs=[{'data': (2**28,4,4), 'parameters': nd.random_normal(shape=(7,)), 'state': nd.random_normal(shape=(1, 4, 1)), 'mode': 'rnn_relu', 'state_size': 1, 'num_layers': 1}], warmup=1, runs=1)
```
the following error was thrown:
```
MXNetError: Check failed: dim_size >= -1 (-2147483640 vs. -1) : shape dim size must be >= -1, while received -2147483640
```

To root cause this issue, I ran the previous command in a Python script with GDB, and found that the underlying problem was in several of the function definitions of `rnn-inl.h`. Several of the data variables (`input_size`, `batch_size`, and `seq_length`) used the `int` dtype when they should have been using `index_t` to properly handle long int dimensions. I switched these variables to `index_t` in the relevant function headers and, after rebuilding, the previous input command displayed the correct output:
```
INFO:root:Begin Benchmark - RNN
INFO:root:Complete Benchmark - RNN
[{'RNN': [{'inputs': {'data': (268435456, 4, 4), 'parameters': '<NDArray 7 @cpu(0)>', 'state': '<NDArray 1x4x1 @cpu(0)>', 'mode': 'rnn_relu', 'state_size': 1, 'num_layers': 1}, 'max_storage_mem_alloc_cpu/0': 27917288.0, 'avg_time_forward_RNN': 1244353.25, 'avg_time_backward_RNN': 1345001.375}]}]
```

However, this only accounted for running the RNN op in `relu` mode. The op currently supports 3 other modes: `rnn_tanh`, `lstm`, and `gru`. To ensure that the op worked with large tensor data in each of these three modes in addition to `relu` mode, I made extensive modifications to `rnn_impl.h`. My modifications involved changing the data type of many function parameters and local variables for forward and backward functions for these three modes.

To ensure completeness and to prevent future breaking changes, I also added a nightly test for the RNN op with large tensor data in `tests/nightly/test_large_array.py`.

### Essentials ###
- [x] Code is well-documented

### Changes ###
- M src/operator/rnn-inl.h
- M src/operator/rnn_impl.h
- M tests/nightly/test_large_array.py

## Comments ##
Tested on r5dn.24xl-ubuntu 16.04 and p2.16xl-ubuntu 16.04 with
1. Individual op run (for RNN op under each of the four modes)
2. Full OpPerf run

## Results ##
The key difference between CPU and GPU tests was the instance type (r5dn.24xl for CPU, p2.16xl for GPU). All relevant build flags remain the same, and both were tested using CPU context.

[Single operator test - RNN ReLU op (GPU)](https://gist.github.com/connorgoggins/1a9c64239783257f7516cdaa931a96b4)
[Single operator test - RNN ReLU op (CPU)](https://gist.github.com/connorgoggins/734d42f5a5ff2553adab69344b451220)
[Single operator test - RNN tanh op (CPU)](https://gist.github.com/connorgoggins/1d44ba81432f7b5e51c02a365247be6a)
[Single operator test - RNN LSTM op (CPU)](https://gist.github.com/connorgoggins/b72ca4b65152b9f602745ea1bcb431d8)
[Single operator test - RNN GRU op (CPU)](https://gist.github.com/connorgoggins/3eae0b40cfa2a23efdf349bb29a7f84e)

[Full OpPerf test (CPU)](https://gist.github.com/connorgoggins/e26b842c08a5bf58af7ce074e5465821)

### DIFF
--- a/src/operator/rnn-inl.h
+++ b/src/operator/rnn-inl.h
@@ -122,7 +122,7 @@ struct RNNParam : public dmlc::Parameter<RNNParam> {
   }
 };
 
-inline int GetRnnParamSize(int num_layer,
+inline index_t GetRnnParamSize(int num_layer,
                            index_t input_size,
                            int state_size,
                            int direction,
--- a/src/operator/rnn-inl.h
+++ b/src/operator/rnn-inl.h
@@ -63,7 +63,7 @@ struct RNNParam : public dmlc::Parameter<RNNParam> {
   bool bidirectional, state_outputs;
   int mode;
   float p;
-  int seq_length_, batch_size_, input_size_;
+  index_t seq_length_, batch_size_, input_size_;
 
   bool use_sequence_length;
   dmlc::optional<int> projection_size;
@@ -123,7 +123,7 @@ struct RNNParam : public dmlc::Parameter<RNNParam> {
 };
 
 inline int GetRnnParamSize(int num_layer,
-                           int input_size,
+                           index_t input_size,
                            int state_size,
                            int direction,
                            int mode,
@@ -182,8 +182,8 @@ inline int GetRnnBiasSize(int num_layer,
  *  - output -> h[t](, c[t] additionally with Lstm) time by time(sz: NxH(x2))
  *  - intermediate y[1...T] as next layer's inputs(sz: TxNxHxD)
  */
-inline size_t GetRNNWorkspaceSize(int seq_length,
-                                  int batch_size,
+inline size_t GetRNNWorkspaceSize(index_t seq_length,
+                                  index_t batch_size,
                                   int hidden_size,
                                   int projection_size,
                                   int direction,
@@ -214,8 +214,8 @@ inline size_t GetRNNWorkspaceSize(int seq_length,
 
 inline size_t GetRNNReserveSpaceSize(int num_layer,
                                      int direction,
-                                     int seq_length,
-                                     int batch_size,
+                                     index_t seq_length,
+                                     index_t batch_size,
                                      int hidden_size,
                                      int mode) {
   size_t size = 0;
@@ -279,9 +279,9 @@ void RNNForwardTraining(DType* ws,
                         bool state_outputs,
                         const int num_layers,
                         const int direction,
-                        const int seq_length,
-                        const int batch_size,
-                        const int input_size,
+                        const index_t seq_length,
+                        const index_t batch_size,
+                        const index_t input_size,
                         const int state_size,
                         DType* x_ptr,
                         DType* hx_ptr,
@@ -321,9 +321,9 @@ void RNNForwardInference(DType* ws,
                          bool state_outputs,
                          const int num_layers,
                          const int direction,
-                         const int seq_length,
-                         const int batch_size,
-                         const int input_size,
+                         const index_t seq_length,
+                         const index_t batch_size,
+                         const index_t input_size,
                          const int state_size,
                          const int projection_size,
                          DType* x_ptr,
@@ -363,9 +363,9 @@ void RNNBackward(DType* ws,
                  DType* rs,
                  const int num_layers,
                  const int direction,
-                 const int seq_length,
-                 const int batch_size,
-                 const int input_size,
+                 const index_t seq_length,
+                 const index_t batch_size,
+                 const index_t input_size,
                  const int state_size,
                  DType* x_ptr,
                  DType* hx_ptr,
--- a/src/operator/rnn-inl.h
+++ b/src/operator/rnn-inl.h
@@ -140,14 +140,14 @@ inline int GetRnnParamSize(int num_layer,
       size *= 3;
       break;
   }
-  int size1 = (input_size + state_size + 2) * size;  // first layer size
-  int size2 = (state_size * direction + state_size + 2) * size;  // other layers size
+  index_t size1 = (input_size + state_size + 2) * size;  // first layer size
+  index_t size2 = (state_size * direction + state_size + 2) * size;  // other layers size
   if (projection_size.has_value()) {
-    int proj_size = projection_size.value();
+    index_t proj_size = projection_size.value();
     size1 = (input_size + proj_size + 2) * size;
     size2 = (proj_size * direction + proj_size + 2) * size;
   }
-  int param_size = size1 + (num_layer - 1) * size2;
+  index_t param_size = size1 + (num_layer - 1) * size2;
   if (projection_size.has_value()) {
     param_size += projection_size.value() * state_size * num_layer * direction;
   }
--- a/src/operator/rnn_impl.h
+++ b/src/operator/rnn_impl.h
@@ -145,7 +145,7 @@ void LstmForwardTraining(DType* ws,
   const int total_layers = D * L;
   Tensor<cpu, 3, DType> hx(hx_ptr, Shape3(total_layers, N, H));
   Tensor<cpu, 3, DType> cx(cx_ptr, Shape3(total_layers, N, H));
-  const int b_size = 2 * H * 4;
+  const index_t b_size = 2 * H * 4;
   const index_t r_size = D * T * N * H * 6;
   const index_t y_offset = T * N * H * 5;
   const index_t cell_size = N * H;
@@ -298,7 +298,7 @@ void LstmForwardInference(DType* ws,
   const int total_layers = D * L;
   Tensor<cpu, 3, DType> hx(hx_ptr, Shape3(total_layers, N, P ? P : H));
   Tensor<cpu, 3, DType> cx(cx_ptr, Shape3(total_layers, N, H));
-  const int b_size = 2 * H * 4;
+  const index_t b_size = 2 * H * 4;
   const index_t cell_size = N * H;
   const index_t projection_size = (P ? P : H) * N;
   DType* y_tmp_ptr = ws + (T + 1) * cell_size * 4 + cell_size * 2;
@@ -553,7 +553,7 @@ void LstmBackward(DType* ws,
   Tensor<cpu, 3, DType> cx(cx_ptr, Shape3(total_layers, N, H));
   Tensor<cpu, 3, DType> dhx(dhx_ptr, Shape3(total_layers, N, H));
   Tensor<cpu, 3, DType> dcx(dcx_ptr, Shape3(total_layers, N, H));
-  const int b_size = 2 * H * 4;
+  const index_t b_size = 2 * H * 4;
   const index_t r_size = D * T * N * H * 6;
   const index_t y_offset = T * N * H * 5;
   const index_t w_size1 = (I + H) * H * 4;      // first layer
--- a/src/operator/rnn_impl.h
+++ b/src/operator/rnn_impl.h
@@ -59,9 +59,9 @@ void LstmForwardTrainingSingleLayer(DType* ws,
                                     DType* rs,
                                     bool state_outputs,
                                     bool bid,
-                                    const int T,
-                                    const int N,
-                                    const int I,
+                                    const index_t T,
+                                    const index_t N,
+                                    const index_t I,
                                     const int H,
                                     const Tensor<cpu, 2, DType> &x,
                                     const Tensor<cpu, 2, DType> &hx,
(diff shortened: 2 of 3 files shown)


connorgoggins:

src/operator/rnn-inl.h
```diff
@@ -123,7 +123,7 @@ struct RNNParam : public dmlc::Parameter<RNNParam> {
 };
 
 inline int GetRnnParamSize(int num_layer,
-                           int input_size,
+                           index_t input_size,
```
access2rohit: Are you sure you don't have to change size1, size2, proj_size and param_size:
https://github.com/apache/incubator-mxnet/pull/17632/files#diff-6dfdca409e69cc495f286170fe1e553eR143
-
https://github.com/apache/incubator-mxnet/pull/17632/files#diff-6dfdca409e69cc495f286170fe1e553eR152

src/operator/rnn-inl.h
```diff
@@ -182,8 +182,8 @@ inline int GetRnnBiasSize(int num_layer,
  *  - output -> h[t](, c[t] additionally with Lstm) time by time(sz: NxH(x2))
  *  - intermediate y[1...T] as next layer's inputs(sz: TxNxHxD)
  */
-inline size_t GetRNNWorkspaceSize(int seq_length,
-                                  int batch_size,
+inline size_t GetRNNWorkspaceSize(index_t seq_length,
+                                  index_t batch_size,
```
access2rohit: Can batch_size be -ve ? @apeforest what do you think ?

src/operator/rnn-inl.h
```diff
@@ -278,9 +278,9 @@ void RNNForwardTraining(DType* ws,
                         bool state_outputs,
                         const int num_layers,
                         const int direction,
-                        const int seq_length,
-                        const int batch_size,
-                        const int input_size,
+                        const index_t seq_length,
+                        const index_t batch_size,
+                        const index_t input_size,
```
access2rohit: Did u check that "seq_length, batch_size, input_size" are index_t in functions LstmForwardTraining, GruForwardTraining, VanillaRNNForwardTraining ? If so can you let me know here, else you may need to update them too

src/operator/rnn-inl.h
```diff
@@ -361,9 +361,9 @@ void RNNBackward(DType* ws,
                  DType* rs,
                  const int num_layers,
                  const int direction,
-                 const int seq_length,
-                 const int batch_size,
-                 const int input_size,
+                 const index_t seq_length,
+                 const index_t batch_size,
+                 const index_t input_size,
```
access2rohit: Did u check that "seq_length, batch_size, input_size" are index_t in functions LstmBackwardTraining, GruBackwardTraining, VanillaRNNBackwardTraining ? If so can you let me know here, else you may need to update them too

tests/nightly/test_large_array.py
```diff
@@ -455,6 +455,21 @@ def npy_instance_norm(data, gamma, beta, axis, eps=1E-5):
...
+        parameters = nd.random_normal(shape=(7,))
+        state = nd.random_normal(shape=(1, 4, 1))
+        mode = 'rnn_relu'
+        state_size = 1
+        num_layers = 1
+
+        out = nd.RNN(data=data, parameters=parameters, state=state, mode=mode,
+                     state_size=state_size, num_layers=num_layers)
+
+        assert out.shape[0] == 268435456
```
access2rohit: Please use constants for constant values. Its good practise and you may need to re-use them in future tests too.

src/operator/rnn-inl.h
connorgoggins: Good point, fixing.

src/operator/rnn-inl.h
connorgoggins: Excellent point, updating now.

src/operator/rnn-inl.h
connorgoggins: Excellent point, updating now.

src/operator/rnn-inl.h
```diff
@@ -140,14 +140,14 @@ inline int GetRnnParamSize(int num_layer,
       size *= 3;
       break;
   }
-  int size1 = (input_size + state_size + 2) * size;  // first layer size
-  int size2 = (state_size * direction + state_size + 2) * size;  // other layers size
+  index_t size1 = (input_size + state_size + 2) * size;  // first layer size
```
access2rohit: Lets prefer size_t for sizes. Or do you think these values can be negative too ?

src/operator/rnn-inl.h
access2rohit: @apeforest  ?

src/operator/rnn-inl.h
access2rohit: sry for changing my opinion again. Since its input size 'size_t' would be better here

src/operator/rnn-inl.h
```diff
@@ -213,8 +213,8 @@ inline size_t GetRNNWorkspaceSize(int seq_length,
 
 inline size_t GetRNNReserveSpaceSize(int num_layer,
                                      int direction,
-                                     int seq_length,
-                                     int batch_size,
+                                     index_t seq_length,
+                                     index_t batch_size,
```
access2rohit: size_t

src/operator/rnn-inl.h
access2rohit: size_t ? make sure API signature doesn't chage. If thats the case then keep it index_t

src/operator/rnn-inl.h
access2rohit: size_t? if its not a breaking chnage

src/operator/rnn-inl.h
access2rohit: size_t? if its not a breaking chnage

src/operator/rnn-inl.h
```diff
@@ -320,9 +320,9 @@ void RNNForwardInference(DType* ws,
                          bool state_outputs,
                          const int num_layers,
                          const int direction,
-                         const int seq_length,
-                         const int batch_size,
-                         const int input_size,
+                         const index_t seq_length,
+                         const index_t batch_size,
+                         const index_t input_size,
```
access2rohit: size_t? if its not a breaking chnage

src/operator/rnn-inl.h
access2rohit: size_t? if its not a breaking chnage

src/operator/rnn_impl.h
```diff
@@ -127,9 +127,9 @@ void LstmForwardTraining(DType* ws,
                          bool state_outputs,
                          const int L,
                          const int D,
```
access2rohit: What are D and L ? can D*L be >5B ?

src/operator/rnn_impl.h
```diff
@@ -146,15 +146,15 @@ void LstmForwardTraining(DType* ws,
   Tensor<cpu, 3, DType> hx(hx_ptr, Shape3(total_layers, N, H));
   Tensor<cpu, 3, DType> cx(cx_ptr, Shape3(total_layers, N, H));
   const int b_size = 2 * H * 4;
```
access2rohit: size_t ?

src/operator/rnn_impl.h
connorgoggins: L is `num_layers`, and D is `direction`. I believe we agreed that we would not support > 2**32 layers, so L should be fine as an `int`. D can have two possible values, 0 or 1, to indicate whether to run the op with bidirectional recurrent layers. Consequently, D*L can never be >5B.

ChaiBapchya: fair to keep D and L as int then.

src/operator/rnn-inl.h
connorgoggins: Agreed - testing changes now.

src/operator/rnn_impl.h
connorgoggins: As discussed offline: `H` represents the LSTM state size, and we are not supporting LSTM states w/dimension >= 2**32, so `b_size` should remain an `int`.

TaoLv:
@zixuanweeei Could you please take a look at the changes? Seems need coordinate with the changes in #17702.

src/operator/rnn_impl.h
zixuanweeei: If I understand it correctly, this is also the reason that  `hidden_size` is remaining to be `int` type, right? If so, `b_size` here, representing the total size of i2h/h2h bias of four gates, still has some overflow risks.

zixuanweeei:
> @zixuanweeei Could you please take a look at the changes? Seems need coordinate with the changes in #17702.

Let's wait for @connorgoggins updating type to `size_t` for some variables representing the size and his feedback.  Overall looks good. We can get this merged first. Due to the fact that the `hidden_size`/`state_size` are still remaining to `int` type, I think there is not much stuff to do for `projection_size` if no further consideration.

src/operator/rnn-inl.h
apeforest: I think we can keep the signed index_t since all the functions being called using signed and the omp loop requires a signed index as well.

connorgoggins:
@zixuanweeei thanks for your feedback! After testing the `size_t` changes on a separate branch, I found that they resulted in a segmentation fault (even on low-dimensional input). When I ran the op on the same input with my `index_t` changes (as they exist in the current state of this PR), the op passed without errors. My `index_t` changes also allow the op to run successfully on large tensor (dimension >= 2^32) input.

With these considerations in mind, we are discussing the best way to move forward.

src/operator/rnn-inl.h
```diff
@@ -123,7 +123,7 @@ struct RNNParam : public dmlc::Parameter<RNNParam> {
 };
 
 inline int GetRnnParamSize(int num_layer,
```
zixuanweeei: Should return `index_t` type? Maybe overflow with a large input size. And the UT only covers cases for large sequence length (first dimension of RNN input data). Would you mind taking some tests for a case with a large input to see whether this function still works?

zixuanweeei:
@connorgoggins Just curious about the reason for the segfault. I don't have much knowledge about that. But I guess it may caused by `for (size_t t = T - 1; t >= 0; --t) {}` in the Backward pass.

src/operator/rnn-inl.h
zixuanweeei: Agree with keeping the signed `index_t` :100:

connorgoggins: I agree @apeforest! I believe the omp loop’s required signed index was the root cause of the segfault when I made the size_t changes.

src/operator/rnn_impl.h
connorgoggins: @zixuanweeei thanks for your feedback. I’m happy to bump b_size up to index_t here if there are overflow concerns.

connorgoggins:
@zixuanweeei you're absolutely right - the segfault was generated on line 2032 of `rnn_impl.h` during the backward pass when I ran the op in ReLU mode. This line lies within the iteration section of the omp loop and, as @apeforest astutely pointed out, the omp loop requires a signed index, which led to errors when the `size_t` changes were implemented.

zixuanweeei:
> @zixuanweeei you're absolutely right - the segfault was generated on line 2032 of `rnn_impl.h` during the backward pass when I ran the op in ReLU mode. This line lies within the iteration section of the omp loop and, as @apeforest astutely pointed out, the omp loop requires a signed index, which led to errors when the `size_t` changes were implemented.

Thanks for trying out the `size_t` type. Let's keep the signed `index_t`.

connorgoggins:
@apeforest @zixuanweeei I believe my latest commit incorporates all of the changes we have discussed. The `b_size` variable and `GetRnnParamSize` function are now of type `index_t`, and the changes have been successfully tested on both small and large tensor input.

PR #17639, opened 2020-02-20
Set Ninja build generator in config.cmake instead of commmand line

leezu:
### Changes ###
- [X] Set Ninja build generator in config.cmake instead of commmand line

## Comments ##
Thanks to @apeforest for the suggestion.

Also remove `--parallel 8` argument from doc, as with ninja the number of cores is autodetected.

### DIFF
--- a/config/darwin.cmake
+++ b/config/darwin.cmake
@@ -55,6 +55,8 @@ set(USE_TVM_OP OFF CACHE BOOL "Enable use of TVM operator build system.")
 #---------------------
 # Compilers
 #--------------------
+set(CMAKE_GENERATOR "Ninja" CACHE STRING "Build Tool Generator used by CMake")
+
 # Compilers are usually autodetected. Uncomment and modify the next 3 lines to
 # choose manually:
 
--- a/config/darwin.cmake
+++ b/config/darwin.cmake
@@ -27,11 +27,11 @@
 #  Next modify the according entries, and then compile by
 #
 #  $ mkdir build; cd build
-#  $ cmake -C ../config.cmake ..
-#  $ cmake --build . --parallel 8
+#  $ cmake  ..
+#  $ cmake --build .
 #
-# You can increase the --parallel 8 argument to match the number of processor
-# cores of your computer.
+# Specify `cmake --build . --parallel N` to set the number of parallel compilation jobs.
+# Default is derived from CPUs available.
 #
 #-------------------------------------------------------------------------------
 
--- a/config/linux.cmake
+++ b/config/linux.cmake
@@ -73,6 +73,8 @@ set(USE_TVM_OP OFF CACHE BOOL "Enable use of TVM operator build system.")
 #---------------------
 # Compilers
 #--------------------
+set(CMAKE_GENERATOR "Ninja" CACHE STRING "Build Tool Generator used by CMake")
+
 # Compilers are usually autodetected. Uncomment and modify the next 3 lines to
 # choose manually:
 
--- a/config/linux.cmake
+++ b/config/linux.cmake
@@ -27,11 +27,11 @@
 #  Next modify the according entries, and then compile by
 #
 #  $ mkdir build; cd build
-#  $ cmake -C ../config.cmake ..
-#  $ cmake --build . --parallel 8
+#  $ cmake ..
+#  $ cmake --build .
 #
-# You can increase the --parallel 8 argument to match the number of processor
-# cores of your computer.
+# Specify `cmake --build . --parallel N` to set the number of parallel compilation jobs.
+# Default is derived from CPUs available.
 #
 #-------------------------------------------------------------------------------
 
--- a/docs/static_site/src/pages/get_started/osx_setup.md
+++ b/docs/static_site/src/pages/get_started/osx_setup.md
@@ -110,9 +110,12 @@ the guide in [Math Library Selection](build_from_source#math-library-selection).
 ```bash
 rm -rf build
 mkdir -p build && cd build
-cmake -C ../config.cmake ..
+cmake ..
 cmake --build .
 ```
+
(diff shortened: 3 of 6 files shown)


docs/static_site/src/pages/get_started/osx_setup.md
```diff
@@ -110,13 +110,10 @@ the guide in [Math Library Selection](build_from_source#math-library-selection).
 ```bash
 rm -rf build
 mkdir -p build && cd build
-cmake -GNinja -C ../config.cmake ..
-cmake --build . --parallel 8
+cmake -C ../config.cmake ..
```
apeforest: Is -C ../config.cmake required?

docs/static_site/src/pages/get_started/osx_setup.md
```diff
@@ -110,13 +110,10 @@ the guide in [Math Library Selection](build_from_source#math-library-selection).
...
+cmake --build .
```
apeforest: Is --build required?

apeforest:
Thanks a lot for enhancing cmake build and instruction. This is almost a dream come true for MXNet developers :)

docs/static_site/src/pages/get_started/osx_setup.md
leezu: Yes. Alternatively you could call `ninja`. Ninja won't work on Windows though, whereas
`cmake --build .` is the same on all configurations.

docs/static_site/src/pages/get_started/osx_setup.md
leezu: Yes

leezu: I think it's reasonable to require specification of the config file. Otherwise users may not understand how to change the config.

apeforest: We have already asked users to `cp config/linux.cmake config.cmake` before this line of instruction. I think it's already very explicit that `config.cmake` will be used to generate the cmake build. This option seems redundant and may cause confusion if users by chance specify a different file other than config.cmake.

docs/static_site/src/pages/get_started/ubuntu_setup.md
```diff
@@ -130,13 +130,10 @@ the guide in [Math Library Selection](build_from_source#math-library-selection).
 ```bash
 rm -rf build
 mkdir -p build && cd build
-cmake -GNinja -C ../config.cmake ..
-cmake --build . --parallel 8
+cmake -C ../config.cmake ..
```
apeforest: Same here. Line 112 has already specify the config.cmake.

docs/static_site/src/pages/get_started/osx_setup.md
leezu: > This option seems redundant and may cause confusion if users by chance specify a different file other than config.cmake.

Maybe on the contrary, it can be useful to enable users to keep `config_1.cmake` and `config_2.cmake`, providing an easy way to switch between the two configs.

Another reason for being explicit here is that our project's `CMakeLists.txt` may not only be used by MXNet "users", but also by third party projects adding MXNet as a dependency.

This use-case is not yet well supported, but modifying our `CMakeLists.txt` to attempt reading `config.cmake` may break the use-case. 

I suggest to investigate this further and improve in a separate PR.

docs/static_site/src/pages/get_started/osx_setup.md
apeforest: If that's the case can we just use the command `ninja` in this doc, since it's for linux-like system. One less option is always better for usability.

docs/static_site/src/pages/get_started/osx_setup.md
```diff
@@ -110,13 +110,10 @@ the guide in [Math Library Selection](build_from_source#math-library-selection).
 ```bash
 rm -rf build
 mkdir -p build && cd build
-cmake -GNinja -C ../config.cmake ..
-cmake --build . --parallel 8
+cmake ..
+cmake --build .
 ```
 
-You can increase the `--parallel 8` argument to match the number of processor
```
apeforest: I think we can leave this as an option in the doc. Some users may not want to use all the processors for compiling.

docs/static_site/src/pages/get_started/ubuntu_setup.md
```diff
@@ -130,13 +130,10 @@ the guide in [Math Library Selection](build_from_source#math-library-selection).
 ```bash
 rm -rf build
 mkdir -p build && cd build
-cmake -GNinja -C ../config.cmake ..
-cmake --build . --parallel 8
+cmake ..
+cmake --build .
 ```
 
-You can increase the `--parallel 8` argument to match the number of processor
```
apeforest: I think we can leave this as an option in the doc. Some users may not want to use all the processors for compiling.

config/linux.cmake
```diff
@@ -27,11 +27,8 @@
 #  Next modify the according entries, and then compile by
 #
 #  $ mkdir build; cd build
-#  $ cmake -C ../config.cmake ..
-#  $ cmake --build . --parallel 8
-#
-# You can increase the --parallel 8 argument to match the number of processor
```
apeforest: I can we can leave this sentence here but remove --parallel 8 from the command line.

config/darwin.cmake
```diff
@@ -27,11 +27,8 @@
 #  Next modify the according entries, and then compile by
 #
 #  $ mkdir build; cd build
-#  $ cmake -C ../config.cmake ..
-#  $ cmake --build . --parallel 8
-#
-# You can increase the --parallel 8 argument to match the number of processor
```
apeforest: I can we can leave this sentence here but remove --parallel 8 from the command line.

apeforest:
@leezu I retriggered CI tests.

PR #17642, opened 2020-02-20
[OpPerf] Fixed Python profiler bug

### LINKED ISSUE #17640: [OPPERF] opperf error out if I use python timer instead of the builtin mxnet profiler
## Description
Seems the python timer did not parse the argument to broadcast_axis properly. It is fine if I use the 'native' profiler.

### Error Message
```
  File "perf.py", line 12, in <module>
    warmup=20, runs=100, profiler='python')
  File "/home/ubuntu/src/incubator-mxnet/benchmark/opperf/utils/benchmark_utils.py", line 177, in run_performance_test
    benchmark_result = _run_nd_operator_performance_test(op, inputs, run_backward, warmup, runs, kwargs_list, profiler)
  File "/home/ubuntu/src/incubator-mxnet/benchmark/opperf/utils/benchmark_utils.py", line 114, in _run_nd_operator_performance_test
    _, _ = benchmark_helper_func(op, warmup, **kwargs_list[0])
  File "/home/ubuntu/src/incubator-mxnet/benchmark/opperf/utils/profiler_utils.py", line 251, in python_profile_it
    modified_args = (args[0], 1, args[2])
```
## To Reproduce
```
#!/usr/bin/python
import mxnet as mx
from mxnet import nd

mx.random.seed(17)
res = run_performance_test(nd.broadcast_axis, run_backward=True, dtype='float32', ctx=mx.gpu(),
                           inputs=[
                               {"data" : (1, 1024, 1), "axis" : (0, 2), "size" : (1024, 8)}
                           ],
                           warmup=20, runs=100, profiler='python')
print(res)
```

connorgoggins:
## Description ##
Currently, the Python profiler for the opperf utility is broken. This fix changes the way args are passed to the underlying op during testing.

Fixes #17640 

### Essentials ###
- [x] Code is well-documented

### Changes ###
- M benchmark/opperf/utils/profiler_utils.py

## Comments ##
Test suite was run with Python profiler on Mac OS X, building the latest version of MXNet (with my fix) from source.

[Full OpPerf test suite - CPU (Native profiler)](https://gist.github.com/connorgoggins/1202852d45e106f9e4d1959e1ffd1985)

[Full OpPerf test suite - CPU (Python profiler)](https://gist.github.com/connorgoggins/2422c1dd1fa8a98b2fad889dc7e9bdd4)

@apeforest

### DIFF
--- a/benchmark/opperf/utils/profiler_utils.py
+++ b/benchmark/opperf/utils/profiler_utils.py
@@ -248,11 +248,17 @@ def python_profile(func):
     @functools.wraps(func)
     def python_profile_it(*args, **kwargs):
         runs = args[1]
+
+        if len(args) > 2:
+            modified_args = (args[0], 1, args[2])
+        else:
+            modified_args = (args[0], 1)
+            
         times = []
 
         for _ in range(runs):
             start_time = time.perf_counter()    # 1
-            res = func(*args, **kwargs)
+            res = func(*modified_args, **kwargs)
             end_time = time.perf_counter()      # 2
             run_time = (end_time - start_time)*1000    # 3
             times.append(run_time)
--- a/benchmark/opperf/utils/profiler_utils.py
+++ b/benchmark/opperf/utils/profiler_utils.py
@@ -248,12 +248,11 @@ def python_profile(func):
     @functools.wraps(func)
     def python_profile_it(*args, **kwargs):
         runs = args[1]
-        modified_args = (args[0], 1, args[2])
         times = []
 
         for _ in range(runs):
             start_time = time.perf_counter()    # 1
-            res = func(*modified_args, **kwargs)
+            res = func(*args, **kwargs)
             end_time = time.perf_counter()      # 2
             run_time = (end_time - start_time)*1000    # 3
             times.append(run_time)
--- a/benchmark/opperf/utils/profiler_utils.py
+++ b/benchmark/opperf/utils/profiler_utils.py
@@ -248,12 +248,7 @@ def python_profile(func):
     @functools.wraps(func)
     def python_profile_it(*args, **kwargs):
         runs = args[1]
-
-        if len(args) > 2:
-            modified_args = (args[0], 1, args[2])
-        else:
-            modified_args = (args[0], 1)
-            
+        modified_args = (args[0], 1)
         times = []
 
         for _ in range(runs):


connorgoggins:

benchmark/opperf/utils/profiler_utils.py
```diff
@@ -248,12 +248,11 @@ def python_profile(func):
     @functools.wraps(func)
     def python_profile_it(*args, **kwargs):
         runs = args[1]
-        modified_args = (args[0], 1, args[2])
```
apeforest: I don't remember the reason why we need to write this way in the first place. @ChaiBapchya could you please review. @connorgoggins Have you run through existing tests to make sure this does not break any existing usage?

connorgoggins: @apeforest yes, ran full OpPerf suite with Python profiler with this change and everything passed.

ChaiBapchya:
preloaded and multi_* ops aren't being tracked for some reason. could you fix that too? @connorgoggins

benchmark/opperf/utils/profiler_utils.py
ChaiBapchya: So the args that are passed to this function are
args[0] = op
args[1] = warmup / runs (number of times to run for warmup or number of times to run)
args[2] - rest of the args

The way it worked for native MXNet CPP profiler is that you could pass the runs
But for Python's time it function, we had to manually run for loop for the number of time to run.
So that's what I did there

we copy the number of runs in a variable in run and then run it that many number of times 
For each run, we use python time it function to time it and then take average, mean, max, etc values for each of those individual python time runs.

Makes sense? @apeforest

apeforest: @ChaiBapchya So basically you are saying we don't need to do this modified_args for python profiler, right? So @connorgoggins change is valid?

apeforest:
> preloaded and multi_* ops aren't being tracked for some reason. could you fix that too? @connorgoggins

Let's fix that in a separate PR.

benchmark/opperf/utils/profiler_utils.py
ChaiBapchya: We **do** need to modify the args to meet the requirement for capturing per run timing info using python's time_it function. This needs to handled in a way that doesn't break existing native profiler.

benchmark/opperf/utils/profiler_utils.py
```diff
@@ -248,12 +248,11 @@ def python_profile(func):
...
         times = []
 
         for _ in range(runs):
             start_time = time.perf_counter()    # 1
-            res = func(*modified_args, **kwargs)
```
ChaiBapchya: @connorgoggins @apeforest 
If we pass the *args as is, it will still have
args[0] as op
args[1] as runs
For eg if user passed runs as 10

So the native profiler would run 10 times and so will the for loop run 10 times (for timing Python profiler)

Coz the func here is nd_forward_backward_profile or nd_forward_profile (both take runs as a parameter)

benchmark/opperf/utils/profiler_utils.py
```diff
@@ -248,7 +248,7 @@ def python_profile(func):
     @functools.wraps(func)
     def python_profile_it(*args, **kwargs):
         runs = args[1]
-        modified_args = (args[0], 1, args[2])
+        modified_args = (args[0], 1)
```
ChaiBapchya: Now it looks good!

Issue #17548, opened 2020-02-07
CI doesn't test cmake build properly


leezu:
Evidence https://github.com/apache/incubator-mxnet/issues/17514 https://github.com/apache/incubator-mxnet/issues/17545

Issue #17641, opened 2020-02-20
OpenMP Error


icemelon9:
## Description
Compiled MxNet has duplicate OpenMP library link to both libomp and libiomp.

### Error Message
(Paste the complete error message. Please also include stack trace by setting environment variable `DMLC_LOG_STACK_TRACE_DEPTH=10` before running your script.)
```
OMP: Error #15: Initializing libiomp5.so, but found libomp.so already initialized.
OMP: Hint This means that multiple copies of the OpenMP runtime have been linked
into the program. That is dangerous, since it can degrade performance or cause
incorrect results. The best thing to do is to ensure that only a single OpenMP
runtime is linked into the process, e.g. by avoiding static linking of the
OpenMP runtime in any library. As an unsafe, unsupported, undocumented
workaround you can set the environment variable KMP_DUPLICATE_LIB_OK=TRUE to
allow the program to continue to execute, but that may cause crashes or silently
produce incorrect results. For more information, please see
http://www.intel.com/software/products/support/.
```

## To Reproduce
I have both Intel MKL and MKLDNN library installed on Ubuntu 18.04. Use the following config to compile MxNet will lead the error shown above.
```
cmake -DUSE_CUDA=0 -DUSE_CUDNN=0 -DUSE_MKLDNN=1 -DCMAKE_BUILD_TYPE=Release -GNinja ..
ninja -v
```

## What have you tried to solve it?

After I deleted 3rdparty/openmp, and recompiled mxnet, this error no longer occurs.

## Environment

Ubuntu 18.04, installed with Intel MKL and MKLDNN library.

sl1pkn07:
that is because `git clone --recursive` pull openmp, and the cmake script don't take care if use system openMP or not

https://github.com/apache/incubator-mxnet/blob/9dcf71d8fe33f77ed316a95fcffaf1f7f883ff70/CMakeLists.txt#L390-L430

TaoLv:
Same issue: https://github.com/apache/incubator-mxnet/issues/17366

sl1pkn07:
this case also same effects with intel-dnnl.  

some distros already provide a package intel-dnnl, but mxnet force download the sources again

leezu:
@cjolivier01 you previously vetoed changing the omp configuration in cmake build, due to a race condition that had not been fixed. As that has been fixed, are you OK with proceeding to prefer system OMP for the CMake build by default? Or what is your recommendation?

Static build should still statically build omp.

[...]

PR #17645, opened 2020-02-21
[CI] Test CMake builds instead of Makefile builds

leezu:
## Description ##
Use cmake build for most CI builds in preparation of dropping Makefile build

Fixes https://github.com/apache/incubator-mxnet/issues/17548

## Comments ##
The following Makefile based builds are preserved
1) staticbuild scripts. Changes to CD pipeline will happen in https://github.com/apache/incubator-mxnet/pull/17646
2) Docs builds. Language binding specific build logic requires further changes
3) Jetson build. Jetpack 3.3 toolchain based on Cuda 9.0 causes 'Internal
   Compiler Error (codegen): "there was an error in verifying the lgenfe
   output!"' errors with cmake. This seems to be a known issue in Cuda 9.0 and
   we need to update Jetpack toolchain to work around it.
4) MKL builds. Waiting for fix of https://github.com/apache/incubator-mxnet/issues/17641

Improvements to CMake build
- Enable -Werror for RelWithDebugInfo build in analogy to "make DEV=1" build
- Add USE_LIBJPEG_TURBO to CMake build
- Improve finding Python 3 executable

Changes to CI setup
- Install protobuf in ubuntu_core.sh and centos_core.sh
- Install up-to-date CMake on Centos 7
- Don't use RelWithDebInfo on Android builds, as gcc 4.9 throws
  -Wdelete-non-virtual-dtor

Code changes
- Locally disable warnings introduced by GCC7


ci/docker/runtime_functions.sh
```diff
@@ -455,36 +437,40 @@ build_ubuntu_cpu() {
...
-        USE_MKLDNN=0                  \
-        USE_DIST_KVSTORE=1            \
-        USE_LIBJPEG_TURBO=1           \
-        USE_SIGNAL_HANDLER=1          \
-        -j$(nproc)
-    make cython PYTHON=python3
+    pushd .
+    cmake \
+        -DCMAKE_BUILD_TYPE="RelWithDebInfo" \
+        -DUSE_TVM_OP=ON \
```
apeforest: I think we should turn off this by default and create a separate to test TVM_ON instead.

ci/docker/runtime_functions.sh
```diff
@@ -455,36 +437,40 @@ build_ubuntu_cpu() {
...
-        USE_TVM_OP=1                  \
-        USE_MKLDNN=0                  \
-        USE_INTEL_PATH=/opt/intel     \
-        USE_DIST_KVSTORE=1            \
-        USE_SIGNAL_HANDLER=1          \
-        -j$(nproc)
+    pushd .
+    cmake \
+        -DCMAKE_BUILD_TYPE="RelWithDebInfo" \
+        -DUSE_TVM_OP=ON \
```
apeforest: I think we should turn off this by default and create a separate to test TVM_ON instead.

ci/docker/runtime_functions.sh
```diff
@@ -617,63 +595,63 @@ build_ubuntu_cpu_clang_tidy() {
...
-        USE_TVM_OP=1                  \
-        USE_BLAS=openblas             \
-        USE_SIGNAL_HANDLER=1          \
-        -j$(nproc)
+    pushd .
+    cd /work/build
+    cmake \
+        -DCMAKE_BUILD_TYPE="RelWithDebInfo" \
+       -DUSE_MKL_IF_AVAILABLE=OFF \
+       -DUSE_TVM_OP=ON \
```
apeforest: I think we should turn off this by default and create a separate to test TVM_ON instead.

ci/docker/runtime_functions.sh
```diff
@@ -617,63 +595,63 @@ build_ubuntu_cpu_clang_tidy() {
...
-        USE_SIGNAL_HANDLER=1          \
-        USE_INTEL_PATH=/opt/intel/    \
-        -j$(nproc)
+    pushd .
+    cd /work/build
+    cmake \
+        -DCMAKE_BUILD_TYPE="RelWithDebInfo" \
+        -DUSE_MKL_IF_AVAILABLE=ON \
+        -DBLAS="MKL" \
+        -DUSE_TVM_OP=ON \
```
apeforest: I think we should turn off this by default and create a separate to test TVM_ON instead.

ci/docker/runtime_functions.sh
```diff
@@ -735,96 +713,93 @@ build_ubuntu_gpu_tensorrt() {
...
-        USE_TVM_OP=1                              \
-        CUDA_ARCH="$CI_CUDA_COMPUTE_CAPABILITIES" \
-        USE_SIGNAL_HANDLER=1                      \
-        -j$(nproc)
+    pushd .
+    cd /work/build
+    cmake \
+        -DCMAKE_BUILD_TYPE="RelWithDebInfo" \
+        -DUSE_MKL_IF_AVAILABLE=OFF \
+        -DUSE_TVM_OP=ON \
```
apeforest: I think we should turn off this by default and create a separate to test TVM_ON instead.

ci/docker/runtime_functions.sh
```diff
@@ -735,96 +713,93 @@ build_ubuntu_gpu_tensorrt() {
...
```
apeforest: I think we should turn off this by default and create a separate to test TVM_ON instead.

ci/docker/runtime_functions.sh
```diff
@@ -735,96 +713,93 @@ build_ubuntu_gpu_tensorrt() {
...
-        CUDA_ARCH="$CI_CUDA_COMPUTE_CAPABILITIES" \
-        USE_SIGNAL_HANDLER=1                      \
-        -j$(nproc)
-    make cython PYTHON=python3
+    pushd .
+    cd /work/build
+    cmake \
+        -DCMAKE_BUILD_TYPE="RelWithDebInfo" \
+        -DUSE_MKL_IF_AVAILABLE=OFF \
+        -DUSE_TVM_OP=ON \
```
apeforest: I think we should turn off this by default and create a separate to test TVM_ON instead.

ci/docker/runtime_functions.sh
```diff
@@ -735,96 +713,93 @@ build_ubuntu_gpu_tensorrt() {
...
-        USE_SIGNAL_HANDLER=1                      \
-        -j$(nproc)
-
-    make cython PYTHON=python3
+    pushd .
+    cd /work/build
+    cmake \
+        -DCMAKE_BUILD_TYPE="RelWithDebInfo" \
+        -DUSE_MKL_IF_AVAILABLE=OFF \
+        -DUSE_TVM_OP=OFF \
```
apeforest: I think we should change this to TVM_OP = ON and rename this test to `build_ubuntu_gpu_cuda101_cudnn7_tvm_op()`

ci/docker/runtime_functions.sh
```diff
@@ -855,7 +830,6 @@ build_ubuntu_gpu_cmake_mkldnn() {
         -DUSE_CUDA=1                            \
         -DUSE_CUDNN=1                           \
         -DUSE_TVM_OP=1                          \
```
apeforest: I think we should turn off this by default and create a separate to test TVM_ON instead.

ci/docker/runtime_functions.sh
```diff
@@ -897,7 +870,6 @@ build_ubuntu_gpu_cmake_no_rtc() {
         -DUSE_CUDA=ON                           \
         -DUSE_CUDNN=ON                          \
         -DUSE_TVM_OP=ON                         \
```
apeforest: I think we should turn off this by default and create a separate to test TVM_ON instead.

leezu:
@apeforest The PR currently attempts to not change any of the configurations tested, but only switch the build system.

Once CI passes, we can consider merging and have a separate PR for changing configurations or iterate on this PR.

apeforest:
Could you first test this in the mxnet-ci-dev environment? If there is anything wrong, we don't want it to break the CI production right?

leezu:
What do you mean? The CI is testing this PR. This has noting to do with the dev environment?

apeforest:
Right, that makes sense.

cmake/Modules/FindMKL.cmake
```diff
@@ -122,10 +126,9 @@ set(INTEL_ROOT "/opt/intel" CACHE PATH "Folder contains intel libs")
     list(APPEND MKL_LIBRARIES ${${__mkl_lib_upper}_LIBRARY})
   endforeach()
 
-
   if(NOT MKL_USE_SINGLE_DYNAMIC_LIBRARY)
-    if (MKL_USE_STATIC_LIBS)
-      set(__iomp5_libs iomp5 libiomp5mt.lib)
+    if (MKL_USE_STATIC_LIBS AND UNIX)
+      set(__iomp5_libs libiomp5.a libiomp5mt.lib)
```
TaoLv: Hmm, why we need static link libiomp5.a?

leezu:
@cjolivier01 this PR fails without fixing OMP. That's because the CI previously didn't test cmake OMP.
The fix in this PR is exactly what we discussed and agreed on. Please revisit your review as it's not technical.

cmake/Modules/FindMKL.cmake
leezu: Otherwise symbols are not resolved at link time and the llvm openmp will be used.

cjolivier01: Not in favor of tail wagging the dog solution, bending to allow mkl to behave foolishly.
Also, this doesn;t help transitive dependencies, i.e. some other clang library has pulled in libomp, and then you still have the problem.  I believe the fix is to squash mkl's bad behavior, possily through static linking or some other way.

leezu: @cjolivier01 I'll remove testing MKL builds from this PR and you can contribute a separate PR that fixes MKL build and enables tests.

leezu:
@cjolivier01 I'll remove the MKL fix and testing MKL builds from this PR and you can contribute a separate PR that fixes MKL build and enables tests.

Please remove your veto.

cmake/Modules/FindMKL.cmake
leezu: Btw, if you look a few line above, you see that we are linking MKL statically.
Even if MKL is statically linked, dynamic linker may load llvm openmp as long as llvm openmp is pulled in by clang (depending on the link order)

3rdparty/mshadow/mshadow/base.h
```diff
@@ -839,7 +840,7 @@ MSHADOW_XINLINE bool MaxValue<bool>(void) {
 /*! \brief maximum value of uint32_t */
 template<>
 MSHADOW_XINLINE uint32_t MaxValue<uint32_t>(void) {
-  return -1;
+  return std::numeric_limits<uint32_t>::max();
```
apeforest: What's the indication of this change? Would it break some existing logic?

leezu: Fixes a compiler warning treated as error due to `-WError` on CI.
Essentially we're currently relying on unsigned integer overflow to obtain the max value. `std::numeric_limits<uint32_t>::max();` gives the max value directly

apeforest: Thanks. I am actually not questioning this change which I think is correct. I am wondering if you have checked places where the return value of "-1" from this function is expected and if this will break that assumption.

cmake/Modules/FindMKL.cmake
TaoLv: @leezu, I'm thinking about static linking MKL libraries and leaving omp dynamic link to mxnet.so. It should not have any issue about unresolved symbol. But please note that if linking libiomp5.so or llvm libomp.so, the MKL library should be `libmkl_intel_thread.a`, otherwise if libgomp.so, the MKL library should be `libmkl_gnu_thread.a`. Refer to the changes to makefile: https://github.com/apache/incubator-mxnet/pull/17693/files

sxjscience:
@ciyongch Would this go to v1.7? I come up to this commit when I'm cherry-picking the related commits one by one

ciyongch:
Hi @sxjscience , sorry for overlooking your message. I don't see there's a strong requirement for this and likely not to include this patch in v1.7.

PR #17648, opened 2020-02-21
Enable users to specify BLAS

apeforest:
## Description ##
Fix a bug where users can't specify BLAS if MKL is installed in the path. This is caused by the logic in https://github.com/apache/incubator-mxnet/blob/master/cmake/ChooseBlas.cmake#L21

@leezu Please review


leezu:
Restarted CI job

PR #17652, opened 2020-02-21
Use multi-tensor sumSQ in clip_global_norm

MoisesHer:
## Description ##
Using multi-tensor sum of squares in gluon: clip_global_norm.
Instead of computing the sum of squares of each input array sequentially, compute them in parallel (multi-tensor). 

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [x] gluon: clip_global_norm was modified to use mxnet.nd.multi_sum_sq op and compute the sumSq of several arrays in parallel
- [x] Test in tests/python/gpu/test_gluon_gpu:test_global_norm_clip_multi_device was extended to have 2 arrays on gpu(0) and 2 arrays on cpu(0), so that multi-tensor sumSq is tested for each of these contexts

### DIFF
--- a/python/mxnet/gluon/utils.py
+++ b/python/mxnet/gluon/utils.py
@@ -31,6 +31,8 @@ import collections
 import weakref
 import requests
 
+import mxnet as mx
+from mxnet import nd
 import numpy as np
 
 from .. import ndarray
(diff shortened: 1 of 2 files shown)


python/mxnet/gluon/utils.py
```diff
@@ -145,12 +143,12 @@ def group_by_ctx(arr_list):
     all_ctx_sum = []
     ctx = arrays[0].context
     for group in arrays_groups:
-        sum_sq = mx.nd.multi_sum_sq(*arrays_groups[group],
-                                    num_arrays=len(arrays_groups[group]))
-        sum_sq = nd.add_n(*sum_sq)
+        sum_sq = ndarray.multi_sum_sq(*arrays_groups[group],
```
apeforest: What’s the difference between mxnet.ndarray and ndarray here?

MoisesHer: I think this is equivalent,
so I modified it to ndarray, since this was already imported for other functions in the file

PR #17672, opened 2020-02-24
[Numpy] more support for boolean indexing/assign and change on implementation

Alicia1529:
## Description ##
1.  implement boolean indexing through advanced indexing by following the below method

>  In general if an index includes a Boolean array, the result will be identical to inserting `obj.nonzero()` into the same position and using the integer array indexing mechanism described above. `x[ind_1, boolean_array, ind_2]` is equivalent to `x[(ind_1,) + boolean_array.nonzero() + (ind_2,)]`.

2. support boolean indexing/assign mixed with other ndarray

3. support boolean assign without restriction on position (previously only support at most one boolean array at the end of the index array)

4. support `True/ array(True)`, `False/array(False)` as part of the indices

### DIFF
--- a/python/mxnet/ndarray/ndarray.py
+++ b/python/mxnet/ndarray/ndarray.py
@@ -39,6 +39,7 @@ from ..base import mx_uint, NDArrayHandle, check_call, DLPackHandle, mx_int, mx_
 from ..base import ctypes2buffer
 from ..runtime import Features
 from ..context import Context, current_context
+from ..util import is_np_array
 from . import _internal
 from . import op
 from ._internal import NDArrayBase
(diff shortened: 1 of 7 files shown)


ciyongch:
@Alicia1529 , can you help to back port this PR to v1.7.x and v1.x branches as well, which will be included in the upcoming 1.7.0 release? As the issue is existing in these two branches so far. Thanks!

Alicia1529:
> @Alicia1529 , can you help to back port this PR to v1.7.x and v1.x branches as well, which will be included in the upcoming 1.7.0 release? As the issue is existing in these two branches so far. Thanks!

Hi, sorry for my late reply. I will do it within this weekend.

ciyongch:
Thanks @Alicia1529 , please also tag me when the new PRs are created.

PR #17682, opened 2020-02-25
Revert "MXNet FFI for Operator Imperative Invocation"

leezu:
Reverts apache/incubator-mxnet#17510

Breaks cmake build: `libmxnet.so: undefined symbol: MXNetFuncListGlobalNames`


leezu:
@hzfan feel free to provide a quick fix and I'll close this. Otherwise let's revert FFI until fixed.

leezu:
Closing this, as the error is specific to a special case where Python binding attempts to load an inconsistent set of shared libraries. On one side, a shared library compiled prior to #17510 as well as some other shared library resulting from #17510.

PR #17685, opened 2020-02-25
for OPECV4.X

chengc843988:
## Description ##
support OPENCV4 for Jetson Nano
### Changes ###
src/io/image_aug_default.cc 
src/io/image_det_aug_default.cc 
src/io/image_io.cc
## Comments ##
spport Jetson Nano, OpenCV 4.X


chengc843988:
Thanks for your reply.

I am recalling what I had done. I followed the procudres on web.
May be I built mxnet on v1.4.1, and the macro not been defined.

Thanks a lot.

PR #17702, opened 2020-02-27
Support projection feature for LSTM on CPU (Only Inference)

zixuanweeei:
## Description ##
[`gluon.rnn.LSTM`](https://mxnet.incubator.apache.org/api/python/docs/api/gluon/rnn/index.html?highlight=lstm#mxnet.gluon.rnn.LSTM) has an argument - `projection_size` - which enables the projection feature for LSTM operator. Previously, this feature is not supported on the CPU backend. This PR aims to add it to the CPU backend. It should be noticed that only the forward pass is ready in this PR. Backward pass and `needs_grads` scenario is not adapted to this feature. When it runs into those contents, it throws an error.

### DIFF
--- a/src/operator/rnn.cc
+++ b/src/operator/rnn.cc
@@ -190,20 +190,19 @@ static std::vector<ResourceRequest> RNNResourceEx(const NodeAttrs& attrs, const 
   return request;
 }
 
+#if MXNET_USE_MKLDNN == 1
 inline static bool RNNStorageType(const nnvm::NodeAttrs& attrs,
                                   const int dev_mask,
                                   DispatchMode* dispatch_mode,
                                   std::vector<int> *in_attrs,
                                   std::vector<int> *out_attrs) {
-  DispatchMode wanted_mode = DispatchMode::kFCompute;
-
-#if MXNET_USE_MKLDNN == 1
-  wanted_mode = DispatchMode::kFComputeEx;
-#endif  // MXNET_USE_MKLDNN == 1
-
-  return storage_type_assign(out_attrs, mxnet::kDefaultStorage,
-                             dispatch_mode, wanted_mode);
+  const RNNParam& param = nnvm::get<RNNParam>(attrs.parsed);
+  const bool support_mkldnn_rnn =
+      !param.projection_size.has_value() && dmlc::GetEnv("MXNET_USE_MKLDNN_RNN", 1);
+  return MKLDNNStorageType(attrs, dev_mask, support_mkldnn_rnn,
+                           dispatch_mode, in_attrs, out_attrs);
 }
+#endif  // MXNET_USE_MKLDNN == 1
 
 struct RNNGrad {
   const char *op_name;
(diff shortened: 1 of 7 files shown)


pengzhao-intel:
@eric-haibin-lin

src/operator/rnn.cc
```diff
@@ -385,7 +382,9 @@ The definition of GRU here is slightly different from paper but compatible with
 })
```
eric-haibin-lin: I don't think the projection support is clear in the documentation. Could you update the documentation with LSTMP support when projection_size is set? You can refer to https://github.com/apache/incubator-mxnet/blob/62a85f365b819829fedb60116f803e0c0a3c554c/python/mxnet/gluon/contrib/rnn/rnn_cell.py#L197 

Thanks!

zixuanweeei: Sure. Thanks for pointing out that.

zixuanweeei: Done. Please take a review again. Thanks.

zixuanweeei:
CI has passed last time. The latest commit just added some documents for the projection feature. Accordingly, it should have no impact on functionality. Let's wait for CI validation. Please take a review. Thanks. @ciyongch @pengzhao-intel

src/operator/rnn.cc
```diff
@@ -385,7 +399,9 @@ The definition of GRU here is slightly different from paper but compatible with
...
 .set_attr<mxnet::FInferShape>("FInferShape", RNNShape)
 .set_attr<nnvm::FInferType>("FInferType", RNNType)
+#if MXNET_USE_MKLDNN == 1
```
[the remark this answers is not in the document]
TaoLv: Make sure you know that, if MKL-DNN is not used, FInferStorageType will not be registered.

src/operator/rnn.cc
```diff
@@ -427,7 +443,9 @@ NNVM_REGISTER_OP(_backward_RNN)
 .set_attr_parser(ParamParser<RNNParam>)
 .set_attr<bool>("TIsLayerOpBackward", true)
 .set_attr<nnvm::TIsBackward>("TIsBackward", true)
+#if MXNET_USE_MKLDNN == 1
```
TaoLv: Same here, merge this check with the one at L450.

src/operator/rnn_impl.h
```diff
@@ -219,7 +220,10 @@ void LstmForwardInferenceSingleLayer(DType* ws,
                                      DType* cy_ptr) {
   using namespace mshadow;
   const Tensor<cpu, 2, DType> wx(w_ptr, Shape2(H * 4, I));
-  const Tensor<cpu, 2, DType> wh(w_ptr + I * H * 4, Shape2(H * 4, H));
+  const Tensor<cpu, 2, DType> wh(w_ptr + I * H * 4, Shape2(H * 4, (P ? P : H)));
+  Tensor<cpu, 2, DType> whr(w_ptr, Shape2(1, 1));
+  if (P > 0)
+    whr = Tensor<cpu, 2, DType>(wh.dptr_ + P * 4 * H, Shape2(P, H));
```
TaoLv: Let's put this into the same line of `if (P > 0)` or add `{. .. }` for it, like what you're doing at L236.

src/operator/rnn.cc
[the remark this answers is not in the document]
zixuanweeei: Done. The default storage type inference function will be executed when FInferStorageType is not registered.

src/operator/rnn.cc
zixuanweeei: Done. Thanks.

src/operator/rnn_impl.h
zixuanweeei: Put them into the same line, as well as L236.

pengzhao-intel:
Thanks for all of your great works. I will merge PR after passed CI.

PR #17725, opened 2020-02-29
Fix get_started pip instructions

leezu:
The class attribute in the respective .md files must match one of the versions
listed in the version dropdown specified in get_started.html

Without this fix, the instructions for installing version 1.5 will always be shown, even if users select another version in the dropdown.

CC @TaoLv, as this error was introduced in #16442.

### DIFF
--- a/docs/static_site/src/_includes/get_started/windows/python/cpu/pip.md
+++ b/docs/static_site/src/_includes/get_started/windows/python/cpu/pip.md
@@ -1,12 +1,12 @@
 Run the following command:
 
-<div class="v1-5-0">
+<div class="v1-5-1">
 
 {% highlight bash %}
 $ pip install mxnet
 {% endhighlight %}
 
-</div> <!-- End of v1-5-0 -->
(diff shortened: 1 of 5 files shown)


docs/static_site/src/_includes/get_started/windows/python/cpu/pip.md
```diff
@@ -70,4 +70,4 @@ $ pip install --pre mxnet -f https://dist.mxnet.io/python/all
 </div> <!-- End of master-->
 
 {% include /get_started/pip_snippet.md %}
-{% include /get_started/gpu_snippet.md %}
```
ChaiBapchya: what has changed in all these gpu_snippets?

ptrendx: I bet it is eol character - those md files had no eol before

leezu: Yes. My editor ensures each file has eol by default

PR #17734, opened 2020-03-01
[MXNET-889] Implement ONNX export for gluon LSTM.

### LINKED ISSUE #15341: Support LSTM operator in the ONNX module
LSTMs are used in most sequence models. There are inference engines that use the ONNX format to serve pre-trained models. This operator needs to be added to the ONNX import and export module.

RuRo:
## Description ##
Implements 3 new node conversion functions, to facilitate the conversion of `gluon.rnn.LSTM` blocks to ONNX. (fixes #15341)

1) `_zeros` (as well as `_ones` and `_full` as a bonus)
    `gluon.rnn.LSTM` uses `sym.zeros` to [initialize the initial recurrent states](https://github.com/apache/incubator-mxnet/blob/10a12d59f67ad21032a94b1721aaf9b96fddac85/python/mxnet/gluon/rnn/rnn_layer.py#L234).
2) `_rnn_param_concat` which is used by mxnet to [concatenate all the flattened parameters](https://github.com/apache/incubator-mxnet/blob/10a12d59f67ad21032a94b1721aaf9b96fddac85/python/mxnet/gluon/rnn/rnn_layer.py#L278), before passing them to the `RNN` node.
3) `RNN` which is the backend op for the actual `LSTM` computation.

I've also implemented/updated a couple helper functions in the [`_op_translations.py`](https://github.com/RuRo/incubator-mxnet/blob/onnx_export_RNN/python/mxnet/contrib/onnx/mx2onnx/_op_translations.py) file in a separate commit to hopefully reduce the boilerplate amount.

### Essentials ###

## Tests ##

I didn't add any tests for this node yet, since I am not 100% sure, how exactly to do that.
Some other PRs with ONNX exports, just add the new node names to a list in [`tests/python-pytest/onnx/test_cases.py`](https://github.com/apache/incubator-mxnet/blob/10a12d59f67ad21032a94b1721aaf9b96fddac85/tests/python-pytest/onnx/test_cases.py#L96), however, I don't see, how would this add any actual tests, without a reference ONNX implementation available.

I would gladly add some unittests, if someone can point the right direction though. For now, I've verified the correctness of my conversion by
1) Running this script:

    <details>
    <summary>test_rnn_export_onnx.py</summary>

    ```python
    #!/bin/env python3
    import numpy as np
    import mxnet as mx
    import onnxruntime as rt

    import logging
    import shutil
    import os
    logging.getLogger().setLevel(logging.INFO)


    # Configuration
    path = 'lstm-model'
    seq_len = 10
    batch_size = 32
    input_size = 16
    hidden_size = 8
    bidirectional = True


    # Create input data
    shape = (seq_len, batch_size, input_size)
    I = mx.nd.random.randn(*shape)


    # Create LSTM network
    seq = mx.gluon.nn.HybridSequential()
    with seq.name_scope():
        seq.add(mx.gluon.rnn.LSTM(hidden_size, bidirectional=bidirectional))
    seq.initialize()
    seq(I)


    # Init all parameters with random values
    for k, p in seq.collect_params().items():
        p.data()[...] = np.random.randn(*p.data().shape)
    seq.hybridize(static_alloc=True, static_shape=True)
    res_mx = seq(I)
    res_mx = res_mx.asnumpy()


    # Export to symbol and then to ONNX and forward with onnxruntime
    seq.export(path)
    mx.contrib.onnx.export_model(
        f'{path}-symbol.json',
        f'{path}-0000.params',
        [shape], np.float32,
        onnx_file_path=f'{path}.onnx',
        verbose=True
    )


    # Forward, using onnxruntime
    sess = rt.InferenceSession(f'{path}.onnx')
    res_onnx, = sess.run(None, {'data': I.asnumpy()})


    # Check results
    diff = np.abs(res_mx - res_onnx)
    print(f"{diff.mean()=}")
    print(f"{diff.max()=}")
    np.testing.assert_almost_equal(res_mx, res_onnx, decimal=4)
    # decimal=4 is the same as mxnet unittests for other ONNX ops


    # Clean up files
    os.remove(f"{path}-symbol.json")
    os.remove(f"{path}-0000.params")
    os.remove(f"{path}.onnx")
    ```

    </details>

    Which generates a model with an `LSTM` node, fills all the parameters with random data, exports it to a `symbol.json` and `0000.params` and then converts the model to `ONNX`. It then loads the model using `onnxruntime` and forwards some random data through both the `mxnet` as well as the `onnx` models and asserts, that the results are close.

2) Using this conversion on a CRNN network, I trained at my job and verifying that the prediction results are the same on a subset of out test set. (unfortunately, I can't give too much details about this model, so you'll have to trust me here)

## Comments ##
There are a few details in my current implementation that should be noted:
1) `_zeros`
    - It seems, that `mx.sym` can broadcast a `_zeros`/`_ones`/`_full` node along an axis with dimension 0 as if it were a 1, while `ONNX` can't do that.
        ```python
        import mxnet as mx
        a = mx.sym.zeros((10, 0, 30))
        b = mx.sym.Variable('b')
        (a * b).eval(b=mx.nd.zeros((10, 20, 30)))[0].shape
        ```
        Currently, my implementation converts all the 0 dims to 1s. Is this correct?
    - The version of `ONNX`, that mxnet uses doesn't support some newer nodes like `ConstantOfShape` and I was not able to find a working onnx implementation, which actually supported the currently deprecated `ConstantFill` op, so I had to actually store the `_zeros` data in a tensor, which ends up being stored in the onnx model file.<br/>
    This also applies to the helper funtions and the other nodes. I've had to settle for some "less than elegant" solutions in some cases, because of the lacking opset support. I might have also missed a better way to do some operations, so if you have any better alternatives, I would gladly change my implementation.<br/>

2) `_rnn_param_concat`
    You can read more about my implementation in the code (comments), but the general outline is that since `ONNX` expects the `RNN` parameters prepared in a completely different way compared to mxnet, I've had to employ some "dark" magic, involving `convert_rnn_param_concat` generating multiple nodes and `convert_RNN` 'guessing' these node names, using regular expressions.<br/>
    As a result, the current conversion mechanism only works, if the parameters of the RNN block are concatenated using `_rnn_param_concat` (and not for example `Concat`). If the conversion of of `mx.sym.RNN` (without `gluon`) is a widely requested feature, we would have to convert the `_rnn_param_concat` node to a regular `Concat`, and then split the parameters back apart in the `RNN` conversion, which is not very nice IMO.<br/>

3) `RNN`
    Even though, this PR implements the conversion of the `RNN` node, currently only the `LSTM` mode is supported and many of the "customization" options are also not implemented.
    **Implemented:**
    - `mode=LSTM`
    - `layout='TNC'`
    - `bidirectional=True` and `False`
    
    **Not Implemented**, but probably supported by ONNX (easy to implement):
    - `mode=GRU, RNN_relu, RNN_tanh` (via GRU and RNN ONNX nodes and the `activations` attribute)
    - `use_sequence_length` (via ONNX `sequence_lens` optional input)
    - custom `states` inputs and `states` outputs (input states might already work, output states - via 'Y_h' and 'Y_c' optional outputs)
    - custom `layout != TNC` (need `SwapAxis` op conversion)
    
    **Not Implemented**, and probably **not** supported by ONNX (hard to implement):
    - `dropout` (the `p` input in ONNX is `peephole`, *not* dropout probability)
    - `state_clip_min` and `state_clip_max` (the `clip` attribute in ONNX seems to produce different results than `state_clip_min`/`max`)
    - `state_clip_nan`
    - `projection_size`
    - `num_layers`

### DIFF
--- a/python/mxnet/contrib/onnx/mx2onnx/_op_translations.py
+++ b/python/mxnet/contrib/onnx/mx2onnx/_op_translations.py
@@ -950,6 +950,31 @@ def convert_concat(node, **kwargs):
     )
     return [concat_node]
 
+@mx_op.register("_zeros")
+@mx_op.register("_ones")
+@mx_op.register("_full")
+def convert_full(node, **kwargs):
+    """Map MXNet's _zeros, _ones and _full operators attributes to onnx's
+    tensors and return the created node.
+    """
+    # ToDo: Use Constant or ConstantOfShape, when Issue #15101 is resolved?
+    name, input_nodes, attrs = get_inputs(node, kwargs)
+    del input_nodes
+
+    # Convert "0"s dimensions to "1"s. This is a workaround for the case, where
+    # mxnet symbols can broadcast "0"s, while ONNX can only broadcast over "1"s
+    shape = convert_string_to_list(attrs["shape"])
+    shape = tuple(dim if dim else 1 for dim in shape)
+
+    value = {
+        '_zeros': 0.0,
+        '_ones': 1.0,
+        '_full': eval(attrs.get('value', '0')),
+    }[node['op']]
+    dtype = attrs.get('dtype')
+    data = np.full(shape, value, dtype)
+
+    return create_helper_tensor_node(data, name, kwargs)
 
 @mx_op.register("transpose")
 def convert_transpose(node, **kwargs):
--- a/python/mxnet/contrib/onnx/mx2onnx/_op_translations.py
+++ b/python/mxnet/contrib/onnx/mx2onnx/_op_translations.py
@@ -950,6 +950,73 @@ def convert_concat(node, **kwargs):
     )
     return [concat_node]
 
+@mx_op.register('_rnn_param_concat')
+def convert_rnn_param_concat(node, **kwargs):
+    """Map MXNet's _rnn_param_concat operator attributes to onnx's Concat
+    operator and return the created node.
+    """
+    name, input_nodes, attrs = get_inputs(node, kwargs)
+    axis = int(attrs.get("dim"))
+
+    # mxnet RNN node and ONNX RNN/LSTM/GRU nodes
+    # use different ways to store their parameters
+
+    # The conversion between these formats is broken into 2 steps
+    # The first step (performed here in _rnn_param_concat) regroups the
+    # flattened parameters according to the table below.
+    # The second step corrects the shapes and orders of gates and is
+    # performed and described in more detail in the RNN node
+
+    # mxnet            [ONNX] -> ONNX (group)
+    # i2h_weights [W (+  WB)] -> W    (input weights)
+    # h2h_weights [R (+  RB)] -> R    (recurrence weights)
+    # i2h_biases [Wb (+ WBb)] -> B = [Wb + Rb (+ WBb + RBb)]
+    # h2h_biases [Rb (+ RBb)] ->      (biases)
+
+    split = len(input_nodes) // 2
+    weights, biases = input_nodes[:split], input_nodes[split:]
+    i2h_weights = weights[::2]
+    h2h_weights = weights[1::2]
+    i2h_biases = biases[::2]
+    h2h_biases = biases[1::2]
+    reordered_biases = [
+        bias
+        for pair in zip(i2h_biases, h2h_biases)
+        for bias in pair
+    ]
+
+    # The order of mxnet parameters in the inputs is:
+    # [
+    #     '{}{}_{}_{}'.format(d, l, g, t)
+    #     for t in ['weight', 'bias']
+    #     for l in range(num_layers)
+    #     for d in ['l', 'r'][:num_directions]
+    #     for g in ['i2h', 'h2h']
+    # ]
+
+    w = onnx.helper.make_node(
+        "Concat",
+        inputs=i2h_weights,
+        outputs=[name + "__W"],
+        axis=axis,
+        name=name + "__W"
+    )
+    r = onnx.helper.make_node(
+        "Concat",
+        inputs=h2h_weights,
+        outputs=[name + "__R"],
+        axis=axis,
+        name=name + "__R"
+    )
+    b = onnx.helper.make_node(
+        "Concat",
+        inputs=reordered_biases,
+        outputs=[name + "__B"],
+        axis=axis,
+        name=name + "__B"
+    )
+    return [w, r, b]
+
 @mx_op.register("_zeros")
 @mx_op.register("_ones")
 @mx_op.register("_full")
--- a/python/mxnet/contrib/onnx/mx2onnx/_op_translations.py
+++ b/python/mxnet/contrib/onnx/mx2onnx/_op_translations.py
@@ -950,6 +950,220 @@ def convert_concat(node, **kwargs):
     )
     return [concat_node]
 
+@mx_op.register("RNN")
+def convert_RNN(node, **kwargs):
+    """Map MXNet's RNN operator attributes to onnx's RNN operator
+    and return the created node.
+    """
+    name, input_nodes, attrs = get_inputs(node, kwargs)
+    nodes = []
+
+    # ============================== Attributes ==============================
+    mode = attrs['mode'].upper()
+    rnn_kwargs = {}
+    if mode != 'LSTM':
+        raise NotImplementedError(
+            "Only LSTM mode RNN conversion to ONNX is currently supported."
+        )
+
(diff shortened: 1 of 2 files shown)


RuRo:
Some of the builds seem to be spuriously failing with weird errors like `Could not connect to mxnetlinux-cpu_.... to send interrupt signal to process` or `requests.exceptions.ReadTimeout: UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)`. Is there something wrong with the current build pipeline or is this somehow my fault?

leezu:
Unfortunately there are issues with CI.

RuRo:
@leezu @szha any feedback?

RuRo:
@anirudhacharya I haven't implemented any unittests yet, but the code **is** tested. I've provided an example test case in my original post, using ONNXRuntime as a reference implementation.

I've looked at how the unittests are implemented for other operators. The `backend_test.py` export test won't work in this case.

1) The current backend unittests don't actually work for nodes, which can **only** be exported from MXNet to ONNX. That is because the current 'strategy' to check exports is to [first import ONNX -> MXNet and then re-export it back to ONNX](https://github.com/apache/incubator-mxnet/blob/713d962359cfe5b014ca24c0f9b3676d86b28569/tests/python-pytest/onnx/backend.py#L104). Since we don't have an implementation for importing LSTMs from ONNX to MXNet, the *export*-only tests will fail, because the *import* for the operator LSTM is not implemented. Implementing an ONNX->MXNet conversion for the LSTM node is a non-trivial task.
2) This PR only implements a limited subset of the ONNX supported LSTM configurations and only a limited subset of MXNet supported RNN configurations. And currently, this conversion only supports Gluon LSTM (not arbitrary MXNet RNNs), so even if we could import the LSTM node from ONNX, the current implementation would fail on unsupported configurations. See "Comments:RNN" section in my original post.

I could add a test case to [test_node.py export_test_cases](https://github.com/apache/incubator-mxnet/blob/713d962359cfe5b014ca24c0f9b3676d86b28569/tests/python-pytest/onnx/test_node.py#L290), but this test only verifies, that a valid onnx graph is exported, not that the conversion is correct. Any kind of test, that involves actually forwarding the LSTM node will **require** implementing the import from ONNX (to use the official test) or using some reference implementation (an LSTM implementation written in numpy or onnxruntime etc).

RuRo:
The CI seems to be borked again btw 😩

anirudhacharya:
> I could add a test case to [test_node.py export_test_cases](https://github.com/apache/incubator-mxnet/blob/713d962359cfe5b014ca24c0f9b3676d86b28569/tests/python-pytest/onnx/test_node.py#L290), but this test only verifies, that a valid onnx graph is exported, not that the conversion is correct

Lets do this for now. we can use the tests from the onnx repo when the lstm import is also implemented.

RuRo:
@anirudhacharya Done

leezu:
thank you. restarted unix-gpu

RuRo:
@leezu did the restart not work or did it just fail again?

szha:
@RuRo It failed again and I restarted the failed pipeline just now. Thanks for the contribution.

For the CI, we will coordinate efforts to improve its status soon.

RuRo:
Nice. All green now.

RuRo:
@szha @Roshrini can you, please, review this PR or maybe nominate somebody else, if you're currently busy? Thanks

RuRo:
@mxnet-bot run ci [unix-cpu, unix-gpu]

RuRo:
@mxnet-bot run ci [unix-gpu]

RuRo:
@mxnet-bot run ci [unix-gpu]

leezu:
Jenkins CI successfully triggered : [centos-gpu]

RuRo:
@szha @Roshrini PTAL

RuRo:
> Feel free to ping me or other committers if a PR that's passing CI check needs attention.

@szha ping )

The CI is passing except for `codecov/project`, which I think claims that the coverage decreased in the following files (?)
```
Changes in src/operator/tensor/dot-inl.h    -2 → +2
Changes in src/io/iter_image_recordio_2.cc  -1 → +1
```
This PR doesn't change anything in `src` so it's not obvious to me, what is wrong here.

I am not too familiar with how `codecov` works so I may be wrong, but I think that this is a false positive.

leezu:
@Ruro codecov is mistakenly displayed on a couple of PRs. Due to instability of the CI, the codecov is not stable and you can ignore it. It won't show up on new PRs again

leezu:
There was some undetected conflict and this broke the tests on at least macOS https://github.com/apache/incubator-mxnet/issues/18596

PR #17749, opened 2020-03-03
Fix races in block scope

### LINKED ISSUE #17612: Gluon block thread safety bug
```
import mxnet as mx
import threading
import queue
import gluonnlp as nlp

q = queue.Queue()
count = 2
net = mx.gluon.nn.Dense((10))
net.initialize(mx.init.Uniform(), ctx=[mx.gpu(0), mx.gpu(1)])
mx.nd.waitall()

def fn(net, ctx):
    with mx.autograd.record():
        x = mx.nd.random.uniform(low=0, high=1000, shape=(10, 128), ctx=ctx)
        q.put(net(x))
    mx.nd.waitall()
    return

threads = []
for i in range(count):
    t = threading.Thread(target=fn, args=(net, mx.gpu()))
    threads.append(t)
for i in range(count):
    threads[i].start()
for i in range(count):
    threads[i].join()

ys = []
for i in range(count):
    ys.append(q.get())
print(ys)
```

Error: 
```
Exception in thread Thread-2:
  File "/home/ubuntu/src/mxnet/python/mxnet/gluon/block.py", line 1154, in forward
    params = {k: v.data(ctx) for k, v in self._reg_params.items()}
  File "/home/ubuntu/src/mxnet/python/mxnet/gluon/block.py", line 1154, in <dictcomp>
    params = {k: v.data(ctx) for k, v in self._reg_params.items()}
  File "/home/ubuntu/src/mxnet/python/mxnet/gluon/parameter.py", line 565, in data
    return self._check_and_get(self._data, ctx)
  File "/home/ubuntu/src/mxnet/python/mxnet/gluon/parameter.py", line 236, in _check_and_get
    "num_features, etc., for network layers."%(self.name))
mxnet.gluon.parameter.DeferredInitializationError: Parameter 'dense0_weight' has not been initialized yet because initialization was deferred. Actual initialization happens during the first forward pass. Please pass one batch of data through the network before accessing Parameters. You can also avoid deferred initialization by specifying in_units, num_features, etc., for network layers.

During handling of the above exception, another exception occurred:

  File "/home/ubuntu/src/mxnet/python/mxnet/gluon/block.py", line 976, in _deferred_infer_shape
    self.infer_shape(*args)
  File "/home/ubuntu/src/mxnet/python/mxnet/gluon/block.py", line 1074, in infer_shape
    self._infer_attrs('infer_shape', 'shape', *args)
  File "/home/ubuntu/src/mxnet/python/mxnet/gluon/block.py", line 1058, in _infer_attrs
    inputs, out = self._get_graph(*args)
  File "/home/ubuntu/src/mxnet/python/mxnet/gluon/block.py", line 928, in _get_graph
    out = self.hybrid_forward(symbol, *grouped_inputs, **params)  # pylint: disable=no-value-for-parameter
  File "/home/ubuntu/src/mxnet/python/mxnet/gluon/block.py", line 94, in __exit__
    self._name_scope.__exit__(ptype, value, trace)
AttributeError: 'NoneType' object has no attribute '__exit__'

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.5/threading.py", line 862, in run
    self._target(*self._args, **self._kwargs)
  File "test.py", line 29, in fn
    q.put(net(x))
  File "/home/ubuntu/src/mxnet/python/mxnet/gluon/block.py", line 696, in __call__
    out = self.forward(*args)
  File "/home/ubuntu/src/mxnet/python/mxnet/gluon/block.py", line 1156, in forward
    self._deferred_infer_shape(x, *args)
  File "/home/ubuntu/src/mxnet/python/mxnet/gluon/block.py", line 980, in _deferred_infer_shape
    raise ValueError(error_msg)
ValueError: Deferred initialization failed because shape cannot be inferred. 'NoneType' object has no attribute '__exit__'
```

hzfan:
## Description ##
Fix races in block scope by
- make _BlockScope thread-local

Thank @eric-haibin-lin  for guidance :)


hzfan:
@eric-haibin-lin Tests added. Seems that ctx is not relevant to this issue, and I just test it on cpu.

python/mxnet/gluon/block.py
```diff
@@ -48,8 +48,9 @@ class _BlockScope(object):
     def __init__(self, block):
         self._block = block
         self._counter = {}
-        self._old_scope = None
-        self._name_scope = None
+        self._local = threading.local()
```
leezu: How about using contextvars here? This will ensure the code is compatible with python coroutines.
https://docs.python.org/3/library/contextvars.html

You need to include https://pypi.org/project/contextvars/ as dependency on Python 3.5, 3.6

hzfan: Thanks for sharing this. Do we have a plan to switch to contextvars in the near future? This pr focuses on the _BlockScope issue. Maybe we can switch all threading.local() to contextvars in another pr.

leezu: It's a bug. MXNet will currently experience undefined behaviour when coroutines are used. There's no immediate plan to fix it, as it requires some backend work as well. However, new code should ideally be written in a correct way to reduce future efforts.

hzfan: Considering that additional dependency on contextvars needs to be introduced for py3.5, a separate pr for contextvars seems better, which adds the dependency, and switch all thread local to contextvars. I may still go with thread local here, but I’ll be glad to switch to contextvars after we add the dependency.

python/mxnet/gluon/block.py
```diff
@@ -91,22 +92,22 @@ def create(prefix, params, hint):
     def __enter__(self):
         if self._block._empty_prefix:
             return self
-        self._old_scope = getattr(_BlockScope._current, "value", None)
+        self._local._old_scope = getattr(_BlockScope._current, "value", None)
         _BlockScope._current.value = self
```
eric-haibin-lin, line 96: Shall we make `_BlockScope._current` also thread-local for consistency?

hzfan: Seems that it has been thread-local ?
https://github.com/apache/incubator-mxnet/blob/f3e8dc14de881a5b8ccc3ff5985167f8d6fa2fb8/python/mxnet/gluon/block.py#L47

chinakook:
This commit may cause rcnn/mask-rcnn cannot be trained.
https://github.com/dmlc/gluon-cv/blob/3c4150a964c776e4f7da0eb30b55ab05b7554c8d/gluoncv/data/transforms/presets/rcnn.py#L162

leezu:
@chinakook gluoncv does not support MXNet 2 development version at this point. You can use MXNet 1

chinakook:
It's also a problem with Tensorflow and Keras, and there are lots of work arounds.
https://github.com/tensorflow/tensorflow/issues/36897

PR #17760, opened 2020-03-04
Optimizing tests in test_loss.py

sleepy-owl:
## Description ##
Changing some parameters in test_loss to reduce the running time of tests

### Essentials ###
- [ ] Changes are complete 

### Changes ###
- [ ] Updated `test_bce_loss_with_pos_weight` and `test_ctc_loss_train`. Reduces the run time of each test from >2 seconds to less than 1.3 seconds locally. Tests pass with multiple seeds (>30).

## Comments ##
- There are multiple opportunities for reducing the running time of tests by tuning some parameters like iterations, learning rate, etc.
- For these changes, I verified that the tests still pass with multiple seeds
- Would you guys be interested in such optimizations? If yes, I can help contribute more and look at optimizing other *more expensive* tests in the test suite. Please let me know if you have any other suggestions or any necessary checks that I should do on these tests.

Thanks!

### DIFF
--- a/tests/python/unittest/test_loss.py
+++ b/tests/python/unittest/test_loss.py
@@ -221,7 +221,7 @@ def test_ctc_loss_train():
     loss = Loss(output, l)
     loss = mx.sym.make_loss(loss)
     mod = mx.mod.Module(loss, data_names=('data',), label_names=('label',))
-    mod.fit(data_iter, num_epoch=200, optimizer_params={'learning_rate': 0.01},
+    mod.fit(data_iter, num_epoch=10, optimizer_params={'learning_rate': 0.0256},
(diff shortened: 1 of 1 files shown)


szha:
Thanks! Could you use MXNET_TEST_COUNT=10000 environment variable to run the tests and see if it passes reliably?

sleepy-owl:
Sure, I can check that. Does this check the test with different seeds? Also, what is the expected output for this?

szha:
@sleepy-owl thanks. It will run repeatedly with different seeds for 10000 times. The expected result is that the test passes successfully.

szha:
@sleepy-owl did you get a chance to verify the change?

PR #17762, opened 2020-03-04
Custom Operator Random Number Generator Support

rondogency:
## Description ##
Add random number generator support for custom operator libraries. 

### Design ###
We pass from MXNet the initialized and seeded states, located on CPU and GPU, to custom library. So user could use those seeds to generate deterministic values from a given seed passed to MXNet. Basically this workflow:

mx.random.seed(128)
r1 = mx.nd.some_custom_random_op(data)
mx.random.seed(128)
r2 = mx.nd.some_custom_random_op(data)
assert (r1 == r2)

### This PR is not ###
Let custom library generate exactly the same sequence of random numbers comparing to MXNet

## Comments ##
This is a continuation of the custom operator project #15921 and #17270


samskalicky:
Thanks for this @rondogency! this is going to be a great addition for customOps.

A couple suggestions on the PR description:

Add random number generator support for custom operator**s** ~libraries~.

Custom ~library~ **operators** now can call MXNet random number generator to get random numbers through OpResource class in forward/backward computation.

include/mxnet/lib_api.h
```diff
@@ -39,7 +39,8 @@
 #include <utility>
 #include <stdexcept>
 
-#define MX_LIBRARY_VERSION 3
+/* Make sure to update the version number everytime you make changes */
+#define MX_LIBRARY_VERSION 4
```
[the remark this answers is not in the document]
rondogency: I know that's why I put that note and will resolve conflict when merging

src/c_api/c_api.cc
```diff
@@ -177,10 +177,39 @@ void CustomFComputeDispatcher(const std::string op_name,
     return static_cast<void*>((*gpualloc)(size));
   };
 
+  // MXNet random number generator lambda to be invoked by custom library
+  // custom library will use gpu rng whenever possible
+  auto rng_caller = [&](RandGenType rand_type, int seed) {
+#if MXNET_USE_CUDA
```
wkcn: Some users may use MXNet built with `MXNET_USE_CUDA=1` on the machine with only CPU. In this case, the GPU random number generator is not available.

It is better to decide which device to choose in the user-side code, namely the code of custom operators.

rondogency: thanks for the timely review. we can let user pass in a "cpu" or "gpu" string to decide it, just like the forward/backward function

Also I am just wondering if we should make a default choice of device?

samskalicky:
> Great. how are we testing this feature?

I think @rondogency's plan is to write a dropout operator

src/c_api/c_api.cc
wkcn: Agree to making a default choice : )

I think it should depend on the context of custom operator, rather that the version (GPU or only CPU) of MXNet. 

If users build a CPU custom operator and use the GPU version of MXNet, on a machine with only CPU, it still work since the CPU random number generator is used.

samskalicky: I dont think we should have a default, user should be explicit which one they want to use. otherwise we introduce possible confusion

rondogency:
in the cmake, the line "target_compile_options(subgraph_lib PUBLIC -shared)" can be removed, as we added the library as shared already

example/extensions/lib_custom_op/relu_lib.cu
```diff
@@ -180,6 +182,75 @@ REGISTER_OP(my_state_relu)
...
+/* ------------------------ Below is noisy relu operator example ---------------------*/
+
+#include <curand_kernel.h>
+#include <random>
+
+#define NumRandomPerThread 64 // mxnet recommended random numbers generated per thread
+
+__global__ void noisy_relu_gpu_forward(float *out, float *in, int64_t N, void *states, int step) {
+    int tid = blockIdx.x * blockDim.x + threadIdx.x;
+    curandStatePhilox4_32_10_t* global_states = (curandStatePhilox4_32_10_t*)states;
```
samskalicky: use static_cast<curandStatePhilox4_32_10_t*>(states)

example/extensions/lib_custom_op/relu_lib.cu
```diff
@@ -180,6 +182,75 @@ REGISTER_OP(my_state_relu)
...
+    curandStatePhilox4_32_10_t thread_state = global_states[tid];
+
+    int start = tid * step;
+    int end = start + step;
+    for (int i=start; i<end && i<N; ++i) {
+        float noise = curand_normal(&thread_state);
+        out[i] = in[i] + noise > 0 ? in[i] + noise : 0;
```
samskalicky: can you add some comments to explain how/why you're calculating this for noisy relu?

example/extensions/lib_custom_op/relu_lib.cu
samskalicky: can you add a comment to explain why you're doing this? where do the states get allocated? how can you guarantee the states pointer is pointing to a big enough array to index with tid?

example/extensions/lib_custom_op/test_relu.py
```diff
@@ -63,22 +63,33 @@
...
 t2 = time.time()
-r2 = mx.nd.my_relu(b)
+r2 = mx.nd.relu(b)
 mx.nd.waitall()
 t3 = time.time()
-r3 = mx.nd.relu(b)
-mx.nd.waitall()
-t4 = time.time()
-print("CPU running time:")
+print("Custom GPU running time:")
```
samskalicky: Custom GPU ==> Custom relu

example/extensions/lib_custom_op/test_relu.py
```diff
@@ -63,22 +63,33 @@
...
 print(t2 - t1)
-print("GPU running time:")
+print("Native GPU running time:")
```
samskalicky: Native GPU ==> Native relu

include/mxnet/lib_api.h
```diff
@@ -437,10 +438,12 @@ class OpResource {
  public:
   OpResource(xpu_malloc_t cpu_malloc_fp, void* cpu_alloc_fp,
              xpu_malloc_t gpu_malloc_fp, void* gpu_alloc_fp, void* stream,
-             sparse_malloc_t sparse_malloc_fp, void* sparse_alloc_fp)
+             sparse_malloc_t sparse_malloc_fp, void* sparse_alloc_fp,
+             void* cpu_states, void* gpu_states)
```
samskalicky: can we call these cpu_rand_states and gpu_rand_states so that its clear that these are for the RNG?

include/mxnet/lib_api.h
```diff
@@ -463,6 +466,19 @@ class OpResource {
                    &(sparse->data), &(sparse->indices), &(sparse->indptr));
   }
 
+  /*! \brief get pointer to cpu random number inited and seeded states */
+  /* this global states are located on cpu, of type std::mt19937 */
+  void* get_cpu_rand_states() {
+    return cpu_rand_states;
```
samskalicky: is there any reason not to return already casted to std::mt19937?

include/mxnet/lib_api.h
```diff
@@ -463,6 +466,19 @@ class OpResource {
...
+  }
+
+  /*! \brief get pointer to gpu random number inited and seeded states */
+  /* this global states are located on gpu, of type curandStatePhilox4_32_10_t */
+  /* note that if you are usign cpu build, then it will return a nullptr */
+  void* get_gpu_rand_states() {
+    return gpu_rand_states;
```
samskalicky: is there any reason not to return already casted to curandStatePhilox4_32_10_t?

include/mxnet/lib_api.h
```diff
@@ -997,7 +1015,8 @@ typedef int (*opCallFComp_t)(fcomp_t fcomp, const char* const* keys,
                              void** in_indices, void** out_indices,
                              void** in_indptr, void** out_indptr,
                              int64_t* in_indices_shapes, int64_t* out_indices_shapes,
-                             int64_t* in_indptr_shapes, int64_t* out_indptr_shapes);
+                             int64_t* in_indptr_shapes, int64_t* out_indptr_shapes,
+                             void* cpu_states, void* gpu_states);
```
samskalicky: can we call these rand_ cpu_states and rand_gpu_states so its more clear that these are for RNG?

src/c_api/c_api.cc
```diff
@@ -248,11 +249,22 @@ void CustomFComputeDispatcher(const std::string op_name,
   // get actual cudaStream_t out of mxnet gpu stream and pass to lib_api.h
   void *cuda_stream = nullptr;
 #if MXNET_USE_CUDA
-  if (inputs[0].ctx().dev_mask() == Context::kGPU) {
+  if (inputs.size() > 0 && inputs[0].ctx().dev_mask() == Context::kGPU) {
     cuda_stream = static_cast<void*>(gpu_stream->stream_);
   }
 #endif
 
+  // get mxnet inited and seeded rng states and pass to lib_api.h
```
samskalicky: inited ==> initialized

example/extensions/lib_custom_op/relu_lib.cu
rondogency: Added comments with details

example/extensions/lib_custom_op/relu_lib.cu
rondogency: Added to comments to explain the logic, and made some type tweaks

rondogency:
@samskalicky resolved your comments, please take another look

rondogency:
@wkcn @eric-haibin-lin this PR is ready for review. I have to reimplement the workflow to get rid of some segfault error so the code is different from last time you reviewed

rondogency:
@sxjscience @yzhliu  I will appreciate if you can take a look at this PR and point out if any random number related use cases that need to be covered or tested

include/mxnet/lib_api.h
```diff
@@ -426,21 +432,30 @@ typedef void (*sparse_malloc_t)(void*, int, int, int, void**, int64_t**, int64_t
...
+  typedef curandStatePhilox4_32_10_t* mx_gpu_rand_pt;
 #else
   typedef void* mx_stream_t;
+  typedef void* mx_gpu_rand_pt;
 #endif
 
+/*! \brief MXNet initialized random states for each device, used for parallelism */
+/* Each thread should generate random number unique sequence out of different states */
+#define NumCPURandomStates 1024
+#define NumGPURandomStates 32768
```
samskalicky: Can we prefix these with MX_ and use all caps?
MX_CPU_RANDOM_STATES
MX_GPU_RANDOM_STATES

include/mxnet/lib_api.h
```diff
@@ -426,21 +432,30 @@ typedef void (*sparse_malloc_t)(void*, int, int, int, void**, int64_t**, int64_t
 
 #if defined(__NVCC__)
   typedef cudaStream_t mx_stream_t;
+  typedef curandStatePhilox4_32_10_t* mx_gpu_rand_pt;
```
samskalicky: since this is a typedef, lets follow the naming style and call it mx_gpu_rand_t

should we typedef std::mt19937 to mx_cpu_rand_t too? or is that going too far?

include/mxnet/lib_api.h
samskalicky: its still called cpu_states here, can we rename it here too?

include/mxnet/lib_api.h
```diff
@@ -1026,7 +1057,8 @@ typedef int (*opCallFStatefulComp_t)(int is_forward, void* state_op,
                                      void** in_indices, void** out_indices,
                                      void** in_indptr, void** out_indptr,
                                      int64_t* in_indices_shapes, int64_t* out_indices_shapes,
-                                     int64_t* in_indptr_shapes, int64_t* out_indptr_shapes);
+                                     int64_t* in_indptr_shapes, int64_t* out_indptr_shapes,
+                                     void* cpu_states, void* gpu_states);
```
samskalicky: can we rename it here too?

include/mxnet/lib_api.h
```diff
@@ -1284,7 +1316,8 @@ extern "C" {
                   int* instypes, int* outstypes, void** in_indices, void** out_indices,
                   void** in_indptr, void** out_indptr,
                   int64_t* in_indices_shapes, int64_t* out_indices_shapes,
-                  int64_t* in_indptr_shapes, int64_t* out_indptr_shapes) {
+                  int64_t* in_indptr_shapes, int64_t* out_indptr_shapes,
+                  void* cpu_states, void* gpu_states) {
```
samskalicky: can we rename it here too?

include/mxnet/lib_api.h
```diff
@@ -1345,7 +1378,7 @@ extern "C" {
     }
 
     OpResource res(cpu_malloc, cpu_alloc, gpu_malloc, gpu_alloc,
-                   cuda_stream, sparse_malloc, sparse_alloc);
+                   cuda_stream, sparse_malloc, sparse_alloc, cpu_states, gpu_states);
```
samskalicky: can we rename it here too?

Just dont want "states" to be confused with stateful ops

include/mxnet/lib_api.h
```diff
@@ -1419,7 +1452,8 @@ extern "C" {
                           int* instypes, int* outstypes, void** in_indices, void** out_indices,
                           void** in_indptr, void** out_indptr,
                           int64_t* in_indices_shapes, int64_t* out_indices_shapes,
-                          int64_t* in_indptr_shapes, int64_t* out_indptr_shapes) {
+                          int64_t* in_indptr_shapes, int64_t* out_indptr_shapes,
+                          void* cpu_states, void* gpu_states) {
```
samskalicky: can we rename it here too?

include/mxnet/lib_api.h
```diff
@@ -1476,7 +1510,7 @@ extern "C" {
     }
 
     OpResource res(cpu_malloc, cpu_alloc, gpu_malloc, gpu_alloc,
-                   stream, sparse_malloc, sparse_alloc);
+                   stream, sparse_malloc, sparse_alloc, cpu_states, gpu_states);
```
samskalicky: can we rename it here too?

src/c_api/c_api.cc
```diff
@@ -248,11 +251,24 @@ void CustomFComputeDispatcher(const std::string op_name,
...
 #if MXNET_USE_CUDA
-  if (inputs[0].ctx().dev_mask() == Context::kGPU) {
+  if (inputs.size() > 0 && inputs[0].ctx().dev_mask() == Context::kGPU) {
     cuda_stream = static_cast<void*>(gpu_stream->stream_);
   }
 #endif
 
+  // get mxnet initialized and seeded RNG states and pass to lib_api.h
+  void *rng_cpu_states = nullptr, *rng_gpu_states = nullptr;
+  if (!is_subgraph_op) {
```
samskalicky: why are we only doing this for non-subgraph ops? Subgraph ops cant get RNG?

example/extensions/lib_custom_op/relu_lib.cu
```diff
@@ -180,6 +182,80 @@ REGISTER_OP(my_state_relu)
...
+    int64_t N = inputs[0].size();
+
+    // below is mxnet recommended workflow to parallel random number generating
+    int num_thread = (N + NumRandomPerThread - 1) / NumRandomPerThread;
+    // we should not launch more threads than mxnet supported random number GPU states
+    int num_thread_need = num_thread < NumGPURandomStates ? num_thread : NumGPURandomStates;
+    // each cuda thread processes [step * tid, step * id + step) snippet of input tensor
+    int step = (N + num_thread_need - 1) / num_thread_need;
+    // this can ensure number of parallel threads less than mxnet supported random number states
+    int num_block = (num_thread_need + NumThreadPerBlock - 1) / NumThreadPerBlock;
```
samskalicky: this is great! will be super helpful for users.

src/c_api/c_api.cc
rondogency: subgraph op is using DefaultSubgraphOpResourceRequest, so if no operator in the subgraph resource request pass has parallel random, resource request will error out

but I think I should check initialized resource types instead of check subgraph op, I will make the change

samskalicky: DefaultSubgraphOpResourceRequest calls DefaultSubgraphOpResourceRequestHelper which goes through the ops in the subgraph and sets all resources requested by any op in the subgraph. One of these could be a custom op who's defining the resource request. 

You have a good point though, we should probably just have subgraph ops define all supported resources anyway.

rondogency:
@samskalicky resolved all your comments

src/c_api/c_api.cc
```diff
@@ -248,11 +250,21 @@ void CustomFComputeDispatcher(const std::string op_name,
   // get actual cudaStream_t out of mxnet gpu stream and pass to lib_api.h
   void *cuda_stream = nullptr;
 #if MXNET_USE_CUDA
-  if (inputs[0].ctx().dev_mask() == Context::kGPU) {
+  if (inputs.size() > 0 && inputs[0].ctx().dev_mask() == Context::kGPU) {
```
samskalicky: while you're making this change, its possible an operator might have zero inputs and some outputs (ie. generator operator, like random), should we also check for outputs if inputs is zero like:
```
if ((inputs.size() > 0 && inputs[0].ctx().dev_mask() == Context::kGPU) ||
    (outputs.size() > 0 && inputs[0].ctx().dev_mask() == Context::kGPU)) {
```

src/c_api/c_api.cc
```diff
@@ -809,11 +822,10 @@ int MXLoadLib(const char *path) {
       regOp.set_attr<mxnet::FInferShape>("FInferShape", DefaultSubgraphOpShape, plevel);
       regOp.set_attr<FInferStorageType>("FInferStorageType",
                                         DefaultSubgraphOpStorageType, plevel);
-      regOp.set_attr<FResourceRequest>("FResourceRequest",
-                                       DefaultSubgraphOpResourceRequest, plevel);
       regOp.set_attr<nnvm::FMutateInputs>("FMutateInputs",
                                           DefaultSubgraphOpMutableInputs, plevel);
     }
+    regOp.set_attr<FResourceRequest>("FResourceRequest", resc_req, plevel);
```
samskalicky: can we put this after line 798 instead? its kinda weird hanging out here in the middle

samskalicky:

src/c_api/c_api.cc
rondogency: put it to 806 for readability

src/c_api/c_api.cc
rondogency: I think we just check for output

wkcn:
@mxnet-bot run ci [unix-gpu, centos-cpu]

wkcn:
@mxnet-bot run ci [unix-gpu]

rondogency:
@mxnet-bot run ci [unix-gpu]

rondogency:
@mxnet-bot run ci [unix-gpu]

rondogency:
@mxnet-bot run ci [unix-gpu, windows-gpu]

PR #17777, opened 2020-03-05
Add alias for np.__version__ and np.dtype

haojin2:
## Description ##
As title.

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/python/mxnet/numpy/fallback.py
+++ b/python/mxnet/numpy/fallback.py
@@ -20,7 +20,10 @@
 
 import numpy as onp
 
+
 __all__ = [
+    '__version__',
+    '_NoValue',
     'allclose',
     'alltrue',
     'apply_along_axis',
@@ -37,6 +40,7 @@ __all__ = [
     'cov',
     'digitize',
     'divmod',
+    'dtype',
     'extract',
     'float_power',
     'frexp',
@@ -106,6 +110,8 @@ __all__ = [
     'vander',
 ]
 
+__version__ = onp.__version__
+_NoValue = onp._NoValue
 allclose = onp.allclose
 alltrue = onp.alltrue
 apply_along_axis = onp.apply_along_axis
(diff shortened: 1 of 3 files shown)


python/mxnet/numpy/fallback.py
```diff
@@ -106,6 +110,8 @@
     'vander',
 ]
 
+__version__ = onp.__version__
```
leezu: Is there any mechanism to ensure `mx.np` is compatible with the `onp.__version__`? If not, should this return the version `mx.np` is compatible with?

haojin2: At least for now the CI have 1.18 installed and we pass all current tests against that version. So, I think we're in good shape here.

leezu: That's great. My question is, once Numpy 1.19 or later versions are released and potentially change API, should `mx.np.__version__` report `1.18`, which is the version you are currently targeting, or should it claim `1.19` which could be the numpy version installed on the users machine.

haojin2: Once a 1.19 comes out our CI would directly report problems if any and we can fix it, I cannot predict what they're changing in the future, and the dilemma here is that the fallback-ed operators will follow the new 1.19.

leezu: Thanks for clarifying the issue. How should this be handled for stable releases? Users may install the release at any time and the numpy version could be incompatible? Should the supported numpy version be explicitly called out and returned by `mx.np.__version__`?

haojin2: I think it should be addressed by setting the correct dependencies in our release version. If we claim that we are compatible with 1.xx version but the pypi packages says that anything up to 1.yy is okay, then I think it's something that whoever is in charge of CD should address.

PR #17779, opened 2020-03-06
[Numpy] FFI Invocation for Unary Ops

hgt312:
| Old FFI (ctypes) (us) | New FFI (cython) (us)
-- | -- | --
unary ops | ~14 | ~7

Using 2x2 tensor as input.

### DIFF
--- a/python/mxnet/ndarray/numpy/_op.py
+++ b/python/mxnet/ndarray/numpy/_op.py
@@ -2076,7 +2076,7 @@ def _unary_func_helper(x, fn_array, fn_scalar, out=None, **kwargs):
     if isinstance(x, numeric_types):
         return fn_scalar(x, **kwargs)
     elif isinstance(x, NDArray):
-        return fn_array(x, out=out, **kwargs)
+        return fn_array(x, out)
     else:
         raise TypeError('type {} not supported'.format(str(type(x))))
 
@@ -2113,7 +2113,7 @@ def sin(x, out=None, **kwargs):
     >>> np.sin(np.array((0., 30., 45., 60., 90.)) * np.pi / 180.)
     array([0.        , 0.5       , 0.70710677, 0.86602545, 1.        ])
     """
-    return _unary_func_helper(x, _npi.sin, _np.sin, out=out, **kwargs)
+    return _unary_func_helper(x, _api_internal.sin, _np.sin, out=out, **kwargs)
 
 
 @set_module('mxnet.ndarray.numpy')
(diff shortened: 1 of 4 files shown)


src/api/operator/ufunc_helper.cc
```diff
@@ -107,4 +107,22 @@ void UFuncHelper(runtime::MXNetArgs args,
...
+                 runtime::MXNetRetValue* ret,
+                 const nnvm::Op* op) {
+  using namespace runtime;
+  nnvm::NodeAttrs attrs;
+  attrs.op = op;
+  NDArray* inputs[] = {args[0].operator NDArray*()};
+  NDArray* out = args[1].operator NDArray*();
+  NDArray** outputs = out == nullptr ? nullptr : &out;
+  int num_outputs = out != nullptr;
+  auto ndoutputs = Invoke(op, &attrs, 1, inputs, &num_outputs, outputs);
```
hzfan: `1` seems like a magic number. Use `int num_inputs = 1` here?

python/mxnet/ndarray/numpy/_op.py
```diff
@@ -1933,7 +1933,7 @@ def _unary_func_helper(x, fn_array, fn_scalar, out=None, **kwargs):
     if isinstance(x, numeric_types):
         return fn_scalar(x, **kwargs)
     elif isinstance(x, NDArray):
-        return fn_array(x, out=out, **kwargs)
```
hzfan: Since legacy operators make heavy use of `_unary_func_helper`, what about leave it as it was? We may have a new helper function for `_api_internal`, which may support `*args` instead of `**kwargs`

hgt312: In fact, these simple unary ops do not support any `kwargs` in MXNet, `kwargs` only need to make scalar's result right and has uniform error handling by using `@wrap_np_unary_func`. So, move `**kwargs` here is a very simple solution.

hzfan: My concern is that in the future, people may want to add other unary operators with kwargs (like tile) with legacy ffi. They may not notice that `**kwargs` are no longer supported by `_unary_func_helper`.

hgt312:
@mxnet-bot run ci [job1, job2, job3]

hgt312:
@mxnet-bot run ci [centos-gpu, unix-cpu]

hgt312:
@mxnet-bot run ci [centos-gpu, unix-cpu]

hgt312:
@mxnet-bot run ci [centos-cpu, centos-gpu, unix-gpu, windows-gpu]

hgt312:

hgt312:

hgt312:
@mxnet-bot run ci [centos-gpu, unix-gpu, unix-cpu, windows-gpu]

PR #17797, opened 2020-03-09
Skip test_multi_worker_dataloader_release_pool

haojin2:
## Description ##
...to unblock many PR on the CI.

### Essentials ###

### Changes ###
- [ ] Skip `test_multi_worker_dataloader_release_pool`

### DIFF
--- a/tests/python/unittest/test_gluon_data.py
+++ b/tests/python/unittest/test_gluon_data.py
@@ -269,6 +269,12 @@ def test_multi_worker_dataloader_release_pool():
     if os.name == 'nt':
         print('Skip for windows since spawn on windows is too expensive.')
         return
+
+    from sys import platform
+    if platform == 'darwin':
+        print('Skip for MacOS due to flaky failures')
+        return
+
     for _ in range(10):
         A = np.random.rand(999, 2000)
         D = mx.gluon.data.DataLoader(A, batch_size=8, num_workers=8)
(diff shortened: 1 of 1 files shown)


leezu:
Does the test fail on any other platform besides OS X? Currently OS X status test is optional, so failing OS X tests doesn't block PRs

haojin2:
@leezu I still think it's still necessary to skip this for now in case it masks the failures of other tests that comes after this test in the test sequence.

leezu:
Thanks for clarification. Let's disable it. I'm still at loss why this only happens on OS X.
Would be great if anyone with a Mac could help debug this issue.

zhreshold:
I think you can check for mac explicitly in test and disable it rather than skip for all platforms

haojin2:
@zhreshold Good idea, I'll make changes in that way.

marcoabreu:
Can we add a tracking issue in this PR as well as in the code?

tests/python/unittest/test_gluon_data.py
```diff
@@ -269,6 +269,12 @@ def test_multi_worker_dataloader_release_pool():
     if os.name == 'nt':
         print('Skip for windows since spawn on windows is too expensive.')
         return
+
+    from sys import platform
+    if platform == 'darwin':
+        print('Skip for MacOS due to flaky failures')
```
leezu:
```suggestion
        print('Skip for MacOS due to https://github.com/apache/incubator-mxnet/issues/17782')
```

leezu: @haojin2 can you merge the suggested change or something similar to address Marco's concern?

PR #17799, opened 2020-03-10
optimization debian package manager tweaks

Rajpratik71:
By default, Ubuntu or Debian based "apt" or "apt-get" system installs recommended but not suggested packages .

By passing "--no-install-recommends" option, the user lets apt-get know not to consider recommended packages as a dependency to install.

This results in smaller downloads and installation of packages .

Refer to blog at [Ubuntu Blog](https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends) .

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

## Description ##

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/cd/python/docker/Dockerfile
+++ b/cd/python/docker/Dockerfile
@@ -25,7 +25,7 @@ FROM ${BASE_IMAGE}
 
 ARG PYTHON_CMD=python3
 RUN apt-get update && \
-    apt-get install -y wget ${PYTHON_CMD}-dev gcc && \
+    apt-get --no-install-recommends install -y wget ${PYTHON_CMD}-dev gcc && \
     wget https://bootstrap.pypa.io/get-pip.py && \
     ${PYTHON_CMD} get-pip.py
 
--- a/ci/README.md
+++ b/ci/README.md
@@ -33,14 +33,14 @@ following script:
 set -e
 set -x
 export DEBIAN_FRONTEND=noninteractive
-apt-get -y install curl
+apt-get --no-install-recommends install -y curl
 curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
 add-apt-repository \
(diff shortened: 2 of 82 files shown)


leezu:
Your assessment is correct, but doesn't take into account that our existing setup may rely on the presence of the recommended packages. Thus just adding `--no-install-recommends` everywhere may not work. Given the CI failing, that seems to be the case.

It's fine to merge this PR for the CI related Dockerfiles, once CI passes. I'm hesitant to merge such changes to Readme's and other files that are not tested by CI. Not only is there a risk that the change will break the instructions, but also you are trading off "less packages to download" against "users need to type more text to install".

marcoabreu:
Agree, that seems like some kind of optimization which yields more harm than good in the worst case. It's traffic vs stability, where I prefer the latter

Rajpratik71:
> Your assessment is correct, but doesn't take into account that our existing setup may rely on the presence of the recommended packages. Thus just adding `--no-install-recommends` everywhere may not work. Given the CI failing, that seems to be the case.
> 
> It's fine to merge this PR for the CI related Dockerfiles, once CI passes. I'm hesitant to merge such changes to Readme's and other files that are not tested by CI. Not only is there a risk that the change will break the instructions, but also you are trading off "less packages to download" against "users need to type more text to install".

will work on it , look for where CI is failing , if it is related , will fix the issue and push the changes according

Rajpratik71:
> Agree, that seems like some kind of optimization which yields more harm than good in the worst case. It's traffic vs stability, where I prefer the latter

will make it stable

marcoabreu:
But I honestly don't see the benefit here, I'm sorry. Saving a few hundred megabytes on a system that processes terabytes per day doesn't seem like a good time investment to me. Furthermore it will create issues whenever people want to add new dependencies because nobody is used to that flag.

Issue #17635, opened 2020-02-20
Windows segmentation faults in GPU tests


larroy:
## Description

Windows GPU tests from the updated environment in https://github.com/aiengines/ci fails with the following:

```
======================================================================
ERROR: Failure: OSError (exception: access violation writing 0x0000000000000000)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Python37\lib\site-packages\nose\failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "C:\Python37\lib\site-packages\nose\loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "C:\Python37\lib\site-packages\nose\importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "C:\Python37\lib\site-packages\nose\importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "C:\Python37\lib\imp.py", line 235, in load_module
    return load_source(name, filename, file)
  File "C:\Python37\lib\imp.py", line 172, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 696, in _load
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "C:\Users\Administrator\mxnet\tests\python\unittest\test_test_utils.py", line 21, in <module>
    import mxnet as mx
  File "C:\Users\Administrator\mxnet\windows_package\python\mxnet\__init__.py", line 33, in <module>
    from . import contrib
  File "C:\Users\Administrator\mxnet\windows_package\python\mxnet\contrib\__init__.py", line 27, in <module>
    from . import autograd
  File "C:\Users\Administrator\mxnet\windows_package\python\mxnet\contrib\autograd.py", line 27, in <module>
    from ..ndarray import NDArray, zeros_like, _GRAD_REQ_MAP
  File "C:\Users\Administrator\mxnet\windows_package\python\mxnet\ndarray\__init__.py", line 20, in <module>
    from . import _internal, contrib, linalg, op, random, sparse, utils, image, ndarray, numpy
  File "C:\Users\Administrator\mxnet\windows_package\python\mxnet\ndarray\numpy\__init__.py", line 23, in <module>
    from . import _register
[...]

PR #17206, opened 2020-01-02
[WIP] Windows dev environment configuration, update install instructions from source in the docs.

larroy:
## Description ##
Update windows build from source documentation with automated scripts.

As part of fixing the windows CI AMI, the scripts to setup the build environment are referenced from the documentation for ease of use.

With this PR seting up the environment in windows consists of executing just a powershell script
```
.\setup.ps1
```
Together with `ci/build_windows.py` this makes 2 commands to build MXNet in a clean windows install.

Before the process to setup the required environment required tons of manual and error prone install and steps.

Chance Bair is also co-author of the installation script.

Verified that the updated instructions lead to a working MXNet build:

```
PS C:\Users\Administrator\incubator-mxnet> C:\Python37\python
Python 3.7.0 (v3.7.0:1bf9cc5093, Jun 27 2018, 04:59:51) [MSC v.1914 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import mxnet as mx
>>> a = mx.nd.array((3,3))
>>> a

[3. 3.]
<NDArray 2 @cpu(0)>
>>>
```

@aaronmarkham @leezu  @Chancebair

PR #17808, opened 2020-03-11
[WIP] Windows dev environment configuration, update install instructions from source in the docs

vexilligera:
## Description ##
Taking over #17206 with fix to #17635, update toolchains to vs2019 and cuda 10.2

### Essentials ###

### Changes ###

## Comments ##


ci/windows_dev_env/windows_deps_headless_installer.py
```diff
@@ -0,0 +1,429 @@
...
+ssl._create_default_https_context = ssl._create_unverified_context
+
+log = logging.getLogger(__name__)
+
+
+DEPS = {
+    'openblas': 'https://windows-post-install.s3-us-west-2.amazonaws.com/OpenBLAS-windows-v0_2_19.zip',
+    'opencv': 'https://windows-post-install.s3-us-west-2.amazonaws.com/opencv-windows-4.1.2-vc14_vc15.zip',
+    'cudnn': 'https://windows-post-install.s3-us-west-2.amazonaws.com/cudnn-9.2-windows10-x64-v7.4.2.24.zip',
+    'nvdriver': 'https://windows-post-install.s3-us-west-2.amazonaws.com/nvidia_display_drivers_398.75_server2016.zip',
```
leezu: Who owns the `windows-post-install` bucket?

ci/windows_dev_env/windows_deps_headless_installer.py
```diff
@@ -0,0 +1,429 @@
...
+
+def install_perl():
+    logging.info("Installing Perl")
+    with tempfile.TemporaryDirectory() as tmpdir:
+        perl_file_path = download(DEPS['perl'], tmpdir)
+        check_call(['msiexec ', '/n', '/passive', '/i', perl_file_path])
+    logging.info("Perl install complete")
+
+
+def install_clang():
```
leezu: Why do we need clang when compiling with visual studio?

ci/windows_dev_env/windows_deps_headless_installer.py
```diff
@@ -0,0 +1,429 @@
...
+            logging.info("Waiting for Visual Studio to install for the last {} seconds".format(str(timer)))
+        sleep(1)
+        timer += 1
+    if vs_still_installing():
+        logging.warning("VS install still running after timeout (%d)", DEFAULT_SUBPROCESS_TIMEOUT)
+    else:
+        logging.info("Visual studio install complete.")
+
+
+def install_cmake():
```
leezu: Unused

ci/windows_dev_env/windows_deps_headless_installer.py
```diff
@@ -0,0 +1,429 @@
...
+        with zipfile.ZipFile(local_file, 'r') as zip:
+            zip.extractall(tmpdir)
+        with remember_cwd():
+            os.chdir(tmpdir)
+            check_call(".\setup.exe -noreboot -clean -noeula -nofinish -passive")
+    logging.info("NVidia install complete")
+
+
+def install_cuda():
+    # CUDA 9.2 and patches
```
leezu: Enforcing Cuda 9.2 may not be very future proof for this script.

docs/static_site/src/pages/get_started/windows_setup.md
```diff
@@ -137,126 +137,37 @@ Check the chart below for other options or refer to [PyPI for other MXNet pip pa
...
-3. Verify that the `DCUDNN_INCLUDE` and `DCUDNN_LIBRARY` environment variables are pointing to the `include` folder and `cudnn.lib` file of your CUDA installed location, and `C:\incubator-mxnet` is the location of the source code you just cloned in the previous step.
-4. Create a build dir using the following command and go to the directory, for example:
-```
-mkdir C:\incubator-mxnet\build
-cd C:\incubator-mxnet\build
-```
-5. Compile the MXNet source code with `cmake` by using following command:
 ```
-cmake -G "Visual Studio 15 2017 Win64" -T cuda=9.2,host=x64 -DUSE_CUDA=1 -DUSE_CUDNN=1 -DUSE_NVRTC=1 -DUSE_OPENCV=1 -DUSE_OPENMP=1 -DUSE_BLAS=open -DUSE_LAPACK=1 -DUSE_DIST_KVSTORE=0 -DCUDA_ARCH_LIST=Common -DCUDA_TOOLSET=9.2 -DCUDNN_INCLUDE=C:\cuda\include -DCUDNN_LIBRARY=C:\cuda\lib\x64\cudnn.lib "C:\incubator-mxnet"
+.\setup.ps1
```
leezu: Is it feasible to provide short non-blackbox instructions to install MXNet?

For example, if users have a standard installation of visual studio, can the installation experience be pretty similar to https://mxnet.apache.org/get_started/ubuntu_setup#build-the-mxnet-shared-library-from-source Step 2 and Step 3?

aaronmarkham:
Let me know when this is ready, and I'll test it out.

leezu:
@vexilligera what is the status of this PR?

vexilligera:
> @vexilligera what is the status of this PR?

I'm trying to update the AMI to VS2019 and CUDA 10.2 and there's a bug with dmlc-core, probably caused by different MSVC standards. I'm working on that.

docs/static_site/src/pages/get_started/windows_setup.md
vexilligera: Yes, we can explicitly use the cmd command invoked from ci/build_windows.py

ci/windows_dev_env/windows_deps_headless_installer.py
vexilligera: I currently don't have access to windows-post-install, so I opened another bucket (mxnet-windows-build). It would be easy to merge them.

ci/windows_dev_env/windows_deps_headless_installer.py
vexilligera: They probably need to support tvm op build on windows as well, which may need this.

ci/windows_dev_env/windows_deps_headless_installer.py
vexilligera: CUDA is updated to 10.2 now.

vexilligera:
> Making it public is redistribution and thus license infringement.

I may also need to upload the VS2019 web installer as in `install_vs`, since the original download page needs redirection to the file url and I couldn't find any short url as for VS2017.

ChaiBapchya:
Updated links for cudnn & VS2019
- https://windows-post-install.s3-us-west-2.amazonaws.com/vs_community__1246179388.1585201415.exe
- https://windows-post-install.s3-us-west-2.amazonaws.com/cudnn-10.2-windows10-x64-v7.6.5.32.zip

ci/windows_dev_env/windows_deps_headless_installer.py
larroy: mxnet-ci-dev

leezu:
@mxnet-bot run ci [windows-cpu, windows-gpu]

ChaiBapchya:
retriggering.. windows

ChaiBapchya:
It's failing only for windows-cpu MKLDNN MKL part with the error
"can't find specified path"
```
[2020-03-27T18:52:40.264Z] "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvarsx86_amd64.bat" && cmake -G Ninja -DCMAKE_C_COMPILER=cl -DCMAKE_CXX_COMPILER=cl -DUSE_CUDA=OFF 
[2020-03-27T18:52:40.264Z] -DUSE_CUDNN=OFF -DENABLE_CUDA_RTC=OFF -DUSE_OPENCV=ON -DOpenCV_RUNTIME=vc15 -DOpenCV_ARCH=x64 -DUSE_OPENMP=ON -DUSE_BLAS=mkl -DUSE_LAPACK=ON -DUSE_DIST_KVSTORE=OFF -DUSE_MKL_IF_AVAILABLE=ON 
[2020-03-27T18:52:40.264Z] -DUSE_MKLDNN=ON -DCMAKE_BUILD_TYPE=Release C:\jenkins_slave\workspace\build-cpu-mkldnn-mkl
```

@larroy is this related to path not being on oneline? or the path itself is incorrect?

ChaiBapchya:
Turns out the path mentioned in ci/build_windows.py is correct
```
VS 2019': r'C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvarsx86_amd64.bat
```
That is exactly where the bat file is.

[verified this by launching an instance using the updated windows AMI]

vexilligera:
WIN_GPU_MKLDNN is having an issue similar to https://github.com/pytorch/pytorch/issues/25393

vexilligera:

vexilligera:
@mxnet-bot run ci [unix-gpu]

vexilligera:

ChaiBapchya:
@marcoabreu Even in this case, @vexilligera is unable to reproduce the error locally. And Windows GPU + MKLDNN is flaky. How do you suggest we proceed trying to resolve the thrust issue?

Fail : http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fwindows-gpu/detail/PR-17808/13/pipeline
Pass : http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fwindows-gpu/detail/PR-17808/12/pipeline

ChaiBapchya:
@vexilligera as discussed offline, lets try testing locally for WIN_GPU and WIN_GPU_MKLDNN build 10 times each (since 1 run takes 20-30mins) to come up with some basis... (ideally would have tried 100 times but given the resource & time constraints)

vexilligera:
> @vexilligera as discussed offline, lets try testing locally for WIN_GPU and WIN_GPU_MKLDNN build 10 times each (since 1 run takes 20-30mins) to come up with some basis... (ideally would have tried 100 times but given the resource & time constraints)

On my local test, the WIN_GPU_MKLDNN is much more flaky than WIN_GPU, as all WIN_GPU builds passed while about 1/3 of WIN_GPU_MKLDNN builds failed.

@haojin2 suggests us removing the WIN_GPU_MKLDNN test entirely since MKLDNN doesn't make much sense as we are running on GPU, and GPU_MKLDNN case is covered on other platforms.

ChaiBapchya:
We can get rid of WIN_GPU_MKLDNN tests altogether but that still leaves us with the flakiness of WIN_GPU as can be seen in these builds
For roughly same code of this PR & same windows AMI, below are the results so far

WIN_GPU | WIN_GPU_MKLDNN | Build Number | Link
-- | -- | -- | -- |
✖︎| ✔︎|15 | http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fwindows-gpu/detail/PR-17808/15/pipeline
✖︎| ✔︎| 14 | http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fwindows-gpu/detail/PR-17808/14/pipeline
✔︎|✔︎| 12 | http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fwindows-gpu/detail/PR-17808/12/pipeline
✔︎| ✖︎| 13 | http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fwindows-gpu/detail/PR-17808/13/pipeline

Ofcourse your tests on local have a different story to tell...

vexilligera:
@larroy do you have any idea on how to solve Win 126 on CI? Basically the binaries aren't in the search path and dlopen fails due to that

marcoabreu:
I'll have a look at the logs when I'm at my computer 

Just to be clear about two things:
1. Retries are fine for now, but must be removed before merging.
2. It is not an option to remove GPU mkldnn. It does make sense for cases where an operator is not supported by the GPU or the user wants to run some parts on CPU. Just because it breaks, it doesn't mean that we can remove it to ease our situation.

vexilligera:
> Sorry but I don't believe that you are testing the development on a VM when this kind of commit is necessary.

Was testing on another instance and made commits on another...manually copied the code and missed some parts...

ChaiBapchya:
>     1. Retries are fine for now, but must be removed before merging.

With the CI blocked on windows-gpu for last 4 days and this PR trying to unblock the CI with an updated AMI, I guess it would be a good way to unblock the CI by allowing retries for now.
I have created issue : https://github.com/apache/incubator-mxnet/issues/17935 for tracking this problem. 

But for unblocking CI do you think it makes sense to add the retries as a temporary hotfix?
@vexilligera tested the retries 10 times to see that adding retries safely ensures successful compilation all the time.

vexilligera:
I printed the path to the missing dll and it was:
C:\jenkins_slave\workspace\ut-python-gpu\windows_package\lib\libmxnet.dll
However, the build log shows this is packed into windows_package as you can find here:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fwindows-gpu/detail/PR-17808/19/

@larroy suggests packing the opencv binary as well but since opencv is already in the environment path of the AMI I guess the real problem should be something else (if opencv is missing and libmxnet.dll present you'll get segmentation fault instead, WinError 126 is triggered from Python due to wrong path)

@marcoabreu do you have any suggestion on this?

ChaiBapchya:
OSError: [WinError 126] The specified module could not be found

This one means the libmxnet.dll is not in the DLL Search path [acc to Stackoverflow]
Fix would be to add it into the environment
https://stackoverflow.com/questions/43987081/openslide-python-import-error

What's surprising is : libmxnet.dll & mxnet_70.dll are being packed into windows_package.7z and unpacked during test phase correctly.
While WIN_CPU tests don't give this error, WIN_GPU fails

ChaiBapchya:
@marcoabreu Also, after adding the retries, from build #18 to #22 all 5 have passed on WIN_GPU and WIN_GPU_MKLDNN compilation stage. That's a compeling reason for using retries to temporarily resolve the intermittent failure I feel.

vexilligera:
> Test in WIN_GPU fail at
> 
> ```
> OSError: [WinError 126] The specified module could not be found
> ```
> 
> This one means the libmxnet.dll is not in the DLL Search path [acc to Stackoverflow]
> Fix would be to add it into the environment
> https://stackoverflow.com/questions/43987081/openslide-python-import-error
> 
> What's surprising is : libmxnet.dll & mxnet_70.dll are being packed into windows_package.7z and unpacked during test phase correctly.
> While WIN_CPU tests don't give this error, WIN_GPU fails

@larroy says he used to run into the problem and it turned out that it was solved by packing opencv binary. In this case the opencv is already in the environment path and it should be found. Although WinError 126 could also occur if a dependent dll is missing, in my local tests when I deliberately created that situation, instead, segmentation fault (access violation at 0x0000000000000000) is thrown.

leezu:
> The only dependent dll of libmxnet.dll is cudart64_102.dll (not including the win32 dlls). For mxnet_70.dll, the libraries of opencv and openblas are already in the ami's environment path.

There should be more dependent dlls? For example, if built with fusion support, the cuda driver should also be a dependent library? https://github.com/apache/incubator-mxnet/issues/17858

https://github.com/apache/incubator-mxnet/blob/56e79853ad5cf98baf84454eb595c7658bef6ee6/CMakeLists.txt#L613-L621

vexilligera:
> > The only dependent dll of libmxnet.dll is cudart64_102.dll (not including the win32 dlls). For mxnet_70.dll, the libraries of opencv and openblas are already in the ami's environment path.
> 
> There should be more dependent dlls? For example, if built with fusion support, the cuda driver should also be a dependent library? #17858
> 
> https://github.com/apache/incubator-mxnet/blob/56e79853ad5cf98baf84454eb595c7658bef6ee6/CMakeLists.txt#L613-L621

They have fixed WinError 126 and now we are looking at a segfault

vexilligera:
@leezu Thank you for checking the instance type and changing the cuda arch to 5.2, the ci is running now. @ChaiBapchya previously the segfault was caused by specifying 7.0 on the g3 instance.

vexilligera:

vexilligera:
@ChaiBapchya I added instance type auto-detection and found out c5 was running the GPU ci, which caused the segfault.

ci/build_windows.py
```diff
@@ -33,16 +33,43 @@
...
+    # https://gitlab.kitware.com/cmake/cmake/issues/18920
     'VS 2015': r'C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\x86_amd64\vcvarsx86_amd64.bat',
-    'VS 2017': r'C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsx86_amd64.bat'
+    'VS 2017': r'C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars64.bat',
+    'VS 2019': r'C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars64.bat'
 }
 
 
+def get_cuda_arch(default='"5.2"'):
+    # cuda arch auto detection in cmake doesn't work on windows ci
```
leezu: The autodetection works fine on Windows. But we are compiling on a CPU instance. Can you remove this function and specify the MXNET_CUDA_ARCH argument as is done before? This function adds unnecessary complexity.

vexilligera: Mistakenly added due to misconception, already removed. Sorry about this.

leezu: No worries. Thanks for your work on this.

marcoabreu:
A missing dll might not mean that the dll itself is missing but that I can not be loaded to a miss in the dependency chain. I'd recommend to get the artifact and inspect it with some resources manager (similar to ldd). That should tell you which dependencies are expected and you can verify that they are available. 

Maybe also starting with a debugger could help to see a more detailed exception which explains which dependency of the dll is missing.

ci/build_windows.py
```diff
@@ -193,14 +238,19 @@ def windows_package(args):
         for dll in dlls:
             logging.info("packing dll: %s", dll)
             shutil.copy(dll, pkgdir_lib)
+
+        if pkgdir_lib.find('gpu') != -1:
+            add_path("C:\\jenkins_slave\\workspace\\ut-python-gpu\\windows_package\\lib")
```
josephevans, line 243: Updating the path should be done in the testing phase - in ci/windows/test_py3_gpu.ps1 so the tests can find all the DLLs.

vexilligera: That's right, forgot to remove that. Will add that to the testing phase once we got the log from ad95025

vexilligera:
@leezu seems like adding both 5.2 and 7.0 and then drop 7.0 would trigger a normal run...

ci/windows/test_py3_gpu.ps1
```diff
@@ -23,7 +23,20 @@ $env:MXNET_STORAGE_FALLBACK_LOG_VERBOSE=0
 $env:MXNET_SUBGRAPH_VERBOSE=0
 $env:MXNET_HOME=[io.path]::combine($PSScriptRoot, 'mxnet_home')
```
ChaiBapchya, line 25: let's print all ENV Vars to compare state of env vars in CI run vs manual powershell test
```Get-ChildItem Env:```

**Comment by vexilligera** [comment]

**Comment by vexilligera** [comment]

**Comment by leezu** [comment]
Connectivity errors appear fixed now by switching to g4 (with faster cpu and network).

Removed the debug statements and rebased on latest master.

**Comment by leezu** [comment]
@marcoabreu Do you still stand by your veto?

> "Retries are fine for now, but must be removed before merging."

I suggest to go ahead and merge this with the retries. We don't have control over the thrust + VS code incompatibility and may need to wait for a fix on nvidia's or microsoft's side. Doesn't seem worth to block all MXNet development until the bug in "thrust + VS code" is fixed, given it's an open issue affecting other projects as well: https://github.com/pytorch/pytorch/issues/25393

File: CMakeLists.txt [hunk 9]
```diff
@@ -163,6 +163,8 @@ if(MSVC)
   add_definitions(-DDMLC_STRICT_CXX11)
   add_definitions(-DNOMINMAX)
   set(CMAKE_C_FLAGS "/MP")
+  # report an accurate value for recent C++ language standards support
+  set(CMAKE_CXX_FLAGS "/Zc:__cplusplus")
```
leezu, line 167: @ChaiBapchya @vexilligera you added this line, but it doesn't have any effect as it is overwritte in the line below

ChaiBapchya: set(flag ${flag} "secondval") does append right?
I thought we are appending to CMAKE_CXX_FLAGS

ChaiBapchya: the typo!!!! `CMAKE_CXX_FLAGS` instead of `CMAKE_C_FLAGS`

ci/build_windows.py
```diff
@@ -106,29 +129,37 @@ class BuildFlavour(Enum):
...
         '-DENABLE_CUDA_RTC=ON '
         '-DUSE_OPENCV=ON  '
+        '-DOpenCV_RUNTIME=vc15 '
+        '-DOpenCV_ARCH=x64 '
         '-DUSE_OPENMP=ON '
         '-DUSE_BLAS=open '
         '-DUSE_LAPACK=ON '
         '-DUSE_DIST_KVSTORE=OFF '
-        '-DMXNET_CUDA_ARCH="5.2" '
+        '-DMXNET_CUDA_ARCH="7.5" '
```
marcoabreu, line 144: This change warrants further discussion. Please change it back.

ci/build_windows.py
```diff
@@ -253,12 +284,8 @@ def main():
     system = platform.system()
     if system == 'Windows':
         logging.info("Detected Windows platform")
-        if 'OpenBLAS_HOME' not in os.environ:
-            os.environ["OpenBLAS_HOME"] = "C:\\Program Files\\OpenBLAS-v0.2.19"
-        if 'OpenCV_DIR' not in os.environ:
-            os.environ["OpenCV_DIR"] = "C:\\Program Files\\OpenCV-v3.4.1\\build"
         if 'CUDA_PATH' not in os.environ:
-            os.environ["CUDA_PATH"] = "C:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v9.2"
+            os.environ["CUDA_PATH"] = "C:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v10.2"
```
marcoabreu, line 288: Same here

ci/safe_docker_run.py
```diff
@@ -38,8 +38,9 @@
 
 from util import config_logging
 
-DOCKER_STOP_TIMEOUT_SECONDS = 3
+DOCKER_STOP_TIMEOUT_SECONDS = 10
```
marcoabreu, line 41: Unrelated change

ci/windows_dev_env/windows_deps_headless_installer.py
```diff
@@ -0,0 +1,397 @@
...
+    local_file = download(DEPS['openblas'])
+    with zipfile.ZipFile(local_file, 'r') as zip:
+        zip.extractall("C:\\Program Files")
+    run_command("PowerShell Set-ItemProperty -path 'hklm:\\system\\currentcontrolset\\control\\session manager\\environment' -Name OpenBLAS_HOME -Value 'C:\\Program Files\\OpenBLAS-windows-v0_2_19'")
+    logging.info("Openblas Install complete")
+
+
+def install_mkl():
+    logging.info("Installing MKL 2019.3.203...")
+    file_path = download("http://registrationcenter-download.intel.com/akdlm/irc_nas/tec/15247/w_mkl_2019.3.203.exe")
```
marcoabreu, line 280: Move links and similar things into constants

marcoabreu:
Connectivity errors were never an issue with Windows slaves. It sounds like you've changed something as part of generating the AMIs. Thus, I'm hesitant to just accept a bunch of more changes (which have not been made transparent) as a result to fix a previously created issue.

Please just revert stuff back to what it was and stop putting duckttape over things. The build used to work and the slaves used to be stable. But I have only seen attempts to change things but no rootcausing (I know about the 2GB limit). Some action resulted in our build breaking and the only sane choice is to revert whatever was done and then do these efforts again.

I will not accept multiple layers of fixes which could have been avoided if the changes were properly developed and tested in a separate environment. At the moment, it seems like open heart surgery on a productive system.

ci/build_windows.py
leezu: This change sets the tested cuda architecture. The CI server now uses T4 GPUs which use 7.5 arch. Thus no, I can't change this back.

ci/build_windows.py
leezu: What is your concern?

ci/build_windows.py
marcoabreu: Well where did we agree that we are dropping g3 instances? How is such a change in the scope of this PR?

ci/build_windows.py
marcoabreu: We didn't agree on increasing the minimum cuda version

leezu:
Discussion continued at https://github.com/apache/incubator-mxnet/pull/17962#issuecomment-608066411

CMakeLists.txt
vexilligera: My bad. I thought it would append...

ci/build_windows.py
leezu: We agreed on using VS 2019. Cuda 10 is a requirement for using VS 2019

ci/build_windows.py
leezu: Let's keep using g3 instances for now then.

ci/build_windows.py
marcoabreu: Are you sure about that? Usually visual studio can be upgraded by installing the vc++ toolset which then grants the possibility to compile other cuda versions. I'm not aware that this categorically excluded.

leezu: https://superuser.com/questions/1506044/installing-cuda-9-2-with-vs-2019

leezu: https://devblogs.microsoft.com/cppblog/cuda-10-1-available-now-with-support-for-latest-microsoft-visual-studio-2019-versions/

vexilligera:
Closing since this is now fixed in #17962

leezu:
@vexilligera it's not fixed. We need a script to generate the working AMI. https://github.com/apache/incubator-mxnet/pull/17962 is only a hotfix.

Will you be working on the automated setup or is someone else taking over the work? Please clarify. Thank you

PR #17815, opened 2020-03-11
[WIP] Add SOVERSION when build shared libmxnet.so library

sl1pkn07:
## Description ##
https://en.wikipedia.org/wiki/Soname
https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html

### Essentials ###

### Changes ###

## Comments ##

this rename the library from `libmxnet.so` to `libmxnet.so.x.x.x`, and cleate symlinks to `libmxnet.so.x` and `libmxnet.so`

~~~
└───╼  ls /usr/lib/libmxnet.so*
lrwxrwxrwx 1 root root        13 mar 11 05:12 /usr/lib/libmxnet.so -> libmxnet.so.2
lrwxrwxrwx 1 root root        17 mar 11 05:12 /usr/lib/libmxnet.so.2 -> libmxnet.so.2.0.0
-rwxr-xr-x 1 root root 583754776 mar 11 05:12 /usr/lib/libmxnet.so.2.0.0

└───╼  objdump -p /usr/lib/libmxnet.so.2.0.0| grep SONAME
  SONAME               libmxnet.so.2

└───╼  readelf -d /usr/lib/libmxnet.so.2.0.0 | grep SONAME
 0x000000000000000e (SONAME)             Nombre-so de la biblioteca: [libmxnet.so.2]
~~~


leezu:
You may need to update the CI settings to be more flexible about the name of the `.so` file:
See the locations here https://github.com/apache/incubator-mxnet/search?q=mx_lib&unscoped_q=mx_lib

Further you'll need to update the staticbuild script https://github.com/apache/incubator-mxnet/blob/0f29ccaf5c34a01bffe5ad3036d4cf8bb00c551e/tools/staticbuild/build_lib_cmake.sh and perhaps the setup.py for the staticbuild https://github.com/apache/incubator-mxnet/tree/0f29ccaf5c34a01bffe5ad3036d4cf8bb00c551e/tools/pip

sl1pkn07:
Hi

That accepts things like 'libmxnet.so.*'?, Because the version depends on the value provided by 'include/mxnet.h'

Greetings

leezu:
Wildcards work. For example, `python/mxnet/_ffi/_cy3/*.so` is already used.

ci/jenkins/Jenkins_steps.groovy
```diff
@@ -23,27 +23,27 @@
...
-mx_lib_cpp_capi_make = 'lib/libmxnet.so, lib/libmxnet.a, lib/libtvm_runtime.so, lib/libtvmop.so, lib/tvmop.conf, libsample_lib.so, lib/libmkldnn.so.1, lib/libmklml_intel.so, 3rdparty/dmlc-core/libdmlc.a, 3rdparty/tvm/nnvm/lib/libnnvm.a, 3rdparty/ps-lite/build/libps.a, deps/lib/libprotobuf-lite.a, deps/lib/libzmq.a, build/cpp-package/example/*, python/mxnet/_cy3/*.so, python/mxnet/_ffi/_cy3/*.so, build/tests/cpp/mxnet_unit_tests'
-mx_lib_cpp_examples_no_tvm_op = 'build/libmxnet.so, build/libcustomop_lib.so, build/libcustomop_gpu_lib.so, build/libsubgraph_lib.so, build/3rdparty/openmp/runtime/src/libomp.so,  build/cpp-package/example/*, python/mxnet/_cy3/*.so, python/mxnet/_ffi/_cy3/*.so'
-mx_lib_cpp_examples_cpu = 'build/libmxnet.so, build/3rdparty/tvm/libtvm_runtime.so, build/libtvmop.so, build/tvmop.conf, build/3rdparty/openmp/runtime/src/libomp.so, build/cpp-package/example/*'
+mx_cmake_lib_debug = 'build/libmxnet.so*, build/3rdparty/tvm/libtvm_runtime.so, build/libtvmop.so, build/tvmop.conf, build/libcustomop_lib.so, build/libcustomop_gpu_lib.so, build/libsubgraph_lib.so, build/tests/mxnet_unit_tests'
```
leezu: Unrelated to your PR, but potential improvement: Have only one variable for makefile build and one for cmake build instead of 10+. If for a particular build more files are listed than are present, the additional ones are just ignored.

CMakeLists.txt
```diff
@@ -1,5 +1,14 @@
 cmake_minimum_required(VERSION 3.13)
 
+file(READ "include/mxnet/base.h" ver)
+string(REGEX MATCH "MXNET_MAJOR ([0-9]*)" _ ${ver})
+set(ver_major ${CMAKE_MATCH_1})
+string(REGEX MATCH "MXNET_MINOR ([0-9]*)" _ ${ver})
+set(ver_minor ${CMAKE_MATCH_1})
+string(REGEX MATCH "MXNET_PATCH ([0-9]*)" _ ${ver})
+set(ver_patch ${CMAKE_MATCH_1})
+set(_PROJECT_VERSION "${ver_major}.${ver_minor}.${ver_patch}")
```
leezu: Why set `_PROJECT_VERSION`?

Instead you may want to use something like `project(mxnet C CXX VERSION "${ver_major}.${ver_minor}.${ver_patch}")` instead of the existing `project(mxnet C CXX)`?

sl1pkn07: because ikd can merge it, is is valid, then I merge it

sl1pkn07: missing "language"

https://cmake.org/cmake/help/latest/command/project.html

noy im not sure if need other changes

CMakeLists.txt
```diff
@@ -704,6 +712,8 @@ if(UNIX)
   target_link_libraries(mxnet PRIVATE mxnet_static)
   target_link_libraries(mxnet_static PUBLIC ${CMAKE_DL_LIBS})
   set_target_properties(mxnet_static PROPERTIES OUTPUT_NAME mxnet)
+  set_target_properties(mxnet PROPERTIES VERSION "${ver_major}.${ver_minor}.${ver_patch}")
```
sl1pkn07: needs this with the last change?

leezu: https://cmake.org/cmake/help/latest/prop_tgt/VERSION.html does not say anything about the target property being initialized from the project version. So it should still be needed. But you can try it out locally: Don's set the target property but print it. I expect it would print an empty value.

Instead of using `ver_major` you can use https://cmake.org/cmake/help/latest/variable/PROJECT_VERSION_MAJOR.html#variable:PROJECT_VERSION_MAJOR etc
 for clarity

sl1pkn07: like this?

~~~cmake
  set_target_properties(mxnet PROPERTIES VERSION "${PROJECT_VERSION}")
  set_target_properties(mxnet PROPERTIES SOVERSION "${PROJECT_VERSION_MAJOR}")
~~~

leezu: You don't need the quotation marks around ${PROJECT_VERSION}

Ref: Github search: https://github.com/search?l=CMake&q=set_target_properties%28+PROPERTIES+VERSION+PROJECT_VERSION%29&type=Code

sl1pkn07: without quotes
<details>

Screenshot_20200318_145003

</details>

with quotes
<details>

Screenshot_20200318_145123

</details>

sl1pkn07: other test, ans this surprise me, i just add
~~~cmake
message(STATUS "Project version: ${PROJECT_VERSION}")
~~~
just before `project()` function, and works

Screenshot_20200318_151002

bus is vanished if put that in the bottom/middle of the CMakeLists

Screenshot_20200318_151423
Screenshot_20200318_151156


cmake is a bit complex for me, IDK where is the problem, so, i think is better keep untounch with `${ver_major}.${ver_minor}.${ver_patch}` and in near future (i i found the problem), fix it

greetings

sl1pkn07: ook, i found the problem

`project()` is call twice in the main CMakeFile

https://github.com/apache/incubator-mxnet/blob/master/CMakeLists.txt#L9
https://github.com/apache/incubator-mxnet/blob/master/CMakeLists.txt#L89

why this?

sl1pkn07: i think is finally done

tested on local and works as spected

leezu: Thank you for fixing it

leezu:
@mxnet-bot run ci [all]

cc @ChaiBapchya @josephevans for some reason the CI wasn't triggered in my merge commit.

leezu:

mseth10:
Seems like this commit breaks the CD pipeline, can you please take a look @sl1pkn07 @leezu 
https://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/restricted-mxnet-cd%2Fmxnet-cd-release-job/detail/mxnet-cd-release-job/1521/pipeline/345

szha:
The CD pipeline needs to be fixed. In addition, it looks like the CD pipeline is receiving libmxnet files with [varying file content](https://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/restricted-mxnet-cd%2Fmxnet-cd-release-job/detail/mxnet-cd-release-job/1524/pipeline/302#step-325-log-6)
```
[2020-08-08T20:16:46.129Z] + md5sum lib/libmxnet.so
[2020-08-08T20:16:46.129Z] da79e007aa331599ab5e474b9ade855d  lib/libmxnet.so
[2020-08-08T20:16:46.129Z] + md5sum lib/libmxnet.so.2
[2020-08-08T20:16:46.689Z] 791b1014c70e5b421285a83f5fee63e7  lib/libmxnet.so.2
[2020-08-08T20:16:46.689Z] + md5sum lib/libmxnet.so.2.0.0
[2020-08-08T20:16:46.689Z] 791b1014c70e5b421285a83f5fee63e7  lib/libmxnet.so.2.0.0
```
I'm reverting this PR to unblock other development. That said, I totally understand that the CD pipeline is unnecessarily complex and it's a bit hard to track what caused the problem. @sandeep-krishnamurthy could your team take up reapplying the patch after revert and fix the CD for it?

cd/mxnet_lib/static/Jenkins_pipeline.groovy
```diff
@@ -26,7 +26,7 @@
 // NOTE: the following variables are referenced in the mxnet_lib_pipeline jenkins file imported bellow
 
 // libmxnet location
-libmxnet = 'lib/libmxnet.so'
+libmxnet = 'lib/libmxnet.so*'
```
mseth10, line 29: With this change, are we planning to package all three binaries (viz. libmxnet.so, libmxnet.so.2, libmxnet.so.2.0.0) into the nightly wheel? @sl1pkn07 @leezu

sl1pkn07:
the library is libmxnet.so.2.0.0. libmxnet.so and libmxnet.so.2 are symlinks of libmxnet.so.2.0.0

szha:
I understand, which makes it weird that the libmxnet.so.2 and libmxnet.so.2.0.0 share the same md5 hash but libmxnet.so doesn't.

leezu:
@mseth10 would you help to fix the `artifact_repository.py` and resubmit the PR as CD is currently not validated by the CI and contributors can't replicate the system.

mseth10:
@sl1pkn07 @leezu thanks for confirming that all three files need to be packaged in the nightly wheel. Will work on fixing the `artifact_repository.py` accordingly. What about @szha 's concern, do the md5 hashes make sense?

szha:
@mseth10 my concern is on the CD pipeline, not on this PR. the problem has to be on the CD side for copying the artifacts.

mseth10:
I'll look into it as well. So all three binary files must share the same md5 hash as they are just pointing to the same binary. Is my understanding correct?

leezu:
Yes, `md5sum` would normally report the hash the file content that is being symlinked, and thus all hashes should be the same:

```
% ls -l test*
-rw-r--r-- 1 lausen domain^users    0 Aug 10 20:40 test
lrwxrwxrwx 1 lausen domain^users    4 Aug 10 20:41 test1 -> test
lrwxrwxrwx 1 lausen domain^users    5 Aug 10 20:41 test1.1 -> test1
% md5sum test*
d41d8cd98f00b204e9800998ecf8427e  test
d41d8cd98f00b204e9800998ecf8427e  test1
d41d8cd98f00b204e9800998ecf8427e  test1.1

```

szha:
@mseth10 in fact we shouldn't need to copy all three files when building the wheel. the wrong md5 sum may indicate that some files are symlink or incomplete files.

mseth10:
Reading about symlink support in wheels, there is no agreed upon solution in the community. There are some workarounds for unix-based platforms, but they dont work for windows. https://discuss.python.org/t/symbolic-links-in-wheels/1945

Would it make sense to copy just the `libmxnet.so` when building the wheels?

leezu:
@mseth10 it's fine to include only a single file in the wheel

leezu:
@mseth10 that's also a valid solution

Issue #17269, opened 2020-01-10
[mxnet 2.0][item 4.8][RFC] Gluon Data API Extension and Fixes(Part 2)


zhreshold:
## Description
This is the part 2 of Gluon Data API extension and fixes, which mainly focus on speeding up the current data loading pipeline using gluon dataset and dataloader.

## Motivation

The current data loading pipeline is the major bottleneck for many training tasks. We can summarize the entire flow as:

```bash
| Dataset.__getitem__ -> 
| Transform.__call__()/forward() ->
| Batchify ->
| (optional communicate through shared_mem) ->
| split_and_load(ctxs) ->
| <training on GPUs>
-> 
```
where there are performance concerns:
- performance of python dataset/transform functions aren't satisfying
- it's not easy to embrace multithreading to speed up dataloading due to global interpreter lock
- python multiprocessing is unfortunately slow and error prune, not to mention the shared memory implementations on different OS are quite difference and very annoying(e.g., it's very likely to run out of shared memory if not properly taken care of)
- currently memory planing for batchify is non-exist, causing frequent alloc/dealloc for large chunk of memory if the batch size is big
- batchify then split and load can be optimized to partial_batchify

## Proposal
To alleviate the existing troubles I propose to use a hybrid solution, that is to 
- provide C++ Datasets that can cover the most usecases
    ```python
    from gluon.data.dataset import TupleDataset, ImageFolderDataset, ArrayDataset
    # as long as TupleDataset, ImageSequenceDataset, ArrayDataset are supported by backend
    dataset = TupleDataset([ImageSequenceDataset(img_paths), ArrayDataset(image_labels)])
    # dataset is an image classification dataset while fully supported in C++
    # with TupleDataset we can combine as many data as possible

    # a C++ backed Dataset can have a magic __handle__ method to return the c++ handle for reference
    class TupleDataset:
        def __init__(self, datasets):
            if all([callable(getattr(dataset, '__handle__')) for dataset in datasets]):
                # all supported by backend
                self._tuple_dataset = check_call(_LIB.MXTupleDatasetCreate([getattr(dataset, '__handle__') for dataset in datasets]))
            else:
                self._tuple_dataset = None

            def __handle__(self):
                return self._tuple_dataset
                    
    ```
- provide common C++ batchify functions that are split and context aware. Batchify with memory planner is TBD.
[...]

PR #17841, opened 2020-03-15
Gluon data 2.0: c++ dataloader and built-in image/bbox transforms

zhreshold:
## Description ##
This is the implementation for proposal #17269

### Changes ###
The major components of this PR is:
- c++ threaded dataloader with native support from the following datasets/batchify_functions
- c++ native support for datasets covering `gluon.data.datasets` and `gluon.data.vision.datasets`
- c++ batchify functions including `Stack`, `Pad`, `Groups`
- Gluon version of transforms that can apply data augmentations to image/bboxes
- Gluon version of data iterator substitutes of `mx.image.ImageIter` and `mx.image.ImageDetIter` for simplest interface for loading image classification and object detection datsets.

### Code Review ###
Since this PR involves lots of backend and frontend changes and can not be easily divided into multiple PRs, so I would like to suggest modular code reviews:

| Modules | notes | suggested reviewers | 
| ------------| -------- | --------------------------- |
| include/mxnet/c_api.h/cc, include/mxnet/io.h,  python/mxnet/gluon/data/_internal.py,  python/mxnet/io/io.py | c api changes for io: datasets and batchify functions, with corresponding python frontends | @szha, @leezu |
| python/mxnet/gluon/data/batchify.py(dataloader.py, dataset.py) | gluon data loader and batchify functions | @sxjscience |
|  python/mxnet/gluon/contrib/data/vision/dataloader.py,  python/mxnet/gluon/contrib/data/vision/transforms/* | substitutes for deprecated mx.image.ImageIter and mx.image.ImageDetIter and old augmenters | all committers |
|  python/mxnet/gluon/nn/basic_layers.py | changes to `nn.sequential` and `nn.HybridSequential` to allow Sequential block to take more than 1 arguments |  all gluon contributors |
|  python/mxnet/image/image.py | numpy compatitility change | @haojin2 |
| src/imperative/* | Minor modification to cached op to allow operators to skip engine in imperative executions | @eric-haibin-lin |
|  src/io/dataloader.cc(dataset.cc, batchify.cc), src/io/iter_sampler.cc | c++ implementations of gluon.data(gluon.data. dataset, gluon.data.batchify) | @leezu @sxjscience  |
|  src/operator/image/* | operators to support many random image transformations | committers familiar with vision operators |

Thank you for everyone who participate in code review and sorry about the size of this PR.


zhreshold:
Ping for reviewers' attention. I will fix lint along with review comments to reduce CI triggers.

src/io/dataset.cc
```diff
@@ -0,0 +1,696 @@
...
+    // declare parameters
+    DMLC_DECLARE_PARAMETER(NDArrayDatasetParam) {
+        DMLC_DECLARE_FIELD(arr)
+            .describe("Pointer to NDArray.");
+    }
+};  // struct NDArrayDatasetParam
+
+DMLC_REGISTER_PARAMETER(NDArrayDatasetParam);
+
+class NDArrayDataset : public Dataset {
```
leezu: Specify all datasets as `final`? This will allow the compiler to perform further optimization

include/mxnet/io.h
```diff
@@ -113,5 +120,105 @@ struct DataIteratorReg
...
+ * data augmentation or transformation processes.
+ */
+class Dataset {
+  public:
+    /*!
+    *  \brief Initialize the Operator by setting the parameters
+    *  This function need to be called before all other functions.
+    *  \param kwargs the keyword arguments parameters
+    */
+    virtual void Init(const std::vector<std::pair<std::string, std::string> >& kwargs) = 0;
```
leezu: As "This function need to be called before all other functions", why not use a constructor instead? That way compiler will ensure initialization happens.

python/mxnet/gluon/data/batchify.py
```diff
@@ -0,0 +1,433 @@
...
+    >>> a = nd.array([[1, 2, 3, 4], [5, 6, 7, 8]])
+    >>> b = nd.array([[5, 8], [1, 2]])
+    >>> batchify.Pad(pad_val=-1)([a, b])
+    [[[ 1.  2.  3.  4.]
+      [ 5.  6.  7.  8.]]
+     [[ 5.  8. -1. -1.]
+      [ 1.  2. -1. -1.]]]
+    <NDArray 2x2x4 @cpu(0)>
+    """
+    def __init__(self, pad_val=None, ret_length=False, dtype=None, use_shared_mem=False):
```
sxjscience: In the new numpy version of GluonNLP, I removed `ret_length` and just call `pad_val` as `val`. Also, I removed the warning. In addition, there is a round_to flag as in https://github.com/dmlc/gluon-nlp/blob/3f7465ab5d0f926c2c6f424644daaa30668570ba/src/gluonnlp/data/batchify/batchify.py#L222

python/mxnet/gluon/data/_internal.py
```diff
@@ -0,0 +1,348 @@
...
+        # check bound
+        if idx < 0 or idx >= self._len:
+            raise IndexError("Index {} out of bound: (0, {})".format(orig_idx, self._len))
+        create_ndarray_fn = _np_ndarray_cls if is_np_array() else _ndarray_cls
+        output_vars = ctypes.POINTER(NDArrayHandle)()
+        num_output = ctypes.c_int(0)
+        check_call(_LIB.MXDatasetGetItems(self.handle,
+                                          ctypes.c_uint64(idx),
+                                          ctypes.byref(num_output),
+                                          ctypes.byref(output_vars)))
```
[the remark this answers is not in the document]
zhreshold: I am thinking of a separate PR that addresses the FFI improvement for all data related APIs. For now, maintaining consistency with other data APIs is preferable.

python/mxnet/gluon/data/batchify.py
```diff
@@ -0,0 +1,433 @@
...
+        else:
+            out = [_arr.array(x) for x in arrs]
+
+    # add batch axis
+    if expand:
+        out = [x.expand_dims(axis=batch_axis) for x in out]
+    return out
+
+
+class Append(object):
```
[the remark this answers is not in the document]
zhreshold: Is python2 explicitly disabled now in master?

python/mxnet/gluon/data/dataloader.py
```diff
@@ -572,16 +564,26 @@ def default_batchify_fn(data):
         unless you are experiencing timeout and you know it's due to slow data loading.
         Sometimes full `shared_memory` will cause all workers to hang and causes timeout. In these
         cases please reduce `num_workers` or increase system `shared_memory` size instead.
+    try_nopython : bool, default is None
```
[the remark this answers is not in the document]
zhreshold: Here's what I set for `try_nopython`:
- `True` to explicitly check for failure cases, e.g. raise warnings if pure c++ loading is not available.
- `False`, do not engage c++, always use python
- `None`, check for c++ and enable silently, if not available, fall back to python silently. This is designed for user who DON'T care

src/imperative/cached_op.h
```diff
@@ -524,15 +528,15 @@ class CachedOp {
...
- private:
   struct DynamicRuntime;
 
   OpStatePtr DynamicForward(
       const Context& default_ctx,
       const std::vector<NDArray*>& inputs,
       const std::vector<NDArray*>& outputs,
       bool use_naive_run = false);
+
+ private:
```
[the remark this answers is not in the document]
zhreshold: Cached op don't allow backward to be called directly. I didn't change the behavior.

include/mxnet/io.h
zhreshold: Due to the argument parsing system mxnet uses, python has to create object first and pass arguments to call `Init`

zhreshold:
> Awesome job! Would you mind also looking into how we adopt the new FFI?


> I am thinking of a separate PR that addresses the FFI improvement for all data related APIs. For now, maintaining consistency with other data APIs is preferable.

include/mxnet/io.h
leezu: `MXDatasetCreateDataset` writes the address of the created dataset to the `out` argument. Why is there be a requirement to have an `Init` function?

python/mxnet/gluon/data/batchify.py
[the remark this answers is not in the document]
leezu: Yes

src/imperative/cached_op.h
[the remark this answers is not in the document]
leezu: Well, previously `DynamicForward` was `private`

include/mxnet/io.h
leezu: This is still outstanding. Do you have a plan to fix the API design issue?

zhreshold:
> Let's use constructor to initialize the classes instead of requiring users to call Init after construction?

@leezu Sorry I missed this comment. If you look at the creator in c_api used for list and create the existing mxnet iterators: https://github.com/apache/incubator-mxnet/blob/master/src/c_api/c_api.cc#L1859, you will find that the registry is used to hold all uninitialized iterators that has been registered to io module: https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/io/io.py#L984

So if I follow the current registeration patter in python module, I have to follow the empty constructor + init function.

leezu:
I'm not convinced. You can change the call signature of the registry and provide `const std::vector<std::pair<std::string, std::string> >& kwargs` to the constructor instead of the Init function. There should be no problem.

include/mxnet/io.h
```diff
@@ -113,5 +120,104 @@ struct DataIteratorReg
...
+  *  \param idx the integer index for required data
+  *  \param ret the returned ndarray items
+  */
+  virtual bool GetItem(uint64_t idx, std::vector<NDArray>* ret) = 0;
+  // virtual destructor
+  virtual ~Dataset(void) {}
+};  // class Dataset
+
+/*! \brief typedef the factory function of dataset */
+typedef std::function<Dataset *()> DatasetFactory;
```
leezu, line 157: Need to change DatasetFactory type to accept kwargs.

src/c_api/c_api.cc
```diff
@@ -1924,6 +1967,194 @@ int MXDataIterGetPadNum(DataIterHandle handle, int *pad) {
...
+
+int MXDatasetCreateDataset(DatasetCreator handle,
+                           uint32_t num_param,
+                           const char **keys,
+                           const char **vals,
+                           DatasetHandle *out) {
+  Dataset *dataset = nullptr;
+  API_BEGIN();
+  DatasetReg *e = static_cast<DatasetReg *>(handle);
+  dataset = e->body();
```
leezu, line 1987: Need to pass kwargs here after changing registry func type.

zhreshold: okay, I got your point, will change to constructor.

leezu: Thanks. For reference you can check https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-complete why this is a good practice.

RuRo:
@zhreshold I see, you increased the timeout for the GPU tests. If you are trying to avoid aborts of the `Python 3: GPU` and `Python 3: GPU (TVM_OP OFF)` tests on `unix-gpu` after 3 hours, this won't help. This is a known bug, where those tests sometimes get completely stuck. These tests should normally complete in about an hour. See #18090 for discussion.

zhreshold:
@leezu Outstanding comments fixed, can you take another look?
@szha @eric-haibin-lin can you please either request change or approve?

tests/python/unittest/test_contrib_gluon_data_vision.py
```diff
@@ -0,0 +1,145 @@
...
+    right = (xy[:, 0] + wh[:, 0])[:, np.newaxis]
+    top = (xy[:, 1] - wh[:, 1])[:, np.newaxis]
+    bot = (xy[:, 1] + wh[:, 1])[:, np.newaxis]
+    boxes = np.maximum(0., np.minimum(1., np.hstack((left, top, right, bot))))
+    cid = np.random.randint(0, 20, size=num)
+    label = np.hstack((cid[:, np.newaxis], boxes)).ravel().tolist()
+    return [2, 5] + label
+
+
+class TestImage(unittest.TestCase):
```
leezu, line 53: Did you check this is completely covered by pytest's unittest support https://docs.pytest.org/en/latest/unittest.html ?

tests/python/unittest/test_numpy_interoperability.py
```diff
@@ -1832,7 +1832,7 @@ def test_scalar_output():
             v2 = vec2.astype(dt)
             OpArgMngr.add_workload('matmul', v1, v2)
             OpArgMngr.add_workload('matmul', v2.T, v1)
-
+
```
leezu, line 1835: Why add the whitespace here (and many other places in this file).

include/mxnet/io.h
```diff
@@ -113,5 +120,98 @@ struct DataIteratorReg
...
+class Dataset {
+ public:
+  /*!
+  *  \brief Get the size of the dataset
+  */
+  virtual uint64_t GetLen(void) const = 0;
+  /*!
+  *  \brief Create a copy of dataset for threaded worker
+  */
+  virtual Dataset* Clone(void) const = 0;
```
leezu, line 139: Is this used? If not, delete. If yes, why not use c++ copy mechanism?

tests/python/unittest/test_numpy_interoperability.py
zhreshold: git merging issue, will delete

tests/python/unittest/test_contrib_gluon_data_vision.py
zhreshold: Yes, it's modified to accomodate the transition from nose to pytest

tests/python/unittest/test_numpy_contrib_gluon_data_vision.py
```diff
@@ -0,0 +1,146 @@
...
+            mx.gluon.contrib.data.vision.ImageBboxDataLoader(3, (3, 400, 400),
+                path_imglist=fname, path_root='', last_batch='rollover'),
+            mx.gluon.contrib.data.vision.ImageBboxDataLoader(3, (3, 400, 400), shuffle=True,
+                path_imglist=fname, path_root='', last_batch='keep')
+        ]
+
+    @use_np
+    def test_bbox_augmenters(self):
+        # only test if all augmenters will work
+        # TODO(Joshua Zhang): verify the augmenter outputs
```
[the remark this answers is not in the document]
zhreshold: Yes, it's verified by visual inspect and training, however the unittest doesn't support checking random augmenters right now. If you have better idea please let me know.

src/io/iter_sampler.cc
```diff
@@ -0,0 +1,183 @@
...
+
+DMLC_REGISTER_PARAMETER(RandomSamplerParam);
+
+class RandomSampler : public IIterator<DataInst> {
+ public:
+  virtual void Init(const std::vector<std::pair<std::string, std::string> >& kwargs) {
+    param_.InitAllowUnknown(kwargs);
+    indices_.resize(param_.length);
+    std::iota(std::begin(indices_), std::end(indices_), 0);  // fill like arange
+    rng_.reset(new common::RANDOM_ENGINE(kRandMagic + param_.seed));
```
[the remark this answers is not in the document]
zhreshold: It's sharing the same random engine with mxnet's inited seed. Adding a separate seed here is can help users to tweak the randomness of this particular sampler. The idea is, 
- if you change seed for mxnet, it affects all internal random generator, including this sampler. 
- If tweaking the seed specified here, only this sampler is affected.

python/mxnet/gluon/data/dataloader.py
```diff
@@ -655,3 +680,119 @@ def __del__(self):
...
+            return False, error_template.format('pass', 'pass', 'fail')
+    elif isinstance(batch_sampler, MXSampler):
+        mx_loader_args['batch_sampler'] = batch_sampler
+    else:
+        return False, error_template.format('pass', 'pass', 'fail')
+    # all good
+    return True, mx_loader_args
+
+
+class MXThreadedDataLoader(object):
```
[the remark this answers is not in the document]
zhreshold: Thanks, changed to private

src/io/iter_sampler.cc
```diff
@@ -0,0 +1,183 @@
...
+    }
+    return false;
+  }
+
+  virtual const DataInst &Value(void) const {
+    return out_;
+  }
+ private:
+  /*! \brief random magic number */
+  static const int kRandMagic = 2333;
```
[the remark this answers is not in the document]
zhreshold: See https://github.com/apache/incubator-mxnet/pull/17841#discussion_r416847315

> It's sharing the same random engine with mxnet's inited seed. Adding a separate seed here is can help users to tweak the randomness of this particular sampler. The idea is,
> 
> * if you change seed for mxnet, it affects all internal random generator, including this sampler.
> * If tweaking the seed specified here, only this sampler is affected.

python/mxnet/gluon/data/dataloader.py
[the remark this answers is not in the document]
zhreshold: Done

python/mxnet/gluon/data/dataloader.py
```diff
@@ -572,16 +564,26 @@ def default_batchify_fn(data):
         unless you are experiencing timeout and you know it's due to slow data loading.
         Sometimes full `shared_memory` will cause all workers to hang and causes timeout. In these
         cases please reduce `num_workers` or increase system `shared_memory` size instead.
+    try_nopython : bool, default is None
+        Try compile python dataloading pipeline into pure MXNet c++ implementation. The benefit is
+        potentially faster iteration, no `shared_memory` usage, and less processes managed by python.
+        The compilation is not gauranteed to support all use cases, but it will fallback to python in
+        case of failure. You can set `try_nopython` to `False` to disable auto-detection of the
+        compilation feature or leave it to `None` to allow MXNet to determine it automatically.
+        If you request `try_nopython` to `True` and the compilation fails, it will raise a warning and
```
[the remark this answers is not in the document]
zhreshold: Modified to Raise RuntimeError instead.

python/mxnet/gluon/contrib/data/vision/dataloader.py
```diff
@@ -0,0 +1,521 @@
...
+    augmenter.add(transforms.image.ToTensor())
+
+    if mean is not None or std is not None:
+        augmenter.add(transforms.image.Normalize(mean, std))
+
+    augmenter.add(transforms.Cast(dtype))
+
+    return augmenter
+
+class ImageDataLoader(object):
```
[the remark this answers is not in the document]
zhreshold: It's a thin wrapper on top of `DataLoader`, the approach you mentioned is exactly what I was thinking of . In fact, anyone can directly use `DataLoader` with `transform` or `transform_first` by manually compose the transforms. 

However, I guess users of the legacy `ImageDetIter` won't be happy with the deprecation of the old iterator so I am trying to keep a contrib version with augmenters wrapped in for convenient transition.

ci/build_windows.py
```diff
@@ -184,6 +184,7 @@ def windows_build(args):
             if ret != 0:
                 build_try += 1
                 logging.info("{} build(s) have failed".format(build_try))
+                sys.exit(1)
```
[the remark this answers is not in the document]
zhreshold: Added to stop windows ci with error status. With https://github.com/apache/incubator-mxnet/pull/18177 I can delete this.

src/io/iter_sampler.cc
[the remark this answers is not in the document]
leezu: I agree about the idea. My point is that I don't think the current code implements this idea, but rather uses a fixed magic seed plus a user-specified seed which defaults to a fixed number as well.

tests/python/unittest/test_numpy_contrib_gluon_data_vision.py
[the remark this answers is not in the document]
leezu: Let's remove the in-line to do then?

src/io/iter_sampler.cc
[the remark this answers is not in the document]
zhreshold: Good catch, will try to align with the random generator resource

src/io/dataset.cc
```diff
@@ -0,0 +1,697 @@
...
+      }
+    }
+    return true;
+  }
+
+ private:
+  /*! \brief parameters */
+  RecordFileDatasetParam param_;
+  /*! \brief recordIO context */
+  std::shared_ptr<dmlc::RecordIOReader> reader_;
```
leezu, line 124: Why not make this `thread_local` given the thread safety issues? Then you may not need to copy around datasets at the level of the sampler

zhreshold:
> Thanks @zhreshold!
> 
> I have two general questions and also left some comments / requested changes inline:
> 
> * What's the plan for python/mxnet/gluon/contrib/data/vision/transform? You mentioned we like to remove vision API and keep it in GluonCV only? Does that not apply to the data API? Will the data API be removed in GluonCV or how is the relation to the one in MXNet?
> * How can users run the pipeline / part of it on GPU? Some of the operators added in this PR do have GPU implementations, but I'm not sure how they could be used. Is there any plan on how to extend the current design to integrate with external libraries such as nvidia dali?
> 
> For the inline comments, the "requested changes" are with respect to CI Windows build script and the random state in the samplers. The others are more open-ended questions

@leezu 

- We are indeed removing CV specific implementation, especially network definitions, in favor of GluonCV. In the meantime some fundmental image processing transforms will still be added to MXNet. For definition of `fundmental`, I refer to the usage in [D2L](https://d2l.ai/) for guidance. 

- For GPU, there's a lot of topics we can discuss, including end-to-end GPU training, DALI integration, hybrid CPU/GPU data pipeline, (de)serialization of data pipeline. I am open to all these discussions, but do not have bandwidth to work on those areas yet.

leezu:
So the fundamental APIs would only be maintained in MXNet side and removed from GluonCV?

zhreshold:
> So the fundamental APIs would only be maintained in MXNet side and removed from GluonCV?

@leezu  Yes

python/mxnet/gluon/data/dataloader.py
```diff
@@ -607,28 +609,51 @@ def __init__(self, dataset, batch_size=None, shuffle=False, sampler=None,
...
+        if use_mx_iter:
+            logging.info("Using MXNet backend ThreadedDataLoader with %s workers "
+                         "instead of python dataloader.", self._num_workers)
+            self._mx_iter = MXThreadedDataLoader(
+                num_workers=self._num_workers,
+                pin_memory=self._pin_memory,
+                pin_device_id=self._pin_device_id,
+                prefetch=self._prefetch, **mx_iter_args)
+        else:
+            if self._num_workers > 0:
```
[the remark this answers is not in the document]
zhreshold: num_workers == 0 is handled in `__iter__` and there's nothing needs to be done in `__init__`

src/io/iter_prefetcher.h
```diff
@@ -86,14 +88,20 @@ class PrefetcherIter : public IIterator<DataBatch> {
...
-                                          Context::CPU(), false,
+                                          ctx, false,
                                           dtype);
           }
         }
         CHECK(batch.data.size() == (*dptr)->data.size());
         // copy data over
         for (size_t i = 0; i < batch.data.size(); ++i) {
+          if ((*dptr)->data.at(i).shape() != batch.data[i].shape_) {
+            // perf warning, dynamic buffer might be slow
```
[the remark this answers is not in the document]
zhreshold: Modified

src/io/iter_sampler.cc
```diff
@@ -0,0 +1,183 @@
...
+
+DMLC_REGISTER_PARAMETER(SequentialSamplerParam);
+
+class SequentialSampler : public IIterator<DataInst> {
+ public:
+  virtual void Init(const std::vector<std::pair<std::string, std::string> >& kwargs) {
+    param_.InitAllowUnknown(kwargs);
+    indices_.resize(param_.length);
+    std::iota(std::begin(indices_), std::end(indices_), 0);  // fill like arange
+    out_.data.resize(2);  // label required by DataBatch, we can use fake label here
```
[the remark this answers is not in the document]
zhreshold: It turns out to be a python iterator limitation, not applicable to the DataBatch, so removed the hacky code, and the DataBatch is now good w/o label.

src/io/iter_sampler.cc
```diff
@@ -0,0 +1,183 @@
...
+  /*! \brief random magic number */
+  static const int kRandMagic = 2333;
+  /*! \brief Stored integer indices */
+  std::vector<int64_t> indices_;
+  /*! \brief current position for iteration */
+  std::size_t pos_;
+  /*! \brief data for next value */
+  DataInst out_;
+  /*! \brief random generator engine */
+  std::unique_ptr<common::RANDOM_ENGINE> rng_;
```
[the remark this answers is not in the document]
zhreshold: Done

python/mxnet/gluon/contrib/data/vision/dataloader.py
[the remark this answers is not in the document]
leezu: Why wouldn't they be happy, given that the composition of transforms appears straightforward?
Having a lot of wrappers around DataLoader will make it harder for us to maintain the codebase in the future, and may also confuse users as it's not clear to them that the manual composition approach is equivalent to using the wrapper. So I wonder if it's better to remove the wrapper and try keep our codebase shorter?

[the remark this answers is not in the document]
zhreshold: A 200 line of code can save each user 5min at least, especially for beginners. Actually it's not duplicating any code block and can handle most usecases for images. 

So I think it's worth the effort to put in `gluon.contrib` and get maintained.

src/io/dataset.cc
```diff
@@ -78,53 +74,44 @@ class RecordFileDataset final : public Dataset {
...
+          out.shape(), cpu::kDevMask, out.dtype(), 0),
+          &dst, Context::CPU(), Context::CPU(), rctx);
     }
     return true;
   }
 
  private:
+  using ReaderPtr = std::unique_ptr<dmlc::RecordIOReader>;
+  using StreamPtr = std::unique_ptr<dmlc::Stream>;
+  using RecordIOPair = dmlc::ThreadLocalStore<std::pair<StreamPtr, ReaderPtr> >;
```
leezu, line 111: ThreadLocalStore has some problems (https://github.com/dmlc/dmlc-core/issues/571), let's just use https://en.cppreference.com/w/cpp/keyword/thread_local ?

src/io/iter_sampler.cc
```diff
@@ -126,7 +126,9 @@ class RandomSampler : public IIterator<DataInst> {
     param_.InitAllowUnknown(kwargs);
     indices_.resize(param_.length);
     std::iota(std::begin(indices_), std::end(indices_), 0);  // fill like arange
-    rng_.reset(new common::RANDOM_ENGINE(kRandMagic + param_.seed));
+    mshadow::Random<cpu> *ctx_rng = ResourceManager::Get()->Request(
+      Context::CPU(), ResourceRequest::kRandom).get_random<cpu, real_t>(nullptr);
+    rng_.reset(new common::RANDOM_ENGINE(ctx_rng->GetSeed() + param_.seed));
```
leezu, line 130: Delete the user-facing `seed` parameter now that you're using the MXNet seed?

src/io/dataloader.cc
```diff
@@ -0,0 +1,188 @@
...
+  /*! \brief batchify function pointer.*/
+  std::intptr_t batchify_fn;
+  /*! \brief pin memory to device id.*/
+  int pin_device_id;
+  // declare parameters
+  DMLC_DECLARE_PARAMETER(ThreadedDataLoaderParam) {
+      DMLC_DECLARE_FIELD(num_workers).set_default(0)
+          .describe("Number of thread workers.");
+      DMLC_DECLARE_FIELD(dataset)
+          .describe("Number of thread workers.");
```
leezu, line 51: describe text is wrong

src/io/dataloader.cc
```diff
@@ -170,8 +164,7 @@ class ThreadedDataLoader : public IIterator<TBlobBatch> {
   /*! \brief batched buffer */
   std::vector<NDArray> batched_buffer_;
   /*! \brief pointer to dataset */
-  // std::shared_ptr<Dataset> dataset_;
-  std::vector<std::shared_ptr<Dataset>> datasets_;
+  std::shared_ptr<Dataset> dataset_;
```
leezu, line 167: Would we have some double free now? Because this shared_ptr would clean up the Dataset at the end of it's lifetime, and we also have a MXDatasetFree C API

leezu: Nevermind, the MXDatasetCreate actually creates a shared pointer on the heap. So we're just reusing that and there is no double free here. `*out = new std::shared_ptr<Dataset>(dataset); `

src/io/dataset.cc
zhreshold: Swapped with `thread_local`

src/io/dataloader.cc
zhreshold: Thanks for catching

src/io/iter_sampler.cc
zhreshold: Removed

PR #17842, opened 2020-03-16
CI: Attempt fixing illegal instruction errors

leezu:
Hypothesis: Caching the source-compiled variants of  libzstd1 and libb2 via our Docker cache can lead to "Illegal instruction" errors when deploying the docker image on an older machines. If so, switching to the distribution provided variants will solve the issue.

Follow-up on https://github.com/apache/incubator-mxnet/pull/17828

PR #17853, opened 2020-03-16
CI: switch to patched ccache 3 version

leezu:
## Description ##
ccache version introduced in #17828 causes illegal instruction errors on various builds. Most builds were fixed by #17842, but "Build / GPU: USE_INT64_TENSOR_SIZE" remains buggy and fails with illegal instruction errors.

Thus, let's switch back to ccache 3 and manually backport the cuda related fixes instead of switching to ccache 4.

### DIFF
--- a/ci/docker/install/deb_ubuntu_ccache.sh
+++ b/ci/docker/install/deb_ubuntu_ccache.sh
@@ -26,6 +26,7 @@ pushd .
 apt update || true
 apt install -y \
     autoconf \
+    gperf \
     xsltproc
 
 mkdir -p /work/deps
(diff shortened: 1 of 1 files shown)


ci/docker/install/deb_ubuntu_ccache.sh
```diff
@@ -50,15 +51,15 @@ export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
 git clone --recursive https://github.com/ccache/ccache.git
 cd ccache
-# Checkout a fixed & tested pre-release commit of ccache 4
-# ccache 4 contains fixes for caching nvcc output: https://github.com/ccache/ccache/pull/381
-git checkout 2e7154e67a5dd56852dae29d4c418d4ddc07c230
+git checkout v3.7.8
+# Backport cuda related fixes: https://github.com/ccache/ccache/pull/381
+git config user.name "MXNet CI"
```
marcoabreu: Can you do that at the place where git is installed?

leezu: No. I'm setting this only for the ccache repository.

leezu:
> It seems like we are doing open heart surgery here. Can we stop experimenting on prod and run a test setup instead? This seems be getting out of hand...

This PR essentially reverts the changes while backporting a few lines of code for fixes.

There has been a lot of unexpected Docker caching related failures over the last few months. This is just one of them. Maybe we should just get rid of it and reduce the docker container build time instead.. The caching is very fragile..

ci/docker/install/deb_ubuntu_ccache.sh
szha: I'd suggest having a patched fork somewhere and we can pick up ccache from that instead.

leezu: That doesn't differ substantially from the current practice (this PR) but will introduce a delay in fixing the problem in the order of 12-18 hours, as any change needs to pass CI again.

leezu: Any commiter has the permission to modify this PR. I don't mind any changes. Just the person modifying it should make sure to take 8 hours after change to retrigger the CI as often as needed.

PR #17860, opened 2020-03-18
libjpeg-turbo: Fix user-assisted execution of arbitrary code

leezu:
Reference https://www.openwall.com/lists/oss-security/2019/11/11/1


access2rohit:
Woah! That's a serious security risk you caught. BTW did you attempt to make a wheel after updating the dependency ?

leezu:
@access2rohit no, but it's tested by CI.

Issue #17818, opened 2020-03-11
RNN operator produces inconsistent gradients for h2h_bias for stacked RNNs


keerthanvasist:
The RNN operator produces different inconsistent values for gradient for h2h_bias in the topmost stack in different mxnet variants. 

This only happens when num_layers > 1. I compared the values after one forward pass and one backward pass. I compared between osx-cpu-mkl and linux-gpu-mkl (both built from source. Snapshot versions are not up-to-date, and the versions available are really flaky in this case). In the cpu variant, the gradient for is all zeros. 

These differences are leading to exception in DJL in CPU context. I also confirmed these differences in the values exist in Python. 

### Error Message
I don't see any error in Python but in DJL we have observed Nan values during the training process. 

## To Reproduce
I used the following script to look at gradients. 

```
import mxnet as mx
from mxnet import gpu, cpu, gluon
from mxnet import np, npx
from mxnet import autograd as ag
from mxnet.gluon import nn

npx.set_np()

def check_param(net):
    param_dict = net.collect_params()
    array = ('lstm0_{}{}_{}_{}'.format(d, l, g, t)
             for t in ['weight', 'bias']
             for l in range(2)
             for d in ['l', 'r'][:1]
             for g in ['i2h', 'h2h'])
    for key in array:
        param = param_dict[key]
        print("checking param: " + str(param))
        print("weight sum: " + str(param.data().sum()))
        print("weight mean: " + str(param.data().mean()))
        print("weight max: " + str(param.data().max()))
        print("weight min: " + str(param.data().min()))
        if param.grad_req != "null":
            print("checking the gradient of para: " + str(param))
            print("grad sum: " + str(param.grad().sum()))
            print("grad mean: " + str(param.grad().mean()))
            print("grad max: " + str(param.grad().max()))
            print("grad min: " + str(param.grad().min()))

def print_ndarray_stats(ndarray, name) :
    print("#####", name, "#####")
    print("checking " + name)
    print("sum: " + str(ndarray.sum()))
    print("mean: " + str(ndarray.mean()))
    print("max: " + str(ndarray.max()))
    print("min: " + str(ndarray.min()))
    print("Shape: " + str(ndarray.shape))

batch = 32
time = 28
channel = 28
state = 64
num_layers = 2

mx.random.seed(1234)
data = np.random.uniform(0, 10, size=(batch, time, channel))
mx.random.seed(1234)
labels = np.random.uniform(0, 1, size=(batch, time, state))

[...]

PR #17872, opened 2020-03-19
Fix issue of zeros gradients w.r.t. RNN bias when num_layers > 1

zixuanweeei:
## Description ##
Patch for the issue #17818. The rnn operator produces zero gradients for bias when num_layers > 1. It is caused by a mistake in calculating the shift of bias pointer, where we used the size of fusion bias (i2h_bias + h2h_bias) but MXNet gives twice (i2h_bias, h2h_bias) as many as the fusion size.

### Changes ###
- [x] Use the correct shift of bias pointer.
- [x] Change the way of sharing the same values of parameters of the fused RNN layers and the stacked one.
- [x] Add check for RNN output states.

### DIFF
--- a/python/mxnet/gluon/rnn/rnn_layer.py
+++ b/python/mxnet/gluon/rnn/rnn_layer.py
@@ -60,6 +60,7 @@ class _RNNLayer(HybridBlock):
         self._lstm_state_clip_nan = lstm_state_clip_nan
         self._dtype = dtype
         self._use_sequence_length = use_sequence_length
+        self.skip_states = None
 
         self._gates = {'rnn_relu': 1, 'rnn_tanh': 1, 'lstm': 4, 'gru': 3}[mode]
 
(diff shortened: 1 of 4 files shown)


ciyongch:
There's still some failure in fusedlstm tests. Please take a check.

stu1130:
@zixuanweeei Thanks for your contribution, could you also cherry-pick the commit to 1.7? DJL LSTM model depends on this commit. Thanks!

pengzhao-intel:
> @zixuanweeei Thanks for your contribution, could you also cherry-pick the commit to 1.7? DJL LSTM model depends on this commit. Thanks!

Sure, please add this requirement in 1.7 roadmap #16864

zixuanweeei:
CI passed. Please take a review. Thanks. @ciyongch @TaoLv @pengzhao-intel 

Besides, we will backport this patch into v1.7 branch as well @stu1130.

Issue #17875, opened 2020-03-19
USE_TVM_OP=1 build broken with DMLC_LOG_FATAL_THROW=0


leezu:
## Description
Setting `-DDMLC_LOG_FATAL_THROW=0` is useful for debugging with `gdb` or `lldb`. Unfortunately we can't compile MXNet with `USE_TVM_OP=1` in that case.

### Error Message
```
[658/658] Linking CXX shared library libmxnet.so
FAILED: libmxnet.so
[...]

PR #17878, opened 2020-03-20
Decouple LOG_FATAL_THROW prepocessor variables between TVM and MXNet

leezu:
## Description ##
Don't share DMLC_LOG_FATAL_THROW preprocessor variable as TVMOP build is not stable. Instead, use MXNET_LOG_FATAL_THROW for MXNet.

Workaround for https://github.com/apache/incubator-mxnet/issues/17875


leezu:
I updated the PR to preserve `DMLC_LOG_FATAL_THROW`, but only set it on the mxnet target.
Previously `DMLC_LOG_FATAL_THROW` was configured globally, affecting both tvm and mxnet build.

leezu:
@yzhliu I find that we always build TVM with OpenMP, even if OpenMP is disabled in MXNet. The comment

```
# Use OPENMP thread pool to be compatible with MXNet
set(USE_OPENMP ON)
```

suggests that is unintended. I added a fix to disable building TVM OpenMP if it's disabled on MXNet side.

yzhliu:
@leezu thanks for the finding and fix.

PR #17882, opened 2020-03-20
Improve performance of broadcast_axis on CPU

access2rohit:
## Description ##
Improves the performance of broadcast axis by reducing ALU operations and caching stride values. This operator is crucial in performance of SSD which gets slowed down by over 50%(total training time) when MXNet is built with Large Tensor Support enabled.

This PR leverages vectorization or SIMD architecture of the CPU. It is described as follows:

### Vectorization ### 
is the term for converting a scalar program to a vector program. Vectorized programs can run multiple operations from a single instruction(SIMD), whereas scalar can only operate on pairs of operands at once.

Consider the following very simple loop that adds the elements of two arrays and stores the results to a third array.
```
for (int i=0; i<16; ++i)
    C[i] = A[i] + B[i];
```
Vectorizing it, produces something like this:
```
for (int i=0; i<16; i+=4)
    C[i:i+3] = A[i:i+3] + B[i:i+3];
```
Note: The key here is that the elements need to be adjacent (so they can be loaded into vector registors. Note:avx 512 ISA does have "gatherd" intstruction for gathering sparse data from scattred locations into avx512 register but that operation is still slow compared to adjacent data being loaded into the register) and there should not be any fundamental dependency(or read after write dependency) in subsequent iterations of the loop. For e.g. the following loop cannot be vectorized because C[i+1] depends on the value of C[i] which needs to bew calculated first.
```
for (int i=0; i<16; ++i)
    C[i] = A[i] + C[i-1];
```
because if we unroll the loop and substitute values:
```
for (int i=0; i<16; i+=4) {
    C[i]   = A[i]   + C[i-1];
     |
     |-------------------------|
                               |
    C[i+1] = A[i+1] + C[i]; <--|
     |
     |---------------------------|
                                 |
    C[i+2] = A[i+2] + C[i+1]; <--|
     |
     |---------------------------|
                                 |
    C[i+3] = A[i+3] + C[i+2]; <--|
}
```
The arrows above show read after write dependency.the above statements cannot be executed in parallel by loading A[1:4] and C[1:4] into vector registers(here 1:4 is just for illustration purposes depending on the word size and register size it can load more data elements).

How is it different from loop unrolling? Unrolling transforms the loop into something that looks like this:
```
for (int i=0; i<16; i+=4) {
    C[i]   = A[i]   + B[i];
    C[i+1] = A[i+1] + B[i+1];
    C[i+2] = A[i+2] + B[i+2];
    C[i+3] = A[i+3] + B[i+3];
}
```

### Side Effects ###
for cases where 1 or very few elements need to be broadcasted to very large no. of locations the performance is not optimized. There is a slowdown when compared to master(int32 w/o large tensor support)

### Essentials ###

### Performance ###

Opperf code used to benchmark
```
add_res = run_performance_test(nd.broadcast_axis, run_backward=True, dtype='float32', ctx=mx.cpu(),
                               inputs=[{'data': (1000, 1, 100, 1), 'axis': (1, 3), 'size': (10, 5)}],
                               warmup=100, runs=1000, profiler='python')
print(add_res)

add_res = run_performance_test(nd.broadcast_axis, run_backward=True, dtype='float32', ctx=mx.cpu(),
                               inputs=[{'data': (1000, 1, 1, 100), 'axis': (1, 2), 'size': (10, 5)}],
                               warmup=100, runs=1000, profiler='python')
print(add_res)

add_res = run_performance_test(nd.broadcast_axis, run_backward=True, dtype='float32', ctx=mx.cpu(),
                               inputs=[{'data': (1, 1, 1, 1), 'axis': (0, 1, 2, 3), 'size': (1000, 10, 100, 5)}],
                               warmup=100, runs=1000, profiler='python')
print(add_res)

add_res = run_performance_test(nd.broadcast_axis, run_backward=True, dtype='float32', ctx=mx.cpu(),
                               inputs=[{'data': (1, 1000, 1, 100, 1), 'axis': (0, 2, 4), 'size': (2, 10, 5)}],
                               warmup=100, runs=1000, profiler='python')
print(add_res)
```

### Results ###

| code version | cases                                       | avg    | p50    | p90    |
|--------------|---------------------------------------------|--------|--------|--------|
|  master LT   | (1000, 1, 100, 1)->(1000, 10, 100, 5)       | 27.5   | 26.05  | 21.47  |
|              | (1000, 1, 1, 100)->(1000, 10, 5, 100)       | 29.68  | 28.59  | 35.93  |
|              | (1, 1, 1, 1)->(1000, 10, 100, 5)            | 129.80 | 123.30 | 169.04 |
|              | (1, 1000, 1, 100, 1)->(2, 1000, 10, 100, 5) | 63.60  | 58.88  | 74.74  |
|              |                                             |        |        |        |
| master no-LT | (1000, 1, 100, 1)->(1000, 10, 100, 5)       | 14.30  | 12.76  | 18.00  |
|              | (1000, 1, 1, 100)->(1000, 10, 5, 100)       | 13.21  | 12.33  | 15.67  |
|              | (1, 1, 1, 1)->(1000, 10, 100, 5)            | 66.80  | 60.77  | 83.80  |
|              | (1, 1000, 1, 100, 1)->(2, 1000, 10, 100, 5) | 31.81  | 29.71  | 40.23  |
|              |                                             |        |        |        |
| new LT       | (1000, 1, 100, 1)->(1000, 10, 100, 5)       | 17.53  | 17.36  | 22.42  |
|              | (1000, 1, 1, 100)->(1000, 10, 5, 100)       | 15.49  | 14.79  | 17.64  |
|              | (1, 1, 1, 1)->(1000, 10, 100, 5)            | 127.74 | 126.06 | 131.19 |
|              | (1, 1000, 1, 100, 1)->(2, 1000, 10, 100, 5) | 39.23  | 38.95  | 39.48  |
|              |                                             |        |        |        |
| new no-LT    | (1000, 1, 100, 1)->(1000, 10, 100, 5)       | 9.41   | 8.38   | 11.84  |
|              | (1000, 1, 1, 100)->(1000, 10, 5, 100)       | 8.29   | 7.65   | 10.61  |
|              | (1, 1, 1, 1)->(1000, 10, 100, 5)            | 67.54  | 63.63  | 86.17  |
|              | (1, 1000, 1, 100, 1)->(2, 1000, 10, 100, 5) | 23.57  | 23.39  | 23.92  |

### DIFF
--- a/src/operator/numpy/np_matmul_op-inl.h
+++ b/src/operator/numpy/np_matmul_op-inl.h
@@ -138,6 +138,7 @@ inline void MatmulImpl(const OpContext& ctx,
   mshadow::Tensor<xpu, 1, DType*> workspace;
   mshadow::Tensor<xpu, 3, DType> ans, mlhs, mrhs;
   mshadow::Stream<xpu> *s = ctx.get_stream<xpu>();
+  bool isCPU = std::is_same<xpu, cpu>::value;
   if (MatmulNeedBroadcast(a_shape, b_shape)) {
     // e.g. a.shape = (2, 3, 1, 4, 2)
     //      b.shape =       (5, 2, 4)
@@ -160,12 +161,21 @@ inline void MatmulImpl(const OpContext& ctx,
         struct ShapeAndStride aux_data_a, aux_data_b;
         PrepareAUXData(&aux_data_a, k_a_shape, k_a_shape_bc, ndim);
         PrepareAUXData(&aux_data_b, k_b_shape, k_b_shape_bc, ndim);
-        Kernel<broadcast_kernel<mshadow_op::identity>, xpu>::Launch(
-          s, bc_size_a, input_a.dptr<IType>(), bc_a_ptr,
-          aux_data_a, OpReqType::kWriteTo, ndim);
-        Kernel<broadcast_kernel<mshadow_op::identity>, xpu>::Launch(
-          s, bc_size_b, input_b.dptr<IType>(), bc_b_ptr,
-          aux_data_b, OpReqType::kWriteTo, ndim);
+        if (isCPU) {
+          Kernel<broadcast_kernel_cpu<mshadow_op::identity>, xpu>::Launch(
+            s, input_a.Size(), input_a.dptr<IType>(), bc_a_ptr,
+            aux_data_a, OpReqType::kWriteTo, ndim);
+          Kernel<broadcast_kernel_cpu<mshadow_op::identity>, xpu>::Launch(
+            s, input_b.Size(), input_b.dptr<IType>(), bc_b_ptr,
+            aux_data_b, OpReqType::kWriteTo, ndim);
+        } else {
+          Kernel<broadcast_kernel_gpu<mshadow_op::identity>, xpu>::Launch(
+            s, bc_size_a, input_a.dptr<IType>(), bc_a_ptr,
+            aux_data_a, OpReqType::kWriteTo, ndim);
+          Kernel<broadcast_kernel_gpu<mshadow_op::identity>, xpu>::Launch(
+            s, bc_size_b, input_b.dptr<IType>(), bc_b_ptr,
+            aux_data_b, OpReqType::kWriteTo, ndim);
+        }
       });
     });
     ans = mshadow::Tensor<xpu, 3, DType>(output.dptr<DType>(),
--- a/src/operator/numpy/np_matmul_op-inl.h
+++ b/src/operator/numpy/np_matmul_op-inl.h
@@ -139,6 +139,7 @@ inline void MatmulImpl(const OpContext& ctx,
   mshadow::Tensor<xpu, 3, DType> ans, mlhs, mrhs;
   mshadow::Stream<xpu> *s = ctx.get_stream<xpu>();
   bool isCPU = std::is_same<xpu, cpu>::value;
+  // Is true if either a or b requires broadcast or not
   if (MatmulNeedBroadcast(a_shape, b_shape)) {
     // e.g. a.shape = (2, 3, 1, 4, 2)
     //      b.shape =       (5, 2, 4)
@@ -162,12 +163,26 @@ inline void MatmulImpl(const OpContext& ctx,
         PrepareAUXData(&aux_data_a, k_a_shape, k_a_shape_bc, ndim);
         PrepareAUXData(&aux_data_b, k_b_shape, k_b_shape_bc, ndim);
         if (isCPU) {
-          Kernel<broadcast_kernel_cpu<mshadow_op::identity>, xpu>::Launch(
-            s, input_a.Size(), input_a.dptr<IType>(), bc_a_ptr,
-            aux_data_a, OpReqType::kWriteTo, ndim);
-          Kernel<broadcast_kernel_cpu<mshadow_op::identity>, xpu>::Launch(
-            s, input_b.Size(), input_b.dptr<IType>(), bc_b_ptr,
-            aux_data_b, OpReqType::kWriteTo, ndim);
+          if (!aux_data_a.shape_changed) {
+            Kernel<direct_copy<mshadow_op::identity>, xpu>::Launch(
+              s, bc_size_a, input_a.dptr<IType>(), bc_a_ptr, OpReqType::kWriteTo);
+            Kernel<broadcast_kernel_cpu<mshadow_op::identity>, xpu>::Launch(
+              s, input_b.Size(), input_b.dptr<IType>(), bc_b_ptr,
+              aux_data_b, OpReqType::kWriteTo, ndim);
+          } else if (!aux_data_b.shape_changed) {
+            Kernel<direct_copy<mshadow_op::identity>, xpu>::Launch(
+              s, bc_size_b, input_b.dptr<IType>(), bc_b_ptr, OpReqType::kWriteTo);
+            Kernel<broadcast_kernel_cpu<mshadow_op::identity>, xpu>::Launch(
+              s, input_a.Size(), input_a.dptr<IType>(), bc_a_ptr,
+              aux_data_a, OpReqType::kWriteTo, ndim);
+          } else {
+            Kernel<broadcast_kernel_cpu<mshadow_op::identity>, xpu>::Launch(
+              s, input_a.Size(), input_a.dptr<IType>(), bc_a_ptr,
+              aux_data_a, OpReqType::kWriteTo, ndim);
+            Kernel<broadcast_kernel_cpu<mshadow_op::identity>, xpu>::Launch(
+              s, input_b.Size(), input_b.dptr<IType>(), bc_b_ptr,
+              aux_data_b, OpReqType::kWriteTo, ndim);
+          }
         } else {
           Kernel<broadcast_kernel_gpu<mshadow_op::identity>, xpu>::Launch(
             s, bc_size_a, input_a.dptr<IType>(), bc_a_ptr,
--- a/src/operator/tensor/broadcast_reduce_op.h
+++ b/src/operator/tensor/broadcast_reduce_op.h
@@ -1085,7 +1085,7 @@ inline void PrepareAUXData(ShapeAndStride *aux_data,
 }
 }  // unnamed namespace
 
-template<typename OP>
+template<typename OP, typename IDXType = index_t>
 struct broadcast_kernel {
   template<typename IType, typename OType>
   MSHADOW_XINLINE static void Map(index_t i,
@@ -1094,15 +1094,16 @@ struct broadcast_kernel {
                                   const ShapeAndStride& aux_data,
                                   const OpReqType req,
                                   const int ndim) {
-    index_t idx = i;
-    index_t in_idx = i;
+    printf("size of IDXType=%d",sizeof(IDXType));
+    IDXType idx = i;
+    IDXType in_idx = i;
 #pragma unroll 4
-    for (int iter = ndim - 1; iter >= 0; --iter) {
-      index_t out_dim_shape = aux_data.output_shape[iter];
-      index_t out_dim_stride = aux_data.out_stride[iter];
+    for (IDXType iter = ndim - 1; iter >= 0; --iter) {
+      IDXType out_dim_shape = aux_data.output_shape[iter];
+      IDXType out_dim_stride = aux_data.out_stride[iter];
       // x % y = x - (x / y) * y
       // speeds up modulo(%) operation in GPU
-      index_t dim_idx = idx - (idx / out_dim_shape) * out_dim_shape;
+      IDXType dim_idx = idx - (idx / out_dim_shape) * out_dim_shape;
       if (aux_data.input_shape[iter] != 1) {
         in_idx += dim_idx * (aux_data.in_stride[iter] - out_dim_stride);
       } else {
@@ -1147,16 +1148,26 @@ inline void BroadcastComputeImpl(const nnvm::NodeAttrs& attrs,
           outputs[0].get_with_shape<xpu, 2, OType>(dst_shape.get<2>(), s);
         Tensor<xpu, 2, IType> data =
           inputs[0].get_with_shape<xpu, 2, IType>(src_shape.get<2>(), s);
-        Kernel<broadcast_kernel<mshadow_op::identity>, xpu>::Launch(
-          s, out.shape_.Size(), data.dptr_, out.dptr_, aux_data, req[0], 2);
+        if (ctx.run_ctx.get_ctx().dev_type == Context::kGPU) {
+          Kernel<broadcast_kernel<mshadow_op::identity, int>, xpu>::Launch(
+            s, out.shape_.Size(), data.dptr_, out.dptr_, aux_data, req[0], 2);
+	} else {
+          Kernel<broadcast_kernel<mshadow_op::identity>, xpu>::Launch(
+            s, out.shape_.Size(), data.dptr_, out.dptr_, aux_data, req[0], 2);
+	}
       } else {
         const int ndim = MXNET_SPECIAL_MAX_NDIM;
         Tensor<xpu, ndim, OType> out =
           outputs[0].get_with_shape<xpu, ndim, OType>(dst_shape.get<ndim>(), s);
         Tensor<xpu, ndim, IType> data =
           inputs[0].get_with_shape<xpu, ndim, IType>(src_shape.get<ndim>(), s);
-        Kernel<broadcast_kernel<mshadow_op::identity>, xpu>::Launch(
-          s, out.shape_.Size(), data.dptr_, out.dptr_, aux_data, req[0], ndim);
+        if (ctx.run_ctx.get_ctx().dev_type == Context::kGPU) {
+          Kernel<broadcast_kernel<mshadow_op::identity, int>, xpu>::Launch(
+            s, out.shape_.Size(), data.dptr_, out.dptr_, aux_data, req[0], ndim);
+	} else {
+          Kernel<broadcast_kernel<mshadow_op::identity>, xpu>::Launch(
+            s, out.shape_.Size(), data.dptr_, out.dptr_, aux_data, req[0], 2);
+	}
       }
     });
   });
--- a/src/operator/tensor/broadcast_reduce_op.h
+++ b/src/operator/tensor/broadcast_reduce_op.h
@@ -1049,6 +1049,7 @@ void ReduceAxesBackwardUseInOut(const nnvm::NodeAttrs& attrs,
   ReduceAxesBackwardUseInOutImpl<xpu, OP, normalize>(ctx, small, inputs, req, outputs);
 }
 
+<<<<<<< HEAD
 namespace {  // unnamed namespace to keep scope of the struct within the file
 struct ShapeAndStride {
   index_t in_stride[MXNET_SPECIAL_MAX_NDIM];
@@ -1087,12 +1088,24 @@ inline void PrepareAUXData(ShapeAndStride *aux_data,
 
 template<typename OP, typename IDXType = index_t>
 struct broadcast_kernel {
+=======
+/**
+ * Changed the thread workload mapping from 1
+ * thread/output element to 1 thread/input to be broadcasted
+ * This approach leverages vectorization when fastest varying
+ * index(stride=1) of the tensor is to be broadcasted.
+ * In other cases it simply performs better by better load balancing.
+ */
+template<typename OP>
+struct broadcast_kernel_cpu {
+>>>>>>> 9919eff50... adding comments explaining code optimizations
   template<typename IType, typename OType>
   MSHADOW_XINLINE static void Map(index_t i,
                                   IType *input,
                                   OType *output,
                                   const ShapeAndStride& aux_data,
                                   const OpReqType req,
+<<<<<<< HEAD
                                   const int ndim) {
     printf("size of IDXType=%d",sizeof(IDXType));
     IDXType idx = i;
@@ -1112,11 +1125,65 @@ struct broadcast_kernel {
       idx /= out_dim_shape;
     }
     KERNEL_ASSIGN(output[i], req, OP::Map(input[in_idx]));
+=======
+                                  const uint32_t ndim,
+                                  const size_t *axes,
+                                  const size_t *out_stride,
+                                  const int num_broadcast_axes) {
+        index_t idx = i;
+        index_t init_off = 0;
+        for (int iter = ndim - 1; idx > 0 && iter >= 0; --iter) {
+          size_t dim_idx = idx % in_shape[iter];
+          init_off += dim_idx * out_stride[iter];
+          idx /= in_shape[iter];
+        }
+        index_t stride_0, stride_1, stride_2;
+        // Each case is based on the number of axis to be broadcasted
+        // (1, 2 or 3) after merging axes.
+        switch (num_broadcast_axes) {
+          // when input shape is amogst one of the form
+          // [(x,1), (x,1,x), (1,x)]
+          // x can be any +ve number >=0 and they need not be equal to each other
+          case 1 :
+            stride_0 = out_stride[axes[0]];
+            for (int l=0; l < out_shape[axes[0]]; l++) {
+              KERNEL_ASSIGN(output[init_off + l*stride_0],
+                  req, OP::Map(input[i]));
+            }
+            break;
+          // when input shape is amogst one of the form
+          // [(x,1,x,1), (1,x,1,x), (x,1,x,1,x)]
+          // x can be any +ve number >1 or =0(the axis ) and they need not be equal to each other
+          case 2:
+            stride_1 = out_stride[axes[1]], stride_0 = out_stride[axes[0]];
+            for (int k=0; k < out_shape[axes[1]]; k++) {
+              for (int l=0; l < out_shape[axes[0]]; l++) {
+                KERNEL_ASSIGN(output[init_off + k*stride_1 + l*stride_0],
+                    req, OP::Map(input[i]));
+              }
+            }
+            break;
+          // when input shape is of the form [(1,x,1,x,1)] and
+          // x can be any +ve number >=0 and they need not be equal to each other
+          case 3:
+            stride_2 = out_stride[axes[2]], stride_1 = out_stride[axes[1]];
+            stride_0 = out_stride[axes[0]];
+            for (int j=0; j < out_shape[axes[2]]; j++) {
+              for (int k=0; k < out_shape[axes[1]]; k++) {
+                for (int l=0; l < out_shape[axes[0]]; l++) {
+                  KERNEL_ASSIGN(output[init_off + j*stride_2 + k*stride_1 + l*stride_0],
+                      req, OP::Map(input[i]));
+                }
+              }
+            }
+            break;
+        }
+>>>>>>> 9919eff50... adding comments explaining code optimizations
   }
 };
 
 template<typename xpu>
-inline void BroadcastComputeImpl(const nnvm::NodeAttrs& attrs,
+inline void BroadcastComputeImplCpu(const nnvm::NodeAttrs& attrs,
                                  const OpContext& ctx,
                                  const std::vector<TBlob>& inputs,
                                  const std::vector<OpReqType>& req,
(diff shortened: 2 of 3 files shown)


ptrendx:
Is this going to concentrate on CPU performance? GPU performance of broadcast_axis/like/etc is really poor too (few days ago I was measuring it and got ~150 GB/s on V100 GPU out of 900 GB/s peak BW.

ptrendx:
The biggest speedup you can get here would be from vectorization, so that you don't need to do those index calculations all the time (just once per vector).

leezu:
@access2rohit can you rebase on latest master to see if that fixes CI?

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1103,21 +1132,52 @@ inline void BroadcastComputeImpl(const nnvm::NodeAttrs& attrs,
...
+          }
       } else {
         const int ndim = MXNET_SPECIAL_MAX_NDIM;
         Tensor<xpu, ndim, OType> out =
           outputs[0].get_with_shape<xpu, ndim, OType>(dst_shape.get<ndim>(), s);
         Tensor<xpu, ndim, IType> data =
           inputs[0].get_with_shape<xpu, ndim, IType>(src_shape.get<ndim>(), s);
-        Kernel<broadcast_kernel<mshadow_op::identity>, xpu>::Launch(
-          s, out.shape_.Size(), data.dptr_, out.dptr_, in_shape, out_shape, req[0], ndim);
+          if (!enable_lt) {
```
apeforest, line 1171: Can we remove this to see how much perf gain we have with the operator speedup alone?

access2rohit:
> Could we also take Da's suggestion by using int32_t for GPU kernels?

@apeforest 
For this op only or all ? For this op only I can take a look.
If for all ops I would like to do that in a separate PR.

src/operator/tensor/broadcast_reduce_op.h
access2rohit: ok

access2rohit:
> @access2rohit can you rebase on latest master to see if that fixes CI?

Was still failing after rebase. I have fixed the failing case and pushed new changes now.

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1058,22 +1058,50 @@ struct broadcast_kernel {
...
-      }
-      idx /= out_shape[iter];
-      in_stride *= in_shape[iter];
-      out_stride *= out_shape[iter];
-    }
-    KERNEL_ASSIGN(output[i], req, OP::Map(input[in_idx]));
+                                  const uint32_t ndim,
+                                  const uint64_t *axes,
+                                  const uint64_t *out_stride,
+                                  const size_t no_axes) {
```
apeforest, line 1064: no_axes is a confusing name. Please use `num_axes`

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1058,22 +1058,50 @@ struct broadcast_kernel {
...
+        index_t idx = i;
+        index_t init_off = 0;
+        for (int iter = ndim - 1; idx > 0 && iter >= 0; --iter) {
+          size_t dim_idx = idx % in_shape[iter];
+          init_off += dim_idx * out_stride[iter];
+          idx /= in_shape[iter];
+        }
+        index_t stride_0, stride_1, stride_2;
+        switch (no_axes) {
```
apeforest, line 1073: Are we only supporting max of 3 axes here?

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1090,6 +1118,7 @@ inline void BroadcastComputeImpl(const nnvm::NodeAttrs& attrs,
   mxnet::TShape src_shape, dst_shape;
   BroadcastReduceShapeCompact(outputs[0].shape_, small, &dst_shape, &src_shape);
   Stream<xpu> *s = ctx.get_stream<xpu>();
+  bool enable_lt = dst_shape.Size() > INT_MAX;
```
apeforest, line 1121: Let's remove this flag and check the performance gain. If indeed we need to handle int64 case differently we need a more holistic design.

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1103,21 +1132,55 @@ inline void BroadcastComputeImpl(const nnvm::NodeAttrs& attrs,
           out_shape[i] = 1;
         }
       }
-      if (dst_shape.ndim() == 2) {
+      uint64_t axes[dst_shape.ndim()], out_stride[dst_shape.ndim()];
+      int iter = dst_shape.ndim() - 1, i = 0;
+      bool flag = false;
```
apeforest, line 1137: what does the flag here mean? Please use a more informative name.

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1103,21 +1132,55 @@ inline void BroadcastComputeImpl(const nnvm::NodeAttrs& attrs,
           out_shape[i] = 1;
         }
       }
-      if (dst_shape.ndim() == 2) {
+      uint64_t axes[dst_shape.ndim()], out_stride[dst_shape.ndim()];
```
apeforest, line 1135: why use uint64_t here?

apeforest:
Please add more details in the PR description to explain the rationale of this change:
(1) Why make the change
(2) How this change improved performance
(3) Any side effect

access2rohit:
> change

@apeforest 
1) Its there in the PR Headiong
2) Leverages vectorization. Can't explain line by line as to how. Too much text to explain. I can share a link about vectorization for you to read.
3) I didn't understand this question. Could you clarify what meant by side-effect.

src/operator/tensor/broadcast_reduce_op.h
access2rohit: ok ... will do

src/operator/tensor/broadcast_reduce_op.h
access2rohit: max only 3 broadcast_axes are possible, since all consecutive broadcast and non-broadcast axes are merged together and we only support total of 5 Dimension are merging axes.

src/operator/tensor/broadcast_reduce_op.h
access2rohit: will do

src/operator/tensor/broadcast_reduce_op.h
access2rohit: this flag ensure if broadcast_axis is supplied with input where input shape = output shape(i.e. no broadcast). In such a case we are simply invoking copy

src/operator/tensor/broadcast_reduce_op.h
access2rohit: will use `size_t` instead

src/operator/tensor/broadcast_reduce_op.h
access2rohit: Here are the results after removing typedef and enable_lt conditional check

| code version | cases                                       | avg    | p50    | p90    |
|--------------|---------------------------------------------|--------|--------|--------|
|  master LT   | (1000, 1, 100, 1)->(1000, 10, 100, 5)       | 28.89   | 26.91  | 34.51  |
|              | (1000, 1, 1, 100)->(1000, 10, 5, 100)       | 28.70  | 26.45  | 34.03  |
|              | (1, 1, 1, 1)->(1000, 10, 100, 5)            | 123.14 | 121.88 | 125.35 |
|              | (1, 1000, 1, 100, 1)->(2, 1000, 10, 100, 5) | 65.18  | 61.52  | 74.58  |
|              |                                             |        |        |        |
| master no-LT | (1000, 1, 100, 1)->(1000, 10, 100, 5)       | 13.82  | 13.24  | 16.83  |
|              | (1000, 1, 1, 100)->(1000, 10, 5, 100)       | 13.46  | 13.00  | 16.18  |
|              | (1, 1, 1, 1)->(1000, 10, 100, 5)            | 52.38  | 51.36  | 53.51  |
|              | (1, 1000, 1, 100, 1)->(2, 1000, 10, 100, 5) | 35.29  | 38.65  | 40.23  |
|              |                                             |        |        |        |
| new LT       | (1000, 1, 100, 1)->(1000, 10, 100, 5)       | 15.91  | 15.79  | 18.23  |
|              | (1000, 1, 1, 100)->(1000, 10, 5, 100)       | 16.44  | 17.50  | 17.97  |
|              | (1, 1, 1, 1)->(1000, 10, 100, 5)            | 124.17 | 125.65 | 126.42 |
|              | (1, 1000, 1, 100, 1)->(2, 1000, 10, 100, 5) | 37.33  | 36.76  | 38.07  |
|              |                                             |        |        |        |
| new no-LT    | (1000, 1, 100, 1)->(1000, 10, 100, 5)       | 9.76   | 9.91   | 11.98  |
|              | (1000, 1, 1, 100)->(1000, 10, 5, 100)       | 9.04   | 9.54   | 9.96  |
|              | (1, 1, 1, 1)->(1000, 10, 100, 5)            | 58.40  | 60.54  | 61.01  |
|              | (1, 1000, 1, 100, 1)->(2, 1000, 10, 100, 5) | 23.45  | 23.11  | 26.79  |

access2rohit:
@ptrendx This is mainly optimizing for CPU. Can you review again?

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1058,22 +1065,58 @@ struct broadcast_kernel {
...
-      in_idx -= dim_idx * out_stride;
-      if (in_shape[iter] != 1) {
-        in_idx += dim_idx * in_stride;
-      }
-      idx /= out_shape[iter];
-      in_stride *= in_shape[iter];
-      out_stride *= out_shape[iter];
-    }
-    KERNEL_ASSIGN(output[i], req, OP::Map(input[in_idx]));
+                                  const uint32_t ndim,
```
apeforest, line 1068: shall we use `int` for ndim as we do for all other places?

src/operator/numpy/np_matmul_op-inl.h
```diff
@@ -157,12 +157,50 @@ inline void MatmulImpl(const OpContext& ctx,
...
-        Kernel<broadcast_kernel<mshadow_op::identity>, xpu>::Launch(
-          s, bc_size_a, input_a.dptr<IType>(), bc_a_ptr,
-          k_a_shape, k_a_shape_bc, OpReqType::kWriteTo, ndim);
-        Kernel<broadcast_kernel<mshadow_op::identity>, xpu>::Launch(
-          s, bc_size_b, input_b.dptr<IType>(), bc_b_ptr,
-          k_b_shape, k_b_shape_bc, OpReqType::kWriteTo, ndim);
+        bool a_shape_changed = false, b_shape_changed = false;
+        size_t axes[ndim], out_stride[ndim], axes_b[ndim], out_stride_b[ndim];
+        int iter = ndim - 1, idx_a = 0, idx_b = 0;
+        out_stride[iter] = 1;
```
apeforest, line 163: should this be named out_stride_a?

src/operator/numpy/np_matmul_op-inl.h
```diff
@@ -157,12 +157,50 @@ inline void MatmulImpl(const OpContext& ctx,
...
+        out_stride_b[iter] = 1;
+        if (k_a_shape[iter] != k_a_shape_bc[iter]) {
```
apeforest, line 165: Can these two if blocks be included in the for loop below?

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1058,22 +1065,58 @@ struct broadcast_kernel {
...
+          init_off += dim_idx * out_stride[iter];
+          idx /= in_shape[iter];
+        }
+        index_t stride_0, stride_1, stride_2;
+        // Each case is based on the number of axis to be broadcasted
+        // (1, 2 or 3) after merging axes.
+        switch (num_broadcast_axes) {
+          // when input shape is amogst one of the form
+          // [(x,1), (x,1,x), (1,x)]
+          // x can be any +ve number >=0 and they need not be equal to each other
```
apeforest, line 1085: what does +ve mean?

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1058,22 +1065,58 @@ struct broadcast_kernel {
...
+          size_t dim_idx = idx % in_shape[iter];
+          init_off += dim_idx * out_stride[iter];
+          idx /= in_shape[iter];
+        }
+        index_t stride_0, stride_1, stride_2;
+        // Each case is based on the number of axis to be broadcasted
+        // (1, 2 or 3) after merging axes.
+        switch (num_broadcast_axes) {
+          // when input shape is amogst one of the form
+          // [(x,1), (x,1,x), (1,x)]
```
apeforest, line 1084: No need to use [] in the comment.

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1058,22 +1065,58 @@ struct broadcast_kernel {
...
+            for (int j=0; j < out_shape[axes[2]]; j++) {
+              for (int k=0; k < out_shape[axes[1]]; k++) {
+                for (int l=0; l < out_shape[axes[0]]; l++) {
+                  KERNEL_ASSIGN(output[init_off + j*stride_2 + k*stride_1 + l*stride_0],
+                      req, OP::Map(input[i]));
+                }
+              }
+            }
+            break;
+        }
```
apeforest, line 1119: Add a default case to error out.

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1058,22 +1065,58 @@ struct broadcast_kernel {
...
+                                  const int num_broadcast_axes) {
+        index_t idx = i;
+        index_t init_off = 0;
+        for (int iter = ndim - 1; idx > 0 && iter >= 0; --iter) {
+          size_t dim_idx = idx % in_shape[iter];
+          init_off += dim_idx * out_stride[iter];
+          idx /= in_shape[iter];
+        }
+        index_t stride_0, stride_1, stride_2;
+        // Each case is based on the number of axis to be broadcasted
```
apeforest, line 1080: Again, could you please explain why there are only 3 cases regarding the number of axes?

apeforest:
> > Please add more details in the PR description to explain the rationale of this change:
> 
> @apeforest
> 
> 1. Its there in the PR Heading. Additionally, its being done to improve BERT performance regression when using Large Tensor Build
> 2. Leverages vectorization. Can't explain line by line as to how. Too much text to write. I can share a link about vectorization for reviewers to read. Let me know if you meant something else. I can write 1 line each for 3 cases which is necessary
> 3. I didn't understand this question. Could you clarify what meant by side-effect.

1. The title tells readers this PR will change the performance of broadcast_axis. However, I am not familiar with which tricks/algorithms you employed to make it faster. It would greatly help people understand code if you could mention it in PR description with a few sentences or a pseudo code.

2. vectorization is a very general term. Please add more details here. You could even copy some of them from your code comment

3. Side effect means any downside or potential problem will this change. e.g. slow down with other inputs, or not supporting certain inputs, backward compatibility etc.

access2rohit:
@apeforest 
SSD 1 epoch time master(int32 w/o large tensor) :                                     230sec
SSD 1 epoch time optimized broadcast_axis for CPU(int64 large tensor) : 190sec

Still unable to fix the case:
(1, 1, 1, 1)->(1000, 10, 100, 5): 125 sec as compared to 65 sec

src/operator/tensor/broadcast_reduce_op.h
access2rohit: WIll update will default choice. Though it should ideally be unreachable.

access2rohit:
Code | SSD 1 Epoch time (sec) | %age Speedup/Slowdown w.r.t Master (large tensor disabled)
-- | -- | --
Master (large tensor disabled) | 226 | 0
Master (large tensor   enabled) | 335 | 48.23% slowdown
Master + CPU Optimized broadcast_axis (large tensor disabled) | 130 | 42.5% speedup
Master + CPU Optimized broadcast_axis   (large tensor enabled) | 184 | 18.5% speedup

access2rohit:
@mxnet-bot run ci [all]

access2rohit:
@apeforest I am investigating the windows CI timeout on my PR. In the meantime can you review ?

src/operator/tensor/broadcast_reduce_op.h
access2rohit: Deafult case with LOG(FATAL) will cause build failure since `cout` cannot be used inside device code. This is due to the way kernel launches are written

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1147,16 +1249,26 @@ inline void BroadcastComputeImpl(const nnvm::NodeAttrs& attrs,
...
       } else {
         const int ndim = MXNET_SPECIAL_MAX_NDIM;
         Tensor<xpu, ndim, OType> out =
           outputs[0].get_with_shape<xpu, ndim, OType>(dst_shape.get<ndim>(), s);
         Tensor<xpu, ndim, IType> data =
           inputs[0].get_with_shape<xpu, ndim, IType>(src_shape.get<ndim>(), s);
-        Kernel<broadcast_kernel<mshadow_op::identity>, xpu>::Launch(
-          s, out.shape_.Size(), data.dptr_, out.dptr_, aux_data, req[0], ndim);
+        if (isCPU) {
+          Kernel<broadcast_kernel_cpu<mshadow_op::identity>, xpu>::Launch(
```
apeforest, line 1266: If it's only CPU you don't need to use the generic kernel, right? You can just use the Kernel<OP, cpu> template.

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1147,16 +1249,26 @@ inline void BroadcastComputeImpl(const nnvm::NodeAttrs& attrs,
           outputs[0].get_with_shape<xpu, 2, OType>(dst_shape.get<2>(), s);
         Tensor<xpu, 2, IType> data =
           inputs[0].get_with_shape<xpu, 2, IType>(src_shape.get<2>(), s);
-        Kernel<broadcast_kernel<mshadow_op::identity>, xpu>::Launch(
-          s, out.shape_.Size(), data.dptr_, out.dptr_, aux_data, req[0], 2);
+        if (isCPU) {
+          Kernel<broadcast_kernel_cpu<mshadow_op::identity>, xpu>::Launch(
+            s, data.shape_.Size(), data.dptr_, out.dptr_, aux_data, req[0], 2);
+        } else {
+          Kernel<broadcast_kernel<mshadow_op::identity>, xpu>::Launch(
```
apeforest, line 1256: Maybe rename broadcast_kernel to broadcast_kernel_gpu for better readability?

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1125,8 +1221,14 @@ inline void BroadcastComputeImpl(const nnvm::NodeAttrs& attrs,
...
   using namespace mxnet_op;
   mxnet::TShape src_shape, dst_shape;
+  // combines 2 or more consecutive broadcast/non-broadcast axes together
+  // e.g. (3,4,1,1,5,1,6,7) (2,3,5) (5,10,9) -> (3*4,1*1,5,1,6*7) (1,3) (5*10, 9)
+  //      -> (12,1,5,1,42) (1,3) (50, 9)
+  //      and this is the new input for broadcast_kernel whose total
+  //      num of dimensions cannot be greater than 5(throws an error otherwise).
   BroadcastReduceShapeCompact(outputs[0].shape_, small, &dst_shape, &src_shape);
   Stream<xpu> *s = ctx.get_stream<xpu>();
+  bool isCPU = ctx.run_ctx.get_ctx().dev_type == Context::kCPU;
```
apeforest, line 1231: You can just use `if (std::is_same<xpu, cpu>::value)` right?

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1114,6 +1124,92 @@ struct broadcast_kernel {
...
+      // x_1, x_2 are size of the dimensions that are not to be broadcasted
+      // here the last axis which is [4] is the one where compiler can vectorize
+      // the code the outer 2 loops improve preformance by avoiding
+      // duplicate stride calculations
+      // for each output location input[i] needs to be written to.
+      case 3:
+        stride_2 = aux_data.out_stride[aux_data.axes[2]];
+        stride_1 = aux_data.out_stride[aux_data.axes[1]];
+        stride_0 = aux_data.out_stride[aux_data.axes[0]];
+        for (index_t j=0; j < aux_data.output_shape[aux_data.axes[2]]; j++) {
```
apeforest, line 1193: nit: add space between "="

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1114,6 +1124,92 @@ struct broadcast_kernel {
...
+      // x_1, x_2, x_3 are size of the dimensions that are not to be broadcasted
+      // in the inner most loop can be vectorized by compiler in outer loops
+      // the performance is improved due avoidance of duplicate stride calculations
+      // for each output location input[i] needs to be written to.
+      case 2:
+        stride_1 = aux_data.out_stride[aux_data.axes[1]];
+        stride_0 = aux_data.out_stride[aux_data.axes[0]];
+        for (index_t k=0; k < aux_data.output_shape[aux_data.axes[1]]; k++) {
+          for (index_t l=0; l < aux_data.output_shape[aux_data.axes[0]]; l++) {
+            KERNEL_ASSIGN(output[init_off + k*stride_1 + l*stride_0],
```
apeforest, line 1178: nit: space round '*'

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1114,6 +1124,92 @@ struct broadcast_kernel {
...
+    switch (aux_data.num_broadcast_axes) {
+      // when input shape is one of the follwing forms
+      // (x_1,1) or (x_1,1,x_2) or (1,x_1)
+      // x_1, x_2 are size of the dimensions that are not to be broadcasted
+      // in case of (x_1,1) the system leverages vectorization but in other 2
+      // the performance is improved due avoidance of duplicate stride calculations
+      // for each output location input[i] needs to be written to.
+      case 1 :
+        stride_0 = aux_data.out_stride[aux_data.axes[0]];
+        for (index_t l=0; l < aux_data.output_shape[aux_data.axes[0]]; l++) {
```
apeforest, line 1162: nit: space around '='

src/operator/tensor/broadcast_reduce_op.h
```diff
@@ -1114,6 +1124,92 @@ struct broadcast_kernel {
...
+      // when input shape is one of the follwing forms
+      // (x_1,1,x_2,1) or (1,x_1,1,x_2) or (x_1,1,x_2,1,x_3)
+      // x_1, x_2, x_3 are size of the dimensions that are not to be broadcasted
+      // in the inner most loop can be vectorized by compiler in outer loops
+      // the performance is improved due avoidance of duplicate stride calculations
+      // for each output location input[i] needs to be written to.
+      case 2:
+        stride_1 = aux_data.out_stride[aux_data.axes[1]];
+        stride_0 = aux_data.out_stride[aux_data.axes[0]];
+        for (index_t k=0; k < aux_data.output_shape[aux_data.axes[1]]; k++) {
```
apeforest, line 1176: nit: space around '='

src/operator/tensor/broadcast_reduce_op.h
apeforest: In that case, could you add an assertion around line: https://github.com/apache/incubator-mxnet/pull/17882/files#diff-b669acf192ce8ff1658871e6d9dcca5dR1095

apeforest: i.e. aux_data->num_broadcast_axes = i;

src/operator/tensor/broadcast_reduce_op.h
access2rohit: good point .. let me try it

src/operator/tensor/broadcast_reduce_op.h
access2rohit: Kernel<broadcast_kernel_cpu<mshadow_op::identity>, cpu> actually causes build failure and gives the following error:
```
../src/operator/nn/./../tensor/broadcast_reduce_op.h(1246): error: no instance of overloaded function "mxnet::op::mxnet_op::Kernel<OP, mxnet::cpu>::Launch [with OP=mxnet::op::broadcast_kernel_cpu<mxnet::op::mshadow_op::identity>]" matches the argument list
            argument types are: (mshadow::Stream<mshadow::gpu> *, mshadow::index_t, float *, float *, mxnet::op::<unnamed>::ShapeAndStride, const mxnet::OpReqType, int)
          detected during:
            instantiation of "void mxnet::op::BroadcastComputeImpl<xpu>(const nnvm::NodeAttrs &, const mxnet::OpContext &, const std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob>> &, const std::vector<mxnet::OpReqType, std::allocator<mxnet::OpReqType>> &, const std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob>> &, const mxnet::TShape &) [with xpu=mxnet::gpu]"
../src/operator/nn/./moments-inl.h(215): here
            instantiation of "void mxnet::op::MomentsBackwardImpl<xpu>(const nnvm::NodeAttrs &, const mxnet::OpContext &, const std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob>> &, const std::vector<mxnet::OpReqType, std::allocator<mxnet::OpReqType>> &, const std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob>> &, const dmlc::optional<mxnet::TShape> &) [with xpu=mxnet::gpu]"
../src/operator/nn/./moments-inl.h(249): here
            instantiation of "void mxnet::op::MomentsBackward<xpu>(const nnvm::NodeAttrs &, const mxnet::OpContext &, const std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob>> &, const std::vector<mxnet::OpReqType, std::allocator<mxnet::OpReqType>> &, const std::vector<mxnet::TBlob, std::allocator<mxnet::TBlob>> &) [with xpu=mxnet::gpu]"
../src/operator/nn/moments.cu(36): here
```

src/operator/tensor/broadcast_reduce_op.h
access2rohit: Done!

access2rohit:
@mxnet-bot run ci [unix-cpu, clang, website, centos-cpu, edge, miscellaneous, unix-gpu, windows-gpu, centos-gpu, windows-cpu]

access2rohit:

access2rohit:
@apeforest can you review. I have addressed your comments

access2rohit:
@zheng-da Can you please review?

access2rohit:

PR #15981, opened 2019-08-23
Disable test coverage of C++ codebase on CI 

stu1130:
## Description ##
Related to #15971
### Root Cause
When linker link the object files, it uses relative addressing method. The error happens when the project is too large for linker to link the target objects using relative addressing with 32-bit address.
### Things we have tried
1. add -fPIC flag to make sure the shared library can be loaded at any address in memory. It end up with the same error message.
2. add -mcmodel=large. GCC default use -mcmodel=small where the program and symbols must be linked within 2 GB which is 32 bit address space. When it is set to large, there is no 2GB limitation. Unfortunately, all the code MXNet statically link including MKLDNN will need to build this way. It also ended up with same failure. So it leaded to our final solution.
3. Removing the test_coverage flag which injects monitoring code every if else branch, while loop, etc.It dramatically increase the size of libmxnet.a with test coverage. After removing the test coverage flag, it passed the Clang MKLDNN CI job.

After discussion with @marcoabreu, we agreed to disable C++ backend test coverage monitoring, which will also boost the CI.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at http://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

## Comments ##
@marcoabreu could you help check if some runtime functions are missing or some functions should not disable? Thanks

PR #17889, opened 2020-03-23
Re-enable coverage on CI for GCC CPU builds

leezu:
As title. Previously https://github.com/apache/incubator-mxnet/pull/15981/ disabled all code coverage builds due to library size limitations on some builds. Let's try to enable code coverage again on the builds not limited by size.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,7 +68,7 @@ option(INSTALL_EXAMPLES "Install the example source files." OFF)
 option(USE_SIGNAL_HANDLER "Print stack traces on segfaults." ON)
 option(USE_TENSORRT "Enable inference optimization with TensorRT." OFF)
 option(USE_ASAN "Enable Clang/GCC ASAN sanitizers." OFF)
-option(ENABLE_TESTCOVERAGE "Enable compilation with test coverage metric output" OFF)
+cmake_dependent_option(ENABLE_TESTCOVERAGE "Enable compilation with test coverage metric output" OFF "NOT MSVC" OFF)
 option(USE_INT64_TENSOR_SIZE "Use int64_t to represent the total number of elements in a tensor" OFF)
 option(BUILD_CYTHON_MODULES "Build cython modules." OFF)
 option(LOG_FATAL_THROW "Log exceptions but do not abort" ON)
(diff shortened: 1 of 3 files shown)


leezu:

leezu:

leezu:

leezu:
@marcoabreu would it make sense to add a coverage check to the status checks listed by Github? Currently we only list CI success / failure. Is there any reason for not adding it earlier (prior to coverage having been disabled)

marcoabreu:
Coverage is not stable due to the excessive use of random data. Our tests do not cover the same paths every single time. Unless tests are expanded to have a consistent coverage, we can't enable it.

leezu:
Thanks for explaining the reasoning. We can enable it, without marking it as required.
This will expose the problem and make it easier to fix it.

marcoabreu:
We could do that, but imo, something even worse than no data is bad data. People will start discarding the results and it will confuse users. 

I'd rather recommend to silently record the results and then take the time to write a script which outputs the lines with inconsistent coverage. Then, create tasks as follow-up. Once coverage is consistent, enable publishing

leezu:
Was there any prior effort on reducing the coverage variance during the past years? I think data was collected silently until a few months ago?

We can at least ensure that any newly introduced code has sufficient code coverage and does not suffer from the variances you describe. Providing test coverage data in PRs, separately listing the coverage data of newly added lines, seems still helpful.

marcoabreu:
No, no activities were done besides activating silent collection. 

The problem with newly added lines is that you can not make that distinction in an automatic Fashion. The new code might be hit by existing tests based on random data with flaky paths. Also, the system is not able to distinguish new hits and existing hits due to them constantly switching. If the coverage would be consistent, then we could enable exactly that Delta.

leezu:
Given that's it may take significant effort and no effort was spent yet, I don't think it's realistic to expect existing test cases would be rewritten soon. Thus if we take this as a blocker to exposing test coverage, we may never expose it.

If new lines of code are covered by existing test cases, that's still valuable information. We can thereby avoid introducing any completely untested lines of code into the codebase.

marcoabreu:
Sure, please feel free to make a poc once publishing is enabled again and then we can evaluate if the data is helpful or not. Generally I'm a big fan of test coverage diff

szha:
@leezu just shared with me the coverage data on codecov: https://codecov.io/gh/apache/incubator-mxnet/branch/master. The coverage looks stable and I think the info should be useful. If we see actual randomness in coverage, then we can start asking people to fix the tests that are causing them. @marcoabreu what do you think?

marcoabreu:
Actually the coverage is not stable at all. Look at this example report https://codecov.io/gh/apache/incubator-mxnet/pull/17905/changes which is based on #17905. The PR adds one new operator, but 132 files have coverage changes with a reported line changes in a four digit space. 

That example is not ideal, but I couldn't find a recent PR which had a successful CI run and was a NO-OP. Maybe just create a few empty PRs, make the CI runs succeed and then you will have a few clear reports which show you exactly which lines are not stable.

Generally we can only consider the coverage report stable if that page reports 0 changes for runs which do not change anything. Otherwise, these numbers are misleading and will confuse the PR authors if we enable the publishing.

image

marcoabreu:
You have to take into account that additionally to randomness in our tests, a stable coverage report requires that the base commit (in the master branch) actually had a successful run. Otherwise, the report is based on partial data.

So I'd say there are two or three action items:

1. Introduce tests for flaky lines
2. Add a retry mechanism for failed CI runs on the master branch
3. Stabilize CI runs to "never". This goal is reached when the master branch CI run rate is close to 100% successful.

PR #17912, opened 2020-03-26
Use x64 toolchain for CI windows build

leezu:
## Description ##
Try using x64 toolchain for CI windows build. We already recommend that in the installation instructions.

This may fix the CI OOM error on windows gpu build. Reference https://github.com/SuperElastix/SuperElastix/issues/190 https://github.com/SuperElastix/SuperElastix/commit/6cd7f1bdad39aa32d49d390b3ca3eb9eaecdc093

BTW, we only need to set this because we're using an outdated windows toolchain. This would be the default on MSVC 2019 version. Reference https://cmake.org/cmake/help/v3.14/generator/Visual%20Studio%2016%202019.html#toolset-selection https://cmake.org/cmake/help/v3.12/generator/Visual%20Studio%2015%202017.html#toolset-selection

### DIFF
--- a/ci/build_windows.py
+++ b/ci/build_windows.py
@@ -54,10 +54,14 @@ class BuildFlavour(Enum):
 
 CMAKE_FLAGS = {
     'WIN_CPU': (
+        '-DCMAKE_C_COMPILER=cl '
+        '-DCMAKE_CXX_COMPILER=cl '
+        '-DOpenCV_RUNTIME=vc15 '
+        '-DOpenCV_ARCH=x64 '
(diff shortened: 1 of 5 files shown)


leezu:
BTW, we only need to set this because we're using an outdated windows toolchain. This would be the default on MSVC 2019 version. Reference https://cmake.org/cmake/help/v3.14/generator/Visual%20Studio%2016%202019.html#toolset-selection

leezu:
Closing as it breaks build. May need to wait for https://github.com/apache/incubator-mxnet/pull/17808/

PR #9558, opened 2018-01-25
Signum with grad compression

yuxiangw:
## Description ##
Added a new grad compression scheme.
It keeps track of the momentum on every worker and sends the sign vector of the momentum.
This is a somewhat hacky version of doing distributed version of the Signum optimizer: https://github.com/apache/incubator-mxnet/pull/9220

### Essentials ###
- [x] Passed code style checking (`make lint`)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- [ ] To the my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #8662, opened 2017-11-15
2bit gradient compression

rahul003:
## Description ##
Implements 2bit gradient compression by quantizing each value in gradient array to 2bits using user specified threshold. Shows about 2x speedup on large models with components like fully connected layers, and LSTM layers.

@eric-haibin-lin @cjolivier01 @anirudh2290 @reminisce 

### Important files to review
GC : gc-inl.h, gc.cc
KVStore local: comm.h
KVStore dist : kvstore_dist.h, kvstore_dist_server.h
Documentation about gradient compression: kvstore.py

### Essentials ###
- [x] Passed code style checking (`make lint`)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage
- [x] For user-facing API changes, API doc string has been updated.
- [x] To my best knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] Gradient compression class
- [x] Reduce operation in kvstore_local / comm.h
- [x] Distributed kvstore changes at worker and server
- [x] Tests for local kvstore, distributed kvstore with predefined and random data. The results have been compared with expected values by implementing this logic in python.
- [x] API changes for Kvstore, Module and Trainer in python
- [x] Addressed comments from last PR

### Problem 
When training large scale deep learning models especially with distributed training, communication becomes a bottleneck for networks whose computation is not high compared to the communication. 

### Approach
We can compress the gradients by considering only those elements that exceed a threshold. Only these elements are encoded and sent. The elements of the gradient that are near zero can safely be delayed by aggregating them in a residual array. When the updated residual with gradient of next iterations exceed the threshold, these values are sent. Effectively these values are updated at a lower frequency. 
On the receiver's end we decompress the data and use the decompressed weights.
Specifically in this PR, 2bit quantization has been implemented.

### Two bit quantization
Any positive value greater than or equal to the threshold is set to one value (say 11), any negative value whose absolute value is greater or equal to the threshold is set to second value (say 10), and others are set to third value (say 0). We need three values to represent data in this fashion and hence two bits. We understand this leads to one bit going waste, but that's an optimization to be done later. The error in quantization is accumulated as residual and carried over to the next iterations. This is added in the next iteration to the gradient before quantizing. 
An example below with thresholds of -2.0 and 2.0. 
This format leads to the reduction of gradient size by 1/16th.
Quantization at work

### Format of compressed gradient
Eac element, represents upto 16 elements in the original array. For the example above, we get an element whose binary representation is 
[...]

PR #17952, opened 2020-04-01
[v1.x][KVStore]1Bit gradient compression

shuo-ouyang:
## Description ##
Added 1bit gradient compression implementation which has similar speedup with 2bit compression. It works with a threshold, values in gradient above the threshold will be quantized to +1, while values below the threshold will be quantizaed to -1. Different with 2bit compression, this implementation of 1bit supports zero threshold. In addition, 1bit compression seems perform better than the current implentation of 2bit compression when batch size increase.

### Important files to review ###
gradient_compression-inl.h
gradient_compression.cc

### Test accuracy of ResNet110 on cifar10 with 1/2/4 workers ###
each worker is equipped with 4 Tesla V100 GPUs
training command: 
```python
python /mxnet/tools/launch.py --launcher ssh -H hosts -n 4 python train_cifar10.py \
--num-epochs 200 --mode hybrid --num-gpus 4 -j 8 --batch-size 128 --wd 0.0001 \
--lr 0.1 --lr-decay 0.1 --lr-decay-epoch 100,150 --model cifar_resnet20_v1 \
--kv-store dist_sync_device --gc-type 1bit --gc-threshold 0 
```
#### 1worker
resnet20_test

#### 2workers
resnet20_test

#### 4workers
resnet20_test

#### 4workers, update_on_kvstore=false
resnet20_test

#### train perplexity
lstm_train

#### test perplexity
lstm_test

### Throughput 
In this part, we use example/imageclassification/benchmark.py to evaluate the performance of gradient compression. We have tried our best to optimize the kernel function of gradient compression operations. In the original kernel function, each thread manipulates 32 bits (4 bytes) of compressed gradients, whereas in our kernel function, each thread writes 8 bits (1 byte) at once. The difference between the original and our implementations are follows.

original kernel implementation:
```c++
struct quantize_2bit {
  MSHADOW_XINLINE static void Map(int out_block_id,
                                  int original_size,
                                  float *out,
                                  float *grad,
                                  float *residual,
                                  const float neg_threshold,
                                  const float pos_threshold) {
    // this block contains the compressed representation of
    // upto 16 values starting from out_block_id*16
    float *compr_block = out + out_block_id;
    // init to 0
    *compr_block = 0;
    // start and end are indices in original grad array
    const int start = out_block_id << 4;
    const int end = (start + 16 <= original_size) ? start + 16 : original_size;
    // cast as char* to manipulate bits of float addresses
    char *block_ptr = reinterpret_cast < char * > (compr_block);
    // masks to set bits when value meets pos_threshold
    // 0xc0 is mask when value is to be represented by the first two bits in a char*
    // 0xc0 means first two bits are set to 11
    const uint8_t posbits[] = {0xc0, 0x30, 0x0c, 0x03};
    // masks to set bits when value meets neg_threshold
    const uint8_t negbits[] = {0x80, 0x20, 0x08, 0x02};
    for (int i = start; i < end; i++) {
      // adds offset to reach appropriate byte
      char *curr_byte = block_ptr + ((i - start) >> 2);
      // adds gradient to existing residual to get updated grad
      residual[i] += grad[i];
      if (residual[i] >= pos_threshold) {
        // set data to 11
        *curr_byte |= posbits[(i & 3)];
        // reduce residual by pos_threshold
        residual[i] -= pos_threshold;
      } else if (residual[i] <= neg_threshold) {
        // set data to 10
        *curr_byte |= negbits[(i & 3)];
        residual[i] -= neg_threshold;
      }
    }
  }
};
```

our kernel implementation:
```c++
struct quantize_2bit {
  MSHADOW_XINLINE static void Map(int out_byte_id,
                                  int original_size,
                                  float *out,
                                  float *grad,
                                  float *residual,
                                  const float neg_threshold,
                                  const float pos_threshold) {
    // this block contains the compressed representation of
    // upto 4 values starting from (char*)out + out_byte_id
    char *compr_byte = reinterpret_cast<char *>(out) + out_byte_id;
    // init to 0
    *compr_byte = 0;
    // start and end are indices in original grad array
    const int start = out_byte_id << 2;
    const int end = (start + 4 <= original_size) ? start + 4 : original_size;

    // masks to set bits when value meets pos_threshold
    // 0xc0 is mask when value is to be represented by the first two bits in a char*
    // 0xc0 means first two bits are set to 11
    const uint8_t posbits[] = {0xc0, 0x30, 0x0c, 0x03};
    // masks to set bits when value meets neg_threshold
    const uint8_t negbits[] = {0x80, 0x20, 0x08, 0x02};
    for (int i = start; i < end; i++) {
      // adds gradient to existing residual to get updated grad
      residual[i] += grad[i];
      if (residual[i] >= pos_threshold) {
        // set data to 11
        *compr_byte |= posbits[(i & 3)];
        // reduce residual by pos_threshold
        residual[i] -= pos_threshold;
      } else if (residual[i] <= neg_threshold) {
        // set data to 10
        *compr_byte |= negbits[(i & 3)];
        residual[i] -= neg_threshold;
      }
    }
  }
};
```
Our optimized implementation performs well when use multiple GPUs on a single machine. Here is a benchmark test of the VGG16 model on a 4 GPUs machine (unit: samples/sec).
| VGG16 | original | our  |
|---|---|---|
| onebit  | ~626  | ~752  |
| twobit | ~705  | ~761  |
| baseline | ~564 ||

**However, our implementation works not well when the compression operator is lunched on the CPU, especially when we use multiple nodes and set `kvstore=dist_sync`.** Under such circumstance, our new implementation may lead to a little throughput reduction. IMHO, one possible solution is adopting different kernel functions for CPU and GPU respectively. In other words, we can use the original kernel for CPU compression operation while the new kernel for GPU operation (still working in process...).

The following are benchmark tests of AlexNet and Resnet-50 on a cluster with at most 4 nodes.
#### Alexnet, batch size=128 on each GPU, dist_sync_device

alexnet-speed

#### ResNet-50, batch size=128 on each GPU, dist_sync_device
resnet50-speed

more performance test will be released recently...

### Related issue or pr ###
signum with grad compression #9558
2bit gradient compression #8662

### Reference ###
[Seide F, Fu H, Droppo J, et al. 1-bit stochastic gradient descent and its application to data-parallel distributed training of speech dnns](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/IS140694.pdf)

### Essentials ###

### Changes ###

## Comments ##

### DIFF
--- a/src/kvstore/gradient_compression.cc
+++ b/src/kvstore/gradient_compression.cc
@@ -41,8 +41,10 @@ void GradientCompression::SetParams(const std::vector<std::pair<std::string, std
                                     & kwargs) {
   GradientCompressionParam params;
   params.InitAllowUnknown(kwargs);
-  CHECK_GT(params.threshold, 0) << "threshold must be greater than 0";
-  if (params.type == "2bit") {
+  if (params.type == "1bit") {
+    SetOneBitCompression(params.threshold);
+  } else if (params.type == "2bit") {
+    CHECK_GT(params.threshold, 0) << "threshold must be greater than 0 for two bit compression";
     SetTwoBitCompression(params.threshold);
   } else {
     LOG(FATAL) << "Unknown type for gradient compression " << params.type;
@@ -55,6 +57,11 @@ CompressionType GradientCompression::get_type() {
 
 std::string GradientCompression::get_type_str() {
   return std::to_string(static_cast<int>(type_));
+}
+
+void GradientCompression::SetOneBitCompression(const float threshold) {
+  type_ = CompressionType::kOneBit;
+  threshold_ = threshold;
 }
 
 void GradientCompression::SetTwoBitCompression(const float threshold) {
@@ -83,7 +90,9 @@ void GradientCompression::DecodeParams(const std::string &s) {
 }
 
 int GradientCompression::GetCompressionFactor() {
-  if (type_ == CompressionType::kTwoBit) {
+  if (type_ == CompressionType::kOneBit) {
+    return 32;
+  } else if (type_ == CompressionType::kTwoBit) {
     return 16;
   } else {
     LOG(FATAL) << "Unsupported compression type: " << get_type_str();
@@ -106,16 +115,34 @@ void GradientCompression::Quantize(const mxnet::NDArray &from, mxnet::NDArray *t
   const int a = from.ctx().dev_mask();
   const int b = to->ctx().dev_mask();
   const float threshold = threshold_;
-  if (type_ == CompressionType::kTwoBit) {
-    if (a == mshadow::cpu::kDevMask && b == mshadow::cpu::kDevMask) {
+  if (a == mshadow::cpu::kDevMask && b == mshadow::cpu::kDevMask) {
+    if (type_ == CompressionType::kOneBit) {
+      mxnet::Engine::Get()->PushSync([from, to, residual, threshold](mxnet::RunContext ctx) {
+        std::vector<mxnet::TBlob> inputs = {from.data(), residual->data(), to->data()};
+        Quantize1BitImpl(ctx.get_stream<mshadow::cpu>(), inputs, threshold);
+      }, from.ctx(), {from.var()}, {to->var(), residual->var()},
+      mxnet::FnProperty::kNormal, priority, "QuantizeCPU");
+    } else if (type_ == CompressionType::kTwoBit) {
       mxnet::Engine::Get()->PushSync([from, to, residual, threshold](mxnet::RunContext ctx) {
         std::vector<mxnet::TBlob> inputs = {from.data(), residual->data(), to->data()};
         Quantize2BitImpl(ctx.get_stream<mshadow::cpu>(), inputs, threshold);
       }, from.ctx(), {from.var()}, {to->var(), residual->var()},
       mxnet::FnProperty::kNormal, priority, "QuantizeCPU");
     } else {
+      LOG(FATAL) << "Unsupported quantization of type " << get_type_str();
+    }
+  } else {
+    if (a == mshadow::gpu::kDevMask && b == mshadow::gpu::kDevMask) {
 #if MXNET_USE_CUDA
-      if (a == mshadow::gpu::kDevMask && b == mshadow::gpu::kDevMask) {
+      if (type_ == CompressionType::kOneBit) {
+        mxnet::Engine::Get()->PushSync([from, to, residual, threshold](mxnet::RunContext ctx) {
+          std::vector<mxnet::TBlob> inputs = {from.data(), residual->data(), to->data()};
+          Quantize1BitImpl(ctx.get_stream<mshadow::gpu>(), inputs, threshold);
+          // Wait GPU kernel to complete
+          ctx.get_stream<mshadow::gpu>()->Wait();
+        }, from.ctx(), {from.var()}, {to->var(), residual->var()},
+        mxnet::FnProperty::kNormal, priority, "QuantizeGPU");
+      } else if (type_ == CompressionType::kTwoBit) {
         mxnet::Engine::Get()->PushSync([from, to, residual, threshold](mxnet::RunContext ctx) {
           std::vector<mxnet::TBlob> inputs = {from.data(), residual->data(), to->data()};
           Quantize2BitImpl(ctx.get_stream<mshadow::gpu>(), inputs, threshold);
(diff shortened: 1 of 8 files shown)


shuo-ouyang:
@mxnet-bot run ci [unix-gpu, centos-gpu]

shuo-ouyang:
@mxnet-bot run ci [unix-gpu]

shuo-ouyang:
@mxnet-label-bot add [KVStore, Distributed]

wkcn:
@mxnet-bot run ci [unix-gpu]

shuo-ouyang:
@wkcn Thanks for your help and support. To be honest, I am not familiar with jenkins workflow, and I dont know why some checks are failed when only one whitespace was deleted in code comments.

wkcn:
@shuo-ouyang It is not related to the code. The reason is that the CI is unstable now.

shuo-ouyang:
@wkcn I get it, thanks!

wkcn:
Hi @rahul003 , could you please help take a review? Thank you!

shuo-ouyang:

wkcn:
Hi @eric-haibin-lin and @szha , could you please help take a reivew?

Thank you!

szha:
@mxnet-bot run ci [all]

wkcn:
Hi @eric-haibin-lin , could you please help take a review?
The PR is related to gradient compression.

Thank you!

wkcn:
Hi @szha , could the PR be merged ?
There seems to be no other committer reviewing this PR, and the PR has been blocked for about 3 months.
The author provides the detailed training result, and I had reviewed the code.

szha:
@wkcn since @eric-haibin-lin recently updated the kvstore interface, it would be best to have a review from him.

@eric-haibin-lin can you help?

szha:
@shuo-ouyang looks like cpp lint failed. You can reproduce it locally with
```
ci/build.py -R --docker-registry mxnetci --platform ubuntu_cpu --docker-build-retries 3 --shm-size 500m /work/runtime_functions.sh sanity_cpp
```

shuo-ouyang:
> @shuo-ouyang looks like cpp lint failed. You can reproduce it locally with
> 
> ```
> ci/build.py -R --docker-registry mxnetci --platform ubuntu_cpu --docker-build-retries 3 --shm-size 500m /work/runtime_functions.sh sanity_cpp
> ```

Sorry, cannot run it on my windows...

szha:
Here are the error messages in the sanity check link
```
[2020-08-09T05:50:42.728Z] =====383/384 cpp-header files passed check=====
[2020-08-09T05:50:42.728Z] src/kvstore/gradient_compression-inl.h: 1 Errors of 1 Categories map={'whitespace': 1}
[2020-08-09T05:50:42.728Z] =====648/649 cpp-source files passed check=====
[2020-08-09T05:50:42.728Z] src/kvstore/gradient_compression.cc: 1 Errors of 1 Categories map={'whitespace': 1}
```

shuo-ouyang:
> What is the end2end speedup when using the 1bit compressor?
> The accuracy drop on resnet20 is also very significant..

Well, in our current experiments (ResNet-20 and ResNet-110 trained on CIFAR-10 dataset), 1bit compressor can achieve similar speed up with an adequate model accuracy compared with 2bit compressor. But CIFAR-10 is a small toy dataset, we will run more experiments on ImageNet dataset and compare these two compressors.

shuo-ouyang:
@eric-haibin-lin 
Very sorry for delayed response...The experiment results are updated, please check it.

src/kvstore/gradient_compression.cc
```diff
@@ -124,14 +151,14 @@ void GradientCompression::Quantize(const mxnet::NDArray &from, mxnet::NDArray *t
         }, from.ctx(), {from.var()}, {to->var(), residual->var()},
         mxnet::FnProperty::kNormal, priority, "QuantizeGPU");
       } else {
-        LOG(FATAL) << "unknown device mask";
+        LOG(FATAL) << "Unsupported quantization of type " << get_type_str();
       }
+    } else {
+      LOG(FATAL) << "unknown device mask";
```
szha, line 157: log what devices are recieved.

src/kvstore/gradient_compression.cc
```diff
@@ -142,35 +169,52 @@ void GradientCompression::Dequantize(const mxnet::NDArray &from, mxnet::NDArray
...
       } else {
-        LOG(FATAL) << "unknown device mask";
+        LOG(FATAL) << "Unsupported dequantization of type " << get_type_str();
       }
+    } else {
+      LOG(FATAL) << "unknown device mask";
+    }
 #else
-      LOG(FATAL) << MXNET_GPU_NOT_ENABLED_ERROR;
+    LOG(FATAL) << MXNET_GPU_NOT_ENABLED_ERROR;
```
szha, line 214: if GPU is not enabled and a wrong device type is received here that is neither gpu nor cpu, this error would be confusing. "gpu not enabled" error only makes sense to users when the input contexts are on gpu.

shuo-ouyang: Thanks for your review. We will solve it as soon as possible.

szha:
The results show that the 2-bit compression results in test perplexity higher than 1-bit compression, which seems surprising. Any explanation on why it's the case?

shuo-ouyang:
The reason may be that the onebit quantizer can exactly capture the direction (sign) of each element in gradients when the threshold is 0, whereas twobit quantizer cannot due to its limination (threshold != 0). By the way, the previous results are trained with additional code modification about error feedback part, which may lead to counterintuitive results. We have corrected them and uploaded the right results.

szha (approved):
Thanks for the explanation. The change looks good to me.

PR #17962, opened 2020-04-02
Fix Windows GPU CI

leezu:
## Description ##
Minimal version of https://github.com/apache/incubator-mxnet/pull/17808 

Update Windows CI to use VS 2019 and enable x64 bit toolchain. Previously we are using an older 32 bit toolchain causing OOM errors during linking.  Switching to x64 bit toolchain on the older VS version previously used by the CI was attempted in https://github.com/apache/incubator-mxnet/pull/17912 and did not work. Update to Cuda 10.2 as it is required by VS 2019. Switch to ninja-build on Windows to speed up build as ninja-build is now preinstalled. Remove logic to install cmake 3.16 on every PR as cmake 3.17 is now preinstalled.

CC: @marcoabreu

### DIFF
--- a/ci/build_windows.py
+++ b/ci/build_windows.py
@@ -37,11 +37,6 @@ from subprocess import check_call
 
 from util import *
 
-
-# Fix for broken PATH with newline inserted presumably by VS studio installation of SQL server or
-# other component which makes visual studio stop working.
-os.environ['PATH']=os.environ.get('PATH').replace('\n','')
-
 KNOWN_VCVARS = {
     # https://gitlab.kitware.com/cmake/cmake/issues/18920
     'VS 2015': r'C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\x86_amd64\vcvarsx86_amd64.bat',
@@ -67,8 +62,6 @@ CMAKE_FLAGS = {
         '-DUSE_CUDNN=OFF '
         '-DENABLE_CUDA_RTC=OFF '
         '-DUSE_OPENCV=ON '
-        '-DOpenCV_RUNTIME=vc15 '
-        '-DOpenCV_ARCH=x64 '
         '-DUSE_OPENMP=ON '
         '-DUSE_BLAS=open '
         '-DUSE_LAPACK=ON '
@@ -84,8 +77,6 @@ CMAKE_FLAGS = {
         '-DUSE_CUDNN=OFF '
         '-DENABLE_CUDA_RTC=OFF '
         '-DUSE_OPENCV=ON '
-        '-DOpenCV_RUNTIME=vc15 '
-        '-DOpenCV_ARCH=x64 '
         '-DUSE_OPENMP=ON '
         '-DUSE_BLAS=open '
         '-DUSE_LAPACK=ON '
@@ -101,8 +92,6 @@ CMAKE_FLAGS = {
         '-DUSE_CUDNN=OFF '
         '-DENABLE_CUDA_RTC=OFF '
         '-DUSE_OPENCV=ON '
-        '-DOpenCV_RUNTIME=vc15 '
-        '-DOpenCV_ARCH=x64 '
         '-DUSE_OPENMP=ON '
         '-DUSE_BLAS=mkl '
         '-DUSE_LAPACK=ON '
@@ -118,8 +107,6 @@ CMAKE_FLAGS = {
         '-DUSE_CUDNN=OFF '
         '-DENABLE_CUDA_RTC=OFF '
         '-DUSE_OPENCV=ON '
-        '-DOpenCV_RUNTIME=vc15 '
-        '-DOpenCV_ARCH=x64 '
         '-DUSE_OPENMP=ON '
         '-DUSE_BLAS=mkl '
         '-DUSE_LAPACK=ON '
@@ -135,8 +122,6 @@ CMAKE_FLAGS = {
         '-DUSE_CUDNN=ON '
         '-DENABLE_CUDA_RTC=ON '
         '-DUSE_OPENCV=ON  '
-        '-DOpenCV_RUNTIME=vc15 '
-        '-DOpenCV_ARCH=x64 '
         '-DUSE_OPENMP=ON '
         '-DUSE_BLAS=open '
         '-DUSE_LAPACK=ON '
@@ -153,8 +138,6 @@ CMAKE_FLAGS = {
         '-DUSE_CUDNN=ON '
         '-DENABLE_CUDA_RTC=ON '
         '-DUSE_OPENCV=ON '
-        '-DOpenCV_RUNTIME=vc15 '
-        '-DOpenCV_ARCH=x64 '
         '-DUSE_OPENMP=ON '
         '-DUSE_BLAS=open '
         '-DUSE_LAPACK=ON '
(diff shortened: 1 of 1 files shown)


ci/build_windows.py
```diff
@@ -106,29 +129,37 @@ class BuildFlavour(Enum):
...
         '-DENABLE_CUDA_RTC=ON '
         '-DUSE_OPENCV=ON  '
+        '-DOpenCV_RUNTIME=vc15 '
+        '-DOpenCV_ARCH=x64 '
         '-DUSE_OPENMP=ON '
         '-DUSE_BLAS=open '
         '-DUSE_LAPACK=ON '
         '-DUSE_DIST_KVSTORE=OFF '
-        '-DMXNET_CUDA_ARCH="5.2" '
+        '-DMXNET_CUDA_ARCH="7.5" '
```
marcoabreu, line 144: Please revert the CUDA change. That does not represent a fix.

ci/build_windows.py
```diff
@@ -106,29 +129,37 @@ class BuildFlavour(Enum):
...
         '-DENABLE_CUDA_RTC=ON '
         '-DUSE_OPENCV=ON '
+        '-DOpenCV_RUNTIME=vc15 '
+        '-DOpenCV_ARCH=x64 '
         '-DUSE_OPENMP=ON '
         '-DUSE_BLAS=open '
         '-DUSE_LAPACK=ON '
         '-DUSE_DIST_KVSTORE=OFF '
-        '-DMXNET_CUDA_ARCH="5.2" '
+        '-DMXNET_CUDA_ARCH="7.5" '
```
marcoabreu, line 162: Please revert the CUDA change. That does not represent a fix.

ci/build_windows.py
```diff
@@ -253,12 +269,8 @@ def main():
     system = platform.system()
     if system == 'Windows':
         logging.info("Detected Windows platform")
-        if 'OpenBLAS_HOME' not in os.environ:
-            os.environ["OpenBLAS_HOME"] = "C:\\Program Files\\OpenBLAS-v0.2.19"
-        if 'OpenCV_DIR' not in os.environ:
-            os.environ["OpenCV_DIR"] = "C:\\Program Files\\OpenCV-v3.4.1\\build"
         if 'CUDA_PATH' not in os.environ:
-            os.environ["CUDA_PATH"] = "C:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v9.2"
+            os.environ["CUDA_PATH"] = "C:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v10.2"
```
marcoabreu, line 273: Please revert the CUDA change. That does not represent a fix.

tests/python/unittest/test_gluon_data.py
```diff
@@ -285,6 +285,11 @@ def test_multi_worker_dataloader_release_pool():
 
 
 def test_dataloader_context():
+    if os.name == 'nt':
+        print("Skipping test_dataloader_context on Windows due to "
+              "https://github.com/apache/incubator-mxnet/issues/17961")
```
marcoabreu, line 290: Please change back to G3.

leezu:
Re https://github.com/apache/incubator-mxnet/pull/17808#issuecomment-608055780

> Connectivity errors were never an issue with Windows slaves

Connectivity issues became a problem after switching to Windows Server 2019. The switch was done as the old AMI apparently can't be rebuilt anymore and a new AMI had to be started. I was not involved in that effort, but I think it's reasonable to get the Windows AMI instructions working again and choose the latest Windows Server for doing that.

Jenkins connectivity issues typically result from system or network load problems. Moving to a g4 instance to run the Windows GPU tests resolved the connectivity issue.

tests/python/unittest/test_gluon_data.py
leezu: Please provide a technical reason for using g3. Currently Jenkins faces connectivity issues on g3 instance.

ci/build_windows.py
leezu: It's a requirement for running on g4.

marcoabreu:
Well the new Ami should not have been moved into production then. Sorry, but changing a hundred knobs to facilitate one change isn't right. 

These are some standards which I do not see aligned with the projects interest.

tests/python/unittest/test_gluon_data.py
marcoabreu: G3 represent the minimum we are verifying. G4 is compatible with G3 but not the other way around. 

Do not change things if they do not work. I want to see some quality check before changes like this are made without checking with the community.

leezu:
Could you elaborate on how the distinction between running tests on a g3 or g4 does not align with the projects interest?

tests/python/unittest/test_gluon_data.py
leezu: We are still verifying g3 on unix gpu tests. There is no difference in the cuda kernels between windows and unix. It's sufficient to have one pipeline run on g3 to verify the compatibility.

ci/build_windows.py
leezu: This represents a fix for the issues experienced on g3 instances, which are blocking the Windows CI at this point in time.

leezu: It's a requirement for running on g4, which is a required fix.

leezu:
Based on offline discussion with Marco, let's use a patched version of the old AMI first to fix the CI. @josephevans helped to install VS Code 2019 on the old AMI. I have further reduced the diff of this PR to include only the minimal changes to switch to VS Code 2019 and the x64 toolchain.

We'll update to the new AMI with update cuda etc, g4 instances at a later point after running it in the dev environment for a while.

ci/build_windows.py
marcoabreu: Usually visual studio does allow more than one cuda version. You just have to install the respective Toolkit.

leezu: What do you mean? Cuda 9.2 does not support VS 2019

leezu: You can read for example https://superuser.com/questions/1506044/installing-cuda-9-2-with-vs-2019

marcoabreu: Oh, they intentionally broke that backwards compatibility feature. Usually it was possible to compile older cuda versions in later vs versions by installing toolkits which make sure that the integration is available. Seems like that caused issues and thus Microsoft and Nvidia decided to not go that route any further. 

In that case, find to proceed. 

But is there still some kind of compatibility mode which checks that the is still compliant with older cuda standards?

leezu: > But is there still some kind of compatibility mode which checks that the is still compliant with older cuda standards?

No. Unix & CentOS tests Cuda 10.1. Windows now tests Cuda 10.2.
But the risk is low that we'd break Cuda 9 support within the next few days. So it's not a one-way door decision. I suggest we discuss on dev if we want to support cuda 9. If we decide to support it, let's switch the CentOS tests to use Cuda 9.

leezu:
> So are going to x version of VS as a default?

Will do

leezu:
@marcoabreu gpu build is still flaky due to thrust + VS2019 issues. Adding back the retries. http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fwindows-gpu/detail/PR-17962/17/pipeline

PR #17970, opened 2020-04-04
fix block.export

chinakook:
```net.hybridize``` may optimize out some ops. These ops are alive in nn.Block(also nn.HybridBlock), but its names are not contained in symbol's ```arg_names``` list. So ignore these ops except that their name are end with 'running_mean' or 'running_var'.

## Description ##

### Essentials ###

### Changes ###

## Comments ##

### CODE REVIEW

**Comment by wkcn** [comment]
Thanks for your contribution!
Some user-defined operator may use other names except of running_mean and running_var to store the augmented variables, which does not need any gradient, such as the memory feature in the memory network.

Is there any weakness to export all augumented variables?

**Comment by chinakook** [comment]
It's complicated for that rarely case, I think we can add a standard suffix or let user to register their suffix for user for this case.

**Comment by chinakook** [comment]
I have this problem just when I export a ICNet of gluoncv. Its results have four feature maps, but we only use the first one, so other ops related to the other three feature maps are optimized out .

**Comment by wkcn** [comment]
I think it is related to `sym.list_auxiliary_states` rather than `block.export`.

`block.export` should store all auxiliary states.

If some feature maps are optimized out, their auxiliary states should be removed in `sym.list_auxiliary_states`.

It seems that the extra auxiliary states are not removed when some ops are removed in the current implementation.

**Comment by chinakook** [comment]
It's not aux states that are optimized out. It's some normal convolution ops and weights unused by the graph. They are in the gluon's parameter dict but their name are not in the symbol's ```sym.list_arguments```. The routine cannot find the name in the arg_list and then find it in the aux_list where the parameter can never be. So the assertion failed.

File: python/mxnet/gluon/block.py [hunk 1]
```diff
@@ -1248,7 +1248,11 @@ def export(self, path, epoch=0, remove_amp_cast=True):
             will be created, where xxxx is the 4 digits epoch number.
         epoch : int
             Epoch number of saved model.
-
+        remove_amp_cast : bool, optional
+            Whether to remove the amp_cast and amp_multicast operators, before exporting the model.
+        allow_extra : bool, optional
+            Whether to save extra parameters whose name not in the result symbol. User can set
+            allow_extra to True to load these parameters with old mxnet.mod.Module.set_params api.
```
szha, line 1255:
```suggestion
            Whether to save extra parameters whose names are not in the result symbol.
            User can set allow_extra to True to load these parameters with old mxnet.mod.Module.set_params API.
```

szha: You may want to backport this to 1.x as we are dropping module support in 2.0

chinakook: Like @wkcn said, someone would like to save extra parameters. We can drop this argument, but it may be useful in some rarely cases.

szha:
cc @leezu

leezu:
@chinakook will you update the PR to fix the root cause of the issue? See the example by @sxjscience in https://github.com/apache/incubator-mxnet/issues/17981

chinakook:
> 
> 
> @chinakook will you update the PR to fix the root cause of the issue? See the example by @sxjscience in #17981

I've tested. It can be export successfully after this fix.

python/mxnet/gluon/block.py
```diff
@@ -1248,7 +1248,11 @@ def export(self, path, epoch=0, remove_amp_cast=True):
             will be created, where xxxx is the 4 digits epoch number.
         epoch : int
             Epoch number of saved model.
-
+        remove_amp_cast : bool, optional
+            Whether to remove the amp_cast and amp_multicast operators, before exporting the model.
+        allow_extra : bool, optional
+            Whether to save extra parameters whose names are not in the result symbol.
+            User can set allow_extra to True to load these parameters with old mxnet.mod.Module.set_params API.
```
leezu, line 1255: `mxnet.mod.Module.set_params` does not exist anymore.

python/mxnet/gluon/block.py
```diff
@@ -1248,7 +1248,11 @@ def export(self, path, epoch=0, remove_amp_cast=True):
             will be created, where xxxx is the 4 digits epoch number.
         epoch : int
             Epoch number of saved model.
-
+        remove_amp_cast : bool, optional
+            Whether to remove the amp_cast and amp_multicast operators, before exporting the model.
+        allow_extra : bool, optional
```
leezu, line 1253: What is the use-case to save parameters that are not used by the symbol? Do we need this option?

python/mxnet/gluon/block.py
```diff
@@ -1271,8 +1275,13 @@ def export(self, path, epoch=0, remove_amp_cast=True):
             if name in arg_names:
                 arg_dict['arg:%s'%name] = param._reduce()
             else:
-                assert name in aux_names
-                arg_dict['aux:%s'%name] = param._reduce()
+                if name.endswith('running_mean') or name.endswith('running_var') \
+                    or name.endswith('moving_mean') or name.endswith('moving_var'):
+                    assert name in aux_names
+                    arg_dict['aux:%s'%name] = param._reduce()
```
leezu, line 1281: Where do we ensure that no aux parameter with a name different to `running_mean`, `running_var`, etc is missing from the resulting params file?

python/mxnet/gluon/block.py
chinakook: I think it's not necessary.

leezu:
@chinakook can you help resolve the conflicts?

python/mxnet/gluon/block.py
```diff
@@ -1343,13 +1343,18 @@ def export(self, path, epoch=0, remove_amp_cast=True):
         arg_names = set(sym.list_arguments())
         aux_names = set(sym.list_auxiliary_states())
         arg_dict = {}
-        for is_arg, name, param in self._cached_op_args:
-            if not is_arg:
-                if name in arg_names:
-                    arg_dict['arg:{}'.format(name)] = param._reduce()
-                else:
+        for name, param in self.collect_params().items():
```
leezu, line 1346: self.collect_params() does not contain all parameters if the block has been optimized for a backend. You need to rely on `self._cached_op_args`

python/mxnet/gluon/block.py
```diff
@@ -1343,13 +1343,18 @@ def export(self, path, epoch=0, remove_amp_cast=True):
...
+            else:
+                if name.endswith('running_mean') or name.endswith('running_var') \
+                    or name.endswith('moving_mean') or name.endswith('moving_var'):
                     assert name in aux_names
-                    arg_dict['aux:{}'.format(name)] = param._reduce()
+                    arg_dict['aux:%s'%name] = param._reduce()
+                else:
+                    warnings.warn('Parameter "{name}" is ignored in saving '
+                                  'as it is not found in the graph. '
+                                  .format(name=name), stacklevel=3)
```
leezu, line 1357: This will ignore all aux parameters that are part of the graph, but are not using the batch norm naming scheme. Why not just remove the `assert name in aux_names` from the original implementation and save all auxiliary parameters that are used in the graph? I don't see a reason for restricting the set of saved parameters in the frontend API.

python/mxnet/gluon/block.py
```diff
@@ -1348,8 +1348,17 @@ def export(self, path, epoch=0, remove_amp_cast=True):
...
+                        or name.endswith('moving_mean') or name.endswith('moving_var'):
+                        if name not in aux_names:
+                            warnings.warn('Parameter "{name}" is not found in '
+                                          'the graph. '
+                                          .format(name=name), stacklevel=3)
+                    else:
+                        warnings.warn('Parameter "{name}" is not found in '
+                                      'the graph. '
+                                      .format(name=name), stacklevel=3)
+                    arg_dict['aux:%s'%name] = param._reduce()
```
leezu, line 1361: Why not

```suggestion
                    if name not in aux_names:
                        warnings.warn('Parameter "{name}" is not found in the graph. '
                                      .format(name=name), stacklevel=3)
                    else:
                        arg_dict['aux:%s'%name] = param._reduce()
```

leezu:

PR #17977, opened 2020-04-05
Relaxing type requirements for broadcast_like

tobecontinued:
## Description ##
Similarly to #14097, broadcast_like op should also not dictate the type of its second argument (since only shape metadata is used there).

### Essentials ###
- [x ] Changes are complete (i.e. I finished coding on this PR)
- [x ] All changes have test coverage:
- [x ] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change


tobecontinued:

src/operator/tensor/broadcast_reduce_op_value.cc
```diff
@@ -138,7 +138,16 @@ NNVM_REGISTER_OP(broadcast_like)
     [](const NodeAttrs& attrs) {
       return std::vector<std::string>{"lhs", "rhs"};
     })
-.set_attr<nnvm::FInferType>("FInferType", ElemwiseType<2, 1>)
+.set_attr<nnvm::FInferType>("FInferType", [](const nnvm::NodeAttrs& attrs,
+                                             std::vector<int> *in_attrs,
+                                             std::vector<int> *out_attrs) {
+  CHECK_EQ(in_attrs->size(), 2) << " in operator " << attrs.name;
+  std::vector<int> checked_in_attrs = { (*in_attrs)[0] };
+  bool ret = !type_is_none((*in_attrs)[1]) &&
```
wkcn, line 146: Thanks for your contribution!

Is it necessary for the condition `!type_is_none((*in_attrs)[1])` ?

tests/python/unittest/test_operator.py
```diff
@@ -3119,6 +3119,15 @@ def test_reshape_like_different_types():
...
     assert_allclose(z.asnumpy(), [[0,0],[0,0],[0,0]])
 
+@with_seed()
+def test_broadcast_like_different_types():
+    x = mx.nd.zeros((2, 1))
+    y = mx.nd.ones((2, 2))
+
+    y = mx.nd.array(y).astype('int32')
+    z = mx.nd.broadcast_like(x, y)
+    assert_allclose(z.asnumpy(), [[0,0],[0,0]])
```
wkcn, line 3129: It is better to add the assertion `assert x.dtype == z.dtype`.

src/operator/tensor/broadcast_reduce_op_value.cc
tobecontinued: Well, I actully just found the issue,  the code is copied from merged PR #14097, which had a similar issue. I think it is better to keep the same requirement for all *_like op.

tobecontinued:
@mxnet-bot run ci [unix-gpu]

tobecontinued:
@mxnet-bot run ci [unix-gpu]

src/operator/tensor/broadcast_reduce_op_value.cc
ptrendx: Yes, it is necessary - what `FInferType` returns is whether it succeeded in inferring all the types (so that if all operators return true we know that all types are inferred). That is why it is important to not lie and return true only if all types are really inferred (even if we do not actually do anything with the other type).

tobecontinued:
@mxnet-bot run ci [unix-gpu]

tobecontinued:
@mxnet-bot run ci [all]

tobecontinued:
@mxnet-bot run ci [all]

wkcn:

wkcn:
The PR has been merged.

Thanks for your contribution! : )

Issue #17968, opened 2020-04-04
[RFC] Raising the toolchain requirements for MXNet 2


leezu:
## Description
I propose to raise our toolchain requirements for the MXNet 2 development branch to require at minimum gcc7 or clang6 on Unix systems and MSVC 2019 on Windows system. All 3 have [reasonable complete C++17 support](https://en.cppreference.com/w/cpp/compiler_support#cpp17) and MSVC 2019 fully supports, so that we can adopt `C++17` as required language standard. gcc7 and clang6 are available on Ubuntu 18.04 LTS release.

The benefits of adopting a more recent C++ standard should be obvious, giving us access to new features and abstractions that the C++ committee has worked on over the course of 6 years. The benefits of adopting a more recent toolchain should also be obvious, as newer compilers will come with more optimizations as older ones.

There are no downsides for MXNet's users, as we can continue to build binary releases of MXNet on CentOS 6 that should work on any major Linux distribution released after 2004. This is possible based on the great work by RedHat to bring new C++ toolchains to old platforms [1]. 

With respect to Windows: MSVC 2019 is the first MSVC that uses an 64bit toolchain by default. You may have noticed that our Windows CI was recently blocked due to the use of a 32bit toolchain and updating it to MSVC 2019 was chosen as remedy (attempts to use the 64bit version of the 2017 toolchain failed). It also appears that MSVC 2019 16.5 is the first release to make proper use of advanced instruction sets, such as AVX2 [2].

## References
1: https://www.softwarecollections.org/en/scls/rhscl/devtoolset-8/
2: https://devblogs.microsoft.com/cppblog/avx2-floating-point-improvements-in-visual-studio-2019-version-16-5/

sandeep-krishnamurthy:
+1 Thanks for bringing this up.

leezu:
https://github.com/apache/incubator-mxnet/commit/fb73a1717acad61caeaeef010faed9e9fcc05f0e implements the proposal, fixing a number of other issues that were blocking. Please see the commit message for a complete list of changes.

As a follow-up item, I suggest to remove the `cpplint` we currently use in favor of `clang-tidy` (which we also use). cpplint enforces the Google's C++ style guide, which is geared toward C++03. Instead we can target https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines with enforcement by clang-tidy (+ potentially MSVC).

Issue #17067, opened 2019-12-13
Scala flaky test: testObjectDetectionExample and GanExampleSuite


eric-haibin-lin:
```
- testObjectDetectionExample

[ScalaTest-main-running-GanExampleSuite] INFO org.apache.mxnetexamples.gan.GanExampleSuite - GPU test only, skipped...

[ScalaTest-main-running-DiscoverySuite] INFO org.apache.mxnetexamples.profiler.ProfilerSuite - Running profiler test...

[ScalaTest-main-running-DiscoverySuite] INFO org.apache.mxnetexamples.profiler.ProfilerSuite - profile file save to /tmp

terminate called after throwing an instance of 'dmlc::Error'

GanExampleSuite:

- Example CI: Test GAN MNIST

  what():  [06:41:27] src/c_api/c_api_profile.cc:141: Check failed: !thread_profiling_data.calls_.empty(): 

Stack trace:

  [bt] (0) /work/mxnet/scala-package/native/target/libmxnet-scala.so(dmlc::LogMessageFatal::~LogMessageFatal()+0x64) [0x7f3cbd6f7a44]

  [bt] (1) /work/mxnet/scala-package/native/target/libmxnet.so(mxnet::on_exit_api()+0x514) [0x7f3bc7355624]

  [bt] (2) /work/mxnet/scala-package/native/target/libmxnet.so(MXNDArrayFree+0x2a) [0x7f3bc73a83fa]

  [bt] (3) [0x7f3d7103f740]

```
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-cpu/detail/PR-17057/1/pipeline/

ChaiBapchya:
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-cpu/detail/PR-16722/5/pipeline

#16722

ChaiBapchya:
Another one - http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-cpu/detail/PR-17482/4/pipeline
#17482

ChaiBapchya:
Another one - http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-cpu/detail/PR-17487/13/pipeline #17487

leezu:
[...]

PR #17984, opened 2020-04-06
Raise toolchain requirements for MXNet 2 

leezu:
## Description ##
As per https://github.com/apache/incubator-mxnet/issues/17968, require C++17 compatible compiler. 
For cuda code, use C++14 mode introduced in Cuda 9. In general there is no issue with linking together objects resulting from different standards as long as they are compiled by the same compiler. 

This PR contains the following specific changes:
- Switch CI pipeline to use gcc7 on Ubuntu and CentOS 
- Switch CD pipeline to CentOS 7 with https://www.softwarecollections.org/en/scls/rhscl/devtoolset-7/ This enables us to build with gcc7 C++17 compiler while keeping a relatively old glibc requirement for distribution.
- Simplify ARM Edge builds
  - Switch to standard Ubuntu / Debian cross-compilation toolchain for ARMv7, ARMv8
  - Switch to https://toolchains.bootlin.com/ toolchain for ARMv6 (the Debian ARMv6 toolchain is for ARMv4 + ARMv5 + ARMv6, but we wish to only target ARMv6 and make use of ARMv6 features)
  - Remove reliance on `dockcross` for cross compilation.
- Simplify Jetson build
  - Use standard Ubuntu / Debian cross-compilation toolchain for ARMv8
  - Upgrade to Cuda 10 and Jetpack 4.3
  - Simplify build setup
- Simplify QEMU ARM virtualization test setup on CI
  - Remove complex "Virtual Machine in Docker" logic and run a QEMU based Docker container instead based on `arm32v7/ubuntu`
- Fix out of bounds vector accesses in
  - SoftmaxGradOpType
  - MKLDNNFCBackward
- Fix use of non-standard `rand_r` function (which is not available on anymore on newer Android toolchains and shouldn't be use in any case).
- Fix reproducibility of RNN with Dropout
- Fix reproducibility of DGL Graph Sampling Operators
- Update tests for Android Edge build to NDK19. The [previously used standalone toolchain is obsolete](https://developer.android.com/ndk/guides/standalone_toolchain).

Those Dockerfiles that required refactoring as part of the effort were refactored based on the following consideration
- Maximize the use of system dependencies provided by the distribution instead of manually installing dependencies from source or from third party vendors. This reduces the complexity of the installation process and essentially pins the dependency versions, increasing CI stability. Further, Dockerfile build speed is improved. To facilitate this, use recent distribution versions. We still ensure backwards compatibility via CentOS7 based build and test stages
- Minimize the number of layers in the Dockerfile. Don't have 5 different script files executed, each calling `apt-get update` and `install`, but just execute once. Speeds up the build and reduces image size. Keep each Dockerfile simple and tailored to a purpose, instead of running 20 scripts to install dependencies for every thinkable scenario, which is unmaintainable.
 
Some more small changes:
- Remove outdated references to Cuda 7 and Cuda 8 in various files.
- Remove C++03 support in mshadow
- Disable broken tests (may fix as part of this PR if I find the time)
  - NumpyBooleanAssignForwardCPU https://github.com/apache/incubator-mxnet/issues/17990
  - test_init.test_rsp_const_init https://github.com/apache/incubator-mxnet/issues/17988


ptrendx:
Note: With this we will be able to change shared_ptr here: https://github.com/apache/incubator-mxnet/blob/master/src/engine/threaded_engine.h#L440 to `unique_ptr` (as it was supposed to be but C++11 did not allow `std::move` to lambda).

leezu:
@larroy this relies on the binfmt fix binary support of linux kernel: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/binfmt_misc.txt?h=linux-4.8.y#n69 There is also a small summary at https://codepyre.com/2019/12/arming-yourself/

leezu:
@mxnet-bot run ci [unix-cpu]

Due to https://github.com/apache/incubator-mxnet/issues/17067

src/operator/contrib/dgl_graph.cc
```diff
@@ -24,6 +24,9 @@
 #include <mxnet/operator_util.h>
```
szha, line 24: @zheng-da could you help review this in #17984?

zheng-da: DGL no longer depends on the contrib graph operators. I would suggest to deprecate this sampling operator and even remove it in MXNet 2.0. Please see my comment here. https://github.com/apache/incubator-mxnet/issues/16167#issuecomment-613160299

Issue #16708, opened 2019-11-02
Training an FPN model using grad_req="add"  causes rapid divergence, while manually implemented gradient accumulation works fine


nickguletskii:
## Description
While working with FPNs and gradient accumulation, I've discovered that using `grad_req=add` with certain models utilizing FPNs causes almost immediate divergence. This issue is similar to #16686, but the results are much more extreme: the toy model I have provided in this issue diverges in just a couple of steps. A proprietary model (which I cannot share) diverges immediately.

### Steps to reproduce
(Paste the commands you ran that produced the error.)

1. Save the following script:

```python
import argparse
from enum import Enum
from typing import Tuple, Iterable

import mxnet as mx
import mxnet.gluon.nn as nn
from gluoncv.model_zoo import resnet18_v1b
from gluoncv.nn.feature import FPNFeatureExpander
from mxnet import autograd
from mxnet.gluon import HybridBlock, Trainer
from mxnet.gluon.loss import SoftmaxCrossEntropyLoss
from mxnet.initializer import Xavier
from mxnet.test_utils import assert_almost_equal


class SimpleModel(HybridBlock):
    def __init__(self, context):
        super().__init__()
        with self.name_scope():
            self.glob_avg_pool = nn.GlobalAvgPool2D()
            self.linear = nn.Dense(2)
            self.loss = SoftmaxCrossEntropyLoss()

    def hybrid_forward(self, F, x, label, *args, **kwargs):
        res = self.linear(self.glob_avg_pool(x))
        return self.loss(res, label)


class FPNModel(HybridBlock):
    def __init__(self, context):
        super().__init__()
        with self.name_scope():
            base_network = resnet18_v1b(pretrained=False, dilated=False, use_global_stats=True, ctx=context)
            features = FPNFeatureExpander(
                network=base_network,
                outputs=['layers1_relu3_fwd', 'layers2_relu3_fwd', 'layers3_relu3_fwd',
                         'layers4_relu3_fwd'], num_filters=[256, 256, 256, 256], use_1x1=True,
                use_upsample=True, use_elewadd=True, use_p6=True, no_bias=False, pretrained=False, ctx=context)
            self.features = features
            self.loss = SoftmaxCrossEntropyLoss()
            self.conv2d = nn.Conv2D(256, kernel_size=(3, 3), weight_initializer=Xavier(magnitude=2))
            self.glob_avg_pool = nn.GlobalAvgPool2D()
            self.linear = nn.Dense(2)

[...]

PR #17995, opened 2020-04-07
Fix ElemwiseSum for more than 4 inputs

### LINKED ISSUE #17989: [Autograd] Very serious bug of grad_req='add'
Minimal reproducible example:

```python
import mxnet as mx
from mxnet.gluon import nn, HybridBlock
import numpy as np
import argparse
mx.npx.set_np()

np.random.seed(123)
mx.random.seed(123)


parser = argparse.ArgumentParser(
        description='Grad req bug minimal example')
parser.add_argument('--addto', action='store_true')
parser.add_argument('--hybridize', action='store_true')
parser.add_argument('--nrepeat', type=int, default=5)
args = parser.parse_args()

class Foo(HybridBlock):
    def __init__(self, prefix=None, params=None):
        super().__init__(prefix=prefix, params=params)
        with self.name_scope():
            self.layer = nn.Dense(16)

    def hybrid_forward(self, F, dat):
        out = dat
        for _ in range(args.nrepeat):
            out = self.layer(out)
        return out

foo = Foo()
if args.hybridize:
   foo.hybridize()
foo.initialize(ctx=mx.gpu())

if args.grad_addto:
    for p in foo.collect_params().values():
        p.grad_req = 'add'
foo.collect_params().zero_grad()


dat = mx.np.random.normal(0, 1, (32, 16), ctx=mx.gpu())
og = mx.np.random.normal(0, 1, (32, 16), ctx=mx.gpu())
with mx.autograd.record():
    out = foo(dat)
    loss = (out * og).sum()
    loss.backward()
for k, v in foo.collect_params().items():
    print(k, mx.np.linalg.norm(v.grad()))
```

The gradient will  be different when `add` is triggered. However, we should initialize the gradient as all zero.

To reproduce the bug, we can use the following command:

new script with zero_grad
```bash
wget https://gist.githubusercontent.com/sxjscience/0bd336c921396b3c66331354e1866886/raw/d618ba69cbecf04d3013db77af86c29d62fe0336/grad_req_addto_bug.py -O grad_req_addto_bug.py
python grad_req_addto_bug.py --addto
python grad_req_addto_bug.py

```
The output is 

```
ubuntu@ip-172-31-27-255:~$ python grad_req_addto_bug.py --addto

foo0_dense0_weight 1.7802463
foo0_dense0_bias 315.05945
ubuntu@ip-172-31-27-255:~$ python grad_req_addto_bug.py
foo0_dense0_weight 0.19310227
foo0_dense0_bias 19.947094
```

ptrendx:
## Description ##
Fixes #17989

It was caused by a bug in `ElemwiseSum` which for more than 4 inputs and `kAddTo` req was counting gradients multiple times.

@sxjscience @zhreshold  Please test if this fixes the issues you saw.

### DIFF
--- a/src/operator/tensor/elemwise_sum.h
+++ b/src/operator/tensor/elemwise_sum.h
@@ -94,7 +94,7 @@ void ElementWiseSumCompute_(const nnvm::NodeAttrs& attrs,
       Kernel<Sum, xpu>::Launch(s, out_size, out_dptr, req[0], in_0_dptr);
       for (size_t i = 1; i < size; ++i) {
         DType* in_dptr = in_data[i].dptr<DType>();
-        Kernel<Sum, xpu>::Launch(s, out_size, out_dptr, req[0], out_dptr, in_dptr);
+        Kernel<Sum, xpu>::Launch(s, out_size, out_dptr, kWriteTo, out_dptr, in_dptr);
       }
       break;
     }
--- a/tests/python/unittest/test_operator.py
+++ b/tests/python/unittest/test_operator.py
@@ -9875,6 +9875,25 @@ def test_im2col_col2im():
         pad         = 1
     )
 
+def test_elemwise_sum_for_gradient_accumulation():
+    for nrepeat in range(1, 10):
+        stored_grad = dict()
+        for grad_req in ['write', 'add']:
+            a = mx.nd.array([1])
+            b = mx.nd.array([2])
+            if grad_req == 'write':
+                a.attach_grad(grad_req='write')
+            elif grad_req == 'add':
+                a.attach_grad(grad_req='add')
+            a.grad[:] = 0
+            with mx.autograd.record():
+                for _ in range(nrepeat):
+                    b = b * a
+                b.backward()
+            stored_grad[grad_req] = a.grad.asscalar()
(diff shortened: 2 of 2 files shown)


leezu:
Should we add a test case in the style of https://github.com/apache/incubator-mxnet/issues/17989#issuecomment-610591835 to ensure no such bug makes it into the codebase again?

sxjscience:
@ptrendx I can confirm that this fixes the issue.

zhreshold:
confirmed it can fix https://github.com/apache/incubator-mxnet/issues/16708#issuecomment-558876214 as well

szha:
@ptrendx thanks for the fix. Let's add a test for this.

szha:
@mxnet-bot run ci [centos-gpu, unix-gpu, windows-gpu]

ptrendx:
Added v1.x label even though this PR is targeted for master in order to track that we want it included in 1.7 release @ciyongch FYI.

ciyongch:
@ptrendx can you help to backport this PR to v1.x branch? Then I'll rebase v1.7.x to v1.x accordingly. I'm adding this to 1.7.0 roadmap https://github.com/apache/incubator-mxnet/issues/16864.

Issue #18005, opened 2020-04-09
[nvidia-docker][1.6] apt-get update failing with gnutls_handshake() failed: Handshake failed


ChaiBapchya:
## Description
Ubuntu 14.04
1.6 branch
Static build Python

### Error Message
```
W: Failed to fetch https://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu1404/x86_64/Packages  gnutls_handshake() failed: Handshake failed
```
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-17993/5/pipeline

## To Reproduce
```
git checkout v1.6.x
ci/build.py --docker-registry mxnetci --nvidiadocker --platform publish.ubuntu1404_gpu --docker-build-retries 3 --shm-size 500m /work/runtime_functions.sh build_static_python_cu101mkl
```

## What have you tried to solve it?

1. Tried it locally. Able to reproduce. Hence ruling out possibility of Jenkins error.
This is a security / certificate issue [likely]. Since apt-get is unable to fetch the package.

While : https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1404/x86_64/Packages is a 404.

https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1404/x86_64 is a valid URL. It has several files [Packages.gz being one of them.

## Comments
Surprisingly, this doesn't happen on master branch or on the previous commits to v1.6 branch

ChaiBapchya:
Discovered in all 5 builds of PR https://github.com/apache/incubator-mxnet/pull/17993

leezu:
Ubuntu 14.04 is no longer supported. Nvidia may now require a newer ssl toolchain than provided by ubuntu 14.04.

The issue doesn't occur on master because it's hidden by the Docker cache

PR #18018, opened 2020-04-10
Fix gnu_tls handshake error on Ubuntu 14.04 Nvidia Docker

ChaiBapchya:
Fixes https://github.com/apache/incubator-mxnet/issues/18005

It doesn't occur on Master branch yet due to docker cache caching the Nvidia docker image for Ubuntu 14.04

This switches the dep link in /etc/apt/sources.list.d/*.list from https to http
Specifically 2 files exit
nvidia-ml.list and cuda.list

With this switch, TLS is no longer needed and hence the handshake error gets circumvented.

@leezu @zachgk Thanks for the help with investigation.
Pl review

### DIFF
--- a/ci/docker/install/ubuntu_publish.sh
+++ b/ci/docker/install/ubuntu_publish.sh
@@ -20,6 +20,10 @@
 # Build on Ubuntu 14.04 LTS for LINUX CPU/GPU
 set -ex
 
+# replace https with http to force apt-get update to use http
+# nvidia-docker no longer supports ubuntu 14.04
+# refer https://github.com/apache/incubator-mxnet/issues/18005
+sudo sed -i 's/https/http/g' /etc/apt/sources.list.d/*.list
 apt-get update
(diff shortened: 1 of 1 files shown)


ChaiBapchya:
@mxnet-label-bot add [build]

ChaiBapchya:
@mxnet-bot run ci [sanity]

To verify if builds are triggered on both jenkins dev & prod.

zachgk:
This is really not that great for security. Especially considering that this is for our release pipeline. I see two other options:

1. We could try building and installing openssl from source so we get a newer version
2. Ultimately, we will need to find a way to migrate this build away from using ubuntu 14.04 at some point. That would also resolve this problem too

leezu:
@zachgk we'll move to CentOS as part of https://github.com/apache/incubator-mxnet/pull/17984/commits/02b9105e6e8921250c9a4fe22e0735d3c1417171

So you can regard this as temporary fix

PR #18054, opened 2020-04-14
Fix tests for ONNX version 1.5.0 bump

RuRo:
## Description ##
This PR was originally proposing to bump the ONNX version from `1.3.0` to `1.5.0`, but the ONNX version got bumped in #18025. However, in 18025 the tests weren't updated, so some of them are failing, so I am repurposing this PR, since it has the fixes necessary to update the tests.

Below is the original text of this PR (before 18025 bumped the ONNX version).

----

----

This is a WIP pull request for increasing the supported `ONNX` version from `1.3.0` to `1.5.0`.
This change is motivated by the fact, that `ONNX v1.5.0` (opset 10) supports new operators, that should allow/simplify the implementation of some `mxnet -> ONNX` conversions, that are currently not implemented.

List of potentially useful operators, which were not available in `1.3.0`, but are available in `1.5.0`:
- `Compress`
- `ConstantOfShape`
- `EyeLike`
- `NonMaxSuppression`
- `NonZero`
- `OneHot`
- `Resize`
- `RoiAlign`
- `Scatter`
- `Where`

There are more operators, but these are the important ones, IMO.

This PR doesn't implement any of the above operators, but just bumps the `ONNX` version in CI and fixes the tests for the newer version.

## Changes ##
1) The `implement onnx translation helpers` commit just implements some helper functions, that I've used in all my PRs for ONNX translations
2) Tests for trigonometric functions `sin`/`cos`, `acos`/`asin`/`atan` now have `[^h]` appended, so that it doesn't accidentally match with **new** ONNX test for the Hyperbolic trigonometric functions, the conversion for which is currently not implemented. (`test_sinh` etc)
3) Test `test_matmul` changed to `test_matmul_`, so that it doesn't accidentally match with the **new** ONNX test for `matmulinteger`, the conversion for which is currently not implemented.
4) Test `test_cast` changed to `test_cast((?!STRING).)*$`, so that it doesn't accidentally match with the **new** ONNX test for casting to/from strings. `((?!STRING).)*$` matches any text until the end of string, as long as it doesn't contain the word `STRING`.
5) Updated the implementation for exporting `topk` and `slice_axis` operators, since they changed their API in the new opset version.
[...]

PR #17827, opened 2020-03-13
[Numpy] Bugfix of slice operator export (MXNet to ONNX)

QueensGambit:
## Description ##
This PR fixes the slice operator export from MXNet into ONNX.

@MoritzMaxeiner reported this problem already: **Incorrect ONNX export of SliceChannel** https://github.com/apache/incubator-mxnet/issues/13061.
The corresponding pull request **ONNX export: Support equal length splits** https://github.com/apache/incubator-mxnet/pull/14121 however, only partially solved the problem.

The output of the slice operator was corrected but it is still not properly handled in the `get_inputs()` function.

The following unit test demonstrates this:

```python
class SplitConcatBlock(HybridBlock):
    """Block which creates two splits and later concatenates them"""
    def __init__(self, name):
        super(SplitConcatBlock, self).__init__(name)

    def hybrid_forward(self, F, x):
        splits = F.split(x, axis=1, num_outputs=2)
        return F.concat(*splits)

def test_onnx_export_slice(self):
     net = nn.HybridSequential(prefix='slice_net')
     with net.name_scope():
         net.add(nn.Dense(100, activation='relu'), SplitConcatBlock("splitConcat"), nn.Dense(10))
     _check_onnx_export(net)
```

Before this PR, the conversion of the model from MXNet into ONNX resulted in the following error:
```python
Traceback (most recent call last):
  File "mxnet_export_test.py", line 134, in test_onnx_export_slice
    _check_onnx_export(net)
  File "mxnet_export_test.py", line 66, in _check_onnx_export
    onnx_file_path=onnx_file_path)
  File "/opt/mxnet/python/mxnet/contrib/onnx/mx2onnx/export_model.py", line 87, in export_model
    verbose=verbose)
  File "/opt/mxnet/python/mxnet/contrib/onnx/mx2onnx/export_onnx.py", line 312, in create_onnx_graph_proto
    checker.check_graph(graph)
  File "/usr/local/lib/python3.6/dist-packages/onnx/checker.py", line 51, in checker
    proto.SerializeToString(), ctx)
onnx.onnx_cpp2py_export.checker.ValidationError: Nodes in a graph must be topologically sorted, however input 'slice_netsplitConcatsplit0' of node: 
input: "slice_netsplitConcatsplit0" input: "slice_netsplitConcatsplit0" output: "slice_netsplitConcatconcat0" name: "slice_netsplitConcatconcat0" op_type: "Concat" attribute { name: "axis" i: 1 type: INT }
[...]

PR #18055, opened 2020-04-14
Implement some ONNX exports, needed for SSD

RuRo:
## Description ##
This is a WIP pull request for implementing exports for some `ONNX` operators, which are needed to successfully export the `SSD` model from `gluocv`. The only 2 operators, which are not yet implemented are `_contrib_box_decode`, which can be sidestepped by using `SSD(minimal_opset=True)` and `_contrib_box_nms`, which could be potentially implemented with the `NonMaxSuppression` `ONNX` operator, but this is a big enough undertaking to be its own separate PR.

This PR assumes that #18054 is merged (ruro/onnx_version_bump), since it uses some `ONNX` operators only available in `v1.5.0`. (The first commit is a squashed version of that PR and will be dropped after it gets merged into master)

If you actually want to try exporting `SSD`, you should also merge #17827 (QueensGambit/onnx_export_slice_fix or ruro/onnx_export_slice_fix).

## Changes ##
- `eagerly check onnx conversion availability in verbose mode`
  The `export_model` method now warns about *all* nodes, that don't have a registered conversion **before** trying to export the model, if `verbose=True`. This is a simple quality of life change, which helps potential contributors to estimate, how many layers are missing until a model can be exported.
- implement `_arange` op
  (doesn't support the `repeat` attribute for now)
- implement `ones_like` and `zeros_like` ops
- implement `_lesser_scalar`, `_greater_scalar` and `_equal_scalar` ops
  also fixes the issue, where the `scalar_op_helper` would fail, if the input is a non-1D initializer
- implement `slice_like` op
  (doesn't support the implicit `axes` variation right now)
- implement `where` op


RuRo:
@mxnet-bot run ci [unix-gpu]

VisionZQ:
Hi, 
Thanks for you kindness.
We have converted the mxnet to onnx 1.5,  and the 'slice_like' is work.
But, When I from onnx convert to TensorRT6.0, the program crash. can you give me any suggestion? 
`While parsing node number 205 [Gather -> "yolov30__mulscalar0__shape__gathered"]:
ERROR: /home/jakin/songzq/onnx-tensorrt/onnx2trt_utils.hpp:412 In function convert_axis:
[8] Assertion failed: axis >= 0 && axis < nbDims
`
Thanks a lot.

RuRo:
@VisionZQ This PR currently has WIP status (it's not ready) and it's main purpose was to implement the needed ops to convert `SSD` models from `mxnet` to `ONNX`. There's been a lot of breakage recently with `onnx2mx`, so some further work may be required. I am not sure, I can help you with converting `yolov3`. I don't currently have the free time to even finish this PR.

I would recommend, that you open an issue, describing, which layers needed for `yolov3` don't have a conversion in `mxnet 1.6.0` and which conversion functions are currently missing in `mxnet 2.0.0`, but I wouldn't get your hopes up, that somebody will implement the conversions for you. Afaik, most `mxnet` contributors are currently **very** busy with `mxnet 2` API changes.

chinakook:
It's a great job!

Aktcob:
Really good job！ 

When I convert yolov3 model in gluoncv to ONNX, I'm faced with the same problem.

PR #18056, opened 2020-04-14
Fix ci/docker_cache.py

leezu:
## Description ##
https://github.com/apache/incubator-mxnet/pull/17984 added a `cache_intermediate` option to `build_docker` enabling developers to speed up local docker builds during development. As no default value was provided, this broke the use of `build_docker` function in other places.

Reference: Failing `ci/jenkins/restricted-docker-cache-refresh` on master after merge: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/restricted-docker-cache-refresh/detail/master/2789/pipeline

Also

Fixes https://github.com/apache/incubator-mxnet/issues/18061

### DIFF
--- a/ci/build.py
+++ b/ci/build.py
@@ -71,7 +71,7 @@ def get_docker_binary(use_nvidia_docker: bool) -> str:
 
 
 def build_docker(platform: str, docker_binary: str, registry: str, num_retries: int, no_cache: bool,
-                 cache_intermediate: bool) -> str:
+                 cache_intermediate: bool = False) -> str:
     """
     Build a container for the given platform
     :param platform: Platform
--- a/ci/docker/Dockerfile.build.ubuntu_nightly_cpu
+++ b/ci/docker/Dockerfile.build.ubuntu_nightly_cpu
@@ -42,6 +42,9 @@ RUN /work/ubuntu_perl.sh
 COPY install/ubuntu_clang.sh /work/
 RUN /work/ubuntu_clang.sh
 
+COPY install/ubuntu_gcc8.sh /work/
+RUN /work/ubuntu_gcc8.sh
+
 COPY install/ubuntu_caffe.sh /work/
 RUN /work/ubuntu_caffe.sh
 
--- a/ci/docker/Dockerfile.build.ubuntu_nightly_gpu
+++ b/ci/docker/Dockerfile.build.ubuntu_nightly_gpu
@@ -42,6 +42,9 @@ RUN /work/ubuntu_perl.sh
 COPY install/ubuntu_clang.sh /work/
 RUN /work/ubuntu_clang.sh
 
+COPY install/ubuntu_gcc8.sh /work/
+RUN /work/ubuntu_gcc8.sh
+
 COPY install/ubuntu_tvm.sh /work/
 RUN /work/ubuntu_tvm.sh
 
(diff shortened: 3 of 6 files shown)


ci/build.py
```diff
@@ -71,7 +71,7 @@ def get_docker_binary(use_nvidia_docker: bool) -> str:
 
 
 def build_docker(platform: str, docker_binary: str, registry: str, num_retries: int, no_cache: bool,
-                 cache_intermediate: bool) -> str:
+                 cache_intermediate: bool = False) -> str:
```
marcoabreu, line 74: This has to be true.

leezu: I don't mind changing it to `True`, but the existing behaviour before introducing the flag was `False`.

leezu: One issue I noted during local development when setting this to `True`, is that my disk can fill up very quickly because Docker will not clean any of the cached layers automatically.

marcoabreu: Well there has been some back and forth between refactors - maybe somebody removed it :/. Initially, I designed it that way that intermediary layers were kept since that's the whole point of the partial cache. This test validates that assumption: https://github.com/apache/incubator-mxnet/blob/master/ci/test_docker_cache.py#L184

Note that this test suite is not run in CI...

marcoabreu: Well you can't have everything - either partial caching and manual cleanup or automatic cleanup but no partial caching. I think it is fine to expect that people use docker purge every now and then.

leezu: Does the CI docker purge automatically? My docker cache quickly grow by 200GB after developing locally with the intermediate caching enabled. It's fine to purge time to time manually, but it's better to automate this.

leezu: > Note that this test suite is not run in CI...

Is it possible to run this suite in CI? If so, would you like to help enable it?

leezu: For reference, you removed the `rm=false` yourself in f107397b753ff08bb19e7572bc1a9ebedd832f88

marcoabreu: CI does not docker purge automatically. But since the lifecycle of unix slaves is very very short, it's not an issue. Also, in 99% of the cases, no new layers are built and only the cache is used, so the likelyhood of a single slave regenerating a lot of new mismatching layers is quite low. On a dev machine though, that's a different case.

marcoabreu: Yeah it should be possible. It will just cause issues if the test fails since that might mean that stuff gets left dangling, but due to the mentioned lifecycle, that shouldn't have a big impact.

Feel free to enable it.

marcoabreu: Good catch! I can't fully recall what was the thing around --rm=false, but I recall that I wrote tests to cover all these unknowns. Rule of thumb is that if the tests pass, we're good to go (given they use the same parameters of course).

leezu: > CI does not docker purge automatically. But since the lifecycle of unix slaves is very very short, it's not an issue

If the intermediate containers are only kept on the slave, and we only generate them in PRs that change the Dockerfile, I think we don't need to use them at all on CI. `--rm=false` then seems only useful for development.

marcoabreu: Yep exactly. But please make sure that the tests pass before merging - otherwise we might break the partial caching.

leezu: This PR does not change the status quo. There is no risk.

PR #18081, opened 2020-04-16
Add ability to assign classes for ImageFolderDataset

dwSun:
there should be a convenient way to assign classes and the order of them.


dwSun:
@mxnet-bot run ci [all]

szha:
@mxnet-bot run ci [all]

Issue #18046, opened 2020-04-14
Proposal to mxnet.metric


acphile:
## Motivation

mxnet.metric provides different methods for users to judge the performance of models. But currently there are some shortcomings which need to be improved in mxnet.metric. We propose to refactor the metrics interface to fix all issues and place the new interface under mx.gluon.metrics.

In general, we want to make the following improvements:

1. Moving the API to the gluon namespace
2. Make the API more user-friendly and pythonic
3. Structure the API to make hybridization of the complete training loop more easily feasible in the future.

### 1. Inconsistency in computational granularity of metrics

Currently there are two computational granularities in mxnet.metric:

1. “macro” level: calculate average performance per batch , like implementation in [MAE](http://mxnet.incubator.apache.org/api/python/docs/api/metric/index.html?highlight=metric#mxnet.metric.MAE)
2. “micro” level: calculate average performance per sample, like implementation in [Accuracy](http://mxnet.incubator.apache.org/api/python/docs/api/metric/index.html?highlight=metric#mxnet.metric.Accuracy), [CrossEntropy](http://mxnet.incubator.apache.org/api/python/docs/api/metric/index.html?highlight=metric#mxnet.metric.CrossEntropy)

Generally, “micro” level is more useful because usually we focus on average performance of data samples in the test set rather than that of testing batches. So here we need to make arrangements between these metrics.

### 2. For future hybridization of the complete training loop

Currently metrics in mxnet.metric receives “list of NDArray” and calculate results by numpy. In fact, many metrics’ computation could be implemented in nn.HybridBlock. Using HybridBlock.hybridize(), the computation could be done in the backend, which could be faster. By refactoring the mxnet.metric, we could one day compile the model with the metric like Tensorflow and do the complete training loop including evaluation fully in the backend. Thus our new API design takes into account the hybridization use-case, so that hybridizing the complete training loop will be easily possible once the backend support is there.

### 3. lacking some useful metrics

Although many metrics are already included, some still need to be implemented.

Apart from the metrics already provided in mxnet.metric: http://mxnet.incubator.apache.org/api/python/docs/api/metric/index.html?highlight=metric#module-mxnet.metric ,  we plan to add the following metrics:

1. F-beta score: (1+beta^2)*precision*recall/(beta^2*precision+recall)
2. binary accuracy with threshold: using a confidence threshold to judge whether the example is positive or negative
3. MeanCosineSimilarity: return the average cosin similarity between predictions and ground truth
[...]

Issue #17886, opened 2020-03-23
Boolean tvm operators broken on gpu


leezu:
## Description
For example, `less_scalar_gpufloat32_2bool_2` is missing causing mxnet to crash.
Works fine for cpu.

### Error Message

```
Traceback (most recent call last):
  File "/home/ubuntu/test.py", line 4, in <module>
    x[x < 2]
  File "/home/ubuntu/src/mxnet-master/python/mxnet/numpy/multiarray.py", line 729, in __getitem__
    return self._get_np_boolean_indexing(key_before_expaned)
  File "/home/ubuntu/src/mxnet-master/python/mxnet/numpy/multiarray.py", line 490, in _get_np_boolean_indexing
    after_mask = _reshape_view(_npi.boolean_mask(data, mask), -1, *remaining_shapes)
  File "<string>", line 53, in boolean_mask
  File "/home/ubuntu/src/mxnet-master/python/mxnet/_ctypes/ndarray.py", line 82, in _imperative_invoke
    check_call(_LIB.MXImperativeInvokeEx(
  File "/home/ubuntu/src/mxnet-master/python/mxnet/base.py", line 246, in check_call
    raise get_last_ffi_error()
mxnet.base.MXNetError: Traceback (most recent call last):
  File "../3rdparty/tvm/src/runtime/module.cc", line 123
  File "../3rdparty/tvm/src/runtime/library_module.cc", line 91
TVMError: Check failed: ret == 0 (-1 vs. 0) : Check failed: f != nullptr: Cannot find function less_scalar_gpufloat32_2bool_2_kernel0 in the imported modules or global registry
```

### Steps to reproduce

Compile MXNet with `USE_TVMOP=1`.

Run 
```
import mxnet as mx

x = mx.np.array([[0, 1], [1, 1], [2, 2]], ctx=mx.gpu())
idx = x < 2
x[idx]
```

ChaiBapchya:
Duplicate of #17840

Issue #17840, opened 2020-03-15
CI unix-gpu build failure in TVM op


apeforest:
## Description
A few recent PRs failed at the same place related to TVM op. 

https://github.com/apache/incubator-mxnet/pull/17835
https://github.com/apache/incubator-mxnet/pull/17795
https://github.com/apache/incubator-mxnet/pull/17531

@yzhliu

leezu:
@apeforest do you mean this error:

```
[2020-03-18T08:37:14.025Z] TVMError: Check failed: ret == 0 (-1 vs. 0) : Check failed: f != nullptr: Cannot find function less_scalar_gpufloat32_2bool_2_kernel0 in the imported modules or global registry
```

apeforest:
Yes, but it seems to be fixed.

leezu:
Why would it be fixed? I got this error 2020-03-18T08:37:14.025Z UTC

ChaiBapchya:
The issue persists. Upon trying unix-gpu build on G4 in CI Dev account. 
http://jenkins.mxnet-ci-dev.amazon-ml.com/blue/organizations/jenkins/mxnet-validation-bapac%2Funix-gpu/detail/update_gpu_toolchain/8/pipeline/414

all 3 failed tests fail in a similar fashion
they fail at 9 tests with the following error
```
TVMError: Check failed: ret == 0 (-1 vs. 0) : Check failed: f != nullptr: Cannot find function  <x> in the imported modules or global registry
```
9 functions [all 3 have same 9 functions]
```
greater_equal_gpufloat32_0float32_0bool_0_kernel0 (x2)
logical_and_gpufloat32_1float32_1bool_1_kernel0 (x2)
equal_gpufloat32_2float32_2bool_2_kernel0 (x2)
sum_gpureduce1st_dim_1req_kWriteTobool_5float32_2float32_2_kernel0 (x3)
cuda_rad2degfloat32_2float32_2_kernel0 (x2)
```

leezu:
Reproducer
Compile MXNet with USE_TVMOP=1.
```
import mxnet as mx

x = mx.np.array([[0, 1], [1, 1], [2, 2]], ctx=mx.gpu())
idx = x < 2
x[idx]
```

leezu:
Has been disabled on CI: https://github.com/apache/incubator-mxnet/pull/18204

Let's track fixing TVMOp in this issue?

yzhliu:
@JinboCi will be helping

PR #18083, opened 2020-04-16
Changes to mxnet.metric

acphile:
## Description ##
change based on https://github.com/apache/incubator-mxnet/issues/18046 
1. make improvements in metric
  a. improve Class MAE (and MSE, RMSE)
  b. improve Class _BinaryClassification
  c. improve Class TopKAccuracy
  d. add Class MeanCosineSimilarity
  e. add Class MeanPairwiseDistance
2. move mxnet.metric to mxnet.gluon.metric

### Essentials ###

### Changes ###

## Comments ##


python/mxnet/gluon/metric.py
```diff
@@ -777,16 +815,113 @@ class F1(EvalMetric):
...
+
+        preds : list of `NDArray`
+            Predicted values.
+        """
+        labels, preds = check_label_shapes(labels, preds, True)
+
+        for label, pred in zip(labels, preds):
+            self.metrics.update_binary_stats(label, pred)
+
+        if self.average == "macro":
```
sxjscience, line 848: In fact, `macro averaging` + F1 does not mean to average the F1 of each batch. I think we should revise it to be the same as https://scikit-learn.org/stable/modules/generated/sklearn.metrics.f1_score.html .

python/mxnet/gluon/metric.py
```diff
@@ -1130,12 +1339,18 @@ def update(self, labels, preds):
                 label = label.reshape(label.shape[0], 1)
             if len(pred.shape) == 1:
                 pred = pred.reshape(pred.shape[0], 1)
-
-            mae = numpy.abs(label - pred).mean()
+            
+            if self.average == "macro":
```
sxjscience, line 1343: It's very strange to have `macro` + MAE. See scikit-learn: https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_absolute_error.html#sklearn.metrics.mean_absolute_error

python/mxnet/gluon/metric.py
acphile: Averaging F1 per batch here previously existed in metric.py before I made changes. This calculation also exists in MAE, MSE, RMSE, and PearsonCorrelation. Should I remove all of them accordingly? For the average "macro" in sklearn, it seems used in calculating F1 score for multiclass/multilabel targets. But currently our F1 only supports binary classification. I think I need to make extensions for F1.

sxjscience: Let's remove it and make it similar to sklearn. This is in fact the reason why I never use the metric class in MXNet.

python/mxnet/gluon/metric.py
```diff
@@ -619,90 +583,100 @@ def update_binary_stats(self, label, pred):
...
     @property
     def fscore(self):
-        if self.precision + self.recall > 0:
-            return 2 * self.precision * self.recall / (self.precision + self.recall)
-        else:
-            return 0.
+        return (1 + self.beta ** 2) * self.precision * self.recall / numpy.maximum(self.beta ** 2 * self.precision + self.recall, 1e-12)
 
     @property
     def global_fscore(self):
```
leezu, line 664: This method should be removed as you dropped the global states?

python/mxnet/gluon/metric.py
```diff
@@ -24,9 +24,9 @@
 
 import numpy
```
leezu, line 25: Instead of using numpy, we can use `mxnet.numpy` as it runs asynchronously and has GPU support.
The summary states of a metric should be stored on CPU, but if for example data and label are on GPU and input to the metric, we can calculate the sufficient statistics on GPU

python/mxnet/gluon/metric.py
acphile: This method actually refers to the micro calculation for F1 and it is not related to original global support.

leezu: Would it make sense to adjust the name?

acphile: I think it is ok to use global_fscore since it is in a private container class.

python/mxnet/gluon/metric.py
```diff
@@ -617,92 +585,102 @@ def update_binary_stats(self, label, pred):
...
-
-        check_label_shapes(label, pred)
-        if len(numpy.unique(label)) > 2:
-            raise ValueError("%s currently only supports binary classification."
-                             % self.__class__.__name__)
+        pred = pred.as_np_ndarray()
+        label = label.as_np_ndarray().astype('int32')
+        if self.class_type == "binary":
+            self._set(1)
+            if len(numpy.unique(label)) > 2:
```
leezu, line 592: This will trigger synchronization (as we need to wait for the result of the np.unique operator). Could we make error checking that triggers synchronization optional?

leezu: https://github.com/apache/incubator-mxnet/blob/83b51703ed354f41024423f140de38df2ba22d50/src/imperative/imperative.cc#L123-L127

acphile:
@mxnet-bot  run ci [centos-cpu, sanity, centos-gpu, windows-cpu, unix-gpu, windows-gpu, unix-cpu]

leezu:
@mxnet-bot run ci [unix-gpu]

leezu:
@mxnet-bot run ci [unix-gpu]

leezu:
@mxnet-bot run ci [unix-gpu]

marcoabreu:
Why are we removing tvm op?

leezu:
It's currently broken. https://github.com/apache/incubator-mxnet/issues/17840 Needs to be fixed before it can be enabled on CI

marcoabreu:
Please disable it in a separate PR then. I'd like to have a better understanding of what exactly is broken , but not have that discussion in this PR.

leezu:
Let's disable it here, because it blocks this PR

leezu:
You can go ahead and try reproduce the issue locally:

The reproducer is available since more than a month at https://github.com/apache/incubator-mxnet/issues/17886#issue-585864921

marcoabreu:
No, it's unrelated and should be a separated and isolated PR. Each PR should serve one propose. That way, we can focus discussions, have single purpose commits and also allow reverting

leezu:
You're very welcome to resolve this issue and create a separate PR

leezu:
Let's disable it in https://github.com/apache/incubator-mxnet/pull/18204

acphile:
@mxnet-bot run ci [unix-gpu]

acphile:
@mxnet-bot run ci [unix-cpu, windows-gpu]

tests/python/tensorrt/test_cvnets.py
```diff
@@ -29,7 +28,12 @@
 def get_classif_model(model_name, use_tensorrt, ctx=mx.gpu(0), batch_size=128):
     mx.contrib.tensorrt.set_use_fp16(False)
     h, w = 32, 32
-    net = gluoncv.model_zoo.get_model(model_name, pretrained=True)
+    model_url = "https://raw.githubusercontent.com/dmlc/web-data/master/gluoncv/models/"
```
leezu, line 31: Please don't hardcode `master` in the URL here. The repository may change and will then break the CI. Instead, use the commit ID: `https://raw.githubusercontent.com/dmlc/web-data/221ce5b7c6d5b0777a1e3471f7f03ff98da90a0a/gluoncv/models/"

acphile:

acphile:
@mxnet-bot run ci [unix-cpu]

mseth10:
@acphile this PR fails nightly CD while running nightly python unit tests. The following tests fail:
test_mcc, test_multilabel_f1, test_binary_f1
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/restricted-mxnet-cd%2Fmxnet-cd-release-job/detail/mxnet-cd-release-job/1119/pipeline/329

We'll need to revert this PR and fix the failures before re-merging it.

leezu:
The reason it fails is that the CI check in this PR was run too long ago. I should have restarted it before merging the PR. Meantime master changed and caused some additional changes to be necessary. They are in https://github.com/apache/incubator-mxnet/pull/18312/files

PR #18087, opened 2020-04-16
CI/CD: Remove cuda 9.0, 9.1

leezu:
## Description ##
Cuda 9.0 and 9.1 lack support for gcc7 host compiler, which is now required by MXNet.
Thus retain only support for Cuda 9.2. An alternative is to build cuda related code with gcc6 and the rest of MXNet with gcc7 or newer, but the added complexity may not be justified.


leezu:
@mxnet-bot run ci [unix-gpu]

leezu:
@mxnet-bot run ci [unix-gpu]

stuck after np true_divide

leezu:
@mxnet-bot run ci [unix-gpu]

stuck after np empty

PR #18093, opened 2020-04-17
CI: unify centos7_base.sh and centos7_core.sh

leezu:
## Description ##
Unifying the files is a mechanism to ensure the same system dependencies are used on CI and CD. For example, previously `openblas-devel` was missing on CD.


leezu:
@mxnet-bot run ci [unix-gpu]

leezu:
@mxnet-bot run ci [unix-gpu]

np truedivide

leezu:
@mxnet-bot run ci [unix-gpu]

leezu:
@marcoabreu having instructions in shell files is a very bad practice. This leads to constant docker cache invalidation and makes for a horrible development experience. We need to refactor all dockerfiles to get rid of the shell scripts. Some references:

* https://stackoverflow.com/questions/48551953/why-does-my-docker-cache-get-invalidated-by-this-copy-command
* Request for reason of cache invalidation: moby/moby#9294
* COPY invalidates cache: moby/moby#21913
* Timestamp part of hash: moby/moby#9391
** Fixed in moby/moby#12031

Instead of using shell scripts, we can structure our Dockerfiles to have a base image that is shared among multiple containers, where each of the multiple containers is a specialization, eg for Python etc.

leezu:
@mxnet-bot run ci [unix-gpu, windows-gpu]

PR #18103, opened 2020-04-20
[Numpy] Fix DeprecationWarning

sxjscience:
Currently there will be deprecation warning if we use advanced indexing in numpy.


sxjscience:
@mxnet-bot run ci [all]

PR #18115, opened 2020-04-20
CI: Consolidate Dockerfiles

leezu:
## Description ##
This PR shows how to get rid of duplicate Dockerfiles and the `install/X.sh` code deduplication strategy on the example of CentOS7 Dockerfiles. Instead, we use a single templated Dockerfile with a multiple targets for different scenarios. The template arguments and respective targets are declared in `docker-compose.yml`.

This has multiple benefits:

1) Having a single Dockerfile instead of a number of scripts is easier to understand and maintain
2) Copying scripts into the Dockerfile as done in the previous approach prevents usage of the https://hub.docker.com/u/mxnetci/ cache on developer machines, as the `COPY` instruction invalidates the cache if the local repository does not match the repository used on CI (eg. development branch instead of master branch). If the cache is invalidated, developers must wait for long docker build time when attempting to reproduce certain CI builds (such as website build, cc @aaronmarkham) or using the docker image to perform the staticbuild.
3) Having fewer steps in the Dockerfile improves the build speed.


marcoabreu:
That looks really neat!

leezu:
@mxnet-bot run ci [miscellaneous, website, centos-gpu, unix-gpu, edge, sanity, centos-cpu, clang, unix-cpu]

ci/docker/runtime_functions.sh
```diff
@@ -983,6 +983,7 @@ sanity_check() {
 # $2 -> python_cmd: The python command to use to execute the tests, python or python3
 cd_unittest_ubuntu() {
     set -ex
+    source /opt/rh/rh-python35/enable
```
szha, line 986: what does this do?

szha: ok this seems to switch the python alias

cd/mxnet_lib/mxnet_lib_pipeline.groovy
```diff
@@ -75,9 +75,9 @@ def get_stash(mxnet_variant) {
 // The environment corresponds to the docker files in the 'docker' directory
 def get_environment(mxnet_variant) {
   if (mxnet_variant.startsWith("cu")) {
-    return "publish.centos7_gpu_${mxnet_variant}"
+    return "centos7_gpu_${mxnet_variant}"
```
szha, line 78: the `_gpu` infix doesn't seem necessary either

ci/docker/runtime_functions.sh
leezu: CentOS7 ships with Python 2 by default. 3.5 is opt in. This is the opt-in. You can find more info https://www.softwarecollections.org/en/

cd/mxnet_lib/mxnet_lib_pipeline.groovy
leezu: It's for consistency with the existing naming scheme. The name used here must match the name in the docker-compose.yml

ci/docker/runtime_functions.sh
leezu: Due to the way the CI runs stuff in the Dockerfile, we can't permanently opt-in when creating the Docker image. We can fix that later. For now need to opt in inside the runtime functions.

PR #18119, opened 2020-04-20
If sanity build is not found, wait until Jenkins recognizes it.

josephevans:
## Description ##
I ran into some problems when activating the staggered build pipeline in prod. It seems there is a slight delay (on a high-load server) between when Jenkins starts the full-build job and having the sanity job being recognized by Jenkins. This change allows it to retry until the job does exist.


josephevans:
@mxnet-bot run ci [unix-gpu]

josephevans:
@mxnet-bot run ci [unix-gpu]

PR #18145, opened 2020-04-23
Revert #17547

roywei:
revert #17547 as it's causing performance regression.
See discussion in:
https://github.com/apache/incubator-mxnet/pull/17547#discussion_r413521892

### DIFF
--- a/src/operator/nn/dropout-inl.h
+++ b/src/operator/nn/dropout-inl.h
@@ -255,13 +255,8 @@ class DropoutOp {
       Stream<xpu> *s = ctx.get_stream<xpu>();
 
       // set dropout state.
-      Random<xpu, unsigned> *prnd = ctx.requested[1].get_random<xpu, unsigned>(s);
-      uint64_t rng_seed = prnd->GetSeed();
-      // reset dropout descriptor if rng seed changed.
-      bool reset = seed_ != rng_seed;
-      seed_ = rng_seed;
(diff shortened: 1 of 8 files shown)


sxjscience:
@roywei @sandeep-krishnamurthy Is there a way to solve this issue? Should we try to accelerate the native dropout instead?

roywei:
cc @ciyongch @TaoLv please don't include #17547  in 1.7.x branch as it will cause performance regression. I'm reverting in master for now, and probably won't have enough time to fix it before 1.7.x code freeze.

src/operator/nn/dropout-inl.h
```diff
@@ -497,7 +492,7 @@ class DropoutOp {
   Context ctx_;
   cudnnDataType_t dtype_;
   cudnnDropoutDescriptor_t dropout_desc_;
-  uint64_t seed_;
+  uint64_t seed_ = 17 + rand() % 4096;  // NOLINT(runtime/threadsafe_fn)
```
leezu, line 495: https://en.cppreference.com/w/cpp/numeric/random

roywei:
Yes we should also consider implementing in cuda instead of cudnn, which will require major change on random module as well.
reference:
https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/cuda/Dropout.cu

ciyongch:
@roywei thanks for point it out, so far this PR is not included in either v1.x or v1.7.x.

PR #18146, opened 2020-04-23
[CI] run pytest in parallel

szha:
## Description ##
run pytest in parallel

### Essentials ###

### Changes ###
- [x] run small tests in parallel
- [x] mark tests that require more resources
- [x] run large tests in serial
- [x] use built-in tmpdir fixtures for temporary files

### DIFF
--- a/tests/python/train/test_resnet_aug.py
+++ b/tests/python/train/test_resnet_aug.py
@@ -23,139 +23,138 @@ import numpy as np
 import os, pickle, gzip
 import logging
 from mxnet.test_utils import get_cifar10
-import tempfile
 
-batch_size = 128
+def test_cifar10(tmpdir):
+    batch_size = 128
 
-# small mlp network
-def get_net():
-    data = mx.symbol.Variable('data')
-    float_data = mx.symbol.Cast(data=data, dtype="float32")
-    fc1 = mx.symbol.FullyConnected(float_data, name='fc1', num_hidden=128)
-    act1 = mx.symbol.Activation(fc1, name='relu1', act_type="relu")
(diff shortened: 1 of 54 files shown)


tests/python/train/test_resnet_aug.py
```diff
@@ -39,14 +40,15 @@ def get_net():
     return softmax
 
 # check data
-get_cifar10()
+path = tempfile.mkdtemp()
```
leezu, line 43: You're not cleaning up these temporary directories. You could use pytests functionality which will handle cleanup https://docs.pytest.org/en/latest/tmpdir.html

szha: thanks. this is useful. I will replace other occurrences.

szha:
will look into it in a follow-up PR.

szha:
@PatricZhao I noticed that the MKL/MKLDNN tests are taking a lot longer than non-MKL builds in the parallel test setting. I will try to run a couple more times to verify so this is just FYI.

szha:
Update: the MKLDNN builds are actually executing a different set of tests which could explain the time difference. However, the MKL build is indeed executing the same unittest as regular python 3 CPU build and it's consistently taking a lot longer.

marcoabreu:
Could that be due to cold booz? Doesn't mkl generate kernels the first time
you invoke them?

Sheng Zha <notifications@github.com> schrieb am Di., 5. Mai 2020, 21:13:

szha:
@marcoabreu indeed that could be a likely cause. I reported my findings in #18244 and we can continue the discussion there.

PR #18147, opened 2020-04-23
[v1.7.x] Backport #17177 to 1.7.x (Fix incorrect calculation results when the C locale is set to a locale that uses commas as the decimal separator)

nickguletskii:
## Description ##
Backport #17177 to the branch for v1.7.x.

Currently, many operators utilize `std::stod` to convert strings into floating point numbers. This causes incorrect calculations (#17140, #16134) when the C locale is set to a locale which uses commas (`,`) as decimal separators.

This pull request replaces calls to `std::stod` and `std::stof` to the locale-invariant `dmlc::stod` and `dmlc::stof` respectively.

### The scope of this patch ###

This patch should fix a large portion of interactions through Python and JVM frontends, since they use locale-invariant serialization in order to pass parameters into MXNet's C API.

However, frontends which utilize C locale-aware serialization (i.e. call `sprintf` or similar) may break when using locales which don't use `.` as the decimal separator. They will have to be fixed in a separate patch. I also suspect that they are already broken, because operators utilising dmlc-core parameter parsing were already using locale-invariant serialization.

### Further steps ###
STL streams are heavily used within the codebase, both for serialization and for forming user-friendly messages. Fortunately, they don't seem to be affected by the C standard library locale settings. However, if someone sets the STL locale by calling [std::locale::global](https://en.cppreference.com/w/cpp/locale/locale/global), MXNet's API will be broken. In order to ensure that this doesn't happen, all streams which are used for serialization will have to be imbued with the "C" locale (not the locale set in the C standard library).

It would be nice to see a more principled approach to serialization in MXNet 2.0, e.g. using a binary format for communication between the frontend and the backend. In addition to solving locale-related issues, this would probably result in a smaller invocation overhead.

### Locale-invariant serialization vs locale-aware serialization ###

As a side-note, using locale-aware serialization is not an option, simply because using `,` as the decimal separator adds ambiguities to tuple serialization, e.g. `(4,4,3)` can be a tuple of 3 integers, or a tuple of 2 floats.

### Essentials ###

### Changes ###
- [x] Replace `std::stod` with `dmlc::stod` and `std::stof with `dmlc::stof`.
- [x] Add a test that tests locale invariance for scalar ops.

## Comments ##
- May break Julia and R code when using a locale that uses `,` as the decimal separator. However, since there was no consistency between the various operators before, it is not unlikely that the code was already broken.

### DIFF
--- a/cpp-package/include/mxnet-cpp/optimizer.h
+++ b/cpp-package/include/mxnet-cpp/optimizer.h
@@ -32,6 +32,7 @@
 #include <string>
 #include <memory>
 #include <functional>
+#include <dmlc/strtonum.h>
 #include "mxnet-cpp/base.h"
 #include "dmlc/logging.h"
 #include "mxnet-cpp/ndarray.h"
@@ -84,7 +85,7 @@ class Optimizer {
   Optimizer *SetLRScheduler(std::unique_ptr<LRScheduler> lrScheduler) {
     CHECK(lrScheduler);
     lrScheduler_ = std::move(lrScheduler);
-    lrScheduler_->SetLR(std::stof(params_["lr"]));
+    lrScheduler_->SetLR(dmlc::stof(params_["lr"]));
     return this;
   }
   /*!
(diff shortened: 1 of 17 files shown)


ciyongch:
Adding this PR to 1.7.0 roadmap https://github.com/apache/incubator-mxnet/issues/16864.

ciyongch:
@leezu can you help to take a review and approve/merge if it looks ok to you :)

Issue #18144, opened 2020-04-23
test_numpy_op.py::test_np_empty_like hangs


szha:
## Description
test_numpy_op.py::test_np_empty_like hangs on unix-gpu

http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-18025/59/pipeline/425

leezu:
In the linked CI run `test_numpy_op.py::test_np_empty_like` is not run and thus can't be responsible for the hang. Thus there must be more triggers besides `test_np_empty_like`.

Related https://github.com/apache/incubator-mxnet/issues/18090

haojin2:
Wow this issue is VERY INTERESTING, in the first [link](http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-18025/59/pipeline/425) given in the issue description I'm not even seeing `test_np_empty_like` being run at all, and the last test run before the final timeout process kill was `test_np_bincount`. Also as @leezu pointed out in the above comment, even removing `test_np_empty_like` does not solve the issue. So to conclude, so far I'm not seeing any solid evidence supporting `test_np_empty_like` to be the root cause for the hang.
To be clear, I'm not saying that I don't think we should re-implement `empty_like` with a native implementation in the future, simply want to suggest that maybe you guys are attacking the wrong target at this moment.

leezu:
@haojin2 you can check #18090 for the evidence.

haojin2:
@leezu I understand the goal, but my point is that we should avoid providing un-related info in the issue's description (the hang in the first provided link is not related at all), shouldn't we? It'd be better if link to #18090 was provided in the first place to avoid such confusions.

leezu:
I agree.

PR #18025, opened 2020-04-11
[DEV] switch nose with pytest

szha:
## Description ##
switch nose with pytest

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented
- [x] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] switch nose with pytest

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #18151, opened 2020-04-23
Disable tests of numpy operators implemented with CustomOp

leezu:
## Description ##
https://github.com/apache/incubator-mxnet/issues/18144, https://github.com/apache/incubator-mxnet/pull/18025 disabled `empty_like` which causes hangs. The other 3 CustomOp based fallback operators are also prone to cause deadlock, as exemplified by http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-18025/59/pipeline/425 where the hang occurs without `np_empty_like` test being run

### DIFF
--- a/tests/python/unittest/test_numpy_op.py
+++ b/tests/python/unittest/test_numpy_op.py
@@ -4382,6 +4382,7 @@ def test_np_randn():
 
 @with_seed()
 @use_np
+@pytest.mark.skip(reason='Test hangs. Tracked in #18144')
 def test_np_multivariate_normal():
     class TestMultivariateNormal(HybridBlock):
         def __init__(self, size=None):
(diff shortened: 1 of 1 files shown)


RuRo:
You might also want to disable other tests, which indirectly use those operators.

RuRo:
The `_np` in `test_np_kron` is `numpy`, not `mxnet.numpy`, so if you disabled that test because of the `_np.unravel_index` calls, then I think we can reenable that one since it's just pure `numpy`.

Nice idea, switching `empty_like` to `ones_like` in `test_np_unary_bool_funcs` though!

leezu:
Thanks for catching that @RuRo

PR #18179, opened 2020-04-27
Enable docker cache build for images listed in docker-compose.yml

leezu:
Enable docker cache build for images listed in docker-compose.yml
Ignore `docker-compose.yml` related files in `ci/build_cache.py`

### DIFF
--- a/ci/Jenkinsfile_docker_cache
+++ b/ci/Jenkinsfile_docker_cache
@@ -38,6 +38,7 @@ core_logic: {
         timeout(time: total_timeout, unit: 'MINUTES') {
           utils.init_git()
           sh "python3 ./ci/docker_cache.py --docker-registry ${env.DOCKER_CACHE_REGISTRY}"
+          sh "cd ci && docker-compose -f docker/docker-compose.yml build --parallel && docker-compose -f docker/docker-compose.yml push "
         }
       }
     }
--- a/tests/README.md
+++ b/tests/README.md
@@ -62,13 +62,21 @@ Ninja is a build tool (like make) that prioritizes building speed. If you will b
     
 ## Runing Python Tests Within Docker
 
-1. To run tests inside docker run the following comamdn
-    ```
-    ci/build.py --platform {PLATFORM} /work/runtime_functions.sh {RUNTIME_FUNCTION}
-    ```
-An example for running python tests would be
+To run tests inside docker, you first need to install `docker` and `docker-compose` on your machine.
+
+On Ubuntu you may install them via `sudo apt-get install docker.io docker-compose`
+and set them up via `sudo usermod $(whoami) -G docker -a`.
+
+Then, to run tests inside docker run the following command
+
 ```
-ci/build.py --platform build_ubuntu_cpu_mkldnn /work/runtime_functions.sh unittest_ubuntu_python3_cpu PYTHONPATH=./python/ pytest tests/python/unittest
+ci/build.py --platform {PLATFORM} /work/runtime_functions.sh {RUNTIME_FUNCTION}
+```
+
+An example for running python tests would be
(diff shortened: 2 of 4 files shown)


ci/Jenkinsfile_docker_cache
```diff
@@ -38,6 +38,7 @@ core_logic: {
         timeout(time: total_timeout, unit: 'MINUTES') {
           utils.init_git()
           sh "python3 ./ci/docker_cache.py --docker-registry ${env.DOCKER_CACHE_REGISTRY}"
+          sh "cd ci && docker-compose -f docker/docker-compose.yml build --parallel && docker-compose -f docker/docker-compose.yml push "
```
marcoabreu, line 41: Can you pull that into docker_cache.py to leverage parallelisation wrt non updated images and also providing one single interface?

leezu: The idea is to delete `docker_cache.py` later, so the `docker-compose -f docker/docker-compose.yml build --parallel && docker-compose -f docker/docker-compose.yml push` would be the unified interface.

leezu: I'm not sure what parallelism you refer to. Is `docker-compose build --parallel` not sufficient?

marcoabreu: I understand. 

It's not sufficient in the sense of that docker-cache and docker compose are run in sequence. Alternatively, create a second job so it runs on another node?

leezu: `docker-cache` will only be used temporarily. I think the current solution is fine given that the first step (`sh "python3 ./ci/docker_cache.py`) will be deleted later.

tests/README.md
```diff
@@ -62,13 +62,21 @@ Ninja is a build tool (like make) that prioritizes building speed. If you will b
...
-1. To run tests inside docker run the following comamdn
-    ```
-    ci/build.py --platform {PLATFORM} /work/runtime_functions.sh {RUNTIME_FUNCTION}
-    ```
+To run tests inside docker, you first need to install `docker` and `docker-compose` on your machine.
+
+On Ubuntu you may install them via `sudo apt-get install docker.io docker-compose`
+and set them up via `sudo usermod $(whoami) -G docker -a`.
+
+Then, to run tests inside docker run the following command
```
ChaiBapchya, line 70: Can you also add `pip3 install -r ci/requirements.txt`
Thanks

leezu:
@mxnet-bot run ci [sanity]

tests/README.md
leezu: Let's add it in a separate PR to unblock merging this fix. We can also consider getting rid of that dependency by moving relying on standard tools instead of using our custom scripts.

ChaiBapchya:
@leezu it's supposed to fail on docker cache refresh?
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/restricted-docker-cache-refresh/detail/master/2830/pipeline

tests/README.md
ChaiBapchya: Makes sense. But custom scripts sort of make job easier for lazy people like me :p having to google steps for getting those standard tools. What do you reckon? Apart from taking the onus of ensuring testing is correct rather than leaving it to users to find their way out of the problems that may come up with setting up those standard tools on their own.

leezu:
The run you point out fails due to flakyness of dockerhub.
Dockerhub suddenly rejects our credentials and we get `[2020-04-28T01:55:45.399Z] denied: requested access to the resource is denied`

After retrigger, that error is gone. However, there's another error to be fixed by
https://github.com/apache/incubator-mxnet/pull/18183

PR #18183, opened 2020-04-28
Fix Dockerfile.publish.test.centos7

leezu:
Delete lines missed in prior commit & add CI test to prevent enforce no such mistakes are made in future

### DIFF
--- a/ci/jenkins/Jenkins_steps.groovy
+++ b/ci/jenkins/Jenkins_steps.groovy
@@ -1764,4 +1764,16 @@ def test_artifact_repository() {
     }]
 }
 
+def misc_test_docker_cache_build() {
+  return ['Test Docker cache build': {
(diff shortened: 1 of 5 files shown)


ChaiBapchya:
Thanks for adding that test so as to ensure docker-cache issues are caught earlier.

leezu:
Mainly the tests ensures that all dockerfiles are valid. Not only the ones used in some CI pipeline, but also the ones only used in CD

ci/jenkins/Jenkins_steps.groovy
```diff
@@ -1764,4 +1764,16 @@ def test_artifact_repository() {
     }]
 }
 
+def misc_test_docker_cache_build() {
```
marcoabreu, line 1767: This is duplicate to the docker cache file. Can you use the same script?

PR #18195, opened 2020-04-29
Specify mxnetci dockerhub user in docker-compose.yml

leezu:
As `docker-compose push` is now used to push the respective images to the dockerhub, we must specify the username in the `docker-compose.yml`.

Reference http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/restricted-docker-cache-refresh/detail/master/2841/pipeline where CI attempts to push the image to the docker.io/library/build.centos7_cpu repository instead of docker.io/mxnetci/build.centos7_cpu

### DIFF
--- a/ci/docker/docker-compose.yml
+++ b/ci/docker/docker-compose.yml
@@ -28,8 +28,9 @@ services:
   ###################################################################################################
   centos7_cpu:
     # The resulting image will be named build.centos7_cpu:latest and will be
-    # pushed to mxnetci under this name
-    image: mxnetci/build.centos7_cpu:latest
+    # pushed to the dockerhub user specified in the environment variable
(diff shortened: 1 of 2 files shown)


ci/docker/docker-compose.yml
```diff
@@ -29,7 +29,7 @@ services:
   centos7_cpu:
     # The resulting image will be named build.centos7_cpu:latest and will be
     # pushed to mxnetci under this name
-    image: build.centos7_cpu:latest
+    image: mxnetci/build.centos7_cpu:latest
```
marcoabreu, line 32: Can we parametrize this? This code is not testable in another environment

leezu: I don't think that's possible

leezu: https://docs.docker.com/compose/compose-file/#variable-substitution

PR #18204, opened 2020-04-30
Disable -DUSE_TVM_OP on GPU builds

leezu:
Need to be fixed first before it can be tested on CI

https://github.com/apache/incubator-mxnet/issues/17886
#17840

### DIFF
--- a/ci/docker/runtime_functions.sh
+++ b/ci/docker/runtime_functions.sh
@@ -723,7 +723,6 @@ build_ubuntu_gpu_mkldnn() {
     CC=gcc-7 CXX=g++-7 cmake \
         -DCMAKE_BUILD_TYPE="RelWithDebInfo" \
         -DUSE_MKL_IF_AVAILABLE=OFF \
-        -DUSE_TVM_OP=ON \
         -DUSE_CUDA=ON \
         -DMXNET_CUDA_ARCH="$CI_CMAKE_CUDA_ARCH" \
         -DUSE_CPP_PACKAGE=ON \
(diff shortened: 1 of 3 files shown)


marcoabreu:
Is this issue caused by using the g4 instances or do we have and clue of why they saw suddenly broken?

leezu:
This has always been broken. Also, CI hasn't switched to G4 yet

leezu:
@mxnet-bot run ci [unix-gpu, windows-gpu, centos-gpu]

marcoabreu:
So how did CI pass if it was broken? 🤔

leezu:
> So how did CI pass if it was broken? thinking

Those features were not tested by CI. For some other tests that do rely on the broken feature, they are disabled, for example

https://github.com/apache/incubator-mxnet/blob/1496c91871b9d81d6a18785bdc8a1c3450bedbca/tests/python/unittest/test_deferred_compute.py#L327-L329

But the functionality in https://github.com/apache/incubator-mxnet/pull/18083 uses the broken features, and at this stage I think we should disable the broken feature instead of changing the way https://github.com/apache/incubator-mxnet/pull/18083 uses the mx.np API

jinboci:
@leezu Hi, I am fixing the issue #17840. Would reverting changes in this PR reenable all the TVMOp tests?

leezu:
Thank you @jinboci for fixing the issue! Reverting this PR will indeed re-enable the previous test setup, but there may be some conflicts during the revert due to subsequent changes to the CI. You can try it though. 

Instead of reverting, my recommendation is to take a look at what tests this PR has removed and think about what the best strategy for testing the TVM_OP feature would be. For example, we can consider to test it only on the CentOS system tests or only on the Ubuntu system tests. The correct testing strategy may also depend on.

Sorry for the late response.

jinboci:
@leezu Thank you very much for your suggestions. I don't have recommendations about testing strategy yet; however, I will take a look at the test code and see if there are better ways for testing. Thanks!

PR #18222, opened 2020-05-02
Fix PyPI Packages and Python Docker Images nightly release

mseth10:
## Description ##
Nightly PyPI packages are currently not getting published to https://dist.mxnet.io/python/all

There is an error in pipeline Build stage for GPU variants [link](http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/restricted-mxnet-cd%2Fmxnet-cd-release-job/detail/mxnet-cd-release-job/1053/pipeline/77). The logs state the following error - `ERROR: No artifacts found for this configuration.`

The binaries for all variants are present in s3, but GPU builds couldn't find them as they are searching at the wrong path. This issue arises because the s3 bucket path contains OS version, which is different during upload and download of GPU binaries. This PR fixes this issue by hard-coding OS to be the same for both upload and download.

This also resolves a similar error in Python docker images pipeline [link](http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/restricted-mxnet-cd%2Fmxnet-cd-release-job/detail/mxnet-cd-release-job/1054/pipeline/124)

There is another error in both pipelines: [link1](http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/restricted-mxnet-cd%2Fmxnet-cd-release-job/detail/mxnet-cd-release-job/1053/pipeline/308) [link2](http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/restricted-mxnet-cd%2Fmxnet-cd-release-job/detail/mxnet-cd-release-job/1054/pipeline/62), which is due to a `SyntaxError` caused by incorrect bash script trigger command. This PR also resolves these bugs.

### DIFF
--- a/cd/utils/artifact_repository.py
+++ b/cd/utils/artifact_repository.py
@@ -396,7 +396,7 @@ def get_s3_key_prefix(args: argparse.Namespace, subdir: str = '') -> str:
     :param subdir: An optional subdirectory in which to store the files. Post-pended to the end of the prefix.
     :return: A string containing the S3 key prefix to be used to uploading and downloading files to the artifact repository
     """
-    prefix = "{git_sha}/{libtype}/{variant}/".format(**vars(args))
+    prefix = "{git_sha}/{libtype}/{os}/{variant}/".format(**vars(args))
     if subdir:
         return "{}{}/".format(prefix, subdir)
     return prefix
--- a/cd/utils/artifact_repository.py
+++ b/cd/utils/artifact_repository.py
@@ -396,7 +396,7 @@ def get_s3_key_prefix(args: argparse.Namespace, subdir: str = '') -> str:
     :param subdir: An optional subdirectory in which to store the files. Post-pended to the end of the prefix.
     :return: A string containing the S3 key prefix to be used to uploading and downloading files to the artifact repository
     """
-    prefix = "{git_sha}/{libtype}/{os}/{variant}/".format(**vars(args))
+    prefix = "{git_sha}/{libtype}/{variant}/".format(**vars(args))
     if subdir:
         return "{}{}/".format(prefix, subdir)
     return prefix
--- a/cd/Jenkinsfile_utils.groovy
+++ b/cd/Jenkinsfile_utils.groovy
@@ -116,12 +116,12 @@ def push_artifact(libmxnet_path, variant, libtype, license_paths = '', dependenc
   if(license_paths == null) license_paths = ''
   if(dependency_paths == null) dependency_paths = ''
 
-  sh "python3 ./cd/utils/artifact_repository.py --push --verbose --libtype ${libtype} --variant ${variant} --libmxnet ${libmxnet_path} --licenses ${license_paths} --dependencies ${dependency_paths}"
+  sh "python3 ./cd/utils/artifact_repository.py --push --verbose --libtype ${libtype} --variant ${variant} --libmxnet ${libmxnet_path} --licenses ${license_paths} --dependencies ${dependency_paths} --os ubuntu18.04"
 }
 
 // pull artifact from repository
 def pull_artifact(variant, libtype, destination = '') {
-  sh "python3 ./cd/utils/artifact_repository.py --pull --verbose --libtype ${libtype} --variant ${variant} --destination ${destination}"
+  sh "python3 ./cd/utils/artifact_repository.py --pull --verbose --libtype ${libtype} --variant ${variant} --destination ${destination} --os ubuntu18.04"
 }
 
 // pulls artifact from repository and places files in the appropriate directories
--- a/cd/python/docker/Jenkins_pipeline.groovy
+++ b/cd/python/docker/Jenkins_pipeline.groovy
@@ -49,21 +49,21 @@ def build(mxnet_variant) {
     ci_utils.docker_run(environment, "cd_package_pypi ${mxnet_variant}", nvidia_docker)
 
     // build python docker images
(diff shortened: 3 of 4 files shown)


leezu:

cd/utils/artifact_repository.py
```diff
@@ -396,7 +396,7 @@ def get_s3_key_prefix(args: argparse.Namespace, subdir: str = '') -> str:
     :param subdir: An optional subdirectory in which to store the files. Post-pended to the end of the prefix.
     :return: A string containing the S3 key prefix to be used to uploading and downloading files to the artifact repository
     """
-    prefix = "{git_sha}/{libtype}/{os}/{variant}/".format(**vars(args))
+    prefix = "{git_sha}/{libtype}/{variant}/".format(**vars(args))
```
marcoabreu, line 399: I don't understand how this is a fix. The OS was put there for a reason apparently, so I'd first like to see a deep dive to understand the reasoning behind that parameter before removing it.

mseth10: @marcoabreu that's a valid concern. Thanks for raising it. Here's what I found after a deep dive:
The binaries for both CPU and GPU builds are pushed using a CPU ubuntu18.04 instance so they are present in the s3 bucket folder ubuntu18.04. The release pipeline uses GPU instances, currently running ubuntu16.04, for the cu* variants and thus fail to pull those artifacts.
The efforts to upgrade the GPU AMIs to ubuntu18.04 are underway. Meanwhile we need to get the nightly packages released due to significant customer impact.

mseth10: Also, having OS in the s3 bucket prefix does not make sense. Please let me know if I am missing something.

leezu: @marcoabreu this is broken because there is no reason for different stages of the CD pipeline the exact same OS version in their AMI. Perhaps the author intended to take the OS used for building the binary into account. But that's separate from the AMI and in either case not needed. Thus deletion of that unneeded complexity seems to be the best path forward.

leezu: Thus please substantiate your statement "The OS was put there for a reason". Unsubstantiated veto is invalid.

marcoabreu: Creating a binary on unix, windows, arm etc will result in different binaries which have to be differentiated. It might be a valid point that the os should not be the differentiator, but it should be properly addressed (or shown) that these cases are properly addressed. 

When removing something, it's up to the PR creator to prove that it will not cause any harm and that the initial design decision are not broken - or otherwise explicitly state it and make an alternative proposal. "fix" is not a valid argument or proposal.

leezu: The existing setup does not build binaries for different OS and different architectures. Even when building binaries for other platforms, there is no reason for the build platform to be the same as the target platform. 

The current design is not designed correctly and just looks at the build platform, ie `cat /etc/os-release`. It needs to be fixed to support the use-cases you mention, though we don't support them at this point.

As it's broken by design, I suggest to remove broken functionality. Though I agree deleting `{os}` does not amount to removing the broken functionality (also `probe_operating_system` etc would need to be removed)

leezu: To exemplify the problem: We're building binaries on CentOS7. However, the current design places them in an `ubuntu1804` bucket and later fails testing the binaries because the test system runs `ubuntu1604` host OS.

marcoabreu: Okay I understand, thanks for elaborating!

Last question: how do we plan to address the different platforms I mentioned as example after this PR? My understanding is that they would no longer be differentiated - speak we could only generate on platform, right?

leezu: I agree with you that we need a naming scheme to differentiate the different platforms. The problem currently is just that the platform name as currently derived is not meaningful. But just as we can delete the `{os}` now, we can reintroduce a platform identifier once we do actually build for multiple platforms on CD. @mseth10 may add a pipeline for Jetson devices later, so that would be a good opportunity to fix the naming scheme.

marcoabreu: My concern here is that we'd be changing the directory structure in s3 - which we then have to fix later on. 

I'd propose to either introduce the platform identifier immediately or replace the os parameter with a matching value (replace ubuntu1804 with Ubuntu 1604) as a hotfix. But I'd prefer if we do not change the directory structure in the underlying system as that'd be something we have to clean up later.

leezu: We can hardcode some value for `{os}` instead of deleting it, sure. Though this is a internal `ARTIFACT_REPOSITORY_BUCKET` s3 bucket, which is private and arguably should use object expiration as access to old objects is not needed (later step of the CD publishes the proper build artifacts to a public bucket). Thus I  don't think we have to make any guarantees about a consistent structure over time.

I have some more concerns about the structure of the CD, but I'll open a separate issue to discuss them.

leezu: https://github.com/apache/incubator-mxnet/issues/18233

mseth10: As per our discussion, I have reverted commit that removes {os} from s3 path. I have hard-coded `os` to `ubuntu18.04` for both upload and download of binaries.
@marcoabreu can you please take a look.

szha: @perdasilva may be able to clarify

szha:
@mseth10 thanks for posting the fix. @marcoabreu @leezu the CI check has passed. How shall we move forward?

PR #18240, opened 2020-05-05
[CI] Enhance debug info

szha:
## Description ##
Enhance debug info

### Essentials ###

### Changes ###
- [x] add pytest-faulthandler
- [x] check for gpu debug build
- [x] correct debug bit in feature list


ci/jenkins/Jenkins_steps.groovy
```diff
@@ -80,278 +80,294 @@ def python3_gpu_ut_cython(docker_container_name) {
 
 //------------------------------------------------------------------------------------------
 
-def compile_unix_cpu_openblas() {
+def compile_unix_cpu_openblas(lib_name) {
```
leezu, line 83: Why make this a parameter? In its current design, lib_name depends on the actual compilation. So it's not interchangeable and you can't just call the same function with different lib_name. 

In a bigger picture, most lib_name should be deleted, consolidated to a single one based on having a single directory in which the build places all artifacts (install target)

docker/install/python.sh
```diff
@@ -24,4 +24,4 @@ apt-get update && apt-get install -y python3-dev
 # the version of the pip shipped with ubuntu may be too lower, install a recent version here
 cd /tmp && wget https://bootstrap.pypa.io/get-pip.py && python3 get-pip.py
 
-pip3 install pylint numpy requests Pillow pytest==5.3.2 pytest-env==0.6.2 pytest-cov==2.8.1 pytest-xdist==1.31.0
+pip3 install pylint numpy requests Pillow pytest==5.3.2 pytest-env==0.6.2 pytest-cov==2.8.1 pytest-xdist==1.31.0 pytest-faulthandler==2.0.1
```
leezu, line 27: "This plugin is now part of pytest core since pytest 5.0, so users should not install this plugin together with that pytest version."  https://github.com/pytest-dev/pytest-faulthandler

Issue #15533, opened 2019-07-13
The problems about SGD with momentum when learning rate changes


wkcn:
Hi, there.
Currently, SGD(Stochastic Gradient Descent) in MXNet is applied by:
```
rescaled_grad = lr * (rescale_grad * clip(grad, clip_gradient) + wd * weight)
state = momentum * state + rescaled_grad
weight = weight - state
```
I found there are two problems in SGD.
1. Loss of accuracy on float-point number
For the SGD with momentum, the variable `state` stores the gradients multiplied by learning rate. However, learning rate is usually a small value, such as 1e-3, enabling the `state` becomes smaller than the gradient. It may loss the accuracy.

2. The case when learning rate changes.
When learning rate changes, the variable `state` stores the gradients multiplied by old learning rate. It is wrong.

Solution:
We should update the implement of SGD with momentum, but we should consider the compatibility with old optimizer states.
```
rescaled_grad = rescale_grad * clip(grad, clip_gradient) + wd * weight
state = momentum * state + rescaled_grad
weight = weight - lr * state
```

ptrendx:
Additional point here is that both TF and pyTorch use that proposed SGD logic, different than MXNet.

I am looking into making that change, I'm not sure if we need to be concerned about the backward compatibility for 2.0.

FYI @shengac since you did a legwork for refactoring optimizers for  2.0, @zhreshold and @eric-haibin-lin  for GluonCV and GluonNLP visibility.

zhreshold:
I also agree that the compatibility of such change is not forced in the upcoming 2.0, and it's extremely rare that you would save/load the states using different mxnet versions.

szhengac:
Putting lr inside the momentum is how Momentum SGD or NAG looks like originally. But putting it outside momentum may be more robust to large batch training in warmup stage.

PR #18285, opened 2020-05-12
Put lr outside momentum for sgd and nag

szhengac:
## Description ##
Change the implementation of momentum SGD and NAG. Specifically, The original implementation is correct but not consistent with Pytorch and TF. 

See #15533 

@zhreshold can you see how many CV models need to be retrained with the change?

### DIFF
--- a/python/mxnet/optimizer/nag.py
+++ b/python/mxnet/optimizer/nag.py
@@ -100,13 +100,13 @@ class NAG(Optimizer):
             mom = state
             if mom is not None:
                 mom[:] *= self.momentum
-                mom[:] -= lr * grad
-                d = self.momentum * mom - lr * grad
+                mom[:] -= grad
+                d = self.momentum * mom - grad
             else:
-                d = -lr * grad
+                d = -grad
 
             # update weight
-            weight[:] += d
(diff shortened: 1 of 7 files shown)


zhreshold:
If both sgd and nag are going to be modified the all models in gluoncv are required to explore new hyperparameters. 

Is it sensible to keep the original versions with the new implementation renamed?

szhengac:
I think it would be ambiguous. It is better to only keep one.

szhengac:
If all models need to be retrained, I think it is better to keep the
current implementation.

nickguletskii:
On one hand, there are existing MXNet models which would require hyperparameter tuning to be done again if this change is implemented.
On the other hand, being inconsistent with PyTorch and other deep learning frameworks makes it harder to port models and the subtle differences might result in slight differences in performance.
Even though having an additional flag to switch between old and new behaviour definitely adds to complexity, I am not sure that it is not worth it.

szha:
Since the change is to mxnet 2.0 where gluoncv is expected to adopt the new numpy-based API, it's expected that model will need to be retrained. Let's merge this change on in 2.0 dev branch and keep it out of 1.x

Issue #18099, opened 2020-04-19
Error in test_contrib_amp.py::test_amp_coverage


szha:
## Description
As part of #18025 I added a waitall() in between tests in amp. This revealed the following error in unix-gpu pipeline which seems to be related to pointwise fusion.

http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-gpu/detail/PR-18025/34/pipeline/417

```
==================================== ERRORS ====================================
____________________ ERROR at teardown of test_amp_coverage ____________________

    def teardown():
>       mx.nd.waitall()

tests/python/gpu/test_contrib_amp.py:42: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
python/mxnet/ndarray/ndarray.py:211: in waitall
    check_call(_LIB.MXNDArrayWaitAll())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

ret = -1

    def check_call(ret):
        """Check the return value of C API call.
    
        This function will raise an exception when an error occurs.
        Wrap every API call with this function.
    
        Parameters
        ----------
        ret : int
            return value from API calls.
        """
        if ret != 0:
>           raise get_last_ffi_error()
E           mxnet.base.MXNetError: Traceback (most recent call last):
E             [bt] (9) /work/mxnet/python/mxnet/../../build/libmxnet.so(std::thread::_State_impl<std::thread::_Invoker<std::tuple<std::function<void (std::shared_ptr<dmlc::ManualEvent>)>, std::shared_ptr<dmlc::ManualEvent> > > >::_M_run()+0x4a) [0x7fb2e5fce4da]
[...]

PR #18292, opened 2020-05-12
Enable test_amp_conversion

leezu:
TVMOP feature is now disabled on GPU builds, which caused this test to fail previously

Fixes https://github.com/apache/incubator-mxnet/issues/18099


leezu:
@mxnet-bot run ci [unix-cpu]

PR #18097, opened 2020-04-18
Change LC_NUMERIC for CentOS CI jobs to verify locale invariance

nickguletskii:
## Description ##
This pull request changes all tests to load the user's locale and changes the CentOS CI jobs to run with `LC_NUMERIC` set to `en_DK.UTF8`. `en_DK.UTF8` is a locale that uses `,` as the decimal separator. This should weed out any bugs similar to #16134, #17140 and #18079. A partial fix for v1.x is available in #17177. Migrating operators and other API calls to the new FFI in MXNet 2.0 should resolve most locale-related issues related to API calls.

This pull request makes many tests fail because they only pass on systems with `LC_NUMERIC` set to a locale that uses `.` as the decimal separator.

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [x] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at https://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [x] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [x] Tests updated to call `locale.setlocale` on initialization.
- [x] CentOS jobs updated to run with `LC_NUMERIC` set to `en_DK.UTF8`.
- [x] xfail/cancel tests which fail due to the incorrect handling of decimal separators.

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #18307, opened 2020-05-13
Fix test_quantization to unblock CD

rondogency:
This PR fixes 2 quantization gpu tests: test_quantize_model and test_quantize_sym_with_calib. Those tests was not marked as xfail in the locale fix in #18097 

Tested in a EC2 and it should unblock CD as CD pipeline fails on those 2 tests

This PR is not integrating test_quantization into Centos CI pipeline, and we should work on it in another PR


rondogency:
@mxnet-bot run ci [unix-cpu]

Issue #17327, opened 2020-01-15
np.ndarray indexing after hybridize


ZheyuYe:
## Description
Indexing is an important feature of numpy that supports complex positioning and value operations as [Numpy | Indexing](https://www.geeksforgeeks.org/numpy-indexing/). The mxnet deepnumpy only supply the basic indexing operation as
```
gathered_data = sequence[indices_x, indices_y]
```
whereas the above operation will fail after hybridize, see [here](https://github.com/apache/incubator-mxnet/pull/16621). Currently, the more effective method is only [mx.npx.pick](https://numpy.mxnet.io/api/deepnumpy/generated/mxnet.npx.pick.html), but it only supports one-dimensional selection instead of providing gather_nd operation( for nd.ndarray) can select array elements multidimensionally.

The temporary solution would be adding the gather_nd operation for mx.np.ndarray, and the hybridized indexing opertaion could be added in future work.
@sxjscience @haojin2

sxjscience:
@ZheyuYe In terms of `mx.npx.pick` we can support multiple-dimensional selection as long as we are selecting the leading ones:

```python
import mxnet as mx
import mxnet.numpy as np
mx.npx.set_np()
a = mx.np.random.uniform(-1, 1, (10, 10, 5))
b = np.random.randint(0, 5, (10,)).broadcast_to((10, 10))
mx.npx.pick(a, b) 
```

Also, would you provide a reproducible code snippet (like above) to mention which type of indexing operation is missing?

ZheyuYe:
@sxjscience 

```
import numpy as np
import mxnet as mx
import numpy.testing as npt

sequence = mx.nd.array(np.random.normal(0, 1, (8, 32, 768)), dtype=np.float32)
# pick_ids: [batch_size, picked_index]
pick_ids = mx.nd.random.randint(0, 31, (8,2), dtype=np.int32)

# original approach with nd.ndarray
idx_arange = mx.nd.contrib.arange_like(pick_ids.reshape((-1, )), axis=0)
batch_idx = mx.nd.broadcast_div(idx_arange, mx.nd.array([2],dtype=np.int32))
# batch_idx_1d = [0,0,1,1,2,2,...]
# pick_ids_1d = [1,2,4,0,3,4,2,3,5...]
batch_idx_1d = batch_idx.reshape((1, -1))
pick_ids_1d = pick_ids.reshape((1, -1))
[...]

PR #18319, opened 2020-05-14
[numpy] symbolic advanced indexing

AntiZpvoh:
## Description ##

Add symbolic advanced indexing for DeepNumPy
Solve the issue https://github.com/apache/incubator-mxnet/issues/17327
For now boolean advanced indexing and new axis case indexing are not supported. Feature request is https://github.com/apache/incubator-mxnet/issues/18627.
### Essentials ###

### Changes ###

## Comments ##


src/operator/numpy/np_indexing_op.cc
```diff
@@ -0,0 +1,572 @@
...
+    const NDArray &idx = inputs[1];
+    const NDArray &out = outputs[0];
+    CHECK_EQ(data.shape()[0], idx.shape()[0]);
+    CHECK_EQ(idx.shape().ndim(), 1U);  // idx is required to be 1-d.
+    // count the number of 1s in `idx`, so that we could know the output dimension
+    size_t idx_size = idx.shape()[0];
+    std::vector<int32_t> prefix_sum(idx_size, 0);
+    size_t valid_num = 0;
+    // Calculate prefix sum
+    MSHADOW_TYPE_SWITCH_WITH_BOOL(idx.dtype(), DType, {
```
sxjscience, line 179: There is no need to use `MSHADOW_TYPE_SWITCH_WITH_BOOL` since we have already tested that it's a bool type.

src/operator/numpy/np_indexing_op.h
```diff
@@ -0,0 +1,84 @@
...
+#include "../tensor/indexing_op-inl.cuh"
+#endif
+
+namespace mxnet {
+namespace op {
+
+namespace np_indexing_ {  // to avoid name conflict
+enum Inputs {kArr, kIdx};
+enum Outputs {kOut};
+}  // namespace np_indexing_
```
sxjscience, line 42: Will we reuse these enums? Can we put them to `.cc`?

AntiZpvoh:
Hi @leezu ! It seems in the deferred compute mode data is imperative and index is symbolic. My operators work only when both data and index are symbolic or imperative. I think these operators can not use in the deferred compute mode.

src/operator/numpy/np_indexing_op.h
AntiZpvoh: We need use them in `.cu` file.

leezu:
@AntiZpvoh on the operator side in the backend there is no distinction of "imperative" or "symbolic". All operators just take some array as input and compute some output. Thus your operators will also work for "imperative" inputs.

The deferred compute provides different approach to record a symbolic graph. You can focus on rewriting the advanced indexing in the imperative API based on your operators and the deferred compute part will work automatically.

sxjscience:
@AntiZpvoh Would you add some tests?

src/operator/numpy/np_indexing_op.cc
```diff
@@ -0,0 +1,544 @@
...
+          Kernel<pick_grad<3, true>, cpu>::Launch(s, inputs[0].data().Size(),
+          outputs[0].data().dptr<DType>(), inputs[0].data().dptr<DType>(),
+          inputs[2].data().dptr<IType>(), M,
+          trailing, Shape3(leading, M, trailing),
+          Shape3(leading, 1, trailing));
+        }
+      });
+    });
+  } else {
+    dmlc::LogMessageFatal(__FILE__, __LINE__).stream()
```
sxjscience, line 455: Is it possible to directly use `LOG(FATAL)`?

src/operator/numpy/np_indexing_op.cu
```diff
@@ -0,0 +1,574 @@
...
+        oshape[i + idxshape.ndim() - 1] = arrshape[i];
+        }
+    }
+
+    const NDArray &out = outputs[0];
+    const_cast<NDArray &>(out).Init(oshape);
+
+    Stream<gpu> *s = ctx.get_stream<gpu>();
+
+    MSHADOW_TYPE_SWITCH_WITH_BOOL(outputs[np_indexing_::kOut].dtype(), DType, {  // output data type
```
sxjscience, line 190: I think we may use the `MSHADOW_TYPE_SWITCH` version without bool. Because the bool type won't be used.

src/operator/numpy/np_indexing_op.cu
```diff
@@ -0,0 +1,574 @@
...
+};
+
+template<typename DType>
+bool CheckIndexOutOfBound(mshadow::Stream<gpu> *s, const DType* data_ptr, size_t data_size,
+                            const DType min, const DType max, char* is_valid_ptr) {
+    using namespace mxnet_op;
+    int32_t is_valid = 0;
+    Kernel<set_zero, gpu>::Launch(s, 1, is_valid_ptr);
+    Kernel<is_valid_check, gpu>::Launch(s, data_size, is_valid_ptr, data_ptr, min, max);
+    CUDA_CALL(cudaMemcpyAsync(&is_valid, is_valid_ptr, sizeof(char),
```
sxjscience, line 47: Here, `is_valid` has dtype=`int32_t`, but the `is_valid_ptr` has dtype=`char`. Thus, you may consider to change the dtype of is_valid to char.

src/operator/numpy/np_indexing_op.cu
```diff
@@ -0,0 +1,574 @@
...
+                                    idx_size,
+                                    stream);
+    size_t buffer_size = idx_size * sizeof(int32_t);
+    temp_storage_bytes += buffer_size;
+    // Allocate memory on GPU and allocate pointer
+    Tensor<gpu, 1, char> workspace =
+        ctx.requested[0].get_space_typed<gpu, 1, char>(Shape1(temp_storage_bytes), s);
+    prefix_sum = reinterpret_cast<int32_t*>(workspace.dptr_);
+    d_temp_storage = workspace.dptr_ + buffer_size;
+    MSHADOW_TYPE_SWITCH_WITH_BOOL(idx.dtype(), IType, {
```
sxjscience, line 132: You may also simplify the `MSHADOW_TYPE_SWITCH_WITH_BOOL` here. This is inside the check of `inputs[np_indexing_::kIdx].dtype() == mshadow::kBool`. Thus, IType will always be bool

src/operator/numpy/np_indexing_op.cc
```diff
@@ -0,0 +1,544 @@
...
+    << "arrays used as indices must be explictly declared as integer (or boolean) type."
+    << "Use np.astype() to cast indices to integer or boolean.";
+  }
+}
+
+NNVM_REGISTER_OP(_npi_advanced_indexing)
+.describe(R"code(
+Given an n-d NDArray data, and a 1-d NDArray index,
+the operator produces an un-predeterminable shaped n-d NDArray out,
+which stands for the rows in x where the corresonding element in index is non-zero.
```
sxjscience, line 465: Revise the documentation.

sxjscience:
@leezu A separate issue has been created in https://github.com/apache/incubator-mxnet/issues/18627 by @AntiZpvoh. Basically, the current advanced indexing misses some feature supports and cannot cover all cases of imperative indexing. Thus, revising the multiarray.py should take some more efforts.

leezu:
@sxjscience @AntiZpvoh thank you for investigating and clarifying the shortcomings. To ensure consistency between the imperative and symbolic interface, how about adding an `if` statement between lines 722 and 723

https://github.com/apache/incubator-mxnet/blob/becb9ca694f51fdc0583d58429ccc943e6462810/python/mxnet/numpy/multiarray.py#L722-L730

to call the operator introduced by @AntiZpvoh when inputs are not boolean. For boolean inputs, we could keep the existing imperative logic. This should be quite straightforward, but please let me know if you disagree.

Thereby we prevent divergence of the two interfaces, minimize the chance of subtle bugs and make the code more maintainable by working towards a fully shared implementation.

AntiZpvoh:

AntiZpvoh:
Hi @leezu. I have added an if statement between lines 722 and 723 in the most recent commit. 
https://github.com/apache/incubator-mxnet/blob/380e122c84bdf512e5f7ff6442900ae0782a5ad6/python/mxnet/numpy/multiarray.py#L722-L744
I think now this PR could ensure consistency between symbolic and imperative advanced indexing.

AntiZpvoh:
@mxnet-bot run ci [unix-gpu]

AntiZpvoh:
@mxnet-bot run ci [unix-gpu]

AntiZpvoh:
@mxnet-bot run ci [unix-gpu]

python/mxnet/numpy/multiarray.py
```diff
@@ -720,6 +720,28 @@ def __getitem__(self, key):
                 key = (_np.newaxis,) + key
             return self._get_np_basic_indexing(key)
         elif indexing_dispatch_code == _NDARRAY_ADVANCED_INDEXING:
+            if prepend == _NDARRAY_NO_ZERO_DIM_BOOL_ARRAY and isinstance(key, (NDArray, _np.ndarray)):
```
leezu, line 723: The `isinstance(key, (NDArray, _np.ndarray)` check here would be equivalent to the `isinstance(key, Symbol)` check in https://github.com/apache/incubator-mxnet/pull/18319/files#diff-e9b526dba36aa6f2e7dbb6c4d49e9822R159-R160

Can you elaborate why you call the `self._get_index_nd` here instead of directly calling `_npi.advanced_indexing(self, key)` as in the Symbol case?

sxjscience: I think this is for boolean indexing.

leezu: So boolean indexing is now supported by `_npi.advanced_indexing`?

sxjscience: I think this is to tell whether there is 0-dim/boolean indexing in the key.

sxjscience: @leezu https://github.com/apache/incubator-mxnet/blob/a9b16f7024878611b236c9f3734ccd37a5a35d38/python/mxnet/ndarray/ndarray.py#L116-L119

leezu: Yes. My question is that it appears that `indexing_key_expand_implicit_axes` messes up the `key` as it supports a much more generic case. It may be easier to have the `if` condition and call to `_npi.advanced_indexing` prior to the `indexing_key_expand_implicit_axes`. In that case the code may be simply

```
if isinstance(key, (NDArray, _np.ndarray) and key.dtype is not _np.bool:
    _npi.advanced_indexing(self, key)
```

But maybe I'm missing something, so clarification is appreciated. Thanks

AntiZpvoh: Thanks for reviewing my code @leezu ! Actually for now my op does not cover newaxis indexing case in backend since the primary purpose of this PR is to solve the issue #17327 (support indexing like sequence[indices_x, indices_y] after hybridized). This is why this if statement is so complex.

AntiZpvoh:
@mxnet-bot run ci [unix-gpu]

AntiZpvoh:

python/mxnet/numpy/multiarray.py
leezu: @AntiZpvoh Thank you for clarifying! However I'm still not clear why the ndarray codepath needs to be more complex than the symbol one for the case that you ensure that the key is either a single ndarray or a tuple of ndarrays. These are the two cases supported in the symbol interface, so you don't need to worry about other cases in the imperative interface.

AntiZpvoh: @leezu My apology I found my mistake. It should be `isinstance(key, tuple)` instead of `isinstance(key, (NDArray, _np.ndarray))`. In that case codepath needs to be complex since we need to ensure the element of tuple is integer NDArray or np.ndarray (cannot be list maybe because list could not be symbolized) and also handle np.newaxis and implicit axis cases. The newest commit has already corrected this mistake.

leezu: > In that case codepath needs to be complex since we need to ensure the element of tuple is integer NDArray or np.ndarray (cannot be list maybe because list could not be symbolized) and also handle np.newaxis and implicit axis cases. The newest commit has already corrected this mistake.

Thanks for fixing the mistake. My concern is that you need to handle the special cases "tuple of NDArray" without `_get_index_nd` and even without `indexing_key_expand_implicit_axes`. Consider your current test file `test_gluon_indexing.py`. It will pass the `key = (array([0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7, 7], dtype=int32), array([27, 30, 24, 26,  1, 23, 12, 25, 29, 22, 21, 31,  6, 31, 11,  6], dtype=int32))`, but after you call `indexing_key_expand_implicit_axes` the key is transformed to `(array([0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7, 7], dtype=int32), array([27, 30, 24, 26,  1, 23, 12, 25, 29, 22, 21, 31,  6, 31, 11,  6], dtype=int32), slice(None, None, None))` and your condition `if is_int_array_tuple` will never be True and `_npi.advanced_indexing_multiple` will never be called..

python/mxnet/numpy/multiarray.py
```diff
@@ -720,6 +720,30 @@ def __getitem__(self, key):
...
+            if is_int_array_tuple:
+                sliced = _npi.advanced_indexing_multiple(self, idcs)
+                # Reshape due to `None` entries in `key`.
+                if new_axes:
+                    final_shape = [sliced.shape[i] for i in range(sliced.ndim)]
+                    for ax in new_axes:  # pylint: disable=invalid-name
+                        final_shape.insert(ax, 1)
+                    return sliced.reshape(tuple(final_shape))
+                else:
+                    return sliced
```
leezu, line 746:
```suggestion

```

Suggest to delete this codeblock and replace by the following in line 702:

```
        all = __builtins__['all']  # `def all` below shadows the all builtin
        if isinstance(key, tuple) and \
           all((isinstance(arr, NDArray) and _np.issubdtype(arr.dtype, _np.integer)) for arr in key):
            return _npi.advanced_indexing_multiple(self, _npi.stack(*key))
```

AntiZpvoh:

AntiZpvoh:

leezu:
@mxnet-bot run ci [unix-cpu]

PR #18324, opened 2020-05-14
[OpPerf] Add example of using opperf with internal op locally

ChaiBapchya:
## Description ##
Currently opperf supports for mx.nd operators alone.
However, locally, one can test internal ops. Documenting it in readme.

@access2rohit @sandeep-krishnamurthy @connorgoggins

### DIFF
--- a/benchmark/opperf/README.md
+++ b/benchmark/opperf/README.md
@@ -167,6 +167,41 @@ Output for the above benchmark run, on a CPU machine, would look something like 
              ]}
 
 ```
+
+## Usecase 5 - Profile internal operators locally
+Currently, opperf supports operators in `mx.nd.*` namespace.
+However, locally, one can profile internal operators in `mx.nd.internal.*` namespace.
+
+#### Changes
+Remove the hasattr check for `op.__name__` to be in `mx.nd`
+
+The resulting diff would look like :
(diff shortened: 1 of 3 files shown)


benchmark/opperf/README.md
```diff
@@ -167,6 +167,41 @@ Output for the above benchmark run, on a CPU machine, would look something like
...
+
+## Usecase 5 - Profile internal operators locally
+Currently, opperf supports operators in `mx.nd.*` namespace.
+However, locally, one can profile internal operators in `mx.nd.internal.*` namespace.
+
+#### Changes
+Remove the hasattr check for `op.__name__` to be in `mx.nd`
+
+The resulting diff would look like :
+```
```
access2rohit, line 179: Do you think rephrasing it to sth like from "old code" and to "new code" would be easier instead on git diff ? Also separating both via code blocks would be easier to read. Do you agree ?

ChaiBapchya: Yup. I took the lazy approach of pasting the diff. addressed.

ChaiBapchya:
@mxnet-bot run ci [unix-cpu]

ChaiBapchya:

PR #18325, opened 2020-05-14
Change SGD with momentum to include momentum correction by default

### LINKED ISSUE #15533: The problems about SGD with momentum when learning rate changes

ptrendx:
## Description ##
Fixes #15533

Notable change needed to happen to LARS optimizer, since it actually requires the lack of momentum correction, so I had to emulate it in the unaggregated case. I also moved and renamed contrib/preloaded_*_sgd -> lars_*_sgd operators (moved out of contrib because non-contrib functionality depends on them).

@wkcn @zhreshold @shengac FYI

### Essentials ###


python/mxnet/optimizer/lars.py
```diff
@@ -252,6 +253,13 @@ def fused_step(self, indices, weights, grads, states):
                 wd = wds[i]
                 lr = lrs[i]
                 lr *= self._get_lars(index, weight, grad, wd)
+                # normal SGD picks up momentum correction by default,
+                # so need to modify the momentum to undo that.
+                # The correction term is previous_lr / current_lr.
+                kwargs['momentum'] = (self.momentum * (self.last_lr / lr)) \
```
wkcn, line 259: Thank you for the fix!

It will reduce the times of getting an item of `self.last_lr`.
```python
kwargs['momentum'] = (self.momentum * (self.last_lr.get(index, lr) / lr)) \
                                     if lr != 0 else \
                                     self.momentum
```

ptrendx: Thanks, done

zhreshold:
@szhengac Can you provide feedback on this?

szhengac:
Overall, it looks good to me. It overlaps partially with #18285.

Issue #18330, opened 2020-05-15
test_metric_performance


leezu:
http://jenkins.mxnet-ci.amazon-ml.com/blue/rest/organizations/jenkins/pipelines/mxnet-validation/pipelines/unix-cpu/branches/PR-18284/runs/6/nodes/364/steps/731/log/?start=0
```

[2020-05-14T23:42:14.444Z] =================================== FAILURES ===================================
[2020-05-14T23:42:14.444Z] ___________________________ test_metric_performance ____________________________
[2020-05-14T23:42:14.444Z] [gw1] linux -- Python 3.6.9 /usr/bin/python3
[2020-05-14T23:42:14.444Z] 
[2020-05-14T23:42:14.444Z]     def test_metric_performance():
[2020-05-14T23:42:14.444Z]         """ unittest entry for metric performance benchmarking """
[2020-05-14T23:42:14.444Z]         # Each dictionary entry is (metric_name:(kwargs, DataGenClass))
[2020-05-14T23:42:14.444Z]         metrics = [
[2020-05-14T23:42:14.444Z]             ('acc', ({}, MetricDataGen)),
[2020-05-14T23:42:14.444Z]             ('top_k_acc', ({'top_k': 5}, MetricDataGen)),
[2020-05-14T23:42:14.444Z]             ('F1', ({}, F1MetricDataGen)),
[2020-05-14T23:42:14.444Z]             ('Perplexity', ({'ignore_label': -1}, MetricDataGen)),
[2020-05-14T23:42:14.444Z]             ('MAE', ({}, MetricDataGen)),
[2020-05-14T23:42:14.444Z]             ('MSE', ({}, MetricDataGen)),
[2020-05-14T23:42:14.444Z]             ('RMSE', ({}, MetricDataGen)),
[2020-05-14T23:42:14.444Z]             ('ce', ({}, MetricDataGen)),
[2020-05-14T23:42:14.444Z]             ('nll_loss', ({}, MetricDataGen)),
[2020-05-14T23:42:14.444Z]             ('pearsonr', ({}, PearsonMetricDataGen)),
[2020-05-14T23:42:14.444Z]         ]
[2020-05-14T23:42:14.444Z]     
[2020-05-14T23:42:14.444Z]         data_size = 1024 * 128
[2020-05-14T23:42:14.444Z]     
[2020-05-14T23:42:14.444Z]         batch_sizes = [16, 64, 256, 1024]
[...]

Issue #18244, opened 2020-05-05
unix-cpu MKL/MKL-DNN Test Time


szha:
# Description
Since #18146 we introduced parallel testing in CI with the hope of reducing test time. However, during that effort we noticed that the MKL and MKLDNN tests run slower than the setting without MKL or MKLDNN. This issue summarizes the set up and current time difference.

## Setup
The results in this issue come from [this CI run](http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-cpu/detail/PR-18240/2/pipeline/), and the time difference is similar in master branch validation.

To show the results, we compare the following test nodes.

#### [Build](https://github.com/apache/incubator-mxnet/blob/c329a524c520ec2b41aa4bff5ee113ff7221b790/ci/docker/runtime_functions.sh#L390-L405)
```
build_ubuntu_cpu_openblas() {
    set -ex
    cd /work/build
    CXXFLAGS="-Wno-error=strict-overflow" CC=gcc-7 CXX=g++-7 cmake \
        -DCMAKE_BUILD_TYPE="RelWithDebInfo" \
        -DENABLE_TESTCOVERAGE=ON \
        -DUSE_TVM_OP=ON \
        -DUSE_CPP_PACKAGE=ON \
        -DUSE_MKL_IF_AVAILABLE=OFF \
        -DUSE_MKLDNN=OFF \
        -DUSE_CUDA=OFF \
        -DUSE_DIST_KVSTORE=ON \
        -DBUILD_CYTHON_MODULES=ON \
        -G Ninja /work/mxnet
    ninja
}
```

#### [Build](https://github.com/apache/incubator-mxnet/blob/c329a524c520ec2b41aa4bff5ee113ff7221b790/ci/docker/runtime_functions.sh#L425-L438)
```
build_ubuntu_cpu_mkl() {
    set -ex
    cd /work/build
    CC=gcc-7 CXX=g++-7 cmake \
        -DCMAKE_BUILD_TYPE="RelWithDebInfo" \
        -DENABLE_TESTCOVERAGE=ON \
        -DUSE_MKLDNN=OFF \
        -DUSE_CUDA=OFF \
        -DUSE_TVM_OP=ON \
        -DUSE_MKL_IF_AVAILABLE=ON \
        -DUSE_BLAS=MKL \
        -GNinja /work/mxnet
    ninja
}
```

[...]

PR #18252, opened 2020-05-06
[CI] run operator tests with naive engine

szha:
## Description ##
run operator tests with naive engine

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] run operator tests with naive engine

PR #18335, opened 2020-05-15
Skip test_metric_performance on MKL builds

leezu:
https://github.com/apache/incubator-mxnet/issues/18244

metrics now use mxnet operators

Fixes https://github.com/apache/incubator-mxnet/issues/18330

### DIFF
--- a/tests/python/unittest/test_metric_perf.py
+++ b/tests/python/unittest/test_metric_perf.py
@@ -15,12 +15,12 @@
 # specific language governing permissions and limitations
 # under the License.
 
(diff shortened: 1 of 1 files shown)


szha:
I'm moving this to benchmark scripts in #18252

ChaiBapchya:
@leezu lets fix the conflicts & remove it from the benchmark script which Sheng moved it to?

leezu:
It should still be useful in the benchmark scripts. But is there any mechanism to verify it's output?

PR #18350, opened 2020-05-18
Add better partial args/aux handling in symbol optimize_for

Kh4L:
## Description ##
Symbol's `optimize_for` was failing when some parameters are missing in `args` or `aux` due to:
- enforcing both in the Python front-end and back-end
- no way to specify attributes of input 
- partial shape inference failing (also related to #18378 and  #18348 fixes)

### Changes
- Add `skip_infer` option to `symbol.optimize_for`, which skips the pre-partition attributes inference. Can be used for backends that does not need this (e.g. TensorRT).
- Add optional `shape_dict`, `type_dict` and `stype_dict` to `optimize_for` to specify input data attributes for the pre-partition attributes inference passes. These can be used when some input and/or parameters are missing in `args`, e.g. input data (similarly to what is done in the `bind` functions).
- Changes both Python and C++ functions to allow partial inference.

CC @samskalicky


samskalicky:
@mseth10 will you please review as original author of the shape inference in optimize_for?

src/c_api/c_api_symbolic.cc
```diff
@@ -1383,47 +1394,78 @@ int MXOptimizeForBackend(SymbolHandle sym_handle,
...
+          args_top++;
+        }
+      }
 
-    // infer shapes
-    g = exec::InferShape(std::move(g), std::move(arg_shapes), "__shape__");
-    // infer dtypes
-    g = exec::InferType(std::move(g), std::move(arg_dtypes), "__dtype__");
-    if (g.GetAttr<size_t>("dtype_num_unknown_nodes") != 0U) {
-      common::HandleInferTypeError(num_forward_inputs, indexed_graph,
```
mseth10, line 1419: why are we removing this error handler?

src/c_api/c_api_symbolic.cc
```diff
@@ -1383,47 +1394,78 @@ int MXOptimizeForBackend(SymbolHandle sym_handle,
...
-                                   g.GetAttr<nnvm::DTypeVector>("dtype"));
-    }
-    // infer stypes
-    g = exec::InferStorageType(std::move(g), std::move(arg_stypes), "__storage_type__");
-    if (g.GetAttr<size_t>("storage_type_num_unknown_nodes") != 0U) {
-      common::HandleInferStorageTypeError(num_forward_inputs, indexed_graph,
```
mseth10, line 1425: same question as above

src/c_api/c_api_symbolic.cc
```diff
@@ -1383,47 +1394,78 @@ int MXOptimizeForBackend(SymbolHandle sym_handle,
...
+          if (in_aux_ptr[aux_top] != nullptr) {
+            const auto &in_arg = *(in_aux_ptr[aux_top]);
+            arg_shapes[i] = in_arg.shape();
+            arg_dtypes[i] = in_arg.dtype();
+            arg_stypes[i] = in_arg.storage_type();
+          } else {
+            auto it_shape = input_shape_map.find(name);
+            if (it_shape != input_shape_map.end()) {
+              arg_shapes[i] = it_shape->second;
+            }
```
mseth10, line 1436: else clause? what if `name` not found in `input_shape_map`? do we error out?

src/c_api/c_api_symbolic.cc
```diff
@@ -1383,47 +1394,78 @@ int MXOptimizeForBackend(SymbolHandle sym_handle,
...
+          aux_top++;
+        } else {
+          CHECK_LT(args_top, args_len)
+            << "Cannot find arg '" << input_names[i] << "' in provided args to optimize_for";
+          if (in_args_ptr[args_top] != nullptr) {
+            const auto &in_arg = *(in_args_ptr[args_top]);
+            arg_shapes[i] = in_arg.shape();
+            arg_dtypes[i] = in_arg.dtype();
+            arg_stypes[i] = in_arg.storage_type();
+          }
```
mseth10, line 1455: else clause? do we error out?

src/c_api/c_api_symbolic.cc
Kh4L: The goal of this PR is to allow the user to be able to skip the inference pass, either explicitly with `skip_infer` or implicitly, relying on partial attribute inference.

Some backends may require the attr inference, some others not, so the error should be handled by the backend.

src/c_api/c_api_symbolic.cc
Kh4L: There is a logic issue (I mixed args and aux...), updating it

src/c_api/c_api_symbolic.cc
Kh4L: Same as above

src/c_api/c_api_symbolic.cc
Kh4L: This logic should be in the other close (for `args`, not aux). I just updated it

src/c_api/c_api_symbolic.cc
Kh4L: AFAIK, we don't need it here, as input can't be in `aux` but only in `args`

python/mxnet/symbol/symbol.py
```diff
@@ -1460,19 +1461,36 @@ def optimize_for(self, backend, args=None, aux=None, ctx=None, **kwargs):
             Input arguments to the symbol, required to infer shapes/types before partitioning
 
             - If type is a list of `NDArray`, the order is the same as that of `list_arguments()`.
+              Non defined arguments' `NDArray`s should be marked as None to respected the order.
```
mseth10, line 1464: nit: respect

Kh4L:

include/mxnet/c_api.h
```diff
@@ -2247,6 +2247,17 @@ MXNET_DLL int MXOptimizeForBackend(SymbolHandle sym_handle,
...
+                                   const char** input_shape_names,
+                                   const int64_t* input_shape_data,
+                                   const uint32_t* input_shape_idx,
+                                   const uint32_t num_input_dtypes,
+                                   const char** input_dtype_names,
+                                   const int* input_dtypes,
+                                   const uint32_t num_input_stypes,
+                                   const char** input_stype_names,
+                                   const int* input_stypes,
+                                   bool skip_infer,
```
ptrendx, line 2260: Please add those new parameters to the docstring

src/c_api/c_api_symbolic.cc
```diff
@@ -1383,47 +1394,80 @@ int MXOptimizeForBackend(SymbolHandle sym_handle,
...
+          args_top++;
+        }
+      }
 
-    // infer shapes
-    g = exec::InferShape(std::move(g), std::move(arg_shapes), "__shape__");
-    // infer dtypes
-    g = exec::InferType(std::move(g), std::move(arg_dtypes), "__dtype__");
-    if (g.GetAttr<size_t>("dtype_num_unknown_nodes") != 0U) {
-      common::HandleInferTypeError(num_forward_inputs, indexed_graph,
```
ptrendx, line 1419: The coverage tool shows that this function no longer is covered (because this was the only place it was used. Considering we do not need it here anymore, could you delete it?

ptrendx: Same for HandleInferStorageTypeError

ptrendx:
I believe it is safe to ignore the codecov failure (see #18497 which disables codecov for reasoning, we could rebase on top of current master and rerun, but I don't think it makes sense to rerun CI just for that).

PR #18354, opened 2020-05-18
cmake: check compiler supports C++17 sufficiently

leezu:
By default cmake accepts any compiler that supports at least some part of C++17.
In practice this leads to hard to understand error messages at later stages
during compilation, as for example gcc5 supports part of C++17 but does not
support C++17 sufficiently to compile MXNet.

cc @apeforest


leezu:
@mxnet-bot run ci [unix-gpu, miscellaneous]

Issue #18120, opened 2020-04-21
MKLDNN header missing in recent mxnet nightly static builds


eric-haibin-lin:
```
root@ip-172-31-37-108:/bps-mkl# pip3 install --pre mxnet-cu100 -f https://dist.mxnet.io/python/cu100  -U
Collecting mxnet-cu100
  Downloading https://repo.mxnet.io/dist/python/cu100/mxnet_cu100-2.0.0b20200312-py2.py3-none-manylinux1_x86_64.whl (808.1MB)
```
```
ls /usr/local/lib/python3.6/dist-packages/mxnet/include/
dlpack  dmlc  mshadow  mxnet  nnvm
```

Since mkldnn is turned on by default in Feb, the headers should be included, too

@TaoLv

leezu:
FYI, I intend to switch the CD to cmake build in the coming days. It's currently blocked by https://github.com/apache/incubator-mxnet/issues/17557 because I'd like the CD to pass at least once prior to changing the build system.. #17557 may be fixed by https://github.com/apache/incubator-mxnet/pull/18025

TaoLv:
I doubt it's caused by the fix in https://github.com/apache/incubator-mxnet/pull/16857. We have lost the mkldnn header folder since then. For example, in https://repo.mxnet.io/dist/python/mkl/mxnet_mkl-1.6.0b20191205-py2.py3-none-manylinux1_x86_64.whl.

leezu:
@TaoLv agree https://github.com/apache/incubator-mxnet/pull/16857 is wrong. Now we only copy the headers on OSX platform which doesn't make any sense

leezu:
I think @mseth10 will fix it as part of his work on fixing CD

wms2537:
The problem seems to persist on the latest nightly builds.
I have tried the followings and the mkldnn headers are not included. This caused problems when installing horovod.

mxnet_cu101-2.0.0b20200506-py2.py3-none-manylinux2014_x86_64.whl
mxnet_cu101-2.0.0b20200508-py2.py3-none-manylinux2014_x86_64.whl

eric-haibin-lin:
mxnet-cu100==2.0.0b20200509 seems to work: 
```
ubuntu@ip-172-31-37-108:~/src$ pip3 list | grep mxnet
mxnet-cu100         2.0.0b20200509

[...]

PR #18355, opened 2020-05-18
Include all mkldnn headers in CD builds

leezu:
Fixes https://github.com/apache/incubator-mxnet/issues/18120 specifically https://github.com/apache/incubator-mxnet/issues/18120#issuecomment-630147645

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,24 +240,26 @@ if(USE_MKLDNN)
     set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /EHsc /Gy /MT")
   endif()
 
-  set(MKLDNN_BUILD_TESTS OFF CACHE INTERNAL "" FORCE)
-  set(MKLDNN_BUILD_EXAMPLES OFF CACHE INTERNAL "" FORCE)
-  set(MKLDNN_ARCH_OPT_FLAGS "" CACHE INTERNAL "" FORCE)
-  set(MKLDNN_ENABLE_JIT_PROFILING OFF CACHE INTERNAL "" FORCE)
-  set(MKLDNN_LIBRARY_TYPE STATIC CACHE INTERNAL "" FORCE)
-  set(DNNL_ENABLE_CONCURRENT_EXEC ON CACHE INTERNAL "" FORCE)
+  function(load_mkldnn)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -240,24 +240,26 @@ if(USE_MKLDNN)
...
 
-  set(MKLDNN_BUILD_TESTS OFF CACHE INTERNAL "" FORCE)
-  set(MKLDNN_BUILD_EXAMPLES OFF CACHE INTERNAL "" FORCE)
-  set(MKLDNN_ARCH_OPT_FLAGS "" CACHE INTERNAL "" FORCE)
-  set(MKLDNN_ENABLE_JIT_PROFILING OFF CACHE INTERNAL "" FORCE)
-  set(MKLDNN_LIBRARY_TYPE STATIC CACHE INTERNAL "" FORCE)
-  set(DNNL_ENABLE_CONCURRENT_EXEC ON CACHE INTERNAL "" FORCE)
-
-  if(NOT USE_OPENMP)
+  function(load_mkldnn)
```
TaoLv, line 243: Why need put them into a function?

CMakeLists.txt
```diff
@@ -240,24 +240,26 @@ if(USE_MKLDNN)
...
 
+    set(CMAKE_INSTALL_INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}/mkldnn")
+    add_subdirectory(3rdparty/mkldnn)
+  endfunction()
+  load_mkldnn()
   include_directories(3rdparty/mkldnn/include)
   include_directories(${PROJECT_BINARY_DIR}/3rdparty/mkldnn/include)
   add_definitions(-DMXNET_USE_MKLDNN=1)
   list(APPEND mxnet_LINKER_LIBS dnnl)
-  set(INSTALL_MKLDNN ON)
```
TaoLv, line 260: Why it's removed? @yzhliu FYI.

leezu: This variable isn't used anywhere but in this line

CMakeLists.txt
leezu: Functions have their own scope in cmake. We don't want to change any of the settings outside of this function (such as the install target name for the include files)

leezu: So it's really about setting the `CMAKE_INSTALL_INCLUDEDIR` differently for mkldnn which is introduced in this PR

Issue #18293, opened 2020-05-13
CD: Missing openblas


leezu:
After enabling cmake staticbuild for CD in https://github.com/apache/incubator-mxnet/commit/c329a524c520ec2b41aa4bff5ee113ff7221b790, `E   OSError: libopenblas.so.0: cannot open shared object file: No such file or directory` https://github.com/leezu/gluon-nlp/runs/669008929?check_suite_focus=true

Need to change the staticbuild script to staticlink openblas or include the library.

PR #18356, opened 2020-05-18
Include lib/libopenblas.so.0 on CD

leezu:
The cmake staticbuild script doesn't link openblas statically, but rather
provides the so file to be included in the distribution. This is intentional, as
it simplifies the build script, but was missed when switching the CD to rely on
cmake.

Fixes https://github.com/apache/incubator-mxnet/issues/18293


leezu:

PR #18328, opened 2020-05-15
Fix memory leaks in Gluon

leezu:
## Description ##
Previously the _BlockScope keeps references to the parameter ndarrays, preventing memory from being freed if a Block is not used anymore. Among other problems, this causes memory usage to increase constantly in unittests (due to testing different blocks and disposing them at the end of the test) until the garbage collector kicks in (which can be too late and the system can run OOM as the parameter arrays can be large).

PR #18358, opened 2020-05-18
[1.7.x] Fix memory leaks in Gluon

leezu:
Backport https://github.com/apache/incubator-mxnet/pull/18328

### DIFF
--- a/python/mxnet/gluon/block.py
+++ b/python/mxnet/gluon/block.py
@@ -23,8 +23,10 @@ __all__ = ['Block', 'HybridBlock', 'SymbolBlock']
 import threading
 import copy
 import warnings
-import re
+import weakref
 from collections import OrderedDict, defaultdict
+
+import re
 import numpy as np
 
 from ..base import mx_real_t, MXNetError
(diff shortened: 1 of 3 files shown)


leezu:
@mxnet-bot run ci [all]

ciyongch:
@TaoLv @pengzhao-intel please take a review and help to merge it :)

TaoLv:
Please also make sure it's fixed on branch v1.x.

ciyongch:
@TaoLv here's the PR for v1.x branch: https://github.com/apache/incubator-mxnet/pull/18359

leezu:
@zhreshold should we revert this PR on the 1.7 and 1.x branch as weakref doesn't work with  deepcopy?

https://github.com/dmlc/gluon-cv/blob/428ee05d7ae4f2955ef00380a1b324b05e6bc80f/gluoncv/data/transforms/presets/yolo.py#L144

(Though I'm surprised that `copy.deepcopy` worked in the first place)

zhreshold:
@leezu Fix to mem leak is nice to have and I don't see any problem with this pariticular PR, however, `deepcopy` of a full network isn't a rare usecase which means a fix to deepcopy is needed anyway. 

In fact, the behavior of `__deepcopy__` of a `Block` has to be overrided with the `_BlockScope` being renewed in the new `Block`, and the suspicious copy behavior of parameters has to be verified and secured.

PR #18362, opened 2020-05-19
fix windows dll search path in python 3.8

yajiedesign:
## Description ##
fix windows dll search path in python 3.8

## Reference  ##
https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew

### DIFF
--- a/python/mxnet/libinfo.py
+++ b/python/mxnet/libinfo.py
@@ -20,7 +20,7 @@
 import os
 import platform
 import logging
-
+import sys
 
 def find_lib_path(prefix='libmxnet'):
     """Find MXNet dynamic library files.
@@ -73,6 +73,11 @@ def find_lib_path(prefix='libmxnet'):
                            'List of candidates:\n' + str('\n'.join(dll_path)))
     if os.name == 'nt':
         os.environ['PATH'] = os.environ['PATH'] + ';' + os.path.dirname(lib_path[0])
+        if sys.version_info >= (3,8):
+            if 'CUDA_PATH' not in os.environ:
+                raise RuntimeError('Cannot find the env CUDA_PATH.Please set CUDA_PATH env with cuda path')
+            os.add_dll_directory(os.path.dirname(lib_path[0]))
+            os.add_dll_directory(os.path.join(os.environ['CUDA_PATH'],'bin') )
     return lib_path
 
 def find_include_path():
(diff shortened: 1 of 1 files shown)


yajiedesign:
@mxnet-bot run ci [ci/jenkins/mxnet-validation/miscellaneous]

yajiedesign:

leezu:

python/mxnet/libinfo.py
```diff
@@ -73,6 +73,11 @@ def find_lib_path(prefix='libmxnet'):
                            'List of candidates:\n' + str('\n'.join(dll_path)))
     if os.name == 'nt':
         os.environ['PATH'] = os.environ['PATH'] + ';' + os.path.dirname(lib_path[0])
+        if sys.version_info >= (3, 8):
+            if 'CUDA_PATH' not in os.environ:
+                raise RuntimeError('Cannot find the env CUDA_PATH.Please set CUDA_PATH env with cuda path')
```
leezu, line 78: Thus mxnet won't work on Windows as long as users don't set up this environment variable? Or is it provided by default by cuda?

python/mxnet/libinfo.py
```diff
@@ -73,6 +73,11 @@ def find_lib_path(prefix='libmxnet'):
                            'List of candidates:\n' + str('\n'.join(dll_path)))
     if os.name == 'nt':
         os.environ['PATH'] = os.environ['PATH'] + ';' + os.path.dirname(lib_path[0])
+        if sys.version_info >= (3, 8):
```
leezu, line 76: It's more pythonic to use try-except syntax and just attempt to call add_dll_directory
It will ensure compatibility with interpreters different to cpython

python/mxnet/libinfo.py
yajiedesign: cuda default provided.

python/mxnet/libinfo.py
yajiedesign: Check the version here because only 3.8 or above needs to call this

leezu: It's nevertheless bad practice as there are other python interpreters with different versioning schemes

422926799:
The reason for this problem is: Python 3.8 changed the dynamic link library (DLL) loading rules under Windows.

The new rules improve security. By default, DLL dependencies can only be loaded from Trusted Locations, which avoids security risks such as DLL hijacking to a certain extent.

Reference link：https://github.com/ynyyn/Miniblink-Python-SimpleDemo/issues/4
```python
import os
mxnetpath=sys.path[6]+"\\mxnet\\libmxnet.dll"
os.add_dll_directory(mxnetpath)
import mxnet
```

PR #18363, opened 2020-05-19
Fix reference cycle in gluon.Trainer

leezu:
Fix reference cycle preventing reference-counting based deallocation of NDArrays due to gluon.Trainer.

cc @eric-haibin-lin

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -28,7 +28,6 @@ import gc
 import os
 import random
 
-import mxnet as mx
 import pytest
 
 
--- a/conftest.py
+++ b/conftest.py
@@ -24,9 +24,11 @@ instantiated before lower-scoped fixtures (such as ``function``).
 """
 
 import logging
+import gc
 import os
 import random
 
+import mxnet as mx
 import pytest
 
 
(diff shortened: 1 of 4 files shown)


conftest.py
```diff
@@ -229,3 +230,50 @@ def doctest(doctest_namespace):
...
+
+@pytest.fixture(scope='session')
+def mxnet_module():
+    import mxnet
+    return mxnet
+
+
+@pytest.fixture()
+# @pytest.fixture(autouse=True)  # Fix all the bugs and mark this autouse=True
+def check_leak_ndarray(mxnet_module):
```
eric-haibin-lin, line 243: why do we intend to use autouse? the mxnet_module function does not seem to do much. How does it help here?

leezu: conftest is run for any pytest session started in the mxnet folder. But some of our tests (such as for website build) do not have mxnet present. So we can't just 'import mxnet' in the top of the file.

Instead, the mxnet_module is a pytest fixture that can be used to get mxnet module. It will only be run if the tests requests it (as here).

Autouse must be used to apply the check_leak_ndarray before and after every other mxnet test. Currently there are too many bugs, so we can't enable it. Later, when the bugs are fixed it can be enabled.

leezu: You can refer to https://docs.pytest.org/en/latest/fixture.html#autouse-fixtures-xunit-setup-on-steroids

PR #18365, opened 2020-05-19
Consider MXNET_BUILD_SHARED_LIBS flag

cbalint13:
## Description ##
1. Consider ```MXNET_BUILD_SHARED_LIBS``` build flag

## Comments ##
Building pure shared version is distro's releng favorite way.

Thank You !


leezu:
Thanks @cbalint13 . Actually we may improve the defaults for upcoming MXNet 2. I suggest to make the "build shared" the default and optionally allow to build only the static library. (ie following the CMAKE_BUILD_SHARED) approach.

What do you think?

cbalint13:
@leezu ,

Let's try, see how will pass CI.

cbalint13:
@leezu ,

UPDATE: Followed your proposal, updated the code.

* The way CI fails is beyond of this PR cause (noticed others fail too, i.e cuda.h not found). 
* Is it still acceptable feature for trunk ?

Let me know if need more work on it.

leezu:
Thanks @cbalint13. I wonder if we should only build the static library if `MXNET_BUILD_STATIC=ON`, instead of building both static and shared libraries. If yes, would `MXNET_BUILD_STATIC` even be needed or should we simply use the value of `CMAKE_BUILD_SHARED` (but default to `CMAKE_BUILD_SHARED=ON`). What do you think?

cbalint13:
@leezu ,

Uhh, Ohh.

* Can't make pure shared build without hurting CI very badly.
* cmake's internal ```CMAKE_BUILD_SHARED``` would be perfect (but it is ```ON``` by default).

* I ended with up using ```MXNET_BUILD_STATIC=ON``` due:
   0. CMAKE_BUILD_SHARED internal by default is ```ON``` (too big change for CI state).
   1. In current static mode a large static object and a small shared surrogate is build, that's it.
   2. CI isn't hurt for now, by switching ```MXNET_BUILD_STATIC=OFF``` one can fix CI side.
   3. distro's releng can handle ```MXNET_BUILD_STATIC=OFF``` for now if they want on linux.
   4. When CI is solved we can replace MXNET_BUILD_STATIC with CMAKE_BUILD_SHARED internal.

leezu:
> In current static mode a large static object and a small shared surrogate is build, that's it now.

Actually not. There is a hack further down in the cmake file where the complete static object is included as is into the shared object. So there is not much difference between the two besides the format.

> Can't make pure shared build without hurting CI very badly.

What do you mean?

cbalint13:
@mxnet-bot run ci [all]

cbalint13:
> > In current static mode a large static object and a small shared surrogate is build, that's it now.
> 
> Actually not. There is a hack further down in the cmake file where the complete static object is included as is into the shared object. So there is not much difference between the two besides the format.
> 
> > Can't make pure shared build without hurting CI very badly.
> 
> What do you mean?



> > In current static mode a large static object and a small shared surrogate is build, that's it now.
> 
> Actually not. There is a hack further down in the cmake file where the complete static object is included as is into the shared object. So there is not much difference between the two besides the format.

I am aware.

> > Can't make pure shared build without hurting CI very badly.
> 
> What do you mean?

Will not pass CI.

leezu:
> Will not pass CI.

Why do you think so. The CI does not rely on the `libmxnet.a`.

cbalint13:
@leezu 

> > Will not pass CI.
> 
> Why do you think so. The CI does not rely on the `libmxnet.a`.

* It seems, had the impression that all test units links ```libmxnet.a```

Can advice me on MacOS CI failure ? Dont get the meaning of error (can't see what symbol):
```
E   AttributeError: dlsym(0x7fdb866c3810, MXGetLastError): symbol not found
```

Thank You !

CMakeLists.txt
```diff
@@ -663,29 +663,46 @@ if(UNIX)
...
-  add_library(mxnet_static STATIC ${SOURCE})
-  target_link_libraries(mxnet_static PUBLIC mshadow)
-  target_compile_definitions(mxnet_static PUBLIC DMLC_LOG_FATAL_THROW=$<BOOL:${LOG_FATAL_THROW}>)
-  add_library(mxnet SHARED ${DUMMY_SOURCE})
-  target_link_libraries(mxnet PRIVATE ${BEGIN_WHOLE_ARCHIVE} $<TARGET_FILE:mxnet_static> ${END_WHOLE_ARCHIVE})
-  target_link_libraries(mxnet PRIVATE mxnet_static)
-  target_link_libraries(mxnet_static PUBLIC ${CMAKE_DL_LIBS})
-  set_target_properties(mxnet_static PROPERTIES OUTPUT_NAME mxnet)
+  if(CMAKE_BUILD_SHARED)
+    add_library(mxnet SHARED ${SOURCE})
```
leezu, line 667: You can remove `SHARED` here and delete the `else` clause below. cmake will use the correct library type based on `CMAKE_BUILD_SHARED`

tests/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ if(NOT GTEST_ROOT)
   endif()
 endif()
 
-if(NOT MSVC)
+if(NOT MSVC AND NOT CMAKE_BUILD_SHARED)
   set(UNITTEST_STATIC_LINK ON)
 endif()
```
leezu, line 21: These 3 lines can be deleted

example/image-classification/predict-cpp/CMakeLists.txt
```diff
@@ -5,11 +5,11 @@ OpenCV version >= 3 should be enabled and found to build image classification ex
...
 
-if(NOT MSVC)
+if(NOT MSVC AND NOT CMAKE_BUILD_SHARED)
   set(IMG_CLASSIFICATION_EXAMPLE_STATIC_LINK ON CACHE BOOL "\
 Link mxnet library statically in the c++ image classification example")
 else()
-  # disable static linking on Windows
+  # disable static linking on Windows & CMAKE_BUILD_SHARED
   set(IMG_CLASSIFICATION_EXAMPLE_STATIC_LINK OFF)
 endif()
```
leezu, line 14: Let's delete this too and let cmake handle things automatically based on `CMAKE_BUILD_SHARED`

CMakeLists.txt
```diff
@@ -663,29 +663,46 @@ if(UNIX)
...
+    list(APPEND MXNET_INSTALL_TARGETS mxnet)
+  else()
+    add_library(mxnet_static STATIC ${SOURCE})
```
leezu, line 670: In particular we should get rid of the separate `mxnet_static` target

leezu:
> It seems, had the impression that all test units links libmxnet.a

That's part of the problem. It comes with huge link, disk space and CI network traffic overheads due to the number of unittests.

> `E   AttributeError: dlsym(0x7fdb866c3810, MXGetLastError): symbol not found`

I'm not yet sure about this error. Could you address above comments first (or state that you disagree) and then let's look more closely into this issue.

szha:
@leezu static object should be useful in building standalone C++ inference programs. If the size of test is a problem, we can strip the binary after compilation.

leezu:
@szha sure. In that case just set CMAKE_BUILD_SHARED=0

cbalint13:
> @leezu static object should be useful in building standalone C++ inference programs. If the size of test is a problem, we can strip the binary after compilation.

- we did't get rid, still can build static objects if one desire
- in my opinion, too big for normal releng purposes (even stripped).

cbalint13:
@leezu ,

* Proposed leap is big, someone need to adjust CI (out of my scope).
* I can re-propose PR it in earlier form (agree not elegant) and see that single one MacOS-CI issue.

leezu:
Adopting the standard `BUILD_SHARED_LIBS` may take additional work to make sure all dependencies + the CI logic correctly handle the result. It would however result into a more "distro-friendly" build you quote, as the dependencies will also be built dynamically.

But to keep things simple in this PR, it's fine to use an MXNet specific setting for now. I modified your PR slightly to reintroduce the MXNet specific option.

cbalint13:
> Adopting the standard `BUILD_SHARED_LIBS` may take additional work to make sure all dependencies + the CI logic correctly handle the result. It would however result into a more "distro-friendly" build you quote, as the dependencies will also be built dynamically.

Wow someone circumvent CI :1st_place_medal: 

> 
> But to keep things simple in this PR, it's fine to use an MXNet specific setting for now. I modified your PR slightly to reintroduce the MXNet specific option.

Thank you very much @leezu !

leezu:
Thanks @cbalint13

PR #18376, opened 2020-05-20
Fix reference leak in Block._children and enable reference leak tests universally

leezu:
## Description ##
Fix reference leak in Block._children and enable reference leak tests universally.
Extends https://github.com/apache/incubator-mxnet/pull/18328 https://github.com/apache/incubator-mxnet/pull/18363


pytest.ini
```diff
@@ -23,6 +23,7 @@ markers =
     gpu: mark a test that requires GPU.
     integration: mark an integration test
     onnx_coverage: ONNX coverage test
+    garbage_expected: this teast leaks ndarray references. The tested functionality is broken.
```
szha, line 26:
```suggestion
    garbage_expected: this test leaks ndarray references. The tested functionality is broken.
```

Issue #18373, opened 2020-05-20
[Numpy] The symbolic export of BatchNorm is wrong


sxjscience:
```python
import mxnet as mx
import json
import pprint
mx.npx.set_np()
net = mx.gluon.nn.BatchNorm(epsilon=2E-5, axis=2)
net.hybridize()
net.initialize()
a = net(mx.np.ones((10, 3, 5, 5)))
net.export('bnorm', 0)
with open('bnorm-symbol.json') as f:
   dat = json.load(f)
   pprint.pprint(dat)
```

Output:
```
           {'attrs': {'__profiler_scope__': 'batchnorm0:',
                      'axis': '1',
                      'eps': '1e-05',
                      'fix_gamma': 'False',
                      'momentum': '0.9',
                      'use_global_stats': 'False'},
            'inputs': [[0, 0, 0], [1, 0, 0], [2, 0, 0], [3, 0, 1], [4, 0, 1]],
            'name': 'batchnorm0_fwd',
            'op': 'BatchNorm'}]}
```
We can find that `eps` and `axis` are not stored.

sxjscience:
I find that issue does not only happen in numpy but also exists in ndarray:
```python
import mxnet as mx
import json
import pprint
#mx.npx.set_np()
net = mx.gluon.nn.BatchNorm(epsilon=2E-5, axis=2)
net.hybridize()
net.initialize()
a = net(mx.nd.ones((10, 3, 5, 5)))
net.export('bnorm', 0)
with open('bnorm-symbol.json') as f:
   dat = json.load(f)
   pprint.pprint(dat)
```

Output:
```
           {'attrs': {'__profiler_scope__': 'batchnorm0:',
                      'axis': '1',
                      'eps': '1e-05',
                      'fix_gamma': 'False',
                      'momentum': '0.9',
                      'use_global_stats': 'False'},
            'inputs': [[0, 0, 0], [1, 0, 0], [2, 0, 0], [3, 0, 1], [4, 0, 1]],
            'name': 'batchnorm0_fwd',
            'op': 'BatchNorm'}]}
```

wkcn:
Hi @sxjscience , is it available to delete the pre-built pip packages impacted by this issue? 

BatchNorm is universally used, and this bug will not raise any exception. Users may install the previous version of MXNet with this bug, and find that the accuracy drops.

sxjscience:
[...]

PR #17679, opened 2020-02-25
[MKL-DNN] BatchNormRelu Fusion

xinyu-intel:
## Description ##
BatchNormRelu Fusion while training.

Main changes for this feature are:
1)	Implement `mx.nd.contrib.BatchNormWithReLU` operator and place it in `src/operator/contrib/` directory;
2)	Add a parameter to gluon interface `nn.BatchNorm(fuse_relu=True, **kwargs)` to call `mx.nd.contrib.BatchNormWithReLU` instead of `mx.nd.BatchNorm`

For users who wants to use this feature, he needs to:
1)	Symbolic: call `mx.sym.contrib.BatchNormWithReLU` directly to replace `mx.sym.BatchNorm` and `mx.sym.Activation`;
2)	Gluon: Add layer `nn.BatchNorm(fuse_relu=True)` instead of `[nn.BatchNorm(), nn.Activation()]`

@pengzhao-intel @TaoLv 

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at https://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #18160, opened 2020-04-24
[1.x] Add BatchNormWithReLU fp32/bf16

xinyu-intel:
## Description ##
@pengzhao-intel @TaoLv 

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [ ] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage:
- Unit tests are added for small changes to verify correctness (e.g. adding a new operator)
- Nightly tests are added for complicated/long-running ones (e.g. changing distributed kvstore)
- Build tests will be added for build configuration changes (e.g. adding a new build option with NCCL)
- [ ] Code is well-documented: 
- For user-facing API changes, API doc string has been updated. 
- For new C++ functions in header files, their functionalities and arguments are documented. 
- For new examples, README.md is added to explain the what the example does, the source of the dataset, expected performance on test set and reference to the original paper if applicable
- Check the API doc at https://mxnet-ci-doc.s3-accelerate.dualstack.amazonaws.com/PR-$PR_ID/$BUILD_ID/index.html
- [ ] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #18377, opened 2020-05-21
Fix BatchNorm

sxjscience:
## Description ##
Fix https://github.com/apache/incubator-mxnet/issues/18373.

The PR https://github.com/apache/incubator-mxnet/pull/17679 introduced the bug @xinyu-intel 

@nttstar Should fix your problem.


sxjscience:
@xinyu-intel I think the fix should also be in 1.7.

xinyu-intel:
@sxjscience 1.7 doesn't contain #17679 , you can cherry-pick this patch to v1.x when #18160 merged.

Issue #16757, opened 2019-11-08
Exceptions thrown in InferType before doing reverse inference


anirudh2290:
## Description
Some operators are missing logic for reverse inference: inferring the type of inputs from outputs. Instead, exception is thrown if input dtype is -1.

Also, InferType logic of operators should throw an exception only if both input and output are defined and incompatible with each other. For example. if inputs to an op are -1 and output is -1, the infertype logic should return false, instead of throwing an exception. It should throw an exception only if the types are incompatible: for example both input types and output types are inferred but input types are not compatible with output types.

The above point is important because the InferType logic does both forward and backward inference. So, sometimes output and input can both be -1 in forward inference but in the backward inference both these values can be inferred.

- [ ] fft
- [ ] ifft
- [ ] deformable_convolution
- [ ] multi_sum_sq
- [ ] SequenceReverse
- [ ] Convolution_v1
- [ ] RNN
- [ ] UpSampling
- [ ] BatchNorm
- [ ] Deconvolution
- [ ] LRN
- [ ] Convolution
- [ ] Embedding
- [ ] split_v2
- [ ] SVMOutput
- [ ] SequenceMask
- [ ] SoftmaxOutput
- [ ] BatchNorm_v1

The existing issues with ops was brought up @ptrendx in https://github.com/apache/incubator-mxnet/pull/16748

samskalicky:
@anirudh2290 Can we make this a feature request and change title to "Improve InferTYpe ..." instead of a bug?

anirudh2290:
It's a bug

anirudh2290:
un-assigning myself since  i am not working on this currently. if someone is interested, please go ahead.

PR #18378, opened 2020-05-21
Add backward Type inference to main NN operators

Kh4L:
## Description ##
As described in #16757 , a list of operators have their FInferType functions were throwing an exception if the input type is not defined (dtype=-1).

This was preventing us from running a partial or backward type inference where some not (yet) defined input dtypes would trigger an exception.

This PR fixes it by making the FInferType functions do backward inference when input is `none` but output is known and return false when both input and output dtype is `none`.


src/operator/contrib/batch_norm_relu.cc
```diff
@@ -84,14 +84,30 @@ static bool BatchNormWithReLUType(const nnvm::NodeAttrs& attrs,
...
-  CHECK_NE(dtype, -1) << "First input must have specified type";
+  const size_t n_out = 4;
   // For float16 input type beta, gamma, mean, and average are stored in float32.
   // For other input types, these parameters have the same type as input
   // NOTE: This requirement is from cuDNN (v. 4 and 5)
   int dtype_param;
-  MSHADOW_REAL_TYPE_SWITCH_EX(dtype, DTypeX, AccRealX, {
+  int dtype = (*in_type)[0];
+  if (type_is_none(dtype)) {
+     if (out_type->size() == 0 || type_is_none((*out_type)[0])) {
```
samskalicky, line 94: what does this mean if this logic is true? can you add a comment in the code?

src/operator/contrib/batch_norm_relu.cc
```diff
@@ -84,14 +84,30 @@ static bool BatchNormWithReLUType(const nnvm::NodeAttrs& attrs,
...
-  const int dtype = (*in_type)[0];
-  CHECK_NE(dtype, -1) << "First input must have specified type";
+  const size_t n_out = 4;
   // For float16 input type beta, gamma, mean, and average are stored in float32.
   // For other input types, these parameters have the same type as input
   // NOTE: This requirement is from cuDNN (v. 4 and 5)
   int dtype_param;
-  MSHADOW_REAL_TYPE_SWITCH_EX(dtype, DTypeX, AccRealX, {
+  int dtype = (*in_type)[0];
+  if (type_is_none(dtype)) {
```
samskalicky, line 93: what does it mean is `type_is_none(dtype)` is true? that the input type is unknown '-1'?

src/operator/contrib/batch_norm_relu.cc
```diff
@@ -84,14 +84,30 @@ static bool BatchNormWithReLUType(const nnvm::NodeAttrs& attrs,
...
+     if (out_type->size() == 0 || type_is_none((*out_type)[0])) {
+       return false;
+     } else {
+       dtype = (*out_type)[0];
```
samskalicky, line 97: what case is this? if input type is known, then just assign output type to input type?

src/operator/contrib/batch_norm_relu.cc
```diff
@@ -84,14 +84,30 @@ static bool BatchNormWithReLUType(const nnvm::NodeAttrs& attrs,
...
+       (*in_type)[0] = dtype;
+       MSHADOW_REAL_TYPE_SWITCH_EX(dtype, DTypeX, AccRealX, {
+         dtype_param = mshadow::DataType<AccRealX>::kFlag; });
+     }
+  } else {
+    MSHADOW_REAL_TYPE_SWITCH_EX(dtype, DTypeX, AccRealX, {
```
samskalicky, line 103: what are we doing in this last case here? can you add a comment?

src/operator/contrib/batch_norm_relu.cc
Kh4L: That's correct

Kh4L:
@samskalicky I added some comments

Issue #18374, opened 2020-05-20
Windows nightly builds broken


leezu:
Last build is from 20200504.

cc: @yajiedesign are you aware? This currently breaks the GluonNLP Windows CI
https://github.com/sxjscience/gluon-nlp/runs/687011402?check_suite_focus=true

yajiedesign:
I already know. Now there are some compilation errors.

PR #18379, opened 2020-05-21
change _Pragma to #pragma

yajiedesign:
## Description ##
change _Pragma to #pragma
_Pragma not work with msvc

### Essentials ###

### DIFF
--- a/src/io/batchify.cc
+++ b/src/io/batchify.cc
@@ -43,9 +43,9 @@ namespace io {
 
 #ifdef _MSC_VER
   #if _MSC_VER < 1925
-   #define omp_parallel(t) __pragma(omp parallel for num_threads(t))
+    #define omp_parallel(t) __pragma(omp parallel for num_threads(t))
   #else
-   #define omp_parallel(t) _Pragma(tostr(omp parallel for num_threads( ## t ## )))
+    #define omp_parallel(t) _Pragma(tostr(omp parallel for num_threads( ## t ## )))
   #endif
 #else
   #define omp_parallel(t) _Pragma(tostr(omp parallel for num_threads( ## t ## )))
--- a/src/io/batchify.cc
+++ b/src/io/batchify.cc
@@ -48,7 +48,7 @@ namespace io {
     #define omp_parallel(t) _Pragma(tostr(omp parallel for num_threads( ## t ## )))
   #endif
 #else
-  #define omp_parallel(t) _Pragma(tostr(omp parallel for num_threads( ## t ## )))
+  #define omp_parallel(t) _Pragma(tostr(omp parallel for num_threads(t)))
 #endif
 
 struct GroupBatchifyParam : public dmlc::Parameter<GroupBatchifyParam> {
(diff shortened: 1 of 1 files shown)


yajiedesign:
https://github.com/apache/incubator-mxnet/issues/18374

zhreshold:
`#pragma` won't work in a macro, what's the version of msvc? At least 2019 supports `_Pragma`

src/io/batchify.cc
```diff
@@ -150,7 +150,7 @@ class StackBatchify : public BatchifyFunction {
         }
         int sbs = static_cast<int>(bs);
         MSHADOW_TYPE_SWITCH_WITH_BOOL(dtype, DType, {
-          _Pragma("omp parallel for num_threads(bs)")
+          #pragma omp parallel for num_threads(bs)
```
zhreshold, line 153: use `__pragma` for msvc if it does not support `_Pragma`

wkcn: We can unify the macro like
```c++
#ifndef _MSC_VER
#define __pragma(id) _Pragma(#id)
#endif
```

yajiedesign:
I'll try upgrading vs

yajiedesign:
Cannot be resolved by upgrade vs.

yajiedesign:
@mxnet-bot run ci [windows-cpu]

src/io/batchify.cc
```diff
@@ -37,6 +37,17 @@
 
 namespace mxnet {
 namespace io {
+
+#ifdef _MSC_VER
+  #if _MSC_VER < 1925
+    #define omp_parallel __pragma(omp parallel for num_threads(bs))
```
zhreshold, line 43: Can you make `bs` argument of the macro?

yajiedesign:
@leezu Because CUDA. The old version of CUDA does not support the new version of msvc.
CI is only one CUDA version. but CD requires multiple versions of CUDA. cannot upgrade MSVC.

leezu:
Thanks for clarifying. In that case, should we either drop the old Cuda versions for MXNet 2 or switch the CI to an older MSVC version? What do you think?

leezu:
(Soon Cuda 11 will be released and we'll drop Cuda 9 support in any case)

yajiedesign:
Keep the MSVC of CI compatible with the minimum supported CUDA version.
Then upgrade MSVC when drop CUDA version support.

leezu:
MXNet 2 will be released after Cuda 11 release, so there is no Cuda 9 support in MXNet 2 currently

leezu:
Yes, that's an option. I'm not convinced though that only fixing this current incompatibility without introducing any mechanism of preventing future problems is helpful. 
So if you think we need to support MSVC 2017, then let's also turn it on in the CI. If not, then let's just use MSVC 2019 on the CD. But supporting MSVC 2017 without having any CI checks would just cause unnecessary problems down the road (with the CD breaking often)

yajiedesign:
@mxnet-bot run ci [macosx-x86_64]

yajiedesign:
@zhreshold

leezu:
@yajiedesign please clarify how you plan to address the CI CD discrepancy?

Issue #18282, opened 2020-05-11
test_composite_embedding_with_one_embedding


leezu:
## Description
`test_composite_embedding_with_one_embedding` fails with `TypeError: an integer is required (got type NoneType)`

## Occurrences
http://jenkins.mxnet-ci.amazon-ml.com/blue/rest/organizations/jenkins/pipelines/mxnet-validation/pipelines/windows-gpu/branches/PR-18186/runs/8/nodes/109/steps/155/log/?start=0

leezu:
http://jenkins.mxnet-ci.amazon-ml.com/blue/rest/organizations/jenkins/pipelines/mxnet-validation/pipelines/windows-gpu/branches/PR-18365/runs/10/nodes/109/steps/155/log/?start=0

PR #18383, opened 2020-05-22
Disable test_composite_embedding_with_one_embedding

leezu:
https://github.com/apache/incubator-mxnet/issues/18282


szha:
contrib.text is going to be deprecated in 2.0

Issue #18353, opened 2020-05-18
[Numpy] Division fails for differnt floating dtypes 


ZheyuYe:
## Description
The current `mx.np.ndarray` produces errors that cannot be tolerated in the face of simple division when the numerator and denominator data types have two differnt floating dtypes `np.float32` and `np.float64` such as 
```
import mxnet as mx
import numpy as np
mx.npx.set_np()
numerator = mx.np.array(5.0).astype(np.float64)
denominator = mx.np.array(2.0)
numerator/denominator
>>>array(0., dtype=float64)
```
whereas the correct answer is obviously 2.5. 

```
import numpy as np
numerator = np.array(5.0).astype(np.float64)
denominator = np.array(2.0)
numerator/denominator
>>>2.5
```
The same calculation steps are reliable under numpy, which forces us to face this problem directly.

## Comments
@sxjscience

sxjscience:
I think we should throw the error. Currently, `float64` + `float32` cannot be mixed for division and we haven't raised the error.

xidulu:
Actually, a warning would be thrown out

```python
>>> np.ones((2,2),dtype='float64') / np.zeros((2,2))
[04:07:01] ../src/operator/numpy/./np_true_divide-inl.h:124: not implemented yet...
array([[0.00000000e+000, 6.95161511e-310],
       [0.00000000e+000, 0.00000000e+000]], dtype=float64)
```
The reason behind the division operation's zero output is that the computation kernel is not launched.

ZheyuYe:
Actually, for this issue, we might expect a proper error that aborts the compuation.

PR #18393, opened 2020-05-23
Fix true_divide logging

sxjscience:
## Description ##
Fix https://github.com/apache/incubator-mxnet/issues/18353


sxjscience:
@ZheyuYe

src/operator/numpy/np_true_divide-inl.h
```diff
@@ -306,7 +306,7 @@ void TrueDivideBroadcastCompute(const nnvm::NodeAttrs& attrs,
     } else {
       if (common::is_float(lhs.type_flag_) && common::is_float(rhs.type_flag_)) {
         // lhs and rhs have different float types, the output is the more precise one
-        LOG(ERROR) << "not implemented yet...";
+        LOG(FATAL) << "not implemented yet...";
```
szha, line 309: We should use
```
LOG(FATAL) << "NotImplementedError: xxx setting for yyy op is not implemented yet.";
```

Issue #18297, opened 2020-05-13
different dtype when calculating division compared with numpy


acphile:
For example:
in mxnet.numpy
&nbsp;&nbsp; for two ndarray a,b with dtype='int64', the dtype of result c = a  / b is 'float32'
in numpy
&nbsp;&nbsp; the dtype is 'float64'
Generally, when calculating division, mxnet.numpy often uses float32 while numpy uses float64.
Using float32 will often lead to the loss of floating point precision.
```
b=mx.np.array(1, dtype='int32')
e=b/3
e.item()
>>> 0.3333333432674408
```

sxjscience:
This is to cope with the DL scenario. The default dtype in DeepNumpy is float32. See also https://github.com/apache/incubator-mxnet/issues/18060 and the current PR: https://github.com/apache/incubator-mxnet/pull/18251

leezu:
Any diverging behaviour with numpy should either be fixed or clearly documented in a central reference location. Let's track either solution here

sxjscience:
@leezu I suggest removing the `Bug` label because this is not a bug and won't be fixed. We could add the Document label instead.

yzhliu:
@JiangZhaoh will be helping to document.

yzhliu:
closed by #18399

PR #18399, opened 2020-05-24
Add docs about default dtype

JiangZhaoh:
## Description ##
relate to issue #18297


ChaiBapchya:
@josephevans why did rest of the pipelines get triggered despite sanity failure?
Curious.. the branch does have staggered pipeline logic in it.

Issue #17240, opened 2020-01-07
[MXNet 2.0] [RFC] Distribution Module


szhengac:
## Description
In MxNet 2.0, we would like to provide a distribution module, analogous to Pytorch distribution. The main difference from theirs is that we use numpy op and it allows hybridization. The current project code can be seen from https://github.com/xidulu/incubator-mxnet/tree/distribution_dev/python/mxnet/gluon/probability. 

The basic skeleton divides into following parts:

1. Stochastic `HybridBlock` and `HybridSequential`: they build upon gluon `HybridBlock` and `HybridSequential` and allows adding extra loss to each layer.
2. Distribution class: it implements a variety of functionalities including `prob`, `log_prob`, `sample`, `broadcast_to`, `mean`, `variance`, etc.
3. KL divergence: `kl_divergence(p, q)` function searches over registered KL divergence functions and performs computation.
4. Transform: transform one distribution to another invertible distribution.
5. Independent: reinterprets some of the batch dims of a distribution as event dims.

Two features that is currently either not supported or kind of broken in MxNet will be very useful to this projects: symbolic shape and control flow.

At the moment, we will implement most of distribution in frontend. We will move the computation to backend when new numpy probability ops such as `chisquare`, `dirichlet` and `multivariate_normal` are introduced into MxNet.

## References
- https://pytorch.org/docs/stable/distributions.html
- https://docs.scipy.org/doc/numpy-1.14.1/reference/routines.random.html

@xidulu @szha @leezu @haojin2

xidulu:
> Two features that is currently either not supported or kind of broken in MxNet will be very useful to this projects: symbolic shape and control flow.

More specifically, control flow helps perform parameter validity check: https://github.com/apache/incubator-mxnet/issues/16868

szha:
This module has been merged and will be included as part of mxnet 2.0. Thank you for the contribution @xidulu @szhengac

Issue #17676, opened 2020-02-24
[RFC] MXNet 2.0 API Deprecation


szha:
As the MXNet community is working on the next major version of MXNet as described in #16167, this RFC seeks to clarify the scope of API deprecation, to inform the community of the replacement API design, and to ensure informed consensus.

Thanks to the long history of MXNet and the accumulated efforts of the community, MXNet now supports a wide range of neural network model training and deployment use cases. Many of these use cases have seen several generations of API design and implementation. Take model training as an example, there have been the Symbol Model API, Symbol Module API, and Gluon Hybrid Block API, all of which coexist in MXNet. Older generations of API often have a significant body of users and thus require time from the community to maintain, even though the supported use cases are mostly covered by a newer generation of API. Such requirement for maintenance not only consumes time and energy of the MXNet community and can distract the community from its longer term goal, but also causes pressure on CI, binary distribution.

In this RFC, we list several candidate API to be deprecated and the corresponding new generation of API as replacement. Unless otherwise stated, these APIs will continue to be supported in the future 1.x releases that happen in parallel to the 2.0 development. On the other hand, participating in the RFC for the new replacement feature of the feature you are interested in is the best way to ensure continued support in 2.0 for that feature. To make it easier to navigate, the replacement feature RFCs are linked in each section.

To make the discussion more productive, I recommend the following actions:

* If a feature in MXNet that you are interested in currently depend on any of the deprecated API, and you plan to switch from 1.x to 2.0, please participate in the RFC for the replacement feature. Please also direct any related questions with respect to how the replacement feature covers your use cases to the replacement feature RFCs.
* If after discussion in the replacement RFCs, you believe that the replacement feature cannot replace the candidate feature for deprecation, please call out in this RFC.
    * Make sure to include your argument on why it’s the case, and clarify what use case cannot be supported in the new feature.
    * If you wish to commit time and sponsor the continued maintenance beyond what’s specified in the following sections, please state so along with your comment.
    * You may also seek other community members to sponsor the feature as comments in this RFC.
    * The group of sponsors needs to collectively clarify what additional support the feature will receive from them, and commit to the cost of maintenance, development, and operations (such as CI).


Please always keep the discussion civilized and informative. Comments otherwise will be folded.

## mxnet.numpy and mxnet.ndarray

[...]

PR #18403, opened 2020-05-25
Gluon.probability

xidulu:
## Description ##
Gluon.probability is part of MXNet 2.0's new features (https://github.com/apache/incubator-mxnet/issues/16167, https://github.com/apache/incubator-mxnet/issues/17240)

It is a module under  `mx.gluon.probability` namespace.

### Module details

This module is made up of three parts:

- Distribution object

- Bijector/Transformation

- StochasticBlock

#### Distribution object

Distribution objects are light-weight object providing useful statistical function of a distribution.

#### Bijector/Transformation

Class representing an invertible or non-invertible transformation.

#### StochasticBlock

Blocks with loss accumlated during forward.

### Demo

A few toy demos can be found in https://github.com/xidulu/mystery_demo, demos still need further refinement, but I believe they can demonstrate some basic functionality of this module.

##### Transformations with learnable parameters
GAN/RealNVP demonstrates how to build a deep generative model with learnable tranformations.
https://github.com/xidulu/mystery_demo/blob/master/GAN.ipynb

https://github.com/xidulu/mystery_demo/blob/master/real_nvp.ipynb

##### Semi-supervised VAE
https://github.com/xidulu/mystery_demo/blob/master/SS-VAE.ipynb

##### Black-box/Auto-differentiation VI for Probabilistic PCA
https://github.com/xidulu/mystery_demo/blob/master/PPCA.ipynb

##### Bayesian Neural Network with local-parameterization trick
https://github.com/xidulu/mystery_demo/blob/master/bnn_demo.ipynb

##### Exclusive KL v.s. Inclusive KL
https://nbviewer.jupyter.org/github/xidulu/mystery_demo/blob/master/KL_demo.ipynb


szha:
nice work on the addition of the new probability module! cc @szhengac @sxjscience @jwmueller @mseeger

python/mxnet/gluon/probability/block/stochastic_block.py
```diff
@@ -0,0 +1,130 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# coding: utf-8
+# pylint: disable=wildcard-import
```
leezu, line 19: Let's not disable pylint flags unnecessarily

python/mxnet/gluon/probability/block/stochastic_block.py
```diff
@@ -0,0 +1,130 @@
...
+    in the forward phase, which is extremely useful in building Bayesian Neural Network,
+    where the loss function is composed of a classification loss and a KL loss.
+
+    """
+
+    def __init__(self, prefix=None, params=None):
+        super(StochasticBlock, self).__init__(prefix=prefix, params=params)
+        self._losses = []
+        self._losscache = []
+        self._count = 0
```
leezu, line 40: Is it possible to just use len(_losscache) instead?

python/mxnet/gluon/probability/block/stochastic_block.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def add_loss(self, loss):
+        self._count += 1
+        self._losscache.append(loss)
+
+    @staticmethod
+    def collectLoss(func):
+        """To accumulate loss during the forward phase, one could first decorate
+        hybrid_forward with `StochasticBlock.collectLos`s`,
```
leezu, line 49: Typo collectLos`s

python/mxnet/gluon/probability/block/stochastic_block.py
```diff
@@ -0,0 +1,130 @@
...
+    def add(self, *blocks):
+        """Adds block on top of the stack."""
+        for block in blocks:
+            self._layers.append(block)
+            self.register_child(block)
+
+    @StochasticBlock.collectLoss
+    def hybrid_forward(self, F, x):
+        for block in self._layers:
+            x = block(x)
```
leezu, line 106: For the `hybrid_forward` API, it seems you forgot to use `F` here?

python/mxnet/gluon/probability/block/stochastic_block.py
```diff
@@ -0,0 +1,130 @@
...
+        self._layers = []
+
+    def add(self, *blocks):
+        """Adds block on top of the stack."""
+        for block in blocks:
+            self._layers.append(block)
+            self.register_child(block)
+
+    @StochasticBlock.collectLoss
+    def hybrid_forward(self, F, x):
```
leezu, line 104: It would be better to override `def forward(self, x)` here, as the `hybrid_forward` is deprecated: https://github.com/apache/incubator-mxnet/blob/5b9aedd933d0fd506de93a1680e107d1f8aa8983/python/mxnet/gluon/block.py#L862-L872

python/mxnet/gluon/probability/block/stochastic_block.py
```diff
@@ -0,0 +1,130 @@
...
+    def __repr__(self):
+        s = '{name}(\n{modstr}\n)'
+        modstr = '\n'.join(['  ({key}): {block}'.format(key=key,
+                                                        block=_indent(block.__repr__(), 2))
+                            for key, block in self._children.items()])
+        return s.format(name=self.__class__.__name__,
+                        modstr=modstr)
+
+    def __getitem__(self, key):
+        layers = list(self._children.values())[key]
```
leezu, line 120: Note the recent change to how HybridSequential stores the _children: https://github.com/apache/incubator-mxnet/pull/18376

You need to take that into account and dereference the weakref here

python/mxnet/gluon/probability/distributions/distribution.py
```diff
@@ -0,0 +1,196 @@
...
+    _validate_args = False
+
+    @staticmethod
+    def set_default_validate_args(value):
+        if value not in [True, False]:
+            raise ValueError
+        Distribution._validate_args = value
+
+    def __init__(self, F=None, event_dim=None, validate_args=None):
+        self.F = F
```
leezu, line 55: Should the F be removed as it's no longer needed?

python/mxnet/gluon/probability/transformation/domain_map.py
```diff
@@ -0,0 +1,109 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# coding: utf-8
+"""Classes for registering and storaging bijection/transformations from
```
leezu, line 19: storing

python/mxnet/gluon/probability/transformation/transformation.py
```diff
@@ -0,0 +1,289 @@
...
+        return self._inv.event_dim
+
+    def __call__(self, x):
+        return self._inv._inverse_compute(x)
+
+    def log_det_jacobian(self, x, y):
+        return -self._inv.log_det_jacobian(y, x)
+
+
+class TransformBlock(Transformation, HybridBlock):
```
leezu, line 118: There may be some issues with multiple inheritance here if constructor arguments for the Block (params, prefix) are used. Did you check it works? But as we plan to remove the Block constructor arguments (https://github.com/apache/incubator-mxnet/pull/18413) it should be fine

python/mxnet/gluon/probability/transformation/transformation.py
```diff
@@ -0,0 +1,289 @@
...
+        raise NotImplementedError
+
+    @property
+    def inv(self):
+        inv = None
+        if self._inv is not None:
+            inv = self._inv()
+        if inv is None:
+            inv = _InverseTransformation(self)
+            self._inv = weakref.ref(inv)
```
leezu, line 68: The idea is to cache the inverse transformation as long as it's in use somewhere else?

src/operator/random/multisample_op.h
```diff
@@ -67,7 +67,7 @@ inline bool MultiSampleOpShape(const nnvm::NodeAttrs& attrs,
   const MultiSampleParam& param = nnvm::get<MultiSampleParam>(attrs.parsed);
   mxnet::TShape sshape = param.shape;
   for (int i = 0; i < sshape.ndim(); ++i) {
-    CHECK_GT(sshape[i], 0) << "shape parameter must be non-zero within each dimension";
+    CHECK_GE(sshape[i], 0) << "shape parameter must be non-zero within each dimension";
```
leezu, line 70: now the error message and the check don't align

tests/python/unittest/test_gluon_probability.py
```diff
@@ -0,0 +1,2326 @@
...
+import scipy.stats as ss
+import scipy.special as scipy_special
+import warnings
+import json
+import unittest
+import random
+import itertools
+from numbers import Number
+
+# set_default_context(mx.gpu(0))
```
leezu, line 42: You should delete this line here and create a separate file in the `gpu` test folder. In that file, import all functions from the current `test_gluon_probability.py` file

tests/python/unittest/test_gluon_probability.py
```diff
@@ -0,0 +1,2326 @@
...
+    sym = mx.sym
+    assert getF(nd.ones((2,2)), nd.ones((2,2))) == nd
+    assert getF(sym.ones((2,2)), sym.ones((2,2))) == sym
+    assert getF(1.0, 2.0) == nd
+
+    # Test exception
+    try:
+        getF(nd.ones((2,2)), sym.ones((2,2)))
+    except TypeError as e:
+        pass
```
leezu, line 78: This doesn't test exception correctly. The test won't fail if no exception is raised. Use pytest.raises

tests/python/unittest/test_gluon_probability.py
```diff
@@ -0,0 +1,2326 @@
...
+            self._func = func
+
+        def hybrid_forward(self, F, low, high, *args):
+            uniform = mgp.Uniform(low, high, validate_args=True)
+            return _distribution_method_invoker(uniform, self._func, *args)
+
+    shapes = [(), (1,), (2, 3), 6]
+
+    # Test log_prob
+    for shape, hybridize in itertools.product(shapes, [True, False]):
```
leezu, line 101: It's better to use pytest.parameterize to enable parallelization

python/mxnet/gluon/probability/transformation/transformation.py
xidulu: I did not check whether setting the prefix would have any issues, but the TransformBlock works well, here is an example https://github.com/xidulu/mystery_demo/blob/master/real_nvp.ipynb

tests/python/unittest/test_gluon_probability.py
xidulu: Oh, I forgot that, I will change it soon.

src/operator/random/multisample_op.h
xidulu: Got that

tests/python/unittest/test_gluon_probability.py
xidulu: You are right, this test is problematic.

python/mxnet/gluon/probability/distributions/distribution.py
xidulu: Why is it no longer needed? Do you mean it will be replaced by deferred computation?

leezu: Yes. Because eventually we like to refactor the backend and replace the nnvm representation. The first step is to remove the explicit reference from the frontend, which implies removing F.

szhengac: In the original design back to last year, we take F into account, and F is not used in every part of the module. If we need to remove F, it will take a lot of effort to redesign. I think removing F can be left to the future.

leezu: > In the original design back to last year, we take F into account, and F is used in every part of the module.

Could you post the reference? Ideally this would have been caught early in a design proposal document. I believe I discussed the plans with Xi in person before.

> If we need to remove F, it will take a lot of effort to redesign.

This PR currently targets the master branch and there wouldn't be any release with the F if we follow https://github.com/apache/incubator-mxnet/issues/17676. So this effort may need to be spent in any case. Now or in the future. However, I think the effort required to fix this PR wouldn't be "a lot". But if it's currently not possible, would it be better to target the 1.x branch and port to the master branch once resources are available?

szhengac: When we started doing this, it was November last year, and Xi told me you mentioned that it is not guaranteed that F can be removed before the release of 2.0. So we did everything based on the original hybrid block, and we cannot test our design if we followed the new api because it was not in MXNET at that time. Also, when the deferred compute was merged, we basically finished all the module except some functionalities. I remembered that I talked to @szha that we can keep F in probability module before.

xidulu: @leezu 
The distribution module kicked off last year, and we had most of the features finished around early January this year when our RFC was opened as one of the items for mxnet 2.0: https://github.com/apache/incubator-mxnet/issues/17240. If we take a look at the RFC for deferred compute, we can see that this RFC was still getting concerns and comments from the community members by late January: https://github.com/apache/incubator-mxnet/issues/16376#issuecomment, and the RFC for deprecating `F` was out late February: https://github.com/apache/incubator-mxnet/issues/17676.
On the other hand, we did respect your effort on deferred compute when we started designing this module, and I remember during one of our earlier offline exchanges, you were not confident enough about finishing this deferred compute feature before 2.0 release. With that piece of information, we eventually decided to proceed with the traditional Gluon (with `F`) in this module. Also, just like what @szhengac said, he reached a consensus with @szha offline about keeping the `F` in this module.
I don’t think it’s reasonable to ask people to predict the future and onboard a new feature that was not guaranteed to be ready.

leezu: @xidulu the timeline of implementation does not have anything to do with ensuring that the final PR integrates well into the overall project. Have you actually taken a look at what steps I requested you to make? I don't think so. The aim of my code-review request here is to make sure that the code you PR is integrated correctly with the Gluon API. For this, you'll need to make a few high-level changes in the way the StochasticBlock is called. Specifically,

- delete self.F = F
- use def forward instead of def hybrid_foward
- Change the checks of isinstance(Symbol) to checking if the Block is hybridized.

These are mostly trivial changes, but I agree there will be an effort to check if any unexpected problems occur. This is nothing unexpected in the case of parallel feature development. The recommendation to start developing against the hybrid_forward API is based on the fact that both APIs are highly related and thus the effort of the final migration is low.

From my perspective it's also fine to keep the F and target the 1.x branch in this PR. Then we may have a separate PR integrating into the master branch. Let me know if you prefer that and I'll be happy to do another review about any issues that the current PR may have with respect to the 1.x API.

xidulu: @leezu 

`StochasticBlock`, in essence, is just a block capable of accumulating loss during the forward propagation. I am very happy to make it more compatible with the new Gluon api. 

Secondly, I don’t think it’s necessary to remove `self.F` in distribution object. In our design, users **do not** need to manually specify `F` when creating a distribution object, as the running mode will be automatically determined by the `getF` function, for example 
https://github.com/apache/incubator-mxnet/pull/18403/files#diff-613421c6ce6fc8ec3338aa29c9535087R48. 

I also made a little modification to the test cases to demonstrate how it works in practice (as well as its compatibility with the new hybridBlock api) https://github.com/apache/incubator-mxnet/pull/18403/files#diff-27fee32587466a570c77da139b7136f6R94-R96

leezu: @xidulu 

> To summarize, the F in the distribution object does not violate the 2.x API. Instead, it can fit well into both 1.x and 2.x API.

Note that the Gluon.HybridBlock in the master branch currently implements both forward and hybrid_forward for transition purpose. The hybrid_forward is deprecated and will be deleted from all HybridBlocks prior to the release. There was just too many code relying on the deprecated interface to delete it outright. This doesn't mean we should add more code relying on the deprecated interface.

Specifically, the `getF` function used in the StochasticBlock would never see a symbolic input. Thus part of the code in your PR can be simplified / updated. Sorry if this background to my previous comment wasn't completely clear.

xidulu: @leezu 

Maybe I did not make it clear in the previous comments, `Distribution` and `getF` both have nothing to do with `Block`, `HybridBlock` or `StochasticBlock`. A `Distribution` object can be used inside any types of blocks or even outside a block. The `getF` function is invoked in the initialization phase of a `Distribution` object to determine *whether to use `sym.np` or `nd.np` for the computation in the member function*, for example:
https://github.com/apache/incubator-mxnet/pull/18403/files#diff-e83cafb33610cdde004c72426d6d1492R82

In short, if the parameters are Symbols, then all the methods in the object will use function under the `sym.np` namespace. If the parameters are NDArray, then we'll choose to use functions under `nd.np`. 

I understand the possibility of `hybrid_forward` getting deleted in the future, but that deprecation hasn't reach community consensus, right? (https://github.com/apache/incubator-mxnet/issues/17676).

Furthermore, as I demonstrate in this test case: https://github.com/apache/incubator-mxnet/pull/18403/files#diff-27fee32587466a570c77da139b7136f6R94-R96, current implementation works perfectly well with the new Gluon API (notice that `F` is not specified in the initialization function), that is to say, distribution object itself **does not rely on** the deprecated interface at all.

What I can do at this moment is:
- Create a new test file based on the new Gluon API. 
- Raise warning to discourage users from explicitly passing `F` when creating a distribution object.

Once `hybrid_forward` got officially deleted from the code base, I would then consider replacing all the `F` in this module with `mx.np` as well as removing `F` from the initialization function.

------------------------------
Your suggestion on StochasticBlock is correct, current implementation is based on the old version of `HybridBlock`, it is necessary to make some modifications.

leezu: @xidulu the consensus in the ticket you reference is to delete `mx.sym` and `mx.nd`. Only `mx.np` and `mx.npx` remain. That's why you don't need the F.

xidulu:

xidulu:

python/mxnet/gluon/probability/block/stochastic_block.py
xidulu: @leezu 
I reimplement the HybridSequential block. The hybrid_forward api is kept because the code is largely borrowed from the implementation of HybridSequential: https://github.com/apache/incubator-mxnet/blob/5b9aedd933d0fd506de93a1680e107d1f8aa8983/python/mxnet/gluon/nn/basic_layers.py#L104
which also overrides the `HybridSequential`.

xidulu:

xidulu:
@mxnet-bot run ci [unix-cpu]

python/mxnet/gluon/probability/block/stochastic_block.py
```diff
@@ -0,0 +1,137 @@
...
+        """Calls forward. Only accepts positional arguments."""
+        for hook in self._forward_pre_hooks.values():
+            hook(self, args)
+        self._losses = []
+        # pylint: disable=no-value-for-parameter
+        out = self.forward(*args)  # out[0]: net output, out[1]: collected loss
+        self._losses.extend(out[1])
+        for hook in self._forward_hooks.values():
+            hook(self, args, out)
+        return out[0]
```
leezu, line 78: You're overwriting `__call__` without calling the parent class implementation:

https://github.com/apache/incubator-mxnet/blob/72a54e7a5f427dc73fbd1cb826ff944d9aa82573/python/mxnet/gluon/block.py#L1321-L1345

python/mxnet/gluon/probability/transformation/transformation.py
szhengac: yes

szhengac:
@leezu any more problems? if not, can you approve it?

@szha

python/mxnet/gluon/probability/block/stochastic_block.py
```diff
@@ -0,0 +1,127 @@
...
+            self._losscache = []
+            return (func_out, collected_loss)
+
+        return inner
+
+    def __call__(self, *args, **kwargs):
+		# pylint: disable=arguments-differ
+        out = super().__call__(*args, **kwargs)
+        self._losses.extend(out[1])
+        return out[0]
```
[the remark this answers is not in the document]
szhengac: @StochasticBlock.collectLoss is only needed when the users want to compute per-block loss such as KL divergence. If such loss is not demanded, the users can simply ignore this decorator or use default HybridBlock.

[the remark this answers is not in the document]
leezu: Thanks for clarifying @szhengac. As per the current implementation, if user ignore this decorator and don't add it when they don't need it, an error will be raised as the `__call__` function always expects two outputs. If the error is correct and expected (ie. adding  `@StochasticBlock.collectLoss` is mandatory), why not remove the decorator for simplicity and handle it's implementation by default (in the `__call__`)?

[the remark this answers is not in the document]
szhengac: If it is not needed, the `out[1]` will be an empty list, so there is no error incurred here. One reason that we use this decorator is that we also need to use it in `StochasticSequential` for collecting losses over multiple blocks. And, as `self.add_loss(kl)` is called inside `forward` or `hybrid_forward`, I think we cannot move the implementation details to `__call__`.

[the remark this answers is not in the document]
leezu: `out[1]` will only be an empty list if the user specifies `@StochasticBlock.collectLoss`. If they don't specify the decorator, `out[1]` will not exist and there'd be an error. We should have a good error message or at least document that users must use `@StochasticBlock.collectLoss`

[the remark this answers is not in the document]
szhengac: Oh yes, you are right. We need to think about it. Thanks for pointing this out.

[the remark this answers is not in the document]
xidulu: @leezu 
Thanks for pointing this out, a error should be raised when users is attempting to call self.add_loss in a function not decorated by `collectLoss`. I will look into this issue.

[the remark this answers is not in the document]
leezu: Thank you @xidulu. Also note that this issue is present when users don't call collectLoss. That's because the __call__ method here is always called when users use StochasticBlock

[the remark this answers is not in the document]
xidulu: @leezu 

I add two checks here: https://github.com/apache/incubator-mxnet/pull/18403/files#diff-85458cf5116b137da8148bf5b38bcfaeR74
https://github.com/apache/incubator-mxnet/pull/18403/files#diff-85458cf5116b137da8148bf5b38bcfaeR78

To make it clearer, I list several possible situations:

1. Users call add_loss inside functions decorated by CollectLoss, add_loss appends losses into _losscache, _losscache would then get cleared in CollectLoss, len(_losscache) becomes 0 when __call__ is invoked.

2. Users call add_loss without using CollectLoss, add_loss appends losses into _losscache, _losscache  still contains value when entering  __call__, in this case, a exception will be raised.

3. Users use CollectLoss without calling add_loss, self._losses = out[1] = []

4. Users use StochasticBlock without calling CollectLoss or add_loss, len(out) == 1, out[1] will not be accessed.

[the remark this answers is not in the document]
xidulu: @leezu 
Update: I made further changes here to avoid confusion. Now the users are forced to use to collectLoss decorator in all cases, otherwise an exception would be raised.

leezu:
@mxnet-bot run ci [unix-gpu]

szha:
@mxnet-bot run ci [unix-gpu]

leezu:
@xidulu any idea why in your PR the `test_np_mixedType_unary_funcs[float16-3-rint-None--5.0-5.0]` test persistently (8 CI runs in a row) is found to leak ndarray references?

xidulu:
@leezu 
I am looking into this issue.

xidulu:
@leezu
Issue solved by adding `garbage_expected` flag: https://github.com/apache/incubator-mxnet/pull/18403/files#diff-a6e158f7742ece308155a95c165505f6R2206

xidulu:
@szha 
Could you help merge this PR?

feevos:
Hi, thank you VERY much for your contribution!!!!!

I have noticed an issue, and am not sure if this is a bug or not (so as to submit a report). Currently, if in the return statement of forward I cannot have a model hybridized and return an object of a distribution. For example, in the VAE code from the mystery_demo, if I want to return the predictions as well as the distribution that I used to get the reconstruction, I get an error when my model is hybridized. 

Example 

```python
import mxnet as mx
mx.npx.set_np()

from mxnet.gluon import HybridBlock
import mxnet.gluon.probability as mgp

class SimpleNet(HybridBlock):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)        
        
    
    def forward(self,input):
        mu=0.0
        scale=1.
        return input, mgp.Normal(mu,scale)

net = SimpleNet()
net.initialize()
net.hybridize() # <=========== if hybridized code exits with error

xx = mx.np.random.rand(3,5)
out = net(xx)
```

error: 

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-6-9e229665d8a2> in <module>
----> 1 out = net(xx)

~/Software/incubator-mxnet/python/mxnet/gluon/block.py in __call__(self, x, *args)
   1645 
   1646             with x.ctx:
-> 1647                 return self._call_cached_op(x, *args)
   1648 
   1649     def forward(self, x, *args):

~/Software/incubator-mxnet/python/mxnet/gluon/block.py in _call_cached_op(self, *args)
   1263     def _call_cached_op(self, *args):
   1264         if self._cached_op is None:
-> 1265             self._build_cache(*args)
   1266 
   1267         if self._first_forward and self._partition_if_dynamic:

~/Software/incubator-mxnet/python/mxnet/gluon/block.py in _build_cache(self, update_graph, *args)
   1136 
   1137     def _build_cache(self, *args, update_graph=True):
-> 1138         data, out = self._get_graph(*args)
   1139         data_names = {data.name: i for i, data in enumerate(data)}
   1140         params = {p.var().name: p for p in self.collect_params().values()}

~/Software/incubator-mxnet/python/mxnet/gluon/block.py in _get_graph(self, *args)
   1132                 return self._get_graph_v1(*args)
   1133             else:  # Gluon 2 based on deferred compute mode
-> 1134                 return self._get_graph_v2(*args)
   1135         return self._cached_graph
   1136 

~/Software/incubator-mxnet/python/mxnet/gluon/block.py in _get_graph_v2(self, *args)
   1122             with autograd.pause(), dc.context():
   1123                 out = super().__call__(*args)
-> 1124             flatten_out, self._out_format = _flatten(out, "output")
   1125             symbol_outputs = dc.get_symbol(flatten_out, sym_cls=type(symbol_inputs[0]))
   1126             self._cached_graph = symbol_inputs, symbol_outputs

~/Software/incubator-mxnet/python/mxnet/gluon/block.py in _flatten(args, inout_str)
    151     fmts = []
    152     for i in args:
--> 153         arg, fmt = _flatten(i, inout_str)
    154         flat.extend(arg)
    155         fmts.append(fmt)

~/Software/incubator-mxnet/python/mxnet/gluon/block.py in _flatten(args, inout_str)
    144 
    145     if not isinstance(args, (list, tuple)):
--> 146         raise ValueError("When hybridized, the input of HybridBlock {}"
    147                          " must be (nested) list of Symbol"
    148                          " or NDArray, "

ValueError: When hybridized, the input of HybridBlock output must be (nested) list of Symbol or NDArray, but got Normal(loc: size (), scale: size (), F: mxnet.ndarray, event_dim: 0) of type <class 'mxnet.gluon.probability.distributions.normal.Normal'>

```

Regards

xidulu:
@feevos 
Hi, in MXNet, you should avoid return anything other than NDArray from the Block, I would say returning a distribution object could be a bad idea.
Also, according to our design, a distribution object is a light-weight on-the-fly wrapper of statistics functions, that is to say, it is not designed to used as the return value or argument of a function.
My suggestion for your case would be first creating a HybridBlock called "encoder" that returns \mu and \sigma or create a function called "get_latents" which returns \mu and \sigma. After you acquire the parameters, you can create a `Normal` object with them and call the function you need.

```python
z_mu, z_sigma = Encoder(x)
qz = Normal(z_mu, z_sigma)
print(qz.sample())
```

feevos:
Thank you @xidulu , I fixed it I was just wondering if it was a bug or not. Again, MANY thanks for your contribution!!!!

Issue #15362, opened 2019-06-25
Reliance on topological ordering for graph inputs


ptrendx:
Currently MXNet relies on topological ordering of the graph when communicating between frontend languages and backend. This makes it hard to apply fusions in the graph, as they could change the topological order of graph inputs and so break passing of the inputs from frontend.

This could be solved by either moving to passing dictionaries (name, value) from frontend to the engine (which requires changing of C API) or making mappings inside Executor and CachedOp from the original topological ordering to the new one.

anirudh2290:
I think this makes sense to me, since the original topological ordering didn't assume mutations like subgraphing and fusion. Related issue: https://github.com/apache/incubator-mxnet/issues/14727 . I think this may require changes to user facing APIs in symbol, module and gluon since at many places ordering of arg_arrays, grad_arrays and aux_arrays depend on the list_arguments and list_inputs which depend on the topo ordering.

PR #18405, opened 2020-05-26
Partition API adding and deleting new params to Block and Symbol

Kh4L:
## Description ##

A custom Partition API backend may add new parameters or delete during the partitioning.

In the Gluon API, the CachedOp needs to received a list of the same size as the number input nodes in the graph. Not deleting the args/aux after partitioning was causing the CachedOp to fail.

### Changes
- Sym: reflects the deletion of parameters in the arg and aux dicts. 
- Gluon: refactor `_build_cache` to allow `_cached_op_args` (the list of parameters passed to the CachedOp) to contain only the subset of parameters needed (excluding the ones deleted by the optimization) from the Block parameters and the ones created during the optimization. The original parameters are not deleted from Block, and _cached_op_args only references the Parameters needed by the model.

### DIFF
--- a/python/mxnet/gluon/block.py
+++ b/python/mxnet/gluon/block.py
@@ -1060,6 +1060,19 @@ class HybridBlock(Block):
                         for name in out.list_auxiliary_states()}
             # Partition the graph.
             out = out.optimize_for(self._backend, arg_dict, aux_dict, ctx, **self._backend_opts)
+            # BFS to delete the delete args/aux from the block Params and its children's Params
+            input_names = out.list_inputs()
+            queue = [self]
+            while len(queue) > 0:
+                curr_block = queue.pop(0)
+                curr_params = curr_block.params if isinstance(curr_block.params, dict) else curr_block.params._params
+                curr_params_names = list(curr_params.keys())
+                for k in curr_params_names:
+                    if k not in input_names:
+                        curr_params.pop(k)
+
+                queue.extend(curr_block._children.values())
+
             #update cached graph with partitioned graph
             self._cached_graph = data, out
 
--- a/python/mxnet/gluon/block.py
+++ b/python/mxnet/gluon/block.py
@@ -1040,29 +1040,16 @@ class HybridBlock(Block):
             warnings.warn("Parameter %s is not used by any computation. "
                           "Is this intended?"%unused, stacklevel=4)
 
-        data_indices = []
-        param_indices = []
-        self._cached_op_args = []
-        for i, name in enumerate(input_names):
-            if name in data_names:
-                data_indices.append(i)
-                self._cached_op_args.append((True, data_names[name]))
-            else:
-                param_indices.append(i)
-                self._cached_op_args.append((False, params[name]))
-        flags = [('data_indices', data_indices), ('param_indices', param_indices)] + \
-                self._flags
-
         args, _ = _flatten(args, "input")
         try:
-            for is_arg, i in self._cached_op_args:
-                if not is_arg:
-                    i.data()
+            for name in input_names:
+                if name in params:
+                    params[name].data()
         except DeferredInitializationError:
             self._deferred_infer_shape(*args)
-            for is_arg, i in self._cached_op_args:
-                if not is_arg:
-                    i._finish_deferred_init()
+            for name in input_names:
+                if name in params:
+                    params[name]._finish_deferred_init()
 
         if self._backend:
             ctx = args[0].context
@@ -1075,6 +1062,20 @@ class HybridBlock(Block):
             out = out.optimize_for(self._backend, arg_dict, aux_dict, ctx, **self._backend_opts)
             #update cached graph with partitioned graph
             self._cached_graph = data, out
+
+        input_names = out.list_inputs()
+        data_indices = []
+        param_indices = []
+        self._cached_op_args = []
+        for i, name in enumerate(input_names):
+            if name in data_names:
+                data_indices.append(i)
+                self._cached_op_args.append((True, data_names[name]))
+            else:
+                param_indices.append(i)
+                self._cached_op_args.append((False, params[name]))
+        flags = [('data_indices', data_indices), ('param_indices', param_indices)] + \
+                self._flags
         self._cached_op = ndarray.CachedOp(out, flags)
 
 
--- a/python/mxnet/gluon/block.py
+++ b/python/mxnet/gluon/block.py
@@ -1051,27 +1051,16 @@ class HybridBlock(Block):
                 if name in params:
                     params[name]._finish_deferred_init()
 
+        arg_dict, aux_dict = dict(), dict()
         if self._backend:
             ctx = args[0].context
             # get list of params in the order of out.list_arguments
-            arg_dict = {name:args[data_names[name]] if name in data_names.keys() else params[name].data()
-                        for name in out.list_arguments()}
-            aux_dict = {name:args[data_names[name]] if name in data_names.keys() else params[name].data()
-                        for name in out.list_auxiliary_states()}
+            arg_dict.update({name:args[data_names[name]] if name in data_names.keys() else params[name].data()
+                             for name in out.list_arguments()})
+            aux_dict.update({name:args[data_names[name]] if name in data_names.keys() else params[name].data()
+                             for name in out.list_auxiliary_states()})
             # Partition the graph.
             out = out.optimize_for(self._backend, arg_dict, aux_dict, ctx, **self._backend_opts)
-            # BFS to delete the delete args/aux from the block Params and its children's Params
-            input_names = out.list_inputs()
-            queue = [self]
-            while len(queue) > 0:
-                curr_block = queue.pop(0)
-                curr_params = curr_block.params if isinstance(curr_block.params, dict) else curr_block.params._params
-                curr_params_names = list(curr_params.keys())
-                for k in curr_params_names:
-                    if k not in input_names:
-                        curr_params.pop(k)
-
-                queue.extend(curr_block._children.values())
 
             #update cached graph with partitioned graph
             self._cached_graph = data, out
@@ -1081,12 +1070,30 @@ class HybridBlock(Block):
         param_indices = []
         self._cached_op_args = []
         for i, name in enumerate(input_names):
+            pair = None
             if name in data_names:
                 data_indices.append(i)
-                self._cached_op_args.append((True, data_names[name]))
+                pair = (True, data_names[name])
             else:
                 param_indices.append(i)
-                self._cached_op_args.append((False, params[name]))
+                if name in params:
+                    param = params[name]
+                else:
+                    assert self._backend, "Parameter " + name + " is missing from block params"
+                    if name in arg_dict or name:
+                        param_data = arg_dict[name]
+                    elif name in aux_dict:
+                        param_data = aux_dict[name]
+                    else:
+                        raise RuntimeError('Expected inputs missing from arg and aux after partioning. '
+                                           'Please check the backend.')
+
+                    param = Parameter(name)
+                    param._load_init(param_data, args[0].context)
+                pair = (False, param)
+
+            self._cached_op_args.append(pair)
+
         flags = [('data_indices', data_indices), ('param_indices', param_indices)] + \
                 self._flags
         self._cached_op = ndarray.CachedOp(out, flags)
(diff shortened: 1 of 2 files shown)


python/mxnet/symbol/symbol.py
```diff
@@ -1554,7 +1563,40 @@ def optimize_for(self, backend, args=None, aux=None, ctx=None, **kwargs):
...
+
+        if isinstance(args, dict):
+            for a_n in deleted_arg_names:
+                if a_n in args:
+                    args.pop(a_n)
+
+        if isinstance(aux, dict):
+            for a_n in deleted_aux_names:
+                if a_n in aux:
+                    aux.pop(a_n)
```
samskalicky, line 1585: Can we do the same if the args/aux are provided as lists?

Kh4L: Actually yes, at first I wanted to give control to the user when params are provided as lists (so he can delete himself the extra params), but then I chose to do it implicitly when params are given as dicts.

I will remove these cumbersome `delete_x` and do the same as what is done for dicts.

Kh4L: Done

python/mxnet/symbol/symbol.py
```diff
@@ -1554,7 +1554,37 @@ def optimize_for(self, backend, args=None, aux=None, ctx=None, **kwargs):
...
+                                     if item not in set(new_aux_names)])
+
+            if isinstance(aux, dict):
+                for a_n in deleted_aux_names:
+                    if a_n in aux:
+                        aux.pop(a_n)
+            elif isinstance(args, list):
+                indices_to_delete = [i for i, name in enumerate(aux_names) if name in deleted_aux_names]
+                for idx in reversed(indices_to_delete):
+                    aux.pop(idx)
```
samskalicky, line 1585: @ptrendx I remember we discovered a problem previously that the ordering of inputs may change after partitioning #15362. What should we do here if the args/aux are a list and not a dict? Should we re-order them so they match the new ordering after calling optimize_for (which can execute graph partitioning, or graph passes)?

python/mxnet/gluon/block.py
```diff
@@ -1037,10 +1024,25 @@ def _build_cache(self, *args):
                          for name in out.list_arguments()]
             aux_array = [args[data_names[name]] if name in data_names.keys() else params[name].data()
                          for name in out.list_auxiliary_states()]
+
             # Partition the graph.
             out = out.optimize_for(self._backend, arg_array, aux_array, ctx, **self._backend_opts)
```
samskalicky, line 1029: I think we need to remove deleted params from the `params` dict too. Otherwise we just remove them from the `arg_array` and `aux_array` but not necessarily from the `params` or `self._params`

python/mxnet/symbol/symbol.py
ptrendx: The API should not accept lists - it is a source of a lot of evil.

ptrendx: Also, @mk-61 FYI.

python/mxnet/symbol/symbol.py
```diff
@@ -1554,7 +1554,37 @@ def optimize_for(self, backend, args=None, aux=None, ctx=None, **kwargs):
...
+            new_arg_names = new_sym.list_arguments()
+            deleted_arg_names = set([item for item in arg_names
+                                     if item not in set(new_arg_names)])
+            if isinstance(args, dict):
+                for a_n in deleted_arg_names:
+                    if a_n in args:
+                        args.pop(a_n)
+            elif isinstance(args, list):
+                indices_to_delete = [i for i, name in enumerate(arg_names) if name in deleted_arg_names]
+                for idx in reversed(indices_to_delete):
```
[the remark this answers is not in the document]
Kh4L: I thought that using list comprehension to get the indices_to_delete was simpler, since the indices list have to be reversed.

```
for idx, name  in reversed(list(enumerate(l))):
   if name in deleted_arg_names:
```

we could do that, but I am not sure if it is more readable and faster.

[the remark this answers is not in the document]
samskalicky: Agreed. Never mind, if we decide to drop list support it won’t matter anyway. Thanks again for this.

python/mxnet/symbol/symbol.py
samskalicky: PR removing support for lists as args/aux https://github.com/apache/incubator-mxnet/pull/18483

python/mxnet/gluon/block.py
Kh4L: Good catch, I added it in the latest commit

python/mxnet/gluon/block.py
```diff
@@ -1073,8 +1060,35 @@ def _build_cache(self, *args):
...
+            # BFS to delete the delete args/aux from the block Params and its children's Params
+            input_names = out.list_inputs()
+            queue = [self]
+            while len(queue) > 0:
+                curr_block = queue.pop(0)
+                curr_params = curr_block.params if isinstance(curr_block.params, dict) else curr_block.params._params
+                curr_params_names = list(curr_params.keys())
+                for k in curr_params_names:
+                    if k not in input_names:
+                        curr_params.pop(k)
```
samskalicky, line 1072: removing params from a python-level block puts the model in bad state, because now it doesnt work in Python. @leezu

samskalicky: instead of deleting and modifying the Python-level state, consider keeping an internal copy of collect_params and modifying that. Would also need to modify export API to use this internal copy if set.

leezu: Thank you @samskalicky for summarizing our discussion!

python/mxnet/gluon/block.py
```diff
@@ -1335,12 +1337,21 @@ def export(self, path, epoch=0, remove_amp_cast=True):
         arg_names = set(sym.list_arguments())
         aux_names = set(sym.list_auxiliary_states())
         arg_dict = {}
-        for param in self.collect_params().values():
-            if param.name in arg_names:
-                arg_dict['arg:%s'%param.name] = param._reduce()
-            else:
-                assert param.name in aux_names
-                arg_dict['aux:%s'%param.name] = param._reduce()
+        if self._cached_op_args:
```
leezu, line 1340: When would self._cached_op_args not be present?

Kh4L: I was missing that export requires hybridization, fixing it now

python/mxnet/gluon/block.py
```diff
@@ -1119,8 +1135,8 @@ def _call_cached_op(self, *args):
                                  .format(fmt, self._in_format))
 
         args_without_none = [ele for ele in args if ele is not None]
-        cargs = [args_without_none[i] if is_arg else i.data()
-                 for is_arg, i in self._cached_op_args]
+        cargs = [args_without_none[i] if is_arg else i
+                 for is_arg, i, _ in self._cached_op_args]
```
leezu, line 1139: Would there be an issue now when passing arrays on a different context? Ie, initially using gpu(0) but later using gpu(1). 
Because self._cached_op_args now contains the .data() array directly instead of the parameter (which may manage arrays on multiple devices)

leezu: https://github.com/apache/incubator-mxnet/blob/d9ec943194dc88942932cc2d9a988b2fc4ce69ee/python/mxnet/gluon/block.py#L1422-L1429

Kh4L:

Kh4L:
@mxnet-bot run ci [centos-gpu, unix-cpu]

Kh4L:

python/mxnet/gluon/block.py
```diff
@@ -1040,41 +1040,62 @@ def _build_cache(self, *args):
...
+        arg_dict, aux_dict = dict(), dict()
         if self._backend:
             ctx = args[0].context
             # get list of params in the order of out.list_arguments
-            arg_dict = {name:args[data_names[name]] if name in data_names.keys() else params[name].data()
-                        for name in out.list_arguments()}
-            aux_dict = {name:args[data_names[name]] if name in data_names.keys() else params[name].data()
-                        for name in out.list_auxiliary_states()}
+            arg_dict.update({name:args[data_names[name]] if name in data_names.keys() else params[name].data()
+                             for name in out.list_arguments()})
```
samskalicky, line 1059: Why do you initialize to `dict()` and then call update on an empty dictionary instead of just assigning?

leezu: `arg_dict` and `aux_dict` could otherwise be undefined below the `assert` in line 1083. This could be a SyntaxError or linter error?

Kh4L: That's right, also Python scopes rules are sometimes a bit unsettling. So I thought that this would make it clearer

samskalicky: makes sense, thanks!

python/mxnet/gluon/block.py
```diff
@@ -1040,41 +1040,62 @@ def _build_cache(self, *args):
...
+        data_indices = []
+        param_indices = []
+        self._cached_op_args = []
+        for i, name in enumerate(input_names):
+            pair = None
+            if name in data_names:
+                data_indices.append(i)
+                pair = (True, data_names[name])
+            else:
+                param_indices.append(i)
```
samskalicky, line 1078: nit: can we move this above the if/else since we do it in both?

python/mxnet/gluon/block.py
```diff
@@ -1040,41 +1040,62 @@ def _build_cache(self, *args):
...
+                if name in params:
+                    param = params[name]
+                else:
+                    assert self._backend, "Parameter " + name + " is missing from block params"
```
samskalicky, line 1082: Is this the case that should never happen? When a param name is not in `params` but `_backend` is not set? If so can we change the error message and maybe leave a comment?

python/mxnet/gluon/block.py
```diff
@@ -1040,41 +1040,62 @@ def _build_cache(self, *args):
...
+                    if name in arg_dict or name:
+                        param_data = arg_dict[name]
+                    elif name in aux_dict:
+                        param_data = aux_dict[name]
+                    else:
+                        raise RuntimeError('Expected inputs missing from arg and aux after partioning. '
```
samskalicky, line 1088: Is this the case when the backend added a param to the graph but not to the arg/aux dict? If so, can we change the error message to say something like "param <x> was added to the graph but the tensor was not added to args/aux"?

python/mxnet/gluon/block.py
```diff
@@ -1437,6 +1460,23 @@ def hybrid_forward(self, F, x, *args, **kwargs):
...
+            Assign Parameter to given context. If ctx is a list of Context, a
+            copy will be made for each context.
+        """
+        params = self.collect_params()
+        if self._cached_op:
+            for p in self._cached_op_args:
+                # resetting parameters creating by the partitioning backend
+                if p.name not in params:
+                    p.reset_ctx(ctx)
+        for p in params.values():
```
samskalicky, line 1478: Can you explain why we need to loop over `_cached_op_args` and only reset the params not in `params`, and then loop again over `params` and then reset them there instead? Is it possible to do the work in the 2nd loop in the first loop?

samskalicky: Although i guess if we delete a param, then it will still be in `params` but not in `_cached_op_args`. And the context check will fail if we dont do all the params, so i guess this makes sense. Maybe we should have a comment that `params` and `_cached_op_args` might contain unique params (ie. there is no superset)?

python/mxnet/symbol/symbol.py
```diff
@@ -1544,8 +1544,35 @@ def optimize_for(self, backend, args=None, aux=None, ctx=None, **kwargs):
...
+                                 if item not in set(new_arg_names)])
+
+        if len(deleted_arg_names) > 0:
+            if args is not None:
+                for a_n in deleted_arg_names:
+                    if a_n in args:
+                        args.pop(a_n)
+            else:
+                warnings.warn('optimize_for deleted some argument. \n' +
+                              'Provide a dictionary to the arg argument to optimize_for')
```
samskalicky, line 1561: Can we clarify this error message to something like:
```
A param was deleted during optimization, but no args dictionary was provided. Please ensure that your model weights match the newly optimized model. 
```
Should we print the names of the deleted_arg_names in the message too (or is that overkill)?

python/mxnet/symbol/symbol.py
```diff
@@ -1544,8 +1544,35 @@ def optimize_for(self, backend, args=None, aux=None, ctx=None, **kwargs):
...
+        deleted_aux_names = set([item for item in aux_names
+                                 if item not in set(new_aux_names)])
+        if len(deleted_aux_names) > 0:
+            if aux is not None:
+                for a_n in deleted_aux_names:
+                    if a_n in aux:
+                        aux.pop(a_n)
+            else:
+                warnings.warn('optimize_for deleted some aux argument. \n' +
+                              'Provide a dictionary to the aux argument to optimize_for')
```
samskalicky, line 1573: Same as above for args

python/mxnet/symbol/symbol.py
Kh4L: Thanks for the suggestion! I think that this new message will be sufficient, printing the list of deleted arg names could be potentially very long and flood the stderr

python/mxnet/gluon/block.py
Kh4L: Done

python/mxnet/gluon/block.py
Kh4L: > Although i guess if we delete a param, then it will still be in params but not in _cached_op_args. 
That is the reason why
We don't want to do any additional reset_ctx: reset is costly because it copies NDArrays 

I will add some comments to clarify

Issue #18412, opened 2020-05-27
Simplifying mxnet.gluon.block APIs


acphile:
## Motivations
Currently the implementation of mxnet.gluon.block is not so pythonic and there are many redundancies 

### 1. overlaps between Block._params and Block._reg_params 
when we want to self-define a model, we currently need to use the code as follows:
```
class Net(nn.HybridBlock):
    def __init__(self, **kwargs):
        super(HybridNet, self).__init__(**kwargs)
        with self.name_scope():
            self.hidden1 = nn.Dense(256, activation='relu')
            self.a=self.params.get('a', shape=(1, ))        
```
There are several shortcomings when using this form of registration:
a. adding parameter ‘a’ will lead to double recordings in both self._params and self._reg_params, which is a redundancy. And there is also a discrepancy in Block:
&nbsp;&nbsp; &nbsp;&nbsp; i. In the method “collect_params”, we use “_params” to get all parameters
&nbsp;&nbsp;&nbsp;&nbsp; ii. while in the method “_collect_params_with_prefix” (and methods “load_parameters” accordingly), we use “_reg_params” to get all parameters.
b. Currently if we do not use “with self.name_scope():” for children blocks, it will lead to wrong name scopes. For the following example, we actually can not get the parameters of self.hidden1 from the result of collect_params
```
class HybridNet(nn.HybridBlock):
    def __init__(self, **kwargs):
        super(HybridNet, self).__init__(**kwargs)
        self.hidden1 = nn.Dense(256, activation='relu')
        with self.name_scope():
            self.hidden2 = nn.Dense(10, activation='relu')

    def hybrid_forward(self, F, x):
        x = self.hidden2(self.hidden1(x))
        return x
    
>>> net = HybridNet()
>>> net.initialize()
>>> print(net.collect_params())
hybridnet0_ (
  Parameter dense0_weight (shape=(256, -1), dtype=float32)
  Parameter dense0_bias (shape=(256,), dtype=float32)
  Parameter hybridnet0_dense0_weight (shape=(10, -1), dtype=float32)
  Parameter hybridnet0_dense0_bias (shape=(10,), dtype=float32)
)
```
From the above example we can also find that the parameter names are not related to the attributes’ names, which is not straightforward.

In all, we find that using name_scope and ParameterDict is not user-friendly. Thus we plan to remove such redundancies and simplify the definitions of children blocks and parameters, like:
```
class Net(nn.HybridBlock):
    def __init__(self, **kwargs):
[...]

PR #18413, opened 2020-05-27
Simplifying mxnet.gluon.block APIs

acphile:
## Description ##
A draft PR for issue https://github.com/apache/incubator-mxnet/issues/18412

### Essentials ###

### Changes ###

## Comments ##


acphile:

tests/python/unittest/test_gluon_estimator.py
```diff
@@ -394,24 +388,6 @@ def test_val_net():
                     val_loss=val_loss,
                     val_net=val_net)
 
-    with pytest.raises(RuntimeError):
-        est.fit(train_data=dataloader,
-                val_data=dataloader,
-                epochs=num_epochs)
-
-    ''' test weight sharing of sequential networks with namescope '''
```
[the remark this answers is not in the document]
acphile: Actually test for new method of weight sharing exists. Originally, there are two tests: "test weight sharing of sequential networks with namescope" and "test weight sharing of sequential networks without namescope". Since in new API, we don't have to define children block inside "with net.name_scope()", I remove one of the tests. You can see in "def _get_test_network(params=None)", I maintain the original test for weight sharing.

leezu:
@mxnet-bot run ci [website, unix-cpu]

python/mxnet/gluon/block.py
```diff
@@ -542,61 +460,60 @@ def load_parameters(self, filename, ctx=None, allow_missing=False,
...
+
+    def load_dict(self, param_dict, filename=None, ctx=None, allow_missing=False,
+                  ignore_extra=False, cast_dtype=False, dtype_source="current"):
+        """Load parameters from dict
+
+        Parameters
+        ----------
+        param_dict : dict
+            Dictionary containing model parameters
+        filename : str, default None
```
leezu, line 476: When making `load_dict` a public function (instead of internal), it shouldn't take a `filename` argument. The use of filename here is just due to how you implement `load_parameters`.

python/mxnet/gluon/parameter.py
```diff
@@ -431,7 +437,10 @@ def initialize(self, init=None, ctx=None, default_init=initializer.Uniform(),
             and :py:meth:`Parameter.init` are ``None``.
         force_reinit : bool, default False
             Whether to force re-initialization if parameter is already initialized.
-
+        structural_name : str, default ""
+            The structural name for the parameter in the block.
+            The value would be accessed in InitDesc.attrs['structure'] by self-defined initializers.
+            Users may want to initialize parameters based on the block's structure
```
leezu, line 443: Unused?

sxjscience:
@leezu This is a major API upgrade and we will need to rewrite + test the majority of gluonnlp/numpy: https://github.com/dmlc/gluon-nlp/tree/numpy

leezu:
@sxjscience Yes.. Though I believe this is unavoidable. Not even the D2L book manages to introduce the existing API correctly.. For example, https://d2l.ai/chapter_deep-learning-computation/model-construction.html is plain wrong in MXNet 1..

RuRo:
**(Disclamer)** I didn't look too deeply into the reasons for these changes, so I may be wrong, but... **(Disclamer)**

In my opinion, the distinction between the "unique/unreadable" parameter names and the "human-readable" parameter names is quite confusing. It's not obvious, where should you expect which kind of parameter name.

For example, currently `.collect_params()` seems to return human-readable names as keys, but the parameter values themselves have "unique" `.name`s.

Exceptions, errors and warnings (for example the "Parameter has not been initialized." error) currently seem to print the "unique" parameter name, which is really unhelpful for debugging purposes, since it doesn't tell you, where is the uninitialized parameter located.

I am not very familiar with the inner workings of Gluon Parameters, but it really feels like there has to be a better way.

leezu:
@RuRo Thanks for pointing out the problem. Generally the structural name should be the "public" or user-visible name. The UUID is mainly to workaround some buggy MXNet operators that have global state based on symbol names. So overall the idea is to get rid of any global topological information `parameter.name`. Let's open an issue to track the problems where the internal name is exposed and improve the behaviour.

acphile:
For warnings in method `intialize()`, I pass parameter `structural_name` which is a readable name based on model structure, and we can replace `self.name` with `self._structural_name`. But I am not sure about other methods of `Parameter`: if those method are executed after `initialize()`, then we can do similar replacements accordingly.

PR #18424, opened 2020-05-28
environment variable handling in unittests

DickJC123:
This PR makes it easy to create unittests that require specific settings of environment variables, while avoiding the pitfalls (discussed in comments section).   This PR can be considered a recasting and expansion of the great vision of @larroy in creating the EnvManager class in https://github.com/apache/incubator-mxnet/pull/13140.

In its base form, the facility is a drop-in replacement for EnvManager, and is called 'environment':

```
with environment('MXNET_MY_NEW_FEATURE', '1'):
    <test with feature enabled>
with environment('MXNET_MY_NEW_FEATURE', '0'):
    <test with feature disabled>
```
Like EnvManager, this facility takes care of the save/restore of the previous environment variable state, including when exceptions are raised.  In addition though, this PR introduces the features:

- A similarly-named unittest decorator: `@with_environment(key, value)`
- The ability to pass in multiple env vars as a dict (as is needed for some tests) in both forms, so for example:
```
with environment({'MXNET_FEATURE_A': '1',
                  'MXNET_FEATURE_B': '1'}):
    <test with both features enabled>
```
- Works on Windows!  This PR includes a wrapping of the backend's setenv() and getenv() functions, and uses this direct access to the backend environment to keep it in sync with the python environment.  This works around the problem that the C Runtime on Windows gets a snapshot of the Python environment at startup that is immutable from Python.  
- `with environment()` has a simple implementation using the @contextmanager decorator
- Tests are included that validate the facility works with all combinations of before_val/set_val, namely unset/unset, unset/set, set/unset, set/set.

There were 5 unittests previously using EnvManager, and this PR shifts those uses to `with environment():`, while converting over 20 other ad-hoc uses of os.environ[] within the unittests.  This PR also enables those unittests that were bypassed on Windows (due to the inability to set environment variables) to run on all platforms.

### Essentials ###
- [X ] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###

## Further Comments ##
Environment variables are a two-edged sword- they enable useful operating modes for testing, debugging or niche applications, but like all features they must be tested.  The correct approach for testing with a particular env var setting is:
```
def set_env_var(key, value):
    if value is None:
        os.environ.pop(key, None)
    else:
        os.environ[key] = value

old_env_var_value = os.environ.get(env_var_name)
try:
    set_env_var(env_var_name, test_env_var_value)
    <perform test>
finally:
    set_env_var(env_var_name, old_env_var_value )
```
The above code makes no assumption about whether the before-test and within-test state of the env var is set or unset, and restores the prior environment even if the test raises an exception.  This represents a lot of boiler-plate code that could be potentially mishandled.  The `with environment()` context makes it simple to handle all this properly.  If an entire unittest wants a forced env var setting, then using the `@with_environment()` decorator avoids the code indent of the `with environment()` approach if used otherwise within the test.


tests/python/unittest/test_base.py
```diff
@@ -17,32 +17,91 @@
 
 import mxnet as mx
 from mxnet.base import data_dir
+from nose.tools import *
```
leezu, line 20: Should this line be deleted as nose is no longer in use?

tests/python/gpu/test_kvstore_gpu.py
```diff
@@ -20,8 +20,8 @@
 import os
 import mxnet as mx
 import numpy as np
-import pytest
-from mxnet.test_utils import assert_almost_equal, default_context, EnvManager
+import unittest
```
leezu, line 23: Should this line be deleted?

python/mxnet/test_utils.py
```diff
@@ -2380,22 +2362,77 @@ def same_symbol_structure(sym1, sym2):
...
+    args:
+        if 2 args are passed:
+            name, desired_value strings of the single env var to update, or
+        if 1 arg is passed:
+            a dict of name:desired_value for env var's to update
+
+    """
+
+    # On Linux, env var changes made through python's os.environ are seen
+    # by the backend.  On Windows though, the C runtime gets a snapshot
```
larroy, line 2386: :+1:   That difference can be nasty.

src/c_api/c_api_test.cc
```diff
@@ -106,3 +106,25 @@ int MXRemoveSubgraphPropertyOpNamesV2(const char* prop_name) {
...
+             const char** value) {
+  API_BEGIN();
+  *value = getenv(name);
+  API_END();
+}
+
+int MXSetEnv(const char* name,
+             const char* value) {
+  API_BEGIN();
+#ifdef _WIN32
```
larroy, line 120: Would it be a good idea to put a mutex? setenv is not thread safe.

tests/python/gpu/test_kvstore_gpu.py
DickJC123: I agree, so I've deleted this line.

src/c_api/c_api_test.cc
DickJC123: The sequence we follow of checkpoint_env_vars, change_env_vars, run_test, reinstate_env_vars, is fundamentally not thread-safe.  I'm not making those guarantees, so no mutex is required I feel.

DickJC123:
I have another clean-up commit or two on this...

DickJC123:
Sorry I got side-tracked on a different PR and let this sit idle for awhile.  At this point, I'm done with the prep of this PR and would be happy to respond to a reviewer comments.

DickJC123:
@MoisesHer Thanks for your patience- I know you've been waiting for this PR's functionality for your PR.  Stay tuned- I'm hopeful you'll be unblocked soon.

DickJC123:
@leezu Thanks for your initial comments from way back when I first posted this PR.  I've responded to those comments and polished the PR a bit so it passes CI.  No major changes in this last go-over.  Do you have time and desire to re-review?

DickJC123:
Thanks @leezu! It will be nice to code up unittests with env vars now without having to special-case the Windows CI runner.

leezu:
Thank you @DickJC123 for your work on this. Would you like to create a tracking issue for that? If there are somewhat straightforward instructions for identifying these tests, it could be a good first issue for new contributors familiar with Python. What do you think?

DickJC123:
In the past, someone would code up a unittest with env vars, it would fail on Windows, they'd eventually figure out it had to do with the immutable env var state, then they'd bypass the Windows runner.  Now for the uninitiated, things should just work...assuming they know to use 'with environment():'.

So in a sense, we should have had the tracking issue alerting to the pitfalls of env var unittests, but now we don't need it.

Or @leezu, are you suggesting we post an issue, something to the effect of: "If you plan on manipulating env vars in unittests, make sure you do it via 'with environment()'?

leezu:
@DickJC123 I misunderstood your previous comment in that I assumed there are more unittests that are disabled on windows and can be enabled now. As you fixed all of them, no need to create an issue. Thanks for your work!

DickJC123:
Right, the PR enables all tests to run on Windows.

PR #18438, opened 2020-05-29
CI: Prime docker cache in "Docker cache build" job

leezu:
Cache priming was previously only handled when using the ./ci/build.py or
./ci/docker_cache.py script. The former is used during most CI jobs. The latter
is used for containers not declared in docker-compose.yml during the "Docker
cache build" job. As more and more containers are declared in
docker-compose.yml, we need to fix the cache priming issue.


leezu:
@mxnet-bot run ci [unix-cpu]

PR #18440, opened 2020-05-29
Fix backward of scalar variants of arctan2 and rarctan2 on GPU

ptrendx:
## Description ##
The used BinaryScalarOp::Compute instead of BinaryScalarOp::Backward. CPU version is using the proper function.

It seems there are no tests for `_scalar` versions of NumPy operators, FYI @haojin2 @sxjscience @yzhliu


yzhliu:
Thanks @ptrendx . Can we add test cases for this particular case to prevent regression?

ptrendx:
Hi @yzhliu, I encountered it while working on a different thing, so do not really have time to do a comprehensive check for all binary scalar operators for numpy, and I believe such check is necessary rather than a spot check of just those operations. That's why I provided a quick fix for the problem, but would prefer if somebody else enhanced the tests.

szha:
let's merge this first and we will try to use code coverage tools to identify where tests are lacking

PR #18441, opened 2020-05-29
Update nccl installation in Dockerfile.build.centos7

leezu:
By moving the NCCL installation before the `COPY runtime_functions.sh` statement, users will not
suffer from cache invalidation due to the COPY and don't have to repeat the
nccl installation locally as there will be a cache hit when using `python ci/build.py --platform centos7_gpu_cu100 /work/runtime_functions.sh build_static_libmxnet cu100`.


leezu:

Issue #18442, opened 2020-05-29
CI: tests hang during test collection stage


leezu:
## Description
```
[2020-05-29T18:49:28.268Z] + pytest --durations=50 --cov-report xml:tests_unittest.xml --verbose tests/python/unittest

[2020-05-29T18:49:31.528Z] ============================= test session starts ==============================

[2020-05-29T18:49:31.528Z] platform linux -- Python 3.6.9, pytest-5.3.5, py-1.8.1, pluggy-0.13.1 -- /usr/bin/python3

[2020-05-29T18:49:31.528Z] cachedir: .pytest_cache

[2020-05-29T18:49:31.528Z] rootdir: /work/mxnet, inifile: pytest.ini

[2020-05-29T18:49:31.528Z] plugins: flaky-3.6.1, xdist-1.31.0, timeout-1.3.4, cov-2.8.1, env-0.6.2, forked-1.1.3

[2020-05-29T18:49:31.528Z] timeout: 1200.0s

[2020-05-29T18:49:31.528Z] timeout method: signal

[2020-05-29T18:49:31.528Z] timeout func_only: False

[2020-05-29T22:44:00.825Z] Sending interrupt signal to process

```
http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Funix-cpu/detail/PR-18391/1/pipeline/366

or 

```
[2020-05-21T18:58:56.773Z] + CI_CUDA_COMPUTE_CAPABILITIES='-gencode=arch=compute_52,code=sm_52 -gencode=arch=compute_70,code=sm_70'

[2020-05-21T18:58:56.773Z] + CI_CMAKE_CUDA_ARCH='5.2 7.0'

[2020-05-21T18:58:56.773Z] + set +x

[2020-05-21T18:58:56.773Z] + export PYTHONPATH=./python/

[2020-05-21T18:58:56.773Z] + PYTHONPATH=./python/

[2020-05-21T18:58:56.773Z] + export DMLC_LOG_STACK_TRACE_DEPTH=10

[2020-05-21T18:58:56.773Z] + DMLC_LOG_STACK_TRACE_DEPTH=10

[2020-05-21T18:58:56.773Z] + python3 tests/python/unittest/onnx/backend_test.py

[2020-05-21T22:52:22.501Z] Sending interrupt signal to process

```

[...]

PR #18443, opened 2020-05-29
Disable test coverage in MKL builds

leezu:
Reference: https://github.com/apache/incubator-mxnet/issues/18244#issuecomment-635911844 

cc: @bgawrych


marcoabreu:
It is expected that test coverage recording introduces slow down. Especially since core logic is run on the cpu for mkl while for GPU the core logic is run on GPU. 

I'd rather exclude the mkl core package from test coverage than disabling it entirely.

For this, I'd recommend looking at where most CPU time is spent.

leezu:
I agree with you that this solution shouldn't be the final solution. However, currently the MKL tests are quite unstable on CI. @bgawrych is looking into it. Until we have a proper solution, I suggest to disable the test coverage for MKL builds to make the CI more stable.

marcoabreu:
The issue is talking about speed but you're talking about stability. Are the test coverage injection causing instability?

leezu:
The stability problem I observed seems unrelated (https://github.com/apache/incubator-mxnet/issues/18442). However, disabling the test coverage at this point is a low-hanging fruit to improve the CI: MKL tests take 3h+ whereas all other tests take <1h.

ci/docker/runtime_functions.sh
```diff
@@ -1014,8 +1001,10 @@ unittest_ubuntu_python3_cpu_mkldnn() {
     export MXNET_SUBGRAPH_VERBOSE=0
     export MXNET_ENABLE_CYTHON=0
     export DMLC_LOG_STACK_TRACE_DEPTH=10
-    # TODO(szha): enable parallel testing and naive engine for ops once #18244 is fixed
-    pytest --durations=50 --cov-report xml:tests_unittest.xml --verbose tests/python/unittest
+    pytest -m 'not serial' -k 'not test_operator' -n 4 --durations=50 --cov-report xml:tests_unittest.xml --verbose tests/python/unittest
```
bgawrych, line 1004: It would be better to set OMP_NUM_THREADS=total_cpu_threads/4. Otherwise multiprocessing module from python will spawn 4 processes which will be using all threads independently (4 x total_cpu_threads).
For serial you can then unset OMP_NUM_THREADS

leezu:
@mxnet-bot run ci [centos-cpu, centos-gpu, unix-cpu]

leezu:
@mxnet-bot run ci [centos-cpu, centos-gpu]

PR #18444, opened 2020-05-30
More clear description to `transform_first`

goldmermaid:
After teaching an MLU-CV class, some students feel confused when they are using the `transform_first` function. So I modified it a bit to make it more clear. :)


wkcn:
The PR has been merged. Thank you!

PR #18450, opened 2020-05-31
[MXNET-1451] Enable MXNet-TRT for building Jetson wheels

TristonC:
## Description ##
It enables build MXNet on Jetson with TensorRT support. Without this, the TensorRT integration source code under mxnet/src/operator/subgraph/tensorrt will not be correctly compiled. 

### Essentials ###
- 
- [x] The PR title starts with [MXNET-1451], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)

### Changes ###
- Turn USE_TENSRORT option on for cmake to enable tensorrt source files under src/operator/subgraph/tenssorrt
- Change the CUDA arch from 5.2 to 5.3

## Comments ##
- This change should only affect MXNet jetson wheels. Current jetson wheel build is done on only one GPU arch.

### DIFF
--- a/ci/docker/Dockerfile.build.jetson
+++ b/ci/docker/Dockerfile.build.jetson
@@ -38,6 +38,9 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
     python3 \
     python3-pip \
     awscli \
+    autoconf \
+    automake \
+    libtool \
     crossbuild-essential-arm64 \
  && rm -rf /var/lib/apt/lists/*
 
--- a/ci/docker/Dockerfile.build.jetson
+++ b/ci/docker/Dockerfile.build.jetson
@@ -42,7 +42,8 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
     automake \
     libtool \
     crossbuild-essential-arm64 \
- && rm -rf /var/lib/apt/lists/*
+ && rm -rf /var/lib/apt/lists/* \
+ && alias python=python3
 
 # cmake on Ubuntu 18.04 is too old
 RUN python3 -m pip install cmake
--- a/ci/docker/Dockerfile.build.jetson
+++ b/ci/docker/Dockerfile.build.jetson
@@ -61,6 +61,7 @@ RUN git clone --recursive -b v0.3.9 https://github.com/xianyi/OpenBLAS.git && \
 # Install aarch64 cross depedencies based on Jetpack 4.3
 # Manually downloaded using SDK Manager tool and placed in a private S3 bucket.
 # We're not allowed to redistribute these files and there is no public version.
+# Install protobuf and protobuf-compiler for TensorRT integration.
 RUN aws s3 cp s3://mxnet-ci-prod-private-slave-data/nvidia/sdkm_downloads/cuda-repo-ubuntu1804-10-0-local-10.0.326-410.108_1.0-1_amd64.deb . && \
     dpkg -i cuda-repo-ubuntu1804-10-0-local-10.0.326-410.108_1.0-1_amd64.deb && \
     rm cuda-repo-ubuntu1804-10-0-local-10.0.326-410.108_1.0-1_amd64.deb && \
@@ -71,6 +72,7 @@ RUN aws s3 cp s3://mxnet-ci-prod-private-slave-data/nvidia/sdkm_downloads/cuda-r
     apt-get update && \
     apt-get install -y -f && \
     apt-get install -y cuda-cross-aarch64 cuda-cross-aarch64-10-0 && \
+    apt-get install -y libprotobuf-dev protobuf-compiler && \
     rm -rf /var/lib/apt/lists/*
 
 ARG USER_ID=0
--- a/ci/docker/Dockerfile.build.jetson
+++ b/ci/docker/Dockerfile.build.jetson
@@ -42,8 +42,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
     automake \
     libtool \
     crossbuild-essential-arm64 \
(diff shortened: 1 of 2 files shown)


TristonC:
The build failed due to "Could NOT find Protobuf (missing: Protobuf_INCLUDE_DIR)". Protobuf needs to be installed for the MXNet-TRT integration.

TristonC:
@mxnet-bot run ci edge

leezu:
@TristonC I suspect the issue is due to using the toolchain file for crosscompilation. It sets `CMAKE_FIND_ROOT_PATH` and thus cmake may look for protobuf in the target toolchain. I didn't look into the recommended setup for this setting, but I suspect you may be able to workaround this issue by manually setting the `Protobuf_INCLUDE_DIR`. If that doesn't work, installing aarch64 protobuf to the target toolchain could be required

TristonC:
> 
> 
> @TristonC I suspect the issue is due to using the toolchain file for crosscompilation. It sets `CMAKE_FIND_ROOT_PATH` and thus cmake may look for protobuf in the target toolchain. I didn't look into the recommended setup for this setting, but I suspect you may be able to workaround this issue by manually setting the `Protobuf_INCLUDE_DIR`. If that doesn't work, installing aarch64 protobuf to the target toolchain could be required

 @leezu  I was looking exactly the same issue. I am comparing it with our internal build process and see if that works for this CI or not. There will be a couple of more commits to make it ready. Do you know how I can access the CI environment for quick iterations?

leezu:
No need to access the CI. Just run `python ci/build.py --platform jetson /work/runtime_functions.sh build_jetson`

leezu:
You may need to update the Dockerfile.jetson first as the cross-toolchain debian packages are currently not public yet.. So you'll need to download them with SDK manager or ask the Jetson team if they can provide you access to their public repo (which apparently is in testing currently)

TristonC:
@leezu Could you help on this error?
`[2020-06-01T22:20:47.379Z] [1/1] cd /work/mxnet/3rdparty/onnx-tensorrt/third_party/onnx/build && python /work/mxnet/3rdparty/onnx-tensorrt/third_party/onnx/onnx/gen_proto.py -p onnx -o /work/mxnet/3rdparty/onnx-tensorrt/third_party/onnx/build/onnx onnx

[2020-06-01T22:20:47.379Z] FAILED: onnx/onnx.proto 

[2020-06-01T22:20:47.379Z] cd /work/mxnet/3rdparty/onnx-tensorrt/third_party/onnx/build && python /work/mxnet/3rdparty/onnx-tensorrt/third_party/onnx/onnx/gen_proto.py -p onnx -o /work/mxnet/3rdparty/onnx-tensorrt/third_party/onnx/build/onnx onnx

[2020-06-01T22:20:47.379Z] /bin/sh: 1: python: not found
`

leezu:
@TristonC yes, that sounds correct. But this may need to be updated in the upstream repository?

TristonC:
> s may need to be updated i
I put it in the runtime_functions.sh to see if it works for now.

TristonC:
> 
> 
> @TristonC yes, that sounds correct. But this may need to be updated in the upstream repository? Or where is PYVER referenced?

I put the PYVER in the runtime_functions.sh to make it have minimum impact for other part for now.

ci/docker/runtime_functions.sh
```diff
@@ -187,13 +187,50 @@ build_dynamic_libmxnet() {
...
+    set -ex
+    # Build ONNX
+    pushd .
+    echo "Installing ONNX."
+    cd 3rdparty/onnx-tensorrt/third_party/onnx
+    rm -rf build
+    mkdir -p build
+    cd build
+    cmake \
+        -DCMAKE_CXX_FLAGS=-I/usr/include/python${PYVER}\
```
leezu, line 200: Suggest you delete this line instead of adding PYVER. Then it should be autodetected

TristonC: Good idea. Will do that.

TristonC: No luck for either way. @leezu How does ninja pick a  python version?

TristonC: @leezu  Is that OK I do alias python in the Dockerfile.build.jetson file? I saw only python3 was installed.

leezu: I think the problem is that we're using a 2 years old version of onnx via the onnx tensorrt. That version seems to require Python 2. I suggest you update the submodule tensorrt onnx submodule to the latest release which hopefully supports Python 3.

This should be unrelated to ninja and above CMAKE_CXX_FLAGS

I would suggest not to introduce such symlink, as the upstream scripts may fail in subtle ways when switching from python2 to python3

TristonC: Make sense. I just did a commit of alias to give it a try. I will revoke that later.

ci/docker/Dockerfile.build.jetson
```diff
@@ -42,7 +42,8 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
     automake \
     libtool \
     crossbuild-essential-arm64 \
- && rm -rf /var/lib/apt/lists/*
+ && rm -rf /var/lib/apt/lists/* \
+ && alias python=python3
```
leezu, line 46: This won't help as the alias will be lost at the end of this RUN command. If you don't plan to fix the onnx update in this PR, you can manually install python2 https://packages.ubuntu.com/focal/python2

TristonC: Yep. Let me revoke the previous changes related to python and then add the python2.

leezu:
@TristonC would you resolve the conflict or could you comment on the missing pieces blocking this PR? Thank you

PR #18452, opened 2020-06-01
[v1.7.x] Fix CI error

ciyongch:
## Description ##
Fix the following CI error:
```
[2020-05-28T17:49:56.861Z] + ln -s /usr/bin/ninja-build /usr/bin/ninja

[2020-05-28T17:49:56.861Z] ln: failed to create symbolic link '/usr/bin/ninja': File exists

[2020-05-28T17:49:57.421Z] The command '/bin/sh -c /work/centos7_core.sh' returned a non-zero code: 1
```
@TaoLv @leezu @pengzhao-intel @ChaiBapchya

### DIFF
--- a/ci/docker/install/arm64_openblas.sh
+++ b/ci/docker/install/arm64_openblas.sh
@@ -29,7 +29,7 @@ git clone https://github.com/xianyi/OpenBLAS.git
 cd OpenBLAS
 make -j$(nproc) TARGET=ARMV8
 make install
-ln -s /opt/OpenBLAS/lib/libopenblas.so /usr/lib/libopenblas.so
-ln -s /opt/OpenBLAS/lib/libopenblas.a /usr/lib/libopenblas.a
-ln -s /opt/OpenBLAS/lib/libopenblas.a /usr/lib/liblapack.a
+ln -sf /opt/OpenBLAS/lib/libopenblas.so /usr/lib/libopenblas.so
+ln -sf /opt/OpenBLAS/lib/libopenblas.a /usr/lib/libopenblas.a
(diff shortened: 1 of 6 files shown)


nickguletskii:
Hi @ciyongch, it seems that the reason for the CI failures in the v1.7.x branch are the same as in v1.x: the ninja package in CentOS now ships with the required symlink, which makes it unnecessary to create it manually. I fixed this issue in v1.x by removing the build step that creates the symlink: https://github.com/apache/incubator-mxnet/pull/18437

ciyongch:
@nickguletskii Thanks for point it out for the fix:), would you mind to backport that PR into v1.7.x branch as well?
Regarding the rest of symlink change in this PR, I suggest to use `ln -sf` instead of `ln -s`, what do you think?

nickguletskii:
> @nickguletskii Thanks for point it out for the fix:), would you mind to backport that PR into v1.7.x branch as well?
> Regarding the rest of symlink change in this PR, I suggest to use `ln -sf` instead of `ln -s`, what do you think?

I've created a backport here: #18456 .
Concerning the other symlink changes: let's see if the CI passes without the ninja link first. Personally, I don't think overwriting symlinks managed by the system's package manager is a very good idea, even in containers, so unless the CI jobs fail without `ln -sf`, I don't think these changes are necessary.

ciyongch:
Thanks for your help and comments @nickguletskii , let's just keep the rest as is if the CI passed. My initial thoughts was to handle all the potential of this kind of issue, as `ln -sf` is used widely in the current build script.

leezu:
#18456

PR #18431, opened 2020-05-28
[website] Redirect Chinese visitors to Apache Chinese CDN provider PART 1

ys2843:
## Description ##
This PR is the part1 of #18422 .To fix slow site speed in China and improve overall loading speed. It adds server side logic to redirect Chinese visitors to Chinese CDN based on Geo IP. This is the most accurate and fastest way to filter visitors.
 
There are 2 checks to each incoming HTTP requests. 
1. Check user Geoip country code, whether it equals to `CN` 
2. Check host name in URL to ensure redirect only happens when user initially attempts to access the website. Avoid any loops and subsequent redirect.

If all conditions match, redirect user to Chinese CDN with same URI. For example, redirect from `https://mxnet.apache.org/features` to `https://mxnet.cdn.apache.org/features`

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:

### Changes ###
- [x] Add logic to redirect Chinese visitors to Apache's Chinese CDN provider

## Comments ##
- Preview: http://ec2-34-219-134-42.us-west-2.compute.amazonaws.com
- Please note: Geo_ip is set up on this preview server. If you visit `http://ec2-34-219-134-42.us-west-2.compute.amazonaws.com/testgeoip.php`, you should see your ip and geolocation data.
Following settings are used to stimulate the process:
```shell
RewriteCond %{ENV:GEOIP_COUNTRY_CODE} ^CN$
RewriteCond %{HTTP_HOST} ec2-34-219-134-42.us-west-2.compute.amazonaws.com
RewriteRule ^(.*) https://mxnet.cdn.apache.org%{REQUEST_URI} [R,L]
```
- Tested by several users in China, redirect works properly and there is a boost of loading speed

PR #18457, opened 2020-06-01
redirect visitors based on geolocation - update host condition

ys2843:
## Description ##
This is a fix for #18431 . The rewrite rule is not redirecting users because of the tight host filter rule. Update the rewrite condition to not only match `mxnet.apache.org`, but reverse it to match all hostname except `mxnet.cdn.apache.org`.

According to Apache Intra team's [response](https://issues.apache.org/jira/browse/INFRA-20203#comment-17121024),
> Try removing the following condition:
RewriteCond %{HTTP_HOST} mxnet.apache.org
Most of your visits go to mxnet.incubator.apache.org

Because we don't know how Apache's server is configured, remove the whole condition could possibly import looping. To safely fix this problem we can first test the updated rewrite rule.

### DIFF
--- a/docs/static_site/src/.htaccess
+++ b/docs/static_site/src/.htaccess
@@ -5,7 +5,7 @@ RewriteOptions AllowNoSlash
 
 # Redirect Chinese visitors to Chinese CDN, temporary solution for slow site speed in China
 RewriteCond %{ENV:GEOIP_COUNTRY_CODE} ^CN$
-RewriteCond %{HTTP_HOST} mxnet.apache.org
+RewriteCond %{HTTP_HOST} !cdn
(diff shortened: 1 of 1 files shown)


ys2843:
> how are you testing it

By using any speedtest website with Chinese server, we can see redirect in the result. Try anyone of the following:
+ https://www.boce.com/ (Chinese speed test website, simply copy url to test) 
+ https://www.webpagetest.org/ (Choose Beijing as location then test)

I created this host to test the redirect rules, it uses the same rules as this PR
+ `http://ec2-34-219-134-42.us-west-2.compute.amazonaws.com/`
```
RewriteCond %{ENV:GEOIP_COUNTRY_CODE} ^CN$
RewriteCond %{HTTP_HOST} !cdn
RewriteRule ^(.*) https://mxnet.cdn.apache.org%{REQUEST_URI} [R,NC,L]
```

PR #18459, opened 2020-06-01
Enable ccache on Github Actions

leezu:
Speed up the OS X tests.

### DIFF
--- a/.github/workflows/os_x_staticbuild.yml
+++ b/.github/workflows/os_x_staticbuild.yml
@@ -8,18 +8,40 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+
+      - name: Compilation cache
+        uses: actions/cache@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/os_x_staticbuild.yml
```diff
@@ -8,18 +8,40 @@ jobs:
...
       - name: Checkout repository
         uses: actions/checkout@v2
+
+      - name: Compilation cache
+        uses: actions/cache@v2
+        with:
+          path: ~/.ccache
+          # We include the commit sha in the cache key, as new cache entries are
+          # only created if there is no existing entry for the key yet.
+          key: ${{ runner.os }}-ccache-${{ github.sha }}
```
szha, line 18: any way to share the cache across commits?

leezu: Yes, via the restore-keys below. It's based on common prefix

PR #18485, opened 2020-06-04
Update Jetson installation guide

mseth10:
## Description ##
Updating Jetson installation guide on MXNet website. Adding config Makefile for Jetson build.

Also removing link to non-ASF wheels as per http://www.apache.org/legal/release-policy.html#what

This PR does not add/modify cmake cross compilation steps to the doc, which will be done in a future PR.


leezu:
@mseth10 I suggest to backport this change to 1.7 / 1.x to ensure the makefile config ships in an actually released version. Currently the plan for the master branch is to delete all makefile build related files prior to the release.

PR #18487, opened 2020-06-05
Consolidate installation instructions on website and add disclaimer for non-ASF ressources

leezu:
## Description ##
Consolidate and update installation instructions for easier maintenance and correctness

Changes
- Merge compile from source instructions for various operating systems into a single document. Previously different instructions went out of sync and were not well maintained.
- Don't link to unreleased versions per ASF policy
- Add disclaimers to names and links pointing to third party ressources per https://lists.apache.org/thread.html/r118075902f9eee1c0f5e9485cd6633ccbcfb6b060c68aebe93cc8de6%40%3Cdev.mxnet.apache.org%3E


ys2843:
Preview link: http://ec2-18-236-134-13.us-west-2.compute.amazonaws.com/

justinmclean:
Please do not link directly to https://downloads.apache.org/incubator/mxnet/ See https://infra.apache.org/release-download-pages for how a download page should follow.

leezu:
Thanks @justinmclean for clarifying. I think https://mxnet.apache.org/get_started/download implements the policy, so I'll replace the links accordingly

justinmclean:
On that page the only hashes that should be accepted  are sha256 and sha512 see https://infra.apache.org/release-distribution.html#sigs-and-sums. Also download links need to go via https://downloads.apache.org/ or https://archive.apache.org/dist see https://infra.apache.org/release-distribution.html#download-links

leezu:
@justinmclean 

> All links to checksums, detached signatures and public keys must reference downloads.apache.org/ using https:// (TLS).

and 

> Old releases should be archived and MAY be linked to from public download pages.

This policy implies that we must not link checksums and detached signatures for old releases as they are no longer available from downloads.apache.org. That seems unintended. What part of the policy am I missing here?

justinmclean:
You would use archive.apache.org for old signatures. The reasoning here is you want to link directly to the signatures and not use the mirror system so you have more confidence they are correct.
You generally don't want multiple versions on downloads.apache.org, only the latest, and this is replicated across all mirrors.

leezu:
@justinmclean thanks. Yes, that's a reasonable setup. My question is if the policy should be updated to reflect what you say? Currently using archive.apache.org for old release signatures is non-compliant according to the policy. That seems to be a mistake in the policy

justinmclean:
There is no contradiction. If it archived its not on the mirror. If you want the wording changed I suggest you speak to Infra as it's their policy.

leezu:
Ok. I ask INFRA to fix their policies at https://issues.apache.org/jira/browse/INFRA-20390
If we wish to take policies seriously, we must ensure that they are accurate.

marcoabreu:
Are you purposefully getting rid of CPU packages? I thought that those were compliant.

leezu:
@marcoabreu which cpu packages are you referring to in particular? In general all CPU packages created using the `tools/staticbuild` script contain GPL code (redistribute libgfortran.so). We can fix that and have first-party convenience binaries.

marcoabreu:
You removed all CPU packages and only mentioned the source packages. 

But yeah, if they contain gpl code, let's get rid of that and make compliant packages asap

szha:
The notice needs to be backported to previous versions too as people who select older versions may see installation guide that are different from the one on master.

cc @ys2843 @sandeep-krishnamurthy

ys2843:
> The notice needs to be backported to previous versions too as people who select older versions may see installation guide that are different from the one on master.
> 
> cc @ys2843 @sandeep-krishnamurthy

If the master install guide is the correct one, I suggest we redirect previous version installation page all to master. Because previous version websites are static artifact, not in a good shape for making these many changes.
And in fact, I believe it is easier to maintain if the current versioned website could be replaced by one single master website with versioned docs & tutorials.

PR #18473, opened 2020-06-02
Dynamically Generate Version Dropdown

ys2843:
## Description ##
To make several website version dropdowns more manageable and easier to update, this PR includes the change to dynamically generate all version dropdown using the `versions` constant in `_config.yml`. This matches the implementation in global search https://github.com/apache/incubator-mxnet/pull/18288#issuecomment-628990327.
**Updated version dropdown in this PR:**
+ Installation guide dropdown
+ General version dropdown (on top nav)

**Existing dynamically generate version dropdown:**
+ 2 global search version dropdown(desktop & mobile)

### Essentials ###
Please feel free to remove inapplicable items for your PR.
- [x] The PR title starts with [MXNET-$JIRA_ID], where $JIRA_ID refers to the relevant [JIRA issue](https://issues.apache.org/jira/projects/MXNET/issues) created (except PRs with tiny changes)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage:

### Changes ###
- [x] Update the general version dropdown on top navigation bar. 
- [x] Update the installation guide version dropdown  

## Comments ##
- Preview: http://ec2-34-219-134-42.us-west-2.compute.amazonaws.com/
- cc @connorgoggins

PR #18488, opened 2020-06-05
revert installation dropdown change

ys2843:
## Description ##
Revert the changes to `get_start` version dropdown in #18473 which caused the dropdown content not responding to version select, will submit another follow up PR with better solution

The reason of the bug is the version dropdown on get start page must use current format, for example "v1.5.1", "v" and "." is mandatory or it will break the parsing rule in `option.js`. But for general version dropdown and global search dropdown, the version format is slightly different from it.

### Changes ###
- [x] Revert changes to get start page version dropdown

## Comments ##
- preview: http://ec2-18-236-134-13.us-west-2.compute.amazonaws.com/get_started?


ys2843:

Issue #18422, opened 2020-05-27
MXNet Website 2.0 - Slow Site Speed in China


ys2843:
## Summary

Due to lack of available CDN in China from Apache, MXNet website is hosted in Japan for Chinese visitors [1]. According to the Q1 2020 Google Analytics report, it took 20 second on average to load the website in such areas and the traffic from China declined by 16% compared to previous quarter. An alternative CDN solution is discussed in this proposal.

## Introduction

MXNet website is the one-stop site for users from all over the world to explore MXNet project. As a part of Apache Software Foundation, MXNet official website is hosted by Apache on its content delivery networks. Because MXNet dev community has no access to Apache’s web server, Google Analytics is used as the tool to track and report website traffic. Based on the statistics from Google Analytics, we are able to better understand our users and improve MXNet website. 

## Problems

According to the statistic (Q1 2020) from Google Analytics, China has the most users using MXNet website which account for 22.95% of the total number of users. But the website load time is very slow in aforementioned area, the average page load time is 20.38s compare to US it is 2.12s. 
The cause of this problem [1] is, according to Apache Infra team, MXNet website is served from Japan for Chinese visitors. Due to the Great Firewall of China [2], it can cause long delay on visiting oversea websites. But at present Apache can not get access to serve from China itself by reason of Chinese government regulations. 
This can cause frustration on our customers in China, and in Q1 2020 there is a 15.69% decline in traffic (measured by sessions) from China compared to previous quarter. 

## Proposed Solution

Based on what Apache Infra team can offer now, this is a compromise solution that can address this issue. A Chinese CDN which is pointed by a different but similar domain name ‘_mxnet.cdn.apache.org_’ is set up by Apache Infra team. It is a Chinese CDN provider and there is a huge improvement on website loading speed in China.
To achieve optimal results, this project is scoped into two parts, both of which aim to increase the usability of MXNet  website and provide better user experience.

**Part 1. Redirect Chinese Visitors**
In .htaccess file, add rewrite rule for Chinese visitors with the help of Apache mod_Geoip module- This is the most accurate and fastest way to redirect.
05/28 - According to Apache Intra team [1], Mod_Geoip is enabled on their server. 

**Part 2. Improve Overall Website Loading Speed [2]**
Eliminate render-blocking resources, defer or lazy-load resources that is unrelated to first render. Update cache policy for static assets

## Reference

1. Slow site speed issue Jira ticket: https://issues.apache.org/jira/browse/INFRA-20203
[...]

Issue #18367, opened 2020-05-19
Google JQuery CDN [https://ajax.googleapis.com] is banned in China


ys2843:
## Description
In MXNet website, JQuery script is loaded from google CDN in `<head>`
```html
<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.3.1/jquery.min.js"></script>
```
But this CDN is known banned in China, as you can see here:
+ [Similar Github issues on other website](https://github.com/kubeflow/website/issues/81)
+ [Things to Consider When Building a Website in China](https://www.labsterx.com/blog/things-to-consider-when-building-a-website-in-china-9-shtml/)
+ [Can webflow team replace “ajax.googleapis.com” with a reachable URL for China?](https://forum.webflow.com/t/can-webflow-team-replace-ajax-googleapis-com-with-a-reachable-url-for-china/29284)

Here is my own test using Chinese servers:

<img width="760" alt="Screen Shot 2020-05-19 at 12 11 59 PM" src="https://user-images.githubusercontent.com/23468760/82372090-30d90e00-99d0-11ea-8262-00f6590d484d.png">

It not only blocks website from loading, but also breaks all the interaction on MXNet website. (e.g. dropdown, installation guide, search etc. )

### Error Message
`ReferenceError: jQuery is not defined[Learn More]`

## Proposed solutions 
While China is the number one country using the MXNet website. We need to solve this problem so that users in China can visit MXNet website without any problem. Here are 3 solutions:
+ Use an alternative CDN as a backup, for example [CDN hosted in China](https://www.bootcdn.cn/) , when google CDN fails then load the resource from backup CDN , as discussed [in this question](https://stackoverflow.com/questions/1014203/best-way-to-use-googles-hosted-jquery-but-fall-back-to-my-hosted-library-on-go)
+ Replace google CDN by another CDN
+ Hosting jQuery locally (**Not preferable** performance-wise, reasons can be found [here](https://meta.stackoverflow.com/questions/258288/use-of-ajax-googleapis-com-not-china-friendly) )

ys2843:
@mxnet-label-bot add [Website]

szha:
jquery has its own CDN which isn't affected by the blocking of google CDN. so the simplest approach to try is to use that https://code.jquery.com/

could you summarize how the current website hosting works? to have more control on this we could use cloudfront, which can also improve stability.

sandeep-krishnamurthy:
[...]

PR #18340, opened 2020-05-16
[Website 2.0] Artifact URL Adjustment

connorgoggins:
## Description ##
This PR moves static artifacts to MXNet website artifacts public bucket.

### Essentials ###
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] Code is well-documented
- [x] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change

### Changes ###
- M docs/static_site/Makefile

## Comments ##
These changes have been tested with a complete website build on Jenkins dev and local hosting. All functionality remains the same.

@sandeep-krishnamurthy @aaronmarkham @ys2843

PR #18512, opened 2020-06-08
Fix Slow Site Loading Speed part2

ys2843:
## Description ##
This PR addressed the rest of the problems listed in #18422. With the fix in part 1, it solved the root of the problem for Chinese visitors using geolocation based redirect. The change in this PR is more of an optimization on overall loading speed.  

### Essentials ###

### Changes ###
- [x] Host `JQuery` script locally (as discussed in #18367, more reliable and faster speed)
- [x] Add cache control (recommended by Apache Infra team to prevent stale contents and better loading speed, see [jira ticket](https://issues.apache.org/jira/browse/INFRA-20203?focusedCommentId=17120675&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17120675))
- [x] Eliminate render blocking resources: `defer` scripts that are not contributing to the first paint
- [x] Move General Version Dropdown CSS from `<head>` to main css. Discussed with @connorgoggins 

## Comments ##
- Preview: http://ec2-34-219-134-42.us-west-2.compute.amazonaws.com/


ys2843:
@mxnet-label-bot add [website]

docs/static_site/src/.htaccess
```diff
@@ -3,6 +3,25 @@ DirectorySlash off
...
+<IfModule mod_expires.c>
+  ExpiresActive on
+
+  # Images
+  ExpiresByType image/png                 "access plus 7 days"
+  ExpiresByType image/jpg                 "access plus 7 days"
+  ExpiresByType image/jpeg                "access plus 7 days"
+  ExpiresByType image/svg+xml             "access plus 7 days"
+
+  # CSS, Javascript, HTML
```
connorgoggins, line 15: Just curious, why are we caching JS & CSS but not HTML?

docs/static_site/src/_sass/generalVersionDropdown.scss
```diff
@@ -0,0 +1,41 @@
+.dropdown-option-active {
```
connorgoggins, line 1: Nice! Definitely cleaner to have this in its own file, excellent work :)

docs/static_site/src/assets/js/jquery-3.3.1.min.js
```diff
@@ -0,0 +1,20 @@
+/*!
```
connorgoggins, line 1: Is keeping this in GitHub the best course of action? Couldn't we store this in an S3 bucket and pull it during build time via Jenkins?

docs/static_site/src/.htaccess
ys2843: Good question, it basically just preserves the current HTML "cache control" (no cache and verify diff by ETag) and extends CSS/JS cache control to 1 day. By not caching HTML, new features/contents can be published without delay.

docs/static_site/src/assets/js/jquery-3.3.1.min.js
ys2843: Good idea, do we have existing S3 for this purpose?

ys2843:
@mxnet-bot run [unix-cpu]

docs/static_site/src/assets/js/jquery-3.3.1.min.js
connorgoggins: Could use the `mxnet-website-static-artifacts` bucket for now, or create a new bucket.

ys2843: Thought this through again and found it could import potential problems by storing it on S3. It will require pulling the script when not only build on Jenkins, but also develop locally. If someone has trouble or high latency accessing our S3, the building process could break.
Found a more general approach would be using node package manager. It helps a lot when managing many dependencies, however we only have this one dependency now, it would import more complexity than benefit.
I would say for now still keep it within the project, ([other projects](https://github.com/apache/incubator-echarts-website/blob/asf-site/en/vendors/jquery/jquery.js) also do), and keep NPM as a candidate, switch to it when we need to handle more dependencies :)

leezu: How about linking to the jquery cdn? https://code.jquery.com/

ys2843: This is the current way discussed in #18367, was to replace google cdn when hosting locally was not a solution. After Apache enabling Chinese CDN, the server is pretty fast and stable. So switch to host the scripts locally.

connorgoggins: Regardless of where the website developer is located geographically, they will always be using Jenkins to build the website - so local latency to S3 isn’t a concern. Since we’re not anticipating changing this file, it seems unnecessary to include it in version control systems and not in S3.

connorgoggins: Agreed that NPM would introduce more complexity than is necessary for including this single file, but pulling the file from S3 at build time would be as simple as adding a `wget` call to the Makefile for building the static site - see [here](https://github.com/apache/incubator-mxnet/pull/18340/files#diff-8938828e17dfd3968ac4048a65029ecfR23) for an example.

ys2843: Agreed on the point of building on Jenkins. However, the problem (local latency) I was concerning is that when develop the main site locally (run `jekyll serve` and website host on 127.0.0.1), it is also the time that the script need to be pulled from S3, and geographical location might affect the accessibility to this script.
If we really want to get rid of JQuery on Git, a way I can think of is that on one hand using S3/CDN for building on Jenkins, and for local dev, updating the doc/makefile to add the wget JQuery command before `jekyll serve`. wdyt?

PR #18513, opened 2020-06-08
Switch staticbuild scripts to Flang Fortran compiler

leezu:
Flang fortran compiler as replacement for gfortran due to the resulting runtime
dependency on libgfortran.so which is GPL-licensed and thus
https://www.apache.org/legal/resolved.html#category-x for distribution.

We need a Fortran compiler to compile the OpenBLAS Lapack implementation.

Note that Flang will be replaced by F18 as part of a future LLVM version. But currently F18 is not yet ready.


ciyongch:
Hi @leeze, as it's tagged with "R1.7.0"/"Backport 1.x" label, so it's a necessary fix for 1.7.0 and will be backported to v1.7.x later, right?

leezu:
It's only for the distribution scripts, so it can be backported to the 1.7 branch at a later point (not a release blocker)

ciyongch:
> It's only for the distribution scripts, so it can be backported to the 1.7 branch at a later point (not a release blocker)

Got it, thanks!

PR #18528, opened 2020-06-09
Update cmake/upstream/FindCUDAToolkit.cmake

leezu:
Previously MXNet includes a hotfix for a cross-compiling bug in upstream FindCUDAToolkit.cmake. Upstream has fixed the bug now. Thus remove MXNet patch and replace by upstream fix

cf https://gitlab.kitware.com/cmake/cmake/-/issues/20572


leezu:
@mxnet-bot run ci [unix-gpu]

PR #18531, opened 2020-06-10
Remove XXOutput loss operators 

eric-haibin-lin:
## Description ##
These operators are often used with mx.module APIs. Removing them for mxnet 2.0. You can find equivalent loss functions in mx.gluon.loss namespace. 

These APIs will still be maintained in the 1.x branch
https://github.com/apache/incubator-mxnet/issues/17676

### Essentials ###

### Changes ###

## Comments ##


fhieber:
Will there be an efficient equivalent for SoftmaxOutput in 2.0 that also supports label smoothing?

sxjscience:
@eric-haibin-lin @fhieber I’ve written a new loss layer in the numpy API: https://github.com/dmlc/gluon-nlp/blob/b714eaccc67619d7bdcbd1574d30be87d9c73f0c/src/gluonnlp/loss.py#L4

I’ve evaluated and the speed + memory cost is better than calculating the cross-entropy between softmax and smoothed label.

fhieber:
@eric-haibin-lin Thanks for the pointer. We had tried custom operators in a separate project (int8 quantization) but found them to be not suitable for performance reasons.
@sxjscience Thanks! I will take a look. I haven't benchmarked SoftmaxOutput and a Gluon implementation recently. Mid 2019 we still observed a ~15-20% slowdown when not using SoftmaxOutput. I will give your label smoothing implementation a try, but my guess is that the combination of Softmax forward pass, (smoothed) cross-entropy gradient backward pass, and support for easy masking without creating an explicit mask, all implemented in a single operator, will still have a slight edge over an implementation with multiple operators.

marcoabreu:
CC @sandeep-krishnamurthy 

Are custom operators actually causing a slow-down? Speak, the same code, one time directly compiled and the other one dynamically loaded are not creating the same performance?

eric-haibin-lin:
@samskalicky is there any evidence that dynamic ops are slower?

samskalicky:
No, dynamic C++ custom ops are significantly faster than Python custom ops. I believe @fhieber and his team found dynamic C++ custom ops to be within 10% of built in operators. There are some overheads that are a result of our design choices (we chose easier implementation rather than speed). But our goal was never equal performance as built in operators.

fhieber:
Just to follow up on this and provide some numbers. I tried the label smoothing implementation by @sxjscience (thanks!) in Sockeye, making sure scaling and normalization is equivalent to how we currently use SoftmaxOutput. 
While this is much faster than a naive implementation of label smoothing with dense, smoothed labels, SoftmaxOutput is still 5-10% faster than the optimized implementation.

PR #18532, opened 2020-06-10
add epsilon to adamax

szhengac:
as in title.


szhengac:
@mxnet-bot run ci [unix-cpu]

szhengac:
@mxnet-bot run ci [unix-cpu]

Issue #17823, opened 2020-03-12
[Operator] Add `index_add` or `index_update` to numpy extension


sxjscience:
We need the functionality to calculate `b = index_add(a, indices, value)`, which mimics the outcome of `a[indices] += value`.

This is similar to the `tensor_scatter_nd_add` in TF: https://www.tensorflow.org/api_docs/python/tf/tensor_scatter_nd_add

Also in JAX: https://jax.readthedocs.io/en/latest/_autosummary/jax.ops.index_update.html#jax.ops.index_update

ZheyuYe:
Besides that, it would be great if `npx.scatter_nd` is implemented for `mx.np.ndarray`.

sxjscience:
@ZheyuYe @JiangZhaoh @yzhliu @haojin2 

To understand the problem, let's consider two use cases. The first one can be solved via `gather_nd` and the second one cannot be solved via the existing MXNet.

## Take elements at specific locations from the input data

`out[i, j, ...] = data[i, positions[i, j], ...]`

In GluonNLP, the `positions` are masked locations in the input that we will need to calculate the loss. `data` is the mapped hidden states of the sequences.

With advanced indexing + imperative API, we can do something like this:

```python
import mxnet as mx
mx.npx.set_np()

data = mx.np.random.normal(0, 1, (5, 5, 5, 5))
positions = mx.np.random.randint(0, 5, (5, 4))
out = data[mx.np.expand_dims(mx.npx.arange_like(data, axis=0), axis=-1), positions]
print(out.asnumpy().shape)
```

In order to make the network hybridizable, we can implement it via `gather_nd`:

```python
@use_np
def select_vectors_by_position(F, data, positions):
    """Select each batch with the given positions.
    Once advanced indexing can be hybridized, we can revise the implementation.
    out[i, j, :] = data[i, positions[i, j], :]
    Parameters
    ----------
    F
    data
        Input tensor of contextualized token embeddings
        Shape (batch_size, seq_length, units)
    positions
        Input tensor of the positions.
        Shape (batch_size, num_sel_positions).
        For each sample in the batch, the values in this tensor must not exceed
        the length of the sequence.
    Returns
    -------
    out
        The selection result.
        Shape (batch_size, num_sel_positions, units)
    """
    # Here, we use gather_nd to select the output from data:
    # Need to compute
    #   out[i, j, :] = in[i, masked_position[i, j], :]
[...]

PR #18545, opened 2020-06-12
add op npx.index_update

JiangZhaoh:
Add npx op index_update.
Relate to #17823 .

### DIFF
--- a/python/mxnet/_numpy_op_doc.py
+++ b/python/mxnet/_numpy_op_doc.py
@@ -699,6 +699,71 @@ def _npx_index_add(a, ind, val):
     pass
 
 
+def _npx_index_update(a, ind, val):
+    """
+    Update values to input according to given indexes.
+    If multiple indices refer to the same location it is undefined which update is chosen; it may choose
+    the order of updates arbitrarily and nondeterministically (e.g., due to concurrent updates on some
+    hardware platforms). Recommend not to use repeate positions.
+    ----------
+    a : ndarray
+        Input data. The array to be updated.
+        Support dtype: 'float32', 'float64', 'int32', 'int64'.
+    ind : ndarray
+        Indexes for indicating update positions.
+        For example, array([[0, 1], [2, 3], [4, 5]] indicates here are two positions to
+        be updated, which is (0, 2, 4) and (1, 3, 5).
+        Note: - 'ind' cannot be empty array '[]', for that case, please use operator 'add' instead.
+              - 0 <= ind.ndim <= 2.
+              - ind.dtype should be 'int32' or 'int64'
+    val : ndarray
+        Input data. The array to update the input 'a'.
+        Support dtype: 'float32', 'float64', 'int32', 'int64'.
+    Returns
+    -------
+    out : ndarray
+        The output array.
+    Examples
+    --------
+    >>> a = np.zeros((2, 3, 4))
+    >>> ind = np.array([[0, 0], [0, 0], [0, 1]], dtype='int32')
+    >>> val = np.arange(2).reshape(2) + 1
+    >>> b = npx.index_update(a, ind, val)
+    >>> b
+    array([[[1., 2., 0., 0.],
+            [0., 0., 0., 0.],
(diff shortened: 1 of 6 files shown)


JiangZhaoh:
@mxnet-bot run ci [unix-gpu, unix-cpu, windows-cpu]

python/mxnet/_numpy_op_doc.py
```diff
@@ -699,6 +699,71 @@ def _npx_index_add(a, ind, val):
     pass
 
 
+def _npx_index_update(a, ind, val):
+    """
+    Update values to input according to given indexes.
+    If multiple indices refer to the same location it is undefined which update is chosen; it may choose
+    the order of updates arbitrarily and nondeterministically (e.g., due to concurrent updates on some
+    hardware platforms). Recommend not to use repeate positions.
```
sxjscience, line 707: You may add a blank line here. (I find that `index_add` misses the blank line also).

src/operator/tensor/index_update.cu
```diff
@@ -0,0 +1,202 @@
...
+                                     const TBlob& ind,
+                                     const mshadow::Shape<MXNET_SPECIAL_MAX_NDIM> grada_pre_stride,
+                                     const int tail_size, const int ind_num, const int ind_ndim,
+                                     const int seg, const int req) {
+  using namespace mxnet_op;
+  using namespace mshadow;
+  mshadow::Stream<gpu> *s = ctx.get_stream<gpu>();
+  MSHADOW_TYPE_SWITCH(grad_a.type_flag_, DType, {
+    size_t temp_mem_size = ind.shape_.Size() * sizeof(int) +
+                           ograd.shape_.Size() * sizeof(DType);
```
sxjscience, line 174: Here, we mix `int` type and `DType`. The second pointer may not be appropriately aligned with `DType`. You may consider to revise the code and add the `PadBytes`: https://github.com/apache/incubator-mxnet/blob/1b02225fefd8ccc93bc73223f0d3cde103fad661/src/operator/tensor/ordering_op-inl.h#L350-L351

The reason to do this is that CUDA forces memory alignment, i.e., we will always have the following assertion:

```c++
DType* ptr;
ASSERT static_cast<size_t>(ptr) % sizeof(DType) == 0.
```

By default, calling cudaMalloc will give you a pointer with at least 256-bytes aligned address. However, after shifting the pointer to `ptr += ind.shape_.Size() * sizeof(int)`, the new address may not be aligned with `sizeof(DType)`.

tests/python/unittest/test_numpy_op.py
```diff
@@ -1477,6 +1477,165 @@ def index_add_bwd(out_grad, a_grad, ind, val_grad, ind_ndim, ind_num, grad_req_a
...
+            configs.append(tuple([shape, ind, tuple(val_shape), ind_ndim, ind_num]))
+
+    dtypes = ['float32', 'float64', 'int32', 'int64']
+    grad_req = ['write', 'null', 'add']
+    for hybridize, grad_req_a, grad_req_val, dtype, indtype in \
+        itertools.product([True, False], grad_req, grad_req, dtypes, ['int32', 'int64']):
+        for a_shape, ind, val_shape ,ind_ndim, ind_num in configs:
+            eps = 1e-3
+            if sys.platform.startswith('linux'):
+                eps = 1e-2
```
leezu, line 1589: eps should not depend on the platform. What's the root-cause here?

src/operator/tensor/index_update.cu
JiangZhaoh: OK, I have revised this segment in [https://github.com/apache/incubator-mxnet/pull/18545/commits/c251837d2703813a254fd75cc426d50139220272](url)

JiangZhaoh: OK, I have revised the code in commit [https://github.com/apache/incubator-mxnet/pull/18545/commits/c251837d2703813a254fd75cc426d50139220272](url)

JiangZhaoh:
@mxnet-bot run ci [unix-cpu]

PR #18555, opened 2020-06-13
Remove check for subgraph with cycles

Kh4L:
## Description ##
Partitioning a graph with potential subgraphs that have a cycle with the input or output fails instead of ignoring these subgraphs. For instance, Faster-RCNN (available in the GluonCV model zoo) optimization failed. 

## Change ##

This PR simply removes the check that raise an exception if the subgraph candidate has a cycle, so LabelSubgraph can return false.

### DIFF
--- a/src/operator/subgraph/build_subgraph.cc
+++ b/src/operator/subgraph/build_subgraph.cc
@@ -308,8 +308,11 @@ void PreSelectSubgraphNodes(const nnvm::Graph& g, SubgraphSelectorV2Ptr subgraph
     success = LabelSubgraph(g, subgraph_selector, label, snid, simple_nodes, subgraph_nodes,
                             &excluded_nodes);
     if (!success) {
+      // Failed to label subgraph due to a cycle
+      // If the number of excluded_nodes didn't change since the last iteration,
+      // this means that there is no possible subgraph for the current node snid, we break
+      // Otherwise, we keep trying (with the excluded nodes tagged)
       if (excluded_nodes.size() == n_excluded_nodes) {
-        // no possible subgraph for the current node snid
         break;
       }
       n_excluded_nodes = excluded_nodes.size();
--- a/src/operator/subgraph/build_subgraph.cc
+++ b/src/operator/subgraph/build_subgraph.cc
@@ -226,9 +226,7 @@ bool LabelSubgraph(const nnvm::Graph& g, SubgraphSelectorV2Ptr subgraph_selector
     std::stack<const nnvm::Node*> s;
     s.push(descendant);
     size_t count = 0;
-    while (!s.empty()) {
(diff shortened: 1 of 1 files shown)


src/operator/subgraph/build_subgraph.cc
```diff
@@ -306,14 +300,19 @@ void PreSelectSubgraphNodes(const nnvm::Graph& g, SubgraphSelectorV2Ptr subgraph
...
+  size_t n_excluded_nodes = 0;
   const size_t max_num_retry = simple_nodes.size() * simple_nodes.size();
   size_t count = 0;
   bool success = false;
   while (!success && count < max_num_retry) {
     success = LabelSubgraph(g, subgraph_selector, label, snid, simple_nodes, subgraph_nodes,
                             &excluded_nodes);
     if (!success) {
-      CHECK(!excluded_nodes.empty());
+      if (excluded_nodes.size() == n_excluded_nodes) {
```
samskalicky, line 311: Do nodes have to be excluded in order to have a possible subgraph?

src/operator/subgraph/build_subgraph.cc
```diff
@@ -306,14 +300,19 @@ void PreSelectSubgraphNodes(const nnvm::Graph& g, SubgraphSelectorV2Ptr subgraph
...
+        // no possible subgraph for the current node snid
+        break;
+      }
+      n_excluded_nodes = excluded_nodes.size();
```
samskalicky, line 315: can we add a comment here, "failed to label subgraph due to loop, excluding nodes to remove loop and retrying"

src/operator/subgraph/build_subgraph.cc
```diff
@@ -306,14 +300,19 @@ void PreSelectSubgraphNodes(const nnvm::Graph& g, SubgraphSelectorV2Ptr subgraph
...
   const size_t max_num_retry = simple_nodes.size() * simple_nodes.size();
   size_t count = 0;
   bool success = false;
   while (!success && count < max_num_retry) {
     success = LabelSubgraph(g, subgraph_selector, label, snid, simple_nodes, subgraph_nodes,
                             &excluded_nodes);
     if (!success) {
-      CHECK(!excluded_nodes.empty());
+      if (excluded_nodes.size() == n_excluded_nodes) {
+        // no possible subgraph for the current node snid
```
samskalicky, line 312: can we extend this comment to something like "failed to label subgraph to loop, excluding nodes does not remove loop"

samskalicky:
Thanks @Kh4L! this is a really great addition to the partitioning logic to be able to skip subgraphs rather than error out when theres a loop. It took me a while to come up to speed with the complexity of the logic here, can you add some more comments so that next time it will be easier?

Kh4L:
@mxnet-bot run ci [unix-cpu]

Kh4L:
@mxnet-bot run ci [clang]

Kh4L:

PR #18562, opened 2020-06-15
[NumPy] allow mixed array types

### LINKED ISSUE #18561: array protocol support on memory sharing
## Description
For arrays that share memory, np.may_share_memory doesn't work for numpy types yet.
```
import mxnet as mx
import numpy as _np
np_array = _np.ones((3,5))
mx_nd_array = mx.nd.from_numpy(np_array, zero_copy=True)
mx_np_array = mx_nd_array.as_np_ndarray()
mx_np_array[0, 0] = 2

In [42]: mx_np_array
Out[42]:
array([[2., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.]], dtype=float64)

In [43]: mx_nd_array
Out[43]:

[[2. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1.]]
<NDArray 3x5 @cpu(0)>

In [44]: np_array
Out[44]:
array([[2., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1.]])
```

Then, if we call `np.may_share_memory(np_array, mx_np_array)` or `np.may_share_memory(np_array, mx_nd_array)`, it returns
```
TypeError: no implementation found for 'numpy.may_share_memory' on types that implement __array_function__: [<class 'numpy.ndarray'>, <class 'mxnet.numpy.ndarray'>]
```

It works for `_np.may_share_memory(mx_nd_array, mx_np_array)` though
```
True
```

szha:
## Description ##
This PR adds mx.npx.from_numpy and enables mixing the usage of mxnet np ndarray and native numpy array.
fixes #18561

### Essentials ###

### Changes ###
- [x] refactor dlpack functionality into dlpack.py
- [x] mx.npx.from_numpy
- [x] enable mixed use of np.ndarray and mx.np.ndarray

### DIFF
--- a/python/mxnet/numpy/multiarray.py
+++ b/python/mxnet/numpy/multiarray.py
@@ -182,7 +182,7 @@ def _reshape_view(a, *shape):  # pylint: disable=redefined-outer-name
 
 def _as_mx_np_array(object, ctx=None):
     """Convert object to mxnet.numpy.ndarray."""
-    if isinstance(object, ndarray):
+    if object is None or isinstance(object, ndarray):
         return object
     elif isinstance(object, _np.ndarray):
         np_dtype = _np.dtype(object.dtype).type
--- a/python/mxnet/numpy/multiarray.py
+++ b/python/mxnet/numpy/multiarray.py
@@ -392,11 +392,17 @@ class ndarray(NDArray):
             out = func(*new_args, **new_kwargs)
             return _as_mx_np_array(out, ctx=cur_ctx)
         else:
-            # Note: this allows subclasses that don't override
-            # __array_function__ to handle mxnet.numpy.ndarray objects
-            if not py_all(issubclass(t, ndarray) for t in types):
-                return NotImplemented
-            return mx_np_func(*args, **kwargs)
+            if py_all(issubclass(t, ndarray) for t in types):
+                return mx_np_func(*args, **kwargs)
+            else:
+                try:
+                    cur_ctx = next(a.ctx for a in args if hasattr(a, 'ctx'))
+                except StopIteration:
+                    cur_ctx = next(a.ctx for a in kwargs.values() if hasattr(a, 'ctx'))
+                new_args = _as_mx_np_array(args, ctx=cur_ctx)
+                new_kwargs = {k: _as_mx_np_array(v, cur_ctx) for k, v in kwargs.items()}
+                return mx_np_func(*new_args, **new_kwargs)
+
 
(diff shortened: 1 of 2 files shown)


python/mxnet/numpy/multiarray.py
```diff
@@ -148,10 +148,10 @@ def _reshape_view(a, *shape):  # pylint: disable=redefined-outer-name
 
 def _as_mx_np_array(object, ctx=None):
     """Convert object to mxnet.numpy.ndarray."""
-    if isinstance(object, _np.ndarray):
-        if not object.flags['C_CONTIGUOUS']:
-            object = _np.ascontiguousarray(object, dtype=object.dtype)
```
sxjscience, line 153: Should we make sure it's c_contiguous?

szha: I'm changing it to make a copy of the numpy array so it's already handled by the usage of `array()`. The reason for not using zero-copy `from_numpy` is that `from_numpy` changes the original array to be read-only, because mxnet execution is asynchronous and thus mutation from numpy side on the same buffer will result in undefined behavior.

szha:
This aims at fixing a more general case for #18561

tests/python/unittest/test_numpy_ndarray.py
```diff
@@ -1401,3 +1401,23 @@ def test_from_numpy_exception():
...
+
+def test_mixed_array_types():
+    np_array = _np.array([[1, 2], [3, 4], [5, 6]], dtype="float32")
+    mx_array = mx.np.ones((3, 1))
+    assert_almost_equal(mx_array + np_array, 1+np_array)
+
+def test_mixed_array_types_share_memory():
+    np_array = _np.array([[1, 2], [3, 4], [5, 6]], dtype="float32")
+    mx_array = mx.npx.from_numpy(np_array)
+    assert _np.may_share_memory(np_array, mx_array)
```
eric-haibin-lin, line 1413: i have a stupid numpy question - usually in what case would `may_share_memory` be useful?

szha: Something like this:
https://github.com/scikit-learn/scikit-learn/blob/36ae6a06d690b992c21280708ca365fce5beec99/sklearn/utils/validation.py#L662-L663

PR #18570, opened 2020-06-15
Reducing the number of calls of the NDArray constructors

andrei5055:
## Description ##
During training, MxNet reinitiates data stored in the instances of `NDArray` class by making a lot of `NDArray` constructor's calls and assigning the newly constructed objects to the existing ones.
First of all, it is not very effective, because 

- (a) usually we call some `NDArray` constructor AND NDArray` copy constructor 
- (b) after each such call the NDArray destructor will be called.

Second, lots of such calls are done "under the hood" and are not effective at all. For instance, the usage of 
```cpp
          std::vector<NDArray> inputs, outputs
```
in lambda expressions generates more `NDArray` constructor calls than it's actually necessary:

- 3 times more for `PushFCompute(...)` and 
- 1.5 time more for `PushFComputeEx(...)`,  `PushOperator(...)`;
 
Third, it's hard to debug.

To fix this problem we propose to use 

- the newly implemented `NDArray::Init(...), NDArray::ReInit(...)` methods
- `std::vector<NDArray *> inputs, outputs;` instead of `std::vector<NDArray> inputs, outputs;`

Our experiments show that these changes reduce the number of `NDArray` constructor calls by approx. 80%:

### Essentials ###

### Changes ###
- [x] The usage of
```
std::vector<NDArray> inputs, outputs
```
in `PushFCompute(...)` , `PushFComputeEx(...)`,  `PushOperator(...)`
was replaced by 
```cpp
std::vector<NDArray> inputs, outputs
```
- [x] The usage of macros, similar to the following ones, which are defined in different ways  for fifferent values of `MXNET_USE_MKLDNN` simplifies the code
```cpp
#if MXNET_USE_MKLDNN == 1
#define INVALIDATE_OUTPUTS(outputs, req) InvalidateOutputs(&outputs, req)
#define INVALIDATE_OUTPUTS_COND(cond, outputs, req) if (cond) INVALIDATE_OUTPUTS(outputs, req)
// add for mkldnn OP + no mkldnn OP
#define CREATE_DEFAULT_INPUTS(cond, attrs, func_call) \
 if (cond) { \
 const auto is_mkldnn = Op::GetAttr<bool>("TIsMKLDNN"); \
 if (!is_mkldnn.get(attrs.op, false)) func_call; \
 }

#else
#define INVALIDATE_OUTPUTS(outputs, ...)       // empty macros
#define INVALIDATE_OUTPUTS_COND(outputs, ...)  // empty macro
#define CREATE_DEFAULT_INPUTS(input, ...)      // empty macro
#endif
```
- [x] Implementation of constructor:
```cpp
  NDArray(const NDArrayStorageType stype, const mxnet::TShape &shape, Context ctx,
          bool delay_alloc = true, int dtype = mshadow::default_type_flag,
          std::vector<int> aux_types = {}, mxnet::ShapeVector aux_shapes = {},
          mxnet::TShape storage_shape = mxnet::TShape(mshadow::Shape1(0)));
```
was moved into newly implemented:
```cpp
void NDArray::ReInit(const NDArrayStorageType stype, const mxnet::TShape &shape,
 Context ctx, int dtype, bool delay_alloc, const std::vector<int> *pAux_types,
 const mxnet::ShapeVector *pAux_shapes, const mxnet::TShape *pStorage_shapes)
```

### DIFF
--- a/src/imperative/imperative_utils.h
+++ b/src/imperative/imperative_utils.h
@@ -37,6 +37,65 @@
 #define MXNET_IMPERATIVE_IMPERATIVE_UTILS_H_
 
 namespace mxnet {
+
+#if MXNET_USE_MKLDNN == 1
+template<typename T>T *pntr(T &obj)           { return &obj; }  // NOLINT
+template<typename T>T *pntr(T *obj)           { return obj; }
+
+template<typename T>
+void InvalidateOutputs(const std::vector<T> *pArrs,
+                        const std::vector<OpReqType> &reqs) {
+  auto arrs = *pArrs;
+  for (size_t i = 0; i < arrs.size(); i++) {
+    if (reqs[i] == kWriteTo || reqs[i] == kNullOp)
+      pntr(arrs[i])->InvalidateMKLDNNData();
+  }
+}
+
+static inline void CreateDefaultInputs(const std::vector<NDArray> &arrs,
+                                       std::vector<NDArray> *out_arrs) {
+  out_arrs->clear();
+  for (size_t i = 0; i < arrs.size(); ++i) {
+    if (arrs[i].IsMKLDNNData())
+      out_arrs->push_back(arrs[i].Reorder2Default());
+    else
+      out_arrs->push_back(arrs[i]);
+  }
+}
+
+static inline void CreateDefaultInputs(std::vector<NDArray> *pArrs) {
+  auto arrs =  *pArrs;
+  for (size_t i = 0; i < arrs.size(); ++i)
+    arrs[i].SelfReorder2Default();
+}
+
+#define INVALIDATE_OUTPUTS(outputs, req) InvalidateOutputs(&outputs, req)
+// kCrossDeviceCopy is used for `_copy_to` operator, which doesn't compute immediately in
+// its FCcomputeEx, but AsyncPush the copy operation to engine.
+// So for the case that A is holding mkldnn memory, and then copy A to B, and then copy B
+// back to A, we shouldn't invalidate outputs for copying B back to A, because at this time,
+// copying A to B may not happen, and will corrupt A's memory.
+#define INVALIDATE_OUTPUTS_COND(cond, outputs, req) if (cond) INVALIDATE_OUTPUTS(outputs, req)
+
+// add for mkldnn OP + no mkldnn OP
+#define CREATE_DEFAULT_INPUTS(cond, attrs, func_call)  \
+                          if (cond) {                                               \
+                            const auto is_mkldnn = Op::GetAttr<bool>("TIsMKLDNN");  \
+                            if (!is_mkldnn.get(attrs.op, false)) func_call;         \
+                          }
+
+#else
+#define INVALIDATE_OUTPUTS(outputs, ...)       // empty macros
+#define INVALIDATE_OUTPUTS_COND(outputs, ...)  // empty macro
+#define CREATE_DEFAULT_INPUTS(input, ...)      // empty macro
+#endif
+
+#if MXNET_USE_MKLDNN == 1
+#else
+#define CREATE_DEFAULT_INPUTS_MKLDNN(in_array, in_array_fallback, attrs)  // empty macro
+#endif
+
+
 namespace imperative {
 
 namespace {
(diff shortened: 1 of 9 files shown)


ptrendx:
@eric-haibin-lin Could you take a look?

szha:
@mxnet-bot run ci [all]

src/imperative/imperative_utils.h
```diff
@@ -35,6 +35,59 @@
...
+    else
+      out_arrs->push_back(arrs[i]);
+  }
+}
+
+static inline void CreateDefaultInputs(std::vector<NDArray> *pArrs) {
+  auto &&arrs = *pArrs;
+  for (size_t i = 0; i < arrs.size(); ++i)
+    arrs[i].SelfReorder2Default();
+}
```
leezu, line 68: Why delete the "Remove helper function after subgraph feature added" comment?

andrei5055: @leezu:  First of all, by replacing NDArray constructors followed by assign operators by some Init function we should see some run time improvement, just because in Init function we do exactly the same as it was done in the constructor, but not doing assigns AND not calling destructors after that.

But the main goal of this replacement was the simplification of the tracing the NDArray objects during the debugging process.

It's extremely difficult to do that when millions and millions of NDArray objects are constructed just to pass to each other some values stored in their members and are destructed after that. 

The tracing of data flows is much easy when corresponding NDArray is reinitialized and reused when needed.

Whit that approach, more than 90% of NDArray constructor calls were eliminated.

andrei5055: > Why delete the "Remove helper function after subgraph feature added" comment?

Because I don't have access to my NVidia computer, I will answer this question later.

src/imperative/imperative_utils.h
```diff
@@ -35,6 +35,59 @@
...
+
+#define INVALIDATE_OUTPUTS(outputs, req) InvalidateOutputs(&outputs, req)
```
eric-haibin-lin, line 70: Why do you make this macros instead of calling the function directly?

andrei5055: As I mentioned in the PR **Description** (see **Changes**), that and a couple of other macros are defined differently for different values of `MXNET_USE_MKLDNN` flag:  
```
#if MXNET_USE_MKLDNN == 1
#define INVALIDATE_OUTPUTS(outputs, req) InvalidateOutputs(&outputs, req)
#else
#define INVALIDATE_OUTPUTS(outputs, ...)       // empty macros
#endif
```

It allows us to eliminate the multiple usages of `#if MXNET_USE_MKLDNN == 1 / #endif lines and simplify the code.

src/imperative/imperative_utils.h
andrei5055: It looks like I lost this comment line when I was moving ```static inline void CreateDefaultInputs``` from one file into another. I just recreated it and rebaised this PR branch.

src/imperative/imperative_utils.h
leezu: Thank you for simplifying the code. Do you think it would be better to rely on `constexpr if` inside the functions (`IsDefaultData`, `InvalidateOutputs`, ...) instead of adding more Macros to the codebase?

src/imperative/imperative_utils.h
leezu: > It looks like I lost this comment line when I was moving `static inline void CreateDefaultInputs` from one file into another. I just recreated it and rebaised this PR branch.

Thank you. The subgraph feature is available now, so the TODO can be addressed (and preserving the comment can be helpful).

> First of all, by replacing NDArray constructors followed by assign operators by some Init function we should see some run time improvement, just because in Init function we do exactly the same as it was done in the constructor, but not doing assigns AND not calling destructors after that.

> But the main goal of this replacement was the simplification of the tracing the NDArray objects during the debugging process.

Thank you for elaborating @andrei5055. Simplifying the tracing of NDArray objects by using `ReInit` is helpful. My point is that the NDArray already has a cheap default constructor which would be invoked by the declaration of `std::vector<NDArray> inputs, outputs;` in `PushOperator` etc.:

https://github.com/apache/incubator-mxnet/blob/7908d7eb56fc9d20c12afffd8ea592b959b80bfc/include/mxnet/ndarray.h#L84-L87

So why add a separate `Init` method with unclear semantics? It seems you're not using the `Init` method and the `ReInit` should be sufficient

Also note that your claim "usually we call some NDArray constructor AND NDArray` copy constructor" may not apply due to mandatory elision of copy/move operations in C++17.

src/imperative/imperative_utils.h
andrei5055: Frankly, I am not familiar with  `constexpr if` notion. I will take a look at it, to see if it is possible to use it as effective as empty macros.

src/imperative/imperative_utils.h
andrei5055: > So why add a separate `Init` method with unclear semantics? It seems you're not using the `Init` method and the `ReInit` should be sufficient.

Both newly implemented `NDArray::ReInit` methods use `NDArray::Init(...)` you were talking about. So, just for not duplicate code, we need to keep it. Perhaps, it makes sense to make it `private` and I did it in my last commit.

src/imperative/imperative_utils.h
andrei5055: > Do you think it would be better to rely on `constexpr if` inside the functions (`IsDefaultData`, `InvalidateOutputs`, ...) instead of adding more Macros to the codebase?

Definitely, it could be done, but I am not sure that it will be as effective as the use of empty macros. My guess is: the compiler is not smart enough to DO NOT generate the functions `IsDefaultData, InvalidateOutputs, ...` when they are empty.  Therefore, they will be defined, called, and do nothing! 

I think that the use of the trick with empty macro is more effective: 

-  we define, compile, and use functions only when they really needed. 
-  the types of parameters of such functions could be defined ONLY when they will be really used.  (Therefore, in general, we don't need even the presence of some header files with these definitions)

But I could implement these functions with  `constexpr if` if you really think that it would be better.

src/imperative/imperative_utils.h
leezu: Making it private addresses my concern. Thank you.

src/imperative/imperative_utils.h
leezu: > My guess is: the compiler is not smart enough to DO NOT generate the functions IsDefaultData, InvalidateOutputs, ... when they are empty. Therefore, they will be defined, called, and do nothing!

My guess was that such small (empty) functions would be inlined. But it's fine to keep the empty macro trick

andrei5055:
Just one test failed because of timeout:
```
1199.93s call     tests/python/unittest/test_io.py::test_inter_methods_in_augmenter[2]
```
I have tried and cannot reproduce it locally.

leezu:
@mxnet-bot run ci [unix-cpu]

PR #18574, opened 2020-06-16
Update the onnx-tensorrt submodule - CI to TRT7

Kh4L:
## Description ##
This PR updates the onnx_tensorrt submodule to the latest commit.

### DIFF
--- a/ci/docker/Dockerfile.build.ubuntu
+++ b/ci/docker/Dockerfile.build.ubuntu
@@ -137,17 +137,27 @@ RUN cd /usr/local && \
 # https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact
 ARG BASE_IMAGE
 RUN export SHORT_CUDA_VERSION=${CUDA_VERSION%.*} && \
+    wget -O nvidia-ml.deb https://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu1804/x86_64/nvidia-machine-learning-repo-ubuntu1804_1.0.0-1_amd64.deb && \
+    dpkg -i nvidia-ml.deb && \
     apt-get update && \
     if [ ${SHORT_CUDA_VERSION} = 10.0 ]; then \
-        apt-get install -y "libnvinfer-dev=5.1.5-1+cuda10.0"; \
+        TRT_VERSION="7.0.0-1+cuda10.0"; \
+        TRT_MAJOR_VERSION=7; \
     elif [ ${SHORT_CUDA_VERSION} = 10.1 ]; then \
-        apt-get install -y "libnvinfer-dev=5.1.5-1+cuda10.1"; \
+        TRT_VERSION="6.0.1-1+cuda10.1"; \
+        TRT_MAJOR_VERSION=6; \
     elif [ ${SHORT_CUDA_VERSION} = 10.2 ]; then \
-        apt-get install -y "libnvinfer-dev=6.0.1-1+cuda10.2"; \
+        TRT_VERSION="7.0.0-1+cuda10.2"; \
+        TRT_MAJOR_VERSION=7; \
     else \
         echo "ERROR: Cuda ${SHORT_CUDA_VERSION} not yet supported in Dockerfile.build.ubuntu"; \
         exit 1; \
     fi && \
+    apt-get install -y libnvinfer${TRT_MAJOR_VERSION}=${TRT_VERSION} \
+                       libnvinfer-dev=${TRT_VERSION} \
+                       libnvinfer-plugin${TRT_MAJOR_VERSION}=${TRT_VERSION} \
+                       libnvinfer-plugin-dev=${TRT_VERSION} && \
+    rm nvidia-ml.deb && \
     rm -rf /var/lib/apt/lists/*
 
 FROM gpu as gpuwithcudaruntimelibs
--- a/ci/docker/Dockerfile.build.ubuntu
+++ b/ci/docker/Dockerfile.build.ubuntu
@@ -50,8 +50,6 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
         build-essential \
         ninja-build \
         git \
-        protobuf-compiler \
-        libprotobuf-dev \
         clang-6.0 \
         clang-tidy-6.0 \
         python-yaml \
@@ -83,6 +81,8 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
         pandoc \
         ## Build-dependencies for ccache 3.7.9
         autoconf \
+        autogen \
+        libtool \
         gperf \
         libb2-dev \
         libzstd-dev && \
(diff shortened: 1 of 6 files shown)


Kh4L:
@mxnet-bot run ci [unix-gpu]

szha:
Thanks, Serge. How is this tested at the moment?

Kh4L:
> Thanks, Serge. How is this tested at the moment?

It was supposed to be tested in the tensorrt tests deleted here in this PR https://github.com/apache/incubator-mxnet/pull/18579 ( related to https://github.com/apache/incubator-mxnet/issues/18729 ). I don't understand why it was deleted.

Also, I am getting this error in the CI
```
[2020-07-17T18:52:46.759Z] ninja: error: unknown target 'onnx/onnx.proto', did you mean 'onnx/onnx-ml.proto'?
```
 not sure of what needs to be changed in ninja

szha:
cc @leezu on the build.

I think the tests were removed as we are deprecating NDArray in 2.0 and there's additional work needed for supporting numpy operator for tensorrt. cc @eric-haibin-lin in case there is more context.

leezu:
> not sure of what needs to be changed in ninja

you need to check what changed in the onnx-tensorrt submodule and update the cmakelists.txt accordingly

ci/docker/Dockerfile.build.ubuntu
```diff
@@ -137,17 +137,27 @@ RUN cd /usr/local && \
 # https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact
 ARG BASE_IMAGE
 RUN export SHORT_CUDA_VERSION=${CUDA_VERSION%.*} && \
+    wget -O nvidia-ml.deb https://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu1804/x86_64/nvidia-machine-learning-repo-ubuntu1804_1.0.0-1_amd64.deb && \
+    dpkg -i nvidia-ml.deb && \
```
leezu, line 141: Why? You shouldn't need this

ci/docker/Dockerfile.build.ubuntu
```diff
@@ -137,17 +137,27 @@ RUN cd /usr/local && \
...
+        TRT_VERSION="7.0.0-1+cuda10.0"; \
+        TRT_MAJOR_VERSION=7; \
     elif [ ${SHORT_CUDA_VERSION} = 10.1 ]; then \
-        apt-get install -y "libnvinfer-dev=5.1.5-1+cuda10.1"; \
+        TRT_VERSION="6.0.1-1+cuda10.1"; \
+        TRT_MAJOR_VERSION=6; \
     elif [ ${SHORT_CUDA_VERSION} = 10.2 ]; then \
-        apt-get install -y "libnvinfer-dev=6.0.1-1+cuda10.2"; \
+        TRT_VERSION="7.0.0-1+cuda10.2"; \
+        TRT_MAJOR_VERSION=7; \
```
leezu, line 151: Why mix version 6 and 7?

Kh4L: TRT7 is not available for 10.1 (TRT team management decision)

ci/docker/Dockerfile.build.ubuntu
Kh4L: This allows us to choose the right version

leezu: My point is that the nvidia ml repository should already be available and you don't need to install it a second time? Prior to your change, there was no issue to run `apt-get install -y "libnvinfer-dev=6.0.1-1+cuda10.2`

ci/docker/docker-compose.yml
```diff
@@ -108,6 +108,16 @@ services:
         BASE_IMAGE: nvidia/cuda:10.1-cudnn7-devel-ubuntu18.04
       cache_from:
         - ${DOCKER_CACHE_REGISTRY}/build.ubuntu_gpu_cu101:latest
+  ubuntu_gpu_cu102:
```
leezu, line 111: Let's delete the ubuntu_gpu_cu101 then?

ChaiBapchya: I recommended serge to add instead of update. I wasn't sure if we should migrate all the existing versions from cu101 to cu102. Do we want to test a few things on cu101 as we continue to support?
Or since this is master [proxy for 2.0], do we move to the cuda 10.2 & then potentially to 11?

Kh4L: It's safer to move only this test on 10.2 for now, I agreed with @ChaiBapchya

leezu: Generally we can't test all possible cuda versions and thus our strategy so far is to test the oldest and newest supported version. For simplicity, I'd recommend to stick with the existing strategy and update all tests to 10.2.

ci/docker/Dockerfile.build.ubuntu
```diff
@@ -110,6 +108,15 @@ COPY install/requirements /work/
...
     python3 -m pip install -r /work/requirements
 
+RUN git clone --recursive -b 3.5.1.1 https://github.com/google/protobuf.git && \
+    cd protobuf && \
+    ./autogen.sh && \
+    ./configure && \
+    make -j$(nproc) install && \
+    cd .. && \
+    rm -rf protobuf && \
+    ldconfig
```
leezu, line 118: Is the version provided by Ubuntu too old?

Kh4L: I am not sure, but onnx is failing and the error seems to be related to the protobuf version.

This is the version we are using in our internal containers

ci/docker/docker-compose.yml
Kh4L: Ok if you guys at Amazon agree, I guess it is fine for us! (actually I pinged you on slack @leezu )

ci/docker/Dockerfile.build.ubuntu
Kh4L: This method guarantees us to have the right nvinfer version no matter what happens to the container in the future

ci/docker/Dockerfile.build.ubuntu
leezu: Do you know that you can reproduce the CI build via `python ci/build.py -R --platform ubuntu_gpu_cu102 /work/runtime_functions.sh build_ubuntu_gpu_tensorrt`?

ci/docker/Dockerfile.build.ubuntu
leezu: Currently the container already contains the machine learning repo in `/etc/apt/sources.list.d/nvidia-ml.list`. Why would the container team decide to break their users by removing the machine learning repo? Is there any other reason for re-installing the already existing repository?

leezu: I noticed the following issue with the container though I'm not sure about the impact https://gitlab.com/nvidia/container-images/cuda/-/issues/82

Kh4L:

Kh4L:
@leezu thanks, I missed those!
I reverted them back

PR #18616, opened 2020-06-24
Update disclaimer wording

leezu:
>  2) Why say "they might contain software components with more restrictive licenses than the Apache License" where they clearly do contain more restrictive licenses.

https://issues.apache.org/jira/browse/LEGAL-515?focusedCommentId=17127138&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17127138


leezu:

PR #18619, opened 2020-06-25
Refactor scope functionality in Python API

leezu:
- Remove deprecated metaclass functionality
- Remove global state in naming
- Switch from threading.local to asyncio compatible contextvars
- Stop exposing UUIDs in parameter name


python/mxnet/gluon/block.py
```diff
@@ -1573,29 +1534,31 @@ def __init__(self, outputs, inputs, params=None):
...
-                    "Please check 'params' is complete!" % name
-                for structured_name in lis:
-                    self._reg_params[structured_name] = param
-            else:
-                self._reg_params[name] = param
+        if params is None:
+            params = {}
+        unused_params = set(params.keys()) - set(arg_params) - set(aux_params)
+        if len(unused_params) > 0:
+            raise ValueError('{} params are unused by the model.'.format(unused_params))
```
acphile, line 1541: I think it is more reasonable to throw out a warning and remove unused parameters instead of raising a ValueError.

leezu: There's the risk that user's misuse the API by mistake and don't see the warning (eg. CI server often doesn't error out for warnings). So IMO it's better to enforce a high standard and ensure that users don't pass an invalid parameter dict. If they do so by mistake and don't see the warning, they end up with randomly initialized parameters instead of the intended values.

acphile: I‘m worried about this case: when a model's forward returns multiple outputs and we want to build SymbolBlock using one of the outputs, it may be difficult for users to precisely pass the intended parameters which the output uses.

sxjscience:
@ZheyuYe FYI. The GluonNLP side needs to wait for this commit.

python/mxnet/gluon/block.py
```diff
@@ -1620,29 +1602,31 @@ def __init__(self, outputs, inputs, params=None):
...
```
szha, line 1609: I think we should probably include two switches to detect parameter set mismatch, ignore_extra_param and ignore_missing_param, with both default to False. if either case happens, we should raise ValueError with the complete information on what's extra and what's missing. Warning will often be visual noise that people ignore.

Issue #17858, opened 2020-03-17
dynamic load libnvrtc.so


leezu:
## Description
Compiling MXNet with `nvrtc` support dynamically links `libcuda.so`. However, `libcuda.so` is part of the cuda driver and not typically available on cpu machines. This means, on cpu machines `libmxnet.so` will have missing dependencies and can't be loaded.

Unfortunately, parts of MXNet compilation require to `dlopen(libmxnet.so)`. In particular `OpWrapperGenerator.py` for the cpp package. This breaks compiling mxnet with nvrtc + cpp-package on cpu machines (as used by our CI).

As a workaround, the stub driver can be added to `LD_LIBRARY_PATH`. A better solution may be to `dlopen` libnvrtc and thereby drop the dependency on `libcuda.so`.

CC: @ptrendx 

## References
- https://github.com/pytorch/pytorch/issues/3395
- https://github.com/NervanaSystems/ngraph/issues/765

ptrendx:
I agree it makes sense to `dlopen` it.

ChaiBapchya:
@mxnet-label-bot add [bug]
This work-around [adding to LD_LIBRARY_PATH] is great. But without the workaround, the mxnet breaks. So I guess, it needs to be fixed and is a bug rather than a feature-request.

ptrendx:
Hi @mseth10, do you think the utilities that you introduced in dynamic library loading support PR could be used for this?

mseth10:
Hi @ptrendx if you need a generic dlopen for this, I'll suggest using ctypes CDLL.

ptrendx:
But that would be just for Python then, right? I would need it to call functions from C++ and the load_lib function from your PR seems like it already handles all the differences in OSs and stuff.

mseth10:
Sorry, I thought you were referring to the python api introduced by the PR. That has a lot of additional functionality.
We can use the c++ utility function - lib_load https://github.com/apache/incubator-mxnet/blob/9d440868603ad26b702e12ddd2587e5c4b56e42b/src/initialize.cc#L113

ptrendx:
I thought about this and the real problem seems to actually be `libcuda.so`, right? `libnvrtc.so` does not depend on `libcuda.so` directly, it is just that to actually use the kernels compiled with `libnvrtc.so` one needs to use the driver API. So it should be fine to link normally to `libnvrtc.so`, and just dlopen `libcuda.so`, right @leezu?

leezu:
[...]

Issue #18280, opened 2020-05-11
[Performance Regression] GPU memory increase for training and inference models


karan6181:
## Description
- There is an MXNet nightly benchmark which runs CV and NLP models on MXNet Nightly pip wheel and report the metrics and it showed a performance regression on GPU Memory.
- After bisecting the PRs, the PR due to which there is an increase in GPU memory of 120-130 MB happen is [17767](https://github.com/apache/incubator-mxnet/pull/17767). I ran the SSD training with and without the PR commit and it showed around 120MB increase in GPU Memory. I haven't ran all those models personally but the reproducing script is basic and it is applicable to all those models.
- List of models that affected are as below. The GPU Memory increase data we got from our internal benchmarking system:

| model                                            | GPU Memory (From:to)                     |
| ------------------------------------------------ | ---------------------------------------- |
| VGG16_training                                   | 9.63k to 9.78k                           |
| SSD_training                                     | 4.75k to 4.91k                           |
| LSTM_inference: Gluon and Module                 | G: 1.55k to 1.71<br />M: 1.47k to 1.64k  |
| Caffenet_inference: Gluon and Module             | G: 2.17k to 2.31k<br />M: 2.0k to 2.16k  |
| YoloV3_GPU: Gluon and Module                     | G: 1.87k to 2.03k<br />M: 1.84k to 2.0k  |
| Resnet50_v2_FP16_inference_GPU: Gluon and Module | G: 1.58k to 1.74k<br />M: 1.55k to 1.71k |
| Resnet50_v2_inference_GPU: Gluon and Module      | G: 1.55k to 1.71k<br />M: 1.48k to 1.64k |
| Resnet152_v2_inference_GPU: Gluon and Module     | G: 1.92k to 2.08k<br />M: 1.86k to 2.02k |
| Inception_Inference_GPU: Gluon and Module        | G: 1.43k to 1.59k<br />M: 1.40 to 1.56k  |
| SSD_inference_GPU: Gluon and Module              | G: 1.65k to 1.81k<br />M: 1.54k to 1.70k |
| A3C_inference_GPU: Gluon and Module              | G: 1.32k to 1.48k<br />M: 1.31k to 1.47k |
| word_language_model_hybrid_p3.16_training        | 2.25k to 2.43k                           |
| Mobile_pose_training                             | 2.38k to 2.54k                           |

## To Reproduce
- Run the below lines of code and monitor GPU usage using `nvidia-smi` command.
```bash
[...]

PR #18205, opened 2020-04-30
Fix nightly CD for GPU builds

mseth10:
## Description ##
This PR fixes nightly CD GPU tests by updating the build toolchain to use cmake static build and updating dnnl headers stash location. It removes 7.5 arch for cu100, cu101, cu102 builds to solve oversized libmxnet.so binary issue with cmake builds.
It also fixes the issue with dnnl headers packaging into nightly build artifacts. Fixes #18120 

Here's the link to a broken CD pipeline: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/restricted-mxnet-cd%2Fmxnet-cd-release-job/detail/mxnet-cd-release-job/1024/pipeline/

Commands used to reproduce and test the fix:
```
alias python=python3

git clone --recursive https://github.com/apache/incubator-mxnet.git
cd incubator-mxnet
pip3 install -r ci/requirements.txt --user
# make changes to mxnet code

# install docker-compose
sudo curl -L "https://github.com/docker/compose/releases/download/1.25.5/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
sudo chmod +x /usr/local/bin/docker-compose
sudo ln -s /usr/local/bin/docker-compose /usr/bin/docker-compose

# build on c5.18xl
python ci/build.py --platform centos7_gpu_cu100 /work/runtime_functions.sh build_static_libmxnet cu100
# test on g3.8xl
python ci/build.py --docker-registry mxnetci --nvidiadocker --platform centos7_gpu_cu100 --docker-build-retries 3 --shm-size 500m /work/runtime_functions.sh cd_unittest_ubuntu cu100
```

Issue #17045, opened 2019-12-11
Relocation truncation issues


leezu:
## Description
`libmxnet.so` is too large (depending on compile options), so that linking fails. This was observed before on CI with test coverage functionality enabled (https://github.com/apache/incubator-mxnet/issues/15971), but can also happen with non-test-coverage builds, such as `-DUSE_INT64_TENSOR_SIZE=ON` build.

I first observe this in the #17031 (http://jenkins.mxnet-ci.amazon-ml.com/blue/rest/organizations/jenkins/pipelines/mxnet-validation/pipelines/unix-gpu/branches/PR-17031/runs/6/nodes/52/steps/84/log/?start=0), but can easily reproduce it on the master branch when building with GCC 7.4.

### Error Message
From the CI

```
/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/crti.o: In function `_init':
(.init+0x7): relocation truncated to fit: R_X86_64_REX_GOTPCRELX against undefined symbol `__gmon_start__'
/usr/lib/gcc/x86_64-linux-gnu/5/crtbeginS.o: In function `deregister_tm_clones':
crtstuff.c:(.text+0x3): relocation truncated to fit: R_X86_64_PC32 against `.tm_clone_table'
crtstuff.c:(.text+0xa): relocation truncated to fit: R_X86_64_PC32 against symbol `__TMC_END__' defined in .nvFatBinSegment section in libmxnet.so
crtstuff.c:(.text+0x1e): relocation truncated to fit: R_X86_64_REX_GOTPCRELX against undefined symbol `_ITM_deregisterTMCloneTable'
/usr/lib/gcc/x86_64-linux-gnu/5/crtbeginS.o: In function `register_tm_clones':
crtstuff.c:(.text+0x43): relocation truncated to fit: R_X86_64_PC32 against `.tm_clone_table'
crtstuff.c:(.text+0x4a): relocation truncated to fit: R_X86_64_PC32 against symbol `__TMC_END__' defined in .nvFatBinSegment section in libmxnet.so
crtstuff.c:(.text+0x6b): relocation truncated to fit: R_X86_64_REX_GOTPCRELX against undefined symbol `_ITM_registerTMCloneTable'
/usr/lib/gcc/x86_64-linux-gnu/5/crtbeginS.o: In function `__do_global_dtors_aux':
[...]

PR #18622, opened 2020-06-25
Use RTC for elementwise and broadcast ops

ptrendx:
## Description ##
As described in https://github.com/apache/incubator-mxnet/issues/18280#issuecomment-627010252, MXNet currently contains too many CUDA kernels, that affect negatively compile time, size of the resulting binary (resulting in issues like #17045 and #18205), and GPU memory consumption (as all of those kernels need to be loaded during the first GPU context creation to GPU memory).

The reason of those problems is the number of templates that need to be instantiated, especially in the case of NumPy operators which need to accept different input/output types - this results in multiple nested `MSHADOW_TYPE_SWITCH` macros and great increase in the number of kernels generated, most of them pretty much never used. For example, executing this command:
```
cuobjdump -symbols -arch sm_70 /usr/local/lib/python3.6/dist-packages/mxnet/libmxnet.so | grep GLOBAL | wc -l
```
on the nightly build of mxnet-cu102 from 6/25 shows 69169 kernels (the same command executed on the library built with this PR at the time of writing gives 51511 kernels).

The proposed approach is to use RTC (runtime compilation) in order to generate the needed kernels at runtime. This saves the ahead-of-time compilation time and binary size as well as the GPU memory utilization (since only the needed kernels are generated, not all combinations).

To test the impact on binary size and memory consumption reduction I compiled MXNet for 5 GPU architectures (5.2, 6.0, 6.1, 7.0, 7.5) using CUDA 11 both from the head of this PR and from the latest master commit included (f872b4340).
Binary size reduction: 292 MB (from 2 GB to 1.7 GB)
Idle GPU memory consumption reduction: 96 MB (from 1442 MB to 1346 MB)
Idle GPU memory consumption reduction was checked by launching Python interpreter and checking GPU memory consumption after calling:
```
import mxnet as mx
a = mx.nd.zeros((1,), ctx=mx.gpu())
```

This PR uses that approach to handle elementwise and broadcast kernels (as well as their backward), which constitute a big portion of the total number of kernels in MXNet.

FYI @leezu @sxjscience @eric-haibin-lin 

### Essentials ###

### Changes ###
 - RTC is now required for using CUDA in MXNet
 - Unary, binary, binary with scalar, binary broadcast ops and their backward counterparts were changed to use RTC 

## Comments ##
- Things left to do:

   - [x] Test the performance impact of cache lookup for kernel code
   - [x] Convert `MixedUnaryBackward` functions
   - [x] Update PR description with the change in GPU memory utilization and binary size resulting from this PR

 - After this PR the next step would be to use the same approach for reduce kernels - this PR already contains a ground work for this as reduction was needed for backward of broadcast ops, but it does not apply that path to standalone reduction ops. Grepping for `reduce_kernel` in the symbols visible in libmxnet.so after application of this PR:
```
cuobjdump -symbols -arch sm_70 /usr/local/lib/python3.6/dist-packages/mxnet/libmxnet.so | grep GLOBAL | grep reduce_kernel | wc -l
```
gives 12057 entries. This would also help with reducing the amount of code duplication that this PR introduces (to maintain both RTC and non-RTC paths).


leezu:
Thank you @ptrendx! As this makes nvrtc feature mandatory, it may be necessary to first/prior to next release also fix https://github.com/apache/incubator-mxnet/issues/17858? It seems that there are a number of users that try to open GPU builds on machines without libcuda.so and this is broken since 1.6 due to https://github.com/apache/incubator-mxnet/issues/17858 (but currently there is the workaround of disabling nvrtc)

ptrendx:
Yeah, we will need to sort it out before the release - I actually thought that if we manage to get everything via RTC (which seems daunting though), we could actually dynamically load both libcuda and libnvrtc and have a single build that supports everything instead of mxnet-cu*. That said, RTC for everything is a big task and I would definitely need help from community if we were to pull it off.

ptrendx:
Also, I thought that you made the compilation use C++17 (so I use `if constexpr`), but now I see that the CUDA part is compiled with C++14 (and fails). What is the reason of that?

leezu:
> Also, I thought that you made the compilation use C++17 (so I use `if constexpr`), but now I see that the CUDA part is compiled with C++14 (and fails). What is the reason of that?

The reason is that CUDA does not support C++17 prior to CUDA11. We can consider requiring CUDA11 for MXNet 2

ptrendx:
Yeah, I wasn't aware of that. I decided to change the `if constexpr` to `#ifdef`s for now as the gpu portion is still compiled with nvcc.

Another thing - it seems that the docker for jetson does not include NVRTC, who can change that?

leezu:
> Another thing - it seems that the docker for jetson does not include NVRTC, who can change that?

You can just edit the Dockerfile. But there might be a problem that NVidia fails to publish a NVRTC cross-compilation package. If so, you'll first need to get it published internally. (That's the case for TensorRT. But hopefully NVRTC is already available)

ptrendx:
Hi @ChaiBapchya could you tell me whether there is any difference in hw/sw of centos-gpu runners compared to unix-gpu runners? I see some failures on centos-gpu that do not happen on unix-gpu.

ChaiBapchya:
Based on the existing CI infra setup, both centos-gpu & unix-gpu run on the same AMI
Same AMI translates essentially to same HW/SW.
However, after the instance is created, it runs the CI on a docker file. That dockerfile setup is different from unix & centos

Centos BUILD: https://github.com/apache/incubator-mxnet/blob/master/ci/docker/Dockerfile.build.centos7
Centos Test: https://github.com/apache/incubator-mxnet/blob/master/ci/docker/Dockerfile.publish.test.centos7

The variables used in these dockerfiles are defined in https://github.com/apache/incubator-mxnet/blob/master/ci/docker/docker-compose.yml

For e.g. 
In your case: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fcentos-gpu/detail/PR-18622/9/pipeline
Fails at 
```
ci/build.py --docker-registry mxnetci --nvidiadocker --platform centos7_gpu_cu92 --docker-build-retries 3 --shm-size 500m /work/runtime_functions.sh unittest_centos7_gpu
```

Thus the platform `centos7_gpu_cu92` defined here

https://github.com/apache/incubator-mxnet/blob/master/ci/docker/docker-compose.yml#L44-L53

ptrendx:
@mxnet-bot run ci [windows-gpu, unix-gpu, miscellaneous, website, centos-gpu, sanity, clang]

ptrendx:
@ChaiBapchya The Jetson CI pipeline fails with not being able to find NVRTC - I confirmed it is part of jetpack, so I guess it is not included in the package of libs you have in S3. Can you update it? (And yes, I'm aware of issue #18637).

leezu:
@ptrendx these are the files available in the S3 bucket. You may need to edit the Dockerfile to install one of the debs:

```
% aws s3 ls --recursive s3://mxnet-ci-prod-private-slave-data/nvidia                                                                                                                                   6s ~
2020-04-11 02:46:49          0 nvidia/
2020-04-11 02:47:05          0 nvidia/sdkm_downloads/
2020-04-11 03:07:04  237414931 nvidia/sdkm_downloads/Jetson-210_Linux_R32.3.1_aarch64.tbz2
2020-04-11 03:03:33  288481398 nvidia/sdkm_downloads/NVIDIA_Nsight_Graphics_L4T_2019.5.19322.deb
2020-04-11 03:24:00    5350765 nvidia/sdkm_downloads/NVIDIA_VisionWorks_References.zip
2020-04-11 03:17:26  113581742 nvidia/sdkm_downloads/NsightSystems-linux-public-2019.6.2.6-3ffb807.deb
2020-04-11 03:24:08    1022926 nvidia/sdkm_downloads/OpenCV-4.1.1-2-gd5a58aa75-aarch64-dev.deb
2020-04-11 03:23:59    9920774 nvidia/sdkm_downloads/OpenCV-4.1.1-2-gd5a58aa75-aarch64-libs.deb
2020-04-11 03:24:13      14240 nvidia/sdkm_downloads/OpenCV-4.1.1-2-gd5a58aa75-aarch64-licenses.deb
2020-04-11 03:24:07    2350538 nvidia/sdkm_downloads/OpenCV-4.1.1-2-gd5a58aa75-aarch64-python.deb
2020-04-11 03:24:11     192038 nvidia/sdkm_downloads/OpenCV-4.1.1-2-gd5a58aa75-aarch64-samples.deb
2020-04-11 03:24:09    1027754 nvidia/sdkm_downloads/OpenCV-4.1.1-2-gd5a58aa75-x86_64-dev.deb
2020-04-11 03:21:58   45701978 nvidia/sdkm_downloads/OpenCV-4.1.1-2-gd5a58aa75-x86_64-libs.deb
2020-04-11 03:24:12      18306 nvidia/sdkm_downloads/OpenCV-4.1.1-2-gd5a58aa75-x86_64-licenses.deb
2020-04-11 03:24:03    2576742 nvidia/sdkm_downloads/OpenCV-4.1.1-2-gd5a58aa75-x86_64-python.deb
2020-04-11 03:24:11     192030 nvidia/sdkm_downloads/OpenCV-4.1.1-2-gd5a58aa75-x86_64-samples.deb
2020-04-11 03:24:11     143245 nvidia/sdkm_downloads/SDKM_logs_JetPack_4.3_Linux_for_Jetson_TX1_2020-04-10_17-43-23.zip
2020-04-11 02:51:29 1263228889 nvidia/sdkm_downloads/Tegra_Linux_Sample-Root-Filesystem_R32.3.1_aarch64.tbz2
2020-04-11 02:51:29  394475774 nvidia/sdkm_downloads/cuda-repo-cross-aarch64-10-0-local-10.0.326_1.0-1_all.deb
2020-04-11 02:51:29 1000333292 nvidia/sdkm_downloads/cuda-repo-l4t-10-0-local-10.0.326_1.0-1_arm64.deb
2020-04-11 02:51:29 1582667320 nvidia/sdkm_downloads/cuda-repo-ubuntu1804-10-0-local-10.0.326-410.108_1.0-1_amd64.deb
2020-04-11 03:22:38   25538902 nvidia/sdkm_downloads/devtools_docs.zip
2020-04-11 03:24:13      16108 nvidia/sdkm_downloads/graphsurgeon-tf_6.0.1-1+cuda10.0_arm64.deb
2020-04-11 03:13:51  161324708 nvidia/sdkm_downloads/libcudnn7-dev_7.6.3.28-1+cuda10.0_arm64.deb
2020-04-11 03:23:55    6627420 nvidia/sdkm_downloads/libcudnn7-doc_7.6.3.28-1+cuda10.0_arm64.deb
2020-04-11 03:13:18  181721612 nvidia/sdkm_downloads/libcudnn7_7.6.3.28-1+cuda10.0_arm64.deb
2020-04-11 03:24:13      12596 nvidia/sdkm_downloads/libnvidia-container-tools_0.9.0_beta.1_arm64.deb
2020-04-11 03:24:12      42744 nvidia/sdkm_downloads/libnvidia-container0_0.9.0_beta.1_arm64.deb
2020-04-11 03:24:12      69900 nvidia/sdkm_downloads/libnvinfer-bin_6.0.1-1+cuda10.0_arm64.deb
2020-04-11 03:20:36   57300292 nvidia/sdkm_downloads/libnvinfer-dev_6.0.1-1+cuda10.0_arm64.deb
2020-04-11 03:24:02    4460940 nvidia/sdkm_downloads/libnvinfer-doc_6.0.1-1+cuda10.0_all.deb
2020-04-11 03:24:08    1546362 nvidia/sdkm_downloads/libnvinfer-plugin-dev_6.0.1-1+cuda10.0_arm64.deb
2020-04-11 03:24:06    1489314 nvidia/sdkm_downloads/libnvinfer-plugin6_6.0.1-1+cuda10.0_arm64.deb
2020-04-11 02:51:29  451737338 nvidia/sdkm_downloads/libnvinfer-samples_6.0.1-1+cuda10.0_all.deb
2020-04-11 03:21:10   56443400 nvidia/sdkm_downloads/libnvinfer6_6.0.1-1+cuda10.0_arm64.deb
2020-04-11 03:24:11     164522 nvidia/sdkm_downloads/libnvonnxparsers-dev_6.0.1-1+cuda10.0_arm64.deb
2020-04-11 03:24:09     578042 nvidia/sdkm_downloads/libnvonnxparsers6_6.0.1-1+cuda10.0_arm64.deb
2020-04-11 03:24:10     538398 nvidia/sdkm_downloads/libnvparsers-dev_6.0.1-1+cuda10.0_arm64.deb
2020-04-11 03:24:08     707550 nvidia/sdkm_downloads/libnvparsers6_6.0.1-1+cuda10.0_arm64.deb
2020-04-11 03:19:50   66561472 nvidia/sdkm_downloads/libvisionworks-repo_1.6.0.500n_amd64.deb
2020-04-11 03:20:07   60780704 nvidia/sdkm_downloads/libvisionworks-repo_1.6.0.500n_arm64.deb
2020-04-11 03:22:59   20597586 nvidia/sdkm_downloads/libvisionworks-sfm-repo_0.90.4_amd64.deb
2020-04-11 03:23:08   20216850 nvidia/sdkm_downloads/libvisionworks-sfm-repo_0.90.4_arm64.deb
2020-04-11 03:23:29   13113238 nvidia/sdkm_downloads/libvisionworks-tracking-repo_0.88.2_amd64.deb
2020-04-11 03:23:56   13107966 nvidia/sdkm_downloads/libvisionworks-tracking-repo_0.88.2_arm64.deb
2020-04-11 03:24:13      14842 nvidia/sdkm_downloads/nvidia-container-csv-cuda_10.0.326-1_arm64.deb
2020-04-11 03:24:14       6176 nvidia/sdkm_downloads/nvidia-container-csv-cudnn_7.6.3.28-1+cuda10.0_arm64.deb
2020-04-11 03:24:13      13686 nvidia/sdkm_downloads/nvidia-container-csv-tensorrt_6.0.1.10-1+cuda10.0_arm64.deb
2020-04-11 03:24:12      29796 nvidia/sdkm_downloads/nvidia-container-csv-visionworks_1.6.0.500n_arm64.deb
2020-04-11 03:24:09     534888 nvidia/sdkm_downloads/nvidia-container-runtime_3.1.0-1_arm64.deb
2020-04-11 03:24:10     501768 nvidia/sdkm_downloads/nvidia-container-toolkit_1.0.1-1_arm64.deb
2020-04-11 03:24:15       2940 nvidia/sdkm_downloads/nvidia-docker2_2.2.0-1_all.deb
2020-04-11 03:18:20   70425180 nvidia/sdkm_downloads/nvidia-l4t-jetson-multimedia-api_32.3.1-20191209225816_arm64.deb
2020-04-11 03:24:14       3632 nvidia/sdkm_downloads/python-libnvinfer-dev_6.0.1-1+cuda10.0_arm64.deb
2020-04-11 03:24:10     363394 nvidia/sdkm_downloads/python-libnvinfer_6.0.1-1+cuda10.0_arm64.deb
2020-04-11 03:24:14       3648 nvidia/sdkm_downloads/python3-libnvinfer-dev_6.0.1-1+cuda10.0_arm64.deb
2020-04-11 03:24:10     360882 nvidia/sdkm_downloads/python3-libnvinfer_6.0.1-1+cuda10.0_arm64.deb
2020-04-11 03:24:11     182539 nvidia/sdkm_downloads/sdkml3_jetpack_l4t_43_ga.json
2020-04-11 03:24:14       3652 nvidia/sdkm_downloads/tensorrt_6.0.1.10-1+cuda10.0_arm64.deb
2020-04-11 03:24:12      37802 nvidia/sdkm_downloads/uff-converter-tf_6.0.1-1+cuda10.0_arm64.deb
2020-04-11 03:23:54    7959372 nvidia/sdkm_downloads/vpi-dev-0.1.0-r2-l4t-aarch64.deb
2020-04-11 03:22:11   26794560 nvidia/sdkm_downloads/vpi-lib-0.1.0-r2-l4t-aarch64.deb
2020-04-11 03:23:58    4491602 nvidia/sdkm_downloads/vpi-samples-0.1.0-r2-l4t-aarch64.deb
```

ptrendx:
@leezu Thanks for listing the files - I am now looking into the contents of the right cross compilation package. I think you are right that everything should already be there and it is possibly just a package that was not installed in the Dockerfile itself, so I should be able to do this in the PR :-).

ptrendx:
I looked at the performance impact of RTC and it adds ~2us of CPU time to the launch, mostly due to string manipulation.

ptrendx:

ci/docker/Dockerfile.build.jetson
```diff
@@ -90,6 +90,9 @@ RUN wget https://repo.download.nvidia.com/jetson/x86_64/pool/r32.4/c/cuda/cuda-c
 # while previously it used to store it at /usr/local/cuda/targets/aarch64-linux/lib/stubs
 RUN ln -s /usr/lib/aarch64-linux-gnu/libcublas.so /usr/local/cuda/targets/aarch64-linux/lib/stubs/libcublas.so
 
+# Need the stub for libcuda.so in a place where host linker can find it to link the unit tests
+RUN ln -s /usr/local/cuda/targets/aarch64-linux/lib/stubs/libcuda.so /usr/lib/aarch64-linux-gnu/libcuda.so
```
leezu, line 94: Do you know why `/usr/local/cuda/targets/aarch64-linux/lib/stubs/libcuda.so` is not detected by cmake? It shouldn't be necessary to create this symlink (it isn't necessary in the non-cross-compile setting). Have you reported a bug in the cmake repository?

ptrendx: It is detected by cmake, but only for CUDA compilation - the problem happens during the compilation of mxnet_unit_tests, which is not CUDA-aware I believe.

leezu: `mxnet_unit_tests` is a cmake target that depends on the `libmxnet` target. The libraries detected when compiling `libmxnet` will also be propagated to `mxnet_unit_tests`. I think the problem is that you just added the `cuda` as dependency, instead of the correct `CUDA::cuda_driver` found by https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindCUDAToolkit.cmake.

leezu: I'm referring to this part of the cmakelists:

```
  list(APPEND mxnet_LINKER_LIBS CUDA::cudart CUDA::cublas CUDA::cufft CUDA::cusolver CUDA::curand)
  if(ENABLE_CUDA_RTC)
    if(CUDA_nvrtc_LIBRARY)
      list(APPEND mxnet_LINKER_LIBS CUDA::nvrtc cuda)
      add_definitions(-DMXNET_ENABLE_CUDA_RTC=1)
    else()
      message(FATAL_ERROR "ENABLE_CUDA_RTC=ON, but failed to find NVRTC. CMake will exit." )
    endif()
  endif()
  list(APPEND SOURCE ${CUDA})
  add_definitions(-DMXNET_USE_CUDA=1)
```

ptrendx: Ok, let me try changing it.

ptrendx: Seems to be working :-). Thanks for the suggestion!

src/common/cuda/rtc.cc
```diff
@@ -0,0 +1,246 @@
...
+
+}  // namespace util
+
+namespace {
+
+// Obtain compilation log from the program.
+std::string GetCompileLog(nvrtcProgram program) {
+  size_t log_size_including_null;
+  NVRTC_CALL(nvrtcGetProgramLogSize(program, &log_size_including_null));
+  // For most std::string implementations, this is probably 1 char bigger than needed.  OK though.
```
leezu, line 74: Why is it needed?

"The elements of a basic_string are stored contiguously, that is, for a basic_string s, &*(s.begin() + n) == &*s.begin() + n for any n in [0, s.size()), or, equivalently, a pointer to s[0] can be passed to functions that expect a pointer to the first element of a null-terminated (since C++11)CharT[] array."

https://en.cppreference.com/w/cpp/string/basic_string

src/common/cuda/rtc.cc
```diff
@@ -0,0 +1,246 @@
...
+                                  0,                                         // num headers
+                                  nullptr,                                   // headers
+                                  nullptr));                                 // include names
+
+    std::string gpu_arch_arg = "--gpu-architecture=compute_" + std::to_string(sm_arch);
+    const char *opts[] = {gpu_arch_arg.c_str(),
+#if NDEBUG == 0
+                          "-G",
+#endif
+                          "--std=c++11"};
```
leezu, line 154: `c++14`?

src/operator/numpy/np_elemwise_broadcast_op_extended.cc
```diff
@@ -201,16 +201,12 @@ MXNET_OPERATOR_REGISTER_NP_BINARY_SCALAR(_npi_copysign_scalar)
 
 MXNET_OPERATOR_REGISTER_NP_BINARY_SCALAR(_npi_rcopysign_scalar)
 .set_attr<FCompute>("FCompute<cpu>", BinaryScalarOp::Compute<cpu, mshadow_op::rcopysign>)
-.set_attr<nnvm::FGradient>("FGradient", ElemwiseGradUseIn{"_backward_npi_rcopysign_scalar"});
+.set_attr<nnvm::FGradient>("FGradient", MakeZeroGradNodes);
```
leezu, line 204: Why change the grad nodes?

ptrendx: Because `rcopysign_grad` is 0 :-): https://github.com/apache/incubator-mxnet/blob/master/src/operator/mshadow_op.h#L777

I noticed this when adding the RTC version, so figured instead of adding a custom code for zero it is better to reuse the existing mechanism to handle the case when gradients are 0.

src/common/cuda/rtc.cc
ptrendx: :+1:

src/common/cuda/rtc.cc
DickJC123: I wrote the code awhile back, but I think I see the logic behind it.  Forgive me, but I prefer to think of it by way of a concrete example:

Let's say the compile log had 100 characters. The nvrtcGetProgramLogSize() call would report a size of 101 and the nvrtcGetProgramLog() call would copy 101 characters to the pointer address passed to it.  The 101 characters are the 100 characters of the log plus a null byte.  This approach is targeting a C interface- you're expected to malloc an array of 101 bytes (the length reported to you) and all is well.

Here though, we want the program log ultimately to look like std::string of 100 characters.  However, do we want to allocate the string of size 100, then hope we can copy 101 bytes into it?  I think most implementations of std::string will add 1 to the size you give at string construction time, because they want an efficient implementation of string::c_str().  So it would probably work, but why risk it?  Basically, we allocate the string with length 101, let nvrtcGetProgramLog() copy the 101 bytes into it, then resize the string to length 100 so we don't have the null character messing up any print or other string manipulations we do down the road.

leezu: Thank you for clarifying @DickJC123.

> I think most implementations of std::string will add 1 to the size you give at string construction time, because they want an efficient implementation of string::c_str(). So it would probably work, but why risk it?

As of C++11, the standard guarantees that "a pointer to s[0] can be passed to functions that expect a pointer to the first element of a null-terminated CharT[] array" according to https://en.cppreference.com/w/cpp/string/basic_string. Ie the pointer returned by `s.c_str()` must be the same as as the pointer to `s[0]`. So there would be no risk in allocating the string of true size and avoid the extra resize.

DickJC123: I studied your link more carefully and think my post above missed your point.  Studying it further, I agree now that the following would work under C++11:
```
std::string GetCompileLog(nvrtcProgram program) {
  size_t log_size_including_null;
  NVRTC_CALL(nvrtcGetProgramLogSize(program, &log_size_including_null));
  std::string log(log_size_including_null-1, '\0');
  // Room for terminating null character ensured since C++11
  NVRTC_CALL(nvrtcGetProgramLog(program, &log[0]));
  return log;
}
```
I guess with the version currently in the PR, I was being conservative, not wanting to research the level of C++11 support in all the compilers MXNet was going to see.  Do you see any difference between &log[0] and log.data() in terms of ensuring the compiler adheres to its contract of providing size()+1 bytes of contiguous storage?

DickJC123: Actually, I see that the non-const version of data() is only available beginning with C++17.  What's the minimum C++ version that this code needs to work against?  Assuming it's C++11 and we go with &log[0] above, do you think that the array-size guarantee of size()+1 is there if the base data address is obtained via &log[0]?  While the Microsoft online documentation generally mirrors the cppreference link you posted, I couldn't find the text  "pointer to s[0] can be passed to functions that expect a pointer to the first element of a null-terminated (since C++11) CharT[] array" in anything from Microsoft, e.g. https://docs.microsoft.com/en-us/cpp/standard-library/basic-string-class?view=vs-2019

ptrendx: @DickJC123 If you read the `data` and `operator[]` paragraphs from the Microsoft documentation that you linked, you will see that they do actually say the same thing (just in a more convoluted way):
 - from `data`: `Prior to C++11, data didn't guarantee the returned string was null-terminated. Since C++11, data and c_str both return a null-terminated string, and are effectively the same.`
 - from `operator[]`: ` An invalid index passed to operator[] results in undefined behavior, but the index equal to the length of the string is a valid index for const strings and the operator returns the null character when passed this index.`

DickJC123: I was assuming we wouldn't be using data() or c_str(), since they return a const pointer (data() has a non-const starting with C++17).  I was worried about a clever compiler that ensured the contiguous size()+1 layout and presence of the null byte lazily upon the user calling data() or c_str().  So assuming we access a non-const pointer to the beginning of the string via &log[0], do you feel Microsoft describes that the size()+1 layout exists at that point?

leezu: We require C++17 for all CPU code (which should include the RTC "managamenet" code):

https://github.com/apache/incubator-mxnet/blob/6bb3d724189e1d5727fc7b2b41cab3863294ff99/CMakeLists.txt#L9-L12

though the CUDA code is compiled with C++14 due to lack of C++17 support in Cuda 10

https://github.com/apache/incubator-mxnet/blob/6bb3d724189e1d5727fc7b2b41cab3863294ff99/CMakeLists.txt#L118-L120

DickJC123: You might be surprised to hear that I'm actually ambivalent which way we go here.  The way I see it, the two approaches are:

1. Existing approach- works, more conservative coding style, has unneeded resize() (2 lines)
2. Proposed approach- simpler, small bit faster due to eliminating resize(), closer to 'bleeding edge' of C++11 standard

I propose we go with approach 2, without the resize(), as suggested by @leezu.  However, let's leave in the comment that I included in my code snippet above.  If the RTC ever starts segfaulting, and I'm pulled in to debug it, the comment would be a sufficient reminder to me to revert to the former approach to eliminate compiler standard non-compliance as a possible cause.

ptrendx:
@eric-haibin-lin Yes. A lot of the code (compilation related as well as the function definitions) was moved around to be able to be reused. Support for NumPy-style mixed types were also added.

ptrendx:

Flaky test test_gluon_dirichlet, tracked in #18809

ptrendx:
@eric-haibin-lin Yes. The overhead comes from preparing a string with kernel options (like the datatypes) and searching for the kernel function in cache. CUDA graph caches the resulting function so the lookup does not occur anymore.

That said, this overhead is lower than the overhead of `cudaLaunchKernel` itself and is barely noticeable - I tried it with a worst case scenario of fully hybridized model that was adding tensors with single element (to be 100% CPU limited) and got ~10% slowdown. More realistic workload with kernels taking longer than a few us would not show any difference. The same CPU limited test with non-hybridized model did not show noticeable slowdown (overheads of imperative mode are way higher than this).

ptrendx:
@mxnet-bot run ci [unix-cpu]

docs/python_docs/python/tutorials/extend/index.rst
```diff
@@ -47,6 +47,12 @@ The following tutorials will help you learn how to customize MXNet.
       How to create new MXNet operators in MXNet's backend using C++.
       An example custom quadratic function op.
 
+   .. card::
+      :title: Using runtime compilation (RTC) to write CUDA kernels in MXNet
+      :link: https://mxnet.apache.org/api/faq/using_rtc
```
szha, line 52: use `:link: /api/faq/using_rtc` instead as the documentation is versioned.

src/common/cuda/rtc/util-inl.h
```diff
@@ -0,0 +1,380 @@
...
+namespace rtc {
+
+const char type_support_string[] = R"code(
+using float32 = float;
+using float64 = double;
+using float16 = half;
+using uint8 = unsigned char;
+using int8 = char;
+using int32 = int;
+using int64 = long long;
```
szha, line 39: will these ever be platform dependent?

tests/python/unittest/test_sparse_operator.py
```diff
@@ -1853,152 +1853,6 @@ def check_broadcast_div(mx_lhs, mx_rhs, np_lhs, np_rhs, dtype):
             check_broadcast_mul(mx_lhs, mx_rhs, np_lhs, np_rhs, np.float32)
             check_broadcast_div(mx_lhs, mx_rhs, np_lhs, np_rhs, np.float32)
 
-@with_seed()
-def test_scatter_ops():
```
szha, line 1857: cc @eric-haibin-lin to confirm

tests/python/unittest/test_operator.py
```diff
@@ -9337,4 +9337,82 @@ def test_elemwise_sum_for_gradient_accumulation():
         assert stored_grad['write'] == stored_grad['add']
         assert stored_grad['write'] == 2 * nrepeat
 
+@with_seed()
+def test_elementwise_ops_on_misaligned_input():
```
szha, line 9341: it would be great to have the numpy equivalent of them too, in tests/python/unittest/test_numpy_op.py

tests/python/unittest/test_sparse_operator.py
ptrendx: Those scatter ops were used long ago - introduced in #7903 to support sparse adagrad I think and then replaced in different implementation of adagrad in #9651. Still, the ops stayed, and since they were always in `_internal` space, I removed them instead of reworking them to RTC.

src/common/cuda/rtc/util-inl.h
ptrendx: Not sure what you mean by that - are you asking about `long long` in particular or all of those? The problem with RTC is that it is not really easy to include anything external, especially system headers, so I need to use built-in types - otherwise I would use stuff like `int64_t` here.

szha: I see. How is the type size determined?

ptrendx: They come from the standard builtin types on a given platform. That said, I see your point - I will add some static asserts to ensure that the size is what we expect.

docs/python_docs/python/tutorials/extend/index.rst
ptrendx: Ok - I was just copy-pasting from the other sections there (like add_op_in_backend). Will update those as well.

ptrendx: Hmm, the toctree also has the full links - will `/api/faq/...` approach work there too?

tests/python/unittest/test_operator.py
ptrendx: :+1:

docs/python_docs/python/tutorials/extend/index.rst
ptrendx: Yeah, I tried putting those relative links in the toctree as well and building of Python docs complains with
```
/work/mxnet/docs/python_docs/python/build/tutorials/extend/index.rst:57: WARNING: toctree contains reference to nonexisting document 'api/faq/new_op'                                                              
/work/mxnet/docs/python_docs/python/build/tutorials/extend/index.rst:57: WARNING: toctree contains reference to nonexisting document 'api/faq/add_op_in_backend'                           
/work/mxnet/docs/python_docs/python/build/tutorials/extend/index.rst:57: WARNING: toctree contains reference to nonexisting document 'api/faq/using_rtc'
```
and those entries are not shown in the final website.

szha: I'm currently fixing a number of issues in https://github.com/apache/incubator-mxnet/pull/18839. You may get conflict from this

ptrendx: Nice :-). I will push the changes to address your other comments then and will get back to the website part once your PR is merged.

ptrendx: #18839 is merged, but it did not tackle the issue of toctree - should I leave the links as they are right now @szha?

PR #18630, opened 2020-06-27
Add recursive wildcards to Makefile

Kh4L:
## Description ##

Source files and includes wildcards in the Makefile were hardcoded, limiting the depth of the globbing and making any modifications prone to forget files in make's dependency graph.

This PR adds a recursive wildcard function to be able to glob the directories recursively.


szha:
thanks for the patch @Kh4L 

As discussed in the 2.0 roadmap, we will stop supporting makefile in 2.0. Makefile will continue to be supported in 1.x so this patch is still useful there. Could you switch the base branch from master to v1.x branch, the dev branch for mxnet 1.x?

PR #18640, opened 2020-06-30
fix julia broken links

ys2843:
## Description ##
On master website api page, currently all language binding api reference links are redirected to v1.6 docs. But only the redirect for Julia does not work as expect. The same rewrite rule that works on a preview host behaves differently on MXNet website. https://mxnet.apache.org/api
After investigation, I found only julia's folder structure changed (was removed) while other languages still preserve the folder with API docs removed and tutorials left. And how Apache's server redirect Julia query is `/api/julia/docs/api/` => `/api/julia.html/docs/api/`, which looks like it searches for `julia` folder first and do not find it, then it try to look for `julia.html` which exists, this redirect executes before our rewrite rules takes effect. While a preview host can't reproduce this with same `htaccess` file, so it is possible a higher level rewrite rule Apache Infra team defined in their server.

To fix this problem, we can 1. put an empty Julia folder and test the result 2. Hard coded the link
In this PR, method 2 is implemented. It hardcoded the endpoint of julia API and tutorials on master directly to `/versions/1.6/api/julia/docs/api`, which should be reverted once new Julia docs completes.

The duplicate `.htaccess` files in each version could also make the redirect work not as expect. Still investigating.

### Essentials ###

### Changes ###
- [x] Remove Julia rewrite rule and hard code the end point

## Comments ##
- Preview: http://ec2-34-219-134-42.us-west-2.compute.amazonaws.com/api

### DIFF
--- a/docs/static_site/src/.htaccess
+++ b/docs/static_site/src/.htaccess
@@ -28,8 +28,6 @@ Redirect 302 /api/scala/docs/api/ /versions/1.6/api/scala/docs/api/
 Redirect 302 /api/java/docs/api/ /versions/1.6/api/java/docs/api/
 Redirect 302 /api/clojure/docs/api/ /versions/1.6/api/clojure/docs/api/
 Redirect 302 /api/cpp/docs/api/ /versions/1.6/api/cpp/docs/api/
-Redirect 302 /api/julia/docs/api/ /versions/1.6/api/julia/docs/api/
-Redirect 302 /api/julia/docs/api/#tutorials /versions/1.6/api/julia/docs/api/#tutorials
 
 # Redirect Chinese visitors to Chinese CDN, temporary solution for slow site speed in China
 RewriteCond %{ENV:GEOIP_COUNTRY_CODE} ^CN$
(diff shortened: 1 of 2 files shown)


ys2843:
@mxnet-label-bot add [Website]

leezu:
> After investigation, I found only julia's folder structure changed (was removed) while other languages still preserve the folder with API docs removed and tutorials left.

Should the other ones be removed for consistency? As they were deleted due to use of removed C-APIs, keeping the tutorials may not be necessary for the master branch.

ys2843:
> > After investigation, I found only julia's folder structure changed (was removed) while other languages still preserve the folder with API docs removed and tutorials left.
> 
> Should the other ones be removed for consistency? As they were deleted due to use of removed C-APIs, keeping the tutorials may not be necessary for the master branch.

I see, wasn't aware of this pending change. So on master branch, all languages tutorials and API (except Python) should be removed from website and point to v1.6, is that correct? 
@leezu @szha It would be good to have a sync on what is expected for website, I can help fix the issues and prevent any bugs  that could affect our users.

Issue #18610, opened 2020-06-23
BatchNorm running variables are not properly synchronized on CPU


anko-intel:
## Description
During investigation of #14357 I have found that running_mean and running_var variables are sometimes not yet calculated when converted to NumPy.

### Error Message
When runing the python test from https://github.com/apache/incubator-mxnet/issues/14357#issuecomment-644887485 sometimes there is information that consecutive reads of the same variable differes, which means that backward pass is calculated after the first read:
`var1 = layer2.running_var.data().asnumpy()`
It shouldn't happen as asnumpy() function should force all needed calculations to get running_var values.
I such case the test shows a  message similar to the one below:
```
seed: 331752807
cpu(0) - DIFF in running_var reads:
   0 : [3.9974644 3.9977627 3.997775 ] 
   2 : [3.9962945 4.0025434 3.9987571]
cpu(0) [3.9962945 4.0025434 3.9987571] [10.001289  9.997483 10.002162]
gpu(0) [3.9962947 4.0025434 3.9987576] [10.001292  9.997481 10.002162]
```
Verified on master branch 959143696

## To Reproduce
run python test from https://github.com/apache/incubator-mxnet/issues/14357#issuecomment-644887485

## What have you tried to solve it?

I already have the fix. I will put it to github soon.

TaoLv:
Fixed by #18064

PR #18644, opened 2020-06-30
Fix BatchNorm backward synchronization

anko-intel:
## Description ##
Fix the issue #18610 - synchronization problem with running variables for backward pass of BatchNorm

### Essentials ###
- [x] To the best of my knowledge, examples are either not affected by this change, or have been fixed to be compatible with this change`

### DIFF
--- a/tests/python/unittest/test_gluon.py
+++ b/tests/python/unittest/test_gluon.py
@@ -666,6 +666,32 @@ def test_pool():
 
 
 @with_seed()
+@pytest.mark.parametrize('variable', ['running_var', 'running_mean'])
+def test_batchnorm_backward_synchronization(variable):
+    """
+    Tests if synchronization of BatchNorm running variables is done correctly.
+    If not, the test sometimes fails - depending on the timing.
+    """
+    ctx = mx.test_utils.default_context()
+
+    for _ in range(20):
+        layer = nn.BatchNorm()
+        layer.initialize(ctx=ctx)
+        for _ in range(3):
+            data = mx.nd.random.normal(loc=10, scale=2, shape=(1, 3, 10, 10), ctx=ctx)
+            with mx.autograd.record():
+                out = layer(data)
+            out.backward()
+
+        # check if each read give the same value
+        var1 = getattr(layer, variable).data().asnumpy()
+        for _ in range(10):
+            var2 = getattr(layer, variable).data().asnumpy()
+            if (var1 != var2).any():
+                raise AssertionError("Two consecutive reads of " + variable + " give different results")
+
+
+@with_seed()
 def test_batchnorm():
     layer = nn.BatchNorm(in_channels=10)
     check_layer_forward(layer, (2, 10, 10, 10))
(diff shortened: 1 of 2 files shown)


tests/python/unittest/test_gluon.py
```diff
@@ -665,6 +665,34 @@ def transpose(shape):
...
+        layer = nn.BatchNorm()
+        layer.initialize(ctx=ctx)
+        for _ in range(3):
+            data = mx.nd.random.normal(loc=10, scale=2, shape=(1, 3, 10, 10), ctx=ctx)
+            with mx.autograd.record():
+                out = layer(data)
+            out.backward()
+
+        # check if each read give the same value
+        var1 = eval(read_op)
```
wkcn, line 689: Thank you for the contribution!
For safety, I suggest to use `var1 = getattr(layer, variable).data().asnumpy()`.

tests/python/unittest/test_gluon.py
```diff
@@ -665,6 +665,34 @@ def transpose(shape):
...
+        for _ in range(10):
+            var2 = eval(read_op)
```
wkcn, line 691: `var2 = getattr(layer, variable).data().asnumpy()`.

tests/python/unittest/test_gluon.py
```diff
@@ -665,6 +665,34 @@ def transpose(shape):
...
 
+@with_seed()
+@pytest.mark.parametrize('cudnn_off', [True, False])
+@pytest.mark.parametrize('variable', ['running_var', 'running_mean'])
+def test_batchnorm_backward_synchronization(cudnn_off, variable):
+    """
+    Tests if synchronization of BatchNorm running variables is done correctly.
+    If not, the test sometimes fails - depending on the timing.
+    """
+    ctx = mx.cpu() if cudnn_off else mx.gpu()
```
wkcn, line 676: It should be `ctx=default_context()`.

In CI, there are tests on CPU and GPU. We don't need to specify the context.

anko-intel: Thanks for the tip @wkcn,  
Is there any command line option or environment value to enforce particular context  when I run pytest locally?

tests/python/unittest/test_gluon.py
anko-intel: Thanks!

anko-intel:
@mxnet-bot [codecov/project]

tests/python/unittest/test_gluon.py
wkcn: There seems to no command line to specify the context.
The default context is CPU.
We can create a python file, like https://github.com/apache/incubator-mxnet/blob/master/tests/python/gpu/test_gluon_gpu.py

'''
import mxnet ad mx
from some_test_file import *
set_default_context(mx.gpu(0))
'''
It will switch to gpu context.

TaoLv:
Thanks for the contribution, @anko-intel. Does the problem also exist on v1.x? If yes, could you please cherry pick the fix to the branch also?

Issue #18600, opened 2020-06-21
Multiple numpy tests fail with numpy 1.19


eric-haibin-lin:
The following four numpy op tests fails with numpy==1.19
- test_np_delete
- test_np_random_beta
- test_np_random_f
- test_np_random_chisquare

The failure happens on the master branch CI http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fwindows-cpu/detail/master/2034/pipeline and blocks recent PRs http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/mxnet-validation%2Fwindows-cpu/detail/PR-18598/7/pipeline 

It only happens recently, and I noticed that the installed numpy version on CI has changed from `numpy-1.18.5` to `numpy-1.19.0` - see installation log in [previous commit](http://jenkins.mxnet-ci.amazon-ml.com/blue/rest/organizations/jenkins/pipelines/mxnet-validation/pipelines/windows-cpu/branches/master/runs/2033/nodes/173/steps/230/log/?start=0) and [master head](http://jenkins.mxnet-ci.amazon-ml.com/blue/rest/organizations/jenkins/pipelines/mxnet-validation/pipelines/windows-cpu/branches/master/runs/2034/nodes/173/steps/234/log/?start=0). 
Also notice that the error log shows that the exceptions happen in the _np.xx functions, which is part of the official numpy APIs instead of mx.numpy APIs. 

@yzhliu FYI

eric-haibin-lin:
Error Log 
```
[2020-06-20T23:26:16.963Z] _______________________________ test_np_delete ________________________________
[2020-06-20T23:26:16.963Z] [gw3] win32 -- Python 3.7.3 C:\Python37\python.exe
[2020-06-20T23:26:16.963Z] 
[2020-06-20T23:26:16.963Z]     @with_seed()
[2020-06-20T23:26:16.963Z]     @use_np
[2020-06-20T23:26:16.963Z]     def test_np_delete():
[2020-06-20T23:26:16.963Z]         class TestDelete(HybridBlock):
[2020-06-20T23:26:16.963Z]             def __init__(self, obj, axis=None):
[2020-06-20T23:26:16.963Z]                 super(TestDelete, self).__init__()
[2020-06-20T23:26:16.963Z]                 self._obj = obj
[2020-06-20T23:26:16.963Z]                 self._axis = axis
[...]

PR #18651, opened 2020-07-01
[v1.x]Add constraint to numpy version to <1.19.0 in CI

ciyongch:
## Description ##
As reported in https://github.com/apache/incubator-mxnet/issues/18600, there're some failures related to numpy operators with numpy 1.19.0 (which was released on Jun 21, 2020). Since numpy doesn't follow semantic versioning, and it requires more integration/functionality tests to make sure MXNet is compatible to the latest numpy version. Before all this works are completed, this PR adds constraint to numpy version < 1.19.0 to enable CI pipeline.

@szha @TaoLv @leezu

### DIFF
--- a/ci/docker/install/centos7_python.sh
+++ b/ci/docker/install/centos7_python.sh
@@ -29,4 +29,5 @@ yum -y install python36u
 # Install PIP
 curl "https://bootstrap.pypa.io/get-pip.py" -o "get-pip.py"
 python3.6 get-pip.py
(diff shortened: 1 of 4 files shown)


leezu:
I'm quite concerned that there has been no further investigation of the scope and amount of work needed to provide compatibility with numpy 1.19. Constraining the numpy version can become a major pain later on (when more and more user systems update to numpy 1.19) and can prevent them from using mxnet.

PR #18683, opened 2020-07-09
Expand NVTX usage

ptrendx:
## Description ##
This PR improves the experience when profiling MXNet under nvprof or Nsight Systems, by expanding NVTX functionality.
It was used successfully for some time in the NVIDIA NGC container.

### Essentials ###

### Changes ###
- [x] Added C API for NVTX ranges push/pop
- [x] Added C API to start/stop profiling via nvprof/NSight Systems
- [x] Added Python API corresponding to the C API
- [x] Added automatic ranges to MXNet operators


szha:
@mxnet-bot run ci [centos-cpu, miscellaneous]

szha:
@ptrendx still WIP?

ptrendx:
Yes, I will return back to it sometime next week probably (@DickJC123 is currently merging 1.7 into our codebase and cherry picking stuff after that merge will be easier).

szha:
What's the status for this change?

ptrendx:
It is currently on hold - I got sidetracked to other things and am not sure exactly when I will have time to finish it - I still intend to do that, though.

ptrendx:
@mxnet-bot run ci [unix-cpu]

python/mxnet/cuda_utils.py
```diff
@@ -0,0 +1,37 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""Provides python interface to cuda-related functions of the mxnet library."""
+from .base import _LIB, mx_uint, c_str, check_call
+
+def nvtx_range_push(name, color=13323030):
```
DickJC123, line 21: I don't enjoy suggesting a code repetition, but isn't the following an improvement over letting python users have to make their own colors:
```
class Nvtx:
    # Palette of colors
    RED = 0xFF0000
    GREEN = 0x00FF00
    BLUE = 0x0000FF
    YELLOW = 0xB58900
    ORANGE  = 0xCB4B16
    RED = 0xDC322F
    MAGENTA = 0xD33682
    VIOLET = 0x6C71C4
    BLUE1   = 0x268BD2
    CYAN    = 0x2AA198
    GREEN  = 0x859900

def nvtx_range_push(name, color=Nvtx.ORANGE):
```

src/common/nvtx.h
```diff
@@ -0,0 +1,104 @@
...
+  static const uint32_t kYellow  = 0xB58900;
+  static const uint32_t kOrange  = 0xCB4B16;
+  static const uint32_t kRed1    = 0xDC322F;
+  static const uint32_t kMagenta = 0xD33682;
+  static const uint32_t kViolet  = 0x6C71C4;
+  static const uint32_t kBlue1   = 0x268BD2;
+  static const uint32_t kCyan    = 0x2AA198;
+  static const uint32_t kGreen1  = 0x859900;
+
+  static void gpuRangeStart(const uint32_t rgb, const std::string& range_name) {
```
DickJC123, line 69: Do we know if the nvtxRangePushEx(&att) makes a copy of att.message.ascii?  I suspect not, in which case I'm worried about the lifetime of range_name.c_str().

gpuRangeStart() is called with a const char *, so won't that require making a short-lived temporary string (with a copy of the char array) in the caller's environment?

src/common/nvtx.h
```diff
@@ -0,0 +1,104 @@
+/*
```
DickJC123, line 1: Should this new file be ./src/common/cuda/nvtx.h ?

src/common/nvtx.h
ptrendx: Not sure I your concern - nvtxRangePush is not an async call. It just puts the name and a timestamp into the database and that's it (and yes, it copies the string in the process).

src/common/nvtx.h
ptrendx: Could be, will change that.

src/common/nvtx.h
DickJC123: So if nvtxRangePush copies the att.message.ascii string (as opposed to the pointer), that answers my primary concern here.

I'm left with one nit then, why don't you have `const char * range_name` above, which avoids the needless string copy and keeps the type consistent between the callers (in c_api.cc and threaded_engine_perdevice.cc) and what att.message.ascii wants?

ptrendx: It is a reference to a string, so there is no copying here (`c_str()` does not copy either). I believe in the threaded engine the nvtx_name is a string anyway, so that is also not doing any copies.

src/common/cuda/nvtx.h
```diff
@@ -0,0 +1,104 @@
...
+  static uint32_t nameToColor(const std::string& range_name, int prefix_len) {
+    static std::vector<uint32_t> colors{kRed, kGreen, kBlue, kYellow, kOrange, kRed1, kMagenta,
+                                        kViolet, kBlue1, kCyan, kGreen1};
+    std::string s(range_name, 0, prefix_len);
+    std::hash<std::string> hash_fn;
+    return colors[hash_fn(s) % colors.size()];
+  }
+
+  // Utility to map a range name to a random color based on its hash
+  static uint32_t nameToColor(const char *range_name) {
```
DickJC123, line 90:
```suggestion
  static uint32_t nameToColor(const std::string& range_name) {
```

src/common/nvtx.h
DickJC123: In threaded engine, the following does a copy that could be avoided if everything was handled as a `const char *`:
```
auto nvtx_name = opr_block->opr->opr_name != "" ? opr_block->opr->opr_name : "Op";
```
so perhaps like:
```
auto nvtx_name = opr_block->opr->opr_name.size() ? opr_block->opr->opr_name.c_str() : "Op";
```
It's a nit as I say, so I'm OK leaving it as: `const std::string& range_name`.  Given that though, the two overloaded nameToColor() signatures should be consistent in their use of std::string (see an additional suggestion).

python/mxnet/cuda_utils.py
DickJC123: If you're not fond of the code repetition, at least the following smaller edit explains the magic number:
```
def nvtx_range_push(name, color=0xCB4B16):    # default color is orange
```

src/common/cuda/nvtx.h
ptrendx: :+1:

python/mxnet/cuda_utils.py
ptrendx: I am refactoring this to be nicer and more Pythonic. Will send an updated version tomorrow, once I test it.

ptrendx: The problem with this is mainly that the standard hash in Python is not deterministic, which would make colors change on subsequent runs. Using a real hashing function like the SHA from Python's hashlib on the other hand seems to be quite a heavy hammer (and most probably introduce additional overheads, which we especially do not want to do when profiling the CPU part of the training).

DickJC123: What I tried to suggest was that the 'None' color would be conveyed to the backend (via -1 or alternate API with no color arg), and the backend would assign the color with the existing backend nameToColor() routine.

ptrendx:
@mxnet-bot run ci [unix-cpu, website]

ptrendx:
@mxnet-bot run ci [unix-cpu, website]

ptrendx:
@mxnet-bot run ci [unix-cpu]

DickJC123 (approved):
Thanks for responding to my comments.  LGTM.

